PR #335, opened 2012-01-09
libfreerdp-codec & wfreerdp

louyihua:


libfreerdp-codec/rfx.c
```diff
@@ -266,7 +266,10 @@ static void rfx_process_message_channels(RFX_CONTEXT* context, STREAM* s)
 
 	stream_read_uint8(s, numChannels); /* numChannels (1 byte), must bet set to 0x01 */
 
-	if (numChannels != 1)
+	/* In RDVH sessions, numChannels will represent the number of virtual monitors 
+	 * configured and does not always be set to 0x01 as [MS-RDPRFX] said.
+	 */
+	if (numChannels < 1)
```
nfedera: @louyihua The spec (https://msdn.microsoft.com/en-us/library/ff635150.aspx) says:
> The TS_RFX_CHANNELS message **MUST contain a single channel**

And also the channelId in the data messages always MUST be 0.

This makes me wonder and I'd like to know if you have seen this in real live or just implemented your conclusion from reading the specs here ?

I know this was a long time ago but I'd be glad If you' remember and report back here. Thanks!

PR #1091, opened 2013-03-13
winpr-makecert implementation

awakecoding:
Implemented self-signed TLS certificate generation for RDP usage with winpr-makecert

winpr-makecert -rdp generates a .crt and .key file readily usable by the FreeRDP servers.


adityasood11:
I am only able to create .key. 
.crt file is not created.

Issue #729, opened 2012-09-19
SSL_read: I/O error (Windows 7)


livelace:
freerdp branch:
1.  master
2. 1.0-stable

software:
1. Gentoo x64
2. openssl 0.9.8x/1.0.0j

WORK for Windows 2008 R2 (full patched):

xfreerdp -u user -p password --app --plugin channels/rail/rail.so --data "%SYSTEMDRIVE%\Windows\system32\notepad.exe" -- 192.168.110.2

FAIL for Windows 7 Professional x64 (1.0-stable brach, full patched):

xfreerdp -u user -p password --app --plugin channels/rail/rail.so --data "%SYSTEMDRIVE%\Windows\system32\notepad.exe" -- 192.168.116.38

loading plugin channels/rail/rail.so
connected to 192.168.116.38:3389
pduType bad
SSL_read: I/O error

FAIL for Windows 7 Professional x64 (master brach, full patched):

xfreerdp -u user -p password --app --plugin channels/rail/rail.so --data "%SYSTEMDRIVE%\Windows\system32\notepad.exe" -- 192.168.116.38

loading plugin channels/rail/rail.so
connected to 192.168.116.38:3389
Unable to detect time zone
expected PDU_TYPE_DEMAND_ACTIVE 0001, got 0007
rdp_client_connect_demand_active failed

PS. User logged out.

UP. Additional debug:

expected PDU_TYPE_DEMAND_ACTIVE 0001, got 0007
DBG_RDP rdp_recv_data_pdu (507): recv Save Session Info Data PDU (0x26), length:610
Local < Remote
0000 03 00 00 09 02 f0 80 21 80                      .......!.
rdp_client_connect_demand_active failed

freechelmi:
We have the same problem here : 
connected to 192.168.2.51:3389
pduType bad SSL_read: I/O error

Connection with remmina is OK though.

hardening:
And does it work with master ?

lkraav:
`net-misc/freerdp-1.2.1_pre20150326`

I'm attempting this against Windows 10 TP 10130, so that oculd also be a reason it's not working.

`
$ [-] xfreerdp /app:"||Skype" /v:ten
[00:29:39:895] [13834:13835] [INFO][com.freerdp.client.common.cmdline] - loading channel rail
[00:29:39:896] [13834:13835] [INFO][com.freerdp.client.x11] - No user name set. - Using login name: leho
Password: 
[...]

PR #1167, opened 2013-04-05
Fix for expected PDU_TYPE_DEMAND_ACTIVE 0001, got 0007

dpoe:
This is related to issue #729 and it is probably #729's cause.

The problem here is that during rdp_client_connect_demand_active, an error PDU may be sent from the server.   The code is set up to handle unexpected PDUs, but it skips over a fixed number of bytes (RDP_PACKET_HEADER_MAX_LENGTH) rather than basing it on the content of the headers (TPKT+X224+T125, etc.).  In summary, it skips over 15 bytes when it should be skipping over 14, and then when the demand_active parsing picks up again, it starts in the wrong place and fails.

The fix is to use the same header-parsing code to handle the error packet header, rather than relying on a fixed offset.  That code is smart enough to compute the offsets correctly.

### DIFF
--- a/libfreerdp/core/capabilities.c
+++ b/libfreerdp/core/capabilities.c
@@ -3222,19 +3222,12 @@ BOOL rdp_read_capability_sets(wStream* s, rdpSettings* settings, UINT16 numberCa
 	return TRUE;
 }
 
-BOOL rdp_recv_demand_active(rdpRdp* rdp, wStream* s)
+BOOL rdp_recv_get_active_header(rdpRdp* rdp, wStream* s, UINT16* pChannelId)
 {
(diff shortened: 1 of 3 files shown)


socrateswithstyle:
Fix broken in Debian Jessie:
xfreerdp --version
This is FreeRDP version 1.1.0-beta1 (git n/a)

xfreerdp /u:user /p:pass /app:"||calc" /v:machine /cert-ignore
loading channel rail
connected to machine:3389
expected PDU_TYPE_DEMAND_ACTIVE 0001, got 0007
SSL_read: I/O error: Success (0)

works obviously if you omit /app

Issue #773, opened 2012-10-16
Numlock led not sync


h113331pp:
Good day, every one:
    I recently use freerdp to connect my windows server 2008 sp1 and WES, and found Numlock led won`t sync if I do these steps:
1. check Numlock is on.
2. connect by freerdp.
3. set Numlock off and then select "disconnect"(not "Log off") to quit rdp session.
4. OK, now my Numlock is off now, and then set Numlock on again.
5. reconnect to rdp session we quit before, Numlock status is on but the function is just act like Numlock is off.

could someone please tell me what I do wrong?

PR #1191, opened 2013-04-15
Fix for issue #773 - Numlock not sync'ed

Nigel62:
I found that mstsc seems to respond to a server TsControlPDU, TS_PDUTYPE_DATAPDU Granted Control message by sending an additional sequence of TsInputPDU packets to ensure that the toggled key states are set correctly. Doing the same with xfreerdp seems to fix the issue, obviously something similar needs to be done for the other supported FreeRDP platforms, but I cannot do this


geiregjo:
@Nigel62 Can you explain why tab up need to be send? I'm trying to figure out why MSTSC is doing this... I know it's been a long time since you wrote this but it would really help me out!

PR #1470, opened 2013-09-10
x11/client: win8/server 2k12 fix for kbd sync

### LINKED ISSUE #773: Numlock led not sync

bmiklautz:
Keyboard indicator sync (caps lock/num lock/..) didn't work properly
on reconnect for windows 8 and windows 2012 server.

fixes #773

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -1407,8 +1407,14 @@ void* xf_thread(void* param)
 		rcount = 0;
 		wcount = 0;
 
+		/*
+		 * win8 and server 2k12 seem to have some timing issue/race condition
+		 * when a initial sync request is send to sync the keyboard inidcators
+		 * sending the sync event twice fixed this problem
+		 */
 		if (freerdp_focus_required(instance))
 		{
+			xf_kbd_focus_in(xfc);
 			xf_kbd_focus_in(xfc);
 		}
 


freerdp-github:
I hate to say this, but I am seeing this issue pop up again when connecting to desktops on 2012 r2 RDS servers using This is FreeRDP version 2.0.0-dev (git 57fe581) from Dec 13th.

dbungert:
The root cause of this issue appears to be a race condition on the MS server regarding synchronization of the lock key states.  This is why the problem never seems to go away - we're fixing it in the wrong location.  All we can do is decide which workaround is the least bad.

freerdp-github:
I know it is unrelated code however I think it it funny how this only happens when I use the /gfx +gfx-progressive combination on the command line. BTW, what is the "least bad workaround" and is it something I can try?

bmiklautz:
@dbungert do you have more information or details on the race condition or on the exact problem?

dbungert:
(dump of everything I know on the topic, I'm pretty sure you know most of this, @bmiklautz)

We use the Synchronize Event PDU to establish the basic, starting, state.  If we're connecting and Num Lock is on, then tell the server that.  Later, subsequent touches of the Num lock perform a toggle via the other keyboard events.  This is why if the key starts out incorrect that future key presses will always be inverted.

The first complicating factor is any difference in behavior when connecting the first time to a server versus reconnecting when the server already has a session active for this user.  On first connection, sending the sync state seems to be "enough" to get correct behavior.  For some reason, reconnect is more difficult.  That is challenging because I'm not aware that the RDP client can even see that difference - connect or reconnect usually isn't relevant info and the server doesn't appear to provide us with info to know that it is a reconnect.

The usage above of sending the sync state twice, back to back, is a significant improvement and matches the observed behavior of MSTSC - that is to say, MSTSC attempts to workaround this problem by sending the sync state twice (along with performing other operations).  And while that helps, it's still possible to see the sync failure in MSTSC, which further supports this being a server issue.

OK, so where do we go from here?

1) There is the nuclear option - just send the sync event with every key event.  Yuck.  Let's try to avoid that one, that would be significant keyboard event bloat.
2) An option we are using is to send sync state when the user interacts with the lock keys.  I like this one.  This one allows a user to recover from initial desynchronization in a natural way - if they're expecting all caps and see lowercase, then the user trying to hit the caps lock to repair this is likely.  2 presses later, local keyboard LEDs and server state should be in agreement with user expectations.  Example - keyboard caps lock led is on, user expects caps lock on, server thinks it is off.  First caps lock press turns caps lock off locally and on server, second press turns caps lock on locally and on server.
3) I hypothesize that the sync event works reliably if sent "late" enough in the sequence - let the reconnect do what it needs, then send the sync event after that.  But when is "late" enough?  We're faking this by sending the sync event when the user interacts with the keyboard the first time.  This does appear to help.

I suggest implementing items 2 and 3 (sync on lock key, sync on first keyboard event) while keeping the existing initialization double-sync.  Sync on first keyboard event seems to improve things and sync on lock key allows the user to recover.

bmiklautz:
@dbungert thank you very much that you took time to put this valuable information together. I know now what you meant with "race condition" :)

PR #1680, opened 2014-02-05
30-bit color depth support

fidoman:
New parameter 'depth' for color manipulation functions, in addition to bpp.
Typical combinations bpp/depth:
8/8 (256 color palette)
16/15 (HiColor)
16/16 (so strange RGB565)
32/24 (TrueColor)
32/30 (DeepColor)
These make clrconv->rgb555 redundant.

Tested with /bpp: 15 and 32 with fine results
/bpp:8 and /bpp:16 have some artifacts, but I cannot track their source

Some fields removed to track their usage with compiler.
Code of cursor representation changed as cursor was bad visible in some apps, it is ugly but functional.


bmiklautz:
As it doesn't apply cleanly anymore, still has sigprof commented out, artifacts with 8 and 16 bit and there wasn't any activity for a while I'll close it. If there is still interest on this please create a new PR for it.

Issue #1691, opened 2014-02-12
multimon flag not working-session always spans monitors ignoring command line parameters


grpomega:
PATCH 1

```
--- orig/client/common/cmdline.c        2014-02-06 09:13:40.000000000 -0600
+++ FreeRDP-1.2/client/common/cmdline.c 2014-02-12 14:26:30.786781952 -0600
@@ -45,6 +45,7 @@ COMMAND_LINE_ARGUMENT_A args[] =
        { "h", COMMAND_LINE_VALUE_REQUIRED, "<height>", "768", NULL, -1, NULL, "Height" },
        { "size", COMMAND_LINE_VALUE_REQUIRED, "<width>x<height>", "1024x768", NULL, -1, NULL, "Screen size" },
        { "f", COMMAND_LINE_VALUE_FLAG, NULL, NULL, NULL, -1, NULL, "Fullscreen mode" },
+       { "span", COMMAND_LINE_VALUE_FLAG, NULL, NULL, NULL, -1, NULL, "Span mode" },
        { "bpp", COMMAND_LINE_VALUE_REQUIRED, "<depth>", "16", NULL, -1, NULL, "Session bpp (color depth)" },
        { "kbd", COMMAND_LINE_VALUE_REQUIRED, "0x<layout id> or <layout name>", NULL, NULL, -1, NULL, "Keyboard layout" },
        { "kbd-list", COMMAND_LINE_VALUE_FLAG | COMMAND_LINE_PRINT, NULL, NULL, NULL, -1, NULL, "List keyboard layouts" },
```

PATCH 2

```
--- orig/libfreerdp/core/gcc.c  2014-02-06 09:13:40.000000000 -0600
+++ FreeRDP-1.2/libfreerdp/core/gcc.c   2014-02-12 14:09:38.313791598 -0600
@@ -402,10 +402,9 @@ void gcc_write_client_data_blocks(wStrea
        gcc_write_client_network_data(s, settings);

        /* extended client data supported */
-
        if (settings->NegotiationFlags & EXTENDED_CLIENT_DATA_SUPPORTED)
        {
-               if (settings->SpanMonitors)
+               if (!settings->SpanMonitors)
                {
                        gcc_write_client_monitor_data(s, settings);
                }
```

PR #1695, opened 2014-02-13
fix multimon and span flags-freerdp always spans monitors follow up to #1691

grpomega:
Add span flag so that /span is recognized in cmdline.c
Add negation to settings->SpanMonitors so that the multimon flag takes effect


bmiklautz:
Was fixed on mater with 65ff98bb

PR #1705, opened 2014-02-27
Update xf_keyboard.c

grpomega:
The power button is trapped by the xfreerdp when in fullscreen mode.  I would like xfreerdp to exit immediately and give the calling program some idea of what happened.  Exit code of 124 is given.  This is the x keycode for the power button.  "Power Button" is sent to stderr.  Either one of these can be inspected by a calling program and the information used for proper control.


realjiangms:
@grpomega 
Maybe following way could be accepted
xfc->disconnect = TRUE;

bmiklautz:
Closing this pull request.
* pull request wasn't update for a while

PR #1713, opened 2014-03-04
Change TZ Daylight Savings Time start date for US

grpomega:
Daylight savings time is now encoded to start the first sunday in march.
US Daylight Standard Time begins on the second sunday in march.  
Changed data for US eastern, US central, US Mountain, US Pacific and Alaska Time Zones.


bmiklautz:
As described above: Superseded with #1726

PR #1930, opened 2014-06-26
add .travis.yml / fix -fPIC for clang

abma:
adds a simple travis-ci.org config file & fixes compile with clang


akallabeth:
Out of date and no longer required (clang compiles work fine in current master)

bmiklautz:
When I got a free moment I'll have a look if we can support the current travis.

PR #2135, opened 2014-10-03
Added lock key synchronization to wfreerdp

erbth:
I've added @nigel62's patches from #1191 to wfreerdp. Only thing is I don't know how to read mouse pointer position in windows, so every time lock keys are synced the pointer is set to left top corner. If someone knowes and has time, I would be so happy if she/he could tell me!

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -738,6 +738,12 @@ DWORD WINAPI wf_client_thread(LPVOID lpParam)
 		rcount = 0;
 		wcount = 0;
 
+		if (freerdp_focus_required(instance))
+		{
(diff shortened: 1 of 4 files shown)


client/Windows/wf_client.c
```diff
@@ -738,6 +738,12 @@ DWORD WINAPI wf_client_thread(LPVOID lpParam)
 		rcount = 0;
 		wcount = 0;
 
+		if (freerdp_focus_required(instance))
+		{
+			wf_event_focus_in(wfc);
+			wf_event_focus_in(wfc);
```
garbb, line 744: Why do this twice?

PR #2240, opened 2014-12-01
xfreerdp: pinch/pan fixes and smart-sizing

nfedera:
- use bilinear filter instead of nearest filter
- fixed invalid transformation matrix
- support non-proportional scaling
- fix calculation of transformed coordinates (rounding errors)
- removed X11 drawing operations which directly target the window drawable
- fixed some errors in the calculation of the required update regions
- removed setting ScalingFactor
- added settings SmartSizingWidth and SmartSizingHeight
- changed option /smart-sizing to optionally support &lt;width&gt;x&lt;height&gt;
- consolidated transformation of input event coordinates
- rdp8 gfx ignored scaling and panning offsets: fixed
- never resize window on panning/pinching
- simplified keyboard multitouch gesture debugging emulation
- disabled keyboard multitouch gesture emulation debug code via define

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -31,6 +31,7 @@
 
 #ifdef WITH_XRENDER
 #include <X11/extensions/Xrender.h>
+#include <math.h>
 #endif
 
 #ifdef WITH_XI
@@ -47,11 +48,6 @@
 
 #ifdef WITH_XI
 #include <X11/extensions/XInput2.h>
-#endif
-
-#ifdef WITH_XRENDER
-#include <X11/extensions/Xrender.h>
-#include <math.h>
 #endif
 
 #include <X11/XKBlib.h>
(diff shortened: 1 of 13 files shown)


Hobby-Student:
I hope, I am allowed to ask in this pull request, also it is closed... But I have a question regarding sizing:

Would it be possible to scale if "/f" (fullscreen) is used?
Everytime I use "/f", "/smart-sizing:WIDTHxHEIGHT" has no effect...

I have some customers with thinclients wanting lower resolutions so the font is bigger and easier to read. the sizing could be an easy solution without struggeling resolution of thin clients and used displays. sometimes it's not possible to use resolutions which keep aspect ratio, because xrandr doesn't allow...

I played around with /size nad /smart-sizing. it is possible to "/size:640x480 /smart-sizing:1600x1200" so the font is biger without touching the host. "/size:640x480 /smart-sizing:100%" also does work to "magnify". (640x480 is just an example and not used by anyone ;) )

nfedera:
@Hobby-Student I've added this feature in PR #2547
Please test and let me know if it works for you

mailinglists35:
sorry for waking this old issue, but could you add an option to keep pixel ratio when resizing window?

I find no use to drag the right edge and only resize horizontally, or drag the bottom edge and resize only vertically. it's only annoying.

I expected to resize proportionally when I drag any edge. ironically, the the native client from microsoft exhibits the same behaviour :(

PR #2290, opened 2014-12-30
rdpdr: use host name for redirected drives if set

bmiklautz:
When a client set the host name, redirected drives would still be shown
as coming from the computer name. For example if the computer the client
run on is called "foo" and the client set the /host-name to "bar" any
redirected drive is shown as "driveX on foo".

Now the host name is used for drives when set.


nfedera:
+1 makes sense

PR #2299, opened 2015-01-11
mfreerdp: fix build with latest cmake

bmiklautz:
Newer versions of cmake seem to use CMAKE_C_FLAGS for objective-c.
Now both flags are set for compatibility.


mfleisz:
+1

PR #2300, opened 2015-01-12
license: fix for corrupted licensing packets

nfedera:
Since commit a228952 FreeRDP generates corrupt licensing packets if the rdp
security layer is used and the peer did not indicate that it is capable of
processing encrypted licensing packets:
That commit changed rdp->sec_flags after the rdp stream was already initialized
with encryption enabled which placed the PDU payload at an incorrect offset.

Instead of directly modifying the rdp->sec_flags this patch temporarily
disables rdp->do_crypt during rdp stream initialization if the peer has not
advertised support for encrypted licensing packets.

### DIFF
--- a/libfreerdp/core/license.c
+++ b/libfreerdp/core/license.c
@@ -160,17 +160,28 @@ void license_write_preamble(wStream* s, BYTE bMsgType, BYTE flags, UINT16 wMsgSi
 wStream* license_send_stream_init(rdpLicense* license)
 {
 	wStream* s;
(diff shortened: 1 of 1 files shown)


mfleisz:
+1

bmiklautz:
+1

bmiklautz:
We just tested the cases with freerdp-server and different clients. Without this pull older versions of rdesktop (that used rdp security) showed the following: 

NOT IMPLEMENTED: license tag 0xRANDOM

Leading to undefined behavior and crashes.

PR #2303, opened 2015-01-13
added set_error_info function

MartinHaimberger:
if an error_info is set, a TS_SET_ERROR_INFO_PDU
will be sent to the client on disconnect with
the error_info


bmiklautz:
+1

mfleisz:
+1

PR #2307, opened 2015-01-15
Move main loop and async channel from client to freerdp library

akallabeth:
This pull replaces the ```select``` and ```*_fds``` functions used by clients with functions allowing to
* Register event handles to main loop 
* Remove event handles to main loop
* A function waiting for an event to happen

The current implementation of async channels has also been moved to the library automatically adding support to all clients.

This pull changes a lot of files and all clients, so please check for breakage.
I'd welcome feedback on the current modifications, if something is missing / could be done better let me know.

### DIFF
--- a/client/Android/FreeRDPCore/jni/android_event.c
+++ b/client/Android/FreeRDPCore/jni/android_event.c
@@ -31,48 +31,27 @@
 #include "android_freerdp.h"
 #include "android_cliprdr.h"
 
-int android_is_event_set(ANDROID_EVENT_QUEUE * queue)
+int android_is_event_set(ANDROID_EVENT_QUEUE* queue)
 {
-	fd_set rfds;
-	int num_set;
-	struct timeval time;
-
-	FD_ZERO(&rfds);
-	FD_SET(queue->pipe_fd[0], &rfds);
-	memset(&time, 0, sizeof(time));
-	num_set = select(queue->pipe_fd[0] + 1, &rfds, 0, 0, &time);
-
-	return (num_set == 1);
+	DWORD ev = WaitForSingleObject(queue->event, 0);
+	return (ev == WAIT_OBJECT_0);
 }
 
-void android_set_event(ANDROID_EVENT_QUEUE * queue)
+void android_set_event(ANDROID_EVENT_QUEUE* queue)
 {
-	int length;
-
-	length = write(queue->pipe_fd[1], "sig", 4);
-
-	if (length != 4)
-		WLog_ERR(TAG, "android_set_event: error");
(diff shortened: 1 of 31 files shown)


bmiklautz:
@akallabeth does that require to update the iOS/android client?

akallabeth:
@bmiklautz yes, already updated the clients but needs some testing.

akallabeth:
@freerdp-bot test

awakecoding:
@akallabeth why were so many empty lines removed? it makes the code all packed

awakecoding:
@akallabeth @bmiklautz please do not merge in the current state... I am not really happy with the whole API breakage. This is a major change and many third-party clients will be severely limited and broken by this. I understand the will to centralize and avoid the repetition of the main loop event waiting and checking, but by forcing it to be all handled by the FreeRDP core we have just made the FreeRDP library much less flexible. Is it possible to reintroduce the previous API in addition to the new API?

awakecoding:
Here's the main problem: FreeRDP isn't the only thing we can wait on. Of course, we can obtain file descriptors or HANDLEs from other libraries and feed them to the new FreeRDP API, but what if we have more than one library behaving the same way (forcing the event wait to be done by the library, not the program using the library)? If you have two such libraries that force such waiting to the hidden away, you have a problem, as they cannot be unified into a single wait. The other problem I can see is that by hiding away the event waiting, we lose fine grained control on how the events get waited on. The first thing you lose is the order in which you check for event handles have been set to do something about it: for instance, you probably want to check for any stop event before checking other events like data waiting to be read, input ready to be sent, etc. The second thing you lose is fine grained control over wait timeouts. You can try and add more options in the common function, but it's easily going to become messy. Not every "duplicated" line of code is worth trying to save, I can easily see how this is going to cause a lot of headaches in the future due to the lack of flexibility in the API that will certainly result in multiple API breakages to attempt filling the gaps.

akallabeth:
@awakecoding Yes, I was expecting some issues, so to address the issues you pointed out:
I could easily add a function to the new API allowing to export all handles internal to FreeRDP. This way each client can decide, if the default wait function is sufficient or implement it on its own.
To address the 'order' of internal handles, I could also add a 'identifier' to each handle added, so that a client needing some specific order can rearrange the handles in any way necessary.
This way the common functionality will still be dropped from the clients but the API would still provide the flexibility currently available.

akallabeth:
@awakecoding Modified the API to resemble the old one with ```WaitForMultipleObjects``` instead of select. This way the client is still in control of the handles and can do whatever is desired.

akallabeth:
Implemented with #2348

PR #2313, opened 2015-01-16
Fixed SetLastError test.

akallabeth:
WLog initialisation checks environment variables, which may not exist
and therefore set an error. So initialise it before doing tests.


mfleisz:
@freerdp-bot test

PR #2314, opened 2015-01-16
Fixed a race condition at thread creation

akallabeth:
The thread structure was incorrectly updated after the thread
function was already running. This caused a race condition with
thread exiting. When the thread function returned before the structure
was correctly updated, the exit routine cleaned up the handle as
it was incorrectly marked not started.

### DIFF
--- a/winpr/libwinpr/thread/thread.c
+++ b/winpr/libwinpr/thread/thread.c
@@ -285,12 +285,13 @@ static void winpr_StartThread(WINPR_THREAD *thread)
 	if (thread->dwStackSize > 0)
 		pthread_attr_setstacksize(&attr, (size_t) thread->dwStackSize);
 
-	pthread_create(&thread->thread, &attr, thread_launcher, thread);
-	pthread_attr_destroy(&attr);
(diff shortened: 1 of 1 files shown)


winpr/libwinpr/thread/thread.c
```diff
@@ -285,12 +285,13 @@ static void winpr_StartThread(WINPR_THREAD *thread)
 	if (thread->dwStackSize > 0)
 		pthread_attr_setstacksize(&attr, (size_t) thread->dwStackSize);
 
-	pthread_create(&thread->thread, &attr, thread_launcher, thread);
-	pthread_attr_destroy(&attr);
+	thread->started = TRUE;
 	reset_event(thread);
+
+	pthread_create(&thread->thread, &attr, thread_launcher, thread);
 	ListDictionary_Add(thread_list, &thread->thread, thread);
```
hardening: Stupid question: thread_list is accessed and modified without taking any lock ?

akallabeth: The list is synchronized, so it automatically locks.

PR #2315, opened 2015-01-16
Clang warning fixes

akallabeth:
This pull fixes a lot of warnings from clang.
Please have a look at the changed files, if you can find any errors or mistakes.

Each changed file is it's own commit, so it is easy to skip controversial changes.

### DIFF
--- a/libfreerdp/core/gateway/tsg.c
+++ b/libfreerdp/core/gateway/tsg.c
@@ -200,13 +200,10 @@ BOOL TsProxyCreateTunnelReadResponse(rdpTsg* tsg, RPC_PDU* pdu)
 {
 	BYTE* buffer;
 	UINT32 count;
-	UINT32 length;
 	UINT32 offset;
-	UINT32 Pointer;
 	PTSG_PACKET packet;
 	UINT32 SwitchValue;
 	UINT32 MessageSwitchValue = 0;
-	UINT32 IsMessagePresent;
 	UINT32 MsgBytes;
 	rdpRpc* rpc = tsg->rpc;
 	PTSG_PACKET_CAPABILITIES tsgCaps;
@@ -217,7 +214,6 @@ BOOL TsProxyCreateTunnelReadResponse(rdpTsg* tsg, RPC_PDU* pdu)
 	if (!pdu)
 		return FALSE;
 
-	length = Stream_Length(pdu->s);
 	buffer = Stream_Buffer(pdu->s);
 
 	if (!(pdu->Flags & RPC_PDU_FLAG_STUB))
@@ -234,6 +230,7 @@ BOOL TsProxyCreateTunnelReadResponse(rdpTsg* tsg, RPC_PDU* pdu)
 
 	if ((packet->packetId == TSG_PACKET_TYPE_CAPS_RESPONSE) && (SwitchValue == TSG_PACKET_TYPE_CAPS_RESPONSE))
 	{
+		UINT32 Pointer;
 		packetCapsResponse = (PTSG_PACKET_CAPS_RESPONSE) calloc(1, sizeof(TSG_PACKET_CAPS_RESPONSE));
 
 		if (!packetCapsResponse)
@@ -256,7 +253,6 @@ BOOL TsProxyCreateTunnelReadResponse(rdpTsg* tsg, RPC_PDU* pdu)
 		{
 			offset += 4; /* MsgID */
 			offset += 4; /* MsgType */
-			IsMessagePresent = *((UINT32*) &buffer[offset]);
 			offset += 4;
 			MessageSwitchValue = *((UINT32*) &buffer[offset]);
 			offset += 4;
@@ -264,7 +260,6 @@ BOOL TsProxyCreateTunnelReadResponse(rdpTsg* tsg, RPC_PDU* pdu)
 
 		if (packetCapsResponse->pktQuarEncResponse.certChainLen > 0)
 		{
-			Pointer = *((UINT32*) &buffer[offset]); /* MsgPtr (4 bytes): 0x00020014 */
 			offset += 4;
 			offset += 4; /* MaxCount (4 bytes) */
 			offset += 4; /* Offset (4 bytes) */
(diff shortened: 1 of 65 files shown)


mfleisz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

bmiklautz:
@akallabeth nice! Thanks for your efforts. I think this should https://ci.freerdp.com/job/FreeRDP%20static%20code%20analyzer%20-%20master/ bring this down a lot ;)

channels/rdpei/rdpei_common.c
```diff
@@ -313,7 +313,7 @@ BOOL rdpei_write_4byte_signed(wStream* s, INT32 value)
 		if (negative)
 			byte |= 0x20;
 
-		Stream_Write_UINT8(s, value);
+		Stream_Write_UINT8(s, byte);
```
awakecoding: this looks like a bug more than a warning fix, but good catch

libfreerdp/core/gateway/tsg.c
```diff
@@ -256,7 +253,6 @@ BOOL TsProxyCreateTunnelReadResponse(rdpTsg* tsg, RPC_PDU* pdu)
 		{
 			offset += 4; /* MsgID */
 			offset += 4; /* MsgType */
-			IsMessagePresent = *((UINT32*) &buffer[offset]);
```
awakecoding: I agree that the variable is not used, but removing it makes you forget there was something there in the first place. These kind of variables are just there to correctly contain data parsed from the packet, even if currently unused.

libfreerdp/core/gateway/rts.c
```diff
@@ -887,7 +887,7 @@ int rts_recv_flow_control_ack_pdu(rdpRpc* rpc, BYTE* buffer, UINT32 length)
 	BYTE ChannelCookie[16];
 
 	offset = 24;
-	offset += rts_flow_control_ack_command_read(rpc, &buffer[offset], length - offset,
```
awakecoding: Even if the offset variable is not used beyond this point, updating it to point to the new offset after the last read operation makes it consistent with the rest. I don't think we want to remove such variable updates.

libfreerdp/core/gateway/tsg.c
```diff
@@ -674,7 +661,6 @@ BOOL TsProxyAuthorizeTunnelReadResponse(rdpTsg* tsg, RPC_PDU* pdu)
 	
 	if (SizeValue == 4)
 	{
-		idleTimeout = *((UINT32*) &buffer[offset]);
```
awakecoding: Again, even if we are not using this IdleTimeout value, it is there as a placeholder for now. I actually had to spend some time figuring out where that unparsed IdleTimeout value was being sent by the server, not having a placeholder will make the next developer waste time figuring out where it should be parsed from and how.

awakecoding:
@akallabeth @mfleisz @bmiklautz while many of the fixes are good, there are many types of "fixes" I have to disagree with. It appears to me that all variable assignments that have no side effect for the rest of the execution of a function have been removed. I understand that in some 100% clean, warning-less code every single variable assignment would have an effect on execution, but in practice this prevents any sort of placeholder variables and assignments. The part which is the most troublesome are all the lines that have been removed in the TS Gateway code: parsing NDR is quite difficult, and you definitely want to do it in a consistent, structured way. This means many portions of the NDR notation are parsed but not necessarily used. Aside from the NDR structure, there are portions of the TS Gateway packet contents that may be parsed but unused at this point, and left there as simple placeholders until we get time to actually make use of it. If we are to remove every single variable assignment with no effect on execution, we lose consistency and structure in many parts of the code along with all the placeholders. Parsing a packet does not mean all of it is fully acted on, it just means we can correctly decode it, removing unused variable assignments is equivalent to saying we shouldn't parse anything that we do not currently act on.

libfreerdp/core/gateway/rts.c
hardening: Well, I don't see why we update variables if we don't use them. Would 
/*offset +=*/ rts_flow_control_ack_command_read(rpc, &buffer[offset], length - offset,
be ok for you @awakecoding ? That would keep the code layout and would please the clang checker...

awakecoding: @hardening the thing is that even if we start commenting out everything, it defeats the purpose. This only becomes annoying because of the clang checker warnings, which only catches certain cases. For instance, our Stream macros update pointers of the wStream* struct on every "call" regardless if they are actually required later on. The last call to a read operation that updates the read pointer will in many cases be discarded unless we do a check for the end read position, but it doesn't mean we should systematically get rid of them. One other way to get rid of many of the warnings would be to define data structures for each of the messages and refactor the code to have serialization/deserialization functions separate from the code acting on the packet contents. Functions that read and populate a data structure do not generate warnings, as there is no way to know if the data populated in the data structure will be used or not in the future. Isn't there a way to tweak the clang warnings?

hardening: I think the comment would be a good compromise between readability (which is the purpose of updating a variable that is not used after) and clang warning (which doesn't understand readability concerns ;) )

awakecoding: Commenting out rather than removing is definitely much better for me. However, you said it well: clang warnings do not understand readability concerns. Are we really in favor of making the code less readable for the sake of not having clang warnings?

bmiklautz:
@awakecoding I think most compilers remove the unneeded assignments anyway but warn.
What about just adding a comment with the line instead of having them in or have a FREERDP_UNUSED to tell the compiler/scanner that we want them by marking them explicitly?

bmiklautz:
@awakecoding while you right with your comment on the other hand every extra line makes it harder to read ;) and more complex.

akallabeth:
@awakecoding I understand your concern, so I'd suggest postponing those changes until we can agree on proper handling of those cases. As for the time being, could we mark those files which contain problematic fixes, then we can skip those for now.

akallabeth:
closing, fixes do not merge any more.

PR #2317, opened 2015-01-18
build: improve wayland detection

bmiklautz:
* add support to check wayland client version
* set minimum required version to 0.95

This fixes build problems when wayland libraries where found (automatically detected as recommended feature)
but the installed version was to old to build the wayland client.


hardening:
Looks good +1

PR #2319, opened 2015-01-20
Channel reconnect fix

akallabeth:
This pull request fixed channel initialisation and cleanup.
Now it is possible to reconnect to a server with channels enabled.

It also fixes some resource initialisation and cleanup issues as well as logging formating relates stuff.


akallabeth:
@freerdp-bot test

mfleisz:
@freerdp-bot test

awakecoding:
@akallabeth I am curious to know, did you run into specific issues that you fixed with this pull request? If so, in what scenarios?

akallabeth:
@awakecoding Yes, I've got an issue with a client that needs to run ```freerdp_connect``` and ```freerdp_disconnect``` without discarding the instance. With current master channel handling is broken, so after the first ```freerdp_disconnect``` the channels are not properly initialized any more.

awakecoding:
@akallabeth that's a welcome improvement :) thanks

channels/rdpsnd/server/rdpsnd_main.c
```diff
@@ -591,24 +595,18 @@ void rdpsnd_server_context_reset(RdpsndServerContext *context)
 
 void rdpsnd_server_context_free(RdpsndServerContext* context)
 {
-	if (!context->priv->StopEvent)
```
hardening: Any reason why you removed that safety check: code trying to free a running channel ?
I don't know any code that would sanely do that but the check what there to ensure it !

akallabeth: @hardening looks like I overlooked that one. Should never be NULL, but the remove was accidental.

PR #2325, opened 2015-01-21
Warning fixes v2

akallabeth:
cherry-picked commits from #2315 that should not remove too much.

### DIFF
--- a/channels/rdpdr/server/rdpdr_main.c
+++ b/channels/rdpdr/server/rdpdr_main.c
@@ -35,7 +35,6 @@ static UINT32 g_ClientId = 0;
 static int rdpdr_server_send_announce_request(RdpdrServerContext* context)
 {
 	wStream* s;
-	BOOL status;
 	RDPDR_HEADER header;
 	ULONG written;
 	WLog_DBG(TAG, "RdpdrServerSendAnnounceRequest");
@@ -48,7 +47,7 @@ static int rdpdr_server_send_announce_request(RdpdrServerContext* context)
 	Stream_Write_UINT16(s, context->priv->VersionMinor); /* VersionMinor (2 bytes) */
 	Stream_Write_UINT32(s, context->priv->ClientId); /* ClientId (4 bytes) */
 	Stream_SealLength(s);
-	status = WTSVirtualChannelWrite(context->priv->ChannelHandle, (PCHAR) Stream_Buffer(s), Stream_Length(s), &written);
+	WTSVirtualChannelWrite(context->priv->ChannelHandle, (PCHAR) Stream_Buffer(s), Stream_Length(s), &written);
 	Stream_Free(s, TRUE);
 	return 0;
 }
@@ -258,7 +257,6 @@ static int rdpdr_server_write_smartcard_capability_set(RdpdrServerContext* conte
 static int rdpdr_server_send_core_capability_request(RdpdrServerContext* context)
 {
 	wStream* s;
-	BOOL status;
 	RDPDR_HEADER header;
 	UINT16 numCapabilities;
 	ULONG written;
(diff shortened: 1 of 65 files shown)


bmiklautz:
@freerdp-bot test.

channels/rdpdr/server/rdpdr_main.c
```diff
@@ -48,7 +47,7 @@ static int rdpdr_server_send_announce_request(RdpdrServerContext* context)
 	Stream_Write_UINT16(s, context->priv->VersionMinor); /* VersionMinor (2 bytes) */
 	Stream_Write_UINT32(s, context->priv->ClientId); /* ClientId (4 bytes) */
 	Stream_SealLength(s);
-	status = WTSVirtualChannelWrite(context->priv->ChannelHandle, (PCHAR) Stream_Buffer(s), Stream_Length(s), &written);
+	WTSVirtualChannelWrite(context->priv->ChannelHandle, (PCHAR) Stream_Buffer(s), Stream_Length(s), &written);
```
awakecoding: this is the kind of change I really have a problem with... yes, the status code was obtained but it was unchecked. the fix shouldn't be to remove the status variable, it should be to check the status variable and return -1 on error. I really, really do not want to play cat and mouse in commits where a future commit will just do the exact opposite of such commits + simply check the status variable correctly for error.

winpr/libwinpr/crt/unicode.c
```diff
@@ -169,7 +168,7 @@ int MultiByteToWideChar(UINT CodePage, DWORD dwFlags, LPCSTR lpMultiByteStr,
 		sourceStart = (const BYTE*) lpMultiByteStr;
 		targetStart = (WCHAR*) NULL;
 
-		result = ConvertUTF8toUTF16(&sourceStart, &sourceStart[cbMultiByte],
+		ConvertUTF8toUTF16(&sourceStart, &sourceStart[cbMultiByte],
```
awakecoding: Again, the problem isn't that the result variable exists in the first place, the problem is that it is unchecked! To get rid of the warnings, check variables, don't remove them please.

winpr/libwinpr/path/shell.c
```diff
@@ -335,7 +334,7 @@ char* GetCombinedPath(const char* basePath, const char* subPath)
 	subPathCpy = _strdup(subPath);
 	PathCchConvertStyleA(subPathCpy, subPathLength, PATH_STYLE_NATIVE);
 
-	status = NativePathCchAppendA(path, length + 1, subPathCpy);
+	NativePathCchAppendA(path, length + 1, subPathCpy);
```
awakecoding: Another case of unchecked variable that was removed instead of checked.

winpr/libwinpr/sspi/CredSSP/credssp.c
```diff
@@ -138,9 +137,7 @@ SECURITY_STATUS SEC_ENTRY credssp_QueryCredentialsAttributesA(PCredHandle phCred
 {
 	if (ulAttribute == SECPKG_CRED_ATTR_NAMES)
 	{
-		SSPI_CREDENTIALS* credentials;
-
-		credentials = (SSPI_CREDENTIALS*) sspi_SecureHandleGetLowerPointer(phCredential);
+		sspi_SecureHandleGetLowerPointer(phCredential);
```
awakecoding: Well that's pretty much useless. The whole point of calling that function is to obtain the pointer, now you've removed the variable that was returned and kept the function alone. Agreed, the variable isn't used, worst case it could be commented out for the moment.

winpr/libwinpr/sspi/Schannel/schannel_openssl.c
```diff
@@ -237,7 +235,7 @@ SECURITY_STATUS schannel_openssl_client_process_tokens(SCHANNEL_OPENSSL* context
 			if (!pBuffer)
 				return SEC_E_INVALID_TOKEN;
 
-			status = BIO_write(context->bioRead, pBuffer->pvBuffer, pBuffer->cbBuffer);
+			BIO_write(context->bioRead, pBuffer->pvBuffer, pBuffer->cbBuffer);
```
awakecoding: the SChannel SSPI module is really a WIP... what often happens with WIP code is that a lot of status codes are unchecked. By removing all of them because of that, they will need to be re-added later when the work on the WIP code continues. Obtaining the status code, even if unused, is a reminder that there is something to be checked there, like a todo.

hardening: just an idea: if this code is WIP, perhaps we should remove it from master and you would keep it on a private branch until it is stabilized. That way we keep master with less warnings, and we don't fool people to think schannel_openssl is functional. Just an idea !

akallabeth:
closing, many fixes do not merge any more.

PR #2327, opened 2015-01-21
Fixes cleanup when closing a dynamic channel.

akallabeth:
The resource cleanup was executed twice for each dynamic channel.
Now cleaning up the resources by array list free callback and only
when closing the channel.

@awakecoding Hope the list cleanup in ```dvcman_close_channel_iface``` is the place to do it?

### DIFF
--- a/channels/drdynvc/client/drdynvc_main.c
+++ b/channels/drdynvc/client/drdynvc_main.c
@@ -28,6 +28,8 @@
 
 #define TAG CHANNELS_TAG("drdynvc.client")
 
+static void dvcman_channel_free(DVCMAN_CHANNEL* channel);
+
 static int dvcman_get_configuration(IWTSListener* pListener, void** ppPropertyBag)
 {
 	*ppPropertyBag = NULL;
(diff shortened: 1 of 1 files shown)


awakecoding:
@akallabeth looks good, but make sure you've tested this in both the regular and reconnection scenario of yours. Yesterday I thought you did and channels weren't getting loaded in xfreerdp.

mfleisz:
@freerdp-bot test

akallabeth:
@awakecoding Initial tests look fine, running some more now. Could you just check ```dvcman_close_channel```, I'm not really sure why the close function pointer there is only executed when status == 0

awakecoding:
@akallabeth I think it may have to do with the channel being successfully open. look for where the status code gets set

awakecoding:
@akallabeth actually... make sure that if the channel does not get opened, you're still freeing everything. I think this "close" really means a virtual channel close.

akallabeth:
@awakecoding thx for the hint, moved cleanup to ```dvcman_close_channel```

PR #2330, opened 2015-01-22
Fixed buffer size checks in h264_decompress

akallabeth:
The buffer size checks in ```h264_decompress``` were broken.
The assumption, that the decoded h264 frame must be smaller than the output buffer is wrong.
Instead check, that the desired output rectangle is a valid region in the decoded h264 frame as well as the output buffer.


mfleisz:
+1

erbth:
+1

PR #2335, opened 2015-01-26
drdynvc: fix segfaults and a compiler warning

nfedera:
If we send a negative creation status while processing the DVC create
response pdu in drdynvc_process_create_request() we must also close
the created channel. Otherwise we will get duplicated channel IDs
in the channel manager's array list because the server will
eventually reuse the channel ids which we have reported as failed in
previous creation responses.
In order to protect the client from duplicated IDs sent by buggy or
malicious servers I've added an appropriate check in
dvcman_channel_new() which will now return NULL in this case.
Also added a missing NULL check for the result of dvcman_channel_new()
in dvcman_create_channel().

### DIFF
--- a/channels/drdynvc/client/drdynvc_main.c
+++ b/channels/drdynvc/client/drdynvc_main.c
@@ -28,7 +28,7 @@
 
 #define TAG CHANNELS_TAG("drdynvc.client")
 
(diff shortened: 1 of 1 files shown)


bmiklautz:
@nfedera pull request builder is currently disabled.

PR #2339, opened 2015-01-27
libwinpr: attempt to fix TerminateThread(...).

eledoux:
The following sequence should not hang forever anymore: TerminateThread(thread, 0); WaitForSingleObject(thread, INFINTE);

This should fix the fact that xfreerdp hangs while signing out of a RDP session after to have used a redirected COM port.

Please double check, I'm not familiar enough with the internal events used by a thread. I just noticed the following line in WaitForSingleObject():

		status = waitOnFd(thread->pipe_fd[0], dwMilliseconds);


eledoux:
Closing the pull request to be able to share more changes on the branch

PR #2343, opened 2015-01-29
Fix build failure on x32.

kilobyte:
Unlike i386, x32 can't accept -march=i686 but wants -fPIC, same as amd64
(both are x86_64 ABIs after all).  Thus, check for the __x86_64__ define
instead of pointer width.


kilobyte:
You'll most likely want to verify only that there's no regression on i386 and amd64.  x32 is a new ABI, and not a single distribution has made a release with it.  Debian, Gentoo and Arch have some work; if you are feeling adventurous you can try my installer at http://debian-x32.org/, but it's not generally fit for human consumption yet.

awakecoding:
@freerdp-bot test.

PR #2346, opened 2015-01-29
Win32 Gateway Fix

awakecoding:
TS Gateway support on Windows was completely broken for possibly a long time. This pull request fixes two major problems:

1) Refactor and improve BIO layering, provide proper event handle to wait for socket read notifications. Fix and improve winsock compatibility.
2) Fix NTLM authentication in TS Gateway. Native SSPI modules on Win32 are broken (we might have a small difference in implementation) but the WinPR SSPI modules work. WinPR SSPI modules are now used instead so it can work correctly.


awakecoding:
@freerdp-bot test.

bmiklautz:
@freerdp-bot test

libfreerdp/core/tcp.c
```diff
@@ -222,12 +294,81 @@ static long transport_bio_simple_ctrl(BIO* bio, int cmd, long arg1, void* arg2)
...
+{
+	WINPR_BIO_SIMPLE_SOCKET* ptr = (WINPR_BIO_SIMPLE_SOCKET*) bio->ptr;
+
+	ptr->socket = socket;
+
+	bio->shutdown = shutdown;
+	bio->flags = BIO_FLAGS_SHOULD_RETRY;
+	bio->init = 1;
+
+	if (ptr->win32)
```
hardening: OMG this is a runtime check instead of some ifdefery ?

awakecoding: yeah... it reduced the amount of #ifdefs necessary, and we're crazy enough to have all the functions in WinPR anyway. I don't strongly care about having it runtime, it just made the code slightly cleaner...

awakecoding: I can remove it in a future pull request though

PR #2348, opened 2015-01-30
FreeRDP Wait Events

awakecoding:
More info here: https://github.com/FreeRDP/FreeRDP/wiki/WaitHandles

This pull request adds support for exposing proper wait event handles from the FreeRDP core as a replacement to freerdp_check_fds. This change is an addition and does not break the existing API. The async channels option is also now deprecated, as it is pretty much the equivalent of a "write pending" event. The async channels option was more useful in earlier versions of FreeRDP where the channel system was still using the deprecated freerdp eventing system. This doesn't mean that channels cannot be processed asynchronously, it just means there is no need to do freerdp_channels_get_fds + freerdp_channels_check_fds is using the new freerdp_get_event_handles + freerdp_check_event_handles API.

The major breakthrough allowing such changes to take was the recent fix for winsock compatibility, so we finally have properly usable event handles that notify us of the socket being ready for read. For a long time those events were simply broken on Windows and we had all sorts of timeouts everywhere to periodically check the socket because of that.

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -506,7 +506,7 @@ BOOL wf_authenticate(freerdp* instance, char** username, char** password, char**
 
 	if (status != NO_ERROR)
 	{
-		WLog_ERR(TAG,  "CredUIPromptForCredentials unexpected status: 0x%08X", status);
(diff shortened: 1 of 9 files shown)


awakecoding:
@freerdp-bot test.

PR #2349, opened 2015-01-30
Forward layout ID of Programmer Dvorak to server

RolKau:
The Programmer Dvorak keyboard layout is supported by Xkb but support in Windows is only available through an open-source add-on driver. It is plausible that those that use this layout in X11 also installs this driver on Windows instead of using the standard Dvorak variant there.

This changeset recognizes Programmer Dvorak as its own variant, and assigns this a layout ID which matches the one used in the Windows driver so that it will be selected when you logon. If this layout is not available, it will now revert to the regular United States layout.

Tested with Ubuntu Precise 12.04 connecting to Windows 7 SP1.


hardening:
LGTM +1

PR #2350, opened 2015-02-01
Fix freerdp_channels_free for clients with multiple active connections

giox069:
When a client disconnects from a server and its channel structures are removed, the global hash g_OpenHandles should not be destroyed. Only freed channels must be removed from the hash.


bmiklautz:
@freerdp-bot test.

bmiklautz:
@freerdp-bot test

PR #2352, opened 2015-02-02
Return FALSE instead of exit() in libfreerdp

weberhofer:
Fixes a warning anpotential problems with libfreerdp:
libfreerdp1_2.x86_64: W: shared-lib-calls-exit /usr/lib64/libfreerdp.so.1.2.0 exit@GLIBC_2.2.5
This library package calls exit() or _exit(), probably in a non-fork() context. Doing so from a library is strongly discouraged - when a library function calls exit(), it prevents the calling program from handling the error, reporting it to the user, closing files properly, and cleaning up any state that the program has. It is preferred for the library to return an actual error code and let the calling program decide how to handle the situation.

### DIFF
--- a/libfreerdp/core/peer.c
+++ b/libfreerdp/core/peer.c
@@ -220,7 +220,7 @@ static BOOL freerdp_peer_initialize(freerdp_peer* client)
 		{
 			WLog_ERR(TAG, "Key sizes > 2048 are currently not supported for RDP security.");
 			WLog_ERR(TAG, "Set a different key file than %s", settings->RdpKeyFile);
-			exit(1);
+			return FALSE;
 		}
 	}
 


bmiklautz:
@weberhofer you are absolutely right exit shouldn't be used in a library and we are aware of that.
Unfortunately just returning FALSE didn't catch all cases back then therefore the exit was intentional.
To really get rid of the exit it needs to make sure that all return paths from that functions are handled correct now.

weberhofer:
As returning false seems to be enough to catch the the "inavlid RDP key file" error, I did not check the invoking code; but a short check shows, that you are right and the return value is possibly not checked.

bmiklautz:
@weberhofer I've added issue #2353 maybe hope there is some time to look into this soon.

bmiklautz:
Closing this pull since more work is required see issue #2353

PR #2354, opened 2015-02-02
Handle Server Pointer Update PDU client side

bmiklautz:
* enhance rdp_pointer to support SetPoisition to make it available for clients
* implement SetPoisition in xfreerdp


nfedera:
+1 very nice

PR #2356, opened 2015-02-02
core/capabilities: remove misplaced/useless code

nfedera:
Commit 0357a38e3129b67093c7c055c7eaa242ce293673 has added some code without any effect.
That commit added code to rdp_read_capability_sets() to check if CAPSET_TYPE_MULTI_FRAGMENT_UPDATE was not received which caused settings->MultifragMaxRequestSize to be set to 0.

- this was done in the wrong place because we do these kind of checks in rdp_recv_confirm_active() by consulting the variable settings->ReceivedCapabilities[]
- the code had no effect at all because MultifragMaxRequestSize gets set to FASTPATH_FRAGMENT_SAFE_SIZE in rdp_recv_confirm_active()  if the CAPSET_TYPE_MULTI_FRAGMENT_UPDATE was not received.

### DIFF
--- a/libfreerdp/core/capabilities.c
+++ b/libfreerdp/core/capabilities.c
@@ -3309,8 +3309,6 @@ BOOL rdp_read_capability_sets(wStream* s, rdpSettings* settings, UINT16 numberCa
 	UINT16 length;
 	BYTE *bm, *em;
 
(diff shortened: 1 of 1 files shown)


bmiklautz:
+1

PR #2357, opened 2015-02-02
Update revision number

hardening:
This allows code using FreeRDP to detect the presence of the skipCompression field in surface commands.


bmiklautz:
@freerdp-bot test

bmiklautz:
@hardening if we want todo this we need also to update the packaging.

bmiklautz:
Superseded by #2360

PR #2358, opened 2015-02-03
core/fastpath: removed flawed slow-path conversion

nfedera:
Commit 0357a38e3129b67093c7c055c7eaa242ce293673 modified the function fastpath_send_update_pdu() to check if the desired update is possible by checking the payload size against the computed maxLength and the clients's advertised max request size.
If the check failed that commit added a workaround which simply copied the payload to a slow path updade.
This workaround is totally flawed and causes protocol errors:
- the fast path update code is not checked and required data format conversions are missing
- depending on the fast path update code rdp_send_data_pdu() would have to be called with differend data pdu type values but the workaround always uses DATA_PDU_TYPE_UPDATE
- the workaround does not check if the total size would exceed the maximum possible size for a slow path update

The check if a fast path output is actually possible with the passed parameters is basically a good idea.
However, if that check fails it would only indicate an error in the server implementation who must not generate updates that exceed the client's max request size.
Even though a slow-path conversion would be possible there is much more involved than simply copying the payload stream.
In addition it is highly doubtful if there is a benefit at all. Even the oldest rdesktop and windows ce clients do support fast path and although some lack the multi-fragment update capability we cannot really send larger updates using slow-path outputs.

For the reasons elucidated above, I have removed the workaround but kept a modified version of the check if a fast-path output is possible at all.

### DIFF
--- a/libfreerdp/core/fastpath.c
+++ b/libfreerdp/core/fastpath.c
@@ -896,29 +896,19 @@ BOOL fastpath_send_update_pdu(rdpFastPath* fastpath, BYTE updateCode, wStream* s
 	totalLength = Stream_GetPosition(s);
 	Stream_SetPosition(s, 0);
 
-	/**
-	 * TEMPORARY FIX
-	 *
-	 * FreeRDP always sends updates using fast-path without regard to the client settings
-	 * for 1) FASTPATH_OUTPUT_SUPPORTED in the extraFlags field of the general capability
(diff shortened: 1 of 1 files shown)


MartinHaimberger:
+1 LGTM

mfleisz:
+1

bmiklautz:
+1

PR #2359, opened 2015-02-03
ringbuffer: add option to enable/disable debugging

bmiklautz:
Add option WITH_DEBUG_RINGBUFFER to enable/disable ringbuffer debugging
at compile time.

Even if it is possible to filter specific wlog tags it's not yet
possible to exclude one or more and ringbuffer adds massive debugging
output if enabled and WLOG_LEVEL is set to DEBUG.


nfedera:
+1

hardening:
+1 too

PR #2361, opened 2015-02-03
xfreerdp: fix problem with SetPosition

bmiklautz:
Since XWarpPointer generates an pointer motion event, as if the mouse
was moved by hand, xfreerdp sends back the received mouse position to
the server. This behavior is wrong since pointer positions set via
server pointer update pdu shouldn't get sent back to the server
(it's not a "real" mouse move).

To fix this problem change the x windows event mask to not handle
pointer motion events before setting the new pointer position and change
it back again afterwards.

Possible downside of this is that some pointer motion events might get
lost but this shouldn't be noticeable.

Thanks to @nfedera for the neat idea with the attributes ;).


hardening:
I love that kind of hack :)

nfedera:
+1

PR #2366, opened 2015-02-05
Fixes Solaris bus error on sparc.

xsumbe00:


bmiklautz:
@freerdp-bot test

bmiklautz:
+1

bmiklautz:
@xsumbe00 are you developing directly on solaris/sparc? If so I wonder if you could try building (cmake option -DBUILD_TESTING=ON) and run the tests (ctest) to see if they work.
There is an open issue for debian on sparc https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=764432  that might rather be related to the platform than to the os.
We unfortunately don't have any sparc machine to test on.

xsumbe00:
Is there any docs how should I run the tests?

bmiklautz:
@xsumbe00 not really. But it's quite straight forward:

* install "ctest" (usually comes with the cmake package)
* configure FreeRDP with the extra cmake option -DBUILD_TESTING=ON and build
* just run ctest in the build directory

I guess you already have some configured and compiling tree checked out. In this case just run the following in your build directory.

```
cmake -DBUILD_TESTING=ON .
make
ctest
```

Would be really great to know if you have the same problem on sparc plus the tests can also help to see if "everything" works.

xsumbe00:
So testing wouldn't build at this moment on Solaris. I might look at it sometime later.

awakecoding:
@xsumbe00 you're going to have a lot of fun... there's a bunch of features that I suspect will be broken on big endian. SPARC was the only big endian system I've ever compiled and ran FreeRDP on years ago. There's a bunch of endian unsafe code in the various codecs and ts gateway code, so I'd be curious about if it even can connect correctly.

xsumbe00:
Actually xfreerdp client works on sparc quite well (there are just swapped two color channels ;-) But my goal is to run reliably freerdp-shadow (from vncserver session). Preferably on sparc. And yes I'm not there yet. Even on intel ;-)

Issue #2333, opened 2015-01-26
serial redirect not work


fedsed:
Version freerdp of 26.01.2015.
Run xfreerdp +clipboard /seria:COM2,/dev/ttyUSB0 /v:server
Through the command 'change port' serial port redirect.
Run a program which does the following actions:
<20150126105130.649 SYS>
COM port is opened
<20150126105130.649 SYS>
Set timeouts: ReadInterval=-1, ReadTotalTimeoutMultiplier=0, ReadTotalTimeoutConstant=0, WriteTotalTimeoutMultiplier=5, WriteTotalTimeoutConstant=50
<20150126105130.649 SYS>
Baud rate 38400
<20150126105130.665 SYS>
RTS on
<20150126105130.665 SYS>
DTR on
<20150126105130.665 SYS>
Data bits=8, Stop bits=1, Parity=None
<20150126105130.665 SYS>
Set chars: Eof=0x00, Error=0x00, Break=0x00, Event=0x00, Xon=0x00, Xoff=0x00
<20150126105130.665 SYS>
Handflow: ControlHandShake=(DTR_CONTROL), FlowReplace=(TRANSMIT_TOGGLE, RTS_CONTROL), XonLimit=0, XoffLimit=0
<20150126105130.665 SYS>
Baud rate 9600
<20150126105130.665 SYS>
RTS off
<20150126105130.665 SYS>
DTR on
<20150126105130.680 SYS>
Data bits=8, Stop bits=1, Parity=None
<20150126105130.680 SYS>
Set chars: Eof=0x00, Error=0x00, Break=0x00, Event=0x00, Xon=0x00, Xoff=0x00
<20150126105130.680 SYS>
Handflow: ControlHandShake=(DTR_CONTROL), FlowReplace=(), XonLimit=0, XoffLimit=0
<20150126105130.883 SYS>
Purge the serial port: TXCLEAR
<20150126105130.883 TX>
<SI>
<20150126105131.087 SYS>
Baud rate 115200
<20150126105131.087 SYS>
RTS off
<20150126105131.087 SYS>
DTR on
<20150126105131.087 SYS>
Data bits=8, Stop bits=1, Parity=None
<20150126105131.087 SYS>
Set chars: Eof=0x00, Error=0x00, Break=0x00, Event=0x00, Xon=0x00, Xoff=0x00
<20150126105131.087 SYS>
[...]

PR #2367, opened 2015-02-06
fixed: serial redirect not work #2333 

eledoux:
fixed: serial redirect not work #2333

Two fixes:
1. xfreerdp doesn't hang anymore on session's Sign Out when a COM port is redirected. This was due to TerminateThread() which was never returning (TeminateThread() in only used by serial_main.c as of today).

2. Fixed how to use select() without a timeout.

One change:

1. Made SerialDriverSerialSys the default underlying 'driver' instead of the more restrictive SerialDriverSerCx2Sys.


awakecoding:
@freerdp-bot test.

PR #2371, opened 2015-02-09
winpr/process: handle PIDs <= 0 in TerminateProcess

bmiklautz:
TerminateProcess shouldn't call kill if the PID is <=0 because this has
unwanted effects (and is not what TerminateProcess should do):

* with PID == 0 any process in the same process group gets the signal
  sent
* with PID == -1 *every* processes that the running users has
  permissions to gets the signal sent
* with PID < -1 the process within the same process group and -PID gets
  the signal send

For more details see kill(2).


bmiklautz:
@freerdp-bot test.

bmiklautz:
@freerdp-bot test.

MartinHaimberger:
+1

nfedera:
+1

PR #2374, opened 2015-02-09
Apply correct exit code to xfreerdp application for certain log offs

bjcollins:
When the user logs off of certain OSs(ie. Windows 2008 R2) then the disconnect reason was being incorrectly treated as an unknown exit code. This change allows the correct exit code to be passed from the application in these cases.


bmiklautz:
@freerdp-bot test.

PR #2377, opened 2015-02-10
wtsapi: add TestWtsApiExtra

bmiklautz:
wtsapi: add TestWtsApiExtra

Move wtsapi tests that only make sense when run within a session
("interactive") to TestWtsApiExtra since they either brake automated
builds or don't make sense if the return value isn't checked.

TestWtsApiExtra is not build by default but can be enabled by specifying
-DTESTS_WTSAPI_EXTRA=ON (this option is only available if BUILD_TESTING
is enabled).

The tests set the ctest label WTSAPI_EXTRA and can therefore also
be run explicitly with ctest -L WTSAPI_EXTRA.


hardening:
+1

PR #2386, opened 2015-02-11
Fix remote app in xfreerdp.

bjcollins:
Remove use of the visibleOffset for local window positioning, this completely breaks the display of all windows except for the main application window. Instead, just maintain a local offset correction of the windowOffset. NOTE: I originally made the visibleOffset change in the past on the main window, but things have been broken now that the window routines for the main window and all other windows have been combined. Basically, all pop-up windows, etc. were always displaying the wrong thing.

Also, apply workaround to determining the workArea for remote app mode. Trying to use NET_WM_WORKAREA seems to somehow prevent the window from being able to get sized correctly. This was very evident if you maximized a window as the bottom of the window would never be updated.

For reference, I did all of my testing on a stock CentOS 6.6 system.

### DIFF
--- a/client/X11/xf_monitor.c
+++ b/client/X11/xf_monitor.c
@@ -163,7 +163,12 @@ BOOL xf_detect_monitors(xfContext* xfc)
 	}
 #endif
 
-	if (!xf_GetWorkArea(xfc))
+	/* WORKAROUND: With Remote Application Mode - using NET_WM_WORKAREA
+ 	 * causes issues with the ability to fully size the window vertically
+ 	 * (the bottom of the window area is never updated). So, we just set
+ 	 * the workArea to match the full Screen width/height.
+	 */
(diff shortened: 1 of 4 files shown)


ablepharus:
Hi,

this seems to fix some issues that seem to be introduced with
https://github.com/FreeRDP/FreeRDP/commit/6857a4f2b745bc56a97731e06cd291287371ad92
and
https://github.com/FreeRDP/FreeRDP/commit/2efa67b4f97218c8215e13da6b3d597e473fab9c

but an (unreported, doing that later) issue remains:

it only works on the upmost,leftmost region of the screen on about
3000x1000px space, outside of that, nothing will be painted. If the
whole window is moved outside that region, it will reapair inoberable in
the leftmost,upmost corner.

Just wanted to mention,

best regards

bmiklautz:
@freerdp-bot test

PR #2392, opened 2015-02-12
winpr: stream ctest and missing read/peek types

nfedera:
- reworked stream.h defines to avoid multiple copies of the same code
- added missing types for Stream_Peek Stream_Read
- added an initial CTest for Stream


bmiklautz:
Really nice! Especially that you added tests \o/
+1

hardening:
+1

PR #2398, opened 2015-02-13
Fix the command line argument array setup for decorations flag

bjcollins:
The initial value was set for the wrong field in the structure. The decorations is correctly initialized to on

anyways, but this corrects the formatting for the array element.


bmiklautz:
@freerdp-bot test.

PR #2403, opened 2015-02-18
Transport, Gateway, Sockets

awakecoding:
Properly abstract TS Gateway layer from rdpTransport
Move socket/transport operations under BIO layer
Implement proper BIO chaining for TS Gateway
Prepare TS Gateway for channel recycling support
Add TS Gateway silent channel reauth on session timeout
Add WSAIoctl() network interface listing
Make client-side NLA event-driven


libfreerdp/core/transport.c
```diff
@@ -898,6 +728,58 @@ void transport_set_nla_mode(rdpTransport* transport, BOOL NlaMode)
...
+	{
+		if (transport->tls)
+		{
+			tls_free(transport->tls);
+			transport->tls = NULL;
+		}
+	}
+
+	transport->frontBio = NULL;
+
```
nfedera: Shouldn't there be a BIO_free(transport->frontBio); ?

PR #2416, opened 2015-02-21
libfreerdp-core: fix dead-end state in transport_read_pdu()

zavadovsky:
Situation: we have fragmented TPKT PDU without two last bytes
(or one last byte - for fast-path) in network stack.

First call to transport_read_pdu() works normally, read
available bytes and exit with status 0 - no whole PDU readed.

Before second call this missed bytes arrive.
Optionally with next PDU.

In second call header parsing code unconditionally read this
two bytes(one byte) despite this is not header bytes.
And increase stream position, so stream now contains whole PDU.

This cause (pduLength - Stream_GetPosition(s)) calculation to be 0.
So transport_read_layer_bytes()-->transport_read_layer() return 0
and transport_read_pdu() exits with "not enough data is available"
status.

If next PDU isn't available next calls to transport_read_pdu()
give same result.
If next PDU arrive - (pduLength - Stream_GetPosition(s)) will be
less than 0. Stream position will grow, grow and grow on each call.
And transport_read_pdu() never signals that PDU is readed.

Caught on Android FreeRDP client with high RDP traffic (several MBytes/s).

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -422,6 +422,7 @@ int transport_read_pdu(rdpTransport* transport, wStream* s)
 		return status;
 	}
 
+	position = Stream_GetPosition(s);
 	header = Stream_Buffer(s);
 
 	if (transport->NlaMode)
@@ -439,7 +440,8 @@ int transport_read_pdu(rdpTransport* transport, wStream* s)
 			{
 				if ((header[1] & ~(0x80)) == 1)
 				{
-					if ((status = transport_read_layer_bytes(transport, s, 1)) != 1)
+					if (position < (2 + 1)
+							&& (status = transport_read_layer_bytes(transport, s, 2 + 1 - position)) != 1)
 						return status;
 
 					pduLength = header[2];
@@ -447,7 +449,8 @@ int transport_read_pdu(rdpTransport* transport, wStream* s)
 				}
 				else if ((header[1] & ~(0x80)) == 2)
 				{
-					if ((status = transport_read_layer_bytes(transport, s, 2)) != 1)
+					if (position < (2 + 2)
+							&& (status = transport_read_layer_bytes(transport, s, 2 + 2 - position)) != 1)
 						return status;
 
 					pduLength = (header[2] << 8) | header[3];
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

bmiklautz:
looks like the bot has troubles fetching the stuff from github.

bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

bmiklautz:
@zavadovsky how do you create the rdp traffic? - Would like to see if this is reproducible with xfreerdp as well.

bmiklautz:
@zavadovsky you are right this is an issue. We couldn't reproduce it with any other client (even if we tried hard :) that maybe the reason why it wasn't seen before.

zavadovsky:
@bmiklautz Xrdp+libfreerdp as RFX encoder+watch Youtube fullscreen in terminal session.
This bug is hard to reproduce - client may work 10 and more minutes without dead-end. Or get stuck in first 30 seconds.

But the code of transport_read_pdu() is potentially buggy - there are must be check that we really need to read some bytes from network layer.

bmiklautz:
@zavadovsky great finding!
+1 for the functionality but I would suggest we use the calculated values (instead of 2+2) for example and add some comment.

zavadovsky:
@bmiklautz added commit with comments and calculated values

nfedera:
@zavadovsky nice finding !!
A great way to test is modifying transport_read_layer_bytes() to always read short like this:
```diff
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -374,7 +374,12 @@ int transport_read_layer(rdpTransport* transport, BYTE* data, int bytes)
 static int transport_read_layer_bytes(rdpTransport* transport, wStream* s, unsigned int toRead)
 {
        int status;
-       status = transport_read_layer(transport, Stream_Pointer(s), toRead);
+       int diff = 0;
+
+       if (!transport->blocking && toRead > 1) {
+               diff = toRead - rand()%toRead;
+       }
+       status = transport_read_layer(transport, Stream_Pointer(s), toRead-diff);
```

I tested your with this "simulation" and it failed horrible without your patch and perfectly fine afterwards.

PR #1392, opened 2013-08-04
fix multimonitor in fullscreen

dontsueme:

PR #2426, opened 2015-02-24
xfreerdp fullscreen enhancements

bjcollins:
Remove override redirect flag for fullscreen with keyboard grab enabled. Window manager should always know about the main window.

Small cleanup of passing around decorations flag.
Limit PercentScreen to single monitor vs. entire desktop. IMO - this is better behavior in a multimonitor environment.

Handle fullscreen windows better:
1. Ensure that size hints are set to allow resizing before setting a window to fullscreen as some window managers do not behave properly.
2. Handle fullscreen toggles without destroying and recreating window.
3. Use NET_WM_STATE_FULLSCREEN Extended Window Manager Hint for fullscreen functionality
4. Use the NET_WM_FULLSCREEN_MONITORS Extended Window Manager Hint when appropriate
5. When a single monitor fullscreen is requested - use the current monitor(as determined from mouse location)
6. Handle cases where there is no local monitor at coordinate 0,0. The Windows server expect there to be a monitor at this location, so we maintain offset if necessary between our local primary monitor and the server side primary monitor located at 0,0.

I did take some of this from another pull request from a long time ago, but added things like handling when no local monitors have a 0,0 coordinate. Please provide comments/suggestions/thoughts.

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -516,7 +516,7 @@ BOOL xf_create_window(xfContext* xfc)
 		xfc->attribs.background_pixel = BlackPixelOfScreen(xfc->screen);
 		xfc->attribs.border_pixel = WhitePixelOfScreen(xfc->screen);
 		xfc->attribs.backing_store = xfc->primary ? NotUseful : Always;
-		xfc->attribs.override_redirect = xfc->grab_keyboard ? xfc->fullscreen : False;
+		xfc->attribs.override_redirect = False;
 		xfc->attribs.colormap = xfc->colormap;
 		xfc->attribs.bit_gravity = NorthWestGravity;
 		xfc->attribs.win_gravity = NorthWestGravity;
@@ -571,7 +571,7 @@ BOOL xf_create_window(xfContext* xfc)
 		}
 #endif
 
-		xfc->window = xf_CreateDesktopWindow(xfc, windowTitle, width, height, xfc->decorations);
+		xfc->window = xf_CreateDesktopWindow(xfc, windowTitle, width, height);
 
 		free(windowTitle);
 
@@ -579,6 +579,10 @@ BOOL xf_create_window(xfContext* xfc)
 			xf_SetWindowFullscreen(xfc, xfc->window, xfc->fullscreen);
 
 		xfc->unobscured = (xevent.xvisibility.state == VisibilityUnobscured);
+
+		/* Disallow resize now that any initial fullscreen window operation is complete */
+		xf_SetWindowSizeHints(xfc, xfc->window, FALSE, xfc->width, xfc->height);
+
 		XSetWMProtocols(xfc->display, xfc->window->handle, &(xfc->WM_DELETE_WINDOW), 1);
 		xfc->drawable = xfc->window->handle;
 	}
@@ -686,16 +690,12 @@ void xf_toggle_fullscreen(xfContext* xfc)
 	rdpContext* context = (rdpContext*) xfc;
 	rdpSettings* settings = context->settings;
 
-	xf_lock_x11(xfc, TRUE);
-
-	xf_window_free(xfc);
-
 	xfc->fullscreen = (xfc->fullscreen) ? FALSE : TRUE;
 	xfc->decorations = (xfc->fullscreen) ? FALSE : settings->Decorations;
 
-	xf_create_window(xfc);
-
-	xf_unlock_x11(xfc, TRUE);
+	xf_SetWindowSizeHints(xfc, xfc->window, TRUE, xfc->width, xfc->height);
+	xf_SetWindowFullscreen(xfc, xfc->window, xfc->fullscreen);
+	xf_SetWindowSizeHints(xfc, xfc->window, FALSE, xfc->width, xfc->height);
 
 	EventArgsInit(&e, "xfreerdp");
 	e.state = xfc->fullscreen ? FREERDP_WINDOW_STATE_FULLSCREEN : 0;
@@ -1697,6 +1697,7 @@ static int xfreerdp_client_new(freerdp* instance, rdpContext* context)
 	xfc->_NET_WORKAREA = XInternAtom(xfc->display, "_NET_WORKAREA", False);
 	xfc->_NET_WM_STATE = XInternAtom(xfc->display, "_NET_WM_STATE", False);
 	xfc->_NET_WM_STATE_FULLSCREEN = XInternAtom(xfc->display, "_NET_WM_STATE_FULLSCREEN", False);
+	xfc->_NET_WM_FULLSCREEN_MONITORS = XInternAtom(xfc->display, "_NET_WM_FULLSCREEN_MONITORS", False);
 	xfc->_NET_WM_WINDOW_TYPE = XInternAtom(xfc->display, "_NET_WM_WINDOW_TYPE", False);
 	xfc->_NET_WM_WINDOW_TYPE_NORMAL = XInternAtom(xfc->display, "_NET_WM_WINDOW_TYPE_NORMAL", False);
 	xfc->_NET_WM_WINDOW_TYPE_DIALOG = XInternAtom(xfc->display, "_NET_WM_WINDOW_TYPE_DIALOG", False);
--- a/client/X11/xf_monitor.c
+++ b/client/X11/xf_monitor.c
@@ -118,12 +118,16 @@ BOOL xf_is_monitor_id_active(xfContext* xfc, UINT32 id)
 BOOL xf_detect_monitors(xfContext* xfc)
 {
 	int i;
-	int nmonitors;
-	int primaryMonitor;
-	int vWidth, vHeight;
+	int nmonitors = 0;
+	int primaryMonitorFound = FALSE;
+	int vX, vY, vWidth, vHeight;
 	int maxWidth, maxHeight;
 	VIRTUAL_SCREEN* vscreen;
 	rdpSettings* settings = xfc->settings;
+
+	int mouse_x, mouse_y, _dummy_i;
+	Window _dummy_w;
+	int current_monitor = 0;
 
 #ifdef WITH_XINERAMA
 	int major, minor;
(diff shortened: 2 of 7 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@bjcollins what window managers did you test this with?

bjcollins:
I tested against metacity on CentOS 6.6. This follows along similar thinking to Change Request 1392 and uses some of its code, but since that CR was never acted on and I am actively looking at xfreerdp again - I decided to push my own change request. I believe in using the EWMH functionality for fullscreen and multimonitor support vs. override_redirects.

Hobby-Student:
Can't test the patch atm, but would really be happy to see it merged (using #1392 for some clients)

nfedera:
@bjcollins Have you tested with /smart-sizing?

client/X11/xf_window.c
```diff
@@ -139,19 +139,57 @@ void xf_SendClientEvent(xfContext* xfc, Window window, Atom atom, unsigned int n
...
+	{
+		xf_SendClientEvent(xfc, window->handle, xfc->_NET_WM_FULLSCREEN_MONITORS, 5,
+				xfc->fullscreenMonitors.top,
+				xfc->fullscreenMonitors.bottom,
+				xfc->fullscreenMonitors.left,
+				xfc->fullscreenMonitors.right,
+				1);
 	}
+
+	window->fullscreen = TRUE;
```
nfedera: Shouldn't this be window->fullscreen = fullscreen ?

client/X11/xf_window.c
```diff
@@ -425,6 +475,17 @@ void xf_ResizeDesktopWindow(xfContext* xfc, xfWindow* window, int width, int hei
...
+		/* Allows the window to resize larger by 1 pixel - so we can
+		 * fullscreen the window with no worries about window manager disallowing based
+		 * on size parameters
+		 */
+		if (can_resize)
+		{
+			size_hints->width_inc = size_hints->height_inc = 1;
+			size_hints->max_width = xfc->width + 1;
+			size_hints->max_height = xfc->height + 1;
+		}
```
nfedera: You're not checking for SmartSizing

client/X11/xf_window.c
```diff
@@ -425,6 +475,17 @@ void xf_ResizeDesktopWindow(xfContext* xfc, xfWindow* window, int width, int hei
...
 
+		/* Allows the window to resize larger by 1 pixel - so we can
+		 * fullscreen the window with no worries about window manager disallowing based
+		 * on size parameters
+		 */
+		if (can_resize)
+		{
+			size_hints->width_inc = size_hints->height_inc = 1;
+			size_hints->max_width = xfc->width + 1;
+			size_hints->max_height = xfc->height + 1;
```
nfedera: Can you give an example for a wm where this was required ? (just curious)

client/X11/xf_client.c
```diff
@@ -686,16 +690,12 @@ void xf_toggle_fullscreen(xfContext* xfc)
...
-
 	xfc->fullscreen = (xfc->fullscreen) ? FALSE : TRUE;
 	xfc->decorations = (xfc->fullscreen) ? FALSE : settings->Decorations;
 
-	xf_create_window(xfc);
-
-	xf_unlock_x11(xfc, TRUE);
+	xf_SetWindowSizeHints(xfc, xfc->window, TRUE, xfc->width, xfc->height);
+	xf_SetWindowFullscreen(xfc, xfc->window, xfc->fullscreen);
+	xf_SetWindowSizeHints(xfc, xfc->window, FALSE, xfc->width, xfc->height);
```
nfedera: SmartSizing is ignored again

client/X11/xf_monitor.c
```diff
@@ -206,30 +233,24 @@ BOOL xf_detect_monitors(xfContext* xfc)
...
-	}
 
-	nmonitors = 0;
-	primaryMonitor = 0;
+		/* Always sets number of monitors from command-line to just 1.
+		 * If the monitor is invalid then we will default back to current monitor
+		 * later as a fallback. So, there is no need to validate command-line entry here.
+		 */
+		settings->NumMonitorIds = 1;
+	}
```
nfedera: good!

bjcollins:
@nfedera I have not tested with smart-sizing(that is kind of new functionality to me and I missed it and how my changes affected it). Now that this patch has been merged, I will plan to work on a new patch to take into account smart-sizing.

client/X11/xf_window.c
bjcollins: Metacity was giving me problems. Tested on CentOS 6.6

client/X11/xf_window.c
bjcollins: Yes. It looks like it should. I do not think that variable is actually used by anything, which is probably why it got overlooked and did not cause me any functional problems.

nfedera:
@bjcollins very cool! However,I think smart-sizing was already partly broken before your commit. AFAIK at commit 415a0a158ec4aab1fa8206bd33fab6c792aa1a67 it was still in perfect shape

nfedera:
@bjcollins Important for smart sizing is that it works with full screen toggle. That is if you start your session with /size:1024x768, then resize the window  to your likings, press ctrl+alt+enter to go into full screen the session must be scaled accordingly. If you press ctrl+alt+enter  again to get out of fullscreen the window should have the same smart-sized dimensions it had before toggeling.
Therefore in the code one should be aware that xfc->width is the rdp session width and not necessarily the with of the window ;)

nfedera:
@bjcollins We do also have code for pinch/zoom and panning. Therefore, xfc->window->width is also not necessarily the width to be scaled to. We might be scaling, but the window has a fixed size (think of your smart phone's display as the window and the rdp session is the picture you're zooming). This is the reason why we need the separate xfc->scaledWidth and xfc->scaledHeight members.

bjcollins:
@nfedera thanks for the information. That all makes sense to me. I doubt I will get to it this week, but I will try to look into smart sizing soon and at the least try to submit a patch to correct my changes to also take into account smart sizing.

client/X11/xf_window.c
bjcollins: @nfedera I realized today that I might have originally hit this with xfwm4 actually. So, there is a possibility metacity does not require it. I know that one of these two WMs have a problem if not both. I used to run xfwm4 on centos5, but now run metacity on centos6.

nfedera:
@bjcollins I've created PR #2530 which reverts your PR and fixes the issues which were already partly broken before your commit.
Once it gets merged it would be cool if you would redo your PR based on the new master coded considering smart-sizing. Also see https://github.com/FreeRDP/FreeRDP/wiki/Quality-Assurance-(QA)#smart-sizing

nfedera:
@bjcollins Have you tested full screen toggling (as in ctrl+alt+enter) with your patch ?

client/X11/xf_monitor.c
```diff
@@ -239,43 +260,114 @@ BOOL xf_detect_monitors(xfContext* xfc)
...
+		settings->MonitorDefArray[nmonitors].orig_screen = i;
 
-		primaryMonitor |= vscreen->monitors[i].primary;
 		nmonitors++;
 	}
 
-	settings->MonitorCount = nmonitors;
-
-	vWidth = vHeight = 0;
-	settings->DesktopPosX = maxWidth - 1;
```
nfedera: If that the only use variable it should be removed

client/X11/xf_monitor.c
```diff
@@ -239,43 +260,114 @@ BOOL xf_detect_monitors(xfContext* xfc)
...
-	settings->DesktopPosY = maxHeight - 1;
```
nfedera: If that was the only real use for the variable maxHeight  it should have been be removed

client/X11/xf_monitor.c
bjcollins: Agreed. That variable is no longer used and should be completely removed to simplify the code.

client/X11/xf_monitor.c
bjcollins: Agreed. That variable is no longer used and should be completely removed to simplify the code.

bjcollins:
@nfedera Yes. I tested with fullscreen toggle to be sure that I could remove the bit of code to destroy the window and create a new one every time there is a fullscreen toggle sequence. Have you seen issues with this?

nfedera:
@bjcollins there are of course issues with smart-sizing which was initialized in xf_create_window . Probably also if the setup changes because xf_create_window now is not called anymore which did also call xf_detect_monitors before. I'm currently looking into this but the complete x11 code (not talking about your patch) is really messy and needs a proper redesign IMHO.

bjcollins:
@nfedera I agree that the X11 code is generally messy. It may make sense to put the destroy/create window bit back if the smart-sizing stuff is too intertwined into those other functions right now. I do not believe that change was a requirement for the user facing functionality in my patch, it just seemed like unnecessary steps to completely destroy and recreate the window so that is why I removed it.

nfedera:
@bjcollins Is it just me or has your pr also has broken /span and /multimon ?

bjcollins:
@nfedera I just looked at the source real quickly for icewm and it does not seem to support the EWMH fullscreen bits, so that is probably the problem there. I would not expect it to just totally not work, were you only seeing an issue in certain scenarios or could you never get it to display on more than one monitor at all?

nfedera:
@bjcollins it seems that also simple /f is now broken too.
If my mouse is on the right monitor xfreerdp is displayed on the left monitor but with the dimensions of the right monitor and vice versa.
I guess this happens because my if screen 0 is right of screen 1.
Can you confirm this issue ?

bjcollins:
@nfedera Testing this functionality was a big part of my PR, so I do not think it is completely broken. However, I do not doubt that that you are having a problem. In an ideal situation, you would be using Xinerama and the WM would support the EWMH(mainly NET_WM_STATE_FULLSCREEN and NET_WM_FULLSCREEN_MONITORS) and I mainly tested with that. So, to help me understand the code path that is being taken can you confirm:

a) whether or not you have compiled xfreerdp with xinerama support
b) the wm you are using and whether it supports the EWMH

I will look at this later today, but wanted to see if I can learn a little more about your environment.

nfedera:
@bjcollins I'm running ubuntu 14.04 on an iMac. The native display is 2560x1440. I've connected an Asus 1920x1080 monitor via ThunderboldToDVI and positioned it left of the iMac.
For debugging I change to the console (Ctrl+Alt+F1) log in as a diferent user and start X11 like this:
 ``` startx /usr/bin/xterm -- /usr/bin/X11 :2```
Now I dow the screen setup via xrandr: ```xrandr --output eDP --auto --rightof DisplayPort-3```
(eDP is the native display and DisplayPort-3 is the ThunderBold)
At this point I can easily start different window managers and test.
The problem described happens with xfce4 and it was working before this PR.

However, don't bother looking into it.
I'm almost done with the new PR and we can test and improve starting from there ....

Thanks

nfedera:
@bjcollins Here it is https://github.com/FreeRDP/FreeRDP/pull/2547 
Please test and check if it breaks any of the features you've added.

Issue #2152, opened 2014-10-16
Wrong graphics, color shifted. Xcode 5.1.1, Mac OS X 10.9.3, cmake 2.8.12.2


palmoff:
Im getting strange picture on iOS 7.1, iOS 8.0.2, on different iOS devices (iPads/iPhones):
![img_1957](https://cloud.githubusercontent.com/assets/8365333/4662797/78185f72-5535-11e4-9993-37cf58c96a3d.PNG)

The same issue on different RDP servers, with different graphic setting.
May be I missing something obvious ? What Im doing wrong ?

wwwzys:
@Elyas-Crimean Has this issue been solved?

wwwzys:
@bmiklautz   “ looks for me like this is related to recent changes in GDI (CLRCONV_INVERT).  ”
How to solve ?

wwwzys:
@bmiklautz   “ looks for me like this is related to recent changes in GDI (CLRCONV_INVERT).  ”
How to solve ?

wwwzys:
@bmiklautz  i want to solve this problem,and i do not how to do. can you help me.

wwwzys:
@palmoff  who has solved this problem ? can you give me a hand ?

DonMag:
No guarantees, but (in iOS client app)...

In:
>    ios_freerdp_ui.m  

in:
>    static void ios_create_bitmap_context(mfInfo* mfi) { ... }

Change:
>    if (gdi->dstBpp == 16)

to:
>    if (gdi->dstBpp == 16 || gdi->dstBpp == 15)

wwwzys:
@DonMag yes, it worked!  You are great! Thank you very much.

kaizirlewagen:
@DonMag works perfect.

bmiklautz:
@DonMag I'll see that we get your fix on master. Thank you.

bmiklautz:
@DonMag could you test if PR #2428 fixes this problem on master?

DonMag:
@bmiklautz looks good to me - I hadn't seen ->bytesPerPixel when I initially looked at the problem.

R3G3N3R4T0R:
I have a similar issue now and it is on Artix x86_64 LXQt with OpenRC. The colors are off like the picture above and the server is running Win10. I have tried KRDC, Remmina and running xfreerdp in the terminal. Only Remmina have the right color from the start but it shifted if I track the window's size to change the resolution. 
[...]

PR #2428, opened 2015-02-26
Fixed bitmap context for 15bit color depth

akallabeth:
Implements the fix for #2152, now using same comparison as mac client for bitmap context creation.
@DonMag thx for the hint.


bmiklautz:
+1

Issue #2104, opened 2014-09-16
Send release event for pressed keys if client loses focus (xfreerdp)


leoluk:
If a shortcut is used to switch away from the xfreerdp window, no release event is send to the remote server, resulting in the keys being stuck.

The same issue has been affecting Remmina: https://github.com/FreeRDP/Remmina/issues/7

It was fixed with this commit: https://github.com/FreeRDP/Remmina/commit/17f023202bed56cf43383d50a8014f8eab77c5ee

Vladimir-csp:
Why isn't it already merged after almost a year? Switching desktops away from xfreerdp window results in stuck Super key inside RDP session.
I will test the patch...

Vladimir-csp:
I've applied the patch to version 1.1.0~git20140921.1.440916e+dfsg1-4 (Debian jessie), replacing "xf_keyboard" with "xf_kbd". But Super key is still stuck when returning to freedesktop window after swithcing from it with keyboard.

leoluk:
Remmina had a similar issue, see https://github.com/FreeRDP/Remmina/issues/7#issuecomment-126452716

Vladimir-csp:
does [this](https://github.com/FreeRDP/Remmina/commit/cc427c785726594bbdfbcdaf36601f4f82b66fe1) commit help? if so, how it can be adapted for xfreerdp?

bmiklautz:
Should be fixed on master. Closing.

PR #2430, opened 2015-02-27
release keys when xfreerdp is unfocused to prevent stuck keys

serayrosh:
fixes issue #2104


ismail:
This also happens on Windows, so another fix is needed for that too.

bmiklautz:
@freerdp-bot test

mathias-pahlen:
Like @ismail, I can confirm that this (still) is a bug in Windows.

Issue #2064, opened 2014-08-27
Scaling in iOS


genxstylez:
Hi, I am trying to scale the RDPSession to fit to iphone screen for the iOS client.

I have tried smart sizing, desktop resize etc. and with no success.

Please help.

Fluke211:
I am also struggling to fit a 1600 x 1200 RDP session on an iPad. 

genxstylez, did you get it to work? Can you post your *.rdp file?

Issue #1701, opened 2014-02-26
Scaling Up Resolution from low DPI to high DPI Question


normanhh3:
Hi folks,

I am trying to use the aFreeRdp android app to connect to a screen with dimensions 800x480 and I would love to have it scale to fill my 1280x720 4.7" smartphone screen.

Unfortunately that does not seem to work, even if I set a custom window size at the requested dimensions.  The image simply takes up the specified screen real estate in normal pixels, not accounting for DPI differences.

I have seen some solutions that require adjusting the DPI on the remote end before connecting and that might be a workable solution.

Is there any solution yet discussed or devised to do some pixel expansion so that the smaller size remote screen will take up the full screen that I have available?

Thank you for any pointers you can provide.

PR #2438, opened 2015-03-04
Enable display scaling in client

irichter:
This is a patch to allow the command-line client to tell the server what display scaling to use. It supports telling the server the following:
- Physical display size (in millimeters)
- Display orientation
- Text scaling to use (in %)

I have tested the it from a linux command-line client connecting to a Windows 8.1 computer, and was only able to get the scaling parameters to have any effect. I do not have time to diagnose this fully, and hope that someone else will be able to work out the issues with the other parameters.

This might be a start toward closing #1701 and #2064.

Future work might attempt to implement [\[MS-RDPEDISP\] Remote Desktop Protocol: Display Update Virtual Channel Extension](https://msdn.microsoft.com/en-us/library/dn366738.aspx), which would allow changing the scaling during a session (on servers that support this).


awakecoding:
@freerdp-bot test.

awakecoding:
@freerdp-bot test.

colemickens:
@irichter, Is it necessary to specify all of those parameters? Or can I just pass a single one to say "Scale at 1.5x"?

colemickens:
I built it locally and am running
`./xfreerdp /u:Default /p:default /v:192.168.122.219 /scale-desktop:200`

but I'm not seeing any changes.

irichter:
> @irichter, Is it necessary to specify all of those parameters? Or can I just pass a single one to say "Scale at 1.5x"?

@colemickens: Per [[MS-RDPBCGR] section 2.2.1.3.2](https://msdn.microsoft.com/en-us/library/cc240510.aspx):
* Both `/pwidth` and `/pheight` (called desktopPhysicalWidth and desktopPhysicalHeight in the specification) must be given valid values or both will be ignored.
* `/orientation` (called desktopOrientation in the specification) can be specified on its own.
* `/scale-desktop` and `/scale-device` (called desktopScaleFactor and deviceScaleFactor in the specification) must both be specified with valid values, or both will be ignored. Alternately, setting `/scale:NNN` is the same as setting `/scale-desktop:NNN /scale-device:NNN`.

daurnimator:
I'd love this feature. managing a windows server from my 4K linux machine is quite the exercise for my eyes.

bmiklautz:
@irichter am I mistaken or must the desktopScaleFactor  always be 100, 140 or 180? In this case the "scale" option as currently implemented makes only sense for those values as well.

Otherwise the pull request looks good to me.
+1

irichter:
@bmiklautz deviceScaleFactor is the one that is limited to 100, 140, or 180 (not desktopScaleFactor), but your statement about the "scale" option is correct.

Perhaps I should have added logic to coerce to the nearest allowed value for deviceScaleFactor, and pass through desktopScaleFactor as provided by the user? Although if I do that, I should probably also coerce desktopScaleFactor to a value between 100 and 500. That said, who knows what additional valid values may be added for Windows 10, so any restrictions implemented now might end up coming back to cause problems later.

bmiklautz:
@irichter I would restrict the values according to the current documentation. If new values getting available the documentation should get updated.

colemickens:
By the way, this is working for me. It must have been user error before, or I wasn't using my custom built freerdp. Either way, I'm using this successfully for connecting to a remote machine from a high dpi client.

There are a few things that aren't scaled properly. I'm not sure if that's due to the Windows implementation of scaling, or due to how xfreerdp sends arguments... I'll attempt to shake it out by comparing freerdp with Remote Desktop from a Windows install somewhere.

Thanks for this patch @irichter, it would be very hard for me to use RDP otherwise!

hardening:
Looks good to me except:

* your git log is not clean with an upstream merge in it;
* Is it safe to send 0 as physical width / height when no argument is given on the command line ?

colemickens:
Sorry to use this as a support channel, but does anyone have any idea what the difference between scale-device and scale-desktop are? I can't tell if it's desirable to set them both, just one - a bit difficult to tell what's going on without maybe taking screenshots and comparing closely.

@irichter Are you going to clean up the git log so this can be pushed it?

colemickens:
@irichter Can I resubmit this with a clean git history if you're not interested in doing so?

@hardening Is there a normal way of handling stale pull requests? I don't want to step on toes, and I'm also unsure of how copyright works in a case of a pull request.

I have this on my master: https://github.com/colemickens/FreeRDP and will keep rebasing when I think of it... I can send a new PR if everyone is okay with that.

bmiklautz:
superseded by #2756

Issue #2427, opened 2015-02-25
UPN Suffix doesn't work at all


Edlanta:
Hi,

i noticed that FreeRDP isn't able to authenticate with a UPN Suffix at all.
If i connect through a TS Gateway Server the following message is displayed:

rts_connect error! Status Code: 401
HTTP/1.1 401 Unauthorized
Content-Type: text/plain
Server: Microsoft-IIS/8.5
WWW-Authenticate: Negotiate
WWW-Authenticate: NTLM
WWW-Authenticate: Basic realm="GATEWAY-SERVER"
X-Powered-By: ASP.NET
Date: Wed, 25 Feb 2015 16:12:26 GMT
Content-Length: 13

rts_connect error!
rpc_connect failed!
Error: protocol security negotiation or connection failure

Without TS Gateway the following message is displayed:

credssp_recv() error: -1
Authentication failure, check credentials.
If credentials are valid, the NTLMSSP implementation may be to blame.
Error: protocol security negotiation or connection failure

I compiled the current version from git as described at "Compilation".

I think this has something to do with username parsing. FreeRDP is splitting the username at the @ symbol and uses the string before @ as username and the string after @ as domainname.
For example:   
UPN Suffix: 123456@TEST.test 
will become
Username: 123456
domain: TEST.test

This is the right behaviour if you use Kerberos, but not the right behaviour if you will logon with UPN Suffix.

I changed the username parsing in the following way (freerdp_parse_username() in cmdline.c):
If a domain AND username is given we shouldn't split the username.
If ONLY the username is given we should split the username at the @ sign. And everything before the @ will become settings->Username, and everything after the @ will become settings->Domain
This change will get UPN Suffix working with a normal server, but unfortunately not with TS Gateway.

I will take a look into the NTLM implementation, because i think there is something wrong with Username handling as well.

Can somebody check if this behaviour is reproducible?

I have checked if this is working with a RDP Client from WIndows 7, and indeed it works, with and without TS Gateway.

Edlanta:
Hi,

i found a solution for this problem. If we are using NTLM in combination with a UPN Suffix, we need to make sure that we don't send the Domain in the NTLM authentication mechanism.

Here is the diff of my changes:

diff --git a/winpr/libwinpr/sspi/sspi.c b/winpr/libwinpr/sspi/sspi.c
index f9ec6a7..47aadc5 100644
--- a/winpr/libwinpr/sspi/sspi.c
[...]

PR #2442, opened 2015-03-06
Added support for UPN Suffixes to FreeRDP. 

Edlanta:
This includes a fix for issue: 
https://github.com/FreeRDP/FreeRDP/issues/2427

These changes are working for me. I can now logon with UPN Suffix to a win2012r2 Server and additionally to a win2012r2 Server which is behind a 2012 TS Gateway Server.
But this changes have some restrictions:
- If you want to use UPN Suffix Logon, you MUST enter a domain name with /d: even if this 
  domain name doesn't exist at all.
- If you want to use UPN Suffix Logon with TS Gateway, you MUST additionally enter a Gateway 
  domain name with /gd: even if this domain name doesn't exist at all.

This restrictions are necessary, because this is the only way we can decide if the username is for Kerberos authentication or for normal authentication with UPN suffix. I explained this in issue #2427

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -1171,6 +1171,10 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings, 
 	DWORD flags;
 	BOOL compatibility;
 	COMMAND_LINE_ARGUMENT_A* arg;
+       char* username = NULL;
+       char* gw_username = NULL;
+       char* user;
+       char* domain;
 
 	compatibility = freerdp_client_detect_command_line(argc, argv, &flags);
 
@@ -1422,13 +1426,7 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings, 
 		}
 		CommandLineSwitchCase(arg, "u")
 		{
-			char* user;
-			char* domain;
-
-			freerdp_parse_username(arg->Value, &user, &domain);
-
-			settings->Username = user;
-			settings->Domain = domain;
+			username = _strdup(arg->Value);
 		}
 		CommandLineSwitchCase(arg, "d")
 		{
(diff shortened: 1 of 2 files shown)


bmiklautz:
@freerdp-bot test

client/common/cmdline.c
```diff
@@ -1962,6 +1953,31 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
+       
+       // Check if Username is set and Domain is missing.
+       // If this is the case we should split Username and Domain at the @ sign (e.g. for Kerberos)
+       // Else we should let the username string as it is (e.g. for UPN Suffix)
+       if (username && !settings->Domain)
+       {
+	  freerdp_parse_username(username, &user, &domain);
+         
+	  settings->Username = user;
+	  settings->Domain = domain;
```
akallabeth: @Edlanta got yourself a memory leak here.

client/common/cmdline.c
```diff
@@ -1962,6 +1953,31 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
+       else
+	  settings->Username = username;
+       
+       // Same as above for Gateway Username and Domain.
+       if (gw_username && !settings->GatewayDomain)
+       {
+	  freerdp_parse_username(gw_username, &user, &domain);
+         
+	  settings->GatewayUsername = user;
+	  settings->GatewayDomain = domain;
```
akallabeth: @Edlanta and here again.

Edlanta:
Sorry for the second commit, wasnt really fit in the morning as i changed this. Now everything should work well again.

akallabeth:
@Edlanta Ok, so you reverted the broken change, still ```freerdp_parse_username``` allocates new memory for the output parameters, but does not free up the input argument. That one is still leaking. So either only ```_strdup``` the values of ```username``` and ```gw_username``` where they are actually assigned to settings or free up the duplicated strings where they are no longer in use.

client/common/cmdline.c
```diff
@@ -1962,6 +1953,51 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
+	  settings->GatewayUsername = _strdup(gw_username);
+       
+       if (user != NULL)
+	 free(user);
+       if (domain != NULL)
+	 free(domain);
+       if (username != NULL)
+	 free(username);
+       if (gw_username != NULL)
+	 free(gw_username);
```
nfedera: You don't have to check pointers passed to free() for NULL. Please see man free ;)

akallabeth:
Implemented this feature with #2701 in a cleaner fashion.
@Edlanta please close your pull.

PR #2447, opened 2015-03-09
Fix poor sound quality with ALSA, add OSS sound system support, FreeBSD patches

ivan-83:
* Fix poor sound quality with ALSA
* decrease "ALSA lib pcm.c:7339:(snd_pcm_recover) underrun occurred"
* little code cleanup


bmiklautz:
@freerdp-bot test

Adonai:
Upvoted

bmiklautz:
@ivan-83 it would be nice if you could split this PR in multiple topic specific ones.

bmiklautz:
@freerdp-bot test

ivan-83:
Retry now.

I specifically here to register to send these patches.
How to split the query into several I have not yet figured out.

bmiklautz:
@freerdp-bot test

CMakeLists.txt
```diff
@@ -358,6 +358,14 @@ if(APPLE)
...
 endif(APPLE)
 
+# BSD
+if(${CMAKE_SYSTEM_NAME} MATCHES "BSD")
+	set(BSD TRUE)
+	if(${CMAKE_SYSTEM_NAME} MATCHES "FreeBSD")
+		set(FREEBSD TRUE)
+	endif()
+endif()
+
```
bmiklautz: If we support more other system we should move that out to some extra file we can include but for now very nice ;).

winpr/libwinpr/synch/wait.c
```diff
@@ -61,7 +61,7 @@
 
 #ifndef _WIN32
 
-#include <alloca.h>
+#include <stdlib.h>
```
bmiklautz: Is alloca.h incuded in stdlib?

ivan-83: Yes, linux include alloca.h in stdlib.h.

In man:
http://linux.die.net/man/3/alloca
#include <alloca.h>

https://www.freebsd.org/cgi/man.cgi?query=alloca&apropos=0&sektion=0&manpath=FreeBSD+10.0-RELEASE&arch=default&format=html
#include <stdlib.h>

include/freerdp/client/rdpsnd.h
```diff
@@ -37,9 +37,7 @@ struct _RDPSND_WAVE
 	UINT16 wTimeStampA;
 	UINT16 wTimeStampB;
 
-	UINT16 wLatency;
```
hardening: I don't know if people using this include have something to say about that but it looks like API + ABI break.

libfreerdp/locale/timezone.c
```diff
@@ -1498,7 +1498,9 @@ static UINT64 freerdp_windows_gmtime()
 	UINT64 windows_time;
 
 	time(&unix_time);
-	windows_time = ((UINT64) unix_time * 10000000) + 621355968000000000ULL;
```
hardening: The original code looked clearer to me....

include/freerdp/client/rdpsnd.h
awakecoding: that variable was practical, and as a matter of fact, it would break it for the wayk client. I'd rather get it back than remove it.

libfreerdp/locale/timezone.c
awakecoding: I kind of agree it wasn't bad in its original state, don't see much of a point in splitting it over multiple statements like that.

PR #2448, opened 2015-03-09
Fixed a regression on the #2333 resolution

eledoux:
Revisited the resolution of #2333 I did a few days ago. It appears I was still confused about the definition of the select() timeout. During the issue's resolution, I understood in the select()'s manual "Indefinitely" more as "undefined" whereas we well need a NULL timeout or a zero-ed timeout according the timeouts values.

The resolution of #2333 is preserved.


bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test.

Issue #2445, opened 2015-03-09
Certficate Name Check Should be Case-INsensitive


KenHolloway:
Since DNS names are not case sensitive, the comparison of the certificate names to the host name (or the certificate name specified on the command line) should also be case-insensitive - www.google.com should match www.Google.com.

WarheadsSE:
I've been hitting this horribly as well. Looking into it at the moment, as I now have people clamoring (rightfully) about this.

WarheadsSE:
This should now be closable.

PR #2455, opened 2015-03-10
Fix case sensitive hostname comparison in tls

WarheadsSE:
To do this I've swapped _strnicmp with memcmp. Seemless, but does lock it to the restrictions of that function.

Addresses https://github.com/FreeRDP/FreeRDP/issues/2445


bmiklautz:
@freerdp-bot test

bmiklautz:
+1

hardening:
+1

PR #2456, opened 2015-03-11
Fix incorrect usage of realloc

zhangzl2013:


client/Windows/wf_cliprdr.c
```diff
@@ -1678,12 +1694,7 @@ static int wf_cliprdr_server_format_data_request(CliprdrClientContext* context,
 					clipboard->file_names[clipboard->nFiles] = (LPWSTR) malloc(cchWideChar * 2);
 					MultiByteToWideChar(CP_ACP, MB_COMPOSITE, p, len, clipboard->file_names[clipboard->nFiles], cchWideChar);
 
-					if (clipboard->nFiles == clipboard->file_array_size)
-					{
-						clipboard->file_array_size *= 2;
-						clipboard->fileDescriptor = (FILEDESCRIPTORW**) realloc(clipboard->fileDescriptor, clipboard->file_array_size * sizeof(FILEDESCRIPTORW*));
-						clipboard->file_names = (WCHAR**) realloc(clipboard->file_names, clipboard->file_array_size * sizeof(WCHAR*));
-					}
+					map_ensure_capacity(clipboard);
```
hardening: I imagine the call to map_ensure_capacity() can fail, can't it ?

client/Windows/wf_cliprdr.c
```diff
@@ -1348,9 +1354,19 @@ static void wf_cliprdr_array_ensure_capacity(wfClipboard* clipboard)
...
-		clipboard->file_names = (WCHAR**) realloc(clipboard->file_names, clipboard->file_array_size * sizeof(WCHAR*));
+		int new_size;
+		FILEDESCRIPTORW **new_fd;
+		WCHAR **new_name;
+
+		new_size = clipboard->file_array_size * 2;
+		new_fd = (FILEDESCRIPTORW**) realloc(clipboard->fileDescriptor, new_size * sizeof(FILEDESCRIPTORW*));
+		new_name = (WCHAR**) realloc(clipboard->file_names, new_size * sizeof(WCHAR*));
+		if (!new_fd || !new_name)
+			return;
```
hardening: if realloc(new_name) fails, we will have potentially an incorrect pointer in clipboard->fileDescriptor, and probably a leak of new_fd

libfreerdp/common/settings.c
```diff
@@ -140,9 +156,16 @@ void freerdp_device_collection_add(rdpSettings* settings, RDPDR_DEVICE* device)
...
-		settings->DeviceArray = (RDPDR_DEVICE**)
-				realloc(settings->DeviceArray, settings->DeviceArraySize * sizeof(RDPDR_DEVICE*));
+		UINT32 new_size;
+		RDPDR_DEVICE **new_array;
+
+		new_size = settings->DeviceArraySize * 2;
+		new_array = (RDPDR_DEVICE**)
+				realloc(settings->DeviceArray, new_size * sizeof(RDPDR_DEVICE*));
+		if (!new_array)
+			return;
```
hardening: Ok with the fix, but I think the function should be changed to return a BOOL (and all callers checking the return value) and we would return FALSE here.

winpr/libwinpr/utils/collections/Reference.c
```diff
@@ -78,10 +81,12 @@ wReference* ReferenceTable_GetFreeEntry(wReferenceTable* referenceTable)
 			return NULL;
 		}
 
-		referenceTable->size *= 2;
-		referenceTable->array = (wReference*) realloc(referenceTable->array,
-				sizeof(wReference) * referenceTable->size);
+		new_size = referenceTable->size * 2;
+		new_ref = (wReference*) realloc(referenceTable->array,
+				sizeof(wReference) * new_size);
```
hardening: new_ref is not check and that would segfault two lines later when ZeroMemorying referenceTable->array

hardening:
Except my comments +1.
Nice job Vic, I know how painful it is to treat these cases.

client/Windows/wf_cliprdr.c
zhangzl2013: yes, I'll fix it right now.

winpr/libwinpr/utils/collections/Reference.c
zhangzl2013: Fixed, thanks.

client/Windows/wf_cliprdr.c
zhangzl2013: Yes, but let's handle the return value issue in other patch series.

libfreerdp/common/settings.c
zhangzl2013: Let's focus return value issue in other patch series, what do you think? ;-)

zhangzl2013:
@hardening Thanks for your elaborative review.

akallabeth:
+1

@zhangzl2013 a hooray for error checks

MartinHaimberger:
+1

bmiklautz:
@zhangzl2013 nice exactly the one we miss! - I'm working on something similar for malloc/calloc (https://github.com/bmiklautz/FreeRDP/commit/f5007e0a671630614bb6708c012a641fea9c4d11 but still work in progress)

bmiklautz:
+1

Issue #2446, opened 2015-03-09
Certificate chain validation fails for certificate chains


KenHolloway:
'''I am connecting to a Windows 2012 server with a valid certificate (issued by Starfield), however, xfreerdp says:

```
xfreerdp --from-stdin -a 24 -T CCS-CW.{removed}.com -g 1024x768 --disable-menu-animations --disable-wallpaper --plugin cliprdr --plugin rdpsnd --plugin rdpdr --data disk:Home:/home/KHolloway -- -d {removed}.com -u KenHolloway --certificate-name cw.{removed}.com CCS-CW.{removed}.com


loading plugin cliprdr
loading plugin rdpsnd
loading plugin rdpdr
password: 
connected to CCS-CW.{removed}.com:3389
Certificate details:
        Subject: OU = Domain Control Validated, CN = cw.{removed}.com
        Issuer: C = US, ST = Arizona, L = Scottsdale, O = "Starfield Technologies, Inc.", OU = http://certs.starfieldtech.com/repository/, CN = Starfield Secure Certificate Authority - G2
        Thumbprint: c7:38:1f:b4:5d:5f:c6:cd:37:91:e8:eb:18:99:5b:5f:67:40:43
The above X.509 certificate could not be verified, possibly because you do not have the CA certificate in your certificate store, or the certificate has expired. Please look at the documentation on how to create local certificate store for a private CA.
Do you trust the above certificate? (Y/N) 
```

I have tried connecting both with "FreeRDP version 1.0.2" (installed on openSUSE from standard openSUSE repo), and with "FreeRDP version 1.2.1-dev (git b33bba0)", downloaded from git a few days ago.

I have checked the server's certificate using openSSL:

```
openssl s_client -connect ccs-cw.{removed}.com:3389 < /dev/null 
CONNECTED(00000003)
depth=2 C = US, ST = Arizona, L = Scottsdale, O = "Starfield Technologies, Inc.", CN = Starfield Root Certificate Authority - G2
verify return:1
depth=1 C = US, ST = Arizona, L = Scottsdale, O = "Starfield Technologies, Inc.", OU = http://certs.starfieldtech.com/repository/, CN = Starfield Secure Certificate Authority - G2
verify return:1
depth=0 OU = Domain Control Validated, CN = cw.{removed}.com
verify return:1
---
Certificate chain
 0 s:/OU=Domain Control Validated/CN=cw.{removed}.com
[...]

PR #2458, opened 2015-03-11
crypto: use PEM as certificate store file format

bmiklautz:
Currently the certificate format expected in FreeRDPs certificate store
is DER (ASN1). On most linux/unix systems the system certificate store
default format is PEM. Which is also the more common format used by CAs
to distribute their certificates.

Changing the default format to PEM allows the usage of system
certificates or published CA certificates without the need to convert them.

This fixes a part of issue #2446.


MartinHaimberger:
+1

hardening:
+1

PR #2459, opened 2015-03-11
Refactored WaitForSingleObject and CloseHandle

akallabeth:
* Every handle now has a structure containing the function pointers used for cleanup after wait, file descriptor retrieval and resource cleanup
* ```WaitForSingleObject``` and ```WaitForMultipleObjects``` now use function pointer to get file descriptors to wait on and clean up after wait


hardening:
@akallabeth I like your patch and I think it's the first step to handle all winpr operations this way (read, writes).
Just one remark: we would save space and CPU time if the callbacks were in a static structure, and the handle struct would have a pointer to structure. I don't think we need to have each operation in the handle struct, but just a pointer to the set of operations.

bmiklautz:
+1 much cleaner now \o/

PR #2468, opened 2015-03-15
fix size error introduced in commit 99b14815a53e556080f459a4998d94244803...

dupondje:
Commit 99b14815a53e556080f459a4998d942448031f08 introduces a bug that causes the client to crash/disconnect.
This was caused by a wrong change in the realloc size.


bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test.

bmiklautz:
+1

PR #2469, opened 2015-03-15
libfreerdp-core: Initialize values correctly

dupondje:
Values were not initialized anymore, causing alot of 'Network Disconnect' error.

### DIFF
--- a/libfreerdp/core/orders.c
+++ b/libfreerdp/core/orders.c
@@ -1314,7 +1314,7 @@ BOOL update_write_line_to_order(wStream* s, ORDER_INFO* orderInfo, LINE_TO_ORDER
 BOOL update_read_polyline_order(wStream* s, ORDER_INFO* orderInfo, POLYLINE_ORDER* polyline)
 {
 	UINT16 word;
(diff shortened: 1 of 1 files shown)


dupondje:
ORDER_FIELD_BYTE for example does not read the date unconditionally.
So if you don't initialize it, it can contain garbage.

bmiklautz:
@dupondje shouldn't they just be initialized to 0?

dupondje:
The PRIMARY_DRAWING_ORDER structure encapsulates a primary drawing order. All primary drawing orders employ special field encoding to reduce the number of bytes sent on the wire. Field encoding maintains a copy of the most recent field values that were used in each primary drawing order, in addition to common state information such as the last bounding rectangle used across all orders and the last order type. Only the fields that have changed are sent on the wire. This implies that all the fields in a primary drawing order are optional, their presence being denoted by the controlFlags and fieldFlags fields.

https://msdn.microsoft.com/en-us/library/cc241586.aspx

So we should ALWAYS use the previous value if the value isn't set in the new update.

PR #2470, opened 2015-03-15
Make handle operations static

hardening:
This patch moves the handle operations in a static struct.

Note that the test with named pipes fails after the patch, but it was already triggering errors in valgrind before. So I guess there's a problem either with pipes or with the test itself.

PR #2472, opened 2015-03-16
Winpr ops rebased

hardening:
Rebased version of https://github.com/FreeRDP/FreeRDP/pull/2470


bmiklautz:
@freerdp-bot test

akallabeth:
+1, reduces size per handle
@hardening nice catch with the reset of fd to -1

bmiklautz:
+1

PR #2474, opened 2015-03-16
Test Fixes, BipBuffer, TSG Cleanup, SCardTransmit Fix

awakecoding:


winpr/libwinpr/utils/collections/BipBuffer.c
```diff
@@ -0,0 +1,435 @@
...
+
+	bb = (wBipBuffer*) calloc(1, sizeof(wBipBuffer));
+
+	if (bb)
+	{
+		SYSTEM_INFO si;
+
+		GetSystemInfo(&si);
+
+		bb->pageSize = (size_t) si.dwPageSize;
```
hardening: The page size is not supposed to change over time, perhaps the GetSystemInfo() could be done just once and the result stored in a static variable...

PR #2476, opened 2015-03-17
fix auidin OSS, add OSS to tsmf, add commandline samples and syntax, add debug messages

ivan-83:


channels/tsmf/client/gstreamer/tsmf_X11.c
```diff
@@ -61,8 +64,8 @@ struct X11Handle
 
 static const char* get_shm_id()
 {
-	static char shm_id[64];
-	snprintf(shm_id, sizeof(shm_id), "com.freerdp.xfreerdp.tsmf_%016X", GetCurrentProcessId());
+	static char shm_id[128];
+	snprintf(shm_id, sizeof(shm_id), "/com.freerdp.xfreerdp.tsmf_%016X", GetCurrentProcessId());
```
hardening: nice catch, I discovered recently that shmIds are supposed to be paths...

libfreerdp/core/fastpath.c
```diff
@@ -466,7 +483,7 @@ static int fastpath_recv_update_data(rdpFastPath* fastpath, wStream* s)
 			if ((fastpath->fragmentation != FASTPATH_FRAGMENT_FIRST) &&
 					(fastpath->fragmentation != FASTPATH_FRAGMENT_NEXT))
 			{
-				WLog_ERR(TAG,  "Unexpected FASTPATH_FRAGMENT_LAST");
+				WLog_ERR(TAG, "fastpath_recv_update_data: Unexpected FASTPATH_FRAGMENT_LAST");
```
hardening: You should use %s and __FUNCTION__ here, as if we rename the function a lot more modifications will be needed.

libfreerdp/core/peer.c
```diff
@@ -455,8 +455,10 @@ static int peer_recv_callback(rdpTransport* transport, wStream* s, void* extra)
 	switch (rdp->state)
 	{
 		case CONNECTION_STATE_INITIAL:
-			if (!rdp_server_accept_nego(rdp, s))
+			if (!rdp_server_accept_nego(rdp, s)) {
+				WLog_DBG(TAG, "peer_recv_callback: CONNECTION_STATE_INITIAL - rdp_server_accept_nego() fail");
 				return -1;
```
hardening: I feel like this log will be very verbose when you connect to a RDP security only server, with 3 connection attempts (NLA, then TLS, and finally RDP)...

hardening:
@ivan-83 this PR is interesting, but it should be cutted in 2: OSS for tsmf  and logs.
In logs you should use \_\_FUNCTION\_\_ if you wanna print the current function name, like that if we rename the function (easy task in eclipse) we don't have to check for all the logs.

hardening:
@freerdp-bot test my friend

hardening:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

libfreerdp/core/peer.c
ivan-83: WLog_DBG display messages only then you run: env WLOG_LEVEL=DEBUG xfreerdp

libfreerdp/core/fastpath.c
ivan-83: Ok.
I will fix it late, in next pull request, by adding:
#define WLOG_ERR(_tag, _fmt, args...)					\
	    WLog_ERR(_tag, __FUNCTION__": "_fmt, ##args)
and same for WLog_DBG.

hardening: WLog_ERR(TAG, "%s: Unexpected FASTPATH_FRAGMENT_LAST", \_\_FUNCTION\_\_);

would be probably better

ivan-83: To many work by hands, very long lines with code:
WLOG_ERR(TAG, "Unexpected FASTPATH_FRAGMENT_LAST");

Also: __FUNCTION": "_fmt - calculated at build.
Also WLOG_ERR in any time can be modified to add line num or other info, or format may be changed.

hardening: The "calculated at build" argument does not really apply when this code is supposed to be almost never called ;)

PR #2482, opened 2015-03-19
tls: fix error handling of no error code.

llyzs:
On Windows WSAGetLastError() in some cases returns no error even though tls got SSL_ERROR_SYSCALL. This was causing unexpected client disconnection, maybe once every 20-30 times. This fix treats this situation as blocking and retry.

@akallabeth @mfleisz

### DIFF
--- a/libfreerdp/crypto/tls.c
+++ b/libfreerdp/crypto/tls.c
@@ -92,16 +92,7 @@ static int bio_rdp_tls_write(BIO* bio, const char* buf, int size)
 				break;
 
 			case SSL_ERROR_SYSCALL:
-				error = WSAGetLastError();
-				if ((error == WSAEWOULDBLOCK) || (error == WSAEINTR) ||
-					(error == WSAEINPROGRESS) || (error == WSAEALREADY))
-				{
-					BIO_set_flags(bio, (BIO_FLAGS_WRITE | BIO_FLAGS_SHOULD_RETRY));
-				}
-				else
-				{
-					BIO_clear_flags(bio, BIO_FLAGS_SHOULD_RETRY);
-				}
+				BIO_set_flags(bio, (BIO_FLAGS_WRITE | BIO_FLAGS_SHOULD_RETRY));
 				break;
 
 			case SSL_ERROR_SSL:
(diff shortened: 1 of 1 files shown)


hardening:
The SSL_ERROR_SYSCALL error code doesn't seems accurate at all when we see all the cases where we choose to ignore it and retry. 
What about setting retry in all cases when we catch a SSL_ERROR_SYSCALL ? 
Because looking at the last error in errno or WSAGetLastError() seems totally wrong when OpenSSL could have used BIOs that don't do any I/O operation (and so don't modify errno). 

Perhaps we should give a try with SSL_ERROR_SYSCALL => always retry. If someone has an example that would break...

llyzs:
I have not reproduced a case yet that SSL_ERROR_SYSCALL happens and WSAGetLastError returns a critical error, but this is according to the OpenSSL document:

--- quote ---

SSL_ERROR_SYSCALL

    Some I/O error occurred. The OpenSSL error queue may contain more information on the error. If the error queue is empty (i.e. ERR_get_error() returns 0), ret can be used to find out more about the error: If ret == 0, an EOF was observed that violates the protocol. If ret == -1, the underlying BIO reported an I/O error (for socket I/O on Unix systems, consult errno for details).

--- unquote ---

llyzs:
I am ok to set it to always retry for SSL_ERROR_SYSCALL if others have no objection.

llyzs:
OK changed as suggested. Please take a look.

akallabeth:
@llyzs Is there really no error set by ```SSL_read``` and ```SSL_write``` or does ```WSAGetLastError``` simply not handle that specific ```errno``` value? ```WSAGetLastError``` is missing a default statement in the switch, so maybe we are just hiding something here...

hardening:
@akallabeth I think OpenSSL's man with SSL_ERROR_SYSCALL is wrong. It doesn't make any sense to look at errno or WSAGetLastError() as the underlying transport is always done by BIOs that can be _not_ socket / file descriptor based. Even worse, if the BIO do some debug printing after having read, errno really won't have the expected value. Today we don't understand clearly what lead openSSL to return a SSL_ERROR_SYSCALL.
@llyzs if you can reproduce the case easily perhaps you could try to add some debug infos with read() and write() calls (underlying socket), so that we could have a clue of what's going on...

If we can't find a situation where it bugs, I think we should take @llyzs's version that just ignore SSL_ERROR_SYSCALL.

llyzs:
@hardening that is certainly not easily reproducible but I don't have a clue how to debug this since when this happens, a retry always works.

I suggest we just take this fix. If we find something in the future, we will be more likely to come back here with better clue.

hardening:
+1

PR #2483, opened 2015-03-19
Feat/freerdp last error name

akallabeth:
Added new functions allowing the retrieval of an error name and description similar to the one already present for ```ERRINFO_*```


libfreerdp/core/errbase.c
```diff
@@ -0,0 +1,88 @@
...
+
+#include "errinfo.h"
+
+#define TAG FREERDP_TAG("core")
+
+#define ERRBASE_DEFINE(_code)	    { ERRBASE_##_code , "ERRBASE_" #_code , ERRBASE_##_code##_STRING }
+
+/* Protocol-independent codes */
+
+#define ERRBASE_POSTCONNECTERROR_STRING \
```
mfleisz: This looks like a leftover? (this might be problematic for some preprocessors because of the \ at the end)

mfleisz:
+1

bmiklautz:
I think this is rather useful ;).
+1

PR #2484, opened 2015-03-19
RDP8 Gateway support

dvincent-devolutions:


awakecoding:
@freerdp-bot test.

awakecoding:
@freerdp-bot test.

awakecoding:
@freerdp-bot test.

libfreerdp/codec/color.c
```diff
@@ -1629,6 +1629,168 @@ void freerdp_clrconv_free(HCLRCONV clrconv)
...
+	nSrcPad = (nSrcStep - (nAlignedWidth / 8));
+	nDstPad = (nDstStep - (nWidth * dstBytesPerPixel));
+
+	if (FREERDP_PIXEL_FORMAT_IS_ABGR(DstFormat))
+		invert = TRUE;
+
+	if (FREERDP_PIXEL_FORMAT_FLIP(DstFormat) == FREERDP_PIXEL_FLIP_VERTICAL)
+		vFlip = TRUE;
+
+	if (dstBytesPerPixel == 4)
```
hardening: Perhaps we should warn loudly here when dstBytesPerPixel != 4 ?

libfreerdp/codec/color.c
```diff
@@ -3312,6 +3474,16 @@ int freerdp_image_copy(BYTE* pDstData, DWORD DstFormat, int nDstStep, int nXDst,
 		status = freerdp_image8_copy(pDstData, DstFormat, nDstStep, nXDst, nYDst,
 				nWidth, nHeight, pSrcData, SrcFormat, nSrcStep, nXSrc, nYSrc, palette);
 	}
+	else if (srcBitsPerPixel == 1)
```
hardening: The mix of if with srcBitsPerPixel and srcBytesPerPixel is a little misleading.
What's the value of srcBytesPerPixel when we hit the case srcBitsPerPixel = [1 | 4] ? If it's 0 then perhaps we could add the appropriate if() and catch srcBitsPerPixel in that statement...

Note: we silently ignore srcBytesPerPixel not in [1, 2, 3, 4] we should at least warn when we hit these cases

winpr/libwinpr/synch/mutex.c
```diff
@@ -54,13 +54,14 @@ static int MutexGetFd(HANDLE handle)
...
-	WINPR_MUTEX *mutex = (WINPR_MUTEX *) handle;
+BOOL MutexCloseHandle(HANDLE handle)
+{
+	WINPR_MUTEX* mutex = (WINPR_MUTEX*) handle;
 
 	if (!MutexIsHandled(handle))
 		return FALSE;
 
-	if(pthread_mutex_destroy(&mutex->mutex))
+	if (!pthread_mutex_destroy(&mutex->mutex))
```
hardening: is it me or we're leaking the handle here ?

awakecoding: no, actually this was changed to prevent a leak. "If successful, the pthread_mutex_destroy() and pthread_mutex_init() functions shall return zero; otherwise, an error number shall be returned to indicate the error." this means that on success the handle was leaked. We could fix it to free it under failure and success.

PR #2485, opened 2015-03-20
core: Fix server side nla

mfleisz:


awakecoding:
@mfleisz can you explain why the fix is apparently in client code when the pull request says it fixes server-side nla?

mfleisz:
Don't merge this pull - apparently it breaks client side NLA.
@awakecoding: The problem is that nla_recv_pdu (and therefore nla_client_recv) is called by the server side nla code through nla_recv (see the auth-loop in nla_server_authenticate). That's the reason why server-side NLA is currently failing. To be honest I am kinda confused with all the different recv functions and wasn't 100% sure how they were supposed to be used. Maybe you can have a short look at the code but I think that's the reason why server side NLA is broken (because it calls code that supposed to be called by the client).

PR #2443, opened 2015-03-20
Incorrect extents calculation in region16_intersect_rect (libfreerdp/codec/region.c)

realjiangms:
There's an issue in libfreerdp/codec/region.c. When there's more than 2 rectangles in the region structure, region16_intersect_rect would calculate extents by all 'intersected' sub rectangles.
But it always extend the extents to (0,0) because it initialize the new extents as (0,0,0,0) and union later rectangles with this empty point by simple MIN/MAX calculation.

In addition, the semantics of rectangle_is_empty is also wrong.

A fix could be as following:

```diff
diff --git a/libfreerdp/codec/region.c b/libfreerdp/codec/region.c
index 95ce193..739b655 100644
--- a/libfreerdp/codec/region.c
+++ b/libfreerdp/codec/region.c
@@ -132,7 +132,7 @@ static RECTANGLE_16 *region16_extents_noconst(REGION16 *region)

 BOOL rectangle_is_empty(const RECTANGLE_16 *rect)
 {
-   return (rect->left + rect->top + rect->right + rect->bottom) ? TRUE : FALSE;
+   return ((rect->left == rect->right) || (rect->top == rect->bottom)) ? TRUE : FALSE;
 }

 BOOL region16_is_empty(const REGION16 *region)
@@ -770,10 +770,17 @@ BOOL region16_intersect_rect(REGION16 *dst, const REGION16 *src, const RECTANGLE
            usedRects++;
            dstPtr++;

-           newExtents.top = MIN(common.top, newExtents.top);
-           newExtents.left = MIN(common.left, newExtents.left);
-           newExtents.bottom = MAX(common.bottom, newExtents.bottom);
-           newExtents.right = MAX(common.right, newExtents.right);
+            if (rectangle_is_empty(&newExtents))
+            {
+                newExtents = common;
+            }
+            else
+            {
+                newExtents.top = MIN(common.top, newExtents.top);
+                newExtents.left = MIN(common.left, newExtents.left);
+                newExtents.bottom = MAX(common.bottom, newExtents.bottom);
+                newExtents.right = MAX(common.right, newExtents.right);
+            }
        }
    }
```

### DIFF
--- a/libfreerdp/codec/region.c
+++ b/libfreerdp/codec/region.c
@@ -132,10 +132,10 @@ static RECTANGLE_16 *region16_extents_noconst(REGION16 *region)
 
 BOOL rectangle_is_empty(const RECTANGLE_16 *rect)
 {
-	/* A rectangle with width <= 0 or height <= 0 should be regarded
+	/* A rectangle with width = 0 or height = 0 should be regarded
 	 * as empty.
 	 */
-	return ((rect->left >= rect->right) || (rect->top >= rect->bottom)) ? TRUE : FALSE;
+	return ((rect->left == rect->right) || (rect->top == rect->bottom)) ? TRUE : FALSE;
 }
 
 BOOL region16_is_empty(const REGION16 *region)
--- a/libfreerdp/codec/region.c
+++ b/libfreerdp/codec/region.c
@@ -132,7 +132,10 @@ static RECTANGLE_16 *region16_extents_noconst(REGION16 *region)
 
 BOOL rectangle_is_empty(const RECTANGLE_16 *rect)
 {
-	return (rect->left + rect->top + rect->right + rect->bottom) ? TRUE : FALSE;
+    /* An rectangle with width <= 0 or height <= 0 should be regarded
+     * as empty.
+     */
+	return ((rect->left >= rect->right) || (rect->top >= rect->bottom)) ? TRUE : FALSE;
 }
 
 BOOL region16_is_empty(const REGION16 *region)
@@ -770,10 +773,21 @@ BOOL region16_intersect_rect(REGION16 *dst, const REGION16 *src, const RECTANGLE
 			usedRects++;
 			dstPtr++;
 
-			newExtents.top = MIN(common.top, newExtents.top);
-			newExtents.left = MIN(common.left, newExtents.left);
-			newExtents.bottom = MAX(common.bottom, newExtents.bottom);
-			newExtents.right = MAX(common.right, newExtents.right);
+            if (rectangle_is_empty(&newExtents))
+            {
+                /* Check if the existing newExtents is empty. If it is empty, use 
+                 * new common directly. We do not need to check common rectangle 
+                 * because the rectangles_intersection() ensures that it is not empty.
+                 */
+                newExtents = common;
+            }
+            else
+            {
+                newExtents.top = MIN(common.top, newExtents.top);
+                newExtents.left = MIN(common.left, newExtents.left);
+                newExtents.bottom = MAX(common.bottom, newExtents.bottom);
+                newExtents.right = MAX(common.right, newExtents.right);
+            }
 		}
 	}
 
--- a/libfreerdp/codec/region.c
+++ b/libfreerdp/codec/region.c
@@ -132,9 +132,9 @@ static RECTANGLE_16 *region16_extents_noconst(REGION16 *region)
 
 BOOL rectangle_is_empty(const RECTANGLE_16 *rect)
 {
-    /* An rectangle with width <= 0 or height <= 0 should be regarded
-     * as empty.
-     */
+	/* A rectangle with width <= 0 or height <= 0 should be regarded
+	 * as empty.
+	 */
 	return ((rect->left >= rect->right) || (rect->top >= rect->bottom)) ? TRUE : FALSE;
 }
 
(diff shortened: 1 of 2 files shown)


libfreerdp/codec/region.c
```diff
@@ -132,7 +132,10 @@ static RECTANGLE_16 *region16_extents_noconst(REGION16 *region)
 
 BOOL rectangle_is_empty(const RECTANGLE_16 *rect)
 {
-	return (rect->left + rect->top + rect->right + rect->bottom) ? TRUE : FALSE;
+    /* An rectangle with width <= 0 or height <= 0 should be regarded
+     * as empty.
+     */
```
hardening: I really don't like the idea that an invalid rectangle with right > left or bottom > top would be considered as empty.

libfreerdp/codec/test/TestFreeRDPRegion.c
```diff
@@ -632,7 +632,7 @@ static int test_norbert_case() {
 		0, 0, 1920, 1080
 	};
 	RECTANGLE_16 expected_inter_extents = {
-		0, 0, 1920, 1078
+		2, 0, 1920, 1078
 	};
```
hardening: any reason to change 0 to 2 ?

realjiangms: I found the expected result is not correct when this test report failure.
The code above this line is:

    RECTANGLE_16 inRectangles[5] = {
            {1680,    0, 1920,  242},
            { 294,  242,  971,  776},
            {1680,  242, 1920,  776},
            {1680,  776, 1920, 1036},
            {   2, 1040,   53, 1078}
    };
=======================
The 'union' of above rects should be 2,0,1920,1078 
=======================
    RECTANGLE_16 screenRect = {
        0, 0, 1920, 1080
    };

=======================
Then intersect with screenRect 0,0,1920.,1080, the result should be 2,0,1920,1078 
=======================
    RECTANGLE_16 expected_inter_extents = {
        2, 0, 1920, 1078
    };

So I think the expected rect should be fixed

hardening:
Your unitary is nice to prove your correction on rectangle_is_empty(). Anyway this single correction is sufficient to have the unitary test to work.
So this lead me to ask for a test that would demonstrate the need for the changes in region16_intersect_rect()

libfreerdp/codec/test/TestFreeRDPRegion.c
```diff
@@ -644,6 +644,9 @@ static int test_norbert_case() {
 			goto out;
 	}
 
+	if (!compareRectangles(region16_extents(&region), &expected_inter_extents, 1) )
```
realjiangms: @hardening 
So I copy the extents test here to prove that before the 'intersect' operation, the result is as my expected {2, 0, 1920, 1078}

libfreerdp/codec/test/TestFreeRDPRegion.c
```diff
@@ -660,6 +663,64 @@ static int test_norbert_case() {
...
+static int test_empty_rectangle() {
+	REGION16 region, intersection;
+	int retCode = -1;
+	int i;
+
+	RECTANGLE_16 emptyRectangles[4] = {
+		{   0,    0,    0,    0},
+		{  10,   10,   10,   11},
+		{  10,   10,   11,   10},
+		{ 100,  100,    0,    0}
```
hardening: Please drop that one that is not an empty rectangle, but an invalid one.

libfreerdp/codec/region.c
realjiangms: I agree an invalid rectangle should be denied.
Actually my original idea is just test "right = left or bottom = top". If we don't like width <0 and height < 0, we can change to just check '='.
Or do you think we should be more strict, only allow (0,0,0,0)? I think that's what the original code thinks, but it returns true when it is NOT empty.

libfreerdp/codec/test/TestFreeRDPRegion.c
realjiangms: OK. I would also fix the 'is_empty' to only check for top=bottom and left=right.

libfreerdp/codec/test/TestFreeRDPRegion.c
realjiangms: @hardening 
Regarding your question of 'a test that would demonstrate the need for the changes in region16_intersect_rect()', I reuse the above 'test_norbert_case' to demonstrate the need for the fix in region16_intersect_rect. Without the fix, the result would be 
{0, 0, 1920, 1078}. The original incorrect 'expected_inter_extents' exactly shows the bug

hardening:
I did some test and norbert_test_case, the unitary test was wrong and {2, 0, 1920, 1078} is really the expended value (BTW if someone was nice enough to add an ascii art describing the test...);

So I would +1 and even merge it after:
* just top == bottom or  left == right in rectangle_is_empty();
* drop the invalid rectangle in the unitary test

hardening:
@freerdp-bot test

realjiangms:
@hardening Thanks.
I don't understand the norbert_test_case at first glance, I think it may be just taken from some real error case when the original developer met his bug.

Let me try to understand the rectangles in inRectangles

hardening:
@realjiangms yes it's a real case we hit with @nfedera with a full screenvideo, hence the name.

hardening:
@freerdp-bot test
please, if you have nothing to do

realjiangms:
@hardening I'm writing the ascii art for the norbert_test_case so that later we can understand it quickly. Is it suitable to be included in this pull request?

hardening:
@freerdp-bot test

hardening:
Ascii art => @realjiangms yeah sure

realjiangms:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

PR #2487, opened 2015-03-23
Fix server side NLA

mfleisz:
This should correctly fix the problem described in #2485


mfleisz:
@awakecoding Please have a look if that's the right way to fix the problem described in #2485 - Thanks!

awakecoding:
@mfleisz did you test both client and server-side NLA with this?

mfleisz:
@awakecoding Yes - works nicely with this fix :+1:

Issue #2338, opened 2015-01-27
async-input appears to be on by default.  Help says off by default


takenit2far:
In master, turning async-input off with '-' fixes not being able to control the screen once a video is playing in a web browser. 

Help states its off by default,  but it would be appear that it is turned on.

bmiklautz:
@takenit2far you are using xfreerdp?

bmiklautz:
Just checked and it's off by default.

takenit2far:
I know I heard that,   but if I turn it on it makes no difference..  Turning it off improves the situation dramatically.   Could anything else be flipping its behaviour?

realjiangms:
@takenit2far 
I think you are using xfreerdp with an rdp server which supports rfx codec. 
I got same issue when I use a thin client (with ATOM 1.6GHZ cpu) and connect to an rdp server with Intel E31220 3.1Ghz, the server side is the shadow server in FreeRDP. I set the fps to 30 to get more stable output.

The default is sync-input, the issue exists and it only response to my keyboard/mouse every 5 seconds. According to the X11 client code, the reason should be poor CPU speed in thin client side. It decode rfx message much more slower than server side encoding the rfx message. Therefore when it returns from decoding current frame, the PDU of next frame has already arrive, and the upper layer function(transport_check_fds() in libfreerdp/core/transport.c) will continue processing next PDU since it exists. Then we run into an endless loop and busy displaying the video. In very rare case we could get a respite (such as, the full black frame in video) and transport_check_fds returns, then the keyboard/mouse event get a chance to be handled.

I would expect that the +async-input, but no, it makes it worse. I added printf in event handler subroutine and it shows that once I press a key, the event is get handled and sent to the server. And in network monitor, when I move mouse and press keyboard, it shows about 100KB/s data is transferred to the server. So I think it works in client side but there may be some race condition and we may violate some rules in protocol, so the server side never handle the event PDU in this case.

@takenit2far 
Finally I come up with a workaround. In libfreerdp/core/transport.c, comment out the following:
```
        /**
         * Loop through and read all available PDUs.  Since multiple
         * PDUs can exist, it's important to deliver them all before
         * returning.  Otherwise we run the risk of having a thread
         * wait for a socket to get signaled that data is available
         * (which may never happen).
         */
        for (;;)
       {
                /**
                 * Note: transport_read_pdu tries to read one PDU from
                 * the transport layer.
[...]

PR #2490, opened 2015-03-23
fix async input.

realjiangms:
The async-input is not really async. We create a thread to peek input event from X, but the event is pushed into FREERDP_INPUT_MESSAGE_QUEUE as message and FREERDP_INPUT_MESSAGE_QUEUE is processed in main client thread. Therefore the input event is still processed in main client thread actually.
The fix make the message also be processed in input thread.

The original issue is discussed in #2338. When we have heavy image update, the input event will hardly  get a chance to be handled. In this case we may want to use async input but it does't actually work in this case.

But I'm not sure if I understand the purpose of async input correctly.

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -1263,7 +1263,6 @@ void* xf_input_thread(void* arg)
 	DWORD nCount;
 	HANDLE events[2];
 	XEvent xevent;
-	wMessage msg;
 	wMessageQueue* queue;
 	int pending_status = 1;
 	int process_status = 1;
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@realjiangms there are different "types" of async (input, update, channel, transport). What did you test your pull request with?

realjiangms:
@bmiklautz I tested with +async-input, without any other 'async' inputs. The fix proposed is for +async-input

Actually I agree that this fix is not really necessary. The issue reported in #2338 could be resolved by using +async-transport or +async-update instead of +async-input. (+async-update may finally panic with out-of-memory as we have unlimited update message queue) 

I raise this pull-request just for discussion on the 'meaning' of async-input, and the difference in detail behavior for 'async-input' between (android/X11) and (Windows/Mac).

These days I kept trying +async-transport to test if it could workaround issue in #2338, unfortunately I found another race condition which could cause thread hung.
After several day's investigation, I've just found the root cause and I'll reorganize wording at this weekend.to report it

bmiklautz:
@realjiangms thanks for the information.

realjiangms:
I would want to close this pull request as it's not really necessary. The issue could be resolved by using +async-transport instead of +async-input. We don't even need +async-input in x11 client.

The hung issue mentioned in previous note is described in #2519

realjiangms:
The hung issue is completely fixed by #3364

PR #2491, opened 2015-03-23
NULL dereference fixes

akallabeth:
Fixed a few missing checks and code cleanup.

### DIFF
--- a/winpr/libwinpr/synch/timer.c
+++ b/winpr/libwinpr/synch/timer.c
@@ -49,7 +49,7 @@ static BOOL TimerIsHandled(HANDLE handle)
 {
 	WINPR_TIMER* pTimer = (WINPR_TIMER*) handle;
 
-	if (!pTimer || pTimer->Type != HANDLE_TYPE_TIMER)
+	if (!pTimer || (pTimer->Type != HANDLE_TYPE_TIMER) || (pTimer == INVALID_HANDLE_VALUE))
 	{
 		SetLastError(ERROR_INVALID_HANDLE);
 		return FALSE;
--- a/winpr/libwinpr/synch/timer.c
+++ b/winpr/libwinpr/synch/timer.c
@@ -49,7 +49,7 @@ static BOOL TimerIsHandled(HANDLE handle)
 {
 	WINPR_TIMER* pTimer = (WINPR_TIMER*) handle;
 
-	if (!pTimer || (pTimer->Type != HANDLE_TYPE_TIMER) || (pTimer == INVALID_HANDLE_VALUE))
+	if (!pTimer || (pTimer->Type != HANDLE_TYPE_TIMER))
 	{
 		SetLastError(ERROR_INVALID_HANDLE);
 		return FALSE;
--- a/winpr/libwinpr/utils/wlog/wlog.c
+++ b/winpr/libwinpr/utils/wlog/wlog.c
@@ -381,39 +381,36 @@ int WLog_ParseFilters()
 	if (!env)
 		return -1;
 
+	nSize = GetEnvironmentVariableA("WLOG_FILTER", env, nSize);
+	count = 1;
+	p = env;
+
+	while ((p = strchr(p, ',')) != NULL)
 	{
-		nSize = GetEnvironmentVariableA("WLOG_FILTER", env, nSize);
-		count = 1;
-		p = env;
-
-		while ((p = strchr(p, ',')) != NULL)
-		{
-			count++;
-			p++;
-		}
-
-		g_FilterCount = count;
-		p = env;
+		count++;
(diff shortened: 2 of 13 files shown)


mfleisz:
+1

bmiklautz:
+1 for the reallocs
~1 for the invalid handles - a lot handle specific functions return NULL and don't really use INVALID_HANDLE_VALUE. Using INVALID_HANDLE_VALUE in those cases might cause problems since it's theoratically valid memory that can be returned. See also http://blogs.msdn.com/b/oldnewthing/archive/2004/03/02/82639.aspx
+1 for the formats check. might not be necessary but also doesn't hurt ;)

mfleisz:
Yes we should only check INVALDI_HANDLE_VALUE for the APIs that use that return value (i.e. the File API) and skip the check on the others (i.e. Event/Mutex API)

akallabeth:
@bmiklautz @mfleisz Removed the ```INVALID_HANDLE_VALUE``` check from handle types not using it.

mfleisz:
@akallabeth Event/Mutex/Semaphore functions don't use INVALID_HANDLE_VALUE either.

akallabeth:
@mfleisz ah, correct. Mixed ```INVALID_HANDLE_VALUE``` with ```ERROR_INVALID_HANDLE```

winpr/libwinpr/synch/timer.c
```diff
@@ -49,7 +49,7 @@ static BOOL TimerIsHandled(HANDLE handle)
 {
 	WINPR_TIMER* pTimer = (WINPR_TIMER*) handle;
 
-	if (!pTimer || pTimer->Type != HANDLE_TYPE_TIMER)
+	if (!pTimer || (pTimer->Type != HANDLE_TYPE_TIMER) || (pTimer == INVALID_HANDLE_VALUE))
 	{
```
nfedera: Don't think INVALID_HANDLE_VALUE is possible here, please check again

winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -353,7 +353,8 @@ int WLog_ParseFilter(wLogFilter* filter, LPCSTR name)
 
 	while ((p = strchr(p, '.')) != NULL)
 	{
-		filter->Names[count++] = p + 1;
+		if ((p + 1) && (count < filter->NameCount))
+			filter->Names[count++] = p + 1;
```
nfedera: I guess (p+1) should be *(p+1) or p[1]

winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -380,34 +381,39 @@ int WLog_ParseFilters()
...
-	while ((p = strchr(p, ',')) != NULL)
-	{
-		strs[count++] = p + 1;
-		*p = '\0';
-		p++;
+		strs[count++] = p;
+
+		while ((p = strchr(p, ',')) != NULL)
+		{
+			if ((p + 1) && (count < g_FilterCount))
```
nfedera: again: p + 1 is always true, probably that should be if (p[1] ...)  or  if (*(p+1) ...)

winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -502,7 +508,8 @@ int WLog_ParseName(wLog* log, LPCSTR name)
 
 	while ((p = strchr(p, '.')) != NULL)
 	{
-		log->Names[count++] = p + 1;
+		if ((p + 1) && (count < log->NameCount))
```
nfedera: again: p + 1 is always true, see above

nfedera:
+1 after fixing my remarks. thanks !

akallabeth:
@nfedera Fixed your remarks, the ```p+1``` was actually ```p[0]``` as in check for end of string to abort.

akallabeth:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -380,34 +381,39 @@ int WLog_ParseFilters()
...
 
-	if (!strs)
-	{
-		free(env);
-		return -1;
+		g_FilterCount = count;
+		p = env;
 	}
+	{
+		count = 0;
```
bmiklautz: Why are there to extra scopes ({}) used there?

bmiklautz:
If the unnecessary scopes are removed +1.

bmiklautz:
@freerdp-bot test.

bmiklautz:
+1

bmiklautz:
I've created a re-based version as #2509

PR #2493, opened 2015-03-25
tls: revert checking SSL_ERROR_SYSCALL which may cause deadlock.

llyzs:
The last commit may cause deadlock when working with mstsc on the server-side and must be reverted. The real cause of SSL_ERROR_SYSCALL has been identified, however. The real solution is on the way and will be submitted as another PR.

### DIFF
--- a/libfreerdp/crypto/tls.c
+++ b/libfreerdp/crypto/tls.c
@@ -58,7 +58,6 @@ long bio_rdp_tls_callback(BIO* bio, int mode, const char* argp, int argi, long a
 
 static int bio_rdp_tls_write(BIO* bio, const char* buf, int size)
 {
-	int error;
 	int status;
 	BIO_RDP_TLS* tls = (BIO_RDP_TLS*) bio->ptr;
 
@@ -96,7 +95,7 @@ static int bio_rdp_tls_write(BIO* bio, const char* buf, int size)
 				break;
 
 			case SSL_ERROR_SYSCALL:
-				BIO_set_flags(bio, (BIO_FLAGS_WRITE | BIO_FLAGS_SHOULD_RETRY));
+				BIO_clear_flags(bio, BIO_FLAGS_SHOULD_RETRY);
 				break;
 
 			case SSL_ERROR_SSL:
@@ -110,7 +109,6 @@ static int bio_rdp_tls_write(BIO* bio, const char* buf, int size)
 
 static int bio_rdp_tls_read(BIO* bio, char* buf, int size)
 {
-	int error;
 	int status;
 	BIO_RDP_TLS* tls = (BIO_RDP_TLS*) bio->ptr;
 
@@ -161,7 +159,7 @@ static int bio_rdp_tls_read(BIO* bio, char* buf, int size)
 				break;
 
 			case SSL_ERROR_SYSCALL:
-				BIO_set_flags(bio, (BIO_FLAGS_READ | BIO_FLAGS_SHOULD_RETRY));
+				BIO_clear_flags(bio, BIO_FLAGS_SHOULD_RETRY);
 				break;
 		}
 	}


hardening:
I'd tend to +1, but the proposed code is not a revert. 
I'm very interested to know the root cause, which we're tracking from a long time !

llyzs:
No it's not supposed to be a direct git revert, because I have to undo an earlier change related to this altogether as well.

llyzs:
I will create an issue for further discussion of this matter, it is not a simple one.

hardening:
@llyzs given that the root cause is a race in the SSL BIO, the SSL_ERROR_SYSCALL error case can be treated like error, can't it ?

llyzs:
Yes this commit treats all SSL_ERROR_SYSCALL as error as it should be; the root cause should be done in a separated PR.

mfleisz:
+1 for this pull and opening a new issue for the concurrency problem

llyzs:
Discussion in #2497

realjiangms:
I just updated the latest code and met following SIGSEGV with +async-update. Not sure whether or not it's related to this check-in:

#0  0x00007ffff3dd7910 in BIO_test_flags () from /lib/x86_64-linux-gnu/libcrypto.so.1.0.0
#1  0x00007ffff5e1a7fe in transport_write () from /usr/local/lib/libfreerdp.so.1.2
#2  0x00007ffff5e1377e in rdp_send_data_pdu () from /usr/local/lib/libfreerdp.so.1.2
#3  0x00007ffff5e1b184 in update_send_frame_acknowledge () from /usr/local/lib/libfreerdp.so.1.2
#4  0x00007ffff5e22356 in update_message_queue_process_message () from /usr/local/lib/libfreerdp.so.1.2
#5  0x00007ffff5e2276d in update_message_proxy_thread () from /usr/local/lib/libfreerdp.so.1.2
#6  0x00007ffff7658051 in thread_launcher () from /usr/local/lib/libwinpr.so.1.1
#7  0x00007ffff564ee9a in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#8  0x00007ffff7324cbd in clone () from /lib/x86_64-linux-gnu/libc.so.6

(gdb) f 1
#1  0x00007ffff5e1a7fe in transport_write () from /usr/local/lib/libfreerdp.so.1.2
(gdb) disassemble $pc
Dump of assembler code for function transport_write:
...
   0x00007ffff5e1a7d8 <+344>:   mov    $0xffffffff,%r14d
   0x00007ffff5e1a7de <+350>:   movl   $0x4,(%rbx)
   0x00007ffff5e1a7e4 <+356>:   jmp    0x7ffff5e1a7b8 <transport_write+312>
   0x00007ffff5e1a7e6 <+358>:   nopw   %cs:0x0(%rax,%rax,1)
   0x00007ffff5e1a7f0 <+368>:   mov    0x8(%rbx),%rdi
   0x00007ffff5e1a7f4 <+372>:   mov    $0x8,%esi                     
   0x00007ffff5e1a7f9 <+377>:   callq  0x7ffff5dad0b0 <BIO_test_flags@plt>
=> 0x00007ffff5e1a7fe <+382>:   test   %eax,%eax
   0x00007ffff5e1a800 <+384>:   je     0x7ffff5e1a786 <transport_write+262>
   0x00007ffff5e1a802 <+386>:   mov    0x80(%rbx),%r15d
   0x00007ffff5e1a809 <+393>:   test   %r15d,%r15d
...

Note that it's actually in calling BIO_test_flags. Code is:
if (!BIO_should_retry(transport->frontBio))
And is actually:
BIO_test_flags(a, BIO_FLAGS_SHOULD_RETRY)

0x00007ffff5e1a7f4 <+372>: mov    $0x8,%esi ;should be BIO_FLAGS_SHOULD_RETRY.
0x00007ffff5e1a7f0 <+368>:   mov    0x8(%rbx),%rdi ;should be transport->frontBio.
But transport seems to be all empty:
(gdb) x /16x $rbx
0x7fffec072a60: 0x00000000      0x00000000      0x00000000      0x00000000
0x7fffec072a70: 0x00000000      0x00000000      0x00000000      0x00000000
0x7fffec072a80: 0x00000000      0x00000000      0x006051f0      0x00000000
0x7fffec072a90: 0x00000000      0x00000000      0x0060ab80      0x00000000

realjiangms:
Hmm, the above SIGSEGV is not related to this check-in. It's because our following code in freerdp_disconnect

BOOL freerdp_disconnect(freerdp* instance)
{
    rdpRdp* rdp;

    rdp = instance->context->rdp;

    rdp_client_disconnect(rdp);  ============> Disconnect transport here
    update_post_disconnect(instance->update);
==============> But update_message_proxy_free is invoked in this line. We should move update_message_proxy_free before rdp_client_disconnect.

The update_message_proxy_new is called after connect, so update_message_proxy_free  should be called before connect. Resource initialization and release should usually be in reverse sequence.

realjiangms:
However, the fix in this pull request increase the disconnect possibility. We seems have big possibility to have SSL_ERROR_SYSCALL and now we always disconnect in SSL_ERROR_SYSCALL .

realjiangms:
After reading the conversation in #2482, I understand where the changes in ssl layer comes from these days. 
I would suggest that always retry at  SSL_ERROR_SYSCALL is not correct. When I use shadow-server with mstsc, I can sometimes hit server side hang with following steps. Not sure if it's exactly what @llyzs described in first note:

1. Connect with cable
2. Connect to rdp server with mstsc
3. Play a video so that there's frequent traffic in network
4. Pull out the cable rudely
5. Then sometimes server side falls in endless retry loop.

hardening:
@realjiangms theoritically we can't hit a SSL_ERROR_SYSCALL, and if we fall in that case:
* surely it's something we have done wrong;
* using errno to take a decision is useless as we have our own BIOs (and so errno can't be used);

realjiangms:
@hardening. I understand your point.
So for the server side hang I mentioned in previous note, we may don't handle some network error correctly. I may try to reproduce it later with a debug build. (I now use release build as I'm really 'using' the shadow-server and the rfx encoder at debug build is slow)

llyzs:
@realjiangms If you hit SSL_ERROR_SYSCALL, there is definitely something wrong with your own code, but not in FreeRDP or OpenSSL. I was not convinced until I built OpenSSL debug symbols and catch this error in live debugging.

The problem I was having is thread synchronization issue, the BIO was accessed from a different thread context, via the FreeRDP main API and instance, such as SurfaceBits or PointerNew etc, which messed the state of the BIO. Such problem is very hard to detect, but it did turned our that it was downstream problem but NOT FreeRDP or OpenSSL problem.

And @hardening was right all the way - errno or WSAGetLastError has nothing to do with SSL_ERROR_SYSCALL. And I guess the only way to find our the cause of your problem is to read the OpenSSL source and do live debugging like I did.

realjiangms:
Thanks for your information. I'm not sure whether or not I really hit SSL_ERROR_SYSCALL. I just met the problem that the server side kept retrying endlessly, so I guess we always have SSL_ERROR_SYSCALL in that case. I need to reproduce my problem to figure out the detail later.
  


     Vic Lee <notifications@github.com> 于 2015年3月30日, 星期一, 下午 12:53 写道：
   

 @realjiangms If you hit SSL_ERROR_SYSCALL, there is definitely something wrong with your own code, but not in FreeRDP or OpenSSL. I was not convinced until I built OpenSSL debug symbols and catch this error in live debugging.The problem I was having is thread synchronization issue, the BIO was accessed from a different thread context, via the FreeRDP main API and instance, such as SurfaceBits or PointerNew etc, which messed the state of the BIO. Such problem is very hard to detect, but it did turned our that it was downstream problem but NOT FreeRDP or OpenSSL problem.And @hardening was right all the way - errno or WSAGetLastError has nothing to do with SSL_ERROR_SYSCALL. And I guess the only way to find our the cause of your problem is to read the OpenSSL source and do live debugging like I did.—

PR #2494, opened 2015-03-25
Fixed OpenH264 linking

akallabeth:
Added the dll link library name to hints


mfleisz:
+1

bmiklautz:
What this definitely changes is that now always the dynamic version is used if found which is a change in behavior. Not sure if this was intended.

bmiklautz:
Since openh264 can't be shipped anyway +1

PR #2498, opened 2015-03-26
Fix memory leak in TSMF video playback

TaGUNil:
Frame data buffers are not freed after displaying, creating a giant memory leak (2+ GB in 40 seconds). Maybe it is FrameEvent callback who should free it, but then what happens if it is NULL?


bmiklautz:
@freerdp-bot test.

hardening:
+1

Issue #2499, opened 2015-03-26
Hangs for 90 seconds after "Do you trust the above certificate"


KenHolloway:
If you answer "N" (either the user enters "N", or input is redirected in a script) to the "Do you trust the above certificate" prompt, FreeRDP hangs for 90 seconds.

Here is the end of the output from
```
strace -ff -e network   xfreerdp 02-VHost4.rdp /bpp:24 /size:1024x768 -menu-anims -wallpaper +clipboard /sound /drive:Home,/home/KHolloway/ /p:Password < /dev/null
```

```
[10:17:21:853] [5026:5027] [INFO][com.freerdp.client.x11] - Certificate details:
[10:17:21:853] [5026:5027] [INFO][com.freerdp.client.x11] -     Subject: CN = VHOST4.inovate.inovate.com
[10:17:21:853] [5026:5027] [INFO][com.freerdp.client.x11] -     Issuer: DC = com, DC = inovate, DC = inovate, CN = xxxxx xxxxx Root Certificate
[10:17:21:853] [5026:5027] [INFO][com.freerdp.client.x11] -     Thumbprint: d9:48:2a:6e:95:58:1b:0f:8f:23:64:9a:3f:fd:3c:7d:c0:c9:bb:8f
[10:17:21:853] [5026:5027] [INFO][com.freerdp.client.x11] - The above X.509 certificate could not be verified, possibly because you do not have the CA certificate in your certificate store, or the certificate has expired. Please look at the documentation on how to create local certificate store for a private CA.
[10:17:21:853] [5026:5027] [INFO][com.freerdp.client.x11] - Do you trust the above certificate? (Y/N) 
[10:17:21:853] [5026:5027] [INFO][com.freerdp.client.x11] - Error: Could not read answer from stdin.
[10:17:21:853] [5026:5027] [INFO][com.freerdp.client.x11] - 
[10:17:21:853] [5026:5027] [ERROR][com.freerdp.crypto] - certificate not trusted, aborting.
[pid  5027] sendto(9, "\25\3\1\0 \17\316\314\210\255:\203\356\341N\27\215\3\177\234p\3/C\277Cg9\6\204\306\210"..., 37, MSG_NOSIGNAL, NULL, 0) = 37
[10:17:21:853] [5026:5027] [ERROR][com.freerdp.core] - freerdp_set_last_error 0x2000B
[...]

PR #2503, opened 2015-03-27
rename and update tls_disconnect

bmiklautz:
tls_disconnect shut down the ssl stream but didn't inform
the BIO(s) about this therefore could happen that a second shut down
was initiated (e.g. in bio_rdp_tls_free) causing rather long delays.

After removing the shut down from tls_disconnect the only thing the
function does is to prepare/set an alert therefore it was renamed to
tls_set_alert.


bmiklautz:
Fixes #2499

bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

PR #2506, opened 2015-03-30
crypto: support buggy Microsoft iOS clients

nfedera:
Microsoft iOS Remote Desktop Clients eventually send NULL-terminated hostnames in SNI which is not allowed in the OpenSSL implementation.

Since we're not using SNI this commit adds an OpenSSL TLS extension debug callback which modifies the SSL context in a way preventing it from parsing this extension.


bmiklautz:
@nfedera does this depend on a specific openssl version?

PR #2507, opened 2015-03-30
nego: fixed X.224 Connection Request PDU parsing

nfedera:
The X.224 Connection Request PDU might contain an optional cookie or routing token before the optional RDP Negotiation Request (rdpNegReq).
If present, both of these fields must be terminated by a 0x0D0A two-byte sequence.
It seems that until now FreeRDP has incorrectly assumed that a token or cookie must always be present.
If that was not the case, FreeRDP was searching for 0x0D0A until it arrived at the end of the stream which prevented the remaining data (RDP Negotiation Request, RDP Correlation Info) from being parsed.

### DIFF
--- a/libfreerdp/core/nego.c
+++ b/libfreerdp/core/nego.c
@@ -626,6 +626,95 @@ int nego_recv(rdpTransport* transport, wStream* s, void* extra)
 	return 0;
 }
 
+
+/**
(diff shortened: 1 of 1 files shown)


bmiklautz:
+1
Nice one!

hardening:
+1 too, nice one !

PR #2508, opened 2015-03-30
crypto: support MSFT iOS clients with faulty SNI

nfedera:
Microsoft iOS Remote Desktop Clients eventually send NULL-terminated hostnames in SNI which is not allowed in the OpenSSL implementation.
Since we're not using SNI this commit adds an OpenSSL TLS extension debug callback which modifies the SSL context in a way preventing it from parsing this extension.

### DIFF
--- a/libfreerdp/crypto/tls.c
+++ b/libfreerdp/crypto/tls.c
@@ -759,6 +759,19 @@ int tls_connect(rdpTls* tls, BIO* underlying)
 	return tls_do_handshake(tls, TRUE);
 }
 
+#ifndef OPENSSL_NO_TLSEXT
+static void tls_openssl_tlsext_debug_callback(SSL *s, int client_server,
+	int type, unsigned char *data, int len, void *arg)
+{
+	/* see code comment in tls_accept() below */
+
+	if (type == TLSEXT_TYPE_server_name) {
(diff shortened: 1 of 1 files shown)


bmiklautz:
+1

nfedera:
Remark: Even if it is real bad practice to terminate strings where a length indication is used I'm not entirely convinced that this is against the specs. [RFC 6066 Section 3](https://tools.ietf.org/html/rfc6066#section-3) defines HostName as *opaque* and [RFC 5246 Section 4.2](https://tools.ietf.org/html/rfc5246#section-4.2) defines opaque as "Single-byte entities containing uninterpreted data". In addition [RFC 6066 Section 3](https://tools.ietf.org/html/rfc6066#section-3) even says that *TLS MAY treat provided server names as opaque data and pass the names and types to the application.* AFAIK latter option is not directly exposed by OpenSSL.
However, if not choosing the MAY then according to RFC 6066 the following rules apply for "HostName":
- contains the fully qualified DNS hostname of the server,  as understood by the client.
- case-insensitive.
- represented as a byte string using ASCII encoding without a trailing dot.
- literal IPv4 and IPv6 addresses are not permitted

Latter is not enforced by OpenSSL (might be considered as a bug).
However, if someone can provide some real good RFC-backed proof as to why this string MUST NOT be null-terminated I'd be glad.

Issue #2399, opened 2015-02-14
xfreerdrp Crashes with segfault in find_keyboard_layout_in_xorg_rules


octoquad:
Hi,

I completed compiling just the freerdrp package on Ubuntu Gnome 15.04, using this guide: https://github.com/FreeRDP/Remmina/wiki/Compile-on-Ubuntu-14.04-and-14.10. The build was successful only after adding libgstreamer-plugins-base1.0-dev as a dependency. I can add this in the page, just let me know.

This particular problem has been around since 2011 and has been reported on Launchpad: https://bugs.launchpad.net/ubuntu/+source/remmina/+bug/753520. Both GDB and Valgrind logs are attached to the bug report. In summary xfreerdp fails to load if the keyboard XKBLAYOUT is set to "za" or "mao" in /etc/default/keyboard, but setting it to "us" seems to make it work again. 

This particular problem is fairly trivial to reproduce, but if you need more information let me know. Just using xfreerdp built from source produces:

```text
(gdb) run +clipboard /sound:rate:44100,channel:2 /v:<ip/host> /u:<user>
Starting program: /usr/local/bin/xfreerdp +clipboard /sound:rate:44100,channel:2 /v:<ip/host> /u:<user>
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7fffecbf1700 (LWP 27493)]

Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffecbf1700 (LWP 27493)]
0x00007ffff5d13232 in find_keyboard_layout_in_xorg_rules (layout=0x7fffe801ad70 "za", variant=0x7fffe801aff0 "") at /home/bruce/dev/desktop/FreeRDP/libfreerdp/locale/xkb_layout_ids.c:913
913				for (j = 0; xkbLayouts[i].variants[j].variant != NULL && strlen(xkbLayouts[i].variants[j].variant) > 0; j++)
(gdb) bt
#0  0x00007ffff5d13232 in find_keyboard_layout_in_xorg_rules (layout=0x7fffe801ad70 "za", variant=0x7fffe801aff0 "") at /home/bruce/dev/desktop/FreeRDP/libfreerdp/locale/xkb_layout_ids.c:913
[...]

PR #2511, opened 2015-03-30
Fix segfault when the xkb layout has no variants

hardening:
This is a fix for #2399: when there's no variants we should not try to scan them.
I have set the RDP US keyboard for the South African layout, if someone has a better layout...

Note: we should probably set something that is not zero for other layouts


bmiklautz:
+1

Issue #2353, opened 2015-02-02
Get rid of "exit" in libfreerdp


bmiklautz:
freerdp_peer_initialize in libfreerdp/core/peer.c uses exit. This shouldn't happen in a library but in this situation it is  unfortunately currently necessary since not all functions along the return path check the values.

Also see PR #2352

akallabeth:
```exit``` only called from client / server.

PR #2512, opened 2015-03-30
Kill more exit() occurences

hardening:
In general we should not have any exit() call in libfreerdp.

Should help for https://github.com/FreeRDP/FreeRDP/issues/2353


bmiklautz:
@hardening I absolutely agree on having exit in a library but I thin just return in peer.c doesn't solve the issue and might lead to undefined behavior.

bmiklautz:
@hardening maybe you can split up this PR. The first part in searchman.c really makes a lot of sense and we should get it in.

hardening:
My last commit fixes the problem with big keys, so it allows to drop the last exit() occurrence.

channels/urbdrc/client/searchman.c
```diff
@@ -219,7 +220,15 @@ USB_SEARCHMAN* searchman_new(void * urbdrc, UINT32 UsbDevice)
...
+		goto out_error_event;
 	sem_init(&searchman->sem_term, 0, 0);
 	
 	return searchman;
+
+out_error_event:
+	pthread_mutex_destroy(&searchman->mutex);
+out_error_mutex:
+	free(searchman);
+	return NULL;
```
bmiklautz: The problem here still might be that the return isn't checked outside after the call.
Maybe you could add the corresponding checks as well.

libfreerdp/core/gcc.c
```diff
@@ -512,12 +512,13 @@ BOOL gcc_read_server_data_blocks(wStream* s, rdpMcs* mcs, int length)
...
 }
 
-void gcc_write_server_data_blocks(wStream* s, rdpMcs* mcs)
+BOOL gcc_write_server_data_blocks(wStream* s, rdpMcs* mcs)
 {
-	gcc_write_server_core_data(s, mcs); /* serverCoreData */
-	gcc_write_server_network_data(s, mcs); /* serverNetworkData */
-	gcc_write_server_security_data(s, mcs); /* serverSecurityData */
-	gcc_write_server_message_channel_data(s, mcs); /* serverMessageChannelData */
+	return
```
bmiklautz: Something messed up here?

bmiklautz:
@hardening regarding key size. I might be mistaken but in my memory freerdp_peer_initialize is called by the clients (Initialize) as well. Maybe we should also add a check there now.

libfreerdp/core/mcs.c
```diff
@@ -716,6 +722,12 @@ BOOL mcs_send_connect_response(rdpMcs* mcs)
 	Stream_Free(server_data, TRUE);
 
 	return (status < 0) ? FALSE : TRUE;
+
+error_stream_s:
+	Stream_Free(gcc_CCrsp, TRUE);
+error_data_blocks:
+	Stream_Free(s, TRUE);
```
bmiklautz: That should be Stream_Free(server_data, TRUE); 
s isn't allocated at that time.

bmiklautz:
Except the problem with Stream_Free (s instead of server_data) big +1.
@hardening you made my day!

hardening:
@bmiklautz should be fixed now

libfreerdp/core/gcc.c
```diff
@@ -1735,14 +1749,18 @@ BOOL gcc_read_server_message_channel_data(wStream* s, rdpMcs* mcs)
...
 }
 
-void gcc_write_server_message_channel_data(wStream* s, rdpMcs* mcs)
+BOOL gcc_write_server_message_channel_data(wStream* s, rdpMcs* mcs)
 {
 	if (mcs->messageChannelId == 0)
-		return;
+		return TRUE;
+
+	if (!Stream_EnsureRemainingCapacity(s, 6 + 4))
```
nfedera: Shouldn't this be 6 + 2 ??

nfedera:
+1 (except the stream length remark)

libfreerdp/core/gcc.c
hardening: Yeah you're right. It doesn't hurt but I will fix it.

PR #2513, opened 2015-04-01
License and security ret values

hardening:
A PR that checks return values in core/[license.c|security.c]

### DIFF
--- a/libfreerdp/core/fastpath.c
+++ b/libfreerdp/core/fastpath.c
@@ -767,6 +767,7 @@ BOOL fastpath_send_multiple_input_pdu(rdpFastPath* fastpath, wStream* s, int iNu
 	rdpRdp* rdp;
 	UINT16 length;
 	BYTE eventHeader;
+	BOOL status;
 
 	/*
 	 *  A maximum of 15 events are allowed per request
@@ -817,7 +818,8 @@ BOOL fastpath_send_multiple_input_pdu(rdpFastPath* fastpath, wStream* s, int iNu
 			Stream_Write_UINT8(s, 0x1); /* TSFIPS_VERSION 1*/
 			Stream_Write_UINT8(s, pad); /* padding */
 
-			security_hmac_signature(fpInputEvents, fpInputEvents_length, Stream_Pointer(s), rdp);
+			if (!security_hmac_signature(fpInputEvents, fpInputEvents_length, Stream_Pointer(s), rdp))
+				return FALSE;
 
 			if (pad)
 				memset(fpInputEvents + fpInputEvents_length, 0, pad);
@@ -829,9 +831,12 @@ BOOL fastpath_send_multiple_input_pdu(rdpFastPath* fastpath, wStream* s, int iNu
 		else
 		{
 			if (rdp->sec_flags & SEC_SECURE_CHECKSUM)
-				security_salted_mac_signature(rdp, fpInputEvents, fpInputEvents_length, TRUE, Stream_Pointer(s));
+				status = security_salted_mac_signature(rdp, fpInputEvents, fpInputEvents_length, TRUE, Stream_Pointer(s));
 			else
-				security_mac_signature(rdp, fpInputEvents, fpInputEvents_length, Stream_Pointer(s));
+				status = security_mac_signature(rdp, fpInputEvents, fpInputEvents_length, Stream_Pointer(s));
+
+			if (!status)
+				return FALSE;
 
 			security_encrypt(fpInputEvents, fpInputEvents_length, rdp);
 		}
@@ -1021,16 +1026,19 @@ BOOL fastpath_send_update_pdu(rdpFastPath* fastpath, BYTE updateCode, wStream* s
 
 			if (rdp->settings->EncryptionMethods == ENCRYPTION_METHOD_FIPS)
 			{
-				security_hmac_signature(data, dataSize - pad, pSignature, rdp);
+				if (!security_hmac_signature(data, dataSize - pad, pSignature, rdp))
+					return FALSE;
 				security_fips_encrypt(data, dataSize, rdp);
 			}
 			else
 			{
 				if (rdp->sec_flags & SEC_SECURE_CHECKSUM)
-					security_salted_mac_signature(rdp, data, dataSize, TRUE, pSignature);
+					status = security_salted_mac_signature(rdp, data, dataSize, TRUE, pSignature);
 				else
-					security_mac_signature(rdp, data, dataSize, pSignature);
+					status = security_mac_signature(rdp, data, dataSize, pSignature);
 
+				if (!status)
+					return FALSE;
 				security_encrypt(data, dataSize, rdp);
 			}
 		}
--- a/libfreerdp/core/fastpath.c
+++ b/libfreerdp/core/fastpath.c
@@ -767,7 +767,6 @@ BOOL fastpath_send_multiple_input_pdu(rdpFastPath* fastpath, wStream* s, int iNu
 	rdpRdp* rdp;
 	UINT16 length;
 	BYTE eventHeader;
-	BOOL status;
 
 	/*
 	 *  A maximum of 15 events are allowed per request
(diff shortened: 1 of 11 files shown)


libfreerdp/crypto/crypto.c
```diff
@@ -145,14 +146,14 @@ CryptoHmac crypto_hmac_new(void)
 	return hmac;
 }
 
-void crypto_hmac_sha1_init(CryptoHmac hmac, const BYTE* data, UINT32 length)
+BOOL crypto_hmac_sha1_init(CryptoHmac hmac, const BYTE* data, UINT32 length)
 {
-	HMAC_Init_ex(&hmac->hmac_ctx, data, length, EVP_sha1(), NULL);
+	return HMAC_Init_ex(&hmac->hmac_ctx, data, length, EVP_sha1(), NULL) == 1;
```
bmiklautz: If I'm not mistaken HMAC_Init_ex was void in older openssl versions.

HenryJacques: Correct, in 2008.
See https://github.com/openssl/openssl/commit/87d52468aa600e02326e13f01331e1f3b8602ed0#diff-0a42cda89821d0691aeb6a4a5d124894

libfreerdp/core/license.c
```diff
@@ -336,15 +340,23 @@ void license_generate_randoms(rdpLicense* license)
...
+	BOOL ret;
+
+	if (
+		/* MasterSecret */
+		!security_master_secret(license->PremasterSecret, license->ClientRandom,
+								license->ServerRandom, license->MasterSecret) ||
+		/* SessionKeyBlob */
+		!security_session_key_blob(license->MasterSecret, license->ClientRandom,
+							  license->ServerRandom, license->SessionKeyBlob))
+	{
```
bmiklautz: Although absolutely correct the formatting looks rather odd for me.

libfreerdp/core/rdp.c
```diff
@@ -429,12 +429,13 @@ void rdp_write_header(rdpRdp* rdp, wStream* s, UINT16 length, UINT16 channelId)
 	Stream_Write_UINT16_BE(s, length); /* userData (OCTET_STRING) */
 }
 
-static UINT32 rdp_security_stream_out(rdpRdp* rdp, wStream* s, int length, UINT32 sec_flags)
+static BOOL rdp_security_stream_out(rdpRdp* rdp, wStream* s, int length, UINT32 sec_flags, UINT32 *pad)
 {
```
bmiklautz: What was the reason to add pad as a parameter?

libfreerdp/core/license.c
hardening: I agree, I wanted to keep the comment, with a not too long line.

libfreerdp/core/rdp.c
bmiklautz: Ah. I see.

hardening: Previously it was the return value, but as this function can now fail it returns a BOOL. The previous return value is now in pad.

libfreerdp/crypto/crypto.c
```diff
@@ -112,19 +112,20 @@ CryptoDes3 crypto_des3_decrypt_init(const BYTE* key, const BYTE* ivec)
...
 
-void crypto_des3_decrypt(CryptoDes3 des3, UINT32 length, const BYTE* in_data, BYTE* out_data)
+BOOL crypto_des3_decrypt(CryptoDes3 des3, UINT32 length, const BYTE* in_data, BYTE* out_data)
 {
 	int len;
-	EVP_DecryptUpdate(&des3->des3_ctx, out_data, &len, in_data, length);
+	int ret = EVP_DecryptUpdate(&des3->des3_ctx, out_data, &len, in_data, length);
 
 	if (length != len)
 		abort(); /* TODO */
```
bmiklautz: Has nothing todo with the pull request itself but the TODO is fitting here . abort isn't much better than exit ;)

libfreerdp/core/rdp.c
bmiklautz: Well done! Perfect trick.

bmiklautz:
+1
@hardening really nice! Thanks.

libfreerdp/crypto/crypto.c
hardening: in this case this should be an assert(). If the 3des decryption doesn't give the expected out bytes we're in serious troubles ;). So we may do some kind of abort anyway.

bmiklautz: yeah sure but within an library abort is as bad as exit.

hardening: That's a philosophic question, but most of the time if we detect something totally unexpected, I think the safest is to just abort(). Just to prevent more damages.

libfreerdp/core/fastpath.c
```diff
@@ -767,6 +767,7 @@ BOOL fastpath_send_multiple_input_pdu(rdpFastPath* fastpath, wStream* s, int iNu
 	rdpRdp* rdp;
 	UINT16 length;
 	BYTE eventHeader;
+	BOOL status;
```
nfedera: suggest to move status to the else tree below where it is used

libfreerdp/core/license.c
```diff
@@ -431,21 +453,23 @@ void license_encrypt_premaster_secret(rdpLicense* license)
...
+		return FALSE;
 	license->PlatformChallenge->length = license->EncryptedPlatformChallenge->length;
 
 	rc4 = crypto_rc4_init(license->LicensingEncryptionKey, LICENSING_ENCRYPTION_KEY_LENGTH);
-
 	if (!rc4)
 	{
 		WLog_ERR(TAG, "unable to allocate a rc4");
-		return;
+		return FALSE;
```
nfedera: missing free(license->PlatformChallenge->data) ?

libfreerdp/core/license.c
```diff
@@ -790,7 +823,7 @@ BOOL license_read_platform_challenge_packet(rdpLicense* license, wStream* s)
 	WLog_DBG(TAG, "MacData:");
 	winpr_HexDump(TAG, WLOG_DEBUG, MacData, 16);
 #endif
-	return TRUE;
+	return ret;
```
nfedera: if license_decrypt_platform_challenge fails some of the pointers printed in the debug lines above are null.
Suggest to do the debug statements inside a  if (ret)

libfreerdp/core/security.c
```diff
@@ -150,42 +150,46 @@ static void security_salted_hash(const BYTE* salt, const BYTE* input, int length
...
-void security_master_secret(const BYTE* premaster_secret, const BYTE* client_random,
+BOOL security_master_secret(const BYTE* premaster_secret, const BYTE* client_random,
 		const BYTE* server_random, BYTE* output)
 {
 	/* MasterSecret = PremasterHash('A') + PremasterHash('BB') + PremasterHash('CCC') */
-	security_premaster_hash("A", 1, premaster_secret, client_random, server_random, &output[0]);
-	security_premaster_hash("BB", 2, premaster_secret, client_random, server_random, &output[16]);
-	security_premaster_hash("CCC", 3, premaster_secret, client_random, server_random, &output[32]);
+	return
+		security_premaster_hash("A", 1, premaster_secret, client_random, server_random, &output[0]) &&
```
nfedera: looks funny ;) suggest to start that directly right of the return

libfreerdp/crypto/crypto.c
nfedera: Why not just return FALSE ?  Is this so unlikely that we must assume a CPU / memory error ?

libfreerdp/crypto/crypto.c
```diff
@@ -145,14 +146,24 @@ CryptoHmac crypto_hmac_new(void)
...
 
-void crypto_hmac_md5_init(CryptoHmac hmac, const BYTE* data, UINT32 length)
+BOOL crypto_hmac_md5_init(CryptoHmac hmac, const BYTE* data, UINT32 length)
 {
+#if (OPENSSL_VERSION_NUMBER >= 0x00909000)
+	return HMAC_Init_ex(&hmac->hmac_ctx, data, length, EVP_md5(), NULL) == 1;
+#else
 	HMAC_Init_ex(&hmac->hmac_ctx, data, length, EVP_md5(), NULL);
+#endif
+	return TRUE;
```
nfedera: should move return TRUE one line up to avoid unreachable code warnings

nfedera:
+1 after the remarks are handled

libfreerdp/crypto/crypto.c
hardening: @nfedera this is definitely an unexpected result, that means some kind of corruption has occured.

bmiklautz: @hardening in crypto_des3_encrypt this isn't checked for example. From the code point of view I like @nfedera s idea to handle it the same way we would handle a general decryption error.

hardening: The problem is that if we hit such situation it's much more than an innocent decryption error. It's really an assertion, we can't be sure that staying will not result in a massive corruption.

hardening:
So anyone to merge it ?

Issue #19, opened 2011-07-15
warning during compilation in cliprdr_format.c


celsius:
Here is a very minor patch :

diff --git a/channels/cliprdr/cliprdr_format.c b/channels/cliprdr/cliprdr_format.c
index 6cce758..8b2d00f 100644
--- a/channels/cliprdr/cliprdr_format.c
+++ b/channels/cliprdr/cliprdr_format.c
@@ -87,7 +87,7 @@ void cliprdr_process_format_list(cliprdrPlugin\* cliprdr, STREAM\* data_in, uint32
        cb_event->formats = (uint32*)xmalloc(sizeof(uint32) \* num_formats);
        cb_event->num_formats = 0;
        if (num_formats \* 36 != dataLen)
-               DEBUG_WARN("dataLen %d not devided by 36!");
-               DEBUG_WARN("dataLen %d not divided by 36!", dataLen);
      for (i = 0; i < num_formats; i++)
      {
              stream_read_uint32(data_in, format);

(sorry, I still have to learn how to use git and github properly)

Issue #16, opened 2011-07-12
cmake and config.h not found


celsius:
Hi

when using cmake, I usually build everything in a separate directory :

mkdir Build
cd Build
cmake path/to/main/cmakelists.txt

since config.h is created using 

configure_file(${CMAKE_CURRENT_SOURCE_DIR}/config.h.in ${CMAKE_CURRENT_BINARY_DIR}/config.h)

CMAKE_CURRENT_BINARY_DIR should be added to the include path in the main CMakeLists.txt :

include_directories(${CMAKE_CURRENT_BINARY_DIR})

Otherwise I get this error :

FreeRDP-1.0/libfreerdp-utils/svc_plugin.c:21:20: error: config.h: File not found

Another solution would be to put config.h in include or at the root

Also, HAVE_CONFIG_H is always false

PR #2519, opened 2015-04-05
Fix X11 hung when we have multiple thread interacts with X11

realjiangms:
Fix X11 hung when we have multiple thread interacts with X11, xfreerdp may have this issue with +async-input / +async-transport / +async-update

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -304,20 +304,20 @@ BOOL xf_sw_end_paint(rdpContext* context)
 			if (gdi->primary->hdc->hwnd->invalid->null)
 				return TRUE;
 
-			xf_lock_x11(xfc, FALSE);
+			xf_lock_x11(xfc, TRUE);
 
 			XPutImage(xfc->display, xfc->primary, xfc->gc, xfc->image, x, y, x, y, w, h);
 
 			xf_draw_screen(xfc, x, y, w, h);
 
-			xf_unlock_x11(xfc, FALSE);
+			xf_unlock_x11(xfc, TRUE);
 		}
 		else
 		{
 			if (gdi->primary->hdc->hwnd->ninvalid < 1)
 				return TRUE;
 
-			xf_lock_x11(xfc, FALSE);
+			xf_lock_x11(xfc, TRUE);
 
 			for (i = 0; i < ninvalid; i++)
 			{
@@ -333,7 +333,7 @@ BOOL xf_sw_end_paint(rdpContext* context)
 
 			XFlush(xfc->display);
 
-			xf_unlock_x11(xfc, FALSE);
+			xf_unlock_x11(xfc, TRUE);
 		}
 	}
 	else
@@ -409,11 +409,11 @@ BOOL xf_hw_end_paint(rdpContext* context)
 			w = xfc->hdc->hwnd->invalid->w;
 			h = xfc->hdc->hwnd->invalid->h;
 
-			xf_lock_x11(xfc, FALSE);
+			xf_lock_x11(xfc, TRUE);
 
 			xf_draw_screen(xfc, x, y, w, h);
 
-			xf_unlock_x11(xfc, FALSE);
+			xf_unlock_x11(xfc, TRUE);
 		}
 		else
 		{
@@ -427,7 +427,7 @@ BOOL xf_hw_end_paint(rdpContext* context)
 			ninvalid = xfc->hdc->hwnd->ninvalid;
 			cinvalid = xfc->hdc->hwnd->cinvalid;
 
-			xf_lock_x11(xfc, FALSE);
+			xf_lock_x11(xfc, TRUE);
 
 			for (i = 0; i < ninvalid; i++)
 			{
@@ -441,7 +441,7 @@ BOOL xf_hw_end_paint(rdpContext* context)
 
 			XFlush(xfc->display);
 
-			xf_unlock_x11(xfc, FALSE);
+			xf_unlock_x11(xfc, TRUE);
 		}
 	}
 	else
@@ -500,16 +500,20 @@ BOOL xf_process_x_events(freerdp* instance)
 
 	while (pending_status)
 	{
-		xf_lock_x11(xfc, FALSE);
+		xf_lock_x11(xfc, TRUE);
 		pending_status = XPending(xfc->display);
-		xf_unlock_x11(xfc, FALSE);
+		xf_unlock_x11(xfc, TRUE);
 
 		if (pending_status)
 		{
+			xf_lock_x11(xfc, TRUE);
+
 			ZeroMemory(&xevent, sizeof(xevent));
 			XNextEvent(xfc->display, &xevent);
 
 			status = xf_event_process(instance, &xevent);
+
+			xf_unlock_x11(xfc, TRUE);
 
 			if (!status)
 				return status;
@@ -1552,20 +1556,20 @@ void* xf_input_thread(void* arg)
 		{
 			do
 			{
-				xf_lock_x11(xfc, FALSE);
+				xf_lock_x11(xfc, TRUE);
 				pending_status = XPending(xfc->display);
-				xf_unlock_x11(xfc, FALSE);
+				xf_unlock_x11(xfc, TRUE);
 
 				if (pending_status)
 				{
-					xf_lock_x11(xfc, FALSE);
+					xf_lock_x11(xfc, TRUE);
 
 					ZeroMemory(&xevent, sizeof(xevent));
 					XNextEvent(xfc->display, &xevent);
 
 					process_status = xf_event_process(instance, &xevent);
 
-					xf_unlock_x11(xfc, FALSE);
+					xf_unlock_x11(xfc, TRUE);
 
 					if (!process_status)
 						break;
--- a/client/X11/xf_gdi.c
+++ b/client/X11/xf_gdi.c
@@ -418,7 +418,7 @@ BOOL xf_gdi_bitmap_update(rdpContext* context, BITMAP_UPDATE* bitmapUpdate)
 			pSrcData = xfc->bitmap_buffer;
 		}
 
-		xf_lock_x11(xfc, FALSE);
+		xf_lock_x11(xfc, TRUE);
 
 		XSetFunction(xfc->display, xfc->gc, GXcopy);
 
@@ -440,7 +440,7 @@ BOOL xf_gdi_bitmap_update(rdpContext* context, BITMAP_UPDATE* bitmapUpdate)
 
 		ret = gdi_InvalidateRegion(xfc->hdc, nXDst, nYDst, nWidth, nHeight);
 
-		xf_unlock_x11(xfc, FALSE);
+		xf_unlock_x11(xfc, TRUE);
 
 		if (!ret)
 			break;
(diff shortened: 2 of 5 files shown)


realjiangms:
More information for the defect:
I suggest to always protect X11 operations with mutex regardless Xthread mechanism is used or not because the X11 multi-thread mechanism is not reliable. Actually this is not FreeRDP bug, it should blame X11 library.

Test done with libx11-6_1.4.99.1-0ubuntu2.2_amd64 on ubuntu. The source code may be (but not very sure) http://www.x.org/releases/X11R7.6/src/lib/libX11-1.4.0.tar.gz or http://www.x.org/releases/X11R7.5/src/lib/libX11-1.3.2.tar.gz.

With debug package, xfreerdp hang as following stacks with +async-transport:

Thread 3 (Thread 0x7fffeb9cc700 (LWP 1634)):
#0  0x00007ffff565589c in __lll_lock_wait () from /lib/x86_64-linux-gnu/libpthread.so.0
#1  0x00007ffff5651065 in _L_lock_858 () from /lib/x86_64-linux-gnu/libpthread.so.0
#2  0x00007ffff5650eba in pthread_mutex_lock () from /lib/x86_64-linux-gnu/libpthread.so.0
#3  0x00007ffff6f27d32 in _XInternalLockDisplay (dpy=0x63b730, wskip=0) at ../../src/locking.c:480
#4  0x00007ffff6f3cab4 in sync_while_locked (dpy=0x63b730) at ../../src/XlibInt.c:217
#5  _XSeqSyncFunction (dpy=0x63b730) at ../../src/XlibInt.c:233
#6  0x00007ffff6f2eb74 in XPutImage (dpy=0x63b730, d=6291459, gc=0x7fffec0728d0, image=0x7fffe40016e0, req_xoffset=0, req_yoffset=0, x=128, y=704, req_width=64, req_height=6
#7  0x00007ffff7b40630 in xf_gdi_surface_bits () from /usr/local/lib/libxfreerdp-client.so.1.2
#8  0x00007ffff5e18d8a in update_recv_surfcmds () from /usr/local/lib/libfreerdp.so.1.2
#9  0x00007ffff5e1713d in fastpath_recv_update () from /usr/local/lib/libfreerdp.so.1.2
#10 0x00007ffff5e17a5b in fastpath_recv_update_data () from /usr/local/lib/libfreerdp.so.1.2
#11 0x00007ffff5e17c63 in fastpath_recv_updates () from /usr/local/lib/libfreerdp.so.1.2
#12 0x00007ffff5e145c0 in rdp_recv_pdu () from /usr/local/lib/libfreerdp.so.1.2
#13 0x00007ffff5e14733 in rdp_recv_callback () from /usr/local/lib/libfreerdp.so.1.2
#14 0x00007ffff5e1a4e0 in transport_check_fds () from /usr/local/lib/libfreerdp.so.1.2
#15 0x00007ffff5e14bc8 in rdp_check_fds () from /usr/local/lib/libfreerdp.so.1.2
#16 0x00007ffff5dfa3e8 in freerdp_check_fds () from /usr/local/lib/libfreerdp.so.1.2
#17 0x00007ffff5dfa4bc in freerdp_check_event_handles () from /usr/local/lib/libfreerdp.so.1.2
#18 0x00007ffff5e192f8 in transport_client_thread () from /usr/local/lib/libfreerdp.so.1.2
#19 0x00007ffff7658061 in thread_launcher () from /usr/local/lib/libwinpr.so.1.1
#20 0x00007ffff564ee9a in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#21 0x00007ffff7324cbd in clone () from /lib/x86_64-linux-gnu/libc.so.6
#22 0x0000000000000000 in ?? ()

Thread 2 (Thread 0x7ffff0abf700 (LWP 1633)):
#0  _XUserLockDisplay (dpy=0x63b730) at ../../src/locking.c:489
#1  0x00007ffff6f3ca76 in sync_while_locked (dpy=0x63b730) at ../../src/XlibInt.c:213
#2  _XSeqSyncFunction (dpy=0x63b730) at ../../src/XlibInt.c:233
#3  0x00007ffff6f2b31f in XPending (dpy=0x63b730) at ../../src/Pending.c:51
#4  0x00007ffff7b529ce in xf_process_x_events () from /usr/local/lib/libxfreerdp-client.so.1.2
#5  0x00007ffff7b532a9 in xf_client_thread () from /usr/local/lib/libxfreerdp-client.so.1.2
#6  0x00007ffff7658061 in thread_launcher () from /usr/local/lib/libwinpr.so.1.1
#7  0x00007ffff564ee9a in start_thread () from /lib/x86_64-linux-gnu/libpthread.so.0
#8  0x00007ffff7324cbd in clone () from /lib/x86_64-linux-gnu/libc.so.6
#9  0x0000000000000000 in ?? ()

realjiangms:
(continue with stacks in previous note)
In src/XlibInt.c:
 208 static
 209 void sync_while_locked(Display *dpy)
 210 {
 211 #ifdef XTHREADS
 212     if (dpy->lock)
 213         (*dpy->lock->user_lock_display)(dpy); ========> Thread 2 at line 213 now
 214 #endif
 215     UnlockDisplay(dpy);
 216     SyncHandle();
 217     InternalLockDisplay(dpy, /* don't skip user locks */ 0); ==========> Thread 3 at line 217 now
 218 #ifdef XTHREADS
 219     if (dpy->lock)
 220         (*dpy->lock->user_unlock_display)(dpy);
 221 #endif
 222 }

But now 
(gdb) p /x *dpy->lock
$64 = {
  mutex = 0x642800,  ====> Xdisplay mutex __owner = 1633, Thread 2
  reply_bytes_left = 0x0,
  reply_was_read = 0x0,
  reply_awaiters = 0x0,
  reply_awaiters_tail = 0x642700,
  event_awaiters = 0x0,
  event_awaiters_tail = 0x642710,
  reply_first = 0x0,
  locking_level = 0x1,
  locking_thread = 0x7fffeb9cc700,  ====> Xdisplay user lock belongs to 0x7fffeb9cc700, thread 3
  cv = 0x6427c0,

Now mutex is held by thread 2. And thread 2 will make locking_level++ in dpy->lock->user_lock_display. Then thread 2 run to line 215 and release mutex. Thread 3 get the mutex and make locking_level--. It should will NOT clear locking_thread because locking_level is 2 (not 1), so locking_thread keep still 0x7fffeb9cc700, the thread 3.

Then thread 2 would stuck forever at line 217 (function _XInternalLockDisplay()) because it found dpy->lock->locking_thread is not itself and it would try to wait until someone release dpy->lock->locking_thread. But actually the lock is held by itself now, the original owner had already gone.

realjiangms:
I just change the fix. The X11 library is OK. However, our usage of X11 library is not correct.
Is there any reason to skip XLockDisplay when calling X11 library? 
I just re-investigated the issue and found if we always enable XLockDisplay, the issue described in previous notes would not happen.

The X11 library requires XLockDisplay/XUnlockDisplay around X11 operations. We calls xf_lock_x11/xf_unlock_x11 correctly, but the real X multi-thread mechanism is somehow skipped by intent, not sure what's the purpose behind.

bmiklautz:
@realjiangms when looking at your fix you just remove the check for display making it an unused variable - instead of removing this check I think it would make sense to identify the real place where the lock is called with display set to false.

realjiangms:
@bmiklautz  Yeah the fix is not final. It's just a minimal change to a) Present my idea. b) Test and prove it works.
My suggestion is to always use XLockDisplay when we calls X operation function. It's required by X11 library. We basically can't tell exactly which X function needs lock and which doesn't. What's the original idea to introduce the 'display' check to skip X lock?

The final fix may be either of following:
1) Discard 'display' argument in completely.
2) If we really want to keep the 'display' check, we can just set 'display' to true when race condition found. For example, this time I found race between XPending and XPutImage, and we just set 'display' to true around these 2 functions. But this leaves other X operations still NOT protected by the X lock. If it cause some other issues next time, we may not be able identify that it's due to lack of multi-thread protection

realjiangms:
@bmiklautz  
Here's more investigation on where we use xf_lock_x11:
Totally 42 calls with display= FALSE and only 3 calls with display=TRUE. 3 calls with display=TRUE is in xf_sw_desktop_resize, xf_hw_desktop_resize and xf_UpdateWindowArea.

For 42 calls with display= FALSE, the X functions which are NOT protected:

XDefineCursor
XUndefineCursor
XFreeCursor
XWarpPointer
XPutImage
XCopyArea
XFlush
XPending
XNextEvent
XSync
XSetFunction
XSetClipRectangles
XSetClipMask
XFillRectangle
XSetFillStyle
XSetForeground
XCopyArea

The above X function should be protected according to X11 source code, because they calls X internal LockDisplay() function. This function may call sync_while_locked which brings race condition.

In addition, xf_draw_screen should always be protected as it contains a lot of X functions.

When going through all the code where xf_lock_x11 is used, I found some code around by xf_lock_x11 doesn't involve X display. They should not be protected by X lock. 

I'll try to figure out everywhere we need X lock and fix them one-by-one and update the pull request.

client/X11/xf_gdi.c
```diff
@@ -1122,14 +1122,14 @@ void xf_gdi_surface_bits(rdpContext* context, SURFACE_BITS_COMMAND* cmd)
 	RFX_MESSAGE* message;
 	xfContext* xfc = (xfContext*) context;
 
-	xf_lock_x11(xfc, FALSE);
```
realjiangms: Here I move the xf_lock_x11 after bitmap data decode so that other thread which need X display could run on other CPU while time-consuming bitmap decode task is running.

client/X11/xf_client.c
```diff
@@ -475,17 +475,21 @@ BOOL xf_process_x_events(freerdp* instance)
...
 	{
-		xf_lock_x11(xfc, FALSE);
+		xf_lock_x11(xfc, TRUE);
 		pending_status = XPending(xfc->display);
-		xf_unlock_x11(xfc, FALSE);
+		xf_unlock_x11(xfc, TRUE);
 
 		if (pending_status)
 		{
+			xf_lock_x11(xfc, TRUE);
```
realjiangms: Protect following X function.
We do correct thing in xf_input_thread, but seems forget here.

bmiklautz:
@realjiangms if I'm not mistaken the display was added for RAIL where multiple windows can exist (using the same drawing functions). In those cases it doesn't necessarily make sens to lock the complete server if some update happens for one window. However I guess this wasn't tested/used with async.

realjiangms:
@bmiklautz 
Thanks for the information. I'm not very clear what RAIL refer to, but does 'multiple windows' means multiple X Display object? If so, the lock on one Display object would not impact lock on another.

If there's only one Display object, the X functions themselves would always lock each other. If we look at the source code of X functions such XPending and XPutImage, they would call 'LockDisplay' which expected to hold a mutex during their operation, so actually they can't be run in parallel. But 'LockDisplay' sometimes calls _XSeqSyncFunction which leads to deadlock (it internally could release the mutex and use some common data). If 2 threads run into LockDisplay in parallel, deadlock could happen.

Accessing same Display Object in parallel is not suggested. But I agree that we need to think about it carefully that the original consideration on RAIL is not broken.

realjiangms:
This issue described in this PR is controversial. Currently just dismiss it

PR #2527, opened 2015-04-08
fixed undefined behaviour of sprintf, buffer overflow, crash

AMV007:


bmiklautz:
@freerdp-bot test

bmiklautz:
@AMV007 where exactly did you get the buffer overflow/crash?

AMV007:
@bmiklautz 
exactly was buffer overflow at :
`sprintf(strContainerId,
		"{%02x%02x%02x%02x-%02x%02x-%02x%02x-%02x%02x-%02x%02x%02x%02x%02x%02x}",
		containerId[0], containerId[1],containerId[2], containerId[3],
		containerId[4], containerId[5], containerId[6], containerId[7],
		containerId[8], containerId[9], containerId[10], containerId[11],
		containerId[12], containerId[13], containerId[14], containerId[15]);`

nfedera:
@AMV007 IO this doesn't fix anything

nfedera:
The real issue in func_container_id_generate() is the first sprintf creating the containerId:
```
sprintf(containerId, "%04X%04X%s", idVendor, idProduct, p);
```
And containerId, idVendor and idProduct are defined like this:
```
char containerId[17];
int idVendor, idProduct;
```
The code makes sure that p is not longer than 8. However, the original author seems to assume that the "%04X" format specifier would always generate 4 characters which is not true. 
So if idVendor/idProduct are negative or >= 2^16  th sprintf will write outside of containerId

channels/urbdrc/client/urbdrc_main.c
```diff
@@ -50,9 +50,9 @@ static int func_hardware_id_format(IUDEVICE* pdev, char(*HardwareIds)[DEVICE_HAR
 	idProduct = pdev->query_device_descriptor(pdev, ID_PRODUCT);
 	bcdDevice = pdev->query_device_descriptor(pdev, BCD_DEVICE);
 
-	sprintf(str, "USB\\VID_%04X&PID_%04X", idVendor, idProduct);
+	sprintf(str, "USB\\VID_%04X&PID_%04X", (unsigned int)idVendor, (unsigned int)idProduct);
```
nfedera: (unsigned int) does not fit into %04X !

channels/urbdrc/client/urbdrc_main.c
```diff
@@ -152,15 +152,15 @@ static int func_container_id_generate(IUDEVICE* pdev, char* strContainerId)
 	else
 		p = path;
 
-	sprintf(containerId, "%04X%04X%s", idVendor, idProduct, p);
+	sprintf(containerId, "%04X%04X%s", (unsigned int)idVendor, (unsigned int)idProduct, p);
```
nfedera: (unsigned int) does not fit into %04X

AMV007:
ye, it's must be unsgined short, my mistake )

AMV007:
should i close this pull and open new one ?

AMV007:
fixed in patch

nfedera:
@AMV007 that should do it. However, I'd personally prefer to use correct variables in the first place (e.g. UINT16 idVendor, idProduct;  or BYTE bDeviceSubClass)  then you don't need to cast a hundred times. And also use snprintf instead of sprintf !

AMV007:
ye, better do work right, but have limited time usually, right now had task to quick fix freerdp crash )
if sprintf crashing - you can easely find crash place, but if snprintf will provide incorrect result-this bug may last years, if generally all will work ?

nfedera:
@AMV007, no offense, but writing beyond valid buffers is NEVER better. For starters http://en.wikipedia.org/wiki/Buffer_overflow

nfedera:
@AMV007 Nevertheless many thanks for helping and reporting the issue!
Don't worry because of your limited time. I've created a modified PR: #2531
Do you mind testing and reporting if if works for you ?

AMV007:
@nfedera, no I haven't problem device in my hands, it was tester issue, and remote debug for find error, but your patch must work.
By the way you forgot couple sprintf change to snprintf,  look in this updated patch, I did some extend to your version.

nfedera:
+1 LGTM

PR #2530, opened 2015-04-10
Revert incomplete pull request #2426 and fix fullscreen + smart-sizing

nfedera:
- revert commit d4a9a2d from @bjcollins (he will come up with an updated PR that considers the
presence of the smart-sizing functionality)
- make fullscreen work with scaling again as it got already broken before @bjcollins's PR

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -543,29 +543,20 @@ BOOL xf_create_window(xfContext* xfc)
 
 		if (xfc->fullscreen)
 		{
-			width = xfc->desktopWidth;
-			height = xfc->desktopHeight;
+			width = WidthOfScreen(xfc->screen);
+			height = HeightOfScreen(xfc->screen);
 		}
 
 #ifdef WITH_XRENDER
 		if (settings->SmartSizing)
 		{
-			if (xfc->fullscreen)
-			{
-				if (xfc->window)
-				{
-					settings->SmartSizingWidth = xfc->window->width;
-					settings->SmartSizingHeight = xfc->window->height;
-				}
-			}
-			else
+			if (!xfc->fullscreen)
 			{
 				if (settings->SmartSizingWidth)
 					width = settings->SmartSizingWidth;
 				if (settings->SmartSizingHeight)
 					height = settings->SmartSizingHeight;
 			}
-
 			xfc->scaledWidth = width;
 			xfc->scaledHeight = height;
 		}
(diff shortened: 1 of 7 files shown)


nfedera:
Also see discussion in FreeRDP/FreeRDP#2426

bjcollins:
@nfedera Just stating my opinion below. I have no problem with what you have done if everyone believes it is the better trade-off in functionality until such a time that a patch to support both sets of functionality is completed.

For the X client, is smart-sizing really a higher priority then fullscreen/multimonitor functionality in general? I agree that they both should be addressed, but I just think that correct multimonitor functionality would be more important to users then smart-sizing with a fullscreen mode in the near term. I will look at this in the future regardless(although I admittedly do not know how soon that will be), so what is deemed as more important should be what we have in master for now until a patch is made to make both things work again.

My opinion is skewed as the user base that I support cares heavily about multimonitor support and smart-sizing would really just be a luxury option. I would think that most users would give up the ability to scale the window on an X desktop if we gave them the ability to correctly redisplay a multimonitor RDP session.

nfedera:
@bjcollins I fully agree with everything you say and of course fullscreen is more important.
However, my understanding was that you did your patch without considering some of the existing features (smart-sizing/pinch/zoom) because you were not aware of their presence.
And in the PR #2426 conversation you offered to rewrite your patch and this PR would give you a perfect starting point. Don't get me wrong but it has already happened so many times during FreeRDP development that one fix introduced multiple other issues or has killed existing features ...
It all depends on your time. It is ok if you say that you won't be able to rewrite your patch in the near future. In that case I can try to fix your PR next week.

bjcollins:
@nfedera  I admit it is not my top priority right now and I do not know if it will be a couple of weeks or months before I can focus back on this. I do want to do the right thing here in the end, as we both do. I think breaking smart-sizing for real multimonitor support is a worthwhile trade-off until such a time that smart-sizing is reworked to work with fullscreen/multimonitor again. However, I see your point that my patch goes down the same trend of fixing something but breaking something else and would have been rejected if that was caught before it was merged in the first place. So, I would like to keep my patch in for what I consider the best functionality, but I completely understand the decision to pull my patch because it was flawed in the first place and I can not commit to fix/resubmit it in a timely manner.

bmiklautz:
Unfortunately we can't automate all of the testing and a lot of tests do require special setups which makes it even more complex. I think a PR that breaks any features of functionality shouldn't make it to master but since no complete test documentation/procedure isn't available (yet) it happens ;(.
Therfore @bjcollins would you add the tests and setup you did for multi monitor fullscreen to https://github.com/FreeRDP/FreeRDP/wiki/Quality-Assurance-(QA) ? - This shouldn't take to much time and will help to make sure the functionality is tested in the future eg. on related pull requests.

bjcollins:
@bmiklautz I will put something together for the QA page when I look at this again.

nfedera:
@bjcollins I'm closing this ... will work on the fixes based on your pr ...

PR #2535, opened 2015-04-13
Fixed NULL dereference.

akallabeth:
In```rdp_security_stream_out``` a pointer was set to ```NULL``` instead of its value being set when using FIPS encryption leading to a crash the very next line.


bmiklautz:
@akallabeth good catch!

hardening:
My fault, sorry.

PR #2538, opened 2015-04-14
transport: added function to get bytes written

MartinHaimberger:
added function to get written bytes on transport.
This is needed to get a more accurate bandwidth
management.

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -576,10 +576,12 @@ int transport_write(rdpTransport* transport, wStream* s)
 {
 	int length;
 	int status = -1;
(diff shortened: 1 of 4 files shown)


libfreerdp/core/transport.c
```diff
@@ -637,8 +639,10 @@ int transport_write(rdpTransport* transport, wStream* s)
 		length -= status;
 		Stream_Seek(s, status);
 	}
+	transport->written += writtenlength;
 
 out_cleanup:
+
```
bmiklautz: Unnecessary line.

bmiklautz:
+1

PR #2547, opened 2015-04-16
Fullscreen fixes and smart sizing resuscitation

nfedera:
- make smart-sizing work again which was killed in previous commits
- renamed xfc->width and xfc->height to sessionWidth / sessionHeight since several contributers in the past repeatedly made the error to treat these as equal with the x11 window width and heigth
- removed unused variables from xf_detect_monitors()
- removed xfc->desktopWidth and xfc->desktopHeight context variables since they were basically unused (from the contexts's point of view) and only served as temporary storage
- removed some weird/unnecessary/ugly code and workarounds
- miscellaneous small fixes
- new feature: restore previous window position when toggling out of  fullscreen mode
- new feature: if /f is specified in combination with /smart-sizing:WxH we run the session in the /smart-sizing dimensions scaled to full screen

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -554,6 +554,9 @@ BOOL xf_create_window(xfContext* xfc)
 				width = settings->SmartSizingWidth;
 			if (settings->SmartSizingHeight)
 				height = settings->SmartSizingHeight;
+
+			xfc->scaledWidth = width;
+			xfc->scaledHeight = height;
 		}
 #endif
 
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -247,11 +247,19 @@ static void xf_desktop_resize(rdpContext* context)
 
 	if (!xfc->fullscreen)
 	{
-		if (xfc->window)
-			xf_ResizeDesktopWindow(xfc, xfc->window, settings->DesktopWidth, settings->DesktopHeight);
+		xf_ResizeDesktopWindow(xfc, xfc->window, settings->DesktopWidth, settings->DesktopHeight);
 	}
 	else
 	{
+#ifdef WITH_XRENDER
+		if (!xfc->settings->SmartSizing)
+#endif
+		{
+			/* Update the saved width and height values the window will be
+			 * resized to when toggling out of fullscreen */
+			xfc->savedWidth = xfc->sessionWidth;
+			xfc->savedHeight = xfc->sessionHeight;
+		}
 		XSetFunction(xfc->display, xfc->gc, GXcopy);
 		XSetFillStyle(xfc->display, xfc->gc, FillSolid);
 		XSetForeground(xfc->display, xfc->gc, 0);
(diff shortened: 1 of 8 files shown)


bjcollins:
@nfedera First, thanks for taking the effort to try to bring this altogether. As far as I can tell, the functionality from my PR is still intact. I do not really know the state of smart-sizing before all of your changes was but it seems to not behave as I would expect from the QA document.

Two examples:

1.  xfreerdp /v:x.x.x.x /smart-sizing:640x480 s/ize:1920x1080
   I would expect to see a 1920x1080 desktop scaled down to a 640x480 window. The window is 640x480 and the desktop is 1920x1080, but no scaling happens. This looks like an initialization issue cause if I toggle fullscreen then back then it behaves as expected.

2. xfreerdp /v:x.x.x.x /smartsizing:640x480 /size:1920x1080
   I would expect my session to be running at 1920x1080 and scaled to my fullscreen dimensions, instead it looks like my session is running at 640x480... Or maybe my session is being scaled down then displayed fullscreen afterwards.

Let me know if this makes sense or if you are seeing different results. I just did some quick testing with CentOS and metacity.

bmiklautz:
@bjcollins if you got a moment could briefly describe your test setup and your tests?

nfedera:
@bjcollins Thanks for testing.
Re 1):
Definitely a bug which is fixed in the last commit fd5f474

Re 2):
This is different. In general (indep. of smart-sizing) /f has always nullified the /size parameter.
So your ```/smartsizing:640x480 /size:1920x1080 /f``` options are actually the same as  ```/smartsizing:640x480 /f```.
Now /f in combination with /smartsizing:WxH is handled special now (based on a feature request) and I'm quoting the QA page here:

> - if /f is specified in combination with /smart-sizing:WxH we run the session in the /smart-sizing dimensions scaled to full screen
> - Example xfreerdp /v:servername /f /smart-sizing:320x200 This should create a 320x200 rdp session scaled to your full screen dimensions

I just confirmed that it is working exactly like this and the result you've described it is the correct behavior.

bjcollins:
@bmiklautz I agree that testing documentation is needed for fullscreen, and I will try to get that started. I think we should write up and verify appropriate behavior with lack of Xinerama and/or lack of NET_WM_STATE_FULLSCREEN and NET_WM_FULLSCREEN_MONITORS hints available.I will try to cover the correct functionality with those things available in the next few days. I think there may be a bug in the absence of those things and I would want to look at it again and test to try to determine what I would consider to be desired functionality in those situations.

bjcollins:
@nfedera  re 2) sorry, should have read the docs closer :) Very confusing to me to have what seems like dual meanings for that parameter. In the non-fullscreen case /smart-sizing is used for the local window sizing. In the fullscreen case, it is used for the remote desktop sizing.

Totally outside the scope of your PR, but just stating my opinion below based on what little I know about the whole smart-sizing thing...

I think it would be better if...
1. Without /smart-sizing and without /f then /size would be used for both the remote desktop size and local window size.
2. Without /smart-sizing but with /f then /size would be totally ignored and the remote desktop size would be calculated based on the local screen size
3. With /smart-sizing but without /f then /size would only refer to the local window size and /smart-sizing would set the remote desktop size
4. With /smart-sizing but with /f then /size would be totally ignored AND /smart-sizing would set the remote desktop size

So, my case 1) above would scale a 640x480 desktop up to a 1920x1080 window and my case 2) above would still work the same. Basically, /smart-sizing would always have a dual purpose of enabling smart-sizing as well as setting the remote desktop size that will be scaled from.

nfedera:
@bjcollins I'd propose to create an issue where we can discuss better or more intuitive options that can be changed after this PR is merged - d'accord ?

Hobby-Student:
@nfedera nice work! Will try to get my test-environment updated this weekend.Will report back.

Hobby-Student:
had time to update my test-environment. could only do some quick test for

/f /smart-sizing:640x480
/f /smart-sizing:1920x1080

Display resolution is 1024x768 (thinclient VM). Both is working as expected. Fullscreen RDP scaled to lower res and scaled to higher res.

Thx for your efforts! If you need further testing, please let me know what you would like to have tested and I try to find some free time.

nfedera:
@Hobby-Student cool, Thx for doing the tests!

Hobby-Student:
@nfedera I'm currently testing wfreerdp for some windows clients (high DPI no fullscreen smart-sizing in Microsoft RDP-client).

Smart-sizing in general is working in wfreerdp working in window mode. E.g. "/size:1024x768 /smart-sizing:640x480" starts a "1024x768" window which can be resized and the resolution of "1024x768" is stretched.

Should there be the feature like in xfreerdp to scale up/down to the "smart-sizing" resolution?

nfedera:
@Hobby-Student AFAIK this is not implemented in the windows client but I guess that it should be quite easy to implement if the basic scaling is already there and working.

PR #2554, opened 2015-04-20
Fix for transport event reset

akallabeth:
The BIO event reset was prone to a race condition.
Using automatic event reset now to avoid loosing events for data read.


nfedera:
LGTM +1

mfleisz:
+1

bmiklautz:
+1
Since this is only windows specific I think it does really make sense to use native functionality instead.
@akallabeth great catch! I could think it was hard to find that race.

PR #2555, opened 2015-04-20
Updated wayland client to new API.

akallabeth:


client/Wayland/wlfreerdp.c
```diff
@@ -104,12 +117,19 @@ BOOL wl_post_connect(freerdp* instance)
 
 	gdi_init(instance, CLRCONV_ALPHA | CLRCONV_INVERT | CLRBUF_32BPP, NULL);
 	gdi = instance->context->gdi;
+	if (!gdi)
+		return FALSE;
 
 	context = (wlfContext*) instance->context;
 	window = wlf_CreateDesktopWindow(context, "FreeRDP", gdi->width, gdi->height, FALSE);
+	if (!window)
+		return FALSE;
```
bmiklautz: In error case gdi memory is leaking.

client/Wayland/wlfreerdp.c
```diff
@@ -104,12 +117,19 @@ BOOL wl_post_connect(freerdp* instance)
...
 
 	 /* fill buffer with first image here */
 	window->data = malloc (gdi->width * gdi->height *4);
+	if (!window->data)
+		return FALSE;
```
bmiklautz: when malloc fails gdi and memory from wlf_CreateDesktopWindow leaks

client/Wayland/wlfreerdp.c
```diff
@@ -117,14 +137,37 @@ BOOL wl_post_connect(freerdp* instance)
 	 /* put Wayland data in the context here */
 	context->window = window;
 
-	freerdp_channels_post_connect(instance->context->channels, instance);
+	if (freerdp_channels_post_connect(instance->context->channels, instance))
+		return FALSE;
```
bmiklautz: If I'm not mistaken in error case here are some possible memory leaks (gdi,  window, window->data)

client/Wayland/wlfreerdp.c
```diff
@@ -117,14 +137,37 @@ BOOL wl_post_connect(freerdp* instance)
...
+
+	if (context->display)
+		wlf_DestroyDisplay(context, context->display);
+
+	if (context->input)
+		wlf_DestroyInput(context, context->input);
+
+	gdi_free(instance);
+	if (context->window)
+		wlf_DestroyWindow(context, context->window);
```
bmiklautz: Although absolutely correct we could leave out the checks for context->display, context->input, context->input in this case since wlf_DestroyDisplay, wlf_DestroyInput respectively wlf_DestroyWindow do proper input validation.
Just an idea but not necessary.

client/Wayland/wlfreerdp.c
```diff
@@ -77,26 +81,35 @@ void wl_end_paint(rdpContext* context)
...
 
-BOOL wl_pre_connect(freerdp* instance)
+static BOOL wl_pre_connect(freerdp* instance)
 {
 	wlfDisplay* display;
 	wlfInput* input;
 	wlfContext* context;
 
-	freerdp_channels_pre_connect(instance->context->channels, instance);
+	if (freerdp_channels_pre_connect(instance->context->channels, instance))
```
hardening: isn't it !freerdp_channels_pre_connect(instance->context->channels, instance) ?

bmiklautz: @hardening that should be fine. freerdp_channels_pre_connect isn't bool it return 0 on success.

akallabeth:
@bmiklautz The resources allocated in pre and post connect are cleaned up by the post disconnect hook.

akallabeth:
Ok, fixed resource cleanup according your comment.

client/Wayland/wlfreerdp.c
hardening: so perhaps freerdp_channels_pre_connect(instance->context->channels, instance) != 0 would be better ...

hardening:
+1

PR #2395, opened 2015-02-13
Fix hang.

xsumbe00:

PR #2557, opened 2015-04-20
shadow fix multi-client hang

realjiangms:
Fix issue #2395.

### DIFF
--- a/server/shadow/shadow_client.c
+++ b/server/shadow/shadow_client.c
@@ -952,6 +952,7 @@ void* shadow_client_thread(rdpShadowClient* client)
 	HANDLE StopEvent;
 	HANDLE ClientEvent;
 	HANDLE ChannelEvent;
+	void* UpdateSubscriber;
 	HANDLE UpdateEvent;
 	freerdp_peer* peer;
 	rdpContext* context;
@@ -983,8 +984,15 @@ void* shadow_client_thread(rdpShadowClient* client)
 	peer->update->SuppressOutput = (pSuppressOutput) shadow_client_suppress_output;
 	peer->update->SurfaceFrameAcknowledge = (pSurfaceFrameAcknowledge) shadow_client_surface_frame_acknowledge;
 
+	if ((!client->StopEvent) || (!client->vcm) || (!subsystem->updateEvent))
+		goto out;
+
+	UpdateSubscriber = shadow_multiclient_get_subscriber(subsystem->updateEvent);
+	if (!UpdateSubscriber)
+		goto out;
+
 	StopEvent = client->StopEvent;
-	UpdateEvent = subsystem->updateEvent;
+	UpdateEvent = shadow_multiclient_getevent(UpdateSubscriber);
 	ClientEvent = peer->GetEventHandle(peer);
 	ChannelEvent = WTSVirtualChannelManagerGetEventHandle(client->vcm);
 
@@ -1001,11 +1009,6 @@ void* shadow_client_thread(rdpShadowClient* client)
 
 		if (WaitForSingleObject(StopEvent, 0) == WAIT_OBJECT_0)
 		{
-			if (WaitForSingleObject(UpdateEvent, 0) == WAIT_OBJECT_0)
-			{
-				EnterSynchronizationBarrier(&(subsystem->barrier), 0);
-			}
-
 			break;
 		}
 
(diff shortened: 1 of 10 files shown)


bmiklautz:
@realjiangms could you add notes on how to test this pull request or reproduce the problem?

server/shadow/shadow_client.c
```diff
@@ -938,7 +939,8 @@ void* shadow_client_thread(rdpShadowClient* client)
 	peer->update->SurfaceFrameAcknowledge = (pSurfaceFrameAcknowledge) shadow_client_surface_frame_acknowledge;
 
 	StopEvent = client->StopEvent;
-	UpdateEvent = subsystem->updateEvent;
+	UpdateSubscriber = shadow_multiclient_get_subscriber(subsystem->updateEvent);
```
nfedera: shadow_multiclient_get_subscriberreturns is not checked for NULL ? Should the thread run in that case ?

server/shadow/shadow_client.c
```diff
@@ -981,9 +978,7 @@ void* shadow_client_thread(rdpShadowClient* client)
 				shadow_client_send_surface_update(client);
 			}
 
-			EnterSynchronizationBarrier(&(subsystem->barrier), 0);
-
-			while (WaitForSingleObject(UpdateEvent, 0) == WAIT_OBJECT_0);
+			shadow_multiclient_consume(UpdateSubscriber);
```
nfedera: shadow_multiclient_consume result is not checked. I guess there is a reason why you made it BOOL ?

server/shadow/shadow_client.c
```diff
@@ -938,7 +939,8 @@ void* shadow_client_thread(rdpShadowClient* client)
 	peer->update->SurfaceFrameAcknowledge = (pSurfaceFrameAcknowledge) shadow_client_surface_frame_acknowledge;
 
 	StopEvent = client->StopEvent;
-	UpdateEvent = subsystem->updateEvent;
+	UpdateSubscriber = shadow_multiclient_get_subscriber(subsystem->updateEvent);
+	UpdateEvent = shadow_multiclient_getevent(UpdateSubscriber);
```
nfedera: shadow_multiclient_getevent is not checked for NULL. Should the thread run in that case ?

realjiangms:
@nfedera Yeah I need to check thereturn

realjiangms:
@nfedera I would add the code to check the return value this weekend. The PR will be updated
@bmiklautz 
This is an issue in shadow server on X11 environment. To reproduce this issue:
1. Launch a shadow server on X11 (I guess it could also be reproduced on MAC-shadow, but win-shadow should be OK
2. Open a video
3. At another Linux, connect the shadow server with several xfreerdp client.
The theory of operation that why the issue happen is described in  #2395 by @xsumbe00

realjiangms:
To reproduce this issue, the shadow-server should be run on an environment with at least 2 CPU threads. It would be better to run on a real machine. 
My server is a CPU with 4 threads, and the issue always occur after several minutes. But when I try to reproduce on a VM (running on same server with 4 cpu assigned), it seems have very low frequency

server/shadow/shadow_client.c
realjiangms: I notice FreeRDP project is cleaning code without return check recently. 
I agree that we should check NULL return, I need to think about how to define the behavior if we have NULL returned. So do the other one.

server/shadow/shadow_client.c
realjiangms: The return value of shadow_multiclient_consume is whether or not the subscriber really consumes the event. It's not cared currently. An explanation would be added, instea :)

realjiangms:
@nfedera 
I've update the PR with return value check.
Thanks for your comments.

xsumbe00:
I just want say that I'm using your original patch and it really is working. No hang since then :-)

realjiangms:
@xsumbe00 I'm very glad that it helps :-)

realjiangms:
Updated the pull request with latest code merged and conflict resolved.

awakecoding:
@freerdp-bot test.

bmiklautz:
@freerdp-bot test

bmiklautz:
@realjiangms it seems that your merge of the latest changes introduced build problems on windows. The preferred way is to do a rebase against the latest master.

hardening:
@freerdp-bot could you test again, I don't understand the last build failure, thank you.

hardening:
Looks like it builds now.
I don't know enough the internals of the shadow server, @awakecoding do you have any opinion on this PR ? If it looks OK to you perhaps we could merge it ?

awakecoding:
@hardening it looks good to me, I know @realjiangms has put some good amount of work into it.

PR #2558, opened 2015-04-21
h264: change encoder api and add multi-thread support.

llyzs:


libfreerdp/codec/h264.c
```diff
@@ -71,7 +71,7 @@ struct _H264_CONTEXT_OPENH264
 {
 	ISVCDecoder* pDecoder;
 	ISVCEncoder* pEncoder;
-	SEncParamBase EncParamBase;
+	SEncParamExt EncParamExt;
```
hardening: I had the impression that SEncParamExt isn't always available depending on the compilation of OpenH264, do I mixed up with something else ?

llyzs: @hardening SEncParamExt is always available. It is defined in public header codec_app_def.h side-by-side along with SEncParamBase and there is no conditional #if whatsoever. So is the InitializeExt API.

nfedera:
LGTM +1

PR #2560, opened 2015-04-21
Add count to ```*_get_event_handles``` function

akallabeth:
* Add number of available handles as argument
* Check return value, expect to return at least 1 handle.

### DIFF
--- a/libfreerdp/core/listener.c
+++ b/libfreerdp/core/listener.c
@@ -269,9 +269,9 @@ int freerdp_listener_get_event_handles(freerdp_listener* instance, HANDLE* event
 
 	for (index = 0; index < listener->num_sockfds; index++)
 	{
-		events[*nCount] = listener->events[index];
-		(*nCount)++;
+		events[index] = listener->events[index];
 	}
+	*nCount = listener->num_sockfds;
 
 	return 0;
 }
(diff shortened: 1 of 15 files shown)


bmiklautz:
When testing this PR the sample server crashes on startup.

akallabeth:
@bmiklautz Ok fixed the handle index update. Thx for the hint.

libfreerdp/core/listener.c
```diff
@@ -258,16 +258,20 @@ static BOOL freerdp_listener_get_fds(freerdp_listener* instance, void** rfds, in
...
+
 	for (index = 0; index < listener->num_sockfds; index++)
 	{
-		events[*nCount] = listener->events[index];
-		(*nCount)++;
+		events[index] = listener->events[index];
 	}
+	*nCount = listener->num_sockfds;
 
 	return 0;
```
bmiklautz: @akallabeth would be nice if we could make this function behave like the other _get_handles function. So what do you think about returning 0 on error and the number of events otherwise?

akallabeth:
@bmiklautz Yea, good idea. fixed that.

bmiklautz:
@akallabeth very nice! I like it. Tested with sample server and xfreerdp. No problems.
+1

cgholam:
Mrhban
بتاريخ Apr 21, 2015 10:30 AM، كتبها "akallabeth" <notifications@github.com>:

cgholam:
بتاريخ Apr 23, 2015 2:06 PM، كتبها "Izabel Davela" <30n90p@gmail.com>:

PR #2564, opened 2015-04-22
Fixed pthread_mutex_timedjoin detection.

akallabeth:
According to the docs, FreeBSD and Solaris both support ```pthread_mutex_timedjoin``` , that was the reason for the static / non static declaration hack.
Now CMake checks, if the function is provided by the platform and only using the drop in replacement if not found.


mfleisz:
+1

akallabeth:
@bceverly Does this fix also work for you?

bceverly:
@akallabeth This fix works for me on OpenBSD 5.6

Thanks for cleaning this up!

bmiklautz:
+1

PR #2568, opened 2015-04-23
codec/rfx: error checking and various fixes

nfedera:
- removed some unneeded null checks for free()
- fixed a memory leak in shadow_client
- removed rfx_compose_message_header from API

Changed the following functions to BOOL, check the result where they are called and handle failures:
- rfx_compose_message
- rfx_compose_message_header
- rfx_write_tile
- rfx_write_message_tileset
- rfx_write_message_frame_begin
- rfx_write_message_region
- rfx_write_message_frame_end
- rfx_write_message

rfx_process_message:
- check memory allocation failures
- verify protocol-conform order of data messages to prevents memory leaks caused by repeated allocations
- verify that header messages were parsed/received before the  data messages
- treat unknown rlgr mode as error
- fixed/added error handling
- fixed all callers to check/handle result

rfx_encode_message:
- fixed incorrect usage of realloc
- missing malloc check
- missing check of CreateThreadpoolWork
- correct cleanup on failure (threadpool, memory)
- check rfx_encode_message result

rfx_encode_messages:
- check rfx_split_message result
- correct cleanup on failure
- prevent memory leak on failure

rfx_write_message_context:
- fixed invalid channelId value (must be 0xFF for WBT_CONTEXT)

rfx_process_message_codec_versions:
- fixed invalid read size of codec_version (it is 16bit)

rfx_process_message_channels:
- verify protocol conform channelId value

rfx_process_message_region:
- replaced invalid reallocs with malloc
- read and verify regionType and numTileSets from stream

rfx_process_message_tileset:
- check allocation results
- fixed incorrect usages of realloc
- ObjectPool_Take result was not checked
- fail if TS_RFX_TILE block type is not CBT_TILE
- CreateThreadpoolWork result was not checked
- post decoding loop code segfaulted in error case

setupWorkers:
- fixed incorrect usages of realloc

rfx_split_message:
- removed dead code
- missing malloc check

rfx_compose_message:
- fixed a memory leak
- check/handle rfx_encode_message result

rfx_decoder_tile_new:
- missing malloc check

rfx_message_free:
- segfault protection

rfx_write_message_tileset:
- segfault protection

### DIFF
--- a/libfreerdp/codec/rfx.c
+++ b/libfreerdp/codec/rfx.c
@@ -3,6 +3,7 @@
  * RemoteFX Codec Library
  *
  * Copyright 2011 Vic Lee
+ * Copyright 2015 Thincast Technologies GmbH
  * Copyright 2015 Norbert Federa <norbert.federa@thincast.com>
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -166,7 +167,11 @@ RFX_TILE* rfx_decoder_tile_new()
 	if (!(tile = (RFX_TILE*) calloc(1, sizeof(RFX_TILE))))
 		return NULL;
 
-	tile->data = (BYTE*) malloc(4096 * 4); /* 64x64 * 4 */
+	if (!(tile->data = (BYTE*) malloc(4 * 64 * 64))) {
+		free(tile);
+		return NULL;
+	}
+
 	tile->allocated = TRUE;
 
 	return tile;
@@ -421,7 +426,7 @@ static BOOL rfx_process_message_sync(RFX_CONTEXT* context, wStream* s)
 {
 	UINT32 magic;
 
-	context->version = 0;
+	context->decodedHeaderBlocks &= ~_RFX_DECODED_SYNC;
 
 	/* RFX_SYNC */
 	if (Stream_GetRemainingLength(s) < 6)
@@ -441,12 +446,13 @@ static BOOL rfx_process_message_sync(RFX_CONTEXT* context, wStream* s)
 
 	if (context->version != WF_VERSION_1_0)
 	{
-		WLog_ERR(TAG, "unknown version number 0x%X", context->version);
+		WLog_ERR(TAG, "invalid version number 0x%04X", context->version);
 		return FALSE;
 	}
 
 	WLog_Print(context->priv->log, WLOG_DEBUG, "version 0x%X", context->version);
 
+	context->decodedHeaderBlocks |= _RFX_DECODED_SYNC;
 	return TRUE;
 }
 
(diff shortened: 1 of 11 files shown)


cgholam:
Oko.
بتاريخ Apr 23, 2015 1:57 PM، كتبها "FreeRDP (build) Bot" <
notifications@github.com>:

libfreerdp/codec/rfx.c
```diff
@@ -162,15 +163,11 @@ RFX_TILE* rfx_decoder_tile_new()
...
-	tile = (RFX_TILE*) malloc(sizeof(RFX_TILE));
+	if (!(tile = (RFX_TILE*) calloc(1, sizeof(RFX_TILE))))
+		return NULL;
 
-	if (tile)
-	{
-		ZeroMemory(tile, sizeof(RFX_TILE));
-
-		tile->data = (BYTE*) malloc(4096 * 4); /* 64x64 * 4 */
-		tile->allocated = TRUE;
```
akallabeth: missing NULL pointer check

libfreerdp/codec/rfx.c
```diff
@@ -722,16 +781,15 @@ static BOOL rfx_process_message_tileset(RFX_CONTEXT* context, RFX_MESSAGE* messa
...
 		WLog_ERR(TAG, "no tiles.");
-		return TRUE;
+		return FALSE;
 	}
 
 	Stream_Read_UINT32(s, tilesDataSize); /* tilesDataSize (4 bytes) */
 
-	context->quants = (UINT32 *)realloc((void*) context->quants, context->numQuant * 10 * sizeof(UINT32));
-	if (!context->quants)
+	if (!(pmem = realloc((void*) context->quants, context->numQuant * 10 * sizeof(UINT32))))
```
akallabeth: Shouldn't context->quants be freed if realloc fails?

libfreerdp/codec/rfx.c
```diff
@@ -1186,21 +1327,21 @@ static BOOL computeRegion(const RFX_RECT* rects, int numRects, REGION16 *region,
...
 	RFX_CONTEXT_PRIV *priv = context->priv;
+	void* pmem;
 
 	if (!context->priv->UseThreads)
 		return TRUE;
 
-	priv->workObjects = (PTP_WORK*) realloc(priv->workObjects, sizeof(PTP_WORK) * nbTiles);
-
-	if (!priv->workObjects)
+	if (!(pmem = realloc((void*) priv->workObjects, sizeof(PTP_WORK) * nbTiles)))
```
akallabeth: Shouldn't priv->workObjects be freed, if realloc fails?

libfreerdp/codec/rfx.c
```diff
@@ -1186,21 +1327,21 @@ static BOOL computeRegion(const RFX_RECT* rects, int numRects, REGION16 *region,
...
 		return FALSE;
 
-	priv->tileWorkParams = (RFX_TILE_COMPOSE_WORK_PARAM*)
-			realloc(priv->tileWorkParams, sizeof(RFX_TILE_COMPOSE_WORK_PARAM) * nbTiles);
+	priv->workObjects = (PTP_WORK*) pmem;
 
-	if (!priv->tileWorkParams)
+	if (!(pmem = realloc((void*) priv->tileWorkParams, sizeof(RFX_TILE_COMPOSE_WORK_PARAM) * nbTiles)))
```
akallabeth: Shouldn't priv->tileWorkParams be freed, if realloc fails?

libfreerdp/codec/rfx.c
```diff
@@ -1386,25 +1526,26 @@ RFX_MESSAGE* rfx_encode_message(RFX_CONTEXT* context, const RFX_RECT* rects, int
...
 
-	if (message->numTiles != maxNbTiles)
+	success = TRUE;
+
+skip_encoding_loop:
+
+	if (success && message->numTiles != maxNbTiles)
 	{
-		message->tiles = realloc(message->tiles, sizeof(RFX_TILE*) * message->numTiles);
+		void* pmem = realloc((void*) message->tiles, sizeof(RFX_TILE*) * message->numTiles);
```
akallabeth: Shouldn't message->tiles be freed, if realloc fails?

libfreerdp/codec/rfx.c
nfedera: @akallabeth I don't see where that would be. Maybe you've looked at the negative/red/old part of the diff ?

libfreerdp/codec/rfx.c
nfedera: @akallabeth I know but since it is in the context that would be unnecessary here. We return FALSE which in turn results in a NULL rfx message delivered to the outmost caller who might either continue to use the rfx context or free it.

libfreerdp/codec/rfx.c
nfedera: @akallabeth Since it is in the context that would be unnecessary here.

libfreerdp/codec/rfx.c
nfedera: @akallabeth Since it is in the context that would be unnecessary here.

libfreerdp/codec/rfx.c
nfedera: Sure and it does get freed via rfx_message_free() below ...

libfreerdp/codec/rfx.c
```diff
@@ -162,15 +163,11 @@ RFX_TILE* rfx_decoder_tile_new()
...
 
-	if (tile)
-	{
-		ZeroMemory(tile, sizeof(RFX_TILE));
-
-		tile->data = (BYTE*) malloc(4096 * 4); /* 64x64 * 4 */
-		tile->allocated = TRUE;
-	}
+	tile->data = (BYTE*) malloc(4096 * 4); /* 64x64 * 4 */
+	tile->allocated = TRUE;
```
akallabeth: Sorry, the last comment was on the wrong place. Here you missed the malloc check ;)

libfreerdp/codec/rfx.c
```diff
@@ -481,7 +478,7 @@ static BOOL rfx_process_message_codec_versions(RFX_CONTEXT* context, wStream* s)
 
 	/* RFX_CODEC_VERSIONT */
 	Stream_Read_UINT8(s, context->codec_id); /* codecId (1 byte) */
-	Stream_Read_UINT8(s, context->codec_version); /* version (2 bytes) */
+	Stream_Read_UINT16(s, context->codec_version); /* version (2 bytes) */
```
bmiklautz: Nice catch!

nfedera:
Don't merge yet. Just found 2 other issues ....

bmiklautz:
@nfedera really nice! -  Except the malloc @akallabeth already mentioned :+1: (+1).
I really do like the way you hardened the parsing with the checks for "expectedDataBlockType" ;)

nfedera:
Thanks for the feedback. Just added another commit. IMHO that should be final now.

PR #2572, opened 2015-04-24
Added support for manpage construction for OpenBSD

bceverly:
Added support for manpage generation and installation into the correct directory on OpenBSD


bmiklautz:
@freerdp-bot test.

PR #2573, opened 2015-04-24
Fixed error on exit about pthread_mutex_destroy with waiters

bceverly:
There was a lingering error message on OpenBSD about destroying a mutex with waiters.  I have made a one-line change (ifdef'ed to just OpenBSD) to correct this.  It might be legit on other platforms but I didn't want to make that call b/c of risk.

### DIFF
--- a/client/X11/xf_tsmf.c
+++ b/client/X11/xf_tsmf.c
@@ -108,7 +108,12 @@ int xf_tsmf_xv_video_frame_event(TsmfClientContext* tsmf, TSMF_VIDEO_FRAME_EVENT
 		return -1001;
 
 	/* In case the player is minimized */
+#ifdef __OpenBSD__
+	/* numVisibleRects is unsigned */
+	if (event->x < -2048 || event->y < -2048)
+#else
 	if (event->x < -2048 || event->y < -2048 || event->numVisibleRects < 0)
+#endif
 	{
 		return -1002;
 	}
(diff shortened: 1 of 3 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@bceverly what error message did you get?

bceverly:
The error message that prints on the command line is:

pthread_mutex_destroy on mutex with waiters

It's a debug message built into OpenBSD and is meant to discourage people
from releasing code where they haven't cleaned up things waiting on a
mutex.  It's a bit annoying honestly but I'd prefer to have the message not
show up on OpenBSD for a released version of an updated FreeRDP port to
avoid people mis-reporting it as an error.

It's pretty easy to generate without the patch in that every time you run
xfreerdp you get it so:

xfreerdp --version

generates the warning on the command-line.  This seems to have fixed it and
I'm not getting any crashes in the testing that I've done.


Thanks,
Bryan

On Fri, Apr 24, 2015 at 1:40 PM, Bernhard Miklautz <notifications@github.com

akallabeth:
@bceverly This patch looks like it will hide an issue. Probably better to trace down the culprit.

akallabeth:
@bceverly Ok, have opened #2579 for a generic solution with a little explanation of why this is necessary. You can close this one now, thx for pointing this out.

PR #2577, opened 2015-04-27
Add the ability to have listener from an existing socket

hardening:
It's sometime useful to be able to have a listener created from an existing socket

### DIFF
--- a/libfreerdp/core/listener.c
+++ b/libfreerdp/core/listener.c
@@ -115,6 +115,12 @@ static BOOL freerdp_listener_open(freerdp_listener* instance, const char* bind_a
 		if ((ai->ai_family != AF_INET) && (ai->ai_family != AF_INET6))
 			continue;
 
+		if (listener->num_sockfds == MAX_LISTENER_HANDLES)
+		{
+			WLog_ERR(TAG, "too many listening sockets");
+			continue;
+		}
+
 		sockfd = socket(ai->ai_family, ai->ai_socktype, ai->ai_protocol);
 
 		if (sockfd == -1)
@@ -182,6 +188,12 @@ static BOOL freerdp_listener_open_local(freerdp_listener* instance, const char* 
 	rdpListener* listener = (rdpListener*) instance->listener;
 	HANDLE hevent;
 
+	if (listener->num_sockfds == MAX_LISTENER_HANDLES)
+	{
+		WLog_ERR(TAG, "too many listening sockets");
+		return FALSE;
+	}
+
 	sockfd = socket(AF_UNIX, SOCK_STREAM, 0);
 
 	if (sockfd == -1)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ endif()
 set(WITH_LIBRARY_VERSIONING "ON")
 set(FREERDP_VERSION_MAJOR "1")
 set(FREERDP_VERSION_MINOR "2")
-set(FREERDP_VERSION_REVISION "2")
+set(FREERDP_VERSION_REVISION "3")
```
nfedera: Master is already on revision 3 ... suggest to rebase

libfreerdp/core/listener.c
```diff
@@ -223,6 +223,27 @@ static BOOL freerdp_listener_open_local(freerdp_listener* instance, const char*
...
+
+	listener->sockfds[listener->num_sockfds] = fd;
+	listener->events[listener->num_sockfds] = CreateFileDescriptorEvent(NULL, FALSE, FALSE, fd);
+	if (!listener->events[listener->num_sockfds])
+		return FALSE;
+
+	listener->num_sockfds++;
+
+	WLog_INFO(TAG, "Listening on socket %d.", fd);
+#endif
```
nfedera: I'd suggest to return FALSE if WIN32 is defined and is missing the implementation

libfreerdp/core/listener.c
```diff
@@ -223,6 +223,27 @@ static BOOL freerdp_listener_open_local(freerdp_listener* instance, const char*
...
+	if (fcntl(fd, F_SETFL, O_NONBLOCK) < 0)
+		return FALSE;
+
+	listener->sockfds[listener->num_sockfds] = fd;
+	listener->events[listener->num_sockfds] = CreateFileDescriptorEvent(NULL, FALSE, FALSE, fd);
+	if (!listener->events[listener->num_sockfds])
+		return FALSE;
+
+	listener->num_sockfds++;
+
```
nfedera: It might be important to check listener->num_sockfds against MAX_LISTENER_HANDLES in the beginning of the function - right ?

bmiklautz:
@hardening any update on @nfedera remarks?

hardening:
@bmiklautz arg, I forgot to treat the remark. As I have copy and paste the code from other listener function I think other listeners have the problem too. I treat it today.

hardening:
So I have also added the checks for other listeners methods. And treated @nfedera 's comments.

hardening:
I think I have pushed too fast modifications in the same branch :( (that's why the build failed)

hardening:
@freerdp-bot re test please

nfedera:
+1

bmiklautz:
Late +1.

PR #2578, opened 2015-04-28
Fix issue when numWindowIds is 0

jonathankap:
I see this issue when loading a remote app that launches a command window -- numWindowIds is 0 at first then updates later to the real value.


hardening:
+1, nice catch

hardening:
@freerdp-bot test

PR #2579, opened 2015-04-28
winpr mutex unlock on destroy

akallabeth:
Since winpr is using windows style ```WaitForSingleObject``` and ```CloseHandle``` the ```pthread_mutex``` objects are usually locked when destroyed. (which is valid behavior with ```CloseHandle```)
```pthread_mutex``` objects must be unlocked on destruction.

This pull uses ```pthread_mutex_trylock``` to ensure the mutex is locked (optionally with a debug stacktrace) and finally unlocks the mutex with ```pthread_mutex_unlock``` before destroying it.

Superseeds #2573

### CODE REVIEW

**Comment by akallabeth** [comment]
Thx to @bceverly for pointing out this issue.

**Comment by bceverly** [comment]
I'll pull things down and test on OpenBSD.  Thanks!


Thanks,
Bryan

**Comment by bceverly** [comment]
Worked perfectly!  Thanks @akallabeth for doing this!


Thanks,
Bryan

File: winpr/libwinpr/synch/mutex.c [hunk 1]
```diff
@@ -57,12 +65,53 @@ static int MutexGetFd(HANDLE handle)
...
 
 	if (!MutexIsHandled(handle))
 		return FALSE;
 
-	if (!pthread_mutex_destroy(&mutex->mutex))
+	rc = pthread_mutex_trylock(&mutex->mutex);
+	switch(rc)
+	{
+		/* If we already own the mutex consider it a success. */
+		case EDEADLK:
```
nfedera: You need to add a case EBUSY which also indicates that it is already locked.
Try running ```./Testing/TestSynch TestSynchMutex``` as proof.

nfedera:
After adding EBUSY as commented above +1. LGTM

akallabeth:
@nfedera Thx, forgot to push that commit.

PR #2582, opened 2015-04-28
Fix unchecked CreateEvent calls and misc fixes

nfedera:
1)
Added missing checks for CreateEvent which also required the following related changes:
- changed freerdp_context_new API to BOOL
- changed freerdp_peer_context_new API to BOOL
- changed pRdpClientNew callback to BOOL
- changed pContextNew callback to BOOL
- changed psPeerAccepted callback to BOOL
- changed psPeerContextNew callback to BOOL

2)
Fixed lots of missing alloc and error checks in the changed code's neighbourhood.

3)
Check freerdp_client_codecs_prepare result to avoid segfaults caused by using non-initialized codecs.

4)
Fixed deadlocks in x11 caused by missing xf_unlock_x11() calls in some error handlers

5)
Some fixes in thread pool:
- DEFAULT_POOL assignment did not match TP_POOL definition
- don't free the pool pointer if it points to the static DEFAULT_POOL
- added error handling and cleanup in InitializeThreadpool


client/Windows/wf_cliprdr.c
```diff
@@ -1975,41 +1975,70 @@ void wf_cliprdr_init(wfContext* wfc, CliprdrClientContext* cliprdr)
...
+	free(clipboard->file_names);
+	clipboard->file_names = NULL;
+fail_file_names:
+	free(clipboard->format_mappings);
+	clipboard->format_mappings = NULL;
+fail_format_mappings:
+	free(wfc->clipboard);
+	wfc->clipboard = NULL;
+
+	return;
```
hardening: Perhaps the function could return a BOOL and indicate failure, with callers checking for the failure...

hardening:
I think the release number should be changed once again as some public functions have different signatures.

client/Windows/wf_cliprdr.c
nfedera: @hardening of course I thought about that but in that case it is completely useless if you look at the calling code ... unless rewriting the windows client which is not what this PR is about ;)

nfedera:
@hardening thanks, api revision is increased now.

MartinHaimberger:
+1 
LGTM

PR #2583, opened 2015-04-28
Two minor source code changes to remove warnings

bceverly:
Fixed linker warning about insecure crypto and fixed compiler warning about unsigned being compared to less than zero.

Both of these are probably safe to have hit all builds if someone wants to do a little testing and remove the #ifdef's.  I didn't want to make that assumption for everyone else however so I played it safe.

### DIFF
--- a/client/X11/xf_tsmf.c
+++ b/client/X11/xf_tsmf.c
@@ -108,7 +108,11 @@ int xf_tsmf_xv_video_frame_event(TsmfClientContext* tsmf, TSMF_VIDEO_FRAME_EVENT
 		return -1001;
 
 	/* In case the player is minimized */
+#ifdef __OpenBSD__
+	if (event->x < -2048 || event->y < -2048)
+#else
 	if (event->x < -2048 || event->y < -2048 || event->numVisibleRects < 0)
+#endif
 	{
 		return -1002;
 	}
--- a/client/X11/xf_tsmf.c
+++ b/client/X11/xf_tsmf.c
@@ -108,11 +108,7 @@ int xf_tsmf_xv_video_frame_event(TsmfClientContext* tsmf, TSMF_VIDEO_FRAME_EVENT
 		return -1001;
 
 	/* In case the player is minimized */
-#ifdef __OpenBSD__
-	if (event->x < -2048 || event->y < -2048)
-#else
-	if (event->x < -2048 || event->y < -2048 || event->numVisibleRects < 0)
-#endif
+	if (event->x < -2048 || event->y < -2048 || event->numVisibleRects == 0)
 	{
 		return -1002;
 	}
--- a/libfreerdp/core/autodetect.c
+++ b/libfreerdp/core/autodetect.c
@@ -144,6 +144,7 @@ BOOL autodetect_send_bandwidth_measure_payload(rdpContext* context, UINT16 paylo
 {
 	wStream* s;
 	UCHAR *buffer = NULL;
+	BOOL bResult = FALSE;
 
 	s = rdp_message_channel_pdu_init(context->rdp);
 
@@ -170,19 +171,27 @@ BOOL autodetect_send_bandwidth_measure_payload(rdpContext* context, UINT16 paylo
 	buffer = (UCHAR *)malloc(payloadLength);
 	if (NULL == buffer)
 	{
+		Stream_Release(s);
 		return FALSE;
 	}
 
 	BCryptGenRandom(NULL, buffer, payloadLength, 0L);
 	Stream_Write(s, buffer, payloadLength);
 
-	return rdp_send_message_channel_pdu(context->rdp, s, SEC_AUTODETECT_REQ);
+	bResult = rdp_send_message_channel_pdu(context->rdp, s, SEC_AUTODETECT_REQ);
+	if (!bResult)
+	{
+		Stream_Release(s);
+	}
+
+	return bResult;
 }
 
 static BOOL autodetect_send_bandwidth_measure_stop(rdpContext* context, UINT16 payloadLength, UINT16 sequenceNumber, UINT16 requestType)
 {
 	wStream* s;
 	UCHAR *buffer = NULL;
+	BOOL bResult = FALSE;
 
 	s = rdp_message_channel_pdu_init(context->rdp);
 
@@ -213,6 +222,7 @@ static BOOL autodetect_send_bandwidth_measure_stop(rdpContext* context, UINT16 p
 			buffer = malloc(payloadLength);
 			if (NULL == buffer)
 			{
+				Stream_Release(s);
 				return FALSE;
 			}
 
@@ -221,7 +231,13 @@ static BOOL autodetect_send_bandwidth_measure_stop(rdpContext* context, UINT16 p
 		}
 	}
 
-	return rdp_send_message_channel_pdu(context->rdp, s, SEC_AUTODETECT_REQ);
+	bResult = rdp_send_message_channel_pdu(context->rdp, s, SEC_AUTODETECT_REQ);
+	if (!bResult)
+	{
+		Stream_Release(s);
+	}
+
+	return bResult;
 }
 
 static BOOL autodetect_send_continuous_bandwidth_measure_stop(rdpContext* context, UINT16 sequenceNumber)
--- a/libfreerdp/core/autodetect.c
+++ b/libfreerdp/core/autodetect.c
@@ -17,8 +17,6 @@
  * limitations under the License.
  */
 
-#include <winpr/bcrypt.h>
-
 #ifdef HAVE_CONFIG_H
 #include "config.h"
 #endif
--- a/libfreerdp/core/autodetect.c
+++ b/libfreerdp/core/autodetect.c
@@ -166,7 +166,11 @@ BOOL autodetect_send_bandwidth_measure_payload(rdpContext* context, UINT16 paylo
 	/* Random data (better measurement in case the line is compressed) */
 	for (i = 0; i < payloadLength / 4; i++)
 	{
+#ifdef __OpenBSD__
+		Stream_Write_UINT32(s, arc4random());
+#else
 		Stream_Write_UINT32(s, rand());
+#endif
 	}
 
 	return rdp_send_message_channel_pdu(context->rdp, s, SEC_AUTODETECT_REQ);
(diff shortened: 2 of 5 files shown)


client/X11/xf_tsmf.c
```diff
@@ -108,7 +108,11 @@ int xf_tsmf_xv_video_frame_event(TsmfClientContext* tsmf, TSMF_VIDEO_FRAME_EVENT
 		return -1001;
 
 	/* In case the player is minimized */
+#ifdef __OpenBSD__
```
hardening: @bceverly can you explain what is specific to OpenBSD here ?
if the check on event->numVisibleRects is useless (which seems to be the case), let's just remove it for all platforms...

bceverly: I'm totally cool with that (and would prefer it) - I just wasn't
comfortable making the call on behalf of everyone else since I am
relatively new here.

My fear is that this was hiding a bug.  Perhaps it needed to be tested to
be == 0 instead of < 0 so I wanted some more experienced eyes to take a
look at it.


Thanks,
Bryan

libfreerdp/core/autodetect.c
```diff
@@ -166,7 +166,11 @@ BOOL autodetect_send_bandwidth_measure_payload(rdpContext* context, UINT16 paylo
 	/* Random data (better measurement in case the line is compressed) */
 	for (i = 0; i < payloadLength / 4; i++)
 	{
+#ifdef __OpenBSD__
+		Stream_Write_UINT32(s, arc4random());
+#else
 		Stream_Write_UINT32(s, rand());
```
akallabeth: Probably a good idea to use ```CryptGenRandom``` from WinPR and do a proper implementation there.

libfreerdp/core/autodetect.c
```diff
@@ -204,7 +208,11 @@ static BOOL autodetect_send_bandwidth_measure_stop(rdpContext* context, UINT16 p
 			/* Random data (better measurement in case the line is compressed) */
 			for (i = 0; i < payloadLength / 4; i++)
 			{
+#ifdef __OpenBSD__
+				Stream_Write_UINT32(s, arc4random());
+#else
```
akallabeth: Probably a good idea to use ```CryptGenRandom``` from WinPR and do a proper implementation there.

libfreerdp/core/autodetect.c
```diff
@@ -166,7 +166,11 @@ BOOL autodetect_send_bandwidth_measure_payload(rdpContext* context, UINT16 paylo
 	/* Random data (better measurement in case the line is compressed) */
 	for (i = 0; i < payloadLength / 4; i++)
 	{
+#ifdef __OpenBSD__
+		Stream_Write_UINT32(s, arc4random());
```
akallabeth: Probably a good idea to use ```CryptGenRandom``` from WinPR and do a proper implementation there.

bceverly:
OK.  I implemented the feedback in the following way:

1.  For the xf_tsmf.c code, I made a platform-neutral change to look for zero visible rectangles as I think that is more in the spirit of the code comment about what it is doing.  Please review this one carefully because if zero is a legitimate value, I just introduced a bug.

2.  For the arc4random() change, I implemented the BCryptGenerateRandom() function in winpr for all platforms and then leveraged it in the two places where rand() was previously being called.  I removed the loop in the two spots and let the BCryptGenerateRandom() function fill a buffer that was passed in.

Anything further or should this be merged?

Thanks for all the feedback!  I'm getting way more comfortable with the code.

libfreerdp/core/autodetect.c
```diff
@@ -201,11 +203,11 @@ static BOOL autodetect_send_bandwidth_measure_stop(rdpContext* context, UINT16 p
 				Stream_Release(s);
 				return FALSE;
 			}
+
 			/* Random data (better measurement in case the line is compressed) */
-			for (i = 0; i < payloadLength / 4; i++)
-			{
-				Stream_Write_UINT32(s, rand());
-			}
+			buffer = malloc(payloadLength);
```
akallabeth: unchecked malloc

libfreerdp/core/autodetect.c
```diff
@@ -163,19 +165,19 @@ BOOL autodetect_send_bandwidth_measure_payload(rdpContext* context, UINT16 paylo
 	Stream_Write_UINT16(s, sequenceNumber); /* sequenceNumber (2 bytes) */
 	Stream_Write_UINT16(s, RDP_BW_PAYLOAD_REQUEST_TYPE); /* requestType (2 bytes) */
 	Stream_Write_UINT16(s, payloadLength); /* payloadLength (2 bytes) */
+
 	/* Random data (better measurement in case the line is compressed) */
-	for (i = 0; i < payloadLength / 4; i++)
-	{
-		Stream_Write_UINT32(s, rand());
-	}
+	buffer = (UCHAR *)malloc(payloadLength);
```
akallabeth: unchecked malloc

akallabeth:
Except my two comments +1

winpr/libwinpr/bcrypt/bcrypt.c
```diff
@@ -68,6 +68,36 @@ NTSTATUS BCryptFinishHash(BCRYPT_HASH_HANDLE hHash, PUCHAR pbOutput, ULONG cbOut
...
+		return -1;
+	}
+
+	/* Loop through the buffer for the number of bytes specified, filling it up. */
+	for (i=0 ; i< cbBuffer ; i++)
+	{
+#ifdef __OpenBSD__
+		random_long = arc4random();
+#else
+		random_long = rand();
```
akallabeth: @bceverly maybe a good idea to also try ```/dev/random``` resp ```/dev/urandom```

bceverly:
Good catch @akallabeth!  I have pushed up another commit that addresses this.  Please let me know if you want me to make any other changes.

libfreerdp/core/autodetect.c
```diff
@@ -201,11 +208,16 @@ static BOOL autodetect_send_bandwidth_measure_stop(rdpContext* context, UINT16 p
...
 				return FALSE;
 			}
+
 			/* Random data (better measurement in case the line is compressed) */
-			for (i = 0; i < payloadLength / 4; i++)
+			buffer = malloc(payloadLength);
+			if (NULL == buffer)
 			{
-				Stream_Write_UINT32(s, rand());
+				return FALSE;
```
akallabeth: ```Stream_Release(s)```, leaking stream here...

libfreerdp/core/autodetect.c
```diff
@@ -163,19 +165,24 @@ BOOL autodetect_send_bandwidth_measure_payload(rdpContext* context, UINT16 paylo
...
 	Stream_Write_UINT16(s, RDP_BW_PAYLOAD_REQUEST_TYPE); /* requestType (2 bytes) */
 	Stream_Write_UINT16(s, payloadLength); /* payloadLength (2 bytes) */
+
 	/* Random data (better measurement in case the line is compressed) */
-	for (i = 0; i < payloadLength / 4; i++)
+	buffer = (UCHAR *)malloc(payloadLength);
+	if (NULL == buffer)
 	{
-		Stream_Write_UINT32(s, rand());
+		return FALSE;
```
akallabeth: Leaking stream here

libfreerdp/core/autodetect.c
```diff
@@ -163,19 +165,24 @@ BOOL autodetect_send_bandwidth_measure_payload(rdpContext* context, UINT16 paylo
...
 	}
 
+	BCryptGenRandom(NULL, buffer, payloadLength, 0L);
+	Stream_Write(s, buffer, payloadLength);
+
 	return rdp_send_message_channel_pdu(context->rdp, s, SEC_AUTODETECT_REQ);
```
akallabeth: When ```rdp_send_message_channel_pdu``` fails, the stream is not released

bceverly:
I believe I have fixed the stream leaks.  If someone could please review and let me know if there is anything else that I need to do before this is merged, I'd appreciate it.

:-)

bceverly:
Actually now that I thinking about it @akallabeth should I be freeing that buffer I malloc'ed before I return from the function?

akallabeth:
@bceverly Yes, propably a good idea ^^

bceverly:
OK.  Added the code to clean up the memory buffer @akallabeth .  Are we good to merge?

akallabeth:
@bceverly Code looks good now.
It would propably be better to use a stronger random implementation if available. ```getentropy``` on openbsd, ```getrandom``` on linux, ```CryptGenRandom ``` on windows and ```/dev/[u]random``` on other unix like systems come to mind.

akallabeth:
@bceverly or for the matter, since freerdp depends on openssl, just use the openssl random engine.

hardening:
Yep RAND_bytes() is definitely the one to use as we don't need platform specific tricks

bceverly:
OK.  I think I've addressed the points raised by @akallabeth and @hardening with this last commit.  Please take one last look at it and, if we are good, let's merge it.

Thanks again everyone!

bceverly:
Hey everyone.  Are we good to merge this one?

akallabeth:
Looks ok for me, +1

bceverly:
Anyone else?  If not, can we please commit?

awakecoding:
@freerdp-bot test.

bmiklautz:
@bceverly just verified locally. The latest version of your fixes breaks tests.

bceverly:
Bernhard,

Do you think it is the thread ID stuff?  If so, let me back it out and see
if that corrects the issue.


Thanks,
Bryan

bmiklautz:
@bceverly I think so but I haven't had the time to check that in detail. But I'll have a look later on.

bceverly:
OK - I pushed another version that puts the code inside of an #ifdef
__OpenBSD__ section.  If you could please re-test I would very much
appreciate it.

Thanks.


Thanks,
Bryan

bmiklautz:
@bceverly with your last fix the test works again. However I'm fairly sure that the real problem here is the && instead of & when you mask the pthread id.

bceverly:
doh!  Thanks.  Wow.  Let me fix it right...


Thanks,
Bryan

bceverly:
Just pushed it.  If you could please retest when you get a chance Bernhard
I would appreciate it.

How do you do an embarrassed emoticon again?  :-)


Thanks,
Bryan

bmiklautz:
@freerdp-bot test

bmiklautz:
+1

bceverly:
Thanks!


Thanks,
Bryan

PR #2584, opened 2015-04-29
Fix build with newer uclibc versions

baruchsiach:
uClibc master branch, as well as uclibc-ng add eventfd_{read,write}
definitions. Instead of testing for __UCLIBC__, have cmake explicitly check
for the existence of eventfd_read and save the result in
WITH_EVENTFD_READ_WRITE.

Fixes build errors like:

.../winpr/libwinpr/synch/event.c:120:12: error: static declaration of 'eventfd_read' follows non-static declaration
 static int eventfd_read(int fd, eventfd_t* value)
            ^
In file included from .../winpr/libwinpr/synch/event.c:39:0:
.../usr/include/sys/eventfd.h:37:12: note: previous declaration of 'eventfd_read' was here
 extern int eventfd_read (int __fd, eventfd_t *__value);
            ^
.../winpr/libwinpr/synch/event.c:125:12: error: static declaration of 'eventfd_write' follows non-static declaration
 static int eventfd_write(int fd, eventfd_t value)
            ^
In file included from .../winpr/libwinpr/synch/event.c:39:0:
.../usr/include/sys/eventfd.h:40:12: note: previous declaration of 'eventfd_write' was here
 extern int eventfd_write (int __fd, eventfd_t __value);


bmiklautz:
+1

PR #2587, opened 2015-05-01
Remove the ARM_FP_ABI option

### LINKED ISSUE #2586: Compiler flags conflict due to ARM_FP_API option
A user reported the following build failure on Gentoo Linux.

https://bugs.gentoo.org/show_bug.cgi?id=547246

```
ld: error: libfreerdp.so.1.2.1 uses VFP register arguments, CMakeFiles/freerdp.dir/codec/rfx_neon.c.o does not
```

This seems to be the result of some source files being built with `-mfloat-abi=softfp`, and others with `-mfloat-abi=hard`. It looks like this strange behavior was introduced in response to #980.

It seems to be a problem that will occur whenever CFLAGS contains `-mfloat-abi=hard` and ARM_FP_ABI is unspecified, causing it to default to softfp. It doesn't really make sense to me to have cmake override CFLAGS from the environment.

I see a couple of solutions here:

1. Remove the cmake logic for ARM_FP_ABI and adds -mfloat-abi to COMPILER_FLAGS. I'm really not sure why that resolved #980 in the first place. It seems like that issue was a result of CFLAGS being set incorrectly in the build environment, or possibly CFLAGS not being respected in that old version of the code.

2. Add some option to make the above mentioned behavior optional.

I can certainly submit a pull request if some some decision can be made on a suitable solution.

floppym:
It's unclear why this option would be necessary, and it causes problems
when people do not match it to their toolchain and CFLAGS.

To set the float abi, either use a toolchain with an appropriate default
or set the float-abi option in the CFLAGS environment variable.

This should resolve #2586.


bmiklautz:
@freerdp-bot test

bmiklautz:
@floppym nice you also updated the package \o/

bmiklautz:
@floppym very nice thank you!
From me basically +1.

PR #2588, opened 2015-05-03
Fix bad color RGB/BGR

anlhord:
Fix virtualbox RDP is displayed in bad color.

### DIFF
--- a/client/Wayland/wlfreerdp.c
+++ b/client/Wayland/wlfreerdp.c
@@ -74,10 +74,19 @@ static BOOL wl_end_paint(rdpContext* context)
 	display = context_w->display;
 	window = context_w->window;
 
-	for (i = 0; i < h; i++)
+	for (i = 0; i < h; i++) {
 		memcpy(window->data + ((i+y)*(gdi->width*4)) + x*4,
-		       gdi->primary_buffer + ((i+y)*(gdi->width*4)) + x*4,
-		       w*4);
+			gdi->primary_buffer + ((i+y)*(gdi->width*4)) + x*4,
+			w*4);
+		char *dta = window->data + ((i+y)*(gdi->width*4)) + x*4;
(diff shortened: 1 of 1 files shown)


anlhord:
It builds. I've just tested in on wayland and x11/weston.

client/Wayland/wlfreerdp.c
```diff
@@ -74,10 +74,19 @@ static BOOL wl_end_paint(rdpContext* context)
 	display = context_w->display;
 	window = context_w->window;
 
-	for (i = 0; i < h; i++)
+	for (i = 0; i < h; i++) {
```
bmiklautz: Please note that this coding style isn't common for FreeRDP. The bracket should be on the next line.

bmiklautz:
@freerdp-bot test

bmiklautz:
@anlhord did you test your changes against any other server than VirtualBox RDP?

hardening:
I confirm bug and correction
Just curious: don't we have function to perform that kind of operation in FreeRDP ?

bmiklautz:
@hardening we do. I didn't look at the code yet but I would assume that gdi is initialized with the wrong color format?

hardening:
@anlhord Have you tried initializing gdi without CLRCONV_INVERT ?

hardening:
@anlhord I've just done a test and removing CLRCONV_INVERT produce the expected output

hardening:
I'm closing this PR as I think your proposal is not the "right way to do it"(TM).

anlhord:
Can I ask:
1. how to fix this
2. how to also implement keyboard/mouse/touch input (preferably without threads since thats hard for me to debug)

hardening:
@anlhord except if I missed something, the color inversion bug is fixed on master. The input bug is something else.

PR #2589, opened 2015-05-03
Don't use unavailable C99 long double math functions

baruchsiach:
uClibc variants do not provide the C99 long double math functions like ceill,
powl, etc.. For future compatibility use check_symbol_exists() to check
whether these functions are available, and keep the result in
HAVE_MATH_C99_LONG_DOUBLE. Use that instead of the fragile Cygwin version
check in triodef.h.

Fixes build failures under uClibc(-ng) like:

../../libwinpr/utils/libwinpr-utils.so.0.1.0: undefined reference to `powl'
../../libwinpr/utils/libwinpr-utils.so.0.1.0: undefined reference to `fmodl'
../../libwinpr/utils/libwinpr-utils.so.0.1.0: undefined reference to `ceill'
../../libwinpr/utils/libwinpr-utils.so.0.1.0: undefined reference to `log10l'
../../libwinpr/utils/libwinpr-utils.so.0.1.0: undefined reference to `floorl'
collect2: error: ld returned 1 exit status

Signed-off-by: Baruch Siach <baruch@tkos.co.il>


bmiklautz:
@freerdp-bot test

PR #2594, opened 2015-05-05
Don't disable C99 long double math on non Unix platforms

baruchsiach:
CMakeLists.txt may enable HAVE_MATH_C99_LONG_DOUBLE only on Unix platforms.
Limit the effect of this symbol accordingly.

### DIFF
--- a/winpr/libwinpr/utils/trio/triodef.h
+++ b/winpr/libwinpr/utils/trio/triodef.h
@@ -313,7 +313,7 @@ typedef void * trio_pointer_t;
 # define TRIO_COMPILER_SUPPORTS_LL
 #endif
 
-#if !defined(HAVE_MATH_C99_LONG_DOUBLE)
+#if defined(TRIO_PLATFORM_UNIX) && !defined(HAVE_MATH_C99_LONG_DOUBLE)
 # define TRIO_NO_FLOORL 1
 # define TRIO_NO_CEILL 1
 # define TRIO_NO_POWL 1


akallabeth:
@baruchsiach Shouldn't the problem be fixed in CMake, aka where the detection fails set the ```HAVE_MATH_C99_LONG_DOUBLE``` manually / special check?

baruchsiach:
@akallabeth The complicated `#ifdef` in `winpr/libwinpr/utils/trio/trio.c` indicates that we can't just assume that those long double math functions are available on other platforms. I just don't know what the right test sequence for these platforms is. This PR leaves the current status of non Unix platforms as is. Of course having direct symbol test for each platform would be preferable.

bmiklautz:
@freerdp-bot test

bmiklautz:
@akallabeth @baruchsiach since the pull request that introduced the HAVE_MATH_C99_LONG_DOUBLE build successfully on all platforms we do build tests for it should be fine for windows which is the only non nix platform. 
I would think the change in this pull request isn't necessary since check_symbol_exists should do it's job on all patforms. As for the complicated ifdefs trio was integrated and didn't use cmake before but not all tests/ifdefs weren't adapted.

baruchsiach:
@bmiklautz Build would succeed event when `HAVE_MATH_C99_LONG_DOUBLE` is not defined. That would currently be the case for any non Unix platform. The thing is that on these platforms `trio.c` now degrades trio_[func] to the non long double version of func. This PR attempts to fix that, and restore the previous behavior.

bmiklautz:
@baruchsiach agreed ;)
+1

PR #2597, opened 2015-05-06
Fix bitstream reading bounds check

clivest:
The bounds checks where previously incorrect causing the macros to read off the end of the array.

Caught with valgrind when investigating another issue.


mfleisz:
@freerdp-bot test
1+ good catch

mfleisz:
@freerdp-bot test

akallabeth:
+1

nfedera:
nice catch!

PR #2599, opened 2015-05-06
Fix colors with the wayland client

hardening:
I think this is a better way to fix it than https://github.com/FreeRDP/FreeRDP/pull/2588


nfedera:
+1, much better ;)

bmiklautz:
Nice!

PR #2600, opened 2015-05-07
h264: set the encoder MaxBitrate field.

llyzs:


llyzs:
Don't merge - some additional changes required.

MartinHaimberger:
@llyzs I dont know if this change is a good idea. after some research this bitrate is used for frame skipping (if encoded frame exeeds this rate, the frame will get dropped) UNSPECIFIED_BIT_RATE disables this behavior.

llyzs:
@MartinHaimberger 

There is another field bEnableFrameSkip set to 0, does that work?

MartinHaimberger:
@llyzs  We use bEnableFrameSkip with 0 and UNSPECIFIED_BIT_RATE, with this settings we got what we wanted ....

llyzs:
@MartinHaimberger 

I think I will just expose a new field instead of setting it to the same as BitRate so that we can tune it outside of the freerdp core.

PR #2601, opened 2015-05-07
Fix some unchecked create calls

nfedera:


server/shadow/shadow_server.c
```diff
@@ -434,16 +434,25 @@ int shadow_server_init_config_path(rdpShadowServer* server)
 
 		if (userLibraryPath)
 		{
-			if (!PathFileExistsA(userLibraryPath))
-				CreateDirectoryA(userLibraryPath, 0);
+			if (!PathFileExistsA(userLibraryPath) &&
+				!CreateDirectoryA(userLibraryPath, 0))
+			{
+				free(userLibraryPath);
```
hardening: Perhaps an error message would be nice here

hardening:
+1

server/shadow/shadow_server.c
nfedera: @hardening done

PR #2602, opened 2015-05-07
h264: add new encoder setting MaxBitRate.

llyzs:


libfreerdp/codec/h264.c
```diff
@@ -182,7 +182,7 @@ static int openh264_compress(H264_CONTEXT* h264, BYTE** ppDstData, UINT32* pDstS
 		sys->EncParamExt.iTargetBitrate = h264->BitRate;
 		sys->EncParamExt.iRCMode = RC_BITRATE_MODE;
 		sys->EncParamExt.fMaxFrameRate = h264->FrameRate;
-		sys->EncParamExt.iMaxBitrate = UNSPECIFIED_BIT_RATE;
+		sys->EncParamExt.iMaxBitrate = h264->MaxBitRate;
```
hardening: Just to be sure, h264->MaxBitRate is initialized with UNSPECIFIED_BIT_RATE ?

MartinHaimberger:
@llyzs As mentioned in #2600, the iMaxBitrate should be UNSPECIFIED_BIT_RATE, otherwise if the encoded frame is bigger than this value, the frame will get skipped. I think the value UNSPECIFIED_BIT_RATE was just fine.

MartinHaimberger:
@llyzs if a frame gets skipped, the result (for example in a terminal if u wanna see the last command, the command may not be displayed if this frame got skipped) is that u dont see the changes on the client. But we need to send a frame we encode in any case, even its too big, then we wait a few frames (future bucket use) before we encode the next frame.

llyzs:
The iMaxBitrate has UNSPECIFIED_BIT_RATE(0) as default value, however when I set it to a larger value, my test result is that as long as I keep bEnableFrameSkip as 0, openh264 never drops a frame even the generated frame is very large. In fact our frontend does NOT allow frame drops.

However this seems to have no effect either. openh264 does not seem to try even a little harder to further compress the image. I set both iBitrate and iMaxBitrate to 60Mbps and use 30fps, the encoded result of a series of random pixel frame is 7xxK bytes per frame - that is 200Mbps already.

PR #2604, opened 2015-05-07
Fix remoteApps

hardening:
The assert that have been removed were preventing RAIL from working when you have an intermediary screen before seeing the application.


nfedera:
+1

PR #2605, opened 2015-05-08
fixed multiple missing gdi return value checks

nfedera:
This PR primarily targets gdi_Create* functions


MartinHaimberger:
+1 LGTM

hardening:
I would have +1

PR #2608, opened 2015-05-11
Add check to protect memcpy(3) from using NULL pointer.

oshogbo:
The ntlm_construct_challenge_target_info function can potentially pass NULL as
argument to the ntlm_av_pair_add function (for example DnsDomainName.Buffer).
This NULL finally lands in the CopyMemory (which is macro to the memcpy(3)
function) which can't handle NULL.

### DIFF
--- a/winpr/libwinpr/sspi/NTLM/ntlm_av_pairs.c
+++ b/winpr/libwinpr/sspi/NTLM/ntlm_av_pairs.c
@@ -21,6 +21,8 @@
 #include "config.h"
 #endif
 
+#include <assert.h>
+
 #include "ntlm.h"
 #include "../sspi.h"
 
@@ -143,6 +145,7 @@ NTLM_AV_PAIR* ntlm_av_pair_add(NTLM_AV_PAIR* pAvPairList, NTLM_AV_ID AvId, PBYTE
 	if (!pAvPair)
 		return NULL;
 
+	assert(Value != NULL);
 	pAvPair->AvId = AvId;
 	pAvPair->AvLen = AvLen;
 	CopyMemory(ntlm_av_pair_get_value_pointer(pAvPair), Value, AvLen);


bmiklautz:
@freerdp-bot test.

bmiklautz:
@oshogbo  For later reference -  in which context did you find this problem?

oshogbo:
@bmiklautz It seems that real problem was here: #2613.

hardening:
@oshogbo so we can close it can't we ?

oshogbo:
@hardening Good question.
It's depend about FreeRDP politic. Do FreeRDP want to handle boundary cases?
- If not we can close it.
- If FreeRDP want it should be merged.
In feature developing passing a null values to the memcpy(3) could cause more problems.

But its depend about politic and I will be happy with any decision. :)

hardening:
@oshogbo I think the question is more "where do we have to put the checks". ntlm_av_pair_add is clearly not supposed to be called with a NULL value, so either we treat the callers (which has been done by https://github.com/FreeRDP/FreeRDP/pull/2613), or we must do something more agressive than silently returning as in your patch (assert or close the connection)

oshogbo:
@Hardening: I'm not a big fun of supposing that otter program will not do
somthing. In my opinion bather assumption is he could do it.

Assertion is the best idea, but not sure if FreeRDP is using it.

On Mon, May 18, 2015 at 02:29:43PM -0700, Hardening wrote:
> @oshogbo I think the question is more "where do we have to put the checks". ntlm_av_pair_add is clearly not supposed to be called with a NULL value, so either we treat the callers (which has been done by https://github.com/FreeRDP/FreeRDP/pull/2613), or we must do something more agressive than silently returning as in your patch (assert or close the connection)
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/FreeRDP/FreeRDP/pull/2608#issuecomment-103217815

--

oshogbo:
@hardening I change my commit to use assert.

Issue #2610, opened 2015-05-11
Async transport thread consumes 100% of one CPU core 


nfedera:
Seen with current master (692df12) with X11 and Mac client

nfedera:
ok ... the code in the transport_client_thread is just sooooo wrong.

nfedera:
Should be fixed via #2611 and #2621

Issue #2609, opened 2015-05-11
Mac OS X native client : 100% CPU usage even when session is idle


nfedera:
Seen with current master (692df124c60ffbfdd51cd576bc3c7cf2735b712a) on OS X 10.9.5.
Does not seem to be related to a specific codec ( /rfx /gfx ...)

shattars3d:
hey I'm having trouble compiling the mac OS X version for the rdp client.. can i get some help off u?. thanks..

nfedera:
Ok the issue is related to +async-transport which is hardcoded in the mac native client.
It also happens with ```xfreerdp /v:host +async-transport```

nfedera:
Created related/root issue #2610

bmiklautz:
@shattars3d please only add comments related to the issues problem. Your build problems are not related to this in any way. Thank you!

nfedera:
Should be fixed via #2611 and #2621

PR #2611, opened 2015-05-11
transport: fixed multiple errors in async thread

nfedera:
Should fix issue #2610 and issue #2609

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -873,6 +873,7 @@ BOOL transport_disconnect(rdpTransport* transport)
 
 static void* transport_client_thread(void* arg)
 {
+	DWORD dwExitCode = 0;
 	DWORD status;
 	DWORD nCount;
 	HANDLE handles[64];
@@ -880,7 +881,7 @@ static void* transport_client_thread(void* arg)
 	rdpContext* context = transport->context;
 	rdpRdp* rdp = context->rdp;
 
-	WLog_DBG(TAG, "Starting transport thread");
+	WLog_DBG(TAG, "Asynchronous transport thread started");
 
 	nCount = 0;
 	handles[nCount++] = transport->stopEvent;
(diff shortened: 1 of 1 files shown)


bmiklautz:
+1.
The part with WAIT_TIMEOUT is quite interesting maybe we can reproduce this somehow.

hardening:
like @bmiklautz,  +1 but the WAIT_TIMEOUT part looks suspicious (sounds like SSL_ERROR_SYSCALL ;) )

nfedera:
It seems to be related to poll(). If I comment ```#define HAVE_POLL_H``` the issue is not present.
BTW: I'm seeing this on Mac OS X 10.9.5 ... still have to check if this also happens on Linux.
For reference it is this poll - https://github.com/FreeRDP/FreeRDP/blob/e3236c2317abf88bb8634f493164367d4708cec9/winpr/libwinpr/synch/wait.c#L154 - which returns 0 although timeout is -1;

nfedera:
Adding my debug output for reference:
```
[09:43:07:288] [24810:09d4d000] [ERROR][com.winpr.sync.wait] - calling poll with timeout = -1
[09:43:07:288] [24810:09c47000] [ERROR][com.winpr.sync.wait] - poll returned 0, dumping pollfds ...
[09:43:07:288] [24810:09c47000] [ERROR][com.winpr.sync.wait] - pollfds[0].fd = 11
[09:43:07:288] [24810:09c47000] [ERROR][com.winpr.sync.wait] - pollfds[0].events = 0x0001 (POLLIN)
[09:43:07:288] [24810:09c47000] [ERROR][com.winpr.sync.wait] - pollfds[0].revents = 0
[09:43:07:288] [24810:09c47000] [ERROR][com.winpr.sync.wait] - pollfds[1].fd = 13
[09:43:07:288] [24810:09c47000] [ERROR][com.winpr.sync.wait] - pollfds[1].events = 0x0001 (POLLIN)
[09:43:07:288] [24810:09c47000] [ERROR][com.winpr.sync.wait] - pollfds[1].revents = 0
```

nfedera:
Weird. I don't see the poll issue on Linux and neither on my iMac which is also still on OS X 10.9.5 like my MacBook where I've seen the issue first.
Will do some further investigation and update this PR to not accept WAIT_TIMEOUT ...

nfedera:
Ok it seems the MacPorts installation is the culprit. otool -L revealed that some libraries (e.g. ssl) get loaded from /opt/local. Maybe the poll symbol gets overridden somewhere which might be quite possible if you look at he history of [broken poll issues on OS X](https://www.google.com/search?q=broken+poll+OS+X).
In addition this line https://github.com/FreeRDP/FreeRDP/blob/f219d341f1f88159cf67578390390a554a4b4083/CMakeLists.txt#L360  causes the issue that the openssl includes are taken from /opt/local/include but freerdp gets linked to the system libraries which might result in build or runtime failures of course. Not sure how to handle that ... @bmiklautz any ideas ?

However, IMHO the correct way is to fail when WaitForMultipleObjects returns WAIT_TIMEOUT if an infinite timeout was specified - contrary to what I've done in this PR.

nfedera:
Update: There are link and runtime issues with MacPorts installed caused by CMakeList.txt addition of /opt/local/include to the head of the include path passed to the compiler.
However, the issue above with WaitForMultipleObjects is a faulty handling of the async transport in mfreerdp. Both, the async transport thread and the mfreerdp client thread check/treat the freerdp transport+channel handles. Of course mfreerdp should only do this if async transport is not enabled.
I'll fix this in an additional PR soon ...

PR #2612, opened 2015-05-12
Function return bool, so lets return FALSE if 0 or TRUE if 1.

oshogbo:
Function return bool, so lets return FALSE if 0 or TRUE if 1.
Style nits.


mfleisz:
+1

hardening:
+1

bmiklautz:
@oshogbo nice catch! Thank you.

PR #2613, opened 2015-05-12
At least at FreeBSD - the gethostname(3) function can fail.

oshogbo:
At least at FreeBSD - the gethostname(3) function can fail.
Check returned value.


hardening:
+1, I'm wondering in which case the function can fail...

hardening:
@freerdp-bot could you test before I merge it, thanks.

bmiklautz:
@hardening I'd think lack of memory or i/o problems could cause the function to fail. Or maybe simply that the hostname can't be fetched ;)

bmiklautz:
+1

oshogbo:
In my cases the problem is sandbox. :)

PR #2621, opened 2015-05-14
transport/mfreerdp: fix async transport

nfedera:
- handle WAIT_TIMEOUT result as error in async transport thread if an INFINITE timeout was specified in WaitForMultipleObjects
- fix mfreerdp's async transport handling to not use freerdp_get_event_handles and  freerdp_check_event_handles if async transport is activated


bmiklautz:
+1

PR #2624, opened 2015-05-18
QueryContextAttributes shouldn't return an error when a user or a domain is empty.

kwitaszczyk:
When a user sends an empty username or domain QueryContextAttributes should return structure with empty strings instead of returning an error.

### DIFF
--- a/winpr/libwinpr/sspi/NTLM/ntlm.c
+++ b/winpr/libwinpr/sspi/NTLM/ntlm.c
@@ -671,21 +671,27 @@ SECURITY_STATUS SEC_ENTRY ntlm_QueryContextAttributesW(PCtxtHandle phContext, UL
 		context->UseSamFileDatabase = FALSE;
 		credentials = context->credentials;
 		ZeroMemory(AuthIdentity, sizeof(SecPkgContext_AuthIdentity));
-		UserA = AuthIdentity->User;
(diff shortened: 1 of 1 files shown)


hardening:
@freerdp-bot could you give a try and test this please ?

winpr/libwinpr/sspi/NTLM/ntlm.c
```diff
@@ -671,21 +671,27 @@ SECURITY_STATUS SEC_ENTRY ntlm_QueryContextAttributesW(PCtxtHandle phContext, UL
...
-									credentials->identity.UserLength,
-									&UserA, 256, NULL, NULL);
 
-		if (status <= 0)
-			return SEC_E_INTERNAL_ERROR;
+		UserA = AuthIdentity->User;
+		if (credentials->identity.UserLength > 0) {
+			status = ConvertFromUnicode(CP_UTF8, 0,
+			    (WCHAR *)credentials->identity.User,
+			    credentials->identity.UserLength, &UserA, 256, NULL,
```
hardening: I'm curious of the 256 here, is it really a safe value ?

kwitaszczyk: @hardening, currently it is safe as it's the upper bound for User and Domain strings in the SecPkgContext_AuthIdentity structure. However, if SecPkgContext_AuthIdentity is changed then this call should be changed as well. It would be good to provide some constant which says what's the upper bound.

PR #2630, opened 2015-05-19
Add checks for some XXX_New and XXX_Add functions

hardening:
Superseeds https://github.com/bmiklautz/FreeRDP/tree/pull/2616

### DIFF
--- a/channels/rail/client/rail_orders.c
+++ b/channels/rail/client/rail_orders.c
@@ -28,58 +28,30 @@
 
 #include "rail_orders.h"
 
-static BOOL rail_read_unicode_string(wStream* s, RAIL_UNICODE_STRING* unicode_string)
+
+static BOOL rail_write_unicode_string(wStream* s, RAIL_UNICODE_STRING* unicode_string)
 {
-	UINT16 new_length;
-
-	if (Stream_GetRemainingLength(s) < 2)
+	if (!Stream_EnsureRemainingCapacity(s, 2 + unicode_string->length))
 		return FALSE;
 
-	Stream_Read_UINT16(s, new_length); /* cbString (2 bytes) */
-
-	if (Stream_GetRemainingLength(s) < new_length)
-		return FALSE;
-
-	if (!unicode_string->string)
-	{
-		unicode_string->string = (BYTE*) malloc(new_length);
-		if (!unicode_string->string)
-			return FALSE;
-		unicode_string->length = new_length;
-	}
-	else
-	{
-		BYTE *new_str;
-
-		new_str = (BYTE*) realloc(unicode_string->string, new_length);
-		if (!new_str)
-			return FALSE;
-		unicode_string->string = new_str;
-		unicode_string->length = new_length;
-	}
-
-	Stream_Read(s, unicode_string->string, unicode_string->length);
-
+	Stream_Write_UINT16(s, unicode_string->length); /* cbString (2 bytes) */
+	Stream_Write(s, unicode_string->string, unicode_string->length); /* string */
 	return TRUE;
 }
 
-static void rail_write_unicode_string(wStream* s, RAIL_UNICODE_STRING* unicode_string)
-{
-	Stream_EnsureRemainingCapacity(s, 2 + unicode_string->length);
-	Stream_Write_UINT16(s, unicode_string->length); /* cbString (2 bytes) */
-	Stream_Write(s, unicode_string->string, unicode_string->length); /* string */
-}
-
-static void rail_write_unicode_string_value(wStream* s, RAIL_UNICODE_STRING* unicode_string)
+static BOOL rail_write_unicode_string_value(wStream* s, RAIL_UNICODE_STRING* unicode_string)
 {
 	if (unicode_string->length > 0)
 	{
-		Stream_EnsureRemainingCapacity(s, unicode_string->length);
+		if (!Stream_EnsureRemainingCapacity(s, unicode_string->length))
+			return FALSE;
+
 		Stream_Write(s, unicode_string->string, unicode_string->length); /* string */
 	}
+	return TRUE;
 }
 
-void rail_send_pdu(railPlugin* rail, wStream* s, UINT16 orderType)
+BOOL rail_send_pdu(railPlugin* rail, wStream* s, UINT16 orderType)
 {
 	UINT16 orderLength;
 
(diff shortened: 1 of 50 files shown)


channels/rail/client/rail_orders.c
```diff
@@ -550,16 +544,21 @@ void rail_send_client_exec_order(railPlugin* rail, RAIL_EXEC_ORDER* exec)
 			exec->workingDir.length +
 			exec->arguments.length;
 
-	s = rail_pdu_init(RAIL_EXEC_ORDER_LENGTH);
-	rail_write_client_exec_order(s, exec);
-	rail_send_pdu(rail, s, RDP_RAIL_ORDER_EXEC);
+	s = rail_pdu_init(length);
```
nfedera: nice catch!

channels/rdpdr/client/devman.c
```diff
@@ -47,14 +47,17 @@ DEVMAN* devman_new(rdpdrPlugin* rdpdr)
 {
 	DEVMAN* devman;
 
-	devman = (DEVMAN*) malloc(sizeof(DEVMAN));
-	ZeroMemory(devman, sizeof(DEVMAN));
+	devman = (DEVMAN*) calloc(1, sizeof(DEVMAN));
```
nfedera: NULL check ?

channels/rdpsnd/client/rdpsnd_main.c
```diff
@@ -1130,39 +1152,37 @@ BOOL VCAPITYPE VirtualChannelEntry(PCHANNEL_ENTRY_POINTS pEntryPoints)
 
 	rdpsnd = (rdpsndPlugin*) calloc(1, sizeof(rdpsndPlugin));
 
-	if (rdpsnd)
-	{
+	if (!rdpsnd)
+		return -1;
```
nfedera: Function is BOOL. Should return FALSE here.

channels/rdpsnd/client/rdpsnd_main.c
```diff
@@ -1130,39 +1152,37 @@ BOOL VCAPITYPE VirtualChannelEntry(PCHANNEL_ENTRY_POINTS pEntryPoints)
...
 
-		rdpsnd_add_init_handle_data(rdpsnd->InitHandle, (void*) rdpsnd);
+	rc = rdpsnd->channelEntryPoints.pVirtualChannelInit(&rdpsnd->InitHandle,
+		&rdpsnd->channelDef, 1, VIRTUAL_CHANNEL_VERSION_WIN2000, rdpsnd_virtual_channel_init_event);
+	if (CHANNEL_RC_OK != rc)
+	{
+		WLog_ERR(TAG, "pVirtualChannelInit failed with %s [%08X]",
+				 WTSErrorToString(rc), rc);
+		free(rdpsnd);
+		return -1;
```
nfedera: Function is BOOL. Should return FALSE here.

client/X11/xf_keyboard.c
```diff
@@ -81,14 +84,13 @@ int xf_keyboard_action_script_init(xfContext* xfc)
...
+		if (ArrayList_Add(xfc->keyCombinations, keyCombination) < 0)
+			return 0;
 	}
 
 	exitCode = pclose(keyScript);
 
-	xf_event_action_script_init(xfc);
-
-	return 1;
+	return xf_event_action_script_init(xfc);
```
nfedera: This function (xf_keyboard_action_script_init) seemed to return 0 on failure and 1 on success. Here you return the result of xf_event_action_script_init() which returns -1 on failure

libfreerdp/core/server.c
```diff
@@ -687,30 +689,46 @@ HANDLE WINAPI FreeRDP_WTSOpenServerA(LPSTR pServerName)
 		return INVALID_HANDLE_VALUE;
 
 	vcm = (WTSVirtualChannelManager*) calloc(1, sizeof(WTSVirtualChannelManager));
+	if (!vcm)
+		return NULL;
```
nfedera: Shouldn't you return INVALID_HANDLE_VALUE here like in the other error cases below ?

libfreerdp/core/server.c
```diff
@@ -944,6 +966,13 @@ HANDLE WINAPI FreeRDP_WTSVirtualChannelOpen(HANDLE hServer, DWORD SessionId, LPS
 	hChannelHandle = (HANDLE) channel;
 
 	return hChannelHandle;
+
+error_queue:
+	Stream_Free(channel->receiveData, TRUE);
+error_receiveData:
+	free(channel);
```
nfedera: It is correct here but I see a little risk here because the memory pointed to by channel is not allocated here and if future error_xx gotos are added it could easily be overseen. If you don't think that this is too paranoid I'd suggest to assign channel to NULL in the declaration, directly assign hChannelHandle in line 945 and move line 966 after line 963 .... or whatever fixes the risk ;)

libfreerdp/core/server.c
hardening: Yes definitely and we should also set the error. Nice catch.

server/shadow/shadow_client.c
```diff
@@ -95,9 +95,7 @@ BOOL shadow_client_context_new(freerdp_peer* peer, rdpShadowClient* client)
 	if (!(client->encoder = shadow_encoder_new(client)))
 		goto fail_encoder_new;
 
-	ArrayList_Add(server->clients, (void*) client);
-
-	return TRUE;
+	return ArrayList_Add(server->clients, (void*) client) >= 0;
```
nfedera: 1. Shouldn't we do the clean-up in that case ?
2. I just noticed that the code in line 101 below is wrong.  Under the fail_encoder_new: label CloseHandle should be called with client->StopEvent and not with client->encoder !

libfreerdp/core/server.c
hardening: I'm sorry norbert I don't see the risk and why it's better to set channel to NULL. Anyway I have catched a case with the calloc of the channel beeing NULL.

winpr/libwinpr/smartcard/smartcard_pcsc.c
```diff
@@ -561,11 +574,20 @@ PCSC_SCARDHANDLE* PCSC_ConnectCardHandle(SCARDCONTEXT hSharedContext, SCARDCONTE
...
 	if (!g_CardHandles)
+	{
 		g_CardHandles = ListDictionary_New(TRUE);
+		if (!g_CardHandles)
+			goto error;
+	}
 
-	ListDictionary_Add(g_CardHandles, (void*) hCard, (void*) pCard);
+	if (!ListDictionary_Add(g_CardHandles, (void*) hCard, (void*) pCard))
+		goto error;
```
nfedera: Maybe move "pContext->dwCardHandleCount++;" from line 574 above directly before the return ?

server/shadow/shadow_client.c
hardening: 1. you're definitely right
2. nice catch

winpr/libwinpr/smartcard/smartcard_pcsc.c
```diff
@@ -634,8 +656,7 @@ BOOL PCSC_AddReaderNameAlias(char* namePCSC, char* nameWinSCard)
 
 	reader->namePCSC = _strdup(namePCSC);
 	reader->nameWinSCard = _strdup(nameWinSCard);
-	ArrayList_Add(g_Readers, reader);
-	return TRUE;
+	return ArrayList_Add(g_Readers, reader) >= 0;
```
nfedera: If ArrayList_Add fails then memory won't be freed - right ?

winpr/libwinpr/smartcard/smartcard_pcsc.c
```diff
@@ -890,12 +911,16 @@ char* PCSC_ConvertReaderNamesToPCSC(const char* names, LPDWORD pcchReaders)
...
 {
 	if (!g_MemoryBlocks)
+	{
 		g_MemoryBlocks = ListDictionary_New(TRUE);
+		if (!g_MemoryBlocks)
+			return FALSE;
+	}
 
-	ListDictionary_Add(g_MemoryBlocks, pvMem, (void*) hContext);
+	return ListDictionary_Add(g_MemoryBlocks, pvMem, (void*) hContext);
```
nfedera: If ListDictionary_Add fails then memory won't be freed - right ?

winpr/libwinpr/smartcard/smartcard_pcsc.c
hardening: Yes of course I will change for that.

nfedera:
Nice PR! Thanks. Definitely +1 (after handling my remarks)

nfedera:
@freerdp-bot test

nfedera:
xf_event_action_script_init is still int (vs. BOOL) in xf_event.h

PR #2633, opened 2015-05-20
Fix rounding error in progressive codec surface new

clivest:
The grid is composed of 64x64 blocks and should not be smaller than the surface. If width or height were not a multiple of 64 the previous rounding resulted in a grid smaller than the surface.


mfleisz:
@freerdp-bot test

PR #2635, opened 2015-05-20
Fix two memory leaks in gdi/gfx

clivest:
clippingRect needs to be uninitialized to free heap-allocated data member.

Noticed using valgrind.


mfleisz:
@freerdp-bot test

PR #2636, opened 2015-05-21
Fix crashes in shadow server

xhaakon:


hardening:
@freerdp-bot would you mind test this PR and see if it compiles cleanly, thank you.

hardening:
It looks good to me, +1.
@xhaakon it would be really great, if you could write a unitary test (complementing an existing one or creating a new one) that would verify your correction.

winpr/libwinpr/synch/barrier.c
```diff
@@ -165,6 +167,8 @@ BOOL WINAPI DeleteSynchronizationBarrier(LPSYNCHRONIZATION_BARRIER lpBarrier)
 	if (!pBarrier)
 		return TRUE;
 
+	while (InterlockedCompareExchange(&pBarrier->count, 0, 0) != 0);
+
```
bmiklautz: For a high count this might cause the CPU to get busy. Maybe a minimal Sleep should be added here.

bmiklautz:
As hardening suggested. If you could add a unit test that reproduces this problem that would be great.

realjiangms:
This also fixes the race condition I pasted at https://github.com/FreeRDP/FreeRDP/pull/2395#issuecomment-77742784

winpr/libwinpr/synch/barrier.c
xhaakon: I've added 100ms sleep into the loop.

xhaakon:
@bmiklautz I've extended TestSynchBarrier to cover also the case of barrier's early deletion. Please let me know if you'd rather prefer a test in some different format.

bmiklautz:
@freerdp-bot test

PR #2638, opened 2015-05-21
Assign peer to settings when a new RDP module is being initialized.

kwitaszczyk:
Otherwise instance in rdpNla is invalid when NLA is used on server-side.

### DIFF
--- a/libfreerdp/core/peer.c
+++ b/libfreerdp/core/peer.c
@@ -635,6 +635,7 @@ BOOL freerdp_peer_context_new(freerdp_peer* client)
 
 	client->context = context;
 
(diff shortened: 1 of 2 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@kwitaszczyk can you give more details how you hit the problem?

bmiklautz:
+1

kwitaszczyk:
@bmiklautz, when a server accepts an NLA connection an rdpNla object is created using settings->instance. In case of a peer this pointer was invalid.

hardening:
+1

PR #2643, opened 2015-05-22
Don't dispose shadow server before all clients have disconnected

xhaakon:
Ensure all client threads are disconnected before we stop the server
thread and thus make it available for deletion. We do this because
shadow_client_context_free() accesses rdpShadowServer structure, which
we have to keep alive as long as it is needed.

Calling shadow_client_context_free() after shadow_server_stop() and
shadow_server_free() were called may have caused an application crash.

### DIFF
--- a/server/shadow/shadow_server.c
+++ b/server/shadow/shadow_server.c
@@ -344,6 +344,14 @@ void* shadow_server_thread(rdpShadowServer* server)
 
 	shadow_subsystem_stop(server->subsystem);
 
+	/* Signal to the clients that server is being stopped and wait for them
+	 * to disconnect. */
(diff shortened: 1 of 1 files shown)


hardening:
@freerdp-bot would you test that PR please

realjiangms:
This fix may be incomplete. The message posted in subsystem->MsgPipe may only be tracked by one client. If we have more than 2 clients, other clients would not be notified.

It's actually common problem. All message posted in subsystem->MsgPipe->out would be tracked by only  one client.

@awakecoding If this is not by intent, we may need to consider an MsgPipe as per-client appliance.
We can have a shared message wparam to save memory from huge amount of client, as use reference count for resource release strategy

PR #2644, opened 2015-05-22
shadow_client: dispose encomsp and remdesk

xhaakon:
Their threads were kept running after the client's exit.


mfleisz:
@freerdp-bot test

Issue #2389, opened 2015-02-12
Problems with serial redirection


StephanKrieger:
Hi @all,

i hope this is the right place for my questions.

I am from germany and wants to use FreeRDP to connect from Linux-ThinClients to a Windows Terminal Server. Starting a RDP Session is not the problem.

I have problems to use the serial ports after the serial port redirection.
I tried two different installations with two different Problems.

First one:
Installed FreeRDP.git like the Compilation-Guide said. 
When i now ridirect the port and will open it on the Windows Terminal Server, Windows says in german: "Die Anforderung wird nicht unterstützt"
And when i close the session, the Linux also says "Attempt to use the unsupported SERIAL_XOFF_CONTINUE"
Can somebody say to me, what thats means?

Second one:
I get a Special vendor package for a Hardware Plattform with an angstrom Distribution.
When i now start there the redirection, My program based on .Net Framwork says in german:
„Error: Fehler beim Öffnen des Waagen-ComPorts - Fehler: System.ArgumentException: Der angegebene Anschlussname ist ungültig. Es kann ein gültiger Anschluss sein, aber kein serieller Anschluss.
Parametername: portName
   bei System.IO.Ports.SerialStream..ctor(String portName, Int32 baudRate, Parity parity, Int32 dataBits, StopBits stopBits, Int32 readTimeout, Int32 writeTimeout, Handshake handshake, Boolean dtrEnable, Boolean rtsEnable, Boolean discardNull, Byte parityReplace)
   bei System.IO.Ports.SerialPort.Open()"
Did somebody know the message and how i can resolve it?

Greetz from germany
Stephan

eledoux:
@StephanKrieger when you says two different installations, do you mean two different devices?

First issue:
That's means your application tried to setup the SERIAL_XOFF_CONTINUE feature that is not supported by the current implementation of the serial port redirection. You can get a bit of information here: https://msdn.microsoft.com/en-us/library/jj680685%28v=vs.85%29.aspx

Second issue:
This looks similar to the first issue. Your program doesn't succeed to setup the port as it would like and returns the error.

If this prevents you to use the serial port you can try to complete the xfreerdp's command line by using for each port something like: /serial:COM1,/dev/ttyS0,Serial,permissive. This turns on a more permissive mode which will pretend a feature such as SERIAL_XOFF_CONTINUE is well supported by the redirection. 

[...]

PR #2646, opened 2015-05-22
fixed a synchronization issue on thread_list

eledoux:
It was possible for a thread to call ExitThread() before it was added to thread_list. In this case a signal was never sent and was never caught by WaitForSingleObject().

Now is ensured that the thread ID is added to thread_list first by the thread itself or by the thread calling pthread_create(). Since thread_list is synchronized, it should not hurt to add two times the thread ID. This is maybe simpler than to check if the ID is already in the list.

winpr-thread:
  - added the unit test: TestThreadExitThread
  - fix: ensure thread_list to be up to date before to call ExitThread()
  - possibly resolved: Problems with serial redirection #2389

### DIFF
--- a/winpr/libwinpr/thread/test/CMakeLists.txt
+++ b/winpr/libwinpr/thread/test/CMakeLists.txt
@@ -6,7 +6,8 @@ set(${MODULE_PREFIX}_DRIVER ${MODULE_NAME}.c)
 
 set(${MODULE_PREFIX}_TESTS
 	TestThreadCommandLineToArgv.c
-	TestThreadCreateProcess.c)
+	TestThreadCreateProcess.c
+	TestThreadExitThread.c)
 
 create_test_sourcelist(${MODULE_PREFIX}_SRCS
 	${${MODULE_PREFIX}_DRIVER}
--- a/winpr/libwinpr/thread/test/TestThreadExitThread.c
+++ b/winpr/libwinpr/thread/test/TestThreadExitThread.c
@@ -19,7 +19,7 @@ int TestThreadExitThread(int argc, char* argv[])
 
 	/* FIXME: create some noise to better guaranty the test validity and
          * decrease the number of loops */
-	for (i=0; i<10000; i++)
+	for (i=0; i<50000; i++)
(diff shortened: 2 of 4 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@eledoux if I'm not mistaken with this pull request it can happen that old thread stick around in the thread_list. List_Remove only removes the first matching element found therefore, if added twice, one item would always stay in the list causing it to grow over time.

akallabeth:
@eledoux Wouldn't it be better to use a wait condition for this synchronisation issue? Then ```thread_launcher``` would wait until ```winpr_StartThread``` updated the list and the race is fixed. No duplicate entries in the thread list.

eledoux:
@bmiklautz you're right there is indeed a possibility to call CloseHandle() before thread_launcher() is done. The @akallabeth's solution is then better. However should we be afraid on a bad impact on performances?

eledoux:
@bmiklautz @akallabeth With this new implementation, I'm not quite sure it is a bright idea to use a winpr's Event object in the WINPR_THREAD structure even if we should not have some circular dependencies here. An implementation based on winpr_sem_t or anything else could have been better?

akallabeth:
@eledoux Would suggest using a pthread_condition, ```CreateEvent``` uses quite a lot of resources.

bmiklautz:
please test

hardening:
Implementation looks good, I can't see any locking problem.
+1

eledoux:
Thanks for the suggestion. I'm getting happy with this implementation which is a bit faster. I also made a try with an implementation that was using only one semaphore (sem_t). This latter was a bit more faster but looking at semaphore.c it seems a such solution is not portable enough, isn't it? (I develop only on Linux)

Otherwise using a semaphore from CreateSemaphore() with WINPR_PIPE_SEMAPHORE still defined was even slower than using an Event.

hardening:
IMHO you have the optimized version now, don't change anything.

akallabeth:
lgtm +1

Issue #2593, opened 2015-05-05
Freerdp iOS issue 


shattars3d:
When compiling the freerdp master build for the iOS client of freerdp, I'm getting the issue of the freerdp settings not being loaded before connecting into the rdp. When running on an iOS device and selecting a server to connect to, xcode has a memory issue at the line     free(inst->settings->HomePath); in the iOS_freerdp.m file.

lgqiang:
The newest code has bug. I have made some change to fix the bug:

lgqiang:
2 files need to change:

freerdp.c line 501, change to below:

	if (!ret) 
		return TRUE;

gdi.c, line 1357, change to below:

if (!instance->context->cache) {
		if (!(cache = cache_new(instance->settings)))
			goto fail_cache;
        instance->context->cache = cache;
    }

after make those change, you can run the program with no crash.

hardening:
@lgqiang if so can you submit a PR with these changes. Thanks.

lgqiang:
@hardening sorry, I don't known how to submit a PR.

shattars3d:
@lgqiang  thanks, always thought the master would be stable ;) but its not lol

shattars3d:
@lgqiang  

when i try to connect to the test server "test service.freerdp.com", it gives me this in the Xcode debugger….

2015-05-14 12:00:33.452 iFreeRDP[2458:60b] Reachability Flag Status: -R ------- networkStatusForFlags
failed to connect to testservice.ifreerdp.com
[12:00:34:430] [2458:04739000] [ERROR][com.freerdp.core.nego] - Protocol Security Negotiation Failure
[12:00:34:430] [2458:04739000] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED [0x2000C]
[12:00:34:430] [2458:04739000] [ERROR][com.freerdp.core.connection] - Error: protocol security negotiation or connection failure
2015-05-14 12:00:34.998 iFreeRDP[2458:6307] ios_run_freerdp: inst->rdp_connect failed

YangLi20:
hello,when i user cmake build freerdp,i met a error ,the reason is " unknown cmake command "set_complex_link_libraries" " , why ?  thank you,my cmake version is 3.3.2

shattars3d:
[...]

PR #2650, opened 2015-05-25
Fixed the issue where iOS would fail to launch a remote desktop due t…

phongleq:
…o memory issue.

"free(inst->settings->HomePath);"

The bug ticket this goes with: https://github.com/FreeRDP/FreeRDP/issues/2593

### DIFF
--- a/libfreerdp/core/freerdp.c
+++ b/libfreerdp/core/freerdp.c
@@ -498,7 +498,7 @@ BOOL freerdp_context_new(freerdp* instance)
 
 	IFCALLRET(instance->ContextNew, ret, instance, instance->context);
 
-	if (ret)
+	if (!ret)
 		return TRUE;
 
 out_error_graphics_new:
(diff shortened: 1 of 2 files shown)


libfreerdp/core/freerdp.c
```diff
@@ -498,7 +498,7 @@ BOOL freerdp_context_new(freerdp* instance)
 
 	IFCALLRET(instance->ContextNew, ret, instance, instance->context);
 
-	if (ret)
+	if (!ret)
```
nfedera: This is definitely wrong. If ret is FALSE here this means that instance->ContextNew is set and failed.
This change would break all clients!

libfreerdp/gdi/gdi.c
```diff
@@ -1354,13 +1354,11 @@ BOOL gdi_init(freerdp* instance, UINT32 flags, BYTE* buffer)
...
-		if (!(cache = cache_new(instance->settings)))
-			goto fail_cache;
-
-		instance->context->cache = cache;
-	}
+    if (!instance->context->cache) {
+        if (!(cache = cache_new(instance->settings)))
+            goto fail_cache;
+        instance->context->cache = cache;
+    }
```
nfedera: I don't see any logical change here and your modifications are against the FreeRDP coding style guidelines

nfedera:
The real issue is that ios_context_new is wrong. Wrong signature and missing error checking.

nfedera:
@phongleq I've created PR #2652.
Please let me know if that fixes your issue. Thanks.

nfedera:
Closing this PR since it would break all other clients.

PR #2652, opened 2015-05-25
client/iOS: fix ios_context_new

nfedera:
- fixed wrong signature since ContextNew is BOOL now
- added error checking


phongleq:
Thanks @nfedera for this fix. My merge request was based off someone else issue that didn't know how to make a merge request. There was no opposition on the bug filed so I made one on their behalf.

PR #2653, opened 2015-05-26
winpr: add intrin.h

bmiklautz:
Exposing lzcnt in crt.h might cause compiler errors (redefinition) with
recent versions of gcc (>=4.9) when winpr is included in other projects.
As lzcnt isn't part of crt according to MSDN and also shouldn't be
exported by default it was moved to intrin.h.

The related test was also moved to the top level directory of winpr.

### DIFF
--- a/winpr/test/CMakeLists.txt
+++ b/winpr/test/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-set(MODULE_NAME "TestWinpr")
+set(MODULE_NAME "TestWinPR")
 set(MODULE_PREFIX "TEST_WINPR")
 
 set(${MODULE_PREFIX}_DRIVER ${MODULE_NAME}.c)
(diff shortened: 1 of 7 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

mfleisz:
+1

winpr/test/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+
+set(MODULE_NAME "TestWinpr")
```
nfedera: I don't really care but the other test is called TestWinPRUtils (uppercase PR).

nfedera:
Not your code but winpr/test/TestIntrinsics.c is broken because TestIntrinsics() always returns 0.

PR #2658, opened 2015-05-27
Fix for #2657

### LINKED ISSUE #2657: FreeRDP closes while running RemoteApp, but server session stays the same.
Hello. Bear with me as I don't understand all the technicalities of Linux, but i'll try to help as much as I can.

The bug that occurs is that FreeRDP closes for no apparent reason while running a RemoteApp. I've managed to replicate this running Word 2016 Preview (and it also happens on another software on our company). On Word it happens when on New File I try to create a new file from template. On the other software (which i'm not able to access at the moment, probably only next week) it does the same when the program tries to load a list from the database.

However if I reopen it works proper until I try to create a template. For the other app I'll also try to get crash info when I'm able to use it again.

Command line used: xfreerdp /v:REDACTED /d:REDACTED.pt.local /load-balance-info:tsv://VMResource.1.Col_WIndows_81 /u:pedro.flores /p:REDACTED /app:||explorer /app-name:Explorer /cert-ignore +fonts -themes -wallpaper /network:lan
(It's called from explorer, and on explorer I open Word. It also crashes if I directly open the app).

The server is running Windows Server 2012 R2 with RDS running on VDI mode. The hosts are also 2012 R2.

Version: This is FreeRDP version 1.2.4-dev (git 58b1995)

OS running FreeRDP: Xubuntu 15.04 amd64, kernel 3.19.0-16-generic, fully patched as of May 27.

CMake parameters: -DCMAKE_BUILD_TYPE=Debug -DWITH_SSE2=ON

hardening:
This fixes #2657 with an empty window title set.

### DIFF
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -470,7 +470,8 @@ static BOOL xf_rail_window_common(rdpContext* context, WINDOW_ORDER_INFO* orderI
 
 	if (fieldFlags & WINDOW_ORDER_FIELD_TITLE)
 	{
-		xf_SetWindowText(xfc, appWindow, appWindow->title);
+		if (appWindow->title)
+			xf_SetWindowText(xfc, appWindow, appWindow->title);
 	}
 
 	if ((fieldFlags & WINDOW_ORDER_FIELD_WND_OFFSET) ||
--- a/libfreerdp/core/peer.c
+++ b/libfreerdp/core/peer.c
@@ -222,7 +222,7 @@ static BOOL freerdp_peer_initialize(freerdp_peer* client)
 
 		if (!settings->RdpServerRsaKey)
 		{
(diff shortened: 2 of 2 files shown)


client/X11/xf_rail.c
```diff
@@ -470,7 +470,8 @@ static BOOL xf_rail_window_common(rdpContext* context, WINDOW_ORDER_INFO* orderI
 
 	if (fieldFlags & WINDOW_ORDER_FIELD_TITLE)
 	{
-		xf_SetWindowText(xfc, appWindow, appWindow->title);
+		if (appWindow->title)
```
bmiklautz: @hardening not sure if it make sense but maybe we should handle this case directly in xf_SetWindowText

hardening: xf_SetWindowText() is called only once from this location.

PR #2660, opened 2015-05-28
tsmf: fix invalid check and a potential segfault

nfedera:
This resuscitates tsmf which was accidently broken in commit f8120919


hardening:
+1 stupid error from me

nfedera:
@hardening I've also missed it while reviewing ;)

PR #2670, opened 2015-06-02
Adjusted config paths

akallabeth:
* Windows configuration and runtime data now in appropriate folders (AppData, Roaming, Local)
* ```<vendor>``` and ```<product>``` can be changed from cmake to avoid configuration collisions of differing FreeRDP builds.
* Configuration now defaults to ```<vendor>/<product>``` subfolders if  ```<vendor>``` and ```<product>``` are not equal. (By default they are)
* Default builds still use the same configuration folders as before (Unix)


libfreerdp/core/settings.c
```diff
@@ -471,7 +472,17 @@ rdpSettings* freerdp_settings_new(DWORD flags)
...
+		base = GetKnownSubPath(KNOWN_PATH_XDG_CONFIG_HOME,
+				       FREERDP_VENDOR_STRING);
+		if (base)
+		{
+			settings->ConfigPath = GetCombinedPath(base, FREERDP_PRODUCT_STRING);
+			if (!PathFileExistsA(base))
+				if (!CreateDirectoryA(base, NULL))
+					settings->ConfigPath = NULL;
+		}
+		free (base);
```
bmiklautz: This would change the default configuration location (~/.config/freerdp) on linux to ~/.config/FreeRDP/FreeRDP as well ;(. Please make sure that the current path doesn't change for compatibility.

bmiklautz:
+1

libfreerdp/core/settings.c
```diff
@@ -195,13 +195,14 @@ void settings_get_computer_name(rdpSettings* settings)
 
 	GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize);
 	settings->ComputerName = (char*) malloc(nSize);
-    if (!settings->ComputerName)
-        return;
+	if (!settings->ComputerName)
```
hardening: Not your code, but the malloc failure silently exists

libfreerdp/core/settings.c
```diff
@@ -471,7 +472,37 @@ rdpSettings* freerdp_settings_new(DWORD flags)
...
-		settings->ConfigPath = GetKnownSubPath(KNOWN_PATH_XDG_CONFIG_HOME, "freerdp");
+
+		/* For default FreeRDP continue using same config directory
+		 * as in old releases.
+		 * Custom builds use <Vendor>/<Product> as config folder. */
+		if (_stricmp(FREERDP_VENDOR_STRING, FREERDP_PRODUCT_STRING))
+		{
+			base = GetKnownSubPath(KNOWN_PATH_XDG_CONFIG_HOME,
+					       FREERDP_VENDOR_STRING);
+			if (base)
```
hardening: I'm perhaps misinterpreting but when base == NULL that means OOM, isn't it ? So an error

libfreerdp/core/settings.c
bmiklautz: Will be fixed in one of my upcoming pull requests.

hardening:
Some non-blocking comments...
+1

libfreerdp/core/settings.c
akallabeth: @hardening Already handled. Next line after the end of the scope is a check for ```settings->ConfigPath == NULL``` which it is as long as it is not set.

bmiklautz:
@awakecoding any objections on changing the paths on windows?

akallabeth:
Will need adjustment if #2675 is merged.

awakecoding:
@bmiklautz no objection

PR #2676, opened 2015-06-03
winpr: fixed some misuses of S_FALSE

nfedera:
Contrary to popular belief S_FALSE is a HRESULT success code and is not used to indicate an error. That is why they prefix with an S_ ;)

PathCchAppend:
- Make this function behave like the Windows 8 implementation and the according MSDN specification.
- return E_INVALIDARG instead of S_FALSE if pszPath is NULL
- return E_INVALIDARG instead of S_FALSE if pszMore is NULL
- return E_INVALIDARG if cchPath is zero
- return E_INVALIDARG if cchPath is greater than PATHCCH_MAX_CCH
- return HRESULT_FROM_WIN32(ERROR_FILENAME_EXCED_RANGE) if the combined  path size exceeds cchPath (this is the same HRESULT which the Windows  implementation returns in this case and which is referred to as  PATHCCH_E_FILENAME_TOO_LONG on msdn)
- extended/fixed the TestPathCchAppend ctest

PathCchStripPrefix:
- fixed to be compatible to the Windows 8 implementation of this function:
- return E_INVALIDARG instead of S_FALSE if pszPath is NULL
- return E_INVALIDARG instead of S_FALSE if cchPath < 4
- return E_INVALIDARG if cchPath > PATHCCH_MAX_CCH
- allow stripping of "\\?\c:" and don't require "\\?\c:\"
- verify that the character at the drive letter position is actually an alphabetic character
- since the passed pszPath must not necessarily be null terminated and we always have enough  space after the memmove we can always ensure the null termination of the stripped result
- extended/fixed the TestPathCchStripPrefixctest

PathCchAddExtension:
- return E_INVALIDARG instead of S_FALSE
- return HRESULT_FROM_WIN32(ERROR_INSUFFICIENT_BUFFER) instead of S_FALSE
- extended/fixed the corresponding ctest

PathCchAddBackslashEx
- return E_INVALIDARG instead of S_FALSE
- return HRESULT_FROM_WIN32(ERROR_INSUFFICIENT_BUFFER) instead of S_FALSE
- extended/fixed the corresponding ctest

PathCchAddBackslash
- return E_INVALIDARG instead of S_FALSE
- return HRESULT_FROM_WIN32(ERROR_INSUFFICIENT_BUFFER) instead of S_FALSE
- extended/fixed the corresponding ctest

More PathCch fixes next week in a separate PR

### DIFF
--- a/winpr/libwinpr/path/path.c
+++ b/winpr/libwinpr/path/path.c
@@ -634,27 +634,29 @@ HRESULT PathCchStripToRootW(PWSTR pszPath, size_t cchPath)
 HRESULT PathCchStripPrefixA(PSTR pszPath, size_t cchPath)
 {
 	BOOL hasPrefix;
-	BOOL deviceNamespace;
 
 	if (!pszPath)
-		return S_FALSE;
+		return E_INVALIDARG;
 
-	if (cchPath < 4)
-		return S_FALSE;
+	if (cchPath < 4 || cchPath > PATHCCH_MAX_CCH)
+		return E_INVALIDARG;
 
 	hasPrefix = ((pszPath[0] == '\\') && (pszPath[1] == '\\') &&
 		(pszPath[2] == '?') && (pszPath[3] == '\\')) ? TRUE : FALSE;
 
 	if (hasPrefix)
 	{
-		if (cchPath < 7)
+		if (cchPath < 6)
 			return S_FALSE;
 
-		deviceNamespace = ((pszPath[5] == ':') && (pszPath[6] == '\\')) ? TRUE : FALSE;
-
-		if (deviceNamespace)
+		if (IsCharAlpha(pszPath[4]) && (pszPath[5] == ':')) /* like C: */
 		{
 			memmove_s(pszPath, cchPath, &pszPath[4], cchPath - 4);
+			/* since the passed pszPath must not necessarily be null terminated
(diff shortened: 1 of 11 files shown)


winpr/libwinpr/path/path.c
```diff
@@ -634,27 +634,29 @@ HRESULT PathCchStripToRootW(PWSTR pszPath, size_t cchPath)
...
+		return E_INVALIDARG;
 
 	hasPrefix = ((pszPath[0] == '\\') && (pszPath[1] == '\\') &&
 		(pszPath[2] == '?') && (pszPath[3] == '\\')) ? TRUE : FALSE;
 
 	if (hasPrefix)
 	{
-		if (cchPath < 7)
+		if (cchPath < 6)
 			return S_FALSE;
```
hardening: In this case it's S_FALSE ?

hardening:
I'm not a win32 expert but it LGTM.

winpr/libwinpr/path/path.c
nfedera: Yes, meaning it doesn't have the prefix. Windows also returns S_FALSE in this case

nfedera: Update: it has the prefix but nothing where the prefix can be removed from. So that's perfectly fine

bmiklautz:
+1

PR #2682, opened 2015-06-08
winpr/path: more PathCch* fixes

nfedera:
- Fixed remaining misuses of S_FALSE which is a HRESULT success code and is not used to indicate an error.
- changed the unimplemented PathCch* functions to return E_NOTIMPL instead of the HRESULT S_OK and print an error message.
- the unimplemented PathCch ctests now also print a warning message.

Also several fixes for PathCchFindExtension:
- fixed multiple implementation errors
- return S_OK if the extension was NOT found
- if no extension was found, ppszExt must point to the string's terminating null
- return E_INVALIDARG if pszPath is not null-terminated within the cchPath range
- return E_INVALIDARG if pszPath is NULL
- return E_INVALIDARG if ppszExt is NULL
- return E_INVALIDARG if cchPath is Zero
- return E_NOTIMPL instead of S_OK in PathPathCchFindExtensionW()
- extended/fixed the TestPathCchFindExtension ctest

Denounce PathAllocCombine - This PR does not fix this function but:
- print a warning message that the function is buggy and added a code comment describing the issues
- fix misuse of the S_FALSE HRESULT in error conditions
- prevent some segfaults
- check result of HeapAlloc

(Fortunately PathAllocCombine is unused in FreeRDP)

### DIFF
--- a/winpr/libwinpr/path/include/PathAllocCombine.c
+++ b/winpr/libwinpr/path/include/PathAllocCombine.c
@@ -6,6 +6,18 @@
 #define PATH_ALLOC_COMBINE	PathAllocCombineA
 */
 
+/**
+ * FIXME: These implementations of the PathAllocCombine functions have
+ * several issues:
+ * - pszPathIn or pszMore may be NULL (but not both)
+ * - no check if pszMore is fully qualified (if so, it must be directly
+ *   copied to the output buffer without being combined with pszPathIn.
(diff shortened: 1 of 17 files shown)


bmiklautz:
@freerdp-bot test

hardening:
LGTM +1

bmiklautz:
+1

Issue #2628, opened 2015-05-19
freerdp-server problem


Hewitt1410:
I use rdp server(on Ubuntu 12.04 64 bit) version 1.2.4-dev (git0067241)
rdp client(shell and control version 6.2.9200, remote desktop protocol 8.0 support, on windows 7 ultimate 32 bit) connect to rdp server success and i can view content from Ubuntu, but after few minutes(about 2 minutes, sometimes more than 2 minutes), window show a message : "Because of an error in data encryption, this session will end. Please try connecting to the remote computer again".
Do you have any suggestions, what I can do?
Thanks,
Hewitt

hardening:
freerdp-server is more a toy to test server-side features than a final product.
Anyway can you give us some details of what is negotiated with your mstsc: TLS / RDP, remoteFx ?

Hewitt1410:
Log trace :
./freerdp-shadow 
[09:07:11:726] [2880:2880] [INFO][com.freerdp.server.shadow.x11] - X11 Extensions: XFixes: 1 Xinerama: 1 XDamage: 0 XShm: 0
[09:07:11:731] [2880:2880] [INFO][com.freerdp.core.listener] - Listening on 0.0.0.0:3389
[09:07:16:724] [2880:2900] [INFO][com.freerdp.core.connection] - Client Security: NLA:1 TLS:1 RDP:0
[09:07:16:724] [2880:2900] [INFO][com.freerdp.core.connection] - Server Security: NLA:0 TLS:1 RDP:1
[09:07:16:724] [2880:2900] [INFO][com.freerdp.core.connection] - Negotiated Security: NLA:0 TLS:1 RDP:0
[09:07:16:790] [2880:2900] [ERROR][com.freerdp.client.shadow] - Failed to check FreeRDP file descriptor
[09:07:16:836] [2880:2905] [INFO][com.freerdp.core.connection] - Client Security: NLA:1 TLS:1 RDP:0
[09:07:16:836] [2880:2905] [INFO][com.freerdp.core.connection] - Server Security: NLA:0 TLS:1 RDP:1
[09:07:16:836] [2880:2905] [INFO][com.freerdp.core.connection] - Negotiated Security: NLA:0 TLS:1 RDP:0
[09:07:16:855] [2880:2905] [INFO][com.freerdp.core.connection] - Accepted client: XXXXXXXXXXX-PC
[...]

PR #2683, opened 2015-06-08
server/shadow: resuscitate shadow server

nfedera:
shadow_client_refresh_rect returned FALSE if the areas parameter was NULL. It may only do so if the count parameter is > 0.

### DIFF
--- a/server/shadow/shadow_client.c
+++ b/server/shadow/shadow_client.c
@@ -274,9 +274,9 @@ BOOL shadow_client_refresh_rect(rdpShadowClient* client, BYTE count, RECTANGLE_1
 			free (wParam);
 			return FALSE;
 		}
-	}
 
-	CopyMemory(wParam->rects, areas, wParam->numRects * sizeof(RECTANGLE_16));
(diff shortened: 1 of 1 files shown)


hardening:
A +1, it was a mistake from me

realjiangms:
We may need one more additional line:

279 CopyMemory(wParam->rects, areas, wParam->numRects * sizeof(RECTANGLE_16));


It should be moved into the above 'if' block.
I'm surprised that the github don't allow me to add comment at 'unchanged' line

realjiangms:
@hardening This is not a mistake in your PR. It's existing error in original shadow code.
The code try to refresh the whole screen when a client is activated, but it actually never.
Even without your PR, it doesn't work. It just not obvious.

realjiangms:
+1 Except one more change in for CopyMemory

nfedera:
@realjiangms you're absolutely right. fixed !

realjiangms:
Test passed. Is it all set to be merged into master?
Resuscitate it :)

@hardening 
I'm not sure whether it fixes #2628. But worth to give a try.
The problem in #2628 may be more complex

PR #2684, opened 2015-06-08
Add bépo layout

Skool:
Bépo is a French ergonomic layout based on the same method as Dvorak.
This layout can be added to a Windows OS with the installer in http://bepo.fr

This layout is also recognized by xorg, xkb and all linux systems natively.

We need to discuss that the LayoutID known in the Windows registry seems to be the same as German Neo layout, who is another ergonomic layout for German that is not installed by default, and the installer for bépo and neo seems use the same installer on Windows.


awakecoding:
@freerdp-bot test.

Issue #2651, opened 2015-05-25
iOS SSL Script issue 


phongleq:
So I've notice that when using the scripts to generate to do the SSL part on iOS, that after building for i386 things something weird is happening. It looks like after copying libcrypto.a and libssl.a to libssl_i386.a and the other file, both files are then removed from their original folder. This prevents from the others being made. So in my instance, it fail to generate one for armv7 and armv7s because the file was not longer there to be copied. 

I'm able to get around this because I have a older version that had generate them correctly, but it looks like this new master script is off. Im not proficient in scripting to know how to fix it.

lemonmojo:
+1

bmiklautz:
@phongleq you are referring to scripts/OpenSSL-DownloadAndBuild.command ?

phongleq:
Yes

bmiklautz:
@phongleq after the script run successfully did you check the folder external/openssl/lib/? In there you should fine two Mach-O universal binaries (libssl.a and libcrypto.a) which contain *all* built architectures. I just tested on our build and test machine. If this doesn't work for you can you please add a more detailed information on the exact error you are getting ("fail to generate one for armv7.." isn't a very exact error description ;()

phongleq:
There are two binaries there, but after i create the Xcode project Xcode complains that the openssl is not build for armv7 or armv7s. I get this when building the script:

```
Building architecture armv7. Please wait ...
Done. Build log saved in BuildLog.darwin-armv7.txt
cp: libcrypto.a: No such file or directory
cp: libssl.a: No such file or directory

Building architecture armv7s. Please wait ...
Done. Build log saved in BuildLog.darwin-armv7s.txt
cp: libcrypto.a: No such file or directory
cp: libssl.a: No such file or directory
```

So im guessing they were not added to the universal binary. So I looked into the logs and it seems the logs had this as the error:

```clang: error: invalid version number in '-miphoneos-version-min=.sd'```

I was previously able to do this last year with Xcode 5 and iOS7. So I didn't think there was a minimum specs needed.

bmiklautz:
@phongleq can you test if the branch ios/openssl in my repository fixes your problem? -https://github.com/bmiklautz/FreeRDP/tree/ios/openssl

DonMag:
[...]

Issue #2680, opened 2015-06-05
arm64 Build OpenSSL


phongleq:
How would I go at making a the ssl for arm64?

bmiklautz:
@phongleq this is more a openssl than FreeRDP specific question. If you search the web you can find a lot of references on this topic.

phongleq:
Thanks for the response. I did tried different version of openssl that would compile a arm64 version of it. My problem was that after doing so I would never be able to connect with it like I would with the script provided in FreeRDP. I know that there is a patch that is applied to the openssl in FreeRDP and wasn't sure if there was something special I would need to do to get a different version of OpenSSL to work with the project.

bmiklautz:
@phongleq the applied patch was only a work around in order to build openssl. Can you test if the branch ios/openssl in my repository fixes your problem? -https://github.com/bmiklautz/FreeRDP/tree/ios/openssl

phongleq:
Cool it works perfectly! Though I am having issues with building the project for ARM64. I added to the arm64 to the iOSToolchain.cmake file like so:

```set (IOS_ARCH arm64 armv7 armv7s)```

Im not sure if there is more to do than that , but I do get a crash with winpr library. It crashes at the line with this:

```*pType = wHandle->Type;```

in the handler.h file

phongleq:
Ok nevermind I found out that you could do this https://github.com/FreeRDP/FreeRDP/issues/2162

phongleq:
So I got it setup using the new script and ran into a issue when trying to connect to a remote desktop

```[11:13:09:971] [219:6119f000] [ERROR][com.freerdp.core.connection] - Error: protocol security negotiation or connection failure
2015-06-10 11:13:09.543 iFreeRDP[219:7104] ios_run_freerdp: inst->rdp_connect failed```

akallabeth:
Current master script builds OpenSSL for arm64

PR #2686, opened 2015-06-09
ios: update openssl build script

bmiklautz:
* remove patch - not required anymore
* add support for arm64 and x86_64 openssl builds
* update documentation


bmiklautz:
Fixes #2651 and #2680

PR #2687, opened 2015-06-10
Store SSL fingerprints with host and port

akallabeth:
* Can now store SSL fingerprints for each <host> <port> combination.
* Using new ```known_host.v2``` file to store.
* Already existing ```known_host``` entries are added on first use to ```known_host.v2``` if the fingerprint matches.
* Added unit test to ensure functionality
* Storing subject and issuer now to display on ```VerifyChangedCertificate``` callback.
* Adjusted arguments of ```VerifyChangedCertificate```

### DIFF
--- a/include/freerdp/crypto/certificate.h
+++ b/include/freerdp/crypto/certificate.h
@@ -44,6 +44,7 @@ struct rdp_certificate_store
 	FILE* fp;
 	char* path;
 	char* file;
+	char* legacy_file;
 	rdpSettings* settings;
 	rdpCertificateData* certificate_data;
 };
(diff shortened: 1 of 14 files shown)


hardening:
+1

bmiklautz:
@akallabeth I haven't had time to look at the whole pull request but I'd personally would find it nice if we call the file know_hosts2 instead of known_host.v2 (similar to ssh).

akallabeth:
@bmiklautz done ;)

bmiklautz:
@llyzs did you test the different possible situations, like if a key was found in the old file, before merging?

llyzs:
I have only tested the new file format which is working fine, but did not actually test the legacy file case.

bmiklautz:
@llyzs thank you. I'll do some more testing for the conversion that is done for the old format.

PR #2688, opened 2015-06-10
Correct the included files to have a working freerdp-nightly-dev package

hardening:


packaging/deb/freerdp-nightly/freerdp-nightly-dev.install
```diff
@@ -1,4 +1,5 @@
-opt/freerdp-nightly/lib/*.so
```
bmiklautz: Without this the development package won't ship the .so files.

packaging/deb/freerdp-nightly/freerdp-nightly.install
```diff
@@ -1,3 +1,3 @@
-opt/freerdp-nightly/lib/*.so.*
```
bmiklautz: This was intentional. The library it self is in the freerdp-nigthly package but the .so files (without version), which are only required for compilation, are shipped with the freerdp-nightly-dev. This is the way debian (and also rpm packages) usually do it.

bmiklautz:
@hardening shipping the .a files is ok but I don't agree on the packaging changes for the so files.

hardening:
@bmiklautz I agree, I'm gonna revert the wrong changes

bmiklautz:
@hardening thanks for your efforts. I'll merge it now but I have also created an issue (#2695) that we don't forget to fix the real problem in cmake.

bmiklautz:
@hardening the packages will be built this night and should be available by tomorrow.

PR #2690, opened 2015-06-11
Fix timer precision for network autodetection on Windows

llyzs:
On Windows, GetTickCount is only at 16ms precision, and it was causing the detected bandwidth much smaller than the actual one. This PR adds a high-precision version of the function and use it in the autodetect module.

### DIFF
--- a/winpr/libwinpr/sysinfo/sysinfo.c
+++ b/winpr/libwinpr/sysinfo/sysinfo.c
@@ -678,6 +678,21 @@ BOOL IsProcessorFeaturePresent(DWORD ProcessorFeature)
 
 #endif //_WIN32
 
+DWORD GetTickCountPrecise(void)
+{
+#ifdef _WIN32
+	LARGE_INTEGER freq;
+	LARGE_INTEGER current;
+
+	QueryPerformanceFrequency(&freq);
+	QueryPerformanceCounter(&current);
+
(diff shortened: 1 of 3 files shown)


MartinHaimberger:
+1 LGTM

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -678,6 +678,21 @@ BOOL IsProcessorFeaturePresent(DWORD ProcessorFeature)
...
+#ifdef _WIN32
+	LARGE_INTEGER freq;
+	LARGE_INTEGER current;
+
+	QueryPerformanceFrequency(&freq);
+	QueryPerformanceCounter(&current);
+
+	return (DWORD) (current.QuadPart * 1000LL / freq.QuadPart);
+#else
+	return GetTickCount();
```
hardening: Linux also have high precision timers ?

llyzs: On Linux, GetTickCount() is implemented using clock_gettime(CLOCK_MONOTONIC) which I believe is already high-precision, so we just have to wrap the new API to it directly. It is only on Windows that GetTickCount() has such problem.

bmiklautz:
+1

PR #2691, opened 2015-06-11
known_host saving host and port.

oshogbo:
In my opinion we should save in known_host pair host and ip (like OpenSSH do this).
This allow us to connect to multiple RDP servers which have same IP but different port.

### DIFF
--- a/include/freerdp/crypto/certificate.h
+++ b/include/freerdp/crypto/certificate.h
@@ -35,6 +35,8 @@ typedef struct rdp_certificate_store rdpCertificateStore;
 struct rdp_certificate_data
 {
 	char* hostname;
+	int port;
+	char* hostnamePort;
 	char* fingerprint;
 };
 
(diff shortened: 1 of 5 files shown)


hardening:
+1 
Looks like what misc people wanted in misc issues.

hardening:
@freerdp-bot would you test that for me please ?

bmiklautz:
@oshogbo there is a other pull request from @akallabeth which does basically the same.

oshogbo:
@bmiklautz When I start to work on it there was no pull request (yesterday), I pull request after testing it which I was able do today and didn't notice that sb also work on same patch.

It's fine you can use @akallabeth patch.

oshogbo:
And i see windows have some problem with snprintf(3) function so maybe @akallabeth better solve that.

akallabeth:
@oshogbo If you are ok with my pull #2687, could you please give it a try, close yours and comment on any issues (or none) with it?

PR #2693, opened 2015-06-11
rdpsnd/pulse: Fix crash if device isn't specified

ondrejholy:
Freerdp crashes if pulseaudio device isn't specified on commandline, i.e.:
xfreerdp --plugin rdpsnd --data pulse

The crash occurs, because data[1] is NULL and it is passed into strlen
function. We have to check whether data[1] was set before calling strlen
to fix this crash.


bmiklautz:
@freerdp-bot test.

PR #2694, opened 2015-06-11
rdpsnd/pulse: Fix crashes in pulseaudio

ondrejholy:
Function pa_stream_trigger has to be called under lock to avoid
following crashes on asserts:

Assertion 'e->mainloop->n_enabled_defer_events > 0' failed at pulse/mainloop.c:257, function mainloop_defer_enable(). Aborting.
Assertion '!e->next' failed at pulsecore/queue.c:104, function pa_queue_pop(). Aborting.
Assertion 'q->front' failed at pulsecore/queue.c:81, function pa_queue_push(). Aborting.

### DIFF
--- a/channels/rdpsnd/pulse/rdpsnd_pulse.c
+++ b/channels/rdpsnd/pulse/rdpsnd_pulse.c
@@ -452,7 +452,9 @@ static void rdpsnd_pulse_start(rdpsndDevicePlugin* device)
 	if (!pulse->stream)
 		return;
 
+	pa_threaded_mainloop_lock(pulse->mainloop);
 	pa_stream_trigger(pulse->stream, NULL, NULL);
+	pa_threaded_mainloop_unlock(pulse->mainloop);
 }
 
 int FreeRDPRdpsndDeviceEntry(PFREERDP_RDPSND_DEVICE_ENTRY_POINTS pEntryPoints)


ondrejholy:
The crashes occurs only in stable-1.0, but the patch would be good to apply also on stable-1.1 and master...

channels/rdpsnd/pulse/rdpsnd_pulse.c
```diff
@@ -449,10 +449,12 @@ static void rdpsnd_pulse_start(rdpsndDevicePlugin* device)
 {
 	rdpsndPulsePlugin* pulse = (rdpsndPulsePlugin*)device;
 
+	pa_threaded_mainloop_lock(pulse->mainloop);
```
bmiklautz: If pulse->stream isn't set wouldn't the following return keep the mainloop locked?
I'd assume that the right place for the lock would be after the check.

ondrejholy:
You are right, I am sorry, I was distracted when pushing the commit... It should be ok now :-)

joakim-tjernlund:
Looks like this is NOT in current master, is it not needed anymore?

ondrejholy:
@joakim-tjernlund I was not able to reproduce with master, so I did not make pull request for master, but it would be probably good to apply it on master for sure... will take a look at it later...

joakim-tjernlund:
I got it[1] with SKY which have some version of freerdp2, I think they have their own hacks in it too.
[1] = Assertion 'q->front' failed at pulsecore/queue.c:81, function pa_queue_push(). Aborting.

joakim-tjernlund:
Using ALSA in pidgin-sipe sometimes produce choppy/distorted audio and I wonder if freerdp
might suffer from the same bug VLC does/did:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=734100

bmiklautz:
@joakim-tjernlund how is the ALSA comment from above related to the pulse audio fix?

joakim-tjernlund:
It doesn't, oops. There is a audio problem in 
  https://github.com/tieto/sipe/issues/107
and I just hand the thought the fix from VLC might be applicable(move locking around) for freerdp

PR #2696, opened 2015-06-12
Fix Microsoft Account authentication failures

akallabeth:
When working from a Microsoft account, it is not possible to authenticate to another machine because the wrong domain context is used.
* This patch sets a sensible default value for the ```domain``` argument.
* Fixes the command line parser, now setting default values if present.


akallabeth:
@awakecoding Could you have a look at this, I'm not quite sure this has no side effects. Would be great to fix this correctly before merging. Thx

PR #2697, opened 2015-06-14
Shadow server: fix channel disposal and add audio support. fix message mechanism in client

realjiangms:
This PR originally is used to add audio support, however some fix in shadow_client is related to it.
So also include several fixes:
1. Introduce message queue in shadow client. No longer use subsytem->MsgPipe->out to deliver message to clients.
We used to use subsytem->MsgPipe->out for messages which need to be sent to client. But it's not correct. Only one client would get the message if multiple client exists
This problem make the fix in PR #2643 incomplete.
Introduced reference count based solution to release resource taken by the message.
Also added APIs for client message delivery.
Also fixed msg pipe in subsystem to clean resource when destroyed.
2. Discard unused StopEvent in client. We actually use quit message instead.
3. Enhance disposal of channels.
Free context for remdesk and encomsp channels. The original fix only stop the threads, but doesn't release resource.
Dispose channels earlier. The channels are built on client->vcm. Disposing channels after client->vcm is closed cause unknown behavior.
Original fix is #2644
4. Start to add audio support.

### DIFF
--- a/server/shadow/X11/x11_shadow.c
+++ b/server/shadow/X11/x11_shadow.c
@@ -350,14 +350,21 @@ void x11_shadow_input_extended_mouse_event(x11ShadowSubsystem* subsystem, UINT16
 
 static void x11_shadow_message_free(UINT32 id, SHADOW_MSG_OUT* msg)
 {
-	if (id == SHADOW_MSG_OUT_POINTER_POSITION_UPDATE_ID)
+	switch(id)
 	{
-		free(msg);
-	}
-	else if (id == SHADOW_MSG_OUT_POINTER_ALPHA_UPDATE_ID)
-	{
-		free(((SHADOW_MSG_OUT_POINTER_ALPHA_UPDATE*)msg)->pixels);
-		free(msg);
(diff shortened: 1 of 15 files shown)


server/shadow/X11/x11_shadow.c
```diff
@@ -348,11 +348,23 @@ void x11_shadow_input_extended_mouse_event(x11ShadowSubsystem* subsystem, UINT16
 #endif
 }
 
+static void x11_shadow_message_free(UINT32 id, SHADOW_MSG_OUT* msg)
+{
+	if (id == SHADOW_MSG_OUT_POINTER_POSITION_UPDATE_ID)
```
hardening: a

hardening:
@freerdp-bot can you do a test build of this, please ?

server/shadow/X11/x11_shadow.c
realjiangms: Yeah all message id should be put in switch block.
Blame my inertial thinking

PR #2698, opened 2015-06-14
server/shadow: Decouple shadow_subsystem and subsystem implementation detail.

realjiangms:
server/shadow: Decouple shadow_subsystem and subsystem implementation detail.

Make shadow_subsystem standalone so that others can implement customized shadow server with shadow library.


realjiangms:
@awakecoding 
This is the PR I mentioned to you before: Make shadow framework more generic and export the shadow_subsystem

PR #2699, opened 2015-06-14
server/shadow: add connect/disconnect anchor in subsystem callback

realjiangms:
server/shadow: add connect/disconnect anchor in subsystem callback. 
Other implementation of subsystem may want to track connect/disconnect of each client

### DIFF
--- a/server/shadow/shadow_client.c
+++ b/server/shadow/shadow_client.c
@@ -250,10 +250,7 @@ BOOL shadow_client_post_connect(freerdp_peer* peer)
 
 	if (subsystem->ClientConnect)
 	{
-		if (subsystem->ClientConnect(subsystem, peer))
-			return TRUE;
-		else
-			return FALSE;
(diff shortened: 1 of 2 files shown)


server/shadow/shadow_client.c
```diff
@@ -248,6 +248,14 @@ BOOL shadow_client_post_connect(freerdp_peer* peer)
 		}
 	}
 
+	if (subsystem->ClientConnect)
+	{
+		if (subsystem->ClientConnect(subsystem, peer))
```
hardening: return subsystem->ClientConnect(subsystem, peer) ?

hardening:
@freerdp-bot I'd like you to test this before doing a +1

server/shadow/shadow_client.c
realjiangms: Yeah It should be. 
I want this callback as the last line in post connect, if subsystem->ClientConnect exists.

PR #2701, opened 2015-06-15
Allowing UPN syntax for user argument.

akallabeth:
* Fixed issues with authentication:
* If ```user@domain``` syntax is used, the domain is set to an empty string instead of ```NULL``` and user to the full string
* If only ```user``` is given, set domain to ```TARGET``` (similar to ```CredUIParseUserName```)
* If ```domain\user``` syntax is used, split the string.

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -492,8 +492,6 @@ BOOL wf_post_connect(freerdp* instance)
 	return TRUE;
 }
 
-static const char wfTargetName[] = "TARGET";
-
 static CREDUI_INFOA wfUiInfo =
 {
 	sizeof(CREDUI_INFOA),
@@ -518,7 +516,9 @@ BOOL wf_authenticate(freerdp* instance, char** username, char** password, char**
 	ZeroMemory(Password, sizeof(Password));
 	dwFlags = CREDUI_FLAGS_DO_NOT_PERSIST | CREDUI_FLAGS_EXCLUDE_CERTIFICATES;
 
-	status = CredUIPromptForCredentialsA(&wfUiInfo, wfTargetName, NULL, 0,
+	status = CredUIPromptForCredentialsA(&wfUiInfo,
+					     instance->settings->ServerHostname,
+					     NULL, 0,
 		UserName, CREDUI_MAX_USERNAME_LENGTH + 1,
 		Password, CREDUI_MAX_PASSWORD_LENGTH + 1, &fSave, dwFlags);
 
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -537,6 +537,8 @@ BOOL wf_authenticate(freerdp* instance, char** username, char** password, char**
 
 	if (strlen(Domain) > 0)
 		*domain = _strdup(Domain);
+	else
+		*domain = _strdup("\0");
 
 	*password = _strdup(Password);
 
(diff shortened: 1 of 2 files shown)


client/common/cmdline.c
```diff
@@ -553,7 +553,7 @@ static char** freerdp_command_line_parse_comma_separated_values_offset(char* lis
 		return NULL;
 	p = t;
 	if (count > 0)
-        MoveMemory(&p[1], p, sizeof(char*) * *count);
+				MoveMemory(&p[1], p, sizeof(char*) * *count);
```
mfleisz: too many tabs here?

mfleisz:
+1 (just a little formatting remark)

client/Windows/wf_client.c
```diff
@@ -537,6 +537,8 @@ BOOL wf_authenticate(freerdp* instance, char** username, char** password, char**
 
 	if (strlen(Domain) > 0)
 		*domain = _strdup(Domain);
+	else
+		*domain = _strdup("\0");
```
bmiklautz: The first is not your code but those strdups should be checked (for the original code this is done in PR #2708)

client/common/cmdline.c
```diff
@@ -795,27 +795,61 @@ int freerdp_client_command_line_post_filter(void* context, COMMAND_LINE_ARGUMENT
 int freerdp_parse_username(char* username, char** user, char** domain)
 {
 	char* p;
-	int length;
+	char* u;
+	int length = 0;
 
 	p = strchr(username, '\\');
+	p = strrchr(username, '@');
+
```
bmiklautz: If I'm not mistaken you would always only handle the @ case.

client/common/cmdline.c
```diff
@@ -795,27 +795,61 @@ int freerdp_client_command_line_post_filter(void* context, COMMAND_LINE_ARGUMENT
...
-		*domain = NULL;
+		if (!*user)
+			return -1;
+
+		/* If only username is given, prefix that with 'TARGET'
+		 * otherwise set the domain to an empty string.
+		 * NOTE: Domain NULL will result in undefined behavior.
+		 */
+		if (!u)
+			*domain = _strdup("TARGET");
```
bmiklautz: Although it seems to work this feels really weird. This would mean that you can't use TARGET as domain name.
@awakecoding do you know something about that or have any ideas?

client/common/cmdline.c
```diff
@@ -1459,13 +1495,7 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
 		CommandLineSwitchCase(arg, "u")
 		{
-			char* user;
-			char* domain;
-
-			freerdp_parse_username(arg->Value, &user, &domain);
-
-			settings->Username = user;
-			settings->Domain = domain;
+			user = _strdup(arg->Value);
```
bmiklautz: Missing strdup check.

client/common/cmdline.c
```diff
@@ -1506,14 +1536,7 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
 		{
-			char* user;
-			char* domain;
-
-			freerdp_parse_username(arg->Value, &user, &domain);
-
-			settings->GatewayUsername = user;
-			settings->GatewayDomain = domain;
-
+			gwUser = _strdup(arg->Value);
```
bmiklautz: strdup return value not handled.

PR #2702, opened 2015-06-15
NLA Use current users credentials if none provided.

akallabeth:
When client side NLA is started without username, the identity is set to NULL.
This way the current user credentials are used on windows.

### DIFF
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -90,14 +90,15 @@
 
 #define TERMSRV_SPN_PREFIX	"TERMSRV/"
 
-BOOL nla_send(rdpNla* nla);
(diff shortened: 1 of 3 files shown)


hardening:
I have not tested it, but I've just touched this file in my "check return value of malloc()" patch. So it looks good to me.

Note: I have done a lot of fixes in this file

libfreerdp/core/nla.c
```diff
@@ -1017,9 +1050,9 @@ BOOL nla_encode_ts_credentials(rdpNla* nla)
 
 	if (nla->settings->DisableCredentialsDelegation)
```
mfleisz: check for nla->identity != NULL here too

mfleisz:
+1

PR #2705, opened 2015-06-17
Sspi fix v2 - Fixes static build

akallabeth:


mfleisz:
+1

CMakeLists.txt
```diff
@@ -229,7 +229,7 @@ if(CMAKE_COMPILER_IS_GNUCC)
 	endif()
 endif()
 
-if("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang")
+if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang")
```
bmiklautz: Side note: This is because of http://www.cmake.org/cmake/help/v3.1/policy/CMP0054.html

PR #2706, opened 2015-06-17
winpr/synch: fix ResetEvent & improve CTest

nfedera:
Unless there is an invalid parameter or internal error, ResetEvent must return TRUE, even if the object is in nonsignaled state.

Also fixed and improved the CTest TestSynchEvent to test for the expected/correct SetEvent and ResetEvent return values.


hardening:
+1 LGTM

mfleisz:
+1

akallabeth:
+1

PR #2711, opened 2015-06-19
core: message channel pdu broken with rdp security

nfedera:
rdp_recv_message_channel_pdu always read the rdp security header even if it was already previously read (which is the case if rdp security is active)

This caused malfunctions and disconnects when heartbeat or bandwidth autodetect packets were sent/received in rdp security mode.

Credit goes to @MartinHaimberger for identifying the broken code part.


bmiklautz:
+1

PR #2713, opened 2015-06-22
core/metrics: prevent division by zero

bmiklautz:
With most compilers this shouldn't happen but on some architectures a division by zero (can happen on the first call of metrics_write_bytes) could still cause a SIGFPE.

### DIFF
--- a/libfreerdp/core/metrics.c
+++ b/libfreerdp/core/metrics.c
@@ -30,8 +30,10 @@ double metrics_write_bytes(rdpMetrics* metrics, UINT32 UncompressedBytes, UINT32
 	metrics->TotalUncompressedBytes += UncompressedBytes;
 	metrics->TotalCompressedBytes += CompressedBytes;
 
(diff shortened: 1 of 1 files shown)


libfreerdp/core/metrics.c
```diff
@@ -30,8 +30,10 @@ double metrics_write_bytes(rdpMetrics* metrics, UINT32 UncompressedBytes, UINT32
 	metrics->TotalUncompressedBytes += UncompressedBytes;
 	metrics->TotalCompressedBytes += CompressedBytes;
 
-	CompressionRatio = ((double) CompressedBytes) / ((double) UncompressedBytes);
-	metrics->TotalCompressionRatio = ((double) metrics->TotalCompressedBytes) / ((double) metrics->TotalUncompressedBytes);
+	if (UncompressedBytes != 0)
+		CompressionRatio = ((double) CompressedBytes) / ((double) UncompressedBytes);
```
realjiangms: Maybe it's redundancy comment: Shouldn't we initialize 'CompressionRatio' since it's possible that we never set it?

PR #2716, opened 2015-06-23
server/shadow: Fix alpha pointer message so that it allows pixel with both 'xor' and 'and'

realjiangms:
server/shadow: Fix alpha pointer message so that it allows pixel with both 'xor' and 'and'.

Original pointer alpha update message is too specific for x11 implementation.
It doesn't allow pointer alpha mask with both 'xor' color and 'and' mask, e.g.: the 'edit' pointer in windows text box.

1. Move shadow_client_convert_alpha_pointer_data to x11 implementation as it is specific for x11.
2. Update message definition to be more generic: to accept 'xor/and' mask directly.
Implementation of subsystem can implement its own way to convert pointer mask data.
3. Fixed fault handling to free the resource allocated.

### DIFF
--- a/server/shadow/shadow_client.c
+++ b/server/shadow/shadow_client.c
@@ -819,85 +819,6 @@ int shadow_client_surface_update(rdpShadowClient* client, REGION16* region)
 	return 1;
 }
 
-int shadow_client_convert_alpha_pointer_data(BYTE* pixels, BOOL premultiplied,
-		UINT32 width, UINT32 height, POINTER_COLOR_UPDATE* pointerColor)
-{
-	UINT32 x, y;
-	BYTE* pSrc8;
-	BYTE* pDst8;
-	int xorStep;
-	int andStep;
-	UINT32 andBit;
-	BYTE* andBits;
-	UINT32 andPixel;
-	BYTE A, R, G, B;
-
-	xorStep = (width * 3);
-	xorStep += (xorStep % 2);
-
-	andStep = ((width + 7) / 8);
-	andStep += (andStep % 2);
-
-	pointerColor->lengthXorMask = height * xorStep;
-	pointerColor->xorMaskData = (BYTE*) calloc(1, pointerColor->lengthXorMask);
-
-	if (!pointerColor->xorMaskData)
-		return -1;
-
-	pointerColor->lengthAndMask = height * andStep;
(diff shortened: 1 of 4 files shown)


server/shadow/shadow_client.c
```diff
@@ -817,85 +817,6 @@ int shadow_client_surface_update(rdpShadowClient* client, REGION16* region)
 	return 1;
 }
 
-int shadow_client_convert_alpha_pointer_data(BYTE* pixels, BOOL premultiplied,
```
realjiangms: Note: This is almost completely moved to X11/x11_shadow.c

bmiklautz:
@freerdp-bot test

server/shadow/shadow_client.c
bmiklautz: I was wondering if it would't have made sense to keep as much in the generic part as possible that different concrete implementations can re-use it?

realjiangms: Sure, how about putting it in shadow_subsystem.c as a common function? Other subsystem implementations could use it.

realjiangms: I'll update the PR.

bmiklautz: @realjiangms was just an idea but your suggestion sounds good for me ;).

realjiangms: I actually hesitated to keep it private for X11 or common for all. Your suggestion give me additional reason to make the decision :)

realjiangms:
@bmiklautz 
Updated PR. We may need to launch test again.

bmiklautz:
@freerdp-bot test

bmiklautz:
@realjiangms looks very good to me. Thanks
+1

bmiklautz:
@freerdp-bot add to whitelist

bmiklautz:
@realjiangms I've white listed you now. Your pulls get now (re-)built automatically.

realjiangms:
@bmiklautz Thanks.
Does it mean that later I can launch test myself by adding a comment with 'freerdp-bot' and 'test'?

bmiklautz:
@realjiangms the bot should automatically build all of your pull requests as well as if you later commit to an open pull request.

realjiangms:
@bmiklautz Can we have this PR merged or we need more reviewer?

PR #2723, opened 2015-06-23
New /from-stdin

akallabeth:
* Adds ```/from-stdin``` to ignore missing ```user```, ```domain``` and ```password```
* Implemented ```GatewayAuthenticate``` callbacks for windows and X11 clients
* Implemented full ```Authenticate``` callback for X11

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -1252,7 +1252,12 @@ static BOOL xf_authenticate_raw(freerdp* instance, BOOL gateway, char** username
 	{
 		size_t username_size = 0;
 		printf("%s", prompt[0]);
-		getline(username, &username_size, stdin);
+		if (getline(username, &username_size, stdin) < 0)
+		{
+			WLog_ERR(TAG, "getline returned %s [%d]", strerror(errno), errno);
+			goto fail;
+		}
+
 		if (*username)
 		{
 			*username = StrSep(username, "\r");
@@ -1264,7 +1269,12 @@ static BOOL xf_authenticate_raw(freerdp* instance, BOOL gateway, char** username
 	{
 		size_t domain_size = 0;
 		printf("%s", prompt[1]);
-		getline(domain, &domain_size, stdin);
+		if (getline(domain, &domain_size, stdin) < 0)
+		{
+			WLog_ERR(TAG, "getline returned %s [%d]", strerror(errno), errno);
+			goto fail;
+		}
+
 		if (*domain)
 		{
 			*domain = StrSep(domain, "\r");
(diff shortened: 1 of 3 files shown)


client/common/cmdline.c
```diff
@@ -169,6 +169,7 @@ COMMAND_LINE_ARGUMENT_A args[] =
 	{ "multitransport", COMMAND_LINE_VALUE_BOOL, NULL, BoolValueFalse, NULL, -1, NULL, "Support multitransport protocol" },
 	{ "assistance", COMMAND_LINE_VALUE_REQUIRED, "<password>", NULL, NULL, -1, NULL, "Remote assistance password" },
 	{ "encryption-methods", COMMAND_LINE_VALUE_REQUIRED, "<40,56,128,FIPS>", NULL, NULL, -1, NULL, "RDP standard security encryption methods" },
+	{ "from-stdin", COMMAND_LINE_VALUE_FLAG, NULL, NULL, NULL, -1, NULL, "print version" },
```
bmiklautz: I guess the help should be different than "print version"

mfleisz:
+1

client/X11/xf_client.c
```diff
@@ -1228,18 +1228,74 @@ static void xf_post_disconnect(freerdp* instance)
...
+			*username = StrSep(username, "\r");
+			*username = StrSep(username, "\n");
+		}
+	}
+
+	if (!*domain)
+	{
+		size_t domain_size = 0;
+		printf("%s", prompt[1]);
+		getline(domain, &domain_size, stdin);
```
MartinHaimberger: result of getline should be checked and on failture the buffers should be freed (see doc)

 If *lineptr is set to NULL and *n is set 0 before the call, then
       getline() will allocate a buffer for storing the line.  This buffer
       should be freed by the user program even if getline() failed.

client/X11/xf_client.c
```diff
@@ -1228,18 +1228,74 @@ static void xf_post_disconnect(freerdp* instance)
...
 
-	if (freerdp_passphrase_read("Password: ", *password, password_size, instance->settings->CredentialsFromStdin) == NULL)
+	if (!username || !password || !domain)
 		return FALSE;
 
+	if (!*username)
+	{
+		size_t username_size = 0;
+		printf("%s", prompt[0]);
+		getline(username, &username_size, stdin);
```
MartinHaimberger: result of getline should be checked and on failture the buffers should be freed (see doc)

If *lineptr is set to NULL and *n is set 0 before the call, then
getline() will allocate a buffer for storing the line. This buffer
should be freed by the user program even if getline() failed.

MartinHaimberger:
besides my remarks LGTM

bmiklautz:
Although I'm very happy we have that again this PR unfortunately changes the default behavior. Usually , when no additional parameter were given, only the password was prompted. Now the domain is prompted as well. But for example if the username is given the Domain isn't asked for.
I'm not sure if we wan't need this in that form.  IMHO we should have a consistent way.
```
xfreerdp /v:127.0.0.1
[11:32:10:651] [25677:25678] [INFO][com.freerdp.client.x11] - No user name set. - Using login name:  xxxx
Domain:    
Password: 
```

PR #2724, opened 2015-06-23
Fix leaks in certificate and identity handling

bmiklautz:
Thanks to @akallabeth for finding/fixing the nla_identity_free.

### DIFF
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -108,9 +108,21 @@ void nla_identity_free(SEC_WINNT_AUTH_IDENTITY* identity)
 {
 	if (identity)
 	{
-		free(identity->User);
-		free(identity->Domain);
-		free(identity->Password);
+		if (identity->User)
+		{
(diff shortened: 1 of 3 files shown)


akallabeth:
+1

mfleisz:
+1

nfedera:
+1 and just two non-blocking remarks
- from a security point of view it might make sense that nla_identity_free clears the user, password, domain blobs before releasing the memory
- nla_identity_free() frees a SEC_WINNT_AUTH_IDENTITY and seems to me independent of nla (not sure if that is true). However, if that's the case it might be good to rename it to sec_winnt_auth_identity_free().

bmiklautz:
@nfedera the password and username is cleared now. Regarding the function name you are right it's not really suiting but all functions are called nla_ in this file and it's a static function therefore I think it's ok.

bmiklautz:
@freerdp-bot test it right this time!

nfedera:
Another +1 now

PR #2727, opened 2015-06-23
Fixed leaks, NULL dereferences and broken init.

akallabeth:
Some fixed found during static code analysis.

### DIFF
--- a/libfreerdp/codec/rfx.c
+++ b/libfreerdp/codec/rfx.c
@@ -966,10 +966,9 @@ static BOOL rfx_process_message_tileset(RFX_CONTEXT* context, RFX_MESSAGE* messa
 			WaitForThreadpoolWorkCallbacks(work_objects[i], FALSE);
 			CloseThreadpoolWork(work_objects[i]);
 		}
-
-		free(work_objects);
-		free(params);
 	}
+	free(work_objects);
+	free(params);
 
 	for (i = 0; i < message->numTiles; i++)
 	{
--- a/libfreerdp/codec/test/TestFreeRDPCodecProgressive.c
+++ b/libfreerdp/codec/test/TestFreeRDPCodecProgressive.c
@@ -134,6 +134,16 @@ static int g_Height = 0;
 static int g_DstStep = 0;
 static BYTE* g_DstData = NULL;
 
+static void sample_file_free(EGFX_SAMPLE_FILE* file)
+{
+	if (!file)
+		return;
+
+	free(file->buffer);
+	file->buffer = NULL;
+	file->size = 0;
+}
+
 static void test_fill_image_alpha_channel(BYTE* data, int width, int height, BYTE value)
 {
 	int i, j;
@@ -940,6 +950,7 @@ int test_progressive_decode(PROGRESSIVE_CONTEXT* progressive, EGFX_SAMPLE_FILE f
 
 int test_progressive_ms_sample(char* ms_sample_path)
 {
+	int i, j, k;
 	int count;
 	int status;
 	EGFX_SAMPLE_FILE files[3][4][4];
@@ -956,12 +967,34 @@ int test_progressive_ms_sample(char* ms_sample_path)
 	status = test_progressive_load_files(ms_sample_path, files);
 
 	if (status < 0)
+	{
+		for(i=0; i<3; i++)
+		{
+			for (j=0; j<4; j++)
+			{
+				for (k=0; k<4; k++)
+					sample_file_free(&files[i][j][k]);
+			}
+		}
+
 		return -1;
+	}
 
 	status = test_progressive_load_bitmaps(ms_sample_path, bitmaps);
 
 	if (status < 0)
+	{
+		for(i=0; i<3; i++)
+		{
+			for (j=0; j<4; j++)
+			{
+				for (k=0; k<4; k++)
+					sample_file_free(&files[i][j][k]);
+			}
+		}
+
 		return -1;
+	}
 
 	count = 4;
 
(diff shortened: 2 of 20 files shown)


libfreerdp/core/redirection.c
```diff
@@ -207,7 +207,7 @@ int rdp_redirection_apply_settings(rdpRdp* rdp)
 			settings->TargetNetAddresses[i] = _strdup(redirection->TargetNetAddresses[i]);
 			if (!settings->TargetNetAddresses[i])
 			{
-				for (--i; i >= 0; --i)
+				for (; i > 0; --i)
```
mfleisz: Good catch - however I think now we miss out on freeing settings->TargetNetAddresses[0] - I'd propose following solution:

```
UINT32 j;
for (j = 0; j < i; ++j)
    free(settings->TargetNetAddresses[j]);
```

mfleisz:
LGTM +1

libfreerdp/codec/rfx.c
```diff
@@ -966,10 +966,9 @@ static BOOL rfx_process_message_tileset(RFX_CONTEXT* context, RFX_MESSAGE* messa
 			WaitForThreadpoolWorkCallbacks(work_objects[i], FALSE);
 			CloseThreadpoolWork(work_objects[i]);
 		}
-
-		free(work_objects);
-		free(params);
 	}
+	free(work_objects);
+	free(params);
```
bmiklautz: work_objects and params are only allocated if threads are used. The free here doesn't hurt but wasn't a leak either before imho.

libfreerdp/codec/test/TestFreeRDPCodecProgressive.c
```diff
@@ -1009,7 +1063,7 @@ int TestFreeRDPCodecProgressive(int argc, char* argv[])
 {
 	char* ms_sample_path;
 
-	ms_sample_path = _strdup("/tmp/EGFX_PROGRESSIVE_MS_SAMPLE");
+	ms_sample_path = GetKnownSubPath(KNOWN_PATH_TEMP, "EGFX_PROGRESSIVE_MS_SAMPLE");
```
bmiklautz: nice!

libfreerdp/common/assistance.c
```diff
@@ -184,7 +184,7 @@ int freerdp_assistance_parse_address_list(rdpAssistanceFile* file, char* list)
 		file->MachinePorts[i] = (UINT32) atoi(q);
 
 		if (!file->MachineAddresses[i])
-			return -1;
+			goto out;
```
bmiklautz: good catch.
We should also clean up the rest (file->MachineAddresses[i]) if a strdup fails maybe you can add that as well.

libfreerdp/core/redirection.c
```diff
@@ -207,8 +207,10 @@ int rdp_redirection_apply_settings(rdpRdp* rdp)
 			settings->TargetNetAddresses[i] = _strdup(redirection->TargetNetAddresses[i]);
 			if (!settings->TargetNetAddresses[i])
 			{
-				for (--i; i >= 0; --i)
-					free(settings->TargetNetAddresses[i]);
+				UINT32 j;
+
+				for (j=0; j < i; j++)
+					free(settings->TargetNetAddresses[j]);
```
bmiklautz: What's wrong with the previous approach in your opinion? - Does effectively the same and saves a variable.

mfleisz: It will loop forever because i is an UINT32 and will never be < 0

winpr/libwinpr/utils/collections/Reference.c
```diff
@@ -157,19 +157,19 @@ wReferenceTable* ReferenceTable_New(BOOL synchronized, void* context, REFERENCE_
...
-
 	referenceTable->context = context;
 	referenceTable->ReferenceFree = ReferenceFree;
 
 	referenceTable->size = 32;
 
+	referenceTable->array = (wReference*) calloc(referenceTable->size, sizeof(wReference));
+	if (!referenceTable->array)
+		goto error_array;
+
```
bmiklautz: Very good catch!

libfreerdp/core/redirection.c
bmiklautz: muhahah. right! - Haven't looked at the type. Nice. Thanks for pointing that out.

libfreerdp/codec/rfx.c
akallabeth: @bmiklautz yes, but it eliminates a false positive in static code analysis.

realjiangms: If thread is not used, the params and work_objects would be NULL.
Not sure if free() abort with NULL

bmiklautz: @realjiangms free works fine with NULL. According to the C standard:
```
If ptr is a null pointer, no action occurs.
```

libfreerdp/core/redirection.c
nfedera: @mfleisz nice finding but what would be wrong with this:
```
while (i) {
	free(settings->TargetNetAddresses[--i])
}
```

mfleisz: @nfedera That's fine as well!

libfreerdp/gdi/dc.c
```diff
@@ -105,7 +105,6 @@ HGDI_DC gdi_CreateDC(UINT32 flags, int bpp)
 
 fail:
 	gdi_DeleteDC(hDC);
-	free(hDC);
```
nfedera: nice catch!

nfedera:
+1

PR #2728, opened 2015-06-24
Fix possible endless loops on cleanup.

bmiklautz:
Some cleanup code possibly create endless loops because an unsigned
type was used as run variable but the check was >= 0 in the for loop.

### DIFF
--- a/libfreerdp/core/message.c
+++ b/libfreerdp/core/message.c
@@ -107,10 +107,10 @@ static BOOL update_message_BitmapUpdate(rdpContext* context, BITMAP_UPDATE* bitm
 		wParam->rectangles[index].bitmapDataStream = (BYTE*) malloc(wParam->rectangles[index].bitmapLength);
 		if (!wParam->rectangles[index].bitmapDataStream)
 		{
-			for (index -= 1; index >= 0; --index)
-			{
-				free(wParam->rectangles[index].bitmapDataStream);
-			}
+			UINT32 i;
+			for (i = 0; i < index; ++i)
+				free(wParam->rectangles[i].bitmapDataStream);
+
 			free(wParam->rectangles);
 			free(wParam);
 			return FALSE;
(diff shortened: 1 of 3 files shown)


mfleisz:
Hehe the joy of copy/paste? ;)
+1

bmiklautz:
@mfleisz don't think so (but don't know) I just checked all for loops in the code with a break condition of <= 0. Just to make sure. And lo and behold I found some ;)

libfreerdp/core/message.c
```diff
@@ -107,10 +107,10 @@ static BOOL update_message_BitmapUpdate(rdpContext* context, BITMAP_UPDATE* bitm
...
 		{
-			for (index -= 1; index >= 0; --index)
-			{
-				free(wParam->rectangles[index].bitmapDataStream);
-			}
+			UINT32 i;
+			for (i = 0; i < index; ++i)
+				free(wParam->rectangles[i].bitmapDataStream);
+
 			free(wParam->rectangles);
```
hardening: Perhaps I'm missing something but as the wParam->rectangles is copied from bitmap->rectangles, in case of OOM we will free the wrong pointers

mfleisz: I don't think we are freeing the wrong pointers - only the ones that have been newly assigned to the BITMAP_DATA struct at line 107 (0 to index - 1).

hardening: you're damn right I commented too fast

hardening:
+1

akallabeth:
+1

nfedera:
+1 thanks for checking !!
However, in order to prevent the additional variable I'd suggest to do the cleanups like this:
```
while (index) {
    free(array[--index].whatever);
}
```

bmiklautz:
@nfedera that's how you meant it?

nfedera:
@bmiklautz perfect!

winpr/libwinpr/clipboard/clipboard.c
```diff
@@ -374,9 +374,12 @@ BOOL ClipboardInitFormats(wClipboard* clipboard)
 
 		if (!format->formatName)
 		{
-			for (--formatId; formatId >= 0; --formatId)
+			int i;
+			for(i = formatId-1; i >= 0; --i)
+			{
+				format = &(clipboard->formats[--clipboard->numFormats]);
```
nfedera: clipboard->numFormats is one to high

bmiklautz: @nfedera thank you. I'v absolutely missed that.

nfedera:
+1 LGTM

PR #2729, opened 2015-06-24
Replaced hard coded registry keys with cmake defines.

akallabeth:
* Using ```FREERDP_VENDOR_STRING``` and ```FREERDP_PRODUCT_STRING``` as base directory now.
* Updated API to allow C char strings (wide character define concatenation is tricky)

NOTE: (default build)
* This moves the registry keys from ```Software\FreeRDP``` to ```Software\FreeRDP\FreeRDP```
* This moves the registry keys from ```Software\WinPR\NTLM``` to ```Software\FreeRDP\FreeRDP\WinPR\NTLM```

### DIFF
--- a/libfreerdp/codec/rfx.c
+++ b/libfreerdp/codec/rfx.c
@@ -61,6 +61,9 @@
 #define RFX_INIT_SIMD(_rfx_context) do { } while (0)
 #endif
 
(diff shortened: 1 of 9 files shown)


hardening:
LGTM +1

mfleisz:
Looks good +1

PR #2735, opened 2015-06-26
[Clipboard] Fixed out of bound array access.

akallabeth:
Found this nasty little bug in clipboard, sending UTF8 string instead of wide character string in ```cliprdr_temp_directory```.
Could anyone with knowledge check if this breaks/fixes anything?


bmiklautz:
+1

PR #2737, opened 2015-06-26
libfreerdp/gdi: fixed gdi return value madness

nfedera:
Mostly booleanization to comply with the MS API


bmiklautz:
@freerdp-bot test this right now, pretty please!

bmiklautz:
@nfedera since the FREERDP_API was modified the version string should get increased.

hardening:
+1

PR #2738, opened 2015-06-26
hardening: check fread and fwrite return values

bmiklautz:


akallabeth:
@bmiklautz Found some more unchecked stuff in ```winpr/libwinpr/utils/image.c``` maybe we should fix them too.

bmiklautz:
@akallabeth I'm not sure why but exactly those in image.c are slipped through for some reason ;). Glad you saw them. Thank you.
I've updated the pull request.

hardening:
+1

akallabeth:
+1 now

libfreerdp/crypto/test/TestKnownHosts.c
```diff
@@ -46,8 +46,9 @@ static int prepare(const char* currentFileV2, const char* legacyFileV2, const ch
 
 	for (i=0; i<sizeof(hosts)/sizeof(hosts[0]); i++)
 	{
-		fwrite(hosts[i], strlen(hosts[i]), sizeof(char), fl);
-		fwrite(hosts[i], strlen(hosts[i]), sizeof(char), fc);
+		if (fwrite(hosts[i], strlen(hosts[i]), sizeof(char), fl) != sizeof(char) ||
+			fwrite(hosts[i], strlen(hosts[i]), sizeof(char), fc) != sizeof(char))
+			goto finish;
```
nfedera: Not your code but the fwrite arguments (size and nmemb) seem to be mixed. Although it works like this the correct way would be:
```
size_t nmemb = strlen(hosts[i]);
size_t size = sizeof(char);
if (fwrite(hosts[i], size, nmemb, fl) != nmemb) ...
```
... because we want to write nmemb characters where each character has sizeof(char) bytes.

libfreerdp/crypto/test/TestKnownHosts.c
```diff
@@ -61,14 +62,19 @@ static int prepare(const char* currentFileV2, const char* legacyFileV2, const ch
...
 	fclose(fl);
 	return 0;
 
 finish:
-	fclose(fl);
-	fclose(fc);
+	if (fl)
+		fclose(fl);
+	if (fc)
+		fclose(fc);
```
nfedera: nice catches! right - contrary to free fclose(NULL) is really undefined.

libfreerdp/crypto/test/TestKnownHosts.c
bmiklautz: @nfedera yes you could say they are mixed. However it works correctly. We write 1 element (sizeof(char)) with the size of strlen(hosts[i]) and check if one element was successfully written.
Anyway the sizeof(char) is really confusing and misplaced - I'll update it tomorrow.

server/shadow/Win/win_wds.c
```diff
@@ -764,21 +764,24 @@ int win_shadow_wds_init(winShadowSubsystem* subsystem)
...
+	size_t size;
 
-		fp = fopen("inv.xml", "w+b");
+	fp = fopen("inv.xml", "w+b");
 
-		if (fp)
+	if (fp)
+	{
+		size = strlen(file->ConnectionString2);
+		if (fwrite(file->ConnectionString2, 1, size, fp) != size || fwrite("\r\n", 1, 2, fp) != 2)
```
nfedera: no problem but the variable passed as fwrite's nmemb parameter is called size J

bmiklautz: @nfedera size J?
I haven't really checked that code since it was if(0) but in order to not have it compiled in I disabled it properly with #if 0.

winpr/libwinpr/utils/image.c
```diff
@@ -316,10 +330,7 @@ int winpr_image_read(wImage* image, const char* filename)
 		image->type = WINPR_IMAGE_PNG;
 		status = winpr_image_png_read_fp(image, fp);
 	}
-	else
-	{
-		fclose(fp);
-	}
+	fclose(fp);
```
nfedera: yes, way better this way. the old impl. was kinda weird

nfedera:
very nice PR. +1

libfreerdp/crypto/test/TestKnownHosts.c
nfedera: @bmiklautz sure, and in this case a simple if (fwrite(str, strlen(str), 1, fp) != 1) is better. And if character size > 1 I'd also want to write one element in this case using fwrite(wstr, wcslen(str)*sizeof(wchar_t), 1, fp != 1)

winpr/libwinpr/utils/image.c
bmiklautz: Just a little ;)

PR #2739, opened 2015-06-27
server/shadow: Enhancement regarding the screen and resolution - addresize support and fix subRect feature

realjiangms:
Detail fixes:
1. Add resize support in shadow framework layer
2. Enhance X11 implementation to detect desktop resolution change
3. Fix the subRect feature.
It seems not completely finished and it looks incompatible in different source code.
Fix it to be consistent in all source code as following:
a. The subRect is only awared in framework layer, subsystem implementation should not be aware of it. It only take effect at shadow_client and corresponding shadow_input.
b. The screen and surface should only represent a monitor screen. They don't need to consider subRect feature.
4. A lobby should be alternative image shown to client when the client is not 'mayView'.
We don't need to have seperate lobby for each client, move it into server structure as 'another surface'

### DIFF
--- a/include/freerdp/server/shadow.h
+++ b/include/freerdp/server/shadow.h
@@ -90,7 +90,6 @@ struct rdp_shadow_client
 	CRITICAL_SECTION lock;
 	REGION16 invalidRegion;
 	rdpShadowServer* server;
-	rdpShadowSurface* lobby;
 	rdpShadowEncoder* encoder;
 	rdpShadowSubsystem* subsystem;
 
@@ -112,6 +111,7 @@ struct rdp_shadow_server
 	wArrayList* clients;
 	rdpShadowScreen* screen;
 	rdpShadowSurface* surface;
+	rdpShadowSurface* lobby;
 	rdpShadowCapture* capture;
 	rdpShadowSubsystem* subsystem;
 
@@ -281,6 +281,8 @@ FREERDP_API int shadow_client_boardcast_quit(rdpShadowServer* server, int nExitC
 
 FREERDP_API int shadow_encoder_preferred_fps(rdpShadowEncoder* encoder);
 FREERDP_API UINT32 shadow_encoder_inflight_frames(rdpShadowEncoder* encoder);
+
+FREERDP_API BOOL shadow_screen_resize(rdpShadowScreen* screen);
 
 #ifdef __cplusplus
 }
--- a/server/shadow/X11/x11_shadow.c
+++ b/server/shadow/X11/x11_shadow.c
@@ -51,6 +51,8 @@
 
 #define TAG SERVER_TAG("shadow.x11")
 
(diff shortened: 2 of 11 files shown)


hardening:
Looks like you did a lot of work.
A remark: we could use Xrandr to know when the screen size has changed instead of constantly pooling to do this...

realjiangms:
But libxrandr-dev does not always exists. So I use XGetWindowAttributes to checking root window size to detect size change. It's a easiest way to detect whether desktop size changed with the basic Xlib function.

hardening:
@realjiangms I don't know what kind of system you have in mind, because all recent X servers have XRender enabled. Even the dev headers are widely deployed.

bmiklautz:
@hardening @realjiangms if I'm not mistaken you can do runtime detection of xrandr. Anyway checking the root window shouldn't have much overhead either ;).

bmiklautz:
@realjiangms one question on your PR - did you test what happens if you have multiple size changes in series? - for example:
```
xrandr -s 800x600; sleep 0.2; xrandr -s 1920x1080;
```

realjiangms:
@hardening
Yeah the XRANDR is enabled everywhere but libxrandr-dev not. (P.S. I'm using OPENSUSE 11 and Ubuntu 12.04)
And in server/shadow/CMakeLists.txt, the XRANDR is 'recommended' not 'required'

    set(XRANDR_FEATURE_TYPE "RECOMMENDED")
    set(XRANDR_FEATURE_PURPOSE "X11 resize, rotate and reflect")
    set(XRANDR_FEATURE_DESCRIPTION "X11 resize, rotate and reflect extension")

So it is not always there.
But I agree that we can have different implementation in ifdef macro for WITH_XRANDR, I may enhance it later.

@bmiklautz 
It works fine in SUSE 11. However it doesn't always work well with xrandr command in Ubuntu. It always report:

X Error of failed request:  BadMatch (invalid parameter attributes)
  Major opcode of failed request:  73 (X_GetImage)

It seems xrandr can take effect at anytime even if I put resize check after XLockDisplay. So when I call XGetImage, the size of root window could have been changed after the previous resize check. So get the BadMatch error.

However, it work very fine with 'gnome-control-center display'.
And interesting, when I close the gnome display control center, xrandr also works better.

But it still screwed up when I invoke following tough test:
while true; do xrandr -s 1024x768; sleep 0.2; xrandr -s 1280x1024; sleep 0.2; done

The rdp client kept resize and eventually disconnected by 'BadMatch' error after a while ;)

And during this test, I found another issue: The x11_shadow_enum_monitors open the display but never close it. I'll update PR with fix of it.

But all in all, it works fine with GUI controller, at least

realjiangms:
PR updated

realjiangms:
@bmiklautz 

I added an X11 error handler to bypass the BadMatch error during catpure.
Now it can survive from touch test as following:
while true; do xrandr -s 1024x768; xrandr -s 1280x1024; xrandr -s 800x600; done

PR #2708, opened 2015-06-18
alloc and strdup checks

bmiklautz:
* handle unchecked occurrences of strdup
* add checks for *alloc
* fix some compiler warnings on android
* fix and update android build
 * SDK 14 is required now
 * update function return values and the jni/JAVA classes accordingly

@hardening did a second round on malloc checks in winpr and
added, along a lot of other things, *alloc checks where missing.

PR #2740, opened 2015-06-28
server/shadow: Fix incorrect handle of makecert_context_process.

realjiangms:
This is introduced in PR #2708 with fault handling for makecert functions.
Other functions return '1' for OK, but makecert_context_process returns '0' for OK, so the current code is incorrect.

### DIFF
--- a/server/shadow/shadow_server.c
+++ b/server/shadow/shadow_server.c
@@ -554,10 +554,10 @@ static BOOL shadow_server_init_certificate(rdpShadowServer* server)
 		if (!makecert)
 			goto out_fail;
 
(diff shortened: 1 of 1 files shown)


realjiangms:
I wonder if we should fix makecert_context_process to keep consistent with other makecert functions. Weird meaning of return values introduces lots of trouble for both coding and reviewing :)

hardening:
LGTM

bmiklautz:
+1
@realjiangms thank you! - Since we do have so much different return type handlings in FreeRDP it's sometimes really hard to keep track ;)

PR #2741, opened 2015-06-28
egfx progressive codec frame handling

awakecoding:
Fix a really tricky bug in egfx progressive related to handling of rects and tiles within the same "frame" that was improperly defined in MS-RDPEGFX. This fixes some graphical artifacts that could be observed in cases with short successions of updates, like full window dragging, menu animations and rich graphics. This issue was discussed here: https://social.msdn.microsoft.com/Forums/en-US/113ac1cf-704d-47e7-93bc-e5d06ec88afd/msrdpegfx-rdpgfxwiretosurfacepdu2-and-codec-context-adm


libfreerdp/codec/progressive.c
```diff
@@ -1755,11 +1780,11 @@ int progressive_decompress(PROGRESSIVE_CONTEXT* progressive, BYTE* pSrcData, UIN
...
 					return -1021;
 
-				if (region->numTiles > progressive->cTiles)
+				if (progressive->cTiles < surface->gridSize)
 				{
 					progressive->tiles = (RFX_PROGRESSIVE_TILE**) realloc(progressive->tiles,
-							region->numTiles * sizeof(RFX_PROGRESSIVE_TILE*));
-					progressive->cTiles = region->numTiles;
+							surface->gridSize * sizeof(RFX_PROGRESSIVE_TILE*));
+					progressive->cTiles = surface->gridSize;
```
nfedera: I suggest to simply allocate progressive->tiles in progressive_surface_context_new() instead and kick progressive->cTiles

Issue #2391, opened 2015-02-12
Mouse click holding bug


Testinos:
Have problem with mouse click, some random time if click to (url, text, window) mouse dont clickup and  holding down the mouse. 
Usage option 

```
xfreerdp /f /multimon +fonts /cert-ignore /bpp:32 /d:AT /sound:sys:pulse /microphone:sys:pulse /kbd:0x00000409 -wallpaper /sec:tls /u:USER /v:win2012-server +auto-reconnect /gfx +gfx-progressive /rfx +async-input +async-update +gfx-h264
```

any idea how fix it?

bmiklautz:
@Testinos could you please add what version of FreeRDP you are using and what system you are connecting to?

Testinos:
@bmiklautz i am using lates version from git, and openh264 lates version from git, Connect to Win server 2012R2. 
I am not sure but i think this bug have only with +gfx-h264 parameter.
I can use xfreerdp without this option but this option significantly increases the speed RDP.

Testinos:
I am very sorry that I do not know с++ code, but i have more question about xfreerdp. 
have next bug with +gfx-h264 parameter
after watch youtube >1 min have bug litle black box image artifacts.

Where can I go to address these issues?

hardening:
@Testinos can you try latest master and see if #2747 fixed the problem for you ?
Thank you

Testinos:
@hardening 
bug still have (after 2 hour work have 2 time this bug)
if open youtube have error in console
```
[13:18:36:092] [7419:7427] [ERROR][com.freerdp.client.x11] - h264_decompress failure: -1
[13:18:36:117] [7419:7427] [ERROR][com.freerdp.client.x11] - h264_decompress failure: -1
[13:18:36:137] [7419:7427] [ERROR][com.freerdp.client.x11] - h264_decompress failure: -1
[13:18:36:164] [7419:7427] [ERROR][com.freerdp.client.x11] - h264_decompress failure: -1
[13:18:36:223] [7419:7427] [ERROR][com.freerdp.client.x11] - h264_decompress failure: -1
```
and this bug appears increasingly

akallabeth:
Closing as GFX AVC modes have improved. If still an issue with current master please reopen.

PR #2747, opened 2015-06-30
Do not discard events when updating EGFX surface mapping

ilammy:
Passing True to XSync() discards any pending X11 events. Occasionally this caused ButtonRelease or KeyRelease to be lost and not forwarded to the remote computed, leading to stuck keys and buttons.

This should resolve issue #2391


bmiklautz:
@freerdp-bot test.

bmiklautz:
@ilammy nice finding!
+1

akallabeth:
+1, nice catch

hardening:
+1

PR #2751, opened 2015-06-30
server/shadow: Fix compile error on win32.

realjiangms:
This is introduced by previous standalone fixes.

### DIFF
--- a/include/freerdp/server/shadow.h
+++ b/include/freerdp/server/shadow.h
@@ -204,7 +204,7 @@ typedef struct _SHADOW_MSG_OUT SHADOW_MSG_OUT;
 typedef void (*MSG_OUT_FREE_FN)(UINT32 id, SHADOW_MSG_OUT* msg);
 #define RDP_SHADOW_MSG_OUT_COMMON() \
 	int refCount; \
-	MSG_OUT_FREE_FN Free; /* function to free SHADOW_MSG_OUT */
+	MSG_OUT_FREE_FN Free /* function to free SHADOW_MSG_OUT */
 
 struct _SHADOW_MSG_OUT
 {
--- a/include/freerdp/server/shadow.h
+++ b/include/freerdp/server/shadow.h
@@ -201,10 +201,10 @@ struct _SHADOW_MSG_IN_SUPPRESS_OUTPUT
 typedef struct _SHADOW_MSG_IN_SUPPRESS_OUTPUT SHADOW_MSG_IN_SUPPRESS_OUTPUT;
 
 typedef struct _SHADOW_MSG_OUT SHADOW_MSG_OUT;
-typedef void (*MSG_OUT_FREE_FN)(UINT32 id, SHADOW_MSG_OUT* msg);
+typedef void (*MSG_OUT_FREE_FN)(UINT32 id, SHADOW_MSG_OUT* msg); /* function to free SHADOW_MSG_OUT */
 #define RDP_SHADOW_MSG_OUT_COMMON() \
 	int refCount; \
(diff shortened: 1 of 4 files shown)


hardening:
LGTM

include/freerdp/server/shadow.h
```diff
@@ -204,7 +204,7 @@ typedef struct _SHADOW_MSG_OUT SHADOW_MSG_OUT;
 typedef void (*MSG_OUT_FREE_FN)(UINT32 id, SHADOW_MSG_OUT* msg);
 #define RDP_SHADOW_MSG_OUT_COMMON() \
 	int refCount; \
-	MSG_OUT_FREE_FN Free; /* function to free SHADOW_MSG_OUT */
+	MSG_OUT_FREE_FN Free /* function to free SHADOW_MSG_OUT */
```
akallabeth: Comments in defines are no good idea, may break certain compilers.

bmiklautz: @akallabeth true that but that was there before as well.

akallabeth: True, but with the missing ```;``` this may lead to problems

akallabeth: @bmiklautz @realjiangms Just tested, VS2013 does not like this kind of define. Would be great, if this could be fixed with this pull.

realjiangms: @akallabeth @bmiklautz 
I'll delete the comments :) And put the comment in MSG_OUT_FREE_FN instead. I think it's clear enough.

It doesn't break compiler on win and Linux, but I agree it looks risky.

bmiklautz:
@realjiangms what's the reason for marking the X11_ShadowSubsystemEntry as FREERDP_API?

realjiangms:
@bmiklautz 
It is not necessary.
The problem is that the corrsponding windows entry point function need 'FREERDP_API' macro so that it could be exported and used by shadow.c
It fails to link on windows platform without this mark.

Actually the X11 function don't need it as Linux defaultly export the functions. But to be consistent, I mark 'FREERDP_API' for all of the 3 functions (Mac/X11/Win)  
The 3 entry point function could be regarded as 'built-in' subsystem sample

Alternative solution is, we completely move the subsystem implementation out of libfreerdp-shadow and make them compiled together with shadow.c.
Therefore there's only shadow framework in libfreerdp-shadow. I may prefer to this approach. But this need changes in CMakeList.txt, and I want to test it on all three platforms.
I may make the effort to do it in later PR as I'm still seting up my MAC development.
 
Any preference?

akallabeth:
LGTM, +1

mfleisz:
+1

mfleisz:
@bmiklautz any objections merging this PR or is the exported X11_ShadowSubsystemEntry a blocker for you? (compiling FreeRDP with WITH_SERVER currently doesn't work with VS without this PR merged)

bmiklautz:
@mfleisz nope I'm fine. But we should definitely handle the exported symbols at some point.

mfleisz:
@bmiklautz agree with that.
@realjiangms maybe you can solve that issue as you proposed in a separate PR?

realjiangms:
@mfleisz
@bmiklautz 
Thanks. 
For the exported symbols: I will post another PR to completely decouple the subsystem implementation and shadow framework (currently is not completed). In that PR I may move the subsystem implementations out of libfreerdp-shadow.  But therefore the user for libfreerdp-shadow can no longer call default subsystem implementation directly, I may need to notify @awakecoding for his comment

PR #2752, opened 2015-07-01
NLA decrypt credentials fixed.

akallabeth:
Fixes a bug in server side credential decryption.
The username was skipped during decryption caused by the password length being checked instead of username length.


hardening:
I think it's my fault, bring the brown paper bag :)

PR #2757, opened 2015-07-02
winpr: fix CreateProcess signal handler issues

nfedera:
The calling thread of CreateProcess can be in any library and can have arbitrary signal masks and handlers.

We now save the caller's mask and  block all signals before forking.
After fork:
- child resets the handlers and unblocks all signals.
- parent restores the caller's original signal mask.

### DIFF
--- a/winpr/libwinpr/thread/process.c
+++ b/winpr/libwinpr/thread/process.c
@@ -80,6 +80,14 @@
 #include "../handle/handle.h"
 #include "../security/security.h"
 
+#ifndef NSIG
+#ifdef SIGMAX
+#define NSIG SIGMAX
+#else
+#define NSIG 64
+#endif
+#endif
+
 char** EnvironmentBlockToEnvpA(LPCH lpszEnvironmentBlock)
 {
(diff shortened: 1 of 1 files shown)


winpr/libwinpr/thread/process.c
```diff
@@ -254,16 +269,33 @@ BOOL _CreateProcessExA(HANDLE hToken, DWORD dwLogonFlags,
...
+		sigset_t set;
+		struct sigaction act;
+
+		/* set default signal handlers */
+		memset(&act, 0, sizeof(act));
+		act.sa_sigaction = SIG_DFL;
+		act.sa_flags = 0;
+		sigemptyset(&act.sa_mask);
+		for (sig = 1; sig < NSIG; sig++)
+			sigaction(sig, &act, NULL);
```
bmiklautz: imho this isn't necessary since it get's reset when exec is called.

bmiklautz:
@nfedera nice. This is really important. I wounder it was missed that long.

PR #2759, opened 2015-07-02
x11/cliprdr: handle empty format names

bmiklautz:
The recently added strdup checks ignored the fact that format names
can be NULL.


bmiklautz:
@dupondje verified the fix.
Thx.

PR #2761, opened 2015-07-03
Updated sample server.

akallabeth:
* Added new command line argument --port to allow starting the
  sample server on a port different than default 3389.
* Using GetKnownSubPath now to determine location of temporary
  directory for server socket.
* Using distinct server socket files for each port.

### DIFF
--- a/server/Sample/sfreerdp.c
+++ b/server/Sample/sfreerdp.c
@@ -28,6 +28,7 @@
 
 #include <winpr/crt.h>
 #include <winpr/synch.h>
+#include <winpr/path.h>
 #include <winpr/winsock.h>
 
 #include <freerdp/channels/wtsvc.h>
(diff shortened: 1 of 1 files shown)


server/Sample/sfreerdp.c
```diff
@@ -873,30 +874,61 @@ int main(int argc, char* argv[])
...
 	freerdp_listener* instance;
+	char* file;
+	char name[MAX_PATH];
+	int port = 3389, i;
+
+	for (i=1; i<argc; i++)
+	{
+		char* arg = argv[i];
+
+		if (strncmp(arg, "--fast", 7) == 0)
```
mfleisz: should be 6 instead of 7

mfleisz:
+1

PR #2762, opened 2015-07-03
Add write event support to HANDLE

akallabeth:
* Added write event support to ```WaitForSingleObject``` and ```WaitForMultipleObjects```
* Modified ```CreateFileDescriptorEvent``` to enable write event support
* Fixed ```WSAEventSelect``` with ```lNetworkEvents == 0``` as argument
* Modified ```SetEventFileDescriptor``` to enable write event support, fixed missing ```bAttached``` to allow file descriptor cleanup.
* Added ```winpr_log_backtrace```

### DIFF
--- a/winpr/libwinpr/synch/event.c
+++ b/winpr/libwinpr/synch/event.c
@@ -376,7 +376,7 @@ int SetEventFileDescriptor(HANDLE hEvent, int FileDescriptor, ULONG mode)
 		return -1;
 
 	event = (WINPR_EVENT*) Object;
-	event->bAttached = TRUE;
+	event->bAttached = FALSE;
 	event->Mode = mode;
 	event->pipe_fd[0] = FileDescriptor;
 	return 0;
--- a/winpr/libwinpr/synch/event.c
+++ b/winpr/libwinpr/synch/event.c
@@ -78,23 +78,22 @@ BOOL EventCloseHandle(HANDLE handle) {
 	if (!EventIsHandled(handle))
 		return FALSE;
 
-    if (!event->bAttached)
-    {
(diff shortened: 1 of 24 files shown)


winpr/libwinpr/synch/event.c
```diff
@@ -360,17 +365,19 @@ int GetEventFileDescriptor(HANDLE hEvent)
...
 	ULONG Type;
-	PVOID Object;
+	WINPR_HANDLE* Object;
 	WINPR_EVENT* event;
 
 	if (!winpr_Handle_GetInfo(hEvent, &Type, &Object))
 		return -1;
 
 	event = (WINPR_EVENT*) Object;
+	event->bAttached = TRUE;
```
MartinHaimberger: setting attached here is right ?

winpr/libwinpr/synch/wait.c
```diff
@@ -193,8 +217,10 @@ DWORD WaitForSingleObject(HANDLE hHandle, DWORD dwMilliseconds)
 	{
 		WINPR_PROCESS *process;
 		process = (WINPR_PROCESS *) Object;
+		int rc;
```
MartinHaimberger: is the int rc here really required ?
Would not be (waitpid(process->pid, &(process->status), 0) != process->pid enought ?

MartinHaimberger:
besides my remarks LGTM

winpr/libwinpr/synch/event.c
akallabeth: Just checked, yes that is a bug.

mfleisz:
+1

PR #2764, opened 2015-07-03
Pull fixes

akallabeth:
* Fixes a missing include in sample server, preventing windows build to find ```snprintf```
* Fixed inverted ```bAttached``` setting.

PR #2765, opened 2015-07-03
Connect abort and unified TCP connect

akallabeth:
* Addes a new event ```abortEvent``` set by the new function ```freerdp_abort_connect``` to allow aborting a ```freerdp_connect``` from another thread.
* Extended ```freerdp_tcp_connect``` arguments to allow waiting for ```abortEvent```
* 
* Unified ```freerdp_tcp_connect_multi``` and ```freerdp_tcp_connect_timeout``` using newly extended WinPR ```WSAEventSelect``` and ```FD_WRITE``` support for ```WaitForMultipleObjects```

### DIFF
--- a/libfreerdp/core/freerdp.c
+++ b/libfreerdp/core/freerdp.c
@@ -181,6 +181,14 @@ freerdp_connect_finally:
 	return status;
 }
 
+BOOL freerdp_abort_connect(freerdp* instance)
+{
+	if (!instance || !instance->context)
+		return FALSE;
+
+	return SetEvent(instance->context->abortEvent);
+}
+
 BOOL freerdp_get_fds(freerdp* instance, void** rfds, int* rcount, void** wfds, int* wcount)
 {
 	rdpRdp* rdp = instance->context->rdp;
@@ -526,6 +534,10 @@ BOOL freerdp_context_new(freerdp* instance)
 
 	update_register_client_callbacks(rdp->update);
 
+	instance->context->abortEvent = CreateEvent(NULL, TRUE, FALSE, NULL);
+	if (!instance->context->abortEvent)
+		goto out_error_abort_event;
+
 	IFCALLRET(instance->ContextNew, ret, instance, instance->context);
 
 	if (ret)
@@ -575,6 +587,9 @@ void freerdp_context_free(freerdp* instance)
 
 	CloseHandle(instance->context->channelErrorEvent);
 	free(instance->context->errorDescription);
+
+	CloseHandle(instance->context->abortEvent);
+	instance->context->abortEvent = NULL;
 
 	free(instance->context);
 	instance->context = NULL;
--- a/libfreerdp/core/freerdp.c
+++ b/libfreerdp/core/freerdp.c
@@ -386,6 +386,8 @@ BOOL freerdp_reconnect(freerdp* instance)
 
 BOOL freerdp_shall_disconnect(freerdp* instance)
 {
+	if (!instance || !instance->context || !instance->context->rdp)
+		return FALSE;
 	return instance->context->rdp->disconnect;
 }
 
--- a/libfreerdp/core/freerdp.c
+++ b/libfreerdp/core/freerdp.c
@@ -102,6 +102,9 @@ BOOL freerdp_connect(freerdp* instance)
 		goto freerdp_connect_finally;
 	}
 
+	if (!status)
+		goto freerdp_connect_finally;
+
 	if (status)
 	{
 		if (instance->settings->DumpRemoteFx)
--- a/libfreerdp/core/freerdp.c
+++ b/libfreerdp/core/freerdp.c
@@ -74,11 +74,6 @@ BOOL freerdp_connect(freerdp* instance)
 	rdp = instance->context->rdp;
 	settings = instance->settings;
 
-	if (rdp)
-		rdp->disconnect = FALSE;
-
-	ResetEvent(instance->context->abortEvent);
-
 	instance->context->codecs = codecs_new(instance->context);
 	IFCALLRET(instance->PreConnect, status, instance);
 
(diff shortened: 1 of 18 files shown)


akallabeth:
Rebased after #2764 were merged, otherwise connect unit test fails.

libfreerdp/core/tcp.c
```diff
@@ -1097,7 +901,9 @@ int freerdp_tcp_connect_multi(char** hostnames, UINT32* ports, int count, int po
 		}
 	}
 
-	waitStatus = WaitForMultipleObjects(count, events, FALSE, timeout * 1000);
+	events[count] = context->abortEvent;
+
+	waitStatus = WaitForMultipleObjects(count + 1, events, FALSE, timeout * 1000);
```
realjiangms: It seems to me that the event[] is never initialized after calloc. That means all events in this array is just NULL and the previous 'WSAEventSelect' doesn't take any effect(it just returns SOCKET_ERROR).

This is not introduced by this PR. It's the original logic in freerdp_tcp_connect_multi for win32.

akallabeth: @realjiangms Just expand the view, then you can see the ```WSAEventSelect``` ;)

MartinHaimberger:
@freerdp-bot would you be so kind and test this nice PR ?

MartinHaimberger:
+1 LGTM

libfreerdp/core/tcp.c
realjiangms: I can see the WSAEventSelect,  it's in the diff ;). What I mean is the array 'events'.
Here's all lines of code involve it: 

        821 HANDLE* events;
	833 events = (HANDLE*) calloc(count + 1, sizeof(HANDLE));
        899 if (WSAEventSelect(sockfd, events[index], FD_READ | FD_WRITE | FD_CONNECT | FD_CLOSE))
      * 918 waitStatus = WaitForMultipleObjects(count + 1, events, FALSE, timeout * 1000);

We can see that the events in 'events' are all NULL.

akallabeth: Ah, ok. Missed that one, now fixed (including cleanup later on)

akallabeth:
@freerdp-bot could you test again please?

libfreerdp/core/freerdp.c
```diff
@@ -526,10 +545,17 @@ BOOL freerdp_context_new(freerdp* instance)
...
+	instance->context->abortEvent = CreateEvent(NULL, TRUE, FALSE, NULL);
+	if (!instance->context->abortEvent)
+		goto out_error_abort_event;
+
 	IFCALLRET(instance->ContextNew, ret, instance, instance->context);
 
 	if (ret)
 		return TRUE;
+
+out_error_abort_event:
```
hardening: If I'm not wrong the action here is not correct (if we haven't manage to create the event, no need to close it)

akallabeth:
@freerdp-bot Could you please test again?

libfreerdp/core/listener.c
```diff
@@ -347,8 +352,10 @@ static BOOL freerdp_listener_check_fds(freerdp_listener* instance)
 
 	for (i = 0; i < listener->num_sockfds; i++)
 	{
+		WSAResetEvent(listener->events[i]);
```
realjiangms: You probably need to keep my '#ifdef _WIN32' here. Reset event on posix platform (Linux, Mac) is not correct. ResetEvent on Linux would break the read buffer because the socket event is the fd itself. We only need it on Windows

 I agree that it's every tricky for us to understand the difference on socket event between windows and posix platform due to the different implementation.

libfreerdp/core/tcp.c
```diff
@@ -145,6 +145,8 @@ static int transport_bio_simple_read(BIO* bio, char* buf, int size)
 
 	BIO_clear_flags(bio, BIO_FLAGS_READ);
 
+	WSAResetEvent(ptr->hEvent);
```
realjiangms: Same as listener.c
We probably need to keep my '#ifdef _WIN32' here to only reset on windows platform

akallabeth:
@realjiangms Have a look at ```WSAResetEvent``` implementation, I've removed the event reset code there because this function is only used for socket event handles that are auto reset on POSIX systems.

libfreerdp/core/listener.c
realjiangms: Oops. You've already re-defined WSAResetEvent. It should be correct now.
Although it still looks a little weird and someone read this line may think it as a bug: why we call a function which does nothing. :)

libfreerdp/core/tcp.c
realjiangms: Sorry for late reply, but it looks great now :)

libfreerdp/core/transport.c
```diff
@@ -750,20 +751,7 @@ int transport_check_fds(rdpTransport* transport)
 	if (!transport)
```
realjiangms: I have to comment here as github don't allow me to add comment on hidden line.
We may need to also remove 'HANDLE event' above as it is never used now.

realjiangms:
The fix ported from #2870 works fine.
I've tested on MAC, Windows7 and Ubuntu12.04.

realjiangms:
Paste my test result from #2870:
Tested on MAC, Windows7 and Ubuntu12.04. 
No 100% cpu issue on all platform.
I tested both client(xfreerdp/MacFreeRDP/wfreerdp) and server(freerdp-shadow).

The CPU usage is very low (client < %1, server < %3).

Client could connect to server successfully (I tested win7 native rdp, shadow server, and virtual box). Mac client sometimes has black screen. This is existing issue and is not related to #2765.

Server could be connected correctly (I tested with mstsc and the xfreerdp). Mac server and Windows server has some image update issue, but it is existing issue and is not related to #2765.

bmiklautz:
@realjiangms thank you for the feedback \o/

PR #2770, opened 2015-07-07
Resetting frontBio after read now.

akallabeth:
Resetting the transport layer event after reading all data.
When the event is not manually reset, the client main loop transforms to a busy loop consuming lots of CPU
This bug was uncovered by #2750 which fixed events getting lost because of automatically reset events and the ```WaitForMultipleObjects``` followed by specific ```WaitForSingleObject``` calls on the bio event handle.

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -674,7 +674,8 @@ DWORD transport_get_event_handles(rdpTransport* transport, HANDLE* events, DWORD
 	{
 		if (events && (nCount < count))
 		{
-			BIO_get_event(transport->frontBio, &events[nCount]);
+			if (BIO_get_event(transport->frontBio, &events[nCount]) != 1)
+				return 0;
 			nCount++;
 		}
 	}
@@ -749,7 +750,8 @@ int transport_check_fds(rdpTransport* transport)
 	if (!transport)
 		return -1;
 
-	BIO_get_event(transport->frontBio, &event);
+	if (BIO_get_event(transport->frontBio, &event) != 1)
+		return -1;
 
 	/**
 	 * Loop through and read all available PDUs.  Since multiple
@@ -758,6 +760,7 @@ int transport_check_fds(rdpTransport* transport)
 	 * wait for a socket to get signaled that data is available
 	 * (which may never happen).
 	 */
+	ResetEvent(event);
 	for (;;)
 	{
 		/**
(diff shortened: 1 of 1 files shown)


realjiangms:
I guess this issue already exists in Linux version because the Linux style of implementation is always 'manual reset', because in winpr/libwinpr/synch/event.c:

    if (!event->bManualReset)
    {
        WLog_ERR(TAG, "auto-reset events not yet implemented");
    }

By the way, I realize that PR #2750 miss the code for Linux:
 		ptr->hEvent = CreateFileDescriptorEvent(NULL, FALSE, FALSE, (int) ptr->socket);

Although it is equivalent to "CreateFileDescriptorEvent(NULL, TRUE, FALSE, (int) ptr->socket);" in current winpr implementation, it is wrong semantically.

akallabeth:
@realjiangms The linux path is fixed by #2765

libfreerdp/core/transport.c
```diff
@@ -744,10 +744,13 @@ int transport_check_fds(rdpTransport* transport)
 	int status;
 	int recv_status;
 	wStream* received;
+	HANDLE event;
 
 	if (!transport)
 		return -1;
 
+	BIO_get_event(transport->frontBio, &event);
```
hardening: return value not checked

hardening:
I have not looked at the transport thing from a long time, but does your patch work also work with async transport ?

akallabeth:
@hardening Should work (have not tested extensivly with async), as it just reads that stuff in its own thread.

llyzs:
+1

Tested and worked well.

HenryJacques:
Does not work for me, see #2790

bmiklautz:
@akallabeth @llyzs did you test on linux? This problem is really quite obvious.

realjiangms:
I think we should not have to reset event in libfreerdp/core/transport.c
What need to be fixed is still in libfreerdp/core/tcp.c:

WSAEventSelect(ptr->socket, ptr->hEvent, FD_READ | FD_WRITE | FD_CLOSE);
==========> 
((It attach to FD_WRITE event. This should not be expected. Because WSAEventSelect attaches to FD_WRITE, it is always in 'set' state even when we drain the read buffer.))

I think it should only attach to FD_READ, same as Linux part:
ptr->hEvent = CreateFileDescriptorEvent(NULL, FALSE, FALSE, (int) ptr->socket, WINPR_FD_READ);


When I looked at this PR firstly, I did realized a question that why we only have problem in windows version. But I didn't think much of it.

realjiangms:
@akallabeth I guess the same 'busy loop issue' also exists in Linux version with PR #2765

realjiangms:
I tried to remove the FD_WRITE in WSAEventSelect, but it doesn't work. The busy loop issue still exists.
The event created by WSACreateEvent need WSAEnumNetworkEvents or WSAResetEvent to reset the event manually. It seems currently the fix in 2770 is the best solution

PR #2772, opened 2015-07-07
tsmf updates/fixes for gstreamer decoder on X11 platform

bjcollins:
- Use decodebin2 instead of old decodebin - decodebin has issues
- Use autovideosink - xvimagesink does not work with cards with no xv ports available and cant be used if wanted to use the fluendo hardware accelerated playback codec
- Use autoaudiosink - let gstreamer choose best audio playback plugin
- Catch when autosinks add known elements so that we can manipulate properties on them
- Adjust caps of various media types to work better with gstreamer, some codecs are picky about having certain fields available
- Remove unneeded plugins such as "ffmpegcolorspace" and "videoscale" - these do not work correctly with fluendo hardware accelerated playback codec
- Name audio/video gstreamer elements better for easier debugging
- Update gstreamer pipeline and element properties to handle playback better
- Detect when valid timestamps are available for buffer from server and try to account for when they are not valid
- Start time is much more reliable then end time from server for various media formats, so use it when possible to make decisions instead of end time
- Do not rebuild gstreamer pipeline for a seek(very expensive), instead reset gstreamer time to 0 and maintain offset between real time and gstreamer time
- Change buffer filled function back to a buffer level function, so that we can use buffer level to make better choices above gstreamer decoder in tsmf
- Remove ack function from gstreamer, instead rely on ack thread to handle acks
- Rework X11 gstreamer code to handle various videosinks which implement the XOverlayInterface and to keep more detailed information on the sub-window that is used for display
- Add check to see if a decoder is available for telling the server the client various media types
- Add in support for M4S2 and WMA1 media types
- Fix flush message handling, they are for individual streams and not the entire presentation
- Delay eos response to try to allow more time for buffers to be loaded into decoder, as we anticipate acks to server and the server will issue stop as soon as we ack eos.
- Fix issue with geometry info being ignored when resent for new streams within existing presentation
- Fixed volume level initialization issue when a stream is stopped and restarted
- Attempt to sync video/audio streams...because we run two different gstreamer pipelines - they can enter pause/playing states at different times and are thus not synchronized. Attempt
to adjust video buffer timestamps based on difference between audio/video running time to account for this difference. This logic accounts for a huge improvement in audio/video sync(ie. lip sync to words)

### DIFF
--- a/channels/tsmf/client/gstreamer/tsmf_X11.c
+++ b/channels/tsmf/client/gstreamer/tsmf_X11.c
@@ -61,6 +61,12 @@ struct X11Handle
 #endif
 	Display *disp;
 	Window subwin;
+	BOOL subwinMapped;
+	GstXOverlay *overlay;
+	int subwinWidth;
+	int subwinHeight;
+	int subwinX;
+	int subwinY;
 };
 
 static const char* get_shm_id()
@@ -70,13 +76,73 @@ static const char* get_shm_id()
 	return shm_id;
 }
 
+static GstBusSyncReply tsmf_platform_bus_sync_handler(GstBus *bus, GstMessage *message, gpointer user_data)
+{
+	struct X11Handle* hdl;
+
+	TSMFGstreamerDecoder* decoder = user_data;
+
+	if (GST_MESSAGE_TYPE (message) != GST_MESSAGE_ELEMENT)
+		return GST_BUS_PASS;
+
+	if (!gst_structure_has_name (message->structure, "prepare-xwindow-id"))
+		return GST_BUS_PASS;
+
+	hdl = (struct X11Handle*) decoder->platform;
+
+	if (hdl->subwin)
+	{
+		hdl->overlay = GST_X_OVERLAY (GST_MESSAGE_SRC (message));
+
+#if GST_VERSION_MAJOR > 0
+		gst_video_overlay_set_window_handle(hdl->overlay, hdl->subwin);
+#else
+		gst_x_overlay_set_window_handle(hdl->overlay, hdl->subwin);
+#endif
+#if GST_VERSION_MAJOR > 0
+		gst_video_overlay_handle_events(hdl->overlay, TRUE);
+#else
+		gst_x_overlay_handle_events(hdl->overlay, TRUE);
+#endif
+
+		if (hdl->subwinWidth != -1 && hdl->subwinHeight != -1 && hdl->subwinX != -1 && hdl->subwinY != -1)
+		{
+#if GST_VERSION_MAJOR > 0
+
+			if (!gst_video_overlay_set_render_rectangle(hdl->overlay, 0, 0, hdl->swubwinWidth, hdl->subwinHeight))
+			{
+				WLog_ERR(TAG, "Could not resize overlay!");
+			}
+
+			gst_video_overlay_expose(hdl->overlay);
+#else
+			if (!gst_x_overlay_set_render_rectangle(hdl->overlay, 0, 0, hdl->subwinWidth, hdl->subwinHeight))
+			{
+				WLog_ERR(TAG, "Could not resize overlay!");
+			}
+
+			gst_x_overlay_expose(hdl->overlay);
+#endif
+			XMoveResizeWindow(hdl->disp, hdl->subwin, hdl->subwinX, hdl->subwinY, hdl->subwinWidth, hdl->subwinHeight);
+			XSync(hdl->disp, FALSE);
+		}
+	} else {
+		g_warning ("Window was not available before retrieving the overlay!");
+	}
+
+	gst_message_unref (message);
+
+	return GST_BUS_DROP;
+}
+
 const char* tsmf_platform_get_video_sink(void)
 {
-	return "xvimagesink";
+	return "autovideosink";
 }
 
 const char* tsmf_platform_get_audio_sink(void)
 {
+	//return "alsasink";
 	return "autoaudiosink";
 }
 
@@ -119,6 +185,12 @@ int tsmf_platform_create(TSMFGstreamerDecoder* decoder)
 		return -4;
 	}
 
+	hdl->subwinMapped = FALSE;
+	hdl->subwinX = -1;
+	hdl->subwinY = -1;
+	hdl->subwinWidth = -1;
+	hdl->subwinHeight = -1;
+
 	return 0;
 }
 
@@ -147,11 +219,15 @@ int tsmf_platform_register_handler(TSMFGstreamerDecoder* decoder)
 
 	bus = gst_pipeline_get_bus(GST_PIPELINE(decoder->pipe));
 
+	gst_bus_set_sync_handler (bus, (GstBusSyncHandler) tsmf_platform_bus_sync_handler, decoder);
+
 	if (!bus)
 	{
 		WLog_ERR(TAG, "gst_pipeline_get_bus failed!");
 		return 1;
 	}
+
+	gst_object_unref (bus);
 
 	return 0;
 }
@@ -189,12 +265,6 @@ int tsmf_window_create(TSMFGstreamerDecoder* decoder)
 	}
 	else
 	{
-#if GST_VERSION_MAJOR > 0
-		GstVideoOverlay *overlay = GST_VIDEO_OVERLAY(decoder->outsink);
-#else
-		GstXOverlay *overlay = GST_X_OVERLAY(decoder->outsink);
-#endif
-
 		if (!decoder)
 			return -1;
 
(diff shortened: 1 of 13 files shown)


channels/tsmf/client/gstreamer/tsmf_X11.c
```diff
@@ -69,13 +75,73 @@ static const char* get_shm_id()
...
 const char* tsmf_platform_get_video_sink(void)
 {
-	return "xvimagesink";
+	return "autovideosink";
 }
 
 const char* tsmf_platform_get_audio_sink(void)
 {
+	//return "alsasink";
 	return "autoaudiosink";
```
hardening: comment to drop

channels/tsmf/client/gstreamer/tsmf_gstreamer.c
```diff
@@ -43,10 +43,14 @@
 #include <inttypes.h>
 #endif
 
+// 1 second
+#define SEEK_TOLERANCE 10000000
```
hardening: no C++ comment
1 * 1000 * 1000 would be more readable ?

channels/tsmf/client/gstreamer/tsmf_gstreamer.c
```diff
@@ -55,8 +59,33 @@ const char* get_type(TSMFGstreamerDecoder* mdecoder)
 
 	if (mdecoder->media_type == TSMF_MAJOR_TYPE_VIDEO)
 		return "VIDEO";
-	else
+	else if (mdecoder->media_type == TSMF_MAJOR_TYPE_AUDIO)
 		return "AUDIO";
+	else
```
hardening: else can be dropped

channels/tsmf/client/gstreamer/tsmf_gstreamer.c
```diff
@@ -750,7 +942,7 @@ BOOL tsmf_gstreamer_ack(ITSMFDecoder* decoder, BOOL (*cb)(void *, BOOL), void *s
 {
 	TSMFGstreamerDecoder* mdecoder = (TSMFGstreamerDecoder *) decoder;
 	DEBUG_TSMF("");
-	mdecoder->ack_cb = cb;
+	mdecoder->ack_cb = NULL;//cb;
 	mdecoder->stream = stream;
```
hardening: drop comment

channels/tsmf/client/tsmf_main.c
```diff
@@ -34,6 +34,38 @@
...
+
+	if (!callback)
+	{
+		DEBUG_TSMF("No callback reference - unable to send eos response!");
+		return;
+	}
+
+	if (callback->stream_id)
+	{
+		s = Stream_New(NULL, 24);
```
hardening: return value not checked

channels/tsmf/client/tsmf_media.c
```diff
@@ -238,6 +274,8 @@ static void tsmf_sample_queue_ack(TSMF_SAMPLE* sample)
 	Queue_Enqueue(sample->stream->sample_ack_list, sample);
 }
 
+// Returns TRUE if no more samples are currently available
+// Returns FALSE otherwise
 static BOOL tsmf_stream_process_ack(void* arg, BOOL force)
```
hardening: no C++ comment

channels/tsmf/client/tsmf_media.c
```diff
@@ -246,22 +284,38 @@ static BOOL tsmf_stream_process_ack(void* arg, BOOL force)
 	BOOL rc = FALSE;
 
 	if (!stream)
-		return FALSE;
+		return TRUE;
```
hardening: are you sure about this one ?

channels/tsmf/client/tsmf_media.c
```diff
@@ -882,7 +1026,7 @@ void tsmf_presentation_set_audio_device(TSMF_PRESENTATION* presentation, const c
 	presentation->audio_device = device;
 }
 
-static void tsmf_stream_flush(TSMF_STREAM* stream)
+void tsmf_stream_flush(TSMF_STREAM* stream)
```
hardening: is the static removed on purpose ?

hardening:
@freerdp-bot would  you be so kind and test this nice PR ?

hardening:
@bjcollins nice PR.

Guys with the merge rights, we have a branch that contains tons of fixes for the TSMF channels, some are possibly conflicting with this PR. I'd suggest that we wait before merging this one.

channels/tsmf/client/gstreamer/tsmf_X11.c
bjcollins: agreed

channels/tsmf/client/gstreamer/tsmf_gstreamer.c
bjcollins: I think we should keep the else, UNKNOWN is a valid value and it is possible for a decoder object to have that value for the media_type parameter if it has not been initialized.

channels/tsmf/client/gstreamer/tsmf_gstreamer.c
bjcollins: Oh, I think I have other comments that need to be multiline as well, I can do your suggestion to make the value more readable.

channels/tsmf/client/gstreamer/tsmf_gstreamer.c
hardening: Well in fact this should be a switch statement !

channels/tsmf/client/gstreamer/tsmf_gstreamer.c
bjcollins: agreed, should probably just remove the whole ack_cb stuff from gstreamer decoder code.

channels/tsmf/client/tsmf_main.c
bjcollins: Ok, just for reference - there are other areas in tsmf that do not check the return value for this call either. I will try to hit all cases with an updated pull request.

channels/tsmf/client/tsmf_media.c
bjcollins: Agreed

channels/tsmf/client/tsmf_media.c
bjcollins: I think this is correct, it matches my description for the return code. I use TRUE to indicate that no samples are currently available, which would be the case if there was no stream. Make sense at all?

channels/tsmf/client/gstreamer/tsmf_gstreamer.c
bjcollins: Ok, no problem. I will change it to a switch statement.

channels/tsmf/client/tsmf_media.c
bjcollins: Yes, I am now calling this function from tsmf_ifman.c t support flush message per stream instead of for the entire presentation.

bjcollins:
I looked at the output and I think my PR is probably flawed in terms of gstreamer 1.0 support as I was focused on gstreamer 0.10, I will look at fixing and resubmitting along with addressing various comments made against the PR.

hardening:
@bjcollins don't invest too much in checks, we have a PR addressing all the checks that should be done and aren't

bjcollins:
Ok. I just did something quick checks based on those few Stream_New calls, finished incorporating your comments also. But need to make fixes and do some Gstreamer 1.0 testing. I made some changes that were not compatible across both versions of gstreamer. Hoping to resubmit today.

bjcollins:
Ok. I made some changes based on the feedback so far and fixed the gstreamer 1.0 build issue from my first commit. Retested both a gstreamer 0.10 and gstreamer 1.0 build.

akallabeth:
lgtm, +1

bjcollins:
Do I need to do something more to resolve the conflict? My original checkin failed, but I have since commited a fix to that.

akallabeth:
@bjcollins Looks like you need to rebase your branch.

bjcollins:
Ok, I updated I fixed this properly now.

@akallabeth Thanks for hint.  I do not use git much, but I also did not realize I was actually in conflict. Just thought it was still saying it was broke due to my first broke commit.

bjcollins:
Rebased this again. Hopefully, it can be reconsidered now.

akallabeth:
@awakecoding @hardening @bmiklautz This branch looks good to me, is there something preventing merging?

hardening:
Nope, It looks good to me. We were waiting for the channel hardening to be in to consider merging this one. Now that it's upstream, let's merge.

PR #2774, opened 2015-07-10
CertStore now opens files on creation and closes on destruction.

akallabeth:
* Fixes a synchronization issue due to the OS flushing new data too late.
* Legacy certificate accept now writes subject and issuer to new entry.
* Improves error checks

### DIFF
--- a/libfreerdp/crypto/certificate.c
+++ b/libfreerdp/crypto/certificate.c
@@ -98,15 +98,28 @@ BOOL certificate_store_init(rdpCertificateStore* certificate_store)
 								(char*) certificate_legacy_hosts_file)))
 		goto fail;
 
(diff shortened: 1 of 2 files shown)


libfreerdp/crypto/certificate.c
```diff
@@ -178,20 +197,27 @@ static int certificate_data_match_legacy(rdpCertificateStore* certificate_store,
...
+		rdpCertificateData* data = certificate_data_new(
+						   hostname,
+						   certificate_data->port,
+						   NULL, NULL,
+						   certificate_data->fingerprint);
 		if (data)
+		{
+			free (data->subject);
+			free (data->issuer);
+			data->subject = _strdup(certificate_data->subject);
```
hardening: strdups not checked here

PR #2779, opened 2015-07-13
Send correct buttons when using a left-handed mouse

RolKau:
If you connect to a Windows server which is set up to use the left button as primary, from an X.org client which is also configured this way, the mouse buttons will appear switched in the RDP session!

This is because RDP expects physical button values to be sent whereas the client currently sends the X.org button event argument which are logical buttons values. The buttons are thus mapped twice.

This changeset implements an indirection where the logical buttons are first unmapped/reverse mapped back to the physical buttons, which are then sent over the wire and can then be mapped by the server. This enables a user to keep the same configuration at both the console and in a remote session.

I have enabled this mapping by default since a remote session should ideally behave like a console session. Left-handed server administrators that only configured their workstation but didn't bother to set the option at the server can use an option to get the back the old behaviour.

For those using a right-handed mouse (at both client and server), there should be no changes.

Tested with Ubuntu Precise 12.04 connecting to Windows 7 SP1.

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -1045,7 +1045,10 @@ static void xf_button_map_init (xfContext* xfc)
 	};
 
 	/* query system for actual remapping */
-	xf_get_x11_button_map (xfc, x11_map);
+	if (!xfc->settings->UnmapButtons)
+	{
+		xf_get_x11_button_map (xfc, x11_map);
+	}
 
 	/* iterate over all (mapped) physical buttons; for each of them */
 	/* find the logical button in X11, and assign to this the       */
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -935,6 +935,46 @@ void xf_check_extensions(xfContext* context)
 #endif
 }
 
+/* Assignment of physical (not logical) mouse buttons to wire flags. */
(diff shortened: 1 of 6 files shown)


bmiklautz:
@freerdp-bot test.

RolKau:
Reduced patch footprint, hopefully without reducing readability. There should be no effective change, but to be on the safe side: @freerdp-bot test

bmiklautz:
@freerdp-bot test.

RolKau:
The logic of the flag was inverse of the wording; updated and squashed into the commit.

akallabeth:
LGTM +1

hardening:
Looks like this PR doesn't apply cleanly anymore. @RolKau any chance that you could update it ?

RolKau:
@hardening Rebased onto 1b177fa (current master)

hardening:
@RolKau I have reviewed again the patch, and I feel like the UnmapButton should be in the ABI stable part of settings, probably in the Input Capabilities section. @akallabeth any comment on this ?

akallabeth:
@RolKau @hardening Yes, I agree. Would suggest the section labeled ```/* Miscellaneous */```.

RolKau:
@akallabeth @hardening Like 6f6f0b0?

PR #2781, opened 2015-07-14
winpr/synch: remove the dependency on winsock.h

bmiklautz:
winsock.h pulls in a lot of defines and dependencies that are not
required and partially unwanted in winpr's core (for parts that are not
related to network). In order to get rid of this dependency and have an
independent defines for extended winpr functions the WINPR_FD_* defines
are used internally (and for exposed functions). Where required, like in
WSAEventSelect, the FD_* is mapped to WINPR_FD_*.


akallabeth:
+1

MartinHaimberger:
+1 LGTM

PR #2784, opened 2015-07-15
Stable 1.1

jenerpeon:


bmiklautz:
@jenerpeon why should we merge master into stable-1.1?

PR #2785, opened 2015-07-15
Channel hardening and fixing

MartinHaimberger:
- introduced WIN32ERROR type
- hardend all channels
- introduced system for error reporting


MartinHaimberger:
@ freerdp-bot please retest

MartinHaimberger:
@freerdp-bot would you be so kind and test this nice PR ?

bmiklautz:
Started with testing/review.

channels/audin/client/audin_main.c
```diff
@@ -348,30 +438,38 @@ static int audin_on_data_received(IWTSVirtualChannelCallback* pChannelCallback,
...
 	DEBUG_DVC("...");
 
 	if (audin->device)
-		IFCALL(audin->device->Close, audin->device);
+	{
+		IFCALLRET(audin->device->Close, error, audin->device);
+		if (error != CHANNEL_RC_OK)
+		{
+			WLog_ERR(TAG, "Close failed with errorcode %lu", error);
+			return error;
```
bmiklautz: We maybe should do the frees below also in error case.

channels/audin/client/oss/audin_oss.c
```diff
@@ -181,6 +191,7 @@ static void* audin_oss_thread_func(void* arg)
 	if ((pcm_handle = open(dev_name, O_RDONLY)) < 0)
 	{
 		OSS_LOG_ERR("sound dev open failed", errno);
+		error = (WIN32ERROR)errno;
```
bmiklautz: Not sure but does this mapping make sense in practice?

channels/audin/server/audin.c
```diff
@@ -132,29 +153,39 @@ static void audin_server_send_formats(audin_server* audin, wStream* s)
...
+				return CHANNEL_RC_NO_MEMORY;
+			}
 
 			Stream_Write(s, audin->context.server_formats[i].data,
 				audin->context.server_formats[i].cbSize);
 		}
 	}
 
-	WTSVirtualChannelWrite(audin->audin_channel, (PCHAR) Stream_Buffer(s), Stream_GetPosition(s), &written);
+	return WTSVirtualChannelWrite(audin->audin_channel, (PCHAR) Stream_Buffer(s), Stream_GetPosition(s), &written) ? CHANNEL_RC_OK : ERROR_INTERNAL_ERROR;
```
bmiklautz: This implicit cast from BOOL to WIN32ERROR is valid but the error code (0x01 for false) would mean ERROR_INVALID_FUNCTION.

bmiklautz: ahh my bad. Haven't seen the line below. Just ignore the comment :)

channels/audin/server/audin.c
```diff
@@ -367,31 +439,55 @@ static void* audin_server_thread_func(void* arg)
...
 				break;
 
 			case MSG_SNDIN_FORMATS:
-				if (audin_server_recv_formats(audin, s, BytesReturned))
-					audin_server_send_open(audin, s);
+				if ((error = audin_server_recv_formats(audin, s, BytesReturned)))
+				{
+                    WLog_ERR(TAG, "audin_server_recv_formats failed with error %lu!", error);
+                    goto out_capacity;
+                }
```
bmiklautz: Formatting.

mfleisz:
I'm getting a bunch of errors when compiling with VS 2010 (x64):

```
Error	1	error C2275: 'WIN32ERROR' : illegal use of this type as an expression	D:\FreeRDP-Server\channels\drive\client\drive_main.c	568
Error	13	error C2275: 'WIN32ERROR' : illegal use of this type as an expression	D:\FreeRDP-Server\channels\rdpdr\server\rdpdr_main.c	818
Error	14	error C2146: syntax error : missing ';' before identifier 'error'	D:\FreeRDP-Server\channels\rdpdr\server\rdpdr_main.c	818
Error	15	error C2065: 'error' : undeclared identifier	D:\FreeRDP-Server\channels\rdpdr\server\rdpdr_main.c	818
....
```

The errors are in drive_main.c and rdpdr_main.c. Seems like a missing include or something like that ...

Compiled with WITH_SERVER and WITH_SERVER_CHANNELS and with server side rdpgfx channel support.

mfleisz:
Interestingly the build works with Visual Studio 2013 (x64) ... so it seems to be a 2010 specific issue

channels/drdynvc/client/drdynvc_main.c
```diff
@@ -381,75 +435,103 @@ int dvcman_create_channel(IWTSVirtualChannelManager* pChannelMgr, UINT32 Channel
...
+		return ERROR_INTERNAL_ERROR;
 	}
 
-	if (channel->status == 0)
+	if (channel->status == CHANNEL_RC_OK)
 	{
 		pCallback = channel->channel_callback;
 		if (pCallback->OnOpen)
-			pCallback->OnOpen(pCallback);
+			if ((error = pCallback->OnOpen(pCallback)))
```
bmiklautz: Maybe we should add some braces around this if as well.

channels/drdynvc/client/drdynvc_main.c
```diff
@@ -381,75 +435,103 @@ int dvcman_create_channel(IWTSVirtualChannelManager* pChannelMgr, UINT32 Channel
...
+			return error;
+		}
 
 		WLog_DBG(TAG, "dvcman_close_channel: channel %d closed", ChannelId);
 
 		ichannel = (IWTSVirtualChannel*) channel;
 
 		if (ichannel->Close)
-			ichannel->Close(ichannel);
+			if ((error = ichannel->Close(ichannel)))
```
bmiklautz: Same as above. Maybe we should add braces in this case.

channels/encomsp/server/encomsp_main.c
```diff
@@ -174,7 +197,7 @@ static void* encomsp_server_thread(void* arg)
 
 	while (1)
 	{
-		status = WaitForMultipleObjects(nCount, events, FALSE, INFINITE);
+		WaitForMultipleObjects(nCount, events, FALSE, INFINITE);
```
bmiklautz: I think the return value should be checked here. In case of memory shortage WaitForMultipleObjects might return an error.

channels/parallel/client/parallel_main.c
```diff
@@ -241,31 +284,52 @@ static void* parallel_thread_func(void* arg)
...
+		if ((error = parallel_process_irp(parallel, irp)))
+		{
+			WLog_ERR(TAG, "parallel_process_irp failed with error %d!", error);
+			break;
+		}
 	}
+	if (error && parallel->rdpcontext)
+		setChannelError(parallel->rdpcontext, error, "parallel_thread_func reported an error");
 
+	ExitThread((DWORD)error);
```
bmiklautz: I wonder if we do need this ExitThread here. If we return the error directly this should be sufficient in my opinion.

channels/parallel/client/parallel_main.c
```diff
@@ -327,22 +401,31 @@ int DeviceServiceEntry(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints)
...
+		{
+			WLog_ERR(TAG, "CreateThread failed!");
+			error = ERROR_INTERNAL_ERROR;
+			goto error_out;
+		}
+	}
 
-error_thread:
+	return CHANNEL_RC_OK;
+error_out:
```
bmiklautz: I think having different lables as before might make sense at least for MessageQueue_Free. If I'm not mistaken MessageQueue_Free won't handle "NULL queues" at the moment.

channels/printer/client/printer_main.c
```diff
@@ -226,7 +270,7 @@ static void printer_free(DEVICE* device)
 	free(printer_dev);
 }
 
-void printer_register(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints, rdpPrinter* printer)
+WIN32ERROR printer_register(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints, rdpPrinter* printer)
 {
 	char* port;
```
bmiklautz: With the new error handling this should be initialized with NULL.

channels/rail/client/rail_orders.c
```diff
@@ -178,22 +204,32 @@ void rail_write_client_status_order(wStream* s, RAIL_CLIENT_STATUS_ORDER* client
...
+	{
+		return error;
+	}
+	if ((error = rail_write_unicode_string_value(s, &exec->workingDir)))
+	{
+		return error;
+	}
+	if ((error = rail_write_unicode_string_value(s, &exec->arguments)))
+	{
+		return error;
```
bmiklautz: what about doing a 
```
return rail_write_unicode_string_value(s, &exec->arguments);
```
here?

channels/rail/client/rail_orders.c
```diff
@@ -297,138 +333,173 @@ void rail_write_langbar_info_order(wStream* s, RAIL_LANGBAR_INFO_ORDER* langbarI
...
+	{
+		WLog_ERR(TAG, "rail_read_handshake_order failed with error %lu!", error);
+		return error;
+	}
 
 	if (context->custom)
 	{
-		IFCALL(context->ServerHandshake, context, handshake);
-		return TRUE;
+		IFCALLRET(context->ServerHandshake, error, context, handshake);
```
bmiklautz: Formatting: for single line if statements the braces should be removed.

channels/rail/client/rail_orders.c
```diff
@@ -297,138 +333,173 @@ void rail_write_langbar_info_order(wStream* s, RAIL_LANGBAR_INFO_ORDER* langbarI
...
+	if ((error = rail_read_handshake_ex_order(s, handshakeEx)))
+	{
+		WLog_ERR(TAG, "rail_read_handshake_ex_order failed with error %lu!", error);
+		return error;
+	}
 
 	if (context->custom)
 	{
-		IFCALL(context->ClientHandshakeEx, context, handshakeEx);
+		IFCALLRET(context->ClientHandshakeEx, error, context, handshakeEx);
```
bmiklautz: Formatting: for single line if statements the braces should be removed.

channels/rail/client/rail_orders.c
```diff
@@ -297,138 +333,173 @@ void rail_write_langbar_info_order(wStream* s, RAIL_LANGBAR_INFO_ORDER* langbarI
...
+	if ((error = rail_read_server_exec_result_order(s, execResult)))
+	{
+		WLog_ERR(TAG, "rail_read_server_exec_result_order failed with error %lu!", error);
+		return error;
+	}
 
 	if (context->custom)
 	{
-		IFCALL(context->ServerExecuteResult, context, execResult);
+		IFCALLRET(context->ServerExecuteResult, error, context, execResult);
```
bmiklautz: Formatting: for single line if statements the braces should be removed.

channels/rail/client/rail_orders.c
```diff
@@ -297,138 +333,173 @@ void rail_write_langbar_info_order(wStream* s, RAIL_LANGBAR_INFO_ORDER* langbarI
...
+	if ((error = rail_read_server_sysparam_order(s, sysparam)))
+	{
+		WLog_ERR(TAG, "rail_read_server_sysparam_order failed with error %lu!", error);
+		return error;
+	}
 
 	if (context->custom)
 	{
-		IFCALL(context->ServerSystemParam, context, sysparam);
+		IFCALLRET(context->ServerSystemParam, error, context, sysparam);
```
bmiklautz: Formatting: for single line if statements the braces should be removed.

channels/rail/client/rail_orders.c
```diff
@@ -297,138 +333,173 @@ void rail_write_langbar_info_order(wStream* s, RAIL_LANGBAR_INFO_ORDER* langbarI
...
+	if ((error = rail_read_server_minmaxinfo_order(s, minMaxInfo)))
+	{
+		WLog_ERR(TAG, "rail_read_server_minmaxinfo_order failed with error %lu!", error);
+		return error;
+	}
 
 	if (context->custom)
 	{
-		IFCALL(context->ServerMinMaxInfo, context, minMaxInfo);
+		IFCALLRET(context->ServerMinMaxInfo, error, context, minMaxInfo);
```
bmiklautz: Formatting: for single line if statements the braces should be removed.

channels/rail/client/rail_orders.c
```diff
@@ -297,138 +333,173 @@ void rail_write_langbar_info_order(wStream* s, RAIL_LANGBAR_INFO_ORDER* langbarI
...
+	if ((error = rail_read_server_localmovesize_order(s, localMoveSize)))
+	{
+		WLog_ERR(TAG, "rail_read_server_localmovesize_order failed with error %lu!", error);
+		return error;
+	}
 
 	if (context->custom)
 	{
-		IFCALL(context->ServerLocalMoveSize, context, localMoveSize);
+		IFCALLRET(context->ServerLocalMoveSize, error, context, localMoveSize);
```
bmiklautz: Formatting: for single line if statements the braces should be removed.

channels/rail/client/rail_orders.c
```diff
@@ -297,138 +333,173 @@ void rail_write_langbar_info_order(wStream* s, RAIL_LANGBAR_INFO_ORDER* langbarI
...
+	if ((error = rail_read_server_get_appid_resp_order(s, getAppIdResp)))
+	{
+		WLog_ERR(TAG, "rail_read_server_get_appid_resp_order failed with error %lu!", error);
+		return error;
+	}
 
 	if (context->custom)
 	{
-		IFCALL(context->ServerGetAppIdResponse, context, getAppIdResp);
+		IFCALLRET(context->ServerGetAppIdResponse, error, context, getAppIdResp);
```
bmiklautz: Formatting: for single line if statements the braces should be removed.

channels/rail/client/rail_orders.c
```diff
@@ -297,138 +333,173 @@ void rail_write_langbar_info_order(wStream* s, RAIL_LANGBAR_INFO_ORDER* langbarI
...
+	if ((error = rail_read_langbar_info_order(s, langBarInfo)))
+	{
+		WLog_ERR(TAG, "rail_read_langbar_info_order failed with error %lu!", error);
+		return error;
+	}
 
 	if (context->custom)
 	{
-		IFCALL(context->ServerLanguageBarInfo, context, langBarInfo);
+		IFCALLRET(context->ServerLanguageBarInfo, error, context, langBarInfo);
```
bmiklautz: Formatting: for single line if statements the braces should be removed.

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -443,16 +487,23 @@ static void* drive_hotplug_thread_func(void* arg)
...
+		ExitThread(ERROR_INTERNAL_ERROR);
+		return NULL;
+	}
 
 	mfd = open("/proc/mounts", O_RDONLY, 0);
 
 	if (mfd < 0)
 	{
 		WLog_ERR(TAG, "ERROR: Unable to open /proc/mounts.");
+		ExitThread(ERROR_INTERNAL_ERROR);
```
bmiklautz: Maybe I've missed it somehow but shouldn't we inform the main thread if we exit?

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -461,7 +512,12 @@ static void* drive_hotplug_thread_func(void* arg)
 	tv.tv_sec = 1;
 	tv.tv_usec = 0;
 
-	handle_hotplug(rdpdr);
+	if ((error = handle_hotplug(rdpdr)))
+	{
+		WLog_ERR(TAG, "handle_hotplug failed with error %lu!", error);
+		ExitThread((DWORD)error);
```
bmiklautz: Same comment as above.

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -973,45 +1171,46 @@ static void* rdpdr_virtual_channel_client_thread(void* arg)
...
-	if (!rdpdr_add_open_handle_data(rdpdr->OpenHandle, rdpdr))
+	if ((error = rdpdr_add_open_handle_data(rdpdr->OpenHandle, rdpdr)))
 	{
-		WLog_ERR(TAG,  "%s: unable to register open handle", __FUNCTION__);
-		return;
+		WLog_ERR(TAG, "rdpdr_add_open_handle_data failed with error %lu!", error);
+		return error;
 	}
 
 	if (status != CHANNEL_RC_OK)
```
bmiklautz: This was already like this before the commit and maybe I'm wrong but shouldn't we do the check for status before rdpdr_add_open_handle_data is done?

channels/encomsp/server/encomsp_main.c
MartinHaimberger: right i will do that

channels/drdynvc/client/drdynvc_main.c
MartinHaimberger: This will break coding style

channels/drdynvc/client/drdynvc_main.c
MartinHaimberger: This will break coding style

channels/parallel/client/parallel_main.c
MartinHaimberger: ExitThread is the preferd methode to close threads in C.

channels/parallel/client/parallel_main.c
MartinHaimberger: will do that

channels/rail/client/rail_orders.c
MartinHaimberger: there are three directorys written not one.

channels/rdpdr/client/rdpdr_main.c
MartinHaimberger: true will do that

channels/rdpdr/server/rdpdr_main.c
```diff
@@ -725,13 +805,12 @@ static BOOL rdpdr_server_send_user_logged_on(RdpdrServerContext* context)
 	status = WTSVirtualChannelWrite(context->priv->ChannelHandle, (PCHAR) Stream_Buffer(s), Stream_Length(s), &written);
 
 	Stream_Free(s, TRUE);
-
-	return status;
+	return status ? ERROR_SUCCESS : ERROR_INTERNAL_ERROR;
```
bmiklautz: I was wondering if there is a reason why ERROR_SUCCESS is used here instead of CHANNEL_RC_OK.

client/X11/xf_gfx.c
```diff
@@ -119,7 +119,7 @@ int xf_OutputUpdate(xfContext* xfc, xfGfxSurface* surface)
 
 int xf_UpdateSurfaces(xfContext* xfc)
 {
-	int count;
+	UINT16 count;
```
bmiklautz: Why the UINT16 here?

client/X11/xf_gfx.c
```diff
@@ -151,7 +151,7 @@ int xf_UpdateSurfaces(xfContext* xfc)
 
 int xf_OutputExpose(xfContext* xfc, int x, int y, int width, int height)
 {
-	int count;
+	UINT16 count;
```
bmiklautz: Same as above I wonder why UINT16 was used here.

bmiklautz:
@MartinHaimberger this is a huge pull request one could say (it is even pushing github to it's diff limits ;).
Big props! - Having better error checking in the channels was long overdue. Very nicely and well done.

I reviewed the PR and tested most plugins (at least on Linux) but due it's size it still might 
be the case that I've overlooked some things. Anyway since it brings so much improvements I'd suggest that we merge it once the comments are integrated and https://github.com/MartinHaimberger/FreeRDP/pull/3 is merged on your branch.

@awakecoding  @mfleisz @hardening @nfedera  any objections/ideas?

client/X11/xf_gfx.c
MartinHaimberger: because GetSurfaceIds returns a UINT16

client/X11/xf_gfx.c
MartinHaimberger: because GetSurfaceIds returns a UINT16

MartinHaimberger:
@freerdp-bot would you be so kind and test this nice PR ?

hardening:
@bmiklautz  I agree with the idea of merging and then we'll see. The PR is so big that it's quite hard to review.

HenryJacques:
@MartinHaimberger On Ubuntu 14.04, your branch mh-channel-fix with usb redirection is unusable. xfreerdp starts then immediately stops.
If I don't redirect an USB device, xfreerdp launches correctly.

```[13:34:07:424] [14547:14548] [INFO][com.freerdp.channels.drdynvc.client] - Loading Dynamic Virtual Channel urbdrc
[13:34:07:424] [14547:14548] [ERROR][com.freerdp.channels.urbdrc.client] - error processing arguments
[13:34:07:426] [14547:14548] [WARN][com.freerdp.channels.urbdrc.client] - bus:0 dev:0 not exist in udevman
[13:34:07:426] [14547:14548] [INFO][com.freerdp.channels.urbdrc.client] - VID: 0x079B, PID: 0x0024
[13:34:07:427] [14547:14548] [WARN][com.freerdp.channels.urbdrc.client] - bus:1 dev:9 not exist in udevman
[13:34:12:486] [14547:14554] [ERROR][com.freerdp.channels.drdynvc.client] - ChannelId 9 not found!
[13:34:12:487] [14547:14554] [ERROR][com.freerdp.channels.drdynvc.client] - dvcman_close_channel failed with error 1359!
[13:34:12:487] [14547:14554] [ERROR][com.freerdp.channels.drdynvc.client] - drdynvc_order_recv failed with error 1359!
[13:34:12:487] [14547:14548] [ERROR][com.freerdp.core] - drdynvc_virtual_channel_client_thread reported an error. Error was 1359
[13:34:12:487] [14547:14548] [INFO][com.freerdp.client.x11] - Network disconnect!
[13:34:12:487] [14547:14548] [ERROR][com.freerdp.client.x11] - Failed to check FreeRDP file descriptor```

hardening:
Looks like this one is my fault

MartinHaimberger:
@awakecoding, @hardening, @bmiklautz, @nfedera 

Ok, how do we proceed now ? Letting the pull request rest forever is no option!

Nice regards!

awakecoding:
@MartinHaimberger I do have a few questions/comments:
This pull request is absolutely huge, and goes beyond the view limit on github. I've never seen something this big. Why choose "WIN32ERROR" as a new type, create a new header for a single typedef and include it everywhere? I have a problem understanding why this is a good idea. Worst case, why not simply use a compatible type to hold the error codes, the type itself doesn't appear to bring any value here, the hardening changes do.

This being said, I agree with hardening changes, but I am definitely not sure about the whole gigantic win32error.h / WIN32ERROR type change. Did I miss something or would the exact same be doable with a simple UINT32 type and no win32error.h?

awakecoding:
@MartinHaimberger would it be possible to do a search & replace on the WIN32ERROR type to change it to a generic type like UINT32, and remove the new win32error.h header? This would isolate this PR as simply adding the hardening code.

bmiklautz:
@awakecoding the return values in all the channel code were a real mess! Every function returned different things sometimes 0 was success sometimes it wasn't, sometimes BOOL was used - and most stuff was void anyway which is't good practice.. I personally think the introduction of a new type makes it much much clearer, consistent and is very useful and I don't see the problem here. 

"... the type itself doesn't appear to bring any value here, ..." - Wrong. I think they really do. If you look at the code carefully you will see that the return values are passed out until the main thread where they can be used to indicate not only that a problem happened but also what kind of problem.

That said I personally wouldn't change the return type and leave it as it is. But we don't necessarily need it in winpr so we could remove it from there and move it to libfreerdp.

nfedera:
@awakecoding I agree with @bmiklautz.
The initial requirement was that all functions should return a [win32 error code](https://msdn.microsoft.com/en-us/library/cc231199.aspx) since this is the most common set of error defines. Unfortunately Microsoft has not defined a type for these kind of functions.
If you grep MSDN for this type of functions (returning ERROR_SUCCESS on success and values greater ERROR_SUCCESS  on failure) you'll find a hell of inconsistency (probably because of historical reasons) in the return types (LONG, UINT, DWORD etc).
The use of alternative, sane return value types like [NTSTATUS](https://msdn.microsoft.com/en-us/library/cc231200.aspx) or [HRESULT](https://msdn.microsoft.com/en-us/library/cc231198.aspx) has been ruled out for several reasons (notably because of unnecessary complexity caused by the required usage of error code conversion and checking macros).
IMHO it is big advantage if the return type implies how the result has to be handled (like with BOOL, HRESULT or NTSTATUS).
Of course it is easy to search and replace the WIN32ERROR type with UINT or DWORD but we'd lose quite some convenience IMHO and there is no easy going back.

vworkspace:
My vote is for keeping it as DWORD and not defining a new type called WIN32ERROR.
1.  DWORD is the accepted way of referring to Win32 error codes as evidenced by the function prototype "DWORD WINAPI GetLastError(void);".  So why define a new type?
2.  I don't see the point of creating a new header file just to have a single type definition.

bmiklautz:
@vworkspace then we use an existing header file if that's your problem. As mentioned above we can pull that out from winpr. 
Also you might want to re-read @nfedera comment above on the types. It's not that @MartinHaimberger chose any arbitrary type just it looked nice. The pull request as it stands now was re-worked in different iteration and the current solution was considered the best approach from all reviewers until now.

It's really "funny" that more than a month good hardening work (that was damn necessary) lays around because of some disagreement about some type, that when you think it all through and review the request carefully, really brings advantages.

awakecoding:
@bmiklautz I think we can all agree on the error checking and hardening changes. the source of strong disagreement is the addition of this new type that gets propagated everywhere. I am willing to merge this if the new type is removed in favour of a generic type. It's "sad" that Microsoft didn't provide a type specifically to contain Win32 error codes, but I don't think we need to "fix" this ourselves in WinPR or FreeRDP.

hardening:
I really don't see why we should do it wrong as microsoft did... The introduction of WIN32ERROR make the return value very clear, while a DWORD can be an error code or just a value.
I'd be ok to remove it from WinPR, but let's be smarter for FreeRDP !

nfedera:
@awakecoding What about the following proposal:
 1. remove the typedef and header file from WinPR
 1. create a DWORD or UINT define for WIN32ERROR in the channel part of FreeRDP 

This would still have the big advantage of the indication how to handle the return value of the individual functions returning win32 error codes but this change doesn't propagate WIN32ERROR everywhere.

awakecoding:
@nfedera I am against adding this new type, may it be in FreeRDP or WinPR. This is way overkill and makes the code harder to read for outsiders. You may see it otherwise, but I just can't accept this.

lemonmojo:
As an "outsider" or "consumer" of the API I'd appreciate the additional hint to what a function returns.
For me, something generic like DWORD, UINT, etc. doesn't convey the message of "Check that return value before you proceed" as much as a dedicated error type.

So for me this would improve readability. But that's just my 2 cent...

awakecoding:
@lemonmojo "readability hints" like this are the kind of thing I don't want to see in FreeRDP/WinPR. It gives the false illusion that the code is more readable. In *all* cases, a programmer has to check what a function can possibly return in order to properly use it. Some people have pointed out that Microsoft made a "mistake" by not defining an explicit error-container data type for the Win32 error codes. This same "mistake" is apparently done by the POSIX API, and I don't see how this is a problem. errno is simply defined as an int. Common operating system functions regularly use int as a return type, where the values indicating error change from function to function. Yes, certain things might be better if they were all made consistent, but in practice things aren't as easy as it seems. It's not the end of the world if the code isn't 100% consistent with all type-enforced consistency. Try to convince me that I won't see a flood of pull requests following this one that will do nothing but change EVERYTHING to this new WIN32ERROR type, because I definitely see that happening. Can we actually work on the real problems here, not the fake ones? The real issue at stake here is improving error checking and making it error consistent, and that doesn't require adding a new type for everything.

lemonmojo:
@awakecoding 
> In all cases, a programmer has to check what a function can possibly return in order to properly use it.

I'm absolutely with you there!
But for me that doesn't mean looking that information up in the actual implementation of a function, which might span hundreds or thousands of lines. I want to see this by looking at the corresponding header file.

Of course that doesn't require an explicit error type, good documentation does the job just fine.
You mentioned the POSIX API which I think is a good example for this. I never had to look at the implementation of a POSIX function to understand what I can expect it to do and return, because the documentation is just good enough to give me that information.

Now of course we all know that it's hard to write good documentation, keep it up to date and make it understandable for "outsiders".
So introducing an error type like this would be a way of giving developers at least a hint at what to expect without writing documentation for each and every function.

In any case, and whatever path you take, let me take this opportunity to thank you all for your wonderful and hard work! Keep it up and don't let discussions like this grind you down! =)

bmiklautz:
@awakecoding I just lol! Return types seem to be more important than having proper error handling..

bmiklautz:
@awakecoding I still don't get all the buzz and problem you have with a single type. Typedefs and defines are all over the place in FreeRDP what's so special or problematic for you about this one? Honestly? I would really like to understand it but I just don't see it. Behind the typedef it's still a standard type.

Current state in FreeRDP is that almost *every* function returns something different. There was absolutely *no* coherence. For sure different subsystem like gfx partially used similar return values which makes sense but in general it was (and is) still a nightmare.
When the hardening was started a lot of function were moved to BOOL when no additional information was needed.
 
Your so called "readability hints" make working with the channel subsystem(s) much much easier properly used a programmer need only to check for *one* well known value to detect an error and still has the possibility to figure out what really happened and reacted based not the value returned. I might not get it but in my opinion this really does make sense, saves time and does make it more readable.

MartinHaimberger:
@awakecoding: I have now removed the WIN32ERROR type and using UINT instead. So nothing should block the merge any more ...

awakecoding:
@MartinHaimberger thank you! I appreciate it :)

PR #2786, opened 2015-07-15
winpr/tools: Fixed x509_get_default_name failing with long computer names

mfleisz:
x509_get_default_name used to fail if the computer name was longer than 15 characters.
Now we will first determine the actual length required to hold the computer name and then we start the actual computer name query with the correct buffer size.
We are also using ComputerNamePhysicalNetBIOS instead of ComputerNameNetBIOS because later will return the NetBIOS name of the cluster virtual server if the computer is a node in a cluster.


MartinHaimberger:
+1 LGTM

PR #2788, opened 2015-07-16
rdpsnd: change plugin load order

bmiklautz:
The recently added OSS support is recommended by default and if found
loaded before ALSA. This possibly breaks old command lines if ALSA
wasn't specified (like if only /sound was used). To make sure this doesn't cause any problems the order
is changed to PULSE,ALSA,OSS.


hardening:
+1

bmiklautz:
Closing in favor of PR #2785 which contains this change as well.

PR #2789, opened 2015-07-16
[Thread] Fixed mutex unlocking and return checks.

akallabeth:
* Fixes a missing unlock mutex in error case.
* Adds checks to pthread mutex lock and unlock functions.


hardening:
+1

bmiklautz:
+1

Issue #2790, opened 2015-07-17
Latest commit 042cb25 - xfreerdp CPU 100% and fails


HenryJacques:
The latest commit 042cb250f3452206140b652370f01649a6a16039 brings a regression on my Ubuntu 14.04 64bits.
When trying to connect to Windows 7 SP1 64bits:

The process goes to 100% during one minute and then it fails.
The length of the logging [INFO] is variable but it always fail in the end.

root@ubuntu:~/FreeRDP-master-17-07# xfreerdp /u:user /d:DOMAIN /p:'password' /v:10.10.10.10 /cert-ignore /sec:nla
[10:00:50:183] [13897:13898] [INFO][com.winpr.sspi.NTLM] - VERSION ={
[10:00:50:183] [13897:13898] [INFO][com.winpr.sspi.NTLM] -      ProductMajorVersion: 6
[10:00:50:183] [13897:13898] [INFO][com.winpr.sspi.NTLM] -      ProductMinorVersion: 1
[10:00:50:183] [13897:13898] [INFO][com.winpr.sspi.NTLM] -      ProductBuild: 7601
[10:00:50:183] [13897:13898] [INFO][com.winpr.sspi.NTLM] -      Reserved: 0x000000
[10:00:50:183] [13897:13898] [INFO][com.winpr.sspi.NTLM] -      NTLMRevisionCurrent: 0x0F

[10:01:50:154] [13897:13898] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_CONNECT_TRANSPORT_FAILED [0x2000D]
[10:01:50:154] [13897:13898] [ERROR][com.freerdp.client.x11] - Authentication only, exit status 1

This is what top gives me:

top - 10:01:00 up 10 days, 15:23,  4 users,  load average: 0,16, 0,07, 0,16
Tasks: 125 total,   1 running, 124 sleeping,   0 stopped,   0 zombie
%Cpu(s): 16,1 us,  9,1 sy,  0,0 ni, 74,9 id,  0,0 wa,  0,0 hi,  0,0 si,  0,0 st
KiB Mem:   6016148 total,  2571960 used,  3444188 free,   296280 buffers
KiB Swap:  6194172 total,        0 used,  6194172 free.  1793176 cached Mem

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
[...]

PR #2791, opened 2015-07-17
transport: fix regression introduced with PR #2770

bmiklautz:
When the event is reset in transport_check_fds xfreerdp doesn't work and
consumes 100% CPU (see #2790). On windows this is require otherwise the
CPU consumption is 100% there.
This quick fix only resets the event on windows. It's a working approach
but definitely not the final solution.


mfleisz:
+1

MartinHaimberger:
+1 LGTM

Issue #2733, opened 2015-06-25
xfreerdp hangs when closing Windows session with redirected smartcard


HenryJacques:
Hello,

I am using FreeRDP from GIT Master branch (from today) on a Ubuntu 12.04 LTS5 to connect to
- a Windows 2012 R2
- a Windows 8.1 Entreprise

Using the following command line:

xfreerdp /v:w8-test +toggle-fullscreen /cert-ignore +compression /compression-level:2 +heartbeat /d:DOMAIN /u:user /p:password /f /kbd:french +auto-reconnect /client-hostname:HOST /sec:nla /smartcard:0f14:0038

On both hosts, redirecting a smartcard causes a problem when I disconnect the session: the screen remains black, xfreerdp hangs forever....

However, I am using a card reader so if I remove the card (or insert it - so if I change the "state" of the reader), it unblocks the process:

The last errors (CHANNEL_RC_NOT_CONNECTED) are printed during the black screen. I interrupt them when removing the card...

[15:39:33:185] [13348:13356] [INFO][com.freerdp.channels.rdpdr.client] - Loading device service smartcard (static)
[15:39:33:199] [13348:13356] [INFO][com.freerdp.channels.rdpdr.client] - Loading device service smartcard (static)
[15:39:33:704] [13348:13356] [INFO][com.freerdp.channels.rdpdr.client] - registered device #1: SCARD (type=32 id=1)
[15:39:33:704] [13348:13356] [INFO][com.freerdp.channels.rdpdr.client] - registered device #2: SCARD (type=32 id=2)
[15:39:34:396] [13348:13356] [INFO][com.freerdp.channels.rdpdr.client] - registered device #1: SCARD (type=32 id=1)
[15:39:34:396] [13348:13356] [INFO][com.freerdp.channels.rdpdr.client] - registered device #2: SCARD (type=32 id=2)
[15:39:35:115] [13348:13364] [WARN][com.freerdp.channels.smartcard.client] - IRP failure: SCardGetStatusChangeW (0x000900A4), status: SCARD_E_CANCELLED (0x80100002)
[15:39:51:892] [13348:13349] [ERROR][com.freerdp.core] - ERRINFO_LOGOFF_BY_USER (0x0000000C):The disconnection was initiated by the user logging off his or her session on the server.
[15:39:51:017] [13348:13349] [ERROR][com.freerdp.core.rdp] - DisconnectProviderUltimatum: reason: 3
[...]

PR #2793, opened 2015-07-20
Fix for transactions support for Smart Cards

rkondratenko:
This patch addresses incorrect  behavior of FreeRDP with Smart Card transactions. Current version had transactions disabled because of these issues. The issues addressed here are:
1. Correcting condition for SCardIsValidContext for it to be true when the context is actually valid.
2. Releasing all contexts on a smartcard to cancel all commands waiting for transactions before waiting for it's thread to exit.
3. Removing "private" context from card handle in PCSC-Lite adapter, so that transactions do unlock when context is released.

### DIFF
--- a/channels/smartcard/client/smartcard_main.c
+++ b/channels/smartcard/client/smartcard_main.c
@@ -150,7 +150,7 @@ static void smartcard_release_all_contexts(SMARTCARD_DEVICE* smartcard) {
 
 			hContext = pContext->hContext;
 
-			if (SCardIsValidContext(hContext))
+			if (SCardIsValidContext(hContext) == SCARD_S_SUCCESS)
 			{
 				SCardCancel(hContext);
 			}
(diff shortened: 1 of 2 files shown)


bmiklautz:
@rkondratenko do you have a test/verification case as well?

rkondratenko:
@bmiklautz
I do but it requires a smart card and an AD configured for Smart Card Logon.
The test case I was using is:
1. enable transactions in the code
2. RDP login to Windows machine by either a Smart card or a password
3. Disconnect (do not log off)
4. Login ther again by the Smart Card, resuming the session.
5. Smart Card is unusable from the session as all calls block due to active transaction. (Try 'certutil -scinfo' as a quick test )
6. Try to disconnect or log off. Application hangs.

This patch addresses this but it has scope in more than just this issue.
This is the most obvious symptom resulting from a combination of problems.

0. When authenticating using a smartcard Windows acquires transaction for that smartcard.
1. When reconnecting an existing session server sends second announce, requiring client to drop all device reference and start sequence from scratch. Windows does not send EndTransaction in this case as it expects client to release all device references (i.e. PCSC contexts).
2. smartcard_init gets called but it does not release contexts, causing that transaction to last forever, making smartcard unusable.
3. When application is about to close it calls smartcard_free that sends Quit signals to all context threads and waits for them to exit. But most of them are blocked due to hanging transaction and this causes a deadlock.

bmiklautz:
@freerdp-bot test

HenryJacques:
This PR fixes my issue #2733 
Thanks!

PR #2798, opened 2015-07-21
This array is so big that it eats whole stack for the thread and makes

psumbera:
the process dump core on some systems (e.g. Solaris).

### DIFF
--- a/server/shadow/shadow_capture.c
+++ b/server/shadow/shadow_capture.c
@@ -80,6 +80,9 @@ int shadow_capture_align_clip_rect(RECTANGLE_16* rect, RECTANGLE_16* clip)
 	return 1;
 }
 
+/* this is too big to consume whole stack on some systems */
+static BOOL grid[1024][1024];
+
 int shadow_capture_compare(BYTE* pData1, int nStep1, int nWidth, int nHeight, BYTE* pData2, int nStep2, RECTANGLE_16* rect)
 {
 	BOOL equal;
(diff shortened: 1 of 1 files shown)


bmiklautz:
@psumbera what about increasing the stack size?

psumbera:
I was thinking about increasing stack size. But didn't went into details yet. It would have to be done probably during pthread_create call. Where should I have look?

psumbera:
It seems that I can add stack size to CreateThread call in x11_shadow_subsystem_start and it helps. But is it sane to have on stack:

BOOL grid[1024][1024];

which is 4MB? Where stack according pthread_create man page on Solaris has by default just 2MB.

hardening:
-1
setting this big array as static is a non-sense. The array is allocated on the stack to prevent constant malloc / free, but perhaps it would be the safest thing to do...

nfedera:
Why not simply delete it. This variable does not seem to be used ;)

nfedera:
Also cols[1024] is only used for some disabled debug code ( by using if(0) ... sick) .
This should be en/disabled using a define.
And instead of using 4 bytes (BOOL) for one flag this shoud be changed to use simple bit magic which would reduce the size down to 1/32.

nfedera:
@psumbera Are you up to this task?

psumbera:
Nice! I can do removal part tomorrow. Not sure about the rest now. But if you can do both I wouldn't object ;-)

bmiklautz:
@nfedera nice solution!

psumbera:
Please see my new commit. I didn't play with bit magic though...

bmiklautz:
@freerdp-bot test

bmiklautz:
+1

Issue #153, opened 2011-10-17
handle smartcard reader reconnection


atong:
For reference, https://github.com/FreeRDP/FreeRDP-old/issues/19

smartcard inserts & removals should work, but reader removal & reconnection isnt handled properly.

bmiklautz:
smartcard has improved a lot since this issue was opened. Closing for new please re-open or create a new issue if there are still problems.

Issue #2631, opened 2015-05-19
The point of specifying the smartcard device id?


HenryJacques:
Hi,

I was wondering why it is important/mandatory to specify the smartcard usb id using the parameter /smartcard:device
I am asking the question because on my setup, I have two smarcard readers but ideally I just want to redirect one. However even if I specify just one reader on the command line, freerdp detects both and redirected them, (see below).

Using mstsc, you cannot redirect just one reader, the option is global.
Does the RDP specifications say something about this?

(btw I don't know if the same behaviour happens with other device types: printer, serial port, ...)

Thanks!

user@host:~$ tsocks xfreerdp /tmp/rdp.rdp +toggle-fullscreen /cert-ignore +compression /compression-level:2 +heartbeat /d:domain /u:userl '/p:password' /sec:nla /kbd:french +auto-reconnect /client-hostname:myhostname /printer:Officejet_100_Mobile_L411 /smartcard:0f14:0038
[18:02:51:767] [2571:2572] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpdr
[18:02:51:768] [2571:2572] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpsnd
[18:02:51:768] [2571:2572] [INFO][com.freerdp.client.common.cmdline] - loading channel cliprdr
[18:02:52:652] [2571:2579] [INFO][com.freerdp.channels.rdpdr.client] - Loading device service printer (static)
[18:02:52:652] [2571:2579] [INFO][com.freerdp.channels.rdpdr.client] - Loading device service smartcard (static)
[18:02:52:652] [2571:2579] [INFO][com.freerdp.channels.rdpdr.client] - Loading device service printer (static)
[18:02:53:670] [2571:2579] [INFO][com.freerdp.channels.rdpdr.client] - registered device #1: SCARD (type=32 id=2)
[18:02:56:532] [2571:2579] [INFO][com.freerdp.channels.rdpdr.client] - registered device #1: PRN1 (type=4 id=1)
[18:02:56:532] [2571:2579] [INFO][com.freerdp.channels.rdpdr.client] - registered device #2: SCARD (type=32 id=2)

HenryJacques:
Hi,

No one?

By the way how can I make FreeRDP more verbose? I compiled it with WITH_DEBUG_ALL and BUILD_TYPE=Debug but I don't seem to have more output....

Thanks

HenryJacques:
Me again,

[...]

PR #2799, opened 2015-07-22
Smartcard Reader Connection Handling

kaliway:
This pull request addresses smartcard reader connection and reconnection issues (see #153)

smartcard_ListReadersA_Call & smartcard_ListReadersW_Call aborts without replying to the server when SCardListReaders return status is not SCARD_SUCCESS. As a result, SCardListReaders calls invoked by the server never get completed. This makes impossible to detect connection of a new smartcard reader when there are no other smartcard readers connected.

smartcard_ListReaders(A/W)_Call should pass the return status of SCardListReaders back to the server in all cases. This is particularly important when there are no smartcard readers in the system and SCardListReaders returns SCARD_E_NO_READERS_AVAILABLE.

Test case:
1. Start FreeRDP with no smartcard readers connected:
        ./xfreerdp -sec-nla /smartcard /v:<rdp-server-address>
2. Connect a smartcard reader;
3. The remote server detects connection of the reader;

### DIFF
--- a/channels/smartcard/client/smartcard_operations.c
+++ b/channels/smartcard/client/smartcard_operations.c
@@ -320,9 +320,6 @@ static UINT32 smartcard_ListReadersA_Call(SMARTCARD_DEVICE* smartcard, SMARTCARD
 		call->mszGroups = NULL;
 	}
 
-	if (status != SCARD_S_SUCCESS)
-		return status;
-
 	smartcard_trace_list_readers_return(smartcard, &ret, FALSE);
 	status = smartcard_pack_list_readers_return(smartcard, irp->output, &ret);
 
@@ -372,9 +369,6 @@ static UINT32 smartcard_ListReadersW_Call(SMARTCARD_DEVICE* smartcard, SMARTCARD
 		free(call->mszGroups);
 		call->mszGroups = NULL;
 	}
-
-	if (status != SCARD_S_SUCCESS)
-		return status;
 
 	smartcard_trace_list_readers_return(smartcard, &ret, TRUE);
 
--- a/channels/smartcard/client/smartcard_operations.c
+++ b/channels/smartcard/client/smartcard_operations.c
@@ -323,11 +323,11 @@ static UINT32 smartcard_ListReadersA_Call(SMARTCARD_DEVICE* smartcard, SMARTCARD
 	smartcard_trace_list_readers_return(smartcard, &ret, FALSE);
 	status = smartcard_pack_list_readers_return(smartcard, irp->output, &ret);
 
-	if (status != SCARD_S_SUCCESS)
-		return status;
-
 	if (mszReaders)
 		SCardFreeMemory(operation->hContext, mszReaders);
+
+	if (status != SCARD_S_SUCCESS)
+		return status;
 
 	return ret.ReturnCode;
 }
@@ -371,14 +371,13 @@ static UINT32 smartcard_ListReadersW_Call(SMARTCARD_DEVICE* smartcard, SMARTCARD
 	}
 
 	smartcard_trace_list_readers_return(smartcard, &ret, TRUE);
-
 	status = smartcard_pack_list_readers_return(smartcard, irp->output, &ret);
-
-	if (status != SCARD_S_SUCCESS)
-		return status;
 
 	if (mszReaders)
 		SCardFreeMemory(operation->hContext, mszReaders);
+
+	if (status != SCARD_S_SUCCESS)
+		return status;
 
 	return ret.ReturnCode;
 }


HenryJacques:
When you say that 
>This makes impossible to detect connection of a new smartcard reader when there are no other smartcard readers connected.

which version of FreeRDP are you referring to?

Using the latest master on my Ubuntu 14.04:
1. Start FreeRDP with not smartcard readers connected to a Windows 7 64bits
2. Connect a a smartcard
3. The remote server detects the card.

All of this without the proposed PR.

Cheers

kaliway:
I am using the latest master on Ubuntu 12.04 (pcscd 1.7.4) and Ubuntu 14.04 (pcscd 1.8.10). I connect to a Windows 2012 server. I start FreeRDP on my Linux conputer with no smartcard readers connected to it. When I say "no smartcard readers" I mean that there are no single reader with or without smartcard connected to my Linux box. When FreeRDP connects, I connect a reader to my Linux computer. Without the PR the remote server does not detect the connection of the reader. It does detect the connection of the reader with PR.

It is important to start without any single reader (even a reader without a smartcard inserted) connected to the client Linux machine because only in this case SCardListReaders will return SCARD_E_NO_READERS_AVAILABLE which will trigger the bug in smartcard_ListReaders(A/W)_Call.

There could be a number of reasons why you experienced a different behaviour of FreeRDP on your computer. For instance, maybe I was unclear that it is important to start without a signle smartcard READER connected and you started with an empty reader (a reader without a smartcard inserted) instead. You may have a virtual smartcard reader driver or you may have a serial smartcard reader configured in /etc/reader.conf.d/. In any case, output of "pcsc_scan" command would be useful to find out what is happening on your computer.

HenryJacques:
Yes, I am formal I start FreeRDP without any smartcard inserted or any readers. pcsc_scan shows nothing.
I am using two kinds of readers:
- a token (reader+card in one)
- a pinpad reader in which a insert a SIM card

both reader works well in Windows when inserted *after* I launch xfreerdp...

On the contrary, with your PR, I am getting additional errors which I didn't have without it:

```
[09:54:33:740] [3024:3031] [WARN][com.freerdp.channels.smartcard.client] - IRP failure: SCardIsValidContext (0x0009001C), status: SCARD_E_INVALID_HANDLE (0x80100003)
[09:54:33:799] [3024:3031] [WARN][com.freerdp.channels.smartcard.client] - IRP failure: SCardDisconnect (0x000900B8), status: SCARD_E_INVALID_HANDLE (0x80100003)
[09:54:33:800] [3024:3031] [ERROR][com.winpr.smartcard] - PCSC_ReleaseCardContext: null pContext!
[09:54:33:800] [3024:3031] [WARN][com.freerdp.channels.smartcard.client] - IRP failure: SCardReleaseContext (0x00090018), status: SCARD_E_INVALID_HANDLE (0x80100003)
[09:54:34:894] [3024:3067] [WARN][com.freerdp.channels.smartcard.client] - IRP failure: SCardStatusA (0x000900C8), status: SCARD_E_INVALID_VALUE (0x80100011)
```

bmiklautz:
@kaliway I just tested the following with the latest master:
  1. make sure *no* reader is connected
  1. start FreeRDP with smart card enabled
  1. start a test application in Windows - no reader is shown
  1. connect a reader - reader is detected
  1. insert a card into the reader - card can be read.

This worked without any problems. For testing I've used a very simple gemalto usb reader.
What and how did you test with on Windows?

bmiklautz:
@freerdp-bot test

bmiklautz:
Github or the build machines seem to have networking troubles. Re-testing it later.

HenryJacques:
@kaliway I've just tested to a Windows 2012 R2 and I think I have a clue...
Can you test it with the complete command? i.e. /smartcard:xxxx:xxxx and not just /smartcard ?
(even if the reader is not connected yet)

kaliway:
Sorry, my bad, looks like I forgot to mention one important thing in my initial test case. I was actually testing logon to a Windows 2012 server with a certificate stored on a smartcard, not just some smartcard application that is launched after successfull username/password login.

During logon, Windows tries to enumerate smart card readers by calling SCardListReaders. When no readers are connected to the client computer, SCardListReaders returns SCARD_E_NO_READERS_AVAILABLE which is normal behaviour of this function (see smartcard_operations.c:364). This return status should be passed back to the server, but unfortunately due to the condition in smartcard_operations.c:376 smartcard_ListReadersW_Call just simply returns without replying back to the server. From the server point of view SCardListReaders never completes which makes impossible for the server to detect insertion of a reader during logon.

Windows tries to call SCardListReaders several times in a row, and since none of the calls completes it gives up. With my PR Windows receives SCARD_E_NO_READERS_AVAILABLE and then calls SCardGetStatusChange with "\\\\?PnP?\\Notification" reader name to detect connection of new readers.

As for the errors in FreeRDP log, I don't think it is a FreeRDP issue. Depending on the situation it may be caused by the application that you use for testing. The reason they appeared may be the fact that with my PR the behavour of SCardListReaders (smartcard_ListReadersW_Call) is actually more accurate which in turn may reveal some minor issues in the application that you use for testing.

Thanks.

kaliway:
@HenryJacques I did the test with /smartcard:xxxx:xxxx 
I see the same bahaviour. Smartcard reader connection does not get detected by the server during logon.

HenryJacques:
@kaliway OK, I asked you because the behaviour is different depending on the target:
- to my windows 7, using /smartcard works, I see my readers (yes, with an application one logged)
- to my Windows 2012 R2, using just /smartcard is not working! I see no smartcard at all once logged.
@bmiklautz : is this a bug? Or using just /smartcard is simply not the way to use the parameter? See also my issue #2631 

However I won't be able to test your scenario as I don't have any workstation with smartcard logon...

kaliway:
@HenryJacques can you test my PR in your scenario with Windows 2012 R2?

HenryJacques:
@kaliway I've done a test but still get the same behaviour (no readers detected)... With Windows 2012 R2 I *have* to specify /smartcard:xxxx:xxxx

bmiklautz:
@freerdp-bot test.

bmiklautz:
@HenryJacques I must admit that I usually only use /smartcard without any parameters. But I also almost never disconnect the reader..

HenryJacques:
@bmiklautz I've done more tests:
* using just /smartcard works if I use xfreerdp with the classic /v: argument to a 2012 R2.
* using just /smartcard doesn't work if I use xfreerdp with a rdp file through a broker. the rdp file contains the redirectsmartcards:i:1 directive. However, in this configuration I have to specify (even a bogus) id with smartcard:xxxx:xxxx

PR #2801, opened 2015-07-23
winpr/thread: small cleanups

bmiklautz:
Also adds error checking of winpr_StartThread in winpr_StartThread.

### DIFF
--- a/winpr/libwinpr/thread/thread.c
+++ b/winpr/libwinpr/thread/thread.c
@@ -327,7 +327,7 @@ exit:
 		if (thread->detached || !thread->started)
 			cleanup_handle(thread);
 	}
-	pthread_exit((void*) (size_t) res);
+	return rc;
 }
 
 static BOOL winpr_StartThread(WINPR_THREAD *thread)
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test!

bmiklautz:
Hmm looks like the build farm has network troubles. Retrying later.

bmiklautz:
@freerdp-bot test

winpr/libwinpr/thread/thread.c
```diff
@@ -331,7 +328,6 @@ static void* thread_launcher(void* arg)
 			cleanup_handle(thread);
 	}
 	pthread_exit((void*) (size_t) res);
```
mfleisz: Shouldn't we just return res here? Not returning anything might cause compiler warnings and simply returning res should result in an implicit pthread_exit(res); call according to docs.

bmiklautz: @mfleisz excellent point. I'll update the PR. Thank you.

mfleisz:
+1

akallabeth:
+1

PR #2811, opened 2015-07-29
Fix race condition in freerdp_channels_client_load that could cause memory leak

clivest:
If g_pInterface is set to NULL outside the critical section in one thread while another thread is in the critical section, the channel client context allocated in the VirtualChannelEntry won't end up getting freed.

The channel client context is normally loaded from g_pInterface in FreeRDP_VirtualChannelInit and stored in the pChannelInitData, then copied from there onto the pChannelOpenData structure in FreeRDP_VirtualChannelOpen and finally freed in freerdp_channels_free.

### DIFF
--- a/libfreerdp/core/client.c
+++ b/libfreerdp/core/client.c
@@ -720,7 +720,6 @@ int freerdp_channels_client_load(rdpChannels* channels, rdpSettings* settings, P
 	EntryPoints.pVirtualChannelClose = FreeRDP_VirtualChannelClose;
 	EntryPoints.pVirtualChannelWrite = FreeRDP_VirtualChannelWrite;
 
-	g_pInterface = NULL;
 	EntryPoints.MagicNumber = FREERDP_CHANNEL_MAGIC_NUMBER;
 	EntryPoints.ppInterface = &g_pInterface;
 	EntryPoints.pExtendedData = data;
@@ -731,6 +730,7 @@ int freerdp_channels_client_load(rdpChannels* channels, rdpSettings* settings, P
 
 	EnterCriticalSection(&g_channels_lock);
 
+	g_pInterface = NULL;
 	g_ChannelInitData.channels = channels;
 	status = pChannelClientData->entry((PCHANNEL_ENTRY_POINTS) &EntryPoints);
 


bmiklautz:
@freerdp-bot test.

akallabeth:
@clivest Nice catch, but will have to look at this in more detail. What I specially find confusing is why is the context for each channel stored in the same variable? Should that not be a list with contexts for each channel loaded?

clivest:
Sure. Took me a while to figure out what was going on.

> why is the context for each channel stored in the same variable?

You mean the `g_pInterface` variable? The context for each channel is only stored in this variable temporarily. It's set in the channel's `VirtualChannelEntry` (called through the `pChannelClientData->entry` function pointer) via assignment to `*EntryPoints.ppInterface`. The channel context is then read from `g_pInterface` in `FreeRDP_VirtualChannelInit`, which is always called back in the scope of the  `VirtualChannelEntry` call, and stored on the channelInitData.

The `g_channels_lock` should be held around modifications to the `g_pInterface` and `g_ChannelInitData` global variables. As I see it, the bug here is that `g_pInterface` is modified outside the critical section. 

I think it's quite an odd design choice, however. I think it looks like the `client.c` code has been forced to fit the VirtualChannel interface, rather than designing the interface to fit the code (an alternative implementation, for example, could add an extra argument to `FreeRDP_VirtualChannelInit` to pass the channel context). I'm guessing this is to model the Windows RDP channel interface.

bmiklautz:
@clivest the client was more or less forced on the interface that's true but this was required in order to have the same  (or at least something that can simply be re-used or converted) Channel API as on windows. But you right it looks odd in that case ;)

bmiklautz:
@clivest generally +1 for your PR. The only thing I was wondering where you have observed multiple threads at this point? - If I'm not mistaken these parts are executed from the main thread when the command line is parsed

clivest:
> these parts are executed from the main thread when the command line is parsed

@bmiklautz Indeed. It looks like all of the current FreeRDP clients call `freerdp_channels_client_load()` from the main thread. I'm playing around with an application that instantiates multiple clients in the same process and ran into this problem when clients start simultaneously.

I think it looks like this area of the code base is intended to be thread-safe - there's a lock that's held around most modifications to the shared state - and could be made so with a straightforward change.

When creating the pull request, I didn't realise that all current clients only run this from the main thread. This pull request, while I think still valid, is less severe/pressing than I initially thought.

bmiklautz:
@akallabeth any objections?

PR #2818, opened 2015-07-31
Error: "Unable to detect time zone" - Add support for tz package

vancepym:
Fixes error message while freerdp connected the remote desktop:
"Unable to detect time zone"

libfreerdp always read timezone from symlink at /etc/localtime at the moment, but sometimes it also may the symlink at /etc/TZ

### DIFF
--- a/libfreerdp/locale/timezone.c
+++ b/libfreerdp/locale/timezone.c
@@ -1539,7 +1539,8 @@ char* freerdp_get_unix_timezone_identifier()
 	 * America/Montreal for example.
 	 */
 	
-	if ((len = readlink("/etc/localtime", buf, sizeof(buf) - 1)) != -1)
+	if ((len = readlink("/etc/localtime", buf, sizeof(buf) - 1)) != -1 ||
+	    (len = readlink("/etc/TZ", buf, sizeof(buf) - 1)) != -1)
 	{
 		int num = 0;
 		int pos = len;


vancepym:
This patch now also placed on the buildroot project at the master branch:
commit: http://git.buildroot.net/buildroot/commit/?id=f986cca8b1967095871065371e75b2498b2185cd

Thanks to Thomas Petazzoni <<thomas.petazzoni@free-electrons.com>>, he told me submit the patch to the official freerdp project.

bmiklautz:
@vancepym I might be mistaken but isn't the format of /etc/localtime and /etc/TZ different?

vancepym:
@bmiklautz Actually the same, but in some system (e.g. buildroot), the file could be /etc/TZ, not /etc/localtime. This patch just for compatible with that situation.

bmiklautz:
@vancepym I recently stumbled on the format used by ulibc for /etc/TZ and I thought that they were different than the "regular" format used.

bmiklautz:
@vancepym any update on my comment?

vancepym:
@bmiklautz YES, both have the same format, except the different file name. So, could you consider merging this patch?

bmiklautz:
@vancepym http://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html whereas /etc/timezone is something like 'Europe/Vienna'

bmiklautz:
@vancepym any comment on my last comment?

bmiklautz:
@vancepym my mistake I looked to /etc/timezone instead of /etc/localtime. If they have the same format this looks good.
+1

bmiklautz:
I'll also pick this to the current master.

PR #2820, opened 2015-07-31
Pdb file install

akallabeth:
* Adds ```pdb``` files to package target on windows (configurable with ```WITH_DEBUG_SYMBOLS```)
* Fixes a name conflict for static compiliation (```args``` arrays of ```client``` and ```makecert``` were global and not static)
* Fixes a naming conflict for ```freerdp-shadow```, binary now renamed on installation and not build time.


winpr/tools/makecert/makecert.c
```diff
@@ -61,7 +61,7 @@ struct _MAKECERT_CONTEXT
 	int duration_months;
 };
 
-COMMAND_LINE_ARGUMENT_A args[] =
+static COMMAND_LINE_ARGUMENT_A args[] =
```
bmiklautz: How is that related to the symbols?

libfreerdp/CMakeLists.txt
```diff
@@ -257,7 +257,7 @@ freerdp_module_add(${PRIMITIVES_SRCS})
 
 if(IPP_FOUND)
 	freerdp_include_directory_add(${IPP_INCLUDE_DIRS})
-    foreach(ipp_lib ${IPP_LIBRARIES})
+		foreach(ipp_lib ${IPP_LIBRARIES})
```
bmiklautz: intentionally or accidental indent?

libfreerdp/CMakeLists.txt
```diff
@@ -21,8 +21,8 @@ set(MODULE_PREFIX "FREERDP")
 # Create imported targets for Intel IPP libraries
 
 if(IPP_FOUND)
-    foreach(ipp_lib ${IPP_LIBRARIES})
-			add_library("${ipp_lib}_imported" STATIC IMPORTED)
+		foreach(ipp_lib ${IPP_LIBRARIES})
```
bmiklautz: Why the reformatting in this file?

channels/CMakeLists.txt
```diff
@@ -155,10 +155,6 @@ macro(add_channel_client_subsystem _channel_prefix _channel_name _subsystem _typ
 endmacro(add_channel_client_subsystem)
 
 macro(add_channel_client_library _module_prefix _module_name _channel_name _dynamic _entry)
-	if(${_dynamic} AND MSVC AND (NOT STATIC_CHANNELS))
```
bmiklautz: What was the reason to remove this? (same for the other two occurrences below).

CMakeCPack.cmake
```diff
@@ -60,7 +60,7 @@ set(CPACK_PACKAGE_ICON "${CMAKE_SOURCE_DIR}/resources\\\\FreeRDP_Install.bmp")
 set(CPACK_NSIS_MUI_ICON "${CMAKE_SOURCE_DIR}/resources\\\\FreeRDP_Icon_96px.ico")
 set(CPACK_NSIS_MUI_UNICON "${CMAKE_SOURCE_DIR}/resource\\\\FreeRDP_Icon_96px.ico")
 
-set(CPACK_COMPONENTS_ALL client server libraries headers)
+set(CPACK_COMPONENTS_ALL client server libraries headers symbols tools)
```
bmiklautz: Since the symbols are only defined on windows with MSVC we should maybe use the same if here as well.
I'm not sure if this can't cause problems otherwise.

winpr/tools/makecert/makecert.c
akallabeth: @bmiklautz If compiling as single binary there are 2 ```args``` arrays globally visible. One in makecert, one in ```client/common```. Making them static resolves that issue.

libfreerdp/CMakeLists.txt
akallabeth: Reformatting by IDE, have to fix that.

libfreerdp/CMakeLists.txt
akallabeth: Same as above.

channels/CMakeLists.txt
akallabeth: @bmiklautz Did prevent the build to succeed (the file is not generated / available), symbols of channels are exported using ```__declspec(dllexport)``` anyway.

CMakeLists.txt
```diff
@@ -287,6 +287,15 @@ if(WIN32)
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -D_CRT_SECURE_NO_WARNINGS")
 	set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DWIN32_LEAN_AND_MEAN")
 
+	set(CMAKE_USE_RELATIVE_PATH ON)
```
bmiklautz: I wouldn't generally change the cmake rpath behavior if not absolutely required. At least maybe this should only be done on windows?

akallabeth: @bmiklautz This is currently in a windows only section.

bmiklautz: @akallabeth ahh right. My bad.

bmiklautz:
+1

PR #2821, opened 2015-08-03
Fullscreen switching fallback for old Window Managers

erbth:
This queries the Window Manager if it supports the Extended Window Manager Hints properties for fullscreen switching and uses them only, if the Window Manager supports them. If it dosn't, the old way of fullscreen switching is done. In my case, Openbox did only use one monitor when switching FreeRDP to fullscreen mode, because it doesn't support _NET_WM_FULLSCREEN_MONITORS property.


bmiklautz:
@erbth did you test this on other window managers than openbox as well? - Fullscreen is always a little bit let's say "tricky" and causes a lot of headache.

bmiklautz:
@freerdp-bot test.

erbth:
@bmiklautz I've just tested with KDE's window manager and with i3. KDE supports _NET_WM_FULLSCREEN_MONITORS, but fullscreen switching doesn't really work well. But I've already had this issue before my changes, just haven't realized it.
I've definitely broken fullscreen on i3. Please give me some time to fix it. Should I also try to fix the issue with KWin?

bmiklautz:
@erbth great! I you got time, yes please. As far as I know a lot of people also use gnome/unity so before we can merge it we definitely need to test this as well.

erbth:
@bmiklautz OK, I'll do that. Thanks!

HenryJacques:
I also have 400 users with fluxbox. I'll try to test this.

erbth:
@HenryJacques OK, Great!

HenryJacques:
At first glance it seems ok with Fluxbox on Ubuntu 12.04 64bits.
But I don't have a multi-monitors environment so my test scenario isn't complete. 
By "fullscreen switching" do you mean going from one monitor to an extended dual-monitor display (and the other way)?

erbth:
@HenryJacques Thank You very much for testing!
With fullscreen switching I just mean switching from fullscreen to windowed mode and the other way round with ctrl + alt + enter. In my original case the problem was even worse, when starting freerdp with /f /multimon in multi-monitor environment, freerdp would only be displayed on one monitor, but the other monitor of the session being invisible. But one could have moved windows there, ...

erbth:
@bmiklautz "tricky" is pretty much the right word! I've decided to give the window manager full control over the window if it supports Enhanced Window Manager Hints. I've removed any move or resize operation, because some window managers (I think Unity/Compiz and KWin) showed odd behavior in some special cases. But I think this works well on all modern window managers. I think EWMH says, the window manager should take care to restore the windows old geometry when leaving fullscreen state. The only thing is that there are small differences where the window is placed when exiting fullscreen. What do you think?

I've tested with Gnome (Mutter), Gnome (Metacity), Unity (Compiz), Xfce (Xfwm4), KDE (KWin4), Openbox, Fluxbox and i3. Fullscreen works on all of them, but on i3 you have to use the window manager's fullscreen function if one would like to use mutple monitors in an RDP session. However, single monitor fullscreen works fine. It is a tilling window manager ... This does not work with Motif at all, and on Window Maker only single monitor fullscreen works. To make it work, I think much more changes would be required. Do you think it is ok if it doesn't work with those two wms?

@HenryJacques it would be really great if you had time to test it again!

HenryJacques:
@erbth  I've just tested your PR in my Ubuntu 14.04 64 bits and that definitely broke something:

windows

As you can see, I am connecting to a Windows 7 64 bits. xfreerdp is launched in fullscreen and you can see that the bottom is cropped. On the top, you can see the remains of the window manager toolbar.
If I switch to non-fullscreen and then immediately to fullscreen, the display of Windows' Desktop is correct. However this is a real issue for me as all xfreerdp session are launched in fullscreen without the toggle-fullscreen option.

nfedera:
@erbth please also verify that your changes don't break smart-sizing by running the qa tests:
https://github.com/FreeRDP/FreeRDP/wiki/Quality-Assurance-(QA)#smart-sizing

HenryJacques:
@erbth I complete my previous post:
I've tested it against a Windows 7 and a Windows 8.1
In both case, using /smart-sizing or not, the result is uncertain: sometimes the fullscreen is immediately correctly displayed, the second time it is as described in my previous post.
Fullscreen toggling gives me also strange result. I also take back what a said about toggling gives me a correct display, it is not. I don't have this behaviour with the master branch.

windows

erbth:
@HenryJacques Thank You very much for testing! I'll try to fix it.

erbth:
@nfedera Thanks, I'll test this!

HenryJacques:
@erbth it seems that the behaviour in my last post was already here without your PR in Master... :(
However with Fluxbox it seems I have a workaround:since I always launch xfreerdp in fullscreen, I can make sure that this specific app doesn't get a titlebar (see https://forums.gentoo.org/viewtopic-t-995292-view-previous.html?sid=164820fc082561e73f0e40f0f23fda17)
But this is only a workaround...

erbth:
@HenryJacques that helps a lot. Would be great if we could fix this bug here. I'm sorry, I haven't that much time at the moment, but I'll keep on working on this issue.

hardening:
That PR looked promising, anyway it requires rebasing now...

erbth:
I'm so sorry, I don't have that much time at the moment, because I'm still at school. But this is my last year and I'll (hopefully) finish school in about four month. Then I will finish this PR. If someone else could solve this fullscreen stuff in the mean time, I would be happy, but if not, I'll work on it again in about four month.

Detegr:
FWIW, I rebased this PR [in my fork](https://github.com/Detegr/FreeRDP). I tested it with i3; the fullscreen functionality works but the changes seem to break ctrl+alt+enter (or i3's fullscreen command) fullscreen toggling, at least with i3.

akallabeth:
Feature should be covered with

akallabeth:
Replaced by #4319

PR #2823, opened 2015-08-04
egfx channel open/close fix, H.264 Media Foundation decoder support

awakecoding:


channels/drdynvc/client/drdynvc_main.c
```diff
@@ -426,7 +430,10 @@ int dvcman_close_channel(IWTSVirtualChannelManager* pChannelMgr, UINT32 ChannelI
 
 	if (!channel)
 	{
-		WLog_ERR(TAG, "ChannelId %d not found!", ChannelId);
+		/**
+		 * Windows 8 / Windows Server 2012 send close requests for channels that failed to be created.
+		 * Do not warn, simply return success here.
+		 */
```
bmiklautz: Nice!

libfreerdp/codec/h264.c
```diff
@@ -59,6 +59,739 @@ static H264_CONTEXT_SUBSYSTEM g_Subsystem_dummy =
...
+
+	if (!sys->MFStartup || !sys->MFShutdown || !sys->MFCreateSample || !sys->MFCreateMemoryBuffer ||
+		!sys->MFCreateMediaType || !sys->MFCreateDXGIDeviceManager)
+		goto error;
+
+	CoInitializeEx(NULL, COINIT_APARTMENTTHREADED);
+
+	if (h264->Compressor)
+	{
+
```
bmiklautz: Is that required? If nothing is executed in the if clause why not using if (!h264->Compressor) and only this case?

libfreerdp/codec/h264.c
```diff
@@ -59,6 +59,739 @@ static H264_CONTEXT_SUBSYSTEM g_Subsystem_dummy =
...
+	h264->pSystemData = (void*) sys;
+
+	if (h264->Compressor)
+	{
+
+	}
+	else
+	{
+
+	}
```
bmiklautz: I'm not sure but do those empty if's have any special purpose?
I would suggest if it's something we currently don't handle but need to add in the future using
#if 0 and a comment would be more nice.

libfreerdp/core/rdp.c
```diff
@@ -34,7 +34,6 @@
 
 #define TAG FREERDP_TAG("core.rdp")
 
-#ifdef WITH_DEBUG_RDP
```
bmiklautz: I personally liked to have a way to disable this even if the WLOG_LEVEL is set to debug. 
Maybe we can extra DEBUG option for that.

bmiklautz:
@awakecoding looks very nice. Would be very nice if you could add some notes how this can be tested  and used that we can update the QA section in the wiki.

Issue #2760, opened 2015-07-02
Authentication Failure Does Not Set LastError to FREERDP_ERROR_AUTHENTICATION_FAILED


itsmikeeng:
Providing credentials with bad password sets context->LastError to FREERDP_ERROR_CONNECT_TRANSPORT_FAILED instead of expected FREERDP_ERROR_AUTHENTICATION_FAILED after call to freerdp_connect().

Using wfreerdp, the code reached the nla_client_begin() method in core\transport.c but succeeds and doesn't call freerdp_set_last_error(context, FREERDP_ERROR_AUTHENTICATION_FAILED);

giox069:
See also issue #2434 (and #2432).

bjcollins:
I encountered the same thing with xfreerdp when using NLA, after researching the spec I found that the server sends a TLS ALERT for ACCESS DENIED. It looks like the server code in the freerdp sources will set this alert, but there is no client side catching of the alert to set the error appropriately. I do not have the resources to work on wfreerdp, but I will try to post a pull request soon which addresses this same issue for xfreerdp.

bjcollins:
https://github.com/FreeRDP/FreeRDP/pull/2824

giox069:
@itsmikeeng: I think it's fixed with merged PR #2824.

akallabeth:
Can we close this?

giox069:
Yes, it works, thank you. You can close it.
There is a related issue to test: if you compile with -DWITH_SANITIZE_ADDRESS=on, you will notice that when an authentication error is correctly detected, the library does not correctly deallocate a lot of memory objects.

akallabeth:
@giox069 Will have a look at it. Is there already a bug entry for that? Anyway, closing this one as the deallocation stuff is a different matter.

giox069:
@akallabeth I never opened an issue for that, and never searched.

Issue #2434, opened 2015-03-01
Wrong error code when NLA auth fails


giox069:
From remmina issue https://github.com/FreeRDP/Remmina/issues/507

Trying to connect to a Windows 7 machine, using xfreerdp + NLA:
```
xfreerdp /sec:nla /v:mypc.remotedom.dom:21017 /u:jsmith /d:domain
```
When I type a **WRONG** password, freerdp sets the wrong error code to  FREERDP_ERROR_CONNECT_TRANSPORT_FAILED [0x2000D]
```
[20:17:03:515] [30711:1583720192] [ERROR][com.freerdp.core] - freerdp_set_last_error 0x2000D
[20:17:03:516] [30711:1583720192] [ERROR][com.freerdp.client.x11] - Authentication only, exit status 1

```

The last working xfreerdp is commit eddfee56a3ebd09d45ecffcfc2a58d7a3e712abd of Feb 15, which correctly calls freerdp_set_last_error(FREERDP_ERROR_AUTHENTICATION_FAILED)  [0x20009].
Here is the output when I type a wrong NLA password in the older freerdp.
```
[20:02:41:159] [29131:-501631232] [ERROR][com.freerdp.core.nla] - nla_recv() error: -1
[20:02:41:159] [29131:-501631232] [ERROR][com.freerdp.core] - freerdp_set_last_error 0x20009
[20:02:41:159] [29131:-501631232] [ERROR][com.freerdp.core.transport] - Authentication failure, check credentials.If credentials are valid, the NTLMSSP implementation may be to blame.
[20:02:41:159] [29131:-501631232] [ERROR][com.freerdp.core.connection] - Error: protocol security negotiation or connection failure
[20:02:41:160] [29131:-501631232] [ERROR][com.freerdp.client.x11] - Authentication only, exit status 1

```

lemonmojo:
+1
`connectErrorCode` is always set to `CONNECTERROR`.

lemonmojo:
Also, when using a Remote Desktop Gateway, there's currently no way of knowing whether the gateway or destination host credentials have been rejected (apart from looking at the log which isn't feasible in GUI-only applications).

giox069:
@lemonmojo: I'm closing this issue thanks to PR #2824 merged. For Remote Desktop Gateway, test it and reopen a new one if the problem is still here.

Issue #2432, opened 2015-03-01
Implementation of errors listed in error.h


tomas-korec:
Hi, I use FreeRDP with Remmina and I think this combination is much better than Windows RDP Client. So good job! :) However, there are som things I didn't like that much, so I decided to tell you about them, so FreeRDP and Remmina can improve.

I think that informing user about errors that happens can be improved. I have created this issue: https://github.com/FreeRDP/Remmina/issues/498 and giox069 (his comment: https://github.com/FreeRDP/Remmina/issues/498#issuecomment-75594933 ) discovered that FreeRDP returns the same error code for any type of error. Could you implement more error codes, so Remmina can inform user about type of error that happened?

Thank you!
Tomas

weberhofer:
see also #4555

akallabeth:
We're quite limited in finding out what exactly caused an error, but #4555 is a first step and has been integrated today.

akallabeth:
this has improved, closing.

PR #2824, opened 2015-08-04
Return FREERDP_ERROR_AUTHENTICATION_FAILED on an authentication failure

bjcollins:
when using NLA with xfreerdp.


akallabeth:
@bjcollins Would it not be easier just to use ```SetLastError(ERROR_MUTUAL_AUTH_FAILED)``` (or similar) instead of all the new functions?

bmiklautz:
@freerdp-bot test

bmiklautz:
@bjcollins any comment on @akallabeth suggestion?

bjcollins:
@bmiklautz @akallabeth I think this patch could cut down, SetLastError can probably just be called a little deeper without passing the failure boolean up the call stack and I called the callback "test_function" which is not really a great name. Not sure about using ERROR_MUTUAL_AUTH_FAILED as I wanted the exit code to reflect that it was an NLA auth failure. Either way, I want to look at this again and try to rework this patch.

bjcollins:
@akallabeth @bmiklautz  Ok, I reworked this. I wanted to make sure the xfreerdp process return code could be used to determine if process had an authentication failure - so I was unsure about trying to use ERROR_MUTUAL_AUTH_FAILED, but the PR is definitely in better shape than my first submission.

akallabeth:
@bjcollins Looks good so far, just a question, why don't you call ```freerdp_set_last_error(rdp->context, FREERDP_ERROR_AUTHENTICATION_FAILED)``` directly instead of using the new boolean value ```nlaFailure```?

bjcollins:
@akallabeth I think the variable might be useful in the future and this approach keeps setting of the error in the same location it already is. I think you can definitely just set call freerdp_set_last_error directly from the callback, but I like using a variable better. I will gladly remove the variable if you want.

akallabeth:
@bjcollins Would be nice, it usually is easier to find a bug if the error is set where it occured.

bjcollins:
@akallabeth Ok, removed the extra variable and just set the error in the callback.

giox069:
I really would like to see this pull request merged. Could please someone merge ?
It also should fix issue #2760 #2434 and helps for #2432

@bjcollins: in issue 2760 you mentioned "TLS ALERT for ACCESS DENIED". Is this the same problem reported [here](https://social.msdn.microsoft.com/Forums/es-ES/d59c3535-43be-45da-a458-19f4003c5294/mscssp-authentication-failure-notification-with-tls-alerts?forum=os_windowsprotocols) by Marc André ?

lemonmojo:
+1 definitely need better error handling in FreeRDP - this would be a first step.

antenore:
:+1: I also join @lemonmojo and @giox069 , this is a must have for several uses cases. 

Great job @bjcollins!

bjcollins:
@giox069 yes it is the same problem. He discovered the problem and put in the server-side code to send the alert. This adds in the code to detect that "Access Denied" alert on the client side and set an exit code accordingly. Sorry, that is not obvious in the code and I did not put in any comments.

I can probably improve the "transport_ssl_cb" function to make that more clear. I am currently looking at the raw numerical value, but there is an ssl function "SSL_alert_desc_string" which can decipher that value into a 2 character string value "AD". That change along with some comments should help make this PR more readable.

giox069:
@bjcollins: thanks for the explanation. I was just looking for the cause of the bad return code, and I came to the same conclusion. I asked to you just for a confirmation.

The only need I have now is... **please merge this PR**. I'm a Remmina developer/maintainer. I would like to update the "remmina-next Ubuntu PPA" where FreeRDP is very old (before issue #2434). But issues #2760 #2434 will eventually create issues in the PPA with remmina password saving, so I'm blocked on these issues and this PR.

Please, please, **merge**. Thank you :)

hardening:
ok ok I'm not so happy with the thing implemented in the OpenSSL callback, but let's merge it as is.

giox069:
Thank you very much :+1:  :+1: :+1:

antenore:
Thanks!

bjcollins:
@hardening Is your issue with the OpenSSL callback similar to what I said about about just using the raw value instead of using a helper function such as "SSL_alert_desc_string" to translate the code? Or is there more? I am just curious as this is something I would like to fix up a little better in the nearish future as I do not really like the callback code either.

PR #2825, opened 2015-08-05
Fixed color pointer decoding.

akallabeth:
The AND mask was wrongly applied for color pointer updates, inverting white pointers and making black ones invisible.

### DIFF
--- a/client/Android/FreeRDPCore/jni/android_freerdp.c
+++ b/client/Android/FreeRDPCore/jni/android_freerdp.c
@@ -1148,7 +1148,7 @@ static void copy_pixel_buffer(UINT8* dstBuf, UINT8* srcBuf, int x, int y, int wi
 	int length;
 	int scanline;
 	UINT8 *dstp, *srcp;
-	
+
 	length = width * bpp;
 	scanline = wBuf * bpp;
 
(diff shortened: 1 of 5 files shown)


bmiklautz:
@akallabeth I'm not convinced that this is really the proper fix for that.

nfedera:
@akallabeth I'm not convinced either. Was this tested against e.g. Windows 7 with different mouse pointer themes? Also this should be tested against Windows XP with different pointer themes.

nfedera:
Also check https://msdn.microsoft.com/en-us/library/windows/hardware/ff556138(v=vs.85).aspx

akallabeth:
@bmiklautz @nfedera Ok, I think I got the right solution now. Had a server and client bug. Server did send an invalid mask (all fields 0xFF) and client decoding routine expected the mask field to always be available.

bmiklautz:
@akallabeth make sense.

bmiklautz:
+1

mfleisz:
+1

hardening:
With my little knowledge of the subject I'd tend to +1. Anyway I'd really like to have @nfedera's opinion on this (our mouse pointer specialist !) before merging.

nfedera:
+1 LGTM

PR #2831, opened 2015-08-12
Fixed GDI_BITMAP, now using custom deallocation function.

akallabeth:
gdi_surface_bits: Now properly discarding old bitmap before creating a new one.


bmiklautz:
@akallabeth the CreateBitmap was rather modeled after https://msdn.microsoft.com/en-us/library/windows/desktop/dd183485(v=vs.85).aspx that's why there isn't a custom function.
When changing the function we basically getting less compatible to the drawing functions on Windows.

bmiklautz:
@akallabeth what was the initial problem that this change is necessary? - All graphic related memory should be allocated aligned anyways.

akallabeth:
@bmiklautz Yes, but that way it is not possible to use an existing bitmap as destination buffer for GDI

akallabeth:
@hardening @awakecoding @bmiklautz Any objections / suggestions on this one?

hardening:
I'm not a GDI specialist, but it looks good to me, +1.

bmiklautz:
+1. Nice idea. Get's much more flexible.

PR #2838, opened 2015-08-19
Added return value checks for smartcard functions.

akallabeth:
* ```smartcard_list_readers_return``` and ```smartcard_pack_list_readers_return``` did not honor error return values and accessed uninitialized memory (leading to crashes if no readers connected)


mfleisz:
+1

Issue #1414, opened 2013-08-12
Clipboard not in sync after program startup


akallabeth:
- Start xfreerdp
- Copy something to the clipboard
- Paste the copied content somewhere in xfreerdp
- Observe, that the clipboard in the xfreerdp window is empty

Note: If the clipboard data is available before program startup, the data is copied as expected.

Issue #2841, opened 2015-08-20
Null Character Appearing at the End of Copied Strings


Enverex:
A user is reporting that when they copy and paste text from a remote Windows box (to their local Linux client) it's not working correctly.

After some investigation it appears that for some reason when the text is copied from the Windows server and then pasted to the Linux client, a null character (%00) is appended to the end of the string (it's not there if they paste within the RDP session, only if they paste to the Linux client). It was hard to track this down as most programs and sites strip characters like this so it didn't affect everything.

The RDP server is Windows 2012 R2 and the client is Arch Linux (Kernel 4.1.5, FreeRDP 1.2.0-beta1).

ilammy:
Actually, sometimes it's not a single character, but a bunch of them (see #2287). I have no idea why Windows does this, but apparently UTF8_STRING format must not include terminating null characters. At least, according to [this draft spec](http://www.pps.univ-paris-diderot.fr/~jch/software/UTF8_STRING/UTF8_STRING.text) (we all know how _well_ X11 is standardized).

I am working on some clipboard improvements in X11 client. ilammy/FreeRDP@a3a72169f86a47474ccb8b146ccfad14dfe91111 should fix the issue with null characters.

Enverex:
Have you submitted that as a pull-request?

ilammy:
Not yet. I am going to submit this together with other clipboard-related changes from that branch so that they can be reviewed in one place.

Enverex:
This isn't happening as of current GIT so can now be closed as resolved.

Issue #2287, opened 2014-12-24
Clipboard Redirection Is Stripping All But One Character


reverendj1:
After playing around more with the latest version, it appears the fix for this bug https://github.com/FreeRDP/FreeRDP/issues/2209 has caused another issue. When copying from Linux to Windows, only the first character is copied.

Version:

```
xfreerdp --version
This is FreeRDP version 1.2.0-dev (git 3b332e7)
```

I use this script to install FreeRDP onto my Ubuntu 14.04 desktops (I've had the same results with several):

```
#!/bin/bash
apt-get install -y build-essential git-core cmake libssl-dev libx11-dev libxext-dev libxinerama-dev libxcursor-dev libxdamage-dev libxv-dev libxkbfile-dev libasound2-dev libcups2-dev libxml2 libxml2-dev libxrandr-dev libgstreamer0.10-dev libgstreamer-plugins-base0.10-dev libxi-dev libavutil-dev libavcodec-dev libcunit1-dev libdirectfb-dev xmlto doxygen libxtst-dev
git clone git://github.com/FreeRDP/FreeRDP.git /tmp/FreeRDP
cd /tmp/FreeRDP
cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_SSE2=ON .
make
make install
echo "/usr/local/lib/freerdp" > /etc/ld.so.conf.d/freerdp.conf
ldconfig
```

I then connect from my Ubuntu machine to a Windows machine (I've tried this on several Windows Server 2003 and 2008 servers), like so:

```
xfreerdp +clipboard /v:$MY_WINDOWS_SERVER /cert-ignore
```

Now, say I copy the text "Amazon" in Ubuntu. When I paste it into Windows, it shows up as "A". Copying from Windows to Ubuntu works fine though.

reverendj1:
@awakecoding I'm just curious, do you know when this might get merged with the master? Thanks again.

nfedera:
@awakecoding there is a new issue (similar to issue #2209) when copying from windows calculator to local (linux) libre office writer:
1.) in rdp session open calc, enter the number  12345 and ctrl+c
2.) on client side open libre office and do a ctrl-v and you get 12345########################...

akallabeth:
Copy paste of text is working on current master 3c7cc555a5a5f02366918be7eb935bb1f22115ef from linux to windows and vice versa.
Reopen if you still experience this bug.

Defenso-QTH:
[...]

PR #2851, opened 2015-08-27
Improve clipboard format support in X11 client

### LINKED ISSUE #1433: Copy / Paste issue with Outlook and HTML text
* Open Outlook on the server. (Should work with every application using HTML text, but not tested so far)
* Copy some HTML text on the client
* Paste the text to outlook 

The clipboard will not be synced properly, pasting will still work with some delay.
The format negotiation will be retried endless.

Copy / Paste of pictures and text works as expected.

ilammy:
Fixes in named formats (copy-paste HTML), raw format transfers (copy-paste anything between remote session), and some other things. See commit messages for more technical details.

Overview
-----------

### Named format support

As one may learn from MS-RDPECLIP spec, clipboard formats are identified either by _format ID_ (32-bit integer) or by _format name_ (arbitrary text string). Fixed-ID formats have fixed IDs assigned to them while named formats can have different IDs on different machines. The tricky part is in that cliprdr protocol is always using IDs to identify clipboard formats in protocol requests, so it is important to use _server-side_ IDs when requesting data from the server. XFreeRDP was doing it wrong, used client-side IDs instead, and expectedly was getting refusals from the server.

These commits revamp format identification and make sure that correct format IDs are used everywhere:

64d5e83 winpr/clipboard: add CF_RAW clipboard format ID
5b511a6 client/X11: improve clipboard format search functions
fed1907 client/X11: improve named clipboard format support

### Miscellaneous fixes

Some miscellaneous minor fixes:

36d3489 client/X11: correctly trim terminating null bytes from strings
f413325 client/X11: send clipboard format data errors correctly
eecd628 client/X11: remove unused function

### Raw transfer support

In the past XFreeRDP had a unique feature: ability to pass arbitrary clipboard formats between two XFreeRDP remote sessions, not limiting them to formats generally supported by X11 applications (basically, text, HTML, and some image formats). For example, this allowed to copy-paste formatted text between MS Office applications without losing any style information. Some parts of this feature still survived, but several other changes have broken it.

These commits restore the missing parts of the raw transfer mechanism:

842ea62 client/X11: refactor raw clipboard transfer indication
e4352bb client/X11: transfer raw clipboard format lists
81a584f client/X11: transfer raw clipboard format data

Generally, the idea stays the same:

  1. Session A receives a format list, assumes clipboard ownership, and advertises raw format list via a special mechanism.

  2. Session B is able to tell that an XFreeRDP session owns the clipboard, extracts raw format list and sends it to the remote side.

  3. Session B receives a format data request, and forwards the request to session A via a special mechanism.

  4. Session A passes the request to its remote side to get the actual data.

  5. The received data is passed intact from session A to session B, and then to the remote side of the session B.

The steps 1 and 5 were broken by previous changes and needed a fix.

Verification
------------

### Named formats

The only named format supported by XFreeRDP is `"HTML Format"`. It can be verified by copying anything from a remote browser into a local application supporting "text/html" (e.g, LibreOffice Writer). HTML should be kept intact: i.e., hyperlinks should be present, bold text should be pasted as bold, etc. Obviously, the same should be true in reverse.

Note that currently FreeRDP _does not remove_ `<!--StartFragment-->` and `<!--EndFragment-->` from HTML data, which results into them being shown as editorial comments by LibreOffice Writer. While additional metadata in comments is [a documented feature of `"HTML Format"`](https://msdn.microsoft.com/en-us/library/aa767917(v=vs.85).aspx), I am not sure that they should be removed by FreeRDP.

### Raw transfers

To check this case it is necessary to open two remote sessions and try transfering formats not supported by XFreeRDP. A good example would be `"Rich Text Format"`, a named format used by WordPad. The point is in that WordPad does not use `"HTML Format"` so it is not possible to confuse them. If raw transfer works correctly, rich formatting should be preserved when text is transferred between remote sessions.

To be absolutely sure that the feature works correctly one can build FreeRDP with enabled cliprdr dumps (`-DWITH_DEBUG_CLIPRDR=ON`) and observe communications over the _cliprdr_ channel.

### Affected known issues

This pull request fixes issues #1414, #2287, #2841.

It may fix #1433, but I am not able to verify it.

### DIFF
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -64,6 +64,8 @@ struct xf_clipboard
 	Atom clipboard_atom;
 	Atom property_atom;
 
+	Atom raw_transfer_atom;
+
 	int numClientFormats;
 	xfCliprdrFormat clientFormats[20];
 
@@ -120,6 +122,51 @@ static BOOL xf_cliprdr_is_self_owned(xfClipboard* clipboard)
 	xfContext* xfc = clipboard->xfc;
 
 	return XGetSelectionOwner(xfc->display, clipboard->clipboard_atom) == xfc->drawable;
+}
+
+static void xf_cliprdr_set_raw_transfer_enabled(xfClipboard* clipboard, BOOL enabled)
+{
+	UINT32 data = enabled;
+	xfContext* xfc = clipboard->xfc;
+
+	XChangeProperty(xfc->display, xfc->drawable, clipboard->raw_transfer_atom,
+			XA_INTEGER, 32, PropModeReplace, (BYTE*) &data, 1);
+}
+
+static BOOL xf_cliprdr_is_raw_transfer_available(xfClipboard* clipboard)
+{
+	Atom type;
+	int format;
+	int result = 0;
+	unsigned long length;
+	unsigned long bytes_left;
+	UINT32* data = NULL;
+	UINT32 is_enabled = 0;
+	Window owner = None;
+	xfContext* xfc = clipboard->xfc;
+
+	owner = XGetSelectionOwner(xfc->display, clipboard->clipboard_atom);
+
+	if (owner != None)
+	{
+		result = XGetWindowProperty(xfc->display, owner,
+			clipboard->raw_transfer_atom, 0, 4, 0, XA_INTEGER,
+			&type, &format, &length, &bytes_left, (BYTE**) &data);
+	}
+
+	if (data)
+	{
+		is_enabled = *data;
+		XFree(data);
+	}
+
+	if ((owner == None) || (owner == xfc->drawable))
+		return FALSE;
+
+	if (result != Success)
+		return FALSE;
+
+	return is_enabled ? TRUE : FALSE;
 }
 
 static BOOL xf_cliprdr_formats_equal(const CLIPRDR_FORMAT* server, const xfCliprdrFormat* client)
@@ -968,7 +1015,7 @@ static UINT xf_cliprdr_server_format_data_request(CliprdrClientContext* context,
 	xfClipboard* clipboard = (xfClipboard*) context->custom;
 	xfContext* xfc = clipboard->xfc;
 
-	if (xf_cliprdr_is_self_owned(clipboard))
+	if (xf_cliprdr_is_raw_transfer_available(clipboard))
 	{
 		format = xf_cliprdr_get_client_format_by_id(clipboard, CF_RAW);
 
@@ -1139,6 +1186,10 @@ xfClipboard* xf_clipboard_new(xfContext* xfc)
 	}
 
 	clipboard->property_atom = XInternAtom(xfc->display, "_FREERDP_CLIPRDR", FALSE);
+
+	clipboard->raw_transfer_atom = XInternAtom(xfc->display, "_FREERDP_CLIPRDR_RAW", FALSE);
+
+	xf_cliprdr_set_raw_transfer_enabled(clipboard, TRUE);
 
 	XSelectInput(xfc->display, clipboard->root_window, PropertyChangeMask);
 
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -828,24 +828,6 @@ UINT xf_cliprdr_send_client_format_list_response(xfClipboard* clipboard, BOOL st
  *
  * @return 0 on success, otherwise a Win32 error code
  */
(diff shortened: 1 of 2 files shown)


bmiklautz:
@freerdp-bot test

hardening:
@ilammy thank you for the contribution, could you rebase you PR on the master ? Thank you.

ilammy:
@hardening I have successfully rebased the changes onto current master, and retested them. Everything seems to be fine.

Resolved conflicts and changes in the original patch set:
- **client/X11: improve clipboard format search functions**
   + Resolved merge conflict around `clipboard->serverFormats` initialization in xf_cliprdr_server_format_list(). Removed the condition on `clipboard->numServerFormats` at the same place as it is always true now due to CF_RAW.
- **client/X11: refactor raw clipboard transfer indication**
   + @nfedera has  removed the old logic of xf_cliprdr_is_self_owned() and \_FREERDP_CLIPRDR_ID property altogether in a4580923e77b4690a5fe90a8e37a042bc5a8fa06. And he did it for a good reason: long time ago xf_cliprdr_is_self_owned() was used for raw transfers, but now it is used only to check whether _the current_ XFreeRDP window owns the clipboard. It should stay so. However, we do need a property to indicate raw transfer capability, so I have reintroduced it in the same way as it was implemented before, but it is now independent of xf_cliprdr_is_self_owned().
    + Updated commit message for this patch accordingly.
- **client/X11: transfer raw clipboard format lists**
   + Added more error logging with WLog_ERR as it is done in surrounding code by a4580923e77b4690a5fe90a8e37a042bc5a8fa06

akallabeth:
LGTM +1

bmiklautz:
@freerdp-bot test

client/X11/xf_cliprdr.c
```diff
@@ -190,14 +271,151 @@ static UINT xf_cliprdr_send_data_response(xfClipboard* clipboard, BYTE* data, in
...
+	}
+
+	if (Stream_GetRemainingLength(s) < sizeof(UINT32))
+	{
+		WLog_ERR(TAG, "too short serialized format list");
+		goto error;
+	}
+
+	Stream_Read_UINT32(s, *numFormats);
+
```
hardening: nitpicking, but perhaps we could check that this is a reasonable value...

client/X11/xf_cliprdr.c
```diff
@@ -190,14 +271,151 @@ static UINT xf_cliprdr_send_data_response(xfClipboard* clipboard, BYTE* data, in
...
+	for (i = 0; i < *numFormats; i++)
+	{
+		if (Stream_GetRemainingLength(s) < sizeof(UINT32))
+		{
+			WLog_ERR(TAG, "unexpected end of serialized format list");
+			goto error;
+		}
+
+		Stream_Read_UINT32(s, formats[i].formatId);
```
hardening: We should check that we don't go after the end of the stream.

hardening:
Except my remark LGTM. 
@ilammy correct this and I will merge it.

akallabeth:
@ilammy Could you fix the two remarks from @hardening? They pervent a merge of your very nice pull ;)

ilammy:
@akallabeth Sorry, I got carried away by... other things :zzz: 

I have pushed a new commit that improves on the issues pointed out by @hardening:
* the number of formats to be read is now checked for sanity (255 should be enough for anybody)
* format names are now read more carefully in order to avoid reading past the buffer's end

client/X11/xf_cliprdr.c
akallabeth: @hardening Isn't the check in line 352 sufficient?

hardening: @akallabeth we have to check at each loop turn as the name has a variable length.

client/X11/xf_cliprdr.c
```diff
@@ -190,14 +273,170 @@ static UINT xf_cliprdr_send_data_response(xfClipboard* clipboard, BYTE* data, in
...
+		formatNameLength = strnlen(formatName, Stream_GetRemainingLength(s));
+
+		if (formatNameLength == Stream_GetRemainingLength(s))
+		{
+			WLog_ERR(TAG, "missing terminating null byte, %zu bytes left to read", formatNameLength);
+			goto error;
+		}
+
+		formats[i].formatName = strndup(formatName, formatNameLength);
+		Stream_Seek(s, formatNameLength + 1);
```
hardening: very nice handling of the \0 in the stream, congrats !

PR #2855, opened 2015-08-28
Memleak fixes

akallabeth:
* Fixes memory leaks found
* Fixes some API misuse

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -408,7 +408,15 @@ static UINT handle_hotplug(rdpdrPlugin* rdpdr)
 			/* copy hotpluged device mount point to the dev_array */
 			if (strstr(word, "/mnt/") != NULL || strstr(word, "/media/") != NULL)
 			{
-				dev_array[size].path = strdup(word);
+				dev_array[size].path = _strdup(word);
+				if (!dev_array[size].path)
+				{
+					fclose(f);
+					free(line);
+					error = CHANNEL_RC_NO_MEMORY;
+					goto cleanup;
+				}
+
 				dev_array[size++].to_add = TRUE;
 			}
 			free(word);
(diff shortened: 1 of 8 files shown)


channels/rdpdr/client/rdpdr_main.c
```diff
@@ -495,14 +500,22 @@ static UINT handle_hotplug(rdpdrPlugin* rdpdr)
...
+	for (i = 0; i < size; i++)
+		free (dev_array[size].path);
+
 	return rdpdr_send_device_list_announce_request(rdpdr, TRUE);
+
+cleanup:
+	for (i = 0; i < size; i++)
+		free (dev_array[size].path);
+
+	return error;
```
MartinHaimberger: would nt it be better to remove the duplicated code above and use something like:

return error ? error : rdpdr_send_device_list_announce_request(rdpdr, TRUE);

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -468,17 +468,21 @@ static UINT handle_hotplug(rdpdrPlugin* rdpdr)
...
 
 			drive->Type = RDPDR_DTYP_FILESYSTEM;
 
-			drive->Path = _strdup(dev_array[i].path);
+			drive->Path = dev_array[i].path;
+			dev_array[i].path = NULL;
+
 			if (!drive->Path)
 			{
 				WLog_ERR(TAG, "_strdup failed!");
```
MartinHaimberger: wrong comment because of removed _strdup

akallabeth: @MartinHaimberger No, the ```_strdup``` just duplicated a string cleaned up later on. Setting the source to NULL prevents that and so no duplication is necessary.

bmiklautz:
+1

hardening:
+1

PR #2861, opened 2015-09-02
rdpsnd/winmm: fix SetVolume when device is not opened.

llyzs:
The client may receive SetVolume PDU before receiving the first audio sample (which triggers the opening of the audio device in winmm plugin). This caused the volume PDU be lost and after some recent changes, now causes the client to be closed unexpectedly. Thanks to the error checking changes so that we can detect this problem.

### DIFF
--- a/channels/rdpsnd/client/winmm/rdpsnd_winmm.c
+++ b/channels/rdpsnd/client/winmm/rdpsnd_winmm.c
@@ -50,6 +50,7 @@ struct rdpsnd_winmm_plugin
 	WAVEFORMATEX format;
 	int wformat;
 	int block_size;
+	UINT32 volume;
 	FREERDP_DSP_CONTEXT* dsp_context;
 };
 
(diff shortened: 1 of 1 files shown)


llyzs:
@akallabeth @mfleisz

akallabeth:
+1

channels/rdpsnd/client/winmm/rdpsnd_winmm.c
```diff
@@ -375,6 +386,8 @@ UINT freerdp_rdpsnd_client_subsystem_entry(PFREERDP_RDPSND_DEVICE_ENTRY_POINTS p
 		return CHANNEL_RC_NO_MEMORY;
 	}
 
+	winmm->volume = 0xFFFFFFFF;
+
```
hardening: Just wondering if it's a good idea to start at full volume...

llyzs: @hardening 

This is the application volume, not master volume. In fact OS default of the application volume is 100% so this should not be a problem.

hardening: Nice then, thank you for the answer.

PR #2864, opened 2015-09-02
Give unique aliases for smartcards with the same name, so that two re…

bjcollins:
…aders do not end up with the same alias.


bjcollins:
Without this patch, two identical readers will end up with same alias and then attempts to talk to the 2 readers will end up talking to just one of the two readers and basically makes it impossible to use either reader.

mfleisz:
I think this one looks ok +1

bmiklautz:
@freerdp-bot test

bmiklautz:
Looks also good to me.
+1

bmiklautz:
@freerdp-bot add to whitelist

PR #2865, opened 2015-09-03
[H264] ignore decoder errors

akallabeth:
Microsoft Windows 8.1 server sometimes sends broken H264 data over GFX.
To avoid disconnections ignore H264 decoding errors.
In case a decoder error occurs, the screen will be distorted due to missing image information,
but better that than no connection :/

Cleaned up GFX code a bit too.

### DIFF
--- a/channels/rdpgfx/client/rdpgfx_codec.c
+++ b/channels/rdpgfx/client/rdpgfx_codec.c
@@ -38,52 +38,12 @@
  *
  * @return 0 on success, otherwise a Win32 error code
  */
-UINT rdpgfx_decode_uncompressed(RDPGFX_PLUGIN* gfx, RDPGFX_SURFACE_COMMAND* cmd)
-{
-	return CHANNEL_RC_OK;
-}
-
-/**
- * Function description
- *
- * @return 0 on success, otherwise a Win32 error code
- */
-UINT rdpgfx_decode_remotefx(RDPGFX_PLUGIN* gfx, RDPGFX_SURFACE_COMMAND* cmd)
-{
-	return CHANNEL_RC_OK;
-}
-
-/**
- * Function description
- *
- * @return 0 on success, otherwise a Win32 error code
- */
-UINT rdpgfx_decode_clear(RDPGFX_PLUGIN* gfx, RDPGFX_SURFACE_COMMAND* cmd)
-{
-	return CHANNEL_RC_OK;
-}
-
-/**
- * Function description
- *
- * @return 0 on success, otherwise a Win32 error code
- */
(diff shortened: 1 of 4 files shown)


nfedera:
I can confirm that Windows 8.1 sends invalid H.264 data.
Currently the only Remote Desktop Client in the world (in addition to FreeRDP) that supports H.264 out of the box is the mstsc on the [Microsoft Surface 2](https://de.wikipedia.org/wiki/Microsoft_Surface#Surface_2) (Nvidia Tegra 4, Windows RT).
If I use the Surface 2 to connect with 640x480 to Windows 8.1 Professional I also get horrible graphical decoding errors but the client does not disconnect.
I've also tested the recorded h264 dump with the latest [Fraunhofer H.264/AVC reference software](http://iphome.hhi.de/suehring/tml/)
```
----------------------------- JM 19.0 (FRExt) -----------------------------
--------------------------------------------------------------------------
 Input H.264 bitstream                  : test.264
 Output decoded YUV                     : test_dec.yuv
 Input reference file                   : test_rec.yuv
--------------------------------------------------------------------------
 Input reference file                   : test_rec.yuv does not exist
                                          SNR values are not available
Profile IDC  : 66
Image Format : 640x480 (640x480)
Color Format : 4:2:0 (8:8:8)
--------------------------------------------------------------------------
POC must = frame# or field# for SNRs to be correct
--------------------------------------------------------------------------
  Frame          POC  Pic#   QP    SnrY     SnrU     SnrV   Y:U:V Time(ms)
--------------------------------------------------------------------------
invalid 16x16 intra pred Mode VERT_PRED_16
```

So +1 for this PR

alantm2:
nfedera: Any idea if the 'Surface 2 -- If I use the Surface 2 to connect with 640x480 to Windows 8.1 Professional I also get horrible graphical decoding errors but the client does not disconnect.' BUG (a) has been reported to Microsoft and (b), a hotfix is in available or being developed, etc. and/or (c) are there any references you could point to? If this is the case, we need the backend components patched by MSFT ASAP to be able to make use of H.264? Yes/no, ideas/thoughts to get this sorted and/or work around it?

nfedera:
@alantm2 (a) no (b) no (c) no - sorry.
The main issue here is that for whatever reason Microsoft does not seem to want that their RDP Servers send H.264 to any of their x86 based clients.
One reference I've found is here: 
https://social.technet.microsoft.com/Forums/windowsserver/en-US/a9fd61fb-0dbf-41f2-be59-9e5c76e6a3a2/rdp-incorectly-using-2008r2-compatible-mode-vdi-on-2012-r2-81-enterpise?forum=winserverTS#60919f2f-03cc-4509-9fdd-beaee83dab9b

So basically H.264/RDP is currently completely unused and bugs regarding their encoder will only be noticed on non-Microsoft devices/clients (except the Surface 2)

Issue #2369, opened 2015-02-08
Mouse pointer incorrect transformed


antonstv:
FreeRDP version 1.2.1-dev (git ed3d952)
VMWare VSphre Client in console guest OS Windows XP cursor disappears (invisible), in putty mouse cursor black and white to be, in Danware Netop Guest cursor disappears. In version 1.0.2 is correctly, but /multimon not available.

В Vsphere client и Netop курсор исчезает, в putty курсор становится черным на черном фоне, а должен быть белым.  В версии 1.0.2 корректно, но нет опции нескольких мониторов.

bmiklautz:
@antonstv can you give us more details what client you use or how you connect?

antonstv:
Ubuntu 10.04 LTS, GPU Intel® Ivybridge Desktop 
FreeRDP 1.2.1 get from https://github.com/FreeRDP/FreeRDP.git
compile: cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_SSE2=ON
connect string(xterminal):  xfreerdp /multimon /u:user033@office.local /v:192.168.103.33

Connecting RDP to Win7 Enterprise.

antonstv:
whith /bpp:32 in VSphre Client and Netop cursor is correctly. In putty coursor is black. In version 1.0.2 cursor dotted white.

nfedera:
I'm pretty sure you're experiencing the issue that inverted color cursors aren't supported under X11.
See https://msdn.microsoft.com/en-us/library/windows/hardware/ff556143%28v=vs.85%29.aspx

In case of putty this is what it should look like:
![image](https://cloud.githubusercontent.com/assets/670606/6102950/7dd8a232-b03e-11e4-8296-c162ad4a41bc.png)

If the AND mask and XOR mask values are both 1 we set that pixel to black since inversion is not supported under X11 (AFAIK).

See https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/codec/color.c#L1505
```
else if (andPixel && xorPixel)
    *pDstPixel++ = 0xFF000000; /* inverted (set as black) */
```

The most simple soulution would be setting white and black alternately, something like this:
```
int bw=0;
...
else if (andPixel && xorPixel)
    *pDstPixel++ = bw++ % 2 ? 0xFF000000 : 0xFFFFFFFF;
```

antonstv:
Thanks

bmiklautz:
[...]

PR #2868, opened 2015-09-03
improve handling of inverted pointer colors

nfedera:
Inverted pointer colors (where individual pixels can change their color to accommodate the background  behind them) only seem to be supported on Windows.
Using a static replacement color for these pixels (e.g. black) might result in invisible pointers depending on the background.
This PR replaces inverted pixels with either black or white, depending on the pixel position.

image


nfedera:
this should fix issue #2369

bmiklautz:
@nfedera nice idea!
+1

akallabeth:
@nfedera Nicely done, +1

PR #2869, opened 2015-09-03
X11 /app: Fix drawing with negative window origin

peterh:
Since REGION16 uses unsigned values, when `appWindow->x` or `appWindow->y`
is negative, the region will have a very large left or top value.

Avoid this problem by clamping to 0 before casting to an unsigned value.


hardening:
+1

bmiklautz:
@freerdp-bot test

bmiklautz:
+1

PR #2873, opened 2015-09-07
makecert: Remove key usage attribute

mfleisz:
The original makecert tool won't add this tribute to a generated
certificate. Adding this attribute might cause problems when using the
certificate with other apps (i.e. Qt SSL).


akallabeth:
+1

hardening:
Nothing against that PR.

PR #2877, opened 2015-09-09
winpr/file refactoring and regular file support

bmiklautz:
* refactor handle creator
* split out client side named pipe handling from generic file handling
* add initial support for regular files.

### DIFF
--- a/winpr/include/winpr/file.h
+++ b/winpr/include/winpr/file.h
@@ -161,6 +161,10 @@
 #define FIND_FIRST_EX_CASE_SENSITIVE		0x1
 #define FIND_FIRST_EX_LARGE_FETCH		0x2
 
+#define STD_INPUT_HANDLE  (DWORD)-10
+#define STD_OUTPUT_HANDLE (DWORD)-11
+#define STD_ERROR_HANDLE  (DWORD)-12
+
 typedef union _FILE_SEGMENT_ELEMENT
 {
 	PVOID64 Buffer;
@@ -316,7 +320,6 @@ WINPR_API BOOL RemoveDirectoryW(LPCWSTR lpPathName);
 #define RemoveDirectory		RemoveDirectoryA
 #endif
 
-
 /* Extra Functions */
 
 typedef BOOL (*pcIsFileHandled)(LPCSTR lpFileName);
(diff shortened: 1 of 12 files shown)


winpr/libwinpr/file/generic.c
```diff
@@ -0,0 +1,613 @@
...
+}
+
+BOOL CreateDirectoryW(LPCWSTR lpPathName, LPSECURITY_ATTRIBUTES lpSecurityAttributes)
+{
+	return FALSE;
+}
+
+BOOL RemoveDirectoryA(LPCSTR lpPathName)
+{
+	if (!rmdir(lpPathName))
```
hardening: return rmdir(lpPathName) == 0 ?

hardening:
I'm not a WIN32 specialist but AFAICT it LGTM

winpr/libwinpr/file/generic.c
bmiklautz: @hardening this was just split moved out from the old file.c I didn't review the code. But your suggestion makes sense.

bmiklautz:
@hardening the one thing need to test on Windows is what happens if you close the handle for stdin/stderr/stdout. And we CreateFile is also missing but something for another pull request then.

mfleisz:
If you close one of the std* handles on Windows then the file descriptor will simply be closed and rendered invalid. You would have to set a new handle (i.e. using SetStdHandle) to reenable the descriptor.
See also http://blogs.msdn.com/b/oldnewthing/archive/2013/03/07/10399690.aspx

mfleisz:
(Can't edit with stupid github app)
So I think the behavior will be the same as on the other platforms right?

bmiklautz:
@mfleisz thanks for pointing that out. On Linux/UNIX closing the file descriptor handles it invalid but the next open call can be re-use the number. For example if a close(2) is done the fd returned by the next open (or similar) might be 2.

mfleisz:
@bmiklautz So on Linux - if you close stdin (usually value 0 iirc) and you open a file then it will get a descriptor id of 0 --> so stdin automatically (without setting it directly or via an API) points to that file? That behavior is different from Windows then ... Windows won't set the stdin handle automatically. the std handles are just global vars that can only be accessed via the Set/GetStdHandle APIs - and if you close that handle it will also make that global variable (used by all the IO functions) useless.

winpr/include/winpr/file.h
```diff
@@ -329,7 +332,9 @@ typedef struct _HANDLE_CREATOR
 	pcCreateFileA CreateFileA;
 } HANDLE_CREATOR, *PHANDLE_CREATOR, *LPHANDLE_CREATOR;
 
-BOOL RegisterHandleCreator(PHANDLE_CREATOR pHandleCreator);
+WINPR_API HANDLE GetStdHandle(DWORD nStdHandle);
+WINPR_API BOOL SetStdHandle(DWORD nStdHandle, HANDLE hHandle);
+WINPR_API BOOL SetStdHandleEx(DWORD dwStdHandle, HANDLE hNewHandle, HANDLE* phOldHandle);
```
MartinHaimberger: this exports should be in the extern "C" { } section ....

bmiklautz:
@mfleisz correct. The standard handle are only the first file handles the process has thats why you usually try to keep them "occupied". What we can do is if the handle is closed to redirect the corresponding fd to /dev/null and if it is set again we re-open it. I think that would come the closest to the behavior on windows. 
I'll create a new pull request for that then.
Thank for your hints!

mfleisz:
@bmiklautz Sounds good :+1:

MartinHaimberger:
+1 LGTM

PR #2879, opened 2015-09-10
xfreerdp: Fix window shape

peterh:
Don't abort the entire xf_rail_window_common function when the window is
already in the correct location.

To reproduce:
 - move an application off the edge of the screen
 - resize the window to cause a shape to be set
 - move the application window fully on the screen
 - resize the application window larger
 - note lack of drawing in newly enlarged portion of window

Bug introduced in abf6d4f71ea901e31ee2d1b4625fa805af6b0cc9 "xfreerdp:
prepare RAIL migration away from libfreerdp-rail" when
xf_rail_MoveWindow was copy-and-pasted into xf_rail_window_common
without noticing that the "return" would omit the rest of the combined
function, not just the portion that was pasted.


hardening:
not an expert but LGTM

bmiklautz:
@freerdp-bot test

akallabeth:
LGTM

PR #2880, opened 2015-09-11
Fixes for OpenBSD port

bceverly:
I had to make two small changes to get the code building properly on OpenBSD again.  If someone can please review and commit them, I would appreciate it.

Thanks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -381,6 +381,7 @@ endif(APPLE)
 
 # OpenBSD
 if(OPENBSD)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -L/usr/local/lib")
 	set(WITH_MANPAGES "ON")
 	set(WITH_ALSA "OFF")
 	set(WITH_PULSE "ON")
--- a/channels/tsmf/client/oss/tsmf_oss.c
+++ b/channels/tsmf/client/oss/tsmf_oss.c
@@ -34,7 +34,11 @@
 #include <libgen.h>
 #include <limits.h>
 #include <unistd.h>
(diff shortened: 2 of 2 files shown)


hardening:
@freerdp-bot you'd be my best friend if you test that PR

channels/tsmf/client/oss/tsmf_oss.c
```diff
@@ -34,7 +34,11 @@
 #include <libgen.h>
 #include <limits.h>
 #include <unistd.h>
-#include <sys/soundcard.h>
+#if defined(__OpenBSD__)
```
akallabeth: This should be solved by including ```OSS_INCLUDE_DIR``` in CMake, then ```soundcard.h``` is in the include path.
See ```cmake/FindOSS.cmake```

CMakeLists.txt
```diff
@@ -381,6 +381,7 @@ endif(APPLE)
 
 # OpenBSD
 if(OPENBSD)
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -L/usr/local/lib")
```
akallabeth: If you have linking issues this may indicate that the libraries detected by CMake have not properly been added. Overwriting ```C_FLAGS``` is not the way to go.

hardening:
+1

hardening:
Sorry I have merged too fast

hardening:
@bceverly could you take in account @akallabeth's remarks and submit another PR ?

bceverly:
Hi - I think I did with this latest PR.  Please take a look and let me know.

Thanks!


Thanks,
Bryan

PR #2883, opened 2015-09-11
Updated OpenBSD changes

bceverly:
Took into account feedback on previous pull request


akallabeth:
@bceverly You should have a look at the ```FindXXX.cmake``` files and fix the search paths there. This way it should not be necessary to add a global link directory. As for ```execinfo``` libraries should be linked by the library or executable actually using them and not globally. Still missing the revert for the ```#ifdef``` that is not necessary if properly adding the include path for the libraries detected.

bceverly:
I believe my latest push addresses everything.  Turns out we didn't need
the library on OpenBSD so I made its inclusion for FreeBSD only.


Thanks,
Bryan

bmiklautz:
@bceverly if possible try not to have merges within your pull request.

bmiklautz:
@freerdp-bot test

akallabeth:
Ok, looks good now ;)

hardening:
LGTM too, so I'm merging it too fast like the last time ;)

PR #2887, opened 2015-09-15
winpr/file: extend API

MartinHaimberger:
Add function GetFileHandleForFileDescriptor to get
a file handle for a valid file descriptor.

### DIFF
--- a/winpr/libwinpr/file/file.c
+++ b/winpr/libwinpr/file/file.c
@@ -33,6 +33,7 @@
 
 #include "../handle/handle.h"
 #include <errno.h>
+#include <fcntl.h>
 
 struct winpr_file
 {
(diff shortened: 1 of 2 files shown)


winpr/libwinpr/file/file.c
```diff
@@ -215,13 +216,34 @@ HANDLE GetStdHandle(DWORD nStdHandle)
...
 }
 
 #endif /* _WIN32 */
 
+/* Extended API */
+
+HANDLE GetFileHandleForFileDescriptor(int fd)
+{
+#ifdef WIN32
+    return INVALID_HANDLE_VALUE;
```
mfleisz: I'd call _get_osfhandle (https://msdn.microsoft.com/en-us/library/ks2530z6.aspx) here

bmiklautz: @mfleisz is that still available on every windows version? - I wasn't sure.

mfleisz: @bmiklautz Since Windows 95: https://msdn.microsoft.com/en-us/library/ks2530z6(v=vs.80).aspx

mfleisz:
LGTM +1

bmiklautz:
+1 for the nice changes on windows. @mfleisz good idea.
@MartinHaimberger thx.

PR #2888, opened 2015-09-15
fix ARM64 data type portability

awakecoding:
Fix size of data sizes on 64-bit ARM


awakecoding:
@freerdp-bot test.

hardening:
LGTM

PR #2889, opened 2015-09-15
Fix RDP8 gateway disconnections

dvincent-devolutions:


awakecoding:
@freerdp-bot test.

awakecoding:
@freerdp-bot test

awakecoding:
the build bot failed for some reason for Win64, retesting

bmiklautz:
@freerdp-bot test

bmiklautz:
Branch/tag couldn't be fetched from github.

mfleisz:
Most of the formatting looks broken (spaces vs tabs)

libfreerdp/crypto/tls.c
```diff
@@ -159,6 +159,11 @@ static int bio_rdp_tls_read(BIO* bio, char* buf, int size)
 				break;
 
 			case SSL_ERROR_SYSCALL:
+                if (WSAGetLastError() == WSAEWOULDBLOCK)
+                {
+                    BIO_set_flags(bio, BIO_FLAGS_READ | BIO_FLAGS_SHOULD_RETRY);
+                    break;
+                }
```
bmiklautz: Didn't we have this a while back and removed it again because it broke stuff (tm)?

mfleisz: Indeed this looks familiar ... @llyzs @hardening what do you say?

awakecoding: That's the part of this pull request which is most subject to discussion. I haven't debugged this myself, but I did look over the problem with Denis, and it turns out that sometimes SSL_read returns with SSL_ERROR_SYSCALL and errno set to EAGAIN/EWOULDBLOCK. We haven't been able to pinpoint the reason of this SSL_ERROR_SYSCALL, and it everything looks good in the RDP8 gateway BIO. I thought if we limit this to just WSAEWOULDBLOCK it wouldn't cause the previous problems that were caused AFAIK by ignoring all SSL_ERROR_SYSCALL, including cases where it is a real error (WSAWOULDBLOCK isn't really an error, just don't know why it ends up there though).

awakecoding:
@mfleisz I will get the formatting fixed, you are correct.

libfreerdp/crypto/tls.c
```diff
@@ -159,6 +159,11 @@ static int bio_rdp_tls_read(BIO* bio, char* buf, int size)
 				break;
 
 			case SSL_ERROR_SYSCALL:
+				if (WSAGetLastError() == WSAEWOULDBLOCK)
```
hardening: We have seen many patches here, they were all hiding real problems.
Reading errno is not reliable in this case.

awakecoding: @hardening I'm jumping on it to find the root cause of this. I'm building against my own openssl build now so I can add printfs in it and find the place in OpenSSL that returns the SSL_ERROR_SYSCALL.

mfleisz: I think the last time we had problems in this area was caused by accessing the SSL BIOs from different threads iirc... maybe that leads you somewhere ...

awakecoding:
I thought maybe the problem was the RDG BIO, but it turns out the problem also affects the TSG BIO. I really can't figure out what's wrong with the BIO, even looking at OpenSSL code. The problem occurs more quickly with OS X for some reason with the stock OpenSSL (0.9.8gz). I have tried figuring out if what we're doing is somehow correct for full duplex usage, and I couldn't find confirmation that we can safely call SSL_read/SSL_write in parallel. This is extremely annoying, as it works well 99.9% of the time with all of the effort we've put in. This stack overflow question affirms that at least with < 1.0.0 we can't do that safely: http://stackoverflow.com/questions/14467630/clarifying-openssl-0-9-8l-concurrency-support-can-ssl-instances-be-used-by-mul. I couldn't find confirmation that it should be safe in any way for later versions. I removed the false fix and added locking around operations that deal with SSL*. I need more extensive testing, but it appears to be stable now. I hate to have to lock SSL*, but I don't see much more options to make a definitive, stable fix. Any thoughts?

awakecoding:
OpenSSL has been making our lives miserable for so long. I found out that a few weeks ago, mbed-tls was relicensed under the Apache license, which would make it usable for us (https://tls.mbed.org). It is apparently the result of ARM buying the rights on PolarSSL and offering it now under Apache 2.0. Building is simple (cmake is supported), and it appears very portable. I would seriously consider at least adding support for mbed-tls, this would alleviate a lot of pain that we currently encounter with OpenSSL.

hardening:
IIRC on MacOS async transport is activated by default, that could explain why it happens more frequently.

bmiklautz:
@awakecoding mbed-tls might be an good alternative but I wouldn't drop openssl completely. Also using a not that wide spread ssl toolkit might make packaging/distribution even harder.

hardening:
About mbed-tls, they don't seem to support reading and writing at the same time. This is quite explicit in the doc: a rule says that until the doc says it's safe you shouldn't do it, and the doc almost never talk of this). Perhaps dropping concurrent read / write would remove us some pain too !

awakecoding:
@bmiklautz @hardening we have tested the new more proper fix that involves locking operations dealing with SSL*, and it appears to be stable now. Aside from the gateway, we have had many user reports similar failures with direct connections in the recent months. I'm thinking we really don't have much of a choice but to lock those operations. Are we good to merge? The current code is definitely not safe, and this new fix is not going to cause another cat and mouse chase between false fixes and true fixes.

awakecoding:
@hardening as for mbed-tls, I think it's at least worth a shot to attempt adding optional support for it. The building procedure would likely involve preparing an mbed-tls build instead of using distribution packages, but it would be statically built in. As for full duplex, I didn't have the time to take a look, but if OpenSSL doesn't do it then it's not really a downside at this point.

bmiklautz:
I haven't tested the changes in rdg but the changes in tls.c look good to me.

awakecoding:
@bmiklautz we've tested the RDG changes, and it definitely resolves the issues we were having.

PR #2897, opened 2015-09-18
Minor rail fixes/tweaks

bjcollins:
1. fix window title for rail as it was not being set on initial window creation
2. remove duplicate call to XStoreName when setting window title
3. expand WITH_XEXT #define for rail window rects as extra unecessary work was being done when WITH_XEXT was not defined


hardening:
@freerdp-bot it looks good, but could you test a compilation of this, Thank you

Issue #2648, opened 2015-05-25
errors like SCARD_E_INSUFFICIENT_BUFFER with smartcard redirection


hendwolt:
I'm using:
This is FreeRDP version 1.2.4-dev (git 36cb1e6)

and this card reader:
kernel: usb 2-1.1: new full-speed USB device number 3 using ehci-pci
kernel: usb 2-1.1: New USB device found, idVendor=0c4b, idProduct=0400
kernel: usb 2-1.1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
kernel: usb 2-1.1: Product: cyberJack e-com plus
kernel: usb 2-1.1: Manufacturer: REINER SCT

When I connect to a Windows7 machine I get the following messages:
[07:16:25:858] [4427:4431] [INFO][com.freerdp.channels.rdpdr.client] - Loading device service smartcard (static)
[07:16:25:858] [4427:4431] [INFO][com.freerdp.channels.rdpdr.client] - Loading device service smartcard (static)
[07:16:25:526] [4427:4431] [INFO][com.freerdp.channels.rdpdr.client] - registered device #1: SCARD (type=32 id=1)
[07:16:25:526] [4427:4431] [INFO][com.freerdp.channels.rdpdr.client] - registered device #2: SCARD (type=32 id=2)
[07:16:28:920] [4427:4450] [WARN][com.freerdp.channels.smartcard.client] - IRP failure: SCardStatusW (0x000900CC), status: SCARD_E_INSUFFICIENT_BUFFER (0x80100008)
[07:16:28:334] [4427:4431] [INFO][com.freerdp.channels.rdpdr.client] - registered device #1: SCARD (type=32 id=1)
[07:16:28:334] [4427:4431] [INFO][com.freerdp.channels.rdpdr.client] - registered device #2: SCARD (type=32 id=2)
[07:16:29:638] [4427:4461] [WARN][com.freerdp.channels.smartcard.client] - IRP failure: SCardStatusW (0x000900CC), status: SCARD_E_INSUFFICIENT_BUFFER (0x80100008)
[07:16:30:777] [4427:4464] [WARN][com.freerdp.channels.smartcard.client] - IRP failure: SCardStatusW (0x000900CC), status: SCARD_E_INSUFFICIENT_BUFFER (0x80100008)
[...]

PR #2898, opened 2015-09-18
Fix use of invalid input causing SCARD_E_INSUFFICIENT_BUFFER errors

bjcollins:
Dont use invalid/bogus input value for cbAtrLen for Smartcard Status Calls, instead just always use the max value here to indicate we have max buffer space available.

According to MS-RDPESC Section 2.2.2.18, the input value of cbAtrLen is unused and should be ignored, so we should not use it to try to calculate buffer size for actual Status calls.


bjcollins:
Issue #2648 should be addressed by this.

bmiklautz:
@freerdp-bot test.

bmiklautz:
Makes sense for me +1.

hardening:
+1

PR #2899, opened 2015-09-21
Standard RDP security server redirection fix

awakecoding:
Fix server redirection with standard security as observed in cases involving a 2012 R2 gateway and a load balanced environment. An optional padding field was considered mandatory, causing a false failure.

### DIFF
--- a/libfreerdp/core/redirection.c
+++ b/libfreerdp/core/redirection.c
@@ -387,8 +387,11 @@ int rdp_recv_enhanced_security_redirection_packet(rdpRdp* rdp, wStream* s)
 	if (status < 0)
 		return status;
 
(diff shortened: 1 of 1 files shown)


hardening:
+1

libfreerdp/core/redirection.c
```diff
@@ -387,8 +387,11 @@ int rdp_recv_enhanced_security_redirection_packet(rdpRdp* rdp, wStream* s)
 	if (status < 0)
 		return status;
 
-	if (!Stream_SafeSeek(s, 1)) /* pad2Octets (1 byte) */
-		return -1;
```
bmiklautz: Nice catch!

PR #2900, opened 2015-09-23
Three small fixes to /app (aka "rail")

peterh:
* xfreerdp: Fix windowOffset X/Y after local move
* xfreerdp: Fix "visible region" offset
* xfreerdp: Fix xf_rail_paint fencepost error

Do you prefer sets of small independent fixes (like this set) to have individual pull requests, or a bundled pull request like this one?

### DIFF
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -194,8 +194,8 @@ void xf_rail_end_local_move(xfContext* xfc, xfAppWindow* appWindow)
 	 * we can start to receive GDI orders for the new window dimensions before we
 	 * receive the RAIL ORDER for the new window size.  This avoids that race condition.
 	 */
-	appWindow->windowOffsetX = windowMove.left;
-	appWindow->windowOffsetY = windowMove.top;
+	appWindow->windowOffsetX = (INT16)windowMove.left;
+	appWindow->windowOffsetY = (INT16)windowMove.top;
 	appWindow->windowWidth = appWindow->width;
 	appWindow->windowHeight = appWindow->height;
 	appWindow->local_move.state = LMS_TERMINATING;
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -196,6 +196,16 @@ void xf_rail_end_local_move(xfContext* xfc, xfAppWindow* appWindow)
 	 */
 	appWindow->windowOffsetX = (INT16)windowMove.left;
 	appWindow->windowOffsetY = (INT16)windowMove.top;
+	if (appWindow->windowOffsetX < 0)
+		appWindow->localWindowOffsetCorrX = 0 - appWindow->windowOffsetX;
+	else
+		appWindow->localWindowOffsetCorrX = 0;
+
+	if (appWindow->windowOffsetY < 0)
+		appWindow->localWindowOffsetCorrY = 0 - appWindow->windowOffsetY;
+	else
+		appWindow->localWindowOffsetCorrY = 0;
+
 	appWindow->windowWidth = appWindow->width;
 	appWindow->windowHeight = appWindow->height;
 	appWindow->local_move.state = LMS_TERMINATING;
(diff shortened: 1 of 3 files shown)


hardening:
I have not tested, nor checked the assertion given in the commit, if the given hypothesis are good, it looks good to me.

hardening:
@freerdp-bot this looks good, but I'd be less worried if you can test a build of this.

client/X11/xf_rail.c
```diff
@@ -194,8 +194,8 @@ void xf_rail_end_local_move(xfContext* xfc, xfAppWindow* appWindow)
 	 * we can start to receive GDI orders for the new window dimensions before we
 	 * receive the RAIL ORDER for the new window size.  This avoids that race condition.
 	 */
-	appWindow->windowOffsetX = windowMove.left;
-	appWindow->windowOffsetY = windowMove.top;
+	appWindow->windowOffsetX = (INT16)windowMove.left;
+	appWindow->windowOffsetY = (INT16)windowMove.top;
```
bjcollins: This is pre-existing, but this is here to proactively update the new location locally that we are sending to the  server immediatey. I think we should also update the value of "appWindow->localWindowOffsetCorrX" and "appWindow->localWindowOffsetCorrY" here as well.

client/X11/xf_window.c
```diff
@@ -942,7 +944,11 @@ void xf_SetWindowVisibilityRects(xfContext* xfc, xfAppWindow* appWindow, RECTANG
 		xrects[i].height = rects[i].bottom - rects[i].top;
 	}
 
-	XShapeCombineRectangles(xfc->display, appWindow->handle, ShapeBounding, 0, 0, xrects, nrects, ShapeSet, 0);
+	/* Translate the server rail window offset to a local offset */
+	translatedWindowOffsetX = (appWindow->visibleOffsetX - appWindow->windowOffsetX);
+	translatedWindowOffsetY = (appWindow->visibleOffsetY - appWindow->windowOffsetY);
```
bjcollins: Does this logic work correctly when "appWindow->windowOffsetX" or "appWindow->windowOffsetY" are negative(such that the appWindow->localWindowOffsetCorr" variables are non-zero)? I would just think those might need to be taken into account as those are used to translate those negative values into our local coordinate system.

client/X11/xf_rail.c
peterh: That sounds like a good idea. Something like 50e4fcbda05fd0a706ed368fa49ff6693776150b ?

client/X11/xf_window.c
peterh: Yes, it works when windowOffsetX is negative. In fact, that's the only time I've ever witnessed translatedWindowOffsetX becoming non-zero. As I wrote in the commit message, this hunk was added to fix exactly the case when the window is partially off the left edge of the screen.

client/X11/xf_rail.c
bjcollins: Yea, that is what I was originally thinking.

client/X11/xf_window.c
bjcollins: I guess the non-negative case would be the one to be concerned about. Either way, I just do not think the logic here is always doing a translation to a local offset. The purpose of the "appWindow->localWindowOffsetCorr" variables was to handle cases where the "appWindow->windowOffset" variables are slightly negative - such as the case when you maximize a window. If they get large negative values when you slide the window off the left side of the screen then I think the overall logic is wrong(and I mean the existing logic beyond the scope of your commit) - as currently we are offseting anything negative into our coordinate system which does not support negative values when maybe we should just be truncating off the negative part. I will try to take a look at this sometime this week also.

bjcollins:
@peterh I created PR 2926 to supercede this PR, I started with your PR and tested and ended finding and making more changes. Please check that out and let me know what you think. Thanks for your changes in this area they pointed out some flaws and the existing code and helped moved things in a better direction.

peterh:
I don't have time to test today, but PR 9296 appears to supersede this PR. Looks good to me. Thanks for working on this.

PR #2901, opened 2015-09-23
Fixed crash when unused Japanese keys are pressed.

volth:
With Japanese physical keyboard and English layout, pressing the extra keys can cause FreeRDP crash.


hardening:
Looks good.

Side note: perhaps the lines above with (keysym == XK_Shift_L) || (keysym == XK_Shift_R)... would be covered by your if too and could be removed ? @awakecoding any opinion on this ?

bmiklautz:
@freerdp-bot test

PR #2908, opened 2015-09-30
Fix a bug in region_union_rect()

hardening:
When region_union_rect() was called with a rect that was above the region without
intersecting it, the first band was created too large. This patch fixes this and
add a unitary test.

All credits go to @nfedera that:
* found the bug;
* found a dataset to reproduce;
* coded the unitary test;
* spotted a suspicious line that was the faulty one.


nfedera:
+1 Thanks

PR #2909, opened 2015-10-02
Speedup for clipboard

akallabeth:
Removes unnecessary copy operation in clipboard client channel


bmiklautz:
@akallabeth looks good to me I wonder why this copy was done in the first place not sure if it hadn't a specific reason.

@awakecoding any ideas?

PR #2910, opened 2015-10-04
FindGStreamer_1_0: fix build failure for new gstreamer versions

dupondje:
On newer GStreamer versions FreeRDP fails to build with the following error.
/usr/include/gstreamer-1.0/gst/gstelement.h:55:27: fatal error: gst/gstconfig.h: Bestand of map bestaat niet

The patch was taken from:
https://github.com/WebKit/webkit/commit/c5a29d1e84dc0d23600a7049d831bd724014010b#diff-277559c7222bb7abe4c5ef38647cce71

I tested this on Ubuntu 15.10 and solves my build error.

### DIFF
--- a/cmake/FindGStreamer_1_0.cmake
+++ b/cmake/FindGStreamer_1_0.cmake
@@ -53,17 +53,17 @@ set(GSTREAMER_1_0_MINIMUM_VERSION 1.0.5)
 # Helper macro to find a Gstreamer plugin (or Gstreamer itself)
 #   _component_prefix is prepended to the _INCLUDE_DIRS and _LIBRARIES variables (eg. "GSTREAMER_1_0_AUDIO")
 #   _pkgconfig_name is the component's pkg-config name (eg. "gstreamer-1.0", or "gstreamer-video-1.0").
-#   _header is the component's header, relative to the gstreamer-1.0 directory (eg. "gst/gst.h").
 #   _library is the component's library name (eg. "gstreamer-1.0" or "gstvideo-1.0")
-macro(FIND_GSTREAMER_COMPONENT _component_prefix _pkgconfig_name _header _library)
+macro(FIND_GSTREAMER_COMPONENT _component_prefix _pkgconfig_name _library)
     # FIXME: The QUIET keyword can be used once we require CMake 2.8.2.
-    pkg_check_modules(PC_${_component_prefix} ${_pkgconfig_name})
 
-    find_path(${_component_prefix}_INCLUDE_DIRS
-        NAMES ${_header}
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot add @dupondje to whitelist

cmake/FindGStreamer_1_0.cmake
```diff
@@ -53,17 +53,17 @@ set(GSTREAMER_1_0_MINIMUM_VERSION 1.0.5)
...
     # FIXME: The QUIET keyword can be used once we require CMake 2.8.2.
-    pkg_check_modules(PC_${_component_prefix} ${_pkgconfig_name})
 
-    find_path(${_component_prefix}_INCLUDE_DIRS
-        NAMES ${_header}
-        HINTS ${PC_${_component_prefix}_INCLUDE_DIRS} ${PC_${_component_prefix}_INCLUDEDIR}
-        PATH_SUFFIXES gstreamer-1.0
-    )
+    string(REGEX MATCH "(.*)>=(.*)" _dummy "${_pkgconfig_name}")
+    if ("${CMAKE_MATCH_2}" STREQUAL "")
```
bmiklautz: As we never pass a extra version comparison here this check regex isn't required for us. But it also doesn't hurt.

cmake/FindGStreamer_1_0.cmake
```diff
@@ -53,17 +53,17 @@ set(GSTREAMER_1_0_MINIMUM_VERSION 1.0.5)
...
+        pkg_check_modules(PC_${_component_prefix} "${_pkgconfig_name} >= ${GStreamer_FIND_VERSION}")
```
bmiklautz: GStreamer_FIND_VERSION isn't set anywhere. Maybe we should use 
GSTREAMER_1_0_MINIMUM_VERSION for the components as well.

bmiklautz:
If all gstreamer pkg-config files exists this works like a charm. 
+1

yan12125:
Bump. Tested on Arch Linux with gstreamer 1.6.0 and commit d277346. Any progress on this PR?

bmiklautz:
@yan12125 generally no objection but I'm still waiting on the answer on the second comment.

yan12125:
Thanks for the information. Hope it merged soon :)

cmake/FindGStreamer_1_0.cmake
akallabeth: @yan12125 Second the comment of @bmiklautz, if you fix the minimum version check the pull is ready to go.

bmiklautz:
We should sort out the GStreamer_FIND_VERSION at some point but as the version is checked specially afterwards it's ok for now.
@dupondje thanks for your efforts.

PR #2912, opened 2015-10-05
Server/Shadow: Minor fix on authentication in shadow_client.c.

realjiangms:
Currently if username or password is not set, the authentication is always failed because the authentication callback is never tried.
Fix to always give a chance to try the subsystem authentication callback even if username / password is not set.


bmiklautz:
+1

hardening:
+1

PR #2913, opened 2015-10-05
Server/Shadow: Fix invalid ALIGN in shadow_client_send_bitmap_update

realjiangms:
Some of the code in shadow looks like ‘ X + X % 4'. I think the purpose is to align variables up to multiples of 4, but it is not correct.

On the other hand, the code in shadow_client_send_bitmap_update which adjusts nXSrc/nYSrc/nWidth/nHeight. I think the purpose is to extent the left/top/right/bottom of the rect to multiples of 4. It does right thing when it adjust nXSrc and nYSrc, but it should not change nXSrc and nYSrc when adjusting nWidth and nHeight


hardening:
LGTM

akallabeth:
LGTM, +1

PR #2914, opened 2015-10-05
Remove warning: HANDLE event variable unused

sriramster:
Throws a warning during compilation. And, see HANDLE event is not used inside function. So removed it.


mfleisz:
+1

PR #2917, opened 2015-10-08
add values-zh/string.xml,that was profiles for chinese

binglinoioij:
add values-zh/string.xml,that was profiles for chinese

### DIFF
--- a/client/Android/FreeRDPCore/res/values-zh/strings.xml
+++ b/client/Android/FreeRDPCore/res/values-zh/strings.xml
@@ -5,7 +5,7 @@
     <string name="no">否</string>
     <string name="cancel">取消</string>
     <string name="cont">继续</string>
-    <string name="login">登陆</string>
+    <string name="login">登录</string>
     <string name="logout">断开</string>
     <!-- Home menu items -->
     <string name="menu_exit">退出</string>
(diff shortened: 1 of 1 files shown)


hardening:
Looks good, but I don't have any Chinese knowledge.
@llyzs perhaps you could give us your opinion.

binglinoioij:
@hardening ,Maybe it's not perfect，but i will continue to improve it.

client/Android/FreeRDPCore/res/values-zh/strings.xml
```diff
@@ -0,0 +1,199 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<resources>
+    <!-- Button labels -->
+    <string name="yes">是</string>
+    <string name="no">否</string>
+    <string name="cancel">取消</string>
+    <string name="cont">继续</string>
+    <string name="login">登陆</string>
```
realjiangms: Shouldn't it be 登录? (according to your other translation for login in this file)

binglinoioij: Yes, you are right. I would change it, and submit later.

binglinoioij: And, i don't have any knowledge of photoshop, so i can't translate the word in the picture now. But i will try to learn some skill of photoshop to translate it. Thank you for your reply.

hardening:
Not a big deal, let's merge that

PR #2920, opened 2015-10-09
WinPR mbed TLS support

awakecoding:
This pull request is the first step towards adding full support for mbedTLS as an alternative to OpenSSL in FreeRDP/WinPR. New wrappers to abstract OpenSSL/mbedTLS are written in libwinpr-crypto and made as generic/clean as possible. It is now possible to generate a WinPR-only project (by generating in the winpr directory, excluding anything FreeRDP) that can build using OpenSSL only. The only remaining modules in WinPR that haven't been ported to mbedTLS yet are SChannel (incomplete anyway) and makecert. For regular FreeRDP builds, OpenSSL remains mandatory at this point, and it should not affect existing functionality.


hardening:
Looks interesting, anyway I think we should not land this before the 2.0 release.

bmiklautz:
@awakecoding it seems that this doesn't build on iOS anymore. What's the minimum openssl version required. If we require an more recent version I can update the build hosts. If a specific version is required I'd suggest we also check for it with cmake.
@hardening yep we definitely should ;)

awakecoding:
@bmiklautz @hardening the build now passes correctly and shouldn't affect existing OpenSSL support. I would like to have this merged now, so that I can at least use the latest FreeRDP branch for its WinPR component in my projects, without OpenSSL. This does not affect functionality in any way for FreeRDP support, and the regular builds still use OpenSSL. I plan on adding complete OpenSSL/mbedtls support in FreeRDP, but this can come at a later point in time, after the 2.0 release. Would this be a fair compromise?

PR #2924, opened 2015-10-13
egfx multimon fixes

awakecoding:
Fix usage of multiple codec contexts for egfx multi monitor support. Aside from ClearCodec, all other codec contexts need to be unique to the surface context. Prior to this pull request, a single codec context was shared across the different surface contexts, causing graphical artifacts and decoding failures.


hardening:
As far as I know the detail of egfx, +1

PR #2926, opened 2015-10-15
Rail fixes

bjcollins:
Remote app improvement and fixes, details in commit logs

1. Fully initialize new windows
2. Update local window on any change to size or position from server to address various race conditions
3. Remove bogus fullscreen toggle support for remote app
4. Always assign remote app windows a type
5. Only use visibility rects for XShape support
6. Change variables to correct signed type per RDP spec
7. Cleanup how we handle misc. X11 Atoms
8. Cleanup mouse event handling for local window move/size 
9. Remove the "localWindowOffsetCorr" variables as they are really unneeded and just complicate the logic.
10. Handle windows that are dragged off the left side of the screen
11. Handle maximizing/unmaximizing remote app windows correctly


bjcollins:
I plan for this PR to supercede PR 2900 from @peterh 
I have included his updates and also made more changes to address other rail related issues in the current baseline.

bjcollins:
It looks like this build is failing compiling tsmf, which is outside the scope of this change set. So, I suspect it is something unrelated to my changes.

bmiklautz:
@freerdp-bot test.

hardening:
As far as I know RAIL it looks good to me.

nfedera:
Also looks good to me. Very nice.

PR #2927, opened 2015-10-16
wlog: fixed return values

MartinHaimberger:
Wlog used to return an int but the only meaning
of the return value was:
* negative ... error
* 0 or positive ... success

but the positve returned value was 1 or some id of some
subsystem, nothing meaningful for the caller.

For a more meaningful returnvalue we now use BOOL.

If something goes wrong FALSE is returned.


bmiklautz:
@freerdp-bot test.

MartinHaimberger:
@freerdp-bot test.

bmiklautz:
+1

hardening:
+1

PR #2929, opened 2015-10-16
cliprdr/server: Don't call CloseHandle on ChannelEvent

mfleisz:
ChannelEvent is queried with WTSVirtualChannelQuery and calling CloseHandle on it will result in  crashes/exceptions.

### DIFF
--- a/channels/cliprdr/server/cliprdr_main.c
+++ b/channels/cliprdr/server/cliprdr_main.c
@@ -1409,12 +1409,6 @@ static UINT cliprdr_server_close(CliprdrServerContext* context)
 		cliprdr->ChannelHandle = NULL;
 	}
 
-	if (cliprdr->ChannelEvent)
-	{
-		CloseHandle(cliprdr->ChannelEvent);
-		cliprdr->ChannelEvent = NULL;
-	}
-
 	return CHANNEL_RC_OK;
 }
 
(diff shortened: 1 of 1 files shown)


llyzs:
@mfleisz 

I think this is going to be a handle leak, so instead of just remove the CloseHandle call, we should move these lines of codes to function cliprdr_server_context_free() in order to make sure that the handle is released before the context memory is freed.

llyzs:
@mfleisz 

Another off-topic matter I spotted: the CliprdrServerPrivate struct is allocated inside cliprdr_server_context_new() but is not freed in cliprdr_server_context_free() (only the context struct is freed, but the private struct is not). It looks like this is another memory leak.

mfleisz:
@llyzs I'm not quite sure about the handle leak ... I'm not an expert with channel code but I just took a look at the other channels and they all never CloseHandle the handle that is returned by WTSVirtualChannelQuery(WTSVirtualEventHandle). So either they all leak that handle or cliprdr incorrectly closes it (in fact the value seems still ok - not like free'd memory but Windows throws an exception that the value doesn't point to a valid handle).

Of course you are right about the private struct (context->handle) leak!

llyzs:
@mfleisz 

Ah, you are right, my mistake. The channel event handle is managed by VCM and should not be closed by the channel implementation. Looks good now!

PR #2932, opened 2015-10-17
GDI: Fix usage of gdi_get_brush_pointer.

realjiangms:
Currently we get color from brush according to the offset in the paint region.
According to MSDN https://msdn.microsoft.com/en-us/library/dd183396(v=vs.85).aspx, it should get color according to dest position instead of offset in paint region.

The issue messes screen display with some program such as AutoCAD 2010
image

### DIFF
--- a/libfreerdp/gdi/16bpp.c
+++ b/libfreerdp/gdi/16bpp.c
@@ -487,7 +487,7 @@ static BOOL BitBlt_PSDPxax_16bpp(HGDI_DC hdcDest, int nXDest, int nYDest, int nW
 			{
 				for (x = 0; x < nWidth; x++)
 				{
-					patp = (UINT16*) gdi_get_brush_pointer(hdcDest, x, y);
+					patp = (UINT16*) gdi_get_brush_pointer(hdcDest, nXDest + x, nYDest + y);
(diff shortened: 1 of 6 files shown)


realjiangms:
Note: The issue impacts only when using software gdi (with /gdi:sw)

libfreerdp/gdi/16bpp.c
```diff
@@ -692,13 +692,10 @@ static BOOL BitBlt_PATCOPY_16bpp(HGDI_DC hdcDest, int nXDest, int nYDest, int nW
...
 	{
-		/* align pattern to 8x8 grid to make sure transition
-		between different pattern blocks are smooth */
-
 		if (hdcDest->brush->style == GDI_BS_HATCHED)
 		{
-			xOffset = nXDest % 8;
-			yOffset = nYDest % 8 + 2; // +2 added after comparison to mstsc
+			xOffset = 0;
+			yOffset = 2; // +2 added after comparison to mstsc
```
realjiangms: Not sure why we need +2. Just keep same logic

hardening:
Not an expert of the subject, but LGTM

akallabeth:
LGTM +1

PR #2933, opened 2015-10-18
GDI: Fix gdi_bitmap_update to check dest buffer size.

realjiangms:
Check the destination size when update bitmap from server. Otherwise the screen may be messed as following or even crash the xfreerdp. The issue only happens with /gdi:sw

image


bmiklautz:
@realjiangms nice catch.
+1

hardening:
LGTM

Issue #2802, opened 2015-07-23
XFreeRdp disconnects after short periods of inactivity


wtfrank:
[17:14:50:809] [6520:6521] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpdr
[17:14:50:809] [6520:6521] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpsnd
[17:14:50:809] [6520:6521] [INFO][com.freerdp.client.common.cmdline] - loading channel cliprdr
[17:14:51:463] [6520:6521] [INFO][com.freerdp.core.gateway.tsg] - TS Gateway Connection Success
[17:14:52:978] [6520:6522] [INFO][com.freerdp.channels.rdpdr.client] - Loading device service drive (static)
[17:15:10:546] [6520:6522] [INFO][com.freerdp.channels.rdpdr.client] - registered device #1: home (type=8 id=1)
[17:19:05:574] [6520:6521] [ERROR][com.freerdp.core.rdp] - rdp_check_fds: tsg_check_event_handles() - -1
[17:19:05:574] [6520:6521] [ERROR][com.freerdp.core] - freerdp_check_fds() failed - 0
[17:19:05:574] [6520:6521] [INFO][com.freerdp.client.x11] - Network disconnect!
[17:19:05:574] [6520:6521] [ERROR][com.freerdp.client.x11] - Failed to check FreeRDP file descriptor

Launch command: client/X11/xfreerdp /g:gateway.xxx /v:xxx /d:xxx /u:xxx /p:xxx +clipboard /w:1600 /h:1000  +home-drive

This occurs with commit bfdf025e88bdd2961bd89b6bc24e6a0bda368ed7 but it has been happening since last year.

Let me know if there are any logs that it would be useful for me to gather.

volth:
I had the same problem and solved it by increasing [TCP_USER_TIMEOUT's value in tcp.c](https://github.com/FreeRDP/FreeRDP/blob/1e1983b7daf88375684854a977549d64f791173f/libfreerdp/core/tcp.c#L1030) from 4000ms to 60000ms.

From the manual: *When the value is greater than 0, it specifies the maximum amount of time in milliseconds that transmitted data may remain unacknowledged before TCP will forcibly close the corresponding connection and return ETIMEDOUT to the application.  If the option value is specified as 0, TCP will to use the system default.*

4000ms is often too small.

wtfrank:
Nice one, this solves the problem for me.

[...]

PR #2936, opened 2015-10-19
Increase TCP_USER_TIMEOUT to avoid disconnections.

volth:
TCP_USER_TIMEOUT value is too small, it is only 4 seconds. 
That causes random disconnections reported in the bug report https://github.com/FreeRDP/FreeRDP/issues/2802
This patch should fix the bug https://github.com/FreeRDP/FreeRDP/issues/2802

### DIFF
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -1027,7 +1027,7 @@ BOOL freerdp_tcp_set_keep_alive_mode(int sockfd)
 #endif
 
 #ifdef TCP_USER_TIMEOUT
-	optval = 4000;
+	optval = 60000;
 	optlen = sizeof(optval);
 
 	if (setsockopt(sockfd, SOL_TCP, TCP_USER_TIMEOUT, (void*) &optval, optlen) < 0)


bmiklautz:
@volth I think 60 seconds are too high on the other side. This would also mean that network problems (disconnects or so) take longer to be "detected". What do you think about that we use 30 for now and increase if it's still too less?

bmiklautz:
@freerdp-bot test.

abma:
> 4000 milliseconds is often too small.
> I think 60 seconds are too high on the other side.

-> make it configureable by the user / at runtime?!

bmiklautz:
@abma making it configurable makes perfect sens of course. But there should also be a sane and working default.

volth:
@amba @bmiklautz Perhaps "0" would the best choice, in this case "system default" value is used (somehow calculated out of the keep-alive timings). I did not tested it.

abma:
i've tested this change: it fixes for me random disconnects of xfreerdp when beeing idle.

idk what would be a reasonable value, but 60s sounds okish, 30s could be to low for wan/vpn connections.

bmiklautz:
@volth the system default timeout is usually rather long (can be up to 20 minutes and more) which caused hanging connections and such. This was the reasons for adding and using TCP_USER_TIMEOUT in the first place. 

@abma finding a good value here isn't that easy. Round trip times for slow connections like gprs are commonly  in the range of 1000ms addin VPN on top doubles or triples them. So the initial choice of 4 seconds even adds up some space (and was tested with a really really slow connection;).
What I wonder is why you are disconnected when idle - I haven't seen this behavior yet. The user timeout should only kick in when the client doesn't receive an ACK for the sent data.  Is there an open issue on that?

abma:
> i've tested this change: it fixes for me random disconnects of xfreerdp when beeing idle.

sorry, forget that, my test was wrong :-| i tested with an old version of xfreerdp and the cause for the random disconnects was some other bug which seems fixed.

bmiklautz:
@volth do you have a reproducible test case where 60 seconds are required?

volth:
@bmiklautz This timeout is not for the maximum round trip time, it is for the age of the oldest unacknowledged packet. My servers are in ~40ms ping from me, and I use cable modem connection, no mobile, not even WiFi. And with "4000" I have few disconnect per day. Not only when idle. I may work in a RDP session experiencing no lags nor freezes then oops - xfreerdp window is closed. 
I consider it is quite normal that there are unacknowledged packets with age of 100 round trip times, even on a good connection.

bmiklautz:
I'm aware that the timeout is not for the round trip but one need to consider the round trip time as well.. Howsoever my experience showed that if the user time out is hit it is mostly a network problem that's also what your above described behavior indicates for me. If everything works well and you are actively working what's the reason that one acknowledgement all at once take more than 4 seconds (which is hell a lot of time in an active tcp stream if the average round trip time around 40 ms)? For sure the server doesn't need to acknowledge a package immediately but I'd say such a delay is rather uncommon in a properly working environment.

I personally wouldn't like to increase the TCP_USER_TIMEOUT to much as it solved quite some issues when I've added it back then.  So let's set it to 16000  per default and use @abma suggestion and make it configurable. @volth would you be up for that?

volth:
Ok, let it be 16000 to close the issue quicker.

As for the future: it does not look convenient if the user would need to configure the timeout beforehand. Especially for the roaming users, who use connections of different quality during the day: fiber optic at the office, wifi at home, 3G on the road... Also, we know appropriate values for our networks, but users may not know them and not willing to know.

Perhaps it would be better to keep the default timeout value around "4000" but not break the connection when the time is over, instead showing a message: "connection issues detected, would you like to `close xfreerdp`, `try to re-connect`, `wait N seconds more`, `wait and increase the threshold for the next message` and auto-disappear the message if new packets arrived when the message on screen. (At the first glance `try to re-connect` may look like a good candidate for the only choice, and this is exactly how xfreerdp works with timeout "4000" and +auto-reconnect, but Windows servers often reject new connections when they see the old one is still active; so our trade over the timeout value is actually the trade over the threshold of automated choice between `try to re-connect` and `wait N seconds more` when the user cannot express its choice).

bmiklautz:
@volth did you have the change to analyze you disconnect problem on a package level? It would be really interesting  if there are a rising number of TCP re-transmissions or similar.

volth:
I think, yes. I recorded some pcap files with the disconnections. I will have a look.

bmiklautz:
@volth very nice! Thank you

bmiklautz:
@volth had any time to have a look into the pcap files?

bmiklautz:
Superseded by #3015

PR #2939, opened 2015-10-20
xfreerdp/clipr: fix self owned test and hardening

nfedera:
- xf_cliprdr_is_self_owned() lied if multiple xfreerdp instances were running.
- fixed a few unchecked callocs
- added/modified and handled some return values in compliance with the new hardened channel api


MartinHaimberger:
+1 LGTM

PR #2952, opened 2015-10-27
cliprdr/server: Provide the server with a possibility to configure capabilities

mfleisz:


akallabeth:
+1

include/freerdp/server/cliprdr.h
```diff
@@ -68,6 +68,12 @@ struct _cliprdr_server_context
 	void* handle;
 	void* custom;
 
+	/* clipboard capabilities - set by server */
```
bmiklautz: @mfleisz am I mistaken or are the following values in the end the combination 
of what is supported by the server and the client?

mfleisz: Exactly - the server first announces its capabilities (i.e. support for long format names) - if the client replies that it does not support that capability the server will revert to use use short format names in that example.

bmiklautz: @mfleisz what do you think about updating the comment to reflect this fact? While reviewing the PR the comment confused me a little ;).

bmiklautz:
Except my comment +1.
Nice that this can be set now \o/

mfleisz:
@bmiklautz Done - hope this is clearer now :)

bmiklautz:
@freerdp-bot test.

PR #2957, opened 2015-10-30
Adds a systemd journal appender

hardening:
In the wlog appender serie, an appender to systemd's journal.


winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -702,6 +702,10 @@ wLog* WLog_GetRoot()
 			else if (_stricmp(env, "SYSLOG") == 0)
 				logAppenderType = WLOG_APPENDER_SYSLOG;
 #endif /* HAVE_SYSLOG_H */
+#ifdef HAVE_JOURNALD_H
+			else if (_stricmp(env, "SYSTEMD") == 0)
```
bmiklautz: For consistency maybe we should use JOURNALD instead of SYSTEMD here.

winpr/libwinpr/utils/wlog/JournaldAppender.c
```diff
@@ -0,0 +1,172 @@
...
+			goto error_env_malloc;
+
+		GetEnvironmentVariableA(name, env, nSize);
+	}
+	else
+	{
+		env = _strdup("winpr");
+	}
+
+	fd = sd_journal_stream_fd(env, LOG_INFO, 1);
```
bmiklautz: If I'm not mistaken env is copied and can therefore be freed after the call.

bmiklautz:
@hardening with buffering disabled it works really well.

Two suggestions that I think that would be nice to have:
* would be great if the WLOG_JOURNALD_ID could also be set differently (maybe additional function like in the FileAppender or we add a generic way to change appender settings)
* for consistency we should name the WLOG_APPENDER value JOURNALD and not SYSTEMD.

Great work
+1

PR #2961, opened 2015-11-03
codec/color: fix segfault in freerdp_image32_copy

nfedera:
Code path for dstBytesPerPixel == 3 moved src and dst pointers beyond their respective buffers.


bmiklautz:
+1

PR #2962, opened 2015-11-03
Allow client build to be disabled

akallabeth:
* Adds ```WITH_CLIENT``` check to build platform dependent client.
* Fixes broken return values for ```PathCchIsRoot```


bmiklautz:
@akallabeth is WITH_CLIENT currently the default? If not we should definitely define the option.

akallabeth:
@bmiklautz Yes, it was already existing and I corrected my pull, ```WITH_CLIENT``` is now a dependent option of ```WITH_CLIENT_COMMON```. This way it is possible to create the client common library without any platform dependent stuff.

nfedera:
Not a cmake expert but LGTM

hardening:
+1

PR #2963, opened 2015-11-04
client/X11: multimon/fullscreen fixes

nfedera:
- fixed wrong calculation of xfc->fullscreenMonitors.[right|bottom]
- only use _NET_WM_FULLSCREEN_MONITORS if at least 2 monitors are involved
- call XMoveWindow before adding the _NET_WM_STATE_FULLSCREEN property
- call XMoveWindow after removing the _NET_WM_STATE_FULLSCREEN property

### DIFF
--- a/client/X11/xf_monitor.c
+++ b/client/X11/xf_monitor.c
@@ -120,7 +120,6 @@ BOOL xf_detect_monitors(xfContext* xfc, UINT32* pMaxWidth, UINT32* pMaxHeight)
 	int i;
 	int nmonitors = 0;
 	int primaryMonitorFound = FALSE;
-	int vX, vY, vWidth, vHeight;
 	VIRTUAL_SCREEN* vscreen;
 	rdpSettings* settings = xfc->settings;
 
@@ -270,10 +269,10 @@ BOOL xf_detect_monitors(xfContext* xfc, UINT32* pMaxWidth, UINT32* pMaxHeight)
 	if (settings->MonitorCount)
 	{
 		/* Initialize bounding rectangle for all monitors */
-		vWidth = settings->MonitorDefArray[0].width;
-		vHeight = settings->MonitorDefArray[0].height;
-		vX = settings->MonitorDefArray[0].x;
-		vY = settings->MonitorDefArray[0].y;
+		int vX = settings->MonitorDefArray[0].x;
+		int vY = settings->MonitorDefArray[0].y;
+		int vR = vX + settings->MonitorDefArray[0].width;
+		int vB = vY + settings->MonitorDefArray[0].height;
 		xfc->fullscreenMonitors.top = xfc->fullscreenMonitors.bottom =
 		xfc->fullscreenMonitors.left = xfc->fullscreenMonitors.right = settings->MonitorDefArray[0].orig_screen;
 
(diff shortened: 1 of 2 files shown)


client/X11/xf_monitor.c
```diff
@@ -285,36 +284,36 @@ BOOL xf_detect_monitors(xfContext* xfc, UINT32* pMaxWidth, UINT32* pMaxHeight)
...
-		vscreen->area.right = vWidth - 1;
+		vscreen->area.right = vR - vX - 1;
 		vscreen->area.top = 0;
-		vscreen->area.bottom = vHeight - 1;
+		vscreen->area.bottom = vB - vY - 1;
 
 		if (settings->Workarea)
 		{
 			vscreen->area.top = xfc->workArea.y;
-			vscreen->area.bottom = (vHeight - (vHeight - (xfc->workArea.height + xfc->workArea.y))) - 1;
```
nfedera: @bjcollins not sure was the original intention was, but vHeight - vHeight is zero - right ?

bjcollins: Your reduction of the calculation looks good to me!
I believe my intent here was to show the operation in(what I considered at the time)a logical format. I imagine that when I worked out what to do here, I took the steps explicitly as I had them and did not realize that  was taking extra unnecessary steps to get the values that I wanted.

client/X11/xf_window.c
```diff
@@ -185,30 +185,27 @@ void xf_SetWindowFullscreen(xfContext* xfc, xfWindow* window, BOOL fullscreen)
...
-
-	/* Only send monitor bounds if they are valid */
-	if ((xfc->fullscreenMonitors.top >= 0) &&
-			(xfc->fullscreenMonitors.bottom >= 0) &&
-			(xfc->fullscreenMonitors.left >= 0) &&
-			(xfc->fullscreenMonitors.right >= 0))
-	{
-		xf_SendClientEvent(xfc, window->handle, xfc->_NET_WM_FULLSCREEN_MONITORS, 5,
+		/* Set monitor bounds */
+		if (settings->MonitorCount > 1)
```
bjcollins: I do not believe this is the correct thing to do. If the user specifies a single monitor on the command-line when they have multiple monitors then that request should be honored, as it specifies exactly which monitor to fullscreen the window on. I think the "settings->MonitorCount variable currently only keeps track of the number of monitors to display on and not the total monitors available. Does this case make sense to you?

bjcollins:
@nfedera Good catch on the bogus calculation for "xfc->fullscreenMonitors.[right|bottom]"! I am surprised I never noticed had an issue as that code was clearly was not doing what it was intended to do.

client/X11/xf_window.c
nfedera: @bjcollins absolutely and basically you're right. The reason for this change is that some window managers (xfwm4 in particular) show a weird behavior if using _NET_WM_FULLSCREEN_MONITORS property where all IDs (top, bottom, right, left) are the same. On the other hand _NET_WM_FULLSCREEN_MONITORS is *not* required for full screen on a single specific monitor because it is and always was sufficient to move the window to the correct position and set _NET_WM_FULLSCREEN afterwards.

nfedera: @bjcollins ... and with this change I got at least single monitor fullscreen (/f) working and displaying on the correct monitor (either with your mouse position detection and/or by specifying a single monitor id using /f /monitors:X) with all window managers I had available (xfce, unity, openbox 3.5/3.6, icewm, awesome)

bjcollins: @nfedera Great! The second case is what I was worried about "specifying a single monitor id using /f /monitors:X", but it sounds like you have that covered. Your reasoning and approach makes sense to me.

nfedera:
I've noticed that restoring the original window position when toggling out of full screen did not work anymore (because the move is ignored in net wm full screen state). I've added commit 7ddd15d to this PR. Now if we're going to add _NET_WM_STATE_FULLSCREEN  XMoveWindow() will be called before setting the property and afterwards if we remove it.

hardening:
Not an expert but LGTM.

michaott:
I have Configuration for 3 Monitors on Gentoo-Linux with NVidia-Driver,

DVI-I-1 connected 1600x1200+0+0 (normal left inverted right x axis y axis) 432mm x 324mm
DP-0 connected primary 1920x1200+1600+0 (normal left inverted right x axis y axis) 519mm x 324mm
DP-1 connected 1600x1200+3520+0 (normal left inverted right x axis y axis) 432mm x 324mm

Output fom xfreerdp /monitor-list
      * [0] 1920x1200   +1600+0
        [1] 1600x1200   +0+0
        [2] 1600x1200   +3520+0

I Use /f /multimon /monitors:0,1

On FreeRDP version 1.2.1-dev  All OK (On Windows 10 Professional)
Windows Mainscreen is on DVI-I-1 with correct resolution 
Windows Enhanced Desktop is On DP-0  with correct Resolution

On FreeRDP version 2.0.0-dev (git be8f8f7)
Windows Mainscreen is in DP-0 with resolution from DVI-I-1 the rest from the Screnn is filled with the Start from Windows Enhanced Desktop from 1600 to 1920, On the other Monitors is Nothing from XFreeRDP. The Output from /monitor-list is the same to FreeRDP 1.2.

PR #2964, opened 2015-11-04
Add an UDP appender to wLog

hardening:
This appender allows to receive the logs over a network connection using UDP packets.
You can see the logs using a listening netcat, for example: nc -ul 127.0.0.1 20000.

### DIFF
--- a/winpr/libwinpr/utils/wlog/UdpAppender.c
+++ b/winpr/libwinpr/utils/wlog/UdpAppender.c
@@ -1,0 +1,209 @@
+/**
+ * WinPR: Windows Portable Runtime
+ * WinPR Logger
+ *
+ * Copyright © 2015 Thincast Technologies GmbH
+ * Copyright © 2015 David FORT <contact@hardening-consulting.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


winpr/libwinpr/utils/wlog/UdpAppender.c
```diff
@@ -0,0 +1,209 @@
...
+		appender->host = _strdup("127.0.0.1:20000");
+		if (!appender->host)
+			goto error_env_malloc;
+	}
+
+	return appender;
+
+error_open:
+	free(appender->host);
+error_env_malloc:
```
nfedera: unimportant remark: rename label to error_host_alloc since it is also targeted from strdup

winpr/libwinpr/utils/wlog/UdpAppender.c
```diff
@@ -0,0 +1,209 @@
...
+
+BOOL Wlog_UdpAppender_SetTarget(wLogUdpAppender* appender, const char *host)
+{
+
+	appender->targetAddrLen = 0;
+	if (appender->host)
+		free(appender->host);
+
+	appender->host = _strdup(host);
+	return (appender->host != NULL) && WLog_UdpAppender_Open(NULL, appender);
```
nfedera: I guess WLog_UdpAppender_Open() would always return FALSE here because NULL is passed for the log parameter which triggers the return FALSE in line 50

hardening: arg you're right

nfedera:
+1 tested, working, nice

PR #2970, opened 2015-11-09
2.0 preparations

bmiklautz:
* cleanup and rework WLog API
* introduce build-config.h
* fix compiler warnings and adapt error levels
* pump version to 2.0


bmiklautz:
@freerdp-bot test.

docs/wlog.md
```diff
@@ -0,0 +1,145 @@
...
+independently.
+
+TODO add more details and configuration examples.
+
+
+
+# Environment variables
+
+* WLOG_APPENDER  - the appender to use
+* WLOG_PREFIX - configure the prefix used for outputting the message (see
```
hardening: suggestion: the type of appender to use

docs/wlog.md
```diff
@@ -0,0 +1,145 @@
...
+The file appender writes the textual output to a file.
+
+Options:
+
+* "outputfilename", value const char*, filename to use
+* "outputfilepath", value const char*, location of the file
+
+### Udp
+
+This appender sends the loging messages to a pre-defined remote host via UDP.
```
hardening: s/loging/logging/

winpr/libwinpr/utils/wlog/Appender.c
```diff
@@ -21,11 +21,77 @@
...
+	if (!appender)
+		return FALSE;
+
+	if (!appender->Open)
+		return TRUE;
+
+	if (!appender->State)
+	{
+		status = appender->Open(log, appender);
+		appender->State = 1;
```
hardening: Again with names: may I suggest Opened or Active instead of State and move to BOOL as we're using only the 0 and 1 value ?

hardening:
I don't see a real benefit with the options set by string name and value in wLog, anyway it doesn't hurt, so +1.
Nice cleanup of wlog, misc warning fixes are also welcome.

1.2.0 is dead, long live the 2.0 !

bmiklautz:
@hardening the benefit is that we don't need to expose any details of the logger. Before we had special functions which was fine as long as *everything* was build all the time. With the addition of optional appenders, like syslog or journad, also those settings functions would have been optional and not available if the appender wasn't compiled into which leads to an inconsistent API.
I've choosen strings because we don't need to modify or extend the wlog header when a new appender is available or new options are available. So the benefit we do have is a stable API/ABI that is still easily extensible without breaking it :).

docs/wlog.md
bmiklautz: @hardening you mean with WLOG_APPENDER?

hardening: Yeah of course, I commented on wrong line, sorry

docs/wlog.md
```diff
@@ -0,0 +1,145 @@
...
+* "ln" - line number
+* "pid" - process id
+* "tid" - thread id
+* "yr" - year
+* "mo" - month
+* "dw" - day of week
+* "hr" - hour
+* "mi" - minute
+* "se" - second
+* "ml" - milliseconds
```
MartinHaimberger: maybe this should be millisecound (like the time description above)

docs/wlog.md
```diff
@@ -0,0 +1,145 @@
...
+# Levels
+
+The WLog are complementary the higher level always includes the lower ones.
+The level list below is top down. Top the highest level.
+
+* WLOG_TRACE - print everything including package dumps
+* WLOG_DEBUG - debug messages
+* WLOG_INFO - general informations
+* WLOG_WARN - warnings
+* WLOG_ERROR - error
```
MartinHaimberger: maybe errors

MartinHaimberger:
+1 LGTM

PR #2971, opened 2015-11-09
cmake: tweak manpage install location

woutershep:
This patch is needed when wanting to install binaries/libraries to other
locations than data. The linux distro Exherbo installs binaries and
libraries to /usr/<chost>/bin, /usr/<chost>/lib respectively but
manpages should still go in /usr/share/man/ because they are
architecture independent, without this patch they go in
/usr/<chost>/share/man unconditionally.

cmake documentation states:
DATAROOTDIR - read-only architecture-independent data root (share)

So this patch makes it use that so that its configurable.


hardening:
LGTM +1

bmiklautz:
@freerdp-bot test.

bmiklautz:
@freerdp-bot test.

bmiklautz:
@woutershep the PR looks good to me thanks. However I'm not entirely sure I understand your description.  When CMAKE_INSTALL_PREFIX is used and set to /usr (what I imagine happens) they would end up in /usr/share/man.

woutershep:
Sorry seeing it above it seems github is eating part of the text that is between &lt and &gt here on github. If you click on the commit the commit message has it properly though. But to to clarify a bit more, we set CMAKE_INSTALL_PREFIX to /usr/x86_64-pc-linux-gnu/ and CMAKE_INSTALL_DATAROOTDIR to /usr/share. This makes manpages end up in /usr/x86_64-pc-linux-gnu/share/man, while we want them in /usr/share/man. Does this clarify it enough for you?

bmiklautz:
@woutershep now it makes sense ;). Thank you very much!

PR #2973, opened 2015-11-10
Fixed build errors that occur when setting WITH_IOSAUDIO build option

mdhorda:
- Replaced } with ) in CMakeLists.txt
- Included header in TPCircularBuffer.c which is required for the TAG define


akallabeth:
@mdhorda Looks good to me, could you just rebase your branch to current master to get rid of the merge commit?
Other than that +1

hardening:
Looks good to me too.

PR #2976, opened 2015-11-12
Added audin support for mac os.

akallabeth:
* Basic microphone redirection for MacOS X (may also work with iOS, but not tested)
* Currently only ```PCM``` format is supported, others need some work.


mfleisz:
+1

hardening:
Looks good AFAI know the subject.

Just a remark: the patch defines WAVE_FORMAT_AAC_MS but never use it

akallabeth:
@hardening Added the format define for later use.
@awakecoding @bmiklautz @hardening any objections to this pull?

awakecoding:
can't hurt, as it wasn't supported before. no objections.

PR #2979, opened 2015-11-17
pkg: update version infos to 2.0

bmiklautz:


hardening:
LGTM
just wondering if the email team@freerdp.com is valid ?

bmiklautz:
@hardening it's valid. Give it a try :)

PR #2983, opened 2015-11-20
Fixes #2982. The idea is to be able to create the socket externally a…

### LINKED ISSUE #2982: Add support for use of existing socket in client connections
The idea is to be able to create the socket externally and pass that socket FD to FreeRDP so that it can be used there.

The idea suggested is to use the following interface:

settings->ServerHostname = "|" 
settings->ServerPort = SocketFD

Related issues are: #609 and  #2577

nitomartinez:
…nd pass that socket FD to FreeRDP so that it can be used there.

The idea suggested is to use the following interface:

settings->ServerHostname = "|"
settings->ServerPort = SocketFD

### DIFF
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -1047,6 +1047,7 @@ int freerdp_tcp_connect(rdpContext* context, rdpSettings* settings,
 	UINT32 optval;
 	socklen_t optlen;
 	BOOL ipcSocket = FALSE;
+	BOOL useExternalDefinedSocket = FALSE;
 
 	if (!hostname)
 		return -1;
@@ -1054,12 +1055,18 @@ int freerdp_tcp_connect(rdpContext* context, rdpSettings* settings,
 	if (hostname[0] == '/')
 		ipcSocket = TRUE;
 
+	if (hostname[0] == '|')
+		useExternalDefinedSocket = TRUE;
+
 	if (ipcSocket)
 	{
 		sockfd = freerdp_uds_connect(hostname);
 
 		if (sockfd < 0)
 			return -1;
+	} else if (useExternalDefinedSocket)
+	{
+	  sockfd = port;
 	}
 	else
 	{
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -1150,7 +1150,8 @@ int freerdp_tcp_connect(rdpContext* context, rdpSettings* settings,
 	settings->ClientAddress = freerdp_tcp_get_ip_address(sockfd);
 	if (!settings->ClientAddress)
 	{
-		close(sockfd);
+	        if (!useExternalDefinedSocket)
+		  close(sockfd);
 		WLog_ERR(TAG, "Couldn't get socket ip address");
 		return -1;
 	}
@@ -1158,7 +1159,7 @@ int freerdp_tcp_connect(rdpContext* context, rdpSettings* settings,
 	optval = 1;
 	optlen = sizeof(optval);
 
-	if (!ipcSocket)
+	if (!ipcSocket && !useExternalDefinedSocket)
 	{
 		if (setsockopt(sockfd, IPPROTO_TCP, TCP_NODELAY, (void*) &optval, optlen) < 0)
 			WLog_ERR(TAG, "unable to set TCP_NODELAY");
@@ -1181,7 +1182,7 @@ int freerdp_tcp_connect(rdpContext* context, rdpSettings* settings,
 		}
 	}
 
-	if (!ipcSocket)
+	if (!ipcSocket && !useExternalDefinedSocket)
 	{
 		if (!freerdp_tcp_set_keep_alive_mode(sockfd))
 		{
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -1064,10 +1064,9 @@ int freerdp_tcp_connect(rdpContext* context, rdpSettings* settings,
 
 		if (sockfd < 0)
 			return -1;
-	} else if (useExternalDefinedSocket)
-	{
-	  sockfd = port;
 	}
+        else if (useExternalDefinedSocket)
+	  sockfd = port;
 	else
 	{
 		sockfd = -1;


hardening:
I like the idea, the implementation looks like a hack. I don't see (yet) a better way to do it, so +1.

giox069:
I too like this idea: currently libfreerdp offers very poor handling of errors when opening a socket, see #2432. Opening the socket on the client (remmina, in my case) will let the user to receive some messages like "Host not found" or "connection refused" which libfreerdp is not currently returning correctly.
But  this implementation is a hack: we cannot reuse the same variable names giving them a different meaning.

hardening:
@giox069 well at least the implementation is done in the same spirit as the local Unix socket implementation. I'm not sure that using _ServerHostname="|\<socket fd\>"_ (and ignore port value) would be a lot better...

nitomartinez:
Happy to change it, 

I just tried to search for the same implementation that was there in the Unix Socket bit, which seemed to follow the same policy, and tried to be as less intrusive as possible (KISS policy). 

Happy to change if preferrable the settings structure if you think that that would be more appropiate, that is: 

- Adding a field to the settings structure called: preExistingSocket 
- Initialize that field to -1 (wehreever this structure is initialized) 
- If that is >=0 use the socket defined here 

Other alternatives might be: 

- still using the ServerHostname = "|" 
- Create an union so that we can use either ServerPort or ExistingSocketFd, this would be clearer in the naming perspective, and it would not need to use any extra data structure 
- Or create a different entry in the settings structure (4 more bytes in the structure which would not be an excessive cost). 

In our case we are using this in the QVD VDI solution ( http://theqvd.com ) and we tunnel everything through https, using curl as the implementation, doing some initialization and then using a websocket protocol upgrade, in this case to RDP, whence reusing the same socket. 

Please let me know if anything like this makes sense. 

Regards, 

Nito 



----- Original Message -----

giox069:
In my opinion (but I'm not a freerdp team member, nor a professional programmer): I would add two new field in the settings struct (file include/freerdp/settings.h)
```
UseExistingSocket
ExistingSocket
 ```
just after "ServerHostname" in the "Core Parameters" section.

The struct settings is initialized with zeroes in libfreerdp/core/settings.c in function freerdp_settings_new(), because it's allocated with calloc() which zeroes it.

Having an extra fields will also help client developers to recognize whether the option is available or not at compile time.

We should also discuss if socket closing, when UseExistingSocket != 0, should be kept inside freerdp or not. I have not idea for that, but I think that also socket closing should be moved outside.

**Comment by nitomartinez** [comment]
Happy with that, 

and the socket should be created/destroyed in the same layer, that is, if it has been created inside FreeRDP then destroyed inside FreeRDP, and if has been created outside, then outside. 

Generally a good practice, forcing also encapsulation strategies, (call it SOLID even if you are talking about C programming):-) 

Nito 

----- Original Message -----

**Comment by nitomartinez** [comment]
Any preference from the FreeRDP team?

Regards

-- 
Nito Martínez
CTO | Qindel Group
E: nito.martinez@qindel.com
M: +34678476906

**Comment by nitomartinez** [comment]
Just reviewed settings.h, and I saw this nice warnings regarding ABI stable zone :-) 

Just reafirming waiting on the FreeRDP team for their input before breaking anything ;-)

<pre>
struct rdp_settings
{
        /**
         * WARNING: this data structure is carefully padded for ABI stability!
         * Keeping this area clean is particularly challenging, so unless you are
         * a trusted developer you should NOT take the liberty of adding your own
         * options straight into the ABI stable zone. Instead, append them to the
         * very end of this data structure, in the zone marked as ABI unstable.
         */
</pre>

**Comment by bmiklautz** [comment]
@nitomartinez  nice idea. I wouldn't add extra parameters at all. Then it's really similar to unix sockets.
Two points:
* Not completely sure about is using |. Although for your use case, when running from another application or within a program this would be fine but in case you run it from a shell somehow one need to be aware that it needs to be quoted (this is often a problem when using remote apps which require ||). Anyway this is only a minor problem.
* The socket would later on be handled as if it was created from FreeRDP itself this includes setting timeouts. Maybe this isn't required/wanted (especially the tcp timeouts). In such case you should also disable those settings like for the icpSocket. Then the calling application can create the socket itself and do the setting itself before passing the fd. - Only exception here is non-blocking which is required.

**Comment by hardening** [comment]
@nitomartinez the patch doesn't touch any of the exported structs so no ABI problem here.
Thinking of it I'd tend to +1 the patch. 

I had proposed "|\<filedescriptor number\>" in ServerHostname, but with the night on it, I figured it would require to parse "again" the server name to extract the number (with all the corresponding error treatment).

**Comment by hardening** [comment]
@freerdp-bot could you have a look at this and do a test, thank you.

**Comment by bmiklautz** [comment]
@freerdp-bot test

File: libfreerdp/core/tcp.c [hunk 1]
```diff
@@ -1054,12 +1055,18 @@ int freerdp_tcp_connect(rdpContext* context, rdpSettings* settings,
...
+
 	if (ipcSocket)
 	{
 		sockfd = freerdp_uds_connect(hostname);
 
 		if (sockfd < 0)
 			return -1;
+	} else if (useExternalDefinedSocket)
+	{
+	  sockfd = port;
```
bmiklautz: Coding style. For single line statements don't use brackets.

hardening: I missed it, but the else if(... should also be on its own line

bmiklautz:
Except the formatting I think this is very useful and good to go.
+1

nitomartinez:
Updated:
* Formatting
* don't try to set sock options like tcp_nodelay and keepalive for external created sockets

hardening:
And merged, @nitomartinez  thanks for the contribution

nitomartinez:
:-)

giox069:
Thank you for the patch :)

antenore:
:+1:

giox069:
I'm sorry to comment a closed PR, but this patch has many problems.
I'm writing here just to leave it as a note:
1. ServerHostname is used a couple of times inside transport.c for TLS and NLA connection (when checking certificate and in building SPN ?). Having overwritten it with a "|" could cause problems.
2. With the externally created socket, all the autoreconnect feature of libfreerdp does not work. No idea why.
3. Did someone test it with remote desktop gateway ?

hardening:
1. Ho yes you're definitely right !
2. well autoreconnect doesn't make any sense in this case, as you're in the case where the socket has been created externaly
3. It won't work with remote gateway as we need more than 1 connection

giox069:
2. I was investigating in "autoreconnect cookies", and I think they will not work with the external socket, because I think they are managed internally libfreerdp. Let's say its a missing feature: when using an external socket, a reconnection to the server will require reentering credentials on the client or having it cached at the client side :)

hardening:
The final solution could probably be to have an interface with calbacks to create sockets (perhaps with the goal: direct server connection, rpc_in, rpc_out, ...). So that clients could override this interface and do what they want to provide a working socket...

giox069:
:+1: for callbacks or similar solution.
But I really don't need them now: my initial need (for Remmina) was only to have a better return code to display to the user when connecting. The most uncaught error is "Host not found", which currently is not shown to the Remmina user: when the user enters a bad hostname, he will only get a generic socket error.
Saying to the user "host not found" would be great. See issue #2432.
Using the external socket was a way to workaround this limitation and capture the socket error inside the client (Remmina)... but now I realize that that was not the perfect solution.

Issue #2302, opened 2015-01-13
Touchpad horizontal scroll translates to back/forward


grawity:
On a laptop, the touchpad supports both vertical and horizontal scrolling using a two-finger gesture. That is, even when I'm scrolling vertically, it is very easy to accidentally activate horizontal scroll.

Unfortunately, with FreeRDP on Linux/X11, the "horizontal scroll" events somehow get interpreted as additional mouse buttons and result in "back"/"forward" navigation. So whenever I'm trying to scroll a list or a page, I keep accidentally navigating backwards or forwards in history.

I don't know if this is a RDP thing or a FreeRDP thing or even an X11 thing (the servers are Windows 2011 SBS and 2012 R2), but if it is somehow possible to disable this behavior, that'd be at the top of my wishlist.

(Currently I just disable horiz scroll system-wide on the client OS, but that's far from ideal...)

colemickens:
This is very, very, very frustrating. How can we disable horizontal scroll events from being transmitted?

colemickens:
Can someone (@hardening, @awakecoding, @akallabeth) point me in the right direction for how I might about ignoring horizontal scroll events?

This issue is a big problem on any laptop. I'd like to see it fixed and help out if I can.

colemickens:
Ping again. This is unusable with libinput becoming the standard. There's no way to disable horizontal scroll events with this new input driver.

bsagal:
The horizontal scroll register with X as buttons 6 and 7

bmiklautz:
@colemickens what version of FreeRDP are you using?

bsagal:
the following patch removes the problem (it affects master, not sending a pull request as the patch is only a workaround and not a real solution)
```diff
--- a/client/X11/xf_event.c
+++ b/client/X11/xf_event.c
@@ -334,14 +334,14 @@ BOOL xf_generic_ButtonPress(xfContext* xfc, int x, int y, int button, Window win
                        flags = PTR_FLAGS_WHEEL | PTR_FLAGS_WHEEL_NEGATIVE | 0x0088;
                        break;
 
-               case 6:         /* wheel left or back */
+//             case 6:         /* wheel left or back */
                case 8:         /* back */
                case 97:        /* Xming */
                        extended = TRUE;
                        flags = PTR_XFLAGS_DOWN | PTR_XFLAGS_BUTTON1;
                        break;
 
-               case 7:         /* wheel right or forward */
[...]

PR #2984, opened 2015-11-24
remove horizontal scroll maping to forward/back

bsagal:
partal fix for #2302


akallabeth:
@bsagal Sorry for the late response, missed that thread ;)
As for mouse wheel mapping, have a look at MS-RDPBCGR 2.2.8.1.1.3.1.1.3 Mouse Event (TS_POINTER_EVENT) for possible event mappings.

hardening:
I feel like this patch could fix things for libinput based system but would break others. @bsagal can you confirm this ?

bsagal:
I have not tested, but don't think it is libinput specify.

I have found various examples with evdev that map left/right scroll to 6/7

Also look at on the comments in the existing `client/X11/xf_event.c`
```
		case 6:		/* wheel left or back */
 		case 8:		/* back */
 
		case 7:		/* wheel right or forward */
 		case 9:		/* forward */
```

6/7 are wheel left/right and 8/9 are forward/back.

akallabeth:
@bsagal You could map the wheel left / right stuff to ```PTRFLAGS_HWHEEL``` events (aka horizontal wheel)
except for that remark, LGTM +1

bsagal:
@akallabeth I think that is the correct solution but not entirely sure how to do it. 

What is the flags 0x0078 and 0x0088 (in cases 4/5)

akallabeth:
@bsagal As far as I can tell they are the number of rotation units for the scroll wheel (have a look at the spec I posted earlier, there is also the definition for ```PTRFLAGS_HWHEEL```, etc)

bsagal:
@abergmeier I have looked at the spec and tried to define ```PTRFLAGS_HWHEEL``` but crashed the client when sending to the server, not had a chance to look more into it. Also i have no idea how to check the Input Capability Set.

akallabeth:
@bsagal I've created a branch based on your pull with these fixes here: https://github.com/akallabeth/freerdp/tree/mouse_scroll
I have no mouse with a horizontal wheel / buttons, so could you check if it works for you?

bsagal:
@abergmeier Tried the branch but discovered i cannot test it as server 2008R2 does not support ```TS_INPUT_FLAG_MOUSE_HWHEEL```.

looking at the code i was wondering why the test in ```libfreerdp/core/capabilities.c``` for
```TS_INPUT_FLAG_MOUSE_HWHEEL``` is in an ```if else``` with ```INPUT_FLAG_FASTPATH_INPUT``` should the test not be in a separate ```if```

bsagal:
Found a server with 2012 to connect to and fixed the horizontal wheel support, this pull will now provide working horizontal.

akallabeth:
LGTM +1

PR #2991, opened 2015-11-28
Fix compilation error under Visual Studio 2010

xdqi:
Visual Studio 2010 use a compiler that supports only C89, which
only supports declaring variable at top of a local scope. Moving
scope variable to the top of function should solve this problem.


hardening:
@freerdp-bot this one is straight forward, but can you test it before I merge it ?

bmiklautz:
@xdqi We use visual studio 2012 (and above) for the compilation tests but not 2010 anymore ;). That's why this problem hasn't shown up before. Thanks for the fix.

PR #3014, opened 2015-12-16
winpr: create a specific winpr nt file handle

bmiklautz:
There exist two definitions of WINPR_FILE: in file/ and in nt/.
Both are different definitions used differently. Therefore split them
into WINPR_FILE and WINPR_NT_FILE.


akallabeth:
LGTM +1

bmiklautz:
I've re-based the branch.

mfleisz:
+1

PR #3015, opened 2015-12-16
tcp: change TCP_USER_TIMEOUT to 16000

bmiklautz:
The initially chosen value of 4 seconds was reported to be
to low for some networks. Therefore 16 seconds was chosen. This should be
short enough to detect broken connections but still long enough to avoid
disconnect because of flaky networks.

For details see issue #2936

### DIFF
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -1027,7 +1027,7 @@ BOOL freerdp_tcp_set_keep_alive_mode(int sockfd)
 #endif
 
 #ifdef TCP_USER_TIMEOUT
-	optval = 4000;
+	optval = 60000;
 	optlen = sizeof(optval);
 
 	if (setsockopt(sockfd, SOL_TCP, TCP_USER_TIMEOUT, (void*) &optval, optlen) < 0)
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -1027,7 +1027,7 @@ BOOL freerdp_tcp_set_keep_alive_mode(int sockfd)
 #endif
 
 #ifdef TCP_USER_TIMEOUT
-	optval = 60000;
+	optval = 9000;
 	optlen = sizeof(optval);
 
 	if (setsockopt(sockfd, SOL_TCP, TCP_USER_TIMEOUT, (void*) &optval, optlen) < 0)


bmiklautz:
@freerdp-bot test

giox069:
TL;DR

A kernel bug seems to impact the way the TCP_USER_TIMEOUT is calculated,
therefore FreeRDP closes the connection when it loose a single keepalive packet.
I propose to increase the timeout to 9 sec on linux platform till the kernel bug will be closed.

### My findings:
I'm investigating (doing packet sniffing) on what is causing disconnections under linux to both freerdp and remmina remote connections.
I almost came to the following conclusion: with current freerdp timing set inside libfreerdp/core/tcp.c (SO_KEEPALIVE=on TCP_KEEPIDLE=5 TCP_KEEPCNT=3 TCP_KEEPINTVL=2 TCP_USER_TIMEOUT=4000), freerdp cannot survive to a single loss of a keepalive packet.
Yes: if you minimize your xfreerdp window, you will see only keepalive packets flowing, one keepalive and one keepalive-ack every 5s.
But if only one of this packet is lost (in WAN links it could happen...), the linux kernel will put the socket in ETIMEDOUT without ever retrying to resend a keepalive retry.

[Kernel bug 108191](https://bugzilla.kernel.org/show_bug.cgi?id=108191) seems to explain why it happens: the TCP_USER_TIMEOUT is calculated erroneously starting from the last good known ack, and not from the 1st unreceived ack (I may be wrong here...).

My proposed workaround is obviously to raise TCP_USER_TIMEOUT with a value above TCP_KEEPIDLE. More precisely we should add TCP_KEEPIDLE to TCP_USER_TIMEOUT to neutralize bug 108191. So the final value for TCP_USER_TIMEOUT under current linux kernel should be be 9000, but only when the platform is linux.

```
#if defined __linux__
optval = 9000;
#else
optval = 4000;
#endif
```
To be changed when bug 108191 will be fixed.

Please consider also the following:
* My big doubt is: why under debian 7 with kernel 3.2.0-4 these disconnections do not occur ? I started to find disconnection problems only with debian 8 and ubuntu with kernel 3.16 and above. I have not tested kernels in between.
* [RFC 5482, section 4.2](https://tools.ietf.org/html/rfc5482#section-4.2) states that when SO_KEEPALIVE is enabled, TCP_KEEPILDE MUST be set to a value larger than TCP_USER_TIMEOUT. And I want to do the opposite in my patch, for __linux__. Right, but now we have kernel bug 108191 which prevent us doing so.
* I don't think that rising further TCP_USER_TIMEOUT will be a good idea. I tested xfreerdp with TCP_USER_TIMEOUT=9000 and I can drop all traffic for 3 seconds without experiencing disconnections. For longer timeouts a user should rely on the +autoreconnect command line switch of freerdp.
* The current code in transport.c has poor error reporting, and will never show that the socket has been disconnected due to TCP_USER_TIMEOUT expiring. I'm going to emit a PR which will add more descriptive error logs to some failed BIO_* funcion calls.

bmiklautz:
@giox069 wow nice finding! I've chosen 16 as a "solution" for  #2936 but I really like your proposal and change it to 9.

Thank  you for your effort!

bmiklautz:
@FreeRDP/contributors any comment/objections?

akallabeth:
+1

PR #3016, opened 2015-12-17
Fix wlog.h includes

akallabeth:
Due to missing ```stdarg.h``` ```wlog.h``` can not be included.


mfleisz:
+1

bmiklautz:
@akallabeth when including into a regular C project this should't be necessary. However it doesn't hurt either. +1

PR #3020, opened 2015-12-21
Add cmake flag -DWITH_SANITIZE_ADDRESS=on to enable AddressSanitizer

giox069:
GCC `-fsanitize=address` is an useful option to be used during debugging for memory leaks discovery. Just install libasan, add `-DWITH_GCC_SANITIZE_ADDRESS=on` to your cmake commandline, compile and run xfreerdp and you will discover some memory leaks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,6 +246,30 @@ if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang")
 	endif()
 endif()
 
+# Enable address sanitizer, where supported and when required
+if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang" OR CMAKE_COMPILER_IS_GNUCC)
+	if(WITH_SANITIZE_ADDRESS)
+		if (DEFINED CMAKE_REQUIRED_FLAGS)
+			set(SAVE_CMAKE_REQUIRED_FLAGS ${CMAKE_REQUIRED_FLAGS})
+		endif()
+		set(CMAKE_REQUIRED_FLAGS "-fsanitize=address")
+		CHECK_C_COMPILER_FLAG ("-fsanitize=address" fsanitize-address)
+		if(fsanitize-address)
+			set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fsanitize=address")
+		endif()
+		if (DEFINED SAVE_CMAKE_REQUIRED_FLAGS)
+			set(CMAKE_REQUIRED_FLAGS ${SAVE_CMAKE_REQUIRED_FLAGS})
+		else()
+			unset(CMAKE_REQUIRED_FLAGS)
+		endif()
+
+		CHECK_C_COMPILER_FLAG ("-fno-omit-frame-pointer" fno-omit-frame-pointer)
+		if(fno-omit-frame-pointer)
+			set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-omit-frame-pointer")
+		endif()
+	endif()
+endif()
+
 if(MSVC)
 	# Remove previous warning definitions,
 	# NMake is otherwise complaining.


akallabeth:
@giox069 Nice idea, but clang also supports ```-fsanitize=address```, would be nice to enable that in that case too.

bmiklautz:
I'd suggest adding a additional CHECK_C_COMPILER_FLAG as well. So if WITH_GCC_SANITIZE_ADDRESS is enabled the flag should be checked and activated if available.
Shouldn't we in case of WITH_GCC_SANITIZE_ADDRESS not also add -fno-omit-frame-pointer as well?

giox069:
@bmiklautz @akallabeth: unfortunately I'm not a big expert on this field (cmake and/or clang).
- yes, I like the idea to have -fsanitize=address built by default in Debug builds. Are there any side effects ?
- Second question: looking at CMakeLists.txt, with my poor knowledge of cmake, I can see that clang is used only in iOS and Android builds. Is there a way to force clang FreeRDP compilation under ubuntu or some other linuxes (just to do some testing) ?
- -fno-omit-frame-pointer: I never used it, but many sources reports it as useful for having nicer stack traces with -fsanitize=address
- yes, I can try to arrange the code for CHECK_C_COMPILER_FLAG

akallabeth:
@giox069: 
* Side effects are the overhead that ```-fsanitize=address``` introduces, the necessity to manually disable it if ```-fsanitize=thread``` is required. (negligible)
* clang is used by default only on iOS, but you can use it for any system the compiler is available for. (testing varies accrodingly). for MacOX and Linux you can enable it by setting the ```CC``` and ```CXX``` variables to your C and C++ compiler binaries before running cmake.
* ```-fno-omit-frame-pointer``` eases human readability of stack frames, without it the stack traces may lack some information.

So, hope this helps.

giox069:
New version:
- Renamed flag to WITH_SANITIZE_ADDRESS, because it's no longer GCC only
- Addedd clang support
- Added `-fno-omit-frame-pointer` to both Gcc and Clang
- use CHECK_C_COMPILER_FLAG
- not yet placed in default DEBUG builds. I'm not sure all developers want this...

Tested on Ubuntu 15.10 with both gcc and clang.

bmiklautz:
@giox069 very nice. Looks good to me! Thank you. 

What I'm not sure about is the use of CMAKE_REQUIRED_FLAGS. What is the reason for using this? CHECK_C_COMPILER_FLAG works without.

If you in mood you could also restructure the PR a little to use one common block for gcc and clang where you put in the flag checks. Something like:
```
if(${CMAKE_C_COMPILER_ID} STREQUAL "Clang" or CMAKE_COMPILER_IS_GNUCC )
```

giox069:
@bmiklautz: CHECK_C_COMPILER_FLAG only applies the test flag during the compilation phase, not during the link phase, and linking the small test program fails complaining that some symbols of libasan are not found. See here: http://blog.quirk.es/2015/05/fixing-bugs-with-gccs-fsanitize.html

Yes, good idea, I can arrange the code to do what you need (one single tests for both compilers). But I need some hours, I will do it this evening at home.

bmiklautz:
@giox069 ahh. Thanks you for pointing that out. Makes perfect sense. In that case as CMAKE_REQUIRED_FLAGS is required it's original value should be saved and restored afterwards as it might not be empty.

giox069:
. Moved code to a single "if" for both GCC and Clang
 . Use temporary variable SAVE_CMAKE_REQUIRED_FLAGS to save global var CMAKE_REQUIRED_FLAGS and restore it. Testing if it's defined before saving. Is there a better name I can use for this variable ?
 . I'm trying to squash the three commits into one, but I'm having some difficulties :)

bmiklautz:
+1

bmiklautz:
@freerdp-bot test.

bmiklautz:
@giox069 sorry for the delay. Could you rebase this to the current master branch then I'll merge it.

giox069:
I did a rebase, compilation failed. I think in the meanwhile you added some commit, so I will repush the "Updata branch" button again,  now.

bmiklautz:
Superseded by #3045 (PR has no merges).

Issue #2828, opened 2015-08-10
TLS 1.2


lemonmojo:
I'm currently trying to get a TLS 1.2 only connection working using FreeRDP.
The target machine is Windows Server 2012 R2 Core and I've deselected TLS 1.0 and TLS 1.1 using IIS Crypto (https://www.nartac.com/Products/IISCrypto).

I'm able to connect from a Windows 8.1 box using mstsc and I've verified that TLS 1.2 is indeed being used by enabling Schannel logging on the target machine (https://support.microsoft.com/en-us/kb/260729).

Unfortunately, I can't say the same about FreeRDP. Just compiled master with OpenSSL 1.0.2d (also tried 1.0.2a) and here's the output I get when trying to connect: https://gist.github.com/lemonmojo/a6f7f3b7680555b2bd1a

On the server there are no Schannel related event entries so it seems to fail early on.

dbungert:
Hey Lemonmojo,

A small tweak to libfreerdp/crypto/tls.c should do it.
You could do this to force it to 1.2

    diff --git a/libfreerdp/crypto/tls.c b/libfreerdp/crypto/tls.c
    index b33bb51..91a4330 100644
    --- a/libfreerdp/crypto/tls.c
    +++ b/libfreerdp/crypto/tls.c
    @@ -753,7 +753,7 @@ int tls_connect(rdpTls* tls, BIO* underlying)
             */
            options |= SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS;
    
    -       if (!tls_prepare(tls, underlying, TLSv1_client_method(), options, TRUE))
    +       if (!tls_prepare(tls, underlying, TLSv1_2_client_method(), options, TRUE))
                    return FALSE;
    
            return tls_do_handshake(tls, TRUE);`

Or this should let it negotiate TLS version.

    diff --git a/libfreerdp/crypto/tls.c b/libfreerdp/crypto/tls.c
    index b33bb51..760dc54 100644
    --- a/libfreerdp/crypto/tls.c
    +++ b/libfreerdp/crypto/tls.c
    @@ -753,7 +753,7 @@ int tls_connect(rdpTls* tls, BIO* underlying)
             */
            options |= SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS;
    
    -       if (!tls_prepare(tls, underlying, TLSv1_client_method(), options, TRUE))
[...]

PR #3036, opened 2016-01-07
Allow autonegotiation of tls/ssl version for tls 1.2 compatibilty

mbbush:
This is the fix referenced in the discussion of issue #2828. I have not tested it outside my own environment, but it works great for me and is reported to be working well by other users in that issue thread.

### DIFF
--- a/libfreerdp/crypto/tls.c
+++ b/libfreerdp/crypto/tls.c
@@ -798,7 +798,7 @@ int tls_connect(rdpTls* tls, BIO* underlying)
 	 */
 	options |= SSL_OP_DONT_INSERT_EMPTY_FRAGMENTS;
 
-	if (!tls_prepare(tls, underlying, TLSv1_client_method(), options, TRUE))
+	if (!tls_prepare(tls, underlying, SSLv23_client_method(), options, TRUE))
 		return FALSE;
 
 	return tls_do_handshake(tls, TRUE);


bmiklautz:
@freerdp-bot test

hardening:
With this patch we allow to do unsecure versions of SSL aka v2 and v3. Looks like a bad idea isn't it ?

bmiklautz:
@mbbush SSLv23_client_method was deprecated recently in openssl. I'd suggest using TLSv1_2_method instead. Could you try this as well?

bmiklautz:
When using TLS_method we could just disable SSLv3 with
```
options |= SSL_OP_NO_SSLv3;
```
This way TLS 1+ works but the Fallback to v3 is disabled. Sounds like the best solution for me.
@hardening @mbbush what do you think?

bmiklautz:
I've just checked and for now there is a compatibility macro available in openssl. So using SSLv23_client_method should be ok but still should disable sslv3 as commented above.

hardening:
SSL_OP_NO_SSLv2 would be nice too

bmiklautz:
@hardening absolutely :).

bmiklautz:
@hardening @mbbush have a look to PR #3046 I've added the options to disable SSL v2 and v3.

PR #3042, opened 2016-01-11
Move wayland client to UWAC

hardening:
In this PR instead of reimplementing all wayland tricks, we rely on the UWAC library that do it for us.
With this PR the wayland client works better, it still remain an issue with keyboard handling (how to choose a RDP locale from the mmap file of wayland...).


bmiklautz:
@hardening to build with uwac we need the lib somehow on the build systems otherwise the detection should disable wayland.

hardening:
Ok so that means that my cmake detection is crappy. Working on this.

bmiklautz:
@hardening the build works again but now we wouldn't do any test builds of the wayland client because uwac isn't available. - right?

hardening:
@bmiklautz yeah right you caught me :)

hardening:
@bmiklautz so this is the version with UWAC bundled in FreeRDP itself. Would be nice if xkbcommon could be installed on the builder...

bmiklautz:
@hardening it's on the list I take care of it as soon as I can :).

bmiklautz:
@freerdp-bot test.

bmiklautz:
@hardening would you rebase this to the latest master?

bmiklautz:
@freerdp-bot test.

hardening:
@freerdp-bot I think you failed the last test, would you retry ?

cmake/FindWayland.cmake
```diff
@@ -28,38 +25,35 @@
...
-          DOC "The Wayland client library"
+find_path(WAYLAND_INCLUDE_DIR wayland-client.h
+    HINTS ${WAYLAND_CLIENT_PKG_INCLUDE_DIRS}
 )
 
-include(FindPackageHandleStandardArgs)
-FIND_PACKAGE_HANDLE_STANDARD_ARGS(WAYLAND DEFAULT_MSG WAYLAND_LIBRARY WAYLAND_INCLUDE_DIR)
+find_library(WAYLAND_LIBS 
+    NAMES "wayland-client"
+    HINTS "${WAYLAND_CLIENT_PKG_LIBRARIES}"
```
akallabeth: @hardening According to cmake documentation this variable only contains the name of the library suitable to use with ```-l <libname>```, you should use ```WAYLAND_CLIENT_PKG_LIBRARY_DIRS``` instead

cmake/FindWayland.cmake
```diff
@@ -28,38 +25,35 @@
...
+    HINTS ${XKBCOMMON_PC_INCLUDE_DIRS}
+)
 
-if(WAYLAND_FOUND)
-  set(WAYLAND_LIBRARIES ${WAYLAND_LIBRARY})
-  set(WAYLAND_INCLUDE_DIRS ${WAYLAND_INCLUDE_DIR})
-endif()
+find_library(XKBCOMMON_LIBS 
+    NAMES xkbcommon
+    HINTS "${XKBCOMMON_PC_LIBRARIES}"
```
akallabeth: @hardening Same issue as above.

akallabeth:
@hardening This pull does not build properly on my debian jessie system. I have wayland libs installed, but they are not detected (see comments above).
Further the failed wayland detection does not disable all uwac related components (or abort the cmake run) and the compilation aborts with missing ```-luwac``` messages.

hardening:
@akallabeth I have fixed the detection, but after deinstalling wayland libs I can't reproduce the compilation bug: for me the wayland client and libs are correctly skipped. Anyway my debian, VM is a wheezy not a jessie.

hardening:
@freerdp-bot what's wrong with this PR, don't you want to test wayland ?

bmiklautz:
@freerdp-bot test

akallabeth:
@hardening can build the library on my machine now, the issues are (mostly) resolved, but you need to add ```${CMAKE_BINARY_DIR}/uwac/libuwac``` to the include path, as the protocol headers are generated there for out of tree builds.

uwac/libuwac/uwac-window.c
```diff
@@ -0,0 +1,616 @@
...
+		return UWAC_ERROR_INTERNAL;
+	}
+
+	data = mmap(NULL, allocSize * nbuffers, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+	if (data == MAP_FAILED)	{
+		ret = UWAC_ERROR_NOMEMORY;
+		goto error_mmap;
+	}
+
+	pool = wl_shm_create_pool(w->display->shm, fd, allocSize * nbuffers);
```
bmiklautz: Missing check?

uwac/libuwac/uwac-tools.c
```diff
@@ -0,0 +1,93 @@
...
+#include <string.h>
+#include <uwac/uwac-tools.h>
+
+/** @brief */
+struct uwac_touch_automata {
+	struct wl_array tp;
+};
+
+void UwacTouchAutomataInit(UwacTouchAutomata *automata) {
+	wl_array_init(&automata->tp);
```
bmiklautz: Does this always succeed?

uwac/libuwac/uwac-input.c
```diff
@@ -0,0 +1,845 @@
...
+}
+
+static void
+seat_handle_name(void *data, struct wl_seat *seat, const char *name)
+{
+	UwacSeat *input = data;
+	if (input->name)
+		free(input->name);
+
+	input->name = strdup(name);
```
bmiklautz: The return value of name should be checked.

client/Wayland/wlfreerdp.c
```diff
@@ -73,21 +90,23 @@ static BOOL wl_end_paint(rdpContext* context)
 	h = gdi->primary->hdc->hwnd->invalid->h;
 
 	context_w = (wlfContext*) context;
-	display = context_w->display;
-	window = context_w->window;
 
+	data = UwacWindowGetDrawingBuffer(context_w->window);
```
bmiklautz: I'm not sure but shouldn't we add a check if data is set here?

bmiklautz:
Looks good for me.
Except my comments +1.

uwac/libuwac/uwac-tools.c
hardening: yes

bmiklautz:
@hardening nice work! - I gonna add some versioning and pkg-config for it \o/

PR #3043, opened 2016-01-11
Build system improvements and changes

bmiklautz:
* get rid of FREERDP_MONOLITHIC_BUILD 
*  cleanup and improve exports (split into freerdp, freerdp-server, freerdp-client)
* don't install unnecessary linker archives when building with STATIC_CHANNELS
* add new command line option /buildconfig


akallabeth:
LGTM, +1

HenryJacques:
@bmiklautz why did you remove MONOLITHIC_BUILD?
We frequently use it!

Thanks

bmiklautz:
@HenryJacques the functionality MONOLITHIC_BUILD offered is default since a "while" (see commit removed0313ca3622baec53425b749e65caa202015894de). There is no non monolithic build anymore. The cmake option was only kept for compatibility.

bmiklautz:
@HenryJacques just make clear this is on master.

HenryJacques:
@bmiklautz  yes you're right. My mistake. I checked with the dev and we only use static libs now.

hardening:
+1

bmiklautz:
@freerdp-bot test.

PR #3046, opened 2016-01-12
tls: enable tls 1+

bmiklautz:
Currently TLS version 1.0 is used implicitly by using the TLSv1_method.
To be able to also use TLS 1.1 and later use SSLv23_client_method
instead. To make sure SSLv2 or SSLv3 isn't used disable them.

Based on PR #3036


mbbush:
This works fine for me to connect to a TLS 1.2-only server.

hardening:
+1

bmiklautz:
@Thunbolt22 what thin client are you using? - Ideally they should integrate the latest branch but at least something that contains the commit from this pull request (aa80f63b4ab19101cbdc376f7e0613ed410fee11)

Thunbolt22:
it is NoTouch by stratodesk.

PR #3049, opened 2016-01-13
Use major version only for SOVERSION

bmiklautz:
For 2.0+ major will be increased if ABI/API isn't compatible anymore.


bmiklautz:
@freerdp-bot test.

akallabeth:
LGTM +1

antenore:
All the .pc files and the include dir still have the major version in the file name, is that normal?

https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/CMakeLists.txt#L317

i.e.
```
configure_file(${CMAKE_CURRENT_SOURCE_DIR}/freerdp.pc.in
${CMAKE_CURRENT_BINARY_DIR}/freerdp${FREERDP_VERSION_MAJOR}.pc @ONLY)
```

I'm having an hard time compiling FreeRDP/Remmina

bmiklautz:
On 01/15/2016 12:05 AM, Antenore Gatta wrote:
> All the .pc files and the include dir still have the major version in the file name, is that normal?
Yes that was intentionally that it is possible to install multiple versions in parallel if necessary.

> https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/CMakeLists.txt#L317
> 
> i.e.
> ```
> configure_file(${CMAKE_CURRENT_SOURCE_DIR}/freerdp.pc.in
> ${CMAKE_CURRENT_BINARY_DIR}/freerdp${FREERDP_VERSION_MAJOR}.pc @ONLY)
> ```
> 
> I'm having an hard time compiling FreeRDP/Remmina
Just use freerdp2/freerdp-client2 instead of the previous one.

PR #3052, opened 2016-01-13
fix #3051: fullscreen switching with CTRL+ALT+ENTER doesn't work any …

### LINKED ISSUE #3051: CTRL+ALT+ENTER doesn't work any more in xfreerdp
this commit broke full screen switching CTRL+ALT+ENTER for me:

https://github.com/FreeRDP/FreeRDP/commit/a9e3368d4473e146c54ede37e3edd331e6e68d59

abma:
KeyboardState is defined as:

BOOL KeyboardState[256];

we can't store keysym there, we must use bool

not sure if the fix is correct, the alternative would be to define KeyboardState as

int KeyboardState[256];

### DIFF
--- a/client/X11/xf_keyboard.c
+++ b/client/X11/xf_keyboard.c
@@ -179,7 +179,7 @@ void xf_keyboard_release_all_keypress(xfContext* xfc)
 
 	for (keycode = 0; keycode < ARRAYSIZE(xfc->KeyboardState); keycode++)
 	{
-		if (!xfc->KeyboardState[keycode])
+		if (xfc->KeyboardState[keycode])
 		{
 			rdp_scancode = freerdp_keyboard_get_rdp_scancode_from_x11_keycode(keycode);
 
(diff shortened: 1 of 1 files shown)


client/X11/xf_keyboard.c
```diff
@@ -179,7 +179,7 @@ void xf_keyboard_release_all_keypress(xfContext* xfc)
 
 	for (keycode = 0; keycode < ARRAYSIZE(xfc->KeyboardState); keycode++)
 	{
-		if (xfc->KeyboardState[keycode] != NoSymbol)
+		if (!xfc->KeyboardState[keycode])
```
hardening: I think it's the opposite here as NoSymbol in the old code means FALSE now

hardening:
@freerdp-bot can you test this, please ?

client/X11/xf_keyboard.c
abma: true, will change / test

akallabeth:
@abma Nice catch, if you can fix the issue @hardening pointed out. Except this (blocking) remark +1

abma:
as far as i can tell, it works for me...

PR #3054, opened 2016-01-14
Leak fixes 

akallabeth:


winpr/libwinpr/file/file.c
```diff
@@ -253,7 +253,7 @@ static BOOL FileLockFileEx(HANDLE hFile, DWORD dwFlags, DWORD dwReserved,
 		DWORD nNumberOfBytesToLockLow, DWORD nNumberOfBytesToLockHigh,
 		LPOVERLAPPED lpOverlapped)
  {
-	int lock;
+	int lock = 0;
```
hardening: lock is always set a few lines below, so I don't think the initialization is useful

hardening:
+1 (non-blocking remark)

winpr/libwinpr/file/file.c
akallabeth: @hardening Yes, did one too much ^^

winpr/libwinpr/file/file.c
```diff
@@ -424,7 +424,7 @@ static HANDLE FileCreateFileA(LPCSTR lpFileName, DWORD dwDesiredAccess, DWORD dw
 	WINPR_FILE* pFile;
 	BOOL create;
 	const char* mode = FileGetMode(dwDesiredAccess, dwCreationDisposition, &create);
-	int lock;
```
mfleisz: I think this lock is also always set when it is used (either if FILE_SHARE_READ or FILE_SHARE_WRITE is set in dwShareMode)

akallabeth: @mfleisz No, look again ;) it is only set if ```dwSharedMode & (FILE_SHARE_READ | FILE_SHARE_WRITE)```

PR #3055, opened 2016-01-14
Created README.md for github.com

Mandar-Shinde:
Created README.md for github.com


hardening:
I like the idea, +1

bmiklautz:
@Mandar-Shinde yeah makes sense :). However we already do have a README so I'm in favor of renaming this to README.md and formating int with markdown.

PR #3059, opened 2016-01-15
Fix memory leak in lodepng_zlib_compress

akallabeth:
The memory in ```outv``` is reallocated and therefore the pointer in ```out``` is no longer valid.
Clean up the reallocated code and update the ```*out``` and ```*outsize``` values.


hardening:
LGTM

mfleisz:
+1

Issue #3058, opened 2016-01-15
the newest code compiled error


ygqzwl:
using centos 7 system
1.git clone git://github.com/FreeRDP/FreeRDP.git
2.the code directoty is:
-rwxrwxrwx 1 root root 358 1月 15 14:54        buildflags.h.in
-rwxrwxrwx 1 root root 2742 1月 15 14:54      ChangeLog
drwxrwxrwx 23 root root 4096 1月 15 14:54    channels
drwxrwxrwx 3 root root 35 1月 15 14:54          ci
drwxrwxrwx 11 root root 4096 1月 15 14:54    client
drwxrwxrwx 6 root root 4096 1月 15 14:54     cmake
-rwxrwxrwx 1 root root 3884 1月 15 14:54     CMakeCPack.cmake
-rwxrwxrwx 1 root root 345 1月 15 14:54       CMakeCPackOptions.cmake.in
-rwxrwxrwx 1 root root 27925 1月 15 14:54   CMakeLists.txt
-rwxrwxrwx 1 root root 2448 1月 15 14:54     config.h.in
drwxrwxrwx 2 root root 133 1月 15 14:54      docs
drwxrwxrwx 2 root root 19 1月 15 14:54        external
drwxrwxrwx 3 root root 53 1月 15 14:54         include
drwxrwxrwx 11 root root 4096 1月 15 14:54   libfreerdp
-rwxrwxrwx 1 root root 11358 1月 15 14:54    LICENSE
drwxrwxrwx 5 root root 56 1月 15 14:54         packaging
drwxrwxrwx 5 root root 139 1月 15 14:54       rdtk
-rwxrwxrwx 1 root root 1245 1月 15 14:54      README
drwxrwxrwx 2 root root 4096 1月 15 14:54     resources
drwxrwxrwx 2 root root 4096 1月 15 14:54     scripts
drwxrwxrwx 7 root root 4096 1月 15 14:54     server
drwxrwxrwx 2 root root 56 1月 15 14:54         third-party
drwxrwxrwx 6 root root 4096 1月 15 14:54     winpr

3.cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_SSE2=ON .

the errors:
CMake Error: install(EXPORT "FreeRDP-ClientTargets" ...) includes target "freerdp-client" which requires target "audin-client" that is not in the export set.
[...]

PR #3062, opened 2016-01-15
build: fix a problem with cmake version 2.8.11

bmiklautz:
Starting with cmake 2.8.10 FreeRDP exports a cmake find module. With 2.8.12
the PRIVATE/PUBLIC keywords were introduced in cmake. When building with
2.8.11 it is not possible to mark link dependencies as private and
therefore they need to be exported.


hardening:
Not a cmake expert, but LGTLM.
If it fixes #3058, let's merge it.

akallabeth:
LGTM +1

PR #3063, opened 2016-01-17
Fixed a memory issue found with AddressSanitizer.

akallabeth:
The strings may be shorter than 32 bytes, therefor use ```strncpy```

Thx @giox069 for pointing that out ;)

### DIFF
--- a/libfreerdp/core/settings.c
+++ b/libfreerdp/core/settings.c
@@ -734,8 +734,8 @@ rdpSettings* freerdp_settings_clone(rdpSettings* settings)
 		_settings->ClientProductId = malloc(32);
 		if (!_settings->ClientProductId)
 			goto out_fail;
-		CopyMemory(_settings->ClientHostname, settings->ClientHostname, 32);
-		CopyMemory(_settings->ClientProductId, settings->ClientProductId, 32);
+		strncpy(_settings->ClientHostname, settings->ClientHostname, 32);
+		strncpy(_settings->ClientProductId, settings->ClientProductId, 32);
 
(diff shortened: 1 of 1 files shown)


hardening:
@giox069 @akallabeth I was really wondering what was wrong with the CopyMemory. Looking at https://github.com/FreeRDP/Remmina/pull/728 I figured why, and that looks more like a misused API error: in the current code ClientHostname is expected to be filled with 32 bytes long chunk.

Let's say the current API is bad, so let's change it to something better (ie take the patch). But I'd suggest to drop the initial allocation of ClientHostname, and just do a _strdup() then. And adjust code using ClientHostname (Ctrl+Shift+G under eclipse on this field and you get them instantly). What do you think ?

giox069:
The problem is not really in CopyMemory, but in `client/common/cmdline.c`, line 1666: `settings->ClientHostname` is allocated with a small buffer via _strdup, and later assumed to be 32 bytes long in other FreeRDP places.
Here is line 1666 of cmdline.c:
`settings->ClientHostname = _strdup(arg->Value)`

Yes, the current API is bad, because having fixed strings at 32bytes is not a good choice: no one will remember this, because everyone is used to have strings allocated on the heap.

But changing it is up to you :) There are *a lot* of 32bytes strings around.

akallabeth:
@hardening Just as @giox069 already said, the problem is that it is not 32 bytes in the first place. I'm happy with every kind of fix, I just used strncpy as it copies only the used parts of the string (and therefore does remove the out of bound access)

akallabeth:
@hardening Ok, looked at this again, the fixed size is a really bad idea ;)
Even our own command line parser does fuck this up (see ```client/common/cmdline.c:1666```)

akallabeth:
Superseeded by #3076

PR #3066, opened 2016-01-18
Fixed SRCCOPY, using memmove now.

akallabeth:
memcpy is not defined, if source and destination overlap.

Thx @giox069 for pointing that out.


hardening:
+1

mfleisz:
+1

PR #3067, opened 2016-01-18
Resetting abort event on freerdp_connect.

akallabeth:
the ```abort``` event may be set by ```freerdp_abort_connect``` from a different thread after ```freerdp_disconnect```.
In that case the next call to ```freerdp_connect``` will not reset the event and ```freerdp_shall_disconnect``` will return true.

With this patch the event is reset on connect avoiding improper connection abort.


mfleisz:
+1

mfleisz:
@hardening Can you have a look at this?

hardening:
LGTM too

PR #3072, opened 2016-01-21
Improve error logging after calling some BIO_ functions

giox069:
Not really a fix, only a logging improvement: when a BIO_* function fails (ie: socket timeout), nothing is currently logged by freerdp.
This PR will add logging of these failures.
Please note that the code from malloc() to free() is completely untested, because I'm not really able to generate a SSL error, not a system error, for testing.

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -371,6 +371,53 @@ BOOL transport_accept_nla(rdpTransport* transport)
 	return TRUE;
 }
 
+#define WLog_ERR_BIO(tag, biofunc, bio) \
+	transport_bio_error_log(tag, biofunc, bio, __FILE__, __FUNCTION__, __LINE__)
+
+static void transport_bio_error_log(LPCSTR tag, LPCSTR biofunc, BIO* bio,
+	LPCSTR file, LPCSTR func, DWORD line)
+{
+	unsigned long sslerr;
+	char *buf;
+	wLog* log;
+	wLogMessage log_message;
+	int saveerrno;
+
+	saveerrno = errno;
+
+	log = WLog_Get(tag);
+	if (!log)
+		return;
+
+	log_message.Level = WLOG_ERROR;
+	if (log_message.Level < WLog_GetLogLevel(log))
+		return;
+
+	log_message.Type = WLOG_MESSAGE_TEXT;
+	log_message.LineNumber = line;
+	log_message.FileName = file;
+	log_message.FunctionName = func;
+
+	if (ERR_peek_error() == 0)
+	{
+		log_message.FormatString = "%s returned a system error %d: %s";
+		WLog_PrintMessage(log, &log_message, biofunc, saveerrno, strerror(saveerrno));
+		return;
+	}
+
+	buf = malloc(120);
+	if (buf)
+	{
+		while((sslerr = ERR_get_error()))
+		{
+			ERR_error_string_n(sslerr, buf, 120);
+			log_message.FormatString = "%s returned an error: %s";
+			WLog_PrintMessage(log, &log_message, biofunc, buf);
+		}
+		free(buf);
+	}
+}
+
 int transport_read_layer(rdpTransport* transport, BYTE* data, int bytes)
 {
 	int read = 0;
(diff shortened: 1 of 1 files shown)


libfreerdp/core/transport.c
```diff
@@ -371,6 +371,37 @@ BOOL transport_accept_nla(rdpTransport* transport)
 	return TRUE;
 }
 
+void transport_bio_error_log(char *biofunc, rdpTransport* transport)
```
hardening: In this file could you do some early returns so that we don't have these cascading if() ?

giox069: Done.
Should I squash the two commits into a single one ?

hardening:
LGTM

libfreerdp/core/transport.c
```diff
@@ -371,6 +371,35 @@ BOOL transport_accept_nla(rdpTransport* transport)
...
+		WLog_ERR(TAG, "%s returned system error %d: %s", biofunc, errno, strerror(errno));
+		return;
+	}
+
+	buf = malloc(120);
+	if (buf)
+	{
+		while((sslerr = ERR_get_error()))
+		{
+			ERR_error_string(sslerr, buf);
```
akallabeth: @giox069 Maybe it would be a good idea to use ```ERR_error_string_n```

libfreerdp/core/transport.c
akallabeth: @giox069 ```WLog_*``` functions also record where they were called from. This wrapper function negates this feature, maybe it would be a good idea to add ```__FUNCTION__, __LINE__ and __FILE__``` as arguments and use ```WLog_PrintMessage``` directly.

akallabeth:
other than my remarks LGTM

libfreerdp/core/transport.c
hardening: @akallabeth I have watched the doc and it says that a buffer of at least 120 bytes is enough for ERR_error_string.

akallabeth: @hardening Read that one too, just do not like magic numbers ;)

giox069: I will add _n and 120 to accomodate @akallabeth paranoid programming style :)

I will also check how to move original `__FUNCTION__`, `__LINE__` and `__FILE__` values inside my function (not yet understood how).

I hope I will be able to do it by tomorrow.

giox069:
Added _n, fixed a missing %s on a format string, squashed.

Now the part from malloc() and free() is tested: just try to enter bad credentials when connecting to a microsoft server and SSL will fail (this is expected):
```
[19:55:32:495] [29343:29344] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_AUTHENTICATION_FAILED [0x20009]
[19:55:32:495] [29343:29344] [ERROR][com.freerdp.core.transport] - biofunc BIO_read error: error:14094419:SSL routines:ssl3_read_bytes:tlsv1 alert access denied
[19:55:32:495] [29343:29344] [ERROR][com.freerdp.client.x11] - Freerdp connect error exit status 1
```
The second line comes from my new function.

@akallabeth: I can't find any reference to `__FUNCTION__`, `__LINE__` and `__FILE__` in WLog_ERR code.
```
#define WLog_ERR(tag, fmt, ...) WLog_Print(WLog_Get(tag), WLOG_ERROR, fmt, ## __VA_ARGS__)
```
It seems that WLog_* functions only make use of TAG, which is correctly defined one time per file. In case of trasport.c:
```
#define TAG FREERDP_TAG("core.transport")
```

akallabeth:
@giox069 Have a look at the define of ```WLog_Print```,

giox069:
Just updated PR with an improved solution with a macro.
How can I the full log output to the commandline with xfreerdp ? I can't see any option for logging with `xfreerdp /?`

akallabeth:
@giox069 Currently there is no command line option (maybe a good idea), you have to set the variable ```WLOG_LEVEL=trace``` and / or to filter some stuff ```WLOG_FILTER=<some tag>:<level>,...```

akallabeth:
Looks great now. +1
Any objections to merge?

bmiklautz:
@giox069 nice! Thank you.

PR #3073, opened 2016-01-21
Allow to specify the raw content of crypto materials

davewheel:
Sometime it's possible that your server application doesn't have access to files
(when running in a very restricted environment for example). This patch allows
to ship the private key and certificate as a string.


hardening:
@freerdp-bot can you test this PR ?
Please.

hardening:
@freerdp-bot you don't wanna test PR anymore ?

hardening:
Good idea, we don't always want files.
LGTM +1

mfleisz:
Nice job @davewheel +1

PR #3075, opened 2016-01-21
Added cmake patch from @david-geiger

akallabeth:
```wtypes.h``` and ```version.h``` from winpr are generated and therefore not found in source directory.

Thx @david-geiger for pointing that out.


hardening:
Good catch +1

bmiklautz:
+1

PR #3076, opened 2016-01-21
Removed fixed size strings.

akallabeth:
* Removes 32 byte fixed size restriction for ```ClientHostname``` and ```ClientProductId``` in settings. (still enforced where required)
* ```ClientHostname``` now converted to proper encoding before being sent. @hardening Was there any specific reason for this ```char*``` to ```WCHAR*``` hack? Should have failed on most encodings.

### DIFF
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -200,12 +200,13 @@ BOOL rdg_send_tunnel_authorization(rdpRdg* rdg)
 	BOOL status;
 	WCHAR* clientName = NULL;
 	UINT16 clientNameLen;
-	UINT32 packetSize = 12 + clientNameLen * 2;
+	UINT32 packetSize;
 
 	clientNameLen = ConvertToUnicode(CP_UTF8, 0, rdg->settings->ClientHostname, -1, &clientName, 0);
 	if (!clientName)
 		return FALSE;
 
+	packetSize = 12 + clientNameLen * 2 + sizeof(WCHAR);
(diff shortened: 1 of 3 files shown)


hardening:
Just a check: in settings_clone() if ClientHostname was not set in the source settings we will fail with OOM, the same with ClientProductId. So are you sure they are always set ?

And about Unicode conversion, I don't know, ClientHostname is supposed to be composed of chars in the ASCII set, so perhaps the conversion is not needed...

akallabeth:
@hardening For the initialization, both are set in ```freerdp_settings_new``` and later modified by the command line parser. It must be initialized by then, otherwise something else is wrong.
As for the unicode conversion, well the old code was buggy too (if a non ASCII char set is used) so we are probably better off with conversion

bmiklautz:
@akallabeth longer names aren't causing any problem later on? Wasn't this limited by the protocol somehow?

akallabeth:
@bmiklautz Length is enforced when used (and truncated if need be), but I will double check that.

akallabeth:
@bmiklautz Ok, rechecked the use of ```ClientHostname``` and ```ClientProductId```, the protocol restrictions are enforced in ```gcc_write_client_core_data```. The string is actually defined as 15 char unicode, so the old version was definitely broken. ```license_send_new_license_request_packet``` does not have such restrictions as far as I can tell from the protocol specs.

mfleisz:
LGTM

bmiklautz:
+1

PR #3078, opened 2016-01-23
Added new command line options for logger.

akallabeth:
The new command line options /log-level:<level> and
/log-filters:<filter>:<level> allow setting default log
level and log module filter to be set.

### DIFF
--- a/winpr/libwinpr/utils/wlog/wlog.c
+++ b/winpr/libwinpr/utils/wlog/wlog.c
@@ -55,7 +55,7 @@ typedef struct _wLogFilter wLogFilter;
  * http://docs.python.org/2/library/logging.html
  */
 
-const char* WLOG_LEVELS[7] =
+LPCSTR WLOG_LEVELS[7] =
 {
 	"TRACE",
 	"DEBUG",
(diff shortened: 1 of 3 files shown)


mfleisz:
+1

winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -302,6 +305,92 @@ DWORD WLog_GetLogLevel(wLog* log)
...
+		p++;
+	}
+
+	pos = g_FilterCount;
+	size = g_FilterCount + count;
+	tmp = (wLogFilter*) realloc(g_Filters, size * sizeof(wLogFilter));
+
+	if (!tmp)
+	{
+		free (g_Filters);
```
hardening: Do we need to free g_Filters ?

winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -302,6 +305,92 @@ DWORD WLog_GetLogLevel(wLog* log)
...
+	cp = (LPSTR)_strdup(filter);
+	if (!cp)
+		return FALSE;
+
+	p = cp;
+	s = cp;
+
+	do
+	{
+		p = strchr(p, ',');
```
hardening: isn't it 
p = strchr(s, ','); 
here ?

May I suggest renaming s to startAt and p to commaPos ?

winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -302,6 +305,92 @@ DWORD WLog_GetLogLevel(wLog* log)
...
+				return FALSE;
+			}
+		}
+		else
+			break;
+
+		if (p)
+		{
+			s = p + 1;
+			p++;
```
hardening: if you take my suggestion above, incrementing p is useless

hardening:
Some non blocking nitpick remarks.
+1

winpr/libwinpr/utils/wlog/wlog.c
akallabeth: @hardening The line is correct (```s``` always points to the beginning of the last string, ```p``` to the starting point for the next search / last found ```,```) but I'll rename them.

winpr/libwinpr/utils/wlog/wlog.c
akallabeth: @hardening If realloc fails that is the way to go, but you pointed out something I missed. Need to reset all static globals.

akallabeth: @hardening on second thought, maybe it is easier your way, just abort before update of the globals and let the uninit handle that...

bmiklautz:
@akallabeth what does take preference now? Command line or environment?

akallabeth:
@bmiklautz Command line for level, filters are merged.

Issue #2975, opened 2015-11-11
"/smartcard" with "/drive:*,hotplug" crash


Hobby-Student:
Hey there,

I have a problem with current master. I try to get away from
```
/drive:USB,/media
```
to
```
/drive:*,hotplug
```

For its own, it's working. But whenever I combine with
```
/smartcard
```

it's crashing:
![freerdp](https://cloud.githubusercontent.com/assets/6012744/11087185/7630ef1a-8858-11e5-9b3a-6482d7bf90b6.png)

Had no time for DEBUG compile... Will do this later, but perhaps someone has an idea before.

Compiled with
```
cmake -DCMAKE_BUILD_TYPE=Release -DWITH_SSE2=ON -DWITH_X11=ON -DWITH_DIRECTFB=OFF -DWITH_ALSA=ON -DWITH_PULSE=ON -DWITH_CUPS=ON -DWITH_PCSC=ON -DWITH_FFMPEG=OFF -DWITH_GSTREAMER=OFF -DWITH_JPEG=ON -DWITH_IPP=OFF -DWITH_XKBFILE=ON -DWITH_XINERAMA=ON -DWITH_XEXT=ON -DWITH_XCURSOR=ON -DWITH_XV=OFF -DWITH_XI=OFF -DWITH_XRENDER=ON
```

Hobby-Student:
ok... faster than I thought. Compiled with debug, but no more infos...

(tried freerdp 3 times... as you can see in the pic. ERROR because I cancelled login on server)

![freerdp-debug](https://cloud.githubusercontent.com/assets/6012744/11087982/ffffe376-885e-11e5-8b15-da1af1c5701b.png)

hyacinthes:
Try "/smartcard:" with "/drive:*,hotplug"

Hobby-Student:
@hyacinthes don't know how you got this, but the simple ":" makes it working 1 out of 20 times... Without the ":" it never works.

Did just some quick tests. Not reliable for production. Will try more next weekend and report back.

Hobby-Student:
damn... not working... I use the mentioned command line options and sometimes it connects to the server, but as soon as I connect an USB-stick, it crashes with "Speicherzugriffsfehler" (segfault?)

so... It is not working with smartcard and "usb hotplug".

hyacinthes:
I'm still digging it. I'll report back if any new information.

sconstantine:
[...]

PR #3079, opened 2016-01-24
Fixed argument checks for drive channel.

akallabeth:
* Fixes issues from #2975, ignore hotplug events for types other than ```RDPDR_DTYP_FILESYSTEM```
* Adds argument checks for functions in drive channel.

### DIFF
--- a/channels/rdpdr/client/devman.c
+++ b/channels/rdpdr/client/devman.c
@@ -6,6 +6,7 @@
  * Copyright 2010-2012 Marc-Andre Moreau <marcandre.moreau@gmail.com>
  * Copyright 2015 Thincast Technologies GmbH
  * Copyright 2015 DI (FH) Martin Haimberger <martin.haimberger@thincast.com>
+ * Copyright 2016 Armin Novak <armin.novak@gmail.com>
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -42,12 +43,18 @@
 
 void devman_device_free(DEVICE* device)
 {
+	if (!device)
+		return;
+
 	IFCALL(device->Free, device);
 }
 
 DEVMAN* devman_new(rdpdrPlugin* rdpdr)
 {
 	DEVMAN* devman;
+
+	if (!rdpdr)
+		return NULL;
 
 	devman = (DEVMAN*) calloc(1, sizeof(DEVMAN));
 
@@ -84,6 +91,9 @@ void devman_unregister_device(DEVMAN* devman, void* key)
 {
 	DEVICE* device;
 
+	if (!devman || !key)
+		return;
+
 	device = (DEVICE*) ListDictionary_Remove(devman->devices, key);
 
 	if (device)
@@ -98,6 +108,12 @@ void devman_unregister_device(DEVMAN* devman, void* key)
 static UINT devman_register_device(DEVMAN* devman, DEVICE* device)
 {
 	void* key = NULL;
+
+	if (!devman || !device)
+		return ERROR_INVALID_PARAMETER;
+
+	if (device->type != RDPDR_DTYP_FILESYSTEM)
+		return CHANNEL_RC_OK;
 
 	device->id = devman->id_sequence++;
 	key = (void*) (size_t) device->id;
@@ -114,6 +130,9 @@ DEVICE* devman_get_device_by_id(DEVMAN* devman, UINT32 id)
 {
 	DEVICE* device = NULL;
 	void* key = (void*) (size_t) id;
+
+	if (!devman)
+		return NULL;
 
 	device = (DEVICE*) ListDictionary_GetItemValue(devman->devices, key);
 
@@ -136,6 +155,9 @@ UINT devman_load_device_service(DEVMAN* devman, RDPDR_DEVICE* device, rdpContext
 	char* ServiceName = NULL;
 	DEVICE_SERVICE_ENTRY_POINTS ep;
 	PDEVICE_SERVICE_ENTRY entry = NULL;
+
+	if (!devman || !device || !rdpcontext)
+		return ERROR_INVALID_PARAMETER;
 
 	if (device->Type == RDPDR_DTYP_FILESYSTEM)
 		ServiceName = DRIVE_SERVICE_NAME;
--- a/channels/rdpdr/client/rdpdr_capabilities.c
+++ b/channels/rdpdr/client/rdpdr_capabilities.c
@@ -128,11 +128,14 @@ static void rdpdr_process_smartcard_capset(rdpdrPlugin* rdpdr, wStream* s)
 	Stream_Seek(s, capabilityLength - 4);
 }
 
-void rdpdr_process_capability_request(rdpdrPlugin* rdpdr, wStream* s)
+UINT rdpdr_process_capability_request(rdpdrPlugin* rdpdr, wStream* s)
 {
 	UINT16 i;
 	UINT16 numCapabilities;
 	UINT16 capabilityType;
+
+	if (!rdpdr || !s)
+		return ERROR_INVALID_PARAMETER;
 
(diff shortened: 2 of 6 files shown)


mfleisz:
+1

BUGgs:
+1

channels/rdpdr/client/devman.c
```diff
@@ -49,6 +53,9 @@ DEVMAN* devman_new(rdpdrPlugin* rdpdr)
 {
 	DEVMAN* devman;
 
+	if (!rdpdr)
```
hardening: Just curious, is there any reason why rdpdr would be NULL ? 
I don't like code that protect against itself, because you feel suspicious about everything. If callers are expected to not provide a NULL pointer, let's adjust the callers...

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -661,13 +671,18 @@ static UINT rdpdr_process_connect(rdpdrPlugin* rdpdr)
 	return error;
 }
 
-static void rdpdr_process_server_announce_request(rdpdrPlugin* rdpdr, wStream* s)
+static UINT rdpdr_process_server_announce_request(rdpdrPlugin* rdpdr, wStream* s)
 {
+	if (!rdpdr || !s)
```
hardening: Same remark here, how would we call that static internal function with invalid parameters ? Perhaps it's not a so nice idea but I feel like the job should be done by the callers (perhaps that these tests aren't even necessary)

hardening:
I gave a deeper look and AFAICT a check in rdpdr_virtual_channel_client_thread for the wStream and rdpdr would allow to remove all these checks (checking the stream and the rdpdr) in subfunctions.

For example looking at the call hierarchy of rdpdr_process_server_announce_request (Ctrl+Alt+H in eclipse), well, a simple check in the top function suffice. And doing so we don't have the strange feeling that the code must check that the callers haven't do something stupid.

@akallabeth there's probably a fix for something, but it would be cool if you could rework the patch to remove the unneeded checks.

akallabeth:
@hardening You're right, these checks should actually check the remaining stream length. (which is the only argument that can be manipulated from the outside)

akallabeth:
@hardening Reworked the checks, the NULL pointer stuff in static functions is gone, but doing tighter stream limit checks now.

channels/rdpdr/client/rdpdr_capabilities.c
```diff
@@ -120,53 +156,77 @@ static void rdpdr_write_smartcard_capset(rdpdrPlugin* rdpdr, wStream* s)
...
-				break;
+		case CAP_GENERAL_TYPE:
+			status = rdpdr_process_general_capset(rdpdr, s);
+			break;
 
-			case CAP_PRINTER_TYPE:
-				rdpdr_process_printer_capset(rdpdr, s);
-				break;
+		case CAP_PRINTER_TYPE:
+			rdpdr_process_printer_capset(rdpdr, s);
```
mfleisz: I think there is a status = missing here and in the lines below

mfleisz:
+1
@hardening can you have a look again?

channels/rdpdr/client/rdpdr_capabilities.c
```diff
@@ -60,12 +61,21 @@ static void rdpdr_write_general_capset(rdpdrPlugin* rdpdr, wStream* s)
 }
 
 /* Process device direction general capability set */
-static void rdpdr_process_general_capset(rdpdrPlugin* rdpdr, wStream* s)
+static UINT rdpdr_process_general_capset(rdpdrPlugin* rdpdr, wStream* s)
 {
 	UINT16 capabilityLength;
 
+	if (!Stream_EnsureRemainingCapacity(s, 2))
```
hardening: It's Stream_GetRemainingLength here and in line below, and then you should return something to say it's bad data (can't find back the error code)

channels/rdpdr/client/rdpdr_capabilities.c
```diff
@@ -120,53 +157,77 @@ static void rdpdr_write_smartcard_capset(rdpdrPlugin* rdpdr, wStream* s)
...
+		return CHANNEL_RC_NULL_DATA;
+
+	if (Stream_GetRemainingLength(s) < 4)
+		return CHANNEL_RC_NO_BUFFER;
+
 	Stream_Read_UINT16(s, numCapabilities);
 	Stream_Seek(s, 2); /* pad (2 bytes) */
 
+	if (Stream_GetRemainingLength(s) < sizeof(UINT16) * numCapabilities)
+		return CHANNEL_RC_NO_BUFFER;
```
hardening: there's possibly an int overflow here, should be:
if (Stream_GetRemainingLength(s) / sizeof(UINT16) < numCapabilities )

channels/rdpdr/client/rdpdr_capabilities.c
```diff
@@ -120,53 +157,77 @@ static void rdpdr_write_smartcard_capset(rdpdrPlugin* rdpdr, wStream* s)
...
+
 	for (i = 0; i < numCapabilities; i++)
 	{
 		Stream_Read_UINT16(s, capabilityType);
```
hardening: Another check should be done here as the check above is only the strict minimum number of bytes that the stream should have.
In fact I think you could remove the check above and only have one here.

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -921,6 +930,12 @@ static UINT rdpdr_process_receive(rdpdrPlugin* rdpdr, wStream* s)
 	UINT32 status;
 	UINT error;
 
+	if (!rdpdr || !s)
```
hardening: IMHO this check should be removed and done outside in rdpdr_virtual_channel_client_thread.

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -1132,14 +1157,15 @@ UINT rdpdr_send(rdpdrPlugin* rdpdr, wStream* s)
 	UINT status;
 	rdpdrPlugin* plugin = (rdpdrPlugin*) rdpdr;
 
+	if (!rdpdr || !s)
```
hardening: I've looked at the caller, and I don't think this check is necessary. s can't be NULL (as we expect the caller to have checked the result of Stream_New), and I can't imagine that caller would pass us a NULL rdpdrPlugin.

hardening:
General remark: I have looked at other channels and  we have used CHANNEL_RC_NO_MEMORY in when we're lacking of memory instead of CHANNEL_RC_NO_BUFFER.

hardening:
I like the patch as it adds missing checks all over the place.

channels/rdpdr/client/rdpdr_main.c
akallabeth: @hardening The function is not private to the file, so I'd assume such things can happen (but you're right, not that much of a threat).

channels/rdpdr/client/rdpdr_main.c
akallabeth: @hardening Well, preference. This way it was just easier to check everything and end in the error return path of ```rdpdr_virtual_channel_client_thread```

akallabeth:
To your general remark: Well, what is then the difference of both? I would have assumed one is for malloc issues the other for short reads / buffer size expectation mismatches. Currently ```CHANNEL_RC_NO_MEMORY``` is used for both.

hardening:
Well at least in the channel that I have treated I have tried to use CHANNEL_RC_NO_MEMORY when it's clearly a lack of memory (fail of malloc, Stream_New, and all functions that fails only because of OOM). When it's a too short stream or unexpected data, ERROR_INVALID_DATA seemed like the right error code. At least that's how I did my hardening of channels, and I'm suggesting for coherency...

akallabeth:
@hardening Ok, adjusted, now I also use ```ERROR_INVALID_DATA``` for short buffers.

PR #3082, opened 2016-01-26
Check server and client capabilities

hardening:
Most of the capabilities are sent by both the client and the server. But for some
the specs specify that they are only supposed to be only send by the server or the client.
This patch ensures this. Without this patch a malicious client can change server settings
and a malicious server can modify client settings.


akallabeth:
LGTM +1

mfleisz:
+1

PR #3083, opened 2016-01-26
Fix #3081, set client string encoding.

### LINKED ISSUE #3081: code page not set in capabilities
In ```rdp_write_order_capability_set``` the code page is always set to ```0```.
According to specs at https://msdn.microsoft.com/en-us/library/cc240556.aspx ```textANSICodePage``` should be set by the client to the codepage used.

akallabeth:


libfreerdp/core/capabilities.c
```diff
@@ -560,7 +560,8 @@ BOOL rdp_write_order_capability_set(wStream* s, rdpSettings* settings)
 	header = rdp_capability_set_start(s);
 
 	/* see [MSDN-CP]: http://msdn.microsoft.com/en-us/library/dd317756 */
-	textANSICodePage = 65001; /* Unicode (UTF-8) */
+    if (!settings->ServerMode)
+    	textANSICodePage = 65001; /* Unicode (UTF-8) */
```
mfleisz: Formatting

mfleisz:
+1 (after formatting is fixed)

hardening:
Arf I've just checked the spec:
extANSICodePage (2 bytes): A 16-bit, unsigned integer. ANSI code page descriptor being
used by the client (for a list of code pages, see [MSDN-CP]). **This field is ignored by the client
and SHOULD be set to zero by the server.**

mfleisz:
@hardening Haha - good find - So I guess we can close the PR .. or maybe change it to add a comment containing your findings?

hardening:
Not sure it's useless, in fact if the client sets it, then the server can read the value, and the answer to the client will be set to 0 and ignored by the client.
So it's a client to server message only...

hardening:
So after the spec bible examination, I'd tend to +1.

PR #3087, opened 2016-01-27
Fix png image loading that needed write access

DavBfr:
Opening a png file don't need the "append" flag to open a file for reading.
This is annoying if the file is in a read-only folder like /usr/share/...


hardening:
Nice catch.

PR #3093, opened 2016-01-28
Check more Stream_EnsureCapacity results when writing capabilities

hardening:
I think I'm done with capacities after this one.


MartinHaimberger:
+1 LGTM

PR #3101, opened 2016-02-02
Stdbool detection fix

akallabeth:
* Moved ```stdbool.h``` and ```stdint.h``` detection to winpr
* Removed obsolete includes in libfreerdp
* Using ```#cmakedefine01``` now to always evaluate to true false


mfleisz:
Fixes previously broke MSVC 2010 build
+1

PR #3103, opened 2016-02-03
gcc: assigned string must not be freed.

llyzs:
This fixed a double-free memory issue. The pointer was assigned but freed, then it will be used and freed again causing memory corruption.


hardening:
OMG how can we have missed that !
+1

PR #3107, opened 2016-02-04
Enable support for SNI on client side

ccpp:
Send server name when connecting to TS Gateway or RDP server


hardening:
LGTM

hardening:
@freerdp-bot what do you think about this, perhaps you could test it ?

akallabeth:
LGTM

PR #3108, opened 2016-02-04
Log level adjustments

akallabeth:
* Reduced log level for some messages

### DIFF
--- a/libfreerdp/core/rdp.c
+++ b/libfreerdp/core/rdp.c
@@ -363,7 +363,7 @@ BOOL rdp_read_header(rdpRdp* rdp, wStream* s, UINT16* length, UINT16* channelId)
 				rdp_set_error_info(rdp, ERRINFO_RPC_INITIATED_DISCONNECT);
 		}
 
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot Be nice and test this through.

libfreerdp/core/rdp.c
```diff
@@ -363,7 +363,7 @@ BOOL rdp_read_header(rdpRdp* rdp, wStream* s, UINT16* length, UINT16* channelId)
 				rdp_set_error_info(rdp, ERRINFO_RPC_INITIATED_DISCONNECT);
 		}
 
-		WLog_ERR(TAG, "DisconnectProviderUltimatum: reason: %d", reason);
+		WLog_DBG(TAG, "DisconnectProviderUltimatum: reason: %d", reason);
```
bmiklautz: Is there now a message printed in those cases? If not we should maybe change this to info.

akallabeth: @bmiklautz Yes, the error info is always printed. This message did just print the more or less cryptic integer version of it.

mfleisz:
+1

bmiklautz:
+1

PR #3109, opened 2016-02-04
fix X11/Xdg.h typedef collision fix

nfedera:
The fix was already there but not working if ```WINPR_HAVE_STDBOOL_H``` or ```WINPR_HAVE_STDINT_H``` was defined

### DIFF
--- a/winpr/include/winpr/wtypes.h.in
+++ b/winpr/include/winpr/wtypes.h.in
@@ -101,6 +101,7 @@
 #endif
 #endif
 
+#ifndef XMD_H /* X11/Xmd.h typedef collision with BOOL */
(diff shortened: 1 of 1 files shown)


akallabeth:
@nfedera So when is ```XMD_H``` defined? At CMake configuration time?

mfleisz:
@akallabeth I guess it is defined when you #include X11/Xdg.h in a project and #include winpr afterwards

nfedera:
@mfleisz right and @akallabeth Xdm.h always wins as their typedefs cannot be reversed (like these in winpr's wtypes.h).

In a perfect world everyone would use stdint.h types ....

akallabeth:
@nfedera Well, utopia is far away but until then this looks acceptable.

mfleisz:
+1

PR #3110, opened 2016-02-04
Mac os ssl warning fix

akallabeth:


mfleisz:
+1

hardening:
@akallabeth I guess it will only build with recent of MacOS X then ?

akallabeth:
@hardening Should only depend on the OpenSSL version used or do you know another reason for this removed const argument?

hardening:
@akallabeth yeah you're right, so the fix should be that the #ifdef would be a #if openssl_version > XXX isn't it ?

akallabeth:
@hardening  Just checked, this (old) API applies to all OpenSSL versions < 1.0.0. But neither gcc nor clang complain about the const argument when compiling against 0.9.8zh

akallabeth:
@hardening Ok, clang actually does complain. Looks like my cache was not up to date, forget the comment above.

mfleisz:
Any objections on merging this one?

hardening:
+1

PR #3111, opened 2016-02-05
Enable display scaling in client

colemickens:
Credit still goes to @irichter.

Tested, works for me.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -175,6 +175,12 @@ static COMMAND_LINE_ARGUMENT_A args[] =
 	{ "buildconfig", COMMAND_LINE_VALUE_FLAG | COMMAND_LINE_PRINT_BUILDCONFIG, NULL, NULL, NULL, -1, NULL, "print the build configuration" },
 	{ "log-level", COMMAND_LINE_VALUE_REQUIRED, "[OFF|FATAL|ERROR|WARN|INFO|DEBUG|TRACE]", NULL, NULL, -1, NULL, "Set the default log level" },
 	{ "log-filters", COMMAND_LINE_VALUE_REQUIRED, "<logger tag>:<log level>[, <logger tag>:<log level>][, ...]]", NULL, NULL, -1, NULL, "Set logger filters" },
+	{ "pwidth", COMMAND_LINE_VALUE_REQUIRED, "<physical width (mm)>", NULL, NULL, -1, NULL, "Physical width of display (in millimeters)" },
+	{ "pheight", COMMAND_LINE_VALUE_REQUIRED, "<physical height (mm)>", NULL, NULL, -1, NULL, "Physical height of display (in millimeters)" },
+	{ "orientation", COMMAND_LINE_VALUE_REQUIRED, "<orientation>", NULL, NULL, -1, NULL, "Orientation of display in degrees (0, 90, 180, 270)" },
+	{ "scale", COMMAND_LINE_VALUE_REQUIRED, "<scale amount (%%)>", "100", NULL, -1, NULL, "Scaling factor of the display (value of 100, 140, or 180)" },
+	{ "scale-desktop", COMMAND_LINE_VALUE_REQUIRED, "<scale amount (%%)>", "100", NULL, -1, NULL, "Scaling factor for desktop applications (value between 100 and 500)" },
+	{ "scale-device", COMMAND_LINE_VALUE_REQUIRED, "<scale amount (%%)>", "100", NULL, -1, NULL, "Scaling factor for app store applications (100, 140, or 180)" },
 	{ NULL, 0, NULL, NULL, NULL, -1, NULL, NULL }
 };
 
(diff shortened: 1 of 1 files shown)


akallabeth:
@colemickens wow, that was fast :) looks nice, but could you do an error return for invalid values? trying to harden the parser.

akallabeth:
@freerdp-bot and you start your testing duty!

colemickens:
Like that?

akallabeth:
Yes, thats it! LGTM, +1

client/common/cmdline.c
```diff
@@ -2289,6 +2295,49 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
+			settings->DesktopPhysicalHeight = atoi(arg->Value);
+		}
+		CommandLineSwitchCase(arg, "orientation")
+		{
+			settings->DesktopOrientation = atoi(arg->Value);
+		}
+		CommandLineSwitchCase(arg, "scale")
+		{
+			int scaleFactor = atoi(arg->Value);
+			if (scaleFactor == 100 || scaleFactor == 140 || scaleFactor == 180) {
```
mfleisz: @colemickens Nitpicking but freerdp's coding style puts { on its own line (i.e. see the CommandLineSwitchCase lines):
```
}
CommandLineSwitchCase(arg, "")
{
```

hardening:
LGTM +1
I'm waiting for @nfedera 's blessing on this PR, as he has looked a lot at that scaling things

client/common/cmdline.c
colemickens: Yeah, I'd rather be consistent then. I've updated the commit. Thanks!

klebed:
Hello! This feature implemented and merged almost year ago, but it seems not working. I'm using freerdp (2.0.0~dev201612151502+dfsg-1ubuntu0~git2b91654) from remmina-next PPA, which should include this commit, but getting nothing out of /scale:180 or /scale-desktop:180 or anything like that. I'm connecting to Win7, should it work with it, or you consider newer versions(then should it be documented somehow)?

akallabeth:
@klebed This feature requires support on server side too. As far as I know display scaling was added with win 8

klebed:
Well, OK, that makes sence... not really. =) What could be easier to translate 960x540 to 1920x1080 on client side and vice versa?
At least there should be notes in help, that some particular commands works only on server side. And determine supported versions.

akallabeth:
@klebed Looks like you're looking for the ```/smart-sizing```, that one is client side scaling only. The ```/scale-*``` options just announce some preference to the server (which then does the scaling).
Agree with the documentation though, but that is being worked on already.

klebed:
@akallabeth I don't see /smart-sizing anywhere in help, and it probably not intended to work, as I expect. I want to multiply server's resolution by 2, for comfortable work with RDP on HDPI screen. Right now everything in RDP window appears so tiny, so I had to strain my eyes, to force them read. I think it will end up with harm to my vision. And that's why I'm looking for solution.
Maybe there is no such thing?

akallabeth:
@klebed The smart-sizing feature is exactly what you're looking for. With ```/size:<width>x<height>``` you set the remote resolution, with ```/smart-sizing:<width>x<height>``` you set the local window size. 
What kind of help are you looking at? There is a man page and ```xfreerdp /?```, both documenting that option.

klebed:
+akallabeth Here's documentation I'm using: https://github.com/FreeRDP/FreeRDP/wiki/CommandLineInterface
Thanks a lot! This, works at last!

akallabeth:
@klebed ah, yes the wiki is due for an update. as for power of 2 just use the 2x resolution locally, no magic there.

dimitry-ishenko:
Sorry for beating a dead horse, but is it possible to do scaling in RemoteApp mode? I've tried adding `/size:1920x1080` and it didn't seem to make a difference. Running Win7 Ultimate.

erenoglu:
Dears, is this even working? I'm running FreeRDP version 2.0.0-dev5 (8954e1646).

Whatever I do, I can't get my 2560x1440 resolution monitor to display my remote Windows 10 machine with sensible font sizes. Everything is very tiny.

I'm also using /scale:180 or /scale-desktop:180, none works. I tried putting /pwidth:5300 /pheight:3100, did not do anything. 

I dont want client side scaling neither as everything becomes blurry.

ChthonVII:
Indeed, /scale is busted. I've got an old nightly installed on the side in /opt and /scale:180 works exactly as desired. If I switch to the v2.0.0 that I got with the upgrade to Debian Buster, keeping the command line otherwise exactly the same, /scale:180 has no effect.

paravz:
> @klebed The smart-sizing feature is exactly what you're looking for. With `/size:<width>x<height>` you set the remote resolution, with `/smart-sizing:<width>x<height>` you set the local window size.
> What kind of help are you looking at? There is a man page and `xfreerdp /?`, both documenting that option.

Thats's the first time i understand what the options mean and how to use it to "zoom" remote window, no combination of /scale or /scale-desktop worked for me to do the same.

roughly full screen xfreerdp window on 4k display with remote host zoomed 2x:
`xfreerdp +fonts +aero /bpp:24 /smart-sizing:3800x2000 /size:1900x1000 /v:host /u:user /gdi:hw +glyph-cache`

PR #3123, opened 2016-02-11
Several fixes for keyboard type 7 (japanese)

nfedera:


hardening:
wow, do you mean that some of the VK for keyboard type 7 were just wrong ?

nfedera:
@hardening yes, and at least for freerdp based servers we should replace this VK_ crap anyway with direct mappings between the codes inside the clients' rdp input PDUs and the backend codes (like evdev). There are VK_ collisions and many scancodes don't even have any official VK_ defines because Windows generates different virtual VK_codes. 

A good example is the scancode 0x70 (hiragana/katakana toggle) from the Type 7 (Japanese) Keyboard: X11 has the `Hiragana_Katakana` symbol, evdev has `<HKTG>` but there is no `VK_Hiragana_Katakana`. So we have to either create an unused define or misuse on of the VK_OEM_ defines which can cause collisions.

bmiklautz:
+1

PR #3124, opened 2016-02-12
libfreerdp-codec: Make region16_rects() handle NULL nbRects

nathankidd:
Now matches header documentation.  Also cleans up related doc grammar.

### DIFF
--- a/include/freerdp/codec/region.h
+++ b/include/freerdp/codec/region.h
@@ -73,10 +73,10 @@ FREERDP_API void region16_init(REGION16 *region);
 /** @return the number of rectangles of this region16 */
 FREERDP_API int region16_n_rects(const REGION16 *region);
 
-/** returns a pointer on rectangles and the number of rectangles in this region.
- * nbRect can be set to NULL if not interested by the numnber of rectangles.
+/** returns a pointer to rectangles and the number of rectangles in this region.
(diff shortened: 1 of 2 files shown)


akallabeth:
Looks good, but if you are at it, could you fix the ```assert(region->data)``` too? It is already checked the line below.

hardening:
+1

Yeah the assert must be removed. And the return statement should be return NULL and not 0 (cosmetic).

I'm wondering why I did it this way (having nbrect optionnal) as it makes no sense to do it in two steps when one suffice (I mean calling region16_n_rects(r), then region_rects(r, NULL)).

hardening:
@freerdp-bot it looks good but would you test a compilation of this, please ?

nathankidd:
Good points, I added the assert and NULL changes.

It makes some sense to handle the NULL nbRects case.  Consider an API like:

`MyRectFrob(int nRects, RECTANGLE_16* rects)` 

It needs to extract the number of rects and the rects pointer separately.  Handling NULL means one doesn't need to have extra stack variables to catch the values and can just inline
`MyRectFrob(region16_n_rects(region), region16_rects(region, NULL))`

hardening:
Well stack variables are there for free. And I can't see any real case where you would get the number of rectangles to not use them immediately after that.

PR #2959, opened 2015-11-02
Add the possibility to redirect specific smartcard readers

HenryJacques:
Until now, /smartcard: had the behaviour to redirect every connected readers. It didn't matter what you used after the parameter /smartcard:xxxx:yyyy.

With this patch you can:
- use /smartcard: "as is" (without argument) to continue to redirect every readers
- use /smartcard:"Reader name" (you have to repeat that for every reader you want to redirect)

The reader name is exactly the one returned by libpcsclite, e.g.

root@host:# opensc-tool -l
# Detected readers (pcsc)
Nr.  Card  Features  Name
0    Yes             Xiring Leo V2 (12547854125036) 00 00
1    Yes             Xiring Leo V2 (52412506854712) 01 00

(Thanks to A. Aing for this PR)

PR #3126, opened 2016-02-12
Add the possibility to redirect specify smartcard readers - 2nd try

HenryJacques:
Same PR as #2959 
But this time it should be clean ;)

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -831,15 +831,26 @@ static int freerdp_client_command_line_post_filter(void* context, COMMAND_LINE_A
 	}
 	CommandLineSwitchCase(arg, "smartcard")
 	{
-		char** p;
-		int count;
-
-		p = freerdp_command_line_parse_comma_separated_values_offset(arg->Value, &count);
-		p[0] = "smartcard";
-
-		status = freerdp_client_add_device_channel(settings, count, p);
-
-		free(p);
+		if (arg->Flags & COMMAND_LINE_VALUE_PRESENT)
+		{
+			char** p;
+			int count;
+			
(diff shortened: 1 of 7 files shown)


akallabeth:
@HenryJacques Sorry, still not up to date with current master...

client/common/cmdline.c
```diff
@@ -830,15 +831,26 @@ static int freerdp_client_command_line_post_filter(void* context, COMMAND_LINE_A
...
-
-		free(p);
+		if (arg->Flags & COMMAND_LINE_VALUE_PRESENT)
+		{
+			char** p;
+			int count;
+			
+			p = freerdp_command_line_parse_comma_separated_values_offset(arg->Value, &count);
+			p[0] = "smartcard";
+			free(p);
```
akallabeth: @HenryJacques This is definitely broken, you do not load the channel and free the arguments...

HenryJacques: @akallabeth  What is broken? On my computer the smartcard channel gets loaded twice (as always) and /smartcard & /smartcard: have the same behaviour.

akallabeth:
the pull is still based on an incompatible branch and when adding some arguments to /smartcard you will enter the branch not loading the channel ;)

HenryJacques:
@akallabeth I'm really sorry to ask this, but how do you know it is still based on an incompatible branch... ?
The files diff (https://github.com/FreeRDP/FreeRDP/pull/3126/files) only show my changes.
And I can guarantee that the channel is loading.

HenryJacques:
@akallabeth nevermind you're right :/
I'll change that.

HenryJacques:
Cancelled

PR #3129, opened 2016-02-14
If Android, get timezone from system command.

pxx02134:
In the case of Android, I changed to get timezone from Android's system command 'getprop persist.sys.timezone' if TZ environmental variable not set. It is because Android doesn't seem to have both "/var/db/zoneinfo" and "/etc/timezone" unlike BSD or Linux.

### DIFF
--- a/libfreerdp/locale/timezone.c
+++ b/libfreerdp/locale/timezone.c
@@ -1584,7 +1584,7 @@ char* freerdp_get_unix_timezone_identifier()
 		return tzid;
 	}
 
-	tzid = freerdp_get_unix_timezone_identifier_from_file(FILE* fp);
+	tzid = freerdp_get_unix_timezone_identifier_from_file();
 	if (tzid != NULL)
 	{
 		return tzid;
--- a/libfreerdp/locale/timezone.c
+++ b/libfreerdp/locale/timezone.c
@@ -1505,10 +1505,70 @@ static UINT64 freerdp_windows_gmtime()
 	return windows_time;
 }
 
+char* freerdp_read_unix_timezone_identifier_from_file(FILE* fp)
+{
+	char* tzid = NULL;
+	
+	fseek(fp, 0, SEEK_END);
+	length = ftell(fp);
+	fseek(fp, 0, SEEK_SET);
+
+	if (length < 2)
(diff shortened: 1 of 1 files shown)


hardening:
LGTM, I'm not an Android expert but the provided command seems to provide the same kind of value than the file we're usually scanning

bmiklautz:
@freerdp-bot test.

libfreerdp/locale/timezone.c
```diff
@@ -1523,7 +1523,9 @@ char* freerdp_get_unix_timezone_identifier()
 		return tzid;
 	}
 
-#if defined(__FreeBSD__) || defined(__OpenBSD__)
+#if defined(ANDROID)
+	fp = popen("getprop persist.sys.timezone", "r");
```
mfleisz: In case of popen we have to use pclose on Android later on to close the fp correctly.

bmiklautz: as @mfleisz said pclose should later be used if popen was used for opening the i/o stream. Otherwise the behavior is undefined.

libfreerdp/locale/timezone.c
```diff
@@ -1558,7 +1572,11 @@ char* freerdp_get_unix_timezone_identifier()
 		if (tzid[length - 1] == '\n')
 			tzid[length - 1] = '\0';
 
-		fclose(fp);
+#if defined(ANDROID)
+			pclose(fp) ;
```
akallabeth: Indentation wrong

akallabeth:
@pxx02134 Could you make the function static (it is only used within the file) and fix the indentation? Except these remarks +1

akallabeth:
@pxx02134 What did you do?

pxx02134:
@akallabeth I modified the following
 - created a function to read contents of a time zone file.
 - created a function to open/close a timezone file which is different from a kind of OSs.
 - fixed indentation.

akallabeth:
@pxx02134 ah, ok. could you sqash these commits and give them a meaningful message?

akallabeth:
@pxx02134 I've taken the liberty to refactor your pull. Could you check it that one works for you?

PR #3130, opened 2016-02-14
Update BookmarkDB.java

pxx02134:
There seems to be two commas forgot.


hardening:
+1

hardening:
@freerdp-bot I don't expect this to fail, but would give a build test anyway ?

PR #3145, opened 2016-02-21
Fix char signedness in glyph offset calculation, fixes #2137

### LINKED ISSUE #2137: Problem with rendering of vertical text in Excel 2007/2010 with /gdi:sw
* Connect to Windows 7 Pro (32 or 64bit) with `xfreerdp /gdi:sw`
  * Open Excel 2007 or 2010, type some text on a cell and rotate it 90° counterclockwise. You will see only the first letter of the text of the cell.

### With /gdi:sw
frdp_sw
### With /gdi:hw
frdp_hw

giox069:
I have just discovered the ["Shocking truth"](http://blog.cdleary.com/2012/11/arm-chars-are-unsigned-by-default/): under ARM gcc char data type is unsigned, when `-fsigned-char` is not used :blush:.
This patch makes an [older commit](https://github.com/FreeRDP/FreeRDP/commit/85b5c5f8909a0e660d8a1b0a23a63883771fa30f) to work independently of default char signedness.

### DIFF
--- a/libfreerdp/cache/glyph.c
+++ b/libfreerdp/cache/glyph.c
@@ -57,7 +57,7 @@ void update_process_glyph(rdpContext* context, BYTE* data, int* index,
 
 		if (offset & 0x80)
 		{
-			offset = data[*index + 1] | ((int)((char)data[*index + 2]) << 8);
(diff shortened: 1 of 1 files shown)


mfleisz:
+1

PR #3149, opened 2016-02-23
Misc fixes

bmiklautz:
* backport  (using /etc/TZ as alternative to /etc/localtime) from stable-1.1 (commit 605d0a85183d2f9d5b0da5629cbadf01173bbb2f)
* fix spelling


akallabeth:
+1 for the fixes, but the timezone patch collides with #3151

bmiklautz:
@hardening it's a one line I guess we will manager to update one of the conflicting pulls ;)

akallabeth:
@bmiklautz you think so? well, then lets merge ;)

PR #3150, opened 2016-02-23
server/shadow: Build libraries for shadow subsystems

xhaakon:
Commit 60ae27b0 decoupled subsystem implementations and shadow framework
core by moving subsystem specific code right into freerdp-shadow-cli.
This, however, doesn't allow applications that would like to link
libfreerdp-shadow, in order to embed RDP shadow server, to reuse also
the subsystem module.

3rd party developers now have to either provide their own subsystem code
(copied from FreeRDP sources or written from scratch) or be limited to
usaging freerdp-shadow-cli executable, which doesn't expose all
functions of the shadow server library (e.g. enumeration of available
monitors).

This change moves the shadow subsystem out of the executable into new
freerdp-shadow-subsystem library, which freerdp-shadow-cli and
potentially other applications can link to.

### DIFF
--- a/server/shadow/CMakeLists.txt
+++ b/server/shadow/CMakeLists.txt
@@ -92,10 +92,10 @@ endif()
 
 set_property(TARGET ${MODULE_NAME} PROPERTY FOLDER "Server/shadow")
 
-# command-line executable
+# subsystem library
 
-set(MODULE_NAME "freerdp-shadow-cli")
-set(MODULE_PREFIX "FREERDP_SERVER_SHADOW_CLI")
+set(MODULE_NAME "freerdp-shadow-subsystem")
+set(MODULE_PREFIX "FREERDP_SERVER_SHADOW_SUBSYSTEM")
 
 if(WIN32)
 	set(WITH_SHADOW_WIN 1)
@@ -225,9 +225,6 @@ if(WITH_SHADOW_MAC)
 	list(APPEND ${MODULE_PREFIX}_MAC_LIBS ${IOKIT} ${IOSURFACE} ${CARBON})
 endif()
 
-set(${MODULE_PREFIX}_SRCS
-	shadow.c)
-
 set(${MODULE_PREFIX}_WIN_SRCS
 	Win/win_rdp.c
 	Win/win_rdp.h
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test.

bmiklautz:
@freerdp-bot test

bmiklautz:
@xhaakon agreed on the change. Thanks.

However freerdp-shadow(-subsystem) isn't exported (pkg-config wise) at the moment  and for the debian package it was planned to be exported as part of freerdp2-shadow-x11 - but we probably should change this then.

bmiklautz:
+1

realjiangms:
@xhaakon This might break the shadow server build on windows. I used to decouple shadow.c and the  subsystem implementations. It worked fine on Mac and X11, but breaks windows. I had to at least add declaration with FREERDP_API on Win_ShadowSubsystemEntry. 

For consistency, I also add FREERDP_API declaration with X11_ShadowSubsystemEntry and the Mac one. But it was asked as not necessary in #2751   :)

We might need to recall declaration with FREERDP_API on Win_ShadowSubsystemEntry :)

xhaakon:
@bmiklautz 

> freerdp-shadow(-subsystem) isn't exported (pkg-config wise) at the moment - but we probably should change this then.

I'd very much appreciate that. There also isn't any header file which would declare `X11_ShadowSubsystemEntry` installed by the package. I always have to do it myself in my code like

`extern int X11_ShadowSubsystemEntry(RDP_SHADOW_ENTRY_POINTS* pEntryPoints);`

Perhaps `server/shadow.h` could include that declaration?

xhaakon:
@realjiangms Well, the automated build on Windows has passed, so I'm also not quite sure :)

Should I update my change right away, just in case, or is there anyone who could say for certain if FREERDP_API is needed?

bmiklautz:
@xhaakon regarding headers I think when @realjiangms did the decoupling it wasn't thought to be exported. Anyway that's something we can simply fix.

Regarding the windows build unfortunately the ci build don't build WITH_SERVER=ON at the moment therefore this code path wasn't tested on windows. I totally forgot about that before merging :(. @realjiangms is of course right and the changes breaks on windows.

realjiangms:
@bmiklautz  Last time we decided not to export the default subsystem implementations. But I would +1 if we want to export them because as @xhaakon mentioned that some 3rd party software might want to use the default implementation directly.

@xhaakon The correct way might be: Add FREERDP_API  declaration in include/freerdp/server/shadow.h if we really want to export them. Then you don't need to declare them in your code :)

bmiklautz:
@realjiangms makes perfect sense. Let's export it. How do you use/link in shadow in your app?
@xhaakon what interface are you using for your application?

I'll can have a look but it would be nice to have some use cases ;).

realjiangms:
I just added 3 decleareation in include/freerdp/server/shadow.h, tried compile on my existing build machine. X11 still fine, however windows failed earlier:

1>win_rdp.obj : error LNK2019: unresolved _freerdp_client_load_addins，in _shw_pre_connect 
1>win_rdp.obj : error LNK2019: unresolved  _freerdp_client_context_new，in _win_shadow_rdp_init 
1>win_rdp.obj : error LNK2019: unresolved  _freerdp_client_start，in _win_shadow_rdp_start 
1>win_rdp.obj : error LNK2019: unresolved  _freerdp_client_stop，in _win_shadow_rdp_stop 

I might need more time to fix it :)

@bmiklautz I write another subsystem implementation for my app, my app need to have an RDP server side interface.It's completely different with X11. I think @xhaakon works with X11_ShadowSubsystemEntry directly.

bmiklautz:
@realjiangms I thought so therefore we might want to expose both interfaces. Maybe we should define a "generic" interface that wraps the subsystems as well.

xhaakon:
@bmiklautz
>what interface are you using for your application?

Not sure what you are asking about, but I pass `X11_ShadowSubsystemEntry` to `shadow_subsystem_set_entry()` and then use the functions from freerdp/server/shadow.h (shadow_enum_monitors(), shadow_server_new(), ...) to interact with the server.

realjiangms:
@bmiklautz Do you mean a function like Default_ShadowSubsystemEntry? It would be fine currently but we might need to consider more than 1 'built-in' subsystem implementation future.

If we don't want to export the detail internal API, we can recall the old style to set an 'built-in' pfnShadowSubsystemEntry implementation by internal name. How about this:

FREERDP_API void shadow_subsystem_set_entry_built_in(const char* name);

And track an internal built-in table like:

static RDP_SHADOW_SUBSYSTEM g_Subsystems[] =
{

#ifdef WITH_SHADOW_X11
       { "X11", X11_ShadowSubsystemEntry },
#endif

#ifdef WITH_SHADOW_MAC
       { "Mac", Mac_ShadowSubsystemEntry },
#endif

#ifdef WITH_SHADOW_WIN
       { "Win", Win_ShadowSubsystemEntry },
#endif

       { "", NULL }
};

The similar code could be found from diff between 202a93 and 131f03.

But anyway I need to investigate on new unresolved / unexported symbol for windows. The issue seems to be introduced before this PR

realjiangms:
I understood the other 4 unresolved symbol. They are exported by freerdp-client. The subsystem library should link to freerdp-client for WIN32.

bmiklautz:
@realjiangms I like the idea. @xhaakon what do you think? We could also export WITH_SHADOW_xyz in build_config.h so by including it can also be checked if the it's available.
I'll have a look at it today or tomorrow.

bmiklautz:
@realjiangms regarding the linking against freerdp-client - I already got a local branch where this is fixed.

xhaakon:
@bmiklautz I'm fine with what @realjiangms suggests too. Changes to my code will be only trivial.

realjiangms:
OK I'll post a PR for it.

@bmiklautz regarding the linking against freerdp-client, Have you posted the PR?

realjiangms:
Created PR#3167. @bmiklautz I didn't include the fix for linking freerdp-client. It would be fixed when you post the fix in your local branch

bmiklautz:
@realjiangms nope haven't created the PR yet but I'll do it based on your last PR.

bmiklautz:
@realjiangms #3167

PR #3151, opened 2016-02-23
Timezone refactor

akallabeth:
* Moves timezone detection to winpr
* Superseeds #3129


akallabeth:
@freerdp-bot could you be a nice bot once and test this properly?

libfreerdp/core/timezone.c
```diff
@@ -76,34 +80,26 @@ void rdp_write_system_time(wStream* s, SYSTEM_TIME* system_time)
...
 {
-	char* str = NULL;
-	TIME_ZONE_INFO* clientTimeZone;
+	LPTIME_ZONE_INFORMATION tz;
 
 	if (Stream_GetRemainingLength(s) < 172)
 		return FALSE;
 
-	clientTimeZone = settings->ClientTimeZone;
+	tz = settings->ClientTimeZone;
```
bmiklautz: Not in the old code but maybe a check for setting->ClientTimeZone makes sense before accessing the members.

libfreerdp/core/timezone.c
```diff
@@ -117,53 +113,37 @@ BOOL rdp_read_client_time_zone(wStream* s, rdpSettings* settings)
...
-	freerdp_time_zone_detect(clientTimeZone);
-
-	standardNameLength = ConvertToUnicode(CP_UTF8, 0, clientTimeZone->standardName, -1, &standardName, 0) * 2;
-	daylightNameLength = ConvertToUnicode(CP_UTF8, 0, clientTimeZone->daylightName, -1, &daylightName, 0) * 2;
+	LPTIME_ZONE_INFORMATION tz;
+	DWORD rc;
 
-	if (standardNameLength > 62)
-		standardNameLength = 62;
+	tz = settings->ClientTimeZone;
```
bmiklautz: Same as above.

libfreerdp/core/timezone.c
```diff
@@ -117,53 +113,37 @@ BOOL rdp_read_client_time_zone(wStream* s, rdpSettings* settings)
...
+	rc = GetTimeZoneInformation(tz);
```
bmiklautz: I might be mistaken but I think GetTImeZoneInformation isn't implemented (as in return 0).

bmiklautz:
If timezone.c is moved around I would also adapt the functions otherwise we do have freerdp_ in winpr.

libfreerdp/core/timezone.c
akallabeth: @bmiklautz This pull is eager to change that.

akallabeth:
@bmiklautz Ok, renamed the functions and added argument checks.

mfleisz:
+1

PR #3152, opened 2016-02-23
Fix a freeze when receiving a deactivateAll PDU

davewheel:
When connecting on FUDO, the xfreerdp client gets stuck when receiving a
deactivateAll PDU. It seems that rdp_recv_deactivate_all() was doing nasty
things as it is called from the receive callback, so when it was calling rdp_check_fds(),
the callback was called again, opening a potential recursive loop. The activation
state machine seems correctly implemented so just returning looks ok and fixes
connections on FUDO hosts.

### DIFF
--- a/libfreerdp/core/activation.c
+++ b/libfreerdp/core/activation.c
@@ -311,18 +311,7 @@ BOOL rdp_recv_deactivate_all(rdpRdp* rdp, wStream* s)
 		while(0);
 	}
 
-	rdp_client_transition_to_state(rdp, CONNECTION_STATE_CAPABILITIES_EXCHANGE);
-
-	while (rdp->state != CONNECTION_STATE_ACTIVE)
-	{
-		if (rdp_check_fds(rdp) < 0)
-			return FALSE;
-
-		if (freerdp_shall_disconnect(rdp->instance))
-			break;
-	}
-
-	return TRUE;
+	return rdp_client_transition_to_state(rdp, CONNECTION_STATE_CAPABILITIES_EXCHANGE) != -1;
 }
 
 BOOL rdp_send_deactivate_all(rdpRdp* rdp)


hardening:
@freerdp-bot the patch makes sense, can you test it ?

bmiklautz:
@davewheel what's "FUDO"?

bmiklautz:
@davewheel it might be that the fix helps with "FUDO" (whatever it is) but I'm not convinced that it's the right fix. While a deactivate-activate sequence is in progress no other deactivation should be received and the client needs to process messages until the sequence completed and the client is active again.
What issue you are exactly hitting? Did you check the messages you received while the deactivation is pending?

-1

davewheel:
FUDO is the product from WheelSystems.

When connecting on a FUDO host (I haven't find why yet), xfreerdp gets stuck when calling again rdp_check_fds(). 
If we don't expect another deactivateAll PDU, then perhaps we could set a bool (not expecting deactivate) and let the state machine do its job ?

bmiklautz:
@davewheel thanks for explaining. Does the FUDO host have an own RDP implementation?

davewheel:
Kinda, it's based on the FreeRDP server code.

bmiklautz:
@davewheel in this case could you figure when exactly this behavior was triggered. We had some similar problems with FreeRDS back then but those only happened in cases when another deactivate-activate sequence was triggered while one was already pending. This could for example happen if the screen resolution changes rapidly and DesktopResize is called while a re-size (which in background triggers a deactivate-activate sequence) was already "running".

hardening:
I was curious about what was saying the spec about the deactivate PDU, and it seems like it can be received:
* during the initial capability + connection finalization phase;
* when we're already in the Active state

So perhaps the fix is to ensure that we're in these two cases. The rdp_check_fds() call in rdp_recv_deactivate_all() is totally insane, as we're called from the received callback that is called by ... rdp_check_fds. So I'm quite sure there's some place for freeze !

bmiklautz:
@hardening yes, the fix is insane. And is by all means not a good and elegant solution (tm). On the other hand it ensures that the client doesn't send out other messages (like key events for example) during the sequence because this can cause disconnects or so on the server side.

hardening:
So that's a client problem only. Looking at the code, it looks like the first negociation is treated as a special case, and then we go in the regular loop. But the desactivateAll PDU can put us back in the state were a part of the negociation must be re-done, and so managing this totally done in rdp_recv_deactivate_all. My suggestion is that we should return an error status that would say to the regular loop "handshake in progress, do not proceed inputs until we're active again" (just like we're returning some special codes for disconnection or redirection). Does it look sane like this ?

bmiklautz:
@hardening sounds like a good idea and might be the best solution but not sure if that is possible for all functions invoked.

PR #3157, opened 2016-02-25
Fix bad malloc size in statvfs

DavBfr:
If using unicode on windows client, the allocated data is too small.

### DIFF
--- a/channels/drive/client/statvfs.c
+++ b/channels/drive/client/statvfs.c
@@ -3,6 +3,8 @@
  * statvfs emulation for Windows
  *
  * Copyright 2012 Gerald Richter
+ * Copyright 2016 Inuvika Inc.
+ * Copyright 2016 David PHAM-VAN <d.phamvan@inuvika.com>
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


mfleisz:
@freerdp-bot test

channels/drive/client/statvfs.c
```diff
@@ -36,7 +38,7 @@ int statvfs(const char *path, struct statvfs *buf)
 	DWORD lpTotalNumberOfClusters;
 
 	len = MultiByteToWideChar(CP_ACP, 0, path, -1, NULL, 0);
-	unicodestr = (LPWSTR) malloc(len);
+	unicodestr = (LPWSTR) malloc(len*sizeof(TCHAR));
```
mfleisz: I think this should be sizeof(WCHAR) rather than sizeof(TCHAR) - otherwise we still have a buffer overrun when we compile with TCHAR=CHAR

DavBfr:
@mfleisz your right, and I forced the unicode version of _GetDiskFreeSpace()_.

channels/drive/client/statvfs.c
```diff
@@ -36,10 +38,10 @@ int statvfs(const char *path, struct statvfs *buf)
 	DWORD lpTotalNumberOfClusters;
 
 	len = MultiByteToWideChar(CP_ACP, 0, path, -1, NULL, 0);
-	unicodestr = (LPWSTR) malloc(len);
+	unicodestr = (LPWSTR) malloc(len*sizeof(WCHAR));
```
akallabeth: @DavBfr You can simplify the whole thing by using ```ConvertToUnicode``` from winpr. Either way you need to check for errors and handle them.

channels/drive/client/statvfs.c
```diff
@@ -29,17 +31,18 @@ int statvfs(const char *path, struct statvfs *buf)
...
 	DWORD lpBytesPerSector;
 	DWORD lpNumberOfFreeClusters;
 	DWORD lpTotalNumberOfClusters;
 
-	len = MultiByteToWideChar(CP_ACP, 0, path, -1, NULL, 0);
-	unicodestr = (LPWSTR) malloc(len);
-	MultiByteToWideChar(CP_ACP, 0, path, -1, unicodestr, len);
+	len = ConvertToUnicode(CP_ACP, 0, path, -1, &unicodestr, 0);
+	if (len < 0)
+		return -1;
```
mfleisz: Style: tab missing

mfleisz:
Aside of @akallabeth's and my remark +1

channels/drive/client/statvfs.c
DavBfr: @mfleisz Where is there a missing tab?

PR #3160, opened 2016-02-25
Stream fixes

akallabeth:
* Fixes a missing stream copy in ```update_message_SurfaceCommand```
* Refactors direct access to stream pointer, use wrapping functions.
* Made stream defines inline functions for type safety.

### DIFF
--- a/libfreerdp/core/message.c
+++ b/libfreerdp/core/message.c
@@ -192,7 +192,7 @@ static BOOL update_message_SurfaceCommand(rdpContext* context, wStream* s)
 	if (!wParam)
 		return FALSE;
 
-	Stream_Copy(wParam, s, Stream_GetRemainingLength(s));
+	Stream_Copy(s, wParam, Stream_GetRemainingLength(s));
 	Stream_SetPosition(wParam, 0);
 
 	return MessageQueue_Post(context->update->queue, (void*) context,
--- a/libfreerdp/core/message.c
+++ b/libfreerdp/core/message.c
@@ -192,7 +192,7 @@ static BOOL update_message_SurfaceCommand(rdpContext* context, wStream* s)
 	if (!wParam)
 		return FALSE;
 
-	Stream_Copy(wParam, s, Stream_Capacity(s));
+	Stream_Copy(wParam, s, Stream_GetRemainingLength(s));
 	Stream_SetPosition(wParam, 0);
 
 	return MessageQueue_Post(context->update->queue, (void*) context,
--- a/libfreerdp/core/message.c
+++ b/libfreerdp/core/message.c
@@ -188,19 +188,12 @@ static BOOL update_message_SurfaceCommand(rdpContext* context, wStream* s)
 {
 	wStream* wParam;
 
-	wParam = (wStream*) malloc(sizeof(wStream));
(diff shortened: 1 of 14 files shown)


akallabeth:
@freerdp-bot Come on, test this please!

hardening:
I like this patch, anyway I heard sometime ago, someone (can't remember who) saying that the static inline function in headers was not always working as expecting with compilers (perhaps it was something with some VC versions). And that the define were the only thing that would work everywhere.
So that person shall talk now or remain silent forever.

For me it's +1

bmiklautz:
I'm basically in favor of inline functions but before we merge it would be interesting to do some bench marking with and without the changes. As for inline the compile is "free" do do it or not iirc.

mfleisz:
+1

mfleisz:
If the compiler is not inlining the functions we could add a new macro FORCEINLINE - most compilers should have some way to declare a function to be inlined no matter what the compiler thinks

hardening:
I'm quite sure static inline in inlined at least by gcc, that's what they use everywhere in the Linux kernel (I guess that's the case too for clang). Don't know for VC.

libfreerdp/core/message.c
```diff
@@ -188,19 +188,12 @@ static BOOL update_message_SurfaceCommand(rdpContext* context, wStream* s)
 {
 	wStream* wParam;
 
-	wParam = (wStream*) malloc(sizeof(wStream));
+	wParam = Stream_New(NULL, Stream_GetRemainingLength(s));
 	if (!wParam)
```
nfedera: Shouldn't this be Stream_Capacity(s) ?

akallabeth: @nfedera No, ```Stream_GetRemainingLength``` as in copy the rest of the stream to a new one.

winpr/include/winpr/stream.h
```diff
@@ -50,35 +50,59 @@ WINPR_API BOOL Stream_EnsureRemainingCapacity(wStream* s, size_t size);
...
+}
+
+static INLINE void Stream_Rewind(wStream* s, size_t _offset)
+{
+	struct _wStream* _s = (struct _wStream*)s;
+	_s->pointer -= (_offset);
+}
+
+#define _stream_read_n8(_t, __s, _v, _p) do { \
+	struct _wStream* _s = (struct _wStream*)__s; \
```
nfedera: Why not use _s directly as it was before ?

akallabeth: @nfedera I've left this in for testing purposes. This way it is quite easy to add ```typedef void* wStream``` and still have the whole thing compile (but complain about direct access to struct members)

nfedera: @akallabeth An opaque wStream type would be nice but since there can't be a type that is both opaque and useable in inline functions I'd suggest to revert these.

nfedera:
@akallabeth maybe a good time to finish the missing stream ctests

akallabeth:
@nfedera Ok, new tests and a little change to ```Stream_Copy``` API, now uses the same syntax as every other function (that is instance first argument, not second)

nfedera:

nfedera:
+1

bmiklautz:
+1

Issue #3007, opened 2015-12-09
Printer channel driver name "hardwired"


szolnoki:
I tested "This is FreeRDP version 2.0.0-dev (git n/a)" version.

I started freerdp with option:
/a:printer,HP4650_PCL,"HP Universal Printing PCL 6"

But HP_4650_PCL always use "Microsoft Publisher Imagesetter" driver on remote Windows. 
("HP Universal Printing PCL 6" installed on remote Windows.)

My hotfix: 
I changed "printer_cups.c"...
From:
cups_printer->printer.driver = "MS Publisher Imagesetter";
To:
cups_printer->printer.driver = "HP Universal Printing PCL 6";

And now, HP_4650 works at remote Windows with the "HP Universal Printing PCL 6" driver! :):):)
My solution works, but not nice... Why not use the specified driver name in command line, instead of default "Imagesettler"...

Why printer drive name "hardwired", when I set this option in command-line?
This problem is independent from printer type.

I would be better, if freerdp use the given driver name if specified instead of "Imagesettler".

bmiklautz:
@szolnoki did you try with the /printer command line already? 
```
/printer:<device>,<driver>
```
I'm not sure that that the "addin syntax" (using /a) parses additional parameters if an extra parameter (like /printer) exists.

szolnoki:
I have tested. But not working. Absolutely same effect as "/a:printer". :(
Driver name only changed, when I modify the source code...
I think not possible to change driver name any other way. Because I searched "printer.driver" string in all source codes. "printer.driver" variable is set only one place, see the "hardwired" line above.

szolnoki:
I continue to investigate this problem, and I found this.
Compiled with cmake options:
-DWITH_CUPS=ON -DCHANNEL_PRINTER=ON

Version: 2.0.0-dev (git df81e84)

Started the xfreerdp with option again and again:
/printer:HP4650_PCL,"HP Universal Printing PCL 6"

In remote Windows Devices&Printer settings, I see:
1. The HP4650_PCL appear with driver "HP Universal Printing PCL 6"
2. Some seconds later HP4650_PCL above disappear
3. Some seconds later HP4650_PCL appear again, but with "MS Publisher Imagesetter" driver.
:(:(:(:(

I added some extra logging "printf" to freeRDP printer channel driver, and I found the problem:
1. I see in log: "[INFO][com.freerdp.channels.rdpdr.client] - Loading device service printer (static)"
[...]

PR #3165, opened 2016-02-26
Printer fix

### LINKED ISSUE #2269: Printer redirection driver defaults to "MS Publisher Imagesetter" even when specified
Using master branch and compiling on Mac OS X 10.9.5 with the following,

```
cmake -D "CMAKE_OSX_ARCHITECTURES:STRING=x86_64" \
"-DWITH_CUPS=ON" . \
&& make
```

The following tests were done connecting to a vanilla Windows Server 2012 R2 server. I've also tested with Windows 7 and Windows XP and got the same results.

## Reference printer drivers installed by default

Before **any** printer redirection the following printer drivers are present on the server.
screen shot 2014-12-10 at 18 34 49

## Test 1 - Printer Redirection without specifying driver

```./MacFreeRDP /printer:HP_LaserJet_P3010_Series```

In this case FreeRDP uses the "MS Publisher Imagesetter" driver, which gets installed automatically on the server.

screen shot 2014-12-10 at 18 36 06

## Test 2 - Printer Redirection specifying driver

```./MacFreeRDP /printer:HP_LaserJet_P3010_Series,"HP LaserJet P3011/P3015 PCL6 Class Driver"```

In this case FreeRDP causes two drivers to be installed: "MS Publisher Imagesetter" and the specified driver "HP LaserJet P3011/P3015 PCL6 Class Driver".

However, the "MS Publisher Imagesetter" driver is the one used for the redirected printer and **not** the specified HP printer driver.

screen shot 2014-12-10 at 19 57 38

## Workaround

By hardcoding the correct printer driver I have got printer redirection working as intended, e.g. the correct driver is installed and set for the printer.

This is done by changing the following line,
https://github.com/FreeRDP/FreeRDP/blob/master/channels/printer/client/printer_cups.c#L228

```
diff --git a/channels/printer/client/printer_cups.c b/channels/printer/client/printer_cups.c
index 0d6fcbe..5507031 100644
--- a/channels/printer/client/printer_cups.c
+++ b/channels/printer/client/printer_cups.c
@@ -225,7 +225,7 @@ static rdpPrinter* printer_cups_new_printer(rdpCupsPrinterDriver* cups_driver, c
        cups_printer->printer.id = cups_driver->id_sequence++;
        cups_printer->printer.name = _strdup(name);
        /* This is a generic PostScript printer driver developed by MS, so it should be good in most cases */
-       cups_printer->printer.driver = "MS Publisher Imagesetter";
+       cups_printer->printer.driver = "HP LaserJet P3011/P3015 PCL6 Class Driver";
        cups_printer->printer.is_default = is_default;
 
        cups_printer->printer.CreatePrintJob = printer_cups_create_printjob;
```

## Solution

Ideally when specifying a specific printer driver, e.g.

```./MacFreeRDP /printer:HP_LaserJet_P3010_Series,"HP LaserJet P3011/P3015 PCL6 Class Driver"```

FreeRDP would a) not install "MS Publisher Imagesetter" and b) set the printer driver to that which was specified

Also, although I have used `MacFreeRDP` for my testing I get the exact same results when just using `xfreerdp`.

akallabeth:
Fixes #2269 and #3007

### DIFF
--- a/channels/printer/client/printer_cups.c
+++ b/channels/printer/client/printer_cups.c
@@ -5,6 +5,7 @@
  * Copyright 2010-2011 Vic Lee
  * Copyright 2015 Thincast Technologies GmbH
  * Copyright 2015 DI (FH) Martin Haimberger <martin.haimberger@thincast.com>
+ * Copyright 2016 Armin Novak <armin.novak@gmail.com>
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -229,10 +230,12 @@ static void printer_cups_free_printer(rdpPrinter* printer)
 		cups_printer->printjob->printjob.Close((rdpPrintJob*) cups_printer->printjob);
 
 	free(printer->name);
+	free(printer->driver);
 	free(printer);
 }
 
-static rdpPrinter* printer_cups_new_printer(rdpCupsPrinterDriver* cups_driver, const char* name, BOOL is_default)
+static rdpPrinter* printer_cups_new_printer(rdpCupsPrinterDriver* cups_driver,
+	const char* name, const char* driverName, BOOL is_default)
 {
 	rdpCupsPrinter* cups_printer;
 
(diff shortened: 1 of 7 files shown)


hardening:
Nice job, I have been working on this but I couldn't find a nice way to do it. I like your solution +1

hardening:
Tested and working as expected.

Note for future work: we should disable the printer channel if we don't have any driver compiled in. And we should probably not fail and shutdown connection in this case (not even saying that the corresponding command lines should not be active). I get caught for the second time, to finally realize that xfreerdp was closing because no printer driver was activated.

mfleisz:
+1

PR #3167, opened 2016-02-26
server/shadow: Export API to set built-in subsystem modules for freerdp-shadow-subsystem

realjiangms:
Fix compile error for server on windows which we discussed in #3150


akallabeth:
@realjiangms This will not work if function visibility is activated (currently working on a patch for that). You need to ```FREERDP_API``` the subsystem function.

akallabeth:
@realjiangms Ok, never mind, fixed that in #3168

realjiangms:
Thanks @akallabeth . The related fix in #3168 resolves the compile issue.
However in #3150 we decide still not export (by default) Win_ShadowSubsystemEntry/X11_ShadowSubsystemEntry/Mac_ShadowSubsystemEntry as they are internal implementation. 

So in this PR we introduce API shadow_subsystem_set_entry_builtin to select the built-in subsystem entries.

akallabeth:
@realjiangms Ok, good. As stated in #3168 for the time being I need to export those functions there. Depending on which one is merged first this or my pull need to be updated.

akallabeth:
LGTM +1 btw

bmiklautz:
+1
@realjiangms nice thanks!
@xhaakon anything still missing for you use case after merging this?

xhaakon:
@bmiklautz this should be sufficient for my usage scenario, thanks!

I'd like to do some more thorough testing once you update your Debian package.

bmiklautz:
@xhaakon Great! I'm updating the package this moment. Comment here once I'm done.

bmiklautz:
@xhaakon pushed the latest freerdp2 package right now. Upstream was merged. There is now a libfreerdp-shadow2 package and freerdp-shadow is "exported" with pkg-config and cmake (in the package freerdp2-dev).

xhaakon:
@bmiklautz Thanks, I've checked the package. It builds and works well for me, but I still found some smaller issues and created separate PRs for them:

https://github.com/FreeRDP/FreeRDP/pull/3182
https://github.com/bmiklautz/debian-freerdp2/pull/3

PR #3097, opened 2016-02-01
Android api upgrade and restructuring

akallabeth:
# What has changed:
* Fixes (some) return value checks for comand line parser
* Fixes wlog and environment issues with android
* Building shared libraries on android now
* Migrated to gradle (android studio) for android java code
* Added build scripts for native libraries
** Now supports ```armeabi armeabi-v7a and x86``` by default
** More architectures as soon as openssl builds there

# Usage:
* run ```./scripts/android-build-freerdp.sh```
* Import ```client/Android/Studio``` in Android Studio
* You're ready to go

Issue #2720, opened 2015-06-23
how to call aFreeRDP in another Android APP to connect to a remote App ? 


cloudfirst:
I am working in an android app that get the hostname, user name and password, as well as remote program name, and use intent mechanism to invoke session activity in aFreeRDP.

however, I can't find any tech document on how to finish this invocation .
can any one help on this ?

bmiklautz:
@cloudfirst there is no real documentation on the Android client at the moment. Have a look to FreeRDP/client/Android/FreeRDPCore to see how the intent(s).

If you have further questions visit our irc channel (#freerdp on freenode) or write a mail to the mailing list but please don't add questions in the issue tracker.

Issue #99, opened 2011-09-14
libfreerdp-gdi ScrBlt


awakecoding:
When multiple ScrBlt messages are fired by the server in a very fast sequence, libfreerdp-gdi starts rendering graphics improperly. This bug does not seem to affect the X11-specific GDI implementation. This bug was observed with the Windows 8 Developer Preview in RemoteFX mode, where the only non-RemoteFX message received was ScrBlt.

PR #3172, opened 2016-02-28
android: Allow freerdp mobile version to be launched from URI (freerdp://)

realjiangms:
It would be good if we have a easy way to call aFreeRDP in another Android APP (Requirement also mentioned in #2720)
We can define a scheme (freerdp://) as unified way to launch FreeRDP from another APP or browser and connect to compatible RDP server
1. Define scheme freerdp://
2. General form could be freerdp://user@hostname:port/title?key1=value&key2=-&key3=%2b&key4=
3. [user] part would be translated to /u:
4. [hostname:port] would be translated to /v:
5. The [user@hostname:port] part would be used as app title, currently it's just the progress dialog title
6. query parameters would be translated to command line arguments. Later same arguments will overwrite the formers:
    a. key1=value: 		=> /key1:value
    b. key2=-: 			=> -key2
    c. key3=%2b 		=> +key3 (%2b is url encoded +)
	d. key4= 			=> /key4
	e. Especially, drive=sdcard will be properly handled with local sdcard path. On my device it will be translated to /drive:sdcard,/storage/emulated/0

Owing to the refactor work in PR #3097, we now pass same command line argument to JNI for freerdp settings.
We just need to make the SessionActivity accept freerdp scheme and translate argument from URI form to command line form.

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/presentation/SessionActivity.java
+++ b/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/presentation/SessionActivity.java
@@ -29,6 +29,7 @@ import android.graphics.Rect;
 import android.graphics.drawable.BitmapDrawable;
 import android.inputmethodservice.Keyboard;
 import android.inputmethodservice.KeyboardView;
+import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
@@ -238,6 +239,11 @@ public class SessionActivity extends ActionBarActivity implements
 			if (progressDialog != null) {
 				progressDialog.dismiss();
 				progressDialog = null;
+			}
+
+			if (session.getBookmark() == null) {
+				// Return immediately if we launch from URI
+				return;
 			}
 
 			// add hostname to history if quick connect was used
@@ -614,9 +620,14 @@ public class SessionActivity extends ActionBarActivity implements
 	}
 
 	private void processIntent(Intent intent) {
-		// get either session instance or create one from a bookmark
+		// get either session instance or create one from a bookmark/uri
 		Bundle bundle = intent.getExtras();
-		if (bundle.containsKey(PARAM_INSTANCE)) {
+		Uri openUri = intent.getData();
+		if (openUri != null) {
+			// Launched from URI, e.g:
+			// freerdp://user@ip:port/connect?sound=&rfx=&p=password&clipboard=%2b&themes=-
+			connect(openUri);
+		} else if (bundle.containsKey(PARAM_INSTANCE)) {
 			int inst = bundle.getInt(PARAM_INSTANCE);
 			session = GlobalApp.getSession(inst);
 			bitmap = session.getSurface().getBitmap();
@@ -648,7 +659,6 @@ public class SessionActivity extends ActionBarActivity implements
 
 	private void connect(BookmarkBase bookmark) {
 		session = GlobalApp.createSession(bookmark, getApplicationContext());
-		session.setUIEventListener(this);
 
 		BookmarkBase.ScreenSettings screenSettings = session.getBookmark()
 				.getActiveScreenSettings();
@@ -673,8 +683,20 @@ public class SessionActivity extends ActionBarActivity implements
 			screenSettings.setWidth(screen_width);
 		}
 
+		connectWithTitle(bookmark.getLabel());
+	}
+
+	private void connect(Uri openUri) {
+		session = GlobalApp.createSession(openUri, getApplicationContext());
+
+		connectWithTitle(openUri.getAuthority());
+	}
+
+	private void connectWithTitle(String title) {
+		session.setUIEventListener(this);
+
 		progressDialog = new ProgressDialog(this);
-		progressDialog.setTitle(bookmark.getLabel());
+		progressDialog.setTitle(title);
 		progressDialog.setMessage(getResources().getText(
 				R.string.dlg_msg_connecting));
 		progressDialog.setButton(ProgressDialog.BUTTON_NEGATIVE, "Cancel",
@@ -699,7 +721,7 @@ public class SessionActivity extends ActionBarActivity implements
 	// binds the current session to the activity by wiring it up with the
 	// sessionView and updating all internal objects accordingly
 	private void bindSession() {
-		Log.v("SessionActivity", "bindSession called");
+		Log.v(TAG, "bindSession called");
 		session.setUIEventListener(this);
 		sessionView.onSurfaceChange(session);
 		scrollView.requestLayout();
@@ -971,6 +993,11 @@ public class SessionActivity extends ActionBarActivity implements
 			bitmap = Bitmap.createBitmap(width, height, Config.RGB_565);
 
 		session.setSurface(new BitmapDrawable(bitmap));
+
+		if (session.getBookmark() == null) {
+			// Return immediately if we launch from URI
+			return;
+		}
 
 		// check this settings and initial settings - if they are not equal the
 		// server doesn't support our settings
--- a/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
+++ b/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
@@ -17,6 +17,7 @@ import com.freerdp.freerdpcore.domain.ManualBookmark;
 
 import android.content.Context;
 import android.graphics.Bitmap;
+import android.net.Uri;
 import android.util.Log;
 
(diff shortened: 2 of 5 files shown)


akallabeth:
@realjiangms Nice! Do you have a simple test app to check if it works as expected?

realjiangms:
I do have a test app built with cordova but it is hard-coded to my local environment. Let me clean it up with single test box and submit botton. Do you know if we can upload some attachment here?

akallabeth:
@realjiangms File upload should work now, just not tested so far... Well, there is a first time for everyting ;)

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
```diff
@@ -273,6 +275,61 @@ public static boolean setConnectionInfo(int inst, BookmarkBase bookmark) {
         String[] arrayArgs = args.toArray(new String[args.size()]);
         return freerdp_parse_arguments(inst, arrayArgs);
     }
+    
+    public static boolean setConnectionInfo(int inst, Uri openUri) {
+        ArrayList<String> args = new ArrayList<String>();
+        HashMap<String, String> map = new HashMap<String, String>();
```
mfleisz: Why are you using a HashMap here instead of directly feeding the parameters into the args list?

realjiangms: For argument overwriting. For example, the later /v argument could overwrite former, which is defined by host:port in uri

realjiangms: I realize it might be not necessary. The parse argument function actually already support overwrite for same argument key.
I will do more test to double check

realjiangms:
@akallabeth Finally I leverage the github page for test: http://realjiangms.github.io/Testsite/freerdp.html
Visit it in browser in Android.Change the freerdp url in the test box and click the button. It will launch aFreerdp if a version with this PR is installed.

By the way, I think it might not be good idea to use entire path component as title. It should be reserved for future extension. The title could just be the entire url.

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/presentation/SessionActivity.java
```diff
@@ -614,9 +620,15 @@ public void onConfigurationChanged(Configuration newConfig) {
...
 	private void processIntent(Intent intent) {
-		// get either session instance or create one from a bookmark
+		// get either session instance or create one from a bookmark/uri
 		Bundle bundle = intent.getExtras();
-		if (bundle.containsKey(PARAM_INSTANCE)) {
+		Uri openUri = intent.getData();
+		if (openUri != null) {
+			// Launched from URI, e.g:
+			// freerdp://user@ip:port/title?sound=&rfx=&p=password&clipboard=%2b&themes=-
+			Log.v(TAG, "Open from URI: " + openUri.toString());
```
akallabeth: This log message is propably dangerous, the URI may contain a password / token

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
```diff
@@ -273,6 +275,61 @@ public static boolean setConnectionInfo(int inst, BookmarkBase bookmark) {
...
+        
+        String user = openUri.getUserInfo();
+        if (user != null) {
+            map.put("u", "/u:" + user);
+        }
+        
+        for (String key: openUri.getQueryParameterNames()) {
+            String value = openUri.getQueryParameter(key);
+            
+            if (value.isEmpty()) {
```
akallabeth: There should be some kind of sanity check to avoid invalid arguemnts (e.g. must be printable char, the keywords are known, ...)

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/presentation/SessionActivity.java
realjiangms: Yes I'm fixing it  ;)

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
realjiangms: The underlying freerdp argument parser would check the arguments. We just simply parse arguments here as if we are calling xfreerdp. Keywords might be extended always and don't want to limit it in aFreeRDP

akallabeth:
Other than my remarks this is working fine. +1

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
realjiangms: I discarded hashmap. It is not necessary

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
akallabeth: @realjiangms Your argument is true, I'm just not comfortable with unchecked full unicode strings from an unknown source. (think separators and escaping them)

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
```diff
@@ -273,6 +274,69 @@ public static boolean setConnectionInfo(int inst, BookmarkBase bookmark) {
...
+        StringBuilder sbArguments = new StringBuilder();
+        for (String argument: args) {
+            if (argument.startsWith("/p:")) {
+                sbArguments.append("/p:*");
+            } else {
+                sbArguments.append(argument);
+            }
+            sbArguments.append(" ");
+        }
+        Log.v(TAG, "Connecting with arguments: " + sbArguments.toString());
```
akallabeth: @realjiangms I'd just remove the logging here entirely, there is not only ```/p``` to hide.

realjiangms: Yeah you are exactly correct. I just checked command line and find following password arguments.
    /pth:<password hash>        Pass the hash (restricted admin mode)
    /p:<password>               Password
    /gp:<password>              Gateway password
    /assistance:<password>      Remote assistance password

I haven't notice the new command line argument these days :). I agree to totally discard android logging for arguments. Because we don't know whether there will be more arguments future.

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
realjiangms: Unknown source with unknown information always looks danger :) But we don't need to worry too much about it because underlying command argument parser will also check. And the launched APP will have same uid with original APP. Even if someone could hack freerdp by some argument (in case we really have bug !-_-), he can't get more permission (I agree there might be some unexpected behavior being allowed unexpectly, though.

On the other hand, bash never do more check for us when someone type 'xfreerdp' in bash console ;)

realjiangms:
I'll post again later with some more fix and include the comments from @akallabeth

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
akallabeth: @realjiangms True, on the other hand bash is not triggered by a link in a brower ;) Think dial URL issues with android and ```*#99``` et al

realjiangms: Yeah link from browser is not 'expected' by the one who is using the browser ;)

akallabeth:
@freerdp-bot test

akallabeth:
@realjiangms I don't like the move of the command line creation out of the ```FreeRDP``` class, move it back there and I'll merge ;) as for my last comment, looks like the wrong command line argument was my fault.

realjiangms:
OK I'll move the arguments back. So you want the detail arguments kept inside FreeRDP lib class.
I'll fix the h264 argument also

akallabeth:
+1

PR #3176, opened 2016-02-29
Export shadow subsystem

bmiklautz:
* fix build on windows (missing freerdp-client)
* correct dll version information
* export FreeRDP-Shadow (pkg-config, cmake)
* build server side parts during CI


akallabeth:
+1

PR #3178, opened 2016-03-01
fix wayland-client.h not found

hendwolt:
fix a build failure in the wayland client because of a missing include directory


hardening:
@freerdp-bot it looks perfectly fine, anyway would you mind test it ?

PR #3182, opened 2016-03-02
shadow/subsystem: fix pkgconfig file

xhaakon:
Add freerdp-shadow-subsystem to 'libs' in order to avoid undefined
reference to shadow_subsystem_set_entry_builtin at runtime.

Remove cyclic dependency on itself.


bmiklautz:
@freerdp-bot test.

bmiklautz:
Completely overlooked that one. Thanks.

PR #3183, opened 2016-03-02
winpr: fix regression in timezone detection

bmiklautz:
ConvertToUnicode can't be used if the destination is a static buffer.
StandardName and DaylightName were invalid and timezone
redirection didn't work.

This regression was introduced with PR #3151


bmiklautz:
@freerdp-bot test.

akallabeth:
a nasty one you fixed! +1

PR #3184, opened 2016-03-02
Fix build on older 32-bit Ubuntu releases

xhaakon:
Some older 32-bit versions of gcc fail on '#elif __x86_64__'.
This applies at least to what Ubuntu 14.04 and 15.04 have as
the default compiler.

### DIFF
--- a/winpr/include/winpr/wtypes.h.in
+++ b/winpr/include/winpr/wtypes.h.in
@@ -285,7 +285,7 @@ typedef void *PVOID64, *LPVOID64;
 #if WINPR_HAVE_STDINT_H
 typedef intptr_t INT_PTR;
 typedef uintptr_t UINT_PTR;
-#elif __x86_64__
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@xhaakon that's really an interesting one. I wonder why this happens we do have ifdef without defined all over the place.

bmiklautz:
Actually I really wonder why our ci build works on 32 bit then. It's an Ubuntu 15.04 at the moment.
For example: https://ci.freerdp.com/job/linux/label=ubuntu32/1117/console

hardening:
Oups perhaps I merged it too fast, slap me if I did.

bmiklautz:
@hardening the fix itself shouldn't hurt but I really wonder why this happens if it works on our build machine ;)

Issue #3185, opened 2016-03-03
Connection spins when attempting to connect through gateway server


jackdozier:
Using FreeRDP version 2.0.0-dev (git 9afa928)
CentOS Linux release 7.2.1511 (Core) x86_64

Trying to connect to Windows 10 behind a gateway. I'm not sure of the version of the gateway server. Software prompts for gateway password and performs handshake. Connection hangs with high cpu usage after debug output says "Tunnel response received". (Full command line and output below.)

Narrowed issue to libfreerdp/core/gateway/rdg.c. These two commits looked related to issue, so I reverted to commit 48c7a8dc1d334bc316f8de466ce94ff91e368e92 and the issue went away.

> commit 106479d6f27cdd34a8d7dd8e654d890542ab1fcb
> Author: Armin Novak <armin.novak@thincast.com>
> Date:   Tue Jan 26 09:34:22 2016 +0100
> 
>     Fixed clientNameLen initialization.
> 
> commit 73ec3d6acaea4ca69dfd43df3cb847cc83310324
> Author: Armin Novak <armin.novak@thincast.com>
> Date:   Thu Jan 21 14:47:10 2016 +0100
> 
>     Removed fixed size strings.

Not wanting to throw away Armin's work, I reverted back to the latest code and applied the [attached](https://github.com/FreeRDP/FreeRDP/files/156145/0001-Fixed-packet-size-issue.txt) changes and it continued to work for me. Basically, removed the unnecessary(?) zero being written to the stream and updated the packet size. I didn't look into the protocol specs, so I have no idea if that zero is necessary. I did test it on two other environments behind gateway servers, though, and they both worked with my changes.

xfreerdp /g:remote.example.com /v:client.example.com /gu:user /gd:example.com /u:user /d:example.com /log-level:DEBUG
[21:19:31:461] [17612:17613] [DEBUG][com.freerdp.client.x11] - Searching for XInput pointer device
[21:19:31:461] [17612:17613] [DEBUG][com.freerdp.client.x11] - Pointer device: 11
[21:19:31:462] [17612:17613] [DEBUG][com.freerdp.core.nego] - Enabling security layer negotiation: TRUE
[21:19:31:462] [17612:17613] [DEBUG][com.freerdp.core.nego] - Enabling restricted admin mode: FALSE
[21:19:31:462] [17612:17613] [DEBUG][com.freerdp.core.nego] - Enabling RDP security: TRUE
[...]

Issue #2918, opened 2015-10-08
Unable to connect through Gateway


MarcusMoeller:
I am trying to connect through a Gateway using this command:

xfreerdp cpub-....rdp /w:1600 /h:1200 /u:myuser /d:d.ethz.ch /p:"..." /gu:myuser /gd:mydomain /gp:"..." /gt:rpc

which leads to the following error:

[13:10:52:909] [23120:23121] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpdr
[13:10:52:909] [23120:23121] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpsnd
[13:10:52:909] [23120:23121] [INFO][com.freerdp.client.common.cmdline] - loading channel cliprdr
[13:11:52:188] [23120:23121] [ERROR][com.freerdp.core.gateway.rpc] - unexpected RTS PDU: Expected CONN/C2
[13:11:52:188] [23120:23121] [ERROR][com.freerdp.core.gateway.tsg] - tsg_check failure
[13:11:52:188] [23120:23121] [ERROR][com.freerdp.core.nego] - Protocol Security Negotiation Failure
[13:11:52:188] [23120:23121] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED [0x2000C]
[13:11:52:188] [23120:23121] [ERROR][com.freerdp.core.connection] - Error: protocol security negotiation or connection failure
[13:11:52:188] [23120:23121] [ERROR][com.freerdp.client.x11] - Authentication only, exit status 1

MarcusMoeller:
I digged a bit deeper and it does not seem to be related to the gateway directly but to the connection broker.

When I try to connect using the rdp file like:

xfreerdp cpub-TST_Desktop-TST_Desktop-CmsRdsh.rdp /u:myuser /d:mydomain /p:"mypass" /gu:mygwuser /gd:mygwdomain /gp:"mygwpass" /gt:rpc

I got the error above. The connection does not seem to go through the gateway.

When I connect directly to the collection:

xfreerdp /g:mygw  /u:myuser /d:mydomain /p:"" /gu:myuser /gd:mydomain /gp:"" /v:collectionname

it works. I am using latest master btw.

MarcusMoeller:
Sorry I meant /v:servername not collectionname

akallabeth:
[...]

PR #3187, opened 2016-03-03
Gateway fix

akallabeth:
* Removed additional '\0' from TUNNEL_AUTH message (clientName).
Thanks to @jackdozier for the patch in #3185

Feedback from those affected by #2918 would be welcome too. I'm sadly currently lacking one to test.

### DIFF
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -206,7 +206,7 @@ BOOL rdg_send_tunnel_authorization(rdpRdg* rdg)
 	if (!clientName)
 		return FALSE;
 
-	packetSize = 12 + clientNameLen * 2 + sizeof(WCHAR);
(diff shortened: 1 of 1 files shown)


libfreerdp/core/gateway/rdg.c
```diff
@@ -206,7 +206,7 @@ BOOL rdg_send_tunnel_authorization(rdpRdg* rdg)
 	if (!clientName)
 		return FALSE;
 
-	packetSize = 12 + clientNameLen * 2 + sizeof(WCHAR);
+	packetSize = 12 + lientNameLen * sizeof(WCHAR);
```
hardening: lientNameLen: typo ?

akallabeth: @hardening Yes, pushed the wrong commit ^ ^ Never push before coffee

mfleisz:
LGTM

Issue #3186, opened 2016-03-03
Unable to connect to some Windows 2003 servers


AurimasF:
git bisect shows:
1b2b1c4ac14ac43f4e475488763d8659bd934eb6 is the first bad commit
commit 1b2b1c4ac14ac43f4e475488763d8659bd934eb6
Author: David FORT <contact@hardening-consulting.com>
Date:   Thu Feb 18 14:19:36 2016 +0100

    Add support for saveSessionInfo PDU
    
    This patch adds some callbacks so that the client can have access to the informations
    stored in this packet. Server-side implementation is also there, so that a server
    can send these informations to the client.

:040000 040000 7a50280ca01604c16e28bfbba112d7bc903a107e cbec4b8e880124e93d966e61efa9ef374e2c756b M      include
:040000 040000 0959d5dd8cb08a160641042c7b425c9aabe31b12 4595ac19de2a7c56ebabf8667cd864af17d3b665 M      libfreerdp

Unable to connect to some Windows 2003 servers, other Windows 2003 servers work.

Trying to connect:
> ./client/X11/xfreerdp -wallpaper -themes +clipboard +heartbeat +auto-reconnect +compression /f /bpp:16 /kbd:US /v:srvt /u:Administrator
[09:36:59:963] [24644:24645] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpdr
[09:36:59:963] [24644:24645] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpsnd
[09:36:59:963] [24644:24645] [INFO][com.freerdp.client.common.cmdline] - loading channel cliprdr
[09:36:59:968] [24644:24645] [WARN][com.freerdp.core.nego] - Error: SSL_NOT_ALLOWED_BY_SERVER
[09:36:59:970] [24644:24645] [WARN][com.freerdp.core.nego] - Error: SSL_NOT_ALLOWED_BY_SERVER
[09:37:04:381] [24644:24645] [ERROR][com.freerdp.core.rdp] - DATA_PDU_TYPE_SAVE_SESSION_INFO - rdp_recv_save_session_info() failed
[09:37:04:382] [24644:24645] [ERROR][com.freerdp.core.rdp] - rdp_recv_data_pdu() failed
[09:37:04:382] [24644:24645] [ERROR][com.freerdp.core.transport] - transport_check_fds: transport->ReceiveCallback() - -1
[...]

PR #3191, opened 2016-03-03
Fix parsing of saveSessionInfo PDU

hardening:
This should fix the case when no username is sent (issue #3186).


akallabeth:
LGTM +1

Issue #3190, opened 2016-03-03
/usr/lib/freerdp directory should be versioned


weberhofer:
To allow installation of multiple versions of the library, also the name of the /usr/lib/freerdp directory should be versioned.

PR #3192, opened 2016-03-03
Updated freerdp plugin path to use major version.

akallabeth:
* Fix for #3190
* Needs some updates for the packaging, @bmiklautz If you have the time and don't mind ^^


bmiklautz:
+1.

PR #3194, opened 2016-03-03
ConvertFromUnicode fixes and misc hardening

nfedera:
- Added missing ConvertFromUnicode checks
- If ConvertToUnicode allocates memory, guarantee the null termination similar to ConvertFromUnicode's implementation
- Fixed some TestUnicodeConversion.c CTest return values
- Added some CTests for ConvertFromUnicode and ConvertToUnicode
- Misc code and protocol hardening fixes in the surrounding code regions
  that have been touched

### DIFF
--- a/libfreerdp/core/info.c
+++ b/libfreerdp/core/info.c
@@ -199,12 +199,24 @@ BOOL rdp_read_extended_info_packet(rdpRdp* rdp, wStream* s)
 	UINT16 cbClientDir;
 	UINT16 cbAutoReconnectLen;
 	rdpSettings* settings = rdp->settings;
+	WCHAR* wstr;
 
 	if (Stream_GetRemainingLength(s) < 4)
 		return FALSE;
 
 	Stream_Read_UINT16(s, clientAddressFamily); /* clientAddressFamily (2 bytes) */
 	Stream_Read_UINT16(s, cbClientAddress); /* cbClientAddress (2 bytes) */
+
+	/* cbClientAddress is the size in bytes of the character data in the clientAddress field.
+	 * This size includes the length of the mandatory null terminator.
+	 * The maximum allowed value is 80 bytes
+	 */
+
+	if ((cbClientAddress % 2) || cbClientAddress < 2 || cbClientAddress > 80)
+	{
+		WLog_ERR(TAG, "protocol error: invalid cbClientAddress value: %u", cbClientAddress);
+		return FALSE;
+	}
 
 	settings->IPv6Enabled = (clientAddressFamily == ADDRESS_FAMILY_INET6 ? TRUE : FALSE);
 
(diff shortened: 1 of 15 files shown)


MartinHaimberger:
+1 LGTM

server/shadow/Win/win_wds.c
```diff
@@ -744,7 +744,7 @@ int win_shadow_wds_init(winShadowSubsystem* subsystem)
 		return -1;
 	}
 
-	subsystem->pInvitation->lpVtbl->get_ConnectionString(subsystem->pInvitation, &bstrConnectionString);
+	hr = subsystem->pInvitation->lpVtbl->get_ConnectionString(subsystem->pInvitation, &bstrConnectionString);
```
hardening: nice catch

libfreerdp/core/info.c
```diff
@@ -724,32 +899,87 @@ BOOL rdp_recv_logon_info_v2(rdpRdp* rdp, wStream* s, logon_info *info)
...
+
+	/* cbUserName is the size in bytes of the Unicode character data in the UserName field.
+	 * The size of the mandatory null terminator is include in this value.
+	 * Note: Since MS-RDPBCGR 2.2.10.1.1.2 does not mention any size limits we assume
+	 *       that the maximum value is 512 bytes, according to the fixed size of the
+	 *       Username field in the Logon Info Version 1 (TS_LOGON_INFO) structure.
+	 */
+
+	if ((cbUserName % 2) || cbUserName < 2 || cbUserName > 512)
+	{
```
hardening: It seems that on Windows 2003 server the username is not always returned leading to cbUserName being 0, so the test cbUserName < 2 should be dropped. The same probably applies for the domain.

libfreerdp/core/info.c
```diff
@@ -671,35 +817,63 @@ BOOL rdp_recv_logon_info_v1(rdpRdp* rdp, wStream* s, logon_info *info)
...
-	if (cbDomain > 52)
-		return FALSE;
-	if (cbDomain)
+
+	/* cbDomain is the size of the Unicode character data (including the mandatory
+	 * null terminator) in bytes present in the fixed-length (52 bytes) Domain field
+	 */
+	if ((cbDomain % 2) || cbDomain < 2 || cbDomain > 52)
 	{
-		ConvertFromUnicode(CP_UTF8, 0, (WCHAR*) Stream_Pointer(s), cbDomain, &info->domain, 0, NULL, FALSE);
```
hardening: Same here I think we should not test the case cbDomain < 2 and consider this as "info not provided".

bmiklautz:
+1

libfreerdp/core/info.c
nfedera: @hardening I cannot reproduce this. And this would be a protocol violation since [MS-RDPBCGR 2.2.10.1.1.2 Logon Info Version 2](https://msdn.microsoft.com/en-us/library/cc240638.aspx) says:
>  cbDomain (4 bytes): A 32-bit, unsigned integer. The size in bytes of the Domain field **(including the mandatory null terminator).**

So IMO the minimum value is 2.  Do you remember how you reproduced that ?

hardening: Well someone complained about that this morning :)

hardening:
@nfedera the corresponding issue from this morning is https://github.com/FreeRDP/FreeRDP/issues/3186 .

libfreerdp/core/info.c
nfedera: Ok, thanks for the info. Adding a reference to the issue: https://github.com/FreeRDP/FreeRDP/issues/3186

Will fix this ...

nfedera:
@hardening cbUserName and cbDomain are allowed to be 0 now

libfreerdp/core/redirection.c
```diff
@@ -238,14 +260,33 @@ BOOL rdp_recv_server_redirection_pdu(rdpRdp* rdp, wStream* s)
...
+	 * structure we will use the following limits in bytes including the null terminator:
+	 *
+	 * TargetNetAddress:     80 bytes
+	 * UserName:            512 bytes
+	 * Domain:               52 bytes
+	 * Password(Cookie):    512 bytes
+	 * TargetFQDN:          512 bytes
+	 * TargetNetBiosName:    32 bytes
+	 */
+
```
hardening: As it's limits that we're choosing arbitrary, perhaps we should Wlog a warning when they are reached instead of just returning FALSE.

hardening:
There's my non-blocking last remark, but +1.

nfedera:
@hardening the mesages are already printed in in rdp_redirection_read_unicode_string

hardening:
@nfedera I missed that, so let's merge.

PR #3200, opened 2016-03-04
winpr: Bump API version to 2.0

dwmw2:
In commit 12bd0ec8 ("winpr: BUMP the API version to 1.2") it was stated
that the intention was to allow multiple winpr versions to be installed
in parallel.

However, it's not sufficient to change just the *minor* version of the
library. The soname is still 'libwinpr.so.1', and thus we still can't
coexist with earlier versions — in particular, we can't install FreeRDP2
in parallel with existing distribution packages of earlier versions.

Bump the soname so it does actually work


bmiklautz:
@freerdp-bot test.

bmiklautz:
@dwmw2 as the commit message says this was basically required for the *cmake
find module*:

```
This is required that the exported cmake find module can be
differentiated from the "old" one. This allows it to have
multiple winpr API version installed in parallel.

find_package should now be used with an explicit version number
set like:

find_package(WinPR 1.2 REQUIRED)
```

The "installed in parallel" refers to cmake find modules.

bmiklautz:
+1
We should do a clean cut for 2.0. Let's also include winpr ;)

akallabeth:
+

Issue #3199, opened 2016-03-04
/cert-ignore is not applied when using rdp file with authentication level set


HenryJacques:
Hello,

With latest master, this is what I get using xfreerdp and a rdp file to connect to a Windows 2012-R2 RDS farm:

the culprit is the `authentication level:i:2` in the rdp file.
If set to 1 or 2, it fails with the following error.
If set to 0, there is no error.

But I have set /cert-ignore in my command line, which is ignored.

> echo 'password' | xfreerdp file.rdp -toggle-fullscreen /cert-ignore +compression /compression-level:2 +heartbeat /d:DOM /u:user /f /sec:nla /from-stdin /kbd:french +auto-reconnect /client-hostname:myhost +multitouch /sound:sys:pulse,quality:high,rate:44100 /sound:latency:50 +async-update +async-transport
[14:14:07:962] [20529:20530] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpdr
[14:14:07:963] [20529:20530] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpsnd
[14:14:07:963] [20529:20530] [WARN][com.freerdp.core.client] - Skipping, channel already loaded
[14:14:07:964] [20529:20530] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpsnd
[14:14:07:965] [20529:20530] [INFO][com.freerdp.client.common.cmdline] - loading channel cliprdr
[14:14:07:965] [20529:20530] [INFO][com.freerdp.client.common.cmdline] - loading channel drdynvc
Certificate details:
        Subject: C = FR, O = MYO, OU = MYOU, CN = BROKER.DOM, emailAddress = certificats@dom.fr
        Issuer: C = FR, O = MYO, OU = MYOU, CN = BROKER.DOM, emailAddress = certificats@dom.fr
        Thumbprint: 67:3e:4e:6b:dc:45:2c:1e:f5:78:16:17:eb:5b:65:9a:ad:61:c6:39
The above X.509 certificate could not be verified, possibly because you do not have
the CA certificate in your certificate store, or the certificate has expired.
Please look at the documentation on how to create local certificate store for a private CA.
[14:14:07:017] [20529:20530] [ERROR][com.freerdp.crypto] - certificate not trusted, aborting.
[14:14:07:017] [20529:20530] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_CONNECT_CANCELLED [0x2000B]
[...]

PR #3204, opened 2016-03-07
Rdp file preference fix

### LINKED ISSUE #2810: option /cert-ignore not working anymore
On today master, the option /cert-ignore seems to not be applied, the mismatche certificate is always prompted to be accepted or refused...

akallabeth:
* Fixes #2810, #3199 

When using a RDP or assistance file the command line was overwritten by the values from the file.
This pull fixes that by reparsing the command line and replacing the defaults set by the file parser with the command line values.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -1474,6 +1474,7 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 
 				CommandLineSwitchCase(arg, "v")
 		{
+			free (settings->ServerHostname);
 			p = strchr(arg->Value, '[');
 			/* ipv4 */
 			if (!p)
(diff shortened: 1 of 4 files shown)


client/common/cmdline.c
```diff
@@ -1474,6 +1474,7 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 
 				CommandLineSwitchCase(arg, "v")
 		{
+			free (settings->ServerHostname);
```
mfleisz: settings->ServerHostname should be NULLed here or we might end up with a double free in case strchr below fails

client/common/cmdline.c
```diff
@@ -1715,16 +1721,19 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
 		}
 		CommandLineSwitchCase(arg, "p")
 		{
+			free (settings->Password);
 			if (!(settings->Password = _strdup(arg->Value)))
 				return COMMAND_LINE_ERROR_MEMORY;
 		}
 		CommandLineSwitchCase(arg, "g")
 		{
+			free (settings->GatewayHostname);
```
mfleisz: settings->GatewayHostname should be NULLed (possible double free later on)

mfleisz:
Except my remark +1

bmiklautz:
@akallabeth just tested this looks like overriding doesn't work (at least with passwords).
Tested with:
```
xfreerdp test.rdp /p:verysecret
```

akallabeth:
@freerdp-bot test

bmiklautz:
With your last commit it works nicely.

PR #3207, opened 2016-03-07
Fixed error handling for channel load failures.

akallabeth:
* Fixes audin channel loading checks, prevents duplicate loading and registration before channel initialized.
* Fixes error handling for channel loading failures

### DIFF
--- a/libfreerdp/core/client.c
+++ b/libfreerdp/core/client.c
@@ -200,8 +200,9 @@ UINT freerdp_drdynvc_on_channel_disconnected(DrdynvcClientContext* context, cons
  * go through and inform all the libraries that we are initialized
  * called only from main thread
  */
-int freerdp_channels_pre_connect(rdpChannels* channels, freerdp* instance)
+UINT freerdp_channels_pre_connect(rdpChannels* channels, freerdp* instance)
 {
+	UINT error = CHANNEL_RC_OK;
 	int index;
 	CHANNEL_CLIENT_DATA* pChannelClientData;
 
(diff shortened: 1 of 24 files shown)


libfreerdp/core/client.c
```diff
@@ -246,7 +252,10 @@ int freerdp_channels_post_connect(rdpChannels* channels, freerdp* instance)
...
 			pChannelOpenData = &channels->openDataList[index];
 
-			pChannelClientData->pChannelInitEventProc(pChannelClientData->pInitHandle, CHANNEL_EVENT_CONNECTED, hostname, hostnameLength);
+			error = pChannelClientData->pChannelInitEventProc(pChannelClientData->pInitHandle,
+					CHANNEL_EVENT_CONNECTED, hostname, hostnameLength);
+			if (error != CHANNEL_RC_OK)
+				goto fail;
 
 			name = (char*) malloc(9);
 			if (!name)
```
mfleisz: If name is NULL we return a signed value although the function's return type is UINT

winpr/include/winpr/wtsapi.h
```diff
@@ -116,7 +116,7 @@ typedef struct tagCHANNEL_PDU_HEADER
 #define VCEXPORT  __export
 #endif
 
-typedef VOID VCAPITYPE CHANNEL_INIT_EVENT_FN(LPVOID pInitHandle,
+typedef UINT VCAPITYPE CHANNEL_INIT_EVENT_FN(LPVOID pInitHandle,
```
mfleisz: Changing this from VOID to UINT breaks compatibility with the Microsoft API: https://msdn.microsoft.com/en-us/library/windows/desktop/aa383568(v=vs.85).aspx
Not sure if we are still aiming for 100% MS API compatibility ..?

bmiklautz: @mfleisz  for wtsapi at least basic functionality still aiming for compatibility
@akallabeth would you be so kind and the PR?

bmiklautz:
@akallabeth just re-tested this branch it seems to cause a double free in some circumstances:
```
xfreerdp /v:192.168.50.32 /cert-ignore +clipboard  /drive:tmp,/tmp /sound:sys:alsa /microphone:sys:alsa
```

akallabeth:
@bmiklautz Yes, nasty one. Thx for pointing that out.

bmiklautz:
@akallabeth greate. Thanks for having a look!

PR #3209, opened 2016-03-08
Reverted translation of internal VALUE arrays.

akallabeth:
Translation of internally used values is really no good idea.
Leave them untranslated and also do not translate them when used.

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/res/values-de/strings.xml
+++ b/client/Android/Studio/freeRDPCore/src/main/res/values-de/strings.xml
@@ -57,11 +57,6 @@
         <item>True Color (24 Bit)</item>
         <item>Highest Quality (32 Bit)</item>
     </string-array>
-    <string-array name="colors_values_array">
-        <item>16</item>
-        <item>24</item>
-        <item>32</item>
-    </string-array>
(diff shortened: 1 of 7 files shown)


client/Android/Studio/freeRDPCore/src/main/res/values-de/strings.xml
```diff
@@ -80,9 +80,9 @@
         <item>1920x1200</item>
     </string-array>
     <string-array name="resolutions_values_array">
-        <item>automatisch</item>
-        <item>an Bildschirm anpassen</item>
-        <item>Benutzerdefiniert</item>
+        <item>automatic</item>
+        <item>fitscreen</item>
+        <item>custom</item>
```
bmiklautz: What's the problem with the German translation? Looks good to me.

akallabeth: @bmiklautz yes, for the strings actually displayed. This is a settings array where translations do break stuff ;)

akallabeth: @bmiklautz But you did give me an idea, looks like non translatable strings can be marked ;)

PR #3210, opened 2016-03-09
Don't require HOME env var to be set for server-side code

hardening:


hardening:
@akallabeth as discussed on IRC

akallabeth:
lgtm, +1

PR #3173, opened 2016-02-28
Fixed argument parsing for audin.

akallabeth:

PR #3212, opened 2016-03-10
audin: Fixed default sys argument for audin

realjiangms:
PR #3173 fixed argument parsing. However it breaks the usage without specifying subsystem: "/microphone"
In particular, aFreerdp microphone redirect is broken as it exactly uses "/microphone"
This PR fixes the regression.
Fix: We should not call CommandLineParseArgumentsA if argc is only 1 (same case as rdpsnd_main.c)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdb-bot test

akallabeth:
@freerdp-bot test

PR #3221, opened 2016-03-14
pkg/deb: adapt channel paths

bmiklautz:
Fix the nightly packages on Ubuntu 12.04 and wheezy.

### DIFF
--- a/packaging/deb/freerdp-nightly/rules
+++ b/packaging/deb/freerdp-nightly/rules
@@ -36,7 +36,7 @@ override_dh_strip:
 
 override_dh_install:
 	mkdir -p debian/tmp/opt/freerdp-nightly/lib/cmake/
-	rm -rf debian/tmp/opt/freerdp-nightly/lib/freerdp/*.a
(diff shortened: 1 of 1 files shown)


packaging/deb/freerdp-nightly/rules
```diff
@@ -36,7 +36,7 @@ override_dh_strip:
 
 override_dh_install:
 	mkdir -p debian/tmp/opt/freerdp-nightly/lib/cmake/
-	rm -rf debian/tmp/opt/freerdp-nightly/lib/freerdp/*.a
+	rm -rf debian/tmp/opt/freerdp-nightly/lib/freerdp2/*.a
```
hardening: Just wondering why we don't have the same with winPR...

bmiklautz: @hardening that's only the .a files that get created when static channels are enabled and the cmake version is to "old". All winPR libs are installed where they belong.

akallabeth:
+1

PR #3225, opened 2016-03-14
freerdp: Fix possible crash when setting error info in server-mode

mfleisz:


bmiklautz:
@freerdp-bot test.

libfreerdp/core/freerdp.c
```diff
@@ -627,7 +627,11 @@ UINT32 freerdp_error_info(freerdp* instance)
 	return instance->context->rdp->errorInfo;
 }
 
-void freerdp_set_error_info(rdpRdp* rdp, UINT32 error) {
+void freerdp_set_error_info(rdpRdp* rdp, UINT32 error)
+{
+	if (!rdp)
+		return;
```
akallabeth: @mfleisz Wouldn't this check be better off in ```rdp_set_error_info```?

mfleisz: freerdp_set_error_info is the "public" API where we should check the input - rdp_set_error_info on the other hand is an internal API (mostly from within the rdp context itself) so I think it's ok to assume a sane rdp input in that case.

akallabeth: @mfleisz you're right with the public API part, but ```rdp_set_error_info``` has a return value, which should reflect such misuses.

Issue #3198, opened 2016-03-04
regenerate timezones


bmiklautz:
After #1726 was merged the timezones should be re-generated in order to contain the proper data.

bmiklautz:
Fixed with #3226

PR #3226, opened 2016-03-14
winpr: Updates time zones and fixes bias values

rjcorrig:
Updates time zone information database in libwinpr/timezone/timezone.c to Microsoft's December 2015 update. This was critical as many of my client stations incorrectly jumped forward to US Daylight Savings Time a week early.

Corrects casing in the scripts/TimeZones.cs generator script to fix build errors when compiling libwinpr/timezone/timezone.c .

Changes Bias values from unsigned to signed integer to properly capture negative bias values. Bias should be the number of minutes to add to local time to get to UTC. Bias has the opposite sign value of glibc's tm_gmtoff and .NET's BaseUtcOffset. For example, America/New_York should have a Bias of 300 and Asia/Katmandu should have a bias of -345. The results can be verified by issuing "w32tm /tz" at a command prompt inside the RDP connection.

In cases where the time zone cannot be looked up in the database, the client attempts to compute the proper Bias value from tm_gmtoff so that the client's local time appears inside the RDP connection. Since this is not strict GMT, the client time zone is labeled "Client Local Time" to avoid confusion.


rjcorrig:
If accepted, this patch addresses issue #3198

akallabeth:
@rjcorrig Could you rebase your pull to master to get rid of the many merges?

rjcorrig:
Yes, sorry. I'm new at this :-)

rjcorrig:
Better?

akallabeth:
@rjcorrig Much better. @bmiklautz What do you think about this?

mfleisz:
@freerdp-bot test

akallabeth:
@rjcorrig If you spend ```winpr_detect_windows_time_zone()``` a ```void ``` argument a +1 from me ;)

rjcorrig:
Done!

akallabeth:
and merged!

rjcorrig:
Thanks! I'll keep an eye out for future time zone updates from Microsoft and submit patches as needed. Glad to help!

bmiklautz:
@rjcorrig nice one thanks!

PR #3249, opened 2016-03-18
Channel loading fixes

akallabeth:
* Do not load sound redirection with device redirection
* Remove obsolete def files.


akallabeth:
@freerdp-bot test

bmiklautz:
@akallabeth don't pin me down but I think vaguely remember that drive redirection requires sound redirection.

HenryJacques:
@bmiklautz If I understand you correctly, no. I use drive redirection (/drive:) without sound redirection (/sound:)

@akallabeth What is exactly the purpose of your PR? Won't I be able to use both redirection (/sound: and /drive:) any more?

bmiklautz:
@HenryJacques currently there is *always* a (fake) sound channel loaded and initialized if you use /drive. So yes. What the PR does to not load the sound channel implicitly when drive redirection is requested. However as mentioned above I have some memories that this is (was) required back then.

bmiklautz:
@akallabeth https://msdn.microsoft.com/en-us/library/cc669276.aspx Section 2.1
So for "older" system this is a requirement.

bmiklautz:
So -1 for the channel loading.
+1 for removing the obsolete file.

HenryJacques:
@bmiklautz Thank you for the clarification

PR #3250, opened 2016-03-18
core: Add support for CredSSP version 3

mfleisz:
This PR contains:
- Adds support for CredSSP version 3 (remains backwards comp. with clients < version 3)
- In case of certain errors during SPNEGO (i.e. password expired) returns error info in CredSSP protocol
- Fixes stream memory leaks in some error cases

### DIFF
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -5,6 +5,7 @@
  * Copyright 2010-2012 Marc-Andre Moreau <marcandre.moreau@gmail.com>
  * Copyright 2015 Thincast Technologies GmbH
  * Copyright 2015 DI (FH) Martin Haimberger <martin.haimberger@thincast.com>
+ * Copyright 2016 Martin Fleisz <martin.fleisz@thincast.com>
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -54,7 +55,8 @@
  * 	version    [0] INTEGER,
  * 	negoTokens [1] NegoData OPTIONAL,
  * 	authInfo   [2] OCTET STRING OPTIONAL,
- * 	pubKeyAuth [3] OCTET STRING OPTIONAL
+ * 	pubKeyAuth [3] OCTET STRING OPTIONAL,
+ * 	errorCode  [4] INTEGER OPTIONAL
  * }
  *
  * NegoData ::= SEQUENCE OF NegoDataItem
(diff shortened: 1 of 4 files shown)


bmiklautz:
@mfleisz is there a way to test v3?

HenryJacques:
@mfleisz I looked on Google but didn't find the information: what version(s) of Windows support CredSSP v3?
Thanks!

mfleisz:
According to the documentation at https://msdn.microsoft.com/en-us/library/cc226764.aspx the following Windows OSes support V2 (and hence do not send the error code field):

```
In Windows XP SP3, Windows Vista, Windows Server 2008, Windows 7, Windows
Server 2008 R2, Windows 8, and Windows Server 2012, only version 2 of the CredSSP Protocol is
supported.

Windows XP SP3, Windows Vista, Windows Server 2008, Windows 7, Windows
Server 2008 R2, Windows 8, and Windows Server 2012 do not implement the errorCode field.
```

mfleisz:
Basically the only difference between v2 and v3 is that extended error information is sent (the errorCode field is present) so that the client can show a more descriptive error message (i.e. password expired, account is disabled) where older versions just showed a generic error message (could not connect).

HenryJacques:
@mfleisz thanks, I saw the doc but didn't find the information :/
So, to resume, only
- Windows 8.1 and 10
- Windows 2012 R2 and 2016
support CredSSP V3

mfleisz:
@HenryJacques According to the documentation yes :)

HenryJacques:
@mfleisz OK, for your info, I've implemented the same thing for all the versions (even with CredSSP <v3). Basically I mimic mstsc's behaviour and try to get a Kerberos ticket from the DC (when the account is in a domain). The error codes are more explicit then. This doesn't work with a local account.

mfleisz:
@HenryJacques That sounds interesting - can we have a PR with your changes? Do you see any problems that our changes might interfere in any way? As far as I understand they should be able to co-exist without any problems right? (your changes being an extension to the client and mine pushing the supported CredSSP server version to 3).

akallabeth:
Looks like the client side code is not always setting the error code.

HenryJacques:
@mfleisz Of course, I'll do a PR with the changes. But in due time ;) We are still in a testing phase.
My changes are mainly in `transport_ssl_cb` here [https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/transport.c#L65](url)

mfleisz:
@HenryJacques Thanks and no rush ;)
@akallabeth Thanks for spotting that issue - will update the PR once I figured out the other issue with disabled Accounts!

akallabeth:
+1 for the pull.
In addition it would be nice to extend the ```freerdp_get_last_error``` to convert these new errors to a proper string.

PR #3252, opened 2016-03-20
Add ctrl-alt-left, ctrl-alt-right, ctrl-alt-tab changes in

mccart:
Add ctrl-alt-left, ctrl-alt-right, ctrl-alt-tab changes to enable key binding 
behavior in FreeRDP/X11 to behave same/similar as OS key bindings.

Make life easier...

http://unix.stackexchange.com/questions/123208/exclude-certain-key-combinations-from-being-passed-via-rdesktop

From patch submitted to rdesktop project, #178 Send some hotkeys to local windows manager instead of remote.

### DIFF
--- a/client/X11/xf_keyboard.c
+++ b/client/X11/xf_keyboard.c
@@ -470,10 +470,33 @@ BOOL xf_keyboard_handle_special_keys(xfContext* xfc, KeySym keysym)
 	{
 		if (keysym == XK_Return)
 		{
+			WLog_DBG(TAG,  "X keycode 0x%x, Ctrl: %d, Alt: %d", keysym, mod.Ctrl, mod.Alt);
 			if (mod.Ctrl && mod.Alt)
 			{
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test.

freisei:
Maybe exactly what i need! Want to View Unity Launcher and start a programm (which should then appear in front), without minimizing xfreerdp in fullscreen-mode.

maurizio-emmex:
This patch works for me. Tested on Puppy Linux Slacko64 6.3.2 based on Slackware 14.1.
Only a note: Ctrl-Alt-Left is not sufficient, I must press Ctrl-Alt-Left+Left (or Ctrl-Alt-Right+Right).

frruit:
Hi,
why is this Issue still open when so many people want this feature and a solution is available and works for some people?

akallabeth:
@frruit There is already the options of pressing ```Right Ctrl``` to ungrab

PR #3254, opened 2016-03-24
Added advanced gfx command line options.

akallabeth:
Allows selective enabling / disabling of GFX-H264 modes.

### DIFF
--- a/channels/rdpgfx/client/rdpgfx_main.c
+++ b/channels/rdpgfx/client/rdpgfx_main.c
@@ -92,19 +92,22 @@ static UINT rdpgfx_send_caps_advertise_pdu(RDPGFX_CHANNEL_CALLBACK* callback)
 	if (gfx->H264)
 		capsSet->flags |= RDPGFX_CAPS_FLAG_AVC420_ENABLED;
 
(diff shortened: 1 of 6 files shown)


ilammy:
I have tested this both `/gfx:AVC420` and `/gfx:AVC444`, it works like a charm.

This option allows to enjoy AVC420 on Windows 10 systems that either do not have AVC444 installed, or this support is disabled in group policies. (See [this MSDN post](https://blogs.msdn.microsoft.com/rds/2016/01/11/remote-desktop-protocol-rdp-10-avch-264-improvements-in-windows-10-and-windows-server-2016-technical-preview/) for more info on how to enable/disable AVC444.)

For some reason, in case when the client sends RDPGFX_CAPSET_VERSION10 (_without_ RDPGFX_CAPS_FLAG_AVC_DISABLED flag) to the server, but the server does not support AVC444, the server then disables AVC420 as well and uses mostly Progressive codec. When `/gfx:AVC420` is specified on the command line then only RDPGFX_CAPSET_VERSION81 is sent and the servers uses AVC420 just fine.

hardening:
LGTM

PR #3257, opened 2016-03-27
allow /p option without value

tchernomax:
If /p is used without value the password is asked by
freerdp_passphrase_read (in command line)

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -74,7 +74,7 @@ static COMMAND_LINE_ARGUMENT_A args[] =
 	{ "vc", COMMAND_LINE_VALUE_REQUIRED, NULL, NULL, NULL, -1, NULL, "Static virtual channel" },
 	{ "dvc", COMMAND_LINE_VALUE_REQUIRED, NULL, NULL, NULL, -1, NULL, "Dynamic virtual channel" },
 	{ "u", COMMAND_LINE_VALUE_REQUIRED, "[<domain>\\]<user> or <user>[@<domain>]", NULL, NULL, -1, NULL, "Username" },
-	{ "p", COMMAND_LINE_VALUE_OPTIONAL, "<password>", NULL, NULL, -1, NULL, "Password" },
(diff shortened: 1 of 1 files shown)


akallabeth:
@maxime1986 May I ask what is wrong with just using ```/from-stdin``` instead of ```/p```?

maxime1986:
I don't have admin access to the RDP server I am connecting to.
From my tests it appears (but I am not sure) that this server doesn't support /sec:nla nor /sec:tls, only /sec:rdp.

To connect to this server I tried `xfreerdp /d:mydomain /u:myusername /from-stdin /app:'||calc' /v:myserver.net` but xfreerdp never asked for the password.
`client_cli_authenticate_raw` nor `freerdp_passphrase_read` was called.

The comment on `client_cli_authenticate_raw` say:
> Callback set in the rdp_freerdp structure, and used to get the user's password, if required to establish the connection.

The *if required to establish the connection* makes me think this only work when NLA is supported.
So I make this patch to clarify things and be able force xfreerdp to ask for the password from stdin.

My interpretation may be wrong. If that so, could you explain to me how to **force** xfreerdp to ask for the password with /from-stdin ?

akallabeth:
@maxime1986 I see, looks like you're right. Still, would be better to fix ```/from-stdin``` than to kind of misuse ```/p```.

maxime1986:
@akallabeth with /sec:rdp is it possible to configure a server to allow client to log in without password (something like guest account) ?
If it's the case, I think /from-stdin can't be fixed because it will not know if it has to ask for a password or not.

Maybe /from-stdin with /sec:rdp should alway ask for password, and disable it if the user leave it blank ?

What do you think ?

maxime1986:
If you think it's better to stay with /from-stdin, is `transport_connect_tls` and `transport_connect_rdp` good places to add `client_cli_authenticate` (or similar) ?

bmiklautz:
@freerdp-bot test.

akallabeth:
@maxime1986 Sorry, don't remember the exact call graph, would have to dig into that myself.

akallabeth:
@tchernomax I've refactored your idea as an extension to ```/from-stdin``` in #4229.

PR #3263, opened 2016-03-31
Added command line option /cert-tofu

akallabeth:


bmiklautz:
@akallabeth tofu = trust on first use? :)

realjiangms:
If just want to accept any cert, what's the difference if I use /cert-ignore?

bmiklautz:
@realjiangms the certificate is saved when encountered for the first time so the real difference is that key changes can be detected.

realjiangms:
@bmiklautz Thanks. That makes sense

PR #3273, opened 2016-04-07
Fix android rotation crash

realjiangms:
Fix the issue in #3258


realjiangms:
I've tested it and it works fine on my Nokia N1.
Without this fix, rotation cause crash

bmiklautz:
@realjiangms nice finding! Thank you very much.

PR #3276, opened 2016-04-11
core: allow ms android client protocol violation

nfedera:
According to [MS-RDPBCGR 2.2.1.11.1.1.1] the TS_EXTENDED_INFO_PACKET structure's cbClientDir field must include the _mandatory_ NULL terminator of the ClientDir field.
However, since version 8.1.31.44, the Microsoft Remote Desktop Client for Android sets cbClientDir to 0.


nfedera:
Update: It wasn't just affecting the Android but also the recent iOS client from MS

PR #3282, opened 2016-04-13
 Added argument checks for exported input functions.

akallabeth:


libfreerdp/core/input.c
```diff
@@ -479,6 +539,9 @@ BOOL input_recv(rdpInput* input, wStream* s)
 {
 	UINT16 i, numberEvents;
 
+	if (!input || !s)
```
mfleisz: I think that stream NULL pointer check is not necessary as this function is called from the transport layer where stream creation is already checked and ensured.

akallabeth: does not hurt either and the symbol is exported (at least libfreerdp internally)

bmiklautz:
+1 if the freerdp-bot is also ok with it ;)

hardening:
I think it's ready to merge when @freerdp-bot will have told us the result of the build

akallabeth:
@freerdp-bot Be a nice bot and test this again

bmiklautz:
@freerdp-bot just test!

Issue #2520, opened 2015-04-05
Problems under Mac OS X 10.5.8 Leopard/PPC (inverted colors without NLA, transport failed with NLA)


czo:
Hi,

Fresh compilation from current master with OS X Leopard 10.5.8 fails to connect to my Windows 7 box. 

When non NLA clients allowed, colors seems to be inverted (screenshot: homokozo.czo.hu/xfreerdp.png ). Googled around it, and fount this message: http://sourceforge.net/p/freerdp/mailman/message/29015196/ , but i can't find the patch. It's merged?

When NLA is requred, i can't connect because:
tornado:~ czo$ xfreerdp /v:172.16.30.252 
[00:42:24:740] [12130:f0081000] [INFO][com.freerdp.client.x11] - No user name set. - Using login name: czo
[00:42:24:879] [12130:f0081000] [ERROR][com.freerdp.client.x11] - Property 174 does not exist
Password: 
[00:42:27:225] [12130:f0081000] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_CONNECT_TRANSPORT_FAILED [0x2000D]
[00:42:27:236] [12130:f0081000] [ERROR][com.freerdp.client.x11] - Authentication only, exit status 1

Anyone has any idea how to make it work? 

versions:
xcode: 3.1.4
cmake: 2.8.12.2
openssl: 1.0.2a 19 Mar 2015

configured with: cmake -DCMAKE_CXX_COMPILER=g++-4.2 -DCMAKE_C_COMPILER=gcc-4.2 -DWITH_X11=ON .

bmiklautz:
@czo what version of FreeRDP dis you try with?

czo:
1.2.0-beta1+android9 from releases. Current master also has this problem, checked it five minutes ago.

current master:
tornado:~ czo$ xfreerdp /version
This is FreeRDP version 1.2.1-dev (git 879ed36)

previous compilation from 1.2.0-beta1+android9
tornado:~ czo$ xfreerdp.old /version
This is FreeRDP version 1.2.1-dev (git n/a)

If u need, i can provide any logs.

czo:
I forgot the arch, i'm trying it on my old PowerPC based notebook.

bmiklautz:
@czo ah ok. Then my guess would be that it is a problem with byte order.
I'm not sure if anyone tested on PowerPC :(.

czo:
[...]

PR #3284, opened 2016-04-13
Endianness fixes

ondrejholy:
Hello, I am trying to fix freerdp to work on big endian architectures (s390, ppc64). See issue #2520 for some additional info. I made some patches, but not sure this is right way to fix it. So I would like to ask for some feedback before I will continue...

The following patch for libwinpr is needed to fix build problems:
http://anonscm.debian.org/cgit/collab-maint/freerdp.git/tree/debian/patches/1004_64-bit-architectures.patch

It seems that bunch of functionality doesn't work (e.g. nfs, compression, colors, string) on big endian machines. I am proposing fixes for some of those problems currently, I will update the branch subsequently. E.g. I modified utf.c to fix strings, so I do not expect that you like this solution, but I am not sure how to do it better to avoid additional memory copies...

Can you give me some feedback?

### DIFF
--- a/libfreerdp/codec/xcrush.c
+++ b/libfreerdp/codec/xcrush.c
@@ -725,7 +725,7 @@ int xcrush_decompress_l1(XCRUSH_CONTEXT* xcrush, BYTE* pSrcData, UINT32 SrcSize,
 		if ((pSrcData + 2) > pSrcEnd)
 			return -1003;
 
-		MatchCount = *((UINT16*) pSrcData);
+		Data_Read_UINT16(pSrcData, MatchCount);
 
 		MatchDetails = (RDP61_MATCH_DETAILS*) &pSrcData[2];
 		Literals = (BYTE*) &MatchDetails[MatchCount];
@@ -736,9 +736,9 @@ int xcrush_decompress_l1(XCRUSH_CONTEXT* xcrush, BYTE* pSrcData, UINT32 SrcSize,
 
 		for (MatchIndex = 0; MatchIndex < MatchCount; MatchIndex++)
 		{
-			MatchLength = MatchDetails[MatchIndex].MatchLength;
-			MatchOutputOffset = MatchDetails[MatchIndex].MatchOutputOffset;
-			MatchHistoryOffset = MatchDetails[MatchIndex].MatchHistoryOffset;
+			Data_Read_UINT16(&MatchDetails[MatchIndex].MatchLength, MatchLength);
+			Data_Read_UINT16(&MatchDetails[MatchIndex].MatchOutputOffset, MatchOutputOffset);
+			Data_Read_UINT32(&MatchDetails[MatchIndex].MatchHistoryOffset, MatchHistoryOffset);
 
 			if (MatchOutputOffset < OutputOffset)
 				return -1005;
--- a/winpr/libwinpr/crt/test/TestUnicodeConversion.c
+++ b/winpr/libwinpr/crt/test/TestUnicodeConversion.c
@@ -292,9 +292,9 @@ BOOL test_unicode_uppercasing(BYTE* lower, BYTE* upper)
 
 BOOL test_ConvertFromUnicode_wrapper()
 {
+	BYTE src1[] = "\x52\x00\x49\x00\x43\x00\x48\x00\x20\x00\x54\x00\x45\x00\x58\x00\x54\x00\x20\x00\x46\x00\x4f\x00\x52\x00\x4d\x00\x41\x00\x54\x00\x40\x00\x40\x00\x40\x00";
+	BYTE src2[] = "\x52\x00\x49\x00\x43\x00\x48\x00\x20\x00\x54\x00\x45\x00\x58\x00\x54\x00\x20\x00\x46\x00\x4f\x00\x52\x00\x4d\x00\x41\x00\x54\x00\x00\x00";
 	/*               00  01  02  03  04  05  06  07  08  09  10  11  12  13  14  15  16  17  18 */
-	WCHAR src1[] = { 'R','I','C','H',' ','T','E','X','T',' ','F','O','R','M','A','T','@','@','@' };
-	WCHAR src2[] = { 'R','I','C','H',' ','T','E','X','T',' ','F','O','R','M','A','T', 0 };
 	CHAR  cmp0[] = { 'R','I','C','H',' ','T','E','X','T',' ','F','O','R','M','A','T', 0 };
 	CHAR* dst = NULL;
 	int i;
@@ -306,7 +306,7 @@ BOOL test_ConvertFromUnicode_wrapper()
 	printf("Input UTF16 String:\n");
 	string_hexdump((BYTE*) src1, 19 * sizeof(WCHAR));
 
-	i = ConvertFromUnicode(CP_UTF8, 0, src1, 16, &dst, 0, NULL, NULL);
+	i = ConvertFromUnicode(CP_UTF8, 0, (WCHAR*)src1, 16, &dst, 0, NULL, NULL);
 	if (i != 16)
 	{
 		fprintf(stderr, "ConvertFromUnicode failure A1: unexpectedly returned %d instead of 16\n", i);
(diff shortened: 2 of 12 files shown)


hardening:
@freerdp-bot can you test a build of this.

hardening:
I gave a quick look and it looks good to me.
Do the unit test pass after the patch (on both little and big endian hosts) ?

bmiklautz:
@ondrejholy very nice!  Big endian architectures have fallen short since we don't have any build or test machine for it.

The debian patch you are referring to is it still required on master if stdint is available?

Thank you for your effort.

ondrejholy:
Thanks that you looked at the patches....

@hardening I tried `ctest .` for a first time (is this what you meant?):
- little endian: `14 tests failed out of 202` with and without my patches, so that is probably ok
- big endian: `23 tests failed out of 202` without my patches, `24 tests failed out of 202` with my patches (1 fixed, 2 new), hmm, I have to look at the tests in detail...

@bmiklautz Yes, the patch is still needed, but it has to be updated for master. There is still the following check for 64bit architectures, which obviously doesn't cover all of the (i.e. ppc64, s390x):
`#if defined(__x86_64__) || defined(__arm64__)`

hardening:
@ondrejholy strange here I have only 2 well-known failing tests: TestCommMonitor and TestIoDevice (BTW we should remove them as they require some special setup to succeed, so they're not really unitary tests).
It could be interesting if you can inquiry on the new failures your patches includes (and it's quite cool if it fixes some)

ondrejholy:
I don't have a real big endian hardware. I was testing the patches on such machine using SSH X11 forwarding. Then I realized that the code is using big endian, however little endian X11 server is still used to draw windows. So I tried VNC instead of X11 forwarding and realized that the colors are still corrupted. 

"Inverted" colors are used on a big endian machine. However "inverted" color for ARGB32 should be BGRA (not ABGR). This logic works for 32 depth, but not for another depths. "Inverted" color for RGB16 is not BGR16, but RGB16 with swapped bytes, i.e. loaded as little endian and stored as big endian.

See commit 52e46cc5. I am not sure now, whether "inverted" colors, which are used in the code currently (i.e. BGR, ABGR) are really needed for some clients, or it is a mistake and it was intended to handle bitmaps on big endian. Can you give me more info about this?

So recently added patches should fix colors for X11 client on big endian with hw gdi. Similar changes are needed for sw gdi also...

ondrejholy:
@hardening we discussed it on irc before several days, but following list of tests fail on big endian (excepting TestComm, TestWtsApiExtra, TestIoDevice) if somebody else is interested in:
```
	 84 - TestImage (Failed)
	 90 - TestStream (Failed)
	122 - TestNTLM (Failed)
	138 - TestUnicodeConversion (Failed)
	171 - TestGdiCreate (Failed)
	180 - TestFreeRDPCodecNCrush (Failed)
	183 - TestFreeRDPCodecPlanar (Failed)
	200 - TestClientRdpFile (Failed)
	166 - TestCommonAssistance (Failed)
```

hardening:
@ondrejholy I think I know what it is about for TestImage, I'll send you some patches to test if you don't mind

ondrejholy:
@hardening Please do so...

Any comments regarding "inverted" colors? @awakecoding ?

bmiklautz:
@ondrejholy not sure if this fixes the problem with inverted colors but @akallabeth hase done some work on GDI color convresion. Maybe you can try to apply your changes on top of his branch and see if the problem still exists: https://github.com/akallabeth/freerdp/tree/color_conversion_fix_v3 (for details see mail on the mailing list)

ondrejholy:
@bmiklautz Hmm, will take a look at the color conversion fixes...

I have added several patches in order to fix NLA authentication. Not sure that commit 12dfc5e is best idea, because I didn't realize before that WCHAR strings are used frequently in winpr and therefor fixes like commit 0a0de16 are needed...

ondrejholy:
I found two public accessible big endian systems, which you can probably use for testing:
s390x: http://www-03.ibm.com/systems/z/os/linux/support/community.html
ppc64be: https://fit-rhlab.rhcloud.com/powerlinux-openpower-development-hosting/

ondrejholy:
Hmm, some data in RDP is sent as BGR(A) according the comment from @akallabeth on mailing list (https://sourceforge.net/p/freerdp/mailman/message/35099835/), which means that the inverted colors are not intended for big endian, so commit 52e46cc is probably wrong and may cause troubles...

ondrejholy:
@bmiklautz I've discussed the color problems on BE with @akallabeth during the last days. It seems that the color overhaul should fix many problems which I tried to fix, because the conversions are done byte per byte, so it seems endian-safe. Although the buffer is accessed as LE, we can use BGRX32 for 24/32bpp instead of XRGB32 on BE. For 15/16bpp will be necessary to add some special BE formats... However I am not able to test his patches currently, because it segfaults for me.

ondrejholy:
I thought previously that it is not a good idea to store the wide char strings always as LE. However I've googled a bit and as far as I can see Windows APIs expects that wide char strings are always stored as LE. Consequently, it makes sense to me to store the strings also as LE (12dfc5e) and require LE in WinPR API also. Do you agree? If so, more fixes like 0a0de16 are needed...

So, I am about to remove color related patches from this branch and prepare some other wide char related patches. Do you think that the rest of the patches are ok to push?

ondrejholy:
So I got rid of color-related patches. Rest of the patches fix other most visible problems I see under big endian (i.e. build, compression, authentication). Is the branch acceptable now?

bmiklautz:
@freerdp-bot test

libfreerdp/codec/xcrush.c
```diff
@@ -725,7 +725,7 @@ int xcrush_decompress_l1(XCRUSH_CONTEXT* xcrush, BYTE* pSrcData, UINT32 SrcSize,
 		if ((pSrcData + 2) > pSrcEnd)
 			return -1003;
 
-		MatchCount = *((UINT16*) pSrcData);
+		Data_Read_UINT16(pSrcData, MatchCount);
 
 		MatchDetails = (RDP61_MATCH_DETAILS*) &pSrcData[2];
```
akallabeth: Is the initialisation here still used? otherwise it is probably better to remove it altogether.

ondrejholy: Which initialization do you have in mind? `MatchDetails` is used for initialization of `Literals` on the next line and `Literals` is used in the following code, so I don't think we can remove it.

akallabeth: @ondrejholy Yes, you're right. Did mix the Data_Read_UINT16 function arguments up.

ondrejholy:
It is worth to mention that TestNTLM and TestCommonAssistance pass thanks to the patches. TestUnicodeConversion and TestStream are not endian-independent, working on fixes currently. TestClientRdpFile fails probably because of various wide char conversions, will take a look later... I haven't checked the rest of the tests, they probably depends on color conversions fixes.

ondrejholy:
@hardening Have you already made some patches related to `TestImage` for testing? I've looked at it briefly and it seems that there are problems with assignments like:
`image->width = bi.biWidth;`
which needs to be replaced by something like:
`Data_Read_UINT32(bi.biWidth, image->width)`

ondrejholy:
So I pushed the fixes for TestStream and TestUnicodeConversions.

bmiklautz:
@freerdp-bot test.

bmiklautz:
@freerdp-bot whitelist ondrejholy

ondrejholy:
Can somebody review the patches please?

akallabeth:
@hardening @mfleisz  @MartinHaimberger Could you have a look and merge? +1 from my side.

winpr/libwinpr/crt/test/TestUnicodeConversion.c
```diff
@@ -292,9 +292,9 @@ BOOL test_unicode_uppercasing(BYTE* lower, BYTE* upper)
 
 BOOL test_ConvertFromUnicode_wrapper()
 {
+	BYTE src1[] = "\x52\x00\x49\x00\x43\x00\x48\x00\x20\x00\x54\x00\x45\x00\x58\x00\x54\x00\x20\x00\x46\x00\x4f\x00\x52\x00\x4d\x00\x41\x00\x54\x00\x40\x00\x40\x00\x40\x00";
+	BYTE src2[] = "\x52\x00\x49\x00\x43\x00\x48\x00\x20\x00\x54\x00\x45\x00\x58\x00\x54\x00\x20\x00\x46\x00\x4f\x00\x52\x00\x4d\x00\x41\x00\x54\x00\x00\x00";
 	/*               00  01  02  03  04  05  06  07  08  09  10  11  12  13  14  15  16  17  18 */
```
hardening: may I suggest that we put as comment the WCHAR array, so we know what the bytes correspond to ?

hardening:
@MartinHaimberger @mfleisz it's up to you to merge it !

ondrejholy:
Cool, thanks! I will focus on the corrupted colors now. It seems that color overhaul from @akallabeth fixes many problems...

@hardening I will propose patch with WCHAR arrays in comments later...

PR #3293, opened 2016-04-19
core: allow TS Gateway protocol violation

nfedera:
According to [MS-RDPBCGR 2.2.1.11.1.1.1] the TS_EXTENDED_INFO_PACKET structure's cbClientAddress field must include the _mandatory_ NULL terminator of the clientAddress field in its byte count.
However, connections proxied via Microsoft's TS Gateway set the cbClientDir value to 0.

### DIFF
--- a/libfreerdp/core/info.c
+++ b/libfreerdp/core/info.c
@@ -210,9 +210,11 @@ BOOL rdp_read_extended_info_packet(rdpRdp* rdp, wStream* s)
 	/* cbClientAddress is the size in bytes of the character data in the clientAddress field.
 	 * This size includes the length of the mandatory null terminator.
 	 * The maximum allowed value is 80 bytes
+	 * Note: Although according to [MS-RDPBCGR 2.2.1.11.1.1.1] the null terminator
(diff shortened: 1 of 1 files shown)


bmiklautz:
@nfedera if it's mandatory maybe we should at least stick a warning (or debug message in) if it violates the written protocol.

bmiklautz:
Otherwise +1

nfedera:
@bmiklautz This is similar to the (maybe) protocol violation of the iOS/Android MS clients with the cbClientDir value. Since in [MS-RDPBCGR 2.2.1.11.1.1.1] the null terminator is mandatory for for the clientDir and clientAddress strings, this also implies that the terminator is mandatory for empty strings. However, the spec does not explicitly mention null resp. not existent strings ... so there is a little room for interpretation.
And since this Microsoft software is already out there and cannot be changed I think we don't need a debug/warning message here.

hardening:
+1 for no warning
LGTM

hardening:
At MS they're really good to not follow their own specifications !

PR #3296, opened 2016-04-19
Fix CB_FILECONTENTS_REQUEST message decoding.

akallabeth:
Do not require optional fields to be present.


bmiklautz:
@akallabeth nice finding! That one must have slipped through while the channel was hardened.

PR #3311, opened 2016-05-02
Don't export/install rdtk

bmiklautz:
Although exported rdtk isn't really mature and stable yet therefore it is
not exported/installed anymore. In case FreeRDP is built without
static libraries librdtk is installed as part or the
FreeRDP-ShadowTargets exports as this is the only place it is used right
now.


hardening:
Looks good on the principle, waiting for @freerdp-bot to agree with us.

PR #3312, opened 2016-05-02
Disable comm tests per default

bmiklautz:
comm tests require a serial device for testing. If the test environment
isn't available the tests will return errors therefore the tests are
now disabled per default. They can be (re-)enabled by using the cmake
option BUILD_COMM_TESTS.


hardening:
We've been bothered by this test for soooo long

bmiklautz:
@hardening same for me but in the end it sooo easy ;)

PR #3316, opened 2016-05-03
rdpgfx: Implementation for server side channel

realjiangms:
rdpgfx: Implementation for server side channel
server/shadow: support h264 codec with gfx channel


channels/rdpgfx/server/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# FreeRDP: A Remote Desktop Protocol Implementation
+# FreeRDP cmake build script
+#
+# Copyright 2012 Marc-Andre Moreau <marcandre.moreau@gmail.com>
```
hardening: Are you sure about the copyright statement ?

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+	{
+	    WLog_ERR(TAG, "Stream_New failed!");
+	    return NULL;
+	}
+
+	/* Write header. Note that actual length will be filled
+	 * after the entire packet has been constructed. */
+	if ((error = rdpgfx_write_header(s, &header)))
+	{
+		WLog_ERR(TAG, "rdpgfx_write_header failed with error %lu!", error);
```
hardening: s is leaked here

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+	{
+		WLog_ERR(TAG, "zgfx_compress_to_stream failed!");
+		error = ERROR_INTERNAL_ERROR;
+		goto out;
+	}
+
+	error = WTSVirtualChannelWrite(context->priv->rdpgfx_channel, (PCHAR) Stream_Buffer(fs), 
+								   Stream_GetPosition(fs), &written) 
+								   ? CHANNEL_RC_OK : ERROR_INTERNAL_ERROR;
+
```
hardening: this is not very clean here, and you should probably check written too ?

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+	/* Fill actual length */
+	Stream_SetPosition(s, sizeof(RDPGFX_HEADER) - sizeof(UINT32));
+	Stream_Write_UINT32(s, SrcSize); /* pduLength (4 bytes) */
+
+	/* Allocate new stream with enough capacity. Additional overhead is
+	 * descriptor (1 bytes) + segmentCount (2 bytes) + uncompressedSize (4 bytes)
+	 * + segmentCount * size (4 bytes) */
+	fs = Stream_New(NULL, SrcSize + 7 
+	   + (SrcSize/ZGFX_SEGMENTED_MAXSIZE + 1) * 4);
+
```
hardening: fs is not checked here

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+	{
+		WLog_ERR(TAG, "rdpgfx_server_packet_new failed!");
+		return CHANNEL_RC_NO_MEMORY;
+	}
+
+	Stream_Write_UINT32(s, pdu->width); /* width (4 bytes) */
+	Stream_Write_UINT32(s, pdu->height); /* height (4 bytes) */
+	Stream_Write_UINT32(s, pdu->monitorCount); /* monitorCount (4 bytes) */
+
+	for (index = 0; index < pdu->monitorCount; index++)
```
hardening: How can you be sure that s will be large enough to hold all monitors informations (if you have many) ?

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+		Stream_Write_UINT32(s, monitor->right); /* right (4 bytes) */
+		Stream_Write_UINT32(s, monitor->bottom); /* bottom (4 bytes) */
+		Stream_Write_UINT32(s, monitor->flags); /* flags (4 bytes) */
+	}
+
+	/* pad (total size must be 340 bytes) */
+	if (Stream_GetPosition(s) > RDPGFX_RESET_GRAPHICS_PDU_SIZE)
+	{
+		WLog_ERR(TAG, "Invalid RDPGFX_RESET_GRAPHICS_PDU data!");
+		return CHANNEL_RC_NO_BUFFER;
```
hardening: s is leaked here

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+				havc420 = &(havc444->bitstream[0]);
+				rdpgfx_write_h264_avc420(context, s, havc420);
+			}
+		}
+		else
+		{
+			Stream_Write(s, cmd->data, cmd->length);
+		}
+
+		assert(Stream_GetPosition(s) <= Stream_Capacity(s));
```
hardening: that kind of assert makes no sense, just ensure that the stream is big enough before writing in it

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+		Stream_Write_UINT32(s, cmd->contextId); /* codecContextId (4 bytes) */
+		Stream_Write_UINT8(s, cmd->format); /* pixelFormat (1 byte) */
+
+		Stream_Write_UINT32(s, cmd->length); /* bitmapDataLength (4 bytes) */
+		Stream_Write(s, cmd->data, cmd->length);
+
+		return rdpgfx_server_packet_send(context, s);
+	}
+	else
+	{
```
hardening: This else block can be moved one level down (as the if above returns) and the "else" removed

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+		WLog_ERR(TAG, "rdpgfx_server_packet_new failed!");
+		return CHANNEL_RC_NO_MEMORY;
+	}
+
+	Stream_Write_UINT16(s, pdu->surfaceId); /* surfaceId (2 bytes) */
+	if ((error = rdpgfx_write_color32(s, &(pdu->fillPixel)))) /* fillPixel (4 bytes) */
+	{
+		WLog_ERR(TAG, "rdpgfx_write_color32 failed with error %lu!", error);
+		return error;
```
hardening: s is leaked

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+	}
+
+	Stream_Write_UINT16(s, pdu->fillRectCount); /* fillRectCount (2 bytes) */
+	for (index = 0; index < pdu->fillRectCount; index++)
+	{
+		fillRect = &(pdu->fillRects[index]);
+		if ((error = rdpgfx_write_rect16(s, fillRect)))
+		{
+			WLog_ERR(TAG, "rdpgfx_write_rect16 failed with error %lu!", error);
+			return error;
```
hardening: s is leaked here too

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+		return CHANNEL_RC_NO_MEMORY;
+	}
+
+	Stream_Write_UINT16(s, pdu->surfaceId); /* surfaceId (2 bytes) */
+	Stream_Write_UINT64(s, pdu->cacheKey); /* cacheKey (8 bytes) */
+	Stream_Write_UINT16(s, pdu->cacheSlot); /* cacheSlot (2 bytes) */
+	if ((error = rdpgfx_write_rect16(s, &(pdu->rectSrc)))) /* rectSrc (8 bytes ) */
+	{
+		WLog_ERR(TAG, "rdpgfx_write_rect16 failed with error %lu!", error);
+		return error;
```
hardening: s is leaked

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+	Stream_Write_UINT16(s, pdu->cacheSlot); /* cacheSlot (2 bytes) */
+	Stream_Write_UINT16(s, pdu->surfaceId); /* surfaceId (2 bytes) */
+	Stream_Write_UINT16(s, pdu->destPtsCount); /* destPtsCount (2 bytes) */
+
+	for (index = 0; index < pdu->destPtsCount; index++)
+	{
+		destPt = &(pdu->destPts[index]);
+		if ((error = rdpgfx_write_point16(s, destPt)))
+		{
+			WLog_ERR(TAG, "rdpgfx_write_point16 failed with error %lu", error);
```
hardening: s is leaked

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+ */
+static UINT rdpgfx_recv_cache_import_offer_pdu(RdpgfxServerContext* context, wStream* s)
+{
+	UINT16 index;
+	RDPGFX_CACHE_IMPORT_OFFER_PDU pdu;
+	RDPGFX_CACHE_ENTRY_METADATA* cacheEntries;
+	UINT error = CHANNEL_RC_OK;
+
+	Stream_Read_UINT16(s, pdu.cacheEntriesCount); /* cacheEntriesCount (2 bytes) */
+	pdu.cacheEntries = (RDPGFX_CACHE_ENTRY_METADATA*)
```
hardening: perhaps you should check that pdu.cacheEntriesCount is not 0 (or you will return with OOM error)

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+	pdu.cacheEntries = (RDPGFX_CACHE_ENTRY_METADATA*) 
+						calloc(pdu.cacheEntriesCount, sizeof(RDPGFX_CACHE_ENTRY_METADATA));
+
+	if (!pdu.cacheEntries)
+	{
+		WLog_ERR(TAG, "calloc failed!");
+		return CHANNEL_RC_NO_MEMORY;
+	}
+
+	for (index = 0; index < pdu.cacheEntriesCount; index++)
```
hardening: You should check that there's enough to read in s

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+ */
+static UINT rdpgfx_recv_caps_advertise_pdu(RdpgfxServerContext* context, wStream* s)
+{
+	UINT16 index;
+	RDPGFX_CAPSET* capsSet;
+	RDPGFX_CAPSET capsSets[3];
+	RDPGFX_CAPS_ADVERTISE_PDU pdu;
+	UINT error = CHANNEL_RC_OK;
+
+	Stream_Read_UINT16(s, pdu.capsSetCount); /* capsSetCount (2 bytes) */
```
hardening: In this function, you must add check so that you're sure that s has enough bytes to read.

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+	pdu.capsSets = (RDPGFX_CAPSET*) capsSets;
+
+	for (index = 0; index < pdu.capsSetCount; index++)
+	{
+		capsSet = &(pdu.capsSets[index]);
```
hardening: I guess the limit for index should be 3, or you offer a nice opportunity for an attacker to rewrite the stack...

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+		Stream_Read_UINT32(s, capsSet->version); /* version (4 bytes) */
+		Stream_Seek(s, 4); /* capsDataLength (4 bytes) */
```
hardening: shouldn't capsDataLength  be checked ? (I guess the spec says this field MUST be as certain size)

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT rdpgfx_recv_qoe_frame_acknowledge_pdu(RdpgfxServerContext* context, wStream* s)
+{
+	RDPGFX_QOE_FRAME_ACKNOWLEDGE_PDU pdu;
+	UINT error = CHANNEL_RC_OK;
+
+	Stream_Read_UINT32(s, pdu.frameId); /* frameId (4 bytes) */
```
hardening: same here don't Stream_Read in s until you're sure the bytes are available

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1239 @@
...
+
+	ExitThread((DWORD)error);
+	return NULL;
+}
+
+static BOOL rdpgfx_server_open(RdpgfxServerContext* context)
+{
+	RdpgfxServerPrivate* priv = (RdpgfxServerPrivate *) context->priv;
+
+	if (!priv->thread)
```
hardening: error cases must be looked at because you leak some priv members depending of where it fails.

include/freerdp/codec/zgfx.h
```diff
@@ -53,8 +57,9 @@ typedef struct _ZGFX_CONTEXT ZGFX_CONTEXT;
 extern "C" {
 #endif
 
-FREERDP_API int zgfx_compress(ZGFX_CONTEXT* zgfx, BYTE* pSrcData, UINT32 SrcSize, BYTE** ppDstData, UINT32* pDstSize, UINT32* pFlags);
 FREERDP_API int zgfx_decompress(ZGFX_CONTEXT* zgfx, BYTE* pSrcData, UINT32 SrcSize, BYTE** ppDstData, UINT32* pDstSize, UINT32 flags);
+FREERDP_API int zgfx_compress(ZGFX_CONTEXT* zgfx, BYTE* pSrcData, UINT32 SrcSize, BYTE** ppDstData, UINT32* pDstSize, UINT32* pFlags);
```
hardening: may I suggest const BYTE* pSrcData ?

channels/rdpgfx/server/CMakeLists.txt
realjiangms: Yeah... Missed that. I checked all .c/.h files but missed that

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: s is expected to be cleaned by rdpgfx_server_packet_send. Will Add comment

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Need to be fixed

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: OK to check written.
What do you mean 'not very clean' besides checking written?

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: All similar points need to be rechecked

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Yes I missed the length calculation. I didn't read the spec correctly when I wrote line 159. It should be fixed.

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: OK, that makes sense

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Just want to have an insurance if future we break the stream length estimation. Currently the length is ensured

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Yes. It is not guaranteed

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Yes. All the 4 read subroutine need to apply this check

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Yes it must be 4.

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: "capsSets" should be dynamically allocated.

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Sure

include/freerdp/codec/zgfx.h
realjiangms: Will do

realjiangms:
Thanks @hardening. It still have compile error even it does work on my Ubuntu.
I'll figure that out. 

I guess I need to add rdpgfx part in workaround channels/server/channels.c, which I used to think  of as obsolete

server/shadow/shadow_client.c
```diff
@@ -30,10 +30,62 @@
...
 
+struct _SHADOW_GFX_STATUS
+{
+	BOOL gfxOpened;
+	BOOL gfxSurfaceCreated;
+};
+typedef struct _SHADOW_GFX_STATUS SHADOW_GFX_STATUS;
+
+static void shadow_client_rdpgfx_new_surface(rdpShadowClient *client)
+{
```
hardening: I guess this function should not be void as it calls some callbacks that can return errors

server/shadow/shadow_client.c
```diff
@@ -400,28 +505,91 @@ BOOL shadow_client_surface_frame_acknowledge(rdpShadowClient* client, UINT32 fra
...
+		regionRect.bottom = cmd.bottom;
+		quantQualityVal.qp = encoder->h264->QP;
+		quantQualityVal.r = 0;
+		quantQualityVal.p = 0;
+		quantQualityVal.qualityVal = 100 - quantQualityVal.qp;
+		avc420.meta.numRegionRects = 1;
+		avc420.meta.regionRects = &regionRect;
+		avc420.meta.quantQualityVals = &quantQualityVal;
+
+		IFCALL(client->rdpgfx->StartFrame, client->rdpgfx, &cmdstart);
```
hardening: I guess all these calls can fail, so...

server/shadow/shadow_encoder.c
```diff
@@ -130,9 +130,6 @@ int shadow_encoder_uninit_grid(rdpShadowEncoder* encoder)
 
 int shadow_encoder_init_rfx(rdpShadowEncoder* encoder)
 {
-	rdpContext* context = (rdpContext*) encoder->client;
-	rdpSettings* settings = context->settings;
-
 	if (!encoder->rfx)
```
hardening: Just wondering how encoder->rfx can be NULL when we fail if rfx_context_new fails (some lines below)...

channels/rdpgfx/server/rdpgfx_main.c
hardening: Forget about cleaness

hardening:
I did a lot of remarks, but don't take it as bashing. After the global hardening we have done, we have increased the requirements for incoming code. So we try to have code that checks all function returns and that doesn't trust the incoming bytes.
Nice work.

server/shadow/shadow_client.c
realjiangms: Reply your comments both this and the next one :)

This is tricky, as I haven't figure out what to do if graphic commands fails. Currently I have to just ignore it (this function, new_surface should be handled, if fail the surface need to be recreated next time. But reset and release, just ignore).

But I agree to have them with return value, just ignore when calling.

server/shadow/shadow_encoder.c
realjiangms: I may misunderstand your question. The rfx_context_new shows if it fail, it always returns NULL.
Are you actually asking "How encoder->rfx is not NULL"? The answer is that it might be called multiple times with 'shadow_encoder_prepare'. So if we've already created context, we don't create again.

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: I released the streams at the end of this function. Let me check whether there's anything missing in zgfx context.

realjiangms:
I agree 'checks all function returns and that doesn't trust the incoming bytes'. I noticed this activity with some big PRs. It should always be appreciated to make quality better. This time I might be a little hurry ;). 
Anyway I still met trouble on windows build, some error doesn't seems in my fix but I still want to figure them out and fix them so that I can make my PR clean build on Mac/Win/X11 (the three shadow server supported platform)

So there will be much fix to do before next update on this PR :)

>

bmiklautz:
@realjiangms let us know if we can help on this.

realjiangms:
@bmiklautz Thanks. I'll definitely past comment and ask for help if I met trouble ;)

server/shadow/shadow_encoder.c
hardening: forget about this comment

realjiangms:
Changes per comments still in progress, but almost done. Will update PR in this week.
Compile error on win32 also resolved. Will also test on MacOS

realjiangms:
@hardening I updated the PR with your comments resolved. Also this time it seems build cleanly on all platform

realjiangms:
Updated with latest code

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -317,28 +410,108 @@ static UINT rdpgfx_send_start_frame_pdu(RdpgfxServerContext* context, RDPGFX_STA
...
+ *
+ * @return estimated size
+ */
+static INLINE UINT32 rdpgfx_estimate_surface_command(RDPGFX_SURFACE_COMMAND* cmd)
+{
+	RDPGFX_AVC420_BITMAP_STREAM *havc420 = NULL;
+	RDPGFX_AVC444_BITMAP_STREAM *havc444 = NULL;
+	UINT32 h264Size = 0;
+
+	/* Create new stream according to codec. */
```
hardening: nitpicking, but a switch() statement would make sense here.

akallabeth:
@freerdp-bot test

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: That makes sense

realjiangms:
@hardening I've fixed to use switch according to your comment. But I found xfreerdp with gfx client doesn't work with shadow with gfx server, the xfreerdp shows black window when gfx channel opened.

I'm investigating on it and try to figure out why. Maybe current implementation has some problem. I'll update the PR later.

gfx client xfreerdp with this version works fine:

commit 45ab97904f862eb336e92caf3ed5759c23404b19

But fails with latest version:

commit 4e66df7228b14e432a0793d32ac2524d37cb88ec

I'm testing on versions between above 2 to figure out why

realjiangms:
It's in https://github.com/FreeRDP/FreeRDP/pull/3328/files, freerdp_client_codecs_prepare was removed from freerdp_client_codecs_reset. 

I need to understand why my implementation depends on this

realjiangms:
I reviewed fix in PR #3328 in detail. The idea is NOT to create codec contexts when creating the surface. Therefore we postpone the creation. Codec contexts are created when we receive a corresponding surface command. Therefore, only freerdp_client_codecs_prepare is called, freerdp_client_codecs_reset is never invoked. That at least breaks the h264 codec as weight and height is 0.

Not sure whether it still work with standard windows RDP server. To make it work again from server side, the only solution is to send a surface command before we send resetGraphic command. But it doesn't make sense.

I'll update this PR with latest fix according to @hardening's latest comment.
We can have another issue opened for client side.

realjiangms:
Posted PR #3376 for client gfx fix. The detail description could be found from notes in #3328

channels/rdpgfx/client/CMakeLists.txt
```diff
@@ -22,8 +22,8 @@ set(${MODULE_PREFIX}_SRCS
 	rdpgfx_main.h
 	rdpgfx_codec.c
 	rdpgfx_codec.h
-	rdpgfx_common.c
-	rdpgfx_common.h)
+	../rdpgfx_common.c
+	../rdpgfx_common.h)
```
akallabeth: Since you are using rdpgfx_common in client and server, maybe it is a good idea to change CMake to create a static lib to avoid duplicate compilation.

realjiangms: It could be better. 
However there's no precedent/sample to have static lib for something used by both channel server and channel client. But there's several existing similar 'common' stuff in other channels such as rdpei and rail.

Well, actually I don't know how to do it gracefully in current compile framework ^_^. Or we should make functions as static inline and put them in to xxxx_common.h? But anyway I think such work would be better resolved in a separate refactor PR for all of the xxx_common.c future.

akallabeth:
@realjiangms Will you update the pull to fix the remaining remarks of @hardening ? Won't get merged with the missing length checks ;)

realjiangms:
@akallabeth I checked the remain remarks in this page which is NOT shown as outdated. All of them is resolved in the latest commit. Not sure why github still display them in this page ;)

realjiangms:
Merged with latest code

akallabeth:
@hardening Anything preventing a merge of this?

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1659 @@
...
+	return rdpgfx_server_single_packet_send(context, s);
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+UINT rdpgfx_send_delete_surface_pdu(RdpgfxServerContext* context,
+									RDPGFX_DELETE_SURFACE_PDU* pdu)
```
hardening: strange indentation here

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1659 @@
...
+ *
+ * @return estimated size
+ */
+static INLINE UINT32 rdpgfx_estimate_surface_command(RDPGFX_SURFACE_COMMAND* cmd)
+{
+	RDPGFX_AVC420_BITMAP_STREAM* havc420 = NULL;
+	RDPGFX_AVC444_BITMAP_STREAM* havc444 = NULL;
+	UINT32 h264Size = 0;
+
+	/* Create new stream according to codec. */
```
hardening: comment is not accurate

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1659 @@
...
+			cmd->codecId == RDPGFX_CODECID_CAPROGRESSIVE_V2)
+	{
+		/* Write RDPGFX_CMDID_WIRETOSURFACE_2 format for CAPROGRESSIVE */
+		Stream_Write_UINT16(s, cmd->surfaceId); /* surfaceId (2 bytes) */
+		Stream_Write_UINT16(s, cmd->codecId); /* codecId (2 bytes) */
+		Stream_Write_UINT32(s, cmd->contextId); /* codecContextId (4 bytes) */
+		Stream_Write_UINT8(s, cmd->format); /* pixelFormat (1 byte) */
+		Stream_Write_UINT32(s, cmd->length); /* bitmapDataLength (4 bytes) */
+		Stream_Write(s, cmd->data, cmd->length);
+	}
```
hardening: perhaps we could return from here so that we wouldn't need the else block (and so less stairs effects)

channels/rdpgfx/server/rdpgfx_main.c
hardening: I really thing the assert here should be removed

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1659 @@
...
+	if (pdu.cacheEntriesCount <= 0)
+	{
+		/* According to the latest spec, capsSetCount <= 3 */
+		WLog_ERR(TAG, "Invalid cacheEntriesCount: %u", pdu.cacheEntriesCount);
+		return ERROR_INVALID_DATA;
+	}
+
+	if (Stream_GetRemainingLength(s) <
+			(pdu.cacheEntriesCount * sizeof(RDPGFX_CACHE_ENTRY_METADATA)))
+	{
```
hardening: using sizeof(in memory struct) is a bad idea to check that size of data available in a stream, as if the compiler does some alignment you end up with RDPGFX_CACHE_ENTRY_METADATA being bigger in memory than in the stream.

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1659 @@
...
+		wStream* s)
+{
+	UINT16 index;
+	RDPGFX_CAPSET* capsSet;
+	RDPGFX_CAPSET capsSets[3];
+	RDPGFX_CAPS_ADVERTISE_PDU pdu;
+	UINT error = CHANNEL_RC_OK;
+	UINT32 capsDataLength;
+
+	if (Stream_GetRemainingLength(s) < sizeof(UINT16))
```
hardening: I think it should be Stream_GetRemainingLength(s) < 2

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1659 @@
...
+	if (pdu.capsSetCount > 3)
+	{
+		/* According to the latest spec, capsSetCount <= 3 */
+		WLog_ERR(TAG, "capsSetCount is greater than 3: %u", pdu.capsSetCount);
+		return ERROR_INVALID_DATA;
+	}
+
+	if (Stream_GetRemainingLength(s) <
+			(pdu.capsSetCount * (sizeof(RDPGFX_CAPSET) + sizeof(UINT32))))
+	{
```
hardening: same remark with using sizeof here

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -0,0 +1,1659 @@
...
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT rdpgfx_recv_qoe_frame_acknowledge_pdu(RdpgfxServerContext* context,
+		wStream* s)
+{
+	RDPGFX_QOE_FRAME_ACKNOWLEDGE_PDU pdu;
+	UINT error = CHANNEL_RC_OK;
+
+	if (Stream_GetRemainingLength(s) < sizeof(RDPGFX_QOE_FRAME_ACKNOWLEDGE_PDU))
```
hardening: same remark with using sizeof here

server/shadow/shadow_client.c
```diff
@@ -173,73 +251,124 @@ void shadow_client_message_free(wMessage* message)
...
+	client = (rdpShadowClient*) peer->context;
+	subsystem = client->server->subsystem;
+
+	IFCALLRET(subsystem->ClientCapabilities, ret, subsystem, client);
+	if (!ret)
+		WLog_WARN(TAG, "subsystem->ClientCapabilities failed");
+
+	/* Recalculate desktop size regardless whether previous call fail 
+	 * or not. Make sure we send correct width/height to client */ 
+	(void)shadow_client_recalc_desktop_size(client);
```
hardening: 2 things here:
* should we call shadow_client_recalc_desktop_size if !ret ?
* any (good) reason to ignore the result of shadow_client_recalc_desktop_size ?

hardening:
My last comments about indentation aren't blocking.
Anyway I really think you should address the problem with the usage of sizeof() when wanting to check the length available in a stream.
After that fixed, we should be able to merge this PR, IMHO.

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Let me find what's our regular good way to get packet size. Maybe a hard coded macro would be better

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Just don't want to introduce any magic number in code. It's OK to change it to 2 as 16bit exactly mean 2 bytes  ;)

server/shadow/shadow_client.c
realjiangms: After CONNECTION_STATE_CAPABILITIES_EXCHANGE stage this callback is invoked. I checked code in libfreerdp/core/peer.c and found the return value of Capabilities doesn't impact the further process. The shadow_client_recalc_desktop_size() is an internal static help function whose return value is '@return TRUE if width/height changed'. In this callback I just use it to ensure the desktop size is correct because the client side WIDTH/HEIGHT is negotiated in CONNECTION_STATE_CAPABILITIES_EXCHANGE stage. So:
1. In any case the shadow_client_recalc_desktop_size need to be called.
2. We don't care whether desktop size is changed in this callback, so the return value of shadow_client_recalc_desktop_size is ignored.

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Because github tab = 8 space, while our non-documented code style define tab=4 space, see
cat scripts/format_code.sh
#!/bin/bash
<snip>
$ASTYLE --lineend=linux --mode=c --indent=force-tab=4 --brackets=linux --pad-header \
                                                           --indent-switches --indent-cases --indent-preprocessor \
                                                           --indent-col1-comments --delete-empty-lines --break-closing-brackets \
                                                           --align-pointer=type --indent-labels --brackets=break \
                                                           --unpad-paren --break-blocks $@
                                                           exit $?
....

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: Not accurate.... It's actually totally wrong :)
I should have check the comments when I split code into different functions

channels/rdpgfx/server/rdpgfx_main.c
realjiangms: I prefer not, because the logic model is not
```
if check1:
   return
if check2:
  return

do something
return
```
The logic model here is actually:
```
if case1:
  handle1
elseif case2:
  handle2
elseif case3:
  handle3
else
  handle4
```
Just currently we happen to have only 2 different cases. :)

hardening: Well I'd prefer the early return, but that's just nitpicking, not blocking at all.

realjiangms:
@hardening Fixed

hardening:
Except my comment it LGTM.

realjiangms:
@hardening Fixed ;)

akallabeth:
@hardening @mfleisz  @MartinHaimberger You think it is ready for merge?

PR #3328, opened 2016-05-11
egfx codec context reset fix, Windows Server 2012 RemoteFX empty tile set fix

awakecoding:
Fix issue with the codec context reset in egfx that caused all codec contexts to be created, including those that aren't used. As a result, if no H.264 support was built-in, the codec context would fail, breaking egfx support entirely even if no H.264 is in use.

Fix issue with a check being too strict on RemoteFX messages containing an empty tile set. Unfortunately, such messages are sometimes sent by Windows Server 2012 (not R2), so we should not fail on them.

### DIFF
--- a/libfreerdp/core/codecs.c
+++ b/libfreerdp/core/codecs.c
@@ -100,13 +100,9 @@ BOOL freerdp_client_codecs_prepare(rdpCodecs* codecs, UINT32 flags)
 	return TRUE;
 }
 
-BOOL freerdp_client_codecs_reset(rdpCodecs* codecs, UINT32 flags,
-				 UINT32 width, UINT32 height)
+BOOL freerdp_client_codecs_reset(rdpCodecs* codecs, UINT32 flags, UINT32 width, UINT32 height)
 {
 	BOOL rc = TRUE;
-
-	if (!freerdp_client_codecs_prepare(codecs, flags))
-		return FALSE;
 
 	if (flags & FREERDP_CODEC_INTERLEAVED)
 	{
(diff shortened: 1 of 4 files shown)


awakecoding:
@freerdp-bot test.

libfreerdp/core/codecs.c
```diff
@@ -100,14 +100,10 @@ BOOL freerdp_client_codecs_prepare(rdpCodecs* codecs, UINT32 flags)
 	return TRUE;
 }
 
-BOOL freerdp_client_codecs_reset(rdpCodecs* codecs, UINT32 flags,
-				 UINT32 width, UINT32 height)
+BOOL freerdp_client_codecs_reset(rdpCodecs* codecs, UINT32 flags, UINT32 width, UINT32 height)
 {
 	BOOL rc = TRUE;
 
-	if (!freerdp_client_codecs_prepare(codecs, flags))
```
realjiangms: This might breaks h264 decoder. The idea is NOT to create codec contexts when creating the surface. Therefore we postpone the creation. Codec contexts are created when we receive a corresponding surface command. Therefore, only freerdp_client_codecs_prepare is called, freerdp_client_codecs_reset is never called. That at least breaks the h264 codec as weight and height is 0.

I think if we don't want to create context for unused codec, we should replace the FREERDP_CODEC_ALL flag in CreateSurface.

Now freerdp_client_codecs_reset in gdi_CreateSurface actually does nothing as no context has been created. 
freerdp_client_codecs_rese in gdi_ResetGraphics only have effect after a gfx->SurfaceCommand is received. But it seems we should have ResetGraphics command before any surface command received.

Does it still work with standard windows RDP server currently?

realjiangms: The codec context should be indeed only created when it is really used. Reset codec should not call prepare as it should only 'reset' the 'existing' codec. Instead, prepare should invoke codec reset subroutines so that we complete the initialization (Although currently only h264 codec really need reset when it is created, we should anyway keep all codec consistent in the semantic level)

I drafted PR #3376 for fix

PR #3337, opened 2016-05-12
shadow: re-enable sharing of multiple monitors

xhaakon:
Commit dea741fc enabled shadow server to share a desktop consisting of all connected monitors (or more generally a rectangle cut-out spanning multiple monitors of the desktop).

After 491978b0, this is no longer possible and only one monitor or its sub-rectangle can be shared.

This commit re-introduces the changes from dea741fc. With them applied, setting desired subRect on rdpShadowServer and enabling shareSubRect allows the remote client to see multiple monitors at once.

### DIFF
--- a/winpr/libwinpr/file/generic.c
+++ b/winpr/libwinpr/file/generic.c
@@ -882,7 +882,7 @@ static BOOL ConvertFindDataAToW(LPWIN32_FIND_DATAA lpFindFileDataA,
 	lpFindFileDataW->dwReserved0 = lpFindFileDataA->dwReserved0;
 	lpFindFileDataW->dwReserved1 = lpFindFileDataA->dwReserved1;
 	unicodeFileName = NULL;
-	length = ConvertToUnicode(CP_UTF8, 0, lpFindFileDataA->cFileName, -1, &unicodeFileName, 0) * 2;
+	length = ConvertToUnicode(CP_UTF8, 0, lpFindFileDataA->cFileName, -1, &unicodeFileName, 0);
 
 	if (length == 0)
 		return FALSE;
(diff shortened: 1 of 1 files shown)


xhaakon:
This PR is mostly meant as a starting point for discussion about enabling this feature. I'm not sure if the change as I'm presenting it is correct with regard to the shadow server architecture, or whether it doesn't create any unexpected side effects.

bmiklautz:
@realjiangms any thoughts?

realjiangms:
I agree with the feature as it is really in use by someone according to ""Share Whole Desktop" does not share all monitors. #66"

The subrect now basically means the 'sub rectangle in the subsystem surface'. The fix looks like, should be in X11 shadow instead of the framework. I need some more time to think about it.

bmiklautz:
@freerdp-bot test.

realjiangms:
This fix looks only works for 'full screen'. 

Think about:
2 monitor, 1024*768 for each, horizontal distribution. If we set subrect as (0,0, 2048, 768), it works fine.
But if we set (48,0,2000,768), it doesn't work as expected.

As a quick fix, I think we can just fix it in x11_shadow.c, if subrect is set, we can make 'x11_shadow_enum_monitors' generate only one single 'virtual' monitor which contains the whole subrect.  This means that if we set 'rect', we ignore 'monitors' argument.

realjiangms:
My previous suggestion might still be confusing. We need to think more about the 'meaning' of surface/screen/monitor/subrect, figure out 'reasonable' meanings

realjiangms:
This time might be a chance redefine the rectangle parameters(x,y,w,h) in surface and subrect.
In current code, the subrect is related rectangle in the selected monitor, and select monitor has exactly same rectangle parameter as surface. 

For example, if we select monitor 1 (x/y/w/h = 1024,0,1024,768), and set subrect as (100,100,100,100), the client will see the screen of (1124,100,100,100).

We can change this definition to following:
1. rectangle parameters in subrect is absolute. And rectangle parameters in monitor definitions are also absolute (this is same as current). The monitors could be considered as predefined special 'subrect'.
2. subrect argument overwrites "monitors" argument. If user choose subrect, the monitor argument will be ignored.
3. Surface need to be created/resized carefully according to monitors and subrect. It should be exactly the rectangle that is seen by client. In previous sample, it is (1124,100,100,100).

I think above is more reasonable than current definition

Then @xhaakon fix matches the new philosophy. But just need some more fix, all x/y/w/h calculation in shadow_input, shadow_screen, and in X11/Mac/Win need to be rechecked and fixed accordingly. Note that currently when we have mouse event (or some client side rectangle related event), we translate the x/y and make it as offset related to surface left/top. And in X11_shadow, translate again to absolute x/y. This need to be updated for new definition. It will be simpler, the x/y from client is just related to surface left/top and we can translate it to absolute. Then in shadow subsystem(x11 shadow) we don't need to translate again.

PR #3339, opened 2016-05-12
libfreerdp-codec: fix xcrush decompression, revert remote control fix

awakecoding:
Fix xcrush decompression on master, revert previous fix


hardening:
That fixes the pb for me

PR #3344, opened 2016-05-13
build: fix win xp builds

bmiklautz:
If doing static builds for windows XP on a newer system there might be
duplicated symbols. With this commit the behavior is changed that
multiple symbols (iif when doing static builds) are allowed.


nfedera:
no luck running the binary on xp (tested xp x86 sp3 and xp x64 sp2):

image

image

nfedera:
Maybe we really need to specify the v120_xp toolset:

See: http://stackoverflow.com/questions/32580151/what-is-the-difference-between-v120-xp-and-v120-on-vs-2013

bmiklautz:
@nfedera I tested it with v120_xp and (v140_xp for vs 2015) same effect (like InitalizeCriticalSectionEx  is missing). There are multiple different functions that are missing (as your image shows).

bmiklautz:
@nfedera is working on much better superseding PR. Closing this one.

PR #3356, opened 2016-05-19
Update mf_mountain_lion.c

rinov:
Remove duplicate return statement in mf_mlion_stop_getting_screen_updates()
This change is no impact anywhere.


hardening:
@freerdp-bot can you test this, but it's so trivial that I should merge it before your answer.

PR #3357, opened 2016-05-20
shadow/rdpsnd: Fix race condition in rdpsnd channel server. 

realjiangms:
The output buffer and format parameters are not protected. This cause some data inconsistency if rdpsnd has own thread.

A typical race scenario is:

1. Stand Alone rdpsnd server thread: In rdpsnd_server_select_format(), Run to line:

context->selected_client_format = client_format_index; // Now context->priv->out_buffer is still 0

2. Another thread: calls rdpsnd_server_send_samples().
It will pass the check of "if (context->selected_client_format < 0)", but will crash when try to copy samples to context->priv->out_buffer

### DIFF
--- a/channels/rdpsnd/server/rdpsnd_main.c
+++ b/channels/rdpsnd/server/rdpsnd_main.c
@@ -314,6 +314,7 @@ static UINT rdpsnd_server_select_format(RdpsndServerContext* context, int client
 	int bs;
 	int out_buffer_size;
 	AUDIO_FORMAT *format;
+	UINT error = CHANNEL_RC_OK;
 
 	if (client_format_index < 0 || client_format_index >= context->num_client_formats)
 	{
@@ -321,6 +322,8 @@ static UINT rdpsnd_server_select_format(RdpsndServerContext* context, int client
 		return ERROR_INVALID_DATA;
 	}
 	
+	EnterCriticalSection(&context->priv->lock);
+
 	context->priv->src_bytes_per_sample = context->src_format.wBitsPerSample / 8;
 	context->priv->src_bytes_per_frame = context->priv->src_bytes_per_sample * context->src_format.nChannels;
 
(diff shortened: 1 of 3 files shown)


channels/rdpsnd/server/rdpsnd_main.c
```diff
@@ -568,18 +588,26 @@ static UINT rdpsnd_server_close(RdpsndServerContext* context)
...
-		if ((error = rdpsnd_server_send_audio_pdu(context, 0)))
+		if (context->selected_client_format < 0)
+		{
+			WLog_ERR(TAG, "Pending audio frame exists while no format selected.");
+			error = ERROR_INVALID_DATA;
+		} 
+		else if ((error = rdpsnd_server_send_audio_pdu(context, 0)))
 		{
 			WLog_ERR(TAG, "rdpsnd_server_send_audio_pdu failed with error %lu", error);
-			return error;
```
bmiklautz: I think error should be set here since it is initialized with CHANNEL_RC_OK.

bmiklautz:
@realjiangms except my comment +1. Thanks.

channels/rdpsnd/server/rdpsnd_main.c
realjiangms: Good eye. I always miss something stupid these days

realjiangms: Good eye. I always miss something stupid these days

server/shadow/shadow_rdpsnd.c
```diff
@@ -62,7 +62,6 @@ static void rdpsnd_activated(RdpsndServerContext* context)
 	}
 
 	context->SelectFormat(context, i);
-	context->SetVolume(context, 0x7FFF, 0x7FFF);
```
akallabeth: Any specific reason to remove this one? Does not look like it is related to the locking issue.

akallabeth:
@realjiangms Except my remark +1

server/shadow/shadow_rdpsnd.c
realjiangms: Actually there's no reason to have this line to set client volume in rdpsnd activated handler.
I copied the code from Windows/wf_rdpsnd.c when I introduce rdpsnd for shadow server last year without much considering.

realjiangms: If a subsystem implementation need to set volume (future), it should post SHADOW_MSG_OUT_AUDIO_OUT_VOLUME message to clients (shadow_client_post_msg/shadow_client_boardcast_msg)

PR #3364, opened 2016-05-24
Limit the time passed in transport check loop

hardening:
This patch make it possible to limit the time that is passed when we call
XXX_check_fds functions. This should smooth the treatment between handling inputs
and handling incoming bitmap updates.
The default maximum time is set to 100 ms.

### DIFF
--- a/libfreerdp/core/test/TestConnect.c
+++ b/libfreerdp/core/test/TestConnect.c
@@ -8,8 +8,8 @@ static HANDLE s_sync = NULL;
 static int runInstance(int argc, char* argv[], freerdp** inst)
 {
 	int rc = -1;
-
 	freerdp* instance = freerdp_new();
+
 	if (!instance)
 		goto finish;
 
@@ -32,19 +32,19 @@ static int runInstance(int argc, char* argv[], freerdp** inst)
 	}
 
 	rc = 1;
+
 	if (!freerdp_connect(instance))
 		goto finish;
 
 	rc = 2;
+
 	if (!freerdp_disconnect(instance))
 		goto finish;
 
 	rc = 0;
-
 finish:
 	freerdp_context_free(instance);
 	freerdp_free(instance);
-
 	return rc;
 }
 
@@ -58,9 +58,7 @@ static int testTimeout(int port)
 		"/v:192.0.2.1:XXXXX",
 		NULL
 	};
-
 	int rc;
-
 	snprintf(arg1, 18, "/v:192.0.2.1:%d", port);
 	argv[1] = arg1;
 	start = GetTickCount();
@@ -71,6 +69,7 @@ static int testTimeout(int port)
 		return -1;
 
 	diff = end - start;
+
 	if (diff > 16000)
 		return -1;
 
@@ -81,7 +80,8 @@ static int testTimeout(int port)
 	return 0;
 }
 
-struct testThreadArgs {
+struct testThreadArgs
+{
 	int port;
 	freerdp** arg;
 };
@@ -95,12 +95,10 @@ static void* testThread(void* arg)
 		"/v:192.0.2.1:XXXXX",
 		NULL
 	};
-
 	int rc;
-	struct testThreadArgs *args = arg;
+	struct testThreadArgs* args = arg;
 	snprintf(arg1, 18, "/v:192.0.2.1:%d", args->port);
 	argv[1] = arg1;
-
 	rc = runInstance(2, argv, args->arg);
 
 	if (rc != 1)
@@ -117,8 +115,8 @@ static int testAbort(int port)
 	HANDLE thread;
 	struct testThreadArgs args;
 	freerdp* instance = NULL;
-
 	s_sync = CreateEvent(NULL, TRUE, FALSE, NULL);
+
 	if (!s_sync)
 		return -1;
 
@@ -126,7 +124,8 @@ static int testAbort(int port)
 	args.arg = &instance;
 	start = GetTickCount();
 	thread = CreateThread(NULL, 0, (LPTHREAD_START_ROUTINE)testThread,
-				&args, 0, NULL);
+	                      &args, 0, NULL);
+
 	if (!thread)
 	{
 		CloseHandle(s_sync);
@@ -137,6 +136,7 @@ static int testAbort(int port)
 	WaitForSingleObject(s_sync, INFINITE);
 	freerdp_abort_connect(instance);
 	status = WaitForSingleObject(instance->context->abortEvent, 0);
+
 	if (status != WAIT_OBJECT_0)
 	{
 		CloseHandle(s_sync);
@@ -147,12 +147,11 @@ static int testAbort(int port)
 
 	status = WaitForSingleObject(thread, 20000);
 	end = GetTickCount();
-
 	CloseHandle(s_sync);
 	CloseHandle(thread);
 	s_sync = NULL;
-
 	diff = end - start;
+
 	if (diff > 1000)
 		return -1;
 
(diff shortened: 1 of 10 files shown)


HenryJacques:
@hardening would it make sense to be able to put the max time value in the command line?

hardening:
@HenryJacques yes we could add something like this and probably remove the async input...

realjiangms:
This fixes the issue we discussed in #2338. I found if we use a thin client which has poor CPU and use a complex codec, e.g. rfx, the client will be put in a 'stuck' status where the keyboard and mouse input never get a chance to be handled. Yes, in worst case, it never get handled because each time it consumes a packet and finishes rendering, a new packet arrives. So transport_check_fds() never returns.

I tried to use async-input, it made input event handled but also cause X11 deadlock. Then I don't get further more solution.

The reason is that we have 'for (;;)'  {now it is while(!freerdp_shall_disconnect(transport->context->instance)) } in libfreerdp/core/transport.c.
I checked history and found it was added at 2014-03-10, but I basically can't figure out why. Following is the reason in the comment (used to be):

```
/**
         * Loop through and read all available PDUs.  Since multiple
         * PDUs can exist, it's important to deliver them all before
         * returning.  Otherwise we run the risk of having a thread
         * wait for a socket to get signaled that data is available
         * (which may never happen).
         */
        for (;;)
       {
                /**
                 * Note: transport_read_pdu tries to read one PDU from

```

Interesting the comment is removed. But it doesn't mean the concern disappears.

@hardening do you know why we introduce infinite loop in transport_check_fds() before? I think we might need to be careful and MAKE SURE that the old concern in the old removed comment is no longer available. Or we might have regression, although I don't understand how it run into the risk described in the above comment

This fix itself looks good. I hope we can resolve it this time. This issue has been bothering me for a year

hardening:
@realjiangms well you're definitely right and something is missing in my PR.

long story: in the client we WaitForMultipleObject and then we call the check_fds function. In the case of a TLS security connection, we(I) have activated the readahead setting of openSSL that makes a SSL_read call read 64k bytes (IIRC) instead of reading 4 bytes (the size of the record) and then the record itself. The consequence is that we get notified of available bytes on the socket once, a 64k block is read, but if we don't read all the PDUs, we won't get notified again that some bytes are buffered for the socket until some activity happens on the socket. So my patch is missing an event that should act as a "read again in the socket" signal.

Looking at this

realjiangms:
@hardening. Thanks for the explanation. I now understand why we need infinite loop when reading.
Looking forward your next update :+1:

bmiklautz:
@hardening did I understand correctly that you'll update this PR?

hardening:
@bmiklautz yes I will, just a matter of hours.

BUGgs:
Hi, any updates on this pull ?

hardening:
@BUGgs yeah I have this lying in a local branch, I still need to do some final tests.

hardening:
About the test, I feel like the TestConnect will fail if 2 of them run in parallel on the same host (as they try to bind 0.0.0.0:3389). @freerdp-bot can you test again to see if I'm right ?

hardening:
@freerdp-bot don't you wanna test again ?

akallabeth:
@freerdp-bot test

hardening:
I have added some changes so that TestConnect has less chance to fail.

libfreerdp/core/test/TestConnect.c
```diff
@@ -241,19 +276,23 @@ static int testSuccess(void)
 
 int TestConnect(int argc, char* argv[])
 {
+	int randomPort;
+
+	randomPort = 3389 + (random() % 20);
```
akallabeth: @hardening Good idea to use a random port, but maybe it would be even better if we spread the port range a little more (and maybe check availability first to avoid collisions with running services)

ropetry:
Great. i will try to test as soon as you give a go 8-)..

libfreerdp/core/transport.c
```diff
@@ -864,8 +895,15 @@ int transport_check_fds(rdpTransport* transport)
 			WLog_ERR(TAG, "transport_check_fds: transport->ReceiveCallback() - %i", recv_status);
 			return -1;
 		}
+
+		now = GetTickCount();
 	}
 
+	if (now > dueDate)
```
realjiangms: Do we miss the event if the loop exits when "now == dueDate"?

realjiangms:
The fix itself looks good to me. The new introduced reread event resolves the issue we discussed before and we no longer need to 'read all PDU' when we receive socket event.

But some code which depends on GetEventHandle seems broken. They don't get reread event.
Maybe we should discard the GetEventHandle API to avoid wrong usage. Also all code with BIO_GET_EVENT need to be reviewed.

Above concern could be handled in another separate PR.

Actually our real problem is that the frontBio event doesn't match our requirement. A prefect event we need is one which is always 'set' when there's still something to be read and always 'reset' when there's nothing to be read. However we use event from the most bottom layer in our tcp/tls stack, so it can't reflect upper layer status correctly. And unfortunately the linux version of event is built directly on socket fd and we basically can't 'set' it manually even if we know 'there is still something to be read'

ropetry:
Hi, is there a way to download a "patched" source, oder just to download this pull request as a diff so that i can apply it by myself? i can´t find a "download" option for the patches or source. perhaps i missed it in the github gui 8-(..
o.k found it.. just apply a .diff to the pullreques id.. i keep this here for others..
https://github.com/FreeRDP/FreeRDP/pull/3364.diff or .patch

thanks
roman

ropetry:
Hello, anyone around with some news about the patch ?
@akallabeth @hardening

hardening:
Ok I have updated the patch according to the comments and rebased it too, so review is welcome.

@realjiangms I agree that GetEventHandle should be removed from the public API

libfreerdp/core/transport.c
```diff
@@ -1075,19 +1112,26 @@ rdpTransport* transport_new(rdpContext* context)
 	    || transport->connectedEvent == INVALID_HANDLE_VALUE)
 		goto out_free_receivebuffer;
 
+	transport->rereadEvent = CreateEvent(NULL, TRUE, FALSE, NULL);
```
akallabeth: @hardening Missing event close in transport_free

hardening: Yep I've seen that too yesterday, patch coming

akallabeth:
@hardening Could you rebase this pull and maybe run ```./scripts/format_code.sh``` on every changed file in this pull to get a consistent style on master?

hardening:
I did some changes in the astyle parameters, the most controversial change must be to set the line width to 100: who still use a 80 column terminal to code these days ? (and BTW when formatting is done by humans that's the kind of limit we have in the code)
I have applied that on my modified files. If it's too intrusive, I will revert and apply the original style.

akallabeth:
@hardening Fine with that, but we should not do that too often ;)

akallabeth:
@freerdp-bot test please

bmiklautz:
@hardening may I suggest we also add a command line parameter that this can be configured? With 0 to disable it completely?

ropetry:
@hardening @bmiklautz 
+1 for the command line parameter ...

hardening:
@ropetry @bmiklautz you mean that 0 would restore the old behaviour, where we treat all the packet we can grab, with the potential bug that will never exit that loop ?

bmiklautz:
@hardening I'd think so. If you don't agree 0 needs to be handles properly at least ;).

realjiangms:
+1 for command line parameter and 0 for old behavior. Actually I think it should be default behavior to be backward compatible. By the way, why do you define 100ms as default value? Is it practically best as default?

hardening:
@bmiklautz I don't really see why we would keep the old behaviour. "Being compatible" doesn't really make sense. You can mimick the old behaviour with setting a big value, but I really don't see why somebody would choose to do so and reduce the responsiveness...

@realjiangms  the 100 ms was choosen by the intuition that it was the maximum acceptable value to handle user inputs. Anyway if someone has a better number I'm ok with it.

bmiklautz:
@hardening you are right. A large value can still "restore" the old behavior but what happens if 0 is passed via the command line parameter?

hardening:
@bmiklautz well for now no command line parameter is available for that, so...

hardening:
Added a command line parameter.

ropetry:
@hardening @bmiklautz i think this patch has introduced a new problem. please see issue 3628
if i set the max-loop-time to 1000 instead of the default 100 the session ( with powerpoint an dother applications) run smoother and not has minimal lags.

realjiangms:
@hardening The problem from @ropetry makes it really interesting :)
 "Being compatible" does really make sense. He is somebody would choose to do so and reduce the responsiveness, but makes server generates less drawing command and works better :)

Issue #2796, opened 2015-07-20
mouse cursor 'skipping back' problem


psumbera:
While using both server and client I see on client following issue. When moving move from one side of screen to another it looks like it's skipping back.

The 'skipping back' problem could always be reproduced in poor network environment.

The problem disappear when I comment out the only call to x11_shadow_pointer_position_update in server/shadow/X11/x11_shadow.c (the server side mouse position would never be sent to client side).

We should have enhancement that the server should not send back the mouse position to the client where user is moving the cursor.

I think the idea to update the mouse position to clients is to notify the clients which are just 'observers'. In multi-client scenario, usually there should be only one 'operator' who is move the mouse or click it. We only need to update the mouse position to the 'observers'.

realjiangms:
Thanks @psumbera 
I'll create a PR when I finish the fix

realjiangms:
Almost a year later, I start to implement a fix. Sorry for late.
The problem could be easily reproduced on client-server where ping reply time > 200ms

realjiangms:
@psumbera I just checked in fix for this issue. Can we close it?

bmiklautz:
Should be fixed on master but no feedback for a while. Closing this issue. Please re-open or create a new one if the problem still exists for you.

PR #3369, opened 2016-05-25
Server/shadow: Fix issue 2796. Mouse cursor draw back issue

realjiangms:
Fix the issue we discussed in #2796

### DIFF
--- a/server/shadow/X11/x11_shadow.c
+++ b/server/shadow/X11/x11_shadow.c
@@ -267,6 +267,7 @@ void x11_shadow_input_mouse_event(x11ShadowSubsystem* subsystem, rdpShadowClient
 
 	server = subsystem->server;
 	surface = server->surface;
+	subsystem->lastMouseClient = client;
 
 	x += surface->x;
 	y += surface->y;
@@ -320,6 +321,7 @@ void x11_shadow_input_extended_mouse_event(x11ShadowSubsystem* subsystem, rdpSha
 
 	server = subsystem->server;
 	surface = server->surface;
+	subsystem->lastMouseClient = client;
 
 	x += surface->x;
 	y += surface->y;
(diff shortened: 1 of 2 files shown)


server/shadow/X11/x11_shadow.c
```diff
@@ -380,7 +386,26 @@ int x11_shadow_pointer_position_update(x11ShadowSubsystem* subsystem)
...
 
-	return shadow_client_boardcast_msg(subsystem->server, NULL, msgId, (SHADOW_MSG_OUT*) msg, NULL) ? 1 : -1;
+	server = subsystem->server;
+
+	ArrayList_Lock(server->clients);
+	for (index = 0; index < ArrayList_Count(server->clients); index++)
+	{
+		client = (rdpShadowClient*)ArrayList_GetItem(server->clients, index);
+		if (client == subsystem->lastMouseClient)
+		{
```
bmiklautz: Coding style. Only line argument shouldn't have {}

server/shadow/X11/x11_shadow.c
```diff
@@ -380,7 +386,26 @@ int x11_shadow_pointer_position_update(x11ShadowSubsystem* subsystem)
...
+			/* Skip the client which send us the latest mouse event */
+			continue;
+		}
+
+		if (shadow_client_post_msg(client, NULL, msgId, (SHADOW_MSG_OUT*) msg, NULL))
+		{
+			count++;
+		}
```
bmiklautz: Coding style. Only line argument shouldn't have {}

bmiklautz:
Except the coding style it looks good to me \o/.

realjiangms:
I understand ^_^

But can I argue that always have {} would makes it have less line difference when someone need to add more lines of code in the 'if' block, and also makes merge easier in Parallel Development.

I would accept the comment as default behavior if there's no one say 'please keep the current style' in 24hrs ^_^, and change the code to following:

```
if (client == subsystem->lastMouseClient)
	continue; /* Skip the client which send us the latest mouse event */


if (shadow_client_post_msg(client, NULL, msgId, (SHADOW_MSG_OUT*) msg, NULL))
	count++;
```

bmiklautz:
@realjiangms all you say is tru. I totally agree with you and personally also prefer to always have {} blocks, even for single lines. However the "unwritten" coding guide for FreeRDP says that single lines should not have {}. If in doubt there is a astyle script in scripts/format_code.sh.

realjiangms:
Then a commonly accepted rule wins, even it is 'unwritten'.

realjiangms:
Having some trouble with astyle...

bmiklautz:
@realjiangms it's of course not required to use the script.

realjiangms:
git push failed. Let me try again

bmiklautz:
Perfect. Thanks a lot.

PR #3370, opened 2016-05-25
winpr/thread: fix TestThreadCreateProcess

nfedera:
On WIN32 TestThread now works and is expected to succeed


bmiklautz:
@freerdp-bot test.

PR #3371, opened 2016-05-25
Server/shadow: Protect X operations on shared subystem X11 display with X11 display lock

realjiangms:


realjiangms:
Lack of X11 Display lock could cause deadlock in X operation eventually

bmiklautz:
@realjiangms did you encounter any dead lock?

realjiangms:
Yes, when I'm testing the mouse draw back fix with an environment that latency > 200ms.
The backtrace shows deadlock happen in XReply(). It seems X11 always cause deadlock if we don't protect XDisplay with XLockDisplay.

realjiangms:
The blocked threads are mouse input and screen capture

bmiklautz:
@realjiangms thanks!

PR #3372, opened 2016-05-26
winpr/library: fix win32 test and some cleanup

nfedera:
On Windows we seem to have to load the TestLibrary[AB] test libraries from in same folder the test executable runs.

Also removed the empty RemoveDllDirectory, SetDefaultDllDirectories, AddDllDirectory tests and the redundant FreeLibrary test.

TestLibrary now works and succeeds on Win32.


hardening:
As far as I know the subject, it seems good to me

hardening:
perhaps the tests that have been removed could have the corresponding files dropped too ?

nfedera:
@hardening thanks! completely forgot to rm . fixed now.

PR #3376, opened 2016-05-29
channels/rdpgfx: Fix for gfx client

realjiangms:
Make freerdp_client_codecs_prepare calls codec reset It fix broken h264 client from #3328. Also it doesn't make sense that we don't need width and height for codec initialization while we need them for codec reset

### DIFF
--- a/client/Windows/wf_gdi.c
+++ b/client/Windows/wf_gdi.c
@@ -406,7 +406,9 @@ BOOL wf_gdi_bitmap_update(rdpContext* context, BITMAP_UPDATE* bitmapUpdate)
 
 			if (bitsPerPixel < 32)
 			{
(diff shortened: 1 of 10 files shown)


hardening:
LGTM

akallabeth:
LGTM, except: ```freerdp_client_codecs_prepare``` should remove the reset code and just call ```freerdp_client_codecs_reset``` after setting up everything.

realjiangms:
@akallabeth It was calling freerdp_client_codecs_reset in my first version.
But consider this scenario:
1. We call prepare with flag A+B+C+D
2. We have already had A/C/D initialized.
In this case,we should only prepare/reset B and leave A/C/D untouched

akallabeth:
@realjiangms Ok, you have a point ;) Good work!

realjiangms:
Thanks

PR #3377, opened 2016-05-30
freerdp/core/gcc: channel name hardening

nfedera:
According to [[MS-RDPBCGR 2.2.1.3.4.1 Channel Definition Structure](https://msdn.microsoft.com/en-us/library/cc240513.aspx)] the channel name must be an 8-byte array containing a null-terminated collection of seven ANSI characters that uniquely identify the channel.

We did not check if the transmitted name was null-terminated which could have the usual severe effects on stability and security since the channel name is used in several functions expecting a null-terminated string (strlen, printf, etc.)

### DIFF
--- a/libfreerdp/core/gcc.c
+++ b/libfreerdp/core/gcc.c
@@ -1428,8 +1428,18 @@ BOOL gcc_read_client_network_data(wStream* s, rdpMcs* mcs, UINT16 blockLength)
 	/* channelDefArray */
 	for (i = 0; i < mcs->channelCount; i++)
 	{
-		/* CHANNEL_DEF */
+		/**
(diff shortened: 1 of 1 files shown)


hardening:
Strange I would have swear that it was not null-terminated on purpose. Is it possible that the specs have changed on this point ?
+1 otherwise

nfedera:
@hardening I don't remember. And regarding null-termination in general the rdp protocol is a real mess with unclear specs and violations by Microsoft's implementations all over the place. However, at least here the specs, [headers](https://msdn.microsoft.com/en-us/library/aa380685(v=vs.85).aspx) and implementations seem sane.

PR #3381, opened 2016-06-01
fix race conditions, tests and some invalid return values

nfedera:
Since the current winpr implementation for overlapped operations is incomplete and buggy, all affected functions will now fail if they are called with a set FILE_FLAG_OVERLAPPED flag or a non-null pointer to a OVERLAPPED structure.

winpr/nt:
- use proper one-time initialization on win32
- fix TestNtCreateFile
- fix broken/incomplete _RtlAnsiStringToUnicodeString
- unimplemented functions return appropriate error codes

winpr/pipe:
- improved TestPipeCreateNamedPipe
- rewrite the completely broken TestPipeCreateNamedPipeOverlapped test

rdtk:
- improve test and don't blindly return success

winpr/synch:
- fix race condition in TestSynchTimerQueue

winpr/ssspi:
- fix TestEnumerateSecurityPackages printf output
- fix TestQuerySecurityPackageInfo printf output

winpr/environment:
- fix GetEnvironmentStrings printf output

winpr/comm:
- unimplemented functions return appropriate error codes

winpr/io:
- unimplemented functions return appropriate error codes

winpr/thread:
- implement SwitchToThread() via sched_yield()

### DIFF
--- a/winpr/libwinpr/synch/wait.c
+++ b/winpr/libwinpr/synch/wait.c
@@ -285,9 +285,9 @@ DWORD WaitForSingleObject(HANDLE hHandle, DWORD dwMilliseconds)
 
 DWORD WaitForSingleObjectEx(HANDLE hHandle, DWORD dwMilliseconds, BOOL bAlertable)
 {
-	WLog_ERR(TAG, "Function not implemented.");
-	assert(0);
-	return WAIT_OBJECT_0;
+	WLog_ERR(TAG, "%s: Not implemented.");
+	SetLastError(ERROR_NOT_SUPPORTED);
+	return WAIT_FAILED;
 }
 
 DWORD WaitForMultipleObjects(DWORD nCount, const HANDLE *lpHandles, BOOL bWaitAll, DWORD dwMilliseconds)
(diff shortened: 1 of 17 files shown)


bmiklautz:
Very nice!

bmiklautz:
@nfedera nice catch.

winpr/libwinpr/synch/wait.c
```diff
@@ -285,9 +285,9 @@ DWORD WaitForSingleObject(HANDLE hHandle, DWORD dwMilliseconds)
 
 DWORD WaitForSingleObjectEx(HANDLE hHandle, DWORD dwMilliseconds, BOOL bAlertable)
 {
-	WLog_ERR(TAG, "Function not implemented.");
-	assert(0);
-	return WAIT_OBJECT_0;
+	WLog_ERR(TAG, "%s: Not implemented.");
```
AlexFdv: Did you forget to use __FUNCTION__ macro here and below?
WLog_ERR(TAG, "%s: Not implemented.", __FUNCTION__);

hardening: Good catch

PR #3382, opened 2016-06-02
Fix some memory leaks in freerdp_settings_free()

giox069:
**freerdp_settings_free()** in libfreerdp/core/settings.c should free almost all pointer members of "struct rdp_settings" (aka: **rdpSettings).** Currently not all members of the setting struct are freed. This PR adds free for some settings.
I did not include the following three settings, because I think they need more careful handling for the free phase
```
TargetNetAddresses
Password51
DeviceArray
```


bmiklautz:
@giox069 thanks.

antenore:
Cool!! Thanks guys!

PR #3388, opened 2016-06-04
winpr: several pool and synch fixes

nfedera:
pool:
- the winpr implementation fallback was not used on older windows editions
- drop useless and conflicting TP_CALLBACK_ENVIRON_V3
- fix race conditions by using use proper one-time initialization
- on win32 WinPR tried to load several pool/callback_environment functions from kernel32.dll but since these are defined as inline functions in the windows headers, no windows edition has ever exported them in any dll.
- removed callback_environment.c and added corresponding static inline  function to pool.h
- fix segfault in TestPoolWork: CloseThreadpoolWork() must not be called if there is a cleanup group associated with the work object since calling CloseThreadpoolCleanupGroupMember() already releases the work object

sync:
- The windows headers incorrectly define InitializeCriticalSectionEx support if  _WIN32_WINNT >= 0x0403 instead of >= 0x0600 (Vista)
- created a compatible define to deal with this issue


hardening:
As far as I know the subject it looks good to me

PR #3392, opened 2016-06-07
winpr/synch: fix barrier deadlock in release build

nfedera:
Let the compiler know that we're comparing a volatile value.
Otherwise the compiler might nuke the comparison operation and produce code that will spin endlessly.


nfedera:

bmiklautz:
Nice!

PR #3402, opened 2016-06-12
Compile without <sys/queue.h>

doughdemon:
Some systems such as linux system using musl libc might not have the include <sys/queue.h>.

### DIFF
--- a/winpr/include/winpr/interlocked.h
+++ b/winpr/include/winpr/interlocked.h
@@ -31,10 +31,12 @@ extern "C" {
 
 #ifndef _WIN32
 
+#ifdef __IOS__
 /* workaround for SLIST_ENTRY conflict */
 
 #include <sys/queue.h>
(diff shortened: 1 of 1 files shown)


hardening:
@freerdp-bot you would be so nice to test that PR !

floppym:
The test fail doesn't seem related to this change.

akallabeth:
@freerdp-bot test

bmiklautz:
@freerdp-bot test.

bmiklautz:
@floppym we are doing some updates/changes on the builder so it might come to some failed builds.

bmiklautz:
@doughdemon sys/queue.h is usually available on linux systems. However as the include shouldn't be required on other systems than iOS +1 for the fix.

bmiklautz:
@freerdp-bot test.

PR #3403, opened 2016-06-13
Fix include path in freerdp-client.pc

TingPing:


hardening:
@freerdp-bot would you be smart enough to test this ?

hardening:
This LGTM, @bmiklautz any opinion ?

bmiklautz:
@hardening the extra / definitely doesn't hurt. 
+1

bmiklautz:
I'll also review the other .pc file and check if similar issues exist.

bmiklautz:
@TingPing thanks for pointing that out.

PR #3406, opened 2016-06-14
winpr/sspi/ntlm: fix GetComputerNameExA parameters

nfedera:
On input, the lpnSize [in, out] parameter for GetComputerNameEx() specifies the total size of the buffer (in characters).
Several functions in ntlm.c were off by one which caused ntlm to fail if the netbios hostname's strlen was exactly MAX_COMPUTERNAME_LENGTH.


bmiklautz:
@nfedera really nice catch!

bmiklautz:
+1 if the build passes.

nfedera:
@byteboon ouch!! you're absolutely right. thanks!
Fixed is in #3428

Issue #3269, opened 2016-04-06
build warning: WARNING: XML failed validation against built-in DTD


hendwolt:
For a little while I get these warnings:
[   83s] [ 76%] Generating xdg-shell-client-protocol.h
[   83s] Scanning dependencies of target audin-client
[   83s] %3Cstdin%3E:342: element entry: validity error : Element entry content does not follow the DTD, expecting (description)?, got (CDATA)
[   83s] %3Cstdin%3E:346: element entry: validity error : Element entry content does not follow the DTD, expecting (description)?, got (CDATA)
[   83s] %3Cstdin%3E:350: element entry: validity error : Element entry content does not follow the DTD, expecting (description)?, got (CDATA)
[   83s] %3Cstdin%3E:356: element entry: validity error : Element entry content does not follow the DTD, expecting (description)?, got (CDATA)
[   83s] *******************************************************
[   83s] *                                                     *
[   83s] * WARNING: XML failed validation against built-in DTD *
[   83s] *                                                     *
[   83s] *******************************************************
[   83s] Scanning dependencies of target audin-client-oss
[   83s] [ 76%] Generating xdg-shell-protocol.c
[   83s] [ 76%] Building C object channels/audin/client/CMakeFiles/audin-client.dir/audin_main.c.o
[   83s] Scanning dependencies of target rdtk
[   83s] %3Cstdin%3E:342: element entry: validity error : Element entry content does not follow the DTD, expecting (description)?, got (CDATA)
[   83s] %3Cstdin%3E:346: element entry: validity error : Element entry content does not follow the DTD, expecting (description)?, got (CDATA)
[   83s] %3Cstdin%3E:350: element entry: validity error : Element entry content does not follow the DTD, expecting (description)?, got (CDATA)
[   83s] %3Cstdin%3E:356: element entry: validity error : Element entry content does not follow the DTD, expecting (description)?, got (CDATA)
[   83s] *******************************************************
[   83s] *                                                     *
[   83s] * WARNING: XML failed validation against built-in DTD *
[   83s] *                                                     *
[   83s] *******************************************************
[   83s] [ 77%] Building C object channels/audin/client/oss/CMakeFiles/audin-client-oss.dir/audin_oss.c.o

I'm not sure, what this means. Otherwise the build completes and the version works for my use cases.

akallabeth:
could you give some details to your os and cmake configuration? pretty hard to guess without these details what could be wrong ;)

[...]

PR #3408, opened 2016-06-14
fix errors in xdg-shell-xml for protocol creation

hendwolt:
This fixes  the build warning: WARNING: XML failed validation against built-in DTD, issue #3269 .

### DIFF
--- a/uwac/protocols/xdg-shell.xml
+++ b/uwac/protocols/xdg-shell.xml
@@ -339,24 +339,32 @@
         0x0000 - 0x0FFF: xdg-shell core values, documented below.
         0x1000 - 0x1FFF: GNOME
       </description>
-      <entry name="maximized" value="1" summary="the surface is maximized">
+      <entry name="maximized" value="1" >
+      <description summary="the surface is maximized">
         The surface is maximized. The window geometry specified in the configure
         event must be obeyed by the client.
+      </description>
       </entry>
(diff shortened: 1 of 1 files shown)


hardening:
@freerdp-bot this PR looks good, but I'd be more confident if you would give a test

bmiklautz:
I saw that these warnings only happen on "older" systems. 
@hendwolt on which distribution did you test create this pull request?
@hardening isn't the xdg-shell.xml copied from wayland - maybe it is an issue with oder
versions of the wayland scanner?

hardening:
@bmiklautz yes the file has been copied from the wayland repo but it's rather old. The proposed changes don't change anything except removing the warning, so...

hendwolt:
I created the request on openSUSE Leap 42.1.

btw.: What's the current state of the wayland client? "GA" or "work in progress" ?
Or, with other words, is it already worth to report bugs for it?
I took it on a test run on openSUSE Tumbleweed with "Gnome 3.20 on Wayland" and "Plasma5 on Wayland" and found a few things.

bmiklautz:
@hendwolt I think it's not much used yet atm. So I'd rather consider it work in progress but having the problems reported wouldn't hurt ;).
@hardening what do you think?

hardening:
Definitely work in progress. But bug reports are welcome.

PR #3410, opened 2016-06-16
Fix win32 linker hell and more

nfedera:
**winpr: fix win32 linking issues**
Depending on the windows target version (_WIN32_WINNT), the used SDK and the build configuration the linker will see multiple libraries exporting the same symbols.
To prevent ugly hacks (e.g. modifying cmake's default system libraries or fragile library linking order chains) we prefix these functions with "winpr_" and create corresponding defines to keep the current api names.

**freerdp: remove static inet_ntop**
inet_ntop is already implemented in winpr

**cmake: remove some void statements and fix winpr source group**

**cmake: replaced STATIC_CHANNELS with BUILTIN_CHANNELS**

**channels/audin/client/winmm: requires winpr**


bmiklautz:
@nfedera wow. Great job. That saves a lot of troubles. I like the renaming from STATIC_CHANNELS to BUILTIN_CHANNELS very much - that makes it more clear what it in fact does.

+1

akallabeth:
nice job!

PR #3417, opened 2016-06-20
Add first support of Kerberos for better error codes handling

HenryJacques:
This PR brings Kerberos support. This gives better error codes when the authentication fails.
xfreerdp can now try to get a "fake" ticket from the KDC, the latter responds with precise error codes we bring back to xfreerdp.

Several conditions are needed:
- compile with the new option  -DWITH_KRB5_SUPPORT=ON
- to compile, install the Heimdal dev packet (heimdal-dev on Ubuntu). This has not been tested against MIT dev libraries
- on the client install libgssapi-krb5-2

To work, a correct client Kerberos configuration is needed (usually in /etc/krb5.conf) for xfreerdp to contact the KDC

(Thanks to A. Aing for this PR)


bmiklautz:
@freerdp-bot test.

bmiklautz:
@HenryJacques thank you for the PR. Just to clarify what the PR basically does it to add additional error information in case a connect failed?

libfreerdp/core/transport.c
```diff
@@ -58,15 +58,158 @@
...
+    	krb5_ccache ccache;
+	krb5_init_creds_context ctx = NULL;
+	
+	memset(address, 0, sizeof(address));
+	
+	strcpy(address, username);
+	strcat(address, "@");
+	strcat(address, domain);
+ 
+	g_KRB5Module = LoadLibrary("libkrb5.so.26");
```
bmiklautz: If used as library this and the following initialization code might get executed multiple times (maybe also by different threads).
I'd suggest to use InitOnceExecuteOnce (for example like done in winpr/libwinpr/pool/pool.c).

What was initially the idea/reason behind loading the library later on instead of directly linking against it?

libfreerdp/core/transport.c
```diff
@@ -58,15 +58,158 @@
...
+	
+	if (!g_KRB5Module) {
+		g_KRB5Module = LoadLibrary("libkrb5.so");
+	}
+
+	if (!g_KRB5Module) {
+		WLog_ERR(TAG, "Failed to load krb5 module\n");
```
bmiklautz: I might be mistaken but is a return missing here? if the library couldn't be loaded the following code won't make sense or work at all.

libfreerdp/core/transport.c
```diff
@@ -58,15 +58,158 @@
...
+	g_KRB5.pKrb5_cc_default 		= (void*) GetProcAddress(g_KRB5Module, "krb5_cc_default");
+	g_KRB5.pKrb5_init_creds_init 		= (void*) GetProcAddress(g_KRB5Module, "krb5_init_creds_init");
+	g_KRB5.pKrb5_init_creds_set_password 	= (void*) GetProcAddress(g_KRB5Module, "krb5_init_creds_set_password");
+	g_KRB5.pKrb5_init_creds_get 		= (void*) GetProcAddress(g_KRB5Module, "krb5_init_creds_get");
+	g_KRB5.pKrb5_init_creds_free 		= (void*) GetProcAddress(g_KRB5Module, "krb5_init_creds_free");
+	g_KRB5.pKrb5_free_context 		= (void*) GetProcAddress(g_KRB5Module, "krb5_free_context");
+
+	/* Create a krb5 library context */
+	if ((ret = g_KRB5.pKrb5_init_context(&context)) != 0) {
+		WLog_ERR(TAG, "krb5_init_context failed with error %d\n", ret);
```
bmiklautz: Here and in the following four function shouldn't an error return as well?

HenryJacques:
@bmiklautz well basically this PR only makes sense with accounts in a Windows domain.
If not, it runs the usual way.

Basically what it does:
- in cases there is a "generic" SSL error (IE/"internal error" see http://linux.die.net/man/3/ssl_alert_type_string) it tries to get a more precise error code than the generic one that freerdp is sending (FREERDP_ERROR_CONNECT_TRANSPORT_FAILED)
- it checks if the username is in a domain and then tries to get a Kerberos ticket using the same credentials as provided in the command line (pKrb5_init_creds_get)
- If a KDC can be reached, it will obviously fail with a Kerberos error code (https://andromeda.rutgers.edu/~sysmail/krb5_error.html). Of course I didn't handle all possible cases.

This is exactly what mstsc is doing (if you use Wireshark in parallel you'll see the same kind of messages between mstsc and the KDC)

I will be able to answer your remarks later (in a few days only)

bmiklautz:
@HenryJacques nice! Thank you very much for explaining!

bmiklautz:
@HenryJacques any updates on the remarks. Would be nice if you got time to take care of them so we can merge it.

HenryJacques:
@bmiklautz I've just came back from my holidays, so I couldn't answer to you.
Yes, we have taken care of the remarks. I will push a new PR in the next few days.

bmiklautz:
@HenryJacques some time passed. Still interested in getting this upstream? If so would you rebase your PR to the latest master and address the remarks? Thank you.

HenryJacques:
@bmiklautz Yes still interested :) Will do that soon!

akallabeth:
@HenryJacques Ok, found the time myself to rebase. Closing your pull, thank you for this nice feature.

HenryJacques:
@akallabeth ouch thanks! Sorry I didn't find the time to do that...

akallabeth:
@HenryJacques No problem. Currently have a little time for cleanup and related stuff.

PR #3418, opened 2016-06-21
rdpsnd/server: decrease audio latency and make it configurable.

llyzs:


channels/rdpsnd/server/rdpsnd_main.c
```diff
@@ -337,26 +337,29 @@ static UINT rdpsnd_server_select_format(RdpsndServerContext* context, int client
 		goto out;
 	}
 
+	if (context->latency <= 0)
+	{
```
hardening: bikeschedding: should drop braces

llyzs:
@hardening OK.

PR #3426, opened 2016-06-27
protocol violation: rdp_read_extended_info_packet

nfedera:
The clientTimeZone and all subsequent fields in the Extended Info Packet ([MS RDPBCGR 2.2.1.11.1.1.1](https://msdn.microsoft.com/en-us/library/cc240476.aspx)) are optional but were not handled as such.


hardening:
I'm wondering how we didn't hit it before...

nfedera:
@hardening because all clients tested until now send all fields up to and including cbAutoReconnectLen and that is what the current code expected to receive.

bmiklautz:
@nfedera nice catch!

PR #3436, opened 2016-07-06
Added support for OpenH264 1.6

akallabeth:


hardening:
As far as I know about OpenH264, it looks good to me

HenryJacques:
@akallabeth : does it still continue to work with 1.5.1?
thanks

akallabeth:
@HenryJacques Yes. Also compiled against every version since 1.3 (did not test connections with these old versions though)

akallabeth:
@freerdp-bot test

PR #3438, opened 2016-07-06
Fixed android libfreerdp install paths.

akallabeth:
* Fixes the installation path of freerdp libraries (now Android Studio can find them)
* Allow building only libfreerdp dependencies by setting ```DEPS_ONLY!=0```


client/Android/Studio/gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.jvmargs=-Xmx4096M
```
bmiklautz: That really scares me ;). 4G for a build tool. O my/

PR #3457, opened 2016-08-03
Send Keystrokes to Desktop 

nickmitchko:
Adds keyboard hotkey that disables/enables keyboard capture on the xfreerdp client window.
Previously in X11, you were stuck in fullscreen mode and graphics would break on the normal CTRL+ALT+ENTER; This made multitasking with workspaces near impossible.

Added Keystrokes (captures both uppercase and lowercase):
Ctrl + Atl + R: Releases keystrokes until window regains focus.
Ctrl + Atl + U: Recaptures keystrokes (they get sent back to rdp), only used if the rdp window doesn't not refocus.

Example Usage:
Case (Ubuntu 16.04, multiple workspaces, RDP on workspace 1 of 4 [2x2]): 
command: xfreerdp /u:user /p:password /d:domain /f /multimon /v:rdphost.somehost.com
- First. user sends keystroke Ctrl + Alt + r
- Then user send Ctrl + Alt + →
  Outcome workspace changes to workspace 2

### DIFF
--- a/client/X11/xf_keyboard.c
+++ b/client/X11/xf_keyboard.c
@@ -465,6 +465,26 @@ BOOL xf_keyboard_handle_special_keys(xfContext* xfc, KeySym keysym)
 	{
 		return TRUE;
 	}
+	// Nicholai Mitchko
+	// Add a keyboard binding to allow switching workspaces in FullScreenMode
+	// Release Keyboard: CTRL + ALT + R/r
+	// Capture Keyboard: CTRL + ALT + U/u
+	if((keysym == XK_u) || (keysym == XK_U)){
+		if(mod.Ctrl && mod.Alt){
+			if(xfc->fullscreen){
+				XGrabKeyboard(xfc->display, xfc->window->handle, TRUE, GrabModeAsync, GrabModeAsync, CurrentTime);
+			}
+		}
+	}
+
+	if((keysym == XK_r) || (keysym == XK_R)){
+		if(mod.Ctrl && mod.Alt){
+			if(xfc->fullscreen){
+				XUngrabKeyboard(xfc->display, CurrentTime);
+			}
+		}
+	}
+	// End Changes
 
 	if(!xfc->remote_app && xfc->fullscreen_toggle)
 	{


client/X11/xf_keyboard.c
```diff
@@ -465,6 +465,26 @@ BOOL xf_keyboard_handle_special_keys(xfContext* xfc, KeySym keysym)
 	{
 		return TRUE;
 	}
+	// Nicholai Mitchko
+	// Add a keyboard binding to allow switching workspaces in FullScreenMode
```
hardening: Please remove the comment about the author and the "end changes" below. When we wanna know wrote what we have git blame for that.

hardening:

caramorsimon:
I am using multiple workspaces (and multiple monitors) on 16.04 gnome 3.20 & do not experience graphics breaking.
 It is true that once focus is on freerdp  all input is directed to the rdp window (which prevents ubuntu handling the super key for instance) & the patch is a worthwhile approach for that without having to use the mouse to change focus.
It is not clear what is meant by "Graphics breaking" above, but I experienced a gnome 3.18 defect  whereby a fullscreen freerdp window would remain visible after I select a different workspace, but putting input focus back into rdp required a switch back to the original workspace [even though the "window" is visible]
Understanding the problem is made harder in fullscreen since the thumbnail for fullscreen rdp windows is missing from the workspace previews  & gnome 3.18 does not support any previews for secondary monitors
I have since upgraded to Gnome 3.20 desktop to get a secondary monitor preview extension installed & resolves the window paint issue on workspace selection. freerdp can now be moved between workspaces as expected & my monitors now display the expected windows only from the selected workspace. 
The issue of missing preview thumbnails is still present.
There is also a slight conflict with my multimonitor extension which has added the gnome notification bar to all monitors & is grabbing the clicks in that area even though it is "hidden" under the xfreerdp fullscreen window, I will either get used to it or workaround by reducing screen dimensions slightly if I want to click in the top of maximised windows without thinking about it..

krissi:
I had the same issue and solved it using the undocumented (?) action.sh-feature.

Create the file /usr/share/freerdp/action.sh with the following contents and make it executable. After that Ctrl+Alt+Left/Right work for workspace switching and Scrolllock for minimizing the current window.

```sh
#!/bin/sh

if [ $# -eq 1 ]; then
  echo "Ctrl+Alt+Left"
  echo "Ctrl+Alt+Right"
  echo "Scroll_Lock"
  exit 0
elif [ $# -eq 2 -a "$1" = "key" ]; then
  if [ "$2" = "Ctrl+Alt+Left" ]; then
    xdotool set_desktop 0
  elif [ "$2" = "Ctrl+Alt+Right" ]; then
    xdotool set_desktop 1
  elif [ "$2" = "Scroll_Lock" ]; then
    xdotool getactivewindow windowminimize
  else
    echo "Unknown hotkey: $2" >&2
  fi
else
  echo invalid usage >&2
  exit 1
fi
```

It is more a workaround than a pretty solution, just some thought-sharing ;)

hardening:
@caramorsimon -grab-keyboard should help too

client/X11/xf_keyboard.c
hardening: And I forgot to mention that we're not using C++ comments in FreeRDP.

Appart from that the patch looks good, any comment from others ?

ilammy:
Hm... I've skimmed through xf_keyboard.c and it seems we've already ungrabbing keyboard when Right Ctrl  is pressed. Why that doesn't work?..

Other than that, the changes seem to be okay. I've tested this with Unity and GNOME 3 from a couple of Ubuntus and it helps with workspace switching (at least until I switch back to the workspace with XFreeRDP and it grabs the keyboard again because mouse cursor has entered the window).

Issue #3461, opened 2016-08-08
clipboard between freerdp/freerdp


FragsterAt:
clipboard corrupted, when copy from one freerdp session and try to paste in other session.
freerdp-2.0.0~git.1463131968.4e66df7 (precompiled for opensuse 42.1)

ilammy:
Hi @FragsterAt,

Could you please elaborate on the use case you are experiencing the issue with?

- What applications do you use in both sessions?
- What kind of data are you trying to copy and paste?
- Are you able to reproduce the corruption when using an intermediate local application as a buffer?

Also, could you please describe the nature of the corruptions?

- Do you see garbage instead of the expected data, or just some minor changes?
- Is the data truncated, or some part of it missing?

Thank you.

I could not reproduce the issue with both latest master (5b2455f3147aca395a441fc65b602938600f0788) and the specified version (4e66df7228b14e432a0793d32ac2524d37cb88ec). I used Mozilla Firefox, MS Office apps, Paint, Notepad, Wordpad. All of them have been able to exchange text, HTML, and images just fine.

The case when the data is transferred directly between two remote sessions is a special one in FreeRDP. In this case FreeRDP does not perform any conversions and only acts an intermediate party between two remote clipboards, simply transferring blobs of data.

The only expected corruption would be truncation of the raw clipboard data when it does not fit into X11 property. This could be fixed, I suppose, by implementing the INCR mechanism for huge amounts of data. 

However, any other kind of corruption would be unexpected in this case and is an indicator of something strange happening.

FragsterAt:
Copy-paste from notepad to notepad. In host system paste into kate (it works).

![screenshot_freerdp](https://cloud.githubusercontent.com/assets/20900647/17581970/1d05d98e-5fba-11e6-802f-2ecc01558f1d.png)

If paste in kate, then copy from kate and paste in 2nd session - it also ok.

meersjo:
@FragsterAt , can you verify contents of all three clipboards (primary, secondary and clipboard) using xsel or similar utilities? Is it possible that the chinese (?) text was taken from the wrong clipboard? Depending on how you copy in X, it may set multiple clipboards to the same value.

ilammy:
@meersjo, I do not think that is the case. FreeRDP uses only CLIPBOARD, so other buffers should not interfere.

[...]

PR #3472, opened 2016-08-16
Fix cached clipboard responses during raw transfers

ilammy:
FreeRDP uses `clipboard->data` to cache the result of the Windows→X11 clipboard format conversion, and `xf_cliprdr_process_selection_request()` immediately provides this result to local applications if they request the same clipboard format again. This saves us a possibly costly conversion in case where the user pastes data repeatedly.

However, this caching mechanism did not support raw clipboard transfers (PR #2851) where the unmodified data is passed between two FreeRDP clients. We use the same XClipboard protocol for this, so the `clipboard->data` is in play. We clear the cached value when we receive new data from the server, so initially raw transfers are fine. But if some local application (e.g., a clipboard manager) asks for some data format before the data is pasted into the second FreeRDP session then `clipboard->data` will contain the _converted_ data. And this converted cached data will be provided to the second FreeRDP session as a part of the raw data transfer. Instead we should have provided the original data.

In order to achieve this we are now caching the original data in the same way as the converted one, and the original data is now correctly provided when the second FreeRDP session asks for a raw data transfer.

This should resolve issue #3461.

### DIFF
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -81,10 +81,12 @@ struct xf_clipboard
 	int requestedFormatId;
 
 	BYTE* data;
+	BYTE* data_raw;
 	BOOL data_raw_format;
 	UINT32 data_format_id;
 	const char* data_format_name;
 	int data_length;
+	int data_raw_length;
 	XEvent* respond;
 
 	Window owner;
@@ -774,6 +776,23 @@ static BOOL xf_cliprdr_process_selection_notify(xfClipboard* clipboard,
 	}
 }
 
+static void xf_cliprdr_clear_cached_data(xfClipboard* clipboard)
+{
+	if (clipboard->data)
+	{
+		free(clipboard->data);
+		clipboard->data = NULL;
+	}
+	clipboard->data_length = 0;
+
+	if (clipboard->data_raw)
+	{
+		free(clipboard->data_raw);
+		clipboard->data_raw = NULL;
+	}
+	clipboard->data_raw_length = 0;
+}
+
 static BOOL xf_cliprdr_process_selection_request(xfClipboard* clipboard,
         XEvent* xevent)
 {
@@ -785,6 +804,7 @@ static BOOL xf_cliprdr_process_selection_request(xfClipboard* clipboard,
 	BYTE* data = NULL;
 	BOOL delayRespond;
 	BOOL rawTransfer;
+	BOOL matchingFormat;
 	unsigned long length;
 	unsigned long bytes_left;
 	CLIPRDR_FORMAT* format;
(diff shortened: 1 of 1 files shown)


client/X11/xf_cliprdr.c
```diff
@@ -1370,9 +1392,21 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext* context
...
 	}
 
+	/* Cache converted and original data to avoid doing a possibly costly
+	 * conversion again on subsequent requests */
 	clipboard->data = pDstData;
 	clipboard->data_length = DstSize;
 
+	/* We have to copy the original data again, as pSrcData is now owned
+	 * by clipboard->system. Memory allocation is not fatal here as this
+	 * is only a cached value. */
```
hardening: I think that at least we should print a message.

hardening:
@freerdp-bot could you test that please ?

ilammy:
I added a warning about the allocation failure into the else branch. It also prints out the amount of memory we have failed to allocate, just in case someone tries to pass `(size_t) -1` there.

bmiklautz:
@freerdp-bot test.

bmiklautz:
@ilammy could you rebase your PR to the current master?

ilammy:
Hi @bmiklautz,

I have rebased the branch onto the current master. I have squashed both commits (the original 10bb24f, and 58d619d with an improvement suggested by @hardening), so it's only one commit now.

The commit 8f1adf64eee1112e2c0942496d7b39841bf1434c ("Refactored ClipboardSetData") introduced some merge conflicts:

- Formatting changes in `xf_cliprdr_process_selection_request()`. No actual conflicts there, just different patch context prevented them from applying cleanly.
- `pSrcData` local variable was removed in `xf_cliprdr_server_format_data_response()`. The original commit was setting it to NULL after freeing. Now we do not need it at all.

I have checked the fix again, it still fixes #3461.

tigerfoot:
Ping ?

tigerfoot:
@bmiklautz is there anything else than need to be done by @ilammy or one of us ?
I believe this pr would be a nice christmas present :-)

akallabeth:
@freerdp-bot test

akallabeth:
@bmiklautz Could you have a look why this pull does not finish the checks?

bmiklautz:
@freerdp-bot test.

tigerfoot:
@bmiklautz @akallabeth seems the bot have sucessfully check the PR.
Now it just need a human review ;-)

tigerfoot:
@bmiklautz @akallabeth sorry to bother, but what is missing to get a human review and approval ?

akallabeth:
@tigerfoot currently mostly time ;) but this one looks good from the code side.

tigerfoot:
At least I'm using that patch to the freeRDP build for openSUSE and has helped to fix 
[boo#1004108](https://bugzilla.opensuse.org/show_bug.cgi?id=1004108)
With report of happy users, I'm also using it since mid-december without any troubles.

bmiklautz:
@nfedera any thoughts from your side?

client/X11/xf_cliprdr.c
```diff
@@ -1343,8 +1361,25 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext*
...
+	 * as this is only a cached value. */
+	clipboard->data_raw = (BYTE*) malloc(size);
+	if (clipboard->data_raw)
+	{
+		CopyMemory(clipboard->data_raw, data, size);
+		clipboard->data_raw_length = size;
+	}
+	else
+	{
+		WLog_WARN(TAG, "failed to allocate %zu bytes for a copy of raw clipboard data", size);
```
nfedera: Please see https://github.com/FreeRDP/FreeRDP/blob/master/docs/PrintFormatSpecifiers.md regarding the print format specifiers.

akallabeth:
@ilammy If you fix the remark from @nfedera I'll merge this pull

ilammy:
Hi. Sorry, I got off track a bit.

`PRIuz` specifier is defined in the upstream WinPR, the base commit of my branch does not contain it. So I have rebased the branch onto the latest master again, and fixed the the format specifier in the warning message.

/cc @nfedera, @akallabeth

akallabeth:
@freerdp-bot test

akallabeth:
@mfleisz @bmiklautz @nfedera Any objections merging this one?

PR #3482, opened 2016-08-27
server/shadow: Fix regression with rdp8.0

realjiangms:
Shadow server crash with rdp8.0 with gfx enabled.
Root Cause: rdp8.0 is not support in shadow gfx and rdpgfx_caps_advertise returns an error. However setChannelError crashs because context->errorDescription is NULL
1. Fix shadow gfx to handle rdp8.0
2. Initialize context->errorDescription for server side new-context


hardening:
LGTM

PR #3492, opened 2016-09-08
Fixed difference between mf_peer_context_new signature and psPeerCont…

okwon78:
Hi ~

I fixed compile error caused by the difference between "mf_peer_context_new" signature and "psPeerContextNew" delaration.

I modified mf_peer_context_new signature from "int mf_peer_init(freerdp_peer\* client)" to "void mf_peer_init(freerdp_peer\* client)"

### DIFF
--- a/server/Mac/mf_peer.c
+++ b/server/Mac/mf_peer.c
@@ -229,8 +229,8 @@ void mf_peer_context_free(freerdp_peer* client, mfPeerContext* context)
 void mf_peer_init(freerdp_peer* client)
 {
 	client->ContextSize = sizeof(mfPeerContext);
-	client->ContextNew = (psPeerContextNew)mf_peer_context_new;
-	client->ContextFree = (psPeerContextFree)mf_peer_context_free;
+	client->ContextNew = mf_peer_context_new;
+	client->ContextFree = mf_peer_context_free;
 	freerdp_peer_context_new(client);
 	
 	info_event_queue = mf_event_queue_new();


okwon78:
Hi ~
I got rid of function pointer casts

okwon78:
Before compilation of code, I completely misunderstood what you request. I will try to fix it.

server/Mac/mf_event.h
```diff
@@ -18,7 +18,6 @@
  */
 
 #ifndef __MF_EVENT_H
-#define __XMF_EVENT_H
```
akallabeth: -1, this is a header include guard! I'll agree that the form is bad (__ is reserved for compiler internal use) so maybe something like ```#ifnded MF_EVENT_H #define MF_EVENT_H```

mfleisz: Yeah - even though the old one was wrong (__XMF instead of __MF) we should fix it by correctly naming the #define and not just deleting it - and rename it as @akallabeth suggests

okwon78:
I had tried to remove casts operation. However, It is not easy. in order to get rid of casts, it requires many modifications, because it influences other server modules for different platforms.

server/Mac/mf_peer.c
```diff
@@ -229,8 +229,8 @@ void mf_peer_context_free(freerdp_peer* client, mfPeerContext* context)
 void mf_peer_init(freerdp_peer* client)
 {
 	client->ContextSize = sizeof(mfPeerContext);
-	client->ContextNew = mf_peer_context_new;
-	client->ContextFree = mf_peer_context_free;
+	client->ContextNew = (psPeerContextNew)mf_peer_context_new;
```
akallabeth: Why did you revert that? If you have warnings your function signatures do not match.

server/Mac/mf_event.h
```diff
@@ -18,6 +18,7 @@
  */
 
 #ifndef __MF_EVENT_H
+#define __MF_EVENT_H
```
akallabeth: Why didn't you remove the ```__``` prefix? As already pointed out these are reserved for compiler internal use.

akallabeth:
@okwon78 I fail to see the point in your latest changes. Could you explain what you're trying to achieve? And why do you base your pull on ```stable-1.1``` and not on master?

okwon78:
The reason why I return code to a previous condition is that "psPeerContextNew" and signature of "mf_peer_context_new" is totally different in a parameter. In other words, rdpContext type in psPeerContextNew and mfPeerContext type in mf_peer_context_new are different structure types but the same position in the function signature. However, rdpContext is shared between other platforms in peer.c while mfPeerContext is only for MacOS.

I am not sure that I have understood correctly. In so far as I see. This part is problematic in terms of polymorphism.  Thus I think that casting operation should remain, even though It does not cause any compile error. 

I'm just trying to understand how FreeRDP works to develop a similar solution. In Addition, when I tried to compile master branch with XCode. I encounter several compile issue. This is why I send pull request on stable branch. Actually, the different signature I fixed in this pull request causes compile error on clang.

akallabeth:
@okwon78 That does not influence the functions. If you require access to platform specific context cast it IN the function, do not change the function pointer arguments.
As for the master branch, what are your issues? Compiles fine here, so I'd be interested in what specifics do fail for you.

okwon78:
If you think so. I will remove casts.
In the near future, I will try to compile master branch. And then, report the issues I face.

I have a Question about H.264. 

As I see, this stable version does not support H.264, while master branch support H.264 codec, does it?

I'm also interested in developing freeRDP.

akallabeth:
@okwon78 Yes, stable-1.1 does lack a lot of features. We're trying to get toward a stable 2.0 branch with master. H264 (the AVC420 and AVC444 features of modern RDP) work with OpenH264 as dependency. To get in touch you can try the mailing list and IRC channels.

bmiklautz:
@freerdp-bot test.

Issue #2707, opened 2015-06-18
multimon smart-sizing correct, but in fullscreen shows display on primary monitor only


user1007017:
Hi 
I'm experiencing the following issue: Multimon smart-sizing is correct but in fullscreen displays whole screen only on primary monitor 
The whole screen 1920x1200+2560x1440 is in fullscreen forced to 1920x1200 (shrinks)

I have to mention that it did work earlier.
What can cause this issue?

BTW Client is Windows 8.1

```
[12:55:56:730] [19161:19162] [INFO][com.freerdp.client.common.cmdline] - loading channel cliprdr
 [12:55:56:751] [19161:19162] [ERROR][com.freerdp.core.gcc] - WARNING: true multi monitor support was not advertised by server!
 [12:55:56:751] [19161:19162] [ERROR][com.freerdp.core.gcc] - Sending multi monitor information anyway (may break connectivity!)


My config:
 xfreerdp --version
This is FreeRDP version 1.2.4-dev (git c8731ea)

 xfreerdp --monitor-list
      * [0] 1920x1200   +0+0
        [1] 2560x1440   +1920+0

/usr/local/bin/xfreerdp /f /rfx /cert-ignore /v:192.168.1.xxx /u:xxxxx /p:xxxxx /sec:rdp  +smart-sizing /multimon:force -sec-nla +clipboard +fonts

Distributor ID: Ubuntu
Description:    Ubuntu 14.04.1 LTS
Release:    14.04
Codename:   trusty
```

bmiklautz:
@b987123 you mean the server is windows 8.1?

b987123:
@bmiklautz 
Thanks, yes, the "server"  (client) with the rdp-server
I have now compiled FreeRDP version 1.1.0-beta1 (git 1.1.0-beta+2013071101-162-g770c67)
and there fullscreen span worked flawless. 

I've tried FreeRDP version 1.2.3-dev and 1.2.4-dev (git c8731ea) but there is the problem as mentioned.

giovannitarter:
Hi. 
I'm experiencing a similar issue on the Linux Client. 
The /multimon parameter makes xfreerdp spawn a fullscreen window only on the primary monitor when using WM openbox 3.5 and 3.6. 
Using metacity as WM, the fullscreen window is displayed on both monitors. 

The regression was introduced on commit bad2f138c183b84c996ec16b4f2e0aaf77b54d7c, the issue is not present on the commits before.

Config:
[...]

PR #3502, opened 2016-09-16
Revert problematic commit which breaks multimonitor-support "client/X…

RubenKelevra:
…11: multimon/fullscreen fixes"

fixing #2707

This reverts commit 48be0815c1f12f1f268135a0a214b327d79a9457.


akallabeth:
-1 from me as the comments from @nfedera and @bjcollins already explained what the issue is (and this just breaks other wm)

PR #3511, opened 2016-09-22
fix for xcode8 clock_gettime

nonamed108:
Credit to @ehlertjd for a relatively portable fix.

Fix comes from this: https://github.com/civetweb/civetweb/pull/358

### DIFF
--- a/winpr/libwinpr/synch/wait.c
+++ b/winpr/libwinpr/synch/wait.c
@@ -67,6 +67,7 @@
 
 #include "../pipe/pipe.h"
 
+/* clock_gettime is not implemented on OSX prior to 10.12 */
 #ifdef __MACH__
 
 #include <mach/mach_time.h>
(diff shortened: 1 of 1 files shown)


hardening:
@freerdp-bot can you test that for me please ?

akallabeth:
@nonamed108 Nice patch! Ccould you apply the wrapping to the other occurrences of ```clock_gettime``` in FreeRDP too?

nonamed108:
@akallabeth Thanks! 

So I looked into further wrapping clock_gettime as necessary for os x. The only occurrences where I see  "mach/mach_time.h" included is in the instance I patched. There are other calls to clock_gettime in winpr/libwinpr/sysinfo/sysinfo.c  - but are defined for linux only. 

The other calls in libfreerdp/primitives/test/measure.h never included mach_time.h and so would never have natively supported os x in the first place. 

I think the existing patch should be sufficient to address the issue in the runtime as is.

PR #3512, opened 2016-09-24
client/X11: Only upconvert < 24 bit RFX tiles

nathankidd:
Fixing this logical typo gives a slight performance boost /
CPU usage reduction on heavy loads.


hardening:
@freerdp-bot it looks good to me, but can you test it ?

PR #3515, opened 2016-09-27
Fixed SetFileTime for mac.

### LINKED ISSUE #3508: Incorrect time stamps with drive redirection
When using drive redirection to copy files from a Windows machine to the host machine (ie. Windows -> macOS) the timestamps are incorrect. They are however correct if copying the other way around (from the host machine to the target (ie. macOS -> Windows)).

In my case, the timestamps seem to be sometime around 2011 all the time even if the copied file has just been created before copying.

I reproduced the issue with both, the X11 client as well as my own macOS based implementation.

For reference, please see the following two screenshots:

Original:
Original, 

After Copy:
After Copy

akallabeth:
Fixes #3508

### DIFF
--- a/winpr/libwinpr/file/file.c
+++ b/winpr/libwinpr/file/file.c
@@ -361,7 +361,7 @@ static BOOL FileSetFileTime(HANDLE hFile, const FILETIME *lpCreationTime,
 	struct stat buf;
 #endif
 /* OpenBSD, NetBSD and DragonflyBSD support POSIX futimens */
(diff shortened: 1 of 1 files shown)


akallabeth:
@lemonmojo Ah, got it. The creation time must be before modification and access time. Since creation time is not handled the times are too old for the system.

lemonmojo:
@akallabeth I see... Any news on a new PR? thx!

PR #3529, opened 2016-10-07
Variadic macro warning fixes

akallabeth:
* C99 compatible variadic macros ```__VA_ARGS__``` (refactored WLog for that)
* Addin loader uses function pointer typedef instead of ```void*``` casts
* Warning fixes (const casts, unused variables)
* Fixed data types for functions (signed / unsigned char casts)


hardening:
Lots of reformating, but it LGTM +1

PR #3531, opened 2016-10-08
client/Mac: Runs Mac password dialog on main UI thread

rjcorrig:
Fixes a bug in which the program terminates with the following error when supplying incomplete credentials to the Mac client:

Terminating app due to uncaught exception 'NSGenericException', reason: '-[NSApplication runModalForWindow:] may only be invoked from the main thread. Behavior on other threads is undefined.'

The submitted change ensures that the runModal dialog runs in the main UI thread and blocks the calling thread until it completes. When the user clicks OK or Cancel, the appropriate return code is saved to a property on the dialog so it can be picked up by the calling thread.


bmiklautz:
@freerdp-bot test.

bmiklautz:
@rjcorrig thanks.

PR #3533, opened 2016-10-10
Glyph cache disable

akallabeth:
deactivate glyph cache for good.

### DIFF
--- a/libfreerdp/core/settings.c
+++ b/libfreerdp/core/settings.c
@@ -434,7 +434,7 @@ rdpSettings* freerdp_settings_new(DWORD flags)
 	settings->NoBitmapCompressionHeader = TRUE;
 	settings->RefreshRect = TRUE;
 	settings->SuppressOutput = TRUE;
-	settings->GlyphSupportLevel = GLYPH_SUPPORT_FULL;
+	settings->GlyphSupportLevel = GLYPH_SUPPORT_NONE;
 	settings->GlyphCache = malloc(sizeof(GLYPH_CACHE_DEFINITION) * 10);
 
 	if (!settings->GlyphCache)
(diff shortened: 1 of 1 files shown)


libfreerdp/core/settings.c
```diff
@@ -272,53 +315,56 @@ rdpSettings* freerdp_settings_new(DWORD flags)
...
 	settings->EncryptionLevel = ENCRYPTION_LEVEL_NONE;
-
 	settings->CompressionEnabled = TRUE;
 
 	if (settings->ServerMode)
-			settings->CompressionLevel = PACKET_COMPR_TYPE_RDP61;
+		settings->CompressionLevel = PACKET_COMPR_TYPE_RDP61;
 	else
-			settings->CompressionLevel = PACKET_COMPR_TYPE_RDP61;
+		settings->CompressionLevel = PACKET_COMPR_TYPE_RDP61;
```
hardening: not your code, but CompressionLevel is always set to PACKET_COMPR_TYPE_RDP61

akallabeth: That is just the default, when using command line arguments you can change that.

akallabeth:
@hardening already added the EXPERIMENTAL warning to the command line option, wouldn't disable it so that someone interested can spot and fix it.

libfreerdp/core/settings.c
hardening: My point was more that the _if (settings->ServerMode)_ is useless as we're setting to PACKET_COMPR_TYPE_RDP61 in all cases.

bmiklautz: @hardening it still can be overridden so it still makes sense.

hardening: @bmiklautz well, my remark still applies: the _if (settings->ServerMode)_ is useless.

PR #3534, opened 2016-10-10
Fixed pointer color mapping.

akallabeth:


akallabeth:
@hardening Needs testing, each arch has it's own color ordering.

Issue #3530, opened 2016-10-08
Slow software rendering  after big October 6 update


giox069:
Hi to all the FreeRDP team (and @antenore). I have seen the big changes in October 6 updates and I'm updating Remmina to support it. I can see some interesting improvements, for example code cleanup by correctly delegating some allocations to libfreerdp and not to the client. Very good.
Currently Remmina make use of libfreerdp's software GDI, like xfreerdp /gdi:sw, and with the commits of october 6, the software GDI rendering is now _noticeably slower_. Are you planning to improve it, or should we try to start to develop some kind of GDI rendering inside Remmina ?
Thank You.

mr-wong:
Not sure about GDI but I'm facing some huge memory leak after last update. Memory usage skyrocketing within 5 mins to 300+mb, sadly I have to make rollback for the time being.

My connection string: xfreerdp /u:user /p:**** /v:host +multitransport +clipboard -decorations /f /bpp:8 +compression /compression-level:2 +bitmap-cache +offscreen-cache +async-update +aero +fonts /gdi:hw /cert-ignore

giox069:
@mr-wong I can confirm memory leaks: I compiled xfreerdp with "-DWITH_GCC_SANITIZE_ADDRESS=on" on cmake commandline. And at the exit, yes, the address sanitizer shows a lot of memory leaks.

hardening:
any chance that you could git-bissect to find the faulty commit ?

giox069:
@hardening: I'm trying to bisect to find the slowness problem. But it's hard, because intermediate steps are sometimes buggy and simply they don't work. For example, a lot of intermediate commit fails when connecting with
```[12:08:25:744] [27852:27853] [ERROR][com.freerdp.core.codecs] - Failed to create h264 codec context```
Is there a temporary workaround for this specific connection error?

akallabeth:
@giox069 Bisecting will probably not be helpful here, mostly rewritten the GDI color handling. A performance profile would be very interesting though.

bmiklautz:
@giox069 if I'm not mistaken does the "the Failed to create h264 codec context" happen if you connect with gfx but don't have openh264 available. So this should be solved if you set Gfx-H264 to FALSE.

Regarding the performance issues it would be quite interesting what functions are generally causing the slowdown. What type of connection do you run? Could you run valgrind with callgrind/cachgrind and see what are the hot spots?

akallabeth:
[...]

PR #3536, opened 2016-10-10
Memory leak fixes

akallabeth:
Fixes 2 nasty memory leaks in pointer and bitmap handling.


akallabeth:
Fixes leaks mentioned in #3530

PR #3541, opened 2016-10-10
Fix nsc client after after color conversion huge fix

realjiangms:
nsc codec client doesn't work:
1. We should mark invalid region for software gdi
2. Checked the code before color conversion fix, the correct color format should be PIXEL_FORMAT_BGRX32_VF (corresponds to old PIXEL_FORMAT_XRGB32_VF)
3. For gdi:hw: xfc->bitmap_buffer is never used/initialized. However gdi->primary_buffer is always maintained. So use primary_buffer to hold the decoded bitmap data

Tested with both gdi:sw and gdi:hw

### DIFF
--- a/client/X11/xf_gdi.c
+++ b/client/X11/xf_gdi.c
@@ -1018,7 +1018,7 @@ static BOOL xf_gdi_surface_bits(rdpContext* context,
 			if (!nsc_process_message(context->codecs->nsc, cmd->bpp, cmd->width,
 			                         cmd->height,
 			                         cmd->bitmapData, cmd->bitmapDataLength,
-			                         xfc->bitmap_buffer, gdi->dstFormat, 0, 0, 0, cmd->width, cmd->height))
+			                         gdi->primary_buffer, gdi->dstFormat, 0, 0, 0, cmd->width, cmd->height))
 			{
 				xf_unlock_x11(xfc, FALSE);
 				return FALSE;
@@ -1026,8 +1026,7 @@ static BOOL xf_gdi_surface_bits(rdpContext* context,
 
 			XSetFunction(xfc->display, xfc->gc, GXcopy);
 			XSetFillStyle(xfc->display, xfc->gc, FillSolid);
-			pSrcData = context->codecs->nsc->BitmapData;
-			pDstData = xfc->bitmap_buffer;
+			pDstData = gdi->primary_buffer;
 			image = XCreateImage(xfc->display, xfc->visual, xfc->depth, ZPixmap, 0,
 			                     (char*) pDstData, cmd->width, cmd->height, xfc->scanline_pad, 0);
 			XPutImage(xfc->display, xfc->primary, xfc->gc, image, 0, 0,
@@ -1042,7 +1041,7 @@ static BOOL xf_gdi_surface_bits(rdpContext* context,
 			XSetFunction(xfc->display, xfc->gc, GXcopy);
 			XSetFillStyle(xfc->display, xfc->gc, FillSolid);
 			pSrcData = cmd->bitmapData;
-			pDstData = xfc->bitmap_buffer;
+			pDstData = gdi->primary_buffer;
 			freerdp_image_copy(pDstData, gdi->dstFormat, 0, 0, 0,
 			                   cmd->width, cmd->height, pSrcData,
 			                   PIXEL_FORMAT_BGRX32_VF, 0, 0, 0, &xfc->context.gdi->palette);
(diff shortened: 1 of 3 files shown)


akallabeth:
@realjiangms Is ```xfc->bitmap_buffer``` actually used anymore?

realjiangms:
I don't see any real reference to xfc->bitmap_buffer. It is always NULL. In previous approach it is initialized with a block of memory. I'm not sure whether it should be removed or not according to your color conversion refactoring

client/X11/xf_gdi.c
```diff
@@ -1018,16 +1018,15 @@ static BOOL xf_gdi_surface_bits(rdpContext* context,
...
+			                         gdi->primary_buffer, gdi->dstFormat, 0, 0, 0, cmd->width, cmd->height))
 			{
 				xf_unlock_x11(xfc, FALSE);
 				return FALSE;
 			}
 
 			XSetFunction(xfc->display, xfc->gc, GXcopy);
 			XSetFillStyle(xfc->display, xfc->gc, FillSolid);
-			pSrcData = context->codecs->nsc->BitmapData;
-			pDstData = xfc->bitmap_buffer;
```
hardening: fun, pSrcData was never used ?

realjiangms: Yes... So remove it. In new approach, image copy is done by nsc_process_message

libfreerdp/codec/nsc.c
```diff
@@ -407,7 +407,7 @@ BOOL nsc_process_message(NSC_CONTEXT* context, UINT16 bpp,
 
 	if (!freerdp_image_copy(pDstData, DstFormat, nDstStride, nXDst, nYDst,
 				width, height, context->BitmapData,
-				PIXEL_FORMAT_BGRA32, 0, 0, 0, NULL))
+				PIXEL_FORMAT_BGRX32_VF, 0, 0, 0, NULL))
```
akallabeth: @realjiangms Could you elaborate how you came to this color format (use case)? I've issues with h264 and windows 10.
See attached image, the menu icons are mirrored upside down
screenshot from 2016-10-11 16 06 27

realjiangms: I got the color format from the code before the 'color conversion' fix. In previous approach, the code is：

        nsc_process_message(xfc->codecs->nsc, cmd->bpp, cmd->width, cmd->height, cmd->bitmapData, cmd->bitmapDataLength);

        XSetFunction(xfc->display, xfc->gc, GXcopy);
        XSetFillStyle(xfc->display, xfc->gc, FillSolid);

        if (xfc->bitmap_size < (cmd->width * cmd->height * 4))
        {
            xfc->bitmap_size = cmd->width * cmd->height * 4;
            xfc->bitmap_buffer = (BYTE*) _aligned_realloc(xfc->bitmap_buffer, xfc->bitmap_size, 16);

            if (!xfc->bitmap_buffer)
            {
                xf_unlock_x11(xfc, FALSE);
                return FALSE;
            }
        }

        pSrcData = xfc->codecs->nsc->BitmapData;
        pDstData = xfc->bitmap_buffer;

        freerdp_image_copy(pDstData, xfc->format, -1, 0, 0,
                    cmd->width, cmd->height, pSrcData, PIXEL_FORMAT_XRGB32_VF, -1, 0, 0, xfc->palette);

================================================
The old PIXEL_FORMAT_XRGB32_VF should corresponds to your new PIXEL_FORMAT_BGRX32_VF

realjiangms: Fortunately, the rfx codec kept PIXEL_FORMAT_XRGB32 in everywhere before the color conversion fix.

akallabeth:
sounds like this could also be solved by adding the _VF flag to destination format in NSC path?

realjiangms:
@akallabeth Yes it does fix the issue if we add VF flag in dstFormat, but it somehow looks confusing, because it is the nsc decompressed bitmap 'flipped', not the destination bitmap. I'm OK with an easy fix anyway.

I just checked the protocol document, nothing mentioned about different format between normal nsc surface_bit command and the gfx clearcodec. Not sure why our client implementation could have this difference. Maybe it's just what mstsc does.

akallabeth:
@realjiangms well, the _VF just means flip vertical (less a color format than a copy instruction) and the thing is somewhere in the spec (hidden really well) it is written that columns are encoded in GDI. Anyway, will push that to #3545 as I've also missed to place the decoded data at the correct image position.

PR #3544, opened 2016-10-11
Multi warn fixes

akallabeth:
Fixed warnings with Release builds (see #3525)


hardening:
I don't understand the rationale with __FILE__ replacement, can you elaborate ?

bmiklautz:
@hardening that's something we have for quite a while but isn't working since some time ;). The idea is if you do a Release build that you get paths relative to the build directory instead of absolute paths on the build machine. One of the nice things about this is that the paths are always the same regardless where you built them.

PR #3545, opened 2016-10-11
Fixed Windows 7 RFX issues.

### LINKED ISSUE #3543: X11 client surface bits with RFX no longer work after color conversion fix
Both gdi:sw and gdi:hw shows black screen with rfx enabled.
For gdi:sw, the reason is that the invalid region is not marked. I fixed it in PR #3541
For gdi:hw, the reason is that the result image is never pasted to X window. 

I tried to fix it by adding XCreateImage/XPutImage, the rfx content appears, however the content doesn't look good. It seems some tiles are not pasted at correct position.

I guess there's something broken with refactor on rfx_process_message. Also it might hit performance. It introduces more memory copy than previous approach.

akallabeth:
Fixes #3543 

* Windows 7 RFX decoding in software GDI used wrong offsets and size.
* X11 decoder used uninitialized region, unified screen update.
* Flipping NSC bitmaps vertically as required (requires #3546 for correct output)

### DIFF
--- a/client/X11/xf_gdi.c
+++ b/client/X11/xf_gdi.c
@@ -999,14 +999,17 @@ static BOOL xf_gdi_surface_bits(rdpContext* context,
 	BYTE* pSrcData;
 	xfContext* xfc = (xfContext*) context;
 	BOOL ret = FALSE;
+	DWORD format;
 	rdpGdi* gdi = context->gdi;
 	xf_lock_x11(xfc, FALSE);
 
 	switch (cmd->codecID)
 	{
 		case RDP_CODEC_ID_REMOTEFX:
+			format = PIXEL_FORMAT_BGRX32;
+
 			if (!rfx_process_message(context->codecs->rfx, cmd->bitmapData,
-			                         PIXEL_FORMAT_BGRX32, cmd->bitmapDataLength,
+			                         format, cmd->bitmapDataLength,
 			                         cmd->destLeft, cmd->destTop,
 			                         gdi->primary_buffer, gdi->dstFormat, gdi->stride,
 			                         gdi->height, NULL))
@@ -1016,10 +1019,12 @@ static BOOL xf_gdi_surface_bits(rdpContext* context,
 			break;
 
 		case RDP_CODEC_ID_NSCODEC:
+			format = FREERDP_VFLIP_PIXEL_FORMAT(gdi->dstFormat);
+
 			if (!nsc_process_message(context->codecs->nsc, cmd->bpp, cmd->width,
-			                         cmd->height,
-			                         cmd->bitmapData, cmd->bitmapDataLength,
-			                         gdi->primary_buffer, gdi->dstFormat, 0, 0, 0, cmd->width, cmd->height))
+			                         cmd->height, cmd->bitmapData, cmd->bitmapDataLength,
+			                         gdi->primary_buffer, format, gdi->stride,
+			                         cmd->destLeft, cmd->destTop, cmd->width, cmd->height))
 				goto fail;
 
 			ret = xf_gdi_update_screen(xfc, cmd, gdi->primary_buffer);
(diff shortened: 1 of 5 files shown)


client/X11/xf_gdi.c
```diff
@@ -970,96 +970,79 @@ static BOOL xf_gdi_surface_update_frame(xfContext* xfc, UINT16 tx, UINT16 ty,
...
-			                                  cmd->height);
-			XSetClipMask(xfc->display, xfc->gc, None);
+
+			if (!freerdp_image_copy(gdi->primary_buffer, gdi->dstFormat, 0, 0, 0,
+			                        cmd->width, cmd->height, pSrcData,
+			                        PIXEL_FORMAT_BGRX32_VF, 0, 0, 0, &xfc->context.gdi->palette))
+				goto fail;
+
+			ret = xf_gdi_update_screen(xfc, cmd, gdi->primary_buffer);
 			break;
```
hardening: maybe all the calls:
_ret = xf_gdi_update_screen(xfc, cmd, gdi->primary_buffer);_ 
could be mutualized at the end of the switch() (excluding the default case of course)

akallabeth: yes, just recognized that the NSCODEC and NONE paths render at the wrong origin (0,0)

client/X11/xf_gdi.c
```diff
@@ -970,96 +970,79 @@ static BOOL xf_gdi_surface_update_frame(xfContext* xfc, UINT16 tx, UINT16 ty,
...
-	REGION16 invalidRegion;
 	xf_lock_x11(xfc, FALSE);
 
 	switch (cmd->codecID)
 	{
 		case RDP_CODEC_ID_REMOTEFX:
 			if (!rfx_process_message(context->codecs->rfx, cmd->bitmapData,
-			                         PIXEL_FORMAT_XRGB32, cmd->bitmapDataLength,
+			                         PIXEL_FORMAT_BGRX32, cmd->bitmapDataLength,
 			                         cmd->destLeft, cmd->destTop,
```
realjiangms: This line should be 0,0. 
Or the bitmap will actually be pasted at cmd->destLeft*2, cmd->destTop*2

realjiangms:
@akallabeth There's something more need to be fixed in libfreerdp/codec/rfx.c. Please check my #3548.
The current offset in tile->data is not correct

client/X11/xf_gdi.c
```diff
@@ -970,96 +970,86 @@ static BOOL xf_gdi_surface_update_frame(xfContext* xfc, UINT16 tx, UINT16 ty,
...
-				ret = FALSE;
-
-			XSetClipMask(xfc->display, xfc->gc, None);
+			ret = xf_gdi_update_screen(xfc, cmd, gdi->primary_buffer);
 			break;
 
 		case RDP_CODEC_ID_NSCODEC:
+			format = FREERDP_VFLIP_PIXEL_FORMAT(gdi->dstFormat);
+
 			if (!nsc_process_message(context->codecs->nsc, cmd->bpp, cmd->width,
```
realjiangms: Since you decide to add VF flag here, would you revert the VF change inside nsc_process_message?

akallabeth: @realjiangms why revert? it is never set in the variable gdi->dstFormat, only in the scoped variable format not used any more

client/X11/xf_gdi.c
```diff
@@ -970,96 +970,86 @@ static BOOL xf_gdi_surface_update_frame(xfContext* xfc, UINT16 tx, UINT16 ty,
...
 	switch (cmd->codecID)
 	{
 		case RDP_CODEC_ID_REMOTEFX:
+			format = PIXEL_FORMAT_BGRX32;
+
 			if (!rfx_process_message(context->codecs->rfx, cmd->bitmapData,
-			                         PIXEL_FORMAT_XRGB32, cmd->bitmapDataLength,
+			                         format, cmd->bitmapDataLength,
 			                         cmd->destLeft, cmd->destTop,
 			                         gdi->primary_buffer, gdi->dstFormat, gdi->stride,
```
akallabeth: @realjiangms will update the pull tomorrow, late here. as for the disadvantage, you can of course always use a pointer directly to the XImage->data element.

realjiangms:
Everything looks fine.
Just one more thing, I think the changes in last function call in nsc_process_message need to be reverted:

    if (!freerdp_image_copy(pDstData, DstFormat, nDstStride, nXDst, nYDst,
                width, height, context->BitmapData,
                PIXEL_FORMAT_BGRX32_VF, 0, 0, 0, NULL))           <============= PIXEL_FORMAT_BGRX32_VF should be reverted to PIXEL_FORMAT_BGRX32
        return FALSE;

realjiangms:
Ignore my previous note ;), already done in #3546 
@akallabeth Merge this PR please

client/X11/xf_gdi.c
realjiangms: Never mind. You've already done it in another PR.

bmiklautz:
@freerdp-bot test.

PR #3547, opened 2016-10-11
Fix nsc server side after color conversion.

realjiangms:
We use new color format system and PIXEL_FORMAT_BGRX32 is the unified buffer color format. Add support for PIXEL_FORMAT_BGRX32 in nsc


akallabeth:
@freerdp-bot test please...

PR #3557, opened 2016-10-15
winpr/mutex: Use PTHREAD_MUTEX_RECURSIVE_NP instead of PTHREAD_MUTEX_RECURSIVE

realjiangms:
Some platform only defines PTHREAD_MUTEX_RECURSIVE_NP without PTHREAD_MUTEX_RECURSIVE in /usr/include/pthread.h


realjiangms:
Oops, seems NP breaks IOS. 
Discard this PR, the original one at least support all platform in jenkins

PR #3562, opened 2016-10-16
client/Mac: Run password dialog as a modal sheet and set window title to server:port

rjcorrig:
The password dialog now allows editing of the domain along with user name and password.
The dialog runs as an application modal sheet attached to the main window in accordance with Apple user interface standards.
The window title, if not set by /t , is set to ServerHostname:ServerPort to give some indication of where we are connected to. Example: server.contoso.com:3389 . This is also used in the "Authenticate to" label on the password dialog.

### DIFF
--- a/client/Mac/MRDPView.m
+++ b/client/Mac/MRDPView.m
@@ -965,9 +965,13 @@ BOOL mac_post_connect(freerdp* instance)
 BOOL mac_authenticate(freerdp* instance, char** username, char** password,
                       char** domain)
 {
+	mfContext* mfc = (mfContext*) instance->context;
+	MRDPView* view = (MRDPView*) mfc->view;
+
 	PasswordDialog* dialog = [PasswordDialog new];
-	dialog.serverHostname = [NSString stringWithCString:
-	                         instance->settings->ServerHostname encoding:NSUTF8StringEncoding];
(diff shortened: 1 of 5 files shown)


akallabeth:
@rjcorrig nice patch. just a thought: modern UI for entering domain is through user field as domain\\user or user@domain. so maybe a proper hint for the field format is sufficient?

rjcorrig:
Thanks. I like that idea. The problem I was trying to solve was if I passed the domain using the /d: flag, there was no indication of that in the UI. Nor if I passed in the user as /u:DOMAIN\\user--in that case, the domain is parsed out into its own field by the time the password dialog is called.

I've tested several combinations of user and domain syntax in the current patch, and I've discovered that DOMAIN\user doesn't work at all: it needs to be broken out into separate user and domain fields. user@domain does work, but only if the domain field is blank.  So I agree that a single user name field is better to reduce the guesswork on the user's part. 

The only issue I see is that very ancient Windows servers may not understand the user@domain syntax, but if I receive it that way in the dialog I can unpack it into separate user and domain fields for mac_authenticate to consume. And if they ignore the hint and use DOMAIN\user anyway, I can handle that too :-) 

Let me grind on this a little more and get back to you.

akallabeth:
Ok, since you are reconsidering, I'll close this pull for now. Feel free to reopen / open another pull if you have something.

PR #3572, opened 2016-10-19
client/mac: Run password dialog as a modal sheet and set window title to server:port

rjcorrig:
Refinement of PR #3562 

The dialog runs as an application modal sheet attached to the main window in accordance with Apple user interface standards.
The window title, if not set by /t , is set to ServerHostname:ServerPort to give some indication of where we are connected to. Example: server.contoso.com:3389 . This is also used in the "Authenticate to" label on the password dialog.

New with this PR:
Password dialog placeholder for user name is username@domain .

The initially displayed text in the user name field is d1\u1 where:
d1 is the domain provided by the /d command line parameter or parsed out of a user name given as /u:domain\user by the command line argument parser and;
u1 is the user string given as /u parsed out by the command line argument parser. This will either be in format "username" or "username@domain" .
If no domain was provided, the user name field is simply u1 .

The user can supply the login credential as either domain\username or username@domain. The resulting login is parsed to break out the domain and user name into separate strings for consumption by the caller.

Should a user supply the domain with both the /d and /u parameters, the user name field will look like d1\d1\u1 or d1\u1@d1 . This is ugly, but maintains consistency with the behavior of the command line argument parser and is indicative of what would actually be passed to the server given those arguments. Seeing the domain name duplicated should lead the user to correct the login name before submitting.

### DIFF
--- a/client/Mac/MRDPView.m
+++ b/client/Mac/MRDPView.m
@@ -965,9 +965,13 @@ BOOL mac_post_connect(freerdp* instance)
 BOOL mac_authenticate(freerdp* instance, char** username, char** password,
                       char** domain)
 {
+	mfContext* mfc = (mfContext*) instance->context;
+	MRDPView* view = (MRDPView*) mfc->view;
+
 	PasswordDialog* dialog = [PasswordDialog new];
-	dialog.serverHostname = [NSString stringWithCString:
-	                         instance->settings->ServerHostname encoding:NSUTF8StringEncoding];
+	dialog.serverHostname = [NSString stringWithFormat:@"%@:%u",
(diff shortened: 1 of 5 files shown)


bmiklautz:
@freerdp-bot test.

rjcorrig:
Having trouble finding astyle 2.4+ . Do you mean 2.04+ ?

akallabeth:
@rjcorrig Yea, typed too fast skipped the 0 ^^

rjcorrig:
OK, got it. Thanks for the tip, indenting was giving me fits.

akallabeth:
@freerdp-bot Care to actually perform a test?

bmiklautz:
@freerdp-bot test.

PR #3578, opened 2016-10-27
Fix server side code to be aware of rereadEvent

realjiangms:


realjiangms:
@hardening that sounds good

libfreerdp/core/peer.c
```diff
@@ -257,6 +256,17 @@ static HANDLE freerdp_peer_get_event_handle(freerdp_peer* client)
...
+{
+	DWORD nCount = 0;
+	nCount += transport_get_event_handles(client->context->rdp->transport, events, count);
+
+	if (nCount == 0)
+		return 0;
+
+	return nCount;
+}
+
```
hardening: This code in this function is really weird.
```
return transport_get_event_handles(client->context->rdp->transport, events, count);
```

would suffice isn't it ?

realjiangms:
@hardening updated

PR #3581, opened 2016-10-28
Fixed surface alignment.

### LINKED ISSUE #3576: Regression: Wrong alignment calculation
Actually this is not a real regression. Because the code before color conversion fix was also wrong.

Code in xf_CreateSurface:
···
    surface->gdi.scanline = surface->gdi.width \* GetBytesPerPixel(
                                surface->gdi.format);

```
    surface->gdi.scanline += (xfc->scanline_pad / 8);
    surface->gdi.scanline -= (surface->gdi.scanline % (xfc->scanline_pad / 8));
```

···

The above code is going to align scanline to scanline_pad
But if scanline_pad is 4(32/8), and width is already multiple of 4bytes, e.g. 5120(1280*4), it will become 5124, which looks not as expected

akallabeth:
Fixes #3576

### DIFF
--- a/client/X11/xf_gfx.c
+++ b/client/X11/xf_gfx.c
@@ -232,8 +232,11 @@ static UINT xf_CreateSurface(RdpgfxClientContext* context,
 
 	if (xfc->scanline_pad > 0)
 	{
-		surface->gdi.scanline += (xfc->scanline_pad / 8);
-		surface->gdi.scanline -= (surface->gdi.scanline % (xfc->scanline_pad / 8));
+		const UINT32 align = xfc->scanline_pad / 8;
+		const UINT32 pad = align - (surface->gdi.scanline % align);
+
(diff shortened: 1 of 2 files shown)


realjiangms:
The code looks good. 
In X11 lib there's more efficient method:

(((nbytes) + ((pad) - 1)) & ~(long)((pad) - 1))

But it only apply to alignment of power of 2.

realjiangms:
Would you mind also fix other places?

    cacheEntry->scanline = (cacheEntry->width + (cacheEntry->width % 4)) * 4;
in gdi_SurfaceToCache

realjiangms:
Can we get this fix checked in?
I think we need approval from some reviewer with write access ;)

akallabeth:
@hardening @bmiklautz @mfleisz @nfedera Could you have a look?

Issue #3422, opened 2016-06-23
rdpgfx or tsmf | freerdp session crashing with all local video formats


T3N4K4:
FreeRDP Master - Git Revision 0d7a6b5
Applied patches http://nopaste.narf.at/show/330331/

All cmake Options:
http://pastebin.com/fwQbSPZx

Connection server -> Windows Server 2012 R2 Standard 64-Bit
Client system -> Linux vagrant-ubuntu-trusty-32 3.13.0-85-generic #129-Ubuntu SMP Thu Mar 17 20:50:41 UTC 2016 i686 GNU/Linux

Local video formats tried: mp4, AVCHD-Video, divx (videoclip), wmv
(played via network drive)

**Command Line:**
/opt/freerdp/bin/xfreerdp /u:$USER /p:$PASS /d:WINDOWS.LOCAL +sec-tls -sec-nla /rfx /gfx /multimedia:decoder:gstreamer /sound:sys:pulse /audio-mode:0 /cert-ignore /v:$SERVER

Errors with **/multimedia:decoder:gstreamer** ->
[11:20:32:675] [1674:1675] [ERROR][com.winpr.timezone] - Unable to find a match for unix timezone: Etc/UTC
[11:20:40:568] [1674:1678] [ERROR][com.freerdp.channels.tsmf.client] - Failed to create new pipe
[11:20:40:569] [1674:1678] [WARN][com.freerdp.channels.tsmf.client] - Format not supported by decoder gstreamer
[11:20:40:596] [1674:1678] [ERROR][com.freerdp.channels.tsmf.client] - Failed to create new pipe
[11:20:40:597] [1674:1678] [WARN][com.freerdp.channels.tsmf.client] - Format not supported by decoder gstreamer
[11:20:40:934] [1674:1678] [ERROR][com.freerdp.channels.drdynvc.client] - ChannelId 10 not found!
[11:20:40:934] [1674:1678] [ERROR][com.freerdp.channels.drdynvc.client] - drdynvc_order_recv failed with error 1359!
[11:20:40:941] [1674:1675] [ERROR][com.freerdp.core] - drdynvc_virtual_channel_client_thread reported an error. Error was 1359
[11:20:40:941] [1674:1675] [ERROR][com.freerdp.client.x11] - Failed to check FreeRDP file descriptor
[11:20:32:581] [1674:1675] [INFO][com.freerdp.client.common.cmdline] - loading channel rdpdr
[...]

PR #3584, opened 2016-11-03
Fix issue #3422, ignore some dynamic virtual channel messages with invalid ids

awakecoding:
MMR on Windows Server 2012 R2 is buggy, and can send some messages for Microsoft::Windows::RDS::Geometry::v08.01 even if the channel wasn't opened on the client. This is a violation of the protocol, but ignoring it works. This fixes https://github.com/FreeRDP/FreeRDP/issues/3422


bmiklautz:
The commit message is a little bit confusing as it would just ignore every invalid channel id.

PR #3591, opened 2016-11-11
Fixed library loader paths for android

akallabeth:
On android ```dlopen``` is quite picky with relative paths.
Since everything is always in the same application folder strip the paths altogether.


akallabeth:
@freerdp-bot come on, please test!

akallabeth:
@freerdp-bot test

akallabeth:
@mfleisz You mean in code or git? As for the dot, we need a valid relative path, when using empty defines we would end up with absolute paths

Issue #3300, opened 2016-04-20
 error! Status Code: 0


tondahermankovi:
When connecting to Microsoft Server 2008 through TS gateway using command 

xfreerdp  /v:server /d:domain /u:user /p:password 

I get the following error 

[23:38:11:377] [12402:12403] [WARN][com.freerdp.core.gateway.http] - http_response_recv: text/plain unexpected body length: actual: 13, expected: 0
[23:38:11:398] [12402:12403] [WARN][com.freerdp.core.gateway.http] - http_response_recv: text/plain unexpected body length: actual: 13, expected: 0
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core.gateway.rpc] - error! Status Code: 0
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core.gateway.http] - Access DeniedHTTP/1.1 200 Success
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core.gateway.http] - Content-Length: 1073741824
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core.gateway.http] - Content-Type: application/rpc
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core.gateway.http] - Server: Microsoft-HTTPAPI/2.0
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core.gateway.http] - Date: Wed, 20 Apr 2016 21:38:11 GMT
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core.gateway.tsg] - tsg_check failure
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core.nego] - Protocol Security Negotiation Failure
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED [0x2000C]
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.core.connection] - Error: protocol security negotiation or connection failure
[23:38:11:416] [12402:12403] [ERROR][com.freerdp.client.x11] - Freerdp connect error exit status 1

Using MS remote desktop on Windows 7/10 or Android works fine with the server.
Need help, what the error status code 0 means and why there is unexpected body lengt. 

xfreerdp --version 
FreeRDP version 2.0.0-dev (git ca2d015)
OS: Linux 14.04.1-Ubuntu x86_64 x86_64 GNU/Linux
[...]

Issue #3479, opened 2016-08-24
Gateway HTTP parsing problem leads to connection failure


ccpp:
I can't connect via gateway anymore, since our company's server changed its 401 page.
I suspect the HTTP body not to be consumed/skipped correctly in http_response_recv (libfreerdp/core/gateway/http.c).

See this line in my log below: "Access DeniedHTTP/1.1 200 Success"
-> Status 200 is not parsed correctly

    [01:09:46:552] [4789:4790] [INFO][com.freerdp.client.x11] - Authentication only. Don't connect to X.
    [01:09:46:938] [4789:4790] [WARN][com.freerdp.core.gateway.http] - http_response_recv: text/html; charset=us-ascii unexpected body length: actual: 341, expected: 0
    [01:09:47:427] [4789:4790] [WARN][com.freerdp.core.gateway.http] - http_response_recv: text/plain unexpected body length: actual: 13, expected: 0
    [01:09:47:510] [4789:4790] [WARN][com.freerdp.core.gateway.http] - http_response_recv: text/plain unexpected body length: actual: 13, expected: 0
    [01:09:47:585] [4789:4790] [ERROR][com.freerdp.core.gateway.rpc] - error! Status Code: 0
    [01:09:47:585] [4789:4790] [ERROR][com.freerdp.core.gateway.http] - Access DeniedHTTP/1.1 200 Success
    [01:09:47:585] [4789:4790] [ERROR][com.freerdp.core.gateway.http] - Content-Length: 1073741824
    [01:09:47:585] [4789:4790] [ERROR][com.freerdp.core.gateway.http] - Content-Type: application/rpc
    [01:09:47:585] [4789:4790] [ERROR][com.freerdp.core.gateway.http] - Server: Microsoft-HTTPAPI/2.0
    [01:09:47:585] [4789:4790] [ERROR][com.freerdp.core.gateway.http] - Date: Wed, 24 Aug 2016 23:09:46 GMT
    [01:09:47:585] [4789:4790] [ERROR][com.freerdp.core.gateway.tsg] - tsg_check failure
    [01:09:47:586] [4789:4790] [ERROR][com.freerdp.core.nego] - Protocol Security Negotiation Failure
    [01:09:47:586] [4789:4790] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED [0x2000C]
[...]

PR #3593, opened 2016-11-13
Bugfix in RDP Gateway HTTP: read/skip full HTTP body

ccpp:
Bugfix in RDP Gateway HTTP: read/skip full HTTP body for non-RPC responses

This fixes issues #3300 and #3479, but it reveals that there might be a timeout issue with /gt:http and /gt:auto

So test with both options /gt:rpc and /gt:http please

### DIFF
--- a/libfreerdp/core/gateway/http.c
+++ b/libfreerdp/core/gateway/http.c
@@ -737,16 +737,6 @@ HttpResponse* http_response_recv(rdpTls* tls)
 
 			response->BodyLength = Stream_GetPosition(s) - payloadOffset;
 
-			if (response->BodyLength > 0)
-			{
-				response->BodyContent = (BYTE*) malloc(response->BodyLength);
-
-				if (!response->BodyContent)
-					goto out_error;
-
-				CopyMemory(response->BodyContent, payload, response->BodyLength);
-			}
-
 			bodyLength = 0; /* expected body length */
 
 			if (response->ContentType)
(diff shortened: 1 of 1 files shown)


akallabeth:
@ccpp Nice catch! Just had a look at the whole code, looks like there is still an issue.
The whole data reading part must be done before any parsing is attempted as ```strtok``` modifies the data and if only a part of the header is read the next attempt will fail.

ccpp:
Hi akallabeth,
I do not think there is a problem, because:

1. The `Content-Length` header MUST have been parsed before this step, otherwise we do not even know how much data is missing.
2. The header parsing has been fully accomplished. This code is in an if-block which even breaks the outer reading loop.
3. Header lines are counted beforehand, therefore `strtok` is only called for the HTTP header lines; HTTP body will never be `strtok`'d (save there's another bug. The code looks a little complicated, but I guess they had to do it this way because you cannot give `strtok` a buffer length)

hardening:
The design of these code should be retreated completely, anyway your fix seems to do the job.

just a note not part of your changes: if the content-length is really big we take the risk to memory explode when reallocating the stream

hardening:
@freerdp-bot test good bot, test !

akallabeth:
@ccpp Yes, you're right. Just don't like the overall state of the code for http gateway as @hardening already stated.
@hardening I'd suggest lets merge it to fix the gateway issues for the time being and create a new issue for refactoring the module?

hardening:
@akallabeth +1

Issue #3592, opened 2016-11-12
Drawing problem with software rendering


giox069:
This may be a regression problem after Oct6 big update: running xfreerdp with /gdi:sw I have blue internal background of progress bar in Acronis Backup and Recovery 10 (under Win2008):

![freerdp mail ipablapieve it_012](https://cloud.githubusercontent.com/assets/5477803/20240643/c7615bfe-a91d-11e6-98b7-4911f0c4408a.png)

Using hardware rendering (without /gdi:sw) all is perfect:

![freerdp mail ipablapieve it_013](https://cloud.githubusercontent.com/assets/5477803/20240648/e0e37f9e-a91d-11e6-9dbc-b250f31d37e6.png)

giox069:
Also on "File Server Resource Manager" of Windows Server 2008:

broken /gdi:sw
![freerdp mail ipablapieve it_014](https://cloud.githubusercontent.com/assets/5477803/20240975/57560da0-a927-11e6-89e7-6a58ee6e1687.png)

/gdi:hw
![freerdp mail ipablapieve it_016](https://cloud.githubusercontent.com/assets/5477803/20240977/66561822-a927-11e6-824a-bdbcf0b64cd0.png)

akallabeth:
@giox069 Maybe there is another way to diagnose the issue. Could you add logging to the callbacks registered in ```gdi_register_update_callbacks``` and shortly connect to a screen where the color issue occurs? That could narrow down the issue.

akallabeth:
@giox069 Does #3680 fix this issue for you?

giox069:
The problem with "File Server Resource Manager", the second post, seems to be already fixed in the past.
And yes, #3680 or one of the last 3-5 days commits also fixes the main issue with Acronis TrueImage on 1st post.
Thank you! :+1:

PR #3594, opened 2016-11-14
Fixed line drawing color format #3592

akallabeth:
The color format for line drawing incorrectly used the source color format instead of the destination color format on color values already converted to destination format.

### DIFF
--- a/libfreerdp/gdi/gdi.c
+++ b/libfreerdp/gdi/gdi.c
@@ -696,14 +696,13 @@ static BOOL gdi_line_to(rdpContext* context, const LINE_TO_ORDER* lineTo)
 {
 	UINT32 color;
 	HGDI_PEN hPen;
-	UINT32 SrcFormat;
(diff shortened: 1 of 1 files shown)


akallabeth:
@giox069 Does this fix #3592?

giox069:
@akallabeth this fixes my second post in issue #3592 (colored lines in FSRM and other tools).

I'm still having problems as in the 1st post of #3592, but the fill color now seems to be different:
freerdp mail ipablapieve it_019

akallabeth:
@giox069 Does this also happen with other applications? Trying to reproduce but don't have acronis around.

giox069:
No, unfortunately I cannot find another application to reproduce. If you can, leave the issue open and I will complete as soon as I will find another application which exposes the same problem as Acronis B&R.

akallabeth:
@freerdp-bot test

PR #3597, opened 2016-11-15
Virtual Channel Mapping Fix

awakecoding:
* remove channels thread-local storage usage
* restore global tables for virtual channel mappings
* add support for extended virtual channel API
* port drdynvc to the extended virtual channel API

This restores proper support for multi-instance in the FreeRDP libraries, especially when thread-local storage is not supported. Usage of global tables was the solution so far, and while suboptimal, it worked. The real solution is to migrate all the internal virtual channels to the extended virtual channel API, such that we have enough parameters to pass custom pointers and avoid the mapping entirely. Both versions of the virtual channel API are supported, and there are 7 remaining virtual channels to port (cliprdr, encomsp, rail, rdpdr, rdpsnd, remdesk).


awakecoding:
@freerdp-bot test.

PR #3603, opened 2016-11-17
X11/client: call gdi resize on hw desktop resize. 

realjiangms:
The gdi primary buffer is used in client/X11/xf_gdi.c, it should be big enough to hold the full screen update. So we need to resize it once the session is resized

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -461,10 +461,17 @@ static BOOL xf_hw_end_paint(rdpContext* context)
 
 static BOOL xf_hw_desktop_resize(rdpContext* context)
 {
+	rdpGdi* gdi = context->gdi;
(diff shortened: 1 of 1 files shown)


akallabeth:
@realjiangms shouldn't that buffer be untouched when using X11 accelerated drawing?

realjiangms:
@akallabeth Unfortunately you decided to use it in xf_gdi.c, bit surface update subroutine. I'm not sure whether it's original intent or not ;)

realjiangms:
@akallabeth Thanks. 
But the automatic test seems stuck, jenkins:

Configuration linux ? ubuntu is still in the queue: ubuntu is offline

Could you help to call @freerdp-bot to retest?

akallabeth:
@freerdp-bot test

PR #3612, opened 2016-11-23
Fixed broken scanline alignment.

akallabeth:


akallabeth:
@freerdp-bot test

realjiangms:
@akallabeth Why revert this back? Does anything in #3576 and #3581 wrong?
Then we still have issue when scanline = 5120(1280*4) and scanline_pad =32. It will finally be 5124! Is this what we expect?

akallabeth:
@realjiangms this was breaking in all cases except yours.

realjiangms:
@akallabeth 
Shouldn't we find a way that it doesn't break any case? Could you give an example which the previous implementation breaks something?

akallabeth:
@realjiangms xfreerdp with default resolution and gfx had lines not aligned to each other.

akallabeth:
of course we schould fix it for your case too.

realjiangms:
I double checked the previous implementation, even take a very strange example:
solution = 1366 *768, format is 8bit color.  Let's assume a very strange xfc->scanline_pad = 24(3bytes)
scanline = 1366 * 1, align = 3

Previous impl:
scanline becomes 1366 + (3-1366%3) = 1368

This PR:
scanline = 1366 + 3 - (1366%3) = 1368

So the previous implementation still works same as this PR's fix.

realjiangms:
@akallabeth 
xfreerdp default resolution is 1024x768, while in 24/32 bit X org, the format is BGRX/BGRA, which is 4 byte.
Then scanline_pad probably is 32.

I think you mean surface->dgi.width = 1024, and surface->dgi.scanline will be 
4096 for previous implementation 
and
4100 after this PR's fix.

Is 4100 correct for 32bit 1024 width?
Usually more buffer could emit some overflow issue. I hope this is not the reason we need to revert it

akallabeth:
@realjiangms no overflow, only a offset calculation error. need to find that to eliminate the issue.

akallabeth:
will check if there is an additional location calculating the scanline. that could explain the offset issue.

PR #3615, opened 2016-11-23
Support for OpenSSL 1.1.0

nfedera:


libfreerdp/common/assistance.c
```diff
@@ -93,18 +92,10 @@ int freerdp_assistance_crypt_derive_key_sha1(BYTE* hash, int hashLength, BYTE* k
 	if (!buffer)
 		goto fail;
 
-	if (!winpr_SHA1_Init(&hashCtx))
-		goto fail;
-	if (!winpr_SHA1_Update(&hashCtx, pad1, 64))
-		goto fail;
-	if (!winpr_SHA1_Final(&hashCtx, buffer, hashLength))
+	if (!winpr_Digest(WINPR_MD_SHA1, pad1, 64, buffer, hashLength))
```
bmiklautz: Nice!

libfreerdp/crypto/tls.c
```diff
@@ -45,6 +46,30 @@
...
+ * Currently the only (runtime) workaround is setting an openssl tls
+ * extension debug callback that sets the SSL context's servername_done
+ * to 1 which effectively disables the parsing of that extension type.
+ *
+ * Nowadays this workaround is not required anymore but still can be
+ * activated by adding the following define:
+ *
+ * #define MICROSOFT_IOS_SNI_BUG
+ */
+
```
bmiklautz: \o/\o/

mfleisz:
Tried to compile with VS 2010 and VS 2013 against OpenSSL 1.1.0c - both fails with 100s of errors in the OpenSSL headers.
Compiling against 1.0.2g works fine.
Did anyone else managed to get a working Windows build?

nfedera:
I've compiled my branch against OpenSSL 0.9.8, 1.0.1 and 1.1.0 and did the following tests with each build (linux only):
- verified via ldd that the correct OpenSSL library is used
- run ctests
- connect with ```/sec:tls```
- connect with ```/sec:nla```
- connect with ```/sec:rdp ```
- connect with ```/sec:rdp /encryption-methods:40```
- connect with ```/sec:rdp /encryption-methods:56```
- connect with ```/sec:rdp /encryption-methods:128```
- connect with ```/sec:rdp /encryption-methods:FIPS```
- all of the above via gateway using ```/g:tsgatwaysrv```

OpenSSL 1.1.0 is not yet on the build bots we know that the Mac, Windows, IOS and Android builds are working too with the older OpenSSL versions.

So I guess one can safely press on merge ...

nfedera:
@akallabeth right. I forgot to disable to disable ERR_remove_thread_state if OpenSSL >= 1.1.0 !
However, the function is still there and only deprecateded. So it should have worked. Maybe your mac build has no deprecation enabled? I've fixed this in the last commit.

PR #3619, opened 2016-11-24
Ignore alpha channel in freerdp_image_copy

### LINKED ISSUE #3616: gfx channel: performance extremely decreased
I don't have the exact numbers right now but with ```/gfx``` or ```/gfx /rfx``` or ```/gfx-h264``` xfreerdp is slower by a factor of 10 compared to a build from march this year.

Will bisect to find the culprit when I have time ...

akallabeth:
When copying image data consider formats that only differ on use
of alpha data equal. This allows using the optimized copy routine
instead of the slower color conversion routine. Fixes #3616

### DIFF
--- a/include/freerdp/codec/color.h
+++ b/include/freerdp/codec/color.h
@@ -100,6 +100,14 @@ typedef struct gdi_palette gdiPalette;
 extern "C" {
 #endif
 
+/* Compare two color formats but ignore differences in alpha channel.
+ */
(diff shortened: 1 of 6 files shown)


bmiklautz:
@akallabeth what factor are we talking about with this change?

akallabeth:
@bmiklautz would say at least 5x (have only done limited cachegrind analysis)

PR #3620, opened 2016-11-24
wtsapi: Fix issue with setting custom API table

mfleisz:


mfleisz:
@bmiklautz 
I.e. when you set the table to the FreeRDP WtsApi using

```
WTSRegisterWtsApiFunctionTable(FreeRDP_InitWtsApi());
```

As soon as you then call WTSOpenServer - the wtsapi wrapper will check if the function table is already initialized (using InitOnceExecuteOnce). As InitOnceExecuteOnce has not been called before it will initialize with the default WtsApi on that platform - in case of Windows that will be the native wtsapi.dll.

As soon as the server now calls a function that uses FreeRDP WtsApi internals (in my case it was WTSVirtualChannelManagerGetFileDescriptor) we have a crash because the channel manager was never initialized (because the native WTSOpenServer function has been called instead of FreeRDP_WTSOpenServer).

bmiklautz:
@mfleisz thanks for explaining. Now we have it in the "history" as well ;)

PR #3622, opened 2016-11-24
Ungrab keyboard feature for X11

khvMX:
Issue: https://github.com/FreeRDP/FreeRDP-old/issues/49
What is missing, is that the key is not configurable (I am using Ctrl+Alt, since I can do Ctrl+Alt+Up/Down to switch workspaces in Gnome3). I have no idea how, but this must be made configurable..
Hope that someone with the insight can do it right!

### DIFF
--- a/client/X11/xf_event.c
+++ b/client/X11/xf_event.c
@@ -496,6 +496,8 @@ static BOOL xf_event_KeyPress(xfContext* xfc, XEvent* event, BOOL app)
 static BOOL xf_event_KeyRelease(xfContext* xfc, XEvent* event, BOOL app)
 {
 	XEvent nextEvent;
+	KeySym keysym;
+	char str[256];
 
 	if (XPending(xfc->display))
 	{
@@ -509,7 +511,8 @@ static BOOL xf_event_KeyRelease(xfContext* xfc, XEvent* event, BOOL app)
 		}
 	}
 
-	xf_keyboard_key_release(xfc, event->xkey.keycode);
+	XLookupString((XKeyEvent*) event, str, sizeof(str), &keysym, NULL);
+	xf_keyboard_key_release(xfc, event->xkey.keycode, keysym);
 	return TRUE;
 }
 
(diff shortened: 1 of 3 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

khvMX:
Hi akallabeth, thanks for feedback. As I said, this must be configurable. Although we could just settle for Right Ctrl which sound eligible to me, it should then just directly ungrab keyboard like VirtualBox does, and not send any additional events. Would that be fine?

akallabeth:
@khvMX Sounds fine too. The current version breaks behaviour with fullscreen toggle ctrl + alt + enter, which now loses keyboard grab.

khvMX:
@akallabeth I've implemented the Right-Ctrl version. I've played with it for a little while and noticed that better behavior would be if it only ungrabbed keyboard when the Right Ctrl key is pressed and directly released, since it could then be still used in IDEs/text editors for Ctrl-moving or Ctrl-Shift-selecting by words. Currently it is not possible because it ungrabs directly after pressing Right Ctrl. Does anybody have an idea on how to detect both "press" and "release" events in the "client/X11/xf_keyboard.c" file?

akallabeth:
@khvMX Well, the ```xf_keyboard_key_press``` and ```xf_keyboard_key_release``` functions are the place you're looking for. Maybe you need to remember some state somewhere to know what was your last key press

akallabeth:
@bmiklautz Could you have a look why the build is not running through?

bmiklautz:
@akallabeth just give it a retest ;)

akallabeth:
@freerdp-bot test please!

Rudedog9d:
@khvMX Glad to see this is going somewhere! Configurable would still be nice, is there already a configuration file that FreeRDP uses?

khvMX:
@freerdp-bot test again? :)

khvMX:
@Rudedog9d not that I know of. At least there is nothing like that in my ~/.freerdp folder. I did not see any global configurations in x11 code as well. Maybe the only way to configure something is CLI arguments. I am not sure how happy would the FreeRDP community be to introduce anything like this in public interface though. I'd suggest opening a new issue for it if this PR gets merged.

akallabeth:
@freerdp-bot test please, don't be like that!
@khvMX Sorry, member only feature ;)

PR #3635, opened 2016-12-01
Updated build tools and readme for android.

akallabeth:
* Updated gradle and android build tools
* Updated readme
* Replaced ```int``` with ```long``` for 64bit architecture support.
* Fixed compiler warnings

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -773,7 +773,7 @@ BOOL rdp_client_connect_auto_detect(rdpRdp* rdp, wStream* s)
 		{
 			if (channelId == rdp->mcs->messageChannelId)
 			{
-				UINT16 securityFlags;
+				UINT16 securityFlags = 0;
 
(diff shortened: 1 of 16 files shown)


akallabeth:
@freerdp-bot test

libfreerdp/core/connection.c
```diff
@@ -773,7 +773,7 @@ BOOL rdp_client_connect_auto_detect(rdpRdp* rdp, wStream* s)
 		{
 			if (channelId == rdp->mcs->messageChannelId)
 			{
-				UINT16 securityFlags;
+				UINT16 securityFlags = 0;
```
bmiklautz: Was this a compiler warning? - As the securityFlags are always initialized because they are read below.

akallabeth: @bmiklautz Yes, there was a warning during android compile. I've just initialized it for every case where the read / write functions are used.

akallabeth:
@bmiklautz ok, done

PR #3637, opened 2016-12-01
codec: Remove src format parameter from rfx_process_message

mfleisz:
This parameter is not required. Default param is BGRX32 and if an other format is required rfx_context_set_pixel_format can be used.


bmiklautz:
@akallabeth better change it now than having it broken or unclear.

PR #3643, opened 2016-12-05
Misc fixes

mfleisz:


libfreerdp/codec/color.c
```diff
@@ -127,7 +127,7 @@ BOOL freerdp_image_copy_from_monochrome(BYTE* pDstData, UINT32 DstFormat,
 		}
 	}
 
-	return 1;
+	return TRUE;
 }
 
 static INLINE UINT32 freerdp_image_inverted_pointer_color(UINT32 x, UINT32 y,
```
mfleisz: In this PR I just changed the return value to fit the function signature. The function generally seems to need some overhaul (i.e. palette is never used, params are not checked, return value of WriteColor is not checked, ...).

akallabeth:
@freerdp-bot test

PR #3653, opened 2016-12-12
cmake: Fix wrong pdb name

mfleisz:
When building without "client interface", cmake incorrectly tried to copy wfreerdp-client.pdb instead of wfreerdp.pdb


akallabeth:
@mfleisz Is there no way to get the name of the target without hard coding it?

mfleisz:
@akallabeth Updated the PR (now querying OUTPUT_NAME for the pdb filename)

PR #3655, opened 2016-12-15
Fix wtypes and format strings

nfedera:
rewrite of wtypes.h
- some simplification
- comply with [MS-DTYP] sizes
- added string format defines if inttypes.h is n/a
- added string format defines for size_t

fix string format specifiers:
- fixed invalid, missing or additional arguments
- removed all type casts from arguments
- added missing (void*) typecasts for %p arguments
- use inttypes defines where appropriate


mfleisz:
@nfedera PR needs rebasing with master

PR #3657, opened 2016-12-16
New argument /auto-reconnect-max-retries:<retries>

akallabeth:
Exposes autoreconnect retries via command line argument.


client/common/cmdline.c
```diff
@@ -2360,6 +2362,13 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 		{
 			settings->AutoReconnectionEnabled = arg->Value ? TRUE : FALSE;
 		}
+		CommandLineSwitchCase(arg, "auto-reconnect-max-retries")
+		{
+			settings->AutoReconnectMaxRetries = atoi(arg->Value);
+
+			if (settings->AutoReconnectMaxRetries <= 0)
```
akallabeth: @nfedera Yes, so we need some sanity limit, what would you suggest? 1000?

PR #3658, opened 2016-12-16
Fix -Wundef with GCC 6.2.1 in wtypes.h

xhaakon:
Undefined preprocessor macro in wtypes.h produces:

  warning: "__ILP64__" is not defined [-Wundef]

This causes unnecessary hassle to people who compile their FreeRDP-based
applications with -Werror.


nfedera:
@freerdp-bot test

PR #3661, opened 2016-12-19
Error status for DNS errors

akallabeth:
* Setting DNS lookup error for connects
* Print an error if more than one last error is set.


akallabeth:
@bmiklautz @mfleisz @nfedera @MartinHaimberger Anyone time to review/merge this one?

PR #3664, opened 2016-12-24
Server/shadow: Fix handling for gfx acknowledge according to spec

realjiangms:
Fix handling for gfx acknowledge according to spec.

[MS-RDPEGFX]:
3.2.5.13  Processing an RDPGFX_FRAME_ACKNOWLEDGE_PDU message
If the queueDepth field is less than 0xFFFFFFFF, the server MUST expect that
RDPGFX_FRAME_ACKNOWLEDGE_PDU messages will continue to be sent by the client.
Furthermore, if the queueDepth field is in the range 0x00000001 to 0xFFFFFFFE the server SHOULD
use this value to determine how far the client is lagging in terms of graphics decoding and then
attempt to throttle the graphics frame rate accordingly.
If the queueDepth field is set to SUSPEND_FRAME_ACKNOWLEDGEMENT (0xFFFFFFFF), the server
MUST clear the Unacknowledged Frames (section 3.2.1.2) ADM element and MUST NOT expect any
further RDPGFX_FRAME_ACKNOWLEDGE_PDU messages from the client. In this mode, the server
MUST NOT wait or block on unacknowledged frames (as the
RDPGFX_FRAME_ACKNOWLEDGE_PDU message is not sent by the client) and MUST assume that
the client is able to decode graphics data at a rate faster than it is receiving frames.

On the other hand, RDPGFX_QOE_FRAME_ACKNOWLEDGE_PDU SHOULD only be used for informational and debugging
purposes and should not be taken into account.

### DIFF
--- a/server/shadow/shadow_client.c
+++ b/server/shadow/shadow_client.c
@@ -594,21 +594,19 @@ static BOOL shadow_client_surface_frame_acknowledge(rdpShadowClient* client,
         UINT32 frameId)
 {
 	shadow_client_common_frame_acknowledge(client, frameId);
+	/*
+	 * Reset queueDepth for legacy none RDPGFX acknowledge
+	 */
+	client->encoder->queueDepth = QUEUE_DEPTH_UNAVAILABLE;
 	return TRUE;
 }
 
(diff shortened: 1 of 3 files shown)


akallabeth:
@nfedera What is your opinion on this one?

PR #3669, opened 2016-12-28
Limit MultifragMaxRequestSize in server mode

xhaakon:
With large desktop resolutions also our preferred max request size
increases, but in practice I've found that at least the client I use
(MS Lync 2013) would show only a black screen when the value is greater
than 0x3EFFFF.

Cap MultifragMaxRequestSize to its initial value, which is currently
0xFFFF or can be user-set at will.

### DIFF
--- a/libfreerdp/core/capabilities.c
+++ b/libfreerdp/core/capabilities.c
@@ -2477,7 +2477,7 @@ static BOOL rdp_write_multifragment_update_capability_set(wStream* s,
 	if (!Stream_EnsureRemainingCapacity(s, 32))
 		return FALSE;
 
-	if (settings->ServerMode)
+	if (settings->ServerMode && settings->MultifragMaxRequestSize == 0)
 	{
 		/**
 		 * In server mode we prefer to use the highest useful request size that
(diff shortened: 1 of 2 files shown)


nfedera:
I don't think this is a good solution. There has been a lot of work and testing done in the past in order to get this badly documented MaxRequestSize value for the fragmented fast path updates right for a large amount of clients (including almost 10 different Windows CE/mobile variants)!

A maximum value of 0xFFFF would probably cause massive unnecessary fast path update fragmentation.

Is your server sending RemoteFX in your black screen case ?

If so then according to [MS-RDPRFX] 1.5  the client MUST use a value greater than or equal to the value we've previously sent in the server to client multi-fragment update capability set.

What value does this client actually send in the client to server  "client multi-fragment update capability set" if you announce a value greater than 0x3EFFFF ?

xhaakon:
[Here](https://github.com/xhaakon/FreeRDP/commit/0cc960ad29659a7995b9d692b1fcef9b3cbfd084) I offer an alternative solution to setting a custom MultifragMaxRequestSize which doesn't change the default behavior, if it would be more acceptable for you.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot stop messing around, test!

nfedera:
@xhaakon +1 for your [alternative solution](https://github.com/xhaakon/FreeRDP/commit/0cc960ad29659a7995b9d692b1fcef9b3cbfd084).

I'll merge once you've updated or replaced this PR.

xhaakon:
@nfedera I've updated this PR to match the alternate solution & rebased the commit to current master.

akallabeth:
@freerdp-bot test this please

PR #3677, opened 2017-01-06
Added missing pointer.SetPosition to corrected crash

alex7ce:
Crash happens if mouse pointer position changed programmatically.

### DIFF
--- a/client/Windows/wf_graphics.c
+++ b/client/Windows/wf_graphics.c
@@ -318,6 +318,14 @@ static BOOL wf_Pointer_SetDefault(rdpContext* context)
 	return TRUE;
 }
 
+static BOOL wf_Pointer_SetPosition(rdpContext* context, UINT32 x, UINT32 y)
+{
+	if (!context)
+		return FALSE;
+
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test.

akallabeth:
@freerdp-bot Oh come on, test properly

client/Windows/wf_graphics.c
```diff
@@ -318,6 +318,14 @@ static BOOL wf_Pointer_SetDefault(rdpContext* context)
 	return TRUE;
 }
 
+static BOOL wf_Pointer_SetPosition(rdpContext* context, UINT32 x, UINT32 y)
+{
+	if (!context)
+		return FALSE;
+
+	return TRUE;
+}
```
hardening: return context != NULL ?

alex7ce: A good tip, however, this is only a placeholder to prevent crash. So inside it there will be more code. It should probably move a cursor on a client machine appropriately.

PR #3688, opened 2017-01-10
Clear codec buffer resize fix

akallabeth:


hardening:
There's misc occurrences of "const UINT32 DstFormat", I don't see the point for these changes.

akallabeth:
@hardening Well, just to ensure that it is not accidentally changed in the function.

hardening:
@akallabeth really I don't see the point to add a const modifier on a value that is passed by copy...

akallabeth:
@hardening A guarantee for the compiler that the value must not be changed within the function body. May be redundant but if it is there it is less likely to be overwritten by accident. (which was the point, I was tracing a invalid colour format, and once there it does not hurt to have it.)

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@hardening @mfleisz Ok, removed the const.

PR #3692, opened 2017-01-11
fix: gdi stride default zero initialization

jcrespo-asg:
when gdi_init_primary gets called with stride=0 then gdi->stride does not get initialize, this can cause problems as it can be init to a positive random value that will get used.

If set to 0 then gdi_CreateBitmapEx will correctly set it to nw*bpp

### DIFF
--- a/libfreerdp/gdi/gdi.c
+++ b/libfreerdp/gdi/gdi.c
@@ -1056,8 +1056,7 @@ static BOOL gdi_init_primary(rdpGdi* gdi, UINT32 stride, UINT32 format,
 	if (format > 0)
 		gdi->dstFormat = format;
 
-	if (stride > 0)
-		gdi->stride = stride;
+	gdi->stride = stride;
 
 	if (!gdi->primary)
 		goto fail_primary;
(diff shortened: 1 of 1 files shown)


akallabeth:
@jcrespo-asg What are you trying to solve?

jcrespo-asg:
when resizing windows using the disp channel:
gdi_resize_ex  and gdi_init_primary gets called with stride 0.
then gdi->stride does not get updated as its not >0 (old value). Because buffers exists then gdi_CreateBitmapEx gets called but it does not update gdi->strade (old value remains).

This causes freerdp_image_copy to break due to bad scanline value.

akallabeth:
@jcrespo-asg Had a in depth look now, so maybe it would be better to calculate the desired stride in ```gdi_resize_ex``` so that the ```gdi->stride``` value is never 0

jcrespo-asg:
gdi_resize_ex can be called from gdi_resize without format (in order to calculate the stride),we could use gdi->dstFormat if 0, but why not leave it as it is?

akallabeth:
@jcrespo-asg because ```gdi.stride``` is used in a lot of places. Feel more comfortable when it has a correct size.

akallabeth:
superseeded

PR #3695, opened 2017-01-11
windows clipboard fix: cliprdr_send_format_list format name

jcrespo-asg:
sometimes (depending on the formatID) wf_cliprdr_server_format_list_response is throwing errors due to the response from the server when sending cliprdr_send_format_list from the client. 

If the clipboard formatId does not have a formatName the whole empty char array is sent. Just validate that the return from GetClipboardFormatNameA, if not send it null.

This fixes the clipboard on windows as the server was not requesting wf_cliprdr_server_format_data_request fue to bad formatlist.


bmiklautz:
@freerdp-bot test.

akallabeth:
@freerdp-bot come on, test please

PR #3696, opened 2017-01-11
Initializing gdi.stride on resize.

akallabeth:
Superseeds #3692


akallabeth:
@jcrespo-asg Simplified your pull, does it what you want?

PR #3703, opened 2017-01-16
[PERFORMANCE] Inlined heavily used functions

akallabeth:


bmiklautz:
@akallabeth the inline will still be quite compiler specific.

PR #3708, opened 2017-01-17
Client channel shutdown fixes

akallabeth:


libfreerdp/core/client.c
```diff
@@ -624,6 +625,7 @@ UINT freerdp_channels_disconnect(rdpChannels* channels, freerdp* instance)
 	/* tell all libraries we are shutting down */
 	for (index = 0; index < channels->clientDataCount; index++)
 	{
+		char* name[9];
```
mfleisz: Shouldn't that be ```char name[9];```?
Also the previous declaration of name should be removed if it's now declared inside the for loop

akallabeth:
@freerdp-bot test please!

PR #3714, opened 2017-01-22
make build reproducible

bmwiedemann:
by allowing build systems to use a constant date instead of the current one
see https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable

This call is designed to work with GNU date. If BSD support is a concern, it can be added.


akallabeth:
@bmwiedemann Thank you for pointing out that flaw in the macro, but if you intended to make our build reproducible there are occurrences of ```__DATE__``` and  ```__TIME__``` in the code.

akallabeth:
@freerdp-bot test please!

akallabeth:
Ok, used your commit and added reproducible build date witn #3715

bmwiedemann:
On 2017-01-23 09:36, akallabeth wrote:
> if you intended to make our build
> reproducible there are occurrences of |__DATE__| and |__TIME__| in the code.

that is fine, because gcc already has a patch to override those.

bmiklautz:
@bmwiedemann @akallabeth I think one of the issues might also be the usage of __FILE__ which might vary depending on the build machine.

bmwiedemann:
@bmiklautz yes, that can be a source of unreproducibility and Debian guys care about that, but at least on the openSUSE side we do not care about that, because we have a standardized build environment that always uses the same build path.

PR #3715, opened 2017-01-23
Reproducible build

akallabeth:


libfreerdp/core/freerdp.c
```diff
@@ -562,8 +562,7 @@ const char* freerdp_get_version_string(void)
 
 const char* freerdp_get_build_date(void)
 {
-	static char build_date[] = __DATE__ " " __TIME__;
```
bmwiedemann: Note that gcc already has patches to override these from SOURCE_DATE_EPOCH.
upstream in gcc7 and backported to gcc6 in Debian and openSUSE, so this part is not really needed

akallabeth:
@bmwiedemann Ok, updated the pull. I've modified your patch in 2 details:
* Use the CMake ```TIMESTAMP``` implementation if ```SOURCE_DATE_EPOCH``` is not used
* Remove the newline created by the ```date``` string

bmiklautz:
@bmwiedemann what's the best way to check if the build is reproducible?

bmwiedemann:
@bmiklautz on openSUSE I have scripts in https://github.com/bmwiedemann/reproducibleopensuse
then I do `osc checkout openSUSE:Factory/freerdp ; cd $_`
and `rb4` (assumed to be in $PATH).

Otherwise just clean-building twice with date set to different days might also give a hint.
My last diff on freerdp is
https://www.zq1.de/~bernhard/linux/reproducibleopensuse/compare.factory-20170110/freerdp-compare.out

cmake/today.cmake
```diff
@@ -3,14 +3,10 @@
...
-    ELSEIF(UNIX)
-        EXECUTE_PROCESS(COMMAND "date" "+%Y-%m-%d" OUTPUT_VARIABLE ${RESULT})
-        string(REGEX REPLACE "(..)/(..)/..(..).*" "\\1/\\2/\\3" ${RESULT} ${${RESULT}})
-    ELSE (WIN32)
-        MESSAGE(SEND_ERROR "date not implemented")
-        SET(${RESULT} 000000)
-    ENDIF (WIN32)
+    if (DEFINED ENV{SOURCE_DATE_EPOCH})
+        EXECUTE_PROCESS(COMMAND "date" "-u" "-d" "@$ENV{SOURCE_DATE_EPOCH}" "+%Y-%m-%d" 
+                        COMMAND "tr" "-d" "'\n'" OUTPUT_VARIABLE ${RESULT})
```
bmwiedemann: `date` is probably only available on UNIX and not WIN32
and https://wiki.debian.org/ReproducibleBuilds/TimestampsProposal#CMake suggests to use OUTPUT_STRIP_TRAILING_WHITESPACE option

cmake/today.cmake
```diff
@@ -3,14 +3,10 @@
...
-    ELSEIF(UNIX)
-        EXECUTE_PROCESS(COMMAND "date" "+%Y-%m-%d" OUTPUT_VARIABLE ${RESULT})
-        string(REGEX REPLACE "(..)/(..)/..(..).*" "\\1/\\2/\\3" ${RESULT} ${${RESULT}})
-    ELSE (WIN32)
-        MESSAGE(SEND_ERROR "date not implemented")
-        SET(${RESULT} 000000)
-    ENDIF (WIN32)
+    if (DEFINED ENV{SOURCE_DATE_EPOCH} AND NOT WIN32)
+        EXECUTE_PROCESS(COMMAND "date" "-u" "-d" "@$ENV{SOURCE_DATE_EPOCH}" "+%Y-%m-%d"
+                        OUTPUT_VARIABLE ${RESULT} OUTPUT_STRIP_TRAILING_WHITESPACE)
```
bmwiedemann: Because other projects are also using cmake's TIMESTAMP function, I proposed
https://gitlab.kitware.com/cmake/cmake/merge_requests/432/diffs
When that is merged, the special case here will not be needed anymore, making everything a lot simpler.

akallabeth:
@bmwiedemann I like that, but for support of older CMake versions this would still be required I guess?

bmwiedemann:
@akallabeth not strictly required. They will just get the current date as before, but people who would update freerdp to get reproducible builds will just as likely update or patch cmake.
Or you merge it as is and clean up in some years. your choice.

Issue #3727, opened 2017-01-27
nightly packages on wheezy and precise broken


bmiklautz:
With PR #3715 the packages for wheezy and precise don't build anymore due the older cmake version.
Fails with
```
CMake Error at cmake/today.cmake:10 (STRING):
  string does not recognize sub-command TIMESTAMP
Call Stack (most recent call first):
  client/X11/CMakeLists.txt:85 (TODAY)
```

PR #3736, opened 2017-01-31
Support old CMake versions without TIMESTAMP #3727

akallabeth:


bmiklautz:
@akallabeth fixed the nightlies again. Thx.

PR #3737, opened 2017-01-31
Flush stdout when asking about certificate

mbaum2000:
Flush stdout when asking for certificate verification.  Without flushing before prompting for input, freerdp will not display the question when trying to automate with QProcess (or possibly something else similar) because in this case, `QProcess::readyReadStandardOutput()` will never be triggered.


akallabeth:
@freerdp-bot test

PR #3741, opened 2017-02-01
codec: Support in-place NSC decoding again

mfleisz:
freerdp_copy_image now also supports in-place flipping

### DIFF
--- a/libfreerdp/codec/color.c
+++ b/libfreerdp/codec/color.c
@@ -505,9 +505,24 @@ BOOL freerdp_image_copy(BYTE* pDstData, DWORD DstFormat,
 					        &srcLine[xSrcOffset], copyDstWidth);
 				}
 			}
-			/* Source and destination are equal... */
-			else
+			/* Source and destination are equal... - maybe a in-place flip is required */
+			else if (vSrcVFlip)
 			{
+				int half = nHeight / 2;
+				int curHeight = nHeight - 1;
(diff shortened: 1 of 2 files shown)


libfreerdp/codec/color.c
```diff
@@ -505,9 +505,23 @@ BOOL freerdp_image_copy(BYTE* pDstData, DWORD DstFormat,
...
+
+				for (y = 0; y < half; ++y)
+				{
+					CopyMemory(tmpBuf, topLine, nDstStep);
+					CopyMemory(topLine, bottomLine, nDstStep);
+					CopyMemory(bottomLine, tmpBuf, nDstStep);
+					topLine += nDstStep;
+					bottomLine -= nDstStep;
+				}
+				_aligned_free(tmpBuf);
```
akallabeth: @mfleisz If you add vSrcVFlip in this branch, it may be a good idea to add this for non overlapping copy too

libfreerdp/codec/nsc.c
```diff
@@ -405,6 +405,9 @@ BOOL nsc_process_message(NSC_CONTEXT* context, UINT16 bpp,
 	context->decode(context);
 	PROFILER_EXIT(context->priv->prof_nsc_decode);
 
+	if (!pDstData)
+		pDstData = context->BitmapData;
+
```
akallabeth: @mfleisz Are you sure this does not screw up if the destination format is not ```PIXEL_FORMAT_BGRA32```?

Issue #3744, opened 2017-02-05
Add command line option to override action script path.


volth:
Path of the action script is [hardcoded](https://github.com/FreeRDP/FreeRDP/blob/master/client/X11/xf_keyboard.c#L62) as ```"/usr/share/freerdp/action.sh"```.
It is inconvenient when there is no ```/usr/share``` (e.g. on [NixOS](http://nixos.org)) or it is not writable by the user.

It would be nice to have a command line option overriding this path.

PS. also, listing the option in help&man would promote a bit this useful but hard to find feature.

bmiklautz:
@volth thanks for the feedback. The feature isn't mentioned as it initially was just added for development and isn't well tested.

volth:
It works very good mostly, except of some short periods of time (I suspect it is when freerdp does autoreconnect of an aborted connection).

akallabeth:
Ok, since your pull is merged I'l close this one.

PR #3746, opened 2017-02-06
Add command line option to override action script path

volth:
It makes ```"/usr/share/freerdp/action.sh"``` overridable with ```/action-script:<filename>``` command line switch
Fix for #3744

### DIFF
--- a/client/X11/xf_event.c
+++ b/client/X11/xf_event.c
@@ -97,7 +97,7 @@ BOOL xf_event_action_script_init(xfContext* xfc)
 		return FALSE;
 
 	ArrayList_Object(xfc->xevents)->fnObjectFree = free;
-	sprintf_s(command, sizeof(command), "%s xevent", xfc->actionScript);
+	sprintf_s(command, sizeof(command), "%s xevent", xfc->context.settings->ActionScript);
 	actionScript = popen(command, "r");
 
 	if (!actionScript)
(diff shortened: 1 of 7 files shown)


akallabeth:
@freerdp-bot test
Anyone knowledgeable around that knows why this was hard coded in the first place?

akallabeth:
@freerdp-bot test

volth:
Ok.
But I would keep ```/usr/share/freerdp/action.sh``` as default.
Changing the default is a breaking change and I am not the only user of the feature out there, other people 
already placed their scripts at ```/usr/share/freerdp/action.sh``` (e.g. https://github.com/FreeRDP/FreeRDP/pull/3457#issuecomment-254855937)

akallabeth:
@volth The default is not set this way, the struct just contains a human readable representation of it (if required)

akallabeth:
@freerdp-bot test

akallabeth:
@bmiklautz You agree with these changes?

bmiklautz:
@volth I share your considerations that people already use this feature. However as mentioned it was more for development but if a command line parameter is available it's "visible" and I could think it's more used as well (once the format is documented ;). Therefore I'd rather have a sensible path in the .config directory than  /usr/share/freerdp/action.sh which isn't usually writable by users.

bmiklautz:
thanks.

PR #3755, opened 2017-02-10
ifdef's for Cygwin compilation

pentagra:
Several ifdef improvements for successful compilation under Cygwin


akallabeth:
@freerdp-bot test

PR #3756, opened 2017-02-10
.RDP Password Attribute

bigpjo:
Allow password to be stored in .RDP file, parsed and settings
updated, this will allow for dynamic .RDP files to be created with
complete login credentials, using this method the username, server and
password will no longer be visible within process lists.

Also fixed bug with username being 
set to null by command line processor if the command line parameter not specified but the value set from the .RDP files.

Usage: create .RDP file, add line 

password:s:XXXXXX where XXXXXX is the password

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -2513,20 +2513,22 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 	}
 	while ((arg = CommandLineFindNextArgumentA(arg)) != NULL);
 
-	free(settings->Username);
-
-	if (!settings->Domain && user)
+	if (!settings->Username && user)
 	{
-		BOOL ret;
-		free(settings->Domain);
-		ret = freerdp_parse_username(user, &settings->Username, &settings->Domain);
-		free(user);
+		free(settings->Username);
+		if (!settings->Domain && user)
+		{
+			BOOL ret;
+			free(settings->Domain);
 
-		if (!ret)
-			return COMMAND_LINE_ERROR;
+			ret = freerdp_parse_username(user, &settings->Username, &settings->Domain);
+			free(user);
+			if (!ret)
+				return COMMAND_LINE_ERROR;
+		}
+		else
+			settings->Username = user;
 	}
-	else
-		settings->Username = user;
 
 	free(settings->GatewayUsername);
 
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test please

client/common/cmdline.c
```diff
@@ -2513,20 +2513,22 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 	}
 	while ((arg = CommandLineFindNextArgumentA(arg)) != NULL);
 
-	free(settings->Username);
-
-	if (!settings->Domain && user)
+	if (!settings->Username && user)
```
akallabeth: This check is no good idea. If I want to override the username provided via RDP file this check will prevent it. A simple ```if (user)``` should do the trick.

akallabeth:
@bigpjo Sorry for the change of mind, just found this issue when retesting your pull.

bmiklautz:
As a remark: Although this is a possible solution we haven't added "password" support in .rdp files yet because you then possibly have files lying around with plain passwords. So if you don't wan't to show the user/server/domain/password in the processes list you probably also don't want to have the password  in files. I think currently the best solution is to create a wrapper script and use /from-stdin.

bigpjo:
I appreciate that storing plain text passwords in the file are not ideal and this is not a suggestion/change I would recommend without a specific goal in mind, there are a number of other open source projects dependent on starting FreeRDP from a command line, it is these projects that would benefit from this change as it would be possible to create .RDP files with a password, start the process and delete the file after. Not ideal but a solution.

If it would be of benefit, I could also add another command line parameter to delete the .RDP file once the process is started to ensure that there are not 100's of .RDP files with passwords left around but this would still be user choice.

In .RDP files there is a hashed password field password 51, the issue with this is that you need a windows API call for the RDP client to decrypt this and send it to establish the report session, this will not work I believe in a linux/mac/unix env making this type of change non portable.

Any comments welcomed, in the grand schema of things, this pull request is not a major contribution to an excellent project

hardening:
My 2 euros cents: previous version of mstsc on mac used to support the password in clear.

akallabeth:
@bigpjo Agree with the necessity of storing the password in some cases.
As for other projects, as @bmiklautz already pointed out providing the password via stdin is probably a better solution (but also far from perfect)
The hashed password option is, as you correctly asserted, not an option without introducing some kind of password manager dependency

That aside, if you fix the check for allowing command line to override RDP username you got my vote ;)

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test, come on

PR #3763, opened 2017-02-14
fix a bug in libwinpr module

seasar:
when enqueue a new entry, we should rearrange wrapped entries if we invoke realloc().
the judgment condition: queue->tail < old_capacity   is wrong, it should be: queue->tail <= queue->head

### DIFF
--- a/winpr/libwinpr/utils/collections/MessageQueue.c
+++ b/winpr/libwinpr/utils/collections/MessageQueue.c
@@ -82,8 +82,9 @@ void MessageQueue_Dispatch(wMessageQueue* queue, wMessage* message)
 		queue->capacity = new_capacity;
 		queue->array = (wMessage*) realloc(queue->array, sizeof(wMessage) * queue->capacity);
 		ZeroMemory(&(queue->array[old_capacity]), old_capacity * sizeof(wMessage));
-
-		if (queue->tail < old_capacity)
+		
+		//rearrange wrapped entries
+		if (queue->tail <= queue->head)
 		{
 			CopyMemory(&(queue->array[old_capacity]), queue->array, queue->tail * sizeof(wMessage));
 			queue->tail += old_capacity;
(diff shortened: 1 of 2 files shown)


winpr/libwinpr/utils/collections/MessageQueue.c
```diff
@@ -82,8 +82,9 @@ void MessageQueue_Dispatch(wMessageQueue* queue, wMessage* message)
 		queue->capacity = new_capacity;
 		queue->array = (wMessage*) realloc(queue->array, sizeof(wMessage) * queue->capacity);
 		ZeroMemory(&(queue->array[old_capacity]), old_capacity * sizeof(wMessage));
-
-		if (queue->tail < old_capacity)
+		
+		//rearrange wrapped entries
```
akallabeth: Please avoid C++ comments.

winpr/libwinpr/utils/collections/Queue.c
```diff
@@ -148,7 +148,8 @@ void Queue_Enqueue(wQueue* queue, void* obj)
 		queue->array = (void**) realloc(queue->array, sizeof(void*) * queue->capacity);
 		ZeroMemory(&(queue->array[old_capacity]), old_capacity * sizeof(void*));
 
-		if (queue->tail < old_capacity)
+		//rearrange wrapped entries
```
akallabeth: Please avoid C++ comments.

akallabeth:
@freerdp-bot test
@seasar Could you please fix the comments and squash your changes into a single commit?

bmiklautz:
@seasar very nice finding! While reviewing I saw that the memory zeroed with is also incorrect. We use old_capacity instead of (new_capacity - old_capacity). Maybe you can fix this in your PR as well if you are in mood (otherwise I'll do a separate pull).

PR #3766, opened 2017-02-14
core: Close channels on context free instead of disconnect

mfleisz:
The current implementation sent the TERMINATE event to channels on
disconnect. The result was that all channels of the freerdp instance were
destroyed. If the same instance was now used to connect again none of the
channels would work because they were already destroyed.


akallabeth:
@freerdp-bot test

PR #3768, opened 2017-02-15
libwinpr-utils: fix 3 logic errors

seasar:
a new pr instead of #3763

(1)when enqueue a new entry, we should rearrange wrapped entries if we invoke realloc().
the judgment condition: queue->tail < old_capacity is wrong, it should be: queue->tail <= queue->head .
(2)collect size of mem in ZeroMemory().
(3)after updating queue->tail,  &(queue->array[queue->tail]) is null. the next statement(message->time = (UINT64) GetTickCount()) will cause a segv.


akallabeth:
@freerdp-bot test

akallabeth:
@seasar Sorry to nag, but your commit does not apply cleanly on master. Could you rebase this pull on current master (no merges) and force push your branch? Thank you.

akallabeth:
@seasar Ok, rebased myself.
Just a not though:
* Please don't open new pull requests for every change, just push to your working branch, this will update the pull. (Force push when rebasing)
* Better work on current master, merges are not that easy to handle if in a pull request.

winpr/libwinpr/utils/collections/MessageQueue.c
```diff
@@ -99,7 +101,6 @@ BOOL MessageQueue_Dispatch(wMessageQueue* queue, wMessage* message)
 	queue->tail = (queue->tail + 1) % queue->capacity;
 	queue->size++;
 
-	message = &(queue->array[queue->tail]);
```
bmiklautz: If this is removed the time is set in the message parameter but not in the message added to the queue.

PR #3773, opened 2017-02-16
Gfx and sample fixes

hardening:
Some nice bits that helped a lot when working with a buggy egfx server.


hardening:
@freerdp-bot I don't see why the connect test would have failed, please re test.

PR #3785, opened 2017-02-20
tsmf dead lock fixed

uplusplus:
As ffmpeg decoder does not have stream->decoder->BufferLevel,So
stream->currentBufferLevel is always equal 1


akallabeth:
@freerdp-bot test

akallabeth:
@uplusplus Does this also work with gstreamer?

mfleisz:
@freerdp-bot test

mfleisz:
@freerdp-bot test

PR #3791, opened 2017-02-20
Kerberos (Rebased #3417)

akallabeth:
Rebased #3417 to current master


HenryJacques:
If I may, I think it will also be useful to add the following three cases in `transport.c` in the `switch (ret)`:

KRB5_KDC_UNREACH: Cannot contact any KDC for requested realm 
KRB5KDC_ERR_CLIENT_REVOKED: Clients credentials have been revoked 
KRB5KDC_ERR_POLICY: KDC policy rejects request

(of course you can add as many Kerberos errors as you want but with these three, you cover 90% of the errors. See https://tools.ietf.org/html/rfc1510#page-84)

akallabeth:
@HenryJacques Ok, will have a look tomorrow.

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,187 @@
...
+UINT32 transport_krb5_check_account(char* username, char* domain, char* passwd)
+{
+	krb5_error_code ret;
+	krb5_context context;
+	krb5_principal principal = NULL;
+	char address[256];
+	krb5_ccache ccache;
+	krb5_init_creds_context ctx = NULL;
+	memset(address, 0, sizeof(address));
+	strcpy(address, username);
```
mfleisz: _snprintf(... , "%s@%s", username, domain) would be safer (checks for buffer size) and we can also get rid of the memset

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,187 @@
...
+
+	if (!g_KRB5Module)
+	{
+		g_KRB5Module = LoadLibrary("libkrb5.so");
+	}
+
+	if (!g_KRB5Module)
+	{
+		WLog_ERR(TAG, "Failed to load krb5 module\n");
+	}
```
hardening: I guess we should exit in this case, not really sure if GetProcAddress(NULL) is safe...

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,187 @@
...
+	g_KRB5.pKrb5_parse_name_flags 		= (void*) GetProcAddress(g_KRB5Module, "krb5_parse_name_flags");
+	g_KRB5.pKrb5_cc_cache_match 		= (void*) GetProcAddress(g_KRB5Module, "krb5_cc_cache_match");
+	g_KRB5.pKrb5_cc_default 		= (void*) GetProcAddress(g_KRB5Module, "krb5_cc_default");
+	g_KRB5.pKrb5_init_creds_init 		= (void*) GetProcAddress(g_KRB5Module, "krb5_init_creds_init");
+	g_KRB5.pKrb5_init_creds_set_password 	= (void*) GetProcAddress(g_KRB5Module,
+	        "krb5_init_creds_set_password");
+	g_KRB5.pKrb5_init_creds_get 		= (void*) GetProcAddress(g_KRB5Module, "krb5_init_creds_get");
+	g_KRB5.pKrb5_init_creds_free 		= (void*) GetProcAddress(g_KRB5Module, "krb5_init_creds_free");
+	g_KRB5.pKrb5_free_context 		= (void*) GetProcAddress(g_KRB5Module, "krb5_free_context");
+
```
hardening: All these should be checked against NULL or we may run into troubles some lines later...

akallabeth:
@mfleisz @hardening Ok, reworked the kerberos library stuff, directly linking it in now. No more ```LoadLibrary``` et al, just plain CMake detection. Also added the error handlers @HenryJacques suggested.

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,130 @@
...
+	else if ((ret = krb5_cc_cache_match(context, principal, &ccache)) != 0)
+	{
+		if ((ret = krb5_cc_default(context, &ccache)) != 0)
+			WLog_ERR(TAG, "krb5 failed to resolve credentials cache with error %d", (int)ret);
+	}
+
+	if (ret != KRB5KDC_ERR_NONE)
+		goto out;
+	/* Create a context for acquiring initial credentials */
+	else if ((ret = krb5_init_creds_init(context, principal, NULL, NULL, 0, NULL, &ctx)) != 0)
```
hardening: the else is useless

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,130 @@
...
+	{
+		WLog_WARN(TAG, "krb5_init_creds_init returned error %d", (int)ret);
+		goto out;
+	}
+	/* Set a password for acquiring initial credentials */
+	else if ((ret = krb5_init_creds_set_password(context, ctx, passwd)) != 0)
```
hardening: the same here

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,130 @@
...
+			transport = (rdpTransport*) SSL_get_app_data(ssl);
 
-		if (!freerdp_get_last_error(transport->context))
-			freerdp_set_last_error(transport->context, FREERDP_ERROR_AUTHENTICATION_FAILED);
+			if (!freerdp_get_last_error(transport->context))
+			{
+				freerdp_set_last_error(transport->context, FREERDP_ERROR_AUTHENTICATION_FAILED);
+			}
+		}
+		else if (ret == 592)
```
hardening: Just noticed that, but it would be neat to have a define or at least a comment for that mysterious 592 value..

HenryJacques: Looking at https://linux.die.net/man/3/ssl_set_info_callback the `ret` value can be interpreted with `ssl_alert_type_string`. This function return F (as Fatal)  

`case SSL3_AL_FATAL:
        return "F";`

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,130 @@
...
+	/* Acquire credentials using an initial credential context */
+	ret = krb5_init_creds_get(context, ctx);
+out:
+
+	switch (ret)
+	{
+		case KRB5KDC_ERR_NONE:
+			break;
+
+		case KRB5_KDC_UNREACH:
```
HenryJacques: KRB5_KDC_UNREACH shouldn't be in this case. In my opinion, it should return a specific message in WLog_WARN indicating it cannot contact the Kerberos server (you can keep the same error FREERDP_ERROR_AUTHENTICATION_FAILED)

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,130 @@
...
+			ret = FREERDP_ERROR_AUTHENTICATION_FAILED;
+			break;
+
+		case KRB5KDC_ERR_KEY_EXP:
+			WLog_WARN(TAG, "krb5_init_creds_get: Password has expired");
+			ret = FREERDP_ERROR_CONNECT_PASSWORD_EXPIRED;
+			break;
+
+		case KRB5KDC_ERR_CLIENT_REVOKED:
+			ret = FREERDP_ERROR_CONNECT_PASSWORD_CERTAINLY_EXPIRED;
```
HenryJacques: I don't agree. FREERDP_ERROR_CONNECT_PASSWORD_CERTAINLY_EXPIRED is only used for non-domain workstations. In this case the Kerberos server was reached. We should define a new error code, for example FREERDP_ERROR_CONNECT_CLIENT_REVOKED

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,130 @@
...
+			break;
+
+		case KRB5KDC_ERR_POLICY:
+			ret = FREERDP_ERROR_INSUFFICIENT_PRIVILEGES;
```
HenryJacques: In my opinion, we could add a WLog_WARN message indicating the reason of this error. During my tests I had this error only when the account was forced to use smartcard logon (and not login/password)

HenryJacques:
@akallabeth I did put 3 single comments in the PR - `in transport.c` (they don't appear in the conversation as they are not reviews I think)

akallabeth:
@mfleisz @hardening @HenryJacques Ok, fixed all your comments and modified the logger to use the faster preloaded one. Don't be fooled by the failing compile, we need to install ```libkrb5-dev``` on the machines before it will go through.

libfreerdp/core/transport.c
```diff
@@ -59,17 +59,146 @@
...
+		default:
+			WLog_Print(transport->log, WLOG_WARN, "krb5_init_creds_get");
+			ret = FREERDP_ERROR_CONNECT_TRANSPORT_FAILED;
+			break;
+	}
+
+	if (ctx)
+		krb5_init_creds_free(context, ctx);
+
+	krb5_free_context(context);
```
mfleisz: if krb5_init_context fails we potentially free an invalid context here

akallabeth:
@freerdp-bot test

hardening:
Would be nice to drop the useless else, anyway it's not a show stopper.

PR #3795, opened 2017-02-21
Limit the number of client announced monitors

hardening:
The specs says that only 16 are allowed, so let's make that limitation a reality.


hardening:
Well the magic number is in the spec...

akallabeth:
@hardening That is what I meant, a link to that would help. Just a suggestion though

MartinHaimberger:
@akallabeth I have not found it directly in the spec for now. But the value is from the group policys. There u can set the max allowed monitors between 1 and 16.

https://blogs.technet.microsoft.com/enterprisemobility/2009/07/01/using-multiple-monitors-in-remote-desktop-session/

hardening:
It's in 2.2.1.3.6 => https://msdn.microsoft.com/en-us/library/dd305336.aspx

PR #3797, opened 2017-02-21
Misc patches around capabilities

hardening:
This PR : 

* removes some warning;
* tries to reformat some parts of the code that have become ugly with the usage of --max-code-length in astyle;
* it adds the very useful ClientCapabilities callback;


akallabeth:
@hardening Just a remark to the formatting issue: Removing the unnecessary comments (What information is provided by them?) would also prevent the line break by astyle

PR #3798, opened 2017-02-22
primitives: Fix compilation with VS 2010

mfleisz:
Fixes compiling issues with VS 2010


mfleisz:
@akallabeth That's not that easy - according to http://stackoverflow.com/questions/15499251/how-to-initialize-m128i-array-statically-in-gcc GCC doesn't like that version for instance so we won't get around defining more than one version ...

PR #3800, opened 2017-02-22
Cleanup channel structs in channels_close to allow instance reuse

mfleisz:


mfleisz:
@nfedera PR changed as we talked about

PR #3754, opened 2017-02-10
NEON and SSSE3 decoder optimisations

akallabeth:
* Optimized NEON primitives for decoding RFX, AVC420, AVC444 and progressive
* Optimized SSSE3 RFX ```yCbCrToRGB_16s8u_P3AC4R```
* Added profiler to tests

PR #3801, opened 2017-02-22
resolve compiler warning

chipitsine:
channels/cliprdr/client/cliprdr_main.c:121:13: warning: ‘cliprdr_print_general_capability_flags’ defined but not used [-Wunused-function]

 static void cliprdr_print_general_capability_flags(UINT32 flags)

             ^

### DIFF
--- a/channels/cliprdr/client/cliprdr_main.c
+++ b/channels/cliprdr/client/cliprdr_main.c
@@ -118,6 +118,7 @@ static UINT cliprdr_packet_send(cliprdrPlugin* cliprdr, wStream* s)
 	return status;
 }
 
+#ifdef WITH_DEBUG_CLIPRDR
 static void cliprdr_print_general_capability_flags(UINT32 flags)
 {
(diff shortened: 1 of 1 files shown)


ilammy:
Why comment out the tests?.. Maybe they should be just reenabled in TestPrimitivesColors. Apparently, they got disabled in #3754.

chipitsine:
tests are called within "ifdef" block

https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/primitives/test/TestPrimitivesColors.c#L284-L301

but, since they are defined outside "ifdef" block, they look as "defined but not used" for compiler

akallabeth:
@chipitsine +1 for the debug define, but with the tests the ifdef should not be done. They are actually a marker of stuff to be done.

chipitsine:
I suspected something like that

chipitsine:
I removed the last commit. is it ok now?

akallabeth:
@freerdp-bot test

PR #3806, opened 2017-02-23
Don't ask astyle to treat long lines as it breaks code readability

hardening:
People are smart enough to not use too long lines, and astyle is too strict
when breaking lines, so let's just remove that reformating.

That will prevent code like this:
	Stream_ReadUINT16(inStream, targetValue); /* targetValue: UINT16 comment...*/
to be changed to:
	Stream_ReadUINT16(inStream,
		targetValue); /* targetValue: UINT16 comment...*/

### DIFF
--- a/scripts/format_code.sh
+++ b/scripts/format_code.sh
@@ -31,7 +31,7 @@ if [ $# -le 0 ]; then
 fi
 
 $ASTYLE --lineend=linux --mode=c --indent=tab=4 --pad-header --pad-oper --style=allman --min-conditional-indent=0 \
-							   --indent-switches --indent-cases --indent-preprocessor -k1 --max-code-length=100 \
+							   --indent-switches --indent-cases --indent-preprocessor -k1  \
 							   --indent-col1-comments --delete-empty-lines --break-closing-brackets \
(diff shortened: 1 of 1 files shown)


hardening:
My point is more that today blindly applying astyle with the current style has broken code readability in many places. BTW the places where it has been broken were written in an optimal way: with code and its specification reference as a comment on the right. 

From my point of view removing max-code is just pragmatic, it breaks things so let's remove it. I don't think we have so much places where we have active code exceeding the 100 limit, and if so we should treat them by hand.

hardening:
To summarize, I'm not for long lines of active code, but today astyle doesn't look like the solution to handle that.

bmiklautz:
+1 for @akallabeth solution. We should do *one* reformat and keep the style (finally - format changes are always a pain to review). However I also don't like the  way astyle breaks the line(s) for this part I do agree with you @hardening either the tool does is nicely or not at all.
@akallabeth as for alternatives we can have a look to clang-format again. It should be available for most distributions. We can also have a look to https://github.com/coala/coala which can wrap a lot of different tools to do things like formatting/enforcing.

akallabeth:
#4350 proposes a solution with a better tool (which hopefully is acceptable by all)

PR #3813, opened 2017-02-24
disable test

chipitsine:
according to @awakecoding "that test should be disabled",
actually that test fails on linux due to "#ifndef",
but it should not run on linux

### DIFF
--- a/winpr/libwinpr/io/test/TestIoDevice.c
+++ b/winpr/libwinpr/io/test/TestIoDevice.c
@@ -6,7 +6,7 @@
 
 int TestIoDevice(int argc, char* argv[])
 {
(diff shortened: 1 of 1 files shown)


chipitsine:
do you define _WIN32 ?

akallabeth:
No, why? It just returns 0 in that case, the define is ```#ifndef _WIN32```

chipitsine:
if _WIN32 is not defined (as it is supposed to be under Linux), the code is being run. and that code does not look like it is supposed to run under linux

akallabeth:
Closing as not required, the comment from @bmiklautz can be done in a new pull.

Issue #3815, opened 2017-02-24
FreeRDP crashes with hostnames > 31 characters


lemonmojo:
I just had a long debugging session with a customer and found out that FreeRDP currently crashes when a hostname with 32 or more characters is used on the machine it is being run on.

This hard limit is defined in [sysinfo.h](https://github.com/FreeRDP/FreeRDP/blob/master/winpr/include/winpr/sysinfo.h#L292), used in [settings.c](https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/settings.c#L255) and the actual problem occurs in [sysinfo.c](https://github.com/FreeRDP/FreeRDP/blob/master/winpr/libwinpr/sysinfo/sysinfo.c#L410).

Not sure about Windows or Linux but on the Mac it's totally possible to have hostnames with 32 characters or more (way more actually). Just do this on a terminal and watch FreeRDP cry like a baby: `sudo scutil --set HostName "abcdefghijklmnopqrstuvwxyz012345"`

akallabeth:
@lemonmojo The 31 character limit is a limitation of the RDP protocol. The error handling for the default is broken though, need a sensible fallback for that.

lemonmojo:
@akallabeth I already suspected that. Any ideas for a proper fix? I'd also take a quick fix as I need to get this out to customers as soon as possible.

akallabeth:
@lemonmojo For a quick fix just truncate the hostname in ```GetComputerNameA``` instead of returning an error

hardening:
Yeah I've hit that bug too some weeks ago with a Ubuntu 14.04 host, who's name was generated from the chip and co and so was way longer than 31 characters. Quite a long debugging to figure that freerdp_settings_new() was failing, and to find that GetComputerNameA was at fault.

lemonmojo:
@akallabeth Just implemented the quick fix, works wonders!
@hardening I didn't even have debugging tools on the customer's machine so that was quite a challenge... ;)

akallabeth:
@lemonmojo Does current master fix the issue as well?

lemonmojo:
@akallabeth 32+ characters hostname works fine with master.

PR #3816, opened 2017-02-24
Fixed GetComputerNameExA calls. #3815

akallabeth:


libfreerdp/core/settings.c
```diff
@@ -252,13 +252,24 @@ void settings_load_hkey_local_machine(rdpSettings* settings)
...
+	if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize))
+		return FALSE;
+
+	computerName = calloc(nSize, sizeof(CHAR));
+
+	if (!computerName)
+		return FALSE;
 
 	if (!GetComputerNameExA(ComputerNameNetBIOS, computerName, &nSize))
 		return FALSE;
```
nfedera: computerName memory leaking here

winpr/libwinpr/sspi/NTLM/ntlm.c
```diff
@@ -108,16 +123,28 @@ int ntlm_SetContextServicePrincipalNameA(NTLM_CONTEXT* context, char* ServicePri
...
+			return -1;
+
+		computerName = calloc(nSize, sizeof(CHAR));
+
+		if (!computerName)
 			return -1;
 
-		name = _strdup(computerName);
+		if (!GetComputerNameExA(ComputerNameNetBIOS, computerName, &nSize))
+			return -1;
```
nfedera: allocated computerName memory is leaking here

libfreerdp/core/settings.c
```diff
@@ -252,13 +252,24 @@ void settings_load_hkey_local_machine(rdpSettings* settings)
...
 
-	settings->ComputerName = _strdup(computerName);
+	if (nSize > MAX_COMPUTERNAME_LENGTH)
+		computerName[MAX_COMPUTERNAME_LENGTH] = '\0';
```
nfedera: According to https://msdn.microsoft.com/en-us/library/windows/desktop/ms724301(v=vs.85).aspx:

> The length of the name may be greater than MAX_COMPUTERNAME_LENGTH characters because DNS allows longer names. 

So why the limitation ?

winpr/libwinpr/sspi/NTLM/ntlm.c
```diff
@@ -47,20 +47,34 @@ int ntlm_SetContextWorkstation(NTLM_CONTEXT* context, char* Workstation)
...
+		if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize))
+			return -1;
+
+		computerName = calloc(nSize, sizeof(CHAR));
+
+		if (!computerName)
+			return -1;
+
 		if (!GetComputerNameExA(ComputerNameNetBIOS, computerName, &nSize))
 			return -1;
```
nfedera: computerName memory leaking here

winpr/libwinpr/sspi/NTLM/ntlm_av_pairs.c
```diff
@@ -167,22 +168,34 @@ NTLM_AV_PAIR* ntlm_av_pair_add_copy(NTLM_AV_PAIR* pAvPairList, NTLM_AV_PAIR* pAv
...
 		return -1;
 
-	name = _strdup(computerName);
+	computerName = calloc(nSize, sizeof(CHAR));
+
+	if (!computerName)
+		return -1;
+
+	if (!GetComputerNameExA(ComputerNameNetBIOS, computerName, &nSize))
+		return -1;
```
nfedera: computerName memory leaking here

libfreerdp/core/settings.c
akallabeth: @nfedera It is a NetBIOS name and I did not want to break anything depending on that limit. (It was enforced before this change, don't know if it breaks anything now)

hardening: Yep I think it's the reason...

nfedera: Right! Sorry for the noise ;)

libfreerdp/core/settings.c
```diff
@@ -252,13 +252,27 @@ void settings_load_hkey_local_machine(rdpSettings* settings)
 
 BOOL settings_get_computer_name(rdpSettings* settings)
 {
-	DWORD nSize = MAX_COMPUTERNAME_LENGTH + 1;
-	CHAR computerName[MAX_COMPUTERNAME_LENGTH + 1];
+	DWORD nSize = 0;
+	CHAR* computerName;
+
+	if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize))
```
nfedera: Here and below: The actual fix must include the GetLastError() check and eventually a size check.
Something like this:
```
if (ComputerNameNetBIOS, NULL, &nSize) || GetLastError() != ERROR_MORE_DATA ||  nSize < 2)
    return FALSE;
```

akallabeth: @nfedera Well, the function will return ```ERROR_BUFFER_OVERFLOW``` as ```GetComputerNameA``` is used internally. Don't know if this is a bug or desired?

nfedera: @akallabeth good catch! This is definitely a bug in FreeRDP's GetComputerNameExA implementation. If it uses GetComputerNameA internally (which correctly returns ERROR_BUFFER_OVERFLOW) it must call `SetLastError(ERROR_MORE_DATA)` in that case.

nfedera:
I've just noticed that GetComputerNameEx does not correctly update the lpnSize parameter.
(That bug was already there).

According to https://msdn.microsoft.com/en-us/library/windows/desktop/ms724301(v=vs.85).aspx:

> lpnSize: On output, receives the number of TCHARs copied to the destination buffer, not including the terminating null character.

Would be great if you could also fix this.
You can find my ctest here: https://github.com/nfedera/FreeRDP/commit/055c4a4276faa17d69b5283bd4b21dd0bf3894ac which I'll create a PR for once your PR is merged.

Issue #3814, opened 2017-02-24
Extra error message when closing the xfreerdp window


hardening:
Since the kerberos error printing, I'm seeing that error when closing the xfreerdp window:

```
[15:44:37:546] [20383:20384] [WARN][com.freerdp.core.transport] - Unhandled SSL error (where=16392, ret=256 [warning, close notify])
```

akallabeth:
@hardening looks like the switch needs another case for this one, found a explanation here:
https://security.stackexchange.com/questions/133575/ssl3-alert-readwarningclose-notify

akallabeth:
@hardening Are these messages gone now? (cold no longer reproduce locally)

hardening:
No messages now, closing thank you @akallabeth

PR #3823, opened 2017-02-27
Fixed SSL error checks in transport_ssl_cb

akallabeth:
Fix error define use. The callback is called from
ssl3_dispatch_alert or dtls1_dispatch_alert where the alert define
is left shifted by 8. Additionally ignore close notifcation. (#3814)

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -154,13 +154,13 @@ out:
 
 static void transport_ssl_cb(SSL* ssl, int where, int ret)
 {
-	if (where | SSL_CB_ALERT)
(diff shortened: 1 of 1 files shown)


ilammy:
One thing that has caught my attention is handling of `where` value on line 157. Shouldn't that be `where & SSL_CB_ALERT` instead of `where | SSL_CB_ALERT`?

hardening:
@akallabeth I think @ilammy is definitely right it should be ```where & SSL_CB_ALERT```, as ```where | SSL_CB_ALERT``` is always true...

akallabeth:
@hardening @ilammy Oops, missed that one. Should be fixed now.

PR #3827, opened 2017-03-01
add  WAVE_FORMAT_MPEG_HEAAC header parser

uplusplus:
read WAVE_FORMAT_MPEG_HEAAC header to get real sample rate.

### DIFF
--- a/channels/tsmf/client/tsmf_codec.c
+++ b/channels/tsmf/client/tsmf_codec.c
@@ -255,6 +255,119 @@ static const TSMFMediaTypeMap tsmf_sub_type_map[] =
 
 };
 
+
+#define WaveFormatTag(_arg_) _arg_
+enum name
+ {
+	WAVE_FORMAT_UNKNOWN                 = WaveFormatTag(0x0000), // Microsoft Corporation
+	WAVE_FORMAT_PCM                     = WaveFormatTag(0x0001), // Microsoft PCM format
+	WAVE_FORMAT_MS_ADPCM                = WaveFormatTag(0x0002), // Microsoft ADPCM
+	WAVE_FORMAT_IEEE_FLOAT              = WaveFormatTag(0x0003), // Micrososft 32 bit float format
+	WAVE_FORMAT_VSELP                   = WaveFormatTag(0x0004), // Compaq Computer Corporation
+	WAVE_FORMAT_IBM_CVSD                = WaveFormatTag(0x0005), // IBM Corporation
+	WAVE_FORMAT_ALAW                    = WaveFormatTag(0x0006), // Microsoft Corporation
+	WAVE_FORMAT_MULAW                   = WaveFormatTag(0x0007), // Microsoft Corporation
+	WAVE_FORMAT_OKI_ADPCM               = WaveFormatTag(0x0010), // OKI
+	WAVE_FORMAT_IMA_ADPCM               = WaveFormatTag(0x0011), // Intel Corporation
+	WAVE_FORMAT_MEDIASPACE_ADPCM        = WaveFormatTag(0x0012), // Videologic
+	WAVE_FORMAT_SIERRA_ADPCM            = WaveFormatTag(0x0013), // Sierra Semiconductor Corp
+	WAVE_FORMAT_G723_ADPCM              = WaveFormatTag(0x0014), // Antex Electronics Corporation
+	WAVE_FORMAT_DIGISTD                 = WaveFormatTag(0x0015), // DSP Solutions, Inc.
+	WAVE_FORMAT_DIGIFIX                 = WaveFormatTag(0x0016), // DSP Solutions, Inc.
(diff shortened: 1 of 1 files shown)


channels/tsmf/client/tsmf_codec.c
```diff
@@ -255,6 +255,119 @@ static const TSMFMediaTypeMap tsmf_sub_type_map[] =
 
 };
 
+
+#define WaveFormatTag(_arg_) _arg_
```
hardening: I don't see any advantage to this, can you elaborate on why you're using that define and use it for enum values ?

channels/tsmf/client/tsmf_codec.c
```diff
@@ -411,6 +524,114 @@ static UINT32 tsmf_codec_parse_VIDEOINFOHEADER(TS_AM_MEDIA_TYPE* mediatype, wStr
 	return 48;
 }
 
+
+const static int avpriv_mpeg4audio_sample_rates[16] = {
+    96000, 88200, 64000, 48000, 44100, 32000,
+    24000, 22050, 16000, 12000, 11025, 8000, 7350
+};
+
+static int get_sample_rates(int idx){
```
hardening: code style mismatch

channels/tsmf/client/tsmf_codec.c
```diff
@@ -411,6 +524,114 @@ static UINT32 tsmf_codec_parse_VIDEOINFOHEADER(TS_AM_MEDIA_TYPE* mediatype, wStr
...
+		*/
+
+		if(audioObjectType!=2){
+			WLog_ERR(TAG, "wrong audioObjectType!");
+		}else{
+			int bits_to_decode = (mediatype->ExtraDataSize-2) * 8;
+			pbAudioSpecificConfig = Stream_Pointer(s) + 2;
+
+			DEBUG_TSMF("SBR: bits_to_decode() %d", bits_to_decode);
+			if(bits_to_decode >= 16){ // left bits is sbr?!
```
hardening: For this big statement with stairs, may I suggest to do:
```
do 
{
    if (bits_to_decode < 16);
         break;
    .....
    if(syncExtensionType != 0x2b7)
         break;
    .....
    if(extensionAudioObjectType != 0x05)
          break;

   .... (and do on)....

} while(0);
```
The code will look nicer with less stairs of if().

Plus you should definitely look at the FreeRDP code style (curly bracket are supposed to be on a single line, binary operators are supposed to have spaces around them, things like that).

channels/tsmf/client/tsmf_codec.c
```diff
@@ -411,6 +524,114 @@ static UINT32 tsmf_codec_parse_VIDEOINFOHEADER(TS_AM_MEDIA_TYPE* mediatype, wStr
...
+		  WORD         wStructType;
+		  WORD         wReserved1;
+		  DWORD        dwReserved2;
+		} HEAACWAVEINFO, *PHEAACWAVEINFO;
+	*/
+	UINT16 wPayloadType;
+	UINT16 wAudioProfileLevelIndication;
+	UINT16 wStructType;
+
+	Stream_Read_UINT16(s, wPayloadType);
```
hardening: You should not Stream_Read() without checking there's enough to read (Stream_GetRemainingLength), so I guess your function should return a BOOL to say if the decoding was ok, and the actual return value should be passed as a pointer on a UINT32. The remark on Stream_Read applies for the rest of the function.

channels/tsmf/client/tsmf_codec.c
```diff
@@ -454,6 +676,10 @@ static BOOL tsmf_read_format_type(TS_AM_MEDIA_TYPE* mediatype, wStream* s, UINT3
 			Stream_Read_UINT16(s, mediatype->BlockAlign);
 			Stream_Read_UINT16(s, mediatype->BitsPerSample);
 			Stream_Read_UINT16(s, mediatype->ExtraDataSize);
+			
+			if(wFormatTag == WAVE_FORMAT_MPEG_HEAAC){
+				tsmf_codec_parse_HEAACWAVEINFOHEADER(mediatype, s);
```
hardening: the return value is ignored, so is it needed ?

hardening:
Thanks for the contribution, do the fixes and I will approve this PR.

akallabeth:
@freerdp-bot test

channels/tsmf/client/tsmf_codec.c
uplusplus: can be ignore or print some warning inside the function

akallabeth:
@uplusplus Could you address the changes so that we can merge?

channels/tsmf/client/tsmf_codec.c
bmiklautz: @hardening I would think you could easily grep for it instead of just searching for the numeric value.

hardening:
@uplusplus any news ?

PR #3829, opened 2017-03-01
String representation of logon_error_info

akallabeth:
* Adds a function converting the logon error info error numbers to a string that can be used to identify the error
* Updated function visibility in ```info.c```


akallabeth:
@freerdp-bot test

PR #3839, opened 2017-03-03
Fixed function pointer casts.

akallabeth:
* Modifies function pointer casts so that C11 compilers do no longer warn when using ```GetProcAddress```
* Modifies addin and channel tables to properly cast content


hardening:
Well it probably makes the compiler happy which human can understand what ```*((PVIRTUALCHANNELENTRY*)&subsystems[j].entry)``` means instead of ```subsystems[j].entry``` ?
Do we have to make the compiler happy, or perhaps we could just disable the warning ?

nfedera:
@hardening I also find it totally unreadable. However it seems to be the only workaround to get rid of ` warning: ISO C forbids conversion of object pointer to function pointer type [-Wpedantic]`.

Quoting http://man7.org/linux/man-pages/man3/dlopen.3.html
```
double (*cosine)(double);
...
cosine = (double (*)(double)) dlsym(handle, "cos");
/*
According to the ISO C standard, casting between function
pointers and 'void *', as done above, produces undefined results.
POSIX.1-2003 and POSIX.1-2008 accepted this state of affairs and
proposed the following workaround:

*(void **) (&cosine) = dlsym(handle, "cos");

This (clumsy) cast conforms with the ISO C standard and will
avoid any compiler warnings.

The 2013 Technical Corrigendum to POSIX.1-2008 (a.k.a.
POSIX.1-2013) improved matters by requiring that conforming
implementations support casting 'void *' to a function pointer.
Nevertheless, some compilers (e.g., gcc with the '-pedantic'
option) may complain about the cast used in this program.
*/
```

I also prefer to ignore or disable this kind of warning.

akallabeth:
@hardening @nfedera Any news on this one? The proposed cast is as far as I can tell the only way to disable this warning. If you prefer to leave that out, I'd create a new pull for the addin changes.

nfedera:
@akallabeth I prefer to ignore these function pointer conversion warnings for the safe of readability

hardening:
@akallabeth same opinion as @nfedera

PR #3845, opened 2017-03-07
Afreerdp db upgrade

akallabeth:
Fixes broken database upgrades.
The old code did only check the main table for upgrades, no doing that for all.


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #3862, opened 2017-03-13
Fix Stack_Peek

uglym8:
Fix simple off-by-one error.


hardening:
@freerdp-bot would you look at this with a test ?

PR #3875, opened 2017-03-21
Client random

hardening:
We need the client_random to enable client reconnection when the client uses RDP security.

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -583,7 +583,14 @@ BOOL rdp_server_establish_keys(rdpRdp* rdp, wStream* s)
 
 	mod = rdp->settings->RdpServerRsaKey->Modulus;
 	priv_exp = rdp->settings->RdpServerRsaKey->PrivateExponent;
-	crypto_rsa_private_decrypt(crypt_client_random, rand_len - 8, key_len, mod, priv_exp, client_random);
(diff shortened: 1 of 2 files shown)


hardening:
@freerdp-bot come on if you stop to test after 4s on windows for sure it will fail. Do it again !

mfleisz:
@freerdp-bot test

mfleisz:
@freerdp-bot test

libfreerdp/core/connection.c
```diff
@@ -583,7 +583,14 @@ BOOL rdp_server_establish_keys(rdpRdp* rdp, wStream* s)
...
 	mod = rdp->settings->RdpServerRsaKey->Modulus;
 	priv_exp = rdp->settings->RdpServerRsaKey->PrivateExponent;
-	crypto_rsa_private_decrypt(crypt_client_random, rand_len - 8, key_len, mod, priv_exp, client_random);
+	if (crypto_rsa_private_decrypt(crypt_client_random, rand_len - 8, key_len, mod, priv_exp, client_random) <= 0)
+	{
+		free(client_random);
+		goto end;
+	}
+
+	rdp->settings->ClientRandom = client_random;
```
hardening: Very good question, checking that.

hardening: And the answer is yes, as server-side it's the only place where it is set. And this function is never called again even during a re-negotiation.

hardening:
Feel free to merge this one if it is ok ;)

Issue #3412, opened 2016-06-17
FreeRDP crashes on system in FIPS mode


ondrejholy:
I enabled FIPS mode on Red Hat Enterprise Linux [1] and configured Microsoft Window Server to require FIPS-compliant encryption [2]. Consequently, xfreerdp crashes with the following error, when I try to connect to the server regardless of used security protocol:
`md4_dgst.c(74): OpenSSL internal error, assertion failed: Digest MD4 forbidden in FIPS mode!`

I see several occurrences of MD4, MD5, RC4 algorithms in the FreeRDP codes. Those algorithms are not FIPS-compliant. However, I suppose that those algorithms should not be used if FIPS-compliant encryption is used...

[1] https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Security_Guide/chap-Federal_Standards_and_Regulations.html
[2] https://technet.microsoft.com/en-us/library/cc770833.aspx

hardening:
I guess you have done your tests with master ?

ondrejholy:
Yes, I did the tests with master and freerdp-1.0, both versions fail in the same way...

bmiklautz:
@ondrejholy i might be mistaken but I think when you enable FIPS compliant security this applies only to the transport itself. MD4, MD5, .. might still be used/required in different other parts like for generating ntlm credentials.

ondrejholy:
As far as I know, FIPS regulations applies on all used crypto algorithms. FreeRDP works correctly with server in FIPS mode, however doesn't work itself on a client system with enabled FIPS mode. NLA security can't be used with FIPS mode probably, because NTLM is not FIPS compliant, because it requires RC4 (see section 5.1 of [MS-NLMP]), and Kerberos is not implemented in FreeRDP. However TLS/RDP security should work with FIPS mode. Triple DES should be used for bulk encryption, and SHA-1 for _any_ hashing operations (e.g. MAC generation). See sections 5.3.5/5.3.6 from [MS-RDPBCGR] for RDP security and section 5.4.1 for Enhanced RDP security. MD5/RC4 is still used in some cases, but I am convinced that it should be replaced by SHA-1 if FIPS encryption is used. Just a note that native Remote Desktop Connection client works correctly on system with enabled FIPS mode.

kfiresmith:
Hello,
[...]

PR #3877, opened 2017-03-21
Fix FreeRDP crashes on system in FIPS mode

ondrejholy:
Crypto algorithms, which aren't FIPS-compliant (i.e. MD5, RC4), can be still used for some functionality, which isn't security-relevant, on systems with enabled FIPS mode. Let's allow weak ciphers in specific cases (i.e. licensing).

Thanks to this patches, FreeRDP works properly with RDP, or TLS security on systems with enabled FIPS mode.

See https://github.com/FreeRDP/FreeRDP/issues/3412 for more details.

### DIFF
--- a/libfreerdp/common/assistance.c
+++ b/libfreerdp/common/assistance.c
@@ -92,10 +92,10 @@ int freerdp_assistance_crypt_derive_key_sha1(BYTE* hash, int hashLength, BYTE* k
 	if (!buffer)
 		goto fail;
 
-	if (!winpr_Digest(WINPR_MD_SHA1, pad1, 64, buffer, hashLength))
+	if (!winpr_Digest(WINPR_MD_SHA1, pad1, 64, buffer, hashLength, FALSE))
 		goto fail;
 
-	if (!winpr_Digest(WINPR_MD_SHA1, pad2, 64, &buffer[hashLength], hashLength))
+	if (!winpr_Digest(WINPR_MD_SHA1, pad2, 64, &buffer[hashLength], hashLength, FALSE))
 		goto fail;
 
 	CopyMemory(key, buffer, keyLength);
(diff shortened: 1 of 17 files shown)


bmiklautz:
@freerdp-bot test.

hardening:
Hi @ondrejholy,
I like the patch but not the way it's done.
I really think the checks for FIPS should not be done at the winPr level but at a higher level. So not adding an argument to ```winpr_HMAC_Init```, but by checking the callers:

* looking at the patch all the calls ```if (!winpr_Digest(WINPR_MD_MD5, ...., FALSE)``` just means ```if (fipsMode) return FALSE;```
* I guess that in many places we can even return false at a higher level (don't even need to start looking at licence things if RC4 is forbidden by FIPS). I gave a look with eclipse, looking at call hierarchies and it would do a much smaller patch (and would let winpr do crypto without caring of FIPS)

Just curious, but I guess lots of things aren't working in RDP with FIPS enabled (can you even do a negotiation ?)

ondrejholy:
@hardening Thanks for the review. I'm not also really happy with the non_fips_allow arguments, but the EVP_CIPH_FLAG_NON_FIPS_ALLOW and EVP_MD_CTX_FLAG_NON_FIPS_ALLOW flags need to be set somehow. I can change the codes to "if (fipsMode) return FALSE" logic, but not sure how to set the flags when needed. I don't really think that it is a good idea to set those flags by default. 

Yes, a lot of things doesn't work under the FIPS mode and can't work apparently. However, some basic functionality may work under the FIPS mode and that's the reason why I proposed those patches. FreeRDP works on systems with enabled FIPS mode with those patches, but /sec:tls, or /sec:rdp has to be specified explicitly. So, yes, the protocol security negotiation needs to be also updated...

Hmm, it seems that PullRequestTester doesn't like my patches. Maybe some special handling is needed if OpenSSL is not built with FIPS support. I will try to solve this somehow...

bjcollins:
@ondrejholy @hardening check out my new pull request #3904, I worked on it some time ago and it takes a different approach then this PR. I think we will have to keep some logic in the winpr to set flags for fips overrides. I have inline documentation for where we override ciphers and since I use an explicit option - I am able to do things like override NLA automatically. One big thing I did different was just to make unique override functions with fips in their name so it is easier to keep track of where we are overriding specific ciphers for fips.

ondrejholy:
Let's close this request in favor of https://github.com/FreeRDP/FreeRDP/pull/3904 .

PR #3882, opened 2017-03-27
some performance fixes

nfedera:
- draw only the updated region in the gdi and x11 surface bits implementation
- don't repeatedly call IsProcessorFeaturePresentEx in rfx rlgr decoder
- fix ugly and unaligned profiler print layout and remove an unnecessary value

### DIFF
--- a/libfreerdp/gdi/gdi.c
+++ b/libfreerdp/gdi/gdi.c
@@ -972,8 +972,13 @@ BOOL gdi_surface_frame_marker(rdpContext* context,
 static BOOL gdi_surface_bits(rdpContext* context,
                              const SURFACE_BITS_COMMAND* cmd)
 {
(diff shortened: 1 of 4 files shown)


libfreerdp/gdi/gdi.c
```diff
@@ -1019,22 +1033,38 @@ static BOOL gdi_surface_bits(rdpContext* context,
...
+		{
+			WLog_ERR(TAG, "Failed to update invalid region");
+			goto out;
+		}
 	}
 
+	result = TRUE;
+out:
+	region16_uninit(&region);
 	return TRUE;
```
akallabeth: @nfedera Should that not be ```return result;```?

nfedera: Right!

PR #3891, opened 2017-03-30
improve rdpegfx drawing for software gdi

nfedera:
process the region's rectangles individually instead of just copying the extents


bmiklautz:
@nfedera nice finding!

PR #3897, opened 2017-04-04
OSX: Updated RPATH settings for each lib.

akallabeth:
Set the RPATH on OSX


bmiklautz:
@akallabeth can you give more specifics for this PR?

akallabeth:
@bmiklautz On OS X all libraries are addressed by full path or ```@rpath/<name>``` style. Relative addressing is not well supported, so this pull configures CMake to create library references and names with the ```@rpath``` prefix

akallabeth:
@mfleisz refactored the pull according to @bmiklautz suggestions.

PR #3864, opened 2017-03-13
Rewrite disk redirection using WinPR

DavBfr:
Goal: Handle properly Unicode characters like '表' (U+8868) or '𤭢' (U+24B62 UTF-16 surrogate pairs) with disk redirection.
- Add some missing functions to WinPR
- Rewrite the disk redirection channel using the WinPR *W functions

PR #3902, opened 2017-04-06
Fixed a memory leak and bad free.

akallabeth:
Fixed issues introduced with #3864

### DIFF
--- a/winpr/libwinpr/file/generic.c
+++ b/winpr/libwinpr/file/generic.c
@@ -475,8 +475,10 @@ BOOL WINAPI GetFileAttributesExA(LPCSTR lpFileName, GET_FILEEX_INFO_LEVELS fInfo
 	WIN32_FIND_DATAA findFileData;
 	HANDLE hFind;
 
-	if ((hFind = FindFirstFileA(lpFileName, &findFileData)) != INVALID_HANDLE_VALUE)
-		FindClose(hFind);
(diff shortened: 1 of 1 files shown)


akallabeth:
@hardening Looks like we missed a leak :/ This one should fix it (clang scanbuild confirms). Just be careful if you rerun, due to the ```INVALID_HANDLE_VALUE``` check it now reports a false positive (or better a rare case where there could actually be a memory leak if the address is ```-1```)

akallabeth:
@mfleisz @bmiklautz @nfedera Anyone time to have a look?

winpr/libwinpr/file/generic.c
```diff
@@ -475,7 +475,7 @@ BOOL WINAPI GetFileAttributesExA(LPCSTR lpFileName, GET_FILEEX_INFO_LEVELS fInfo
 	WIN32_FIND_DATAA findFileData;
 	HANDLE hFind;
 
-	if ((hFind = FindFirstFileA(lpFileName, &findFileData) != INVALID_HANDLE_VALUE))
+	if ((hFind = FindFirstFileA(lpFileName, &findFileData)) != INVALID_HANDLE_VALUE)
```
nfedera: nice finding! however, return false on error is required too

akallabeth: @nfedera done.

akallabeth:
@freerdp-bot test please

Issue #3532, opened 2016-10-10
Implement Clipboard filesharing


lanoxx:
Opening this issue here as a prerequirement of this issue FreeRDP/Remmina/issues/323

I would like to ask that freedrp implements support for file sharing via the clip board. This is a feature that has long been possible in the windows implementation.

I want to take any file on Windows and copy it to the clip board (e.g. with CTRL+C) and then I want to paste it on Linux (e.g. into nautilus). This should also work the other way around.

I am not sure whats required to support this but it would be pretty awesome if this worked. This feature it currently the only reason why I still keep a Windows 10 VM around, because I often need to copy a file and using FTP or some other tool is often to complicated, when I just want to quickly copy a single file. Besides I have many remote windows servers where no FTP or other file sharing services are installed.

akallabeth:
duplicate

Issue #3302, opened 2016-04-24
Files copying issue


mikhaXD:
If copy file from PC to a RDP = its works fine. But when from RDP to PC - file is copied but size is 0bytes. No data inside. 
Im using latest windows nightly build

bmiklautz:
Please have a look to https://github.com/FreeRDP/FreeRDP/wiki/BugReporting and add more information.

akallabeth:
Closing due to lack of feedback.

ilammy:
That's actually a feature request: #3532

akallabeth:
@ilammy Ah, ok :) Sorry, did update a lot of issues and misread that one. But since there is already a more precisely formulated issue let's keep this one closed.

Issue #2424, opened 2015-02-23
clipboard and files


azhel12:
Clipboard works, but I can not copy any file via it. Copy/paste text works fine. This options works in standart windows rdp-client(mstsc). So I think that it must work in FreeRDP too.

bmiklautz:
@azhel12 the clipboard channel doesn't support file copies yet.
If you want to copy from your remote machine to the local machine you need to use the drive channel.

azhel12:
I thought the same, but I tried to do it in Windows default client and it works here! Also I have written small program on C# with using MSTSCLib. My program (it is about 10 code lines) supports copy files via clipboard.

meanbunny:
This would be an awesome feature!

ilammy:
Well, I have an implementation of file clipping in X11 client somewhere. IIRC, it worked pretty well for me. (Except for some issues with large files.) Because reasons it did not make it to public, but I can try find it and rebase onto current master.

azhel12:
I think you are speak russian then write russian. If I was wrong, write me
about it, pls. Доброго времени суток. Пишу по поводу Вашего ответа "Well, I
have an implementation of file clipping in X11 client somewhere. IIRC, it
worked pretty well for me. (Except for some issues with large files.)
Because reasons it did not make it to public, but I can try find it and
rebase onto current master."
Посмотрел Ваш аккаунт на github и сделал вывод, что Вы носитель русского
языка, поэтому пишу по-русски) Делаю небольшой проект, где бы классно было
использовать FreeRDP, но нужна передача файлов через буфер обмена. Я так
понял, что у Вас это реализовано. Где я могу это найти? В вашем форке и
ветке master?

2015-04-01 15:09 GMT+03:00 ilammy <notifications@github.com>:

bmiklautz:
@azhel12 please not that the official language for FreeRDPs issues is English. Thank you.

ilammy:
@azhel12, yes, I know Russian, but others here maybe do not, so I am writing in English. Drop me an email if you are more comfortable with Russian.

The file clipping implementation is not anywhere on Github, it was somewhere on my hard drive about  a year ago. I will look for it and keep you (and others) posted.

bmiklautz:
@ilammy thanks!

azhel12:
Sorry, I wanted to write only to ilammy. But made a mistake( Ok, my email: azhel12reg@gmail.com

ilammy:
[...]

PR #3905, opened 2017-04-09
Local-to-remote file clipping for xfreerdp

ilammy:
This adds support for file transfer via clipboard redirection to WinPR and teaches xfreerdp how to use it, making some progress with issues #2424, #3302, #3532, FreeRDP/Remmina#323, FreeRDP/Remmina#730.

Only local-to-remote transfers are implemented at the moment. That is, the users can select and copy/cut files in their local file managers (e.g., Nautilus) and then paste them inside the RDP session window. Implementing the remote-to-local transfers is not so trivial, so this will be coming later.

The patchset has been tested by me on Debian 8 and Ubuntu 12.04 with Nautilus and Thunar file managers on the local side and Windows 7 and Windows 10 VMs on the remote side. You can find a rough check list [here](https://github.com/ilammy/FreeRDP/pull/4#issuecomment-292360152).

The overall design has been outlined [on the mailing list](https://sourceforge.net/p/freerdp/mailman/message/35694035/), I won't repeat it here. The detailed comments on the implementation can be found in the commit messages. I tried making the individual commits small enough to ease the code review, but that's 1.6k lines of new code however you look at it. The whole patch sequence can be split into the following parts:

* Preparatory refactoring of the common code and some utilities:
  - winpr: define file attribute flags
  - channels/cliprdr: CLIPRDR_FILELIST utilities
  - wClipboard: improve error handling
* Parsing and processing the text/uri-list format:
  - wClipboard: local file subsystem boilerplate
  - wClipboard/posix: basic file list handling
  - wClipboard/posix: conversion to FILEDESCRIPTORs
  - wClipboard/posix: parse text/uri-list format
  - wClipboard/posix: decode percent-encoding
  - wClipboard/posix: add local files to file list
  - wClipboard/posix: add directories to file list
* Implementing the interface for retrieving file contents:
  - wClipboard/posix: basic delegate interface
  - wClipboard/posix: implement file size retrieval
  - wClipboard/posix: implement file range retrieval
  - wClipboard: track sequence numbers of file lists
* Adding support of file tranfer to xfreedp:
  - x11/cliprdr: impove error handling
  - x11/cliprdr: handle text/uri-list format
  - x11/cliprdr: provide file content to the server
  - x11/cliprdr: negotiate file streaming support
* Fixing issues that have been found during testing:
  - channels/cliprdr: fix CLIPRDR_FILECONTENTS_REQUEST
  - channels/cliprdr: fix CLIPRDR_FILECONTENTS_RESPONSE
  - channels/cliprdr: handle cliprdr 2 GB size limit
  - wClipboard: disallow Windows reserved names

I would love to have feedback from @akallabeth (as a person familiar with clipboard redirection) and from @hardening (as a person good at finding stupid bugs early, C is hard).

As for the future, I plan to implement a GVFS subsystem for local files as an example of asynchronous fetching of file content. This should fix support for remote-mapped filesystems in Nautilus and Thunar in GNOME-based environments. After that I am going to take on implementation of remote-to-local file transfers via FUSE. Or instead of that...


akallabeth:
@freerdp-bot test

akallabeth:
@ilammy Looks promising (and a very welcome addition), but I will need some time to check this one. In the mean time, could you fix the issues preventing a build with vs2010 (that is very old dialect of C)?

ilammy:
On 11 April 2017 at 09:11, akallabeth <notifications@github.com> wrote:

> @ilammy <https://github.com/ilammy> Looks promising (and a very welcome
> addition), but I will need some time to check this one. In the mean time,
> could you fix the issues preventing a build with vs2010 (that is very old
> dialect of C)?
>
Sure. From what I see in the build log it seems that MSVC does not like
when function prototypes include argument names.

I did not check the build on Windows as I though I haven't touched anything
that would be enabled there. But in fact I did. I'll check the Windows
build and a fix.

ilammy:
Well, I'm kinda ashamed to admit it, but I'm still having trouble setting up the development environment on Windows which is able to build FreeRDP (fighting with OpenSSL at the moment). However, I have given another look at the [build logs][1] and it seems I have found the issue:

```diff
diff --git a/include/freerdp/channels/cliprdr.h b/include/freerdp/channels/cliprdr.h
index ba7fb1b..45fae0c 100644
--- a/include/freerdp/channels/cliprdr.h
+++ b/include/freerdp/channels/cliprdr.h
@@ -23,6 +23,8 @@
 #include <freerdp/api.h>
 #include <freerdp/types.h>
 
+#include <winpr/shell.h>
+
 #define CLIPRDR_SVC_CHANNEL_NAME	"cliprdr"
 
diff --git a/include/freerdp/channels/cliprdr.h b/include/freerdp/channels/cliprdr.h
index 45fae0c..2f1d027 100644
--- a/include/freerdp/channels/cliprdr.h
+++ b/include/freerdp/channels/cliprdr.h
@@ -86,6 +86,19 @@ struct _CLIPRDR_MFPICT
 };
 typedef struct _CLIPRDR_MFPICT CLIPRDR_MFPICT;
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+FREERDP_API UINT cliprdr_parse_file_list(const BYTE* format_data, UINT32 format_data_length,
+		FILEDESCRIPTOR** file_descriptor_array, UINT32* file_descriptor_count);
+FREERDP_API UINT cliprdr_serialize_file_list(const FILEDESCRIPTOR* file_descriptor_array,
+		UINT32 file_descriptor_count, BYTE** format_data, UINT32* format_data_length);
+
+#ifdef __cplusplus
+}
+#endif
+
 /* Clipboard Messages */
```
the compiler seems to treat FILEDESCRIPTOR as an identifier name when it follows `const` in the declaration of `cliprdr_serialize_file_list` function. This is because `<winpr/shell.h>` defines FILEDESCRIPTOR only for non-Windows platforms. I guess it should include appropriate Windows headers when WinPR is built on Windows.

I have added the patch to the pull request which does this, but I did not even build-test it. @freerdp-bot, can you please do this for me? :crossed_fingers:

[1]: https://ci.freerdp.com/job/windows_msbuild_vs2010/arch=win32,label=vs2010/143/console

akallabeth:
@freerdp-bot test

ilammy:
Ugh, another blind fix based on scrutilizing the build log. No idea why I have renamed the field in 	6ad05d5 (it used to have the correct name).

bmiklautz:
@freerdp-bot test

client/X11/xf_cliprdr.c
```diff
@@ -1391,8 +1455,156 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext*
 	return CHANNEL_RC_OK;
 }
 
+UINT xf_cliprdr_server_file_size_request(xfClipboard* clipboard,
```
akallabeth: Could be static

client/X11/xf_cliprdr.c
```diff
@@ -1391,8 +1455,156 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext*
...
+	if (fileContentsRequest->cbRequested != sizeof(UINT64))
+	{
+		WLog_WARN(TAG, "unexpected FILECONTENTS_SIZE request: %"PRIu32" bytes",
+			fileContentsRequest->cbRequested);
+	}
+
+	return clipboard->delegate->ClientRequestFileSize(clipboard->delegate, &request);
+}
+
+UINT xf_cliprdr_server_file_range_request(xfClipboard* clipboard,
```
akallabeth: could be static

client/X11/xf_cliprdr.c
```diff
@@ -1391,8 +1455,156 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext*
...
+	request.streamId = fileContentsRequest->streamId;
+	request.listIndex = fileContentsRequest->listIndex;
+	request.nPositionLow = fileContentsRequest->nPositionLow;
+	request.nPositionHigh = fileContentsRequest->nPositionHigh;
+	request.cbRequested = fileContentsRequest->cbRequested;
+
+	return clipboard->delegate->ClientRequestFileRange(clipboard->delegate, &request);
+}
+
+UINT xf_cliprdr_send_file_contents_failure(CliprdrClientContext* context,
```
akallabeth: could be static

client/X11/xf_cliprdr.c
```diff
@@ -1391,8 +1455,156 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext*
...
+	ZeroMemory(&response, sizeof(response));
+
+	response.msgFlags = CB_RESPONSE_FAIL;
+	response.streamId = fileContentsRequest->streamId;
+	response.dwFlags = fileContentsRequest->dwFlags;
+
+	return context->ClientFileContentsResponse(context, &response);
+}
+
+UINT xf_cliprdr_server_file_contents_request(CliprdrClientContext* context,
```
akallabeth: could be static

client/X11/xf_cliprdr.c
```diff
@@ -1391,8 +1455,156 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext*
...
+	{
+		WLog_ERR(TAG, "failed to handle CLIPRDR_FILECONTENTS_REQUEST: 0x%08X", error);
+
+		return xf_cliprdr_send_file_contents_failure(context, fileContentsRequest);
+	}
+
+	return CHANNEL_RC_OK;
+}
+
+UINT xf_cliprdr_clipboard_file_size_success(wClipboardDelegate* delegate,
```
akallabeth: could be static

client/X11/xf_cliprdr.c
```diff
@@ -1391,8 +1455,156 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext*
...
+	response.msgFlags = CB_RESPONSE_OK;
+	response.streamId = request->streamId;
+	response.dwFlags = FILECONTENTS_SIZE;
+	response.cbRequested = sizeof(UINT64);
+	response.requestedData = (BYTE*) &fileSize;
+
+	return clipboard->context->ClientFileContentsResponse(clipboard->context, &response);
+}
+
+UINT xf_cliprdr_clipboard_file_size_failure(wClipboardDelegate* delegate,
```
akallabeth: could be static

client/X11/xf_cliprdr.c
```diff
@@ -1391,8 +1455,156 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext*
...
+	ZeroMemory(&response, sizeof(response));
+
+	response.msgFlags = CB_RESPONSE_FAIL;
+	response.streamId = request->streamId;
+	response.dwFlags = FILECONTENTS_SIZE;
+
+	return clipboard->context->ClientFileContentsResponse(clipboard->context, &response);
+}
+
+UINT xf_cliprdr_clipboard_file_range_success(wClipboardDelegate* delegate,
```
akallabeth: could be static

client/X11/xf_cliprdr.c
```diff
@@ -1391,8 +1455,156 @@ static UINT xf_cliprdr_server_format_data_response(CliprdrClientContext*
...
+	response.msgFlags = CB_RESPONSE_OK;
+	response.streamId = request->streamId;
+	response.dwFlags = FILECONTENTS_RANGE;
+	response.cbRequested = size;
+	response.requestedData = (BYTE*) data;
+
+	return clipboard->context->ClientFileContentsResponse(clipboard->context, &response);
+}
+
+UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate,
```
akallabeth: could be static

cmake/ConfigOptions.cmake
```diff
@@ -121,6 +121,7 @@ option(WITH_DEBUG_TRANSPORT "Print transport debug messages." ${DEFAULT_DEBUG_OP
 option(WITH_DEBUG_THREADS "Print thread debug messages, enables handle dump" ${DEFAULT_DEBUG_OPTION})
 option(WITH_DEBUG_MUTEX "Print mutex debug messages" ${DEFAULT_DEBUG_OPTION})
 option(WITH_DEBUG_TIMEZONE "Print timezone debug messages." ${DEFAULT_DEBUG_OPTION})
+option(WITH_DEBUG_WCLIPBOARD "Print WinPR clipboard debug messages" ${DEFAULT_DEBUG_OPTION})
```
akallabeth: Would remove this, WLog supports filtering.

winpr/libwinpr/clipboard/posix.c
```diff
@@ -0,0 +1,906 @@
...
+	return result;
+}
+
+static BOOL process_uri(const char* uri, size_t uri_len, wArrayList* files)
+{
+	BOOL result = FALSE;
+	char* name = NULL;
+
+#ifdef WITH_DEBUG_WCLIPBOARD
+	WLog_DBG(TAG, "processing URI: %.*s", uri_len, uri);
```
akallabeth: As mentioned above, would use ```WLog_TRACE``` without define and just use WLog filters.

winpr/libwinpr/file/file.c
```diff
@@ -814,6 +814,86 @@ BOOL GetDiskFreeSpaceW(LPCWSTR lpwRootPathName, LPDWORD lpSectorsPerCluster,
...
+ *
+ * Some file names are not valid on Windows. See "Naming Files, Paths, and Namespaces":
+ * https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx
+ */
+BOOL ValidFileNameComponent(LPCWSTR lpFileName)
+{
+	LPCWSTR c = NULL;
+
+	/* CON */
+	if ((lpFileName[0] != L'\0' && (lpFileName[0] == L'C' || lpFileName[0] == L'c')) &&
```
akallabeth: Missing NULL argument check

ilammy:
Hi @akallabeth. Sorry, I've been busy with some stuff as well. Finally I've pushed the improvements to the branch. All your suggestions make very much sense. Thank you.

As for adding `static` to callbacks, there are some other files in X11 client which could have some of their functions marked as static as well, but I've confined the changes only the clipboard-related stuff.

akallabeth:
@freerdp-bot test

akallabeth:
@hardening could you have a look? Would really like this feature ;)

channels/cliprdr/client/cliprdr_format.c
```diff
@@ -336,3 +336,181 @@ UINT cliprdr_process_format_data_response(cliprdrPlugin* cliprdr, wStream* s, UI
...
+		WLog_ERR(TAG, "invalid packed file list");
+
+		result = ERROR_INCORRECT_SIZE;
+		goto out;
+	}
+
+	Stream_Read_UINT32(s, count); /* cItems (4 bytes) */
+
+	if (Stream_GetRemainingLength(s) < count * CLIPRDR_FILEDESCRIPTOR_SIZE)
+	{
```
hardening: to prevent integer overflow it should be 
```
if (Stream_GetRemainingLength(s) /  CLIPRDR_FILEDESCRIPTOR_SIZE < count)
```

channels/cliprdr/client/cliprdr_format.c
```diff
@@ -336,3 +336,181 @@ UINT cliprdr_process_format_data_response(cliprdrPlugin* cliprdr, wStream* s, UI
...
+		Stream_Read_UINT32(s, file->dwFlags); /* flags (4 bytes) */
+		Stream_Seek(s, 32); /* reserved1 (32 bytes) */
+		Stream_Read_UINT32(s, file->dwFileAttributes); /* fileAttributes (4 bytes) */
+		Stream_Seek(s, 16); /* reserved2 (16 bytes) */
+		Stream_Read_UINT64(s, lastWriteTime); /* lastWriteTime (8 bytes) */
+		file->ftLastWriteTime = uint64_to_filetime(lastWriteTime);
+		Stream_Read_UINT32(s, file->nFileSizeHigh); /* fileSizeHigh (4 bytes) */
+		Stream_Read_UINT32(s, file->nFileSizeLow); /* fileSizeLow (4 bytes) */
+		for (c = 0; c < 260; c++) /* cFileName (520 bytes) */
+			Stream_Read_UINT16(s, file->cFileName[c]);
```
hardening: Isn't there a proper way to read WCHARs ?

channels/cliprdr/client/cliprdr_format.c
```diff
@@ -336,3 +336,181 @@ UINT cliprdr_process_format_data_response(cliprdrPlugin* cliprdr, wStream* s, UI
...
+		 * There is a known issue with Windows server getting stuck in
+		 * an infinite loop when downloading files that are larger than
+		 * 2 gigabytes. Do not allow clients to send such file lists.
+		 *
+		 * https://support.microsoft.com/en-us/help/2258090
+		 */
+		if ((file->nFileSizeHigh > 0) || (file->nFileSizeLow >= CLIPRDR_MAX_FILE_SIZE))
+		{
+			WLog_ERR(TAG, "cliprdr does not support files over 2 GB");
+			result = ERROR_FILE_TOO_LARGE;
```
hardening: Perhaps the caller should perform such checks ?

channels/cliprdr/client/cliprdr_format.c
```diff
@@ -336,3 +336,181 @@ UINT cliprdr_process_format_data_response(cliprdrPlugin* cliprdr, wStream* s, UI
...
+
+		Stream_Write_UINT32(s, file->dwFlags); /* flags (4 bytes) */
+		Stream_Zero(s, 32); /* reserved1 (32 bytes) */
+		Stream_Write_UINT32(s, file->dwFileAttributes); /* fileAttributes (4 bytes) */
+		Stream_Zero(s, 16); /* reserved2 (16 bytes) */
+		lastWriteTime = filetime_to_uint64(file->ftLastWriteTime);
+		Stream_Write_UINT64(s, lastWriteTime); /* lastWriteTime (8 bytes) */
+		Stream_Write_UINT32(s, file->nFileSizeHigh); /* fileSizeHigh (4 bytes) */
+		Stream_Write_UINT32(s, file->nFileSizeLow); /* fileSizeLow (4 bytes) */
+		for (c = 0; c < 260; c++) /* cFileName (520 bytes) */
```
hardening: Same remark as above for writing WCHARs. Perhaps it's the way to do it but I feel like we have some functions to do that. If not just forget that comment.

channels/cliprdr/client/cliprdr_format.c
mfleisz: I agree - should use Stream_Read here

channels/cliprdr/client/cliprdr_format.c
mfleisz: Should use Stream_Write here

ilammy: I initially used Stream_{Read,Write} there but then I realized that they do not handle endianness properly. Sadly, the spec vaguely says 'Unicode' without specifying the exact encoding. While we can (?) assume this will always be little-endian UTF-16, I believe we cannot assume that all FreeRDP clients will be running on a little-endian hardware, thus simply memcpy()ing that array would not work.

channels/cliprdr/client/cliprdr_format.c
ilammy: Maybe. I put it there because I guessed that all clients will have to perform such checks if they want to deal with that issue, so why not put this failsafe in the common function. However, it still requires cooperation from the client. If the client does not fill in the file size immediately in FILEDESCRIPTOR list then the server can still break later when it explicitly requests the file size and proceeds with content retrieval (as there are not checks there).

I placed that check before any file content transfer in order to abort the whole clipboard paste if there are files like that. If we fail it later then remote Windows will just stop, show no error messages, and paste random files which it has managed to download so far. I believe this is wrong user experience as it suggests successful transfer for the user. But it is mainly caused by the fact that xfreerdp does not show any error messages. I guess there would be no problem with aborting the transfer halfway if we (more prominently) tell the user what's broke.

Also, I did not put these checks into wClipboard stuff because I felt that a 'generic clipboard' should not deal with peculiarities of cliprdr protocol and its server-side implementation. However, it may be a good place for them, as the final frontier where we can prevent a disaster on the server.

Another option will be to remove this failsafe entirely. I do not remember which systems actually fail catastrophically if they are faced with 2GB+ files. Maybe it's only the oldest ones.

I have no problem with moving this check into the client just before calling cliprdr_serialize_file_list(), just don't see any practical merit in it at the moment.

@akallabeth, what's your opinon on this?

channels/cliprdr/client/cliprdr_format.c
ilammy: Stream_GetRemainingLength() returns size_t, maybe we could simply cast count to size_t so that multiplication result is size_t instead of UINT32?

ilammy: Or we'd better do that when printing an error message, with divison in the condition as you suggest.

channels/cliprdr/client/cliprdr_format.c
hardening: So perhaps a ConvertToUnicode with destination Stream_Pointer(s) would  do the trick ?

ilammy: I'm afraid ConvertToUnicode() cannot convert between different endian variations (at least because it ignores the code page argument). And there is no BOM which can tell whether it is little or big endian, so I don't get what we could do with ConvertToUnicode() even if it could convert endianness for us.

channels/cliprdr/client/cliprdr_format.c
hardening: No tricks with size_t, an explicit check is always better.

ilammy: Yeah, I've done as above: explicit division in the condition where the check is performed. I've still used size_t in the error message as we print byte count from Stream_GetRemainingLength() and it seems natural to compare it to the expected number of bytes.

channels/cliprdr/client/cliprdr_format.c
hardening: I guess that
```
ConvertFromUnicode(CP_UTF8, 0, Stream_Pointer(s), 520, &file->cFileName, 520, NULL, NULL);
```
should do the trick

channels/cliprdr/client/cliprdr_format.c
hardening: I'm quite surprised of your comment because we widely use Convert[To|From]Unicode in misc places where there is unicode data (usually paths).

channels/cliprdr/client/cliprdr_format.c
akallabeth: Well, it is a server issue which can easily fixed there. I don't like restricting the client too much because some server implementation is broken.

channels/cliprdr/client/cliprdr_format.c
akallabeth: Just checkec, there is actually no real good way to read/write unicode in little endian format from/to a stream. Will create a pull for that one though.

akallabeth: Ok, added pull #3991 to address this.

akallabeth:
@frerdp-bot test

akallabeth:
@freerdp-bot test

channels/cliprdr/client/cliprdr_format.c
ilammy: @hardening, that functions are doing quite well for UTF-8—UTF-16 conversions, but here we want to convert between UTF-16 bytes and WCHARs, keeping the WCHARs because that's the type of the cFileName field of the FILEDESCRIPTOR struct.

The file name is Windows file name so Windows client using this clipboard function can directly use the FILEDESCRIPTOR struct without any additional work, while UNIX clients will have to convert UTF-16 into UTF-8 *and then* also replace backslashes with slashes. ConvertFromUnicode() won't do this for the client.

On Windows we use native FILEDESCRIPTOR definition. On non-Windows platforms we use our own struct from WinPR which is kept identical to the one on Windows. While it is totally possible to use a custom struct like WINPR_FILEDESCRIPTOR in both cases, I'd like to keep it named FILEDESCRIPTOR as in the spec and have it accurately represent the data received or sent to the server (which means UTF-16 file names with Windows-style separators). Non-Windows clients will not be able to use FILEDESCRIPTOR directly anyway, they will have to have their own custom structs for representing the files on the clipboard. Thus they can do any file name conversions they need by themselves.

Thus I do not see any problem with doing Stream_Read_UINT16() in a loop as that's exactly what we do here. @akallabeth's PR makes it prettier and more self-explanatory, so it would be nice to use those functions.

akallabeth:
@hardening Do you like the changes? Would like to merge this one.

PR #3906, opened 2017-04-10
Addin cast fix

akallabeth:
* Fixes the table generation (proper function pointer definitions and function arguments)
* Fixed some warnings concerning argument casts


akallabeth:
@freerdp-bot test

channels/urbdrc/client/searchman.h
```diff
@@ -57,7 +57,7 @@ struct _USB_SEARCHMAN
 	/* show all device in the list */
 	void (*show) (USB_SEARCHMAN* self);
 	/* start searchman */
-	BOOL (*start) (USB_SEARCHMAN* self, void * func);
+	BOOL (*start) (USB_SEARCHMAN* self, void* (*func)(void*));
 	/* close searchman */
```
hardening: Perhaps we could add a typedef for void *(*fun)(void *)

Issue #3910, opened 2017-04-11
Large black rectangle with /gfx in multi-monitor setup


T3N4K4:
Is it normal that I get a large black rectangle on my second monitor when using /gfx (it does not matter which gfx option I use)?
This only happens when I use /f /multimon:force to force the windows session (windows 2012r2) to be extended instead of stretched (<=which works).

xfreerdp /u: /p: /d: /v: /gdi:hw /f /multimon:force /bpp:32 -sec-tls +sec-nla /gfx:RFX -bitmap-cache /sound:sys:pulse,quality:high,latency:50 -compression /network:lan -fonts -aero -window-drag -menu-anims -themes +wallpaper -clipboard

I see the mouse on the second monitor but I cannot see any application which opens there or is moved on the second screen.

nfedera:
@T3N4K4 That's definitely not normal. Does it work with `/gdi:sw` instead of `/gdi:hw` ?

T3N4K4:
@nfedera Yes it works with /gdi:sw.

nfedera:
@T3N4K4 I can reproduce the issue and have already identified the culprit. A fix is in progress.

T3N4K4:
@nfedera Thank you :)

nfedera:
@T3N4K4 The fix is in master now. Please confirm and close the issue if it works for you.

T3N4K4:
@nfedera Thank you for your very fast response, works for me.

PR #3912, opened 2017-04-11
client/x11: fix gfx drawing bug with /multimon

nfedera:
The surface's damage region is not relative to the output but lives in its own universe starting at origin 0,0.
Also fixed the drawing coordinates used in the XRENDER code path which is used with /smart-sizing

This should fix the problem reported in issue #3910


nfedera:
@freerdp-bot please test again and ignore this false connect test error if you don't mind.

PR #3920, opened 2017-04-13
Disconnect/Reconnect channels if they are already connected before a server redirect

bjcollins:
Handle the case where a server redirection happens after channels are already set up by disconnecting the channels and running the post connect on the channels after the redirection connection is made. Also, free up dependent devices in rdpdr before freeing up their dependent resources.

This case is hit a 2012R2 Connection Broker redirects the client based on load-balance-info when NLA is not used for the connection, the symptom is that none of the channels are properly connected(no smartcard/audio/media/drive/etc. redirection). 

When NLA is enabled then the client is redirected before channels are connected, so the problem is not hit in that scenario.

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -1652,6 +1652,13 @@ static UINT rdpdr_virtual_channel_event_disconnected(rdpdrPlugin* rdpdr)
 {
 	UINT error;
 
+	/* Remove all dependent devices before tearing down resources */
+	if (rdpdr->devman)
+	{
+		devman_free(rdpdr->devman);
+		rdpdr->devman = NULL;
+	}
+
 	if (MessageQueue_PostQuit(rdpdr->queue, 0)
 	    && (WaitForSingleObject(rdpdr->thread, INFINITE) == WAIT_FAILED))
 	{
@@ -1685,12 +1692,6 @@ static UINT rdpdr_virtual_channel_event_disconnected(rdpdrPlugin* rdpdr)
 	{
 		Stream_Free(rdpdr->data_in, TRUE);
 		rdpdr->data_in = NULL;
-	}
-
-	if (rdpdr->devman)
-	{
-		devman_free(rdpdr->devman);
-		rdpdr->devman = NULL;
 	}
 
 	return error;
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -314,7 +314,17 @@ BOOL rdp_client_disconnect(rdpRdp* rdp)
 BOOL rdp_client_redirect(rdpRdp* rdp)
 {
 	BOOL status;
+	BOOL reconnect_channels = FALSE;
 	rdpSettings* settings = rdp->settings;
+
+	/* Ensure existing channels are disconnected */
+	rdpContext* context = rdp->context;
+	rdpChannels* channels = context->channels;
+	if (channels->connected)
+	{
+		reconnect_channels = TRUE;
+		freerdp_channels_disconnect(channels, context->instance);
+	}
 
 	rdp_client_disconnect(rdp);
 	if (rdp_redirection_apply_settings(rdp) != 0)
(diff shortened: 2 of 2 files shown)


bjcollins:
It seems like this might be the case where the TestConnect failure might not actually be from the change in this PR, since the test passed on the ubuntu build but not the ubuntu32 build.

akallabeth:
@freerdp-bot test
@bjcollins Looks good on a first glance, could you just add a little description here on how to test these changes?

libfreerdp/core/connection.c
```diff
@@ -314,8 +314,18 @@ BOOL rdp_client_disconnect(rdpRdp* rdp)
 BOOL rdp_client_redirect(rdpRdp* rdp)
 {
 	BOOL status;
+	BOOL reconnect_channels = FALSE;
 	rdpSettings* settings = rdp->settings;
 
+	/* Ensure existing channels are disconnected */
+	rdpContext* context = rdp->context;
+	rdpChannels* channels = context->channels;
+	if (channels->connected)
```
akallabeth: No need to check if the channels are already connected, see ```rdp_client_reconnect```

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -1652,6 +1652,13 @@ static UINT rdpdr_virtual_channel_event_disconnected(rdpdrPlugin* rdpdr)
 {
 	UINT error;
 
+	/* Remove all dependent devices before tearing down resources */
+	if (rdpdr->devman)
```
akallabeth: Are you sure this does not add a race condition? Would need to check that my self in depth, but I assume you just want to clear the device list first?

akallabeth: Something like ```void devman_clear(DEVMAN* devman)
{
	ListDictionary_Clear(devman->devices);
}```
should do what you want.

libfreerdp/core/connection.c
bjcollins: We need to know if the channels were connected in the first place, so that we can only call freerdp_channels_post_connect() if needed. What can happen otherwise it that we call it here, then it gets called again later(if redirect happens before normal call) which gives the error "CHANNEL_RC_ALREADY_CONNECTED". Once we call freerdp_channels_disconnect() then we cant tell since it resets that variable. It is possible, that rdp_client_reconnect may have the same potential problem.

channels/rdpdr/client/rdpdr_main.c
bjcollins: The goal is to make sure the "Free" for the devices gets called so that there threads will stop running, since they are dependent on the rdpdr channel. I am having trouble reproducing this issue this morning, but when I was having an issue I had devices getting an CHANNEL_NOT_CONNECTED error (it looks like the smartcard device code tries to handles this specific case gracefully). It seemed like it would make sense to try to free these resources earlier, but maybe you are right about the race condition - maybe it should happen right after stopping the rdpdr thread but before disconnecting the channel. I agree this needs to be looked at closer.

bjcollins:
For testing, I have been connecting to two different 2012R2 Connection Brokers - one which redirects to VMS on a Virtualization Host and one which redirects to a Session Host. I was using the load-balance-info to get the redirection to happen. The redirection happens before channels are connected when using NLA and after channels are connected when not using NLA. The original issue I was trying to resolve was freerdp not reconnecting any channels to the new server when NLA was disabled.

channels/rdpdr/client/rdpdr_main.c
akallabeth: @bjcollins That is why I mentioned the ```ListDictionary_Clear(devman->devices); }``` function. It removes all devices and calls the cleanup function registered without eliminating the list (which is assumed existing)

libfreerdp/core/connection.c
akallabeth: @bjcollins Good point. Either way it may be a good idea to also have a look at ```rdp_client_reconnect``` and keep it synced with this function.

akallabeth:
Will be merged in the upcoming fix for #4511

akallabeth:
Superseeded by #4548

PR #3924, opened 2017-04-19
Added a GFX area update callback.

akallabeth:
Added a new callback function allowing to update decoded GFX image data to some backend buffer. (e.g. CPU to GPU update)


akallabeth:
@freerdp-bot test

PR #3927, opened 2017-04-25
Bugfix: wrong test regarding the first GetPrinter

DavBfr:
This first call will always return FALSE because the goal of this call is to know the size needed.
https://msdn.microsoft.com/en-us/library/windows/desktop/dd144911(v=vs.85).aspx


bmiklautz:
@freerdp-bot test.

bmiklautz:
Exactly when size is set to 0 it's just used to figure out the memory to use.

PR #3932, opened 2017-04-29
resolve issues identified by cppcheck

chipitsine:
[channels/drive/client/drive_file.c:125]: (error) Memory leak: path_slash
[server/Windows/wf_dxgi.c:195]: (error) Invalid number of character '(' when these macros are defined: 'WITH_DXGI_1_2'.

### DIFF
--- a/channels/drive/client/drive_file.c
+++ b/channels/drive/client/drive_file.c
@@ -112,6 +112,10 @@ static BOOL drive_file_remove_dir(const WCHAR* path)
 	WCHAR* fullpath;
 	WCHAR* path_slash;
 	UINT32 base_path_length;
+
+	if (!path)
+		return FALSE;
+
 	base_path_length = _wcslen(path) * 2;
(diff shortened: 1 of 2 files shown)


channels/drive/client/drive_file.c
```diff
@@ -122,7 +122,10 @@ static BOOL drive_file_remove_dir(const WCHAR* path)
 	}
 
 	if (!path)
```
ilammy: Well, this is kinda bad. Note that `path` is referenced on the line 115. The compiler is likely to remove this condition entirely on the basis that NULL check is redundant because passing NULL to _wcslen() is undefined behavior thus it must never be NULL.

I'd move this check to the start of the function, before doing anything, and drop the `free()` call.

chipitsine: it makes sense.
done

bmiklautz:
@freerdp-bot test.

PR #3937, opened 2017-05-02
Fix h264 in multimon scenario

hardening:
h264 contexts are specific to each surface, so in multi-monitor (with 2 surfaces and so 2 h264 contexts) decoding was not working correctly.


hardening:
anyone to merge this one ?

akallabeth:
Merging as I'm using this patch for ~3 days in production now.

Issue #3939, opened 2017-05-03
Remote files are listed with huge wrong size under 32bit client platforms


AngelNone:
Good day. There was a problem. When you connect to RDP from Linux Mint, the files on the attached drive from the local computer increase the size from 2 bytes to 8 gigabytes. What is the problem with this problem? We use the latest version. When connecting with xreerdp the same problem, I attach a screenshot it may be easier to understand what is happening.

You can also view the reference on the link https://github.com/FreeRDP/Remmina/issues/1166 where the screenshots of the problem are indicated

giox069:
I have been able to reproduce it on intel 32bit platforms. Not on 64bits.

AngelNone:
On a 64 bit system, everything works well. There are no comments, only there is a 32 bit bit.

ilammy:
I'd reckon this has to do something with the disk drive redirection refactoring in #3864.

AngelNone:
I'm unfortunately not familiar with the internal mechanisms of the RDP client. Tell me how to solve this problem?

giox069:
@AngelNone I don't think you can fix it by yourself if you are not a C programmer.
You should wait that someone of the FreeRDP team will fix it, and then we will push the fix to the Remmina PPA.

akallabeth:
May be related to #3913. Fixed in https://github.com/akallabeth/FreeRDP/tree/shift_warn_fix
Can anyone give it a try and report back?

AngelNone:
@akallabeth I'm new to Linux. I downloaded the folder from the link. Unzip it ... Then went to the link https://github.com/FreeRDP/FreeRDP/wiki/Compilation and step by step done what is said there except download. I made these steps from the folder in which I unzipped the downloaded archive. Everything went smoothly. But the problem did not dare I do not deny that maybe something did wrong.

antenore:
@AngelNone maybe you can follow our guide? 

https://github.com/FreeRDP/Remmina/wiki/Compile-on-Ubuntu-16.04
Just change the FreeRDP repository with the akallabeth one.

Be sure to remove any previous version of FreeRDP and Remmina you have installed.

I don't have a 32bit platform so I cannot test myself and I think Giovanni is busy now.

AngelNone:
[...]

Issue #3913, opened 2017-04-11
right shift count warning in winpr/libwinpr/file/generic.c for 32 bit builds


hendwolt:
This is a follow up for #3903, where this was the second of two issues.
After the first issue got fixed, this is still open:

[   60s] /home/abuild/rpmbuild/BUILD/FreeRDP-master/winpr/libwinpr/file/generic.c: In function 'FindNextFileA':
[   60s] /home/abuild/rpmbuild/BUILD/FreeRDP-master/winpr/libwinpr/file/generic.c:979:53: warning: right shift count >= width of type [-Wshift-count-overflow]
[   60s]     lpFindFileData->nFileSizeHigh = fileStat.st_size >> 32;
[   60s]                                                      ^~

Happens in 32 bit builds only.
[osTWbuild_log_20170411.zip](https://github.com/FreeRDP/FreeRDP/files/914319/osTWbuild_log_20170411.zip)

hendwolt:
I investigated this a little further and think, it may be a fallout of an openSUSE bug (#1007123), where someone (I hope by accident) switched off the support for files larger than 2GB in some programs and libraries for 32bit.

TL;DR:
To change this, a compiler flag _FILE_OFFSET_BITS=64 needs to be set. 

On the other hand, I don't know, if there are any architectures out there with no large file support. If such architectures exists, this 32 bit shift should not be done unconditionally.

hendwolt:
If I want to enable _FILE_OFFSET_BITS=64 for gcc (only), I use the attached patch
[cmake_64bit.patch.zip](https://github.com/FreeRDP/FreeRDP/files/962201/cmake_64bit.patch.zip)

bmiklautz:
@hendwolt should be fixed with the current master if I'm not mistaken. Can you verify that?

hendwolt:
Yes, it is fixed.

PR #3942, opened 2017-05-05
Fixed right shift on 32bit platforms.

akallabeth:


hendwolt:
I do still suggest, to set the appropriate compiler options, to make fileStat.st_size a uint64 value:
-D_FILE_OFFSET_BITS=64
instead of casting.

ilammy:
I'm just curious why exactly is that an issue if file size is 32-bit?

I know that signed overflow when left-shifting is undefined, and there is an issue with right shifting negative values (which are never transformed to zero when shifted right by any amount). However, both of these are actual whether off_t is 32-bit signed integer or 64-bit signed integer. And file sizes should never be negative in the first place.

I believe the issue is with that off_t is _signed_, so if we are going to split it into low and high parts we should cast it to something unsigned for the bit magic to work. But I don't understand what issue this fixes.

hendwolt:
If off_t is a 32-bit type the maximum file size, that can be handled by this program, is limited to 4GB. 
But 32-bit (Linux) systems are capable of handling much larger files. But this works only, if **all** involved programs and libraries use the switch  "_FILE_OFFSET_BITS=64" to make off_t a 64-bit type.
See https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html for more details.

akallabeth:
@hendwolt Not sure this can be added without side effects. Read the docs and this only works if the system provides 64 bit file support, correct?

hendwolt:
@akallabeth 
This flag _FILE_OFFSET_BITS=64 is necessary on 32-bit Linux systems. They usually provide "Large File Support" for many years. On 64-bit Linux this flag is simply ignored; according to the GNU documentation. 
I am not sure about other platforms and if it works for other compilers like clang.
I'm currently using the patch attached to #3913 to build freerdp for openSUSE Tumbleweed i586 on OBS (only for gcc) and it works for me.

Without the patch: If a program with a 32-bit off_t "hits" a file larger than 4GB, the function, that should fill the structure "fileStat", will not do this, but return an error instead (EOVERFLOW).

akallabeth:
@hendwolt Added the define for linux systems only, hope that fixes this :)

akallabeth:
This fixes #3939 where multiple users already confirm that the issue is fixed with this pull.

hardening:
Everyone agrees, let's merge it.

PR #3948, opened 2017-05-10
Remove library prefix override for rdpei-server

hypherion2:
This library was missed in commit
059374457d801351d02ac81badb4aabbfa7689e7, so it was the only one that
still had a lib prefix.


akallabeth:
@freerdp-bot test

PR #3950, opened 2017-05-11
Fix handle leaks in encomsp and remdesk server

hypherion2:
Ensure proper thread cleanup when these channel servers are stopped.


hardening:

PR #3952, opened 2017-05-12
Added support for Set Keyboard IME Status

akallabeth:


akallabeth:
@freerdp-bot test, come on!

PR #3953, opened 2017-05-12
Drop some annoying warning

hardening:
Probably that some unused functions should be removed, but at least it makes it compile with no warnings.

### DIFF
--- a/channels/encomsp/client/encomsp_main.c
+++ b/channels/encomsp/client/encomsp_main.c
@@ -916,31 +916,6 @@ static void encomsp_process_connect(encomspPlugin* encomsp)
 {
 }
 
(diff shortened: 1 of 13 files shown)


hardening:
@akallabeth I've removed what looked really unused, there's some parts of code that will need to be there if we implement the feature (like in capabilities with gdiplus for example). These parts of code are nice and there so I didn't want to remove them.

akallabeth:
@hardening Figured as much. Still, then it is better to have the warning. That way the code at least compiles and issues with changed data structures et al are caught by the pull request tester. If commented they will not be tested and will end up broken anyway.

hardening:
Ok so I've removed the #if 0.

PR #3416, opened 2016-06-20
redirect specific smartcard readers

HenryJacques:
Until now, /smartcard: had the behaviour to redirect every connected readers. It didn't matter what you used after the parameter /smartcard:xxxx:yyyy.

With this patch you can:
-    use /smartcard: "as is" (without argument) to continue to redirect every readers
-    use /smartcard:"Reader name" (you have to repeat that for every reader you want to redirect)

The reader name is exactly the one returned by libpcsclite, e.g.

> root@host:# opensc-tool -l
> Detected readers (pcsc)
> 
> Nr. Card Features Name
> 0 Yes Xiring Leo V2 (12547854125036) 00 00
> 1 Yes Xiring Leo V2 (52412506854712) 01 00

(Thanks to A. Aing for this PR)

PR #3956, opened 2017-05-15
redirect specific smartcard readers

dodo040:
Rebased pull-request #3416 : redirect specific smartcard readers.

Until now, /smartcard: had the behaviour to redirect every connected readers. It didn't matter what you used after the parameter /smartcard:xxxx:yyyy.

With this patch you can:

use `/smartcard:` "as is" (without argument) to continue to redirect every readers
use `/smartcard:"Reader name"` (you have to repeat that for every reader you want to redirect)

The reader name is exactly the one returned by libpcsclite, e.g.

    root@host:# opensc-tool -l
    Detected readers (pcsc)

    Nr. Card Features Name
    0 Yes Xiring Leo V2 (12547854125036) 00 00
    1 Yes Xiring Leo V2 (52412506854712) 01 00

### DIFF
--- a/channels/smartcard/client/smartcard_main.c
+++ b/channels/smartcard/client/smartcard_main.c
@@ -45,7 +45,6 @@ void* smartcard_context_thread(SMARTCARD_CONTEXT* pContext)
 	SMARTCARD_OPERATION* operation;
 	UINT error = CHANNEL_RC_OK;
 	smartcard = pContext->smartcard;
-
 	nCount = 0;
 	hEvents[nCount++] = MessageQueue_Event(pContext->IrpQueue);
 
@@ -516,7 +515,6 @@ static void* smartcard_thread_func(void* arg)
 	wMessage message;
 	SMARTCARD_DEVICE* smartcard = (SMARTCARD_DEVICE*) arg;
 	UINT error = CHANNEL_RC_OK;
-
 	nCount = 0;
 	hEvents[nCount++] = MessageQueue_Event(smartcard->IrpQueue);
 	hEvents[nCount++] = Queue_Event(smartcard->CompletedIrpQueue);
@@ -739,6 +737,15 @@ UINT DeviceServiceEntry(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints)
 			smartcard->path = name;
 		else
 			smartcard->name = name;
+	}
+
+	LONG status;
+	status = SCardAddReaderName(&smartcard->thread, (LPSTR) name);
+
+	if (status != SCARD_S_SUCCESS)
+	{
+		WLog_ERR(TAG, "Failed to add reader name!");
+		goto error_device_data;
 	}
 
 	smartcard->IrpQueue = MessageQueue_New(NULL);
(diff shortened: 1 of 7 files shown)


akallabeth:
@freerdp-bot test

channels/smartcard/client/smartcard_main.c
```diff
@@ -741,6 +739,15 @@ UINT DeviceServiceEntry(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints)
 			smartcard->name = name;
 	}
 
+	LONG status;
```
hardening: freerdp-bot complains here about LONG

client/common/cmdline.c
```diff
@@ -779,7 +777,7 @@ static char** freerdp_command_line_parse_comma_separated_values_offset(
 
 	p = t;
 
-	if (count)
+	if (*count)
```
akallabeth: This looks wrong. The old version did check for NULL pointers, this dereferences...

akallabeth:
@freerdp-bot test

client/common/cmdline.c
dodo040: Checking pointer here is redundant because we already did it in `freerdp_command_line_parse_comma_separated_values ` with `assert(NULL != count);`

dodo040:
@akallabeth what's wrong with Windows builds ?

hardening:
@freerdp-bot looks like you could re-test for windows...

dodo040:
@akallabeth what about the merge ?

akallabeth:
@dodo040 wanted to test the stuff with hardware first, but failed to allocate time

client/common/cmdline.c
```diff
@@ -853,13 +851,24 @@ static int freerdp_client_command_line_post_filter(void* context,
...
 	CommandLineSwitchCase(arg, "smartcard")
 	{
-		char** p;
-		int count;
-		p = freerdp_command_line_parse_comma_separated_values_offset(arg->Value,
-		        &count);
-		p[0] = "smartcard";
-		status = freerdp_client_add_device_channel(settings, count, p);
-		free(p);
+		if (arg->Flags & COMMAND_LINE_VALUE_PRESENT)
```
akallabeth: This new version is just more complex than the old implementation.

client/common/cmdline.c
```diff
@@ -853,13 +851,24 @@ static int freerdp_client_command_line_post_filter(void* context,
...
+			status = freerdp_client_add_device_channel(settings, count, p);
+			free(p);
+		}
+		else
+		{
+			char* p[2];
+			int count;
+			count = 2;
+			p[0] = "smartcard";
+			p[1] = "";
```
dodo040: When we don't have any argument, this allows us to just write `/smartcard ` instead of `/smartcard:`. Note the "**:**".

akallabeth: Then there is something else wrong with argument parsing. When you add the ```:``` it is interpreted as an (empty) argument. The case without should work nonetheless.

akallabeth:
@freerdp-bot test

dodo040:
> @dodo040 wanted to test the stuff with hardware first, but failed to allocate time

@akallabeth Have you managed to allocate time for this PR ?

akallabeth:
@dodo040 Yes, did some tests finally.
Following two things I've notices:
* ```smartcard_init``` releases all device contexts, looks wrong (and prevents any successful tests on my side)
* When connecting without a reader name I get errors when trying to connect with gemalto PCSC Bridge

Oh, and could you please rebase instead of the merge you did? Would remove the additional merge entries in the git log.

akallabeth:
@freerdp-bot test

PR #3957, opened 2017-05-15
Use a preallocated logger instance for GFX

akallabeth:
The static logger functions of WLog have quite a heavy performance
penalty. Use a pointer to an allocated logger to speed things up.


akallabeth:
@hardening yes, the debug messages in normal execution path slow down frame ack and lead to reduced image quality with ms servers

Issue #3962, opened 2017-05-16
drive redirection not working with a slash at the end of the path


hendwolt:
Short version:

works: /drive:hendhome,/home/hendrik
works NOT: /drive:hendhome,/home/hendrik/

Long version:
I'm using the master version of yesterday (2017-05-15) with the following options: 
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX:PATH=/usr -DCMAKE_INSTALL_LIBDIR:PATH=lib -DCMAKE_SKIP_INSTALL_RPATH=ON -DWITH_CUPS=ON -DWITH_CLIENT=ON -DWITH_PCSC=ON -DWITH_ALSA=ON -DWITH_PULSE=ON -DWITH_DIRECTFB=OFF -DWITH_JPEG=ON -DWITH_GSM=ON -DWITH_GSTREAMER_1_0=ON -DWITH_IPP=OFF -DWITH_OPENSSL=ON -DWITH_CHANNELS=ON -DWITH_X11=ON -DWITH_XCURSOR=ON -DWITH_XEXT=ON -DWITH_XKBFILE=ON -DWITH_XI=ON -DWITH_XINERAMA=ON -DWITH_XRENDER=ON -DWITH_XV=ON -DWITH_ZLIB=ON -DWITH_KRB5=ON -DBUILTIN_CHANNELS=OFF -DWITH_SERVER=OFF -DWITH_SSE2=ON -DWITH_DEBUG_SCARD=OFF -DWITH_SMARTCARD_INSPECT=ON -DWITH_WAYLAND=ON -DWITH_LIBSYSTEMD=ON -DCMAKE_NO_BUILTIN_CHRPATH=ON .

My clients: xfreerdp on openSUSE Tumbleweed 32 bit, openSUSE Leap 42.2 64 bit
Targets: Windows7 Prof. (64 bit), Windows10 Pro (Insider Preview, 32 bit)

symptoms: When I use a path with a slash at the end (/home/hendrik/) I see the share "hendhome" on the target system, but the folder is empty. Without the trailing slash everything works and I see the content of the folder.

PR #3965, opened 2017-05-17
Drive fixes unicode

akallabeth:


akallabeth:
@freerdp-bot test

hendwolt:
I tested the patch in my OBS repository for openSUSE Tumbleweed. Removing the trailing slash works and solves #3962 for me.

PR #3966, opened 2017-05-18
Add a callback to provide NTLM hashes on server-side

davewheel:
Adds a callback that allows servers to compute NTLM hashes by themselves. The typical
use of this callback is to provide a function that gives precomputed hash values.


hardening:
Interesting, that's the server-side part, perhaps the path-the-hash option could be implemented that way with client... Anyway @freerdp-bot can you give a test ?

akallabeth:
@freerdp-bot test

Issue #3983, opened 2017-05-30
Clipboard with HTML data


Vladimir-csp:
If clipboard contains HTML data, it can not be pasted into applications expecting formatted text.
To reproduce: on Linux machine open browser (I tested with Firefox and Thunderbird's formatted mail), copy formatted text. Connect to windows machine with freerdp, open application expecting formatted text (I tested with LibreOffice Writer), paste. Nothing or partial garbled content is pasted.
This bug is observed with freerdp `2.0.0~git20161130.1.e60d0d5+dfsg1-1` on Debian jessie and stretch.
This bug is not observed with freerdp `2.0.0+0~20151219024829.202~1.gbpdf81e8` on debian jessie

ilammy:
@Vladimir-csp could you please build FreeRDP with cmake flags like these:
`cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_DEBUG_CLIPRDR=ON .`
then run FreeRDP with this log filter:
`WLOG_FILTER=com.freerdp.channels.cliprdr.client:DEBUG ./client/X11/xfreerdp ...`
and paste the standard output here?

I have tried reproducing the issue. FreeRDP is announcing HTML format for the remote desktop, but the remote application still requests Unicode text for some reason.

From what I see in the history from df81e84 to e60d0d5 I'd blame #2851 for introducing a bug.

ilammy:
Okay, I have updated Firefox (on my Debian) and LibreOffice (on Windows VM) and it seems that I have managed to reproduce the issue. When I copy page fragments in Firefox and try to paste them into LO Writer in the remote session I only see `<` to be pasted.

The data sent to the server is valid `"HTML Format"`, and it really contains only `<` character. Debugging shows that Firefox provides clipboard data in UTF-16 and FreeRDP expects it to be null-terminated UTF-8. This can cause various weird issues with encoding. Props to me for my 'great' testing effort in #2851

To fix this I believe we will have to teach FreeRDP to guess the encoding from BOM or something like that.

ilammy:
Actually, FreeRDP already does that detection, it just was not able to do that because of a 'helpful' fixup of clipboard data length.

@Vladimir-csp, please check whether #3985 fixes the issue for you.

Vladimir-csp:
Thanks!
I've tested it and it seems to be working.

akallabeth:
Closing as reported as fixed with #3985

PR #3985, opened 2017-05-31
Do not fixup HTML data length

ilammy:
The data provided by local applications can be actually encoded in UTF-16 (e.g., Firefox does this to HTML). UTF-16 allows embedded null bytes so we should not use strlen() to fix up the data. The HTML format synthesizer can handle trailing null bytes just fine and can detect whether it deals with UTF-8 or UTF-16.

This should resolve issue #3983.


akallabeth:
@freerdp-bot test

Issue #3229, opened 2016-03-15
freerdp Windows Server 2008 R2 - Barcode Scanner USB - missing double characters 


ropetry:
Hello, we are using freerdp 2.0.0-dev with our 2008 R2 Terminal Servers. Attached to the Linux Device is a USB Barcode scanner from OPTICON.
If we scan Barcodes with Double Numbers via freerdp , like
02612295 or 02422288 we are loosing in 9 from 10 scans the double 22 or 88.. It just prints only one 2 or one 8.
Label:
02422288
barcode scan:
024228

.. If i do this on the Linux client in a xterm it works without a problem.. also rdesktop works without any problems. it seeams to be a timing issue ? i already has configured the usb barcode reader to use a bigger delay between the intercharacters, but it doesn´t help.

Output of dmesg:
[12396.196149] usb 1-1: USB disconnect, device number 2
[12402.708041] usb 1-1: new full-speed USB device number 3 using uhci_hcd
[12403.089079] usb 1-1: New USB device found, idVendor=065a, idProduct=0001
[12403.089088] usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[12403.089094] usb 1-1: Product: USB Code Reader
[12403.089099] usb 1-1: Manufacturer: Optoelectronics Co., Ltd.
[12403.097724] input: Optoelectronics Co., Ltd. USB Code Reader as /devices/pci0000:00/0000:00:1d.0/usb1/1-1/1-1:1.0/0003:065A:0001.0009/input/input13
[12403.098422] hid-generic 0003:065A:0001.0009: input,hidraw0: USB HID v1.00 Keyboard [Optoelectronics Co., Ltd. USB Code Reader] on usb-0000:00:1d.0-1/input0
~ #
output of the second reader:

Kernel is 3.16.7-29

if i change the opticon device and use an other usb barcode device the behavior is the same..
[12672.608058] usb 1-1: new low-speed USB device number 4 using uhci_hcd
[12672.969091] usb 1-1: New USB device found, idVendor=040b, idProduct=6543
[12672.969099] usb 1-1: New USB device strings: Mfr=0, Product=0, SerialNumber=0
[12672.984681] hid-generic 0003:040B:6543.000A: hiddev0,hidraw0: USB HID v1.10 Device [HID 040b:6543] on usb-0000:00:1d.0-1/input0
[...]

PR #3992, opened 2017-06-05
  Fix issue #3229 Barcode Scanner USB - missing double characters 

johnjmullen:
With some USB barcode scanners, repeated characters do not appear in the FreeRDP session.
It looks like this is because the KeyRelease signal is not sent for the first character.

This fixes the problem, but I'm not sure if there is a reason these KeyRelease events need to be ignored

### DIFF
--- a/client/X11/xf_event.c
+++ b/client/X11/xf_event.c
@@ -492,18 +492,6 @@ static BOOL xf_event_KeyRelease(xfContext* xfc, XEvent* event, BOOL app)
 	KeySym keysym;
 	char str[256];
 
-	if (XPending(xfc->display))
-	{
-		ZeroMemory(&nextEvent, sizeof(nextEvent));
-		XPeekEvent(xfc->display, &nextEvent);
-
-		if (nextEvent.type == KeyPress)
-		{
-			if (nextEvent.xkey.keycode == event->xkey.keycode)
-				return TRUE;
-		}
-	}
-
 	XLookupString((XKeyEvent*) event, str, sizeof(str), &keysym, NULL);
 	xf_keyboard_key_release(xfc, event->xkey.keycode, keysym);
 	return TRUE;


hardening:
hum, could you write something in your commit log about this commit ?

johnjmullen:
It's possible that this will break stuff. I'm not sure whether there is a reason to ignore some KeyRelease events.  It looks like those lines have always been there, so they might be necessary.

There was a patch in one of the comments on the bug report that sent a key release in xf_event_KeyPress, but I was concerned it might cause issues with modifier keys or holding a key down.  I could try that instead if you think it would be better.

johnjmullen:
Is this code meant for disabling auto repeat? I couldn't figure out why it was there, but I didn't think of that until now.

akallabeth:
@johnjmullen That may be possible, I'm not quite sure myself, need to dig into that.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

johnjmullen:
This change fixes the Intermec SG20B barcode scanner, and I have not been able to find any problems with the keyboards I have.  I think the code I removed might be related to autorepeat, but everything appears to work correctly without it.

Is there anything else that I should test to confirm it will not cause problems?

johnjmullen:
@akallabeth 
I asked our QA team to run their keyboard tests against this patch and approved it:
"Approved. Tested French, German, Spanish (LA), Simplified and Traditional Chinese, and Japanese layouts. Also tested hotkey support in Windows 2008 and 2012. The only functions that didn't work were the Calculator and the Play/Pause Media Keys."

Let me know if there is anything else that needs to be tested or if you need me to make any changes.

johnjmullen:
@akallabeth Were you going to merge this?

PR #4000, opened 2017-06-12
Command line ignore empty if flag set

akallabeth:
When CommandLineParseArgumentsA is called with flag
COMMAND_LINE_IGN_UNKNOWN_KEYWORD return success if the
command line is empty.


akallabeth:
@hardening @mfleisz @bmiklautz Looks like this one is now necessary, as the more strict argument checking fails for microphone/sound now.

PR #4013, opened 2017-06-21
fix FreeRDP some memory leak problem

weizhenwei:
using valgrind to check these memory leak, and fix them.

### DIFF
--- a/channels/drive/client/drive_main.c
+++ b/channels/drive/client/drive_main.c
@@ -869,7 +869,6 @@ out_error:
 	MessageQueue_Free(drive->IrpQueue);
 	ListDictionary_Free(drive->files);
 	drive_free(drive);
-	free(drive);
 	return error;
 }
 
--- a/channels/drive/client/drive_main.c
+++ b/channels/drive/client/drive_main.c
@@ -757,6 +757,7 @@ static UINT drive_free(DEVICE* device)
 	ListDictionary_Free(drive->files);
 	MessageQueue_Free(drive->IrpQueue);
 	Stream_Free(drive->device.data, TRUE);
+	free(drive->path);
 	free(drive);
 	return error;
 }
@@ -867,6 +868,7 @@ UINT drive_register_drive_path(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints,
 out_error:
 	MessageQueue_Free(drive->IrpQueue);
 	ListDictionary_Free(drive->files);
+	free(drive->path);
 	free(drive);
 	return error;
 }
--- a/channels/drive/client/drive_main.c
+++ b/channels/drive/client/drive_main.c
@@ -868,7 +868,7 @@ UINT drive_register_drive_path(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints,
 out_error:
 	MessageQueue_Free(drive->IrpQueue);
 	ListDictionary_Free(drive->files);
-	free(drive->path);
+	drive_free(drive);
 	free(drive);
 	return error;
 }
--- a/channels/drive/client/drive_main.c
+++ b/channels/drive/client/drive_main.c
@@ -866,8 +866,6 @@ UINT drive_register_drive_path(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints,
 
 	return CHANNEL_RC_OK;
 out_error:
-	MessageQueue_Free(drive->IrpQueue);
-	ListDictionary_Free(drive->files);
 	drive_free(drive);
 	return error;
 }
--- a/channels/rail/client/rail_main.c
+++ b/channels/rail/client/rail_main.c
@@ -100,23 +100,14 @@ static void rail_client_clean_exec_order(RAIL_EXEC_ORDER* exec)
     if (!exec)
         return;
 
-    if (exec->exeOrFile.string)
-    {
-    	free(exec->exeOrFile.string);
-    	exec->exeOrFile.string = NULL;
-    }
+    free(exec->exeOrFile.string);
+    exec->exeOrFile.string = NULL;
 
-    if (exec->workingDir.string)
-    {
-		free(exec->workingDir.string);
-		exec->workingDir.string = NULL;
-    }
+    free(exec->workingDir.string);
+    exec->workingDir.string = NULL;
 
-    if (exec->arguments.string)
-    {
-		free(exec->arguments.string);
-		exec->arguments.string = NULL;
-    }
+    free(exec->arguments.string);
+    exec->arguments.string = NULL;
 }
 
 /**
--- a/channels/rail/client/rail_main.c
+++ b/channels/rail/client/rail_main.c
@@ -97,9 +97,26 @@ UINT rail_send_channel_data(railPlugin* rail, void* data, size_t length)
  */
 static void rail_client_clean_exec_order(RAIL_EXEC_ORDER* exec)
 {
-	free(exec->exeOrFile.string);
-	free(exec->workingDir.string);
-	free(exec->arguments.string);
+    if (!exec)
+        return;
+
+    if (exec->exeOrFile.string)
+    {
+    	free(exec->exeOrFile.string);
+    	exec->exeOrFile.string = NULL;
+    }
+
+    if (exec->workingDir.string)
+    {
+		free(exec->workingDir.string);
+		exec->workingDir.string = NULL;
+    }
+
+    if (exec->arguments.string)
+    {
+		free(exec->arguments.string);
+		exec->arguments.string = NULL;
+    }
 }
 
 /**
--- a/channels/rail/client/rail_main.c
+++ b/channels/rail/client/rail_main.c
@@ -104,6 +104,7 @@ static UINT rail_client_execute(RailClientContext* context,
                                 RAIL_EXEC_ORDER* exec)
 {
 	char* exeOrFile;
+	UINT error;
 	railPlugin* rail = (railPlugin*) context->handle;
 	exeOrFile = exec->RemoteApplicationProgram;
 
@@ -122,11 +123,11 @@ static UINT rail_client_execute(RailClientContext* context,
 	                              &exec->workingDir); /* ShellWorkingDirectory */
 	rail_string_to_unicode_string(exec->RemoteApplicationArguments,
 	                              &exec->arguments); /* RemoteApplicationCmdLine */
-	UINT ret = rail_send_client_exec_order(rail, exec);
+	error = rail_send_client_exec_order(rail, exec);
 	free(exec->exeOrFile.string);
 	free(exec->workingDir.string);
 	free(exec->arguments.string);
-	return ret;
+	return error;
 }
 
 /**
--- a/channels/rail/client/rail_main.c
+++ b/channels/rail/client/rail_main.c
@@ -122,7 +122,11 @@ static UINT rail_client_execute(RailClientContext* context,
 	                              &exec->workingDir); /* ShellWorkingDirectory */
 	rail_string_to_unicode_string(exec->RemoteApplicationArguments,
 	                              &exec->arguments); /* RemoteApplicationCmdLine */
-	return rail_send_client_exec_order(rail, exec);
+	UINT ret = rail_send_client_exec_order(rail, exec);
+	free(exec->exeOrFile.string);
+	free(exec->workingDir.string);
+	free(exec->arguments.string);
+	return ret;
 }
 
 /**
--- a/channels/rail/client/rail_main.c
+++ b/channels/rail/client/rail_main.c
@@ -628,13 +628,13 @@ static void* rail_virtual_channel_client_thread(void* arg)
 		{
 			data = (wStream*) message.wParam;
 
-			if ((error = rail_order_recv(rail, data)))
+			error = rail_order_recv(rail, data);
+			Stream_Free(data, TRUE);
+			if (error)
 			{
-				Stream_Free(data, TRUE);
 				WLog_ERR(TAG, "rail_order_recv failed with error %"PRIu32"!", error);
 				break;
 			}
-			Stream_Free(data, TRUE);
 		}
 	}
 
--- a/channels/rail/client/rail_main.c
+++ b/channels/rail/client/rail_main.c
@@ -92,6 +92,17 @@ UINT rail_send_channel_data(railPlugin* rail, void* data, size_t length)
 }
 
 /**
+ * used by rail_client_execute() to free RAIL_EXEC_ORDER's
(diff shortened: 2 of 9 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@weizhenwei thanks. Would you update your pull request to not contain merges?

weizhenwei:
remove all merge commits, and fix compile error on win64/vs2010 platform.

weizhenwei:
@bmiklautz , please help to check.

bmiklautz:
@freerdp-bot test

weizhenwei:
I checked Jenkins output: "[WS-CLEANUP] Deleting project workspace...
ERROR: [WS-CLEANUP] Cannot delete workspace: remote file operation failed: c:\Devel\workspace\windows_msbuild\arch\win32\label\vs2013 at hudson.remoting.Channel@4429bfe2:JNLP4-connect connection from frdp-win7.in.freerdp.com/10.20.50.14:49168: java.io.IOException: Unable to delete 'c:\Devel\workspace\windows_msbuild\arch\win32\label\vs2013\build\CMakeCache.txt'. Tried 3 times (of a maximum of 3) waiting 0,1 sec between attempts."  What could I do about this?

bmiklautz:
@weizhenwei I'll have a look.

libfreerdp/core/update.c
```diff
@@ -2139,6 +2139,17 @@ void update_free(rdpUpdate* update)
 		free(update->primary);
 		free(update->secondary);
 		free(update->altsec);
+		free(update->window->monitored_desktop.windowIds);
+		free(update->window->window_state.titleInfo.string);
```
hardening: may I suggest to create function to free the window_state (it could be used below) ?

libfreerdp/core/window.c
```diff
@@ -350,6 +352,14 @@ BOOL update_read_window_state_order(wStream* s, WINDOW_ORDER_INFO* orderInfo, WI
 
 BOOL update_read_window_icon_order(wStream* s, WINDOW_ORDER_INFO* orderInfo, WINDOW_ICON_ORDER* window_icon)
 {
+	if (window_icon->iconInfo)
+	{
+		free(window_icon->iconInfo->bitsColor);
```
hardening: same remark here, it would be neat to have a function to free the icon info and also use it above.

libfreerdp/core/update.c
weizhenwei: OK, I will create a function to free window_state and call it at multiple place.

bmiklautz:
@freerdp-bot test.

weizhenwei:
@hardening , @bmiklautz , I refined the patch to remove redundant code copy, please check, thank you very much.

winpr/libwinpr/file/generic.c
```diff
@@ -938,6 +938,7 @@ HANDLE FindFirstFileW(LPCWSTR lpFileName, LPWIN32_FIND_DATAW lpFindFileData)
 		}
 	}
 
+	free(fd);
 	return h;
 }
```
ilammy: It's a matter of style, but with functions that deal with resources I personally like to have a single return point and structure error handling with goto. This ensures that the resources are always freed.

Like this:
```diff
 	if (h != INVALID_HANDLE_VALUE)
 	{
 		if (!ConvertFindDataAToW(fd, lpFindFileData))
 		{
 			SetLastError(ERROR_NOT_ENOUGH_MEMORY);
-			free(fd);
-			return INVALID_HANDLE_VALUE;
+			h = INVALID_HANDLE_VALUE;
+			goto out;
  		}
  	}
-
+out:
+	free(fd);
  	return h;
  }
```
Though maybe it's an overkill in this case.

channels/rail/client/rail_orders.c
```diff
@@ -650,7 +650,9 @@ UINT rail_order_recv(railPlugin* rail, wStream* s)
 		case RDP_RAIL_ORDER_EXEC_RESULT:
 			{
 				RAIL_EXEC_RESULT_ORDER execResult;
-				return rail_recv_exec_result_order(rail, &execResult, s);
+				error = rail_recv_exec_result_order(rail, &execResult, s);
+				free(execResult.exeOrFile.string);
```
ilammy: This field is not actually initialized in execResult. If `rail_recv_exec_result_order()` fails before filling in the string (e.g., in `rail_read_server_exec_result_order()`) the we'll get heap corruption here.

I'd recommend to zero-initialize execResult (and other structures in this function, while we're at it). I guess you can try adding `= { 0 }` to declarations, but if that fails then explicit `ZeroMemory` calls should do the trick.

libfreerdp/core/window.c
```diff
@@ -294,6 +294,7 @@ BOOL update_read_window_state_order(wStream* s, WINDOW_ORDER_INFO* orderInfo, WI
 
 		Stream_Read_UINT16(s, windowState->numWindowRects); /* numWindowRects (2 bytes) */
 
+		free(windowState->windowRects);
 		size = sizeof(RECTANGLE_16) * windowState->numWindowRects;
 		windowState->windowRects = (RECTANGLE_16*) malloc(size);
```
ilammy: I guess we can do realloc() here instead of free() + malloc(). You have to be careful to handle realloc() failure, but I think this field will be eventually freed, so we can simply return here if realloc fails, just as we do now for malloc().

libfreerdp/core/window.c
```diff
@@ -328,6 +329,7 @@ BOOL update_read_window_state_order(wStream* s, WINDOW_ORDER_INFO* orderInfo, WI
 
 		Stream_Read_UINT16(s, windowState->numVisibilityRects); /* numVisibilityRects (2 bytes) */
 
+		free(windowState->visibilityRects);
 		size = sizeof(RECTANGLE_16) * windowState->numVisibilityRects;
 		windowState->visibilityRects = (RECTANGLE_16*) malloc(size);
```
ilammy: Here is the same realloc() thing as above.

channels/rail/client/rail_orders.c
ilammy: Oh, nevermind. I missed the `ZeroMemory` call in rail_recv_exec_result_order(). It is too sneaky placed near the variable declarations.

So there are no errors here, the free() call is fine.

channels/rail/client/rail_main.c
```diff
@@ -620,9 +620,11 @@ static void* rail_virtual_channel_client_thread(void* arg)
 
 			if ((error = rail_order_recv(rail, data)))
 			{
+				Stream_Free(data, TRUE);
 				WLog_ERR(TAG, "rail_order_recv failed with error %"PRIu32"!", error);
 				break;
 			}
+			Stream_Free(data, TRUE);
```
weizhenwei: Thank you for  the suggestion, it's good to me, I will refactor to avoid dulicating Stream_Free.

libfreerdp/core/window.c
weizhenwei: I took reference on other realloc() call, and rewrote like this:
		size = sizeof(RECTANGLE_16) * windowState->numWindowRects;
		newRect = (RECTANGLE_16*)realloc(windowState->windowRects, size);
		if (!newRect)
		{
			free(windowState->windowRects);
			windowState->windowRects = NULL;
			return FALSE;
		}
		windowState->windowRects = newRect;

channels/drive/client/drive_main.c
```diff
@@ -867,6 +868,7 @@ UINT drive_register_drive_path(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints,
 out_error:
 	MessageQueue_Free(drive->IrpQueue);
 	ListDictionary_Free(drive->files);
+	free(drive->path);
```
akallabeth: This looks broken in context. Should use ```drive_free``` to remove duplication (and possible missing cleanup)

channels/rail/client/rail_main.c
```diff
@@ -92,6 +92,17 @@ UINT rail_send_channel_data(railPlugin* rail, void* data, size_t length)
 }
 
 /**
+ * used by rail_client_execute() to free RAIL_EXEC_ORDER's
+ * internal malloced memory;
+ */
+static void rail_client_clean_exec_order(RAIL_EXEC_ORDER* exec)
+{
+	free(exec->exeOrFile.string);
```
akallabeth: Missing NULL pointer check, set pointers to ```NULL``` afterward.

libfreerdp/core/update.c
```diff
@@ -2041,6 +2041,13 @@ static void update_free_queued_message(void* obj)
 	update_message_queue_free_message(msg);
 }
 
+static void update_free_window_state(WINDOW_STATE_ORDER window_state)
+{
+	free(window_state.titleInfo.string);
```
akallabeth: Would suggest makeing ```window_state``` a pointer and set it to ```NULL``` after free. That way an invalid access will crash the application reliably.

libfreerdp/core/window.c
```diff
@@ -581,6 +597,14 @@ BOOL update_recv_desktop_info_order(rdpUpdate* update, wStream* s, WINDOW_ORDER_
 	return result;
 }
 
+void update_free_window_icon_info(ICON_INFO* iconInfo)
+{
+	free(iconInfo->bitsColor);
```
akallabeth: Missing NULL pointer check. Set freed pointers to NULL afterward

channels/drive/client/drive_main.c
weizhenwei: I will replace free(drive->path) with drive_free(drive)  and test.

libfreerdp/core/window.c
weizhenwei: I will do all theNULL pointer check and set freed pointers to NULL afterward.

channels/rail/client/rail_main.c
```diff
@@ -97,9 +97,26 @@ UINT rail_send_channel_data(railPlugin* rail, void* data, size_t length)
  */
 static void rail_client_clean_exec_order(RAIL_EXEC_ORDER* exec)
 {
-	free(exec->exeOrFile.string);
-	free(exec->workingDir.string);
-	free(exec->arguments.string);
+    if (!exec)
+        return;
+
+    if (exec->exeOrFile.string)
```
mfleisz: NULL checks before free() are useless because free() already does that check

weizhenwei:
@akallabeth , I added all the  NULL pointer check and set freed pointers to NULL afterward, please help to review.

channels/rail/client/rail_main.c
weizhenwei: as suggested, I added these NULL checks. please give some tips if there are better ways to do so.

channels/drive/client/drive_main.c
```diff
@@ -868,7 +868,7 @@ UINT drive_register_drive_path(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints,
 out_error:
 	MessageQueue_Free(drive->IrpQueue);
 	ListDictionary_Free(drive->files);
-	free(drive->path);
+	drive_free(drive);
 	free(drive);
```
akallabeth: double free now (drive is already freed in ```drive_free```)

channels/rail/client/rail_main.c
akallabeth: As @mfleisz mentioned, not necessary to check the pointers to be freed, only the data structure dereferenced.

libfreerdp/core/update.c
```diff
@@ -2041,11 +2041,28 @@ static void update_free_queued_message(void* obj)
...
-static void update_free_window_state(WINDOW_STATE_ORDER window_state)
+static void update_free_window_state(WINDOW_STATE_ORDER* window_state)
 {
-	free(window_state.titleInfo.string);
-	free(window_state.windowRects);
-	free(window_state.visibilityRects);
+    if (!window_state)
+        return;
+
+    if (window_state->titleInfo.string)
```
akallabeth: Same as above.

libfreerdp/core/window.c
```diff
@@ -599,9 +600,24 @@ BOOL update_recv_desktop_info_order(rdpUpdate* update, wStream* s, WINDOW_ORDER_
 
 void update_free_window_icon_info(ICON_INFO* iconInfo)
 {
-	free(iconInfo->bitsColor);
-	free(iconInfo->bitsMask);
-	free(iconInfo->colorTable);
+    if (iconInfo->bitsColor)
```
akallabeth: same as above

channels/drive/client/drive_main.c
weizhenwei: sorry for my stupid fault.

weizhenwei:
@akallabeth , @mfleisz , I removed the uselsess NULL pointer check and rewrote the drive_free() call, please help to check. Thank you for your help and patience.

akallabeth:
@freerdp-bot test

libfreerdp/core/window.c
```diff
@@ -329,9 +335,14 @@ BOOL update_read_window_state_order(wStream* s, WINDOW_ORDER_INFO* orderInfo, WI
 		Stream_Read_UINT16(s, windowState->numVisibilityRects); /* numVisibilityRects (2 bytes) */
 
 		size = sizeof(RECTANGLE_16) * windowState->numVisibilityRects;
-		windowState->visibilityRects = (RECTANGLE_16*) malloc(size);
-		if (!windowState->visibilityRects)
+		newRect = (RECTANGLE_16*)realloc(windowState->visibilityRects, size);
+		if (!newRect)
```
bmiklautz: This can cause a double free if windowState->visibilityRects is 0.

libfreerdp/core/window.c
```diff
@@ -295,9 +296,14 @@ BOOL update_read_window_state_order(wStream* s, WINDOW_ORDER_INFO* orderInfo, WI
 		Stream_Read_UINT16(s, windowState->numWindowRects); /* numWindowRects (2 bytes) */
 
 		size = sizeof(RECTANGLE_16) * windowState->numWindowRects;
-		windowState->windowRects = (RECTANGLE_16*) malloc(size);
-		if (!windowState->windowRects)
+		newRect = (RECTANGLE_16*)realloc(windowState->windowRects, size);
+		if (!newRect)
```
bmiklautz: This can cause a double free if windowState->numWindowRects is 0. (And also breaks rails).

libfreerdp/core/window.c
bmiklautz: @weizhenwei no problem ;) - We are happy about the improvements. Just wanted to mention it that we have it in one place.

Issue #4019, opened 2017-06-27
gst_x_overlay_set_window_handle not yet supported on RHEL 6


scroogie:
In /channels/tsmf/client/gstreamer/tsmf_X11.c line 112 gst_x_overlay_set_window_handle() is called. However, this function is only available since gstreamer 0.10.31. RHEL doesn't provide this. 
Could you maybe consider to call 

gst_x_overlay_set_xwindow_id(hdl->overlay, hdl->subwin);

instead? This makes it compile on RHEL6 for me.

Cheers
André

bmiklautz:
@scroogie no gstreamer 1.0 available either?

scroogie:
Nope, not on RHEL 6, there 0.10.29 is the latest. RHEL 7.x has gstreamer1-* packages, though.

bmiklautz:
@scroogie maybe you can create a pull request that uses the working code path depending on the installed gstreamer version. If < 0.10.31 use your suggestion the current version otherwise.

bmiklautz:
closed with PR #4021

PR #4021, opened 2017-06-27
Fix compilation on RHEL 6.x by macro-guarding two functions

scroogie:
Those two functions were added in gstreamer 0.10.31 (RHEL 6.x being on 0.10.29).

Compare #4019 

I was unsure about the init, though. Should I maybe add a static?


bmiklautz:
@freerdp-bot test.

bmiklautz:
@scroogie first of thanks. 
If I'm not mistaken GST_CHECK_VERSION checks for one specific version. So it would fix the problem only for 0.10.31. I Suggest we change the check to use the "old" functions if the version is smaller or equal to the version you are having troubles with but use the current functions otherwise.

scroogie:
@bmiklautz Nope, GST_CHECK_VERSION is, although the name might suggest otherwise, exactly for this. It checks greater than or equal the passed version:

https://gstreamer.freedesktop.org/data/doc/gstreamer/head/gstreamer/html/gstreamer-GstVersion.html

Greetings

bmiklautz:
@scroogie thanks!

PR #4029, opened 2017-07-04
server/shadow: Fix incorrect bitmap fragment update.

realjiangms:
Legacy bitmap update might fail with 'fast path update size (xxxxx) exceeds the client's maximum request size (xxxxx)'
Original code might update last fragment with exceeded fragment size incorrectly. Fix the logic to prevent it.


akallabeth:
@freerdp-bot test

PR #4032, opened 2017-07-05
Fix clipboard POSIX build because of basename conflict

blino:
FreeRDP/winpr/libwinpr/clipboard/posix.c:397:20: error: conflicting types for ‘basename’
 static const char* basename(const char* name)
                    ^
In file included from FreeRDP/winpr/include/winpr/collections.h:25:0,
                 from FreeRDP/winpr/libwinpr/clipboard/posix.c:37:
/usr/include/string.h:599:14: note: previous declaration of ‘basename’ was here
 extern char *basename (const char *__filename) __THROW __nonnull ((1));


hardening:
@freerdp-bot can you give a build test ?

PR #4033, opened 2017-07-05
Add fullscreen support in Wayland clients

blino:


hardening:
@freerdp-bot can you test that patch that looks good ?

akallabeth:
@freerdp-bot test

PR #4034, opened 2017-07-05
Reuse evdev/X11 keymap for wayland

blino:


hardening:
So if I understand correctly the patch in hybrid environment where there's X and wayland (namely gnome3) your patch allows to find the keymap from the X environment even if we're running a wayland app ?

hardening:

blino:
This is actually for a pure wayland environment, for which there was no keycode conversion at all previously.
This uses the same keymap as X11, which is easy to do since the X11 keycodes are the Linux kernel (evdev) keycodes + 8.

hardening:
That's a long time I've not played with wayland and keyboard, but from what I remember there was the need to look at the mapped xkb file sent to us by the compositor, so that we can send the correct RDP keyboard to the server. But there's no easy way to do that.

I may be wrong, but if your code works for you it may be because: 

* you have the default US keyboard;
* with your changes, we detect the keyboard layout from the X emulation of your compositor, and by chance (or purpose) it's the same layout that the one used by the wayland part. Can you tell what kind of test environment you're using ?

blino:
Actually, in my case, this is with a TV remote control, so the keyboard mapping does not matter that much. But yes, it would default to a US keyboard mapping.
A better way would be to handle the compositor keymap, use the xkbcommon library, and write a xkb -> RDP mapping.

Even if not totally correct, I think this change is still an improvement, since previously the wayland keyboard mapping looked pretty random.

With this change, the wayland client is on par with the X11 client when libxkbfile support is not enabled.

hardening:
So I will test on F25 which has gnome3 with wayland and X to see if it improves things with my french keyboard. 
Basically we would need to know the features (layout, variant ) of the xkb file that is sent to us by the compositor. So that we can find a RDP layout that matches (as much as possible) the xkb file, the problem is that in wayland, we get a mapped file and we have lost all the informations about that file. So using X to know which file we get could be a hack to achieve the final goal (but would not work in environment where there's no X support at all).

hardening:
So I've tested under GNOME3 under wayland (so a mixed environment that also have X), and connecting with wlfreerdp to a windows hosts with my french keyboard works. I will have a test with your patch applied to see if it still works.

akallabeth:
@hardening Any update on this?

PR #4038, opened 2017-07-10
Fixed sample server's wrong rfx pixel format

nfedera:


nfedera:
image

hardening:
Fun I didn't know that it was actually displaying anything.

PR #4044, opened 2017-07-17
Disable h264 if no backend compiled in

akallabeth:
* Disable client side announcement of H264 capability
* Server clears H264 caps on response
* Removed dummy backend (only crashing client/server if used)
* Added compiler warnings/errors for backends with partial/missing implementation (X264, FFMPEG)


akallabeth:
@freerdp-bot test

bmiklautz:
I've tested with cmake 3.0.2, having WITH_FFMPEG=ON but WITH_GFX_H264=OFF is still off.

PR #4047, opened 2017-07-18
Always use software GFX for windows client.

akallabeth:
Allows windows client to work in GFX mode without adding ```/gdi:sw```


bmiklautz:
@akallabeth how come?

bmiklautz:
@akallabeth my suggestion would be to add a warning that gfx doesn't use hw gdi when /gdi:hw is used.

PR #4052, opened 2017-07-20
Mac drive fixes

akallabeth:


channels/rdpdr/client/rdpdr_main.c
```diff
@@ -430,19 +431,25 @@ static UINT handle_hotplug(rdpdrPlugin* rdpdr)
...
+		}
 
 		for (i = 0; i < size; i++)
 		{
-			if (strstr(device_ext->path, dev_array[i].path) != NULL)
+			if (strstr(path, dev_array[i].path) != NULL)
 			{
 				dev_found = TRUE;
 				dev_array[i].to_add = FALSE;
 				break;
```
ilammy: It seems we are leaking `path` here.

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -430,19 +431,25 @@ static UINT handle_hotplug(rdpdrPlugin* rdpdr)
 		if (device_ext->path == NULL)
 			continue;
 
+		ConvertFromUnicode(CP_UTF8, 0, device_ext->path, 0, &path, 0, NULL, FALSE);
```
ilammy: I guess we should check the return value here. If conversion fails, it may not allocate `path` and further calls to strstr() will segfault.

akallabeth:
@bmiklautz @hardening anyone time for review/merge?

PR #4060, opened 2017-07-26
Using ICU instead of custom unicode conversion.

akallabeth:
If WITH_ICU CMake option is set the unicode conversion routines
use ICU instead of custom conversion code.

### DIFF
--- a/winpr/libwinpr/crt/unicode.c
+++ b/winpr/libwinpr/crt/unicode.c
@@ -30,7 +30,15 @@
 
 #ifndef _WIN32
 
+#if defined(WITH_ICU)
+#include <unicode/ucnv.h>
+#include <unicode/ustring.h>
+#else
 #include "utf.h"
+#endif
+
+#include "../log.h"
+#define TAG WINPR_TAG("unicode")
 
 /**
  * Notes on cross-platform Unicode portability:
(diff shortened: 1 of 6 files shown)


ilammy:
Why not replace `ConvertUTF8toUTF16()` and `ConvertUTF16toUTF8()` implementations with ICU usage? It seems to be a more logical plug-and-play replacement. Or your idea is to leverage ICU for legacy-to-Unicode conversions in the future?

winpr/libwinpr/crt/unicode.c
```diff
@@ -30,7 +30,15 @@
 
 #ifndef _WIN32
 
+#if defined(WITH_ICU)
```
ilammy: I think instead of all these ugly #ifdefs we can do it in a cleaner way. Why not provide a separate **unicode_icu.c** file with ICU-based implementations for all the functions from <winpr/string.h>?

akallabeth:
@ilammy Because ```MultiByteToWideChar``` and ``` WideCharToMultiByte```are exported symbols. Of course there could be an improved version of ```ConvertUTF8toUTF16``` and ```ConvertUTF16toUTF8``` but this version was the easiest.
As for the defines, separate files could be done, but it was just a preference not to do it.

akallabeth:
@bmiklautz @mfleisz  anyone time for review/merge?

PR #4063, opened 2017-07-28
Fixed leaks, certificate comparison and channel context cleanup

akallabeth:


ilammy:
I looked through this, seems okay to me.

libfreerdp/core/nla.c
```diff
@@ -1215,7 +1217,7 @@ static BOOL nla_read_ts_credentials(rdpNla* nla, PSecBuffer ts_credentials)
 {
 	wStream* s;
 	int length;
-	int ts_password_creds_length;
+	int ts_password_creds_length = 0;
 	BOOL ret;
```
hardening: IMHO not need as this field is not used (and probably should be...)

libfreerdp/crypto/certificate.c
```diff
@@ -184,7 +184,9 @@ static int certificate_data_match_legacy(rdpCertificateStore* certificate_store,
 					hostname, pline);
 			else if (strcmp(hostname, certificate_data->hostname) == 0)
 			{
-				match = strcmp(pline, certificate_data->fingerprint);
+				const int diff = strcmp(pline, certificate_data->fingerprint);
```
hardening: const not needed

akallabeth:
@hardening @bmiklautz anyone interested in merging this?

hardening:
@akallabeth sorry to bother but could you remove the ```const int``` ?

akallabeth:
@hardening Sorry, don't get it. What is the problem with a const variable?

hardening:
@akallabeth well it's just that I don't see any utility for that, and that it's not the style of FreeRDP code...

akallabeth:
@hardening ```const``` ist not style? The variable is constant, so there is no reason to hide that fact from the compiler...

hardening:
well we've not used any const on scalar type. Ok merging it.

akallabeth:
@hardening thx.

PR #4064, opened 2017-07-28
Fix gstreamer-1.0 detection

floppym:


floppym:
To elaborate on this, before, the code was essentially looking for `gstreamer-1.0 >= (blank)`. This is a malformed dependency that would fail with recent versions of `pkgconf`.

After the change, it looks for `gstreamer-1.0 >= 1.0.5`.

bmiklautz:
@floppym thanks! - I've fixed the same issue but hadn't created the PR :).

bmiklautz:
For reference: the version check was definitely broken but that wasn't a problem for older cmake version. 
I've observed the problem with cmake 3.9

PR #4069, opened 2017-08-01
Build on Solaris

yurashek:
Please adopt the (least) changes to successful build on Solaris.

### DIFF
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -991,6 +991,10 @@ BOOL freerdp_tcp_set_keep_alive_mode(int sockfd)
 	}
 
 #endif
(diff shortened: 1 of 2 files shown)


hardening:
@freerdp-bot can you test that even if we're missing a solaris host...

yurashek:
Tested on Solaris trunk (12.0 formerly). i386 works fine both (client/server). Only the server was tested on SPARC because I do not have a SPARC desktop. Both (client/server) can be built for SPARC but server can not be connected from Win8 -> do not work. I would like to merge this change-set as a "niveau" to be able to build the code seamlessly. Then I will continue in Solaris support improvements. Thank you very much.

PR #4070, opened 2017-08-02
Sun keyboard - cursor keys

yurashek:
May I pleased for review of cursor movements keys fix on Sun keyboard? It looks a bit strange to me I have found only this place which looks like a Japanese keyboard support so the comments are appreciated.
 This patch works fine between Windows and Solaris in both directions and also from Solaris as the client of the VirtualBox "head" and of coarse Solaris to Solaris.


hardening:
@freerdp-bot my friend can you test that for me please ?

bmiklautz:
@yurashek thanks. Looks basically good for me but I'd prefer if you could turn the logic around. Use #ifndef __sun for the default and the else part for __sun.

yurashek:
I prefer to use #if defined(__sun) because the other mutually exclusive branches can be added by #elif defined(i.e. _HPUX) then #else collects "the others" (unspecified) case. Note: #if defined(val.) might be shortened as #ifdef val.

bmiklautz:
@yurashek ok. My main intention was to have the most common code to be on top. Anyway I see your point as well. Let's leave it as is.

+1.

winpr/libwinpr/input/keycode.c
```diff
@@ -450,6 +477,7 @@ DWORD KEYCODE_TO_VKCODE_EVDEV[256] =
 	0, /* <PAST> 143 */
 	0, /* <FIND> 144 */
 	0, /* <CUT> 145 */
+#endif
```
bmiklautz: One more remark. 137-145 are the same if I haven't missed anything. So why not just have 136 in the ifdef/else?

yurashek: These keys: "Stop", "Again", ..., "Paste", "Find", "Cut" are Sun keyboard native (those are placed on the left side below the "Help" key) and I suppose there are no such keys on "PC-104" kbd. It is a "heart issue" to find its re-mapping to the Virtual-Keys in freerdp to be able to map it back in case of the RDP client and also the server running both on Solaris. Currently my knowledge about these mappings in freerdp are very poor so please let me this small place-holder/reminder. :-)

yurashek:
I removed the patch of servicing the Openlook extensions keys. Reason is the support of these 10keys is currently removed also in Gnome 3 in Solaris trunk which replaced the original Gnome 2.2 so I can not test it.

akallabeth:
@freerdp-bot test

yurashek:
I do not understand to this:
 - I asked for commit of changeset with 2 changes
 - bmiklautz asked me to remove 2.nd change which concerned the 10 extra keys on Sun kbd
 - I have found the support of these keys is not available in Solaris trunk so I decided to remove (only the) 2nd change
 - I have pushed my new changeset (consists of the only 1 change) and tested on Solaris (same conditions as before) -> everything is OK
 - Jenkins failed on the change which was yesterday OK to him

Is Jenkins broken? How to restart the Jenkins build?

hardening:
@freerdp-bot come on bot, do a correct test please !

yurashek:
@freerdp-bot please test again my friend but try to set the same conditions as in build# 2201 which was successful

yurashek:
@bmiklautz I have removed the change you have disputed because it is not actual now (support of the Sun keyboard ext. keys is now removed in Gnome 3)

yurashek:
I can not re-run the bot so I merged my repo with origin (pull --rebase FreeRDP/FreeRDP) and pushed back into my clone.
Push always restart the Jenkins to me. I can not restart Jenkins by the message addressed ti him.

yurashek:
@freerdp-bot test

yurashek:
@hardening @akallabeth @bmiklautz maybe I do not have the permissions to run test maybe the daemon has changed. Can anyone help me?

akallabeth:
@freerdp-bot test

yurashek:
@bmiklautz I have removed a piece of code you have disputed so remove your hold please.

bmiklautz:
@yurashek thanks.

PR #4079, opened 2017-08-09
Dynamic channel signal disconnect after the fact.

akallabeth:


akallabeth:
@MartinHaimberger I've refactored the pull, now the call sequence during a channel disconnect is not changed.
What did change though is that the cleanup code is always called now when the channel is disconnected by whatever reason (which was not the case previously)

Issue #4081, opened 2017-08-09
Second monitor not showing FreeRDP


TeruYuuji:
Hi All,
I'm using **FreeRDP version 2.0.0-dev**, from the source code here.
On Lubuntu based on **Ubuntu 16.04.3 LTS**, **kernel 4.8.0-36-generic**.

On the (Windows 2008r2) server I can see it has detected the second monitor, the FreeRDP window does not appear on my second screen.
At the start, I see a flicker on the second screen but 1 second later just the normal (Lubuntu) desktop.

When I try multi monitor or span, with different sub commands like  /multimon:force or with /monitor:0,1 or /span I get the same results.

I have compiled the source code with this command:

> cmake -DCMAKE_BUILD_TYPE=Debug -DWITH_SSE2=ON -DWITH_ALSA=ON -DCHANNEL_URBDRC_CLIENT=ON  -DCHANNEL_URBDRC=ON .

I'm not sure if I have missed a command.

_Before this I had run the FreeRDP version 1.1.0-beta1, which worked with dual monitor.
Only there I don't have the ability to share audio and add a disk to the server from the client.
(And USB redirection was also a reason to go to use the source code. )_

hardening:
IIRC you need to go fullscreen (/f)

TeruYuuji:
Fullscreen with one screen works on both monitors. But when I try /mulitmon or /span it does not. When I had the /f I get same results.

It seems to be connected to Openbox I found another thread where they are discussion the same issue,  I think [#2707](https://github.com/FreeRDP/FreeRDP/issues/2707).

T3N4K4:
@TeruYuuji the source from our company is currently in the devel branch. The master branch is just a clone of FreeRDP-master if that helps.

TeruYuuji:
@T3N4K4 Ahh that helps! Thanks.
I now get a windowed screen, but it is visible on multiple monitors.
Can I set it to always be on top and undecorated it by the cli?

hardening:
@TeruYuuji reading #2707 you have the answer to your bug: TLDR; Openbox is broken regarding modern fullscreen support, anyway there's a way to workaround it but nobody took time to do it. So it's up to you to do a contribution ;)

T3N4K4:
[...]

PR #4087, opened 2017-08-11
Possible fix for #4081 and #2707

T3N4K4:
I tried to create a fix for #4081 and #2707 where _NET_WM_FULLSCREEN_MONITORS gets detected automatically instead of the version before where you have to set +openbox.

I guess there are two approaches. Either search for _NET_WM_FULLSCREEN_MONITORS in _NET_WM_SUPPORTED, or read _NET_WM_FULLSCREEN_MONITORS out directly with xf_GetWindowProperty().

### DIFF
--- a/client/X11/xf_window.c
+++ b/client/X11/xf_window.c
@@ -145,6 +145,9 @@ void xf_SetWindowFullscreen(xfContext* xfc, xfWindow* window, BOOL fullscreen)
 	UINT32 height = window->height;
 	window->decorations = xfc->decorations;
 	xf_SetWindowDecorations(xfc, window->handle, window->decorations);
+	unsigned long nitems, bytes;
+        BYTE* prop;
+        BOOL status;
 
 	if (fullscreen)
 	{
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

client/X11/xf_window.c
```diff
@@ -203,10 +206,16 @@ void xf_SetWindowFullscreen(xfContext* xfc, xfWindow* window, BOOL fullscreen)
...
 	}
 
-	/* Set the fullscreen state */
-	xf_SendClientEvent(xfc, window->handle, xfc->_NET_WM_STATE, 4,
-	                   fullscreen ? _NET_WM_STATE_ADD : _NET_WM_STATE_REMOVE,
-	                   xfc->_NET_WM_STATE_FULLSCREEN, 0, 0);
+	status = xf_GetWindowProperty(xfc, DefaultRootWindow(xfc->display),
+	                              xfc->_NET_WM_FULLSCREEN_MONITORS, 1, &nitems, &bytes, &prop);
+
+	if (status)
```
hardening: Shouldn't we also check the value of the property ?

T3N4K4: Do you mean by checking _NET_WM_FULLSCREEN_MONITORS with XInternAtom() if it is "None" ?

hardening: I was meaning, checking the value of the retrieved property (prop) but perhaps I'm missing something...

T3N4K4: Ah I understand. I will rework it next week on monday or @nfedera could check it.

I was a bit confused.
 _NET_WM_FULLSCREEN_MONITORS is the property.
&prop is the property which gets returned.

akallabeth:
@freerdp-bot test

abma:
this change very likely introduced https://github.com/FreeRDP/FreeRDP/issues/4096 :

fullscreen mode doesn't work any more.

PR #4095, opened 2017-08-16
Fix scrolling crash in wayland client

mmeyer724:
An "axis discrete" handler is required since protocol version 5:
https://code.woboq.org/qt5/include/wayland-client-protocol.h.html#_M/WL_POINTER_AXIS_SOURCE_SINCE_VERSION

Since this was not defined, wlfreerdp would crash with the error message "listener function for opcode 8 of wl_pointer is NULL" if the user attempted to scroll.


hardening:

PR #4097, opened 2017-08-17
fix clang warnings, directly include wtypes.h

bmiklautz:
* Directly include winpr/wtypes.h where _fseeki64 or _ftelli64 is used.
* fix clang 5 warnings.


libfreerdp/codec/bitmap.c
```diff
@@ -435,8 +435,8 @@
 #define TEST_MIX \
 		((last_line == 0 && pixel == mix) || \
 				(last_line != 0 && pixel == (ypixel ^ mix)))
-#define TEST_FOM (TEST_FILL || TEST_MIX)
-#define TEST_COLOR (pixel == last_pixel)
+#define TEST_FOM TEST_FILL || TEST_MIX
```
hardening: I don't see why it's a gain to have all these added parenthesis below...

bmiklautz:
@freerdp-bot test.

libfreerdp/codec/bitmap.c
bmiklautz: In fact it's not really a gain but with -Wparentheses-equality there are some warnings but I think it's worth to have the warnings enabled in general.

bmiklautz:
@freerdp-bot test test test.

PR #4098, opened 2017-08-17
Fixed RSA generation for OpenSSL >= 1.1

### LINKED ISSUE #4094: Shadow server segfaults when starting
I tried freerdp-shadow-cli from the Fedora26 official repository, from my own build and from your Jenkins CI server. They all crash the same:

```
$ /opt/freerdp-nightly/bin/freerdp-shadow-cli
Segmentation fault (core dumped)
```

It happens immediately after start.

akallabeth:
Should also fix #4094


bmiklautz:
@akallabeth nice catch. I really really wonder ...

PR #4099, opened 2017-08-17
wlfreerdp: mirror display resolution in fullscreen mode

mmeyer724:
This removes the need to specify both `/w:1920 /h:1080` and `/f`. If obtaining the resolution fails the default will be used instead.

### DIFF
--- a/client/Wayland/wlfreerdp.c
+++ b/client/Wayland/wlfreerdp.c
@@ -103,6 +103,8 @@ static BOOL wl_pre_connect(freerdp* instance)
 {
 	rdpSettings* settings;
 	wlfContext* context;
+	UwacOutput *output;
+	UwacSize resolution;
 
 	if (!instance)
 		return FALSE;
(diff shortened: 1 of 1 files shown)


hardening:
@freerdp-bot can you test that nice patch please ?

client/Wayland/wlfreerdp.c
```diff
@@ -146,6 +148,21 @@ static BOOL wl_pre_connect(freerdp* instance)
 	PubSub_SubscribeChannelDisconnected(instance->context->pubSub,
 	                                    (pChannelDisconnectedEventHandler) wlf_OnChannelDisconnectedEventHandler);
 
+	if (settings->Fullscreen)
+	{
+		// Use the resolution of the first display output
+		output = UwacDisplayGetOutput(context->display, 1);
```
hardening: shouldn't it be ```UwacDisplayGetOutput(context->display, 0);``` ?

mmeyer724: That doesn't seem to work for me :-/ I think it might be a quirk in UWAC?

mmeyer724: It doesn't error or anything, the resolution just comes back as 0px by 0px. Both of my test machines work great with 1 in that spot (one is Gnome and the other is Weston).

mmeyer724:
@hardening CI failed due to the TestConnect unit-test which I think is flaky. Can you kick off another build for me? 😬

bmiklautz:
@freerdp-bot test.

mmeyer724:
@hardening is this free to be merged?

PR #4100, opened 2017-08-17
wlfreerdp: improve performance by making the window opaque

mmeyer724:
I don't have any numbers.. but this seems to have improved wlfreerdp's performance on both of my test machines.
A similar improvement was made to glmark2-wayland, for example: https://bugs.launchpad.net/glmark2/+bug/1219764


hardening:

mmeyer724:
@hardening I just realized this function uses both `window` and `context->window`. Obviously this isn't a problem because they both point to the same thing, but I'm wondering if you would like me to completely switch it over to one or the other?

bmiklautz:
@freerdp-bot test.

mmeyer724:
@hardening is this free to be merged?

Issue #3918, opened 2017-04-13
Socks proxy implementation


EliotWealth:
https://github.com/FreeRDP/Remmina/issues/196
External defined socket works fine. Socks proxy support should be implemented on client.

EliotWealth:
[07:33:51:707] [12966:12966] [ERROR][com.freerdp.core.transport] - BIO_read returned a system error 11: Resource temporarily unavailable
[07:33:51:707] [12966:12966] [ERROR][com.freerdp.core.nego] - Protocol Security Negotiation Failure
[07:33:51:708] [12966:12966] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED [0x000
2000C]
[07:33:51:708] [12966:12966] [ERROR][com.freerdp.core.connection] - Error: protocol security negotiation or connection failure

EliotWealth:
Socks4 log says it is actually connecting to target RDP:
`SOCK4.1080 00093 - %source_ip%:25856 %dest_ip%:3389 41 19 0 CONNECT_%dest_ip%:3389`

And on target RDP netstat -a says that there is connection from proxy is ESTABLISHED:
`TCP    %rdp_ip%:3389   %proxy_ip%:51372   ESTABLISHED`

But still after some delay about 30sec i think it says that ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED 

I think that some transport or nego internal timeouts can play some role?
Any thoughts anybody?

akallabeth:
#4603 has been merged.

PR #4102, opened 2017-08-19
SOCKS proxy support

yurashek:
Supports SOCKS5 proxy from client directly on tcpip transport level. Originally was motivated by the fact the xfreerdp do not work with socksify on Solaris what is irritating me for a longer time. 
Main use-case is to be used with ssh -D and I also do not have any other SOCKS5 proxy to test on it.
Constraints:  here is no URL-parser in command line nor the credentials cache so the other auth. methods than "NO AUTH" are not implemented in SOCKS5 auth. and it will be the subject of the further effort.
Please review, test and merge.
Thank you,
Jiri

### DIFF
--- a/libfreerdp/core/proxy.c
+++ b/libfreerdp/core/proxy.c
@@ -29,7 +29,8 @@
 #define CRLF "\r\n"
 #define TAG FREERDP_TAG("core.proxy")
 
-BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port);
+static BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port);
+static BOOL socks_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port);
 void proxy_read_environment(rdpSettings* settings, char* envname);
 
 BOOL proxy_prepare(rdpSettings* settings, const char** lpPeerHostname, UINT16* lpPeerPort,
@@ -160,13 +161,16 @@ BOOL proxy_connect(rdpSettings* settings, BIO* bufferedBio, const char* hostname
 		case PROXY_TYPE_HTTP:
 			return http_proxy_connect(bufferedBio, hostname, port);
 
+		case PROXY_TYPE_SOCKS:
+			return socks_proxy_connect(bufferedBio, hostname, port);
+
 		default:
 			WLog_ERR(TAG, "Invalid internal proxy configuration");
 			return FALSE;
 	}
 }
 
-BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
+static BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
 {
 	int status;
 	wStream* s;
@@ -255,3 +259,136 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
 	return TRUE;
 }
 
+/* SOCKS Proxy auth methods by rfc1928 */
+#define AUTH_M_NO_AUTH   0
+#define AUTH_M_GSSAPI    1
+#define AUTH_M_USR_PASS  2
+
+static BOOL socks_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
+{
+	int status;
+	BYTE buf[280], hostnlen = 0xff & strlen(hostname);
+	/* CONN REQ replies in enum. order */
+	static const char *rplstat[] = {
+	  "succeeded",
+	  "general SOCKS server failure",
+	  "connection not allowed by ruleset",
+	  "Network unreachable",
+	  "Host unreachable",
+	  "Connection refused",
+	  "TTL expired",
+	  "Command not supported",
+	  "Address type not supported"
+	};
+
+	/* select auth. method */
+	memset(buf, '\0', sizeof(buf));
+	buf[0] = 5; /* SOCKS version */
+	buf[1] = 1; /* #of methods offered */
+	buf[2] = AUTH_M_NO_AUTH;
+	status = BIO_write(bufferedBio, buf, 3);
+	if (status != 3)
+	{
+		WLog_ERR(TAG, "SOCKS proxy: failed to write AUTH METHOD request");
+		return FALSE;
+	}
+
+	/* Read result SOCK reply */
+	memset(buf, '\0', sizeof(buf));
+	for(;;) {
+	  status = BIO_read(bufferedBio, buf, sizeof(buf));
+	  if (status > 0) break;
+	  else if (status < 0)
+	  {
+	    /* Error? */
+	    if (BIO_should_retry(bufferedBio))
+	    {
+	      USleep(100);
+	      continue;
+	    }
+
+	    WLog_ERR(TAG, "Failed reading AUTH reply from SOCKS proxy (Status %d)", status);
+	    return FALSE;
+	  }
+	  else if (status == 0)
+	  {
+	    /* Error? */
+	    WLog_ERR(TAG, "Failed reading AUTH reply from SOCKS proxy (BIO_read returned zero)");
+	    return FALSE;
+	  }
+	}
+
+	if (buf[0] != 5)
+	{
+	  WLog_ERR(TAG, "SOCKS Proxy version is not 5 (AUTH)");
+	  return FALSE;
+	}
+
+	if (buf[1] != AUTH_M_NO_AUTH)
+	{
+	  WLog_ERR(TAG, "SOCKS Proxy: (NO AUTH) method was not selected by proxy");
+	  return FALSE;
+	}
+
+	/* CONN request */
+	memset(buf, '\0', sizeof(buf));
+	buf[0] = 5; /* SOCKS version */
+	buf[1] = 1; /* command = connect */
+	/* 3rd octet is reserved x00 */
+	buf[3] = 3; /* addr.type = FQDN */
+	buf[4] = hostnlen; /* DST.ADDR */
+	memcpy(buf +5, hostname, hostnlen);
+	/* follows DST.PORT in netw. format */
+	buf[hostnlen +5] = 0xff & (port >> 8);
+	buf[hostnlen +6] = 0xff & port;
+	status = BIO_write(bufferedBio, buf, hostnlen +7);
+	if (status != (hostnlen +7))
+	{
+		WLog_ERR(TAG, "SOCKS proxy: failed to write CONN REQ");
+		return FALSE;
+	}
+
+	/* Read result SOCK reply */
+	memset(buf, '\0', sizeof(buf));
+	for(;;) {
+	  status = BIO_read(bufferedBio, buf, sizeof(buf));
+	  if (status > 0) break;
+	  else if (status < 0)
+	  {
+	    /* Error? */
+	    if (BIO_should_retry(bufferedBio))
+	    {
+	      USleep(100);
+	      continue;
+	    }
+
+	    WLog_ERR(TAG, "Failed reading reply from SOCKS proxy (Status %d)", status);
+	    return FALSE;
+	  }
+	  else if (status == 0)
+	  {
+	    /* Error? */
+	    WLog_ERR(TAG, "Failed reading reply from SOCKS proxy (BIO_read returned zero)");
+	    return FALSE;
+	  }
+	}
+
+	if (buf[0] != 5)
+	{
+	  WLog_ERR(TAG, "SOCKS Proxy version is not 5 (CONN REQ)");
+	  return FALSE;
+	}
+
+	if (buf[1] == 0)
+	{
+	  WLog_INFO(TAG, "Successfully connected to %s:%d", hostname, port);
+	  return TRUE;
+	}
+
+	if (buf[1] > 0 && buf[1] < 9)
+	  WLog_INFO(TAG, "SOCKS Proxy replied: %s", rplstat[buf[1]]);
+	else
+	  WLog_INFO(TAG, "SOCKS Proxy replied: %d status not listed in rfc1928", buf[1]);
+
+	return FALSE;
+}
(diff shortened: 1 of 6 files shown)


yurashek:
@freerdp-bot please test

hardening:
@yurashek @freerdp-bot only respond to his masters, when we ask him to test !

bmiklautz:
@freerdp-bot test.

yurashek:
Could it be possible to add me into the bot's list please?
Thanks,
Jura
PS - I apologize for posting this here but I can not find how to write a PM

Combain:
There is no need to pull proxy support to the core lib since external defined sockets works fine.
#3918

As to code review: this code does not support SOCKS4, SOCKS4a and SOCKS5 with auth.
So you should write that it is supports only SOCKS 5 without any auth.

yurashek:
done

yurashek:
@Combain done
@akallabeth I am using client on Solaris against Win10, Solaris desktop (server) and VirtualBox. I will try it also on Linux whenever the school holidays of my son will ends (he have the Linux desktop). Thanks

ccpp:
@yurashek Attention, you are passing TRUE as the `isHTTPS` parameter to proxy_prepare. This has the effect that the environment variable `$http_proxy` is considered for SOCKS proxy (not tested)

libfreerdp/core/transport.c
```diff
@@ -404,14 +405,25 @@ BOOL transport_connect(rdpTransport* transport, const char* hostname,
 	}
 	else
 	{
-		sockfd = freerdp_tcp_connect(context, settings, hostname, port, timeout);
+		UINT16 peerPort;
+		const char *peerHostname;
+		BOOL isProxyConnection = proxy_prepare(settings, &peerHostname, &peerPort, TRUE);
```
ccpp: Attention, the last argument should be FALSE here

akallabeth:
@yurashek any update on this?

yurashek:
This is really useful feature saving me all of the overhead with doing all the ssh tunnels to lab (mostly the virtual) machines and I currently can not imagine my work without it. Is there a reason why the community do not want to adopt it?

akallabeth:
@yurashek mostly lack of time to test (for me) as there have been many other issues I've had to track down/fix.

libfreerdp/core/proxy.c
```diff
@@ -253,3 +257,136 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
...
+	  "Network unreachable",
+	  "Host unreachable",
+	  "Connection refused",
+	  "TTL expired",
+	  "Command not supported",
+	  "Address type not supported"
+	};
+
+	/* select auth. method */
+	memset(buf, '\0', sizeof(buf));
```
hardening: this memset is not needed

libfreerdp/core/proxy.c
```diff
@@ -253,3 +257,136 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
...
+	if (status != 3)
+	{
+		WLog_ERR(TAG, "SOCKS proxy: failed to write AUTH METHOD request");
+		return FALSE;
+	}
+
+	/* Read result SOCK reply */
+	memset(buf, '\0', sizeof(buf));
+	for(;;) {
+	  status = BIO_read(bufferedBio, buf, sizeof(buf));
```
hardening: I guess you will run into troubles if the response packet is splitted in multiple small packets. So I guess you should loop until you have read a certain amount of bytes.
So in that sense, the memset on line 295 is not needed.

libfreerdp/core/proxy.c
```diff
@@ -253,3 +257,136 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
...
+/* SOCKS Proxy auth methods by rfc1928 */
+#define AUTH_M_NO_AUTH   0
+#define AUTH_M_GSSAPI    1
+#define AUTH_M_USR_PASS  2
+
+static BOOL socks_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
+{
+	int status;
+	BYTE buf[280], hostnlen = 0xff & strlen(hostname);
+	/* CONN REQ replies in enum. order */
```
hardening: usually we would have written `strlen(hostname) & 0xff`

libfreerdp/core/proxy.c
```diff
@@ -253,3 +257,136 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
...
+	status = BIO_write(bufferedBio, buf, hostnlen +7);
+	if (status != (hostnlen +7))
+	{
+		WLog_ERR(TAG, "SOCKS proxy: failed to write CONN REQ");
+		return FALSE;
+	}
+
+	/* Read result SOCK reply */
+	memset(buf, '\0', sizeof(buf));
+	for(;;) {
```
hardening: You should put that reading loop in a separate function. And use it and above.

libfreerdp/core/proxy.c
yurashek: I apologize but I was busy on garden during the weekend.
buf[] is the "automatic variable" where the initial value can not be assumed so for portability reason it is needed. However the memset() here is performed once/per session. What do you think it will "save" when it will not be here?

yurashek: ...or did you meant line 295? memset() can be removed from there but if you debug (might happen) then it is also better to zero a buffer before use. "Costs" ale low but usefulness is high.

libfreerdp/core/proxy.c
yurashek: reply on select auth req. is very short so the packet fragmentation can not happen here. It is also the first received packet so here the data are in sync.

akallabeth: @yurashek Sadly that is up to the implementation, so better never assume anything beyond the indications returned.

libfreerdp/core/proxy.c
yurashek: Why not.

yurashek:
Please help. During my tests I have merged with project master and I have a problem:
[00:14:13:965] [29312:00000002] [ERROR][com.winpr.synch.timer] - InitializeWaitableTimer: os specific implementation is missing

do you know how to switch on the POSIX_TIMERS ? I tried it in my last commit but I was not successful.

libfreerdp/core/transport.c
ccpp: Oops.. that's a bug in my implementation of `proxy_prepare`. Probably `isHTTP` should be handled correctly there! Would you mind adding such an if-statement to the PR?

But, just for the sake of "semantically correct code", why do you pass TRUE then at all, if it changes nothing? Do you explicitly want the environment variables to be picked up? And, what is the initial comment you are referring to?

A decent url parser would be a good idea indeed. It should be centralized for both command line and environment variable. It is not implemented like this now, only because either a public symbol would be needed in libfreerdp, or the rdpSettings have to be changed to pass the unparsed URL instead of type/host/port. I guess, this would be a task for a separate issue / PR :)

yurashek: Call of proxy_prepare() is needed to set the global variable defining the proxy to ensure the isProxy will report TRUE. This is needed to avoid the hostname resolving on local but passing the hostname via the proxy to make hostname resolve on the remote end.

My idea was also to support the ALL_PROXY env.var as curl can do but for socks5 proxy only. Without the URL parser it can make a lot of unwanted effects so it is not in scope of this issue.

My usecase is very simple. As part of my work I am using lab setups mostly on virtual machines including also the Windows VMs. These "boxes" are on non-public IPs (like 192.168...) where I can ssh onto a machine in lab having also the accessible IP. ssh can act as the sock5 anonymous proxy. Anything else is out of scope of this issue.

yurashek:
I have added delta https://github.com/yurashek/FreeRDP/commit/b8e93cb9fd15a29c9cfa050883e02adae06c6e71 with the changes proposed by you. Here is also the merge with 'master' (command line options has moved to cmdline.h)

I have built (on SunOS) and tested. Proxy connection goes OK but working connection will fail on issue #4592 which has nothing to do with this issue. See output:

 [16:39:26:634] [909:00000002] [INFO][com.freerdp.client.common.cmdline] - loading channelEx cliprdr
[16:39:26:646] [909:00000002] [WARN][com.freerdp.core] - setsockopt() IPPROTO_TCP, TCP_KEEPIDLE
[16:39:26:709] [909:00000002] [INFO][com.freerdp.core.proxy] - Successfully connected to w10:3389
[16:39:27:673] [909:00000002] [INFO][com.freerdp.gdi] - Local framebuffer format  PIXEL_FORMAT_BGRX32
[16:39:27:673] [909:00000002] [INFO][com.freerdp.gdi] - Remote framebuffer format PIXEL_FORMAT_RGB16
[16:39:27:863] [909:00000002] [INFO][com.winpr.clipboard] - initialized POSIX local file subsystem
[16:39:27:865] [909:00000002] [ERROR][com.winpr.synch.timer] - InitializeWaitableTimer: os specific implementation is missing

so I assume this issue as fixed and I need a help from someone who is familiar with linux implementation.

hardening:
@yurashek we've not implemented yet the timers for SunOS and so the message `InitializeWaitableTimer: os specific implementation is missing`

yurashek:
Yes I undestand. But I am asking for help how to switch-on WITH_POSIX_TIMER because I tried it here https://github.com/FreeRDP/FreeRDP/pull/4102/commits/41b35539d2c21d1cb19d9c5f76f015a7f37f80a3 + defined -DWITH_POSIX_TIMER in cmake args but both without the success.

libfreerdp/core/proxy.c
```diff
@@ -255,3 +259,117 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
...
+	buf[2] = AUTH_M_NO_AUTH;
+	status = BIO_write(bufferedBio, buf, 3);
+	if (status != 3)
+	{
+		WLog_ERR(TAG, "SOCKS proxy: failed to write AUTH METHOD request");
+		return FALSE;
+	}
+
+	status = recv_socks_reply(bufferedBio, buf, sizeof(buf), "AUTH REQ");
+	if (status < 0) return FALSE;
```
akallabeth: Coding style violation, please run the formatting script.

winpr/libwinpr/synch/synch.h
```diff
@@ -32,6 +32,10 @@
 #define WITH_POSIX_TIMER	1
 #endif
 
+#ifdef __sun
```
akallabeth: Maybe we should move this test to a CMake script checking for the necessary API, so we don't have to do this for every platform.

libfreerdp/core/proxy.c
```diff
@@ -255,3 +259,117 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
...
+	buf[hostnlen +6] = 0xff & port;
+	status = BIO_write(bufferedBio, buf, hostnlen +7);
+	if (status != (hostnlen +7))
+	{
+		WLog_ERR(TAG, "SOCKS proxy: failed to write CONN REQ");
+		return FALSE;
+	}
+
+	status = recv_socks_reply(bufferedBio, buf, sizeof(buf), "CONN REQ");
+	if (status < 0) return FALSE;
```
akallabeth: Missing return value checks, did we really receive the expected number of bytes?

libfreerdp/core/proxy.c
```diff
@@ -255,3 +259,117 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
 	return TRUE;
 }
 
+static int recv_socks_reply(BIO* bufferedBio, BYTE * buf, int len, char* reason)
+{
+	int status;
+
+	ZeroMemory(buf, len);
+	for(;;) {
```
akallabeth: Coding style violation, please run the formatting script.

libfreerdp/core/proxy.c
```diff
@@ -255,3 +259,117 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
...
+	  status = BIO_read(bufferedBio, buf, len);
+	  if (status > 0) break;
```
akallabeth: Coding style violation, please run the formatting script.

libfreerdp/core/transport.c
```diff
@@ -404,14 +405,25 @@ BOOL transport_connect(rdpTransport* transport, const char* hostname,
...
+		else
+		  sockfd = freerdp_tcp_connect(context, settings, hostname, port, timeout);
 
 		if (sockfd < 1)
 			return FALSE;
 
 		if (!transport_attach(transport, sockfd))
 			return FALSE;
 
+		if (isProxyConnection)
```
akallabeth: Coding style violation, please no nested `if` without braces

libfreerdp/core/proxy.c
```diff
@@ -255,3 +259,117 @@ BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
...
+	buf[1] = 1; /* #of methods offered */
+	buf[2] = AUTH_M_NO_AUTH;
+	status = BIO_write(bufferedBio, buf, 3);
+	if (status != 3)
+	{
+		WLog_ERR(TAG, "SOCKS proxy: failed to write AUTH METHOD request");
+		return FALSE;
+	}
+
+	status = recv_socks_reply(bufferedBio, buf, sizeof(buf), "AUTH REQ");
```
akallabeth: Missing length check (return may be smaller than the data touched)

winpr/libwinpr/synch/synch.h
akallabeth: Created #4608 to fix the timer issue, please remove here.

akallabeth:
thank you for your contribution, we added it to #4603

PR #4103, opened 2017-08-21
wlfreerdp: fix double-free race condition

mmeyer724:
This line of code was 'randomly' causing wlfreerdp to crash with a double free or corruption memory error:
https://github.com/FreeRDP/FreeRDP/blob/b2c29158be5114f3b98335baeb844824dd4ce3ab/libfreerdp/codec/region.c#L671

Removing the second call to wl_end_paint in wlfreerdp.c caused this error to go away without noticeable impact on display or memory usage.

### DIFF
--- a/client/Wayland/wlfreerdp.c
+++ b/client/Wayland/wlfreerdp.c
@@ -228,9 +228,6 @@ static BOOL handle_uwac_events(freerdp* instance, UwacDisplay* display)
 				context = (wlfContext*)instance->context;
 				context->waitingFrameDone = FALSE;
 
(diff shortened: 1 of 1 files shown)


hardening:
With a quick look that fix looks wrong, anyway perhaps I missed something.

akallabeth:
@hardening could you have another look, from what I see the call is wrong, but should not be removed but replaced by ```wl_update_content```

akallabeth:
@freerdp-bot test

mmeyer724:
@akallabeth Interesting, I'll have try that out

akallabeth:
Superseeded by #4240

PR #4116, opened 2017-09-03
Fixed BadValue with XCreateWindow

ChickenPeanut:
xf_FixWindowCoordinates occasionally set the dimensions of the window to invalid values (0) because the minimum value check was done at the beginning of the method rather than at the end, which caused it to crash.

Attached sample log from crash
[log.txt](https://github.com/FreeRDP/FreeRDP/files/1273319/log.txt)


ChickenPeanut:
Pardon me, I'm still getting accustomed to git

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

Issue #4117, opened 2017-09-04
Segfault in xf_Bitmap_SetSurface


ondrejholy:
xfreerdp segfaults for me on one Window 7 machine, when Paint application is opened. No special arguments are used when connecting. It is reproducible with master, but even with stable-1.0, so nothing new. Attaching full backtrace and debug log from one of the runs:
[backtrace.txt](https://github.com/FreeRDP/FreeRDP/files/1275275/backtrace.txt)
[log.txt](https://github.com/FreeRDP/FreeRDP/files/1275276/log.txt)

Btw  -offscreen-cache can be used as workaround. 

Can you give me some hint how may I debug it further?

ondrejholy:
(The attached patches just prevent segfaults, but it fails anyway...)

akallabeth:
@ondrejholy Could you add your command line and freerdp version / build configuration?
From a first glance it looks like the requested offscreen surface is missing, maybe you could add some debug logging for ```offscreen_cache_get.put/delete``` and see if the index should be valid.

ondrejholy:
It can be reproduced without any special arguments, i.e. xfreerdp -u USER -p PASS SERVER, respective xfreerdp /u:USER /p:PASS /v:SERVER.

I don't think that version/configuration is interesting in this case, because I can reproduce e.g. with freerdp-2.0.0-33.20170724gitf8c9f43.fc27.x86_64 (Fedora Rawhide), freerdp-1.0.2-10.el7 (RHEL 7), but here is cmake output for my build:
[cmake.txt](https://github.com/FreeRDP/FreeRDP/files/1276257/cmake.txt)

Yes, I am convinced that server is requesting surface which is missing, but I will check it...

ondrejholy:
Yes, it is invalid:
[cache.txt](https://github.com/FreeRDP/FreeRDP/files/1276285/cache.txt)

ondrejholy:
It is a bit more stable now, maybe thanks to the newly attached patches (the Paint app window has to be moved/resized/focused several times in order to trigger this), but still exits unexpectedly. It seems that the problems begin after `Create Offscreen Bitmap Alternate Secondary Drawing Order`, but I don't see any problems in `update_read_create_offscreen_bitmap_order` according to the spec...

Attaching some examples of unexpected exits:
[examples.txt](https://github.com/FreeRDP/FreeRDP/files/1276710/examples.txt)

Any ideas?

akallabeth:
[...]

PR #4121, opened 2017-09-07
Various fixes from #4117

ondrejholy:
Proposing various fixes as a side-effect of debugging https://github.com/FreeRDP/FreeRDP/issues/4117

### DIFF
--- a/client/X11/xf_graphics.c
+++ b/client/X11/xf_graphics.c
@@ -202,6 +202,10 @@ static BOOL xf_Bitmap_SetSurface(rdpContext* context, rdpBitmap* bitmap,
                                  BOOL primary)
 {
 	xfContext* xfc = (xfContext*) context;
+
+	if (!context || (!bitmap && !primary))
+		return FALSE;
+
 	xf_lock_x11(xfc, FALSE);
 
 	if (primary)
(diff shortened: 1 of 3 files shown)


client/X11/xf_graphics.c
```diff
@@ -202,10 +202,16 @@ static BOOL xf_Bitmap_SetSurface(rdpContext* context, rdpBitmap* bitmap,
...
 	xfContext* xfc = (xfContext*) context;
+
+	if (!context)
+		return FALSE;
+
 	xf_lock_x11(xfc, FALSE);
 
 	if (primary)
 		xfc->drawing = xfc->primary;
+	else if (!bitmap)
```
hardening: this check must be done some line above, or you leave the function with X11 locked. 
I'd be curious to see when that situation happens (I mean perhaps the check is to be done by the callers ?)

libfreerdp/core/orders.c
```diff
@@ -2460,7 +2460,7 @@ static BOOL update_read_create_offscreen_bitmap_order(wStream* s,
 		if (deleteList->cIndices > deleteList->sIndices)
 		{
 			UINT16* new_indices;
-			new_indices = (UINT16*)realloc(deleteList->indices, deleteList->sIndices * 2);
+			new_indices = (UINT16*)realloc(deleteList->indices, deleteList->cIndices * 2);
```
hardening: good catch !

client/X11/xf_graphics.c
ondrejholy: Dammit, sorry, updated. Will be also checked by callers thanks to the following patch...

ondrejholy:
@hardening Is anything expected from my side?

PR #4129, opened 2017-09-12
Add description for available plugins

ondrejholy:
I am not sure you are interested in stable-1.0 patches, but I am going to use this downstream, so I think it is fair to propose it here first...

The man pages lack info about available plugins and its functionality. Add description for available plugins and their parameters.

It is based on info from upstream wiki pages:
https://github.com/FreeRDP/FreeRDP/wiki/Plugins
https://github.com/FreeRDP/FreeRDP/wiki/CommandLineInterface

### DIFF
--- a/client/X11/xfreerdp.1.xml
+++ b/client/X11/xfreerdp.1.xml
@@ -438,10 +438,142 @@
         </listitem>
       </varlistentry>
       <varlistentry>
-        <term>--plugin <replaceable class="parameter">pluginname</replaceable></term>
+        <term>--plugin <replaceable class="parameter">plugin</replaceable> [--data <replaceable class="parameter">data</replaceable> --]</term>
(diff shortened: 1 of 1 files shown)


bmiklautz:
@ondrejholy thanks for the PR. We won't do much changes to 1.0/1.1 as we really trying to get to 2.0. However as it is "only" documentation I'll merge it. If you got time would be nice to have it for master as well ;).

ondrejholy:
@bmiklautz Great, thanks. Master has a listing of the plugins thanks to options like /smartcard, /drive, /printer and has also some examples, in contrary to stable-1.0 where the available plugins were not listed at all... but yes it would be nice to add more details. I will try to propose some improvements, but I don't promise anything because the wiki doesn't contain much info about the new syntax...

Issue #4134, opened 2017-09-15
xfreerdp doesn't work correctly with /bpp:32 on X11 with 32 depth 


ondrejholy:
Wallpaper/background color is not shown (regardless ` /wallpaper`) and the window is transparent instead, there are some other glitches also, see toolbar in the attachment:

![screenshot](https://user-images.githubusercontent.com/1051384/30482339-e74c30d2-9a22-11e7-9ede-feff626f55a5.png)

```
[15:43:05:916] [32126:32127] [INFO][com.freerdp.gdi] - Local framebuffer format  PIXEL_FORMAT_RGBA32
[15:43:05:916] [32126:32127] [INFO][com.freerdp.gdi] - Remote framebuffer format PIXEL_FORMAT_RGBA32
```

Tested on Fedora rawhide (with X11 `DefaultDepth 32`) with system version and freerdp master. Just `xfreerdp /u:... /p:... /v:... /bpp:32 `, nothing special (/gdi:sw/hw is same). Tested against Windows 7 server.

akallabeth:
@ondrejholy oh, real transparency :) Does this issue disappear when using ```PIXEL_FORMAT_RGBX32``` as local framebuffer format?

hardening:
Yeah, I guess most of us would not complain about full transparency working so well :)

ondrejholy:
@akallabeth If I recall correctly, it didn't help, but I will try again for sure...

ondrejholy:
Nope, it doesn't help...
```
[09:13:54:329] [8773:8774] [INFO][com.freerdp.gdi] - Local framebuffer format  PIXEL_FORMAT_BGRX32
[09:13:54:329] [8773:8774] [INFO][com.freerdp.gdi] - Remote framebuffer format PIXEL_FORMAT_BGRA32
```

akallabeth:
@ondrejholy Sorry for the delay. Wanted to ask if you could also try other formats (```PIXEL_FORMAT_ARGB32```, ```PIXEL_FORMAT_XRGB32``` et al)

ondrejholy:
[...]

PR #4135, opened 2017-09-15
Fix colors on big endian

ondrejholy:
Let's fix the colors on big endian finally https://github.com/FreeRDP/FreeRDP/issues/2520. See commit descriptions...

Tested all combinations of 16/24/32 depth of X11 and /bpp:16/24/32 on systems with little/big endian. It seems it is working correctly, just /bpp:32 doesn't work correctly on X11 with 32 depth, but this is a problem even with master, see https://github.com/FreeRDP/FreeRDP/issues/4134. I did not test with /bpp:8 and lower systematically, but it seems that there are still some minor issues...

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -381,6 +381,9 @@ static BOOL xf_sw_desktop_resize(rdpContext* context)
 		goto out;
 	}
 
+	xfc->image->byte_order = LSBFirst;
+	xfc->image->bitmap_bit_order = LSBFirst;
+
 	ret = xf_desktop_resize(context);
 out:
 	xf_unlock_x11(xfc, TRUE);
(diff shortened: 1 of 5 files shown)


ondrejholy:
@akallabeth To be honest, I've never tried that options, but I will try them...

ondrejholy:
I tried xfreerdp /u:... /p:... /v:... /rfx and xfreerdp /u:... /p:... /v:... /gfx:RFX (isn't this the same as /rfx?) on big endian with 16/24/32 depth of X11 and it seems working correctly. I did not try AVC420|AVC444, because I don't see any available h264 codec in s360x repositories...

hardening:
@akallabeth feel free to merge it if it seems ok for you

akallabeth:
@ondrejholy No, ```/rfx``` uses old style transport, ```/gfx``` uses the newer gfx channel transport. Codec wise they are (mostly) the same though. As for H264 options, openh264 works quite fine and for a test a custom compile should suffice (although I assume it is working too, if any GFX option is working)

akallabeth:
@ondrejholy Ok, looks good, I'll merge it. Just another test I forgot to mention: did you try ```/gdi:sw``` too?

ondrejholy:
@akallabeth Great, thanks! Yes, I forgot to mention that I tested everything also with /gdi:sw/hw.

PR #4136, opened 2017-09-18
Fixed endless loop when RDP Server sends SERVER_DENIED_CONNECTION

tditsch:


libfreerdp/core/transport.c
```diff
@@ -1053,6 +1053,10 @@ int transport_check_fds(rdpTransport* transport)
 		transport->haveMoreBytesToRead = TRUE;
 	}
 
+  if (freerdp_shall_disconnect(transport->context->instance)) {
```
hardening: I'd suggest to move that statement in the while loop above and remove the corresponding test in the while consition (so that the test is only done once)

bmiklautz:
@freerdp-bot test.

akallabeth:
@hardening ready for merge?

PR #4138, opened 2017-09-19
Check and invalidate handles on free.

akallabeth:


akallabeth:
@freerdp-bot test please

PR #4150, opened 2017-09-25
Unlink file after binding to it.

akallabeth:
When unlinking the file before binding, a new entry is created
in the file system after binding. This is not desireable, so
unlink it after binding to remove the temporary file after the process
closes.

### DIFF
--- a/libfreerdp/core/listener.c
+++ b/libfreerdp/core/listener.c
@@ -185,9 +185,9 @@ static BOOL freerdp_listener_open_local(freerdp_listener* instance, const char* 
 
 	addr.sun_family = AF_UNIX;
 	strncpy(addr.sun_path, path, sizeof(addr.sun_path));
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

hardening:
@akallabeth does it mean that the file will exist only when the fd is opened ?

akallabeth:
@hardening yes

kakaroto:
Sorry, but this was wrong and it actually creates a bug. The listener creates the socket file when you call 'bind'. If the file already exists, the bind will fail with "address already in use", which is why the unlink was there before the bind. By moving the unlink after the bind, you are basically deleting the socket right after its creation. This prevents anyone from connecting to the server listener socket ("file not found" errors), rendering the server/listener unusable.
I've fixed it in PR #4201.

PR #4154, opened 2017-09-26
Fix raw surfaces displaying + misc other changes

hardening:


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

hardening:
youhou, after tons of failed tests, it finally works !

bmiklautz:
@hardening what did you use for re-formatting? - astyle does ugly thing - right but it at least enforces some kind of standard ;)

hardening:
Well I think we've accepted ugly things introduced by astyle too quickly. This was a rollback attempt. We all agreed that astyle's output with 80 char / line limitation doesn't provide any gain in terms of readability.

akallabeth:
@hardening It does, think 3-way diff ;)

bmiklautz:
@hardening @akallabeth I'd suggest we sort out the formatting and what we really want once and for all including a script/config that needs to be used and format all at one shot (as in pull request).

bmiklautz:
If the last commit in this PR is just formatting you maybe want to remove it and we do it as suggested in a different one.

hardening:
he he my git skills are improving, I managed to do the commit drop !

bmiklautz:
@hardening nice \o/. Thanks.

PR #4155, opened 2017-09-26
TestConnect: Extend timeout, only listen locally

akallabeth:


akallabeth:
@hardening Looks like a too tight timeout check was at fault

PR #4156, opened 2017-09-26
Added delay for connect abort

akallabeth:
```freerdp_connect``` resets the abort event. This opens a race window for ```TestConnect```. Fix it by waiting until at least freerdp_connect is called.


akallabeth:
@hardening Yay, another race :/

PR #4162, opened 2017-10-02
Misc fix and improvements for the DR channel

hardening:


channels/drive/client/drive_main.c
```diff
@@ -592,13 +614,16 @@ static UINT drive_process_irp_query_directory(DRIVE_DEVICE* drive, IRP* irp)
 	if (!drive || !irp || !irp->Complete)
 		return ERROR_INVALID_PARAMETER;
 
+	if (Stream_GetRemainingLength(irp->input) < 28)
```
bmiklautz: Confused here: UINT32+UINT8+UINT32+23*UINT8 isn't that 32?

channels/drive/client/drive_main.c
```diff
@@ -274,10 +274,13 @@ static UINT drive_process_irp_read(DRIVE_DEVICE* drive, IRP* irp)
 	if (!drive || !irp || !irp->output || !irp->Complete)
 		return ERROR_INVALID_PARAMETER;
 
+	if (Stream_GetRemainingLength(irp->input) < 6)
```
bmiklautz: I think that should be < 12

channels/drive/client/drive_main.c
```diff
@@ -328,11 +331,14 @@ static UINT drive_process_irp_write(DRIVE_DEVICE* drive, IRP* irp)
 	if (!drive || !irp || !irp->input || !irp->output || !irp->Complete)
 		return ERROR_INVALID_PARAMETER;
 
+	if (Stream_GetRemainingLength(irp->input) < 26)
```
bmiklautz: Shouldn't that be < 32?

channels/drive/client/drive_main.c
hardening: yes of course !

channels/drive/client/drive_main.c
hardening: yep

channels/drive/client/drive_main.c
hardening: You're right again

hardening:
So I've fixed all my wrong computations (looks like late patch aren't a so nice idea)

channels/drive/client/drive_main.c
akallabeth: @hardening Sure you don't want to change that too?

hardening:
@bmiklautz @akallabeth I think I've fixed everything

bmiklautz:
@giox069 could you re-test the current pull request if it's working for you?

bmiklautz:
@hardening nice work!

giox069:
I just tested it with the same couple of win2003 servers and a win10 server: I copied some files from the remote desktop to \\tsclient\shareddir without problems: it worked with no crashes and no file corruptions.

bmiklautz:
@giox069 thanks.

PR #4163, opened 2017-10-05
winpr: Convert time zone generator programs to C# scripts

rjcorrig:
Converts the C# programs (.cs modules) used to generate the time zone structures in winpr/libwinpr/timezone/timezone.c to C# scripts (.csx files). This change allows the files to be run directly using the C# interpreter (csi.exe) without having to stand up an MSBuild or Visual Studio project.

As a convenience to the developer running the script, the resulting output .txt files are .gitignored.

As an added convenience, the WindowsZones.csx script downloads the CLDR timezone data directly from unicode.org instead of relying on a previously downloaded windowsZones.txt file.


akallabeth:
@freerdp-bot test

akallabeth:
Looks like the scripts do what they are supposed to do.
@rjcorrig One question though, ```csi.exe``` is available with which versions of visual studio?

mfleisz:
@akallabeth Should be available since Visual Studio 2015 Update 1

akallabeth:
@bmiklautz @hardening Any objections to merging?

PR #4164, opened 2017-10-05
Pause playback on queue empty, free buffers on close.

akallabeth:
Fixes an issue on mac os that audio stops playing after closing the first time. (open browser, play something, close, reopen -> no more sound)


mgibson2:
@mfleisz 
@akallabeth 
Can you both please take care to get this into tcvier ASAP and make a new dmg for testing.
But please first test that this fix does not brake anything on Linux and Windows

hardening:
@mgibson2 the changes are in some mac specific code, so no need to worry.

akallabeth:
@mgibson2 The changes are in files only compiled when the target is iOS or MacOS, so no chance it breaks something on other platforms.

mgibson2:
@akallabeth 
OK, thx

PR #4172, opened 2017-10-10
egfx: fix disconnection caused by invalid cache entries due to wrong announced cache size

hardening:
Added some checks so that when setting a cache entry fails, we close connection (or
we fail later when trying to use that empty entry).
The small cache egfx capability has also been fixed.

### DIFF
--- a/channels/rdpgfx/client/rdpgfx_main.c
+++ b/channels/rdpgfx/client/rdpgfx_main.c
@@ -74,7 +74,10 @@ static UINT rdpgfx_send_caps_advertise_pdu(RDPGFX_CHANNEL_CALLBACK* callback)
 	if (gfx->ThinClient)
 		capsSet->flags |= RDPGFX_CAPS_FLAG_THINCLIENT;
 
-	if (gfx->SmallCache)
+	/* in CAPVERSION_8 the spec says that we should not have both
+	 * thinclient and smallcache (and thinclient implies a small cache)
+	 */
+	if (gfx->SmallCache && !gfx->ThinClient)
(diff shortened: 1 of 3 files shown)


channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -1716,11 +1724,8 @@ UINT DVCPluginEntry(IDRDYNVC_ENTRY_POINTS* pEntryPoints)
 		if (gfx->H264)
 			gfx->SmallCache = TRUE;
 
-		if (gfx->SmallCache)
-			gfx->ThinClient = FALSE;
-
-		gfx->MaxCacheSlot = (gfx->ThinClient) ? 4096 : 25600;
-		context = (RdpgfxClientContext*) calloc(1, sizeof(RdpgfxClientContext));
+		gfx->MaxCacheSlot = gfx->SmallCache ? 4096 : 25600;
```
akallabeth: Your comment above implies that the cache is also small if RDPGFX_CAPS_FLAG_THINCLIENT is set, correct?

hardening: Yes, that's what the spec says. thinclient implies the use of remoteFx (instead of progressive) and the cache to fit in 16Mb (so a small cache).

akallabeth: Just asking, because here you only check ```gfx->SmallCache``` and not ```gfx->ThinClient```

hardening:
Basically this PR fixes some missing checks when setting cache entries. And it also fixes the announced cache size: we had some case where the server was considering that we had a big cache while it was a small one (and so it was setting entries out-of-bounds)

PR #4177, opened 2017-10-12
Android: Fix tab key mapping for hardware keyboard

### LINKED ISSUE #2486: Android client maybe not passing Tab key correctly?
I'm working on getting aFreeRDP running for a Motorola MC32 device as part of a client hardware evaluation for our internal applications delivered over RDS. 
This device has an integrated barcode imager that sends keyboard input events to the active Activity. 
Our standard is to set the imager to send a Tab key after the decoded barcode to advance the cursor in the remote application to the next field. 
However, it doesn't seem to work consistently, because the Android client is misinterpreting the Tab key.

I found that the KeyboardMapper class in the Android FreeRDPCore project was receiving a KEYCODE_TAB KeyEvent in processAndroidKeyEvent. 
When it takes a trip into getVirtualKeyCode, it finds no vkcode for the Tab code, and falls through the rest of the if-else cases until it sends the key as a UnicodeKey with value 9. 

Of the applications I tested on our server, only Notepad would correctly interpret the tab. Other applications didn't receive the event at all. 

When I added a mapping from KEYCODE_TAB to VK_TAB in the keymapAndroid array, all remote applications behave correctly. 
It looks like KeyboardMapper is expecting the Tab key to come in as a custom or extended key, but that's not happening on my hardware. 
Is it acceptable to add the KEYCODE_TAB -> VK_TAB in keymapAndroid? 

Details on setup:
aFreeRDP Client from git master, March 15 2015
RDS Server version: Windows 2012 R2
Android Version: 4.1.1

rjcorrig:
Some handheld computers, such as the Motorola MC32 series, have a hardware keyboard with a Tab key. 

If the KEYCODE_TAB -> VK_TAB mapping is not present, the key is passed to the server as a Unicode key down event with no matching key up event. It appears most Windows programs (with the notable exception of Notepad) are looking for a keydown/keyup sequence before processing the Tab keypress. 

Adding the mapping to the keymapAndroid array, as done here, causes the Tab to be sent as a regular pair of keydown/keyup keyboard events.

fixes #2486

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/utils/KeyboardMapper.java
+++ b/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/utils/KeyboardMapper.java
@@ -271,6 +271,7 @@ public class KeyboardMapper {
         keymapAndroid[KeyEvent.KEYCODE_DEL] = VK_BACK;
         keymapAndroid[KeyEvent.KEYCODE_ENTER] = VK_RETURN;
         keymapAndroid[KeyEvent.KEYCODE_SPACE] = VK_SPACE;
+        keymapAndroid[KeyEvent.KEYCODE_TAB] = VK_TAB;
 //		keymapAndroid[KeyEvent.KEYCODE_SHIFT_LEFT] = VK_LSHIFT;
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

Issue #3439, opened 2016-07-07
multiple issues with fullscreen mode on two monitors


nshustov:
xfreerdp /multipon /u:&lt;skipped&gt; /v:&lt;skipped&gt;

Ubuntu 16.04 LTS (Kubuntu) x64
2 monitors

Both in  the latest stable and nightly builds of freerdp:

freerdp-nightly          2.0.0+0~20160707014830.282~1.gbp89c252       amd64
freerdp-x11              2.0.0~dev~git20160331+dfsg-0xenial3                   amd64
1. Client does not span on multiple monitors in full screen mode - it just sticks to one monitor. 
   It behaves this way with both same resolution monitors and different resolution monitors. **It used to work some time ago (sorry, not sure what package upgrade broke it).**
2. When switched to windowed mode, an attempt to move the window makes it minimized and there is no way to restore the window/maximize it.
3. After window went minimized as described in the above step (and has active input focus) attempting to go full screen mode (via keyboard shortcut) crashes freerdp with the following diagnostics:

(nightly build):

X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  12 (X_ConfigureWindow)
  Value in failed request:  0x0
  Serial number of failed request:  24794
  Current serial number in output stream:  24797

(stable):

X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  12 (X_ConfigureWindow)
  Value in failed request:  0x0
  Serial number of failed request:  61016
  Current serial number in output stream:  61019

vadimseyn:
I have the same problem as you on OpenSUSE 42.1

FragsterAt:
same in freerdp-2.0.0~git.1463131968.4e66df7 precompiled for opensuse 42.1
on server side (win 10m win server 2012 etc) there are 2 monitors, but freerdp window show only "primary"

T3N4K4:
Have you tried the parameters "/span" or "/multimon:force"?
Or the combination of both "/f /multimon /span"?
Maybe it even helps to give the actual size of both monitors together to "/size:<width>x<height>".

And I don't know if it is relevant, but it sometimes worked for me to explicitly enable tls "+sec-tls".

FragsterAt:
with /span remote system thinks that I have a big monitor,  but freerdp displays one full screen window on one of the monitors and a half of "big" screen
Other combinations, including +sec-tls, /monitors:0,1 etc, do not work

[...]

PR #4186, opened 2017-10-17
FreeRDP multimonitor: Use first command line element, then primary, …

T3N4K4:
...then do a fallback to 0,0.

I tried to address the issue of #3439 and some more of the multimonitor / primary monitor problems for myself.
The FreeRDP code makes always a fallback to 0,0 since the primary monitor is always not found.

My fix may be still work in progress and I don't know if it is as FreeRDP intended this. But I am open for advice how to do it properly.

Currently I look up what the user specified on the command line with /monitors:x,y... and take x as the default monitor (primary). If no command line option was given I use the default monitor which I get with
```
screen = DefaultScreenOfDisplay(xfc->display);
monitor_index = XScreenNumberOfScreen(screen);
```
and last we try to make a fallback to 0,0 if still no primary monitor was found (I heard that it is a thing in newer *nix versions that there is no primary monitor set).

In my tests on ubuntu 15.10 / xfce4:
/monitors:0, /monitors:1, /monitors:1,0 /multimon:force (our server does not advertise multimonitor), /monitors:0,1 /multimon:force or /multimon:force only - did work.

### DIFF
--- a/client/X11/xf_monitor.c
+++ b/client/X11/xf_monitor.c
@@ -115,12 +115,14 @@ BOOL xf_detect_monitors(xfContext* xfc, UINT32* pMaxWidth, UINT32* pMaxHeight)
 {
 	int i;
 	int nmonitors = 0;
+	int monitor_index = 0;
 	BOOL primaryMonitorFound = FALSE;
 	VIRTUAL_SCREEN* vscreen;
 	rdpSettings* settings = xfc->context.settings;
 	int mouse_x, mouse_y, _dummy_i;
 	Window _dummy_w;
 	int current_monitor = 0;
+	Screen* screen;
 #ifdef WITH_XINERAMA
 	int major, minor;
 	XineramaScreenInfo* screenInfo = NULL;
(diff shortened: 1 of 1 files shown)


akallabeth:
@T3N4K4 looks like #4187 fixes the same issue? Sorry to have looked at that one first, but it was the highest in my listing.

T3N4K4:
@akallabeth no problem, I submitted mine first, reworked the history and saw later that @hardening was on the same issue with an own commit.
My fix uses the "default"-monitor given back from the xinerama output though instead of only using /monitors:x1,x2... before doing a fallback to 0,0. I don't know though if my fix is best practice ;-)

akallabeth:
@T3N4K4 Ok, I've merged #4187 first, as both your and @hardening s pull address two slightly different issues. Could you rebase to current master, as the last does have some conflicting changes? Sorry for the comment in the closed pull, did manage to overlook this one at my first search :/

T3N4K4:
Sure I'll work on it as soon as I can :-)

T3N4K4:
@akallabeth it could take a while ... unfortunately rebasing or merging hates me :-(

akallabeth:
@T3N4K4 I know that feeling, take your time

T3N4K4:
@akallabeth I hope I fixed it now.

akallabeth:
@T3N4K4 Any update on this?

T3N4K4:
@akallabeth no sorry, I haven't had the time to check your mentioned errors .. :-(, and I have to read a bit more on sanitizing arguments ;)

T3N4K4:
@akallabeth just a question. Why strtol() ? In the cmdline.c file the MonitorIds are set with atoi().

akallabeth:
@T3N4K4 atoi can not properly report if the argument is an integer or not. Should be replaced everywhere.

akallabeth:
@T3N4K4 Have a look at #4237 
I've removed ```atoi``` in that pull completely.

akallabeth:
@freerdp-bot test

akallabeth:
@mfleisz @hardening Anyone time for a review of this one?

hardening:
It's not clear for me what this PR is trying to solve, can you elaborate @T3N4K4 ?

T3N4K4:
@hardening as said in my first post.
FreeRDP (before your PR) used the 0,0 point to look for the first monitor. But if you wanted to start it on the primary monitor there was no option for it.
/monitors did not work either.

In my PR we look if the user used the /monitors command line, then if not given use the primary monitor (which can be anything else) and as a fallback use the 0,0 point.

So if your primary monitor has 1920x1080+1920+0 and your second 1920x1080+0+0 always the second one was used if /monitors was not entered. In my opinion the primary monitor should be used if no /monitors are entered.

mfleisz:
@T3N4K4 
What I don't understand is the need for the for-loop if the "default monitor" is not originated at 0/0. As far as I understand the "default monitor" id is either passed using the /monitors argument or is determined using XScreenNumberOfScreen() to get the id of the primary monitor (in case /monitors is absent). In both cases we want to use that monitor - no matter if it is at 0/0 or not?

T3N4K4:
@mfleisz true. But I happen to stumble upon some test-machines which did not have a primary screen. At least not one which was set and could be read with tools like xrandr. I did not try this fix on those machines but I remembered this state and guessed that a fallback is better. I don't know in which state we would land if /monitors is not set and XScreenNumberOfScreen fails (or sth. like that).

But we can change it sure :-)

mfleisz:
@T3N4K4 
Ah ok - didn't know that the API might return something stup ... I'm fine leaving it as is for now ...

PR #4187, opened 2017-10-17
Multimonitor fix

hardening:
Using the /monitor: option was not working as expected.


akallabeth:
@hardening Could you check #4186 (which addresses the same issue) but does a few details differently? Would be nice to know which one to prefer ;)

hardening:
@akallabeth why choosing when you need both ? I think the 2 PR are needed, they just fix different issues. Mine fixes the setting of the primary monitor when using /monitors. While  #4186 is usefull when you have _not_ used /monitors.

hardening:
anyone to merge this one ?

PR #4189, opened 2017-10-17
codec/h264: fix potential segfault and mf_init

nfedera:


bmiklautz:
@nfedera looks like MFCreateDXGIDeviceManager isn't used at all.

nfedera:
@bmiklautz I've noticed that too. Will ask around to be sure before nuking it.

PR #4197, opened 2017-10-22
X11 Client: fix UTF8 window title

dreamadv:


client/X11/xf_window.c
```diff
@@ -451,7 +451,12 @@ xfWindow* xf_CreateDesktopWindow(xfContext* xfc, char* name, int width,
 		            settings->DesktopPosY);
 	}
 
+	const size_t i = strlen(name);
```
akallabeth: Why not simply use ```xf_SetWindowText```? That function already does this properly.

dreamadv: xf_SetWindowText never call from other places.
I tried to use it by call from xf_CreateDesktopWindow, but the window var has different types (xfWindow and xfAppWindow)

akallabeth: @dreamadv Ah, ok, you're right. Then I'd suggest the following:
Add a static function in ```xf_window.c``` with the function signature ```(xfContext* xfc, Window window, const char* name)``` that can be used by both, therefore eliminating the duplicate name setting (and possible inconsistences after future changes)

akallabeth:
@dreamadv I've reworked this pull, therefore closing yours. Thanks for pointing out this issue.

PR #4201, opened 2017-10-27
Do not delete the listener socket right after creating it.

kakaroto:
The listener server socket file needs to be deleted before we bind it
otherwise it's an "address already in use" error. But it was getting
deleted after the bind, causing the file to disappear, and preventing
anyone from connecting to the listener socket since the socket stops
existing.

This is caused by commit 884e87fde4ece132c3b7bdf947de044ecb3d4b14
The bug was introduced via https://github.com/FreeRDP/FreeRDP/pull/4150

### DIFF
--- a/libfreerdp/core/listener.c
+++ b/libfreerdp/core/listener.c
@@ -186,8 +186,8 @@ static BOOL freerdp_listener_open_local(freerdp_listener* instance, const char* 
 	addr.sun_family = AF_UNIX;
 	strncpy(addr.sun_path, path, sizeof(addr.sun_path));
 
-	status = _bind(sockfd, (struct sockaddr*) &addr, sizeof(addr));
 	unlink(path);
+	status = _bind(sockfd, (struct sockaddr*) &addr, sizeof(addr));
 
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test.

akallabeth:
Will need to find another way to clean up files created by shadow server when closing.

kakaroto:
Thanks for the merge.
I think cleaning up the files should be as simple as doing the unlink right after calling closesocket()

akallabeth:
@kakaroto Yes, but for that the filename is necessary. Don't want to modify the listener structs for that, but it looks like there is no descriptor to filename conversion available :/

kakaroto:
well, it can be done within that same function for all the closesocket error cases that happen after the bind. As for the regular closesocket, I think it's possible to use getsockname to get the path. That's what the man page says anyways, but I haven't tested this. So on closing it, you could do getsockname, check if it's AF_UNIX and unlink the path return in the sockaddr struct. This way, you don't have to keep a copy of the filename.

PR #4203, opened 2017-10-30
winpr: _IoCreateDeviceEx: fix mkdir error check

floppym:
The mkdir(2) function returns 0 on success, and -1 on error.

This resolves an error in TestIoDevice when /tmp/.device/ does not
exist.

Bug: https://bugs.gentoo.org/635838

### DIFF
--- a/winpr/libwinpr/io/device.c
+++ b/winpr/libwinpr/io/device.c
@@ -138,7 +138,7 @@ NTSTATUS _IoCreateDeviceEx(PDRIVER_OBJECT_EX DriverObject, ULONG DeviceExtension
 
 	if (!PathFileExistsA(DeviceBasePath))
 	{
-		if (!mkdir(DeviceBasePath, S_IRUSR | S_IWUSR | S_IXUSR))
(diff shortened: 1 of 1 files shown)


winpr/libwinpr/io/device.c
```diff
@@ -138,7 +138,7 @@ NTSTATUS _IoCreateDeviceEx(PDRIVER_OBJECT_EX DriverObject, ULONG DeviceExtension
 
 	if (!PathFileExistsA(DeviceBasePath))
 	{
-		if (!mkdir(DeviceBasePath, S_IRUSR | S_IWUSR | S_IXUSR))
+		if (mkdir(DeviceBasePath, S_IRUSR | S_IWUSR | S_IXUSR))
```
akallabeth: Maybe a good idea to write something like ```if (mkdir(DeviceBasePath, S_IRUSR | S_IWUSR | S_IXUSR) != 0)``` to indicate the integer nature of the return

hardening: +1

floppym: I force-pushed a revised commit.

PR #4205, opened 2017-10-31
Fixed #4199: ConvertFindDataAToW string length

### LINKED ISSUE #4199: Length of filename or path problem
Hi there,

I mount a local folder in Linux remotely to a Windows 7 desktop. And I notice that the length of a file in the folder has a limit of 130 characters on the remote Windows 7 desktop. It is the same whether the file is put into sub-folder or not. 

So my question is "Is this designed on purpose?"

When I use a very long file name for the testing file, such as `abcdefghijklmnopqrstuvwxyzAabcdefghijklmnopqrstuvwxyzBabcdefghijklmnopqrstuvwxyzCabcdefghijklmnopqrstuvwxyzDabcdefghijklmnopqrstuvwxyzE.txt`, which has 140 characters including the extension, it shows only up to `Dabcdefghijklmnopqrstuv` (totally 130 characters) with the rest `wxyzE.txt` being trimmed out. If I put this file in a sub-folder named `test`, its name is still trimmed to that same position.

If I use this 140 characters with `.txt` removed (136 then) to name a directory and put the 140-character-named file into that folder, then on the remote Windows 7, the folder is shown with name `abcdefghijklmnopqrstuvwxyzAabcdefghijklmnopqrstuvwxyzBabcdefghijklmnopqrstuvwxyzCabcdefghijklmnopqrstuvwxyzDabcdefghijklmnopqrstuv` (130 characters). And no file in that folder (I think it is because this folder is treated as new folder other than the long name one).

Thanks a lot!

akallabeth:


winpr/libwinpr/file/generic.c
```diff
@@ -890,7 +890,7 @@ static BOOL ConvertFindDataAToW(LPWIN32_FIND_DATAA lpFindFileDataA,
 	if (length > MAX_PATH)
 		length = MAX_PATH;
 
-	CopyMemory(lpFindFileDataW->cFileName, unicodeFileName, length);
+	CopyMemory(lpFindFileDataW->cFileName, unicodeFileName, length * sizeof(WCHAR));
```
mfleisz: length is already multiplied by 2 at the end of the ConvertToUnicode call.
I think what is wrong is the check if length > MAX_PATH which should be length > MAX_PATH * sizeof(WCHAR) - and of course instead of 2 we should also multiply with sizeof(WCHAR) at the end of ConvertToUnicode.

mfleisz: The same applies for the cAlternateFileName conversion too

akallabeth: @mfleisz Yes, you're correct. Fixed that.

Issue #4209, opened 2017-11-04
NLA Authentication error handling broken


lemonmojo:
Since https://github.com/FreeRDP/FreeRDP/commit/367bddd7adbb2f93f8ceca72d83e0281a30aaf5d error handling for NLA is pretty much broken. The problem is that instead of just passing on the NLA error code to `freerdp_set_last_error`, the generic `FREERDP_ERROR_CONNECT_FAILED` is now used in most cases (all errors except for password must change, password expired and account disabled).

This presents a huge problem because it makes it impossible to differentiate between authentication errors (except for the few handled in the switch statement) and, say, a more generic failure like the remote host couldn't be resolved.

In the connection result handler of my FreeRDP implementation I have the following code (which now doesn't work anymore):

```
if (conErrorCode == 0 &&
    result != 0 &&
    lastErrorCode != 0) {
    if (lastErrorCode == STATUS_LOGON_FAILURE) {
        conErrorCode = STATUS_LOGON_FAILURE;
    } else if (lastErrorCode == STATUS_WRONG_PASSWORD) {
        conErrorCode = STATUS_WRONG_PASSWORD;
    } else if (lastErrorCode == STATUS_PASSWORD_EXPIRED) {
        conErrorCode = STATUS_PASSWORD_EXPIRED;
    } else if (lastErrorCode == STATUS_PASSWORD_MUST_CHANGE) {
        conErrorCode = STATUS_PASSWORD_MUST_CHANGE;
    } else if (lastErrorCode == STATUS_ACCESS_DENIED) {
        conErrorCode = STATUS_ACCESS_DENIED;
    } else if (lastErrorCode == STATUS_ACCOUNT_DISABLED) {
        conErrorCode = STATUS_ACCOUNT_DISABLED;
    } else if (lastErrorCode == STATUS_ACCOUNT_RESTRICTION) {
        conErrorCode = STATUS_ACCOUNT_RESTRICTION;
    } else if (lastErrorCode == STATUS_ACCOUNT_LOCKED_OUT) {
        conErrorCode = STATUS_ACCOUNT_LOCKED_OUT;
    } else if (lastErrorCode == STATUS_ACCOUNT_EXPIRED) {
        conErrorCode = STATUS_ACCOUNT_EXPIRED;
    } else if (lastErrorCode == STATUS_LOGON_TYPE_NOT_GRANTED) {
        conErrorCode = STATUS_LOGON_TYPE_NOT_GRANTED;
[...]

PR #4210, opened 2017-11-06
Added additional NLA error mappings.

akallabeth:
Added additional NTSTATUS mappings suggested in #4209


hardening:
@akallabeth that looks like a very bad idea to change values of error code that are in a public header (ERRCONNECT_ACCOUNT_DISABLED and ERRCONNECT_PASSWORD_MUST_CHANGE). What do you think ?

lemonmojo:
@hardening LGTM. Just tested some of the common scenarios like wrong pw, using a non-existing username, password expired, disabled account, account must change pw, etc. and they all came back correctly. Thx!

PR #4053, opened 2017-07-20
Implemented FFMPEG based encoder.

akallabeth:
Does work rudimentary, will need some FFMPEG parameter adjustments.

PR #4212, opened 2017-11-06
Fix libavcodec encoding errors and set tunables.

SriRamanujam:
Opening a PR for the code mentioned in this comment: https://github.com/FreeRDP/FreeRDP/pull/4053#issuecomment-323414737

I managed to get the code into a state where both FreeRDP and Windows' RDP software decoder are able to render the stream. Windows' RDP hardware decoder for some reason is a lot less successful (I've seen black screens, horrible amounts of frame delay, and spurious disconnections due to decoding failures), honestly not sure what to do about that but at least you can force software decoding in Group Policy.

I want to stress that I am by no means a libavcodec or H.264 tuning genius or anything of the sort, I only barely know what the options do and I mostly just adapted recommendations from the ffmpeg wiki article on live streaming. There is almost certainly lots more that can be done.

### DIFF
--- a/libfreerdp/codec/h264_ffmpeg.c
+++ b/libfreerdp/codec/h264_ffmpeg.c
@@ -246,7 +246,11 @@ static int libavcodec_compress(H264_CONTEXT* h264, BYTE** ppDstData, UINT32* pDs
 	if (!libavcodec_create_encoder(h264))
 		return -1;
 
+#if LIBAVUTIL_VERSION_INT >= AV_VERSION_INT(55, 39, 100)
 	av_packet_unref(&sys->packet);
+#else
+	av_free(sys->packet.data);
+#endif
 	av_init_packet(&sys->packet);
 	sys->packet.data = NULL;
 	sys->packet.size = 0;
@@ -256,7 +260,9 @@ static int libavcodec_compress(H264_CONTEXT* h264, BYTE** ppDstData, UINT32* pDs
 #if LIBAVUTIL_VERSION_INT >= AV_VERSION_INT(52, 48, 100)
 	sys->videoFrame->colorspace = AVCOL_SPC_BT709;
 #endif
+#if LIBAVUTIL_VERSION_INT >= AV_VERSION_INT(52, 92, 100)
 	sys->videoFrame->chroma_location = AVCHROMA_LOC_LEFT;
+#endif
 	sys->videoFrame->data[0] = h264->pYUVData[0];
 	sys->videoFrame->data[1] = h264->pYUVData[1];
 	sys->videoFrame->data[2] = h264->pYUVData[2];
--- a/libfreerdp/codec/h264_ffmpeg.c
+++ b/libfreerdp/codec/h264_ffmpeg.c
@@ -140,13 +140,28 @@ static BOOL libavcodec_create_encoder(H264_CONTEXT* h264)
 	{
 		1, h264->FrameRate
 	};
-	sys->codecEncoderContext->gop_size = 0;
+
+	av_opt_set(sys->codecEncoderContext, "preset", "veryfast", AV_OPT_SEARCH_CHILDREN);
+	av_opt_set(sys->codecEncoderContext, "tune", "zerolatency", AV_OPT_SEARCH_CHILDREN);
+
+	sys->codecEncoderContext->flags |= CODEC_FLAG_LOOP_FILTER;
+	sys->codecEncoderContext->me_cmp |= 1;
+	sys->codecEncoderContext->me_subpel_quality = 3;
+	sys->codecEncoderContext->me_range = 16;
+	sys->codecEncoderContext->gop_size = 60;
+	sys->codecEncoderContext->keyint_min = 25;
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
The server error was ```Resource temporarily unavailable [-11]```

akallabeth:
It is working with the current changes (although the image quality is really miserable)

SriRamanujam:
@akallabeth I'm a bit confused, sorry, does the code work for you on Debian stable? To me, your comments seem to contradict each other...

Regarding the terrible image quality: Yeah, I know, I was just trying to get it working with whatever minimum settings I could with the Windows hardware decoder. The image quality can definitely be tuned up, at the expense of the Windows hardware decoder blowing up in ever more spectacular ways. I suspect we are not sending our AVC420 frames exactly the way Windows would encode them itself, but again, I'm not enough of a codec expert to even begin troubleshooting that issue.

akallabeth:
@SriRamanujam No, your changes do not work at all (the ```Resource temporarily unavailable [-11]``` error), but current master does with the mentioned bad image quality

SriRamanujam:
So I've fixed the issue @akallabeth brought up... it turns out that the new libavcodec API returns `EAGAIN` when it wants more input frames. To accommodate this, I've therefore slightly tweaked the avc420_compress API to be able to signal that it requires more input frames and to not try to send a `SurfaceFrameCommand`. 

The h264 subsystems behind `avc420_compress` right now only return a negative status to indicate an error or 1 to indicate success. I decided to use `0`, therefore, to indicate that the subsystem wants more input frames, and, crucially, that the code should not try to send a `SurfaceFrameCommand` because the codec did not give us a valid frame to send.

libfreerdp/codec/h264_ffmpeg.c
```diff
@@ -246,7 +262,12 @@ static int libavcodec_compress(H264_CONTEXT* h264, BYTE** ppDstData, UINT32* pDs
 
 	if (status < 0)
 	{
-		WLog_ERR(TAG, "Failed to encode video frame (%s [%d])",
+		if (status == AVERROR(EAGAIN))
+		{
+			// codec requires more input frames
+			return 0;
```
akallabeth: That is actually an error if in streaming mode. We need 1:1

server/shadow/shadow_client.c
```diff
@@ -823,10 +824,21 @@ static BOOL shadow_client_send_surface_gfx(rdpShadowClient* client,
...
+								 nWidth, nHeight, &avc444.bitstream[0].data,
+								 &avc444.bitstream[0].length);
+
+		if (status < 0)
+		{
+			WLog_ERR(TAG, "avc420_compress failed for avc444");
 			return FALSE;
+		}
+
+		if (status == 0)
```
akallabeth: Same here, need more frames is a configuration error

server/shadow/shadow_client.c
```diff
@@ -866,9 +878,20 @@ static BOOL shadow_client_send_surface_gfx(rdpShadowClient* client,
...
+		status = avc420_compress(encoder->h264, pSrcData, cmd.format, nSrcStep,
+		                nWidth, nHeight, &avc420.data, &avc420.length);
+
+		if (status < 0)
+		{
+			WLog_ERR(TAG, "avc420_compress failed");
 			return FALSE;
+		}
+
+		if (status == 0)
```
akallabeth: Same again

SriRamanujam:
Okay, I removed the commit that introduced EAGAIN handling so it's just the tunables again. Hopefully this should be all good now.

libfreerdp/codec/h264_ffmpeg.c
```diff
@@ -231,7 +246,7 @@ static int libavcodec_compress(H264_CONTEXT* h264, BYTE** ppDstData, UINT32* pDs
 	if (!libavcodec_create_encoder(h264))
 		return -1;
 
-	av_free(sys->packet.data);
+	av_packet_unref(&sys->packet);
```
akallabeth: This breaks old FFMPEG versions (e.g. ubuntu 14.04)
Do a version AVCODEC check >= 55.39.100 the new style, otherwise the old code you deleted

libfreerdp/codec/h264_ffmpeg.c
```diff
@@ -241,14 +256,14 @@ static int libavcodec_compress(H264_CONTEXT* h264, BYTE** ppDstData, UINT32* pDs
 #if LIBAVUTIL_VERSION_INT >= AV_VERSION_INT(52, 48, 100)
 	sys->videoFrame->colorspace = AVCOL_SPC_BT709;
 #endif
+	sys->videoFrame->chroma_location = AVCHROMA_LOC_LEFT;
```
akallabeth: Same here, please add a version check to disable on older versions
LIBAVUTIL 52.92.100

SriRamanujam:
@akallabeth I've added the `#ifdef`s.

akallabeth:
@freerdp-bot test

PR #4224, opened 2017-11-09
codec/nsc: fix memory corruption in case of chroma subsampling

krisztian-kovacs-balabit:
For odd number of rows, the memory copy operation was broken: after exiting
the loop, yplane points to the end of the last row data, and thus (yplane +
rw) points *after* the end of the last row.

### DIFF
--- a/libfreerdp/codec/nsc_encode.c
+++ b/libfreerdp/codec/nsc_encode.c
@@ -237,9 +237,13 @@ static void nsc_encode_argb_to_aycocg(NSC_CONTEXT* context, const BYTE* data,
 
 	if (context->ChromaSubsamplingLevel && (y % 2) == 1)
 	{
-		CopyMemory(yplane + rw, yplane, rw);
-		CopyMemory(coplane + rw, coplane, rw);
-		CopyMemory(cgplane + rw, cgplane, rw);
+		yplane = context->priv->PlaneBuffers[0] + y * rw;
+		coplane = context->priv->PlaneBuffers[1] + y * rw;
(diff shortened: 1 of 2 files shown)


hardening:
@freerdp-bot what do think of it, would you mind give a test ?

libfreerdp/codec/nsc_encode.c
```diff
@@ -237,9 +237,13 @@ static void nsc_encode_argb_to_aycocg(NSC_CONTEXT* context, const BYTE* data,
 
 	if (context->ChromaSubsamplingLevel && (y % 2) == 1)
 	{
-		CopyMemory(yplane + rw, yplane, rw);
-		CopyMemory(coplane + rw, coplane, rw);
-		CopyMemory(cgplane + rw, cgplane, rw);
+		yplane = context->priv->PlaneBuffers[0] + y * rw;
```
mfleisz: Maybe I missed something but isn't ```context->priv->PlaneBuffers[0] + y * rw``` pointing at the end of the buffer here too? (with ```y == context->height``` at this point)

krisztian-kovacs-balabit: Well, you're right that y == context->height at this point, and context->height is odd if we enter this if.

However, if context->ChromaSubsamplingLevel > 0, then the plane has to be padded, which means an extra row (allocated in nsc_context_initialize_encode by ROUND_UP_TO(context->height, 2). Thus, context->priv->PlaneBuffers[0] + y * rw actually points to the first byte of the padding row.

mfleisz: Ah OK - got it :)

PR #4225, opened 2017-11-09
core/connection: use redirection password when reconnecting

krisztian-kovacs-balabit:
According to MS-RDPBCGR the server might send a password in the Redirection PDU
that then must be sent by the client to the server as password.

Since the field either contains a password string (unicode) or a binary cookie,
we try to convert the password from unicode and use it only if conversion
succeeds.


hardening:
@freerdp-bot would you mind give a test ?

PR #4226, opened 2017-11-09
libfreerdp/core/certificate: open key file for reading only

krisztian-kovacs-balabit:
There's no point in writing the key file for read-write, and it makes it
impossible to run the shadow server with the key file being read only.


akallabeth:
What happens if the file does not exist?

krisztian-kovacs-balabit:
The same as previously, as far as I can see -- key_new() fails.

PR #4228, opened 2017-11-10
Fixed uninitialized values.

akallabeth:


libfreerdp/core/gcc.c
```diff
@@ -837,7 +837,8 @@ BOOL gcc_read_client_core_data(wStream* s, rdpMcs* mcs, UINT16 blockLength)
 		                                     RNS_UD_CS_SUPPORT_MONITOR_LAYOUT_PDU) ? TRUE : FALSE;
 
 	if (settings->SupportStatusInfoPdu)
-		settings->SupportStatusInfoPdu = (earlyCapabilityFlags & RNS_UD_CS_SUPPORT_STATUSINFO_PDU) ? TRUE : FALSE;
+		settings->SupportStatusInfoPdu = (earlyCapabilityFlags & RNS_UD_CS_SUPPORT_STATUSINFO_PDU) ? TRUE :
+		                                 FALSE;
```
hardening: The new formatting is not that nice, could we replace that with expr > 0 instead of expr ? TRUE : FALSE

hardening:
I've looked again and we never hit the uninitialized case.

akallabeth:
@hardening I've corrected the formatting. As for we never hit that case, that may be the case (assuming ```is_primary``` is always set) but it still triggers a false positive on static code analysis that can be easily avoided this way.

PR #4229, opened 2017-11-10
Option to force password prompt before connection

akallabeth:
The idea is based on #3257.
If a non NLA connection is requested the password callbacks are not executed
and there is currently no way to read from stdin.
This extension of /from-stdin allows reading the passwords
before the connection is established.


akallabeth:
@hardening what do you think about this one?

seamus-45:
I waited for this patch for a year.
Tested with linux key retention service and Windows LDAP (sssd). Now it is possible to be authorized transparently. Works great! Thx for this patch :+1:

PR #4232, opened 2017-11-13
Added compat define for missing format.

akallabeth:
nightly builds on wheezy and precise don't have that color format defined, replace it with something that is available.


akallabeth:
@bmiklautz @hardening Should fix the broken nightlies, please have a look.

PR #4236, opened 2017-11-14
Scanbuild warning fixes

### LINKED ISSUE #4234: New clang-scanbuild warnings
Fix scanbuild warnings found after the latest merges

akallabeth:
* Fixes a lot of warnings found during scanbuild
* Fixes some formatting issues in files touched
* Fixes #4234

### DIFF
--- a/channels/sshagent/client/sshagent_main.c
+++ b/channels/sshagent/client/sshagent_main.c
@@ -67,7 +67,7 @@ struct _SSHAGENT_LISTENER_CALLBACK
 	IWTSVirtualChannelManager* channel_mgr;
 
 	rdpContext* rdpcontext;
-        const char *agent_uds_path;
+	const char* agent_uds_path;
 };
 
 typedef struct _SSHAGENT_CHANNEL_CALLBACK SSHAGENT_CHANNEL_CALLBACK;
@@ -80,7 +80,7 @@ struct _SSHAGENT_CHANNEL_CALLBACK
 	IWTSVirtualChannel* channel;
 
 	rdpContext* rdpcontext;
-        int agent_fd;
+	int agent_fd;
 	HANDLE thread;
 	CRITICAL_SECTION lock;
 };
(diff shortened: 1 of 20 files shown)


channels/sshagent/client/sshagent_main.c
```diff
@@ -236,23 +243,21 @@ static UINT sshagent_on_data_received(IWTSVirtualChannelCallback* pChannelCallba
...
 	if (WaitForSingleObject(callback->thread, INFINITE) == WAIT_FAILED)
 	{
 		UINT error = GetLastError();
 		WLog_ERR(TAG, "WaitForSingleObject failed with error %"PRIu32"!", error);
 		return error;
 	}
 
-        CloseHandle(callback->thread);
+	CloseHandle(callback->thread);
+	LeaveCriticalSection(&callback->lock);
```
mfleisz: Good catch!

client/X11/xf_cliprdr.c
```diff
@@ -1211,21 +1213,19 @@ static UINT xf_cliprdr_server_capabilities(CliprdrClientContext* context,
 static UINT xf_cliprdr_server_format_list(CliprdrClientContext* context,
         CLIPRDR_FORMAT_LIST* formatList)
 {
-	int i, j;
-	CLIPRDR_FORMAT* format;
+	UINT32 i;
+	int j;
```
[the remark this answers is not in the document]
akallabeth: @mfleisz Yes, could move the loop counters to the scopes too. As for the additional scope with format, I think it is better visible now that this is intended and not some uninitialized variable used after the loop

channels/sshagent/client/sshagent_main.c
akallabeth: Not mine, thank astyle ;)

client/X11/xf_cliprdr.c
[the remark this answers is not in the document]
mfleisz: @akallabeth OK agree on the format variable handling - maybe you can ditch the global loop counter when rebasing? 👍

libfreerdp/core/nla.c
```diff
@@ -1140,7 +1145,7 @@ SECURITY_STATUS nla_decrypt_public_key_echo(rdpNla* nla)
 		ap_integer_decrement_le(public_key2, public_key_length);
 	}
 
-	if (memcmp(public_key1, public_key2, public_key_length) != 0)
+	if (!public_key1 || !public_key2 || memcmp(public_key1, public_key2, public_key_length) != 0)
 	{
```
hardening: if public_key1 or public_key2 are NULL that means that nla->packageName is not one of the one we know, so that means a real big bug that will have triggered some error long before. So these checks and initializations aren't needed.

hardening:
I'm bothered that this PR has been merged so fast.
Static analysers have their good part, but I don't think we should change the code to please them when it's just some case that they don't detect correctly.
I'm really annoyed by all these initializations and checks that are sometime unneeded, but just added to have a clean ```clang-scanbuild``` result.

PR #4237, opened 2017-11-14
Replaced atoi

akallabeth:
* Replaced ```atoi``` with ```strtol```, ```strtol```, ```_strtoui64``` and ```_strtoi64```
* Added range checks wherever possible
* Added conversion error checks wherever possible

### DIFF
--- a/libfreerdp/common/assistance.c
+++ b/libfreerdp/common/assistance.c
@@ -21,6 +21,8 @@
 #include "config.h"
 #endif
 
+#include <errno.h>
+
 #include <winpr/wtypes.h>
 #include <winpr/crt.h>
 #include <winpr/crypto.h>
@@ -78,7 +80,6 @@ int freerdp_assistance_crypt_derive_key_sha1(BYTE* hash, int hashLength, BYTE* k
 	BYTE* buffer;
 	BYTE pad1[64];
 	BYTE pad2[64];
-
 	memset(pad1, 0x36, 64);
 	memset(pad2, 0x5C, 64);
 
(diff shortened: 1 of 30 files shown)


akallabeth:
@freerdp-bot test

libfreerdp/common/assistance.c
```diff
@@ -974,26 +964,26 @@ int freerdp_assistance_parse_file_buffer(rdpAssistanceFile* file, const char* bu
...
-
 		r = (char*) malloc(length + 1);
 
 		if (!r)
 			return -1;
 
 		CopyMemory(r, p, length);
 		r[length] = '\0';
+		errno = 0;
+		{
```
hardening: Why are you setting errno here ?

libfreerdp/crypto/tls.c
```diff
@@ -23,6 +23,7 @@
 
 #include <assert.h>
 #include <string.h>
+#include <errno.h>
```
hardening: not needed ?

libfreerdp/common/assistance.c
akallabeth: to ensure no other function has set ERRNO (tried to do that consistently before calling  ```strto*```)

libfreerdp/crypto/tls.c
akallabeth: correct, looks like a leftover

libfreerdp/common/assistance.c
akallabeth: hmm, comment gone? Ah well, I tried to set ```errno = 0``` before ```strto*``` calls wherever possible

PR #4240, opened 2017-11-16
Fixed wayland issues.

akallabeth:


client/Wayland/wlf_input.c
```diff
@@ -25,22 +25,30 @@
 
 #include "wlf_input.h"
 
-BOOL wlf_handle_pointer_enter(freerdp *instance, UwacPointerEnterLeaveEvent *ev) {
-	rdpInput* input = instance->input;
+BOOL wlf_handle_pointer_enter(freerdp* instance, UwacPointerEnterLeaveEvent* ev)
+{
+	if (!instance || !ev || !instance->input)
+		return FALSE;
```
hardening: I don't see how we can get ev == NULL, the same for instance.
The ```rdpInput* input = instance->input``` was nice for the rest of the code IMHO

client/Wayland/wlfreerdp.c
```diff
@@ -315,20 +333,18 @@ static int wlfreerdp_run(freerdp* instance)
 		return -1;
 	}
 
-	handle_uwac_events(instance, context->display);
-
 	while (!freerdp_shall_disconnect(instance))
 	{
 		handles[0] = context->displayHandle;
-		count = freerdp_get_event_handles(instance->context, &handles[1], 63);
```
hardening: I don't understand in which case freerdp_get_event_handles() returns no event handle.

client/Wayland/wlf_input.c
akallabeth: It is a callback function. Arguments should be checked and a proper error code returned. Compiler will optimize unnecessary checks anyway.
for the ```instance->input``` part, that I can do again.

client/Wayland/wlfreerdp.c
akallabeth: Not currently, correct. But this is client code and the core library may change over time, so better safe than sorry

akallabeth:
@hardening Would not call that self protecting, all checks return proper error codes. I don't like segmentation faults (best case), as they can often lead to security vulnerabilities. Having a proper error return is usually the better option.

client/Wayland/wlf_input.c
hardening: Well IMHO that's exactly the opposite, a callback is supposed to be called with some pre-requirements. Having instance != NULL and ev != NULL are the requirements, even  testing instance->input is questionable.

Look at the call hierarchy for ```wlf_handle_pointer_enter```  (Ctrl+Alt+H under eclipse) you'll see that NULL in instance or ev are just not possible. So it had some micro tests that for sure the compiler will keep as is as there's no place for optimization here. But worst, when you read the code you can be suspicious on the instance or ev parameters when there's no reason for that.

hardening:
When you add unneeded checks: checks for things that can't happen, it's self protecting code. That means that there's no convention or requirements between the callee and the caller. This can make sense for public API, but here we're not in that case.
I don't like segfaults either, even if I prefer them to corruption because a valgrind run or even gdb will catch a segfault very easily. The security argument has nothing to do here, adding some unneeded checks doesn't improve security, and it's not a proper error handling either.

akallabeth:
@hardening Sorry, forgot about this one.
To your argument:
* The API is the public API of the core library, the checks are done in the client.
I can add some warning/error messages if you like, but I still prefer handling this without killing the client in case we changed some core API and did not properly update the client (which may happen, since some are rarely tested)

mfleisz:
@hardening @akallabeth 
Same here as with #4322

akallabeth:
@hardening Removed the checks in functions without return value, but keeping them where we can signal an error.

hardening:
@akallabeth I'm totally against unneeded checks for reasons that I've already explained multiple times. So my proposal is that I'm approving this PR and I'll submit a PR that will do a big cleanup with checks that have been added for no reason. Do you agree ?

akallabeth:
@hardening I'm against removing checks as I already pointed out for various reasons.
Would agree with adding log messages for easier debugging but no removing perfectly fine parameter checks. (that is what a return value is for)

client/Wayland/wlfreerdp.c
```diff
@@ -245,7 +263,7 @@ static BOOL handle_uwac_events(freerdp* instance, UwacDisplay* display)
 				context = (wlfContext*)instance->context;
 				context->waitingFrameDone = FALSE;
 
-				if (context->haveDamage && !wl_end_paint(instance->context))
+				if (context->haveDamage && !wl_update_content(context))
 					return FALSE;
```
MartinHaimberger: Good catch

akallabeth:
@hardening @mfleisz @bmiklautz @MartinHaimberger Could we merge this or should I close it?

Issue #4227, opened 2017-11-10
Convert settings->Password to binary blob


akallabeth:
Currently the ```settings->Password``` field contains a UTF-8 string. Since the authentication token can be either a WCHAR string or a binary blob it would be a good idea to convert command line/external values to WCHAR immediately to eliminate the necessity of handling binary blobs separately.

PR #4241, opened 2017-11-16
core/connection: handle trailing zeroes in redirection password

krisztian-kovacs-balabit:
According to MS-RDPBCGR the Password field in the Redirection PDU
should be null-terminated. In that case the trailing zeroes are part of
settings->RedirectionPasswordLength.

Unfortunately, if that is the case, ConvertFromUnicode() fails if we use
RedirectionPasswordLength as the length of the password because the
trailing zeroes cannot be converted.

Since the redirection PDU is always copied in
rdp_redirection_apply_settings() so that the buffer contains an extra
trailing zero wchar_t, it's safe to pass -1 as the length of the string
to ConvertFromUnicode().

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -373,7 +373,7 @@ BOOL rdp_client_redirect(rdpRdp* rdp)
 		/* If unicode conversion fails, this might be because the Password field
 		 * might contain a non-unicode cookie value. Simply ignore in this case. */
 		if (ConvertFromUnicode(CP_UTF8, 0,
-			 (WCHAR*) settings->RedirectionPassword, settings->RedirectionPasswordLength,
+			 (WCHAR*) settings->RedirectionPassword, -1,
 			 &password, 0,
 			 NULL, NULL) > 0)
 		{


hardening:

akallabeth:
@krisztian-kovacs-balabit Does that really work? I've opened #4227 to track this, as the only real solution I can think of is keeping the password in the format to be sent.

krisztian-kovacs-balabit:
Yes, it seems to work for the subset of cases where RedirectionPassword is indeed a password.

Quoting the relevant part of MS-RDPBCGR: "A variable-length array of bytes containing the password used by the user in Unicode format, including a null-terminator, or a cookie value that MUST be passed to the target server on successful connection."

My assumption was that if RedirectionPassword can be decoded from Unicode, then we can safely handle it as a password, as converting back to Unicode will lead to the same representation as before. I think MS-RDPBCGR is not quite clear what the cookie value might be -- but as long as the contents of the field is valid UTF-16 we don't lose information, so at least the password case is handled correctly now.

It would probably be useful to ask for some clarification of the documentation from the DocHelp team at Microsoft.

krisztian-kovacs-balabit:
Indeed, it's still a subset but still better than the current behaviour where RedirectionPassword is not used at all when reconnecting.

akallabeth:
@krisztian-kovacs-balabit Ok, looks like I missed something in #4225, better option would be to check where ```settings->Password``` is used and do the ```settings->RedirectionPassword``` check there.

krisztian-kovacs-balabit:
OK, I'm looking into using RedirectionPassword directly instead.

krisztian-kovacs-balabit:
I've created PR #4303 instead, which also reverts the change merged in #4225.

Issue #4183, opened 2017-10-16
Add aFreeRDP to F-Droid?


ccpp:
Was it ever considered to contribute aFreeRDP to the free-open-source app repository of F-Droid?
So, aFreeRDP could be installed from [F-Droid](https://f-droid.org/) instead of Google Play.

Refering [their contribution page](https://f-droid.org/en/contribute/):
- As far as I can see it, the FreeRDP project perfectly matches their [inclusion policy](https://f-droid.org/en/docs/Inclusion_Policy/).
- They can be either asked to include aFreeRDP by opening an issue to their "Requests For Packaging" repo, or, a pull request including build instructions can be opened at their "Data" repository

akallabeth:
@ccpp We did not find the time for that, but if someone is willing to do that it would be great.

ccpp:
@akallabeth @bmiklautz 
I created [a contribution branch for fdroiddata](https://gitlab.com/c7np6r/fdroiddata/commit/9798d71d52784119239f86c5ef59b335a835e5e6) which contributes aFreeRDP version version 2.0.0-beta1+android11, and is tested to compile and run.

Can you please have a look and see if I can submit that branch as a pull request? I added following modifications:
- Fix a compilation bug where sysroot is not set automatically, in case GCC is used
- Fix gradle build without RELEASE_STORE_FILE / RELEASE_KEY_ALIAS (not needed because fdroid signs the package in a separate step)
- Added security upgrade of OpenSSL to version 1.1.0f
- I tagged the build "ndk=r10e" because of the FreeRDP documentation. Is this OK or should this be updated?
- I used the "android-build-32.conf" configuration only; obviously there is no easy way to compile both x32 and x64... is there? Would you suggest a different compilation step?

akallabeth:
@ccpp Nice work! Here some notes:
* Better use branch ```2.0.0-rc0``` which is the latest tag we have
* The NDK stuff is a bit outdated, I've successfully compiled with more recent versions
* Building both, 32bit and 64bit versions can be done by running the script twice, once with the 32bit configuration, second time with the 64bit configuration followed by the final gradle build step

akallabeth:
[...]

PR #4242, opened 2017-11-17
Increase APK versionCode for aFreeRDP 2.0-rc1

ccpp:
versionCode should be different for each version. Regarding #4183, this is necessary to get the version out at all


akallabeth:
@freerdp-bot test

Issue #3267, opened 2016-04-06
drivestoredirect parameter in rdp file is not correctly parsed


HenryJacques:
Using latest master and a rdp file containing the `drivestoredirect`parameter.
As stated in MS' specifications, if no value is specified the client must not redirect any drives.
Using mstc, the following line has the effect to not redirect any drive:

> drivestoredirect:s:

However FreeRDP redirects the root "/" even if no drive is specified.

bmiklautz:
@HenryJacques would you add the rdp file your resting with for reference?

HenryJacques:
@bmiklautz of course:

> redirectclipboard:i:0
redirectprinters:i:1
redirectcomports:i:0
redirectsmartcards:i:0
redirectdrives:i:0
drivestoredirect:s:
session bpp:i:32
prompt for credentials on client:i:1
span monitors:i:1
use multimon:i:1
remoteapplicationmode:i:1
server port:i:3389
allow font smoothing:i:1
promptcredentialonce:i:0
videoplaybackmode:i:1
audiocapturemode:i:0
gatewayusagemethod:i:0
gatewayprofileusagemethod:i:1
gatewaycredentialssource:i:0
full address:s:MYBROKER.MYDOM
alternate shell:s:||iexplore
remoteapplicationprogram:s:||iexplore
remoteapplicationname:s:MyApp
remoteapplicationcmdline:s:-k http://myapp.mydom/
workspace id:s:MYBROKER.MYDOM
use redirection server name:i:1
loadbalanceinfo:s:tsv://MS Terminal Services Plugin.1.MYAPP
alternate full address:s:MYBROKER.MYDOM

AlexDarkf0x:
I can confirm I have the same issue. Master with commit  a0d9969
Removing the line from .rdp file, works for me.

PR #4244, opened 2017-11-17
Fix parsing of drivestoredirect (#3267)

akallabeth:


client/common/file.c
```diff
@@ -1237,7 +1237,8 @@ BOOL freerdp_client_populate_settings_from_rdp_file(rdpFile* file, rdpSettings*
 		 * Very similar to DevicesToRedirect, but can contain a
 		 * comma-separated list of drive letters to redirect.
 		 */
-		freerdp_set_param_bool(settings, FreeRDP_RedirectDrives, TRUE);
+		const BOOL empty = file->DrivesToRedirect || (strlen(file->DrivesToRedirect) == 0);
```
mfleisz: I think that should be ```!file->DrivesToRedirect ``` - otherwise we might hit a strlen(NULL)

akallabeth:
@mfleisz oops...

Issue #4101, opened 2017-08-18
Problem with 16 bit and /rfx


bmiklautz:
If the local X server is started with 16 bit and /rfx is used then the resulting session is distorted (artifacts and half of the screen is black). This happens with sw and hw GDI.

Steps to reproduce:
* start a X server with 16 bit (e.g. Xephyr :200 -screen 1024x768x16)
* launch xfreerdp on this xserver with /rfx

akallabeth:
Fixed ```/gdi:sw``` RFX decoding, with ```/gdi:hw``` there are distortions left, but they seem to originate from some other callback.

PR #4255, opened 2017-11-20
Fix #4253: Rfx decode tile width.

akallabeth:


akallabeth:
@hardening Got time to check this one? #4101 has a nice little test description to check if it works.

PR #4264, opened 2017-11-21
Fixed NLA for Negotiate and NTLM authentication

cedrozor:
Regression due to the recent addition of kerberos support


akallabeth:
@freerdp-bot test

PR #4267, opened 2017-11-22
Enable FIPS mode automatically

ondrejholy:
FreeRDP aborts if OpenSSL operates in FIPS mode and +fipsmode is not manually specified. Let's prevent the abortion and enable the necessary options in that case automatically.

### DIFF
--- a/winpr/libwinpr/utils/ssl.c
+++ b/winpr/libwinpr/utils/ssl.c
@@ -346,6 +346,15 @@ BOOL winpr_CleanupSSL(DWORD flags)
 	return TRUE;
 }
 
+BOOL winpr_FIPSMode(void)
+{
+#if (OPENSSL_VERSION_NUMBER < 0x10001000L)
+	return FALSE;
+#else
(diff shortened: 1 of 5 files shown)


winpr/libwinpr/utils/ssl.c
```diff
@@ -346,6 +346,15 @@ BOOL winpr_CleanupSSL(DWORD flags)
 	return TRUE;
 }
 
+BOOL winpr_FIPSMode()
```
akallabeth: Missing void, empty argument is variable arguments in plain C

winpr/libwinpr/utils/ssl.c
```diff
@@ -358,4 +367,9 @@ BOOL winpr_CleanupSSL(DWORD flags)
...
```
akallabeth: Missing void, empty argument is variable arguments in plain C

ondrejholy:
@akallabeth Sure, done.

Issue #4231, opened 2017-11-11
freerdp-2.0.0_rc0: 5 tests failed out of 184 on ppc


ernsteiswuerfel:
[LastTest.log.gz](https://github.com/FreeRDP/FreeRDP/files/1464035/LastTest.log.gz)
[freerdp_build.log.gz](https://github.com/FreeRDP/FreeRDP/files/1464036/freerdp_build.log.gz)

The following tests FAILED:
	 66 - TestIoDevice (Failed)
	 71 - TestImage (Failed)
	163 - TestFreeRDPCodecNCrush (Failed)
	165 - TestFreeRDPCodecZGfx (SEGFAULT)
	182 - TestClientRdpFile (Failed)
Errors while running CTest

akallabeth:
@ernsteiswuerfel We're lacking PPC hardware for testing, so I guess you'll need to get your hands dirty on those. gdb backtrace of ```TestFreeRDPCodecZGfx``` would be a start.

ernsteiswuerfel:
No problem, here you are.

`# coredumpctl gdb
           PID: 10131 (TestFreeRDPCode)
           UID: 250 (portage)
           GID: 250 (portage)
        Signal: 11 (SEGV)
     Timestamp: Tue 2017-11-14 22:41:18 CET (1min 37s ago)
  Command Line: /var/tmp/portage/net-misc/freerdp-2.0.0_rc0/work/freerdp-2.0.0_rc0_build/Testing/TestFreeRDPCodec TestFreeRDPCodecZGfx
    Executable: /var/tmp/portage/net-misc/freerdp-2.0.0_rc0/work/freerdp-2.0.0_rc0_build/Testing/TestFreeRDPCodec
 Control Group: /user.slice/user-1000.slice/session-c1.scope
          Unit: session-c1.scope
         Slice: user-1000.slice
       Session: c1
     Owner UID: 1000 (ef)
       Boot ID: 4a691e6e80cc4bff88c30f9b8561bdd2
    Machine ID: a11f6dbd63d44c20941f9f48a8fbb3e4
      Hostname: T800
       Storage: /var/lib/systemd/coredump/core.TestFreeRDPCode.250.4a691e6e80cc4bff88c30f9b8561bdd2.10131.1510695678000000.lz4
       Message: Process 10131 (TestFreeRDPCode) of user 250 dumped core.
                
                Stack trace of thread 10131:
[...]

PR #4268, opened 2017-11-23
Fix ZGfx decoder on big endian

akallabeth:
#4231 pointed out some issues on big endian, here a fix for one of them.


bmiklautz:
nicely solved ;). Thanks.

PR #4282, opened 2017-11-27
Fix #4281: Added option to prefer IPv6 over IPv4

### LINKED ISSUE #4281: Force ipv6 usage
If host names are used currently ipv4 is preferred if a A record is found.
In order to disable this we should add a /ipv6 switch to force ipv6.

akallabeth:
What this does:
By default FreeRDP prefers IPv4 addresses when resolving hostnames.
With this additional argument FreeRDP will use IPv6 addresses whenever one is available.

How to test:
1. Connect to a server with a URL (not an IP address as ```/v:<url>``` argument)
1. Connect once with ```/ipv6``` and once without
1. The connection must always succeed for valid hosts
   1. With a IPv4 only host
      * Connection must succeed, always the same address
   1. Mixed IPv4/IPv6 host
      * With the parameter the IPv6 address must be used, otherwise the IPv4 address
   1. IPv6 only host
      * The IPv6 address must always be used
1. To test the unix socket support
   1. Start the shadow server with ```/ipc-socket:<local path>``` argument
   1. Connect with ```/v:<local path>```
   1. The client address must be resolved to ```127.0.0.1```


akallabeth:
@freerdp-bot test

PR #4301, opened 2017-12-06
geometry: a skeleton for the MS-RDPEGT channel

hardening:
The base for implementing the MS-RDPEGT client channel that allows to track window geometry.

### DIFF
--- a/channels/geometry/client/geometry_main.c
+++ b/channels/geometry/client/geometry_main.c
@@ -1,0 +1,365 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ * Geometry tracking Virtual Channel Extension
+ *
+ * Copyright 2017 David Fort <contact@hardening-consulting.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


channels/geometry/client/geometry_main.c
```diff
@@ -0,0 +1,365 @@
...
+	Stream_Read_INT32(s, rgndata->boundingRect.x);
+	Stream_Read_INT32(s, rgndata->boundingRect.y);
+	Stream_Read_INT32(s, right);
+	Stream_Read_INT32(s, bottom);
+
+	rgndata->boundingRect.width = right - rgndata->boundingRect.x;
+	rgndata->boundingRect.height = bottom - rgndata->boundingRect.y;
+
+	len -= 32;
+	if (len / 4 < rgndata->nRectCount)
```
mfleisz: I think this should be ```len / 16``` (4 * sizeof(UINT32))

include/freerdp/channels/geometry.h
```diff
@@ -0,0 +1,60 @@
...
+enum {
+	GEOMETRY_UPDATE	= 1,
+	GEOMETRY_CLEAR	= 2
+};
+
+enum {
+	RDH_RECTANGLE = 1
+};
+
+struct _FREERDP_RGNDATA {
```
mfleisz: ```{``` Formatting (same for the enums above)

channels/geometry/client/geometry_main.c
hardening: good catch

hardening:
@mfleisz 's remarks have been addressed

hardening:
@mfleisz are the changes ok for you ?

PR #4302, opened 2017-12-06
Fix #3602 by preventing to poll stdin

### LINKED ISSUE #3602: Gateway connection blocks indefinitely with RDG (/gt:http, /gt:auto)
There seems to be a timeout when connecting via TDG ("http")
Connection works with /gt:rpc

Here's a log without any /gt argument:

    Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.
    [00:29:26:137] [23617:23617] [INFO][com.freerdp.utils] - Registering signal hook...
    [00:29:26:140] [23617:23618] [INFO][com.freerdp.client.common.cmdline] - loading channel cliprdr
    [00:29:27:443] [23617:23618] [INFO][com.winpr.sspi.NTLM] - VERSION ={
    [00:29:27:443] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	ProductMajorVersion: 6
    [00:29:27:443] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	ProductMinorVersion: 1
    [00:29:27:443] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	ProductBuild: 7601
    [00:29:27:443] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	Reserved: 0x000000
    [00:29:27:443] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMRevisionCurrent: 0x0F
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - negotiateFlags "0xE2898235"
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_56 (0),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_KEY_EXCH (1),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_128 (2),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_VERSION (6),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_TARGET_INFO (8),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_EXTENDED_SESSION_SECURITY (12),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_TARGET_TYPE_DOMAIN (15),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_ALWAYS_SIGN (16),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_NTLM (22),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_SEAL (26),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_SIGN (27),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_REQUEST_TARGET (29),
    [00:29:27:529] [23617:23618] [INFO][com.winpr.sspi.NTLM] - 	NTLMSSP_NEGOTIATE_UNICODE (31),

    (... a lot of NTLM lines...)

Then, connection hangs.

ccpp:
This pull request avoids that standard input is polled by error.

Fixes #3602

There might be another bug inside the SSL BIO handling where a FD is not initialized correctly / "on time", which is fixed by this.

### DIFF
--- a/libfreerdp/crypto/tls.c
+++ b/libfreerdp/crypto/tls.c
@@ -735,6 +735,15 @@ int tls_do_handshake(rdpTls* tls, BOOL clientMode)
 		pollfds.events = POLLIN;
 		pollfds.revents = 0;
 
+		/*
+		 * This is a hotfix for #3602.
+		 * The condition fd == 0 should actually never hold
+		 */
+		if (fd == STDIN_FILENO)
+		{
+			continue;
+		}
+
 		do
 		{
 			status = poll(&pollfds, 1, 10 * 1000);


ccpp:
OK I dug a bit further into the code, and what I found is;
- This happens after establishing the RDG in and out proxy connections, when the tunnel is being established over NLA.
- In this case I never observed any FD other than 0, so probably there is no FD (obviously there is no real network socket) which could be `select`ed or `poll`ed. (prove me wrong if that's not correct!)
- With my fix, `BIO_should_retry` evaluates to true for many iterations (busy waiting), until finally `BIO_do_handshake` succeeds.

...which actually means to me that my patch is actually not a workaround but an actual bug fix (just the busy should be dealt with): it just makes sure that `BIO_do_handshake` is being continued also in the case where `select`/`poll` is unavailable.

What do you think?

akallabeth:
@ccpp Somebody suggested it may be caused by the fact that the remote host is resolved twice and input/output connections may end up being connected to different hosts (if a load balancer is used)
Could that be the issue here?

ccpp:
@akallabeth I think I had the same finding as you.
However, this is not the problem. In my case the gateway host resolves to exactly one IP

akallabeth:
@ccpp We have a `custom` BIO (`BIO_s_buffered_socket`) that implements the call in `transport_bio_buffered_ctrl`. It should return the socket of the underlying `BIO_s_simple_socket` implementing the call in `transport_bio_simple_ctrl` and returning the actual socket.

ccpp:
@akallabeth see https://github.com/FreeRDP/FreeRDP/issues/3602#issuecomment-378741070 :)

akallabeth:
@ccpp As it looks like this is not the correct fix I'll close this pull and will merge one that implements the stuff you're currently on

PR #4303, opened 2017-12-06
Use redirection password

krisztian-kovacs-balabit:
This pull request implements using the redirection password (if present) in NLA authentication performed after a redirection.

This is required because according to MS-RDPBCGR the password field of a Redirection PDU is:

> A variable-length array of bytes containing the password used by the user in
Unicode format, including a null-terminator, or a cookie value that MUST be passed to the target
server on successful connection.

Also, with this change it's no longer meaningful to try to set settings->Password to an UTF-8 encoded representation of settings->RedirectionPassword so that change is reverted.

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -385,22 +385,6 @@ BOOL rdp_client_redirect(rdpRdp* rdp)
 			return FALSE;
 	}
 
-	if (settings->RedirectionFlags & LB_PASSWORD)
-	{
-		char *password = NULL;
-
-		/* If unicode conversion fails, this might be because the Password field
-		 * might contain a non-unicode cookie value. Simply ignore in this case. */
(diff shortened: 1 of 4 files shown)


akallabeth:
@freerdp-bot test

krisztian-kovacs-balabit:
Some additional information: since MS-RDPBCGR is not quite clear if the Password field of RDP_REDIRECTION_PACKET can contain non-Unicode values (and how exactly those should be used by the client upon reconnecting) I've asked clarification in email from the Microsoft Dochelp team.

Also, based on the OPN files for MS-RDPBCGR we've found out that Microsoft Message Analyzer simply assumes that the Password field of the redirection PDU contains a UTF16-encoded password.

akallabeth:
Would also make #4227 obsolete.

akallabeth:
@mfleisz @hardening Anyone time for a review?

PR #4305, opened 2017-12-08
Fix XRandr for old systems and MacOSX

hardening:
We need a recent enough version of XRandr to correctly detect monitor.
Also this patch adds some typo or style fixes.


hardening:
@diget @akallabeth I finally found what was wrong with MacOS X, but it was also affecting systems with old versions of XRandr

libfreerdp/core/rdp.c
```diff
@@ -604,7 +604,7 @@ BOOL rdp_send_data_pdu(rdpRdp* rdp, wStream* s, BYTE type, UINT16 channel_id)
 	Stream_SetPosition(s, length);
 	Stream_SealLength(s);
 
-	WLog_DBG(TAG, "%s: sending data (type=0x%x size=%d channelId)", __FUNCTION__, type, Stream_Length(s), channel_id);
+	WLog_DBG(TAG, "%s: sending data (type=0x%x size=%d channelId=%d)", __FUNCTION__, type, Stream_Length(s), channel_id);
```
akallabeth: Looks like there are a few issues here (not your fault, but none the less):
* ```Stream_GetPosition``` returns ```sizt_t```
  * Type of ```length``` is ```UINT16```
  * Type of ```sec_hold``` is ```int```
* Debug format string uses wrong format specifiers

hardening: ok will change that

hardening:
I took the opportunity to treat all callers of Stream_GetPosition()

libfreerdp/core/rdp.c
```diff
@@ -604,7 +604,7 @@ BOOL rdp_send_data_pdu(rdpRdp* rdp, wStream* s, BYTE type, UINT16 channel_id)
 	Stream_SetPosition(s, length);
 	Stream_SealLength(s);
 
-	WLog_DBG(TAG, "%s: sending data (type=0x%x size=%d channelId=%d)", __FUNCTION__, type, Stream_Length(s), channel_id);
+	WLog_DBG(TAG, "%s: sending data (type=0x%x size=%d channelId=%"PRIu16")", __FUNCTION__, type, Stream_Length(s), channel_id);
```
akallabeth: Wrong format string argument, size_t has its own defined ```PRIzd``` (or similar)

hardening:
@akallabeth is it good for you now ?

akallabeth:
@hardening Nice work ^^

PR #4307, opened 2017-12-09
Simple code style change

cpsw:
A very simple code style change to isolate the attribution from the checking. I'm submitting this issue after see some discussion around this point, which I agree and I would change in my own code. It is just a suggestion!


hardening:
@freerdp-bot I guess the test is gonna work don't you ?

akallabeth:
@cpsw A link to the discussion would be nice ;)
As @hardening already mentioned, it is not really necessary and more or less preference.

PR #4313, opened 2017-12-14
disp: implement dynamic resolution for X11

hardening:
If the display channel is available we use it to allow the user to resize the xfreerdp window. When the window is resized we announce a new monitor layout and the server reacts by doing a reactivation sequence to the new size.
The minimum window size is limited to 300x300 as 2012 servers crash horribly if we send them a smaller layout.

### DIFF
--- a/channels/disp/client/disp_main.c
+++ b/channels/disp/client/disp_main.c
@@ -98,7 +98,6 @@ UINT disp_send_display_control_monitor_layout_pdu(DISP_CHANNEL_CALLBACK* callbac
 	type = DISPLAY_CONTROL_PDU_TYPE_MONITOR_LAYOUT;
 
 	s = Stream_New(NULL, length);
-
 	if(!s)
 	{
 		WLog_ERR(TAG, "Stream_New failed!");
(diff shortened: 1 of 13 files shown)


akallabeth:
@hardening Did some tests, works (mostly) as expected.
Only thing that I've noticed is that the requests are sent too fast, so when resizing the window the resolution set may be different from the window size set locally.
Could probably be solved by ignoring all resize events until some grace timeout is expired.

hardening:
@akallabeth yes right I'm having the same. My problem is that I'm tracking window resizes done with the mouse by looking at configure events, but you also get configure event when the window is resized by a XWindowResize. And that's exactly what happens when the reactivation sequence takes place:

* you start dragging the window corner with the mouse, we get a configure event with the new window size;
* we send a new layout via the display channel;
* during that time you're probably continuing to move the mouse;
* the server does a reactivation sequence to the new size, so we XWindowResize the main window, and we get a configure event to a size that does not correspond to the mouse position;

I'll look at delaying the sending of the layout, but that makes the channel much more complicated.

akallabeth:
@hardening I think a simple timer should suffice. Start a timer with the current resolution, if a new one arrives restart the timer with the new resolution. So, if you're still moving/resizing the window, all events are ignored, but if you stop for e.g. half a second, a new resolution is requested from the server

akallabeth:
Merging as the resize issue is an annoyance but the feature itself is working stable.
The issues can be worked out in follow up pulls.

PR #4318, opened 2017-12-19
Dynamic resolution improvements

hardening:
This PR introduces a new Activated event that is triggered when a (re)activation sequence has completed. This is needed as the DesktopResize callback is called just after capabilities, that means that the activation sequence is not completed (and so we theoretically can't send any packet).
As the goal was to limit the number of resize requests, we also introduced a Timer event that is triggered every 100ms, it is used by the display implementation for X11 to send resizes not more than every 200 ms. But the timer event could be used by any other channel.

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -86,6 +86,7 @@
 #include <winpr/synch.h>
 #include <winpr/file.h>
 #include <winpr/print.h>
+#include <winpr/sysinfo.h>
 #include <X11/XKBlib.h>
 
 #include "xf_gdi.h"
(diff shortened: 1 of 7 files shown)


hardening:
@freerdp-bot I think you failed testing on the windows host, can you try again ?

akallabeth:
@freerdp-bot be so kind and test!

mfleisz:
@freerdp-bot test

mfleisz:
@akallabeth @hardening Sorry - had to update cmake on that build machine and it looks like the x86 version was installed before. Will try to fix this asap!

mfleisz:
@freerdp-bot test please

akallabeth:
@freerdp-bot does not like testing...

PR #4319, opened 2017-12-20
Fullscreen without _NET_WM_FULLSCREEN_MONITORS

### LINKED ISSUE #3439: multiple issues with fullscreen mode on two monitors

akallabeth:
Some window managers do not support _NET_WM_FULLSCREEN_MONITORS.
In that case multimonitor fullscreen does not properly work, so
add a path resizing the window over all screens instead.
Based on @erbth pull request, adding proper X11 atom checks.

Fixes #3439, #2707

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -30,6 +30,7 @@
 
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
+#include <X11/Xatom.h>
 
 #ifdef WITH_XRENDER
 #include <X11/extensions/Xrender.h>
@@ -1738,6 +1739,19 @@ static int xfreerdp_client_stop(rdpContext* context)
 	return 0;
 }
 
+static Atom get_supported_atom(xfContext* xfc, const char* atomName)
+{
+	unsigned long i;
+	const Atom atom = XInternAtom(xfc->display, atomName, False);
+
+	for (i = 0;  i < xfc->supportedAtomCount;  i++)
+	{
+		if (xfc->supportedAtoms[i] == atom)
+			return atom;
+	}
+
+	return None;
+}
 static BOOL xfreerdp_client_new(freerdp* instance, rdpContext* context)
 {
 	xfContext* xfc = (xfContext*) instance->context;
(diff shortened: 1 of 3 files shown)


hardening:
@T3N4K4 can you check if that removes the need of your /openbox hack ?

hardening:
@akallabeth BTW you need a rebase for this one

akallabeth:
@hardening Don't know for @T3N4K4 but tests with openbox on my debian jessie show a working multimonitor fullscreen (as opposed to the current nightly build)

T3N4K4:
@hardening @akallabeth Sorry for the delay, tight schedule this week.
Seems to be working for me.

PR #4320, opened 2017-12-20
Coverity Scan fixes

ondrejholy:
Proposing various fixes based on Coverity Scan results, see commit descriptions for more details.

### DIFF
--- a/libfreerdp/core/rdp.c
+++ b/libfreerdp/core/rdp.c
@@ -615,17 +615,11 @@ BOOL rdp_send_data_pdu(rdpRdp* rdp, wStream* s, BYTE type, UINT16 channel_id)
 BOOL rdp_send_message_channel_pdu(rdpRdp* rdp, wStream* s, UINT16 sec_flags)
 {
 	UINT16 length;
-	UINT32 sec_bytes;
-	size_t sec_hold;
 	UINT32 pad;
 
 	length = Stream_GetPosition(s);
 	Stream_SetPosition(s, 0);
 	rdp_write_header(rdp, s, length, rdp->mcs->messageChannelId);
-	sec_bytes = rdp_get_sec_bytes(rdp, sec_flags);
-	sec_hold = Stream_GetPosition(s);
-	Stream_Seek(s, sec_bytes);
-	Stream_SetPosition(s, sec_hold);
 
 	if (!rdp_security_stream_out(rdp, s, length, sec_flags, &pad))
 		return FALSE;
--- a/libfreerdp/core/security.c
+++ b/libfreerdp/core/security.c
@@ -524,9 +524,9 @@ static void fips_expand_key_bits(BYTE* in, BYTE* out)
 		p = b / 8;
 		r = b % 8;
 
-		if (r == 0)
+		if (r <= 1)
 		{
-			out[i] = buf[p] & 0xfe;
+			out[i] = (buf[p] << r) & 0xfe;
 		}
 		else
 		{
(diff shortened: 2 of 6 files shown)


libfreerdp/crypto/per.c
```diff
@@ -328,8 +329,8 @@ BOOL per_read_object_identifier(wStream* s, BYTE oid[6])
 		return FALSE;
 
 	Stream_Read_UINT8(s, t12); /* first two tuples */
-	a_oid[0] = (t12 >> 4);
```
akallabeth: Where do you get the change from the shift by 4 (division by 16) to division by 40?

libfreerdp/crypto/per.c
```diff
@@ -352,11 +353,12 @@ BOOL per_read_object_identifier(wStream* s, BYTE oid[6])
  * Write PER OBJECT_IDENTIFIER (OID)
  * @param s stream
  * @param oid object identifier (oid)
+ * @warning It works correctly only for limited set of OIDs.
  */
 
 void per_write_object_identifier(wStream* s, BYTE oid[6])
 {
-	BYTE t12 = (oid[0] << 4) & (oid[1] & 0x0F);
+	BYTE t12 = oid[0] * 40 + oid[1];
```
akallabeth: Where do you get the change from the shift by 4 (division by 16) to division by 40?

libfreerdp/core/security.c
```diff
@@ -524,9 +524,9 @@ static void fips_expand_key_bits(BYTE* in, BYTE* out)
 		p = b / 8;
 		r = b % 8;
 
-		if (r == 0)
+		if (r <= 1)
 		{
-			out[i] = buf[p] & 0xfe;
+			out[i] = (buf[p] << r) & 0xfe;
```
akallabeth: Why did you add the shift here?

libfreerdp/crypto/per.c
mfleisz: Afaict this is described in https://msdn.microsoft.com/en-us/library/windows/desktop/bb540809 - the previous calculation with shifting seems to be wrong

libfreerdp/core/security.c
mfleisz: Good question! :)

libfreerdp/crypto/per.c
akallabeth: Ok, so that explains it, perfect.

libfreerdp/core/security.c
akallabeth: Ok, missed the commit message in overview mode, so this looks good too.

ondrejholy: In order to handle the case when r=1 (without access into buf[p+1])...

libfreerdp/crypto/per.c
ondrejholy: First tuple is encoded as 40m + n if m < 3 and n < 40, see:
https://msdn.microsoft.com/en-us/library/windows/desktop/bb540809
https://msdn.microsoft.com/en-us/library/cc240836.aspx

Sorry that I did not explain it fully in the commit message.

libfreerdp/core/security.c
mfleisz: Ah - alright - understood!

libfreerdp/core/rdp.c
```diff
@@ -615,17 +615,11 @@ BOOL rdp_send_data_pdu(rdpRdp* rdp, wStream* s, BYTE type, UINT16 channel_id)
...
-	size_t sec_hold;
 	UINT32 pad;
 
 	length = Stream_GetPosition(s);
 	Stream_SetPosition(s, 0);
 	rdp_write_header(rdp, s, length, rdp->mcs->messageChannelId);
-	sec_bytes = rdp_get_sec_bytes(rdp, sec_flags);
-	sec_hold = Stream_GetPosition(s);
-	Stream_Seek(s, sec_bytes);
-	Stream_SetPosition(s, sec_hold);
```
bmiklautz: Nice finding. Looks like a rather old copy 'n' paste error ;)

bmiklautz:
@ondrejholy nice one! Thank you.

PR #4322, opened 2017-12-20
Rail fixes for GFX mode

akallabeth:
* Added proper argument checks in RAILS channel
* Fixed screen update in GFX mode
* Fixed some missing NULL pointer checks

### DIFF
--- a/channels/rail/client/rail_main.c
+++ b/channels/rail/client/rail_main.c
@@ -38,6 +38,10 @@
 RailClientContext* rail_get_client_interface(railPlugin* rail)
 {
 	RailClientContext* pInterface;
+
+	if (!rail)
+		return NULL;
+
 	pInterface = (RailClientContext*) rail->channelEntryPoints.pInterface;
 	return pInterface;
 }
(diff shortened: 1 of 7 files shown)


channels/rail/client/rail_main.c
```diff
@@ -38,6 +38,10 @@
 RailClientContext* rail_get_client_interface(railPlugin* rail)
 {
 	RailClientContext* pInterface;
+
+	if (!rail)
+		return NULL;
+
```
hardening: Check not needed, see the stack call of this function, rail can't be NULL.

channels/rail/client/rail_main.c
```diff
@@ -124,13 +126,18 @@ static UINT rail_client_execute(RailClientContext* context,
...
+	rail = (railPlugin*) context->handle;
 	exeOrFile = exec->RemoteApplicationProgram;
 
 	if (!exeOrFile)
 		return ERROR_INVALID_PARAMETER;
 
-	if (strlen(exeOrFile) >= 2)
+	if (strnlen(exeOrFile, MAX_PATH) >= 2)
 	{
```
hardening: In this file it's the only pertinent change, all others are unneeded checks. 
Verify the call stacks and you'll see that the rail argument can never be NULL (or it would have failed a long before) and the same apply for the second parameter of each function (called with a ref on local stack variable).

akallabeth:
@freerdp-bot tes

akallabeth:
@freerdp-bot test

mfleisz:
@hardening @akallabeth 
Would be nice to get this merged - can the two of you come to an agreement concerning the checks.
I have no problem merging the PR as is ...

hardening:
My position is still that this PR is really fine with e23af1d removed from the history. Unneeded checks:

* take time at runtime;
* makes the code harder to read;
* give no protection at all;
* give the impression to the reader that the checks that are done indicate legitimate values. Or at least a feeling of self-protecting code, which is an indication of a lack of control of the callers;

The best image I have for that, is "do you take snow chains when going in summer in the Sahara ?"

akallabeth:
@hardening Removed the checks you referenced as the function is returning void and can not propagate the misuse.

mfleisz:
@hardening Is this one good to go now from your side?

PR #4328, opened 2017-12-22
Activated address sanitizer for nightly builds.

akallabeth:
Since we ship debug symbols with the nightly builds it would be nice to get proper stack traces on memory issues without users having to do anything but copy the console output.


akallabeth:
@hardening @bmiklautz Anyone interested in merging this? Will give nice stack traces for free to anyone installing the debug symbols ;)

PR #4338, opened 2018-01-07
Ensure HTML Format clipboard synthesizer always converts number strings as if they are in base 10

2asoft:
`StartHTML` and `EndHTML` values can be left-padded with an arbitrary number of the `0` character ( per https://msdn.microsoft.com/en-us/library/windows/desktop/ms649015(v=vs.85).aspx ).

`strtol` and friends treat this as base-8 if base is specified as `0`.

Because these values are always sent in base-10 (not stated but implied in the article linked above), the fix is to always use base-10.


ilammy:
Seems good. Nice catch.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot hey, new year and already not testing?

Issue #2006, opened 2014-07-28
Eliminate race conditions by using InitOnceExecuteOnce


nfedera:
FreeRDP has several potential race conditions in different places doing a singleton init that could in theory race (contrary to c++).

All of these initializers should be converted to use the newly available InitOnceExecuteOnce() function.

Please log all occurrences here in this issue.

PR #4340, opened 2018-01-08
Fix remaining issues with #2006

akallabeth:
See #2006


akallabeth:
@freerdp-bot test test test

Issue #4341, opened 2018-01-08
Make GFX mode default for windows 10 and up


akallabeth:
As mentioned in #4326 newer windows 10 versions do no longer support legacy RDP modes.
Change FreeRDP to announce GFX by default.

Unitl #4266 is resolved it may be a good idea to only announce 10.2 capabilities by default.

lemonmojo:
FYI: As mentioned in https://github.com/FreeRDP/FreeRDP/issues/4326, newer Insider builds (I tested 17093 and 17101) don't require egfx anymore, so it might make sense to postpone switching GFX on by default.

akallabeth:
Fixed with https://github.com/FreeRDP/FreeRDP/pull/6786

PR #4342, opened 2018-01-08
New defaults for command line

akallabeth:
* Fix a bug in the command line parser not using defaults provided
* Add a function to check, if a specific parameter has a value that needs to be evaluated.
* Added GFX capability announcement parameter
* Windows 7 RFX mode now enabled by default
* GFX mode now enabled by default (caps 10, 10.2 and 10.3, windows 10 only)
* By default enable font smoothing

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -48,6 +48,7 @@
 #include <freerdp/log.h>
 #define TAG CLIENT_TAG("common.cmdline")
 
+
 BOOL freerdp_client_print_version(void)
 {
 	printf("This is FreeRDP version %s (%s)\n", FREERDP_VERSION_FULL,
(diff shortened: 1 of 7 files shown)


akallabeth:
Fixes #4341

client/common/cmdline.c
```diff
@@ -43,11 +44,168 @@
 #include <freerdp/version.h>
 
 #include "compatibility.h"
-#include "cmdline.h"
 
 #include <freerdp/log.h>
 #define TAG CLIENT_TAG("common.cmdline")
 
+static COMMAND_LINE_ARGUMENT_A args[] =
```
hardening: Isn't that array defined in cmdline.h ?

akallabeth: oh, seems like an error in rebasing :/

PR #4350, opened 2018-01-10
New code formatting rules

akallabeth:
This pull contains a new code formatting utilities:
* a script for clang-format
* a CMake target to reformat the project
* Removes the old ```astyle``` based scripts

Rationale:
* Due to some deficiencies of ```astyle``` formatting I want to change to this one.
* The script settings try to follow the current style of the project as good as possible.
* It contains rules for all languages used in the project (C, Java, ObjectiveC)
* Please have a look at the reformatted code and comment with style improvements/wishes
  * When the style has been accepted, the whole project will be reformatted with it once and for all
  * The pull request tester will be adjusted to check for style violations on new contributions from that point on


akallabeth:
@freerdp-bot test please!

hardening:
Here I get some indent of 2 with clang-format 3.8, which looks too small.
I've also noticed that curly braces style has been changed to
```C
if (something) {
   // do something
}
```
while it was 
```C
if (something) 
{
   // do something
}
```
before.

I personally prefer the first form, but I'm not sure it's a shared opinion here...

hardening:
Don't know if it's clang-format but I got strange things in libfreerdp/core/capabilities.c:
```C
static const char *const CAPSET_TYPE_STRINGS[] = {"Unknown",
                                                  "General",
                                                  "Bitmap",
                                                  "Order",
```
in one place and
```C
static const GUID CODEC_GUID_REMOTEFX = {
    0x76772F12,
    0xBD72,
    0x4463,
    {0xAF, 0xB3, 0xB7, 0x3C, 0x9C, 0x6F, 0x78, 0x86}};
```
in another. 
The second one has my preference, I guess it's because of the long line that the alignment is nicer. Don't know if we can do something for that...

In this file I've also looked at the numerous lines with ```Stream_Read() /* comment */```, there's a few damages but it looks acceptable.

akallabeth:
@hardening Ok, just realized there are one or two options not supported on older clang-format versions in the config. Will update accordingly.

hardening:
@akallabeth the solution can also be to force the clang-format version...

akallabeth:
@hardening true, but the stuff you've seen is the clang-format default style (not the one from the config) as it is ignored due to unsupported options.
only issue is that the options I've been playing around with occurred only recently (that is 5 or 6) :/

akallabeth:
@hardening Ok, did some tests, looks like only clang-format-6.0 is currently capable of providing a coding style that resembles something that is similar to ours.

bmiklautz:
@hardening the coding style should stay the same as it is now. 
@akallabeth clang 6 is still under development and might change, what's the drawbacks of using 5?

akallabeth:
@bmiklautz Some options do not exist:
* ```AfterExternBlock: true``` This one is the most visible, all global variables will have the opening brace on the same line.
* ```IncludeBlocks```
* ```IndentPPDirectives```
* ```RawStringFormats```

bmiklautz:
@hardening would you re-try your tests with clang 6 (released and stable ;) if you agree?

bmiklautz:
My suggestion would be to use these https://github.com/bmiklautz/FreeRDP/blob/clang-format/.clang-format settings. Also have a look at commit bmiklautz/FreeRDP@56181ca98146ef6e6913ba8dab4af8c961901328 that has the format already applied to see what changes it would introduce.

I did some further testing and those settings showed the least "impact" on the whole code base (~80k lines)

Some remarks:
* The line length isn't set on purpose - changing it to 120 would introduce 10k and to 100 20k more changes. - I personally don't mind either length and see it more as "soft" requirement and nothing that should be enforced.
* In case of do {} while() the while is kept on the same line - haven't found any setting to get around that - that concerns about 200 lines
* For formatting macros, like PRIu32, extra spaces are put in front and after (around 3000 places). Here I haven't found any setting to prevent that either. It looks a little bit odd but isn't "wrong". Example:
```
-			printf("%02"PRIX8"", ClientAddress->Address[i]);
+			printf("%02" PRIX8 "", ClientAddress->Address[i]);
```
* A little bit annoying, but currently a missing feature of clang-format, are define indentations:
```
#define hugo     ABCD
#define oh         EFGH
```
gets re-formatted to
```
#define hugo ABCD
#define oh EFGH
```
* There are definitely some other places where the applied format looks odd or even wrong but those can be fixed on the go by disabling clang-format and formatting manually.

I think it's worth to live with those little "quirks" in favor of having a common, apply able and testable style.

@mfleisz @nfedera @MartinHaimberger what's your thinking on this?

MartinHaimberger:
Overall a good idea, but the -> and logic statements look a little bit wired.

bmiklautz:
@mfleisz good point. **BreakBeforeBinaryOperators** was set to All but should be None. Yes that's one of the quirks I was talking about. Current clang format version do add the " " before (and possibly) after a macro (what the CommandLineSwitchCase is)
@MartinHaimberger ^^

mfleisz:
@bmiklautz Alright then - if we change ```BreakBeforeBinaryOperators = None``` I'm happy with the formatting (the other minor quirks can be fixed later manually) 👍

bmiklautz:
@mfleisz @akallabeth - re applied against master with 120 columns and  BreakBeforeBinaryOperators = None bmiklautz/FreeRDP@7a5fbf5224eb486cfaf26233e967a1eed940fd70

akallabeth:
:+1

mfleisz:
@bmiklautz Did you change anything else with the line breaks too? The macro stuff looks like this now:
```
		CommandLineSwitchStart(arg) CommandLineSwitchCase(arg, "dev")
```
and some other parts look weird as well ... 🤔

bmiklautz:
@mfleisz that's basically because of the line size .. If it seems that clang-format tries to use it as mach as possible. And as our macros don't have a ';' they are not identified.. So I quirk for us - but cmdline.h is definitely something where we want to keep the manual formatting. Once we have settled on a format I'll go through that.

mfleisz:
@bmiklautz Ah yes - that makes sense - alright :)

akallabeth:
@bmiklautz ok, would be great if you could provide the (modified) formatting script you have and we finally merge this one.

hardening:
The PR looks good, I think it is just lacking a cmake option directive so that we see it when doing a ccmake.
I'm also seeing a compilation breakage after code reformatting probably because of includes reordering.

akallabeth:
@hardening If you´re happy with the current `clang-format` settings feel free to merge this one.

hardening:
@akallabeth I will merge it after you have added cmake option for `WITH_CLANG_FORMAT` so that it's not an hidden option.

akallabeth:
@hardening ok, done ;)

PR #4355, opened 2018-01-11
Fix smartcard login by ignoring Status_Call.cbAtrLen as is required by [MS-RDPESC]

jukeks:
The [[MS-RDPESC]](https://msdn.microsoft.com/en-us/library/cc242596.aspx) clearly states in section [2.2.2.18](https://msdn.microsoft.com/en-us/library/cc242813.aspx) that Status_Call's cbAtrLen must not be used when generating Status_Return response. This is also how FreeRDP 1.1 behaves. This fixes logging in with smartcards.


hardening:

PR #4364, opened 2018-01-15
Gateway refactor

akallabeth:
* Made functions static where possible
* Fixed various memory leaks
* Added response size limit for HTTP parser
* Use a single buffer for a http response


akallabeth:
Anyone merging this one?

PR #4376, opened 2018-01-23
Implement MS-RDPEVOR

hardening:
This PR proposes an implementation of MS-RDPEVOR. It changes the API of the geometry plugin to make it more convenient for other plugin to interact with it.
The video packets are reassembled and decoded (only H264). Then rendering is done using a new YUV renderer, that may use multiple threads to do the work (each thread doing the YUV420->RGB conversion on a small part of the picture).
A very trivial feedback algorithm is implemented to inform the server about the decoding capabilities.
To test: use WMP (don't know minimal version) with at least a 2012 server, download your favourite trailer and enjoy.
TODO: the spec suggests that we should be able to synchronize audio and video using the provided timestamps but I've not dig the subject.

### DIFF
--- a/channels/drdynvc/client/drdynvc_main.c
+++ b/channels/drdynvc/client/drdynvc_main.c
@@ -706,7 +706,9 @@ static UINT drdynvc_send(drdynvcPlugin* drdynvc, wStream* s)
 	switch (status)
 	{
 		case CHANNEL_RC_OK:
+			return CHANNEL_RC_OK;
 		case CHANNEL_RC_NOT_CONNECTED:
+			Stream_Free(s, TRUE);
 			return CHANNEL_RC_OK;
 
 		default:
@@ -839,8 +841,7 @@ static UINT drdynvc_send_capability_response(drdynvcPlugin* drdynvc)
 		return CHANNEL_RC_NO_MEMORY;
 	}
 
-	Stream_Write_UINT16(s,
-	                    0x0050); /* Cmd+Sp+cbChId+Pad. Note: MSTSC sends 0x005c */
+	Stream_Write_UINT16(s, 0x0050); /* Cmd+Sp+cbChId+Pad. Note: MSTSC sends 0x005c */
 	Stream_Write_UINT16(s, drdynvc->version);
 	status = drdynvc_send(drdynvc, s);
 
@@ -1190,6 +1191,7 @@ static void VCAPITYPE drdynvc_virtual_channel_open_event_ex(LPVOID lpUserParam, 
 	if (!drdynvc || (drdynvc->OpenHandle != openHandle))
 	{
 		WLog_ERR(TAG, "drdynvc_virtual_channel_open_event: error no match");
+		Stream_Free((wStream*) pData, TRUE);
 		return;
 	}
 
--- a/client/X11/xf_video.c
+++ b/client/X11/xf_video.c
@@ -1,0 +1,631 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ * Video Optimized Remoting Virtual Channel Extension for X11
+ *
+ * Copyright 2017 David Fort <contact@hardening-consulting.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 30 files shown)


client/X11/xf_video.h
```diff
@@ -0,0 +1,39 @@
...
+struct _xfVideoContext;
+typedef struct _xfVideoContext xfVideoContext;
+
+void xf_video_geometry_init(xfContext *xfc, GeometryClientContext *geom);
+void xf_video_control_init(xfContext *xfc, VideoClientContext *video);
+
+void xf_video_data_init(xfContext *xfc, VideoClientContext *video);
+void xf_video_data_uninit(xfVideoContext *xfVideo);
+
+xfVideoContext *xf_video_new();
```
akallabeth: missing ```void```

hardening:
I've fixed a bug found by @bmiklautz, anyway when compiled with ASAN I'm getting strange error like if there were a concurrency issue.

hardening:
@bmiklautz @akallabeth with the latest changes the leaks and segfaults are gone

hardening:
@akallabeth well I think I have these leaks without my PR (at least here ASAN reports lost bytes with WLog and other places)...

akallabeth:
@hardening No, specifically the video channel new and free functions in pre/post connect and disconnect.

akallabeth:
@hardening Yes, sure of that. The interesting part is that the leaks only occur when NOT adding ```/video``` but only ```/gfx```

akallabeth:
@hardening got it, you call ```xf_video_new``` in ```xf_post_connect``` but you don't clean that up.

client/X11/xf_video.c
```diff
@@ -652,6 +655,7 @@ void xf_video_control_init(xfContext *xfc, VideoClientContext *video)
 void xf_video_control_uninit(xfContext *xfc, VideoClientContext *video)
 {
 	xf_video_free(xfc->xfVideo);
+	xfc->xfVideo = NULL;
```
akallabeth: Free here is unnecessary, post_disconnect is always called.
Or could the whole stuff be moved to the control init/uninit avoiding runtime impact when not used?

channels/drdynvc/client/drdynvc_main.c
```diff
@@ -706,7 +706,9 @@ static UINT drdynvc_send(drdynvcPlugin* drdynvc, wStream* s)
 	switch (status)
 	{
 		case CHANNEL_RC_OK:
+			return CHANNEL_RC_OK;
 		case CHANNEL_RC_NOT_CONNECTED:
+			Stream_Free(s, TRUE);
```
akallabeth: nice catch

channels/drdynvc/client/drdynvc_main.c
```diff
@@ -1187,6 +1188,7 @@ static void VCAPITYPE drdynvc_virtual_channel_open_event_ex(LPVOID lpUserParam,
 	if (!drdynvc || (drdynvc->OpenHandle != openHandle))
 	{
 		WLog_ERR(TAG, "drdynvc_virtual_channel_open_event: error no match");
+		Stream_Free((wStream*) pData, TRUE);
 		return;
```
hardening: @akallabeth in fact the leak was here. In this function we have ownership on the stream, so we must release it.

T3N4K4:
Basically it seems to be working with all "video"-material and is not only based on WMP.
When I play a video (without fullscreen) in internet explorer, edge or vlc-player after a while a pop-out occurs in which the optimized video redirection takes place. Unfortunately the pop-out is not on top of the original video.

akallabeth:
@hardening Does not work with `/smart-sizing`

akallabeth:
Would also be nice to have a generic implementation like with GFX that does not depend on X11

akallabeth:
@hardening Sorry for the late feedback.
Tested your current branch, the following issues still apply:
* I can see an offset of the video if I start youtube in chrome (`/gfx /video`)
* The video is not scaling (`/gfx /video /smart-sizing`)
* `/gdi:sw` is ignored, as in it still calls the X11 client specific code and does not fall back to copying the video data to the gdi framebuffer

hardening:
Yeah there's an issue with geometry, I'm currently on it.

akallabeth:
@hardening Another issue, clang-scanbuild tells there is a use after free in `VideoFrame_free`

hardening:
@akallabeth yep, already fixed here

akallabeth:
@hardening Fixed a few issues I found, but the geometry stuff is still not really working.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@hardening Did some further testing, looks like we miss some `FREERDP_API` prefixes in exported header files.
Other than that, are you still looking into unifying GFX and VIDEO surfaces?

hardening:
@akallabeth unifying => nope not for now. I guess we could merge that PR and unify things after.

akallabeth:
@hardening added the missing header export, could you do the rebase? Don't want to screw that one up ^^

PR #4380, opened 2018-01-24
  Fixed SuppressOutput update.

### LINKED ISSUE #4371: black screen with gfx and suppress output in hardware gdi
When hardware GDI is enabled and gfx used  supress output leads
to a black screen until a update would happen from the server.

Happens on 2.0.0-dev2.

Steps to reproduce:
* run xfreerdp /gdi:hw /gfx 
* switch workspace
* switch back
* as result xfreerdp stays black

### LINKED ISSUE #4377: Client desktop taskbar stays on top of session when xfreerdp toggles full-screen after maximize
Toggle full-screen using CTRL-ALT-ENTER and then minimizing the session works. However, when restoring the session and then toggling full-screen again shows the client taskbar on top of the remote session. Using the Layer menu of the UI decorations and forcing it to be on top solves the problem. However, this is not the correct behavior. 

Latest master git version does this. I have went back to a known working git version:
FreeRDP-097f8edd97735bc8df4bce6ee1605a4ac16b20e3

Lubuntu 14.04.5/16.04.3, amd64, openbox/xfce

/usr/local/freerdp/bin/xfreerdp /gfx +gfx-progressive /u:user /p:xyz -themes -wallpaper /f +clipboard /d:DOMAIN /sec:nla /cert-ignore /v:RDS-SERVER

Connecting to W2K12R2 RDS Server

This issue is also referenced in #4371 because the GFX mods causing the black screen may have something to do with this.

akallabeth:
* Fixes #4371: Sending suppress output PDU whenever the window state changes

NOTE: As a side effect the output is suppressed when minimized and no preview is shown (e.g. cinnamon taskbar)

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -289,8 +289,6 @@ static BOOL xf_desktop_resize(rdpContext* context)
 static BOOL xf_sw_begin_paint(rdpContext* context)
 {
 	rdpGdi* gdi = context->gdi;
-	gdi->primary->hdc->hwnd->invalid->null = TRUE;
-	gdi->primary->hdc->hwnd->ninvalid = 0;
 	return TRUE;
 }
 
@@ -303,6 +301,10 @@ static BOOL xf_sw_end_paint(rdpContext* context)
 	HGDI_RGN cinvalid;
 	xfContext* xfc = (xfContext*) context;
 	rdpGdi* gdi = context->gdi;
+
+	if (gdi->suppressOutput)
+		return TRUE;
+
 	x = gdi->primary->hdc->hwnd->invalid->x;
 	y = gdi->primary->hdc->hwnd->invalid->y;
 	w = gdi->primary->hdc->hwnd->invalid->w;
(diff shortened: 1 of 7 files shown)


client/X11/xf_event.c
```diff
@@ -41,51 +41,178 @@
...
 #endif
 
+static BOOL xf_send_suppress_output(xfContext* xfc, XEvent* event, BOOL suppress)
+{
+	RECTANGLE_16 rect;
+	rdpSettings* settings;
+	rdpUpdate* update;
+
+	if (!xfc || !xfc->context.settings || !xfc->context.update)
+		return FALSE;
```
hardening: I don't see how xfc could be null here, not sure this check is necessary

client/X11/xf_event.c
```diff
@@ -41,51 +41,178 @@
...
-	"GenericEvent",
-};
+	switch (event)
+	{
+		case KeyPress:
+			return "KeyPress";
+
+		case KeyRelease:
+			return
+			    "KeyRelease";
```
mfleisz: Formatting looks weird - why put each string on a separate line?

akallabeth: @mfleisz you're right, fixed.

client/X11/xf_client.c
```diff
@@ -289,8 +289,6 @@ static BOOL xf_desktop_resize(rdpContext* context)
 static BOOL xf_sw_begin_paint(rdpContext* context)
 {
 	rdpGdi* gdi = context->gdi;
```
nfedera: unused variable

PR #4391, opened 2018-01-30
SSSE3 optimized RGB to AVC444 YUV conversion

akallabeth:
* Split optimized YUV functions to separate files
* Added optimized  ```RGBToAVC444YUVv2```
* Fixed generic ```RGBToAVC444YUVv2``` UV average for luma frame


akallabeth:
@hardening looks like I forgot to disable optimized primitive function calls if they are not enabled.

PR #4393, opened 2018-01-31
disabled expensive debug call in channel send

nfedera:
I suppose @hardening forgot to disable this before committing a1329223


hardening:
@nfedera you're damn right, I have even added some other debug statement that aren't useful for the general case.

nfedera:
@hardening allright. you might want to kick these in a separate pr?

PR #4397, opened 2018-02-01
Use libavcodec's VA-API to accelerate H.264 decoding.

SriRamanujam:
Leverages libavcodec's hw decode support to provide VA-API based hardware decoding. Depends on the local build of ffmpeg having hardware VA-API support compiled in and the appropriate libva drivers and libraries installed.

I've confirmed that ffmpeg is built with VA-API support in Ubuntu Trusty onwards and Fedora 26 onwards. I've gated this code behind an optional `WITH_VAAPI` build flag in any case, so it's easy to build with the feature turned off. 

I would appreciate guidance on how to determine the precise AV_VERSION macro guards, I'm not familiar with that aspect of developing against ffmpeg libraries and so I thought that rather than doing it wrong I'd ask for help and get it right the first time :) 

I thought that this code would get me AVC420 decoding as well, but for some reason it doesn't look like the codepath is followed with AVC420. Really not sure why that is.


hardening:
@freerdp-bot give it a test please

hardening:
@SriRamanujam I used to have the same kind of patch lying around and it was working fine with AVC420. IIRC you should try with Fedora 25 to get a version that needs more code to work.
Apart from that the code looks good to me (as it's almost the same as I had)

hardening:
That looks good to me, especially as I had the same piece of code lying around.
About macros at least 56.60.100 (as installed on the FreeRDP CI) doesn't have access to ```libavutils/hwcontext.h```.
You could make them work but these older versions requires a lot of tricks and ffmpeg internal interactions to use VAAPI, so limiting the version would be nice.
In my tests, AVC420 was working fine too with VAAPI.

SriRamanujam:
@hardening understood, I'll take another look at the AVC420 codepath and make sure I wasn't mistaken. 

I'll also dig through ffmpeg and see what version introduced `libavutils/hwcontext.h` and gate the code accordingly.

libfreerdp/codec/h264_ffmpeg.c
```diff
@@ -197,16 +209,54 @@ static int libavcodec_decompress(H264_CONTEXT* h264, const BYTE* pSrcData,
...
+		{
+			status = av_frame_copy(sys->videoFrame, sys->hwVideoFrame);
+		}
+
+		break;
+	}
+
+#else
+
 	do
```
hardening: Please keep the original loop add #ifdefs statement before and another after for the VAAPI case

SriRamanujam:
Okay, I've guarded `hwcontext.h` behind the version of libavutil that initially introduced VA-API support (https://github.com/FFmpeg/FFmpeg/commit/551c6775abb5e0ad34c26d7e23bc6fbbe8ccc9d4) and folded the code back into the existing loop structures.

akallabeth:
@freerdp-bot test

akallabeth:
@SriRamanujam Nice job btw

akallabeth:
Additional comment: May be a good idea to provide a fallback in case no VAAPI device is available during runtime

hardening:
@akallabeth  AFAICT the fallback is already implemented

libfreerdp/codec/h264_ffmpeg.c
```diff
@@ -415,6 +503,25 @@ static BOOL libavcodec_init(H264_CONTEXT* h264)
...
+
+		if (!sys->hwctx)
+		{
+			int ret = av_hwdevice_ctx_create(&sys->hwctx, AV_HWDEVICE_TYPE_VAAPI, VAAPI_DEVICE, NULL, 0);
+
+			if (ret < 0)
+			{
+				WLog_ERR(TAG, "Could not initialize hw decoder: %s", av_err2str(ret));
+				goto EXCEPTION;
+			}
```
hardening: here I think we should go to fallback without hardware support. Otherwise that means that if the permissions aren't correct for accessing the DRM device, we won't be able to run at all.

SriRamanujam:
@akallabeth 

Yeah, seems like the `hw_device_ctx` field was only added last year (https://github.com/FFmpeg/FFmpeg/commit/c1a5fca06f75cc0e7b9b2808fecaa0c1b424da50). WIthout it, life becomes more difficult as you have to manually set up the VA-API context and libav frames: https://lists.ffmpeg.org/pipermail/ffmpeg-user/2016-December/034530.html. The `struct vaapi_context` mentioned in the mailing list has been deprecated since 2016 as well (https://github.com/FFmpeg/FFmpeg/commit/2a4a8653b6ca77ed1c9cec1336c55d291ef545b6). I'll take a stab at setting it up this way and see how it goes. 

It might be easier, though, to gate the VA-API feature behind the libavcodec version that introduces `hw_device_ctx` and call it done. Let me know if that's okay. 

Regarding `find_feature`: should I extend `FindFFmpeg` or create a separate `FindVAAPI` that checks the libavcodec/libavutil versions specifically for VA-API support? I'm not sure how to check whether libavcodec was built with VA-API support through `find_feature`, if you have any ideas on that, I'd appreciate it.

SriRamanujam:
Further update: I managed to get the decoding going using `hw_frames_ctx` without having to manually create a `struct vaapi_context`, so that's good. Those changes are pushed, @akallabeth please let me know if it builds and works for you now. 

The only outstanding item now is the `find_feature` changes, which I'll start looking into now.

akallabeth:
@freerdp-bot test
@SriRamanujam Working quite fine here now, also when hardware acceleration is not available.

akallabeth:
@freerdp-bot come on, test!

bmiklautz:
@SriRamanujam very nice! Thanks.

PR #4411, opened 2018-02-08
Fixed IPv4/IPv6 to string

akallabeth:


cedrozor:
Nice, thanks :)

libfreerdp/core/tcp.c
```diff
@@ -667,19 +667,25 @@ BIO_METHOD* BIO_s_buffered_socket(void)
 static char* freerdp_tcp_get_ip_address(int sockfd, BOOL* pIPv6)
 {
 	socklen_t length;
-	char ipAddress[INET6_ADDRSTRLEN + 1];
-	struct sockaddr_in sockaddr;
-	length = sizeof(sockaddr);
-	ZeroMemory(&sockaddr, length);
+	char ipAddress[INET6_ADDRSTRLEN + 1] = { 0 };
+	struct sockaddr_in sockaddr = { 0 };
```
mfleisz: Shouldn't this be a sockaddr_in6 (which is most likely bigger in size than a sockaddr_in) or we might get a truncated address from getsockname? Same goes for the length assignment.

akallabeth:
@mfleisz Already updated, see http://pubs.opengroup.org/onlinepubs/000095399/basedefs/netinet/in.h.html

PR #4413, opened 2018-02-09
Memleak queue fixes

akallabeth:
* Fix a memory leak in gfx on error
* Now clearing message queue on free just as Hash and ArrayList
* dynamic channel message queue now frees objects still in the queue
* Clear the client channel queue and free data

### DIFF
--- a/winpr/libwinpr/utils/collections/MessageQueue.c
+++ b/winpr/libwinpr/utils/collections/MessageQueue.c
@@ -227,6 +227,8 @@ void MessageQueue_Free(wMessageQueue* queue)
 	if (!queue)
 		return;
 
+	MessageQueue_Clear(queue);
+
 	CloseHandle(queue->event);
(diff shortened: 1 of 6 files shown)


winpr/libwinpr/utils/collections/MessageQueue.c
```diff
@@ -227,6 +227,8 @@ void MessageQueue_Free(wMessageQueue* queue)
 	if (!queue)
 		return;
 
+	MessageQueue_Clear(queue);
+
```
hardening: good catch

hardening:
@akallabeth just a matter of taste but I prefer
```
status = op()
if (status)
  .....
```
over
```
status = op()

if (status)
  .....
```

IMHO it looks better when the operation and the test are grouped in the same block.

akallabeth:
@mfleisz Already fixed.

mfleisz:
@freerdp-bot test

PR #4414, opened 2018-02-09
Pthread cleanup fix

akallabeth:
`pthread_once` destructor functions are only called when a thread exits with `pthread_exit`.
This was not true for the main application thread leading to noise in valigrind and address sanitizer

To resolve this issue, a new function `winpr_exit` is introduced that must be called on main thread exit.
* It deletes the memory allocated by WLog
* It triggers the pthread destructors

While not strictly necessary for production use, it greatly improves the usefullness of valgrind/address sanitizer by reducing false positives.

### DIFF
--- a/client/DirectFB/dfreerdp.c
+++ b/client/DirectFB/dfreerdp.c
@@ -422,7 +422,7 @@ int main(int argc, char* argv[])
 	if (!(g_sem = CreateSemaphore(NULL, 0, 1, NULL)))
 	{
 		WLog_ERR(TAG, "Failed to create semaphore");
-		exit(1);
+		return winpr_exit(1);
 	}
 
 	instance = freerdp_new();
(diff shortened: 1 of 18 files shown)


llogar:
Calling `winpr_exit()` from main thread makes the process exit with exit status 0, thus the true exit code gets lost...

akallabeth:
@llogar Your comment doesn't make any sense. `winpr_exit` is just a wrapper around `pthread_exit` and will exit with whatever return code you supply.

llogar:
According to [http://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_exit.html](url) 

> The process exits with an exit status of 0 after the last thread has been terminated. The behaviour is as if the implementation called exit() with a zero argument at thread termination time

and this is what I am observing at least on Linux `xfreerdp` executable exits with status 0, regardless of the _dwExitCode_? 

Therefore it is my understanding that the main thread should not be terminated via _pthread_exit()_...

akallabeth:
Ah, I see what you mean, missed that part obviously. Need to call pthread_exit to trigger the cleanup handlers though...

Issue #2128, opened 2014-09-29
centos xfreerdp to win 7 fail: Failure in SSL library


lxia7212:
This is FreeRDP version 1.0.2

Linux ????.com 2.6.32-279.el6.x86_64 #1 SMP Fri Jun 22 12:19:21 UTC 2012 x86_64 x86_64 x86_64 GNU/Linux
# xfreerdp 192.168.1.27 -u someone -f --sec nla

connected to 192.168.1.27:3389
Password: 
SSL_read: Failure in SSL library (protocol error?)
Authentication failure, check credentials.
If credentials are valid, the NTLMSSP implementation may be to blame.

Is there anyone can help me?

relphj:
I get the same failure:

$ xfreerdp --version
This is FreeRDP version 1.0.2

$ xfreerdp --sec nla win7host.nodc.noaa.gov
connected to win7host.nodc.noaa.gov:3389
Password:
SSL_read: Failure in SSL library (protocol error?)
Authentication failure, check credentials.
If credentials are valid, the NTLMSSP implementation may be to blame.

(By the way "--sec nla" is the correct form for xfreerdp 1.0.2.)

If I use the same command to a Windows Server 2003 system, it works fine.

relphj:
I am successfully able to connect to my Windows 7 host from another Windows 7 client using Remote Desktop Connection. The Windows 7 host has NLA enabled, no TLS or other authentication methods are enabled. This appears to be a reproducible problem in xfreerdp.

aroundthfur:
I have the same issue. 

$ xfreerdp /u:jovanam /p:Memory09 /v:hostname
connected to hostname:3389
SSL_read: Failure in SSL library (protocol error?)
SSL_read: error:14094419:SSL routines:SSL3_READ_BYTES:tlsv1 alert access denied
credssp_recv() error: -1
Authentication failure, check credentials.
If credentials are valid, the NTLMSSP implementation may be to blame.
Error: protocol security negotiation or connection failure

Funny that some times it works and some times it does not. I can't seem to find any reason for the behaviour..

relphj:
I have tried using --no-tls, --ignore-certificate, and pretty much every other option I can think of. Still no connect to a Windows 7 system from centos using xfreerdp 1.0.2. (Does anybody read these reports?)

trinacriax:
[...]

PR #4417, opened 2018-02-09
tls: enable tls 1+

ondrejholy:
Currently TLS version 1.0 is used implicitly by using the TLSv1_method.
To be able to also use TLS 1.1 and later use SSLv23_client_method
instead. To make sure SSLv2 or SSLv3 isn't used disable them.

Commit aa80f63b from master was backported by Ondrej Holy <oholy@redhat.com>.

https://github.com/FreeRDP/FreeRDP/issues/2128

### DIFF
--- a/libfreerdp-core/tls.c
+++ b/libfreerdp-core/tls.c
@@ -25,8 +25,9 @@
 boolean tls_connect(rdpTls* tls)
 {
 	int connection_status;
+	int options = 0;
 
-	tls->ctx = SSL_CTX_new(TLSv1_client_method());
+	tls->ctx = SSL_CTX_new(SSLv23_client_method());
 
 	if (tls->ctx == NULL)
 	{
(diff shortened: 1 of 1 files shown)


TomerAdmon:
@ondrejholy  - Thanks! works perfectly!
Tested both CentOS and Ubuntu with Windows 10 and Windows 2016.
Is it possible to merge this code also to the 1.1-stable branch?
@akallabeth - Can you please add the stable-1.1 tag as well? 
👍 👍 👍

ondrejholy:
I have been under the impression that stable-1.1 has never been officially released, nor supported... @akallabeth, is it worth to create new merge request?

bmiklautz:
@ondrejholy 1.1 was actually never left the development state (partially because of library naming and conflicts) however if you in mood to create a PR we'll merge since 1.1 is used quite a lot.
For the same reason we'll also merge this to 1.0 .

Thanks for your effort.

bmiklautz:
Pull request tester doesn't work with 1.0 and 1.1. Tested build manually.

PR #4426, opened 2018-02-13
Check if xfreerdp window != NULL

akallabeth:
When xfreerdp aborts a connection it may be before xfc->window was allocated.
Don't access it in that case.


akallabeth:
@freerdp-bot test and stop fooling around!

PR #4432, opened 2018-02-14
Fixed PubSub function pointer casts.

akallabeth:
Removed all function pointer casts for the `PubSub` system to allow the compiler to warn us about argument mismatches.


include/freerdp/client/geometry.h
```diff
@@ -61,7 +61,7 @@ struct _geometry_client_context
 };
 
 #ifdef __cplusplus
-extern "C"
+extern "C" {
 #endif
```
hardening: Good catch !

PR #4434, opened 2018-02-14
Fixed winpr digest type.

akallabeth:
there is an enum for the digest types. use it


akallabeth:
@freerdp-bot care to really test?

akallabeth:
@freerdp-bot test

PR #4440, opened 2018-02-15
Disable unused variable warning from profiler.

akallabeth:
Get rid of these annoying unused variable warnings and not break VS2010


mfleisz:
@freerdp-bot test

PR #4437, opened 2018-02-14
Support for gatewayaccesstoken / PAA

jaccobraat:
I've been testing with a F5 APM appliance for remote access to a Microsoft VDI farm (W10 desktops)
MS RDP client on W10 and OSX work fine. On Linux I could not get it working, and found this was caused by lack of PAA support. These patches add PAA to FreeRDP.

Add option for "gatewayaccesstoken:s:<token>" in rdp file.
Add option for "/gat:<token>" command-line option

Changes to RD Gateway connection setup flow:
Select PAA Extendend Authentication mode when gatewayaccesstoken is present.
Skip NTLM authentication on RDG CHANNELS, request PAA authentication.
Add PAA token to TUNNEL_CREATE packet

PR #4441, opened 2018-02-15
[cleanup] Support for gatewayaccesstoken / PAA

akallabeth:
This is a cleaned up version of #4437 
(fixup commits split/merged to original commits, rebased to master)
@jaccobraat thank you for the original contribution.


client/common/file.c
```diff
@@ -893,6 +896,12 @@ BOOL freerdp_client_populate_settings_from_rdp_file(rdpFile* file, rdpSettings*
 		free(host);
 	}
 
+	if (~((size_t) file->GatewayAccessToken))
+	{
```
hardening: I can't understand what is tested here ?

akallabeth: @hardening this checks if `file->GatewayAccessToken != 0xFFFFFFFF` (the initial value only changed if a new value is set)
Looks strange, but is used all over the place in rdp file parser.

mfleisz: That code really isn't a beauty ... but as @akallabeth mentioned it's all over the rdp parser ... maybe one day we should "fix" that ugliness in a separate PR

PR #4450, opened 2018-02-19
Assistance and RDP file parser fix

akallabeth:
Fixed .rdp and .msrcIncident checks.

The command line detection fails, if only one of the aforementioned
files is used as an argument. Detect those first and ignore command
line detection if found.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -1294,12 +1294,25 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 	char* str;
 	size_t length;
 	int status;
-	DWORD flags;
+	BOOL ext = FALSE;
+	BOOL assist = FALSE;
+	DWORD flags = 0;
 	BOOL promptForPassword = FALSE;
-	BOOL compatibility;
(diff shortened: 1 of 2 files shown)


jaccobraat:
this changes the behavior of the client. a command line with a .rdp file and additional arguments is not allowed any more.

DMWORD flags = 0;

still 0 when you get to:
		CommandLineClearArgumentsA(args);
		status = CommandLineParseArgumentsA(argc, (const char**) argv, args, flags,
		                                    settings,
		                                    freerdp_client_command_line_pre_filter,
		                                    freerdp_client_command_line_post_filter);


should be something like this from freerdp_client_detect_command_line:
	/* Default is POSIX syntax */
	*flags = COMMAND_LINE_SEPARATOR_SPACE;
	*flags |= COMMAND_LINE_SIGIL_DASH | COMMAND_LINE_SIGIL_DOUBLE_DASH;
	*flags |= COMMAND_LINE_SIGIL_ENABLE_DISABLE;

akallabeth:
@jaccobraat See #4477

PR #4455, opened 2018-02-22
[WIP] Changed default SSL error, added additional log messages.

akallabeth:
* The default of `FREERDP_ERROR_CONNECT_PASSWORD_CERTAINLY_EXPIRED` was very misleading, falling back to `FREERDP_ERROR_CONNECT_TRANSPORT_FAILED` as that is the only thing we know for certain.
* Added debug log messages with SSL error states for additioal information
* Fixed registration of calback, `BIO_callback_ctrl` uses a different callback function type, use proper `SSL_CTX_set_info_callback`

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -152,11 +152,25 @@ out:
 }
 #endif /* WITH_GSSAPI */
 
-static void transport_ssl_cb(SSL* ssl, int where, int ret)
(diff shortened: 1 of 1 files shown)


byteboon:
Just wondering why this wasn't merged to master... the code seems correct, and is certainly desirable. After a failed freerdp_connect, with this patch connectErrorCode will be CONNECTERROR whereas currently connectErrorCode is zero.

PR #4457, opened 2018-02-27
Fix /size: /w: /h: with /monitors: (Fix custom sizes)

T3N4K4:
I don't know if that was planned or part of a rework, but I noticed that you could not use /size:WxH or /w:W or /h:H with /monitors:0 or 1 or x anymore, so that a custom sized window will be rendered on a specific monitor. Instead it would be rendered where the mouse cursor is, even though this should be a fallback if /monitors:x is not used I guess.

This PR should fix that problem.

### DIFF
--- a/client/X11/xf_monitor.c
+++ b/client/X11/xf_monitor.c
@@ -4,6 +4,7 @@
  *
  * Copyright 2011 Marc-Andre Moreau <marcandre.moreau@gmail.com>
  * Copyright 2017 David Fort <contact@hardening-consulting.com>
+ * Copyright 2018 Kai Harms <kharms@rangee.com>
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

T3N4K4:
Okay, I may have found another problem.

akallabeth:
@T3N4K4 Sorry to ask, but from your last comment, is this pull ready for merge or are you working on a update?

T3N4K4:
@akallabeth I try to work on an update but I still don't know if I am on the right file. /size: or /w: or /h: with /monitors: is working again with that pull, but without /monitors: /size: is not behaving like it should.

bmiklautz:
@T3N4K4 what are the open issues you have?

T3N4K4:
@bmiklautz when I use /size: without /monitors: I get a weird rectangle window which is not what it should be.
For example /size:1024x768 /monitors:0 is working fine. /size:1024x768 shows a rectangle window over both monitors.. like freerdp renders one part on the first screen and the other part on the second screen.
But I am stuck on that.

client/X11/xf_monitor.c
```diff
@@ -281,8 +282,15 @@ BOOL xf_detect_monitors(xfContext* xfc, UINT32* pMaxWidth, UINT32* pMaxHeight)
...
+	else if (settings->DesktopWidth || settings->DesktopHeight)
+	{
+		if (settings->DesktopWidth)
+			*pMaxWidth = settings->DesktopWidth;
+		if (settings->DesktopHeight)
+			*pMaxHeight = settings->DesktopHeight;
+	}
 
-	if (!settings->Fullscreen && !settings->Workarea && !settings->UseMultimon)
+	if (!settings->Fullscreen && !settings->Workarea && !settings->UseMultimon && !(settings->DesktopWidth || settings->DesktopHeight))
```
akallabeth: @T3N4K4 This is the culprit for your weired rectangle, the new checks are wrong.

T3N4K4:
@freerdp-bot test

T3N4K4:
@akallabeth you were right. My own mistake. I removed the checks now and it worked like it should I guess. Tested /size:hxw, /h:, /w:, /h: /w:, /size:%%.

akallabeth:
@freerdp-bot test

PR #4464, opened 2018-03-02
Misc changes for Windows client

perkerk:
These are fixes from creating a client based on wfreerdp. The H.264 codec changes here handle just not trapping if the mfplat.dll load fails, which happened testing this on a Windows N system

### DIFF
--- a/client/Windows/wf_event.c
+++ b/client/Windows/wf_event.c
@@ -567,7 +567,7 @@ LRESULT CALLBACK wf_event_proc(HWND hWnd, UINT Msg, WPARAM wParam,
 			break;
 
 		case WM_SETCURSOR:
-			if (LOWORD(lParam) == HTCLIENT)
+			if (wfc && LOWORD(lParam) == HTCLIENT)
 				SetCursor(wfc->hDefaultCursor);
 			else
 				DefWindowProc(hWnd, Msg, wParam, lParam);
--- a/client/Windows/wf_event.c
+++ b/client/Windows/wf_event.c
@@ -566,14 +566,6 @@ LRESULT CALLBACK wf_event_proc(HWND hWnd, UINT Msg, WPARAM wParam,
 			PostQuitMessage(WM_QUIT);
 			break;
 
-		case WM_SETCURSOR:
-			if (wfc && LOWORD(lParam) == HTCLIENT)
-				SetCursor(wfc->hDefaultCursor);
-			else
-				DefWindowProc(hWnd, Msg, wParam, lParam);
-
-			break;
-
 		case WM_SETFOCUS:
 			DEBUG_KBD("getting focus %X", hWnd);
 
--- a/client/Windows/wf_graphics.c
+++ b/client/Windows/wf_graphics.c
@@ -131,11 +131,8 @@ static void wf_Bitmap_Free(rdpContext* context, rdpBitmap* bitmap)
 		DeleteObject(wf_bitmap->bitmap);
 		DeleteDC(wf_bitmap->hdc);
 
-		if (wf_bitmap->_bitmap.data)
-		{
-			_aligned_free(wf_bitmap->_bitmap.data);
-			wf_bitmap->_bitmap.data = NULL;
-		}
+		_aligned_free(wf_bitmap->_bitmap.data);
+		wf_bitmap->_bitmap.data = NULL;
 	}
 }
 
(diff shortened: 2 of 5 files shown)


akallabeth:
@freerdp-bot test

client/Windows/wf_event.c
```diff
@@ -567,7 +567,7 @@ LRESULT CALLBACK wf_event_proc(HWND hWnd, UINT Msg, WPARAM wParam,
 			break;
 
 		case WM_SETCURSOR:
-			if (LOWORD(lParam) == HTCLIENT)
+			if (wfc && LOWORD(lParam) == HTCLIENT)
```
akallabeth: Sure the else branch is still correct?

perkerk: I was just trying to avoid the exception, but in looking at this, the WM_SETCURSOR case where wfc != NULL has already been handled above, so this will never do anything. Plus hDefaultCursor is never referenced anywhere else, and even if it was, we wouldn't be able to access it here when wfc is NULL so I'll just take out this entire case statement

perkerk:
@akallabeth Added in one more fix I ran across

akallabeth:
@perkerk Could you mark your pull with `[WIP]` until you have finished it?

perkerk:
@akallabeth last commit was it, nothing more for now

akallabeth:
@freerdp-bot test

libfreerdp/codec/h264_mf.c
```diff
@@ -494,12 +494,15 @@ static BOOL mf_init(H264_CONTEXT* h264)
 	if (!sys)
 		goto error;
 
-	h264->pSystemData = (void*) sys;
 	/* http://decklink-sdk-delphi.googlecode.com/svn/trunk/Blackmagic%20DeckLink%20SDK%209.7/Win/Samples/Streaming/StreamingPreview/DecoderMF.cpp */
 	sys->mfplat = LoadLibraryA("mfplat.dll");
-
 	if (!sys->mfplat)
+	{
```
akallabeth: What was this change necessary for?
`sys` was cleaned up in `mf_uninit` below.

client/Windows/wf_graphics.c
```diff
@@ -130,6 +130,12 @@ static void wf_Bitmap_Free(rdpContext* context, rdpBitmap* bitmap)
 		SelectObject(wf_bitmap->hdc, wf_bitmap->org_bitmap);
 		DeleteObject(wf_bitmap->bitmap);
 		DeleteDC(wf_bitmap->hdc);
+
+		if (wf_bitmap->_bitmap.data)
+		{
```
akallabeth: good catch

libfreerdp/codec/h264_mf.c
perkerk: before, if the block was allocated mf_uninit blindly tries to access some fields that don't end up getting initialized, MFShutdown for one

client/Windows/wf_graphics.c
perkerk: heh - not exactly a catch. my client plays back video and each successive playback was 25% or more slower than the last. only then did I see the memory growing

libfreerdp/codec/h264_mf.c
mfleisz: The free() here is not the way to go imho.
The problem is the current mf_uninit implementation:
- First it should not call MFShutdown without checking if the pointer is valid
- Secondly MFShutdown should not be called after mfplat.dll was freed with FreeLibrary

With these changes a goto error; should be sufficient here ...

perkerk: @mfleisz ok but there are more logic problems in the init/uninit than just those. I'll make a another pass and clean it up

mfleisz: @perkerk thanks!

perkerk:
Restored the original mf_init code and changed mf_uninit to clean up. @mfleisz Not sure if you wanted me to change the NULL check before _aligned_free or if that was just info, but did it anyway

libfreerdp/codec/h264_mf.c
```diff
@@ -470,17 +476,22 @@ static void mf_uninit(H264_CONTEXT* h264)
...
 		if (sys->mfplat)
 		{
+			if (mf_plat_loaded(sys))
+				sys->MFShutdown();
+
 			FreeLibrary(sys->mfplat);
 			sys->mfplat = NULL;
+
+			if (mf_plat_loaded(sys))
+				CoUninitialize();
```
mfleisz: This could be called right after MFShutdown

mfleisz:
@akallabeth If you can have a quick look again - if you are happy with the changes we can merge this one.

mfleisz:
@freerdp-bot test

PR #4471, opened 2018-03-07
Thread and Threadpool work function fix

akallabeth:
* Removed unnecessary and harmful function pointer casts for `CreateThread` and `CreateThreadpoolWork`
* Fixed return, call and parameter type of thread and threadpool work functions.

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -1332,7 +1332,7 @@ static int xf_logon_error_info(freerdp* instance, UINT32 data, UINT32 type)
 	return 1;
 }
 
-static void* xf_input_thread(void* arg)
+static DWORD WINAPI xf_input_thread(LPVOID arg)
 {
 	BOOL running = TRUE;
 	DWORD status;
@@ -1411,7 +1411,7 @@ static void* xf_input_thread(void* arg)
 
 	MessageQueue_PostQuit(queue, 0);
 	ExitThread(0);
-	return NULL;
+	return 0;
 }
 
 static BOOL xf_auto_reconnect(freerdp* instance)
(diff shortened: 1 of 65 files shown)


client/X11/xf_client.c
```diff
@@ -1463,7 +1463,7 @@ static BOOL xf_auto_reconnect(freerdp* instance)
 *  @param instance - pointer to the rdp_freerdp structure that contains the session's settings
 *  @return A code from the enum XF_EXIT_CODE (0 if successful)
 */
-static void* xf_client_thread(void* param)
+static DWORD WINAPI xf_client_thread(LPVOID param)
 {
 	BOOL status;
 	int exit_code;
```
mfleisz: Should have type UINT/DWORD

libfreerdp/core/message.c
```diff
@@ -2861,7 +2861,7 @@ static void* update_message_proxy_thread(void* arg)
 		WLog_ERR(TAG, "update=%p, update->queue=%p", (void*) update,
 		         (void*)(update ? update->queue : NULL));
 		ExitThread(-1);
-		return NULL;
+		return -1;
```
mfleisz: This will certainly cause a warning - should set/return error to 1 instead.

winpr/libwinpr/pipe/test/TestPipeCreateNamedPipe.c
```diff
@@ -90,7 +90,7 @@ static void* named_pipe_client_thread(void* arg)
 	return NULL;
```
mfleisz: Should return 0 instead of NULL

winpr/libwinpr/synch/test/TestSynchCritical.c
```diff
@@ -82,18 +82,18 @@ static PVOID TestSynchCritical_Test1(PVOID arg)
...
 		LeaveCriticalSection(&critical);
 		return (PVOID)1;
 	}
-	return (PVOID)0;
+	return 0;
 }
 
 
-static PVOID TestSynchCritical_Main(PVOID arg)
+static DWORD WINAPI TestSynchCritical_Main(LPVOID arg)
```
mfleisz: Some paths still return (PVOID)0 or (PVOID)1

PR #4475, opened 2018-03-08
[WIP] Fix Windows 8, 10 cursors drawing as black

perkerk:
Fixes 2 issues:
- xor buffer was misaligned because scanline size was calculated as width * GetBitsPerPixel
- CreateBitmap did not create a usable bitmap for the hbmColor field

### DIFF
--- a/client/Windows/wf_graphics.c
+++ b/client/Windows/wf_graphics.c
@@ -237,24 +237,35 @@ static BOOL wf_Pointer_New(rdpContext* context, const rdpPointer* pointer)
 		            pointer->height);
 		info.hbmMask = CreateBitmap(pointer->width, pointer->height, 1, 1, pdata);
 		_aligned_free(pdata);
-		pdata = (BYTE*) _aligned_malloc(pointer->width * pointer->height *
-		                                GetBitsPerPixel(gdi->dstFormat), 16);
 
-		if (!pdata)
+		UINT32 srcFormat;
+		switch (pointer->xorBpp)
+		{
+		case 16:
+			srcFormat = PIXEL_FORMAT_RGB15;
+			break;
+		case 24:
+			srcFormat = PIXEL_FORMAT_BGR24;
+			break;
+		case 32:
+			srcFormat = PIXEL_FORMAT_BGRA32;
+			break;
+		default:
+			srcFormat = gdi->palette.format;
+			break;
+		}
+		info.hbmColor = wf_create_dib((wfContext*)context, pointer->width, pointer->height, srcFormat, NULL, &pdata);
+
+		if (!info.hbmColor)
 			goto fail;
 
 		if (!freerdp_image_copy_from_pointer_data(pdata, gdi->dstFormat, 0, 0, 0,
-		        pointer->width, pointer->height,
-		        pointer->xorMaskData, pointer->lengthXorMask,
-		        pointer->andMaskData, pointer->lengthAndMask, pointer->xorBpp, &gdi->palette))
+			pointer->width, pointer->height,
+			pointer->xorMaskData, pointer->lengthXorMask,
+			pointer->andMaskData, pointer->lengthAndMask, pointer->xorBpp, &gdi->palette))
 		{
-			_aligned_free(pdata);
 			goto fail;
 		}
-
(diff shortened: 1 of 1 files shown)


client/Windows/wf_graphics.c
```diff
@@ -237,24 +237,35 @@ static BOOL wf_Pointer_New(rdpContext* context, const rdpPointer* pointer)
...
 		info.hbmMask = CreateBitmap(pointer->width, pointer->height, 1, 1, pdata);
 		_aligned_free(pdata);
-		pdata = (BYTE*) _aligned_malloc(pointer->width * pointer->height *
-		                                GetBitsPerPixel(gdi->dstFormat), 16);
 
-		if (!pdata)
+		UINT32 srcFormat;
+		switch (pointer->xorBpp)
+		{
+		case 16:
```
akallabeth: Is this always 15bpp?
Case for 8bpp and 15bpp is missing or are they not supported at all?

akallabeth:
@freerdp-bot test

akallabeth:
Thank you for the contribution, the windows client does really need some love :+1:

client/Windows/wf_graphics.c
```diff
@@ -237,24 +237,35 @@ static BOOL wf_Pointer_New(rdpContext* context, const rdpPointer* pointer)
 		            pointer->height);
 		info.hbmMask = CreateBitmap(pointer->width, pointer->height, 1, 1, pdata);
 		_aligned_free(pdata);
-		pdata = (BYTE*) _aligned_malloc(pointer->width * pointer->height *
-		                                GetBitsPerPixel(gdi->dstFormat), 16);
 
-		if (!pdata)
+		UINT32 srcFormat;
```
akallabeth: Declarations at beginning of a scope only, older compilers (hello VS2010) are not really C99 conformant...

client/Windows/wf_graphics.c
```diff
@@ -237,24 +237,35 @@ static BOOL wf_Pointer_New(rdpContext* context, const rdpPointer* pointer)
...
 
 		if (!freerdp_image_copy_from_pointer_data(pdata, gdi->dstFormat, 0, 0, 0,
-		        pointer->width, pointer->height,
-		        pointer->xorMaskData, pointer->lengthXorMask,
-		        pointer->andMaskData, pointer->lengthAndMask, pointer->xorBpp, &gdi->palette))
+			pointer->width, pointer->height,
+			pointer->xorMaskData, pointer->lengthXorMask,
+			pointer->andMaskData, pointer->lengthAndMask, pointer->xorBpp, &gdi->palette))
 		{
-			_aligned_free(pdata);
```
akallabeth: possible memory leak in error case here

client/Windows/wf_graphics.c
perkerk: Well 8bpp was falling through to palette format below. This switch is just a reverse map of what's in the freerdp_image_copy_from_pointer_data() routine, and other resolutions aren't supported by it. I should be more specific in this case statement. 

But, can I just get rid of this case statement and using  freerdp_image_copy_from_pointer_data altogether? Do you know why it's even in there. The and and xor data should already be in a usable format for CreateBitmap and CreateDIBSection (other than flipping it to a top-down bitmap). I should be able to call flip_bitmap on the xor data and pass it directly to CreateDIBSection instead

client/Windows/wf_graphics.c
perkerk: got it

client/Windows/wf_graphics.c
perkerk: I will look into what bit depths are supported for color cursors, maybe some <24bpp formats do need to be converted to an RGB (24 or 32) format. I'll also check what happens in the 16bpp case

akallabeth: @perkerk You're probably right, the `freerdp_image_copy_from_pointer_data` is used on non windows platforms to convert to usable bitmap formats.

perkerk:
@akallabeth Won't be able to get to this until sometime next week

akallabeth:
@perkerk any update?

perkerk:
@akallabeth fixed in my fork but I have to merge up, will try to get to this over the weekend

PR #4477, opened 2018-03-09
Detect command line flags in case RDP or MSINCIDENT files are used.

akallabeth:
Command line detection sets flags for the parser.
Run it when the first argument is a `.rdp` or `.msIncident` file with `argc - 1` and `&argv[1]` to avoid breaking the detector.


akallabeth:
@bmiklautz @hardening Could anyone have a look at this one, I think working `.rdp` and assistance support would be niche for the next rc

PR #4478, opened 2018-03-10
Updated project site link to HTTPS

smalleel:
A very small pull request that simply changes the project site in the README document from "http://www.freerdp.com/" to "https://www.freerdp.com/". Directing more users to the HTTPs version of the site will help improve user security a little.

There's a good explanation on the Google Developer guides why it's a good idea to provide and use HTTPS, even for static sites that don't handle credentials or sensative data: https://developers.google.com/web/fundamentals/security/encrypt-in-transit/why-https

### DIFF
--- a/README
+++ b/README
@@ -8,7 +8,7 @@ interoperability can finally liberate your computing experience.
 Resources
 ---------
 
-Project website: http://www.freerdp.com/
+Project website: https://www.freerdp.com/
 Issue tracker: https://github.com/FreeRDP/FreeRDP/issues
 Sources: https://github.com/FreeRDP/FreeRDP/
 Downloads: https://pub.freerdp.com/releases/


akallabeth:
@freerdp-bot test

bmiklautz:
@catleeball thank you for your contribution.
You are right using https would the much better choice and we are aware oft that. However we do use github pages for a while now for www.freerdp.com and until now github doesn't officially support https for custom domains. As it seems they are working on it and in our case it actually already works nicely with https but until it is not officially available we will need to stick to http in the documentation.

I'll leave the PR open. Once github announces the support officially I'll merge it.
Thank you!

akallabeth:
@bmiklautz As far as I can tell https is now supported?

Issue #3518, opened 2016-09-29
Long term test - Freerdp disconnects with fastpath error


T3N4K4:
**FreeRDP Master** - Git Revision 24aa71e6cb547c02fbdf724258c3e3c0932ef0b9

**All cmake Options:**
sudo cmake -DCMAKE_INSTALL_PREFIX=/opt/freerdp -DWITH_JPEG=ON -DWITH_THIRD_PARTY=ON -DBUILTIN_CHANNELS=OFF -DWITH_CUPS=ON -DWITH_PULSE=ON -DWITH_PCSC=ON -DWITH_CHANNELS=ON -DCHANNEL_SERIAL=ON \ 
 -DCHANNEL_SMARTCARD=ON -DWITH_CLIENT_INTERFACE=ON -DWITH_XINERAMA=ON -DWITH_XEXT=ON -DWITH_XV=ON -DWITH_XI=ON -DWITH_XRENDER=ON -DWITH_SSE2=ON -DCHANNEL_PRINTER=ON -DCHANNEL_PRINTER_CLIENT=ON \
-DCHANNEL_RDPDR=ON -DCHANNEL_DRIVE=ON -DCHANNEL_CLIPRDR=ON -DCHANNEL_URBDRC_CLIENT=ON -DCHANNEL_SMARTCARD_CLIENT=ON -DCHANNEL_URBDRC=ON -DWITH_WAYLAND=OFF -DWITH_FFMPEG=ON -DWITH_X11=ON \
-DWITH_OPENH264=ON -DWITH_GSM=ON -DWITH_GSTREAMER_0_10=ON

Connection server -> Windows Server 2012 R2 Standard 64-Bit
Client system -> Linux ubuntu-trusty-32 3.13.x

**Command Line:**
/usr/bin/xfreerdp /u:"$USER" /p:"$PASS" /v:"$SERVER" /d:"$DOMAIN" /gdi:hw /f /multimon /bpp:24 /kbd:0x0404 +async-input +async-update +sec-rdp +sec-tls -sec-nla /rfx +grab-keyboard /smartcard +compression /network:lan +clipboard /cert-ignore

```
[12:02:57:719] [24952:24953] [WARN][com.freerdp.core.client] - Skipping, channel already loaded
[12:02:57:754] [24952:24953] [ERROR][com.winpr.timezone] - Unable to find a match for unix timezone: Etc/UTC
[...]

PR #4481, opened 2018-03-13
Fastpath uncompressed length issue

akallabeth:
After the last feedback from #3518 I've found the following issue:
* In case of a compressed update a stream is created with the capacity/size of the single update to be processed.
* In case of an uncompressed stream the capacity/size matched the whole received data, which could contain multiple independent updates.


akallabeth:
@mfleisz I've refactored the pull, now we reuse a single buffer for everything, that should eliminate all drawbacks of creating a copy.
@MartinHaimberger Should be no performance issue in there any more and the pull is now actually quite simple, could you review?

mfleisz:
LGTM

PR #4482, opened 2018-03-13
Winpr exit

akallabeth:
Return the actual exit code instead of always 0 (side effect of `pthread_exit`)


akallabeth:
@mfleisz Hmm, with #4529 in place `winpr_exit` is actually no longer necessary, maybe we should remove that again before the next rc?

mfleisz:
@akallabeth Yeah - let's get rid of winpr_exit 👍

PR #4501, opened 2018-03-20
Removed xrdp workarounds.

akallabeth:
These workarounds limit functionality with other server implementations
and should be fixed in xrdp.

I've notified the xrdp people (https://github.com/neutrinolabs/xrdp/issues/1068)


bmiklautz:
+1 for removal but let's postpone the merge until the  xrdp issue has been resolved

Issue #4503, opened 2018-03-21
New warnings after implementation of newer versions of CredSSP 


hardening:
I get this:
```
[27/81] Compiling C object 'libfreerdp/freerdp@sha/core_nla.c.o'.
../libfreerdp/core/nla.c: In function ‘nla_encrypt_public_key_hash’:
../libfreerdp/core/nla.c:1132:35: warning: pointer targets in passing argument 2 of ‘winpr_Digest_Update’ differ in signedness [-Wpointer-sign]
  if (!winpr_Digest_Update(sha256, hashMagic, strlen(hashMagic) + 1))
                                   ^
In file included from ../libfreerdp/core/nla.h:30:0,
                 from ../libfreerdp/core/nla.c:48:
.././winpr/include/winpr/crypto.h:671:16: note: expected ‘const BYTE * {aka const unsigned char *}’ but argument is of type ‘const CHAR * {aka const char *}’
 WINPR_API BOOL winpr_Digest_Update(WINPR_DIGEST_CTX* ctx, const BYTE* input, size_t ilen);
                ^
../libfreerdp/core/nla.c: In function ‘nla_decrypt_public_key_hash’:
../libfreerdp/core/nla.c:1371:35: warning: pointer targets in passing argument 2 of ‘winpr_Digest_Update’ differ in signedness [-Wpointer-sign]
  if (!winpr_Digest_Update(sha256, hashMagic, strlen(hashMagic) + 1))
                                   ^
In file included from ../libfreerdp/core/nla.h:30:0,
                 from ../libfreerdp/core/nla.c:48:
.././winpr/include/winpr/crypto.h:671:16: note: expected ‘const BYTE * {aka const unsigned char *}’ but argument is of type ‘const CHAR * {aka const char *}’
 WINPR_API BOOL winpr_Digest_Update(WINPR_DIGEST_CTX* ctx, const BYTE* input, size_t ilen);
                ^
```

PR #4499, opened 2018-03-20
cssp: Add support for protocol version 6

mfleisz:
This PR implements CredSSP protocol version 6 (refer to issue #4488 for more details).

PR #4504, opened 2018-03-21
cssp: Fix warnings (#4503)

mfleisz:
Fix warnings introduced in #4499 as reported in #4503.

### DIFF
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -115,8 +115,26 @@ static void nla_identity_free(SEC_WINNT_AUTH_IDENTITY* identity);
 #define ber_sizeof_sequence_octet_string(length) ber_sizeof_contextual_tag(ber_sizeof_octet_string(length)) + ber_sizeof_octet_string(length)
 #define ber_write_sequence_octet_string(stream, context, value, length) ber_write_contextual_tag(stream, context, ber_sizeof_octet_string(length), TRUE) + ber_write_octet_string(stream, value, length)
 
-static const CHAR ClientServerHashMagic[] = "CredSSP Client-To-Server Binding Hash\0";
-static const CHAR ServerClientHashMagic[] = "CredSSP Server-To-Client Binding Hash\0";
+/* CredSSP Client-To-Server Binding Hash\0 */
+static const BYTE ClientServerHashMagic[] =
+{
+	0x43, 0x72, 0x65, 0x64, 0x53, 0x53, 0x50, 0x20,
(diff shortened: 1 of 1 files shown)


libfreerdp/core/nla.c
```diff
@@ -1129,7 +1129,7 @@ SECURITY_STATUS nla_encrypt_public_key_hash(rdpNla* nla)
 		goto out;
 
 	/* include trailing \0 from hashMagic */
-	if (!winpr_Digest_Update(sha256, hashMagic, strlen(hashMagic) + 1))
+	if (!winpr_Digest_Update(sha256, (const BYTE*)hashMagic, strlen(hashMagic) + 1))
```
akallabeth: Why not just declare the buffer as `BYTE` and use `ARRAYSIZE` for the length?

akallabeth: and if we are at cleanups, have a look, there are still a lot of broken/wrong/invalid `(int)` casts around

mfleisz:
@akallabeth Changed the handling of the hash magic and fixed some int/size_t casts

libfreerdp/core/nla.c
```diff
@@ -115,8 +115,26 @@ static void nla_identity_free(SEC_WINNT_AUTH_IDENTITY* identity);
 #define ber_sizeof_sequence_octet_string(length) ber_sizeof_contextual_tag(ber_sizeof_octet_string(length)) + ber_sizeof_octet_string(length)
 #define ber_write_sequence_octet_string(stream, context, value, length) ber_write_contextual_tag(stream, context, ber_sizeof_octet_string(length), TRUE) + ber_write_octet_string(stream, value, length)
 
-static const CHAR ClientServerHashMagic[] = "CredSSP Client-To-Server Binding Hash\0";
-static const CHAR ServerClientHashMagic[] = "CredSSP Server-To-Client Binding Hash\0";
+/* CredSSP Client-To-Server Binding Hash\0 */
+static const BYTE ClientServerHashMagic[] =
```
akallabeth: Does VS complain when initializing `BYTE` (aka `unsigned char`) simply as string?

mfleisz: VS 2017 has no complaint about it - but then there are other compilers out there as we know. Also this way we ensure that there is definitely just one '\0' at the end of the string - compilers are supposed to add one at the end when defining strings this way ... supposed to.

akallabeth: @mfleisz  ok, keeping the magic :)

PR #4508, opened 2018-03-21
Fix ccache support

h3xx:
Setting the compiler launcher to "ccache" is the recommended way of enabling ccache for the build.

If cmake is run with it defined, it causes an error when ccache tries to run:

    ccache: error: Recursive invocation (the name of the ccache binary must be "ccache")

This was because the compiler was getting invoked as `ccache ccache [COMPILER]`

Reproduced by running cmake:

    mkdir BUILD && cd BUILD &&
    cmake \
        -DCMAKE_C_COMPILER_LAUNCHER=ccache \
        -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
        ..
    make

This results in the following error:

    $ make
    Scanning dependencies of target winpr
    [  0%] Building C object winpr/libwinpr/CMakeFiles/winpr.dir/synch/address.c.o
    ccache: error: Recursive invocation (the name of the ccache binary must be "ccache")
    make[2]: *** [winpr/libwinpr/CMakeFiles/winpr.dir/build.make:63: winpr/libwinpr/CMakeFiles/winpr.dir/synch/address.c.o] Error 1
    make[1]: *** [CMakeFiles/Makefile2:250: winpr/libwinpr/CMakeFiles/winpr.dir/all] Error 2
    make: *** [Makefile:152: all] Error 2

(This patch also takes into account the fact there might be ccache alternatives in the future, or even crazy custom settings like `distcc` & `ccache` somehow working together via the compiler launcher)


akallabeth:
@freerdp-bot test
@h3xx Do you have a link at hand, can not find the proper documentation for the variable in the `CMake` docs
All I can find are mailing list entries that are at most 2 years old, so this most likely needs some compatibility for `CMake-2.8` versions

h3xx:
The documentation for `CMAKE_<LANG>_COMPILER_LAUNCHER` is [here](https://cmake.org/cmake/help/latest/prop_tgt/LANG_COMPILER_LAUNCHER.html). It mentions cmake and distcc as examples.

The [documentation for `RULE_LAUNCH_COMPILE`](https://cmake.org/cmake/help/latest/prop_gbl/RULE_LAUNCH_COMPILE.html) is here.

akallabeth:
@h3xx Thanks. So `CMAKE_<LANG>_COMPILER_LAUNCHER` is available from `CMake 3.4` onward? Should fall back to the old version if `CMake` is too old, otherwise this looks good.

Thanks for your contribution!

h3xx:
Okay, I added backwards-compatibility for cmake <3.4. I looked up all the tags in their git repo where it wasn't supported.

akallabeth:
@freerdp-bot test

bmiklautz:
@h3xx thanks.

Issue #4502, opened 2018-03-20
nla-connection to Windows7 with CredSSP patch stopped working again (de83f4df2)


hendwolt:
With current master (de83f4df2) nla connections to a Windows7 Prof. with CredSSP patch stopped working. With the older version f8baeb724 it still works. 
Connecting with /sec:tls to Windows7 works with the new version.
Connecting with /sec:nla to Winsows10 (Insider) works with the new version.

xfreerdp /buildconfig :
This is FreeRDP version 2.0.0-dev2 (de83f4df2)
[...]

PR #4510, opened 2018-03-23
cssp: Separate client/server version handling (#4502)

mfleisz:
This PR fixes CredSSP issues with non-version 6 servers as described in #4502. Tested with patched and unpatched Win7 Ultimate x64.


mdunc:
I tested it on a patched Windows 2012 Server and it works perfect.

matthewpapa07:
Works for me again with latest master. Thanks for the fix!

PR #4513, opened 2018-03-25
initial travis-ci support

chipitsine:
this is very initial travis-ci support

travis-ci is not a replacement for ci.freerdp.com, it is rather quick PR test thing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,54 @@
+sudo: required
+dist: trusty
+
+os: linux
+
+language: c
+
+compiler:
+  - gcc
+
+matrix:
+  include:
+    - os: linux
+      compiler: gcc
+    - os: linux
+      compiler: clang
+  exclude:
+    - compiler: gcc
+
+addons:
+  apt:
+    packages:
+      - gdb
+      - libx11-dev
+      - libxrandr-dev
+      - libxi-dev
+      - libxv-dev
+      - libcups2-dev
+      - libxdamage-dev
+      - libxcursor-dev
+      - libxext-dev
+      - libxinerama-dev
+      - libxkbcommon-dev
+      - libxkbfile-dev
+      - libxml2-dev
+      - libasound2-dev
+      - libgstreamer1.0-dev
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

.travis.yml
```diff
@@ -0,0 +1,54 @@
...
+      - libavutil-dev
+      - libx264-dev
+      - libxext-dev
+
+before_script:
+  - ulimit -c unlimited -S 
+
+script:
+  - sudo hostname travis-ci.local
+  - cmake -G "Unix Makefiles" -C ci/cmake-preloads/config-linux-all.txt -D CMAKE_BUILD_TYPE=Debug -DWITH_LIBSYSTEMD=OFF -DWITH_WAYLAND=OFF .
```
akallabeth: Would use `Ninja` build, is a lot faster ;)
Just a question though, why the additional settings (systemd, etc)? Just add them to the deps and the build will succeed.

.travis.yml
```diff
@@ -0,0 +1,54 @@
...
+  - make
+  - make test
```
akallabeth: `cmake --build . --target test` would work with any generator used.

.travis.yml
chipitsine: I had a look to current jenkins build, and picked cmake from there


as for systemd - it is ubuntu 14.04, no systemd

(well, I agree that it might be detected in automated way, however, I think even such a build that I provided is somewhat good, let other contributors to improve it later ?)

as for my own plans, I'd like to build on "osx" and I'd like to integrate memory, address, thread sanitizers (I already have several tests failing with thread sanitizer)

bmiklautz:
@chipitsine thanks for your contribution.

We won't enable it for the integrations tests but it's nice to have for regular tests/builds.

chipitsine:
So, you will setup cron jobs on travis-ci.org?

And PR jobs?

bmiklautz:
@chipitsine no don't plan to. But it's can be convenient for testing your own repo/branch before doing a pull request.

PR #4519, opened 2018-03-29
cssp: Fix handling of nonce

mfleisz:
The nonce must not change during the message exchange.
Currently the nonce would be newly generated (and leaking memory) if the authentication is performed in more than 1 step.


libfreerdp/core/nla.c
```diff
@@ -2315,6 +2312,13 @@ rdpNla* nla_new(freerdp* instance, rdpTransport* transport, rdpSettings* setting
 	}
 
 	return nla;
+
```
akallabeth: wouldn't calling `nla_free` be the better option (and check, if that does proper ceanup)?

mfleisz:
@akallabeth Agree - this shoud do better

akallabeth:
@bmiklautz @hardening Anyone up for review?

PR #4522, opened 2018-03-30
Fix SCardStatusW function

UndefBehavior:
Terminating null character was inserted in the middle of readerNames instead of
last position in the unicode version of SCardStatus function.
This commit fix it.


hardening:

bmiklautz:
@UndefBehavior nerver had a situation where the unicode version of the function was used while testing ;). Thanks for your contribution.

PR #4527, opened 2018-04-01
PCB support

jkaivo:
This commit adds support for the "pcb" field in .RDP files generated by the VM interface in Project Honolulu. Project Honolulu is the web interface to Windows Server management that will be included with Windows Server 2019. When using the web interface to manage Hyper-V virtual machines, there is an option for Connect which generates an .RDP file to connect directly to the virtual machine console (similar in function to the freerdp /vmconnect switch). The .RDP file includes the necessary change to "server port" and "negotiate security layer". This patch simply parses the "pcb" string from the file and adds it to the settings so the connection actually works.


akallabeth:
@freerdp-bot test

PR #4529, opened 2018-04-03
Fix #4524: Initialize with cleanup handler

### LINKED ISSUE #4524: Commit 7e26c548668508ca59b5b204b4973d2a4ef8d4ce causes crash in multi-client service.
freerdp crashes on logging calls, because  WLog_Uninit() calls WLog_Free(root) while other thread is in the middle of WLog* call.
Suggest removing WLog_Uninit() (WLog_Init() seems harmless)

akallabeth:
Use singleton initializer and register cleanup handler for logger.

manually cleaning up WLog by calling `WLog_Uninit` is not really feasable, do it at program termination.

### DIFF
--- a/winpr/libwinpr/utils/wlog/wlog.c
+++ b/winpr/libwinpr/utils/wlog/wlog.c
@@ -78,6 +78,10 @@ static int WLog_ParseLogLevel(LPCSTR level);
 static BOOL WLog_ParseFilter(wLogFilter* filter, LPCSTR name);
 static BOOL WLog_ParseFilters(void);
 
+#if !defined(_WIN32)
+static void WLog_Uninit_(void) __attribute__((destructor));
+#endif
+
 static void WLog_Uninit_(void)
 {
 	DWORD index;
@@ -152,7 +156,9 @@ static BOOL CALLBACK WLog_InitializeRoot(PINIT_ONCE InitOnce, PVOID Parameter, P
 	if (!WLog_SetLogAppenderType(g_RootLog, logAppenderType))
 		goto fail;
 
+#if defined(_WIN32)
 	atexit(WLog_Uninit_);
+#endif
 	return TRUE;
 fail:
 	free(g_RootLog);
--- a/winpr/libwinpr/utils/wlog/wlog.c
+++ b/winpr/libwinpr/utils/wlog/wlog.c
@@ -28,6 +28,7 @@
 #include <winpr/print.h>
 #include <winpr/debug.h>
 #include <winpr/environment.h>
+#include <winpr/wlog.h>
 
 #if defined(ANDROID)
 #include <android/log.h>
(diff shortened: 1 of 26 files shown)


winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -66,12 +66,103 @@ LPCSTR WLOG_LEVELS[7] =
...
+
+static BOOL CALLBACK WLog_InitializeRoot(PINIT_ONCE InitOnce, PVOID Parameter, PVOID* Context)
+{
+	char* env;
+	DWORD nSize;
+	DWORD logAppenderType;
+
+	if (!g_RootLog)
+	{
+		LPCSTR appender = "WLOG_APPENDER";
```
hardening: AFAICT when this function is called g_RootLog is always NULL, so the `if(!g_RootLog)` test is useless.

akallabeth: @hardening right, classic copy paste error ^^

akallabeth:
@freerdp-bot test please

winpr/libwinpr/utils/wlog/wlog.c
```diff
@@ -66,12 +66,99 @@ LPCSTR WLOG_LEVELS[7] =
...
+		else if (_stricmp(env, "UDP") == 0)
+			logAppenderType = WLOG_APPENDER_UDP;
+
+		free(env);
+	}
+
+	if (!WLog_SetLogAppenderType(g_RootLog, logAppenderType))
+		goto fail;
+
+	atexit(WLog_Uninit_);
```
mfleisz: I think this could get us into trouble. MSDN says:

"The code in the atexit function should not contain any dependency on any DLL which could have already been unloaded when the atexit function is called."

If someone uses a custom logger with cleanup code like i.e. WLog_UdpAdapter_Free (which calls socket APIs) this could lead to crashes on app exit.

Not sure what you guys think about this - maybe I'm just too pessimistic here ... ^^

akallabeth: @mfleisz That would also crash if `WLog_*` functions are called...
Don't think that the dependent libs are already unloaded there (unless someone really wants to get hurt)

mfleisz: @akallabeth WLog_* functions won't be a problem as they are part of winpr.dll and are still loaded when the atexit() handler is called (on dll unload of winpr). There is just no guarantee the adapter free handlers won't call any functions from other dlls that might have been already unloaded before winpr.

akallabeth: @mfleisz My understanding of this is that it is only relevant when using `WinPR` with `dlopen`/`dlclose` (or `LoadLibrary`/`CloseHandle`)

PR #4532, opened 2018-04-03
Fix spelling of network commandline values

T3N4K4:
Found that one some time ago, guess it should be named high instead of higg ;-)


akallabeth:
@freerdp-bot test

Issue #4523, opened 2018-03-30
X11/xf_graphics.c comparing a nested structure with NULL ?


hendwolt:
When I compile the current master for Linux (openSUSE) I get compiler warnings about :
client/X11/xf_graphics.c:173:22: warning: comparison between pointer and integer
client/X11/xf_graphics.c:176:19: warning: assignment makes integer from pointer without a cast [-Wint-conversion]

In function xf_Bitmap_free:
...
        if (xbitmap->pixmap != NULL)
        {
                XFreePixmap(xfc->display, xbitmap->pixmap);
                xbitmap->pixmap = NULL;
        }
...
As far as I see, pixmap is a nested structure and not a pointer:
(xfreerdp.h)
struct xf_bitmap
{
        rdpBitmap bitmap;
        Pixmap pixmap;
        XImage* image;
};

I'm not sure, what this code actually does; if it works.

hendwolt:
Sorry for misinterpreting Pixmap. After looking into the X11 headers it seems to be an unsigned 64 bit integer.
But the code seems odd anyway.
NULL will probably be resolved to (int)0.

PR #4533, opened 2018-04-03
fix warnings about discarded const and comparing int with null

hendwolt:
This fixes a few warnings about discarded const qualifiers and comparing an integer with null and setting it to null; including the case described in #4523.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -589,7 +589,7 @@ static char** freerdp_command_line_parse_comma_separated_values_ex(const char* n
 
 			if (p)
 			{
-				p[0] = name;
+				p[0] = (char*)name;
 				*count = 1;
 				return p;
 			}
--- a/winpr/libwinpr/utils/cmdline.c
+++ b/winpr/libwinpr/utils/cmdline.c
@@ -289,7 +289,7 @@ int CommandLineParseArgumentsA(int argc, LPCSTR* argv, COMMAND_LINE_ARGUMENT_A* 
 
 				if (value)
 				{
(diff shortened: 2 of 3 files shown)


akallabeth:
@freerdp-bot test

client/common/cmdline.c
```diff
@@ -589,7 +589,7 @@ static char** freerdp_command_line_parse_comma_separated_values_ex(const char* n
 
 			if (p)
 			{
-				p[0] = name;
+				p[0] = (char*)name;
```
akallabeth: This is something that needs to be fixed, so don't cast the warning away

winpr/libwinpr/utils/cmdline.c
```diff
@@ -289,7 +289,7 @@ int CommandLineParseArgumentsA(int argc, LPCSTR* argv, COMMAND_LINE_ARGUMENT_A*
 
 				if (value)
 				{
-					options[j].Value = value;
+					options[j].Value = (char*)value;
```
akallabeth: Same as above

akallabeth:
@hendwolt Thank you for the contribution, added the XPixmap fixes to #4535

client/common/cmdline.c
hendwolt: A few more lines down there is already such an assignment with the cast. In both cmdline.c files.
I tested both versions and both work.

akallabeth: @hendwolt yes, there are still many such things to be sorted out :/ They, of course, work but silence the compiler (which makes it easy to forget to fix)

PR #4543, opened 2018-04-06
Fix variable passsed to HashCallback with MIC.

oshogbo:
Fix value where the MessageIntegrityCheck is stored.
The value in the context is not set yet and we need one from
authentication message.

### DIFF
--- a/winpr/libwinpr/sspi/NTLM/ntlm.h
+++ b/winpr/libwinpr/sspi/NTLM/ntlm.h
@@ -250,6 +250,7 @@ struct _NTLM_CONTEXT
 	NTLM_NEGOTIATE_MESSAGE NEGOTIATE_MESSAGE;
 	NTLM_CHALLENGE_MESSAGE CHALLENGE_MESSAGE;
 	NTLM_AUTHENTICATE_MESSAGE AUTHENTICATE_MESSAGE;
+	UINT32 MessageIntegrityCheckOffset;
 	SecBuffer NegotiateMessage;
 	SecBuffer ChallengeMessage;
 	SecBuffer AuthenticateMessage;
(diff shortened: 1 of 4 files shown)


mfleisz:
@akallabeth Just had a quick look over the code and it is true that context->MessageIntegrityCheck is never initialized/set before the callback is called. However I agree that the proposed fix doesn't look like a real solution to that problem ...

akallabeth:
@mfleisz Looks like it should be initialized at the end of `ntlm_read_AuthenticateMessage` in client mode?

mfleisz:
@akallabeth Yeah I guess a CopyMemory should to the trick there (as it is already done for the other fields). 
For the server side I think it should be enough to ensure that `nltm_compute_message_integrity_check` is called before `ntlm_compute_XX_v2_response` in `ntlm_server_AuthenticateComplete`?

akallabeth:
@oshogbo Could you adjust the pull according to the changes discussed?

oshogbo:
Yes I plan to do so just - need some more time also to test it. ;)
Thanks for your patience.

oshogbo:
I did a little bit research and it turns out we cant ensure that nltm_compute_message_integrity_check is called before ntlm_compute_XX_v2_response in ntlm_server_AuthenticateComplete because it's depends on them.

The SessionBaseKey is calculated based on NTLM hash. The KeyExchangeKey is a copy of SessionBaseKey. The RandomSessionKey is generated base on the KeyExchangeKey. The ExportedSessionKey is a copy of the RandomSessionKey. Finally to calculate MIC we use ExportedSesssionKey which is based on SessionBaseKey which i just proved.

This is the chicken and egg problem IMHO. I would stay with this version of code.

akallabeth:
@oshogbo So, then `context->MessageIntegrityCheck` is unused? Would remove it then to avoid confusion.

oshogbo:
@akallabeth Done.

akallabeth:
@freerdp-bot test

PR #4544, opened 2018-04-06
Don't restrict the hostname size.

oshogbo:
There is no reason to restrict nSize to 2 the hostname can be empty on
UNIX-like machines.

### DIFF
--- a/libfreerdp/core/settings.c
+++ b/libfreerdp/core/settings.c
@@ -255,8 +255,7 @@ BOOL settings_get_computer_name(rdpSettings* settings)
 	DWORD nSize = 0;
 	CHAR* computerName;
 
-	if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || (GetLastError() != ERROR_MORE_DATA) ||
-	    (nSize < 2))
+	if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || GetLastError() != ERROR_MORE_DATA)
 		return FALSE;
 
 	computerName = calloc(nSize, sizeof(CHAR));
--- a/winpr/libwinpr/sspi/NTLM/ntlm.c
+++ b/winpr/libwinpr/sspi/NTLM/ntlm.c
@@ -52,8 +52,7 @@ static int ntlm_SetContextWorkstation(NTLM_CONTEXT* context, char* Workstation)
 
 	if (!Workstation)
 	{
-		if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || (GetLastError() != ERROR_MORE_DATA) ||
-		    (nSize < 2))
+		if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || GetLastError() != ERROR_MORE_DATA)
 			return -1;
 
 		computerName = calloc(nSize, sizeof(CHAR));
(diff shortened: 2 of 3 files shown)


hardening:

hardening:
@akallabeth sure, but I guess this prevents xfreerdp from working in containers with empty host name or other restricted environments.

libfreerdp/core/settings.c
```diff
@@ -255,8 +255,7 @@ BOOL settings_get_computer_name(rdpSettings* settings)
 	DWORD nSize = 0;
 	CHAR* computerName;
 
-	if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || (GetLastError() != ERROR_MORE_DATA) ||
-	    (nSize < 2))
+	if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || GetLastError() != ERROR_MORE_DATA)
```
akallabeth: This will still fail if the name is empty (see the `if (!computerName)` check after calloc)

winpr/libwinpr/sspi/NTLM/ntlm.c
```diff
@@ -52,8 +52,7 @@ static int ntlm_SetContextWorkstation(NTLM_CONTEXT* context, char* Workstation)
 
 	if (!Workstation)
 	{
-		if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || (GetLastError() != ERROR_MORE_DATA) ||
-		    (nSize < 2))
+		if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || GetLastError() != ERROR_MORE_DATA)
```
akallabeth: This will still fail if the name is empty (see the if (!computerName) check after calloc)

winpr/libwinpr/sspi/NTLM/ntlm_av_pairs.c
```diff
@@ -180,8 +180,7 @@ int ntlm_get_target_computer_name(PUNICODE_STRING pName, COMPUTER_NAME_FORMAT ty
 	DWORD nSize = 0;
 	CHAR* computerName;
 
-	if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || (GetLastError() != ERROR_MORE_DATA) ||
-	    (nSize < 2))
+	if (GetComputerNameExA(ComputerNameNetBIOS, NULL, &nSize) || GetLastError() != ERROR_MORE_DATA)
```
akallabeth: This will still fail if the name is empty (see the if (!computerName) check after calloc)

libfreerdp/core/settings.c
akallabeth: This will still fail if the name is empty (see the if (!computerName) check after calloc)

winpr/libwinpr/sspi/NTLM/ntlm.c
akallabeth: This will still fail if the name is empty (see the if (!computerName) check after calloc)

winpr/libwinpr/sspi/NTLM/ntlm_av_pairs.c
akallabeth: This will still fail if the name is empty (see the if (!computerName) check after calloc)

oshogbo:
@akallabeth For empty string the calloc will alloc exactly 1 bit for '\0' because lenght is returned strlen() + 1. So in all your cases nSize will be equals 1. So non of those calloc will fail. If you refer exactly for this:
```
                if (!computerName)
                        return -1;
```
You refer to the adders that is returned by calloc not to the value.
So everything lgtm.

Issue #4222, opened 2017-11-09
Gateway connection fails when gateway name is associated to more than a single IP


andreesteve:
The gateway I connect to is setup to send a challenge to my phone for 2FA (two factor authentication) after the credentials (user name and password) are validated. I assume the rdp client has no knowledge about it and just waits for the gateway to reply with the authentication result.

The weird thing is that more often than not, when trying to establish a connection, I don't get the phone challenge, so I need to kill the client and try to connect again, until one of the tries will get me a challenge on the phone.

I noticed I get different traces on the client when the gateway sends me a challenge vs. when it doesn't.

When no challenge is sent, this is the trace the client logs:
```
[22:34:23:818] [11344:11345] [INFO][com.freerdp.client.common.cmdline] - loading channelEx rdpdr
[22:34:23:818] [11344:11345] [INFO][com.freerdp.client.common.cmdline] - loading channelEx rdpsnd
[22:34:23:818] [11344:11345] [DEBUG][com.freerdp.channels.cliprdr.client] - VirtualChannelEntryEx
[22:34:23:818] [11344:11345] [INFO][com.freerdp.client.common.cmdline] - loading channelEx cliprdr
[22:34:23:818] [11344:11345] [DEBUG][com.freerdp.channels.drdynvc.client] - VirtualChannelEntryEx
[22:34:23:818] [11344:11345] [INFO][com.freerdp.client.common.cmdline] - loading channelEx drdynvc
[22:34:23:820] [11344:11345] [DEBUG][com.freerdp.client.x11] - Searching for XInput pointer device
[22:34:23:820] [11344:11345] [DEBUG][com.freerdp.client.x11] - Pointer device: 6
[22:34:23:820] [11344:11345] [DEBUG][com.freerdp.core.nego] - Enabling security layer negotiation: TRUE
[22:34:23:820] [11344:11345] [DEBUG][com.freerdp.core.nego] - Enabling restricted admin mode: FALSE
[22:34:23:820] [11344:11345] [DEBUG][com.freerdp.core.nego] - Enabling RDP security: TRUE
[22:34:23:820] [11344:11345] [DEBUG][com.freerdp.core.nego] - Enabling TLS security: TRUE
[22:34:23:820] [11344:11345] [DEBUG][com.freerdp.core.nego] - Enabling NLA security: TRUE
[22:34:23:820] [11344:11345] [DEBUG][com.freerdp.core.nego] - Enabling NLA extended security: FALSE
[...]

PR #4547, opened 2018-04-07
Gateway client (RDG) must connect to same IP/server for both channels

andreesteve:
## Problem description
 * Gateway connection fails when gateway name is associated to more than a single IP #4222 

Currently RDG first opens the OUT channel and the subsequently the IN channel. Each channel calls the underlying **tcp.h** available functions to resolve the peer address from its name (settings->GatewayHostname). The issue arises when the gateway name points to multiple IPs and the IN channel is opened against one IP and the OUT channel against a different one.

For example, imagine a large organization that wants to scale the number of concurrent users they can have through their gateway. It can set their gateway name have multiple [A-records](https://en.wikipedia.org/wiki/List_of_DNS_record_types) so they can have different IPs and gateway servers serving connections to a single gateway hostname.

From my observations, the RDG protocol does not support the IN and OUT channels to be handled by different gateway servers. Thus the need of guaranteeing that we pick a single server/IP to connect for both channels.

## Proposed solution

1. Added **freerdp_tcp_get_peer_address** to *tcp.c/h* that allows one to resolve the peer (remote) address from a socket
2. When opening the IN channel in *rdg.c* I first take the already connected socket for the OUT channel, get the peer address using (1) and use it to create the IN channel

## Notes

When a proxy is used between client and server, my understanding is that the proxy is the one to resolve the gateway hostname, thus no changes were made for this scenario.

### DIFF
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -1027,8 +1027,10 @@ static BOOL rdg_tls_in_connect(rdpRdg* rdg, const char* hostname, UINT16 port, i
 	BIO* bufferedBio = NULL;
 	rdpSettings* settings = rdg->settings;
 	const char* peerHostname = settings->GatewayHostname;
-	int peerPort = settings->GatewayPort;
+	UINT16 peerPort = settings->GatewayPort;
 	BOOL isProxyConnection = FALSE;
+	int outChannelSocket = 0;
+	char* gatewayAddress = NULL;
 	assert(hostname != NULL);
 
 	if (settings->ProxyType)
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

andreesteve:
Updated with new commit. Let me know if I should squash instead.

akallabeth:
@freerdp-bot test

libfreerdp/core/gateway/rdg.c
```diff
@@ -1037,17 +1039,43 @@ static BOOL rdg_tls_in_connect(rdpRdg* rdg, const char* hostname, UINT16 port, i
 		peerPort = settings->ProxyPort;
 		isProxyConnection = TRUE;
 	}
+	else
+	{
+		// if settings->GatewayHostname is a name, it may be mapped to more than one
```
akallabeth: Sorry to bother again, but we use plain C, so no C++ comments allowed.

andreesteve:
Updated comment to c-style.
Also rebased (w/ squash) again to take care of conflict.

akallabeth:
@freerdp-bot test

PR #4549, opened 2018-04-09
Fix checking of krb in encrypt public key echo.

oshogbo:
In commit 0e1a073384ddd1382f6a0619177cf0b601ad973c there was a mistake -
originally code said different then kerberos. Because of that NLA authentication
of server side didn't work for me.


akallabeth:
@freerdp-bot test

bmiklautz:
@oshogbo thanks

PR #4551, opened 2018-04-09
Fix setting of negotiated security protocol.

oshogbo:
PROTOCOL_RDP is equals 0 so if we and it with anything is still 0.
Correct way of doing it is to compare two values.


akallabeth:
@freerdp-bot test

PR #4563, opened 2018-04-11
Don't assume that the pointer function are set.

oshogbo:


hardening:
@freerdp-bot can you test that please

hardening:
Thanks for the contribution Mariusz

PR #4570, opened 2018-04-12
android Toolchain fix

akallabeth:
@ccpp Hope it is ok to take your patch from the fdroiddata repo?

### DIFF
--- a/client/Android/Studio/build.gradle
+++ b/client/Android/Studio/build.gradle
@@ -4,6 +4,19 @@ File file = new File('release.properties')
 if (file.canRead()) {
     properties.load(new FileInputStream(file))
 }
+   
+if (!hasProperty('RELEASE_STORE_FILE')) {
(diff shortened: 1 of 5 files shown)


ccpp:
@akallabeth Of course it's OK to use this. :)

ccpp:
@akallabeth Another thing which would be nice is to make the presence of the variables RELEASE_STORE_FILE and RELEASE_STORE_PASSWORD non-mandatory.
In other terms, if I want an unsigned APK (as for F-Droid), I do not want to add them as empty values, as in
```
prebuild=echo -e "RELEASE_STORE_FILE=\nRELEASE_KEY_ALIAS=" >>../gradle.properties
```

akallabeth:
@ccpp Ok, done. `RELEASE_STORE_*` variables get initialized if not available from environment/user properties.

PR #4576, opened 2018-04-18
Fix polling in RDG

ccpp:
This solves the issues discussed in #3602:
- It fixes the timeout passed to `poll`
- It implements `BIO_get_fd` for RDGs in a correct way

### DIFF
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -1526,7 +1526,7 @@ static int rdg_bio_gets(BIO* bio, char* str, int size)
 
 static long rdg_bio_ctrl(BIO* bio, int cmd, long arg1, void* arg2)
 {
-	int status = 0;
+	int status = -1;
 	rdpRdg* rdg = (rdpRdg*) BIO_get_data(bio);
 	rdpTls* tlsOut = rdg->tlsOut;
 	rdpTls* tlsIn = rdg->tlsIn;
@@ -1536,14 +1536,6 @@ static long rdg_bio_ctrl(BIO* bio, int cmd, long arg1, void* arg2)
 		(void)BIO_flush(tlsOut->bio);
 		(void)BIO_flush(tlsIn->bio);
 		status = 1;
-	}
-	else if (cmd == BIO_C_GET_EVENT)
-	{
-		if (arg2)
-		{
-			BIO_get_event(rdg->tlsOut->bio, arg2);
-			status = 1;
-		}
 	}
 	else if (cmd == BIO_C_SET_NONBLOCK)
 	{
@@ -1583,9 +1575,10 @@ static long rdg_bio_ctrl(BIO* bio, int cmd, long arg1, void* arg2)
 		else
 			status = 1;
 	}
-	else if (cmd == BIO_C_GET_FD)
+	else if (cmd == BIO_C_GET_EVENT || cmd == BIO_C_GET_FD)
 	{
 		/*
+		 * A note about BIO_C_GET_FD:
 		 * Even if two FDs are part of RDG, only one FD can be returned here.
 		 *
 		 * In FreeRDP, BIO FDs are only used for polling, so it is safe to use the outgoing FD only
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -1583,6 +1583,17 @@ static long rdg_bio_ctrl(BIO* bio, int cmd, long arg1, void* arg2)
 		else
 			status = 1;
 	}
+	else if (cmd == BIO_C_GET_FD)
+	{
+		/*
+		 * Even if two FDs are part of RDG, only one FD can be returned here.
+		 *
+		 * In FreeRDP, BIO FDs are only used for polling, so it is safe to use the outgoing FD only
+		 *
+		 * See issue #3602
+		 */
+		return BIO_get_fd(tlsOut->bio, NULL);
+	}
 
 	return status;
 }
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -1592,7 +1592,7 @@ static long rdg_bio_ctrl(BIO* bio, int cmd, long arg1, void* arg2)
 		 *
 		 * See issue #3602
 		 */
-		return BIO_get_fd(tlsOut->bio, NULL);
+		status = BIO_ctrl(tlsOut->bio, cmd, arg1, arg2);
 	}
 
 	return status;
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

ccpp:
Please, check again!

libfreerdp/crypto/tls.c
```diff
@@ -737,10 +737,14 @@ static int tls_do_handshake(rdpTls* tls, BOOL clientMode)
 
 		do
 		{
-			status = poll(&pollfds, 1, 10 * 1000);
+			status = poll(&pollfds, 1, 10);
 		}
 		while ((status < 0) && (errno == EINTR));
 
+		if (status == 0) {
+			status = WAIT_TIMEOUT;
```
p-pautov: WAIT_TIMEOUT is used below only in case of "_WIN32" and I believe "poll" is not going to be used in that case, so there seem to be no need for such fixup.

libfreerdp/core/gateway/rdg.c
```diff
@@ -1583,6 +1583,17 @@ static long rdg_bio_ctrl(BIO* bio, int cmd, long arg1, void* arg2)
...
+		 * Even if two FDs are part of RDG, only one FD can be returned here.
+		 *
+		 * In FreeRDP, BIO FDs are only used for polling, so it is safe to use the outgoing FD only
+		 *
+		 * See issue #3602
+		 */
+		status = BIO_ctrl(tlsOut->bio, cmd, arg1, arg2);
+	}
 
 	return status;
```
p-pautov: Makes sense to initialize "status" with "-1" on line 1529 to fail for non implemented commands.

P.S. I can't leave comment for line 1529 for some reason.

hardening:

libfreerdp/crypto/tls.c
ccpp: I am not so sure about whether HAVE_POLL_H is always undefined in Win32.
IMO, this statement at worst has no effect, and at best avoids an error in Windows. @akallabeth wat do you think?

libfreerdp/core/gateway/rdg.c
ccpp: I made a test run and it seems now all commands required for the test run are handled in this function. The only difference would be in the command `BIO_C_GET_EVENT`, where `status` can be left untouched (which did not happen in my test run).

I agree, Initializing with -1 would be a reasonable future-proof improvement. @akallabeth any objections? Or is the PR good to go? :)

libfreerdp/crypto/tls.c
p-pautov: @ccpp This statement creates unnecessary confusion as if "poll" is supported on _WIN32 and it's not.
In case of _WIN32 the code on line 724 does "BIO_get_event(tls->bio, &event);". And "poll" uses "fd", which is left uninitialized. If poll returns "1", the _WIN32 specific check on line 767 will fail as well.
I think, these series of ifndefs need to be rearranged to more clearly indicate supported configurations, but in mean time we shouldn't make it even more confusing.

ccpp: @p-pautov fair point!

libfreerdp/core/gateway/rdg.c
p-pautov: Looks like, handling of BIO_C_GET_EVENT also needs to be changed to
    status = BIO_ctrl(tlsOut->bio, cmd, arg1, arg2);

akallabeth: @p-pautov Yes, you're right, that will be required to get this stuff working on windows

ccpp:
I incorporated the points recommended by @p-pautov in https://github.com/FreeRDP/FreeRDP/pull/4576#pullrequestreview-113378805 and https://github.com/FreeRDP/FreeRDP/pull/4576#pullrequestreview-113381733 in the recent commits.
The PR may change the behavior of FreeRDP with a RDG connection; it has been successfully tested in X11 client, but should be tested on Windows, too.

libfreerdp/core/gateway/rdg.c
ccpp: @p-pautov @akallabeth Done. The cases `BIO_C_GET_EVENT` and `BIO_C_GET_FD` are now handled together as they are unique now.
One might ask why a switch/case is not used here instead of else-ifs, however I'd rather not change it anymore because it would produce a lot of changed LOC

ccpp:
The latest commit fixes the same bug for TSG (/gt:rpc)

libfreerdp/core/gateway/tsg.c
```diff
@@ -1920,7 +1920,7 @@ static int transport_bio_tsg_gets(BIO* bio, char* str, int size)
 
 static long transport_bio_tsg_ctrl(BIO* bio, int cmd, long arg1, void* arg2)
```
p-pautov: It makes sense to implement BIO_C_GET_FD here as well.

libfreerdp/core/gateway/rdg.c
```diff
@@ -1583,6 +1575,18 @@ static long rdg_bio_ctrl(BIO* bio, int cmd, long arg1, void* arg2)
 		else
 			status = 1;
 	}
+	else if (cmd == BIO_C_GET_EVENT || cmd == BIO_C_GET_FD)
+	{
+		/*
+		 * A note about BIO_C_GET_FD:
```
p-pautov: This equally applies to BIO_C_GET_EVENT, i.e. there are two events and only OUT connection event is returned.
Not sure, though, if such a comment is even necessary here, IMHO just mentioning the issue number in commit message is enough.

akallabeth:
@freerdp-bot test

libfreerdp/core/gateway/tsg.c
p-pautov: I'd expect it to be called when RDG RPC transport is used (/gt:rpc). That single BIO_get_fd call (from tls.c) should end up here in such case.

ccpp: @p-pautov I'd expect the same. But please test it! I did, and I asked to test for a reason... (e.g. things might work differently on other OSs or clients.)

My observations on this:
- Returning -1 would make the test fail anyway, so... it's not been called! (again, please test! If it does fail then I accept that I might be wrong here)
- Digging deeper, the actually invoked ctrl is that of RDG and not -- as we expected -- TSG.
- Probably, TLS is not running on top of TSG, but on top of RDG. But I do not know about these details.

p-pautov: Apparently, TSG/RPC transport operates in blocking mode during tls_do_handshake() call (in tls.c), so whole handshake is done after single BIO_do_handshake() and the statement with BIO_get_fd() is not even reached.
I guess, that means BIO_C_GET_FD implementation in RPC can be omitted for now...

PR #4586, opened 2018-04-24
Fix inconsistent cchReaderLen value between WinScard and pcsc-lite.

rubendv:
On windows we found that we were only getting half of the reader name when using the StatusW smartcard operation, the reason was that the value of cchReaderLen means the number of characters in WinScard, while in pcsc-lite it means the number of bytes.


akallabeth:
@freerdp-bot test

PR #4588, opened 2018-04-24
Fix comments where is cookie and where is token.

oshogbo:


hardening:
@freerdp-bot can you test these trivial changes ?

PR #4596, opened 2018-04-26
RDG related fixes for better compatibility with mstsc

p-pautov:
RDG related fixes for better compatibility with mstsc and thus F5 APM.

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -353,8 +353,8 @@ BOOL transport_connect(rdpTransport* transport, const char* hostname,
 	BOOL status = FALSE;
 	rdpSettings* settings = transport->settings;
 	rdpContext* context = transport->context;
-	transport->async = settings->AsyncTransport;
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

libfreerdp/core/transport.c
```diff
@@ -380,7 +381,7 @@ BOOL transport_connect(rdpTransport* transport, const char* hostname,
 			}
 		}
 
-		if (!status && settings->GatewayRpcTransport)
+		if (!status && settings->GatewayRpcTransport && rpcFallback)
```
akallabeth: Sure this is correct? Don't you want something like `if (!status && (settings->GatewayRpcTransport || rpcFallback))`?

p-pautov: Nope, RPC is supposed to be used only if it's enabled via GatewayRpcTransport AND fallback to RPC makes sense (rpcFallback == TRUE).

Issue #4597, opened 2018-04-26
FreeBSD strtol() behaves different -- /network:lan doesn't work


adriaandegroot:
FreeBSD's strtol() and family behave differently from (Linux) GLibC implementations. This affects the way they set (or don't re-set) errno on erroneous input, and then parameter handling by the freerdp executable throws an error, leading to .. well, "no worky". This is most obvious with the /network option, since that has both numeric and non-numeric (e.g. "lan") values to handle.

There are downstream bugs at
 - https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=225099
 - https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=225559

There are patches (by me) attached to the latter bug, which fix **some** of the issues, but not all of them. I don't have much to test with / against, so basically I'm chasing down-downstream bug reports and running the freerdp with various command-line options until it stops giving me command-line-parsing errors.

The affected version is 2.0.0-rc1; I haven't tried -rc2.

```
xfreerdp /network:lan /port:3389 /v:127.0.0.1
```

This produces -- on FreeBSD -- output that indicates that the command-line parameters are wrong. Removing the /network option fixes it (in the sense that it then doesn't give an error about command-line parameters, and goes on to try to connect to an RDP server I don't have).

The main issue is that errno isn't reset by successful calls to strtol(), while it is with (Linux) GLibC. I'll see if I can whip up an example program to show that.

kevans91:
I mention this in the mentioned PR 225559, but all of the strtol, stroul, strold usage throughout FreeRDP should be corrected to set errno beforehand if they're going to do an errno check after the fact, as per POSIX recommendations [1]. POSIX [1] also backs up Adriaan's last statement as POSIXly correct behavior- "These functions shall not change the setting of errno if successful."

Whipping something up with Coccinelle to correct these invocations throughout FreeRDP might be a good thing to do given the number of calls- a quick `grep` seemed to indicate that they all basically follow the same pattern of fail if `errno != 0` without having reset errno beforehand.

[1] http://pubs.opengroup.org/onlinepubs/9699919799/functions/strtol.html

PR #4598, opened 2018-04-27
Fix strto* usage by setting errno before-hand and interpreting EINVAL properly

kevans91:
For glibc, the previous usage of these things all worked. As pointed out in #4597, these are not POSIXly correct invocations. (See the first two comments for a description of the problem and further diagnosis)

Further examination after issue #4597 revealed that not *all* invocations were incorrect; some were assumed to be on numeric strings and thus didn't check errno, while others were properly setting errno beforehand. This PR adds more errno resets and fixes the two instances of not accepting EINVAL that existed in the cmdline parsing.

I wrote a small Coccinelle patch (available here [1]) to find some of these, while others I found by hand. My Coccinelle skills suck, though, so I ended up with some false positives that I had to correct.

[1] https://people.freebsd.org/~kevans/freerdp-strto.cocci

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -1373,8 +1373,9 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 
 				if (p)
 				{
+					unsigned long val;
 					errno = 0;
-					unsigned long val = strtoul(&p[1], NULL, 0);
+					val = strtoul(&p[1], NULL, 0);
 
 					if ((errno != 0) || (val == 0) || (val > UINT16_MAX))
 						return COMMAND_LINE_ERROR_UNEXPECTED_VALUE;
@@ -1411,8 +1412,9 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 
 				if (*(p2 + 1) == ':')
 				{
+					unsigned long val;
 					errno = 0;
-					unsigned long val = strtoul(&p2[2], NULL, 0);
+					val = strtoul(&p2[2], NULL, 0);
 
 					if ((errno != 0) || (val == 0) || (val > UINT16_MAX))
 						return COMMAND_LINE_ERROR_UNEXPECTED_VALUE;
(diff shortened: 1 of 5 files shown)


kevans91:
@akallabeth Thanks! I've corrected all instances of mid-scope declaration, I believe.

winpr/libwinpr/sysinfo/cpufeatures/cpu-features.c
```diff
@@ -659,6 +659,7 @@ get_elf_hwcap_from_proc_cpuinfo(const char* cpuinfo, int cpuinfo_len)
 
 	if (cpuArch)
 	{
+		errno = 0;
```
p-pautov: No one is checking "errno" after "strtol" here.

libfreerdp/core/gateway/http.c
```diff
@@ -454,6 +454,7 @@ static BOOL http_response_parse_header_status_line(HttpResponse* response, char*
 		if ((errno != 0) || (val < 0) || (val > INT16_MAX))
 			return FALSE;
 
+		errno = 0;
 		response->StatusCode = strtol(status_code, NULL, 0);
```
p-pautov: It should be "response->StatusCode = val;" instead, "status_code" was already parsed into "val" on line 452.

client/common/cmdline.c
```diff
@@ -1917,9 +1949,10 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 		{
 			long type;
 			char* pEnd;
+			errno = 0;
 			type = strtol(arg->Value, &pEnd, 10);
 
-			if (errno != 0)
+			if ((errno != 0) && (errno != EINVAL))
```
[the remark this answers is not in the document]
kevans91: I wasn't aiming to make any functional changes here- just to make it work with a more POSIX-compliant (in this regard) libc.

I can make this change if you'd like, though- the original submitter of the previously mentioned issue had a patch that also validated type before it was later passed to a function in case it failed both numeric parsing and wasn't one of the present values, but I had intentionally left that out as I considered it ESCOPE for the moment.

libfreerdp/core/gateway/http.c
kevans91: Right, I didn't apply any thought to whether the context made any sense- this was kind of a low priority for me as we're preparing for an upcoming release over in our world. =) ETIME, as it goes. =(

akallabeth:
@freerdp-bot test

akallabeth:
@kevans91 Thank you for your contribution, I've created #4610 to simplify your changes.
We do not want to reset `errno` unnecessarily (that might hide issues in the library)

Issue #4592, opened 2018-04-24
xfreerdp can not built with WITH_POSIX_TIMER on __sun


yurashek:
After establishing the TCP connection the xfreerdp fail on non-existing implementation of timer
...
[16:39:26:709] [909:00000002] [INFO][com.freerdp.core.proxy] - Successfully connected to w10:3389
[16:39:27:673] [909:00000002] [INFO][com.freerdp.gdi] - Local framebuffer format  PIXEL_FORMAT_BGRX32
[16:39:27:673] [909:00000002] [INFO][com.freerdp.gdi] - Remote framebuffer format PIXEL_FORMAT_RGB16
[16:39:27:863] [909:00000002] [INFO][com.winpr.clipboard] - initialized POSIX local file subsystem
[16:39:27:865] [909:00000002] [ERROR][com.winpr.synch.timer] - InitializeWaitableTimer: os specific implementation is missing

bash-4.4$ xfreerdp /buildconfig
This is FreeRDP version 2.0.0-dev3 (n/a)
[...]

PR #4605, opened 2018-05-01
Allow POSIX timer usage on FreeBSD as well

kevans91:
We, too, have a POSIX timer implementation. =)


akallabeth:
Looks like we should move that check out of source into a CMake detection (there is #4592 for sun too)

akallabeth:
Thank you for the contribution, but this pull is superseeded by #4608 which should solve this issue for all OS that implement posix timer.

PR #4606, opened 2018-05-01
Pull in the LibreSSL compatibility patches from FreeBSD

kevans91:
While I don't personally run LibreSSL, these patches were submitted by folks that do use LibreSSL and have tested them for a while now on FreeBSD.


akallabeth:
@freerdp-bot test

PR #4607, opened 2018-05-02
FreeBSD: epoll/timerfd support

kevans91:
FreeBSD has a libepoll-shim port that provides the epoll API as a wrapper of kevent. As such, some extra work is required to locate the headers to add to the include directories and some extra libraries are needed for linking.

5e9c8c4 is included here because DragonflyBSD is FreeBSD-compatible here and actually uses a modified version of our ports tree, so things in our tree are slightly tangled and this was included to allow the epoll-shim bits to work for DragonflyBSD as well.

aca530b is also only tangentially related- it was introduced to our tree at the same time as the epoll support, but it opts to use shm support instead of tmp files here for performance reasons.


akallabeth:
@freerdp-bot test

PR #4610, opened 2018-05-02
Fix #4597: Do string argument checks before a possible strtol

### LINKED ISSUE #4597: FreeBSD strtol() behaves different -- /network:lan doesn't work

akallabeth:
strtol sets errno, which must be reset if it is no abort condition.
Invert the comparisons to avoid that.
Superseeds #4598 

We do not reset `errno` as there must never be a condition where it is not handled.
If somewhere a function sets a code that is not reset it must be handled there.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -1965,23 +1965,22 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 		}
 		CommandLineSwitchCase(arg, "gateway-usage-method")
 		{
-			long type;
+			long type = 0;
 			char* pEnd;
-			type = strtol(arg->Value, &pEnd, 10);
 
-			if (errno != 0)
-				return COMMAND_LINE_ERROR_UNEXPECTED_VALUE;
-
-			if (type == 0)
+			if (_stricmp(arg->Value, "none") == 0)
+				type = TSC_PROXY_MODE_NONE_DIRECT;
+			else if (_stricmp(arg->Value, "direct") == 0)
+				type = TSC_PROXY_MODE_DIRECT;
+			else if (_stricmp(arg->Value, "detect") == 0)
+				type = TSC_PROXY_MODE_DETECT;
+			else if (_stricmp(arg->Value, "default") == 0)
+				type = TSC_PROXY_MODE_DEFAULT;
+			else
 			{
-				if (_stricmp(arg->Value, "none") == 0)
-					type = TSC_PROXY_MODE_NONE_DIRECT;
-				else if (_stricmp(arg->Value, "direct") == 0)
-					type = TSC_PROXY_MODE_DIRECT;
-				else if (_stricmp(arg->Value, "detect") == 0)
-					type = TSC_PROXY_MODE_DETECT;
-				else if (_stricmp(arg->Value, "default") == 0)
-					type = TSC_PROXY_MODE_DEFAULT;
+				type = strtol(arg->Value, &pEnd, 10);
+				if (errno != 0)
+					return COMMAND_LINE_ERROR_UNEXPECTED_VALUE;
 			}
 
 			freerdp_set_gateway_usage_method(settings, (UINT32) type);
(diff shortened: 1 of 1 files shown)


kevans91:
All of the other strto* calls with a following errno check still need to be fixed as per my other patch. One strto* failure cascades into other failures without those, so not all of the errno resets I included were redundant -- including these ones.

akallabeth:
@kevans91 There shouldn't be any cascade as whenever a `errno` is set the command line parser must return a error condition (and the client abort)

kevans91:
Ah, so not these two- there were still a couple other instances outside of cmdline that didn't look like they would cause an abort ?

akallabeth:
@kevans91 Those are channel command line parsing and gateway related (or some other command line utils)

kevans91:
*nod* Fair enough, then. =)

client/common/cmdline.c
```diff
@@ -1965,23 +1965,22 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
 			{
-				if (_stricmp(arg->Value, "none") == 0)
-					type = TSC_PROXY_MODE_NONE_DIRECT;
-				else if (_stricmp(arg->Value, "direct") == 0)
-					type = TSC_PROXY_MODE_DIRECT;
-				else if (_stricmp(arg->Value, "detect") == 0)
-					type = TSC_PROXY_MODE_DETECT;
-				else if (_stricmp(arg->Value, "default") == 0)
-					type = TSC_PROXY_MODE_DEFAULT;
+				type = strtol(arg->Value, &pEnd, 10);
```
p-pautov: May be go a step further and drop numeric values support altogether? I just can't think of a good reason for anyone to use numeric value here.

p-pautov:
I think, "strto*" calls still need to be revisited (in a separate effort). The usage of this API "as is" seems to be too error prone. As was mentioned in original PR it might not reset "errno" on both success and failure. Say, some other call before "strto*" might set "errno" to failure (which might not be critical for that call) and then we might discard a valid number.
It seems that the only reliable way to test "strto*" success/failure is to check whether "pEnd/endptr" pointer has reached the end of input string.

akallabeth:
@p-pautov `errno` values must be handled when they are set. Of course that can be quite taxing, but it is the only way to go. doing some other kind of hackish error checks is not really a good idea.

p-pautov:
@akallabeth I'm not arguing against checking of `errno`. The problem here is that `strtol()` sets it rather frivolously (at least on Linux), making `errno` check after `strtol()` call _almost_ useless. Just checking `errno` is not enough.

```C
#include <stdlib.h>
#include <errno.h>
#include <stdio.h>

void parse_num(const char* str)
{
	strtol(str, NULL, 0);

	if (errno == 0) {
		printf("'%s' is a number\n", str);
	} else {
		printf("'%s' is NOT a number\n", str);
	}
}

int main()
{
	parse_num("banana");

	fopen("non-existent-file", "r");

	parse_num("5");
}
```
The above program prints following on Ubuntu 16.04
```
'banana' is a number
'5' is NOT a number
```
Here is how more reliable wrapper around `strtol()` could look like:
https://github.com/openssh/openssh-portable/blob/master/openbsd-compat/strtonum.c
https://gist.github.com/gocha/623eb5611bd9648fb3024146d6ecc055/

mfleisz:
Merging this as it fixes the initial issue #4597 
We should open a separate issue for the general problem of converting string to numbers as @p-pautov pointed out.

PR #4612, opened 2018-05-02
Follow OpenBSD convention when installing manpages on FreeBSD

kevans91:
Both OS install third party man pages into ${PREFIX}/man/man${SECTION}, rather
than the convention established in the base system of share/man/man${SECTION}.


hardening:
@freerdp-bot can you test this trivial patch.

PR #4614, opened 2018-05-02
Abstract away CMake module installation setting, install differently on FreeBSD

kevans91:
As indicated by 824ace4:

Different platforms/systems may install CMake modules into different places. For
instance, FreeBSD will install modules into ${PREFIX}/share/cmake/Modules while
Linux distributions might install into ${PREFIX}/lib/cmake.

This leaves us with a useful abstraction- consumers only need to care about
the name of their subdirectory, and only one place needs to be patched to change
where this subdirectory lands.


akallabeth:
@freerdp-bot test

PR #4617, opened 2018-05-03
Upstream the rest of our local FreeBSD patching

kevans91:
FreeBSD/DragonflyBSD are also UNIX platforms, as far as trio is concerned, and we don't need to set the CMAKE_INSTALL_RPATH on FreeBSD.


akallabeth:
@freerdp-bot test

saper:
Big thanks to @kevans91 and everyone on working on this. Today I was able to build master (c5b84db) on FreeBSD just with `cmake .` and `make` and it is working as expected. My previous issues related to NLA with Windows Server 2016 (probably similar to #4487) disappeared.

Thank you for all the work!

kevans91:
@saper Good to hear, thanks! =) It got kind of tiring rerolling these patches in the ports tree for every update.

PR #4619, opened 2018-05-03
Corrected a bug in scripts/autoformat: interpreting a bash script with sh!

informatimago:
Renamed bash and sh scripts removing nefarious file extensions.

Unix scripts should not have any file extension.

    They unnecessarily expose implementation detail (breaking encapsulation).
    They uselessly and incompletely mimic detail from the #! line.
    They capture insufficient detail to be useful at the system level (and aren't used).
    They clash with recommended Unix (and Linux) practice.
    They add noise to the command-level API.
    They are very commonly technically incorrect for the script.
    They give incorrect impressions about the use of the files they adorn.
    They aren't validated even for what little info is present in them.
    They interfere with switching scripting languages.
    They interfere with changing scripting language versions.
    They interfere with changing to presumably-faster compiled forms.
    They encourage naively running scripts with the extension-implied interpreter.
    They infect novice scripters with misinformation about Unix scripting.
    They ironically are only a problem when interpreted by humans.

See more at:

    http://www.talisman.org/~erlkonig/documents/commandname-extensions-considered-harmful

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -881,6 +881,16 @@ find_feature(FAAC ${FAAC_FEATURE_TYPE} ${FAAC_FEATURE_PURPOSE} ${FAAC_FEATURE_DE
 
 find_feature(GSSAPI ${GSSAPI_FEATURE_TYPE} ${GSSAPI_FEATURE_PURPOSE} ${GSSAPI_FEATURE_DESCRIPTION})
 
+if (WITH_FFMPEG AND NOT FFmpeg_FOUND)
+	message(FATAL_ERROR "FFMPEG support requested but not detected")
+endif()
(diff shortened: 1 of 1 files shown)


informatimago:
I have a following commit 3280482 on autoformat ; perhaps you'll want to merge it in the same PR?

akallabeth:
@freerdp-bot test

informatimago:
The name with .sh is used in the jenkins script.
We may either correct the jenkins script, or leave the android script with the .sh extension.
What do you want to do?

akallabeth:
@informatimago Would wait with additional commits on this one, feedback sometimes takes quite some time :/
@freerdp-bot test

informatimago:
See followup PR #4699

Issue #4593, opened 2018-04-25
Unable to share drives


mahmoodn:
On centos 7, I want to connect to a windows server via xfreerdp. The following command works properly

`# xfreerdp -u administator 10.1.1.2`

However, I can not share the host drives. 

```
# xfreerdp +drives -u administator 10.1.1.2
transport_connect:getaddrinfo (Name or service not known)
Error: protocol security negotiation failure
```

The version is 1.0.2 and the freerdp-plugins package has been installed.

ondrejholy:
This new syntax doesn't work in stable-1.0 versions. Try  `--plugin rdpdr --data disk:<Name>:<Path> --` instead of `+drives`.

By the way, I've updated xfreerdp man page by commit https://github.com/FreeRDP/FreeRDP/commit/0c050bb8cde65b0ac368f06bfd5ec36cccf87a83 and the update is part of RHEL 7.5...

ondrejholy:
Hmm, https://github.com/FreeRDP/FreeRDP/wiki/Plugins contains the following:

> Note: With freerdp version 1.0.2 the parameter "disk" is going to be replaced by "drive" for compatibility reasons to MS RDP.

But 1.0.2 still obviously uses `--plugin rdpdr --data disk:<Name>:<Path> --`...

And also please be aware of:

> If you want any redirection to work with Windows Server 2012 you MUST use --plugin rdpsnd before you use any rdpdr options.

(Windows Server 2012 or newer)

ondrejholy:
It should be before, not instead of it, so `xfreerdp --plugin rdpsnd --plugin rdpdr --data disk:home:/home -- -u administrator 10.1.1.2` if I am not mistaken.

> What should I write in the network map page of windows? is that the sharename only (home) or \sharename (\home)?

I don't know and don't have time to test, maybe somebody from maintainers would know...?

ondrejholy:
Also be sure that you properly logged out the rdp session (e.g. over start menu) before you are going to try the redirection.

mahmoodn:
OK. That command says no error.

```
# xfreerdp --plugin rdpsnd --plugin rdpdr --data disk:home:/home -- -u administrator 10.1.1.2
loading plugin rdpsnd
loading plugin rdpdr
connected to 10.1.1.2:3389
Password:
registered device #1: home (type=8 id=1)

```
However, still I can not access the drive in windows. In map network drive, I tried
\\home
and 
\\10.1.1.1\home
[...]

PR #4626, opened 2018-05-04
Fix drive redirection in manpage

ondrejholy:
Man page has been updated according to the wiki. Unfortunately, wiki
mentioned "drive" subplugin for 1.0.2, but stable-1.0 still uses "disk".
Let's fix the man page accordingly.

Closes: https://github.com/FreeRDP/FreeRDP/issues/4593


bmiklautz:
Thanks!

Issue #4636, opened 2018-05-10
Maintain consistent help wording with +/- command line attributes


ghost:
For the release version of FreeRDP, the wording of the help text should be made consistent with respect to the +/- attributes. Some of the descriptions use negation while others are just explanations with default settings. Others appear to be incorrect. I will provide several examples of this.

// here is a case where there is an enable explanation with default settings
+aero                       Enable desktop composition (default:off)

// here is a case where there is no enable/disable explanation with default settings
+async-input                Asynchronous input (default:off)

// here is a case where there is an enable explanation and it should be disable with default settings
-bitmap-cache               Enable bitmap cache (default:on)

I realize this is a trivial item, however I thought it might be confusing to newcomers. I would suggest you stick to one way of explaining the option and updating the incorrect ones.

Great work!

PR #4657, opened 2018-05-16
make help wording consistent with +/- command line attributes 

hendwolt:
make help wording consistent with +/- command line attributes (implements #4636)
- I synced +/- with enable/disable.
- Every line with +/- contains now one of the two words enable/disable.

Please check the results. There are a few lines, which sound "not good"; at least to me.
Especially:
   +gestures                   Enable consume multitouch input locally (default:off)
   +unmap-buttons              Enable letting the server see real physical pointer button (default:off)


hendwolt:
That means, #4636 is invalid?

hendwolt:
Or should we use both words?
instead of "Disable bitmap cache"
something like "Enable(+) or disable(-) bitmap cache"
?

akallabeth:
The latter sounds more reasonable

akallabeth:
@freerdp-bot test

akallabeth:
@hendwolt Thanks for the pull, and since I don't only want to complain, I've created a patched version here https://github.com/akallabeth/FreeRDP/tree/wording_consistent
what do you think? Wording might still be changed, but it is no longer necessary to change the text when the default changes (and that is often forgotten)

hendwolt:
@akallabeth 
I like the idea, that enable/disable is set automatically according to the default values.
Will it be clear enough - from enable/disable, respective "+"/"-" - what the actual default is?

hendwolt:
For some BOOL options, which did not have enable/disable until now, I'd consider little text changes. I'm not a native English speaker, but e.g.:
"Enable Redirect multitouch input" sounds not good to me.
I like "Enable redirection of multitouch input" better.

akallabeth:
@hendwolt you're right, the text needs refinement :)

akallabeth:
Replaced by #4681

PR #4701, opened 2018-06-06
Cleaned up const char** -> char** for argv, since we definitely do modify the argv!

informatimago:
We overwrite the password and pin arguments.
This implies changes in the argument parsing tests that now must pass a mutable argv
(copied from the statically declared test argvs).
Some other const inconsistency have been dealt with too.

Note there remains a const problem with structure literals declared in ntlm.c; I'll post an issue about it.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -589,7 +589,14 @@ static char** freerdp_command_line_parse_comma_separated_values_ex(const char* n
 
 			if (p)
 			{
-				p[0] = name;
+				p[0] = strdup(name);
+
+				if (!p[0])
+				{
+					free(p);
+					return NULL;
+				}
+
 				*count = 1;
 				return p;
 			}
@@ -1037,7 +1044,7 @@ int freerdp_map_keyboard_layout_name_to_id(char* name)
 }
 
 static int freerdp_detect_command_line_pre_filter(void* context, int index,
-        int argc, LPCSTR* argv)
+        int argc, LPSTR* argv)
 {
 	int length;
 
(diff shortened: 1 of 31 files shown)


client/common/cmdline.c
```diff
@@ -589,7 +589,14 @@ static char** freerdp_command_line_parse_comma_separated_values_ex(const char* n
 
 			if (p)
 			{
-				p[0] = name;
+				p[0] = strdup(name);
```
akallabeth: why the duplication here?

informatimago: Because name is declared cons char*, but the function returns char** p.
name comes from calls to freerdp_command_line_parse_comma_separated_values_offset with either literal arguments, or arg->Name.  So it's better to create a mutable string in all cases.

The alternative would be to return a const char**  but this result is used in a dozen or more places, where const would have to be added.

Granted, most of the strings in those argv arrays are never mutated, only the passwords are (AFAIK), but if we want the help of the compiler validating that, using const or not, we need to do this. (unfortunately, we cannot specify that some slots of the array are mutable and some are not; we'd have to use a struct for that).

winpr/libwinpr/utils/test/TestCmdLine.c
```diff
@@ -65,7 +65,9 @@ int TestCmdLine(int argc, char* argv[])
 	long height = 0;
 	COMMAND_LINE_ARGUMENT_A* arg;
 	flags = COMMAND_LINE_SIGIL_SLASH | COMMAND_LINE_SEPARATOR_COLON | COMMAND_LINE_SIGIL_PLUS_MINUS;
-	status = CommandLineParseArgumentsA(testArgc, testArgv, args, flags, NULL, NULL, NULL);
+	int testArgc = string_list_length(testArgv);
```
mfleisz: VS 2010 doesn't like mid-code variable declarations.

informatimago:
@mfleisz corrected (I'm so used to C11 now, I couldn't even see the problem).

akallabeth:
@freerdp-bot test

hardening:
@akallabeth the PR looks globally good to me, was the answer of @informatimago ok for you (and if so merging would be nice) ?

client/common/cmdline.c
akallabeth: @informatimago that is a memory leak, sorry.

akallabeth:
Fixed the remaining issue with the memory leak in #4711.
Thank you for your contribution.

PR #4706, opened 2018-06-08
X11: fixed Right-Ctrl ungrab feature (PR #3622)

khvMX:
Since last week I have been having issues with a feature I introduced in this PR: https://github.com/FreeRDP/FreeRDP/pull/3622

Seems like XOrg update has changed the behavior somehow, and either more events got generated by mouse than before, or something similar. Adding this line fixed the issue for me.

Testing: on X11 client (not Wayland), connect to RDP server. Press Right Ctrl once. After it is released, your keyboard grab must be in your host system.
These are my most popular scenarios to use this feature, after releasing the grab:
- Alt+F4 to disconnect
- Alt+Tab or Super to switch to other window in the host
- Ctrl+Alt+Up/Down to switch workspaces (I'm using Gnome 3)

This must be working in both Fullscreen and windowed modes. Works for me.

### DIFF
--- a/client/X11/xf_keyboard.c
+++ b/client/X11/xf_keyboard.c
@@ -596,7 +596,7 @@ void xf_keyboard_handle_special_keys_release(xfContext* xfc, KeySym keysym)
 		{
 			xf_toggle_control(xfc);
 		}
-
+		xfc->mouse_active = FALSE;
 		XUngrabKeyboard(xfc->display, CurrentTime);
 	}
 


khvMX:
@codemuse have you tried to apply this patch? I am also using arch and this also was an xorg update which broke the right-ctrl ungrab for me.. It's working for me with this patch, but I am really not an X11 expert to tell what is actually wrong..

codemuse:
@khvMX just tried and yep, the patch works for me as well!

akallabeth:
@freerdp-bot test

QiangF:
This doesn't work for me. I press Alt-Tab short after right Ctrl, but the key combination is not received by my window manager (EXWM). Is it possible to extend the ungrab time?

xenotropic:
I'm currently experiencing right-Ctrl ungrab as not working with Ubuntu 22.10, X11 v 1.21.1.4, FreeRDP 2.8.1 . Don't really have time (or skills) to debug but wanted to put it out there in case anyone else can confirm.

ghollisjr:
> I'm currently experiencing right-Ctrl ungrab as not working with Ubuntu 22.10, X11 v 1.21.1.4, FreeRDP 2.8.1 . Don't really have time (or skills) to debug but wanted to put it out there in case anyone else can confirm.

It seems to be working for me on Ubuntu 23.10

PR #4710, opened 2018-06-19
Reconnect fixes

akallabeth:
* Fixes unnecessary `Stream_Free` in channels (The stream is pool collected after the call returns with `Stream_Release`)
* Separates channel from client state (Follow up to redirect fixes in #4631): 
  1. If the client is ready to utilize the channels, call `freerdp_channels_post_connect` on reconnect or redirect
  1. If the client is not readily initialized, skip that


include/freerdp/freerdp.h
```diff
@@ -240,7 +240,9 @@ struct rdp_freerdp
 								   Callback for context deallocation
 								   Can be set before calling freerdp_context_free() to have it executed before deallocation.
 								   Must be set to NULL if not needed. */
-	UINT64 paddingC[48 - 35]; /* 35 */
+	UINT64 paddingC[47 - 35]; /* 35 */
+
+	ALIGN64 ConnectionCallbackState; /* 48 */
```
hardening: So the offset is 47, isn't it ? (IIRC the number in the comment is used by some scripts)

libfreerdp/core/connection.c
```diff
@@ -437,21 +435,19 @@ BOOL rdp_client_reconnect(rdpRdp* rdp)
...
-	connected = channels->connected;
 
 	if (!rdp_client_disconnect_and_clear(rdp))
 		return FALSE;
 
 	status = rdp_client_connect(rdp);
 
-	if (status && connected)
+	if (status && (context->instance->ConnectionCallbackState == 2))
 		status = (freerdp_channels_post_connect(channels, context->instance) == CHANNEL_RC_OK);
```
hardening: Not your code, but wouldn't it be nice to have some enum instead of raw values for ConnectionCallbackState ?

include/freerdp/freerdp.h
hardening: not having a variable name here is weird isn't it ?

akallabeth:
@freerdp-bot test

PR #4711, opened 2018-06-19
Command line cleanup rework

akallabeth:
Fixed version of #4701


akallabeth:
@hardening could you check and merge this? I've only fixed the memory leak @informatimago added in the command line parsing.

PR #4724, opened 2018-07-02
Wave2 server side support

akallabeth:
Automatically switches to `WAVE2 PDU` if client supports it.


channels/rdpsnd/server/rdpsnd_main.c
```diff
@@ -65,8 +66,8 @@ UINT rdpsnd_server_send_formats(RdpsndServerContext* context, wStream* s)
 		Stream_Write_UINT16(s, context->server_formats[i].nChannels); /* nChannels */
 		Stream_Write_UINT32(s,
 		                    context->server_formats[i].nSamplesPerSec); /* nSamplesPerSec */
-		Stream_Write_UINT32(s, context->server_formats[i].nSamplesPerSec *
-		                    context->server_formats[i].nChannels *
+		Stream_Write_UINT32(s, context->server_formats[i].nSamplesPerSec*
+		                    context->server_formats[i].nChannels*
```
mfleisz: imo the old formatting was more clear in this case ...

channels/audin/client/audin_main.c
```diff
@@ -221,25 +220,33 @@ static UINT audin_process_formats(IWTSVirtualChannelCallback* pChannelCallback,
...
 		Stream_Read_UINT16(s, format.cbSize);
-		format.data = Stream_Pointer(s);
 
 		if (Stream_GetRemainingLength(s) < format.cbSize)
 			return ERROR_INVALID_DATA;
 
-		Stream_Seek(s, format.cbSize);
+		if (format.cbSize > 0)
+		{
+			Stream_Seek(s, format.cbSize);
```
hardening: Don't we have a pb here, shouldn't you seek _after_ doing the copy ?

akallabeth:
@hardening Thx, you spotted some nasty issue.

akallabeth:
@hardening and the common part, I've added it before I realized we only need a common header.

PR #4738, opened 2018-07-10
Automatic timeout reconnect after callbacks

akallabeth:
The following scenario currently does not work:
1. Connect to a server `xfreerdp /v:foo /u:bar`
1. Wait `>1m` to accept the certificate or enter the password
1. See the connection not being established

With this pull 
* `freerdp_reconnect` is capable of handling situations, where `PostConnect` callback has not been called.
* Certificates are remembered to not call `VerifyCertificate` et al if already accepted on reconnect
* Sample code has been added to the X11 client to handle automatic reconnect.


libfreerdp/crypto/tls.c
```diff
@@ -1384,6 +1502,15 @@ int tls_verify_certificate(rdpTls* tls, CryptoCert cert, char* hostname,
 		crypto_cert_subject_alt_name_free(alt_names_count, alt_names_lengths,
 		                                  alt_names);
 
+	if (verification_status > 0)
```
mfleisz: verification_status is a BOOL

PR #4741, opened 2018-07-11
Prevent FindFirstFileA from accessing parent directory

akallabeth:
`FindFirstFileA` did open the parent of a directory to be opened.
This is fine on most systems, but if AppArmor or SELinux is used for access control this might prevent the function from succeeding.

Now we check if the path is an existing directory and set the `pFileSearch->lpPath` and `pFileSearch->lpPattern` variables accordingly.

### DIFF
--- a/winpr/libwinpr/file/generic.c
+++ b/winpr/libwinpr/file/generic.c
@@ -833,9 +833,8 @@ static BOOL FindDataFromStat(const char* path, const struct stat* fileStat,
 
 HANDLE FindFirstFileA(LPCSTR lpFileName, LPWIN32_FIND_DATAA lpFindFileData)
 {
-	LPSTR p;
-	size_t index;
-	size_t length;
+	int oldErrno;
+	BOOL isDir = FALSE;
 	struct stat fileStat;
 	WIN32_FILE_SEARCH* pFileSearch;
 	ZeroMemory(lpFindFileData, sizeof(WIN32_FIND_DATAA));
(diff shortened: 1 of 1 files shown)


winpr/libwinpr/file/generic.c
```diff
@@ -801,58 +846,81 @@ HANDLE FindFirstFileA(LPCSTR lpFileName, LPWIN32_FIND_DATAA lpFindFileData)
 		return INVALID_HANDLE_VALUE;
 	}
 
-	/* Separate lpFileName into path and pattern components */
-	p = strrchr(lpFileName, '/');
-
-	if (!p)
-		p = strrchr(lpFileName, '\\');
+	oldErrno = errno;
```
mfleisz: Why do we preserve errno here?

akallabeth: right, could remove the restore and simply set to `0`

akallabeth:
@freerdp-bot test

winpr/libwinpr/file/generic.c
```diff
@@ -785,11 +785,54 @@ struct _WIN32_FILE_SEARCH
 };
 typedef struct _WIN32_FILE_SEARCH WIN32_FILE_SEARCH;
 
+static BOOL FindDataFromStat(const char* path, const struct stat* fileStat,
+                             LPWIN32_FIND_DATAA lpFindFileData)
+{
+	UINT64 ft;
+	char* lastSep;
+	lpFindFileData->dwFileAttributes = 0;
```
MartinHaimberger: Since its an api, we need to check the pointer parameters at least for NULL

if (!lpFindFileData)
 		return FALSE;

winpr/libwinpr/file/generic.c
```diff
@@ -785,11 +785,54 @@ struct _WIN32_FILE_SEARCH
...
+
 HANDLE FindFirstFileA(LPCSTR lpFileName, LPWIN32_FIND_DATAA lpFindFileData)
 {
-	LPSTR p;
-	size_t index;
-	size_t length;
+	BOOL isDir = FALSE;
 	struct stat fileStat;
 	WIN32_FILE_SEARCH* pFileSearch;
 	ZeroMemory(lpFindFileData, sizeof(WIN32_FIND_DATAA));
```
MartinHaimberger: Since its an api, we need to check the pointer parameters at least for NULL

if (!lpFindFileData)
return INVALID_HANDLE_VALUE;

winpr/libwinpr/file/generic.c
```diff
@@ -962,7 +1027,6 @@ BOOL FindNextFileA(HANDLE hFindFile, LPWIN32_FIND_DATAA lpFindFileData)
 	char* fullpath;
 	size_t pathlen;
 	size_t namelen;
-	UINT64 ft;
 	ZeroMemory(lpFindFileData, sizeof(WIN32_FIND_DATAA));
```
MartinHaimberger: need check here too

winpr/libwinpr/file/generic.c
akallabeth: Internal function, already checked

winpr/libwinpr/file/generic.c
akallabeth: Already done by the call to `stat` (returns a corresponding error code)

PR #4758, opened 2018-07-19
Audio/microphone fixes

akallabeth:
Changes:
* Fixes a bug where `freerdp_dsp_decode` and `freerdp_dsp_encode` overwrite data written previously to the output stream
* Reintroduce dynamic logging
* Fix some codec mappings for FFMPEG
* Adjust receive buffers
* Fix sample buffering for microphone redirection (FFMPEG)

TODO: (maybe in follow up pull)
* `FFMPEG ADPCM` codecs do not work
* `DSP libgsm, libfaac, liblame` path does not buffer samples

Help:
* Anyone who is annoyed by bad sound/microphone quality, thest this! `WITH_DSP_FFMPEG` should give you perfect sound!

### DIFF
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -696,6 +696,13 @@ static UINT audin_on_new_channel_connection(IWTSListenerCallback* pListenerCallb
 static UINT audin_plugin_initialize(IWTSPlugin* pPlugin, IWTSVirtualChannelManager* pChannelMgr)
 {
 	AUDIN_PLUGIN* audin = (AUDIN_PLUGIN*) pPlugin;
+
+	if (!audin)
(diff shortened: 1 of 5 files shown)


channels/audin/client/audin_main.c
```diff
@@ -607,14 +626,14 @@ static UINT audin_on_close(IWTSVirtualChannelCallback* pChannelCallback)
 	AUDIN_CHANNEL_CALLBACK* callback = (AUDIN_CHANNEL_CALLBACK*) pChannelCallback;
 	AUDIN_PLUGIN* audin = (AUDIN_PLUGIN*) callback->plugin;
```
MartinHaimberger: Why is the check 

	if (!callback || !data)

done in audin_on_data_received and not in audin_on_close?

should all functions be checked which can be called from outside, or via callback ?

PR #4759, opened 2018-07-19
audin: Fix server-side dsp decoding

mfleisz:
The decoder format was never set and so dsp_decode always failed
on the server side.


channels/audin/server/audin.c
```diff
@@ -82,6 +82,12 @@ static UINT audin_server_select_format(audin_server_context* context,
 	}
 
 	context->selected_client_format = client_format_index;
+	if (!freerdp_dsp_context_reset(audin->dsp_context,
+								   &audin->context.client_formats[client_format_index]))
```
akallabeth: Indentation looks broken here

Issue #3666, opened 2016-12-24
Freezes the screen xfreerdp in LinuxMint when switching from full screen mode (Ctrl+Alt+Enter)


ava57r:
Use - version This is FreeRDP version 2.0.0~dev (git n/a)
OS - Linux Mint 18.1 Serena
Desktop - Cinnamon 3.2.7

Connect to Windows server 2008 RDP gateway with option `/f`
When switching from full screen mode pressing (Ctrl+Alt+Enter) 3 times 
1. Full screen mode -> window mode
2. Window mode -> Full screen mode
3. Full screen mode - window mode
*FreeRDP freezes*
Lost icons and panels, window freerdp becomes transparent

syslog - empty
run terminal (Ctrl+Alt+F1) and kill proccess

akamuza:
Ubuntu 16.04, Intel graphics G33 , FreeRDP version 2.0.0-dev (latest nightly build).
Same problem.
After switching several times to/from fullscreen - screen just freezes and nothing helps but reboot. 
Using +toggle-fullscreen doesn't help.

wolfstefan:
I figured out that this issue was introduced with the commits 48be0815c1f12f1f268135a0a214b327d79a9457 and 7ddd15d8deab94a8a2c1d61c792188fa7b6e21a2. To be more precise, calling XMoveWindow before xf_SendClientEvent seems to break it.

The attached patch removes all calls of XMoveWindow in this case and fixes the problem as far as I have tested it. Moreover, it prevents the window from getting lower after each switch between fullscreen and non-fullscreen. However, I think that there is a reason for the XMoveWindow calls and therefore this patch might break something else. It would be nice if one of the developers could take a look at it.

P.S.: thanks to all developers for the amazing work on FreeRDP :)

[fullscreen.diff.txt](https://github.com/FreeRDP/FreeRDP/files/1282309/fullscreen.diff.txt)

akamuza:
In my case XMoveWindow  is a good idea, because when I exit fullscreen for the very first time then the top bar of my rdp window is hidden by Ubuntu's top toolbar, and thus I can not see my close,minimize buttons of the rdp window.
And because of that I have to enter/exit fullscreen once more in order to move rdp window a bit lower.

Of cause that's not a big deal, jast an additional information on XMoveWindow

wolfstefan:
[...]

PR #4763, opened 2018-07-22
Fix for #3666

mmattes:
According to https://specifications.freedesktop.org/wm-spec/wm-spec-latest.html#idm140200472531472 

> _NET_WM_STATE_FULLSCREEN indicates that the window should fill the entire screen and have no window decorations. **Additionally the Window Manager is responsible for restoring the original geometry after a switch from fullscreen back to normal window.**

I think that the cinnamon desktop handles this and that the additional move to position 0, 0 causes issues here. Not moving the window after the event is send fixes this issue in cinnamon. I additionally tested the change in XFCE and Mate and it seems to work without any problems.

### DIFF
--- a/client/X11/xf_window.c
+++ b/client/X11/xf_window.c
@@ -209,12 +209,6 @@ void xf_SetWindowFullscreen(xfContext* xfc, xfWindow* window, BOOL fullscreen)
 		                   fullscreen ? _NET_WM_STATE_ADD : _NET_WM_STATE_REMOVE,
 		                   xfc->_NET_WM_STATE_FULLSCREEN, 0, 0);
 
-		if (!fullscreen)
-		{
-			/* leave full screen: move the window after removing NET_WM_STATE_FULLSCREEN */
-			XMoveWindow(xfc->display, window->handle, startX, startY);
-		}
-
 		/* Set monitor bounds */
 		if (settings->MonitorCount > 1)
 		{
--- a/client/X11/xf_window.c
+++ b/client/X11/xf_window.c
@@ -209,6 +209,16 @@ void xf_SetWindowFullscreen(xfContext* xfc, xfWindow* window, BOOL fullscreen)
 		                   fullscreen ? _NET_WM_STATE_ADD : _NET_WM_STATE_REMOVE,
 		                   xfc->_NET_WM_STATE_FULLSCREEN, 0, 0);
 
+		if (!fullscreen)
+		{
+			/* Fix for Cinnamon WM see ISSUE 3666 */
+			if (startX == 0)
+				startX = 1;
+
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

mmattes:
> The window is not restored to its previous position after applying your patch.
> On multimonitor setups it even restores on the wrong monitor.
Hm... on the website i referenced it says that the window manager is responsible for restoring the window to its previous state. So isnt your discovery then a bug within cinnamon which FreeRDP should **not** aim to fix? 

The only other solution I could think of is to add some sleep time before moving the window, i can give that a try this afternoon.

akallabeth:
@mmattes It might indeed be a bug in window managers (or disagreement, pick your poison ;)), sadly they all misbehave (the ones I tried at least). This way the behaviour is at least somewhat the one expected.

Anyway, the referenced issue with the transparent/frozen window seems to be of a different nature, as the window may be moved by code at any time without the wm to do block something, so this fix is just hiding what is causing the issue anyway.

Have you checked we are not in some kind of recursive loop through X11 event generation in an X11 event handler? That might be the actual cause of the issue and be fixable without manipulating fullscreen switch behaviour.

mmattes:
I haven't had a look for an recursive event loop yet, will have a look later. What I can say so far is that it would work if the window is moved to x = 1 and y = 0. Rather then x = 0 and y = 0.

akallabeth:
@mmattes Ah, ok. Might simply be an issue if we don't actually move the window then?

mmattes:
@akallabeth this is, after a lot of trying and reading the only solution i could come up with. It keeps the old behavior and fixes the issue with Cinnamon WM. 

Please note that toggling full-screen back and forth many times will slowly move the window down as XMoveWindow moves the position of the window including decorator (green mark blow) but the ConfigureNotify event gives the top/left position of the window without decorator (red mark below). I haven't found anything which tells me how large the decorator is so i left this behavior as it is. 

image

akallabeth:
@mmattes played a bit and found (again) issues with the fullscreen stuff.
Could you try https://github.com/akallabeth/FreeRDP/tree/fullscreen_cinnamon and report if it works for you?
Should fullscreen and restore on the correct monitor now (using cinnamon here on fedora 28)

mmattes:
@akallabeth anything i can do to get this pushed forward to get a solution?

akallabeth:
@mmattes trial and error so far, have not found something that works better in all cases

dannymcc:
@mmattes Hi, did you ever have any luck with this? We are using FreeRDP to access software from VetZ and seem to be struggling with windows resizing crashing the session. Did you ever make any further progress?

Neustradamus:
Have you progressed on it?

akallabeth:
is this issue still relevant?
Is this still an issue with `HIGH DEF` rails support?

PR #4765, opened 2018-07-24
Removed double ifdef WITH_XI, XInput2.h already included

mmattes:
Not really an issue but i just saw it and cleaned it


akallabeth:
@freerdp-bot test

PR #4772, opened 2018-07-30
macosx: Bring newly launched freerdp window to front

Mandar-Shinde:
To bring newly launched freerdp window on top of all the windows.


akallabeth:
@freerdp-bot test

PR #4788, opened 2018-08-08
Fixed #4786: Added fullscreen handling for dynamic resolution.

### LINKED ISSUE #4786: Full-screen Multimonitor Bug
I've been using the latest master version for months now to access my home computer with 3 monitors using the following command from Ubuntu 16.04:
xfreerdp +auto-reconnect /frame-ack:0 /bpp:16 -decorations /dynamic-resolution /f +fonts /gdi:hdi +home-drive -themes -wallpaper +clipboard /u:*** /multimon:3 /p:*** /rfx /rfx-mode:video +drives /disp /v:201.**.**.**:3389
It has been working perfectly but when I downloaded the latest updates to the master branch today and compiled I started having a bug where starting the remote connection with the command above works perfectly, but when I exit full-screen mode FreeRDP would switch to single monitor mode and it would never work on multiple monitors again until I restart the whole application.
This definitely didn't happen up to two weeks ago.

akallabeth:
Fullscreen was currently not properly handled for dynamic resolution.
With this pull the fullscreen flag is checked and the correct monitor layout sent to the server.
In case of switching to windowed mode the existing behaviour is used (single fake monitor with the window resolution)

### DIFF
--- a/client/X11/xf_disp.c
+++ b/client/X11/xf_disp.c
@@ -47,10 +47,13 @@ struct _xfDispContext
 	int targetWidth, targetHeight;
 	BOOL activated;
 	BOOL waitingResize;
+	BOOL fullscreen;
 	UINT16 lastSentDesktopOrientation;
 	UINT32 lastSentDesktopScaleFactor;
 	UINT32 lastSentDeviceScaleFactor;
 };
+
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

Issue #4756, opened 2018-07-18
Connection Bar for xfreerdp same as windows mstsc client


gnshkl1:

mmattes:
I started to work on this. See https://github.com/mmattes/FreeRDP/tree/feature/fullscreen-bar for an initial version. It already can do close, minimize or toggle full-screen and resizing and dragging it around also works. I still do have stuff to do on it to get it ready to be pulled in e.g. auto-hide or make sure that i fulfill the coding guidelines of the project. 

BTW: this is actually the first time i am coding something in C. Any feedback if i made somewhere major mistakes in regards to coding in C would be appreciated.

akallabeth:
@mmattes Looks good overall, just 2 little hints:
* We try to stay compatible with older compilers, so no C++ comments or variable declarations mid code
* Use `./scripts/format_code.sh` (astyle) script which formats code according freerdp standards

gnshkl1:
@mmattes Good work can you make it configurable from freerdp commandline

gnshkl1:
@mmattes active by default and command line switch to disable it will be good option

mmattes:
@gnshkl1 have a look at https://github.com/mmattes/FreeRDP/commit/37d84751b331cac1e53f3bac66f2a0b382f1131c where i added a new option to disable the floatbar. But i got some queries. 
- settings.h seems to be, at least partially, auto generated. I couldn't figure out in detail ho,  why i am not sure if the new option i added is correct?
- cmdline.h seems to be common for all clients, i have seen that there is a floatbar in the windows client why i also implemented changes there to apply the new option, but I cant verify them. See commit https://github.com/mmattes/FreeRDP/commit/20316cacdeeacb8f9825a11244418ebdea387397

mmattes:
Totally forgot, i have no devices to test if my floatbar implementation within the X11 client is working properly in multimonitor mode and not shown at all when it is a RemoteApp. Can anyone verify that?

drnuriev:
Hi. How to disappear connection bar on Thinstation (xfreerdp)? Is there some parametr? Thanks.

PR #4805, opened 2018-08-20
Feature/fullscreen bar for X11

mmattes:
Here the PR for the floatbar for x11. This fixes Issue #4756 
What still needs to be verified is: 
- does it work in multimonitor mode?
- does it work with remote app, what means the floatbar should not be there at all?
- are the changes I made to the windows client correct so that the floatbar is deactivated if the new setting is used? 

Also let me know if any of my coding is bad as this is the first thing i really implemented using C


client/X11/resource/close.xbm
```diff
@@ -0,0 +1,9 @@
+#define close_width 24
+#define close_height 24
+static unsigned char close_bits[] = {
```
akallabeth: Formatting, run `./scripts/format_code.sh`

client/X11/resource/lock.xbm
```diff
@@ -0,0 +1,9 @@
+#define lock_width 24
+#define lock_height 24
+static unsigned char lock_bits[] = {
```
akallabeth: Formatting, run `./scripts/format_code.sh`

client/X11/resource/minimize.xbm
```diff
@@ -0,0 +1,9 @@
+#define minimize_width 24
+#define minimize_height 24
+static unsigned char minimize_bits[] = {
+   0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+   0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
+   0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x3f, 0x00, 0xfc,
```
akallabeth: Formatting, run `./scripts/format_code.sh`

client/X11/resource/restore.xbm
```diff
@@ -0,0 +1,9 @@
+#define restore_width 24
+#define restore_height 24
+static unsigned char restore_bits[] = {
```
akallabeth: Formatting, run `./scripts/format_code.sh`

client/X11/resource/unlock.xbm
```diff
@@ -0,0 +1,9 @@
+#define unlock_width 24
+#define unlock_height 24
+static unsigned char unlock_bits[] = {
```
akallabeth: Formatting, run `./scripts/format_code.sh`

client/X11/xf_event.c
```diff
@@ -1014,6 +1014,14 @@ BOOL xf_event_process(freerdp* instance, XEvent* event)
 		}
 	}
 
+	/* Forward event to the floatbar event handler, if the window
```
akallabeth: You should check the floatbar is actually used here before dispatching events

include/freerdp/settings.h
```diff
@@ -1127,6 +1127,7 @@ struct rdp_settings
 	ALIGN64 BOOL   AsyncChannels;           /* 1546 */
 	ALIGN64 BOOL   AsyncTransport;          /* 1547 */
 	ALIGN64 BOOL   ToggleFullscreen;        /* 1548 */
+	ALIGN64 BOOL   Floatbar;
```
akallabeth: The numbers in the comments are there for a reason, to maintain ABI compatibility replace one of the reserved spots, do not just insert new variables

client/X11/xf_floatbar.c
```diff
@@ -0,0 +1,609 @@
...
+}
+
+void xf_floatbar_hide_and_show(xfContext* xfc)
+{
+	xfFloatbar* floatbar;
+	floatbar = xfc->window->floatbar;
+
+	if (!floatbar->locked)
+	{
+		if (floatbar->mode == 0 && floatbar->last_motion_y_root > 10 && floatbar->y > (FLOATBAR_HEIGHT * -1))
```
akallabeth: Might be a good idea to throw in some braces

akallabeth:
Did some refactoring in https://github.com/akallabeth/FreeRDP/tree/pr/4805, you can cherry pick the commits from that branch as a start.

mmattes:
@akallabeth thank you for your help and insights. Can you explain me why you moved the xfFloatbar struct from the header to the code file? Currently the hide doesnt work anymore and i need  to get your input on why you did that to fix it. 

Also please note that the event handling for the floatbar needs to happen before the main window handles its events and that i on purpose return the function with true afterwards. If i dont return the function then the events for the floatbar and its buttons are handled also by the main window, this has side effects. For example dragging the floatbar around also will cause a drag and drop action in the main window aswell.

akallabeth:
Did move the stuff from the header as it is not used elsewhere (and shouldn't) so basically containment. As for event handling, ok, did not take that to account.

mmattes:
@akallabeth are you happy the way it is now or is there still something i need to have a look at?

client/X11/xf_event.c
```diff
@@ -366,6 +366,9 @@ static BOOL xf_event_MotionNotify(xfContext* xfc, XEvent* event, BOOL app)
 	if (xfc->use_xinput)
 		return TRUE;
 
+	if(xfc->floatbar && !(app))
```
akallabeth: This stuff is part of XEvent handling, couldn't that be done in that function?

client/X11/xf_window.c
```diff
@@ -577,7 +577,7 @@ xfWindow* xf_CreateDesktopWindow(xfContext* xfc, char* name, int width,
 		            settings->DesktopPosY);
 	}
 
-	window->floatbar = xf_floatbar_new(xfc, window->handle, window->width);
+	window->floatbar = xf_floatbar_new(xfc, window->handle, xfc->workArea.width);
```
akallabeth: This must be the center of the screen that the window is actually on (which might change on multimonitor systems)

akallabeth:
@mmattes Tested your pull, looks good except the wrong positioning.
Would suggest you clean that up and stash all changes in a single commit.

client/X11/xf_event.c
mmattes: I need to know in the floatbar where the mouse pointer is within the root window to decide wether or not the floatbar needs to be shown or hidden. That value is checked outside of any Event handling as the main loop calls the xf_floatbar_hide_and_show function regulary. 
Alternativeley i could store the last x, y postion of the pointer outside of my xfFloatbar struct, but i did not want to mess around to much with the existing structures.

akallabeth: @mmattes Ah, ok, then that is a valid case.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@mmattes sorry, did take some time to check this.

mmattes:
@akallabeth cool, i will then create a new branch and PR as this branch is somehow messed up when i tried to squash the changes into one commit.

akallabeth:
@mmattes Looks like you have some clash with redefinitions in the windows client.
Might also be a good idea to refactor the commit history and clean that up before we merge.

akallabeth:
@mmattes Ok, forget the last comment, yours just showed up above mine (damn auto refresh, or lack thereof sometimes)

mmattes:
@akallabeth can you point out to me what the problem is in regards to the windows client? I just compiled it on a windows machine and it looks like everything works?

akallabeth:
Struct member floatbar redefinition see https://ci.freerdp.com/job/windows_msbuild_vs2010/1220/arch=win64,label=vs2010/console for the build log

mmattes:
thank you will have a look at it, couldnt find the log in jenkins as i am not used to it.

mmattes:
im closing this see #4865

PR #4810, opened 2018-08-21
No proxy support

akallabeth:


libfreerdp/core/proxy.c
```diff
@@ -75,14 +75,23 @@ void proxy_read_environment(rdpSettings* settings, char* envname);
 BOOL proxy_prepare(rdpSettings* settings, const char** lpPeerHostname, UINT16* lpPeerPort,
                    const char** lpProxyUsername, const char** lpProxyPassword)
 {
+	if (settings->ProxyType == PROXY_TYPE_IGNORE)
+		return FALSE;
+
 	/* For TSGateway, find the system HTTPS proxy automatically */
-	if (!settings->ProxyType)
+	if (settings->ProxyType == PROXY_TYPE_NONE)
```
mfleisz: Command line output says that env variable is now called http_proxy

PR #4812, opened 2018-08-21
Smartcard logon rdp 2

informatimago:
Rebased smartcard-logon-rdp.


akallabeth:
@freerdp-bot test

gofish543:
This pull request still contains the SCARD_E_SUFFICIENT_BUFFER exception. This can be fixed by extending the pbAtr from 32 to 34. Would you like me to make the change and request a PR into your branch?

gofish543:
The issue I raised a while back still exists within this pull request. A smartcard buffer overflow will occur if the buffer size for pbAtr is not increased from 32 to 34. Additionally if this is done, the smart card will not have any of it's contents forwarded and Windows will not recognize any valid certificates on the smart card.

Also, how does this actually work with freerdp? I can't seem to get it to work at all. What is the command that's suppose to be run? I read the documentation page, but it isn't making sense.

informatimago:
@gofish543 See: https://github.com/informatimago/FreeRDP/wiki/smartcard-logon

winpr/libwinpr/utils/cmdline.c
```diff
@@ -322,6 +322,16 @@ int CommandLineParseArgumentsA(int argc, LPSTR* argv, COMMAND_LINE_ARGUMENT_A* o
...
 						options[j].Flags |= COMMAND_LINE_VALUE_PRESENT;
 					}
+					else
+					{
+						/*
+						We need to reset the option Value and Flag
+						to process correctly the cases where the same
+						option is given several times like: /foo:bar /foo
+						*/
+						options[j].Value = NULL;
```
akallabeth: looks like a misuse of the command line, usually it should be `/option:foo,bar,bla`

akallabeth:
@informatimago I've implemented the changes you've done to `LinkedList` in #4814 in a FreeRDP 'compatible' way, could you rebase on that? (Assuming I didn't forget something you need?)
You can set your comparison functions with `LinkedList_Object(list)->fnObjectEquals = foo`
If you want the contents to be copied in the list, do `LinkedList_Object(list)->fnObjectNew = _strdup` and `LinkedList_Object(list)->fnObjectFree = free`

.gitignore
```diff
@@ -132,7 +133,7 @@ TAGS
 *.tar.gz
 
 # packaging related files
-!packaging/scripts/prepare_deb_freerdp-nightly.sh
```
akallabeth: please remove this change, changes to these scripts require changes in the build infrastructure.

akallabeth:
@informatimago The parameter parsing is still broken/different from the rest of options, the syntax is `/smartcard:dev1,dev2,...` and not multiple `/smartcard` options

informatimago:
@akallabeth This is false. For example, in TestCmdLine.c, we see the case:
```
		{"xfreerdp", "--plugin", "rdpsnd", "--plugin", "rdpdr", "--data", "disk:media:"DRIVE_REDIRECT_PATH, "--", "test.freerdp.com", 0},
```
If what you say: was true it would be "--plugin:rdpsnd,rdpdr"!
Another example is the /drive option, since the comma doesn't separate drives, but different kind of parameters. You have to give one drive option per drive.

The patch you quoted correct a bug in those case, if /plugin or /drive can meaningfully be given without argument (which is the case of /smartcard). So I will leave it.

However, I will put back parsing of /smartcard:str1,str2,...  but this will prevent currently the use of comma in the smartcard filtering. (We could add it back using an escape character).

akallabeth:
@informatimago this is true for the old style syntax that will be dropped.

PR #4813, opened 2018-08-21
Bunch of fixes based on covscan results

ondrejholy:
Here is a bunch of fixes based on covscan results. The most of them fix leaks in various corner-cases, but there are also some fixes for potential buffer overflow, usage of released memory...

The set of patches in this branch eliminates 131 defects marked as important from the [scan-results-imp.err](https://github.com/FreeRDP/FreeRDP/files/2306685/scan-results-imp.err.before.txt) generated for commit 735ab2e8b22840c33d5cab368ac168243e96f33e.

However, there are still some remaining issues, which I think are mostly false positives, see [scan-results-imp.err](https://github.com/FreeRDP/FreeRDP/files/2306753/scan-results-imp.err.after.txt) for this branch, but there are some which are worth to be investigated (i.e. the one with `USE_AFTER_FREE`) where I am not sure how to fix it...

### DIFF
--- a/client/X11/generate_argument_docbook.c
+++ b/client/X11/generate_argument_docbook.c
@@ -11,65 +11,83 @@ LPSTR tr_esc_str(LPCSTR arg, bool format)
 	LPSTR tmp = NULL;
 	size_t cs = 0, x, ds, len;
 	size_t s;
-	if(NULL == arg)
+
+	if (NULL == arg)
 		return NULL;
+
 	s = strlen(arg);
+
 	/* Find trailing whitespaces */
-	while((s > 0) && isspace(arg[s-1]))
+	while ((s > 0) && isspace(arg[s - 1]))
 		s--;
+
 	/* Prepare a initial buffer with the size of the result string. */
 	ds = s + 1;
-	if(s)
+
+	if (s)
 		tmp = (LPSTR)realloc(tmp, ds * sizeof(CHAR));
-	if(NULL == tmp)
+
+	if (NULL == tmp)
 	{
 		fprintf(stderr,  "Could not allocate string buffer.\n");
 		exit(-2);
 	}
+
 	/* Copy character for character and check, if it is necessary to escape. */
 	memset(tmp, 0, ds * sizeof(CHAR));
-	for(x=0; x<s; x++)
+
+	for (x = 0; x < s; x++)
 	{
-		switch(arg[x])
+		switch (arg[x])
 		{
 			case '<':
 				len = format ? 13 : 4;
 				ds += len - 1;
 				tmp = (LPSTR)realloc(tmp, ds * sizeof(CHAR));
-				if(NULL == tmp)
+
+				if (NULL == tmp)
 				{
 					fprintf(stderr,  "Could not reallocate string buffer.\n");
 					exit(-3);
 				}
+
 				if (format)
-					strncpy (&tmp[cs], "<replaceable>", len);
-				else				
-					strncpy (&tmp[cs], "&lt;", len);
+					strncpy(&tmp[cs], "<replaceable>", len);
+				else
+					strncpy(&tmp[cs], "&lt;", len);
+
 				cs += len;
 				break;
+
 			case '>':
 				len = format ? 14 : 4;
 				ds += len - 1;
 				tmp = (LPSTR)realloc(tmp, ds * sizeof(CHAR));
-				if(NULL == tmp)
+
+				if (NULL == tmp)
 				{
 					fprintf(stderr,  "Could not reallocate string buffer.\n");
 					exit(-4);
 				}
+
 				if (format)
-					strncpy (&tmp[cs], "</replaceable>", len);
-				else				
-					strncpy (&tmp[cs], "&lt;", len);
+					strncpy(&tmp[cs], "</replaceable>", len);
+				else
+					strncpy(&tmp[cs], "&lt;", len);
+
 				cs += len;
 				break;
+
 			case '\'':
 				ds += 5;
 				tmp = (LPSTR)realloc(tmp, ds * sizeof(CHAR));
-				if(NULL == tmp)
+
+				if (NULL == tmp)
 				{
 					fprintf(stderr,  "Could not reallocate string buffer.\n");
 					exit(-5);
 				}
+
 				tmp[cs++] = '&';
 				tmp[cs++] = 'a';
 				tmp[cs++] = 'p';
@@ -77,14 +95,17 @@ LPSTR tr_esc_str(LPCSTR arg, bool format)
 				tmp[cs++] = 's';
 				tmp[cs++] = ';';
 				break;
+
 			case '"':
 				ds += 5;
 				tmp = (LPSTR)realloc(tmp, ds * sizeof(CHAR));
-				if(NULL == tmp)
+
+				if (NULL == tmp)
 				{
 					fprintf(stderr,  "Could not reallocate string buffer.\n");
 					exit(-6);
 				}
+
 				tmp[cs++] = '&';
 				tmp[cs++] = 'q';
 				tmp[cs++] = 'u';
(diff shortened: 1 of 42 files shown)


akallabeth:
@ondrejholy Wow, that is a nice bunch of fixes, thank you.
Will take some time to review though.
@freerdp-bot test

client/X11/generate_argument_docbook.c
```diff
@@ -42,8 +42,10 @@ LPSTR tr_esc_str(LPCSTR arg, bool format)
 					exit(-3);
 				}
 				if (format)
+					// coverity[buffer_size]
```
akallabeth: Please no C++ comments

client/X11/generate_argument_docbook.c
```diff
@@ -42,8 +42,10 @@ LPSTR tr_esc_str(LPCSTR arg, bool format)
...
 					strncpy (&tmp[cs], "<replaceable>", len);
 				else				
+					// coverity[buffer_size]
```
akallabeth: Please no C++ comments

client/X11/generate_argument_docbook.c
```diff
@@ -57,8 +59,10 @@ LPSTR tr_esc_str(LPCSTR arg, bool format)
 					exit(-4);
 				}
 				if (format)
+					// coverity[buffer_size]
```
akallabeth: Please no C++ comments

client/X11/generate_argument_docbook.c
```diff
@@ -57,8 +59,10 @@ LPSTR tr_esc_str(LPCSTR arg, bool format)
...
 					strncpy (&tmp[cs], "</replaceable>", len);
 				else				
+					// coverity[buffer_size]
```
akallabeth: Please no C++ comments

libfreerdp/codec/planar.c
```diff
@@ -958,6 +958,7 @@ BYTE* freerdp_bitmap_planar_compress_plane_rle(const BYTE* inPlane,
 	UINT32 outBufferSize;
 	UINT32 nBytesWritten;
 	UINT32 nTotalBytesWritten;
+	BOOL allocate = !outPlane;
```
akallabeth: Why free that here? The buffers are allocated and kept in the planar context struct.

winpr/libwinpr/winsock/winsock.c
```diff
@@ -1191,7 +1191,7 @@ SOCKET _socket(int af, int type, int protocol)
 
 	fd = socket(af, type, protocol);
 
-	if (fd < 1)
+	if (fd < 0)
```
akallabeth: `stdin` is also invalid here, needs further check.

winpr/libwinpr/utils/ntlm.c
```diff
@@ -110,7 +113,11 @@ BYTE* NTOWFv2W(LPWSTR Password, UINT32 PasswordLength, LPWSTR User,
 
 	/* Compute the HMAC-MD5 hash of the above value using the NTLMv1 hash as the key, the result is the NTLMv2 hash */
 	if (!winpr_HMAC(WINPR_MD_MD5, NtHashV1, 16, buffer, UserLength + DomainLength, NtHash, WINPR_MD4_DIGEST_LENGTH))
+	{
+		if (allocate)
+			free(NtHash);
 		result = NULL;
+	}
```
akallabeth: If `NtHash` was allocated here we have to free it in any case at the end, correct?

libfreerdp/core/gateway/rpc.c
```diff
@@ -739,7 +739,7 @@ static int rpc_channel_tls_connect(RpcChannel* channel, int timeout)
 	sockfd = freerdp_tcp_connect(context, settings, peerHostname,
 	                             peerPort, timeout);
 
-	if (sockfd < 1)
+	if (sockfd < 0)
```
akallabeth: Needs a check if `stdin` is valid here.

libfreerdp/core/transport.c
```diff
@@ -419,7 +419,7 @@ BOOL transport_connect(rdpTransport* transport, const char* hostname,
 		else
 			sockfd = freerdp_tcp_connect(context, settings, hostname, port, timeout);
 
-		if (sockfd < 1)
```
akallabeth: Needs a check if stdin is valid here.

libfreerdp/core/gateway/rpc.c
```diff
@@ -739,7 +739,7 @@ static int rpc_channel_tls_connect(RpcChannel* channel, int timeout)
 	sockfd = freerdp_tcp_connect(context, settings, peerHostname,
 	                             peerPort, timeout);
 
-	if (sockfd < 1)
```
akallabeth: Needs a check if stdin is valid here

libfreerdp/core/gateway/rdg.c
```diff
@@ -680,7 +681,7 @@ static BOOL rdg_tls_connect(rdpRdg* rdg, rdpTls* tls, const char* peerAddress, i
 	                             peerAddress ? peerAddress : peerHostname,
 	                             peerPort, timeout);
 
-	if (sockfd < 1)
```
akallabeth: Needs a check if stdin is valid here.

akallabeth:
As for the `UwacSeatDestroy` issue, @hardening could you have a look? This seems wrong ;)

client/X11/generate_argument_docbook.c
ondrejholy: `/* coverity[buffer_size] */` will be ok?

libfreerdp/codec/planar.c
ondrejholy: The function seems to be a little bit unfortune. You can pass `outPlane`, which is modified and returned. But you can also pass `NULL` and `malloc` is used consequently (though not sure that this case is ever used). In case of failure, `NULL` is returned and pointer to newly allocated memory is lost, but it should be freed. Maybe the function should rather return `BOOL` and expect that `outPlane` is always non-`NULL`.

winpr/libwinpr/winsock/winsock.c
ondrejholy: `0` may be returned from `socket()` if `stdin` is closed and consequently the descriptor is leaked, but it is rather theoretical...

winpr/libwinpr/utils/ntlm.c
ondrejholy: This is more or less the same as `freerdp_bitmap_planar_compress_plane_rle`. `NtHash` may be passed in, or `NULL` is passed in. But it seems that there is a bug, because the newly allocated `NtHash` should be also returned...

client/X11/generate_argument_docbook.c
akallabeth: yes, please. (sorry, but old C compilers complain horribly about these)

libfreerdp/codec/planar.c
akallabeth: @ondrejholy you're right, the whole allocation stuff should go. (`freerdp_bitmap_planar_context_reset` allocates the planes)

ondrejholy:
> Please run ./scripts/format_code.sh (astyle) on changed files.

Hmm, the format script does a lot of non-related changes, wouldn't be better to push some mass change of format before you force people to use this script? 

I suppose that I should apply only the related to my changes and squash them into the corresponding commits, am I right?

winpr/libwinpr/utils/ntlm.c
ondrejholy: It seems that the same problems have also other functions in that file. May I change all the functions in that file to return `BOOL` to be obvious that you have to provide non-`NULL`?

akallabeth:
@ondrejholy yes, we should do the reformatting sometime before 2.0, for now we do it whenever we change a file. (that at least the space/tab stuff is correct)
I've checked, yes you have some files `winsock.c`, `generate_argument_docbook.c`, ... which were not properly formatted before.
Best reformat them first in their own commit, then apply your patches.

ondrejholy:
So I have force pushed the requested changes apart from the `sockfd < 1` related patches... what you suggest to do with them?

akallabeth:
@ondrejholy I would say lets use the fixed version (and if some issues occur fix that)
My gut tells me that if there is an issue, there is a bigger bug hidden anyway.

akallabeth:
@freerdp-bot test

winpr/libwinpr/utils/ntlm.c
```diff
@@ -77,52 +73,51 @@ BYTE* NTOWFv1A(LPSTR Password, UINT32 PasswordLength, BYTE* NtHash)
...
 	BYTE NtHashV1[16];
-	BYTE* result = NtHash;
+	BOOL result = FALSE;
 
-	if ((!User) || (!Password))
-		return NULL;
+	if ((!User) || (!Password) || (!NtHash))
+		return FALSE;
 
 	if (!NtHash && !(NtHash = (BYTE*) malloc(WINPR_MD4_DIGEST_LENGTH)))
```
ondrejholy: This line should be also dropped :-/ I scheduled covscan again to be sure that I do not introduce another defect and this has been found :-) I am going to update it...

uwac/libuwac/uwac-window.c
```diff
@@ -500,7 +501,7 @@ UwacReturnCode UwacDestroyWindow(UwacWindow** pwindow)
 		wl_region_destroy(w->opaque_region);
 
 	if (w->input_region)
-		wl_region_destroy(w->opaque_region);
+		wl_region_destroy(w->input_region);
```
hardening: good catch

ondrejholy:
Thanks for merging! Would you like me to file a new issue with the list of remaining defects reported by covscan? I am unsure about their solution and thus I am not going to propose any fixes for them. But as I mentioned before, I recommend you to look at the `USE_AFTER_FREE` defect at least...

akallabeth:
@ondrejholy Thanks for the fixes ;)
And yes, a new issue with the remaining report would be nice.

PR #4814, opened 2018-08-22
Linked list add object for user defined functions

akallabeth:
#4812 introduced the requirement for user defined comparison/new/delete callbacks for `LinkedList` values.
This pull implements that:
* breaks ABI of `wLinkedList` (now opaque) and `wLinkedListNode` (now internal)
* Adds `LinkedList_Object` to get a pointer to `wObject` to set the user defined callbacks
* Modifies `typedef void* (*OBJECT_NEW_FN)(void);` to `typedef void* (*OBJECT_NEW_FN)();` to allow new functions to have arguments

### DIFF
--- a/winpr/include/winpr/collections.h
+++ b/winpr/include/winpr/collections.h
@@ -35,7 +35,9 @@
 extern "C" {
 #endif
 
-typedef void* (*OBJECT_NEW_FN)(void);
+/* We don't know if the new function will require an argument.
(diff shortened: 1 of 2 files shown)


akallabeth:
@hardening The only breaking points are the now opaque type and removed internal structs.
They shouldn't have been used anyway (and the LinkedList was not in heavy use)

PR #4837, opened 2018-09-03
Smartcard logon rdp

informatimago:
Added /smartcard-logon:rdp option.
cf. https://github.com/informatimago/FreeRDP/wiki/smartcard-logon (it will be copied to FreeRDP/FreeRDP wiki once the merge is completed).

Basically; sets the autologin flag which along the smartcard redirection allows MS-Windows to use Smartcard Logon.

This patch only works with RDP negotiation.

Several substrings can be given to the /smartcard option to select the Smartcard readers that will be redirected to the remote workstation.  This allows to omit the Smartcard reader that may interfere with the Smartcard Logon process or that we don't want to give access to the remote workstation.


include/freerdp/settings.h
```diff
@@ -417,6 +418,12 @@ typedef struct rdp_monitor rdpMonitor;
 #define RDPDR_DTYP_FILESYSTEM		0x00000008
 #define RDPDR_DTYP_SMARTCARD		0x00000020
 
+/* TODO: OUTCH! cf. https://github.com/FreeRDP/FreeRDP/issues/4833
```
akallabeth: Does not belong to this pull, remove.
Pull requests for `2.0.0 + 1` are of course welcome.

include/freerdp/settings.h
```diff
@@ -1088,13 +1102,18 @@ struct rdp_settings
...
-	ALIGN64 UINT32 Password51Length; /* 1281 */
-	UINT64 padding1344[1344 - 1282]; /* 1282 */
+	ALIGN64 BYTE*  Password51;          /* 1280 */
+	ALIGN64 UINT32 Password51Length;    /* 1281 */
+	ALIGN64 BOOL   SmartcardLogon;      /* 1282 */
+	ALIGN64 char*  LogonReaderName;     /* 1283 */
+	ALIGN64 char*  Pin;                 /* 1284 */
+	ALIGN64 BOOL   PinLoginRequired;    /* 1285 */
+	ALIGN64 BOOL   PinPadIsPresent;     /* 1286 */
+	UINT64  padding1344[1344 - 1288];   /* 1288 */
```
akallabeth: Looks like a calculation error, all following fields are now at offset `+1`

channels/smartcard/client/smartcard_main.c
```diff
@@ -258,7 +315,8 @@ static void smartcard_release_all_contexts(SMARTCARD_DEVICE* smartcard)
 static UINT smartcard_free(DEVICE* device)
 {
 	UINT error;
-	SMARTCARD_DEVICE* smartcard = (SMARTCARD_DEVICE*) device;
+	CHECKED_SMARTCARD_DEVICE(smartcard, device);
```
akallabeth: Good idea, but maybe name it so it is obvious that this macro does a `return` in the error case?

informatimago:
@akallabeth 

-    Please run the formatting script on all changed files, there are multiple issues (and unrelated changes) with that. Maybe also check with a difftool against master.

I did before commit, it reformated some files.  It did again now, and it issued no output.
```
scripts/autoformat.sh
```
I don't find anything significant with a diff.

-     As already mentioned in the last pull, what is the reason for the smartcard filtering? Is this required for authentication or an extra feature? Just want to know what we're supposed to merge here.

This is explained in https://github.com/informatimago/FreeRDP/wiki/smartcard-logon

We have tokens that look like Smartcards, but they don't contain certificate allowing logon on MS-Windows.  If those tokens are redirected along with our smartcard reader, MS-Windows gets confused and cannot perform the Smartcard logon at all.

Another case would be https://github.com/FreeRDP/FreeRDP/issues/4743 where the reader or the smartcard returns a name longer than the 32 characters allowed by the protocol, which is thus unusable thru RDP.  We would want to avoid redirecting this reader to the remote MS-Windows workstation, but still redirect a reader that work. (assuming we have several readers connected, which is the case here).

So how else can I explain it, what is wrong with the explanations in https://github.com/informatimago/FreeRDP/wiki/smartcard-logon ?


-     Since we finally got rid of strcpy I'm against reintroducing it (I know it can be used safely, but it is a lot easier if it is not used at all)

Ok, I propose to use strcat as an alternative, but it is O(n^2) instead of O(n).

-     rdp_write_info_packet introduced new flags without conditions, sure this is intended and not something forgotten?

Do you want me to introduce corresponding command line flags to set them?

channels/smartcard/client/smartcard_operations.c
```diff
@@ -524,8 +531,12 @@ static LONG smartcard_ListReadersW_Call(SMARTCARD_DEVICE* smartcard, SMARTCARD_O
 	cchReaders = SCARD_AUTOALLOCATE;
 	status = ret.ReturnCode = SCardListReadersW(operation->hContext,
 	                          (LPCWSTR) call->mszGroups, (LPWSTR) &mszReaders, &cchReaders);
+	mszStringsLog("Readers BEFORE filtering: ", TRUE, mszReaders);
+	/* Remove the card readers that are not specified on the command line */
+	mszFilterStrings(TRUE, mszReaders, & cchReaders, smartcard->filter);
+	mszStringsLog("Readers AFTER  filtering: ", TRUE, mszReaders);
 	ret.msz = (BYTE*) mszReaders;
-	ret.cBytes = cchReaders * 2;
```
akallabeth: Sure that the amount of bytes here is not times `2`? `mszReaders` is of type `LPWSTR`

informatimago: Yes, sure. cchReader is already a number of bytes.  Also, the traces show the right values..

akallabeth: @informatimago Ok, could you tell the details of your setup? When I check `PCSC_SCardListReadersW` (which is called from `SCardListReadersW`) `cchReaders` is set to the number of characters just as https://docs.microsoft.com/en-us/windows/desktop/api/winscard/nf-winscard-scardlistreadersw tells too...

informatimago: ( As for spaces/tab, yes, mixups are white it's better to use 100% whitespaces, and also, programmers who use whitespaces to indent tend to earn more$ ;-) )

```
find . -name \*.[hc] -exec grep -nH -e '^ \| 	' {} +|sed -e 's/:.*//'|sort -u > ../a
git diff 25e2ab1..a393bc7 | grep 'diff --git'|sed -e 's-.* b/--'|sort -u > ../b
grep -nH -e '^ \| 	' $(grep -v -F -x -f ../a ../b)
```
The only matches are 
1- lines in comments, eg. " * foo";
2- parameter list continuations.  My editor is configured to indent them with only one tab, but autoformat.sh indents them with spaces aligned to the opening parenthesis.
3- rdpdr.h in which autoformat.sh indents the enums with 4 spaces. (I let my editor indent it with tabs, but autformat.sh reverted it to spaces).

informatimago: @akallabeth PCSC_SCardListReadersW is not the last function having modified cchReaders; it's mszFilterStrings.  This one is specified to return a number of bytes in all cases (cf. str.h).

akallabeth:
@informatimago Which version of `astyle` do you use? I can't reproduce the changes you commit with `astyle 3.1` on fedora nor with `astyle 2.06` on debian (and most of the files are still not properly formatted or contain changes I can't reproduce at all without manual intervention)

informatimago:
@akallabeth I have astyle 2.03 on GNU/Linux ubuntu 14.04   (and no, we cannot upgrade since we have to use pkcs#11 libraries that are not available on newer versions).  But I could compile and install a newer astyle.  In general, it's a lot of work since I will probably have to port also a bunch of newer versions of its dependencies...

akallabeth:
@informatimago Ok, then I'll do the reformatting, looks like your version does not support some of the switches. (Thought with running debain I'm the one with the oldest versions ^^)

akallabeth:
@informatimago Ok, nice try, but we'll decide when it is time to do the reformatting ;)

As you've already seen it is easier to get small changes approved fast, doing too much in one pull is error prone and hard to review.

Have taken the smartcard filtering idea from your pull and implemented that with #4840 and `strstr` instead of a custom comparison function.

akallabeth:
Ok, split your pull in #4840 for the filtering and #4842 for the logon flag.
I've stripped all changes that are future work (smartcard pin, ...).

Thanks for the effort!

PR #4842, opened 2018-09-04
Added /smartcard-logon option to set flag. (Stripped version of #4837…

akallabeth:
… by @informatimago)

This pull takes the idea of #4837 and only adds the `smartcard-logon` command line option without the filtering (split to #4840).
Stripped all unused additions (Smartcard variables currently unused)


libfreerdp/core/info.c
```diff
@@ -43,6 +43,67 @@ static const char* const INFO_TYPE_LOGON_STRINGS[4] =
...
+	for (i = 0; i < ARRAYSIZE(info_flags); i ++)
+	{
+		if (info_flags[i].flag & flags)
+		{
+			strcat(result, info_flags[i].label);
+			strcat(result, "|");
+		}
+	}
+
+	result[strlen(result) - 1] = '\0'; /* remove last "|" */
```
mfleisz: Potential OOB write if result is empty (no flags specified).

libfreerdp/core/info.c
```diff
@@ -43,6 +43,68 @@ static const char* const INFO_TYPE_LOGON_STRINGS[4] =
...
+
+	if (!result)
+		return 0;
+
+	for (i = 0; i < ARRAYSIZE(info_flags); i ++)
+	{
+		if (info_flags[i].flag & flags)
+		{
+			strcat(result, info_flags[i].label);
+			strcat(result, "|");
```
mfleisz: Potential OOB write if all flags are set (strcat adds a terminating \0 to the end of the string).

client/common/cmdline.c
```diff
@@ -1288,6 +1288,18 @@ static BOOL ends_with(const char* str, const char* ext)
 
 	return strncmp(&str[strLen - extLen], ext, extLen) == 0;
 }
+
+static void activate_smartcard_logon_rdp(rdpSettings* settings)
+{
+	settings->SmartcardLogon = TRUE;
+	settings->RdpSecurity = TRUE;
+	settings->TlsSecurity = FALSE;
```
hardening: Just curious, isn't smartcard authentication possible with TLS security ?

libfreerdp/core/settings.c
```diff
@@ -611,6 +611,7 @@ rdpSettings* freerdp_settings_new(DWORD flags)
 		goto out_fail;
 
 	settings->ActionScript = _strdup("~/.config/freerdp/action.sh");
+	settings->SmartcardLogon = FALSE;
 	return settings;
```
hardening: nitpicking but `settings` is calloced so SmartcardLogon is already `FALSE`

client/common/cmdline.c
akallabeth: @hardening Hmm, right. Should also be possible, will check.

Issue #4731, opened 2018-07-07
Length limit on title in wfreerdp?


jol64:
I am using wfreerdp.exe version 2.0.0-dev3 (b34e3bcf7) 64bit. I am using a commandline like the following:

> wfreerdp.exe /v:host.some.domain "/t:joachim8-sandisk-ultra-ii-240gb-cc-e8-a3-86.20180707-020455-897-success.vhdx" /u:Joachim /p:password /vmconnect:13FE501B-C517-437D-AAC6-ABE48CE5B780;EnhancedMode=1 /audio-mode:0 2>"C:\ProgramData\Traces\wfreerdp 20180707 204005.tracefile"

as you can see the title is pretty long (generated by one of my programs).

The title is displayed as follows: 
![image](https://user-images.githubusercontent.com/24666601/42413744-0d1b4306-8227-11e8-86e4-d7d897d9f6fd.png)
Looks like the tail of the title is somehow interpreted as another character set. If I truncate the title in the middle, everything looks nice, but information relevant to my users is lost. 
Maybe there is an upper limit to the title length?

hardening:
I gave a quick look and from what I've seen the problem does not seem to be in FreeRDP core, perhaps in the windows specific part...

mfleisz:
Currently the buffer used for the window title could only handle up to 63 characters. I increased the buffer size in #4848.

mfleisz:
Fixed with #4848

PR #4848, opened 2018-09-10
wfreerdp: Increase buffer size for window title (#4731)

mfleisz:


mfleisz:
@hardening You are right - the sprintfs were buggy as well (snprintf doesn't \0 terminate if the buffer is not big enough to hold the data). Should be fixed now.

PR #4850, opened 2018-09-11
Reject unknown connection type values for /network command line option.

r-barnett:
Don't silently ignore unknown values for /network:.

I'm new to GitHub and fairly new to git in general, so starting with something quite simple.  Let me know if I should be doing something differently.


akallabeth:
@freerdp-bot test

PR #4855, opened 2018-09-14
Map a particular disconnect situation triggered by a user logging off…

r-barnett:
This commit adds detection of a disconnect resulting from a user logging off and maps it to a user log off code.  This was found with Windows XP, so maybe there's little interest, but I'm not sure that it is limited to that.

Before the commit...
```
# freerdp /v:winxphost
...
<log off via start menu>
[09:17:53:773] [16826:16827] [INFO][com.freerdp.core] - ERRINFO_RPC_INITIATED_DISCONNECT (0x00000001):The disconnection was initiated by an administrative tool on the server in another session.
[09:17:53:776] [16826:16827] [ERROR][com.freerdp.core.rdp] - rdp_recv_tpkt_pdu: rdp_recv_deactivate_all() fail
[09:17:53:776] [16826:16827] [ERROR][com.freerdp.core.transport] - transport_check_fds: transport->ReceiveCallback() - -1
# echo $?
1
```
After the commit...
```
# freerdp /v:winxphost
...
<log off via start menu>
[09:27:14:477] [18140:18141] [INFO][com.freerdp.core] - ERRINFO_RPC_INITIATED_DISCONNECT (0x00000001):The disconnection was initiated by an administrative tool on the server in another session.
[09:27:14:478] [18140:18141] [INFO][com.freerdp.core.rdp] - Error info says user did not initiate but MCS ultimatum sa they did; treat this as a user logoff
[09:27:14:478] [18140:18141] [INFO][com.freerdp.core] - ERRINFO_LOGOFF_BY_USER (0x0000000C):The disconnection was initiated by the user logging off their session on the server.
[09:27:14:479] [18140:18141] [ERROR][com.freerdp.core.rdp] - rdp_recv_tpkt_pdu: rdp_recv_deactivate_all() fail
[09:27:14:479] [18140:18141] [ERROR][com.freerdp.core.transport] - transport_check_fds: transport->ReceiveCallback() - -1
# echo $?
12
```

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -71,6 +71,13 @@
 #include <freerdp/codec/color.h>
 #include <freerdp/codec/bitmap.h>
 
+/*
+ * X11/Xlib.h defines Status, but winpr/rpc.h, which is eventually included by
+ * mcs.h, declares a few functions that have a parameter named Status.
+ */
+#undef Status
+#include <libfreerdp/core/mcs.h>
+
 #include <freerdp/utils/signal.h>
 #include <freerdp/utils/passphrase.h>
 #include <freerdp/client/cliprdr.h>
@@ -100,6 +107,7 @@
 #include "xf_input.h"
 #include "xf_channels.h"
 #include "xfreerdp.h"
+
 
 #include <freerdp/log.h>
 #define TAG CLIENT_TAG("x11")
@@ -1631,7 +1639,17 @@ static DWORD WINAPI xf_client_thread(LPVOID param)
 	}
 
 	if (!exit_code)
+	{
 		exit_code = freerdp_error_info(instance);
+
+		if (exit_code == XF_EXIT_DISCONNECT && freerdp_get_dpu_reason(context) == MCS_Reason_user_requested)
+		{
+			/* This situation might be limited to Windows XP. */
+			WLog_INFO(TAG,
+			          "Error info says user did not initiate but MCS ultimatum says they did; treat this as a user logoff");
+			exit_code = XF_EXIT_LOGOFF;
+		}
+	}
 
 disconnect:
 
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -71,13 +71,6 @@
 #include <freerdp/codec/color.h>
 #include <freerdp/codec/bitmap.h>
 
-/*
- * X11/Xlib.h defines Status, but winpr/rpc.h, which is eventually included by
- * mcs.h, declares a few functions that have a parameter named Status.
- */
-#undef Status
-#include <libfreerdp/core/mcs.h>
-
 #include <freerdp/utils/signal.h>
 #include <freerdp/utils/passphrase.h>
 #include <freerdp/client/cliprdr.h>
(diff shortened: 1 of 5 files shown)


akallabeth:
@freerdp-bot test

libfreerdp/core/rdp.c
```diff
@@ -373,6 +373,14 @@ BOOL rdp_read_header(rdpRdp* rdp, wStream* s, UINT16* length, UINT16* channelId)
 			else
 				rdp_set_error_info(rdp, ERRINFO_RPC_INITIATED_DISCONNECT);
 		}
+		else if ((rdp->errorInfo == ERRINFO_RPC_INITIATED_DISCONNECT) && (reason == MCS_Reason_user_requested))
```
akallabeth: I'd map the `ERRINFO_RPC_INITIATED_DISCONNECT` code to a return code instead of setting a (unrelated) error info code.

r-barnett: This function (`rdp_read_header()`) has a return type of BOOL, so little information can be passed through that.  Also, the `rdp_set_error_info()` call is the same as that used a few lines above for `reason == NCS_Reason_user_requested`, so it seems at least somewhat consistent.

However, I suspect I'm missing your point.  Can you elaborate?

akallabeth: @r-barnett but the error code set will not be overridden. That can be checked and converted to a proper exit code.

r-barnett: @akallabeth Yes, that's what happens with `ERRINFO_LOGOFF_BY_USER`, but...you already understand that, so I'm still missing something.

Are you suggesting the use of a different error code or maybe the creation of a new error code?

akallabeth:
@r-barnett Might be a good idea to remember the `disconnect provider ulimatum reason` in the `rdpRdp` struct though to handle it at client exit.

r-barnett:
Maybe I got it now.
I should add a field to `struct rdp_rdp` (or better `struct rdp_context`?) to hold the reason.  Remove the new test I added to `rdp_read_header()`, but store the reason in the new field instead.  Then I should check the reason in the exit handling code in `xf_client_thread()` to determine the appropriate exit code.
Correct?

r-barnett:
@akallabeth is this what you had in mind?

client/X11/xf_client.c
```diff
@@ -71,6 +71,13 @@
 #include <freerdp/codec/color.h>
 #include <freerdp/codec/bitmap.h>
 
+/*
+ * X11/Xlib.h defines Status, but winpr/rpc.h, which is eventually included by
+ * mcs.h, declares a few functions that have a parameter named Status.
+ */
+#undef Status
+#include <libfreerdp/core/mcs.h>
```
akallabeth: Sorry, no includes from the `libfreerdp` folder. They are not part of the exposed API

include/freerdp/freerdp.h
```diff
@@ -360,6 +361,9 @@ FREERDP_API void freerdp_free(freerdp* instance);
 FREERDP_API BOOL freerdp_focus_required(freerdp* instance);
 FREERDP_API void freerdp_set_focus(freerdp* instance);
 
+FREERDP_API int freerdp_get_dpu_reason(rdpContext* context);
+FREERDP_API void freerdp_set_dpu_reason(rdpContext* context, int reason);
```
akallabeth: are you sure the setter needs to be exported? This is a protocol internal variable, so this should be hidden.

client/X11/xf_client.c
r-barnett: Okay.  This commit does add the reason to the exposed API though.  Does that mean that the `MCS_Reason` enum should be mvoed into one of the headers in include/freerdp?  Maybe constants.h where a few other constants are defined?  Or maybe freerdp.h where the `rdp_context` structure that contains the reason field is defined?

include/freerdp/freerdp.h
r-barnett: Good point.  I was just making it symmetric without thinking.  I can just remove the setter altogether.  There don't seem to be many others anyway, so that would be consistent.

client/X11/xf_client.c
akallabeth: @r-barnett good question, the name is bound to the `mcs` module. Might be a good idea to move and rename that to `freerdp.h`

include/freerdp/freerdp.h
akallabeth: @r-barnett maybe `freerdp_get_disconnect_ultimatum`? seems short enough and descriptive

r-barnett:
Getting closer perhaps.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #4867, opened 2018-09-19
Fixed smartcard filter for empty arguments.

akallabeth:


frispete:
This fixed the issue, as noted on the ML:

[FreeRDP-devel] Smartcard support in latest git is lost for Kobil token

Thank you, guys.

PR #2919, opened 2015-10-09
Sec/NLA: Support passwordless (blank password) login with NLA.

realjiangms:
It was supported in freerdp 1.0.2 but not supported in lastest master.
We should take empty password if it is explicitly specified with /p option.
If a password is not specified, we could first try SAM file. If the user entry does not exist, prompt for password.

Fixes #2783

PR #4870, opened 2018-09-19
[NLA] Password prompt is shown even if empty password is provided

corux:
FreeRDP shows the "Enter your credentials" prompt, even if an empty password was specified on the command line. It should accept the empty password and not show the prompt.

The pull request #2919 had already fixed this issue, but it got changed again in b81f168f0e26241907035eb942204c7fbfd01a77.

For testing this change: run FreeRDP with the option `/p:""`. No prompt should be shown in that case.


akallabeth:
@freerdp-bot test

PR #4874, opened 2018-09-21
MonitorLocalShiftX,Y should be set when 1st monitor from MonitorIds s…

r-barnett:
…elected as primary
The MonitorLocalShiftX,Y fields do not currently get set when the primary monitor is set to the first one from the /monitors: command line argument.  (There might be some other cases missed as well, but I do not understand all of the possibilities and objectives here well enough to deal with them all.)

### DIFF
--- a/client/X11/xf_monitor.c
+++ b/client/X11/xf_monitor.c
@@ -340,6 +340,8 @@ BOOL xf_detect_monitors(xfContext* xfc, UINT32* pMaxWidth, UINT32* pMaxHeight)
 		if (i == settings->MonitorIds[0])
 		{
 			settings->MonitorDefArray[nmonitors].is_primary = TRUE;
+			settings->MonitorLocalShiftX = settings->MonitorDefArray[nmonitors].x;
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@r-barnett Could you add a short description on what to expect with this change and how to test? Multimonitor support is a quite delicate piece.

r-barnett:
I found this problem when I tried to move a patch we have to get MMR to work from an older snapshot to RC3.  The MMR window was offset from the Windows Media Player window because the `MonitorLocalShiftX,Y` variables were not being set when we used /monitors to specify the list of monitors to use.

Besides our patch, I have not found any visible effects of not setting `MonitorLocalShiftX,Y`.  Can someone explain the comment at around line 470: "Subtract monitor shift from monitor variables for server-side use.  We maintain monitor shift value as Window requires the primary monitor to have a coordinate of 0,0"?  Is there some particular version of Windows that requires this?  What are the consequences of not doing that?  Is there some other code that is compensating for the loop not doing what the comment says it does?

Issue #4876, opened 2018-09-23
Copy to clipboard in remote windows 10 disconnects freerdp


borisneubert:
- Local host: Ubuntu 18.04 Bionic Beaver with X (not Wayland)
- Remote host: Windows 10 Professional
- FreeRDP: nightly, version 2.0.0-dev4 (3c158299a)

A remote desktop session to a windows 10 host crashes from time to time when copying something to the clipboard on the remote host.

To reproduce: start a RDP session to windows 10 host, fire up Excel, and start copying around cells or formats with Ctrl+C and Ctrl+V.

Command line: 

`/opt/freerdp-nightly/bin/xfreerdp /v:$vbox  /size:$size /bpp:24 /kbd:US /t:vbox02 /usb:auto  /sound +clipboard +fonts +aero /u:$user /p:$password`

Happens with and without the `+clipboard` parameter.

Build config:
```
This is FreeRDP version 2.0.0-dev4 (3c158299a)
[...]

PR #4879, opened 2018-09-24
Fixes for clipboard #4876

akallabeth:


libfreerdp/locale/keyboard_layout.c
```diff
@@ -219,99 +219,96 @@ static const RDP_KEYBOARD_IME RDP_KEYBOARD_IME_TABLE[] =
...
-			}
+			if (!layouts[num].name)
+				goto fail;
 		}
 	}
 
 	ZeroMemory(&layouts[num], sizeof(RDP_KEYBOARD_LAYOUT));
-
-	return layouts;
+	return &layouts[1];
```
mfleisz: Why do we return ```&layouts[1]``` here? I think this will cause trouble when we call free() on this address ...

akallabeth: @mfleisz Good you spottet that, missed to clean that one before commit.

PR #4885, opened 2018-09-24
Fixed #3423: Process xevents when in reconnect mode.

### LINKED ISSUE #3423: Host system freezes when target network connectivity is lost.
Hello,

Just wanted to report my findings.  I am running Ubuntu 16.04 x64, and RDP to a Windows 7 desktop for reference.

Whenever there is a loss of connectivity to the target machine, the host machine completely freezes up, forcing me to have to hard shutdown my computer.  It seems there is no safe-disconnect or cancel upon a certain number of reconnect tries on accidental network connectivity loss?

Problem can be easily reproduced by two methods:
1) On the target Windows device i made a batch script that runs ipconfig /release and ipconfig /renew (to ensure I get connectivity back).  Connect to the target via RDP and run the script.  This will cause the aforementioned problem.  I found this to be reproducible on accident.

2) I differentiate between two different IP networks, and sometimes have to disable one network to ensure proper network connection to various assets.  Simply disconnecting from the network where I am connected via RDP to a target machine will cause the aforementioned problem.  This one is avoidable but its a habit to just switch network connections before closing the RDP session.

Hope this helps.

Jordan

akallabeth:
* Adds a callback function to process local window events to reconnect function.
* Fixed crashes due to unloaded disp channel in reconnect mode.

To test:
1. Connect with `+auto-reconnect`
1. Disable network connection
1. Wait 15 seconds until reconnect triggers
1. Try to resize/fullscreen/close the local window

Old behaviour:
The window does not process any events and is frozen

### DIFF
--- a/client/X11/xf_disp.c
+++ b/client/X11/xf_disp.c
@@ -58,12 +58,7 @@ static UINT xf_disp_sendLayout(DispClientContext* disp, rdpMonitor* monitors, in
 
 static BOOL xf_disp_settings_changed(xfDispContext* xfDisp)
 {
-	rdpSettings* settings;
-
-	if (!xfDisp || !xfDisp->xfc || !xfDisp->xfc->context.settings)
-		return FALSE;
-
-	settings = xfDisp->xfc->context.settings;
+	rdpSettings* settings = xfDisp->xfc->context.settings;
 
 	if (xfDisp->lastSentWidth != xfDisp->targetWidth)
 		return TRUE;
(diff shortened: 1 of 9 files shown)


client/X11/xf_disp.c
```diff
@@ -341,11 +383,16 @@ UINT xf_DisplayControlCaps(DispClientContext* disp, UINT32 maxNumMonitors,
...
+BOOL xf_disp_init(xfDispContext* xfDisp, DispClientContext* disp)
 {
-	rdpSettings* settings = xfc->context.settings;
-	xfc->disp = disp;
-	disp->custom = (void*) xfc->xfDisp;
+	rdpSettings* settings;
+
+	if (!xfDisp || !xfDisp->xfc || !disp)
+		return FALSE;
+
```
hardening: I don't understand the rational for these checks, I can't see any situation that would lead any of these to be NULL.

akallabeth:
@hardening which are?

client/X11/xf_disp.c
bmiklautz: @hardening I agree with some of these checks but especially in this function which is called from a handler and therefore possibly externally invoked. It makes sense imho.

akallabeth:
@hardening Reworked the checks in `xf_disp.c`, removed the ones already checked and unified the ones for the function pointers.

PR #4904, opened 2018-10-01
Fixed floatbar X11 leak.

akallabeth:


client/X11/xf_floatbar.c
```diff
@@ -204,6 +204,9 @@ xfFloatbar* xf_floatbar_new(xfContext* xfc, Window window)
 	xfFloatbar* floatbar;
 	XWindowAttributes attr;
 	int i, width;
+	if (!xfc)
+		return NULL;
+
```
hardening: `xfc` can't be NULL here

PR #4907, opened 2018-10-02
Transport write leak fix

akallabeth:
* Handling stream allocation and free for buffers used for `transport_write` was quite error prone, this pull tries to address this.
* `StreamPool_Take` did not reset used stream length to current capacity letting `Stream_SafeSeek` and other operations dependent on length fail.


akallabeth:
@hardening @bmiklautz @mfleisz @nfedera Anyone time for review?

PR #4911, opened 2018-10-04
Move from libavresample to libswresample

vp1981:
- The libavresample as part of FFmpeg was declared as deprecated in
    favor of libswresample on Dec 25 2017 in commit https://github.com/FFmpeg/FFmpeg/commit/c29038f3041a4080342b2e333c1967d136749c0f  
```
lavr: deprecate the entire library
    
    As major Linux distributions (RPM, DEB and Archlinux) provide now
    FFmpeg and libswresample as its part use the libswresample instead of
    libavresample.

Signed-off-by: Vladimir Lomov <vladimir@smoon7.vl-lomov.ru>
```


hardening:
I'm afraid that will need both support for `libavresample` and `libswresample`, any opinions ?

akallabeth:
@hardening yes, we'll need both
@vp1981 thank you for this patch, will integrate it in a new pull

PR #4913, opened 2018-10-04
Fixed audio recording with opensles.

akallabeth:
* Simplify recording (eliminate unnecessary thread)
* Eliminate buffer copy
* Utilize a fixed 2 buffer setup to remove unnecessary allocations


akallabeth:
@freerdp-bot test

PR #4915, opened 2018-10-04
Add cmake option to use standard OpenSSL libraries on Android

awakecoding:
This small PR makes it possible to override the default and not use "freerdp-openssl" and instead rely on a standard or vendor-agnostic OpenSSL library. In my case, I provide my own OpenSSL build that is shared between multiple internal projects, so I not want to use a special FreeRDP-only OpenSSL build.


cmake/FindOpenSSL.cmake
```diff
@@ -71,7 +71,7 @@ IF(WIN32)
   endif()
 ENDIF(WIN32)
 
-IF(ANDROID)
+IF(ANDROID AND WITH_FREERDP_OPENSSL)
```
akallabeth: This seems broken as now in the android case no library detection is done, if not also `WITH_FREERDP_OPENSSL` is set.
Would suggest to do this differently by providing a default name `ANDROID_SSL_LIBRARY_NAME` that can be overridden from external sources.

akallabeth:
Implemented with release `2.1.0`

PR #4924, opened 2018-10-12
winpr: add some functions to use wStream in a static way

hardening:
It's sometime useful to create a stream aliasing a buffer on the stack, and it's nice if we don't need some extra malloc for this.

Example use:
```C
   BYTE buffer[20];
   wStream s;

   Stream_StaticInit(&s, buffer, sizeof(buffer));
   Stream_Write_UINT16(&s, 0xff01);
   Stream_StaticFree(&s, FALSE);
```

### DIFF
--- a/winpr/include/winpr/stream.h
+++ b/winpr/include/winpr/stream.h
@@ -52,6 +52,9 @@ WINPR_API BOOL Stream_EnsureRemainingCapacity(wStream* s, size_t size);
 WINPR_API wStream* Stream_New(BYTE* buffer, size_t size);
 WINPR_API void Stream_Free(wStream* s, BOOL bFreeBuffer);
 
+WINPR_API BOOL Stream_StaticInit(wStream *s, BYTE* buffer, size_t size);
+WINPR_API void Stream_StaticFree(wStream *s, BOOL bFreeBuffer);
+
 static INLINE void Stream_Seek(wStream* s, size_t _offset)
 {
 	s->pointer += (_offset);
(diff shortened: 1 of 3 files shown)


hardening:
This would help in many situations like for most packets where we have packets structured like `<type, packet len, data>`, that allows to easily (no check needed) build a sub `wStream` that covers `data` and is passed to the function treating the sub packet. 
Sample code (`s` is the main `wStream`):
```c
void treat_packet(wStream *s) {
...
}

...
wStream substream;

Stream_Read_UINT8(s, packetLen);
Stream_StaticInit(&substream, Stream_Pointer(s), packetLen);

treat_packet(&substream); /* do the treatment using the packet */

Stream_Seek(s, packetLen);
```

That avoids playing with the position pointer of the main stream which is quite error prone, not even talking of tricky computations that go with it (hard to read and understand).

hardening:
@akallabeth what about if `Stream_StaticInit` requires a valid `buffer` (and so doesn't have to do any allocation), so that we don't have to worry about freeing (and so I can drop the according free function) ?

akallabeth:
@hardening yes, that sounds good. Like `wStream s = Stream_WrapBuffer(buffer, size);` and be done with it.

akallabeth:
@hardening Although there needs to be taken care of Stream functions trying to resize the buffer.

hardening:
@akallabeth you're right about resizing. Anyway I'm bothered with changing the wStream struct (adding a isStatic bool at the end of the struct ?). Closing the PR as I think it needs more discussion.

Issue #4925, opened 2018-10-14
color.c : freerdp_image_copy : unchecked memory bounds leads to crash


openconcerto:
If a server send wrong or out of client screen coordinates command parameters (which can be valid),
xfreerdp is seriously "garbarging" memory.

This issue causes serious issues of compatibility and can lead to exploit from a malicious server.

The problem is present in the master branch.
For example, the parameters in  freerdp_image_copy (codec/color.c) and other free_rdp functions MUST check these parameters :
nDstStep, nXDst, nYDst,
nWidth, nHeight,
nSrcStep, nXSrc, nYSrc

(It took me some hours to find the cause of random crashes of xfreerdp and Remmina)

Please note that in the RDP GDI Specification ([MS-RDPEGDI]), coordinates MUST be considered as signed values (see 2.2.2.2.1.1.1.1), but your implementation is "converting" the INT32s from the order structure to UINT32 parameters in free_rdp* functions.

akallabeth:
@openconcerto thank you for the report.
Do you have any further details you want/can share?
Will add some checks in the decoder.

openconcerto:
@akallabeth I think you could add some tests on libfreerdp/gdi/test/
 and libfreerdp/codec/test to check negative parameters 
and parameters greater than the screen/bitmap width or height.
I saw a lot of situations where the buffer manipulations can be "out of bound".

I saw gdi_get_bitmap_pointer calls (gdi.h)  using int32 instead of uint32,
in the case of gdi_LineTo, it seems safe from my point of view, but it bypass negatif checks or truncations.
I have to admit, that's it's not clear what a client should do for a line (-10,-10) to (20,20),
should it discard it or draw (0,0) to (20,20). But even if it will not crash the client, it should be investigated.

akallabeth:
@openconcerto I've added checks in the lower layer for all primary orders in #4926 so the coordinates from the wire will be valid or the session is aborted if the server sends garbage.

akallabeth:
@openconcerto The coordinates should be converted from relative format to absolute format correctly.
Would be interesting to see which orders have such invalid offsets and why they use them.
Do you have any logs from that? (that is, `/log-level:debug` or similar)

openconcerto:
Here is the last lines of the client log
 
[04:33:25:672] [21269:21270] [DEBUG][com.freerdp.core.update] - Switch Surface Alternate Secondary Drawing Order (0x00)
[...]

PR #4926, opened 2018-10-15
Order data validation and correction

akallabeth:
Related to fixing #4925 

* Adds checks to only allow orders that have been activated by client and server during capability exchange
* Various crash fixes found during fuzzing run
* Coordinate adjustments for orders, cut invalid ones before passing them on
* Split order handling in `support check`, `decoding` and `processing`

### DIFF
--- a/libfreerdp/core/orders.c
+++ b/libfreerdp/core/orders.c
@@ -212,100 +212,89 @@ static const char* altsec_order_string(BYTE orderType)
 	return buffer;
 }
 
-static BOOL freerdp_primary_check_bound(rdpContext* context, INT32 left, INT32 top, INT32 right,
-                                        INT32 bottom)
-{
-	if (!context || !context->settings)
-		goto fail;
-
-	if (left < 0)
-		goto fail;
-
-	if (top < 0)
-		goto fail;
-
-	if (right < 0)
-		goto fail;
-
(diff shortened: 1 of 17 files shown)


openconcerto:
This patch fixes memory corruption, but corrupts the display...

akallabeth:
@freerdp-bot test, come on!

libfreerdp/core/orders.c
```diff
@@ -3020,320 +3629,404 @@ static BOOL update_recv_primary_order(rdpUpdate* update, wStream* s, BYTE flags)
...
+	}
+
 	if (flags & ORDER_BOUNDS)
 	{
-		IFCALL(update->SetBounds, context, NULL);
+		rc = IFCALLRESULT(FALSE, update->SetBounds, context, NULL);
 	}
 
-	return TRUE;
+	return rc;
```
MartinHaimberger: cant be false ... but ok

akallabeth:
@MartinHaimberger thx, removed the log messages and simplified the check.
Reading but not processing an order does not really make sense, so removed that option.

MartinHaimberger:
@freerdp-bot test

openconcerto:
Remmina is no more compatible with xrdp
BitBlt is broken.
Good job.

PR #4932, opened 2018-10-16
winpr: add some functions to use wStream in a static way

hardening:
It's sometime useful to create a stream aliasing a buffer on the stack, and
it's nice if we don't need some extra malloc for this.

Example use:
```c
   BYTE buffer[20];
   wStream s;

   Stream_WrapBuffer(&s, buffer, sizeof(buffer));
   Stream_Write_UINT16(&s, 0xff01);
   Stream_Free(&s, FALSE);
```

### DIFF
--- a/winpr/libwinpr/utils/stream.c
+++ b/winpr/libwinpr/utils/stream.c
@@ -22,6 +22,7 @@
 #include "config.h"
 #endif
 
+#include <assert.h>
 #include <winpr/crt.h>
 #include <winpr/stream.h>
 
@@ -46,7 +47,17 @@ BOOL Stream_EnsureCapacity(wStream* s, size_t size)
 
 		position = Stream_GetPosition(s);
 
-		new_buf = (BYTE*) realloc(s->buffer, new_capacity);
+		if (!s->isOwner)
+		{
+			new_buf = (BYTE *)malloc(new_capacity);
+			CopyMemory(new_buf, s->buffer, s->capacity);
+			s->isOwner = TRUE;
+		}
+		else
+		{
+			new_buf = (BYTE*) realloc(s->buffer, new_capacity);
+		}
+
 		if (!new_buf)
 			return FALSE;
 		s->buffer = new_buf;
@@ -74,10 +85,8 @@ wStream* Stream_New(BYTE* buffer, size_t size)
 		return NULL;
 
 	s = malloc(sizeof(wStream));
-
 	if (!s)
 		return NULL;
-
 
 	if (buffer)
 		s->buffer = buffer;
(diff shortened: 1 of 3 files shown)


akallabeth:
@hardening Just one thought about allocating a `wStream` struct on stack:
This will prevent us from making `wStream*` opaque.
Having a `wStream* Stream_Wrap(BYTE* buffer, size_t size)` allocating/taking from an internal pool would still allow that.

hardening:
Well making wStream opaque would prevent fron inlining functions, so not a so good idea. And allocating from an internal pool still need some checks to see if it succeeded.

hardening:
Lots of stupid typos in this one !

hardening:
This time it's not me, please @freerdp-bot  can you test properly on mac ?

winpr/libwinpr/utils/stream.c
```diff
@@ -96,13 +98,23 @@ wStream* Stream_New(BYTE* buffer, size_t size)
...
 	s->count = 0;
-
+	s->isStatic = FALSE;
 	return s;
 }
 
+void Stream_WrapBuffer(wStream *s, BYTE *buffer, size_t size)
+{
+	assert(buffer);
+	s->buffer = s->pointer = buffer;
```
akallabeth: Might be a good idea to free previously allocated buffers if the `wStream*` was created by `Stream_New`

winpr/libwinpr/utils/stream.c
```diff
@@ -96,13 +98,23 @@ wStream* Stream_New(BYTE* buffer, size_t size)
 
 	s->pool = NULL;
 	s->count = 0;
-
+	s->isStatic = FALSE;
```
akallabeth: Might be a good idea to distinguish between static buffer and static `wStream`

winpr/libwinpr/utils/stream.c
hardening: My intent was that you either use `Stream_WrapBuffer` or `Stream_New`, not recycling a `wStream` retrieved from `Stream_New`. There's `Stream_SetBuffer` if you want to change where a dynamically allocated stream points.

hardening: perhaps that the function name should be more explicit on the fact that we target a static stream, like `Stream_StaticInit` like it was before.

winpr/libwinpr/utils/stream.c
hardening: @awakecoding proposed some modifications from wayk now where there is a `isOwner` flag, that allows to resize statically allocated streams by doing a copy of the buffer on the heap.

hardening:
Don't hesitate to merge that nice patch !

PR #4938, opened 2018-10-18
Warning fixes and TSG abort condition fix

akallabeth:
Fixes warnings occuring on current master.

### DIFF
--- a/libfreerdp/core/gateway/http.c
+++ b/libfreerdp/core/gateway/http.c
@@ -84,7 +84,7 @@ struct _http_response
 	wStream* data;
 };
 
-static char* string_strnstr(const char* str1, const char* str2, size_t slen)
+static char* string_strnstr(char* str1, const char* str2, size_t slen)
 {
 	char c, sc;
 	size_t len;
@@ -110,10 +110,10 @@ static char* string_strnstr(const char* str1, const char* str2, size_t slen)
 		str1--;
 	}
 
-	return ((char*) str1);
+	return str1;
 }
 
-static BOOL strings_equals_nocase(void* obj1, void* obj2)
+static BOOL strings_equals_nocase(const void* obj1, const void* obj2)
 {
 	if (!obj1 || !obj2)
 		return FALSE;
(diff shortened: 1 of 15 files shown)


libfreerdp/core/gateway/http.c
```diff
@@ -362,7 +362,7 @@ static BOOL http_encode_print(wStream* s, const char* fmt, ...)
 {
 	char* str;
 	va_list ap;
-	size_t length, used;
+	int length, used;
```
hardening: I don't understand what is done here, you change these variables to `int` and then they are cast to `size_t` in the rest of the function. What was wrong before ?

akallabeth: @hardening The return value of `vsnprintf` is int, but other functions require `size_t` as argument.

hardening: Then perhaps just casting the calls to `vsnprintf` would introduce less changes and would improve readability ?

akallabeth: it is a return value of `vsnprintf`...

libfreerdp/core/gateway/tsg.c
```diff
@@ -670,14 +676,7 @@ static BOOL TsProxyCreateTunnelReadResponse(rdpTsg* tsg, RPC_PDU* pdu,
...
 			case TSG_ASYNC_MESSAGE_REAUTH:
 				{
-					size_t loffset = Stream_GetPosition(pdu->s);
-					UINT32 offset = (UINT32)loffset;
-
-					if (loffset > UINT32_MAX)
-						goto fail;
-
-					if (!Stream_SafeSeek(pdu->s, rpc_offset_align(&offset, 8) ||
-					                     (Stream_GetRemainingLength(pdu->s) < 8)))
```
nfedera: - Don't know the internals but the new tsg_stream_align always uses `4` while we had `8`.
- Not sure if the removed checks were important

akallabeth:
@freerdp-bot test

PR #4945, opened 2018-10-18
transferEncoding can be NULL, correct if happens in line 348

mmattes:
I am trying to use freerdp to connect to a remote app over a remote gateway which doesn't work yet. While debugging i found a false check on transferEncoding being NULL in the method rdg_build_http_request. The right IF happens a bit further in the method in line 348 and is working fine.


akallabeth:
@freerdp-bot test

PR #4946, opened 2018-10-18
added missing check for remote app where no floatbar exists

mmattes:
Found a missing check for a floatbar method which caused the method xf_floatbar_hide_and_show to be called although there was no floatbar created as the context was running in remote app mode. This happens when xfreerdp is started with a rdp file.


akallabeth:
@freerdp-bot test

PR #4947, opened 2018-10-18
Fixed http gateway body length read.

akallabeth:
@mmattes looks like I've found it, `bodyLength` is usually `0` in the responses, then use the length actually read.

### DIFF
--- a/libfreerdp/core/gateway/http.c
+++ b/libfreerdp/core/gateway/http.c
@@ -742,10 +742,12 @@ HttpResponse* http_response_recv(rdpTls* tls)
 
 	response->ContentLength = 0;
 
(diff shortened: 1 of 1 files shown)


mmattes:
@akallabeth for what was this? The request I made in the irc channel? If so then I unfortunately got disconnected and could not read any reply. I changed this local but it did not have any effect to my problem.

akallabeth:
@mmattes That was an issue I had testing with `http` gateway

mmattes:
@akallabeth unfortunately this does not work for me, i will have another look may i can find something else what causes the disconnect

mmattes:
@akallabeth cool, seems all the changes you did also solved my problems :+1:

PR #4955, opened 2018-10-19
Fix static drives 'a:, c:' not redirected

jlafontaine-devolutions:
Allow FreeRDP to correctly redirect static drives like 'A:' and 'C:'.

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -115,6 +115,10 @@ static UINT rdpdr_send_device_list_remove_request(rdpdrPlugin* rdpdr,
 
 #ifdef _UWP
 
+void first_hotplug(rdpdrPlugin* rdpdr)
+{
+}
+
 static DWORD WINAPI drive_hotplug_thread_func(LPVOID arg)
 {
 	return CHANNEL_RC_OK;
@@ -135,6 +139,10 @@ BOOL check_path(char* path)
 		return FALSE;
 
 	return GetVolumeInformationA(path, NULL, 0, NULL, NULL, NULL, NULL, 0);
+}
+
+void first_hotplug(rdpdrPlugin* rdpdr)
+{
 }
 
 LRESULT CALLBACK hotplug_proc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lParam)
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test

akallabeth:
@freerdp-bot test

jlafontaine-devolutions:
I tested on all platforms and the expected drives were redirected. 
Windows now redirects my C drive and has the same redirected drives as mstsc.exe.

akallabeth:
Will recheck and post the results tomorrow. Github does not like us today :/

akallabeth:
@jlafontaine-devolutions I've checked with `+drives` (fedora 28):
1. Hotplug events are handled, ok
1. Initial population of mount points is not working, so only drives attached after the client is started are redirected/removed

jlafontaine-devolutions:
@akallabeth 
I've double checked +drives on Ubuntu 18.04.1.
1. They are.
2. You are right, on linux, hotplugable drives don't show up upon connection. They do on Windows.
I will fix that.

jlafontaine-devolutions:
Initial population is now working with the latest commit.

akallabeth:
@freerdp-bot test

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -1089,8 +1061,6 @@ static UINT rdpdr_process_connect(rdpdrPlugin* rdpdr)
 					WLog_ERR(TAG, "CreateThread failed!");
 					return ERROR_INTERNAL_ERROR;
 				}
-
-				continue;
```
akallabeth: Why remove this? Adding a pseudo device for a hotplug wildcard does not really make sense or does it?

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -1079,7 +1051,7 @@ static UINT rdpdr_process_connect(rdpdrPlugin* rdpdr)
 		{
 			RDPDR_DRIVE* drive = (RDPDR_DRIVE*)device;
 
-			if (drive->Path && (strcmp(drive->Path, "*") == 0))
+			if (!rdpdr->hotplugThread && drive->Path && (strcmp(drive->Path, "*") == 0))
```
akallabeth: Might be good, if you allow more than one wildcard.
Still, adding wildcard devices (see below) does not seem to make sense.

channels/rdpdr/client/rdpdr_main.c
jlafontaine-devolutions: The first_hotplug and hotplug thread on Windows exclude fixed drives but adding the speudo device redirects all my drives correctly.
Would it be better if check_path allowed more drive types?

channels/rdpdr/client/rdpdr_main.c
jlafontaine-devolutions: That was some defensive programming. 
If only one wild card could ever happen, I can remove the check.

channels/rdpdr/client/rdpdr_main.c
akallabeth: yes, might be the most simple solution. If that works for you too.

jlafontaine-devolutions:
Added drive_fixed in the list of valid hot plug devices.

jlafontaine-devolutions:
Also pushed a similar fix for MacOS where Macintosh HD is a symlink to / and wasn't redirected.

PR #4961, opened 2018-10-24
Fixed #2039: Check for overflow in calculations.

### LINKED ISSUE #2039: Integer overflow in _aligned_offset_malloc
In winpr/libwinpr/crt/alignment.c::_aligned_offset_malloc(), the size computation passed to `malloc` can overflow, causing `malloc` to allocate a buffer which is too small:

```
 82         /* minimum alignment is pointer size */
 83         if (alignment < sizeof(void*))
 84                 alignment = sizeof(void*);
 85 
 86         /* malloc size + alignment to make sure we can align afterwards */
 87         base = malloc(size + alignment + sizeof(WINPR_ALIGNED_MEM));
```

In the same function, the comment does not match the following check.

```
 74         /* alignment must be a power of 2 */
 75         if (alignment % 2 == 1)
 76                 return NULL;
```

akallabeth:


winpr/libwinpr/crt/alignment.c
```diff
@@ -86,13 +89,36 @@ void* _aligned_offset_malloc(size_t size, size_t alignment, size_t offset)
...
+		return NULL;
+
+	alignmem = alignment + sizeof(WINPR_ALIGNED_MEM);
+
+	if (size > SIZE_MAX - alignmem)
+		return NULL;
+
+	alignsize = size + alignmem;
+
+	if (offset > SIZE_MAX - alignmem)
```
mfleisz: I think this should be ```SIZE_MAX - alignsize```

bmiklautz:
While basically functional correct the changes, especially the new variable alignmem (vs. parameter alignment) makes it much harder to read and to follow.

Big +1 for the checks
-1 for the new variables

akallabeth:
@mfleisz could you recheck? The new code should be more readable.

PR #4968, opened 2018-10-25
Simplified timezone update handling:

### LINKED ISSUE #4965: TimeZone generator script creates a file with >3000 compiler warnings
The generated file `timezone.c` is missing initializer values for `SYSTEMTIME` structs.
See currently committed file for a proper initialization.

NOTE: All warnings are visible when using `-Wextra`

akallabeth:
* Fix #4965: Adjusted the timezone update script to initialize
  structs properly.
* Updated the scripts to generate the files directly
* Added a small HOWTO
* Split zone information to separate files to allow automatic
  update without further manual editing.


akallabeth:
@bmiklautz added a note to the `README.timezones`.
As this is not run during compile time (only manually on a windows machine) I don't think that is a big issue.

PR #4970, opened 2018-10-25
uwac: change seat API to not reference disposed UwacSeat

hardening:
This is a false positive returned by covscan, but that doesn't hurt to fix it.


ondrejholy:
The defect is still reported by covscan and also new warning appeared: `FreeRDP-76c91ee51ae456b2238a3448761b0b336a46d25f/uwac/libuwac/uwac-display.c:269:4: warning: Value stored to 'seat' is never read`. I'm convinced that you have to change `static UwacSeat* display_destroy_seat(UwacDisplay* d, uint32_t name)` to not return the destroyed seat to fix this fully :-)

PR #4973, opened 2018-10-29
Allow fixed drive redirection (Windows) and fix macos drive duplication.

jlafontaine-devolutions:
Allow fixed drive redirection on Windows (a:, c:, ...). 
Prevent redirected drives duplication when a hot pluggable device is added or removed.


bmiklautz:
@jlafontaine-devolutions thanks. Sorry for the delay :)

PR #4974, opened 2018-10-30
Remove unsupported call to fseek when file is opened with popen

jlafontaine-devolutions:
Fix pclose hanging indefinately if a call to fseek was made on a file opened with popen.


bmiklautz:
@freerdp-bot test

PR #4976, opened 2018-10-31
remotefx: fix rlgr last bits of encoder

hardening:
The last bits of the bitstream were not flushed in the output, this patch fixes it.


libfreerdp/codec/rfx_bitstream.h
```diff
@@ -77,6 +77,13 @@ typedef struct _RFX_BITSTREAM RFX_BITSTREAM;
 			bs->byte_pos++; \
 		} \
 	} } while (0)
+#define rfx_bitstream_flush(bs) do { \
```
akallabeth: Might be a good idea to make this a `static INLINE` function (adds some type safety)

PR #4979, opened 2018-10-31
license: support CAL license

hardening:
This patch simplifies the licensing code mutualizing encryption / decryption routines. It also adds the support for client_info packet that allows to send a previously saved CAL file.

### DIFF
--- a/libfreerdp/core/license.c
+++ b/libfreerdp/core/license.c
@@ -4,6 +4,7 @@
  *
  * Copyright 2011-2013 Marc-Andre Moreau <marcandre.moreau@gmail.com>
  * Copyright 2014 Norbert Federa <norbert.federa@thincast.com>
+ * Copyright 2018 David Fort <contact@hardening-consulting.com>
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +25,9 @@
 
 #include <winpr/crt.h>
 #include <winpr/crypto.h>
+#include <winpr/shell.h>
+#include <winpr/path.h>
+
 #include <freerdp/log.h>
 
 #include "redirection.h"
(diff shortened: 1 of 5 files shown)


bmiklautz:
@hardening what servers have you tested against? I'll do some testing against our test setup as well.

hardening:
I guess any server configured with license server configured in per device mode should trigger the added code.

hardening:
Hum thinking of it perhaps in per user mode it is triggered too, not tested that.

hardening:
@akallabeth we've discussed of that with @bmiklautz (privately). The changes are useful with a server license configured in per device mode (the server seems to link the MAC address and the CAL file).

akallabeth:
@hardening then I'll check with @bmiklautz tomorrow and approve it ;)

akallabeth:
@hardening nice improvement, btw

hardening:
option for old behaviour added (off by default)

bmiklautz:
@hardening thanks. Looks good for me. When we do have a second review I opt in to merge it.

libfreerdp/core/license.c
```diff
@@ -132,21 +139,178 @@ static void license_print_product_info(LICENSE_PRODUCT_INFO* productInfo)
...
+
+	if (written != 1)
+	{
+		DeleteFile(filepathNew);
+		goto fail_file_open;
+	}
+
+	ret = MoveFileEx(filepathNew, filepath, MOVEFILE_REPLACE_EXISTING);
+
+fail_file_open:
```
akallabeth: initializing variables with `NULL` could avoid having more than one `goto` label

libfreerdp/core/license.c
```diff
@@ -132,21 +139,178 @@ static void license_print_product_info(LICENSE_PRODUCT_INFO* productInfo)
...
+		goto error_read;
+
+	*dataLen = length;
+
+	fclose(fp);
+	free(calPath);
+	free(licenseStorePath);
+	return ret;
+
+error_read:
```
akallabeth: initializing variables with `NULL` can avoid having to use multiple `goto` labels

hardening:
I think `LICENSE_NULL_CLIENT_RANDOM` and `LICENSE_NULL_PREMASTER_SECRET`are flags to behave like `rdesktop` (yes it's not secure on the license part).

libfreerdp/core/license.c
hardening: good catch, changing this.

libfreerdp/core/license.c
hardening: here I can simplify a little, but not completely getting rid of the 2 paths

hardening:
@akallabeth I've fixed what you pointed out

Issue #4983, opened 2018-11-05
free(): invalid next size (fast) in aecc77cd77335526aeefcba1bb36727e9cbb117b


informatimago:
**Describe the bug**

```
commit aecc77cd77335526aeefcba1bb36727e9cbb117b
xfreerdp /v:${windows_host} /u:${user}
```
ends in:
> free(): invalid next size (fast)
followed by a SIGABORT.

[freerdp.log](https://github.com/FreeRDP/FreeRDP/files/2549319/freerdp.log)
[compile-freerdp.txt](https://github.com/FreeRDP/FreeRDP/files/2549352/compile-freerdp.txt)

**To Reproduce**
- checkout current HEAD (aecc77cd77335526aeefcba1bb36727e9cbb117b)
- compile with command in compile-freerdp.txt
- run xfreerdp /v:${windows_host} /u:${user}

**Expected behavior**
Open a RDP window.

**Desktop:**
 - OS: GNU/Linux ubuntu 18.08
Linux L0253344 4.15.0-36-generic #39-Ubuntu SMP Mon Sep 24 16:19:09 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

informatimago:
Compiling with  ` export CFLAGS='-fsanitize=address -fsanitize=bounds -fsanitize=object-size -fsanitize=null -g -g3 -ggdb -O0 '`
gives the following:

> ==30358==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60300003c9f0 at pc 0x7fc2e2dde77a bp 0x7fc2c9302fd0 sp 0x7fc2c9302778
> WRITE of size 92 at 0x60300003c9f0 thread T1
>     #0 0x7fc2e2dde779  (/usr/lib/x86_64-linux-gnu/libasan.so.4+0x79779)
>     #1 0x7fc2df41ce99 in kerberos_EncryptMessage /home/pjb/src/span/sources/informatimago-FreeRDP/winpr/libwinpr/sspi/Kerberos/kerberos.c:681
>     #2 0x7fc2df41f2c8 in negotiate_EncryptMessage /home/pjb/src/span/sources/informatimago-FreeRDP/winpr/libwinpr/sspi/Negotiate/negotiate.c:507
>     #3 0x7fc2df42c0e5 in winpr_EncryptMessage /home/pjb/src/span/sources/informatimago-FreeRDP/winpr/libwinpr/sspi/sspi_winpr.c:1482
[...]

PR #4985, opened 2018-11-06
Fix Issue #4983 : Increased size of buffer to encrypt hash with kerbe…

informatimago:
So, I would propose this patch which seems to work, but in general, I find it unsatisfactory that the input buffers be overwritten by gss when the output can be of different size. This SEC_BUFFER API looks dumb to me.


hardening:
@freerdp-bot can you test that

bmiklautz:
@informatimago the API itself is modeled after https://docs.microsoft.com/en-us/windows/desktop/secauthn/using-secbufferdesc-and-secbuffer

PR #5015, opened 2018-11-14
cleanup based on PVS analyzer findings

chipitsine:


hardening:

PR #5018, opened 2018-11-15
cmdline: add window-position argument to set initial window position

hardening:
This is useful if you want the window to be at a given position. The patch also mutualizes the parsing of <xpos>x<ypos> or <width>x<height> arguments.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -220,6 +220,7 @@ static int freerdp_client_command_line_pre_filter(void* context, int index,
 	return 0;
 }
 
+
 BOOL freerdp_client_add_device_channel(rdpSettings* settings, int count,
                                        char** params)
 {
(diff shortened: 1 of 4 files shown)


client/common/cmdline.c
```diff
@@ -1315,6 +1316,45 @@ static void activate_smartcard_logon_rdp(rdpSettings* settings)
...
+		return FALSE;
+	if (v1)
+		*v1 = v;
+
+	xcharpos = strchr(input, 'x');
+	if (!xcharpos || xcharpos != endPtr)
+		return FALSE;
+
+	errno = 0;
+	v = strtoul(xcharpos + 1, &endPtr, 10);
```
akallabeth: Might be a good idea to split the `strtoul` off to a function too (can unify all the uses in command line parser this way)

akallabeth: Would suggest something like `BOOL parse_value(const char* input, unsigned long* result, unsigned long max)`

akallabeth:
@freerdp-bot test

bmiklautz:
@freerdp-bot test

PR #5023, opened 2018-11-17
call xf_SetWindowTitle before XMapWindow

ileGITimo:
Need to set window name before mapping, so window managers like fvwm, which allow mapping windows to a  virtual desktop/page based on the window name, can act on it.


akallabeth:
@freerdp-bot test

PR #5025, opened 2018-11-18
fix for connection with openssl 1.1.1

DustPuppyNet:
openssl 1.1.1 (from debian/testing) connects with tls1.3 and does not revert to lower tls versions, if the server announces them. Instead, the connection is dropped with "SSL_ALERT wrong version".
The following fixes at least tls_connect() to remote servers.

* change obsolete call to SSLv23_client_method in tls_connect() with TLS_client_method()
* set min TLS Version


akallabeth:
@freerdp-bot test

bmiklautz:
@DustPuppyNet thanks!

PR #5039, opened 2018-11-20
Fixed recreate activity issue with bluetooth devices.

mcsong:
It will fix recreation activity issue with bluetooth devices.


akallabeth:
@freerdp-bot test

akallabeth:
Sorry for asking this, but what exactly is the issue and how can we test that?

mcsong:
After connected with a remote desktop, I tried to use bluetooth keyboard and mouse. When I tried to connect a bluetooth keyboard, SessionActivity will be recreate. And disconnection will be recreate SessionActivity, again. So, this patch will be fix recreation issue with bluetooth devices.

PR #5055, opened 2018-11-22
Fixed #2021: Removed *get_fds functions and use.

### LINKED ISSUE #2021: Remove remaining select in client code and fix *get_fds functions
- Client code currently uses `select` instead of `WaitForMultipleObjects`
- `*get_fds` functions return a file descriptors instead of `HANDLE`
- `*get_fds` functions currently only work with static array sizes as arguments.
- Not all relevant event sources are returned with `*get_fds` functions

Already tried to fix this with #1910, but split code.
- Created pull requrests #1963 and #1964 to prepare winpr and move some common client code to libfreerdp.
- Working branch https://github.com/akallabeth/FreeRDP/tree/client-main-loop
  - Replaces client code `select`
  - Replaces server code `select`
  - Replaces `*get_fds` with `*get_handles`, which reallocs space as necessary to return handles.
  - Return additional event sources (transport thread, update thread, ...) allowing `INFINTE` wait in client.

akallabeth:
* Removed all useages of `*_get_fds` functions
* Renamed `*check_fds` functions to `*check_event_handles`
* Added compatibility defines for `*check_fds`
* Still keeping the funtions around for compatibility

### DIFF
--- a/server/Windows/wf_peer.c
+++ b/server/Windows/wf_peer.c
@@ -167,43 +167,37 @@ BOOL wf_peer_accepted(freerdp_listener* instance, freerdp_peer* client)
 
 static DWORD WINAPI wf_peer_socket_listener(LPVOID lpParam)
 {
-	int i, fds;
-	int rcount;
-	int max_fds;
-	void* rfds[32];
-	fd_set rfds_set;
 	wfPeerContext* context;
 	freerdp_peer* client = (freerdp_peer*) lpParam;
-	ZeroMemory(rfds, sizeof(rfds));
+
+	if (!client || !client->context)
+	{
+		ExitThread(1);
+		return 1;
+	}
+
 	context = (wfPeerContext*) client->context;
 
 	while (1)
 	{
-		rcount = 0;
+		DWORD status;
+		HANDLE handles[64];
+		DWORD usedHandles = client->GetEventHandles(client, handles, ARRAYSIZE(handles));
 
-		if (client->GetFileDescriptor(client, rfds, &rcount) != TRUE)
+		if (usedHandles == 0)
 		{
 			WLog_ERR(TAG, "Failed to get peer file descriptor");
 			break;
 		}
 
-		max_fds = 0;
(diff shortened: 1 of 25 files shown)


akallabeth:
No, it just removes the possibility to use get a fd directly, there is still the other function returning a HANDLE

server/Windows/wf_peer.c
```diff
@@ -167,50 +167,45 @@ BOOL wf_peer_accepted(freerdp_listener* instance, freerdp_peer* client)
...
-	int rcount;
-	int max_fds;
-	void* rfds[32];
-	fd_set rfds_set;
 	wfPeerContext* context;
 	freerdp_peer* client = (freerdp_peer*) lpParam;
-	ZeroMemory(rfds, sizeof(rfds));
+
+	if (!client || !client->context)
+	{
```
hardening: no self-check in not exported static function

hardening:
So how external project will be able to select fds (the infamous `WaitForMultipleObjects` not being an option for most projects )?

akallabeth:
@hardening `WaitForMultipleObjects` uses `poll` internally with `select` as fallback if former is not available. So, adding `fd` to the list should be no problem.

hardening:
My point is more that I don't see projects using that when you have your own eventloop, `WaitForMultipleObject` is a real pain to use.

hardening:
updated: `winpr_Handle_getFd` should allow to still retrieve the underlying fd.

akallabeth:
@hardening No, `winpr_Handle_getFd` is not exposed.

akallabeth:
@hardening @bmiklautz Ok, so we agree on `WTSVirtualChannelManagerGetFileDescriptor` being required.
Which functions / function pointers do you think will still be required additionally?

akallabeth:
@hardening Had a little talk with @bmiklautz and updated the pull accordingly.
Basically, this removes all usage of the functions but keeps them around for external projects to use.
Could you have a second look please?

PR #5057, opened 2018-11-22
Unified initialization of OrderSupport

akallabeth:
Added a library internal function freerdp_settings_set_default_order_support
which initializes the OrderSupport array of settings.
Now clients no longer need to set this up on their own, if they
do not implement their own hardware accelerated order processing.


akallabeth:
@hardening updated the pull to accommodate the new sample client. You still approve?

hardening:
@akallabeth  am I wrong if I say that with your changes, third party softwares using freerdp can't set specific order support anymore (as the default setter is always called in `freerd_client_connect`)?

akallabeth:
@hardening yes, you're wrong. The function is called before preconnect where it was historically. Add it there again and you have the old behaviour.

PR #5062, opened 2018-11-23
Fix #5061: Initialize output buffer transparent.

### LINKED ISSUE #5061: freerdp_image_copy_from_pointer_data does not initialize buffer
alpha data is not correctly initialized to transparent

akallabeth:
If a destination buffer is not allocated with `calloc` but some `malloc` or other, not every pixel is initialized properly.
Fix this by setting the buffer to `0` initially.

### DIFF
--- a/libfreerdp/codec/color.c
+++ b/libfreerdp/codec/color.c
@@ -179,6 +179,12 @@ BOOL freerdp_image_copy_from_pointer_data(
 	if (nDstStep <= 0)
 		nDstStep = dstBytesPerPixel * nWidth;
 
+	for (y = nYDst; y < nHeight; y++)
+	{
+		BYTE* pDstLine = &pDstData[y * nDstStep + nXDst * dstBytesPerPixel];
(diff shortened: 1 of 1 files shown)


mfleisz:
Afaict this function allows to copy the pointer data anywhere onto an image i.e. we can specify x/y offsets and the destination image's area could be bigger than the actual pointer data is.

In that case the memset is wrong as it doesn't account for correct x/y offset handling.

I think the function is correct as it is ... the caller just wants to copy the pointer data somewhere into the destination buffer. If he wants to have a cleared image with just the pointer data in it it is the callers responsibility in my opinion to clear the image before.

akallabeth:
@mfleisz You're right about the offsets, did not take those into account.
The other issue is not valid, you can simply try to allocate a cairo surface and see the result of this function with and without initialization with a background color.

akallabeth:
@hardening we're talking about a mouse pointer here, small tiles

akallabeth:
@mfleisz Can we merge this one?

PR #5064, opened 2018-11-23
Fixed a broken length check in rdg_process_packet

akallabeth:
HTTP gateway connections aborted due to this.
Additionally add more verbose error logging in RDG.


akallabeth:
@Olen thx for the fast feedback, really helps to spot stuff before the next release :)

PR #5066, opened 2018-11-26
Fix #4768: Set SSL verify purpose to ANY

### LINKED ISSUE #4768: Public-Signed Certificate Validation Error
I recently upgraded to openSUSE Leap 15.0 on my laptop, and now I am getting a "certificate could not be verified" warning when connecting to servers that are using a public-signed certificate.

I redownloaded and recompiled on both openSUSE Leap 42.3 and openSUSE Leap 15.0 today - it works on 42.3, but not on the newer 15.0.

I am attaching the output from a strace for both operating systems. It looks to me like both are checking for /var/lib/ca-certificates/openssl/4bfab552.0 (which exists, and is the correct CA) - but on 15.0, it never opens that file?

[openSUSE-Leap-42.3-Works.TXT](https://github.com/FreeRDP/FreeRDP/files/2229158/openSUSE-Leap-42.3-Works.TXT)
[openSUSE-Leap-15-Error.TXT](https://github.com/FreeRDP/FreeRDP/files/2229159/openSUSE-Leap-15-Error.TXT)

akallabeth:
Found https://stackoverflow.com/questions/40685111/why-openssls-pkcs7-verify-requires-smimesign-certificate-purpose which explains the references behaviour.

Set to server since we verify the certificate of the server and not the client in that case only.

### DIFF
--- a/libfreerdp/crypto/crypto.c
+++ b/libfreerdp/crypto/crypto.c
@@ -756,6 +756,8 @@ char* crypto_cert_issuer(X509* xcert)
 
 BOOL x509_verify_certificate(CryptoCert cert, char* certificate_store_path)
 {
+	int purpose;
+	X509_VERIFY_PARAM* verify_param;
 	X509_STORE_CTX* csc;
 	BOOL status = FALSE;
 	X509_STORE* cert_ctx = NULL;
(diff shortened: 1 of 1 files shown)


bmiklautz:
+1 but I would like (even if not fully correct) that we add the previous purpose (s/mime) again. Then it's no change in functionality

PR #5068, opened 2018-11-26
Refactored BufferPool API

akallabeth:
* `BufferPool` structs now opaque
* Replaced `int` types with `SSIZE_T` or `size_t` variants
* Added proper overflow checks for allocation sizes


akallabeth:
Broken, will redo when there is time.

PR #5073, opened 2018-11-28
Fix #4764: Second try, use X509_STORE_CTX_set_purpose

### LINKED ISSUE #4764: Compiling FreeRDP for MACOSX
Hey,

Where can I find the latest documentation to compile FreeRDP Mac OSX.

I noticed the instructions are for mountain lion?

thanks

akallabeth:
@KenHolloway Could you check this works for you?

### DIFF
--- a/libfreerdp/crypto/crypto.c
+++ b/libfreerdp/crypto/crypto.c
@@ -756,7 +756,6 @@ char* crypto_cert_issuer(X509* xcert)
 
 BOOL x509_verify_certificate(CryptoCert cert, char* certificate_store_path)
 {
-	int purpose;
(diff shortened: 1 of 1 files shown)


KenHolloway:
Yes, this seems to work for me.

I found a reference that I think says that "Any Purpose" will not work:

https://serverfault.com/questions/111625/openssl-x509-purpose-flag-any-purpose-what-is-this

> 'The "Any Purpose" setting is something which lets anything through and performs no checks at all. It was put there originally as a way for people to use broken certificates if they had no other choice and could live with the consequences... Since then CA checks have been made mandatory in the code even if "Any Purpose" is set. So if you actually tried to use that certificate as a CA it would be rejected.'

PR #5075, opened 2018-11-28
Fixed clipboard channel cleanup.

akallabeth:


akallabeth:
@mfleisz as one message can be broken up in separate packets it might still be necessary.

mfleisz:
@akallabeth Ah yes - that's true 👍

PR #5094, opened 2018-12-04
Make "suppress output" setting available on command line

akallabeth:
Replaces #5090 
Naming of original command line option was misleading, correct that.
Thank you @praimmugen for your contribution.


akallabeth:
@freerdp-bot test

PR #5096, opened 2018-12-04
Fix #5019: Added channel up/downmixing code for mono/stereo.

### LINKED ISSUE #5019: Channel up/downsampling in freerdp_dsp_encode
Implement channel number conversion in `freerdp_dsp_encode`

Use case:
* `freerdp_dsp_encode` is used by `audin` and server side `sound` channels.
* Input is ususally in a fixed format (e.g. `PCM 16bit, 44100Hz, 2 channels) but a client may request a different format (e.g. PCM 8bit, 22050Hz, 1 channel or PCM 16bit, 48000Hz, 5 channels)
* Currently only the conversion of sample rate and PCM 16bit to PCM 8bit is working, but the channels can not be changed.

akallabeth:


libfreerdp/codec/dsp.c
```diff
@@ -111,6 +111,90 @@ struct _FREERDP_DSP_CONTEXT
...
+	size_t samples;
+	size_t x, y;
+
+	if (!context || !data || !length)
+		return FALSE;
+
+	if (srcFormat->wFormatTag != WAVE_FORMAT_PCM)
+		return FALSE;
+
+	bpp = srcFormat->wBitsPerSample > 8 ? 2 : 1;
```
mfleisz: This will cause problems with formats that use less than 8 bits per sample (i.e. ADPCM)

akallabeth:
@mfleisz No, this is raw pcm from a source, output might be adpcm but not input

bmiklautz:
@mfleisz it it's ok for you and tested any objection for merging?

Issue #5078, opened 2018-11-29
RAIL windows hidden once can never be shown again


hualet:
**Describe the bug**

Rail windows hidden once can never be shown again.

**To Reproduce**
1. make the test application attached a remote app and give it a fixed command line argument 5000 or time interval in milliseconds.
2. run the remote app with xfreerdp.
3. you can see the main window can show once, hide and never be shown again.

**Expected behavior**
The rail window should do showing and hiding in every 5s or the time interval you specified.

**Application details**
* FreeRDP 2.0.0
* Command line used
freerdp /sec:rdp /u:Administrator /p:'larry+1s' /app:'||test_window' /v:10.0.13.216
* output of `/buildconfig`

> This is FreeRDP version 2.0.0-rc4 (e21b72c95+debian)
[...]

PR #5097, opened 2018-12-04
Fixed RAIL window can't show again once hidden

hualet:
The cause is very simple: we didn't map the xwindow on receiving
WINDOW_SHOW. but doing that causes another problem that you can't
hide a window anymore, and that is because whlie window hiding, the
_NET_WM_STATE and WM_STATE properies of the xwindow may change, in
the function `xf_event_PropertyNotify` we just assume that windows
not maximized, not minimized, yet not showing normally should be
corrected to be shown, we just need to consider the situation that
the window is hidden here.

fix: #5078


akallabeth:
@hualet Thank you, will have a look at it. Don´t know if I manage today but from an initial glance it should be ok.

hualet:
Wow, PRs get merged so quickly, I'm so impressed :) Good job @akallabeth @hardening

Issue #5108, opened 2018-12-06
Support RAIL notification icon


hualet:
**Is your feature request related to a problem? Please describe.**
Applications that rely on notification icons to work well is not usable in RAIL mode, for example, IM Apps, downloaders etc.

**Describe the solution you'd like**
Provide basic notification icon support in RAIL mode.

**Describe alternatives you've considered**
NO.

**Additional context**
NO

hualet:
Actually, I'm working an implementation, it can work now, not perfectly working thro. 
And my question is it OK that I file a PR now? or I should wait until it's perfectly done?

akallabeth:
@hualet you can create a pull and prefix that with `[WIP]`. That way we can let our build bot check your changes too.
Also please run the code formatter script before committing changes.

And thank you for this :)

PR #5110, opened 2018-12-06
[WIP] RAIL notification icon support

hualet:
Now the implementation can work as:

1. show a tray icon for remote apps.
2. support basic button clicks.

but it also have known issues:

1. the icon is poorly drawn, the background is black.
2. the local tray icon windows sometimes are not embeded into the system tray.
3. right click on the tray icon, the context menu shows at somewhere else.

trying to fix: #5108

### DIFF
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -907,13 +907,13 @@ static BOOL xf_rail_notify_icon_common(rdpContext* context,
 	if (orderInfo->fieldFlags & WINDOW_ORDER_STATE_NEW)
 	{
 		notifyIcon = (xfAppNotifyIcon*) calloc(1, sizeof(xfAppNotifyIcon));
-		notifyIcon->xfc = xfc;
-		notifyIcon->windowId = orderInfo->windowId;
-		notifyIcon->notifyIconId = orderInfo->notifyIconId;
 
 		if (!notifyIcon)
 			return FALSE;
 
+		notifyIcon->xfc = xfc;
+		notifyIcon->windowId = orderInfo->windowId;
+		notifyIcon->notifyIconId = orderInfo->notifyIconId;
 		HashTable_Add(xfc->railNotifyIcons, (void*)(UINT_PTR) orderInfo->notifyIconId,
 		              (void*) notifyIcon);
 		xf_appNotifyIconCreate(xfc, notifyIcon);
@@ -1310,6 +1310,7 @@ static void rail_notify_icon_free(void* value)
 
 	XUnmapWindow(icon->xfc->display, icon->handle);
 	XDestroyWindow(icon->xfc->display, icon->handle);
+	free(icon);
 }
 
 int xf_rail_init(xfContext* xfc, RailClientContext* rail)
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -970,7 +970,7 @@ static BOOL xf_rail_notify_icon_common(rdpContext* context,
 			XDestroyImage(notifyIcon->image);
 
 		notifyIcon->image = image;
-		// bgraPixels will be freed by XDestroyImage.
+		/* bgraPixels will be freed by XDestroyImage. */
 	}
 
 	if (orderInfo->fieldFlags & WINDOW_ORDER_CACHED_ICON)
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -880,6 +880,35 @@ static BOOL xf_rail_window_cached_icon(rdpContext* context,
 static BOOL xf_rail_notify_icon_common(rdpContext* context,
                                        WINDOW_ORDER_INFO* orderInfo, NOTIFY_ICON_STATE_ORDER* notifyIconState)
 {
+	xfContext* xfc = NULL;
+	xfAppNotifyIcon* notifyIcon = NULL;
+	xfc = (xfContext*) context;
+
+	if (orderInfo->fieldFlags & WINDOW_ORDER_STATE_NEW)
+	{
+		notifyIcon = (xfAppNotifyIcon*) calloc(1, sizeof(xfAppNotifyIcon));
+		notifyIcon->xfc = xfc;
+		notifyIcon->windowId = orderInfo->windowId;
+		notifyIcon->notifyIconId = orderInfo->notifyIconId;
+
+		if (!notifyIcon)
+			return FALSE;
+
+		HashTable_Add(xfc->railNotifyIcons, (void*)(UINT_PTR) orderInfo->notifyIconId,
+		              (void*) notifyIcon);
+		xf_appNotifyIconCreate(xfc, notifyIcon);
+	}
+	else
+	{
+		notifyIcon = (xfAppNotifyIcon*) HashTable_GetItemValue(xfc->railNotifyIcons,
+		             (void*)(UINT_PTR) orderInfo->notifyIconId);
+	}
+
+	if (!notifyIcon)
+	{
+		return FALSE;
+	}
+
 	if (orderInfo->fieldFlags & WINDOW_ORDER_FIELD_NOTIFY_VERSION)
 	{
 	}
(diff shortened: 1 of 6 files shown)


client/X11/xf_rail.c
```diff
@@ -698,6 +708,23 @@ static BOOL convert_icon_color_to_argb(ICON_INFO* iconInfo, BYTE* argbPixels)
 	       );
 }
 
+static void convert_argb_to_bgra(BYTE* argbPixels, BYTE* bgraPixels, int len)
+{
```
akallabeth: Why not use `freerdp_image_copy`?

client/X11/xf_window.c
```diff
@@ -1133,3 +1132,84 @@ xfAppWindow* xf_AppWindowFromX11Window(xfContext* xfc, Window wnd)
...
+	return TRUE;
+}
+
+xfAppNotifyIcon* xf_AppNotifyIconFromX11Window(xfContext* xfc, Window wnd)
+{
+	int index;
+	int count;
+	ULONG_PTR* pKeys = NULL;
+	xfAppNotifyIcon* notifyIcon;
+	count = HashTable_GetKeys(xfc->railNotifyIcons, &pKeys);
```
akallabeth: `xfc->railNotifyIcons == NULL` if not used with RAIL

client/X11/xf_rail.c
```diff
@@ -880,6 +907,35 @@ static BOOL xf_rail_window_cached_icon(rdpContext* context,
 static BOOL xf_rail_notify_icon_common(rdpContext* context,
                                        WINDOW_ORDER_INFO* orderInfo, NOTIFY_ICON_STATE_ORDER* notifyIconState)
 {
+	xfContext* xfc = NULL;
+	xfAppNotifyIcon* notifyIcon = NULL;
+	xfc = (xfContext*) context;
+
+	if (orderInfo->fieldFlags & WINDOW_ORDER_STATE_NEW)
+	{
+		notifyIcon = (xfAppNotifyIcon*) calloc(1, sizeof(xfAppNotifyIcon));
```
akallabeth: unchecked allocation

client/X11/xf_rail.c
```diff
@@ -1214,6 +1311,17 @@ static void rail_window_free(void* value)
...
 
+static void rail_notify_icon_free(void* value)
+{
+	xfAppNotifyIcon* icon = (xfAppNotifyIcon*) value;
+
+	if (!icon)
+		return;
+
+	XUnmapWindow(icon->xfc->display, icon->handle);
+	XDestroyWindow(icon->xfc->display, icon->handle);
```
akallabeth: memory leak, `xfAppNotifyIcon` was allocated with `calloc`

client/X11/xf_rail.c
```diff
@@ -880,6 +907,35 @@ static BOOL xf_rail_window_cached_icon(rdpContext* context,
...
+	{
+		notifyIcon = (xfAppNotifyIcon*) calloc(1, sizeof(xfAppNotifyIcon));
+		notifyIcon->xfc = xfc;
+		notifyIcon->windowId = orderInfo->windowId;
+		notifyIcon->notifyIconId = orderInfo->notifyIconId;
+
+		if (!notifyIcon)
+			return FALSE;
+
+		HashTable_Add(xfc->railNotifyIcons, (void*)(UINT_PTR) orderInfo->notifyIconId,
```
akallabeth: '0' is a valid notification ID but `HashTable_Add` will not accept that.

hardening:
@akallabeth could be nice to report the crash conditions to fedora (as an X client is not supposed to crash the whole server)

akallabeth:
@hardening still trying to get a proper crash dump ;)

akallabeth:
The tray icons should be set by `libappindicator` or similar and not as a simple X window if possible.

hualet:
Thank you for your reviewing, I'll update the PR soon.

> The tray icons should be set by `libappindicator` or similar and not as a simple X window if possible.

The reason that I didn't use libappindicator is that I don't know if it's proper to introduce new dependencies :P, if that's acceptable, I'll try that @akallabeth

hualet:
Oh, libappindicator seems to be only working with GTK,  so I don't think it's a very good choice.

client/X11/xf_rail.c
hualet: Does that situation that '0' as a notification ID exist in practical ? How do I suppose to fix this ? :sweat_smile:

akallabeth:
@hualet problem with x11 windows is that they look like the app has 2 entries in most environments then but still no notification icon. 
New deps are no problem, just check it still works without them activated ;)

hualet:
> New deps are no problem, just check it still works without them activated ;)

I'm reading gtk's source code to see how can I fix the tray icon problem.
because introducing libappindicator/gtk means we need another main loop, really overwhelming I think @akallabeth

client/X11/xf_rail.c
akallabeth: yes it does. @hardening pointed me in a simple direction, just set a comparison function that compares the pointers by type (UINT32) and use something like `&notifyIcon->notifyIconId` as key

client/X11/xf_rail.c
```diff
@@ -955,40 +948,29 @@ static BOOL xf_rail_notify_icon_common(rdpContext* context,
...
-		                     ZPixmap, 0, (char*) pixels, icon.width, icon.height,
+		                     ZPixmap, 0, (char*) bgraPixels, icon.width, icon.height,
 		                     xfc->scanline_pad, 0);
 
 		if (notifyIcon->image)
 			XDestroyImage(notifyIcon->image);
 
 		notifyIcon->image = image;
-		free(argbPixels);
+		// bgraPixels will be freed by XDestroyImage.
```
akallabeth: No C++ comments please

hualet:
@akallabeth I've fixed all of the mistakes you mentioned above, except:

1. hashtable don't accept uint32, which I've created a new PR https://github.com/FreeRDP/FreeRDP/pull/5120 trying to follow up the direction you told me.
2. the tray icon background is black, I still don't have a very good way to fix that.

can you review my code again, thanks?

akallabeth:
@hualet 
* The hash table issue can be solved just like #5117 
* I'll have a look at it.

akallabeth:
@hualet I've fixed a few remaining issues and rebased to current master. See https://github.com/akallabeth/FreeRDP/tree/rail_notify_icons (note: I squashed your changes to a single commit)

hualet:
> @hualet I've fixed a few remaining issues and rebased to current master. See https://github.com/akallabeth/FreeRDP/tree/rail_notify_icons (note: I squashed your changes to a single commit)

OK, so I need to close this PR now?

akallabeth:
@hualet if you want to. I'm still struggling with getting the notification icon to work though.
I can see the window icon in the window bar (gnome that is with the extension for it) but:
* No icon for the notification
* Everything I expect from the notification icon is not working, aka context menu, ...

hualet:
> @hualet if you want to. I'm still struggling with getting the notification icon to work though.
> I can see the window icon in the window bar (gnome that is with the extension for it) but:
> 
> * No icon for the notification
> * Everything I expect from the notification icon is not working, aka context menu, ...

Hmmm.... I tested that with Mate, DDE and stalonetray, all seems good,  can you test with stalonetray ?

hualet:
Another thing I must mention is that the context menu is not working perfectly here too, the context menu shows at random position :(

akallabeth:
@hualet I've tried with cinnamon and mate and only in mate I could see an `icon` (with a width of 1 pixel)

hualet:
> @hualet I've tried with cinnamon and mate and only in mate I could see an `icon` (with a width of 1 pixel)

Which application are you testing with? I can test it here too.

akallabeth:
@hualet Yes, with KDE plasma I can see an icon too. Application not relevant, all behave the same in this regard.
Still, I get many problems with nearly all WM so this will need some work before getting merged.

PR #5111, opened 2018-12-06
Use windowID + 1 as key for hash table as 0 is an invalid key.

akallabeth:


hardening:
@akallabeth instead of that hack, I'd suggest to index the hashtable on `railWindow->windowId` and to have a comparison function of the hashtable that operate on `UINT32 *` (and compare the pointed value). Having the 0 value as forbidden always leads to problems and corner cases.

akallabeth:
@hardening hmm, might work if a `keyClone` function is set and we use the address of the integer as `HashTable_Add` argument.

akallabeth:
@hardening Ok, had a look at it and the reason for this `hack` is the following:
* Currently we don't need to allocate space for the key as it fits in a pointer value
* With the requested change the integer must be allocated to be able to compare

so, either we use it this way or we need to change the way `HashTable` works if we want to avoid allocations. (`NULL` values as pointer to the key are rejected for a reason)

client/X11/xf_rail.c
```diff
@@ -327,14 +327,14 @@ static BOOL xf_rail_window_common(rdpContext* context,
 			return FALSE;
 		}
 
-		HashTable_Add(xfc->railWindows, (void*)(UINT_PTR) orderInfo->windowId,
+		HashTable_Add(xfc->railWindows, (void*)(UINT_PTR) orderInfo->windowId + 1,
 		              (void*) appWindow);
```
hardening: My suggestion would be to do
```C
HashTable_Add(xfc->railWindows, &appWindow->windowId, (void*) appWindow);
```
plus you add a comparison function on the `railWindows` that compares `UINT32 *`.
and getting values becomes
```C
return (xfAppWindow*) HashTable_GetItemValue(xfc->railWindows, &windowId);
```

hardening:
@akallabeth what about https://github.com/hardening/FreeRDP/tree/rail_hash ? (I've not been able to fully tested as my RAIL setup is broken here, but the windows I had seemed to work as expected)

side note: the  same should be done for gfx surfaces

client/X11/xf_rail.c
akallabeth: @hardening ok, you got me :+1 missed the obvious solution because there are local variables used when finding existing windows.

PR #5117, opened 2018-12-07
rails: allow a window with id == 0

hardening:


client/X11/xf_rail.c
```diff
@@ -1232,12 +1231,13 @@ int xf_rail_init(xfContext* xfc, RailClientContext* rail)
 	rail->ServerMinMaxInfo = xf_rail_server_min_max_info;
 	rail->ServerLanguageBarInfo = xf_rail_server_language_bar_info;
 	rail->ServerGetAppIdResponse = xf_rail_server_get_appid_response;
-	xfc->railWindows = HashTable_New(TRUE);
 
+	xfc->railWindows = HashTable_New(TRUE);
 	if (!xfc->railWindows)
 		return 0;
-
+	xfc->railWindows->keyCompare = (HASH_TABLE_KEY_COMPARE_FN)rail_window_key_equals;
```
akallabeth: Better not cast function pointers, that silences the compiler when the signature changes. Better do the casts in the implementation

bmiklautz: +1 for not casting. In case we have changes on the API the compiler wouldn't complain.

bmiklautz:
@hardening nice solution!

client/X11/xf_rail.c
hardening: I will do the change, but who would be mad enough to change the API. Well it could be me to change the equals_key function to take `const void *` in argument instead of `void *`.

akallabeth: @hardening I had the same thought ;)

akallabeth:
@hardening Oops, missed one thing: You did not set a proper hash function, the default one works on pointers and so you can't compare different memory locations for the same integer.

akallabeth:
@hardening got a fix in https://github.com/akallabeth/FreeRDP/tree/rail_notify_icons but need to test the other changes before creating a new pull for that.

Issue #5121, opened 2018-12-10
Failure to get clipboard data when copying Windows Update info


staples1347:
Local Host: Gentoo x86_64 with X
Tested Remote Hosts: Windows 7, Windows 8, Windows Server 2008 R2, and Windows Server 2012
FreeRDP: Version: 2.0.0_rc2, 2.0.0_rc4, latest master: https://github.com/FreeRDP/FreeRDP/commit/fc12b9336f6c6b5b49bdfc564ce8e4443d9e82ef

To reproduce: Go to "Control Panel\All Control Panel Items\Windows Update" in Windows Explorer then after clicking on an update to get to "Control Panel\All Control Panel Items\Windows Update\Select updates to install", right-click on the text under the Name column and click "Copy details"

Command Line: xfreerdp /u:user /p:pwd

freerdp exits with message:
[17:34:01:151] [9552:9595] [ERROR][com.freerdp.client.x11] - failed to get clipboard data in format UTF8_STRING [source format CF_UNICODETEXT]
[17:34:01:151] [9552:9595] [ERROR][com.freerdp.channels.cliprdr.client] - ServerFormatDataResponse failed with error 1359!
[17:34:01:151] [9552:9595] [ERROR][com.freerdp.channels.cliprdr.client] - cliprdr_process_format_data_response failed with error 1359!
[17:34:01:151] [9552:9595] [ERROR][com.freerdp.channels.cliprdr.client] - cliprdr_order_recv failed with error 1359!
[17:34:01:151] [9552:9553] [ERROR][com.freerdp.core] - cliprdr_virtual_channel_client_thread reported an error. Error was 1359

PR #5122, opened 2018-12-10
[WIP] Workaround for #5121: Ignore invalid unicode symbols.

akallabeth:
As reported the strings returned from the server sometimes contain
unmappable characters. Ignore them and continue to fix this issue.
Still need to find a proper fix.

### DIFF
--- a/winpr/libwinpr/crt/unicode.c
+++ b/winpr/libwinpr/crt/unicode.c
@@ -268,7 +268,11 @@ int MultiByteToWideChar(UINT CodePage, DWORD dwFlags, LPCSTR lpMultiByteStr,
  * to the output buffer, and should match the value obtained on the first call to WideCharToMultiByte.
  *
  */
-
+/* TODO: https://github.com/FreeRDP/FreeRDP/issues/5121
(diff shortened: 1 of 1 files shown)


bmiklautz:
@akallabeth do we have any test cases?

akallabeth:
@bmiklautz the referenced issue contains a reproducible issue. I've connected to windows 2012 and followed the instructions and got invalid characters in the unicode string.

akallabeth:
@bmiklautz surprisingly this issue occurs with both, the custom unicode converter in freerdp and libicu

bmiklautz:
@akallabeth weird. Could we possibly pack this into a test case?

akallabeth:
@bmiklautz could dump the unicode string and check if conversion is successful for it, other than that I don't know a way...

akallabeth:
@bmiklautz Ok, the main reason seems to be that the clipboard data contains ~200k bytes but the string is actually much shorter (with garbage at the end...) need to check if a valid `\0` byte is ignored or the string is simply broken.

PR #5142, opened 2018-12-12
Define scancodes for virtual keys, used by VNC servers

astrand:
The list of fake keycodes can be found at:

https://cgit.freedesktop.org/xkeyboard-config/tree/keycodes/evdev#n290

This commit adds support for AltGr when xfreerdp is used in
TigerVNC/Xvnc and ThinLinc sessions.


akallabeth:
@freerdp-bot test

Issue #4820, opened 2018-08-23
Covscan defects


ondrejholy:
I've fixed multiple defects reported by covscan as a part of https://github.com/FreeRDP/FreeRDP/pull/4813. However, there are still some remaining defects which I am unsure about their solution and thus I am not going to propose any fixes for them. Some of them look like false-positives, but some of them seem to be worth looking at, e.g. the one with `USE_AFTER_FREE`. Here is the list of important defects on top of the mentioned pull request: [scan-results-imp.err](https://github.com/FreeRDP/FreeRDP/files/2315544/scan-results-imp.err.txt).

ondrejholy:
Btw there is also a list containing much more issues, which are not marked as important: [scan-results.err](https://github.com/FreeRDP/FreeRDP/files/2315551/scan-results.err.txt). There is also a lot of interesting defects, e.g. `NO_EFFECT`, `DEADCODE`. If I find some time I will propose a fix for some of those also.

akallabeth:
@ondrejholy We've fixed a lot of warnings (and bugs) in the last weeks, so I assume we can close this one?

ondrejholy:
The scan based on the current master (commit d3c1db1746b99c05d515de626d2c515e49debfdd) revealed more issues than the previous one. There is still the `USE_AFTER_FREE` defect, which would be nice to fix, but I am not sure how. I don't see anything else very important. See [scan-results-imp.err](https://github.com/FreeRDP/FreeRDP/files/2514439/scan-results-imp.err.txt) and [scan-results.err](https://github.com/FreeRDP/FreeRDP/files/2514438/scan-results.err.txt) for details.

akallabeth:
@ondrejholy looks like there were some socket leaks handling error cases of `tcp_connect`.
The `USE_AFTER_FREE` is currently harmless, but @hardening is changing the return there to a seat ID to avoid the confusion (both, developer and compiler)

ondrejholy:
Thanks that you took look at those defects. The seat id defect is not still fixed, see my comment on https://github.com/FreeRDP/FreeRDP/pull/4970. I can confirm that some socket leaks disappeared. If you are interested see updated [scan-results-imp.err](https://github.com/FreeRDP/FreeRDP/files/2553667/scan-results-imp.err.txt) and [scan-results.err](https://github.com/FreeRDP/FreeRDP/files/2553665/scan-results.err.txt).

akallabeth:
[...]

PR #5144, opened 2018-12-13
uwac: Do not return destroyed UwacSeat

ondrejholy:
This is needed to fix the following defects reported by Covscan:
use_after_free: Using freed pointer "seat".
warning: Value stored to 'seat' is never read

This is follow up for https://github.com/FreeRDP/FreeRDP/pull/4970

https://github.com/FreeRDP/FreeRDP/issues/4820


akallabeth:
@freerdp-bot test

PR #5147, opened 2018-12-13
Clarified defaults of /floatbar flag

jphein:
Just a simple clarification of the new /floatbar flag in the help file.


akallabeth:
@freerdp-bot test
@jphein thx

PR #5149, opened 2018-12-14
New option to disable user certificate dialog

akallabeth:
The new option +cert-deny aborts a connection automatically if
the certificate can not be validated by OpenSSL or via known hosts.


bmiklautz:
We do have now multiple options for certificate handling. We should bring them all together to one setting.

akallabeth:
@hardening @bmiklautz @mfleisz anyone want to merge this one?

PR #5151, opened 2018-12-14
Updates to default settings

akallabeth:
* Disable `offscreen-cache` by default
* Disable `bitmap-cache` by default
* Enable `heartbeat` by default


akallabeth:
@hardening @bmiklautz @mfleisz anyone want to merge this one?

bmiklautz:
Basically agree and the PR looks good. However since the default still is to use GDI (not rfx/gfx) compression might save at least a little bit. I'd opt in for leaving it on.

bmiklautz:
@hardening, @mfleisz  what do you think?

PR #5158, opened 2018-12-18
Improved wayland mouse support

akallabeth:
* Implements horizontal mouse wheel
* Adds support for mouse button 4 and 5


akallabeth:
@hardening Little improvement in handling mouse for wayland client.
Getting a new mouse soon, so I'll do a follow up to improve mouse wheel handling (calculate the actual delta instead of the fixed `0x78` step)

PR #5159, opened 2018-12-18
Implemented wayland display resize channel.

akallabeth:
* Implements the display resize for wayland
* Currently only single monitor
* Requires #5202 to actually display the changed resolution correctly

### DIFF
--- a/client/Wayland/wlf_disp.c
+++ b/client/Wayland/wlf_disp.c
@@ -318,9 +318,6 @@ BOOL wlf_disp_handle_configure(wlfDispContext* disp, int32_t width, int32_t heig
 	if (!disp)
 		return FALSE;
 
-	if (!disp)
-		return FALSE;
-
 	disp->targetWidth = width;
 	disp->targetHeight = height;
 	return wlf_disp_sendResize(disp);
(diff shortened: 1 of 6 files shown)


client/Wayland/wlf_disp.c
```diff
@@ -0,0 +1,377 @@
...
+
+	ret = IFCALLRESULT(CHANNEL_RC_OK, disp->SendMonitorLayout, disp, nmonitors, layouts);
+	free(layouts);
+	return ret;
+}
+
+BOOL wlf_disp_handle_configure(wlfDispContext* disp, int32_t width, int32_t height)
+{
+	if (!disp)
+		return FALSE;
```
hardening: duplicated check

akallabeth: exported in `wlf_disp.h`, no internal static

hardening: @akallabeth well I don't like the check (as it's not really exported as just used internally, it's present in the header file just for convenience). My point was more that you have exactly the same check 2 lines below.

akallabeth: @hardening ok, that duplication was not intentional

PR #5166, opened 2018-12-21
cmake: made FindGSSAPI.cmake able to detect kerberos versions such as…

informatimago:
Scanning for kerberos version numbers didn't allow dash-suffices such as in:
  "Kerberos 5 release 1.18-prerelease".
This is corrected.


akallabeth:
@freerdp-bot test

PR #5168, opened 2018-12-24
fix button_set_locked no return statement

ckelsel:


hardening:
@freerdp-bot can you test that please

PR #5173, opened 2018-12-29
enable fonts smoothing per default, part two

### LINKED ISSUE #5169: default value for fonts (enabled) does not work
With the current master (2.0.0-dev5 (5b24dc1ac)):
The changed default value for "fonts", introduced with #5133, does not work for me.
If I read the source code right, the defaults are hard coded in libfreerdp/core/settings.c. 
The change in cmdline.h seems not to be enough.
On the other hand, the "fonts" setting depends on the "network" setting. Is this still desired?

hendwolt:
This is a follow-up for #5133 and fixes #5169.


akallabeth:
@freerdp-bot test

PR #5175, opened 2019-01-02
fix(crypto/tls): typo

simon04:


hardening:
@freerdp-bot can you test that PR please ?

PR #5176, opened 2019-01-02
add ninja to gitignore

chipitsine:


hardening:
@freerdp-bot would you test that trivial PR for me please ?

Issue #3411, opened 2016-06-16
Connect via Connectionstring


vcordes79:
Hello,

I was wondering if it is possible to connect to a session created by the RDPSession class (https://msdn.microsoft.com/en-us/library/mt446078%28v=vs.85%29.aspx) via a connectionstring or if this could be implemented. That would be great.

Thanks,
Volker

aijyo:
I also wondring now....
So, anyone can answer this question?

Because of the rdpserver cannot run correctly on the windows(If anyone run the rdp server & client success.. tell me). I have create the rdpsever locally...by the  windows rdpsession. But I need connect the server on the linux platform(i.e phone , mac osx)...

fodinabor:
I thought so too.. but no, they are different..
As far as I understand it, there is only the part returned, that usually is stored in the LHTICKET (but decrypted), so that should equal to the Remote Assistance Connection String 2: https://msdn.microsoft.com/en-us/library/cc240132.aspx

Another observation I made: in my connection string a value `CE=` exists in the `<A` part. I haven't found any documentation on this one...

Also the curernt implementation of freerdp_assistance_parse_connection_string2 does not seem to be standard compliant, as the (optional) KH2 is not handled.

Here's a small forum thread mentioning some differences:
https://social.msdn.microsoft.com/Forums/windowsserver/en-US/281594b2-c664-4646-80e2-7715c4f1118e/windows-desktop-share-rdp-sessions?forum=os_windowsprotocols

fodinabor:
oups.. missed #5089

PR #5178, opened 2019-01-03
Support standard Remote Assistance Connection String 2

### LINKED ISSUE #5089: [assistance] unsupported options KH2 and others
Just as the title, I have try to shared my windows client as a server, and connnect and view it on some linux device, but when I run wfreerdp-client.exe and try to connect to the freerdp-shadow-cli.exe  something went wrong...(I'll show the error later, what is not my concern now).

I have already read a part of  "freerdp-shadow-cli" code , and found it implement with IRDPSession by convert IRDPSession connectstring to msrcIncident file.  

When I use wfreerdp-client.exe connect to a windows server with msrcIncident file, everything is okay.

When I use IRDPViewer connect to the freerdp-shadow-cli.exe server (use connectionstring that I hooked from the win_wds.c::int win_shadow_wds_init(winShadowSubsystem* subsystem)) everthing is okay.
image

**What I really concern  is : If it is possible to connect to the IRDPSession use freerdp-client  and if it is possible connect to the shared windows by connectionstring  created by IRDSession not only .msrcIncident  file or .rdp file...**

**So the problem that cannot be connected currently is simply because it has not been developed yet or because it cannot be supported at all...**

fodinabor:
As in #5089 and #3411 (and maybe others) mentioned, the WDS API gives a connection string, that should be used to initialize a connection. This is a Remote Assistance Connection String 2. This is also used in the assistance invitations, but in encrypted form and as part of the whole.

I just added a new case, if the check for an remote assistance invitation failed, it is now checked, whether the file looks like a RACS2. If so, the existing function `freerdp_assistance_parse_connection_string2` is used for parsing the file.

As the `Username` field is an extension by the invitation format, it is not necessary to establish the connection.

Further extensions should fully implement issues mentioned in #5089 including the KH2 check (just a _safer_ server-identity verification, than KH - both of these should maybe be used some day...)

## Test
Requires an application that opens a IRDPSRAPISharingSession (.NET: RDPSession), creates an invitation with a password and spits out the connection string from the invitation.
This connection string should be saved to `Test.msrcIncident`.

Now execute `wfreerdp.exe Test.msrcIncident /assistance:password`

### DIFF
--- a/libfreerdp/common/assistance.c
+++ b/libfreerdp/common/assistance.c
@@ -743,236 +743,256 @@ int freerdp_assistance_parse_file_buffer(rdpAssistanceFile* file, const char* bu
 	size_t length;
 	p = strstr(buffer, "UPLOADINFO");
 
-	if (!p)
-		return -1;
-
-	p = strstr(p + sizeof("UPLOADINFO") - 1, "TYPE=\"");
-
-	if (!p)
-		return -1;
-
-	p = strstr(buffer, "UPLOADDATA");
-
-	if (!p)
-		return -1;
-
-	/* Parse USERNAME */
-	p = strstr(buffer, "USERNAME=\"");
-
 	if (p)
 	{
-		p += sizeof("USERNAME=\"") - 1;
-		q = strchr(p, '"');
+		p = strstr(p + sizeof("UPLOADINFO") - 1, "TYPE=\"");
 
-		if (!q)
+		if (!p)
 			return -1;
 
-		length = q - p;
-		file->Username = (char*) malloc(length + 1);
+		p = strstr(buffer, "UPLOADDATA");
 
-		if (!file->Username)
+		if (!p)
 			return -1;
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test.

libfreerdp/common/assistance.c
```diff
@@ -1078,9 +1098,6 @@ BOOL freerdp_assistance_populate_settings_from_assistance_file(rdpAssistanceFile
 	if (freerdp_set_param_string(settings, FreeRDP_RemoteAssistancePassword, file->password) != 0)
 		return FALSE;
 
-	if (freerdp_set_param_string(settings, FreeRDP_Username, file->Username) != 0)
```
akallabeth: You'll break a lot of connection strings with this ;)

akallabeth:
@freerdp-bot test

PR #5180, opened 2019-01-04
resolve several issues found by coverity

chipitsine:


libfreerdp/core/gateway/rdg.c
```diff
@@ -601,10 +604,16 @@ static BOOL rdg_handle_ntlm_challenge(rdpNtlm* ntlm, HttpResponse* response)
 	}
 
 	if (!ntlm_authenticate(ntlm, &continueNeeded))
+	{
+		free(ntlmTokenData);
```
mfleisz: ntlm_client_set_input_buffer won't copy the content of ntlmTokenData but only save the pointer in the ntlm structure. Therefore if we free ntlmTokenData here the pointer in the ntlm struct will be left pointing to an already freed address causing a double free on cleanup.

chipitsine:
OK, I will remove this particular patch

akallabeth:
@freerdp-bot test

PR #5187, opened 2019-01-09
Mingw build fixes part1

astrand:
This pull request contains unobtrusive fixes for building on MinGW/GCC systems. This includes 
avoding mixed case includes and library names etc.

Issue #4930, opened 2018-10-15
[CrossCompiling] Fix build and update wiki for MinGW compilation


jhonesk:
I followed this instructions (https://github.com/FreeRDP/FreeRDP/wiki/Compilation) for cross compiling FreeRDP to be used on windows hosts. I'm running Arch Linux. 

I compiled OpenSSL and MingGw just as the wiki suggests, and on the final step, when I try to cmake, the compilation fails and I receive the following message:

ps: Yes, I made sure that all the paths are correct before issuing the command.
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
FREERDP_VERSION=2.0.0-dev4
-- Git Revision 00af869cd
-- GCC default symbol visibility: hidden
CMake Error at CMakeLists.txt:490 (message):
  Unknown generator Unix Makefiles

-- Configuring incomplete, errors occurred!
See also "/home/jcruz/FreeRDP/mingw_build/CMakeFiles/CMakeOutput.log".
See also "/home/jcruz/FreeRDP/mingw_build/CMakeFiles/CMakeError.log".
-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=

I get this from the FreeRDP/CMakeLists.txt -> Line 490;

****************************************************************************************
   if (${CMAKE_GENERATOR} MATCHES "NMake Makefile*" OR ${CMAKE_GENERATOR} MATCHES "Ninja*")
		set(CMAKE_PDB_BINARY_DIR ${CMAKE_BINARY_DIR})
	elseif (${CMAKE_GENERATOR} MATCHES "Visual Studio*")
		set(CMAKE_PDB_BINARY_DIR "${CMAKE_BINARY_DIR}/\${CMAKE_INSTALL_CONFIG_NAME}")
	else()
		message(FATAL_ERROR "Unknown generator ${CMAKE_GENERATOR}")
	endif()
********************************************************************************************

And this from the /home/jcruz/FreeRDP/mingw_build/CMakeFiles/CMakeError.log : 

********************************************************************************************
Building C object CMakeFiles/cmTC_01eb5.dir/CheckIncludeFile.c.obj
[...]

PR #5188, opened 2019-01-09
Mingw build fixes part2

astrand:
This PR builds upon #5187. With these commits, I can successfully build wfreerdp.exe in our Linux based MinGW environment, which is based on mingw-w64-v3.3.0 and gcc-5.1.0.


astrand:
See also #4930.

akallabeth:
@freerdp-bot test

libfreerdp/codec/rfx_rlgr.c
```diff
@@ -80,7 +80,9 @@ static INIT_ONCE rfx_rlgr_init_once = INIT_ONCE_STATIC_INIT;
 
 static BOOL CALLBACK rfx_rlgr_init(PINIT_ONCE once, PVOID param, PVOID *context)
 {
+#ifndef __MINGW32__
```
akallabeth: Why this? If the symbol is not native, it should be replaced by `WinPR` version of it.

libfreerdp/codec/rfx_rlgr.c
```diff
@@ -101,7 +103,9 @@ static INLINE UINT32 lzcnt_s(UINT32 x)
 		return n - x;
 	}
 
+#ifndef __MINGW32__
```
akallabeth: Why this? If the symbol is not native, it should be replaced by WinPR version of it.

winpr/libwinpr/sspi/NTLM/ntlm.h
```diff
@@ -79,13 +79,15 @@ typedef enum _NTLM_STATE NTLM_STATE;
 
 enum _NTLM_AV_ID
 {
+#ifndef __MINGW32__
```
akallabeth: Sure this breaks nothing? Will change the numerical id of all elements for such builds.

winpr/libwinpr/CMakeLists.txt
```diff
@@ -133,6 +133,8 @@ endif()
 
 add_definitions(${WINPR_DEFINITIONS})
 if (WIN32)
+        # With some versions of msvcrt, you need msvcr100 to get wmemmove_s
+        list(APPEND WINPR_LIBS "msvcr100")
```
akallabeth: How can this work if the runtime is not the one linked to? (See pull request tester failure)

mfleisz:
Why did you remove client/Windows/wfreerdp.rc (it's referenced in client/Windows/cli/CMakeLists.txt)?

akallabeth:
Is it correct that all commits of #5187 are contained in this pull?
Might close the former then (otherwise this one would need rebasing after merging it anyway)

astrand:
> Why did you remove client/Windows/wfreerdp.rc (it's referenced in client/Windows/cli/CMakeLists.txt)?

The intent was not to remove it, but correct the file type and encoding. Currently it is:

$ file wfreerdp.rc 
wfreerdp.rc: C source, Little-endian UTF-16 Unicode text, with CRLF, CR line terminators

The MinGW resource compiler cannot deal with that, so I changed it to a normal text file. However, apparently, I need to convince git that the file should not be considered binary. I also need to check the path separator; my mistake. I'll take a look at this.

astrand:
> Is it correct that all commits of #5187 are contained in this pull?
> Might close the former then (otherwise this one would need rebasing after merging it anyway)

That is correct. I wanted to separate the unintrusive parts, which can be easily accepted, and the more difficult ones. Thus, we can either merge #5187 and then rebase and work more on this one. Or, you can close #5187 in favour of this one.

(Probably, I won't have time to look at the other issues until next week.)

akallabeth:
@astrand I'd be careful with the resource files, the `UCS-2` encoding is because the `VisualStudio` resource compiler likes it that way ;)

akallabeth:
@astrand any update on this?

astrand:
> @astrand I'd be careful with the resource files, the `UCS-2` encoding is because the `VisualStudio` resource compiler likes it that way ;)

I've checked this now. The MinGW resource compiler (at least the version from mingw-w64-v3.3.0) does not support UCS-2 (or UTF-16). I know that VisualStudio prefers it, but I would be surprised if the MS "rc" does not work with ordinary text files. For example, the TurboVNC vncviewer.rc is a plain ASCII file, and can be built with "Microsoft Visual C++ 2005 or later".

I'm unsure about the paths, however. MinGW requires forward slashes. MSVC supports this in C/C++ files, but I'm not sure about resource files. Do not have access to a MSVC for testing now, unfortunately. If MSVC does not support forward slashes, one needs to move files around so that the .rc is in the same directory as the referenced files.

libfreerdp/codec/rfx_rlgr.c
astrand: Right, will fix.

akallabeth:
yes, you can use plain ascii files, but then you're limited to ascii... utf-8 does not work properly with older ms compilers.

winpr/libwinpr/CMakeLists.txt
astrand: > How can this work if the runtime is not the one linked to? (See pull request tester failure)

The Jenkins log is gone, will check again after new pull.

astrand:
What's up with Jenkins?

bmiklautz:
@freerdp-bot test

astrand:
I don't have much time to spend on this issue, and to be honest, I don't know what the "correct" solution would be to the wmemmove_s / msvcr100 problem. How about this solution?:

```
add_definitions(${WINPR_DEFINITIONS})
if (WIN32)
	if (NOT MSVC)
            # With MinGW, you need msvcr100 to get wmemmove_s
            list(APPEND WINPR_LIBS "msvcr100")
	endif()
	target_link_libraries(${MODULE_NAME} ${PUBLIC_KEYWORD} ${WINPR_LIBS})
else()
	target_link_libraries(${MODULE_NAME} ${PRIVATE_KEYWORD} ${WINPR_LIBS})
endif(WIN32)
```

akallabeth:
@astrand I´d check `WinPR/crt.h` and the define blocks there. Then no linking to some runtime is required in that case.

pfeatherstone:
Is there any way of resolving the conflicts ?

Neustradamus:
Have you progressed on it?

hardening:
AFAICT we've fixed all the issues with mingw (at least that works perfectly here), @astrand perhaps this PR can be closed ?

PR #5191, opened 2019-01-10
Implemented full mouse support.

akallabeth:
Add missing mouse buttons and wheel handling for mac client.

### DIFF
--- a/client/Mac/mf_client.m
+++ b/client/Mac/mf_client.m
@@ -143,6 +143,80 @@ void mf_scale_mouse_event(void* context, rdpInput* input, UINT16 flags,
 	}
 }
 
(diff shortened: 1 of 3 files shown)


client/Mac/mf_client.m
```diff
@@ -143,6 +143,80 @@ void mf_scale_mouse_event(void* context, rdpInput* input, UINT16 flags,
 	}
 }
 
+void mf_scale_mouse_event_ex(void* context, rdpInput* input, UINT16 flags,
```
mfleisz: The code in this function is a bit hard to understand ...
I.e. I don't really get why we only check if the height is equal [```if (wh != dh)```] but don't do the same check for the width?

akallabeth:
@mfleisz true, did not check that function. Unified the scaling calculations in its own function now and adjusting x and y equally.

PR #5194, opened 2019-01-14
Fixed #5185 broken file size initialization.

### LINKED ISSUE #5185: File corrupted when copied via clipboard and drive redirection server →client
**Description**
Hi, I can reproduce file corruption that happens when files are copied from an inside RDP session to the local client via clipboard copy + drive redirection. Only certain file sizes are affected (for sure 16KB and 32KB, always reproducible). Note that an isolated clipboard copy (solely inside a FreeRDP session) or an isolated copy via drive redirection is not affected (but not 100% about the latter).

**To Reproduce**
1. Use FreeRDP current master (aa442de05) or recent Remmina using freerdp-libs based on freerdp-2.0.0-47.rc4.fc29.src.rpm
2. Connect to a Windows RDP server (in my case: Server2016) while redirecting a local folder (`/drive:...`)
3. Inside the RDP session, connect to another Windows RDP server (in my case: Server2016, maybe it's sufficient to mstsc to the very same server to reproduce) using mstsc.exe with enabled clipboard (file) copy
4. Clipboard-copy a 16KB or 32KB file (with all NUL bytes, didn't check if file contents matter) from that 'inside' RDP session directly to the redirected drive on the 'outside' RDP session.
5. Observe that the copied file on your local FreeRDP client's local folder is 0 bytes.

**Expected behavior**
- Copy as-is :)

**Application details**
* Version of FreeRDP: 2.0.0-dev5 (aa442de05) / 2.0.0-47.rc4
* Command line used `xfreerdp /kbd:German +fonts +window-drag +decorations -themes -wallpaper +clipboard +auto-reconnect /disp /dynamic-resolution /drive:bla,/home/bla/blub +compression /sec:nla /bpp:16 /rfx /gfx +gfx-progressive -bitmap-cache +multitransport /u:bla /v:$rdp_host:$rdp_port`
* output of `/buildconfig`
* OS version connecting to: Server 2016
* If available the log output from a run with `/log-level:trace`: N/A (let me know if it was helpful)

**Desktop:**
- OS: Fedora 29 x64

akallabeth:


akallabeth:
@hardening sorry :)

PR #5196, opened 2019-01-16
fix [winpr/hash]: initialize ssl

bmiklautz:
winpr-crypt is used in winpr for hash generation but currently it's
still required to initialize openssl in the application itself.
winpr-hash didn't do that therefore the generated hashes were useless.


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #5198, opened 2019-01-17
[WIP] Smartcard cleanup and fixes

akallabeth:
Inspired by #5195 I've started cleaning up the code of the smartcard channel.


akallabeth:
@informatimago I've integrated your changes to smartcard status update and cleaned up the code a bit. Could you check, if this is working for you as well?

akallabeth:
@bmiklautz could you check if this now works with yubikey and certificates?

bmiklautz:
I haven't had time to go and test all changes in detail but the CARD_GET_STATUS_CHANGE_[AW] changes look a little odd. First they only initialize the first element in the list. And secondly they are received from the server and passed on to the PCSC backend. Another point is that the current state is also reset to 0.

PR #5201, opened 2019-01-18
Wayland refresh fix

akallabeth:
* Fixes an issue with screen no longer updating after focus loss
* Unifies screen invalidation and update
* Refreshes the whole screen on configure event.

### DIFF
--- a/client/Wayland/wlfreerdp.c
+++ b/client/Wayland/wlfreerdp.c
@@ -69,32 +69,23 @@ static BOOL wl_begin_paint(rdpContext* context)
 	return TRUE;
 }
 
-
-static BOOL wl_end_paint(rdpContext* context)
+static BOOL wl_update_buffer(wlfContext* context_w, UINT32 x, UINT32 y, UINT32 w, UINT32 h)
 {
 	rdpGdi* gdi;
 	char* data;
-	wlfContext* context_w;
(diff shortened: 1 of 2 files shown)


client/Wayland/wlfreerdp.c
```diff
@@ -214,6 +224,17 @@ static void wl_post_disconnect(freerdp* instance)
 		UwacDestroyWindow(&context->window);
 }
 
+static BOOL wl_refresh_display(wlfContext* context)
+{
+	rdpGdi* gdi;
+
+	if (!context || !context->context.gdi)
+		return FALSE;
+
```
hardening: unneeded checks

uwac/libuwac/uwac-window.c
```diff
@@ -641,6 +641,7 @@ UwacReturnCode UwacWindowSubmitBuffer(UwacWindow* window, bool copyContentForNex
 	if (window->pendingBuffer)
 	{
 		/* we already have a pending frame, don't do anything*/
+		UwacSubmitBufferPtr(window, window->pendingBuffer);
```
mfleisz: That comment above should be changed ...

akallabeth: done.

PR #5202, opened 2019-01-18
license: implement server-side management

hardening:
Add server-side management of the licensing workflow. The default behaviour is to accept the client, but if a server wants to implement full licensing support as in MS-RDPELE it is possible by defining a callback.

### DIFF
--- a/include/freerdp/license.h
+++ b/include/freerdp/license.h
@@ -1,0 +1,160 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ * Licensing API
+ *
+ * Copyright 2018 David Fort <contact@hardening-consulting.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef FREERDP_LICENSE_H
+#define FREERDP_LICENSE_H
+
+#include <freerdp/api.h>
+
+
+typedef struct rdp_license rdpLicense;
+
+/** @brief Licensing Packet Types */
+enum
+{
+	LICENSE_REQUEST		= 0x01,
+	PLATFORM_CHALLENGE 	= 0x02,
+	NEW_LICENSE 		= 0x03,
+	UPGRADE_LICENSE 	= 0x04,
+	LICENSE_INFO		= 0x12,
+	NEW_LICENSE_REQUEST	= 0x13,
+	PLATFORM_CHALLENGE_RESPONSE	= 0x15,
+	ERROR_ALERT			= 0xFF
+};
+
+#define LICENSE_PKT_CS_MASK	(LICENSE_INFO | NEW_LICENSE_REQUEST | PLATFORM_CHALLENGE_RESPONSE | ERROR_ALERT)
+#define LICENSE_PKT_SC_MASK	(LICENSE_REQUEST | PLATFORM_CHALLENGE | NEW_LICENSE | UPGRADE_LICENSE | ERROR_ALERT)
+#define LICENSE_PKT_MASK	(LICENSE_PKT_CS_MASK | LICENSE_PKT_SC_MASK)
+
+#define LICENSE_PREAMBLE_LENGTH			4
+
+/* Cryptographic Lengths */
+
+#define CLIENT_RANDOM_LENGTH			32
+#define SERVER_RANDOM_LENGTH			32
+#define MASTER_SECRET_LENGTH			48
+#define PREMASTER_SECRET_LENGTH			48
+#define SESSION_KEY_BLOB_LENGTH			48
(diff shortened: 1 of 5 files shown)


include/freerdp/peer.h
```diff
@@ -67,6 +67,16 @@ typedef int (*psPeerVirtualChannelWrite)(freerdp_peer* peer, HANDLE hChannel, BY
 typedef void* (*psPeerVirtualChannelGetData)(freerdp_peer* peer, HANDLE hChannel);
 typedef int (*psPeerVirtualChannelSetData)(freerdp_peer* peer, HANDLE hChannel, void* data);
 
+/** @brief the result of the license callback */
+typedef enum {
```
akallabeth: Formatting, braces

include/freerdp/license.h
```diff
@@ -0,0 +1,159 @@
...
+ * The most significant byte of the PlatformId field contains the operating system version of the client.
+ * The second most significant byte of the PlatformId field identifies the ISV that provided the client image.
+ * The remaining two bytes in the PlatformId field are used by the ISV to identify the build number of the operating system.
+ *
+ * 0x04010000:
+ *
+ * CLIENT_OS_ID_WINNT_POST_52	(0x04000000)
+ * CLIENT_IMAGE_ID_MICROSOFT	(0x00010000)
+ */
+enum {
```
akallabeth: Formatting, braces

libfreerdp/core/license.c
```diff
@@ -1551,8 +1556,9 @@ BOOL license_send_platform_challenge_response_packet(rdpLicense* license)
  * @param license license module
  */
 
-BOOL license_send_valid_client_error_packet(rdpLicense* license)
+BOOL license_send_valid_client_error_packet(rdpRdp* rdp)
```
akallabeth: Any specific reason to change the argument?

hardening: Yep, we don't have access to the `rdpLicence` in the `rdpRdp`.

akallabeth: `rdpLicense` is opaque there, correct. But you can still access all members of `rdpRdp` from where it is called.
Anyway, no blocker, just curious.

include/freerdp/peer.h
```diff
@@ -67,6 +67,17 @@ typedef int (*psPeerVirtualChannelWrite)(freerdp_peer* peer, HANDLE hChannel, BY
...
 typedef int (*psPeerVirtualChannelSetData)(freerdp_peer* peer, HANDLE hChannel, void* data);
 
+/** @brief the result of the license callback */
+typedef enum
+{
+	LICENSE_CB_INTERNAL_ERROR, 	/** an internal error happened in the callback */
+	LICENSE_CB_ABORT,			/** licensing process failed, abort the connection */
+	LICENSE_CB_IN_PROGRESS,		/** incoming packet has been treated, we're waiting for further packets to complete the workflow */
+	LICENSE_CB_COMPLETED		/** the licensing workflow has completed, go to next step */
+} LicenseCallbackResult;
```
akallabeth: Formatting

include/freerdp/peer.h
```diff
@@ -67,6 +67,17 @@ typedef int (*psPeerVirtualChannelWrite)(freerdp_peer* peer, HANDLE hChannel, BY
...
+
+typedef LicenseCallbackResult (*psPeerLicenseCallback)(freerdp_peer* peer, wStream *s);
```
akallabeth: pointer alignment

libfreerdp/core/peer.c
```diff
@@ -567,16 +567,26 @@ static int peer_recv_callback(rdpTransport* transport, wStream* s, void* extra)
 			break;
 
 		case CONNECTION_STATE_LICENSING:
-			if (!license_send_valid_client_error_packet(rdp->license))
-			{
-				WLog_ERR(TAG,
-				         "peer_recv_callback: CONNECTION_STATE_LICENSING - license_send_valid_client_error_packet() fail");
+		{
+			LicenseCallbackResult res = client->LicenseCallback(client, s);
```
akallabeth: Sure this callback is always set? The struct is exposed in `peer.h`, so a NULL check with an appropriate error message could be helpful.

hardening: make sense

include/freerdp/peer.h
hardening: What do you mean by `pointer alignment` ? ̀

mfleisz: I think he means ```type* var``` vs ```type *var```

libfreerdp/core/peer.c
```diff
@@ -567,16 +567,34 @@ static int peer_recv_callback(rdpTransport* transport, wStream* s, void* extra)
...
 		case CONNECTION_STATE_LICENSING:
-			if (!license_send_valid_client_error_packet(rdp->license))
+		{
+			LicenseCallbackResult res;
+			if (!client->LicenseCallback)
 			{
-				WLog_ERR(TAG,
-				         "peer_recv_callback: CONNECTION_STATE_LICENSING - license_send_valid_client_error_packet() fail");
+				WLog_ERR(TAG, "peer_recv_callback: LicenseCallback has been removed, assuming licensing is ok (please fix your app)");
+				res = LICENSE_CB_COMPLETED;
```
akallabeth: Shouldn't this be an error condition?

hardening: Well, it could but in practice we don't really care. To be in that case, we must have an app that would have replaced, on purpose, the callback by `NULL`. Looks very unlikely.

akallabeth: or like a beginners mistake ;)
anyway, not really that important.

PR #5209, opened 2019-01-22
Wayland protocol update and keyboard inhibit support

akallabeth:
* Updates `xdg-shell` to stable API
* Updates `fullscreen-shell` to `fullscreen-shell-unstable-v1` API
* Add `keyboard-shortcuts-inhibit-unstable-v1` support to inhibit local keyboard shortcuts (`-grab-keyboard` to disable) Currently no release key implemented, only mouse can release again.
* Add `xdg-decorations` support to enable server side decorations if available
* Add `kde-decorations` support to enable server side decorations if available (fallback option)

### DIFF
--- a/uwac/libuwac/uwac-window.c
+++ b/uwac/libuwac/uwac-window.c
@@ -384,6 +384,30 @@ UwacBuffer* UwacWindowFindFreeBuffer(UwacWindow* w)
 	return &w->buffers[i];
 }
 
+static UwacReturnCode UwacWindowSetDecorations(UwacWindow *w)
+{
+	if (!w || !w->display)
+		return UWAC_ERROR_INTERNAL;
+
+	if (w->display->deco_manager) {
+		w->deco = zxdg_decoration_manager_v1_get_toplevel_decoration(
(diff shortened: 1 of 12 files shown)


hardening:
Looks perfectly fine.

uwac/libuwac/uwac-window.c
```diff
@@ -155,10 +157,11 @@ static void xdg_handle_configure(void* data, struct xdg_surface* surface,
 	}
 
 ack:
-	xdg_surface_ack_configure(surface, serial);
+	return;
 }
```
hardening: lonely return ? (unneeded here)

hardening:
@akallabeth remove the erroneous return and I'll merge this PR

uwac/libuwac/uwac-window.c
akallabeth: @hardening necessary for the label to jump to, did not want to refactor the whole function (but then again it is not that large)

akallabeth: ok, done :)

Issue #5211, opened 2019-01-23
wlfreerdp segfaults 


hendwolt:
The Wayland client from the current master segfaults on Plasma5 on Wayland.
Even without any parameters wlfreerdp segfaults.

operating system: openSUSE Tumbleweed

freerdp version and config:
This is FreeRDP version 2.0.0-dev5 (281c85598)
[...]

PR #5213, opened 2019-01-24
Wayland and warning fixes

akallabeth:
* Removed frame done check in wayland client (it is unreliable and might only be used to reduce frame updates according to https://wayland.freedesktop.org/docs/html/apa.html `wl_surface::frame - request a frame throttling hint` description
* Removed buffer submit multi buffer handling in UWAC, use a single buffer for all updates (@hardening this is working on my test machines without glitches, as soon as I reenable the old behaviour I get missed screen updates on mass, could you confirm or point out any error?)
* Fixes a crash when clipboard is registered due to not all resources being initialized yet.
* Fixes some compiler and static analyzer warnings

### DIFF
--- a/channels/drive/client/drive_file.c
+++ b/channels/drive/client/drive_file.c
@@ -54,8 +54,7 @@
 static void drive_file_fix_path(WCHAR* path)
 {
 	size_t i;
-	size_t length;
-	length = (int) _wcslen(path);
+	size_t length = _wcslen(path);
 
 	for (i = 0; i < length; i++)
 	{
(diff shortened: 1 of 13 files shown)


akallabeth:
@hardening solves maximizing a window and get a proper screen if no dynamic channel is used for me. (then the `wl_buffer` is larger with different stride than the `gdi` buffer).
Same for when the window is smaller than the `gdi` buffer.

hendwolt:
I tested this PR and can confirm, that it fixes the crash, I reported in #5211.

PR #5216, opened 2019-01-25
change SetTimer elapse from 10 to USER_TIMER_MINIMUM 

ckelsel:
If uElapse is less than USER_TIMER_MINIMUM (0x0000000A), the timeout is set to USER_TIMER_MINIMUM. If uElapse is greater than USER_TIMER_MAXIMUM (0x7FFFFFFF), the timeout is set to USER_TIMER_MAXIMUM.


akallabeth:
@freerdp-bot test

PR #5219, opened 2019-01-25
Wayland smart sizing

akallabeth:

PR #5221, opened 2019-01-25
Added wayland mouse cursor 

akallabeth:
* Implements mouse cursor callbacks in wayland client
* Adds UWAC function to update mouse cursors

TODO:
* Add scaling support #5219 required


akallabeth:
@hardening yes, that area might need improvement.

hardening:
@akallabeth multiseat => yes makes sense, as it requires some changes and I don't really know how we can handle multiple pointers from different seats in FreeRDP...

PR #5223, opened 2019-01-28
Scanbuild null fixes

akallabeth:
* Fixes all `NULL` dereference issues found by clang scanbuild
* Fixes all uninitialized value issues found by clang scanbuild

### DIFF
--- a/client/X11/xf_event.c
+++ b/client/X11/xf_event.c
@@ -213,7 +213,7 @@ static BOOL xf_event_execute_action_script(xfContext* xfc, XEvent* event)
 	char buffer[1024] = { 0 };
 	char command[1024] = { 0 };
 
-	if (!xfc->actionScriptExists || !xfc->xevents)
+	if (!xfc->actionScriptExists || !xfc->xevents || !xfc->window)
 		return FALSE;
 
 	if (event->type > LASTEvent)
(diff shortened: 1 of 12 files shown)


client/X11/xf_event.c
```diff
@@ -477,6 +477,9 @@ static BOOL xf_event_KeyRelease(xfContext* xfc, XEvent* event, BOOL app)
 }
 static BOOL xf_event_FocusIn(xfContext* xfc, XEvent* event, BOOL app)
 {
+	if (!xfc || !event || !xfc->window)
+		return FALSE;
```
hardening: how the hell can we have `xfc == NULL` or `event == NULL`, if this happens I guess this is a problem with the caller (static function). Even the check on `xfc->window` is questionable.

hardening:
As usual I'm not sure it's a good idea to make static analyzers happy at all costs when they report false positives

akallabeth:
@hardening These issues all have proper stack traces by the analyzer, they are correct.
The 'false'  positives are the 4 memory leaks that result from the invalid HANDLE values that are not NULL

akallabeth:
@hardening removed the checks for `xfc` and `event` as requested.
The rest of the changes are quite valid (uninitialized memory, ...)

PR #5237, opened 2019-02-02
client/common/cmdline.c: remove redundant check

chipitsine:
[client/common/cmdline.c:390] -> [client/common/cmdline.c:399]: (warning) Identical inner 'if' condition is always true.


akallabeth:
@freerdp-bot test

PR #5244, opened 2019-02-10
Add OpenBSD to TRIO_PLATFORM_UNIX

ledeuns:
OpenBSD is family of NetBSD, FreeBSD & DragonFlyBSD so treat as the same.


akallabeth:
@freerdp-bot test

akallabeth:
@ledeuns no longer interested in getting this merged?

ledeuns:
@akallabeth after discussing with OpenBSD team, it might give the false impression that we don't an issue with the timer thing anymore :)

PR #5246, opened 2019-02-12
Resolve Gfx mode missing surface updates from cache.

wintersandroid:
In the android client library.
We found that while using Gfx mode, some screen updates (usually when not much is changing) were being missed, for example when viewing a clock running seconds.

I did some really deep digging and had to compare against the X11 version which the updates would show correctly. 
I found that some of the android_end_paint callbacks had no regions to update, and it appeared that the reason was that sometimes the android_begin_paint callback was returning false.

The windows and X11 clients don't implement the begin_paint callback, so in removing all code from the android_begin_paint and just returning TRUE, the android_end_paint would have regions to update and we would no-longer miss surface updates which were in the cache.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,4 @@ packaging/deb/freerdp-nightly/freerdp-nightly-dbg
 
 # VisualStudio Code
 .vscode
+cache/
--- a/client/Android/android_freerdp.c
+++ b/client/Android/android_freerdp.c
@@ -130,24 +130,6 @@ static void android_OnChannelDisconnectedEventHandler(
 
 static BOOL android_begin_paint(rdpContext* context)
 {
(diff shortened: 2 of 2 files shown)


akallabeth:
@freerdp-bot test

.gitignore
```diff
@@ -150,3 +150,4 @@ packaging/deb/freerdp-nightly/freerdp-nightly-dbg
 
 # VisualStudio Code
 .vscode
+cache/
```
akallabeth: please remove.

wintersandroid: This is the directory where the script/android-build-freerdp.sh puts the archives it downloads from github for re-use.

akallabeth: @wintersandroid ah, ok, point taken.

Issue #5250, opened 2019-02-12
freerdp-2.0.0_rc4: TestNTLM, TestFreeRDPCodecInterleaved fail on ppc64


ernsteiswuerfel:
Gentoo downstream bugreport (ppc): [https://bugs.gentoo.org/672744](url)

Re-tested on a Talos II (ppc64) with git-master. _cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=ON . && make -j34 && ctest -j 34 --test-load 32_:
```
[...]
99% tests passed, 2 tests failed out of 186

Total Test time (real) =  21.54 sec

The following tests FAILED:
	110 - TestNTLM (Failed)
	169 - TestFreeRDPCodecInterleaved (Failed)
Errors while running CTest
```
Output of the 2 failing tests:
```
 # ./Testing/TestFreeRDPCodec TestFreeRDPCodecInterleaved
 # ./Testing/TestSspi TestNTLM
NTLM_NEGOTIATE (length = 40):
[20:46:23:036] [29262:29262] [WARN][com.winpr.sspi] - AcceptSecurityContext status SEC_E_INTERNAL_ERROR [0x80090304]
AcceptSecurityContext status: SEC_E_INTERNAL_ERROR (0x80090304)
test_ntlm_server_authenticate failure
 # ./Testing/TestFreeRDPCodec TestFreeRDPCodecInterleaved
 # 
```

akallabeth:
@ernsteiswuerfel big endian I assume? we lack proper hardware to test that and qemu is quite slow for that kind of emulation. (still surprised that only that few tests fail ;))

akallabeth:
@ernsteiswuerfel for the `TestFreeRDPCodecInterleaved` issue the culprit is buffer to `UINT16` and `UINT32` casts.
If you have the opportunity, could you fix the `GETPIXEL16`, `GETPIXEL32`  macros for your endianess and check?

ernsteiswuerfel:
Yes, Big Endian. ;)

So I replaced the funtions in _libfreerdp/codec/bitmap.c_. I think you forgot some "(" here, or else it won't build:
return (((UINT32)((const BYTE*)d)[0]) << 24) | **"("**((UINT32)((const BYTE*)d)[1]) << 16) | **"("**((UINT32)((const BYTE*)d)[2]) << 8) | ((const BYTE*)d)[3];

But the tests still fail the same way:
```
[...]

PR #5252, opened 2019-02-13
Fixed endianess issue with GETPIXEL16 and GETPIXEL32

akallabeth:
The buffer read functions did not work on big endian systems.
This tries to fix that.


akallabeth:
Fixes part of #5250, confirmed by initial reporter

PR #5263, opened 2019-02-20
libfreerdp/codec/progressive.c: remove useless function call

chipitsine:
function progressive_context_reset() is only useful for its return value, which is ignored


akallabeth:
@chipitsine the reset function is part of the API and a placeholder for a later point (complete implementation of all parts of the codec, same as `progressive_compress`.

PR #5267, opened 2019-02-20
Resource management redesign

sss123next:
portion of changes to make libfreerdp and channels addins more friendly to external resource management (external thread pools, event libraries)

this changes introducing DisableThreads option in rdpSetting structure.

currently this setting turning off threads in rfx and yuv codecs.

also implemented DisableThreads option support for cliprdr channel addin for client (currently used in our project) which demonstrates how to use this new option.


libfreerdp/gdi/gfx.c
```diff
@@ -85,11 +85,25 @@ static UINT gdi_ResetGraphics(RdpgfxClientContext* context,
 	}
 
 	free(pSurfaceIds);
+	{
+		UINT32 flags = FREERDP_CODEC_ALL;
 
-	if (!freerdp_client_codecs_reset(gdi->context->codecs, FREERDP_CODEC_ALL,
-	                                 gdi->width, gdi->height))
-		goto fail;
+		if (settings->RemoteFxCodec == FALSE)
```
akallabeth: What is this for? No real reason to do that magic

libfreerdp/core/connection.c
```diff
@@ -184,8 +185,19 @@ static BOOL rdp_client_reset_codecs(rdpContext* context)
 	if (!context->codecs)
 		return FALSE;
 
-	return freerdp_client_codecs_prepare(context->codecs, FREERDP_CODEC_ALL,
-	                                     settings->DesktopWidth, settings->DesktopHeight);
+	if (settings->RemoteFxCodec == FALSE)
```
akallabeth: What is this for?

akallabeth:
@freerdp-bot test

libfreerdp/core/connection.c
sss123next: to avoid initialization of codecs->rfx, codecs->nsc, if it disabled in settings, at best it does memory allocations, at worst, creating threads, mutexes, i think it does not make sense if it is unused at all and also harmful for resources

libfreerdp/gdi/gfx.c
sss123next: you are right, this flags does not make much sense on reset as it check for null ptr of each codec context,  looks like i am i bit overdone here.

sss123next: ah, it's just typo, i missed  what flags var not used at all...

sss123next: sorry for delay, i was very busy during last week,  typo fixed now, changes rebased against latest master.

akallabeth:
@sss123next I´ve had a look at the changes you´re doing and one thing I did recognize:
The channel structure is usually: A thread handling events in a queue
It might be a better approach to fix this incrementally.
* Would suggest first creating a pull unifying the channel threads (provide a common push message and init message handler API that can either use a thread or handle the message directly, depending on configuration)
* It would be better to wait with these changes until https://github.com/FreeRDP/FreeRDP/milestone/12 is complete. we try not to break the API for 2.0 any more (not too successfully, but we try :))

sss123next:
>     * Would suggest first creating a pull unifying the channel threads (provide a common push message and init message handler API that can either use a thread or handle the message directly, depending on configuration)
 
can you describe more precisely  what exactly you want to see please ?

as i understand currently all threading implemented in libwinpr.

so, should i extend libwinpr with new apis ?, or i need to do it inside libfreerdp using existing apis of libwinpr and do threading magic based on settings here ?

akallabeth:
@sss123next That was for the channels in the `channels`:
They have a thread that blocks (`WaitForMultipleObjects`) and waits for the designated messge queue to have something added.
Then it processes that and returns to step 1

This boiler plate code consists roughly of:
* CreateThread
* CreateMessageQueue
* Thread Handler with:
  * WaitForMultipleObjects
  * MessageQueue_Take
  * process

For that there could be a generic:
* channel_message_handler_init(thecallback_for_processing_a_single_message)
* channel_message_handler_add_message()
* channel_message_hanlder_uninit()

in `channels/client` which can use threads and message queues or just skip that altogether

akallabeth:
@sss123next Note: This is only valid for the `channels/<xxx>/client` stuff, not for the server side implementations

sss123next:
understood,
i will see what i can do.
to be honest, changes i already made is enough for our task at current moment of time, but our team is very interested to have all required changes to be available in upstream, so i guess i have little choice.

sss123next:
1. if i understand correctly, "CHANNEL_ENTRY_POINTS_FREERDP_EX" is a place to add this new callbacks, maybe something like "pVirtualChannelSetDataHandlers" or "pVirtualChannelSetMessageHandlers" ?

2. if i understand correctly i also need to move out threading and raw data handling logic from channels to  somewhere in "libfreerdp/core/client.c" ?

sss123next:
also i think "pVirtualChannelWriteEx" can be replaced by this high level api

akallabeth:
@sss123next Sorry, missed the last comment.
* The channel entry API is crude, but sadly currently used by other projects, so we can't easily change it.
Current way to pass stuff in there is by `rdpContext` (mostly `rdpSettings`)

* The common client threading function should just replace the channel code that does
1. Create a message queue and a thread
1. Wait in the thread function for new messages
1. Pass messages from transport/dynamic channel to own thread implementation
1. Tear down thread and message queue

With minor modifications the current thread functions can be reduced to just process a new message and the thread and message queue handling to a common core.

These would not be function pointers but just some common functions like:
1. `channel_client_create_handler(ctx, void (*fkt)(wStream* stream))`: Create message queue and thread or not, depending on settings
1. `channel_client_post_message(ctx, wStream* stream)`: post a message in the queue or directly call the processing handler
1. `channel_client_quit_handler(ctx)` Tear down queue and thread

Not really sure about which exact type the arguments should be, just to demonstrate the rough direction

sss123next:
ok, i can't change old api, but i can extend structure with new apis ?

this should keep binary compatibility for old external addins...

akallabeth:
@sss123next for your purposes it should be sufficient to create a new setting (threading or whatever it will be called).
That can be checked and acted upon accordingly.
Or you can do it compile time, whichever you prefer.

akallabeth:
@sss123next and asking questions is always easier than doing rework, agree there ^^

sss123next:
ok, understood, so just add three new functions for client which will handle threading if needed.

akallabeth:
@sss123next Might be a good idea to use a bitmask as a setting, that way it can be used for more fine grained control later on (external thread/limit number/...)

akallabeth:
@sss123next yes, just these 3 and all threading related settings are contained in there.

sss123next:
ok, i will change it to uint32 and add some flags, also i think setting name should be renamed to more appropriate name in this case.

sss123next:
draft implementation of this new api is done and used for cliprdr channel

but i am not sure if i have placed it in correct place...

channels/client/addin.h
```diff
@@ -18,3 +18,18 @@
...
+//#include <freerdp/freerdp.h>
+
+void
+channel_client_create_handler(rdpContext *ctx, HANDLE thread, LPVOID userdata,
+	UINT (*msg_handler)(LPVOID userdata,  wStream* data), wMessageQueue* queue);
+
+UINT
+channel_client_post_message(rdpContext *ctx, LPVOID pData, UINT32 dataLength,
+	UINT32 totalLength, UINT32 dataFlags, wStream* channel_data_in,
+	wMessageQueue* queue, UINT (*msg_handler)(LPVOID userdata,  wStream* data),
```
akallabeth: Why do you add the function pointer for each post? Should be sufficient to add it during `channel_client_create_handler` or do I miss something?

channels/client/addin.c
```diff
@@ -390,3 +390,186 @@ PVIRTUALCHANNELENTRY freerdp_channels_load_static_addin_entry(LPCSTR pszName, LP
...
+			data = (wStream*) message.wParam;
+
+			if ((error = args->msg_handler(args->userdata, data)))
+			{
+				WLog_ERR(TAG, "msg_handler failed with error %"PRIu32"!", error);
+				break;
+			}
+		}
+	}
+	ExitThread(error);
```
akallabeth: Missing the `channelError` checks

channels/client/addin.c
```diff
@@ -390,3 +390,186 @@ PVIRTUALCHANNELENTRY freerdp_channels_load_static_addin_entry(LPCSTR pszName, LP
...
+		{
+			WLog_ERR(TAG, "cliprdr_plugin_process_received: read error");
+			return ERROR_INTERNAL_ERROR;
+		}
+
+		channel_data_in = NULL;
+		Stream_SealLength(data_in);
+		Stream_SetPosition(data_in, 0);
+
+		if (!MessageQueue_Post(queue, NULL, 0, (void*) data_in, NULL))
```
akallabeth: Just call the handler directly if no threads are used? No queue necessary in that case...

sss123next:
done.

channels/client/addin.h
sss123next: it was just draft, to confirm what we understand each other, look at latest changes.

bmiklautz:
@freerdp-bot test.

sss123next:
check it.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@sss123next looks like you missed a few calls to refactor.

sss123next:
check it.

kubistika:
@freerdp-bot test

libfreerdp/codec/test/TestFreeRDPCodecRemoteFX.c
```diff
@@ -851,7 +851,9 @@ int TestFreeRDPCodecRemoteFX(int argc, char* argv[])
 	BYTE* dest = NULL;
 	size_t stride = FORMAT_SIZE * IMG_WIDTH;
 
-	context = rfx_context_new(FALSE);
+    /* use default threading options here, pass zero as
```
akallabeth, line 854: formatting, run `clang-format` please

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

sss123next:
rebased on master, squashed

akallabeth:
@freerdp-bot test

akallabeth:
@sss123next back from holidays, so let me catch up before I´ll check this.

channels/drdynvc/client/drdynvc_main.c
```diff
@@ -1378,8 +1383,9 @@ static DWORD WINAPI drdynvc_virtual_channel_client_thread(LPVOID arg)
 		if (message.id == 0)
 		{
 			data = (wStream*)message.wParam;
+			UINT32 ThreadingFlags = TRUE;
```
akallabeth, line 1386: Breaks older compilers, move to beginning of scope

libfreerdp/core/connection.c
akallabeth: Ok, then it would be best to create a function `DWORD freerdp_settings_codec_mask(const rdpSettings* settings)` or similar to get the mask and use it instead of duplicating the whole code.

libfreerdp/gdi/gfx.c
akallabeth: See below in how to unify this

akallabeth:
@freerdp-bot test

akallabeth:
@sss123next `clang-format` please ;)

akallabeth:
@freerdp-bot test

channels/client/addin.h
```diff
@@ -16,3 +16,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/* we need rdpContext */
+//#include <freerdp/freerdp.h>
```
akallabeth, line 20: dead code?

channels/client/addin.h
```diff
@@ -16,3 +16,14 @@
...
+
+void* channel_client_create_handler(rdpContext* ctx, LPVOID userdata,
```
akallabeth, line 22: these should most likely have `FREERDP_API` prefixed as channels might be compiled as separate libraries

channels/client/addin.h
```diff
@@ -16,3 +16,14 @@
...
+                                    UINT (*msg_handler)(LPVOID userdata, wStream* data),
+                                    const char* channel_name);
+
+UINT channel_client_post_message(void* MsgsHandle, LPVOID pData, UINT32 dataLength,
```
akallabeth, line 26: Better use a specific typedef for the `MsgHandle`, so that the compiler can warn you when you pass in a wrong type (even if it is opaque)

server/shadow/shadow_encoder.c
```diff
@@ -131,8 +131,9 @@ static int shadow_encoder_uninit_grid(rdpShadowEncoder* encoder)
 
 static int shadow_encoder_init_rfx(rdpShadowEncoder* encoder)
 {
+	/* TODO: pass ThreadingFlags somehow */
```
akallabeth, line 134: Maybe settings?

server/Windows/wf_update.c
```diff
@@ -187,7 +187,8 @@ void wf_update_encoder_reset(wfInfo* wfi)
 		}
 		else
 		{
-			wfi->rfx_context = rfx_context_new(TRUE);
+			/* TODO: pass ThreadingFlags somehow */
```
akallabeth, line 190: maybe settings?

libfreerdp/common/settings.c
```diff
@@ -925,3 +925,18 @@ BOOL freerdp_settings_set_value_for_name(rdpSettings* settings, const char* name
 	}
 	return FALSE;
 }
+
+UINT32 freerdp_settings_get_codecs_flags(rdpSettings* settings)
```
akallabeth, line 929: `const rdpSettings*` does not hurt

include/freerdp/settings.h
```diff
@@ -1559,6 +1562,9 @@ struct rdp_settings
 	                                   default value - currently UNUSED! */
 	ALIGN64 char* ActionScript;
 	ALIGN64 DWORD Floatbar;
+
+	/* resource management related options */
+	ALIGN64 UINT32 ThreadingFlags;
```
akallabeth, line 1567: Better move to API stable zone with a specific position. Run the `update-*` scripts under `scripts` to generate getter/setter

akallabeth:
Looks like you used `snprintf`? microsoft is a bit special, use `_snprintf` instead, then it should work.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth (dismissed):
Review LGTM ,
still need to test functionality.

sss123next:
let's finish with this PR, i am a bit tired of resolving conflicts every half year....

akallabeth:
@freerdp-bot test

server/shadow/shadow_encoder.c
sss123next: explain please.

server/Windows/wf_update.c
sss123next: explain please.

akallabeth:
@sss123next will have a look at this this week, but currently otherwise occupied.

akallabeth:
@freerdp-bot test

sss123next:
@akallabeth good, note what this is only client part, and only for few channels new api is used, here is already request for this optimization for server side channels architecture and i will need it anyway as i need to disable threading in my project, but i think it will be better to finish with client-side part first.

akallabeth:
@freerdp-bot test

server/shadow/shadow_encoder.c
akallabeth: `rdpShadowEncoder` contains `rdpSettings` for client and server instances.
You can add a setting and check that here if required.

sss123next: understood, will do

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth (commented):
Changes compile now and seem to be in a state that can be considered to be merged.
Please properly rebase this to current master (there are many changes that are note related to the pr and seem to be from merges) and best squash the commits

sss123next:
done

akallabeth:
@sss123next there are still many changes in files you did not touch, please remove those. (They might result from running `clang-format` but you can ignore all changes in files you did not touch)

akallabeth:
@freerdp-bot test

akallabeth (approved):

Issue #5268, opened 2019-02-21
Embedded window using ParentWindowId setting does not show FreeRDP client


anbukart:
I am trying to embed FreeRDP windows client inside of a panel in my windows forms application using C#. The rdp context is created fine, but when I start the client using client_start function call, it is not showing up inside the parent window. Without the embed option, it works. 
I have a fix for embed option in my local.

**This is what I have done so far (Steps to reproduce):**
1. Downloaded rc4 source
2. Compiled FreeRDP source using CMake and built using visual studio
3. I had to change the "client\Windows\wfreerdp-client.vcxproj" to create a dll instead of exe since I want to consume it's functions in my windows application
4. I call these functions of the wfreerdp-client.dll -> "freerdp_client_context_new" function to get context and then "freerdp_client_start" function with rdp settings like Desktop_Height, Desktop_Width, ParentWindowId, Hostname set
5. The FreeRDP windows client is not visible since client_width and client_height is 0 when I debug the code in "wf_client_thread" function in "wf_client.c" file

**Fix proposed**
After making this change here in this commit, it seems to work - https://github.com/anbukart/FreeRDP/commit/b1921714349affd04d4a1bcc61ce24d97473f033

I have this fix I am using. Can you please check.

akallabeth:
Hi @anbukart your fix looks good. Care to create a pull request for that?

anbukart:
Hi, @akallabeth I have created a pull request here-> https://github.com/FreeRDP/FreeRDP/pull/5269
Thanks :)

akallabeth:
actually thank you ;)

akallabeth:
oh, and a little trick, `-DWITH_CLIENT_INTERFACE=ON` for the cmake run should automatically create a library for you

anbukart:
Oh ok. Cool. Thanks

PR #5269, opened 2019-02-21
Embedded window client

anbukart:
Client width and height getting set as 0 when embedded using ParentWindowId setting
This is where the issue is explained in detail -> https://github.com/FreeRDP/FreeRDP/issues/5268


akallabeth:
@freerdp-bot test

PR #5271, opened 2019-02-21
Create event exa scanbuild fix

akallabeth:


akallabeth:
@hardening No other way (the sanitizer is actually correct about the possible leak) but the handles are specified that way...

Issue #5092, opened 2018-12-03
Not releasing modifiers keys on focus gain?


Arokha:
To continue the long trend of these issues (#2104, #4662, #1335, #1010, #540), here this is again. Appears to be a regression fairly recently.

**To Reproduce**
1. Give freerdp window focus
2. ~~Use any keyboard combinations with ALT that cause the window to lose focus (alt-tab, workspace switching combos, etc)~~ See first comment.
3. Return to freerdp window, find that ALT is continually held until both (always both) alt keys on keyboard are tapped (remote OS does not see this, only local)

**Expected behavior**
I have -grab-keyboard, and freerdp should drop keys when losing focus. There were various patches to fix this over the years in the above-mentioned issues, but it seems to have been lost.

* Ver: 2.0.0-rc4
* Distro: Manjaro (so, Arch)
* Windows 10 remote host (not sure that's relevant in this case though)

**Additional context**
This did not always happen, something recent changed this, though I don't know what. I looked through recent PRs and don't see anything that might.

Arokha:
Actually, after a little experimentation, this appears to be the opposite of the above issues?

Losing focus seems fine. It's GAINING focus with a modifier key pressed that pins it.

So for e.g.:
xfreerdp open on workspace 1
Switch to workspace 2 with Ctrl-Alt-2
Return to workspace 1 with something other than kb shortcuts = works fine

xfreerdp open on workspace 1, currently viewing workspace 2
Switch to workspace 1 with Ctrl-Alt-2
Alt pinned in xfreerdp until both alt keys pressed

akallabeth:
Duplicate of #4966 
Looks like this might be something of a window manager / X11 event combination but so far I could never reproduce. (all keys are released on focus out and the current state is read on focus in)
If you have additional information please attach that to the issue referenced.

yeroc:
This ticket and the subsequent analysis by @Arokha perfectly describe the symptoms I'm experiencing on AlmaLinux 8 with Gnome.  I understand this may well be a duplicate of #4966 but quite frankly this ticket describes the issue much more succinctly.  If I can provide more info to be of help let me know.

Issue #4966, opened 2018-10-24
Keys being "stuck" when switching desktops


StewPedassle:
Hi All,

I'm rather inexperienced when it comes to reporting/tracing bugs, so please bear with me (feel free to treat me like a toddler when it comes to making/fetching logs etc.)

I've found a bug or two on my machine.  I am submitting these both under one because they appear to be intertwined behaviors.  

First unexpected behavior: I can switch workspaces while xfreerdp is active using only "Lctrl+alt+[U/D]" (i.e., I don't have to hit my escape key -- Rctrl).  Of course, the expected behavior is for xfreerdp/windows to do nothing with this combination unless I hit Rctrl before. 

Second unexpected behavior: switching workspaces back to xfreerdp's workspace with "Lctrl+alt+[U/D]" will cause xfreerdp to capture certain keys.  E.g., "c" is non-functional and "enter" alone will toggle fullscreen of xfreerdp until xfreerdp receives either a Rctrl or an enter.  Again, expected behavior is obvious.

** Additional Context ** 
- This behavior does not occur when switching back to the workspace using other means to navigate (e.g., Super+home or Super + mouse to select workspace)
- After switching, hitting "enter" alone will behave like "ctrl+alt+enter" (toggle fullscreen).  This will happen only once.  It will not happen if I have hit enter or Rctrl after switching.
- After switching, I will lose all use of the "c" key without losing any other character keys.  This is remedied once I hit Rctrl or enter.
- The on-screen keyboard on the remote machine does not indicate anything is being activated until I actually press keys (except "c" won't show actuation, of course).
- I don't know exactly what update caused this to happen.  I have been experiencing issues with copy/paste for a month or more (fixed with reboot of remote machine), and our IT service may or may not have updated our server recently (xfreerdp started displaying the computer name in a tab at the top without any changes to my login file)

** Application details **
* OS: Ubuntu 18.04.1 LTS
* Version of FreeRDP:  2.0.0-dev4 (d3c1db174)
* Command line used:  GNU bash, version 4.4.19(1)-release (x86_64-pc-linux-gnu)
* OS version connecting to: Windows 7
* output of `/buildconfig`:
[...]

PR #5280, opened 2019-02-24
potential fix for ENTER (modifiers) stuck after CTRL-ALT-ENTER (toggle fullscreen), clearing the buffer in xk_keyboard_update_modifier_keys.

DraganNSDK:
A possible fix for issues FreeRDP/FreeRDP#4966 FreeRDP/FreeRDP#5092 .

Tested heavily on the Linux/Ubuntu side (mostly LinuxMint) as a HOST and Win10 as remote. 
This fix works on my end and is solving the issue I was having.

Issue: 
On fullscreen toggle (CTRL-ALT-ENTER) any subsequent ENTER (w/o any modifiers pressed) toggles the screen - i.e. user is unable to edit anything remotely.

Solution:
Problem seems to be the modifiers buffer that for some reason doesn't refresh, while xk_keyboard_update_modifier_keys() is called on focus successfully it doesn't seem to do anything (in some cases at least). Forcing the xk_keyboard_update_modifier_keys doesn't help either, but clearing the buffer does the trick. That solution is adapted from a long history of this issue.
My first solution was to put the modifiers refresh on every press call (which should be ok enough), but that alone didn't fix it - forcing the 'xf_keyboard_clear(xfc)' on refresh is what actually solves it.

Fix is a one-liner and it should be quite safe to introduce into the code w/o major testing - as it 'just' adds extra initialization step (w/o any hit on the performance or similar).

### DIFF
--- a/client/X11/xf_keyboard.c
+++ b/client/X11/xf_keyboard.c
@@ -326,7 +326,6 @@ static void xk_keyboard_update_modifier_keys(xfContext* xfc)
 	                 XK_Control_L, XK_Control_R, XK_Super_L, XK_Super_R
 	                };
 
-	// CTRL-ALT-ENTER FIX: we need to clear the buffer here beforehand
 	xf_keyboard_clear(xfc);
 
 	state = xf_keyboard_read_keyboard_state(xfc);
--- a/client/X11/xf_keyboard.c
+++ b/client/X11/xf_keyboard.c
@@ -325,6 +325,10 @@ static void xk_keyboard_update_modifier_keys(xfContext* xfc)
 	int keysyms[] = {XK_Shift_L, XK_Shift_R, XK_Alt_L, XK_Alt_R,
 	                 XK_Control_L, XK_Control_R, XK_Super_L, XK_Super_R
 	                };
+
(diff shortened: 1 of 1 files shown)


client/X11/xf_keyboard.c
```diff
@@ -325,6 +325,10 @@ static void xk_keyboard_update_modifier_keys(xfContext* xfc)
 	int keysyms[] = {XK_Shift_L, XK_Shift_R, XK_Alt_L, XK_Alt_R,
 	                 XK_Control_L, XK_Control_R, XK_Super_L, XK_Super_R
 	                };
+
+	// CTRL-ALT-ENTER FIX: we need to clear the buffer here beforehand
```
akallabeth: Please no C++ comments

akallabeth:
@freerdp-bot test

client/X11/xf_keyboard.c
DraganNSDK: no problems, removed the comment on my end, is that all from me?

akallabeth:
@freerdp-bot test
@DraganNSDK thank you.

DraganNSDK:
@akallabeth thank you for the FreeRDP, 
glad if I could be of some help, 
Best

PR #5285, opened 2019-02-26
Channel, log debug and semaphore related fixes

akallabeth:


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test, come on

winpr/libwinpr/utils/print.c
```diff
@@ -33,23 +33,35 @@
...
 	{
-		WLog_ERR(tag, "malloc(%"PRIuz") failed with [%d] %s", blen, errno, strerror(errno));
+		WLog_Print(log, WLOG_ERROR, "malloc(%"PRIuz") failed with [%"PRIuz"] %s", blen, errno,
+		           strerror(errno));
 		return;
 	}
 
 	while (offset < length)
 	{
-		pos += trio_snprintf(&buffer[pos], blen - pos, "%04x ", offset);
```
mfleisz: This line looks fishy ... ```%04zu``` won't print a hex-value to the string ... this will screw the heap

akallabeth: so you think the `trio_snprintf` has some bounds check issues?

akallabeth: ok, seems you were correct nice catch!

mfleisz: No - what I mean is that the format specifier is incorrect. We are not printing hex digits anymore (missing X) and we might write more than 4 characters to the string i.e.

```
size_t x = 0xFFFF;
printf("%04u\n", x);
printf("%04x\n", x);
```
output:
```
65535
ffff
```

mfleisz: And that's what screwed the stack I think ...

akallabeth: @mfleisz No, that is on purpose. (4 digits line count, if you have more this function is really awfully slow)
Problem was that if for some reason there is an error condition `-1` is returned by `trio_snprintf` and that translates to `0xFFFFFFFF` unsigned...

mfleisz: @akallabeth Yes but %04 only guarantees that it's at least 4 digits in size - not at max! And since this changed from hex reprenstation (x) to decimal now we will reach the case where it is > 4 (and as a result of that trio_snprintf to fail) way sooner than before. Obviously this happened during the NTLM test case ...

akallabeth: @mfleisz it only reserves space, that is correct. I still don´t see any instance where there is a broken bounds check though, the arguments for the follow up messages are calculated on used space and `snprintf` should truncate if there is not enough space left...

mfleisz: The stack issue was fixed with your latest commit (checking snprintf's return value before adding it to pos). However now we silently fail and don't get a hex dump i.e. in the case of the NTLM test (due to the format specifier change from %04x to %04u which causes snprintf to fail).

IMO we should check the input length and if it is > MAX_UINT16 then we should print an error (buffer to large to dump) and exit. Otherwise we can continue and leave the format of the offset as hex (which I'd prefer) - or if you really want to have the offset in decimal increase the buffer so that it can hold numbers with up to 5 digits instead of 4.

akallabeth: @mfleisz That is a bug in `trio_snprintf`. Need to check where exactly that occurs.
The format string is `[1234] 01 02 03 04   abcd` which translates to:
`[` + `<max 20 char linenumber>` + `] ` +  data as hex `02 ` + <3 spaces>` + <buffer line as ASCII>` + zero terminator
Lines are limited to 16 characters per line, that is `91` characters per line maximum!
Since the buffer is large enough it means that `trio_snprintf` does something wrong as `%04z` only means that it will reserve at least 4 characters for the line number, but use optionally more if required.
As we get some `-1 ` return it means that is broken...
see `http://pubs.opengroup.org/onlinepubs/7908799/xsh/fprintf.html`

akallabeth: @mfleisz ok, checked with a dummy buffer of size `0xffffff`  and it works flawlessly now. Only way to get a crash is to dump beyond the buffer boundaries...

PR #5301, opened 2019-03-13
Changed caps lock into control

cizra:
It's not like anybody uses it anyway, and there's plenty of feature requests asking for this.


lhzw:
Thanks for the hint, I will use this (caps <--> ctrl) to compile one for myself.

PR #5303, opened 2019-03-14
Added optional PreFrameAck callback to RdpgfxClientContext 

m4ntis:
## Motivation
When implementing an RDP proxy I came by a potential sync issue between the proxy's client and target server. My implementation of the GFX proxy registers all `RdpgfxClientContext` callbacks from the target server and sends the PDUs using an `RdpgfxServerContext` serving the proxy's client. Ideally, I could register the same `RdpgfxServerContext`'s FrameAck callback and send it back using the `RdpgfxClientContext`, but `RdpgfxClientContext` doesn't expose a method for sending the ack. Instead, `RdpgfxClientContext`'s implementation sends the ack automatically when an EndFrame PDU is received.

This behaviour could prove to be an issue in the context of a proxy, where there can be small latency between the proxy and target server, maybe being together in the same LAN, and big latency towards the client, perhaps connecting via the internet.

## My solution
After discussion on the IRC, I added an optional callback called just before the FrameAck is sent. This lets you to optionally register the callback and block for however long you need, in my case waiting for an ack from the proxy's client, while maintaining the default behavior where you don't need to concern yourself with acking.

## Testing
Testing the implementation is pretty simple, you could register a method in one of the RDP clients that sleeps for a second, connect with GFX and watch the frame updating once a second. 

Example test callback:
```c
#include <unistd.h>

BOOL test_PreClientAck(RdpgfxClientContext* context, RDPGFX_FRAME_ACKNOWLEDGE_PDU* ack)
{
	sleep(1);
	return TRUE;
}
```

### DIFF
--- a/channels/rdpgfx/client/rdpgfx_main.c
+++ b/channels/rdpgfx/client/rdpgfx_main.c
@@ -643,7 +643,7 @@ static UINT rdpgfx_recv_end_frame_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
 			ack.queueDepth = QUEUE_DEPTH_UNAVAILABLE;
 
 			if ((error = rdpgfx_send_frame_acknowledge_pdu(callback, &ack)))
-				WLog_Print(gfx->log, WLOG_DEBUG, "rdpgfx_send_frame_acknowledge_pdu failed with error %"PRIu32"",
+				WLog_Print(gfx->log, WLOG_ERROR, "rdpgfx_send_frame_acknowledge_pdu failed with error %"PRIu32"",
 				           error);
 		}
 	}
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -625,23 +625,27 @@ static UINT rdpgfx_recv_end_frame_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
 	gfx->TotalDecodedFrames++;
 	ack.frameId = pdu.frameId;
 	ack.totalFramesDecoded = gfx->TotalDecodedFrames;
+	IFCALLRET(context->PreFrameAck, sendAck, context, &ack);
```
akallabeth: Dereferencing of `context` is a possible `NULL` dereference

PR #5309, opened 2019-03-18
Fix NTLM AvPair lists

### LINKED ISSUE #5250: freerdp-2.0.0_rc4: TestNTLM, TestFreeRDPCodecInterleaved fail on ppc64

smaeul:
There were two main issues here: First, the `ntlm_av_pair_add` and
`ntlm_av_pair_add_copy` were not adding a new `MsvAvEOL` to the end of
the list to replace the one they overwrote. This caused the second call
to one of those functions to fail (since it couldn't find the
terminator), which was the source of the test failure. It also caused
`ntlm_av_pair_list_length` and `ntlm_print_av_pair_list` to read out of
bounds until they happened to find the right word.

Second, several bounds checks were wrong or missing. For example,
`ntlm_av_pair_add` does not ensure that the value fits inside the list.
And `ntlm_av_pair_get_len` and `ntlm_av_pair_get_value_pointer` can
return error codes or NULL, but those error returns were ignored, and
the values used anyway (such as in `ntlm_av_pair_add_copy`).

This fixes the list handling code to have the invariant that all
functions returning `NTLM_AV_PAIR*` only return non-`NULL` if the entire
returned `AvPair` is within bounds. This removes the need for the length
parameter in functions that only operate on a single `AvPair`. This
check is performed by the new `ntlm_av_pair_check` helper, which is
added in some new places and used to simplify the code in others.

Other issues fixed along the way include:
 - `ntlm_av_pair_list_length` did not cast to `PBYTE`, so it was
   returning the number of `NTLM_AV_PAIR`-sized chunks (which was
   possibly not even an integer) instead of the number of bytes
 - I removed an impossible check for `offset <= 0` in
   `ntlm_av_pair_get_next_pointer`
 - The assertion that `Value != NULL` and the call to `CopyMemory` are
   only necessary if `AvLen` is nonzero
 - `ntlm_av_pair_get_next_pointer` (renamed to `ntlm_av_pair_next`)
   could be declared `static`

With this commit, TestNTLM now passes on powerpc64.

```
$ ./Testing/TestSspi TestNTLM
NTLM_NEGOTIATE (length = 40):
NTLM_CHALLENGE (length = 168):
NTLM_AUTHENTICATE (length = 352):
$ echo $?
0
```

Fixes #5250

### DIFF
--- a/winpr/libwinpr/sspi/NTLM/ntlm_av_pairs.c
+++ b/winpr/libwinpr/sspi/NTLM/ntlm_av_pairs.c
@@ -54,14 +54,15 @@ static const char* const AV_PAIR_STRINGS[] =
 	"MsvChannelBindings"
 };
 
-static NTLM_AV_PAIR* ntlm_av_pair_get_next_pointer(NTLM_AV_PAIR* pAvPair, size_t* pcbAvPair);
+static BOOL ntlm_av_pair_check(NTLM_AV_PAIR* pAvPair, size_t cbAvPair);
+static NTLM_AV_PAIR* ntlm_av_pair_next(NTLM_AV_PAIR* pAvPairList, size_t* pcbAvPairList);
 
-static void ntlm_av_pair_set_id(NTLM_AV_PAIR* pAvPair, UINT16 id)
+static INLINE void ntlm_av_pair_set_id(NTLM_AV_PAIR* pAvPair, UINT16 id)
 {
 	Data_Write_UINT16(&pAvPair->AvId, id);
 }
 
-static void ntlm_av_pair_set_len(NTLM_AV_PAIR* pAvPair, UINT16 len)
+static INLINE void ntlm_av_pair_set_len(NTLM_AV_PAIR* pAvPair, UINT16 len)
 {
 	Data_Write_UINT16(&pAvPair->AvLen, len);
 }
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

PR #5310, opened 2019-03-18
Close file handle correctly.

cerg2010cerg2010:
We have found bugs using [PVS-Studio](https://www.viva64.com/en/pvs-studio-download/) tool. It is a static code analyzer for C, C++, C#, and Java.

You can find other FreeRDP bugs in this article: [Checking FreeRDP with PVS-Studio](https://www.viva64.com/en/b/0617/).


hardening:

PR #5311, opened 2019-03-18
Fix Wire to Surface 2 pixel format decode on client

m4ntis:
Came across this when tunneling GFX messages from `RdpgfxClientContext` to `RdpgfxServerContext` in the proxy we're implementing. The client just copies the `pixelFormat` field instead of parsing properly, while the server implementation does this correctly, which threw an error when tunneling the message.

This parsing of the same pixel format field is already implemented correctly in `recv_w2s1_pdu`, I just took the parse code from there. Should I maybe implement a conversion method that could be reused instead of having this big switch statement multiple times? If so, where would be the correct place for it, color.h where the pixel formats are defined?


akallabeth:
@freerdp-bot test

PR #5313, opened 2019-03-20
Fix update_end_paint wrong log level

kubistika:
Just a small fix for a debug log that is currently an error log. On most use-cases this function isn't called, but in our scenario as a proxy, testing connection to a windows 7 target server, this log spams the proxy unnecessarily.


hardening:

PR #5315, opened 2019-03-22
Add null terminator to RoutingToken in nego_set_routing_token

kubistika:
While working on the proxy we're implementing, We used the routing token as a temporary solution for passing the target server from the client. After running some manual tests, I noticed that `nego->RoutingToken` isn't null terminated correctly in `nego_set_routing_token`. While it is being parsed, the function `nego_read_request_token_or_cookie` replaces the CRLF with zeros:

```c
if (crlf == 0x0A0D)
{
	Stream_Rewind(s, 2);
	len = Stream_GetPosition(s) - pos;
	Stream_Write_UINT16(s, 0); /* Here the CRLF bytes are being replaced with zeros */
	if (strlen((char*)str) == len)
	{
		if (isToken)
			result = nego_set_routing_token(nego, str, len);
		else
			result = nego_set_cookie(nego, (char*)str);
	}
}
```
The problem is that `nego_set_routing_token` calls CopyMemory with Length = `nego->RoutingTokenLength` which is the length of the routing token without the null terminator byte.

I chose to fix it by adding the null terminator "manually" instead of calling CopyMemory with `RoutingTokenLength + 1` because I found it more readable and understandable.

### DIFF
--- a/libfreerdp/core/nego.c
+++ b/libfreerdp/core/nego.c
@@ -1294,12 +1294,13 @@ BOOL nego_set_routing_token(rdpNego* nego, BYTE* RoutingToken, DWORD RoutingToke
 
 	free(nego->RoutingToken);
 	nego->RoutingTokenLength = RoutingTokenLength;
-	nego->RoutingToken = (BYTE*) malloc(nego->RoutingTokenLength);
+	nego->RoutingToken = (BYTE*) malloc(nego->RoutingTokenLength + 1);
 
 	if (!nego->RoutingToken)
 		return FALSE;
 
 	CopyMemory(nego->RoutingToken, RoutingToken, nego->RoutingTokenLength);
+	nego->RoutingToken[nego->RoutingTokenLength] = 0;
 	return TRUE;
 }
 


akallabeth:
@freerdp-bot test

akallabeth:
@kubistika I understand that it might help to have a `\0` terminator appended, but what kind of issue did you have?
The whole routing token is not a string but a byte array and no termination is required. (and even then you have the length argument which limits the size of the buffer)
Sounds like there is some other issue hiding beneath

kubistika:
@akallabeth As I said, we use the routing token as a solution to pass the target host from the client. Meaning that settings->ServerHostname is set to the client's routing token. Therefore, without the null byte, sometimes the connection couldn't be opened because the hostname wasn't terminated and that's caused a bug. I understand that no null-termination is **required**, but I don't think there are any disadvantages of adding it. It will just add the option to use the RoutingToken as a string.

We can solve this by copying the routing token to a buffer and add a null terminator, if you don't think that this change should be merged.

hardening:
Same as @akallabeth , the spec doesn't require null termination and you have a way to achieve what you want without any changes. IMHO this PR is useless.

kubistika:
:+1:

PR #5322, opened 2019-04-04
Fix double call to rdpgfx_server_close on RdpgfxServerContext

m4ntis:
I just noticed that `rdpgfx_server_close` is both exposed as `RdpgfxServerContext`'s `Close` method and is called from `rdpgfx_server_context_free` unconditionally. Surely enough, when adding a print at the beginning of `rdpgfx_server_close`, it is printed twice consecutively in shadow. The second close doesn't do anything as it checks if everything it tries to free has already been freed, so I just added a check if the server has already been closed before calling it in `rdpgfx_server_context_free`.

### DIFF
--- a/channels/rdpgfx/server/rdpgfx_main.c
+++ b/channels/rdpgfx/server/rdpgfx_main.c
@@ -1614,10 +1614,15 @@ out_free:
 
 void rdpgfx_server_context_free(RdpgfxServerContext* context)
 {
-	rdpgfx_server_close(context);
-
 	if (context->priv)
+	{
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

m4ntis:
It is indeed called serially from the same thread, this is not a sync issue. Just thought it is logically weird to call `close` on the server twice, depending on the `close` method's implementation to retain wanted behavior.

I guess that this PR doesn't have a concrete reason, should have probably opened an issue asking about it beforehand :)

kubistika:
@akallabeth I guess this one can be closed

PR #5325, opened 2019-04-05
fix a bunch of warnings

hardening:
This kills almost all warnings here

### DIFF
--- a/channels/rail/client/rail_orders.c
+++ b/channels/rail/client/rail_orders.c
@@ -948,6 +948,7 @@ out:
 	return error;
 }
 
+#if 0
 /**
  * Function description
  *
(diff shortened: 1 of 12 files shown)


akallabeth:
@hardening Did you see #5325? Might be a good idea to merge that one before fixing stuff already handled ;)

hardening:
@akallabeth rebased with your warning PR applied

akallabeth:
@hardening if I see this correct, then the only remaininig changes are `#if 0` ?
Would opt for keeping these warnings or removing the functions all togeher (as they have never been used)

hardening:
@akallabeth well we have some working code that is just not used, but may be used in the future, better keep it IMHO

akallabeth:
@hardening Then I´m leaning toward letting these warnings stand (unused function) as that way the code at least compiles

PR #5330, opened 2019-04-05
FreeRDP#5329 if using OldLicenseBehaviour, don't try to save the Cal …

byteboon:
…since we're not going to try to load it


akallabeth:
@freerdp-bot test

akallabeth:
@byteboon Thank you for this patch!

PR #5331, opened 2019-04-07
Fix segfault when IniFile_Load is called with invalid input

kubistika:
Just a small fix to a bug that caused a segfault to occur when `IniFile_Load` is called with an invalid input. The function's signature implies that an invalid INI input would return `-1` while the actual functionality is a segfault. For example, when calling `IniFile_Load` with this input, a segfault occurs.

```ini
[General]
Host = "0.0.0.0"
Port = 3389
LocalOnly = 0

[Input]
Mouse = 1
Keyboard = 1

[Graphics]
GFX = 1
BitmapUpdate = 1
Some Illegal Key

[Security]
NlaSupport = 0
TlsSupport = 0
RdpSupport = 0

```
That happens because the code assumes that the file is a valid INI. Therefore, there was no check for `strchr`'s result in line 392. I've added another test case to ensure the problem was fixed.

**I also ran the code formatting script.**

### DIFF
--- a/winpr/libwinpr/utils/ini.c
+++ b/winpr/libwinpr/utils/ini.c
@@ -390,6 +390,10 @@ int IniFile_Load(wIniFile* ini)
 		else
 		{
 			separator = strchr(line, '=');
+
+			if (separator == NULL)
+				return -1;
+
 			end = separator;
 
 			while ((&end[-1] > line) && ((end[-1] == ' ') || (end[-1] == '\t')))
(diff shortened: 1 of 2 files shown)


kubistika:
@akallabeth Thanks, can you check it again?

akallabeth:
@freerdp-bot test

akallabeth:
@kubistika Thank you for this nice fix!

PR #5339, opened 2019-04-11
Unified update->BeginPaint and update->EndPaint

akallabeth:
Since these functions were called from 2 different threads
(main thread or dynamic channel) depending on fastpath or
gfx channel use lock between these.
If not locked the invalid region may be accessed from both
threads and lead to crashes as experienced with nightly df280a7ff

Testcase:
`xfreerdp /rfx /gfx /network:auto` to a windows 10 machine sometimes sends fastpath `EndPaint` events that may crash the client. (best seen with nightly and address sanitizer)


akallabeth:
@bmiklautz @mfleisz anyone time to have a look?

PR #5350, opened 2019-04-18
Rail exec order

m4ntis:
This PR might be a bit out of nowhere without context, so I'll explain shortly. In the RDP proxy we're developing (that I've mentioned in a few of my PRs already), we want to support RAIL proxy. For that to work, I am currently working on implementing RAIL server support for FreeRDP (protocol level support, not an actual implementation for a server like shadow). While implementing receiving a RAIL Exec order from the client, I noticed that the API for the Exec order is inconsistent with the rest of the orders:
- The strings in the order are unicode strings, but are instead represented as `char*` in the order struct.
- The methods that deal with the order don't even use the order struct, they just receive the struct's properties instead.

The point of this PR is to unify the handling of orders and to make it easier to work with exec order (using one struct and not multiple strings and values)

### DIFF
--- a/channels/rail/client/rail_main.c
+++ b/channels/rail/client/rail_main.c
@@ -105,46 +105,22 @@ UINT rail_send_channel_data(railPlugin* rail, void* data, size_t length)
  * @return 0 on success, otherwise a Win32 error code
  */
 static UINT rail_client_execute(RailClientContext* context,
-                                const RAIL_EXEC_ORDER* exec)
+                                RAIL_EXEC_ORDER* exec)
 {
-	char* exeOrFile;
-	UINT error;
 	railPlugin* rail;
-	UINT16 flags;
-	RAIL_UNICODE_STRING ruExeOrFile = { 0 };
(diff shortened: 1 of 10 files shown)


akallabeth:
@freerdp-bot test

m4ntis:
@freerdp-bot test

akallabeth:
@m4ntis @freerdp-bot only listens to its masters ;)

akallabeth:
@freerdp-bot test please

m4ntis:
> @m4ntis @freerdp-bot only listens to its masters ;)

:D 
Thought it would, just wanted to test whether I could save the masters some time :P

m4ntis:
@akallabeth oh yeah sorry I'm on a small vacation, will look at this again at the start of next week

speidy:
@m4ntis can we close this one ?

m4ntis:
Yeah, will include this in my RAIL server PR as I discussed with @akallabeth

m4ntis:
Yeah, will include this in my RAIL server PR as I discussed with @akallabeth

PR #5354, opened 2019-04-18
xf_floatbar.c: fix build without Xfixes

ffontaine:
Remove unneeded include on Xfixes.h as it is not always available and
not used in xf_floatbar.c

Fixes:
 - http://autobuild.buildroot.org/results/69245e574787bada718c52c805ec137041dc233d

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


akallabeth:
@freerdp-bot test

Issue #5360, opened 2019-04-25
Segfault using /smartcard without any readers


r-barnett:
**Describe the bug**
A segmentation fault results when using /smartcard:scard when there are no smartcard devices attached to the local host when the remote host has smartcard software installed or when such a device is detached while xfreerdp is running.
**To Reproduce**
Run "xfreerdp /smartcard:scard /v:host" where host is a host that has smartcard software installed and the local host does not have any smartcard devices attached.

**Expected behavior**
No segmentation fault.

** Application details
* Version of FreeRDP
Built from master branch today.
* Command line used
xfreerdp /smartcard:scard /v:host
* output of `/buildconfig`
This is FreeRDP version 2.0.0-dev5 (e1ef302a8)
[...]

PR #5361, opened 2019-04-25
5360 segfault using smartcard without any readers

informatimago:
Fixed issue #5360
by testing for null readers as returned by ConvertFromUnicode when out of memory.


hardening:

PR #5372, opened 2019-05-01
First version of an RDP proxy

m4ntis:
#### Hey all,

As some of you maintainers probably know by now, for the last couple of months me and 2 colleagues of mine (@kubistika and @speidy) have been working on implementing an RDP proxy using the library. This is a PR to introduce what we've been working on up till now, containing a working base version of the proxy.

The changes and number of commits are quite vast, so I'll start with the motivation, why we chose to implement it in this way and why the *h* you might want all of this massive PR. Later on I'll get to what is implemented here, how to use it, future goals and where we want to get to with this proxy.

## Motivation
The proxy came about for a number of reasons, mostly security related:
- There is no real good RDP proxy out there, enabling you to finely select the specific features and channels you want to allow, filter or right out disable for all connections passing through. Lets say you have some isolated network where you want to minimize all traffic going in and out while still enabling easy access via RDP, you probably don't want any channels passing through but the bare minimum, say GFX from the server and basic input events from the client.
- There is no proxy that lets you monitor and analyze RDP traffic going through it. We might want to log certain events or look for anomalies, such as suspicious amounts of certain events or specific channel messages.
- We wanted a proxy that will let you capture and record all traffic passing through. You might use this feature by placing the proxy as the only gateway to some sensitive server such as a DC, being able to view all passed connections filtered by time and username.
- We wanted to have a proxy that has an API for extra authentication, letting us implement a 2FA towards the proxy and disallowing connections to the target server without authentication.

## Why we chose FreeRDP
So the reasoning for choosing to implement the proxy as a part of the FreeRDP library is quite obvious. FreeRDP is by far the most maintained and up to date OS implementation of the protocol, and with the server side implementations only getting bigger, it only seemed fitting. As we **do** parse the messages going through the proxy (the whole point of the proxy), in order to support updated versions of the protocol we need to use an updating implementation of RDP. There is a tiny, basic proxy MITM in [pyrdp](https://github.com/GoSecure/pyrdp) but it's too small, not extensive and with way lower chances of being properly maintained in the long run.

## Benefits to the library
Well first of all, we hope that the proxy in itself is a useful thing to have on it's own. But further than that, as you've seen in the last month or so, me and @kubistika already issued a few small PRs that fixed some of the server side implementation of the protocol. This is just the beginning, and this PR already has a few other fixes to GFX parsing bundled inside it (I'll elaborate below). The proxy tests both client and server side implementations of the protocol without having to implement a client nor server, as you can easily use mstsc and Microsoft's RDP TS server going talking via the proxy to quickly find any parsing errors.

I'm sure we'll continue to squish bugs as we continue to other parts of the protocol :wink:.

Additionally, the proxy is an incentive for us to implement the server side of a few other channels that aren't implemented yet, as we want to parse and proxy them too (I am currently working on a RAIL server implementation, might be a few other coming after that).

## So what do we have here
This is the first usable version of the proxy, so most of our goals aren't yet fully implemented (some not at all). What we do have is a proxy that can parse and proxy connections, passing graphics messages and keyboard / mouse events. GFX and bitmap updates are fully working, GDI still has some issues but I think that it's obvious why GDI is a low priority right now.

**Basic list of features**:
- GFX, bitmap updates and input events parsing and proxying.
- Settings sync: A client connects to and negociates with the proxy, the proxy takes the negociated settings and connects with them to the server, and then renegociates with the client using the final settings.
- GFX capabilities sync: The client and server capabilities are properly synced.
- Basic configuration (example below). Not all features in the config file are fully implemented yet, but soon to be.

There are a few changes to the library itself too:
#### Bugfixes:
1. In `client/rdpgfx_main.c:rdpgfx_recv_caps_confirm_pdu()`, the caps data length was incorrectly read into a local variable.
2. In `client/rdpgfx_main.c:rdpgfx_recv_map_surface_to_scaled_window_pdu()` and `scaled_output_pdu`, the stream length magic number was incorrect, as the methods were probably copy+pasted from `recv_map_surface_to_window/output` without changing the length.
3. Same issue as the one above in the server side, in `server/rdpgfx_main.c:rdpgfx_send_map_surface_to_scaled_window_pdu`, stream length is incorrect.
4. Same again with `scaled_output_pdu` on the server side, but on this PDU the values in the struct were named wrongly by the spec, `targetX` and `targetY` instead of `targetWidth` and `targetHeight`.

#### Changes to API:
1. In order to properly be able to forward GFX frame acking and to sync capabilities between the client and server, we needed to have a way to have control over sending FrameAck and CapsConfirm messages, so we exposed these methods as part of the `RdpgfxClientContext`'s API.
2. Additionally, we needed some way of disabling the client's automatic frame acking / caps confirmation, without hindering default functionality. We added and optional `OnOpen` callback with flags that control whether to `do_frame_acks` and `do_caps_confirm`. These values default to `TRUE`.

## Usage
To compile with cmake, the `-DWITH_SERVER=on` flag will compile the proxy. Running it is as simple as `./pfreerdp` (there are no CLI options yet). There is a `config.ini` present in `server/proxy` that currently has to be in the working dir of the proxy. Lets go over the config:

```ini
[Server]
Host = "0.0.0.0"
Port = 3389
LocalOnly = 0

[Target]
UseLoadBalanceInfo = 1
Host = "Target server"
Port = 3389

[Input]
Mouse = 1
Keyboard = 1

[Graphics]
GFX = 1
BitmapUpdate = 1

[Security]
NlaSecurity = 0
TlsSecurity = 1
RdpSecurity = 1

[Channels]
WhitelistMode = 0
AllowedChannels = "cliprdr,Microsoft::Windows::RDS::Video::Control"
DeniedChannels = "Microsoft::Windows::RDS::Geometry"
```
All sections but the `[Target]` and `[Channels]` are quite obvious (host config and allowing / disallowing RDP features).

The `[Target]` config option regards the method of conveying which server shall the client connecting to the proxy be forwarded to. The simple option is configuring a `Host` and `Port` value and setting `UseLoadBalanceInfo` to `0`. This fixes all connections to a single configured destination. The second method is to dynamically detect the target for each incoming connection. The way we figured out is to use the routing token to send that info (this is one of the things we would like to hear your opinion on). This allows the proxy to receive the target server info from any normal rdp client. For mstsc, you could use the `loadbalanceinfo:s:` option in an rdp file, and in any of the FreeRDP clients you could use the `/load-balance-info:` flag. The format of the routing token is `Cookie: msts=<token>`, so as an example, using with xfreerdp would look like this:

`xfreerdp /v:<proxy addr> /load-balance-info:"Cookie: msts=<destination server>"`
 
The `[Channels]` section is not yet implemented, but expresses the two modes we want the proxy to be able to work in, blacklist and whitelist. The first would allow all channels excluding the ones configured in `DeniedChannels`. The latter would allow only the channels configured in `AllowedChannels`. Currently, only the channels explicitly supported by the proxy are allowed. In order to support a blacklist mode, we would have to find a way of proxing channels without inspecting the data. We thought about implementing a "dummy" channel that just exposes receiving and sending data (we'd love hearing suggestions about this too).

### Small caveat
NLA isn't supported yet, but will be very soon.

## Looking forward
I'll keep this short, but here's a short summary of the features we want to implement:

1. A hooking type of API for events in the connection, probably loading "so" files that register hooks the proxy exposes.
2. Implementing more channels.
3. Implementing a "blacklist" channel mode (explained just above)
4. Adding some sort of API for a third party authentication logic.
5. Connection recording functionality.

This list can change / grow, we're willing to hear suggestions for features / how to implement them.

Please note that this is just the first initial PR for this proxy. We are and will be actively developing this proxy, adding support for more channels and features. We wanted to do this PR and merge early so that we could get a lot of feedback and suggestions from the maintainers, and so that you follow and review our work.

Thanks for taking the time to read and review this big PR :)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -625,24 +652,23 @@ static UINT rdpgfx_recv_end_frame_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
 	gfx->TotalDecodedFrames++;
 	ack.frameId = pdu.frameId;
 	ack.totalFramesDecoded = gfx->TotalDecodedFrames;
-	IFCALLRET(context->PreFrameAck, sendAck, context, &ack);
```
akallabeth: Why did you remove this callback?
Don't remember who it was but the idea was to control the callback sending by external parameters

libfreerdp/core/nego.h
```diff
@@ -92,6 +92,34 @@ enum RDP_NEG_MSG
 #define REDIRECTED_AUTHENTICATION_MODE_REQUIRED		0x02
 #define CORRELATION_INFO_PRESENT				0x08
 
+struct rdp_nego
```
akallabeth: would prefer it if this struct stays opaque and you implrement access to the routing token via getter/setter

akallabeth:
getting some warnings like
`server/proxy/pf_rdpgfx.c:97:39: warning: passing argument 2 of ‘server->CreateSurface’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  return server->CreateSurface(server, createSurface);`
warnings, might check that

akallabeth:
Created a patch to fix the warnings mentioned above.

[0001-Fixed-const-correctness-of-gfx-function-pointer.zip](https://github.com/FreeRDP/FreeRDP/files/3136771/0001-Fixed-const-correctness-of-gfx-function-pointer.zip)

hardening:
Yes looks interesting.
Out of curiosity, did you have a look at ogon, I feel like you can implement the same kind of functionality with it but with a bigger range (I also understand the need to just proxy one server) ?
I'm citing ogon, because I think we've addressed a lot of the problems you may hit when implementing your proxy (blocking writes and bandwidth management specifically).

channels/rdpgfx/client/rdpgfx_main.c
m4ntis: It was mine actually, I initially didn't want to expose FrameAck sending and some flag to control automatic frame acking, but when coming to implement it in the proxy, alongside needing to manually control capability sending, our current solution made more sense and was a cleaner, simpler implementation.
As the PreFrameAck was my idea in the first place, I was confident enough to know it can be removed.

akallabeth:
@freerdp-bot test

m4ntis:
It's the old msbuild failing about local variable decleration, probably all over the code, fixing now

akallabeth:
@freerdp-bot test

m4ntis:
@freerdp-bot You really like to spoil the fun don't you :P
Think I found the last culprit

hardening:
@freerdp-bot test again my grumpy buddy ;)

hardening:
@freerdp-bot would you test corrections ?

akallabeth:
Created another patch to minimize the changes to the core library.

I'm still checking the changes to the gfx channel (which are currently blocking inclusion for 2.0 release)
Would it be possible to write this differently so that the callbacks would not have to be modified (new client callbacks, old ones removed)?
These are breaking changes that we'd like to prevent.
So, if you know a way to avoid these, we could merge to 2.0

[0001-Fixed-const-correctness-of-gfx-function-pointer.zip](https://github.com/FreeRDP/FreeRDP/files/3142747/0001-Fixed-const-correctness-of-gfx-function-pointer.zip)

m4ntis:
If necessary in order to be merged now, we'll definitely be thinking about implementing things in a different way. Just to be clear, the removed callback is one that I've added in order to solve our problem exactly, and as we've found what we considered a cleaner, nicer solution, I thought it could be removed in favor on the new callbacks. Moreover, we made sure the new callbacks don't hinder current functionality, all callbacks being optional and the flags they control defaulting to maintain current behavior. Anyway, the removal of the callback is not at all mandatory for this PR and can be reverted if necessary.

akallabeth:
@m4ntis ok, then it will most likely not break someons use case ^^
Sorry, didn't remember you were the one. 
Will check if the remaining changes are compatible with the current api on monday (we try to only change the const stuff or add new stuff)
Other than that the pr looks good.

akallabeth:
Maybe squashing the commits and splitting in gfxfix, api extension and new shiny proxy could be helpful.

akallabeth:
Ok, rechecked and recognised I've attached the wrong patch in the last comments :/

here is the new one (instead of the old on once again)
[0001-Removed-nego-struct-from-direct-access.zip](https://github.com/FreeRDP/FreeRDP/files/3145320/0001-Removed-nego-struct-from-direct-access.zip)

m4ntis:
Oh I was just about to tell you that the patch files are the same. Thx for the patch :)
We're working on splitting the changes to the library into different PRs that'll probably have to all be merged before the proxy's PR can be rebased and compile.
And could we keep the proxy's history in tact and not squash the whole pr? The history has some meaning to us as we're continuing to work on it.

akallabeth:
@speidy you're free to change that (would require some getter for the nego struct too)
this was just my approach to that.
Just want the structs hidden away so that nobody is even tempted using direct access (eliminates really a lot of problems we encountered with this project)

akallabeth:
@freerdp-bot test

m4ntis:
Right, finished moving all the changes to the GFX library to PRs outside of one, and removing them from here. Of course, this means the PR won't compile until the others are merged and this one is rebased on master with the changes:
```sh
$ cmake --build . 2>&1 | grep error
server/proxy/pf_rdpgfx.c:277:15: error: ‘RdpgfxClientContext’ {aka ‘struct _rdpgfx_client_context’} has no member named ‘CapsAdvertise’
server/proxy/pf_rdpgfx.c:286:15: error: ‘RdpgfxClientContext’ {aka ‘struct _rdpgfx_client_context’} has no member named ‘FrameAcknowledge’
server/proxy/pf_rdpgfx.c:313:5: error: ‘RdpgfxClientContext’ {aka ‘struct _rdpgfx_client_context’} has no member named ‘OnOpen’
server/proxy/pf_rdpgfx.c:314:5: error: ‘RdpgfxClientContext’ {aka ‘struct _rdpgfx_client_context’} has no member named ‘OnClose’
server/proxy/pf_rdpgfx.c:315:5: error: ‘RdpgfxClientContext’ {aka ‘struct _rdpgfx_client_context’} has no member named ‘CapsConfirm’
```

libfreerdp/core/nego.c
```diff
@@ -1440,3 +1440,12 @@ void nego_free_nla(rdpNego* nego)
 	nla_free(nego->transport->nla);
 	nego->transport->nla = NULL;
 }
+
+const BYTE* nego_get_routing_token(rdpNego* nego, DWORD* RoutingTokenLength)
+{
+	if (!nego)
```
hardening: Just an idea but perhaps it would be nice to also set `*RoutingTokenLength` to 0 ?

bmiklautz:
@freerdp-bot test

server/proxy/freerdp_proxy.c
```diff
@@ -0,0 +1,68 @@
...
+
+		case CONFIG_INVALID:
+			goto fail;
+	}
+
+	if (config->WhitelistMode)
+	{
+		WLog_INFO(TAG, "Channels mode: WHITELIST");
+
+		for (i = 0; i < config->AllowedChannelsCount; i++)
```
akallabeth: iterator sign warning

server/proxy/freerdp_proxy.c
```diff
@@ -0,0 +1,68 @@
...
+			WLog_INFO(TAG, "Allowing %s", config->AllowedChannels[i]);
+	}
+	else
+	{
+		WLog_INFO(TAG, "Channels mode: BLACKLIST");
+
+		for (i = 0; i < config->BlockedChannelsCount; i++)
```
akallabeth: iterator sign warning

server/proxy/pf_config.h
```diff
@@ -0,0 +1,71 @@
...
+#define CONFIG_PARSE_ERROR 	 1
+#define CONFIG_INVALID 		 2
+
+#include <winpr/ini.h>
+
+struct proxy_config
+{
+	/* server */
+	char* Host;
+	INT32 Port;
```
akallabeth: Lot of warnings originate from the port being a signed type (and targets expecting unsigned)
Maybe just modify the check when reading instead of exposing this with a type that does not match.

server/proxy/pf_config.h
```diff
@@ -0,0 +1,71 @@
...
+	BOOL  LocalOnly;
+
+	/* target */
+	BOOL UseLoadBalanceInfo;
+	char* TargetHost;
+	INT32 TargetPort;
```
akallabeth: Lot of warnings originate from the port being a signed type (and targets expecting unsigned)
Maybe just modify the check when reading instead of exposing this with a type that does not match.

server/proxy/pf_server.c
```diff
@@ -0,0 +1,423 @@
...
+		*target = malloc(len + 1);
+		CopyMemory(*target, routing_token + prefix_len, len);
+		*(*target + len) = '\0';
+		colon = strchr(*target, ':');
+		WLog_INFO(TAG, "Target [parsed from routing token]: %s", *target);
+
+		if (colon)
+		{
+			/* port is specified */
+			*port = strtoul(colon + 1, NULL, 10);
```
akallabeth: No error checks here, types also do not match (32bit to 64bit on some platforms)

server/proxy/pf_server.c
```diff
@@ -0,0 +1,423 @@
...
+}
+
+static BOOL pf_server_parse_target_from_routing_token(freerdp_peer* client,
+        char** target, DWORD* port)
+{
+#define TARGET_MAX	(100)
+#define ROUTING_TOKEN_PREFIX "Cookie: msts="
+	const char* routing_token;
+	char* colon;
+	int len;
```
akallabeth: type mismatches

server/proxy/pf_server.c
```diff
@@ -0,0 +1,423 @@
...
+	int prefix_len;
+	DWORD routing_token_length;
+	routing_token = freerdp_nego_get_routing_token(client->context, &routing_token_length);
+	prefix_len = strlen(ROUTING_TOKEN_PREFIX);
+
+	if (routing_token &&
```
akallabeth: some braces might make this more readable

server/proxy/freerdp_proxy.c
```diff
@@ -0,0 +1,68 @@
...
+#include "pf_log.h"
+
+#define TAG PROXY_TAG("server")
+
+int main(int argc, char* argv[])
+{
+	int status = 0;
+	int i;
+	proxyConfig* config = calloc(1, sizeof(proxyConfig));
+	status = pf_server_load_config("config.ini", config);
```
akallabeth: might be a good idea to allow overriding this with command line arguments

server/proxy/pf_config.c
```diff
@@ -0,0 +1,183 @@
...
+		{
+			WLog_ERR(TAG, "Configuration value for `Target.Port` is not valid");
+			return FALSE;
+		}
+	}
+
+	return TRUE;
+}
+
+DWORD pf_server_load_config(char* path, proxyConfig* config)
```
akallabeth: path can be made const

server/proxy/pf_channels.h
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+
+#ifndef FREERDP_SERVER_PROXY_PFCHANNELS_H
+#define FREERDP_SERVER_PROXY_PFCHANNELS_H
+
+#include <freerdp/freerdp.h>
+#include <freerdp/client/channels.h>
+
+int pf_on_channel_connected(freerdp* instance, const char* name,
```
akallabeth: Please remove unused functions

server/proxy/pf_common.c
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+#include "pf_common.h"
+
+BOOL pf_common_connection_aborted_by_peer(proxyData* pdata)
+{
+	return WaitForSingleObject(pdata->connectionClosed, 0) == WAIT_OBJECT_0;
+}
+
+void pf_common_copy_settings(rdpSettings* dst, rdpSettings* src)
```
akallabeth: Why not use `freerdp_settings_clone`?

server/proxy/pf_client.c
```diff
@@ -0,0 +1,415 @@
...
+#include "pf_context.h"
+#include "pf_log.h"
+
+#define TAG PROXY_TAG("client")
+
+/**
+ * Re-negociate with original client after negociation between the proxy
+ * and the target has finished.
+ */
+void proxy_server_reactivate(rdpContext* client, rdpContext* target)
```
akallabeth: Check internal functions, if possible make them `static`

server/proxy/pf_config.c
```diff
@@ -0,0 +1,183 @@
...
+	{
+		tokens = realloc(tokens, tokens_count * sizeof(char*));
+	}
+
+	*length = tokens_count;
+	free(s);
+	return tokens;
+}
+
+BOOL pf_server_is_config_valid(proxyConfig* config)
```
akallabeth: Check internal functions, if possible make them `static`

server/proxy/pf_context.c
```diff
@@ -0,0 +1,81 @@
...
+	if (context)
+	{
+		WTSCloseServer((HANDLE) context->vcm);
+	}
+}
+
+BOOL init_p_server_context(freerdp_peer* client)
+{
+	client->ContextSize = sizeof(pServerContext);
+	client->ContextNew = (psPeerContextNew) client_to_proxy_context_new;
```
akallabeth: casts are evil, they prevent the compiler from checking if the functions assigned have the same signature as the pointer.

akallabeth:
@m4ntis ok, hope this does not look too picky, but I'd like to get new stuff in clean.
Have tested basic proxy functionality and it works. (as mentioned, a simple command line argument to specify which configuration file to load would be awesome)

include/freerdp/server/rdpgfx.h
```diff
@@ -21,6 +21,7 @@
 #define FREERDP_CHANNEL_RDPGFX_SERVER_RDPGFX_H
 
 #include <freerdp/channels/rdpgfx.h>
+#include <freerdp/freerdp.h>
```
akallabeth: unnecessary change, better remove.

akallabeth:
Ok, checked the server and already did the fixes ;)

[0001-Cleaned-up-proxy-server-code.zip](https://github.com/FreeRDP/FreeRDP/files/3157048/0001-Cleaned-up-proxy-server-code.zip)

bmiklautz:
@freerdp-bot test

akallabeth:
@freerdp-bot test

include/freerdp/server/rdpgfx.h
m4ntis: Umm, is it? It doesn't compile without it. There is usage of the `rdpContext` struct without including `freerdp.h`. Surely it's a better solution, rather than relying on including `freerdp.h` before `server/rdpgfx.h` each time

akallabeth:
@kubistika @m4ntis @speidy Only remaining question, anything against squashing the commits?
120 is quite a number ;)
nothing you have to do, can do that when merging automatically. (learned from my previous mistakes, makes debugging later on a lot easier if the commit actually compiles and works)

m4ntis:
I guess you have a point. We'd rather not, as we use it to see certain changes to files from time to time, but after all it isn't the most important thing, and if you find squashing more appropriate do as you wish 😉

akallabeth:
@hardening @bmiklautz any one of you time for a second opinion?

speidy:
LGTM, lets merge

PR #5376, opened 2019-05-05
rdpgfx: Fix GFX v10.6 PDUs parsing and naming according to the spec

kubistika:
This PR fixes the new GFX 10.6 client and server PDUs parsing and naming according to the spec.
This change is taken out of our proxy PR as @akallabeth asked to do. The proxy's functionality depends on this one.


akallabeth:
@freerdp-bot test

PR #5377, opened 2019-05-05
client/rdgpfx_main.c: Fix rdpgfx_recv_caps_confirm_pdu caps length read

m4ntis:
Just another PR with a fix from the proxy, GFX caps sync doesn't work properly without this one

### DIFF
--- a/channels/rdpgfx/client/rdpgfx_main.c
+++ b/channels/rdpgfx/client/rdpgfx_main.c
@@ -206,7 +206,6 @@ static UINT rdpgfx_recv_caps_confirm_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
         wStream* s)
 {
 	RDPGFX_CAPSET capsSet;
-	UINT32 capsDataLength;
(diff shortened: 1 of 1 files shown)


channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -218,7 +217,7 @@ static UINT rdpgfx_recv_caps_confirm_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
 	}
 
 	Stream_Read_UINT32(s, capsSet.version); /* version (4 bytes) */
-	Stream_Read_UINT32(s, capsDataLength); /* capsDataLength (4 bytes) */
+	Stream_Read_UINT32(s, capsSet.length); /* capsDataLength (4 bytes) */
```
akallabeth: for current use most likely ok, but with capssets that have differing lengths `gfx->ConnectionCaps` should be adjusted to be a dynamically allocated copy.

akallabeth:
@freerdp-bot test

PR #5381, opened 2019-05-06
Added image scaling api for software drawing.

akallabeth:
For future GFX channel functions an image scaling function is required.
This moves the implementation from wayland client to core library
and adds support for the much faster SWScale library.

### DIFF
--- a/cmake/ConfigOptions.cmake
+++ b/cmake/ConfigOptions.cmake
@@ -145,11 +145,14 @@ endif(WITH_FFMPEG)
 
 option(USE_VERSION_FROM_GIT_TAG "Extract FreeRDP version from git tag." OFF)
 
-if(ANDROID)
-include(ConfigOptionsAndroid)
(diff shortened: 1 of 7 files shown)


akallabeth:
@bmiklautz @mfleisz anyone time to have a look?

cmake/ConfigOptions.cmake
```diff
@@ -149,7 +149,14 @@ if(ANDROID)
 include(ConfigOptionsAndroid)
 endif(ANDROID)
 
-if(IOS)
-include(ConfigOptionsiOS)
+option(WITH_CAIRO    "Use CAIRO image library for screen resizing" OFF)
+option(WITH_SWSCALE  "Use SWScale image library for screen resizing" OFF)
+
+if (ANDROID)
```
mfleisz: The same code is just 3 lines above

akallabeth: ok, fixed.

Issue #4934, opened 2018-10-16
In seamless mode, moving a window shrinks it 14x7


chrislong:
When running in seamless mode, moving any window causes that window to shrink by 14 pixels horizontally and 7 pixels vertically. Each move causes this, until the window starts moving across the screen by itself (to the right) when it has focus.

**To Reproduce**
1. Start xfreerdp in seamless mode
2. Drag an application by its title bar, watching its size (preferably your window manager reports this to you)
3. You should notice the window has gotten smaller. If you measure (e.g., by grabbin the application again if your WM shows you sizes when you do) you will find it's smaller now by 14x7 pixels.
4. If you drag it again, it will shrink the same amount again. Repeat as desired.

**Expected behavior**
Moving the window does not automatically shrink it.

**Screenshots**
I can provide these if needed.

** Application details
* Version of FreeRDP
This is FreeRDP version 2.0.0-rc3 (2.0.0-rc3)

* Command line used
xfreerdp /v:foo.bar.com /u:auser /p:mypassword /cert-ignore /app:'%windir%\explorer.exe'

* output of `/buildconfig`
This is FreeRDP version 2.0.0-rc3 (2.0.0-rc3)
[...]

PR #5395, opened 2019-05-09
Fix for #4934 - In seamless mode, moving a window shrinks it 14x7

chrislong:
This PR fixes the bug reported in #4934. With a lot of help from Microsoft tech support, I determined that the window margins have to be used as offsets on position and size values returned by the server.

To test it, run in single app mode, for example: `xfreerdp /v:HOST /u:USER /p:PWD /app:'c:\windows\explorer.exe' `. Move the window around. You should see the window stay the same size.

This PR includes a bunch of new debugging with the DEBUG_X11 macro that I put in to diagnose what was going on. Normally they compile to nothing, so I left them in, in case someone needs to debug this stuff later, but if you don't want them let me know.

I also added the function `xf_rail_window_new` to `xf_rail.c` to simplify `xf_rail_window_common`. It makes the PR noisier than it strictly would have to be to fix the bug, but I think it's an improvement since `xf_rail_window_common` is so long. But I can undo it if desired.


akallabeth:
@freerdp-bot test

akallabeth:
@chrislong wow, nice timing :)
Hope to have time for a review tomorrow, thx for this nice patchset!

akallabeth:
@chrislong would suggest keeping the old names in public header.
Variable declarations old C style please (beginning of scope)

chrislong:
@akallabeth thanks for the quick feedback. Just a few questions to make sure I understand the changes you want...

> @chrislong would suggest keeping the old names in public header.
You mean like `resizeMarginLeft`? Ok, I'll change it.

> Variable declarations old C style please (beginning of scope)
Do you mean the `for` loop declarations that caused the Android build to break? I'll fix that. (I didn't realize anything got built any more with pre-C99 rules.)

include/freerdp/window.h
```diff
@@ -72,6 +72,9 @@
 #define WINDOW_ORDER_FIELD_DESKTOP_ZORDER          0x00000010
 #define WINDOW_ORDER_FIELD_DESKTOP_ACTIVE_WND      0x00000020
 
+extern const UINT32 WINDOW_ORDER_VALUES[];
```
akallabeth: No need to export these, only used internally and if necessary wrap it in a function.

include/freerdp/window.h
```diff
@@ -72,6 +72,9 @@
...
+extern const char* const WINDOW_ORDER_STRINGS[];
```
akallabeth: No need to export these, only used internally and if necessary wrap it in a function.

include/freerdp/window.h
```diff
@@ -194,6 +197,10 @@ struct _WINDOW_STATE_ORDER
 	INT32 clientOffsetY;
 	UINT32 clientAreaWidth;
 	UINT32 clientAreaHeight;
+	UINT32 windowLeftResizeMargin;
+	UINT32 windowRightResizeMargin;
+	UINT32 windowTopResizeMargin;
+	UINT32 windowBottomResizeMargin;
```
akallabeth: As already mentioned, please keep the old names.
we try to keep API changes to a minimum now as we're really close to 2.0

libfreerdp/core/window.c
```diff
@@ -30,6 +30,124 @@
 
 #define TAG FREERDP_TAG("core.window")
 
+const UINT32 WINDOW_ORDER_VALUES[] =
```
akallabeth: Can be `static`

libfreerdp/core/window.c
```diff
@@ -30,6 +30,124 @@
...
+	WINDOW_ORDER_FIELD_DESKTOP_NONE,
+	WINDOW_ORDER_FIELD_DESKTOP_HOOKED,
+	WINDOW_ORDER_FIELD_DESKTOP_ARC_COMPLETED,
+	WINDOW_ORDER_FIELD_DESKTOP_ARC_BEGAN,
+	WINDOW_ORDER_FIELD_DESKTOP_ZORDER,
+	WINDOW_ORDER_FIELD_DESKTOP_ACTIVE_WND,
+	0
+};
+
+const char* const WINDOW_ORDER_STRINGS[] =
```
akallabeth: Can be `static`

channels/rail/client/rail_orders.c
```diff
@@ -1597,7 +1597,11 @@ UINT rail_send_client_window_move_order(railPlugin* rail, const RAIL_WINDOW_MOVE
 	error = rail_write_client_window_move_order(s, windowMove);
 
 	if (error == ERROR_SUCCESS)
+	{
+		WLog_Print(rail->log, WLOG_TRACE, "sending new RAIL window corners: 0x%x (%d, %d), (%d, %d)",
```
akallabeth: Please check the format string and use the appropriate `PRIx32` or similar C macros

libfreerdp/core/window.c
```diff
@@ -538,7 +657,7 @@ static BOOL update_recv_window_info_order(rdpUpdate* update, wStream* s,
 	else if (orderInfo->fieldFlags & WINDOW_ORDER_STATE_DELETED)
 	{
 		update_read_window_delete_order(s, orderInfo);
-		WLog_Print(update->log, WLOG_DEBUG, "WindowDelete");
+		WLog_Print(update->log, WLOG_DEBUG, "WindowDelete 0x%x", orderInfo->windowId);
```
akallabeth: Please use the appropriate format string macros

client/X11/xf_window.c
```diff
@@ -935,9 +934,16 @@ void xf_MoveWindow(xfContext* xfc, xfAppWindow* appWindow, int x, int y,
 	appWindow->height = height;
 
 	if (resize)
+	{
+		DEBUG_X11("setting new internal & X geometry: %dx%d+%d+%d\tshowState=%d", width, height, x, y,
```
akallabeth: Please use the appropriate format string macros

hardening:
I'm interested by this PR too. @chrislong could you do the changes asked by @akallabeth, especially the squash ?

chrislong:
@hardening Sorry for being slow on this. I had to address some things on other projects. I hope to get this cleaned up today, or if not then at least by some time Monday.

include/freerdp/window.h
chrislong: Ok.

include/freerdp/window.h
chrislong: Ok.

libfreerdp/core/window.c
chrislong: Right, thanks.

libfreerdp/core/window.c
chrislong: Right, thanks.

channels/rail/client/rail_orders.c
chrislong: Ok (I wasn't familiar with those macros).

libfreerdp/core/window.c
chrislong: Ok

client/X11/xf_window.c
chrislong: For variables declared with no bit size (e.g., `int x`), should I use `PRId32`? I didn't see a macro that uses the right size for just plain `int`.

include/freerdp/window.h
chrislong: Ok, I'll update the names. (FYI, I didn't arbitrarily rename existing fields; when I originally forked these didn't exist.)

chrislong:
I think I've addressed all the comments. I couldn't get rebasing on this branch to clean things up nicely (maybe due to inadequate git-foo, I don't know), so I made a new branch and a new PR with cleaner commit history: #5411.

akallabeth:
ok, then I´m closing this and we continue in #5411

PR #5403, opened 2019-05-16
Fixed fastpath/slowpatth pause event.

akallabeth:
@kubistika since we seem to misunderstand here the code of what I meant ;)

### DIFF
--- a/include/freerdp/input.h
+++ b/include/freerdp/input.h
@@ -31,6 +31,7 @@ typedef struct rdp_input rdpInput;
 
 /* keyboard Flags */
 #define KBD_FLAGS_EXTENDED		0x0100
+#define KBD_FLAGS_EXTENDED1		0x0200
 #define KBD_FLAGS_DOWN			0x4000
 #define KBD_FLAGS_RELEASE		0x8000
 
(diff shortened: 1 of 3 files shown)


kubistika:
That's what I thought you mean, but as you see you added the flag `KBDFLAGS_EXTENDED1` to fastpath while according to the spec it is used in slowpath. Moreover, the same change should be applied in `input.c: input_recv_keyboard_event`

Thanks for the help btw :)

akallabeth:
@kubistika Ah, ok now I see your point.
We're using the slow path defines also as the `public` export so you only have to check for these flags.
No need for a change in `input_recv_keyboard_event` as there the flag is already set (client sends it, correct?)

akallabeth:
@kubistika when sending the received data from `KeyboardEvent` again it is converted back to the flags used in `input_send_fastpath_keyboard_event` or `input_send_keyboard_event`
No need for special handling of the pause sequence here to proxy it (just pass it on?)

kubistika:
@akallabeth Oh... Well that's the missing part for me. Sorry for the misunderstanding and thank you!

PR #5404, opened 2019-05-19
Proxy: Redirect errors back to client

kubistika:
This PR adds support for errors redirection to freerdp-proxy.


akallabeth:
@freerdp-bot test

PR #5411, opened 2019-05-21
Bugfix 4934

chrislong:
Fix for issue #4934 (In seamless mode, moving a window shrinks it 14x7). Replaces PR #5395.


akallabeth:
Please run the formatter script on changed files and revert the public header changes properly (the variable order changed ;))

akallabeth:
@chrislong doing a `git checkout origin/master -- include/freerdp/window.h` should revert the file. just need to commit that then again.

akallabeth:
@freerdp-bot test

akallabeth:
@chrislong oh, and you need to stay C compliant (the older one, so variable declarations are only allowed at the beginning of a scope

hardening:
@freerdp-bot hey bot, can you test again ?

akallabeth:
@hardening will not build because of in code variable declaration ;)

hardening:
@chrislong looking forward for this PR, but please fix build errors and rebase your branch (we've merged a branch that make rail logs more verbose and conflicts with yours)

libfreerdp/core/window.c
```diff
@@ -32,6 +32,123 @@
...
+		if (flags & WINDOW_ORDER_VALUES[i])
+		{
+			totalLength += strlen(WINDOW_ORDER_STRINGS[i]) + 1;
+		}
+	}
+
+	if (totalLength > 0)
+	{
+		s = malloc(totalLength);
+		char* next = s;
```
akallabeth: C violation, must be declared at beginning of scope

include/freerdp/window.h
```diff
@@ -194,6 +194,10 @@ struct _WINDOW_STATE_ORDER
 	INT32 clientOffsetY;
 	UINT32 clientAreaWidth;
 	UINT32 clientAreaHeight;
+	UINT32 resizeMarginLeft;
+	UINT32 resizeMarginRight;
```
akallabeth: pleaase revert to original

chrislong:
Ok, I fixed the variables order and the variable declaration, and rebased.

I saw the formatting system has changed and ran `cmake --build . --target clangformat`. I got over 1300 files changed. I tried it on the head from freerdp/master and got the same thing. Do you want me to reformat and include over 1300 reformatted files as part of this PR?

akallabeth:
@chrislong thank you for the fix.
We already merged the new formatting conventions, but will enforce them at a later time. Please don't commit these changes.

chrislong:
@akallabeth Ok, I didn't think you'd want that many files changed. :) Let me know if there are any remaining issues.

chrislong:
FYI, I've now realized this fix isn't as good as I originally thought. There are 3 issues:

1. It does not work on all windows, specifically some dialog boxes, because they don't report window margins to the client.
1. Dialog boxes and menus show a drawing artifact: a gray box on their left side. It does not appear to affect functionality, just look weird.
1. Mouse registration is slightly off (I think by the margin amount, 7 pixels). This seems to affect all mouse events. Since the delta is small, I don't think it's a big deal, but it is a bug.

I am working to address all of these. Unfortunately, the documentation about RDP and the margin is scant, so I need to experiment before I know how to fix them.

I think this PR is still worth merging as-is because on balance I think the program is more usable with it than without. But if you want to wait, I understand.

Let me know if you want me to push fixes for the above issues to this branch and PR or a new one.

chrislong:
It didn't take as long to fix those issues as I expected, so I pushed them to this branch for this PR. I think it's all good now. If there are issues, just let me know.

akallabeth:
@freerdp-bot test

client/X11/xf_rail.c
```diff
@@ -269,90 +303,153 @@ void xf_rail_paint(xfContext* xfc, INT32 uleft, INT32 utop, UINT32 uright,
 
 /* RemoteApp Core Protocol Extension */
 
-static BOOL xf_rail_window_common(rdpContext* context,
-                                  const WINDOW_ORDER_INFO* orderInfo, const WINDOW_STATE_ORDER* windowState)
+static BOOL resizeMarginXSet = FALSE;
```
akallabeth: Please do not use static variables for this, add it to the rail context.

client/X11/xf_rail.c
```diff
@@ -269,90 +303,153 @@ void xf_rail_paint(xfContext* xfc, INT32 uleft, INT32 utop, UINT32 uright,
...
-			return FALSE;
+	if (!appWindow)
+		return NULL;
 
-		appWindow->dwStyle = windowState->style;
-		appWindow->dwExStyle = windowState->extendedStyle;
+    appWindow->dwStyle = windowState->style;
+    appWindow->dwExStyle = windowState->extendedStyle;
+
+    if ((orderInfo->fieldFlags & WINDOW_ORDER_FIELD_RESIZE_MARGIN_X) == 0 && resizeMarginXSet) {
```
akallabeth: Kind of useless, `appWindow` is created using `calloc` which initialized everything with `0`

client/X11/xf_rail.c
chrislong: Ok, no problem. Do you mean the `_rail_client_context` (from `include/freerdp/client/rail.h`)? Or the `xfContext` (aka `xf_context`, from `client/X``/xfreerdp.h`)? Partly why I kept it static was I wasn't sure where it should go.

client/X11/xf_rail.c
chrislong: I don't understand which part is useless. Those lines (331-334) don't use any values from appWindow. `windowState->style` and `extendedStyle` aren't always 0, are they? (Since they come from the server I didn't think we could assume that, anyway.)

akallabeth: The static variables above are kind of useless.
You get the resize margins from the window state (an argument to this function)
If the flag is set, use it, otherwise just ignore the field

chrislong: Actually, now that I think about it some more, I don't think line 334 is needed at all. If the new window has its own margins, those will get set later (starting at 409) and override the ones set here, anyway. I'll clean it up.

In case it's not clear, the idea with `resizeMarginXSet` is to enable remembering the first margins that are encountered so they can be used as default margins for later windows that don't have any specified. Another approach would be to always use the most recent margins as the default for new windows that don't have them. I don't know which is best.

akallabeth: I wouldn´t do any client side magic as there are many types of windows that have no margins on purpose (think menus and tooltips, ...)

client/X11/xf_window.h
```diff
@@ -109,6 +109,10 @@ struct xf_app_window
 	INT32 clientOffsetY;
 	UINT32 clientAreaWidth;
 	UINT32 clientAreaHeight;
+    UINT32 windowLeftResizeMargin;
```
akallabeth: Might be nice to fix indentation here if you are at it ;)

chrislong: Sure, no problem. (My editor put in spaces instead of a tab.)

client/X11/xf_rail.c
chrislong: I would agree, except that unfortunately, things break without any client side magic. For example, if margins are only used for windows that report them, then the Font dialog box for the notepad application doesn't work right. It has the moving and resizing behavior that started this issue. If the client remembers the margin from the previous window that had one and reuses it for the Font dialog box, it works. I've seen the same behavior with a dialog box for view options in explorer.exe. It appears that the server expects margins to exist on dialog boxes but does not specify them in its messages to the client.

I understand your concern about possible effects of this on windows that perhaps should not have margins. Experimentally, though, there does not seem to be any ill effect on menus or tooltips.

akallabeth: you might want to take a look at how `ClientWindowMove` is used.
I don´t have the time currently as I have work to be done on some other projects, but what I can tell is that the server is sending the correct margins and if reported correctly does not start to resize the window.
Might need some delay magic as the `ClientWindowMove` should only be used if the window is moved without the server noticing (by window manager or similar) and if both, the server try to move and the client reports back too often there might be a mixup.

chrislong: I will look at `ClientWindowMove` and investigate whether delay magic might help. I have some questions about some of your comments:

1. "the server is sending the correct margins". How can I find out what the "correct" margins are?
1. "if reported correctly does not start to resize the window". Could you elaborate? If what is "reported correctly"? When using the nightly build, any time I move a window with the window manager or by grabbing the title bar, it shrinks.

akallabeth: 1. The margins that are announced for a window ID are the correct ones. No magic necessary (just like written in the protocol spec, default is no margin)
1. This is the exact issue we´re talking about here. 
  * During normal operation the server is placing the window where it should be (you resize with mouse or drag the title bar)
  * When dragging via window manager the server does not know where the window is and for this situation `ClientWindowMove` is responsible
  * `ClientWindowMove` needs to announce the window size with the margins
  * The server announces the winow size without margins (the margins must be manually added using the values previously announced)
So, the issue is somewhere one (or both) of two things happen:
1. The window size is announced or applied with the wrong margins
1. The `ClientWindowMove` is sending new window geometry while the server is still trying to apply other settings (network lag or similar)

chrislong: Here's what I observed with dialog boxes. This example uses the Font dialog box from notepad.exe, but I've also seen it with a view option dialog box from explorer.exe.

I'll attach the full output log, but for clarity I have only included what I think are the relevant bits below.

I started up with `xfreerdp /v:xxx /u:xxx /p:xxx /app:'c:\windows\notepad.exe' /span /log-filters:com.freerdp.client.x11:debug,com.freerdp.channels.rail.client:debug,com.freerdp.core.update:debug`.

The main window showed up with this output:
```
[12:20:01:166] [2272:2285] [DEBUG][com.freerdp.core.update] - WindowCreate windowId=0x131332 owner=0x0 [ex]style=<0x14cf0000, 0x110 visible thickframe border caption> show=current title clientOffsetX=851 clientOffsetY=825 resizeMarginLeft=7 resizeMarginRight=7 resizeMarginTop=0 resizeMarginBottom=7 windowOffsetX=425 windowOffsetY=387 windowClientDeltaX=426 windowClientDeltaY=438 windowWidth=842 windowHeight=351 windowRects=((0,0,842,351)) visibleOffsetX=425 visibleOffsetY=387 visibilityRects=((0,0,842,351))
```
Notice that margins are set.

I brought up the Font dialog box and got this output:
```
[12:32:37:102] [2272:2285] [DEBUG][com.freerdp.core.update] - WindowCreate windowId=0x655614 owner=0x20104 [ex]style=<0x94c800c4, 0x10101 popup visible border caption> show=current title clientOffsetX=835 clientOffsetY=1065 windowOffsetX=417 windowOffsetY=517 windowClientDeltaX=418 windowClientDeltaY=548 windowWidth=427 windowHeight=471 windowRects=((0,0,427,471)) visibleOffsetX=417 visibleOffsetY=517 visibilityRects=((0,0,427,471))
```
Notice that **no** margins are set.

I used `wmctrl` to move the dialog box window to (0,0):
```
[12:34:04:127] [2272:2285] [DEBUG][com.freerdp.client.x11] - updated internal geometry from X: 427x471+0+0
```

The window moved instantly but its size did not change, and no further output was generated until the window got focus, and then:
```
[12:35:40:857] [2272:2285] [DEBUG][com.freerdp.client.x11] - forwarding geometry from X/internal to RDP server: 427x471+0+0	(0,0), (427,471) {this is the ClientWindowMove}
[12:35:40:878] [2272:2285] [DEBUG][com.freerdp.core.update] - WindowUpdate windowId=0x655614 clientOffsetX=15 clientOffsetY=31 windowOffsetX=7 windowOffsetY=0 windowClientDeltaX=8 windowClientDeltaY=31 windowWidth=413 windowHeight=464 windowRects=((0,0,413,464)) visibleOffsetX=7 visibleOffsetY=0 visibilityRects=((0,0,413,464))
[12:35:40:878] [2272:2285] [DEBUG][com.freerdp.client.x11] - setting new internal & X geometry: 413x464+7+0	showState=5
[12:35:40:894] [2272:2285] [DEBUG][com.freerdp.client.x11] - updated internal geometry from X: 413x464+7+0
```
The client did not adjust the position or size of the window in the ClientWindowMove, because no margins were specified by the server for this window.

After the ClientWindowMove, the server immediately responded with a WindowUpdate that had a location and size that were different from the ClientWindowMove, even though no margin has been specified for this window. The amount of the difference is exactly the margins specified for the main window.

As shown in the log, over 30 seconds elapsed between operations, so I don't see a way to improve the situation by adding delay to the client somewhere.

I think the server is using the same margins for the dialog box as for the main window even though it does not send those margins to the client when the dialog box is created.

I am open to other explanations and/or suggested fixes.

chrislong: I tried to attach the full log but it's not working. I'll try again later. (Tried with Firefox and Chrome, both drag & drop and dialog box.)

akallabeth: hmm, you can only attach zip and image files ;)
As for the resizing, which os version are you connecting to?
the resize_margin flag is only supported with newer windows versions if I read the spec correctly.
might need to take the client area and offsets into account too.

chrislong: > hmm, you can only attach zip and image files ;)

I zipped it and tried that, but still doesn't work. Do I need some kind of permission to attach files to this issue tracker? I couldn't find anything helpful in the github help.

> As for the resizing, which os version are you connecting to?
> the resize_margin flag is only supported with newer windows versions if I read the spec correctly.

Windows Server 2016 Datacenter. Also, the server does send window margins for the main window.

> might need to take the client area and offsets into account too.

Hmmm. Sure, could be. Looking at the values the server reports for those, I don't understand what they mean or how they could help. The RDP spec is not very informative. Do you know of any documentation with more info?

chrislong:
I just heard from Microsoft support about the mysterious geometry differences between server & client for dialog boxes:

> It turns out that the reason the menu and dialog do not have resize margins reported is that they aren’t resizable and don’t have resize margins. However, the size difference is due to a thin border that the server doesn’t report on. We are investigating how this may be worked around.

What would you like to do with this PR while we wait to hear about this "thin border" issue? I'd prefer to merge it now with the "wrong" fix for dialog boxes and do a new PR later to update with the "right" dialog box fix once we find out what that is.

akallabeth:
I´d remove the magic borders. Might be a good idea to check the window state for being resizable instead before announcing geometry changes.

chrislong:
Ok. Good idea about checking the resizable window state.

chrislong:
Do you know what determines whether a window is resizable? I looked at the window styles and extended window styles and there isn't one for resize. Is it `WS_BORCER`?

chrislong:
More news from Microsoft tech support: There's a bug in the server. It should include resize margins for all windows that have them, even if they're not resizable. They're going to fix it and update the docs, and try to come up with a suggestion for what clients should do in the meantime. I guess this PR could wait until they have a suggestion for us, but I don't know if or when that will happen so I would rather not.

It seems to me the best fix for FreeRDP for now is to use the heuristic I've implemented (i.e., remembering the last resize margins and reusing them on windows that don't declare any). If the "remembered" margins should be in a struct instead of static, that's totally fine by me, although I could use guidance on which struct would be the best one to put them in. (I don't know which struct is "the rail context".)

If you would rather not include this heuristic in the main build, I would very much like to use `#ifdef`s and a new option in the `CMakeLists.txt` file so I can enable this feature for my builds. I want the best experience for our users and dialog boxes magically resizing and moving is not good.

BTW, it occurred to me that this shouldn't matter for menus since the issue only comes up if/when the user moves or resizes a window, and you can't do that with menus, at least as far as I can tell.

akallabeth:
@bmiklautz @hardening Looks like we have some decision to make here, what do you think?

akallabeth:
@chrislong I´ve been playing a bit with and without your pr, as well as with and without your margin hack.
Working better with your changes, but I can´t get the window to shrink with my current test machine (and master current master), you have a simle test setup? Don´t want to miss the fun ;)
(been moving the window locally by wm and it always stays the correct size)

chrislong:
@akallabeth What do you mean by "I can´t get the window to shrink"? Like drag the side or corner to resize? I've been unable to do that a lot of the time, also, but I'm not sure if that ever worked reliably. I figure one problem at a time. (If your local window manager has resizng commands of its own, that should work.)

Tachiorz:
This is actually fixes Options window in visual studio trying to move away to the right all by itself.

akallabeth:
@infinytum address the issues with this pr, e.g. fix the boundary hack to be useful for all kinds of windows, I see many glitches with menu windows here.
(there are already enough problems with rails, we don´t need to trade one kind for another)

hardening:
closing, too old

PR #5415, opened 2019-05-24
Fixed format security issues.

akallabeth:
reported by @giox069 
* Fixes the missing format string
* Simplifies `DUMP_APPEND` macro
* Ensures the buffers are `\0` terminated by providing only `sizeof(buffer) - 1` as length to print functions.


hardening:
Hum the previous security problem is not very clear, anyway, could you change the macro so that we don't have to pass `buffer` and `bufferSize` all the time ?

Issue #5400, opened 2019-05-16
Exporting Monitor Layout PDU


kubistika:
Hey!

While working on adding multimon support for the proxy (#5372), I saw that the code for sending monitor layout PDU is buried inside this method in `libfreerdp/core/activation.c`:
```c
BOOL rdp_server_accept_client_font_list_pdu(rdpRdp* rdp, wStream* s)
```

In order to achieve full multimon support in the proxy we need, of course, to forward the monitor layout PDU from the target server to the client. As of now, we can't do it because it is not exported.

I want to open a PR which exports a function that sends the monitor layout PDU, but don't know where would it be the best to export it from. What do you think that is the best way to export this logic?

kubistika:
#5416

PR #5416, opened 2019-05-26
Add multimon support to proxy server

kubistika:
This PR adds multimon support to the proxy.

* In order to send MonitorLayoutPdu, we had to export it. It was located in acitvation.c (which is not the location for this message, so we moved it to a new file display.c as the it appears in the docs under "Display update notifications"). Please refer to #5400.

### DIFF
--- a/libfreerdp/core/activation.c
+++ b/libfreerdp/core/activation.c
@@ -22,6 +22,7 @@
 #endif
 
 #include "activation.h"
+#include "display.h"
 
 /*
 static const char* const CTRLACTION_STRINGS[] =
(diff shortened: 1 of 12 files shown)


akallabeth:
@freerdp-bot test

libfreerdp/core/display.h
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef FREERDP_LIB_CORE_DISPLAY_H
+#define FREERDP_LIB_CORE_DISPLAY_H
+
+#include <freerdp/display.h>
+#include "rdp.h"
+
+FREERDP_LOCAL BOOL display_convert_rdp_monitor_to_monitor_def(UINT32 monitorCount,
+        const rdpMonitor* monitorDefArray, MONITOR_DEF** result);
+FREERDP_LOCAL BOOL rdp_write_monitor_layout_pdu(wStream* s, UINT32 monitorCount,
```
speidy: rm, as we don't have this func anymore

we use the exported one instead

speidy:
re run test?

server/proxy/pf_context.c
```diff
@@ -77,5 +77,45 @@ rdpContext* p_client_context_create(rdpSettings* clientSettings,
 	settings->ServerPort = port;
 	settings->SoftwareGdi = FALSE;
 	settings->RedirectClipboard = FALSE;
+	/* Client Monitor Data */
+	settings->MonitorCount = clientSettings->MonitorCount;
```
speidy: maybe we can move the monitor stuff to its own func.

speidy: or to` pf_common_copy_settings`

kubistika: not get your reasoning to move to common settings, as those settings are one-sided (only client settings). i'd like not to change it until will open a PR that uses `freerdp_settings_clone`, then we will remove it

akallabeth:
@freerdp-bot test

libfreerdp/core/activation.c
```diff
@@ -341,22 +342,25 @@ BOOL rdp_server_accept_client_font_list_pdu(rdpRdp* rdp, wStream* s)
 	    peer->AdjustMonitorsLayout(peer))
 	{
 		/* client supports the monitorLayout PDU, let's send him the monitors if any */
-		wStream* st = rdp_data_pdu_init(rdp);
-		BOOL r;
+		MONITOR_DEF* monitors = (MONITOR_DEF*) calloc(settings->MonitorCount, sizeof(MONITOR_DEF));
```
hardening: Just curious, can't we have `settings->MonitorCount == 0`, because in this case we will end up with  ̀calloc`  returning NULL (and so error path) ?

kubistika: don't understand your question, the if
```c
	if (settings->SupportMonitorLayoutPdu && settings->MonitorCount && peer->AdjustMonitorsLayout &&
	    peer->AdjustMonitorsLayout(peer))

```

won't allow monitor count to be 0
and in this case the monitor layout pdu won't be sent
isn't this ok?

rebased on master

kubistika:
@hardening rebase on master, can you re run tests?

hardening:
@freerdp-bot can you test ?

kubistika:
@hardening  @akallabeth  can we merge?:)

libfreerdp/core/activation.c
hardening: ok that was my question. Merging.

PR #5420, opened 2019-05-31
channels: ssh-agent: cmake: turn off server side channel by default in `ChannelsOptions.cmake`

speidy:
CMake v3.5.1 throws the following error if this is not set:
```
CMake Error at channels/CMakeLists.txt:48 (if):
  if given arguments:

    "OFF" "OR"

  Unknown arguments specified
  channels/sshagent/ChannelOptions.cmake:6 (define_channel_options)
  channels/CMakeLists.txt:273 (include)
```


akallabeth:
@freerdp-bot test

PR #5423, opened 2019-06-04
rdpUpdate: add autoBitmapDataCalculate

kubistika:
I found a bug in our proxy that occurs when the target server is win7:
the target sends bitmap update data with flags=0 and the server code changes it to 0x401 instead of sending it as is, then the client (mstsc) closes the connection (because it got an uncompressed bitmap with flags indicating that it is compressed, I guess)

I didn't want to break the API so I though of a way to fix it and wanted to hear what do you guys think of it
@akallabeth @hardening 

BTW, wasn't sure about the name `autoBitmapDataCalculate`, because it actually calculate only three members of the struct. I noted this in the comment, before the declaration of this struct member. So a new name for this member / a different idea for a solution will be welcomed

### DIFF
--- a/include/freerdp/update.h
+++ b/include/freerdp/update.h
@@ -253,6 +253,11 @@ struct rdp_update
 	rdpBounds currentBounds;
 	rdpBounds previousBounds;
 	CRITICAL_SECTION mux;
+
+	/* if autoCalculateBitmapData is set to TRUE, the server automatically 
+	 * fills BITMAP_DATA struct members: flags, cbCompMainBodySize and cbCompFirstRowSize.
+	*/
+	BOOL autoCalculateBitmapData;
 };
 
 #endif /* FREERDP_UPDATE_H */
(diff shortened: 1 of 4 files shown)


include/freerdp/update.h
```diff
@@ -243,6 +243,11 @@ struct rdp_update
 	rdpPcap* pcap_rfx;
 	BOOL initialState;
 
+	/* if autoCalculateBitmapData is set to TRUE, the server automatically 
+	 * fills BITMAP_DATA struct members: flags, cbCompMainBodySize and cbCompFirstRowSize.
+	*/
+	BOOL autoCalculateBitmapData;
+
```
hardening: would be nice to move that new field at the end of the struct for binary compatibility

hardening:
@freerdp-bot can you test

kubistika:
I'll fix the rebase later, sorry for that
@hardening i moved it to the end of the struct

kubistika:
@hardening fixed it, can you re run tests?

akallabeth:
@freerdp-bot test

kubistika:
@hardening  your opinion?

kubistika:
@akallabeth @hardening can we merge?

PR #5435, opened 2019-06-14
primitives: add openCL support and autodection of best implementation

hardening:
This PR adds support for OpenCL in primitives, for now only YUVtoRGB primitives are implemented.


akallabeth:
@hardening did some cleanups here: https://github.com/akallabeth/FreeRDP/tree/opencl
One thing that may improve speed a lot is implementing `YUV444ToRGB_8u_P3AC4R` as `YUV420ToRGB_8u_P3AC4R` is currently only used by the video redirection (`AVC420` is more or less dead)

libfreerdp/primitives/primitives.c
```diff
@@ -23,65 +24,483 @@
...
+
+	if (!gotGPU)
+	{
+		WLog_ERR(TAG, "openCL: no GPU found");
+		return FALSE;
+	}
+
+	snprintf(sourcePath, sizeof(sourcePath), "%s/primitives.cl", OPENCL_SOURCE_PATH);
+
+	f = fopen(sourcePath, "r");
```
akallabeth: Why import that at runtime? While nice to allow changing the implementation it is a hindrance if you compile a single binary (like `wfreerdp.exe`)

hardening: well to my knowledge there's no way to have a precompiled version as it depends on the driver, it's like a shader. The other solution would be to embed the source in the freerdp binary...

akallabeth: this could, of course, be done optionally ;)

akallabeth:
@hardening any update on this? would love to merge ;)

hardening:
> `FindOpenCL.cmake` is shipped with `CMake` 3.2 and above.
> The variables differ to your script though, so maybe it would be good to fix that.
> https://cmake.org/cmake/help/latest/module/FindOpenCL.html

Unfortunately for now the minimum cmake version that we require is 2.8. Do you suggest changing my file so that it match the shipped script ?

akallabeth:
@hardening would use the variables of the module available since `3.2` and add a check to use the cmake provided version for these cases. (maybe like the `cmake/compat*` folders?)

libfreerdp/primitives/primitives.c
```diff
@@ -23,72 +24,377 @@
...
 }
+
+primitives_t* primitives_get_by_type(DWORD type)
+{
+	InitOnceExecuteOnce(&generic_primitives_InitOnce, primitives_init_generic_cb, NULL, NULL);
+	switch (type)
+	{
+#if defined(WITH_OPENCL)
+		case PRIMITIVES_ONLY_GPU:
+			InitOnceExecuteOnce(&gpu_primitives_InitOnce, primitives_init_cpu_cb, NULL, NULL);
```
akallabeth: should be `primitives_init_gpu_cb`

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #5458, opened 2019-07-08
Declare support for multimon in proxy server.

kubistika:
Unfortunately, by mistake, in the DISP PR the multimon was broken (the `UseMultimon` was set to FALSE, then `gcc_write_client_monitor_data` wasn't called).
This PR fixes it.


akallabeth:
@freerdp-bot test

PR #5471, opened 2019-07-11
Bugfix/objectpool

kubistika:
Came across this while testing our proxy. Discussed with @akallabeth


akallabeth:
@freerdp-bot test

PR #5472, opened 2019-07-15
Rdp parser fixes

akallabeth:


mfleisz:
Just noticed that setting the `RDP_FILE_LINE_FLAG_STANDARD` flag in `freerdp_client_rdp_file_set_integer` is broken (it is always set).

akallabeth:
@m4ntis @speidy @kubistika Could you have a look on the proxy changes?
Using standard settings clone instead of a custom one now to avoid code duplication (and issues if settings are added)

kubistika:
@akallabeth I'll check it, but please don't merge this PR yet, because when we tried using `freerdp_settings_clone` instead of our custom function, the proxy's client failed to connect (was a problem with some fields that shouldn't be copied, something with certificates)

akallabeth:
@kubistika Ah, ok. If you have issues with that I´d suggest doing it the opposite way, using `freerdp_settings_clone` and resetting what you don´t need

kubistika:
@akallabeth yes, that's what we need to do.
The problem is that that's not so easy (we need to reset something, check if its working, if not, check what should be reset and try again), and I'm working on the proxy alone for now

mfleisz:
@kubistika Is it ok for you if we merge this one (without the proxy changes)?

akallabeth:
@mfleisz see irc, tracking the clone issue here: https://github.com/FreeRDP/FreeRDP/issues/5474

mfleisz:
@akallabeth alright merging - thx!

PR #5473, opened 2019-07-15
server: proxy: implement NLA fallback

kubistika:
This PR contains a feature we called "NLA Fallback", which means that if the proxy's client fails to connect with NLA, it then tries to connect without NLA (using only TlsSecurity). We need this because there's a domain policy that disables NTLM then there are some computers that the proxy can't connect to them (because it tries NLA and fails).

### DIFF
--- a/server/proxy/pf_client.c
+++ b/server/proxy/pf_client.c
@@ -208,7 +208,14 @@ static void pf_client_post_disconnect(freerdp* instance)
 	PubSub_UnsubscribeErrorInfo(instance->context->pubSub, pf_OnErrorInfo);
 	gdi_free(instance);
 
-	SetEvent(pdata->connectionClosed);
+	if (!context->during_connect_process)
+	{
+		/* proxy's client failed to connect, and now it's trying to connect without NLA, no need to shutdown
+		 * the connection between proxy's server and the original client.
+		 */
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

server/proxy/pf_client.c
```diff
@@ -222,16 +229,37 @@ static void pf_client_post_disconnect(freerdp* instance)
...
 	if (!freerdp_connect(instance))
 	{
-		WLog_ERR(TAG, "connection failure");
-		return 0;
+		if (instance->settings->NlaSecurity)
+		{
+			WLog_ERR(TAG, "freerdp_connect() failed, trying to connect without NLA");
+			/* disable NLA, enable TLS */
+			instance->settings->NlaSecurity = FALSE;
+			instance->settings->TlsSecurity = TRUE;
```
speidy: Should we explicitly enable RdpSecurity too?

kubistika: it is already set to true, but for the code to be more readable, yes i'll add it

kubistika:
@akallabath can you please test again? :)

akallabeth:
@freerdp-bot test

kubistika:
@akallabeth lets merge  :)

PR #5475, opened 2019-07-15
Pass on cert validation failure, set freerdp error in all use cases.

akallabeth:


akallabeth:
@mfleisz no, the change is mostly to allow `VerifyX509Certificate` to return an error (only with `ExternalCertificateManagement`)

akallabeth:
@mfleisz additionally `verification_status`  is set to `-1` in case of abort if `ExternalCertificateManagement` is not set.

akallabeth:
There are 2 places. The other one is for connect.

PR #5476, opened 2019-07-15
server: proxy: Use freerdp_settings_clone

### LINKED ISSUE #5474: server: proxy: use freerdp_settings_clone() instead of a custom function
As discussed with @akallabeth, instead of using a custom function to copy the client settings, we should use the library function `freerdp_settings_clone`. The problem is that it copies some fields that shouldn't be copied, for example, the certs.

kubistika:
This PR removes the custom function that the proxy used in order to copy the settings struct.
In order to use it, the proxy must revert some fields (regarding certificates, and `ServerMode` field), so it does it in a function `pf_context_copy_settings` that underline calls `freerdp_settings_clone`.

Important note: because we're cloning the settings in runtime, there are some pointers to the settings struct that remains old (for example, rdpRdp* has a pointer to rdpSettings*). The proxy couldn't update them itself because rdpRdp* is not exported, so I created and implemented a function named `freerdp_update_settings` that updates all necessary pointers (not sure about her name).

@akallabeth hope it made you happy as it made me ;)
fixes #5474

### DIFF
--- a/libfreerdp/core/certificate.c
+++ b/libfreerdp/core/certificate.c
@@ -826,6 +826,44 @@ out_free:
 	return NULL;
 }
 
+rdpRsaKey* key_clone(const rdpRsaKey* key)
+{
+	rdpRsaKey* _key = (rdpRsaKey*) calloc(1, sizeof(rdpRsaKey));
+
+	if (!_key)
+		return NULL;
+
+	CopyMemory(_key, key, sizeof(rdpRsaKey));
+
+	if (key->Modulus)
+	{
+		_key->Modulus = (BYTE*) malloc(key->ModulusLength);
+
+		if (!_key->Modulus)
+			goto out_fail;
+
+		CopyMemory(_key->Modulus, key->Modulus, key->ModulusLength);
+	}
+
+	if (key->PrivateExponent)
+	{
+		_key->PrivateExponent = (BYTE*) malloc(key->PrivateExponentLength);
+
+		if (!_key->PrivateExponent)
+			goto out_fail;
+
+		CopyMemory(_key->PrivateExponent, key->PrivateExponent, key->PrivateExponentLength);
+	}
+
+	return _key;
+out_fail:
+	free(_key->Modulus);
+	free(_key->PrivateExponent);
+	free(_key);
+	return NULL;
+}
+
+
 void key_free(rdpRsaKey* key)
 {
 	if (!key)
--- a/include/freerdp/settings.h
+++ b/include/freerdp/settings.h
@@ -1559,6 +1559,7 @@ extern "C" {
 
 FREERDP_API rdpSettings* freerdp_settings_new(DWORD flags);
 FREERDP_API rdpSettings* freerdp_settings_clone(rdpSettings* settings);
+FREERDP_API BOOL freerdp_settings_copy(rdpSettings* dst, const rdpSettings* src);
 FREERDP_API void freerdp_settings_free(rdpSettings* settings);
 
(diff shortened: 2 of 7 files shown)


akallabeth:
@freerdp-bot test

kubistika:
@akallabeth the change in filter_demo.c should be reverted, I'll do it when I'll be on my PC

kubistika:
@akallabeth fixed it

akallabeth:
@freerdp-bot test

include/freerdp/freerdp.h
```diff
@@ -492,6 +492,7 @@ FREERDP_API void setChannelError(rdpContext* context, UINT errorNum,
 FREERDP_API BOOL checkChannelErrorEvent(rdpContext* context);
 
 FREERDP_API const char* freerdp_nego_get_routing_token(rdpContext* context, DWORD* length);
+FREERDP_API void freerdp_update_settings(rdpContext* context);
```
akallabeth: after looking through all that stuff I´d prefer a function `BOOL freerdp_settings_copy(rdpSettings* dst, const rdpSettings* src)` instead of this one. (don´t like the pointer magic, got a feeling that this is too error prone)

server/proxy/filters/filter_demo.c
```diff
@@ -33,7 +33,7 @@ static PF_FILTER_RESULT demo_filter_mouse_event(connectionInfo* info, void* para
 
 	if (event_data->x % 100 == 0)
 	{
-		return FILTER_DROP;
+		return FILTER_PASS;
```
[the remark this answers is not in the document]
kubistika: i already reverted this change

libfreerdp/core/certificate.c
```diff
@@ -826,6 +826,46 @@ rdpRsaKey* key_new(const char* keyfile)
...
 }
 
+rdpRsaKey* key_clone(const rdpRsaKey* key)
+{
+	rdpRsaKey* _key = (rdpRsaKey*) calloc(1, sizeof(rdpRsaKey));
+
+	if (!_key)
+		return NULL;
+
+	CopyMemory(_key, key, sizeof(rdpRsaKey));
```
akallabeth: easier is using `*_key = *key` ;)

libfreerdp/core/certificate.c
```diff
@@ -826,6 +826,46 @@ rdpRsaKey* key_new(const char* keyfile)
...
+
+	if (key->Modulus)
+	{
+		_key->Modulus = (BYTE*) malloc(key->ModulusLength);
+
+		if (!_key->Modulus)
+			goto out_fail;
+
+		CopyMemory(_key->Modulus, key->Modulus, key->ModulusLength);
+		_key->ModulusLength = key->ModulusLength;
```
akallabeth: Not necessary, already done on struct copy

libfreerdp/core/certificate.c
```diff
@@ -826,6 +826,46 @@ rdpRsaKey* key_new(const char* keyfile)
...
+
+	if (key->PrivateExponent)
+	{
+		_key->PrivateExponent = (BYTE*) malloc(key->PrivateExponentLength);
+
+		if (!_key->PrivateExponent)
+			goto out_fail;
+
+		CopyMemory(_key->PrivateExponent, key->PrivateExponent, key->PrivateExponentLength);
+		_key->PrivateExponentLength = key->PrivateExponentLength;
```
akallabeth: Not necessary, already done on struct copy

akallabeth:
@freerdp-bot test

kubistika:
@akallabeth can we merge?

PR #5480, opened 2019-07-17
Fix broken clipboard after file copy

akallabeth:
thx to amosbird in IRC for pointing out the error cause.


akallabeth:
@mfleisz the error was the premature return

mfleisz:
@akallabeth Yes I know - but I meant the `DstSize = 0;` line ... :-)

PR #5481, opened 2019-07-17
Extract whole certificate chain to PEM format.

akallabeth:
PEM extraction used as argument for `VerifyX509Certificate` did not extract intermediate certificates.
This way no proper validation was possible until now.


akallabeth:
@freerdp-bot test

PR #5487, opened 2019-07-18
Fixed a bunch of conversion warnings.

akallabeth:


client/common/file.c
```diff
@@ -199,143 +199,146 @@ struct rdp_file
  *
  */
 
-static int freerdp_client_rdp_file_set_integer(rdpFile* file, const char* name, int value,
-                                               int index)
+static BOOL freerdp_client_rdp_file_set_integer(rdpFile* file, const char* name, long value,
+											   SSIZE_T index)
 {
-	int standard = 0;
+	BOOL standard = FALSE;
```
mfleisz: The name of this var is quite misleading as it does exactly the opposite ^^

client/common/file.c
```diff
@@ -896,195 +913,117 @@ static size_t freerdp_client_write_setting_to_buffer(char** buffer, size_t* buff
...
-		res = freerdp_client_write_setting_to_buffer(&buffer, &size, fmt_, param_); \
+		SSIZE_T res = freerdp_client_write_setting_to_buffer(&buffer, &size, fmt_, param_); \
 		if (res < 0)                                                                \
-			return -1;                                                              \
-		totalSize += res;                                                           \
+			return 0;                                                              \
+		totalSize += (size_t)res;                                                           \
 	}
 
+#define WRITE_SETTING_INT(fmt_, param_) \
```
mfleisz: Good one  👍

client/common/file.c
akallabeth: Ok, fixed that (including description)

client/common/file.c
```diff
@@ -195,176 +195,179 @@ struct rdp_file
 /*
  * Set an integer in a rdpFile
  *
- * @return 0 if a standard name was set, 1 for a non-standard name, -1 on error
+ * @return FALSE if a standard name was set, TRUE for a non-standard name, -1 on error
```
mfleisz: Function won't return -1

client/common/file.c
```diff
@@ -1740,7 +1690,7 @@ int freerdp_client_rdp_file_set_integer_option(rdpFile* file, const char* name,
 			return -1;
 		}
 
-		if (freerdp_client_rdp_file_set_integer(file, (char*)name, value, index) < 0)
+		if (freerdp_client_rdp_file_set_integer(file, (const char*)name, value, index) < 0)
```
mfleisz: This is now a BOOL function and won't return values `< 0`

PR #5492, opened 2019-07-21
server: proxy: fix rdpgfx close race

kubistika:
This fixes a race that caused the proxy to crash:
- Client started channels termination, and raised OnClose callback
- pf_rdpgfx_on_close was called from client thread, which caused Close to be called on the server side
- pf_server called rdpgfx_server_context_free when the connection was terminated

Both client & server threads tried to call zgfx_context_free then a race caused a double free and a segfault to occur.


bmiklautz:
@freerdp-bot test

kubistika:
@bmiklautz rebased on master

bmiklautz:
@freerdp-bot test.

PR #5500, opened 2019-07-25
wlfreerdp: Bring Wayland support into a working condition

kennylevinsen:
wlfreerdp was unable to start. This was tested under sway, but would be the case for any compositor that adheres to the spec.

This PR brings wlfreerdp into a working state (i.e. can open an RDP connection, paint and interact with it), fixing the following issues:
1. Surface roles were never configured due to not ack'ing configure on our xdg_surface. This caused a window to never appear.
2. pendingBuffer was incorrectly set to a non-NULL value on configure, causing painting to bail as a non-NULL value indicated that a paint has occurred and that it should wait on a frame callback that would never arrive to clear the value. This caused the window to never be painted to.
3. Re-enable damage tracking code that was `#if 0`'d. This ensures that we're not giving the compositor unnecessary work.

### DIFF
--- a/client/Wayland/wlfreerdp.c
+++ b/client/Wayland/wlfreerdp.c
@@ -114,7 +114,7 @@ static BOOL wl_update_buffer(wlfContext* context_w, INT32 ix, INT32 iy, INT32 iw
 	if (UwacWindowAddDamage(context_w->window, x, y, w, h) != UWAC_SUCCESS)
 		return FALSE;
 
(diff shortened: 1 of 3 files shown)


kennylevinsen:
A few nice-to-haves are included now.

bmiklautz:
@freerdp-bot test.

akallabeth:
@kennylevinsen Nice one, thank you :)

PR #5509, opened 2019-08-01
server: proxy: sync channels close

### LINKED ISSUE #5495: server: proxy: crashes when target closes the connection
In short, the reason for the crash is that in a case when proxy's client disconnects before that the server shuts down the connection, the server doesn't know that the client has disconnected. Then the server tries to send messages to the target, using the proxy's client, which is already free()ed.

## Deeper explanation

A reasonable scenario that causes the crash is:
* A client connects to a TS broker through the proxy.
* The broker redirects proxy's client to a single TS server, which closes the RDP connection with an error about low memory.
* `rdp_client_disconnect` is called in proxy's client thread (and frees client's stuff, its channels etc).
* Meanwhile, the server thread which doesn't know yet that the client exited (`PostDisconnect` callback wasn't called yet!), keeps receiving messages from its client, and tries to send them to the target using the proxy's client, which at this point, is already freed.

## Proposal for a solution
* Add a `PreDisconnect` callback to `rdp_freerdp` struct, which will be called at the very begining of `rdp_client_disconnect`.
* When this callback will be called, set `pdata->abort_event`, which will cause the server thread to close the connection with its client before proxy's client will be disconnected. Then, before returning from the callback, wait for an acknowledgement from the server (can be implemented as another event, called `server_finished` or something. This fixes the problem by making sure that before proxy's client disconnection, the proxy's server will not get any more messages and won't even need to use the proxy's client.

@akallabeth @hardening if this sounds okay to you, please let me know how should I add the function pointer to the callback in `rdp_freerdp` struct (I saw some alignment stuff and comments about offsets).

kubistika:
fixes #5495

- Whenever a client channel is closed, the suitable server channel is closed too
    - Also fixes TS redirection that caused a crash
- Before the server free()s its channels, it waits for client termination

### DIFF
--- a/server/proxy/pf_channels.c
+++ b/server/proxy/pf_channels.c
@@ -41,14 +41,13 @@
 
 #define TAG PROXY_TAG("channels")
 
-void pf_OnChannelConnectedEventHandler(void* context,
+void pf_OnChannelConnectedEventHandler(void* data,
                                        ChannelConnectedEventArgs* e)
 {
-	pClientContext* pc = (pClientContext*) context;
+	pClientContext* pc = (pClientContext*) data;
 	pServerContext* ps = pc->pdata->ps;
-	RdpgfxClientContext* gfx;
-	RdpgfxServerContext* server;
-	WLog_DBG(TAG, "Channel connected: %s", e->name);
+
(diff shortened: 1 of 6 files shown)


akallabeth:
@freerdp-bot test

PR #5511, opened 2019-08-02
client: xf_channels: remove unused settings variable

kubistika:
Can't and don't want to see a compiler warning about this even one more time ):


akallabeth:
@freerdp-bot test

PR #5512, opened 2019-08-03
Proxy/config loader refactor

kubistika:
## Regarding configuration

- Config loading code is cleaner.
- Only open allowed channels (do not allow disp if DisplayControl is not allowed in config).

## Regarding filters
- Make sure that `filter_init` is exported on filters compilation.
- Instead of dropping the connection in filter_demo, just print some info to stdout.

### DIFF
--- a/server/proxy/config.ini
+++ b/server/proxy/config.ini
@@ -18,19 +18,14 @@ Port = 3389
 Mouse = 1
 Keyboard = 1
 
-[Graphics]
-GFX = 1
-BitmapUpdate = 1
-
 [Security]
 NlaSecurity = 0
 TlsSecurity = 1
 RdpSecurity = 1
 
 [Channels]
-WhitelistMode = 0
-AllowedChannels = "cliprdr,Microsoft::Windows::RDS::Video::Control"
-DeniedChannels = "Microsoft::Windows::RDS::Geometry"
+GFX = 1
+DisplayControl = 1
 
 [Filters]
 ; FilterName = FilterPath
--- a/server/proxy/filters/filters_api.h
+++ b/server/proxy/filters/filters_api.h
@@ -22,7 +22,10 @@
 #ifndef FREERDP_SERVER_PROXY_FILTERS_API_H
 #define FREERDP_SERVER_PROXY_FILTERS_API_H
 
+#include <freerdp/freerdp.h>
 #include <winpr/winpr.h>
+
+#define PROXY_API FREERDP_API
 
 enum pf_filter_result {
     FILTER_PASS = 0,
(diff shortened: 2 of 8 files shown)


server/proxy/config.ini
```diff
@@ -18,19 +18,14 @@ Port = 3389
 Mouse = 1
 Keyboard = 1
 
-[Graphics]
-GFX = 1
-BitmapUpdate = 1
```
speidy: we don't want that one anymore?

server/proxy/pf_config.c
```diff
@@ -30,195 +30,143 @@
...
 #define TAG PROXY_TAG("config")
 
-#define CHANNELS_SEPERATOR ","
+static inline BOOL pf_config_get_bool(wIniFile* ini, const char* section, const char* key)
+{
+	return IniFile_GetKeyValueInt(ini, section, key);
+}
 
-wArrayList* parse_string_array_from_str(const char* str)
+static BOOL pf_config_get_uint16(wIniFile* ini, const char* section, const char* key, UINT16* port)
```
speidy: port -> value

server/proxy/config.ini
kubistika: nope, gfx moved to channels section and we never used BitmapUpdate value

server/proxy/pf_config.c
kubistika: just a second after you comment, renamed to result

server/proxy/filters/filters_api.h
```diff
@@ -22,6 +22,20 @@
 #ifndef FREERDP_SERVER_PROXY_FILTERS_API_H
 #define FREERDP_SERVER_PROXY_FILTERS_API_H
 
+#if defined _WIN32 || defined __CYGWIN__
```
hardening: IIRC we have something equivalent somewhere else, can you check that  ?

hardening:
@freerdp-bot can you test ?

hardening:
side note: VS2010 shows some interesting warnings that may/should be fixed

server/proxy/filters/filters_api.h
kubistika: We have FREERDP_API/WINPR_API
Maybe i can #define PROXY_API FREERDP_API

kubistika:
@hardening changed the definition of PROXY_API and changed the inline function to be a macro, that should fix the msbuild errors

could you test again?

server/proxy/filters/filters_api.h
hardening: sounds nice

kubistika: ok please test again

hardening:
@freerdp-bot can you test ?

kubistika:
@hardening can we merge this one?

PR #5513, opened 2019-08-05
server: proxy: cliprdr support

kubistika:
This one adds `cliprdr` support to the proxy, allowing to decide whether to allow files copy or not, and limit text size copy.

- Added autoInitializationSequence for the proxy to be able to sync cliprdr caps between client and target (like we did in GFX).
- Changed the implementation of server_recv_capabilities to call the ClientCapabilities callback, for the proxy to be able to forward the caps.
- Added `lastRequestedFormatId` for both client & server, so the proxy will know which format is requested in every FormatDataResponse PDU (to be able to filter it). It is quite well documented in the commit message.

### DIFF
--- a/channels/cliprdr/client/cliprdr_format.c
+++ b/channels/cliprdr/client/cliprdr_format.c
@@ -295,7 +295,7 @@ UINT cliprdr_process_format_data_request(cliprdrPlugin* cliprdr, wStream* s, UIN
 
 	Stream_Read_UINT32(s, formatDataRequest.requestedFormatId); /* requestedFormatId (4 bytes) */
 
-
+	context->lastRequestedFormatId = formatDataRequest.requestedFormatId;
 	IFCALLRET(context->ServerFormatDataRequest, error, context, &formatDataRequest);
 	if (error)
 		WLog_ERR(TAG, "ServerFormatDataRequest failed with error %"PRIu32"!", error);
--- a/channels/cliprdr/client/cliprdr_main.c
+++ b/channels/cliprdr/client/cliprdr_main.c
@@ -184,6 +184,7 @@ static UINT cliprdr_process_general_capability(cliprdrPlugin* cliprdr,
 		return ERROR_INTERNAL_ERROR;
 	}
 
+	capabilities.msgType = CB_CLIP_CAPS;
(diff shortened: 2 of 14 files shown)


kubistika:
rebased on master

include/freerdp/channels/cliprdr.h
```diff
@@ -116,14 +116,14 @@ typedef struct _CLIPRDR_HEADER CLIPRDR_HEADER;
 struct _CLIPRDR_CAPABILITY_SET
 {
 	UINT16 capabilitySetType;
-	UINT16 capabilitySetLength;
```
akallabeth: I don´t like renaming symbols in public headers for no apparent reason.

kubistika: Then I'll revert. Just noticed that this is inconsistent with the documentation.

akallabeth:
@kubistika had a look at the server side channel and fixed it here: https://github.com/akallabeth/FreeRDP/tree/clip_server_const
you might want to cherry pick and add it to your pr.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@kubistika updated my pr, now only printing a warning that the parameter has the wrong `msgType` and using the expected one automatically. Working well so far here.

akallabeth:
@freerdp-bot test

channels/cliprdr/client/cliprdr_format.c
```diff
@@ -291,11 +291,11 @@ UINT cliprdr_process_format_data_request(cliprdrPlugin* cliprdr, wStream* s, UIN
 
 	formatDataRequest.msgType = CB_FORMAT_DATA_REQUEST;
 	formatDataRequest.msgFlags = msgFlags;
-	formatDataRequest.dataLen = dataLen;
+	formatDataRequest.dataLen = 4;
```
akallabeth: Please check my updated pr, this should not be necessary.

kubistika:
@hardening do you want to have a look?

kubistika:
@akallabeth lets merge?

PR #5515, opened 2019-08-06
server: proxy: declare GlyphSupportLevel

kubistika:
- Fixes a bug that causes the connection to be closed (after client <> proxy server negotiation, GlyphSupportLevel was updated to be 0x3, then the client, who's settings are copied from the server's settings, connected with GlyphSupportLevel = 0x3, instead of 0x0 (GLYPH_SUPPORT_NONE - we currently do not support GDI orders [`OrderSupport` is zeroed], so the library checks caused the connection to be closed).
- Fixes a bug introduced in #5509: server dynamic channel (for eg. GFX) may be opened before server's dynvc is in ready state.

### DIFF
--- a/server/proxy/pf_client.c
+++ b/server/proxy/pf_client.c
@@ -90,6 +90,8 @@ static void pf_OnErrorInfo(void* ctx, ErrorInfoEventArgs* e)
  */
 static BOOL pf_client_pre_connect(freerdp* instance)
 {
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

kubistika:
@akallabeth can we merge this one? other PRs depend on this one

Issue #5514, opened 2019-08-05
Sound skips and rattles heavily from Windows to Linux (xrdp) or Windows Server (RDS)


vscfreire:
Using the latest FreeRDP 2 builds (build 648) and connecting from Windows to either Linux (xrdp 0.9.10) or Windows Server (RDS 2016) results in sound skipping and rattling frequently. This problem is present for some time now.

I’m using the following command for the connection with the Windows and Linux server:
wfreerdp.exe /v:hostname /cert-tofu /u:Administrator /p:password /sound:format:0x0001 /w:1000 /h:600

Several optimization tactics, like sound formats, quality, latency and network detection were used with very limited success, usually only delaying the sound skipping and rattling.

Also, using sound formats (e.g. WAVE_FORMAT_MPEGLAYER3 0x0055) other than PCM (0x0001) while connecting to Linux (xrdp 0.9.10) server will result in client immediately exiting when playing any kind of sound with [com.freerdp.client.common] - Network disconnect.  Debugging shows “rdpsnd_virtual_channel_client_thread reported an error. Error was 13”

akallabeth:
@vscfreire
* `for some time now` is how long? last commit there was ages back. (we´d appreciate some windows maintainer/tester for early feedback)
* Could you describe your setup with a bit more detail? I sometimes use the windows client and did not have that particular problem (but most of the time just listen to some video in background)
* Your 'optimization` is not really one, better not do that (`format` option that is). (most of the time that only leads to more issues, should really remove that option ;))
* the `xrdp` `mp3` abort is expected, you can´t use a codec that is not compiled into the client. (and it is part of the `EXPERIMENTAL` ones, so they are buggy anyway) (not really a bug as the only way you can force this is by forceful command line override)

akallabeth:
@vscfreire just tried the current nightly build https://ci.freerdp.com/job/freerdp-nightly-windows/arch=win64,label=vs2013/lastSuccessfulBuild/artifact/build/Release/wfreerdp.exe against various servers (rds and direct connections) and did not experience rattling or similar.
Could you provide a short description of your test routine to allow reproducing this?

akallabeth:
@vscfreire Ok, did finally manage to reproduce, it was the missing latency estimation.
This still has some small glitches when reconnecting to a e.g. browser running a youtube video but the problems occurring with 'rattles' seem to be fixed.

vscfreire:
[...]

PR #5517, opened 2019-08-12
Added proper latency estimation for winmm sound backend.

akallabeth:
Fixes some issues as reported in #5514 with links that have variable latency.


akallabeth:
@hardening Improved latency estimation a bit and implemented missing/broken winpr `GetTickCount64` for windows XP. (yea, I know...) as well as some compile issues with windows 7 linking against directX.
If you still like/apporove the changes, please merge :)

PR #5518, opened 2019-08-12
Return a proper error if redirection fails.

akallabeth:


kubistika:
I think adding a debug log indicating that a redirection is starting can be useful (saying this after having a hard time solving proxy issues related to redirection, without knowing that they were related to redirection, at first).

PR #5519, opened 2019-08-12
client: proxy: allow passing SOCKS5 proxies as env. vars

sbraz:
Hi,
For some reason, when SOCKS5 support was added, SOCKS5 proxies could not be passed via environment variables. This PR fixes that shortcoming.


akallabeth:
@freerdp-bot test

sbraz:
> Nice patch, but for consistency could you also update the `proxy_prepare` environment variable checks?

Hi, I'm not sure what is missing from it. Should we parse another variable like `socks_proxy`?

PR #5520, opened 2019-08-12
Do not try to free a colormap that FreeRDP did not create.

r-barnett:
I think the call to XFreeColormap() should not be here since it tries to free the default colormap that FreeRDP did not create.


hardening:

PR #5521, opened 2019-08-13
Proxy rdpsnd support

kubistika:
This one adds rdpsnd support for the proxy.

Please do not merge this one before #5513 is merged


include/freerdp/client/rdpsnd.h
```diff
@@ -51,6 +52,8 @@ struct rdpsnd_device_plugin
 	pcStart Start;
 	pcClose Close;
 	pcFree Free;
+
+	rdpContext* rdpcontext;
```
akallabeth: Not required, already part of `rdpsnd_plugin`. 
better export that in a private header only available in the sound channel, then you don´t need to change the public API.

kubistika: not sure i understood how to solve this. rdpsnd_plugin is not exported, that's why i added it

kubistika: @akallabeth ?

akallabeth: sorry, forgot about this one. instead of adding an additional pointer to a struct you could just use a getter (`FREERDP_API rdpContext* freerdp_rdpsnd_get_context(rdpsndPlugin* rdpsnd);`) so you don´t change the existing structs but allow access to internal members.
that way existing plugins would not need to be recompiled for this function to work.

kubistika: got you, thanks.

kubistika:
@akallabeth rebased on master + added the settings getter in include/freerdp/client/rdpsnd.h

kubistika:
@akallabeth for some reason the commits order has changed (locally it seems okay...) `server: proxy: rdpsnd support` is supposed to be the last commit. i think it has something to do with the CI.

PR #5522, opened 2019-08-13
server: proxy: pf_context.c: Fix nla crash

kubistika:
Details in commit message

### DIFF
--- a/libfreerdp/core/settings.c
+++ b/libfreerdp/core/settings.c
@@ -789,6 +789,11 @@ BOOL freerdp_settings_copy(rdpSettings* _settings, const rdpSettings* settings)
 	CHECKED_STRDUP(AllowedTlsCiphers); /* 1101 */
 	CHECKED_STRDUP(NtlmSamFile); /* 1103 */
 	CHECKED_STRDUP(PreconnectionBlob); /* 1155 */
+	CHECKED_STRDUP(TargetNetAddress); /* 1217 */
+	CHECKED_STRDUP(RedirectionUsername); /* 1220 */
+	CHECKED_STRDUP(RedirectionDomain); /* 1221 */
+	CHECKED_STRDUP(RedirectionTargetFQDN); /* 1224 */
+	CHECKED_STRDUP(RedirectionTargetNetBiosName); /* 1225 */
 	CHECKED_STRDUP(RedirectionAcceptedCert); /* 1231 */
 	CHECKED_STRDUP(KerberosKdc); /* 1344 */
 	CHECKED_STRDUP(KerberosRealm); /* 1345 */
@@ -991,6 +996,33 @@ BOOL freerdp_settings_copy(rdpSettings* _settings, const rdpSettings* settings)
 
 	CopyMemory(_settings->ClientTimeZone, settings->ClientTimeZone,
 			   sizeof(TIME_ZONE_INFORMATION));
+
+	_settings->RedirectionPasswordLength = settings->RedirectionPasswordLength;
+	if (settings->RedirectionPasswordLength > 0)
+	{
+		_settings->RedirectionPassword = malloc(_settings->RedirectionPasswordLength);
+		if (!_settings->RedirectionPassword)
+		{
+			_settings->RedirectionPasswordLength = 0;
+			goto out_fail;
+		}
+
+		CopyMemory(_settings->RedirectionPassword, settings->RedirectionPassword, _settings->RedirectionPasswordLength);
+	}
+
+	_settings->RedirectionTsvUrlLength = settings->RedirectionTsvUrlLength;
+	if (settings->RedirectionTsvUrlLength > 0)
+	{
+		_settings->RedirectionTsvUrl = malloc(_settings->RedirectionTsvUrlLength);
+		if (!_settings->RedirectionTsvUrl)
+		{
+			_settings->RedirectionTsvUrlLength = 0;
+			goto out_fail;
+		}
+
+		CopyMemory(_settings->RedirectionTsvUrl, settings->RedirectionTsvUrl, _settings->RedirectionTsvUrlLength);
+	}
+
 	_settings->TargetNetAddressCount = settings->TargetNetAddressCount;
 
 	if (settings->TargetNetAddressCount > 0)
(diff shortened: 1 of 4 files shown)


kubistika:
@akallabeth well, do you have any idea where this should be?
these pointers must be set to NULL, otherwise a double free occurs
and the copy is what causing the pointers duplication issue, so i though to reset the pointers to NULL here.

kubistika:
@akallabeth i'll move the set NLA = TRUE to pf_client_pre_connect.

kubistika:
@akallabeth rebased on master and did the change

server/proxy/pf_context.c
```diff
@@ -98,10 +103,21 @@ void pf_context_copy_settings(rdpSettings* dst, const rdpSettings* src, BOOL is_
...
 
-	if (is_server)
+	if (is_dst_server)
 	{
 		free(dst->ServerCertificate);
 		dst->ServerCertificateLength = before_copy->ServerCertificateLength;
+
+		/* all of these should be marked as NULL, because they are already freed in the source
+		 * settings struct */
+		dst->TargetNetAddress = NULL;
```
akallabeth: Just checked, if they are not `NULL` here some memory has been allocated in `freerdp_settings_copy` that needs to be cleaned up - or do I miss the point?

server/proxy/pf_context.c
```diff
@@ -98,10 +103,21 @@ void pf_context_copy_settings(rdpSettings* dst, const rdpSettings* src, BOOL is_
...
+		dst->RedirectionTargetFQDN = NULL;
```
akallabeth: This one should be copied in `freerdp_settings_copy`

server/proxy/pf_context.c
```diff
@@ -98,10 +103,21 @@ void pf_context_copy_settings(rdpSettings* dst, const rdpSettings* src, BOOL is_
...
+		dst->RedirectionTargetNetBiosName = NULL;
```
akallabeth: This one should be copied in `freerdp_settings_copy`

server/proxy/pf_context.c
```diff
@@ -98,10 +103,21 @@ void pf_context_copy_settings(rdpSettings* dst, const rdpSettings* src, BOOL is_
...
+		dst->RedirectionUsername = NULL;
```
akallabeth: This one should be copied in `freerdp_settings_copy`

server/proxy/pf_context.c
```diff
@@ -98,10 +103,21 @@ void pf_context_copy_settings(rdpSettings* dst, const rdpSettings* src, BOOL is_
...
+		dst->RedirectionDomain = NULL;
```
akallabeth: This one should be copied in `freerdp_settings_copy`

server/proxy/pf_context.c
```diff
@@ -98,10 +103,21 @@ void pf_context_copy_settings(rdpSettings* dst, const rdpSettings* src, BOOL is_
...
+		dst->RedirectionPassword = NULL;
```
akallabeth: This one should be copied in `freerdp_settings_copy`

server/proxy/pf_context.c
```diff
@@ -98,10 +103,21 @@ void pf_context_copy_settings(rdpSettings* dst, const rdpSettings* src, BOOL is_
...
+		dst->RedirectionTsvUrl = NULL;
+		dst->RedirectionAcceptedCert = NULL;
```
akallabeth: This one has been allocated in `freerdp_settings_copy`

server/proxy/pf_context.c
```diff
@@ -98,10 +103,21 @@ void pf_context_copy_settings(rdpSettings* dst, const rdpSettings* src, BOOL is_
...
+
+		/* all of these should be marked as NULL, because they are already freed in the source
+		 * settings struct */
+		dst->TargetNetAddress = NULL;
+		dst->RedirectionTargetFQDN = NULL;
+		dst->RedirectionTargetNetBiosName = NULL;
+		dst->RedirectionUsername = NULL;
+		dst->RedirectionDomain = NULL;
+		dst->RedirectionPassword = NULL;
+		dst->RedirectionTsvUrl = NULL;
```
akallabeth: This one should be copied in freerdp_settings_copy

server/proxy/pf_context.c
kubistika: no, they will be freed when context_free will be called in connection termination
they are set to null only in the destination struct, but the src struct contains those pointers and they will be freed

akallabeth:
@kubistika There seems to be some use after free left.
Try a `clang` scanbuild (just set `CC=<path to>/ccc-analyzer` before running `cmake` and build with `scan-build -V cmake --build .`

server/proxy/pf_context.c
```diff
@@ -86,7 +91,7 @@ void pf_context_copy_settings(rdpSettings* dst, const rdpSettings* src, BOOL is_
 	free(dst->ClientRandom);
```
akallabeth: Not a good idea to free that here and copy to that pointer a few lines below.

kubistika:
@akallabeth would you mind checking the changes?

libfreerdp/core/settings.c
```diff
@@ -991,6 +996,33 @@ BOOL freerdp_settings_copy(rdpSettings* _settings, const rdpSettings* settings)
 
 	CopyMemory(_settings->ClientTimeZone, settings->ClientTimeZone,
 			   sizeof(TIME_ZONE_INFORMATION));
+
+	_settings->RedirectionPasswordLength = settings->RedirectionPasswordLength;
+	if (settings->RedirectionPasswordLength > 0)
+	{
+		_settings->RedirectionPassword = malloc(_settings->RedirectionPasswordLength);
```
akallabeth: Variable not set to NULL if nothing allocated (should already be NULL in source settings, but just to be sure)
(same below)

kubistika: so if settings->RedirectionPasswordLength == 0 -> _settings->RedirectionPassword needs to be set to NULL. right?

kubistika: @akallabeth

kubistika: by the way, i think this is okay, because at the begining of the function all pointers are copied (the deref copy *_settings = *settings), so the length will be 0 and the pointer will remain NULL

kubistika:
@akallabeth tested and works perfect. If you want me to set NULL if length = 0 for the code to be more easy to understand, please let me know. My opinion: it is not needed..

PR #5523, opened 2019-08-14
Rdp parser null checks

akallabeth:


client/common/file.c
```diff
@@ -881,11 +888,20 @@ static SSIZE_T freerdp_client_write_setting_to_buffer(char** buffer, size_t* buf
...
 	va_list ap;
 	SSIZE_T len;
-	char* buf = *buffer;
-	size_t bufSize = *bufferSize;
+	char* buf;
+	size_t bufSize;
+
+	if (!buffer || !bufferSize || !fmt)
+		return -1;
+
```
hardening: This is a static function only called in this file, so such checks (if really needed) should be done by the caller, aren't they ?

mfleisz: Actually this check is wrong because buffer and bufferSize can be NULL. In this case vsnprintf returns the required buffer size. This is actually used by `freerdp_client_write_rdp_file`. `freerdp_client_write_rdp_file_buffer` already checks the for valid parameters (either buffer and bufferSize both are NULL or non-NULL).

PR #5528, opened 2019-08-16
 client/common: Remove incorrect prompt for creds parsing

mfleisz:
"prompt for credentials on client" was incorrectly merged together with
"prompt for credentials" into a single setting. However the first option
determines if the client should prompt for credentials if the remote
server does not support server authentication. "prompt for credentials"
on the other hand determines if the client should use any previously
stored credentials or if it should always prompt for them.


mfleisz:
Removed formatting commit

PR #5533, opened 2019-08-19
server: proxy: sync cliprdr opening

kubistika:
Only open cliprdr for client if it is allowed in configuration and original client (client <> proxy's server) connected with cliprdr. Fixes a crash when connecting with xfreerdp without `/clipboard` flag.

### DIFF
--- a/server/proxy/pf_server.c
+++ b/server/proxy/pf_server.c
@@ -233,7 +233,6 @@ static DWORD WINAPI pf_server_handle_client(LPVOID arg)
 	pdata->config = client->ContextExtra;
 	config = pdata->config;
 	client->settings->UseMultimon = TRUE;
-	client->settings->RedirectClipboard = config->Clipboard;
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

server/proxy/pf_server.c
```diff
@@ -233,7 +233,6 @@ static DWORD WINAPI pf_server_handle_client(LPVOID arg)
 	pdata->config = client->ContextExtra;
 	config = pdata->config;
 	client->settings->UseMultimon = TRUE;
-	client->settings->RedirectClipboard = config->Clipboard;
 	client->settings->SupportGraphicsPipeline = config->GFX;
 	client->settings->SupportDynamicChannels = TRUE;
 	client->settings->CertificateFile = _strdup("server.crt");
```
akallabeth: Missed this, but it might be a good idea to check `shadow` server to have a proper file path appended.

PR #5534, opened 2019-08-19
(FreeBSD) Build fixes for urbdrc

kevans91:
- error is defined as both int and UINT32 in urbdrc_search_usb_device... remove int definition, as context would have me believe that UINT32 is more proper.
- CreateFileDescriptorEvent takes a mode argument now; match the Linux version of this.


akallabeth:
@freerdp-bot test

hardening:

PR #5541, opened 2019-08-23
Rdp file full drivestoredirect parsing

akallabeth:


client/common/cmdline.c
```diff
@@ -3173,6 +3216,77 @@ BOOL freerdp_client_load_addins(rdpChannels* channels, rdpSettings* settings)
...
+	if (settings->DrivesToRedirect)
+	{
+
+		/*
+		 * Drives to redirect:
+		 *
+		 * Very similar to DevicesToRedirect, but can contain a
+		 * comma-separated list of drive letters to redirect.
+		 */
+		const BOOL empty = !settings->DrivesToRedirect || (strlen(settings->DrivesToRedirect) == 0);
```
mfleisz: We already checked that DrivesToRedirect is not NULL. We can just do `if (strlen(settings->DrivesToRedirect) > 0)` here

akallabeth: done

client/common/cmdline.c
```diff
@@ -48,6 +48,102 @@
...
+	if (name)
+	{
+		if (!(drive->Name = _strdup(name)))
+			goto fail;
+	}
+	else /* We need a name to send to the server. */
+		if (!(drive->Name = _strdup(path)))
+			goto fail;
+
+	if (!path || !freerdp_sanitize_drive_name(drive->Name, "\\/:", "___"))
```
mfleisz: The : should not be sanitized/replaced with _

akallabeth: are you sure? a name with a `:` might otherwise be interpreted as path.

PR #5542, opened 2019-08-23
keyboard: don't catch Ctrl+Alt-C if encomsp is not activated

hardening:


akallabeth:
@hardening nice catch

PR #5547, opened 2019-08-26
Fix #5543: 16 bit color to 32bit color conversion.

### LINKED ISSUE #5543: Incorrect colors
Hi,

several Royal TSX customers have reported an issue to me that I wasn't taking very seriously until I started some more in-depth debugging today.

It seems like color rendering is slightly off compared to MSRDP and what I'm seeing when connecting directly to the console (via VMware) of the machine.
This is most easily seen when something should be 100% white. When using FreeRDP it's a light gray instead of 100% white.

Here are three screenshots:

**VMware Console**
VMwareConsole

**MSRDP**
MSRDP

**FreeRDP**
FreeRDP

Now if you use a color picker and pick the white color in the user icon you will notice that in the first two screenshots it's 100% white but FreeRDP renders it as R:248 G:252 B:248.

I have only been able to reproduce when connecting to Windows Server 2012 R2 machines and older. Also, as soon as I turn on /gfx the issue is gone.

Any idea what's going on?

thx,
felix

akallabeth:
When converting 16 bit color (5 or 6 bit per color) to 32bit
(8 bit per color) add a value dependent correction factor so that
value 0 -> 0 and 31 -> 255.

### DIFF
--- a/include/freerdp/codec/color.h
+++ b/include/freerdp/codec/color.h
@@ -365,13 +365,25 @@ static INLINE void SplitColor(UINT32 color, UINT32 format, BYTE* _r, BYTE* _g,
 		/* 16bpp formats */
 		case PIXEL_FORMAT_RGB16:
 			if (_r)
-				*_r = (BYTE)(((color >> 11) & 0x1F) << 3);
+			{
+				const UINT32 c = (color >> 11) & 0x1F;
+				const UINT32 val = (c << 3) + c / 4;
+				*_r = (BYTE) (val > 255 ? 255 : val);
+			}
 
 			if (_g)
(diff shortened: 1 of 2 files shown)


akallabeth:
@lemonmojo looks like the fix was easier than expected. could you test if this fixes the problem for you too?

hardening:
Would be nice to have a unitary test for all this...

akallabeth:
@hardening ok, added the missing check to interleaved codec (which uses the functions heavily and provides an appropriate test case for the rest)

lemonmojo:
@akallabeth I'm sorry I will only be able to test the changes next week.

lemonmojo:
@akallabeth Works perfectly! Many thx for the quick turnaround! =)

PR #5552, opened 2019-08-28
Bugfix/blank screen

### LINKED ISSUE #5530: Server proxy - blank screen
When connecting to a server proxy, I get a blank screen.

**To Reproduce**
1. Build FreeRDP with WITH_SERVER=ON option
2. run freerdp-proxy with the config.ini file specifying a target
3. On a separate machine build FreeRDP
4. Run xfreerdp connecting to the proxy host, ensuring the keys in server/proxy/ folder are in your local folder
5. You will get a blank screen.

kubistika:
Fixes #5530.

Problem was that proxy's client was waiting for server's drynvc to be in ready state no matter if original client connected with any dynamic channel.

Fixed by: only waiting when needed. For disp, which doesn't have an early capability flag (like GFX) indicating that it will be opened, we first call Open and check if ERROR_NOT_FOUND was returned. If it wasn't the error, then we wait and call it again.

### DIFF
--- a/server/proxy/pf_channels.c
+++ b/server/proxy/pf_channels.c
@@ -43,6 +43,13 @@
 
 #define TAG PROXY_TAG("channels")
 
+static void pf_channels_wait_for_server_dynvc(pServerContext* ps)
+{
+	WLog_DBG(TAG, "pf_channels_wait_for_server_dynvc(): waiting for server's drdynvc to be ready");
+	WaitForSingleObject(ps->dynvcReady, INFINITE);
+	WLog_DBG(TAG, "pf_channels_wait_for_server_dynvc(): server's drdynvc is ready!");
+}
+
 void pf_OnChannelConnectedEventHandler(void* data,
                                        ChannelConnectedEventArgs* e)
 {
@@ -57,6 +64,8 @@ void pf_OnChannelConnectedEventHandler(void* data,
 	}
 	else if (strcmp(e->name, RDPGFX_DVC_CHANNEL_NAME) == 0)
 	{
+		pf_channels_wait_for_server_dynvc(ps);
+
 		if (!ps->gfx->Open(ps->gfx))
 		{
 			WLog_ERR(TAG, "failed to open GFX server");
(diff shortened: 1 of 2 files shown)


server/proxy/pf_channels.c
```diff
@@ -68,13 +77,28 @@ void pf_OnChannelConnectedEventHandler(void* data,
 	}
 	else if (strcmp(e->name, DISP_DVC_CHANNEL_NAME) == 0)
 	{
-		if (ps->disp->Open(ps->disp) != CHANNEL_RC_OK)
```
akallabeth: Why this retry? I don't really understand your explanation, could you elaborate?

kubistika: yep.
first, i try to open disp. if it fails, i need to check if ERROR_NOT_FOUND is returned. it that's the case, then nothing should be done (client did not open disp). otherwise, it might not opened because server's dynvc wasn't in ready state yet. then, i wait for it to be in ready state and try again.

i must first check for ERROR_NOT_FOUND because drynvc will be in ready state only if a drdynvc caps sent from client.

server/proxy/pf_channels.c
```diff
@@ -57,6 +64,8 @@ void pf_OnChannelConnectedEventHandler(void* data,
 	}
 	else if (strcmp(e->name, RDPGFX_DVC_CHANNEL_NAME) == 0)
 	{
+		pf_channels_wait_for_server_dynvc(ps);
```
akallabeth: Why not just wait on server `drdynvc` when the client `drdynvc` event is triggered?

kubistika: Because client drdynvc event will be triggered even if original client did not connect with any dynamic channel. This is because proxy's client always opens DISP, because it has no idea whether if original client connected with disp or not.

kubistika: @akallabeth

akallabeth: @kubistika sorry, did want to think a bit about it, but did not have time.
lets merge this one and (hopefully) come up with a better solution later.

kubistika: I'd really appreciate if you will have the time to help me figure another solution. I don't like it too but this is the only solution I found for this :/ (that makes sure channels will be opened)

akallabeth: I'll keep an eye open for this, but currently a bit busy :/

PR #5553, opened 2019-08-29
Reattach to cmd if wfreerdp was started from one.

akallabeth:
Found a solution for the long standing no output on console issue.
With `AttachConsole(ATTACH_PARENT_PROCESS)`  we reattach `wfreerdp.exe` to the calling console.
If used as launcher for `rdp` files no new console process is created, so this should cover all use cases.

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -65,6 +65,8 @@ static BOOL wf_create_console(void)
 
 	freopen("CONOUT$", "w", stdout);
 	freopen("CONOUT$", "w", stderr);
+	freopen("CONIN$", "r", stdin);
+	clearerr(stdin);
 
 	WLog_INFO(TAG,  "Debug console created.");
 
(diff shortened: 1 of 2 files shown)


hardening:
@akallabeth sorry I don't know enough windows internals to comment on this.

akallabeth:
@hardening workings are simple:
`AttachConsole` returns ` TRUE` if it can attach to a parent console (when started from e.g. `cmd.exe`) but `FALSE` when started as file handler for `rdp` files.
So now you can see the output of wfreerdp when started from command line.

There is a side effect though, `wfreerdp >log.txt 2>&1`  no longer works :/

client/Windows/wf_client.c
```diff
@@ -58,15 +58,18 @@
...
+static BOOL wf_create_console(void)
 {
-	if (!AllocConsole())
-		return 1;
+	if (!AttachConsole(ATTACH_PARENT_PROCESS))
+		return FALSE;
 
 	freopen("CONOUT$", "w", stdout);
 	freopen("CONOUT$", "w", stderr);
+	freopen("CONIN$", "w", stdin);
```
mfleisz: Shouldn't this be `"r"`?

client/Windows/wf_client.c
```diff
@@ -494,64 +505,149 @@ static BOOL wf_gw_authenticate(freerdp* instance,
...
+fail:
+	free(buffer);
+	free(caption);
+
 	/* return 1 to accept and store a certificate, 2 to accept
 	 * a certificate only for this session, 0 otherwise */
-	return 2;
+	switch(what)
+	{
+	case IDOK:
```
mfleisz: Should be `IDYES`

client/Windows/wf_client.c
```diff
@@ -494,64 +505,149 @@ static BOOL wf_gw_authenticate(freerdp* instance,
...
+	what = MessageBoxW(NULL, buffer, caption, MB_YESNOCANCEL);
+fail:
+	free(buffer);
+	free(caption);
+
+	/* return 1 to accept and store a certificate, 2 to accept
+	 * a certificate only for this session, 0 otherwise */
+	switch(what)
+	{
+	case IDOK:
```
mfleisz: Should be `IDYES`

client/Windows/wf_client.c
```diff
@@ -930,6 +1024,15 @@ static void wfreerdp_client_global_uninit(void)
 
 static BOOL wfreerdp_client_new(freerdp* instance, rdpContext* context)
 {
+	wfContext* wfc = (wfContext*)context;
+	if (!wfc)
+		return FALSE;
+
+	// AttachConsole and stdin do not work well.
+	// Use GUI input dialogs instead of command line ones.
+	//wfc->isConsole = wf_create_console();
```
mfleisz: `isConsole` is nowhere initialized

client/Windows/wf_client.c
```diff
@@ -494,64 +505,149 @@ static BOOL wf_gw_authenticate(freerdp* instance,
...
-	/* TODO: ask for user validation */
-#if 0
-	input_handle = GetStdHandle(STD_INPUT_HANDLE);
-	GetConsoleMode(input_handle, &mode);
-	mode |= ENABLE_ECHO_INPUT | ENABLE_LINE_INPUT;
-	SetConsoleMode(input_handle, mode);
-#endif
+	int rc;
+	size_t size = 1024;
+	WCHAR* buffer = calloc(size, sizeof(WCHAR));
```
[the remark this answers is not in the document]
akallabeth: @mfleisz well, it is the same as you suggest (with a 'default' of 1024 bytes already allocated).
`vswprintf_s` returns the required characters if the buffer is too small anyway.

PR #5561, opened 2019-09-02
Winmm play fix

### LINKED ISSUE #5412: WfreeRDP audio cuts out on close of another instance
WfreeRDP audio cuts out on close of another instance.

When connected to 2 different devices (in my case Windows 10 remote servers and I'm also connecting using windows 10) using audio, the "/sound" command line option, closing one wfreerdp window cuts out the audio for the remaining instance.

**To Reproduce**
1. Open powershell
2. cd to the wfreerdp.exe
3. Open a WFreeRDP instance by entering ".\wfreerdp /v:<hostName> /u:<userName> /p:<password> /sound"
4. Repeat steps 1-3 so now you have 2 instances open (doesn't matter if you connect to the same server or not, you can even use a different wfreerdp exe in another location)
5. Play an audio file on server 1, or use windows narrator to make sounds (a useful thing to have open at this stage is the windows Sound devices dialog with the device output visible)
6. Close the second instance by logging off, shutting down, disconnecting or simply closing the program

**Expected behavior**
Audio continues to work for the first instance

**Actual behavior**
Audio stops working / playing, although I can still see the audio output in the windows sound dialog

**Application details**
* Version of FreeRDP = latest available from here https://cloudbase.it/freerdp-for-windows-nightly-builds/
* Command line used = ".\wfreerdp /v:<hostName> /u:<userName> /p:<password> /sound"
* OS version connecting to = Windows 10 and windows server 2016 tested

**Desktop (please complete the following information):**
 - OS: Windows 10

**Additional Information**
Having made a Visual Studio project using CMAKE and had a poke around, through sheer trial and error, I've noticed that Line 1148 of rdpsnd_main.c 

```
SetEvent(rdpsnd->stopEvent);
```

being commented out stops the problem from occurring. I'm sure doing this causes all sorts of other problems, but it might be a helpful place to start.

akallabeth:
Fixed #5412

### DIFF
--- a/channels/rdpgfx/client/rdpgfx_main.c
+++ b/channels/rdpgfx/client/rdpgfx_main.c
@@ -760,7 +760,7 @@ static UINT rdpgfx_recv_start_frame_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
 	Stream_Read_UINT32(s, pdu.frameId); /* frameId (4 bytes) */
 	DEBUG_RDPGFX(gfx->log, "RecvStartFramePdu: frameId: %"PRIu32" timestamp: 0x%08"PRIX32"",
 	           pdu.frameId, pdu.timestamp);
-	gfx->StartDecodingTime = GetTickCountPrecise();
+	gfx->StartDecodingTime = GetTickCount64();
 
 	if (context)
 	{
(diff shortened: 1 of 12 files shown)


akallabeth:
@mfleisz true, text file dos/unix mismatch, reverted.

PR #5569, opened 2019-09-05
Dynamic printer backend

akallabeth:
Allows loading of dynamic printer backend libraries.


channels/printer/client/cups/printer_cups.c
```diff
@@ -309,23 +354,53 @@ static rdpPrinter* printer_cups_get_printer(rdpPrinterDriver* driver,
...
+	rdpCupsPrinterDriver* cups_driver = (rdpCupsPrinterDriver*) driver;
+	if (cups_driver->references <= 1)
+	{
+		free(cups_driver);
+		cups_driver = NULL;
+	}
+	else
+		cups_driver->references--;
+}
+
```
hardening: I'd prefer something like:
```c
#ifdef BUILTIN_CHANNELS
rdpPrinterDriver *cups_freerdp_printer_client_subsystem_entry(void)
#else
FREERDP_API freerdp_printer_client_subsystem_entry(void)
#endif
```

user-72632:
@akallabeth and @mfleisz , this series of commits breaks my Printer Redirection now using the following syntax:
"/printer:MFC-8910DW,Brother MFC-8910DW Printer" 
If you need logs then I would be happy to provide.

Issue #5577, opened 2019-09-09
win_shadow_input_synchronize_event & win_shadow_input_mouse_event


Zoson:
In win_shadow.c, I found that The Function of "win_shadow_input_synchronize_event" returns False directly and why The fuction of win_shadow_input_mouse_event returns 0 at the end,  should it be 1?

PR #5582, opened 2019-09-10
Fix windows shadow server issues reported in #5577

akallabeth:


akallabeth:
@hardening documentation is a nice thing, especially if half of the truth is missing ^^

PR #5589, opened 2019-09-11
Tighter error checking for integer options in RDP file.

akallabeth:


client/common/file.c
```diff
@@ -567,6 +569,40 @@ BOOL freerdp_client_parse_rdp_file_buffer(rdpFile* file, const BYTE* buffer, siz
 	return freerdp_client_parse_rdp_file_buffer_ex(file, buffer, size, NULL);
 }
 
+static char* trim(char* value, size_t length)
```
mfleisz: I think this function should use `isspace` instead of all these switch/cases

client/common/file.c
```diff
@@ -567,6 +569,22 @@ BOOL freerdp_client_parse_rdp_file_buffer(rdpFile* file, const BYTE* buffer, siz
...
+{
+	if (value)
+	{
+		BOOL found = TRUE;
+		char* end = &value[length - 1];
+		while((value != end) && isspace(end))
```
mfleisz: Should be `isspace(*end)`

client/common/file.c
```diff
@@ -567,6 +569,22 @@ BOOL freerdp_client_parse_rdp_file_buffer(rdpFile* file, const BYTE* buffer, siz
...
+				*end-- = '\0';
+	}
+
+	while(isspace(value))
```
mfleisz: Should be `isspace(*value)`

PR #5595, opened 2019-09-14
avoid blurry filter when scaling factor is 2x, 3x, etc

volth:
Useful when the client has high-dpi monitor.

For example, when the client has 4k (3160x2160) monitor, `xfreerdp /w:1920 /h:1080 /smart-sizing /f` and `xfreerdp /w:1280 /h:720 /smart-sizing /f` will create crispy 1920x1080 and 1280x720 RDP sessions.

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -123,6 +123,7 @@ static void xf_draw_screen_scaled(xfContext* xfc, int x, int y, int w, int h)
 	double yScalingFactor;
 	int x2;
 	int y2;
+	const char* filter;
 	rdpSettings* settings = xfc->context.settings;
 
 	if (xfc->scaledWidth <= 0 || xfc->scaledHeight <= 0)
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@volth nice one, thank you!

hardening:

ikeboy:
this isn't working for me on Ubuntu 19.10. Tried the latest nightly and using your exact command on a 4k monitor (precision 5530). Logging into a Windows server.

bparker06:
Same issue here for me.

volth:
It was refactored in https://github.com/FreeRDP/FreeRDP/pull/5657
probably it broke something
cc @mfleisz

volth:
As far as remember, `0.001` was used because `DBL_EPSILON` is too small, those scaling factor values were not exact `3840 / 1280`  but something 1 pixel more or less, so that `0.001` was chosen as "at least (1 / screenwidth)"

PR #5597, opened 2019-09-15
channels: cliprdr: get rid of duplicated code

kubistika:
This one refactors the cliprdr channel, by getting rid of a lot! code duplication. I just couldn't see so much functions duplicated (in cliprdr, many messages are bi-directional, and server's code was copy-pasted from client's code in freerdp). It also fixes a problem that occured in file copying though the proxy because a wrong check done in cliprdr server (dwFlags is only needed for request!).

By the way, it also fixes some Stream_Read that were done without making sure stream capacity is enough (in server code).

### DIFF
--- a/channels/cliprdr/client/CMakeLists.txt
+++ b/channels/cliprdr/client/CMakeLists.txt
@@ -21,7 +21,10 @@ set(${MODULE_PREFIX}_SRCS
 	cliprdr_format.c
 	cliprdr_format.h
 	cliprdr_main.c
-	cliprdr_main.h)
+	cliprdr_main.h
+	../cliprdr_common.h
+	../cliprdr_common.c
+	)
 
 add_channel_client_library(${MODULE_PREFIX} ${MODULE_NAME} ${CHANNEL_NAME} FALSE "VirtualChannelEntryEx")
 
--- a/channels/cliprdr/client/cliprdr_format.c
+++ b/channels/cliprdr/client/cliprdr_format.c
@@ -33,6 +33,7 @@
 
 #include "cliprdr_main.h"
 #include "cliprdr_format.h"
+#include "../cliprdr_common.h"
 
 /**
  * Function description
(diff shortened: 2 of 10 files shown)


kubistika:
@akallabeth can this one be merged?

akallabeth:
@kubistika waiting for feedback on some other chanels if the removed public structs are not used

kubistika:
@akallabeth do you mean channels outside freerdp's src tree? Okay

kubistika:
@akallabeth any progress on this? I can remove the commit that removes the unused structs.

akallabeth:
Sorry, been busy. Yes, probably the safe route.

kubistika:
@akallabeth did it.

kubistika:
@akallabeth can we merge this one?

akallabeth:
@kubistika Did some testing and the pr looks good.
Only thing I noticed is that `cliprdr_packet_new` and `cliprdr_server_packet_new` are the same and the whole common part can be greatly improved when you just return the allocated `wStream` for most of the calls.
Same is true for `cliprdr_packet_send` and `cliprdr_server_packet_send`

kubistika:
@akallabeth do you mean that instead of writing to wStream pointer, all writing functions should just create a new stream and write to it?

akallabeth:
@kubistika well, won´t work for all, but you could modify your write functions to allocate the stream and write the blocks and return this stream, so code could shrink quite a bit with that change. (no *_new and error checking anymore everywhere)

akallabeth:
like
```
s = cliprdr_packet_new(CB_LOCK_CLIPDATA, 0, 4);
@@ -807,11 +785,17 @@ static UINT cliprdr_client_lock_clipboard_data(CliprdrClientContext* context,
		return ERROR_INTERNAL_ERROR;
		return ERROR_INTERNAL_ERROR;
	}
	}


	Stream_Write_UINT32(s, lockClipboardData->clipDataId); /* clipDataId (4 bytes) */
	if ((error = cliprdr_write_lock_clipdata(s, lockClipboardData)))
		goto fail;
```

could be merged to a single `cliprdr_packet_lock_clipdata_new` function

kubistika:
Got it, I'll do that. Thank you.

kubistika:
@akallabeth Took some time, but I really think it looks so much better than before this PR.
Tested again functionality, seems fine. Would be great if you can test too.

PR #5621, opened 2019-09-26
Trim all RDP file string options except passwords.

akallabeth:
make parser more compatible with broken RDP files (spaces before and after strings)

### DIFF
--- a/client/common/file.c
+++ b/client/common/file.c
@@ -22,6 +22,7 @@
 #endif
 
 #include <errno.h>
+#include <ctype.h>
 
 #include <freerdp/client/file.h>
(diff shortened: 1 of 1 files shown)


client/common/file.c
```diff
@@ -579,6 +580,85 @@ BOOL freerdp_client_parse_rdp_file_buffer(rdpFile* file, const BYTE* buffer, siz
...
+		str++;
+
+	while((end > str) && isspace(*end))
+		end--;
+	end[1] = '\0';
+	if (start == str)
+		*strptr = str;
+	else
+	{
+		*strptr = _strdup(str);
```
mfleisz: missing NULL check

PR #5628, opened 2019-10-01
Fix WINPR_EVENT and WINPR_TIMER structs.

akallabeth:
Leads to breakage when library is used as server.


akallabeth:
@hardening they break ogon ;)
Look at the getfiledescriptor function that returns the wrong value because of that

PR #5632, opened 2019-10-02
rail: Update to latest spec version 28.0

mfleisz:


channels/rail/client/rail_orders.c
```diff
@@ -261,14 +261,14 @@ static UINT rail_read_server_get_appid_resp_order(wStream* s,
 	if (!s || !getAppidResp)
 		return ERROR_INVALID_PARAMETER;
 
-	if (Stream_GetRemainingLength(s) < 516)
+	if (Stream_GetRemainingLength(s) < 524)
```
akallabeth: sure this will not break older servers?

winpr/include/winpr/stream.h
```diff
@@ -180,6 +180,12 @@ static INLINE void Stream_Write_UINT8(wStream* _s, UINT8 _v)
 	*_s->pointer++ = (UINT8)(_v);
 }
 
+static INLINE void Stream_Write_INT16(wStream* _s, INT16 _v)
```
akallabeth: Not necessary, the `Stream_Write_UINT16`   does the same ;)

channels/rail/client/rail_orders.c
mfleisz: To my understanding this was a documentation error and the PDU actually contained 520 bytes of application id data from the beginning. Inf fact the current spec is still erroneous as it still shows the PDU with 512 bytes in the Protocol Examples chapter.

winpr/include/winpr/stream.h
mfleisz: Yes but that takes a UINT16 and I don't wanted to introduce new warnings/casts

channels/rail/client/rail_orders.c
mfleisz: Just checked legacy rail to a Win7 machine and it definitely sends a 520 bytes application id.

winpr/include/winpr/stream.h
akallabeth: true.

channels/rail/client/rail_orders.c
akallabeth: ah, thanks for the clarification.

PR #5639, opened 2019-10-03
Added function to get an error category for better error handling.

akallabeth:
Groundwork to be able to categorize an error so that a error message with 'who to blame' instructions can be displayed by clients.


mfleisz:
@freerdp-bot test

PR #5660, opened 2019-10-18
Fixes Auto USB Redirection

AzureusNation:
Base 0 does not work when converting the string to long int. This ends up not showing the correct device id and sometimes the wrong bus id. By changing it to base 10 conversion instead of base 0 this fixes the issue of auto redirect usb devices.


akallabeth:
@freerdp-bot test

AzureusNation:
Theres a couple other bits that still need fixing I believe but this is the big step forward to getting to the end goal.

PR #5665, opened 2019-10-22
server: proxy: implement GFX capture

kubistika:
https://streamable.com/f86z0 :)

In order to implement this, proxy's GFX client side must do two things:
* Decode GFX orders and build the drawing frame.
* Proxy all GFX orders to original client.

In order to do that, I exported an API to create a GFX client, and created another client (for decoding).


kubistika:
@akallabeth any help with the CI will be very appreciated  ^^
https://ci.freerdp.com/job/windows_msbuild_vs2010/arch=win64,label=vs2010/2330/console

akallabeth:
@freerdp-bot test

server/proxy/pf_rdpgfx.c
```diff
@@ -241,7 +401,7 @@ static UINT pf_rdpgfx_caps_advertise(RdpgfxServerContext* context,
 		 * TODO: Have a better way of expressing max supported GFX caps version
 		 * by FreeRDP.
 		 */
-		if (currentCaps->version <= RDPGFX_CAPVERSION_106)
+		if (currentCaps->version <= RDPGFX_CAPVERSION_102)
```
kubistika: should revert, here by mistake (was testing something)

akallabeth: ok, I´ll wait for the revert before merging ^^

kubistika: reverted :)
i'll be glad if while reviewing, you'll take a deeper look in pf_capture.c (i want to make sure that there are no problems with how the path is built).

server/proxy/pf_capture.c
```diff
@@ -0,0 +1,77 @@
...
+ * it is important to call this function only after the connection is fully established, as it uses
+ * settings->Username and settings->ServerHostname values to create the directory. After the
+ * connection established, we know that those values are valid.
+ */
+BOOL pf_capture_create_session_directory(pClientContext* pc)
+{
+	proxyConfig* config = pc->pdata->config;
+	rdpSettings* settings = pc->context.settings;
+	SYSTEMTIME localTime;
+	char tmp[MAX_PATH];
```
akallabeth: Might be a good idea to remove use of `MAX_PATH` as this is just the size of a single path segment (and under certain circumstances not even that is true)
`sprintf` functions have the neat property that they return the length required if called with `NULL` ;)

server/proxy/pf_capture.h
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+#ifndef FREERDP_SERVER_PROXY_CAPTURE_H
+#define FREERDP_SERVER_PROXY_CAPTURE_H
+
+#include <winpr/image.h>
+#include <winpr/sysinfo.h>
+#include <winpr/path.h>
+#include <winpr/file.h>
```
akallabeth: maybe move the includes to the implementation, they are not required in the header.

server/proxy/pf_capture.c
kubistika: hmm, so you're suggesting to call sprintf with NULL, get the size, allocate a buffer with this size and then call sprintf again?

server/proxy/pf_capture.h
kubistika: sure

server/proxy/pf_capture.c
kubistika: and what about the usage of MAX_PATH in pf_capture_save_frame?

akallabeth: correct, yes.
this is of course true for every use. (I know we still use it in some places in freerdp, but that is one thing to take care of, introducing it in new code is no good idea)

kubistika:
@akallabeth what do you think?

PR #5670, opened 2019-10-23
Fix #5669: If cairo is detected as part of other dependencies ignore it

### LINKED ISSUE #5669: Unable to build without libcairo
I want to compile FreeRDP without libcairo support. So I specify `-DWITH_CAIRO=OFF` in my cmake args. Unfortunately that doesn't seem to be enough as cmake still looks for it and unfortunately finds it `-- Found Cairo: /usr/local/lib/libcairo.dylib`.

Any way to get `-DWITH_CAIRO=OFF` to actually work and force it to be disabled?

thx,
felix

akallabeth:
If WITH_CAIRO=OFF it was still used when the library was detected as
part of some other dependency. With this commit the dependency is
only added if WITH_CAIRO=ON

### DIFF
--- a/libfreerdp/CMakeLists.txt
+++ b/libfreerdp/CMakeLists.txt
@@ -89,20 +89,18 @@ if (WITH_CAIRO)
     find_package(Cairo REQUIRED)
 endif(WITH_CAIRO)
 
-if (SWScale_FOUND)
+# Prefer SWScale over Cairo, both at the same time are not possible.
+if (WITH_SWSCALE)
     add_definitions(-DSWSCALE_FOUND=1)
     include_directories(${SWScale_INCLUDE_DIR})
     freerdp_library_add(${SWScale_LIBRARY})
-else(SWScale_FOUND)
-
-    if (CAIRO_FOUND)
-        add_definitions(-DCAIRO_FOUND=1)
-        include_directories(${CAIRO_INCLUDE_DIR})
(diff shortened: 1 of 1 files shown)


libfreerdp/CMakeLists.txt
```diff
@@ -89,20 +89,18 @@ if (WITH_CAIRO)
     find_package(Cairo REQUIRED)
 endif(WITH_CAIRO)
 
-if (SWScale_FOUND)
+# Prefer SWScale over Cairo, both at the same time are not possible.
+if (WITH_SWSCALE)
```
mfleisz: By far not a cmake expert but shouldn't we still check for `SWScale_FOUND` and only set `SWSCALE_FOUND=1` if both swscale was found and `WITH_SWSCALE` is enabled (same for cairo)?

akallabeth: @mfleisz yes, look 4 lines above ;)

mfleisz: @akallabeth Yes that's where we search for swscale ... but if I define `WITH_SWSCALE` then `-DSWSCALE_FOUND=1` below is defined ... no matter if the library was found or not ... or am I missing something 😅 ?

akallabeth: @mfleisz it is defined as `CMake` variable, not for `C` preprocessor symbol

PR #5685, opened 2019-10-28
COMMAND_LINE_ARGUMENT structs contain parser results, use one per instance

akallabeth:
Remove the old global structs as the parser modifies them. When using
multiple instances in the same process space this could break parsing.


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

Issue #5712, opened 2019-11-11
Remote Assistance (with file invitation.msrcIncident) over proxy


boom1:
**Remote Assistance by invitation file does not work via proxy.**

Tested with xfreerdp as RDP client on Fedora 30 with FreeRDP from git repo, master branch of Nov. 7th.
Target system for assistance was Windows 10.

For testing, start Remote Assistance under Windows 10, with invitation by file :
> msra.exe /saveasfile invitation test-password

Copy the resulting file invitation.msrcincident to a Linux PC which can reach the Windows PC only via a proxy (in below example a SOCKS5 proxy at proxy.test.org, port 1080).

Start xfreerdp on the Linux PC, with proxy settings:
> xfreerdp invitation.msrcincident /proxy:socks5://proxy.test.org:1080 /assistance:test-password

This does not work (at least in my case it does not use the proxy).
And this patch should fix this: [FreeRDP.patch.txt](https://github.com/FreeRDP/FreeRDP/files/3831388/FreeRDP.patch.txt)

Basically the patch moves code that tries the multiple destinations given in an invitation file (AFAIK at least IPv4, IPv6 and DNS name of the Windows PC) from a lower layer to a higher layer. This makes proxies work, because they are handled in an intermediate layer which now is used for each destination from the invitation file that is tried.

"higher layer" is nego_tcp_connect() in nego.c
"intermediate layer" is transport_connect() in transport.c
"lower layer" is freerdp_tcp_connect_multi() in tcp.c

akallabeth:
@boom1 so, does `proxy_prepare` not return `true` in your case?

boom1:
From my memory: proxy_prepare() does return true and returns the proxy settings from the command line (because I have no environment variables for proxy set).

akallabeth:
@boom1 checked it, looks like the glue is not correct for your use case.
The part with connecting to the proxy is lacking the alternative addresses.
the patch you provided is a nice start, but will fail for other cases (everything non NLA).
Mark it as todo.

boom1:
I agree, the patch still fails for other cases (I did not test it, but gateways seems not to go through the proxy as well).
Thank you for marking it as todo.

akallabeth:
@boom1 did some fixes in https://github.com/akallabeth/FreeRDP/tree/proxy_cleanup
does that start working for you?

boom1:
[...]

PR #5718, opened 2019-11-13
Fixed cleanup of remdesk in CHANNEL_EVENT_WRITE_COMPLETE

akallabeth:
Fix an issue reported in #5712:
`pData` is the event specific user data that was passed to `VirtualChannelWriteEx` as user data.


akallabeth:
@freerdp-bot test

PR #5720, opened 2019-11-13
fix [build]: don't export rdp2tcp as cmake target

bmiklautz:
rdp2tcp was wrongly exposed as cmake target. This could lead to errors
like the following when the FreeRDP2 cmake file was required:

CMake Error at ../lib/cmake/FreeRDP2/FreeRDPTargets.cmake:73 (message):

  The imported target "rdp2tcp-client" references the file
     ".../lib/freerdp2/librdp2tcp-client.a"
  but this file does not exist.  Possible reasons include:
* The file was deleted, renamed, or moved to another location.
* An install or uninstall procedure did not complete successfully.
* The installation package was faulty and contained
     ".../lib/cmake/FreeRDP2/FreeRDPTargets.cmake"
  but not all the files it references.


bmiklautz:
@freerdp-bot test!

PR #5744, opened 2019-11-21
urbdrc: fix some usb devices redirection failed

nanpuyue:
If `DeviceDescription` get by `control_query_device_text` contains
extra L'\0', the returned `bufferSize` do not match the actual value.
This will cause the USB device redirection to fail.

### DIFF
--- a/channels/urbdrc/client/libusb/libusb_udevice.c
+++ b/channels/urbdrc/client/libusb/libusb_udevice.c
@@ -1017,7 +1017,7 @@ static int libusb_udev_control_query_device_text(IUDEVICE* idev, UINT32 TextType
 			}
 			else
 			{
-				*BufferSize = ret;
+				*BufferSize = _wcslen((WCHAR*)Buffer) * 2;
 			}
 
 			break;


bmiklautz:
@freerdp-bot test

nanpuyue:
@akallabeth now fix in `libusb_udev_control_query_device_text`.

akallabeth:
@nanpuyue what exactly is the difference of `ret` to `wcslen *  2` ? this looks like a quick hack without knowing what is wrong, so could you check if this is a off by one or similar?

akallabeth:
@nanpuyue checked the values in the debugger, that is not really fixing anything (despite using unsafe string functions on possible non `\0` terminated buffers.

nanpuyue:
It should be `*BufferSize = _wcsnlen((WCHAR*)Buffer, ret / 2) * 2;`.

akallabeth:
@nanpuyue you still fail to explain what the difference here is. I've run this with a check for multiple devices and ret/2==wcslen for all requests so far.

nanpuyue:
The device is `096e:0209 Feitian Technologies, Inc. ROCKEY ARM`:
https://www.ftsafe.com.cn/products/rockey/ROCKEY-ARM/ROCKEY-ARM

akallabeth:
could you check the contents of the first 2 bytes of the buffer?
The returned data should consist of the data outlined here: https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/usbspec/ns-usbspec-_usb_string_descriptor

nanpuyue:
@akallabeth

nanpuyue:
`bLength` = `30`

akallabeth:
@nanpuyue thanks, I´ve added the check to the pr I´ve just opened (a rewrite of the usb redirection) here #5745

PR #5745, opened 2019-11-22
Urbdrc rewrite

akallabeth:
This PR is a major rewrite of the USB redirection channel.
* Removes the insane thread use
* Fixes the (very) misleading log messages

hope somebody will find it useful ;)


docds:
4APT

akallabeth:
@bmiklautz @hardening @mfleisz @nfedera anyone want to review/merge this?

channels/urbdrc/client/data_transfer.c
```diff
@@ -21,419 +21,361 @@
...
 }
 
-#if ISOCH_FIFO
-
-static int func_check_isochronous_fds(IUDEVICE* pdev)
+static UINT urb_write_completion(IUDEVICE* pdev, URBDRC_CHANNEL_CALLBACK* callback, BOOL noAck,
+                                 wStream* out, UINT32 InterfaceId, UINT32 MessageId,
+                                 UINT32 RequestId, UINT32 usbd_status, UINT32 OutputBufferSize)
 {
-	int ret = 0;
```
hardening: I'd suggest `out` to be `wStream **` instead of `wStream *`, so that this function can clear the pointer reference.
Perhaps this stream could also come from a `StreamPool`, as we will call it multiple times with stream with quite the same size...

channels/urbdrc/client/data_transfer.c
```diff
@@ -21,419 +21,361 @@
...
-	BYTE* data_temp;
-	UINT32 size_temp, process_times = 2;
-	ISOCH_CALLBACK_QUEUE* isoch_queue = NULL;
-	ISOCH_CALLBACK_DATA* isoch = NULL;
-	URBDRC_CHANNEL_CALLBACK* callback;
+	if (!out)
```
hardening: not sure it's pertinent to check `out` in this static function (I guess the caller has already checked that)

channels/urbdrc/client/data_transfer.c
```diff
@@ -21,419 +21,361 @@
...
+		return ERROR_INVALID_PARAMETER;
 
-	isoch_queue = (ISOCH_CALLBACK_QUEUE*)pdev->get_isoch_queue(pdev);
-
-	while (process_times)
+	if (Stream_Capacity(out) < OutputBufferSize + 36)
 	{
```
hardening: why not using `Stream_EnsureRemainingCapacity()` ?

akallabeth: size should already be known (and allocated) at this point

AzureusNation:
When will this be merged? This is one of the most wanted PRs to fix usb redirection

PR #5756, opened 2019-11-28
Added data length check for RDP_CODEC_ID_NONE

akallabeth:


client/X11/xf_gdi.c
```diff
@@ -1076,7 +1084,6 @@ static BOOL xf_gdi_surface_bits(rdpContext* context, const SURFACE_BITS_COMMAND*
 
 		default:
 			WLog_ERR(TAG, "Unsupported codecID %" PRIu16 "", cmd->bmp.codecID);
-			ret = TRUE;
 			goto fail;
```
hardening: shall we fail and disconnect when we receive an unknown codec ?

akallabeth: well, yes.
otherwise you'll end up with a very distorted screen anyway.

hardening: hum, ok I'm seeing it more like "I've not announced that codec, so why are you using it with me ?". Ok.

PR #5760, opened 2019-11-29
primitives: use host pointers for openCL

hardening:
Using host pointers may skip the need for copying buffers.

### DIFF
--- a/libfreerdp/primitives/primitives.c
+++ b/libfreerdp/primitives/primitives.c
@@ -247,7 +247,7 @@ static BOOL primitives_autodetect_best(primitives_t* prims)
 			goto out;
 		}
 
-		WLog_DBG(TAG, " * %s\t= %" PRIu32, cur->name, cur->count);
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

libfreerdp/primitives/primitives.c
```diff
@@ -247,7 +247,7 @@ static BOOL primitives_autodetect_best(primitives_t* prims)
 			goto out;
 		}
 
-		WLog_DBG(TAG, " * %s\t= %" PRIu32, cur->name, cur->count);
+		WLog_DBG(TAG, " * %s= %" PRIu32, cur->name, cur->count);
```
akallabeth: I´ve run into issues with the performance test.
On one of my machines opencl did not initialize and the whole test is failing the session.
Could you make this fail silent?

hardening: ok will look at that.

PR #5768, opened 2019-12-06
Timezone update

### LINKED ISSUE #5767: Windows server can't recognize time zone
**Describe the bug**
When connected to a Windows server, it fails to detect the local timezone, resulting in a wrong timeclock (+1 hour from the correct time)

**To Reproduce**
Connect to an windows server (any version) using freerdp and tz set to America/Sao_paulo

**Expected behavior**
It should detect the correct timezone config from the local one.
In my case the timezone is America/Sao_paulo (UTC -03:00), this may be related to a recent change in my country rules, which now doesn't have daylight saving anymore (since this year).

I'm sure this is a freerdp problem, since connecting with another software (rdesktop) it'll detect the correct timezone

**Screenshots**
Freerdp
freerdp

Rdesktop
rdesktop

** Application details
Tested with manjaro and debian 10
v: freerdp-1:2.0.0_rc4-7 (Manjaro)
v: 2.0.0~git20190204.1.2693389a+dfsg1-1 (debian 10)

* Command line used
xfreerdp /v: ip /size:90% /u: user /p: password

* OS version connecting to
Windows server 2008 R2 / 2012 /  2012 Standard

akallabeth:
Fixes #5767


akallabeth:
@bmiklautz @hardening anyone interested in merging this?

PR #5780, opened 2019-12-12
Progressive refactor

akallabeth:
* Removes a lot of direct data casts from wire data
* Adds state tracking discarding frames that are out of order
* Lots of additional checks


akallabeth:
@hardening @nfedera this pr adds the ability to decode `PROGRESSIVE` without ` Reduce-Extrapolate DWT` so it should improve compatibility with ogon in that regard (there are the artefacts the spec is talking about left but the session image is now usable)

libfreerdp/codec/progressive.c
```diff
@@ -361,18 +362,80 @@ static INLINE BOOL progressive_set_surface_data(PROGRESSIVE_CONTEXT* progressive
 	return TRUE;
 }
 
-static INLINE void* progressive_get_surface_data(PROGRESSIVE_CONTEXT* progressive, UINT16 surfaceId)
+static INLINE PROGRESSIVE_SURFACE_CONTEXT*
+progressive_get_surface_data(PROGRESSIVE_CONTEXT* progressive, UINT16 surfaceId)
 {
-	ULONG_PTR key;
+	void* key = (void*)(((ULONG_PTR)surfaceId) + 1);
 	void* pData = NULL;
```
hardening: Would be nice to index the `progressive->SurfaceContexts` hashtable with the address of `PROGRESSIVE_SURFACE_CONTEXT->id` like we did in https://github.com/FreeRDP/FreeRDP/blob/master/channels/geometry/client/geometry_main.c#L484 and https://github.com/FreeRDP/FreeRDP/blob/master/channels/geometry/client/geometry_main.c#L229 . That would allow us to remove the `+1` hack for the surfaceId.

akallabeth:
@mfleisz @bmiklautz anyone wants to merge this one? connections to ogon with only `/gfx` are much more usable that way ;)

libfreerdp/codec/progressive.c
```diff
@@ -392,38 +455,103 @@ static PROGRESSIVE_SURFACE_CONTEXT* progressive_surface_context_new(UINT16 surfa
...
+	}
 
-		if (tile->data)
-			_aligned_free(tile->data);
+	t = &surface->tiles[zIdx];
 
-		if (tile->sign)
-			_aligned_free(tile->sign);
+	if (upgrade)
+	{
```
hardening: I can't see any difference between the `if` and the `else` statement, am I missing something ?

akallabeth: @hardening some of the struct members are different. was too lazy to extract them from the if/else struct

PR #5781, opened 2019-12-13
Revert breaking changes from #5751

akallabeth:
The struct definitions must follow the MS definitions to stay
source compatible.

### DIFF
--- a/winpr/include/winpr/wtypes.h.in
+++ b/winpr/include/winpr/wtypes.h.in
@@ -309,7 +309,7 @@ typedef union _ULARGE_INTEGER
 	{
 		DWORD LowPart;
 		DWORD HighPart;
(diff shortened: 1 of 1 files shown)


akallabeth:
@hardening I´ve seen you talked about this with @nfedera . Is this pr still required or can it be closed?

bmiklautz:
@hardening any objections?

akallabeth:
@hardening waiting for your comment ;)

hardening:
I've nothing against reverting even the ms doc says that we should keep my changes:
https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-ularge_integer~r1

akallabeth:
@hardening ok, closing then.

PR #5783, opened 2019-12-16
Added /gfx and /gfx-h264 option mask=<value>

### LINKED ISSUE #5771: Commandline switch for RDP Version
The current implementation of the GFX parameters is confusing assuming you are connecting to a Windows 2016/2019 server and want to use the RemoteFX Codec. This is a common use case, since not all servers can handle AVC444 load and RemoteFX Progressive Codec is too slow on thin clients.

To activate RemoteFX (non-progressive) Codec you have to set /gfx:AVC420.

Reason:
RDP versions > 8.1 ignore the RDPGFX_CAPS_FLAG_THINCLIENT flag. But the only way to tell the server to use RemoteFX (non-progressive) Codec is to set the RDPGFX_CAPS_FLAG_THINCLIENT flag, which is on by default in FreeRDP.
The parameter /gfx:AVC420 is the only parameter to set RDP Version to 8.1 max. Every other GFX parameter advertises RDP10.x capabilities. 

A command line parameter to set GfxCapsFilter could solve this issue. But in this case no gfx parameter should filter out RDP versions (e.g. AVC420 ;))

1. An alternative would be to filter out RDP10.x capabilities if /gfx-thin-client is set. But in this case the new flag RDPGFX_CAPS_FLAG_AVC_THINCLIENT needs a new command line switch. Something like /gfx-avc-thin-client. 
1. /gfx:rfx could filter out RDP10.x and only advertise RDP8.1

akallabeth:
With this new option the gfx capability set can be masked to only
announce older versions of the protocol if required.
(Fixes #5771)


client/common/cmdline.c
```diff
@@ -2506,12 +2528,36 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
+						settings->GfxAVC444 = TRUE;
+					}
+					else if (_strnicmp("AVC420", val, 7) != 0)
+						rc = COMMAND_LINE_ERROR;
+					else if (_strnicmp("mask:", val, 5) == 0)
+					{
+						ULONGLONG v;
+						const char* uv = &val[5];
+						if (!value_to_uint(uv, &v, 0, UINT32_MAX))
+							rc = COMMAND_LINE_ERROR;
```
kubistika: I think you can exit the loop here.

client/common/cmdline.c
```diff
@@ -2506,12 +2528,36 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
 				{
-					settings->GfxAVC444 = TRUE;
+					const char* val = p[x];
+
+					if (_strnicmp("AVC444", val, 7) == 0)
+					{
+						settings->GfxAVC444 = TRUE;
+					}
+					else if (_strnicmp("AVC420", val, 7) != 0)
+						rc = COMMAND_LINE_ERROR;
```
kubistika: I think you can exit the loop here.

client/common/cmdline.c
```diff
@@ -2450,28 +2450,50 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
+						settings->GfxAVC444 = FALSE;
+						settings->GfxH264 = FALSE;
+						settings->RemoteFxCodec = TRUE;
+					}
+					else if (_strnicmp("mask:", val, 5) == 0)
+					{
+						ULONGLONG v;
+						const char* uv = &val[5];
+						if (!value_to_uint(uv, &v, 0, UINT32_MAX))
+							rc = COMMAND_LINE_ERROR;
```
kubistika: I think you can exit the loop here.

akallabeth: No, the arguments may appear in any order.

client/common/cmdline.c
akallabeth: No, the arguments may appear in any order.

kubistika: I meant that you can exit in this case because rc = COMMAND_LINE_ERROR, so the whole operation will fail either way - there is no way to continue looping the modifing the settings struct

client/common/cmdline.h
```diff
@@ -132,10 +132,10 @@ static const COMMAND_LINE_ARGUMENT_A args[] = {
...
 	  "Consume multitouch input locally" },
 #ifdef WITH_GFX_H264
-	{ "gfx", COMMAND_LINE_VALUE_OPTIONAL, "[RFX|AVC420|AVC444]", NULL, NULL, -1, NULL,
-	  "RDP8 graphics pipeline" },
-	{ "gfx-h264", COMMAND_LINE_VALUE_OPTIONAL, "[AVC420|AVC444]", NULL, NULL, -1, NULL,
-	  "RDP8.1 graphics pipeline using H264 codec" },
+	{ "gfx", COMMAND_LINE_VALUE_OPTIONAL, "[[RFX|AVC420|AVC444],mask=<value>]", NULL, NULL, -1,
+	  NULL, "RDP8 graphics pipeline" },
+	{ "gfx-h264", COMMAND_LINE_VALUE_OPTIONAL, "[[AVC420|AVC444],mask=<value>]", NULL, NULL, -1,
+	  NULL, "RDP8.1 graphics pipeline using H264 codec" },
```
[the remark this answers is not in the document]
akallabeth: @mmorvai good catch

akallabeth:
@bmiklautz @mfleisz anyone want to merge this?

PR #5794, opened 2019-12-18
Fix pkg-config check for libswscale

floppym:

### DIFF
--- a/cmake/FindSWScale.cmake
+++ b/cmake/FindSWScale.cmake
@@ -2,7 +2,7 @@
 include(FindPkgConfig)
 
 if (PKG_CONFIG_FOUND)
-	pkg_check_modules(SWScale swscale)
+	pkg_check_modules(SWScale libswscale)
 endif()
 
 find_path(SWScale_INCLUDE_DIR libswscale / swscale.h PATHS $ {SWScale_INCLUDE_DIRS})


akallabeth:
@floppym where did you need the lib prefix for pkgconfig?
But there seems to be an issue with the find_path below indeed.

floppym:
ffmpeg installs `libswscale.pc`, not `swscale.pc`. A quick test confirms this:
```
% pkg-config --libs swscale
Package swscale was not found in the pkg-config search path.
Perhaps you should add the directory containing `swscale.pc'
to the PKG_CONFIG_PATH environment variable
Package 'swscale', required by 'virtual:world', not found
% pkg-config --libs libswscale
-lswscale
```
Also, you can confirm this by comparing the cmake output with/without my patch:
```
% cmake -DWITH_SWSCALE=ON -B build
...
-- Checking for module 'swscale'
--   Package 'swscale', required by 'virtual:world', not found
```
```
% cmake -DWITH_SWSCALE=ON -B build
...
-- Checking for module 'libswscale'
--   Found libswscale, version 5.5.100
```

akallabeth:
@floppym I'm asking because detection works on my dev machine (afk, so can not verify which pc file it detects)
So, which os / version? Will check tomorrow.

floppym:
This is on Gentoo Linux with ffmpeg-4.2.1 installed.
```
% cat /usr/lib64/pkgconfig/libswscale.pc
prefix=/usr
exec_prefix=${prefix}
libdir=/usr/lib64
includedir=/usr/include

Name: libswscale
Description: FFmpeg image rescaling library
Version: 5.5.100
Requires:
Requires.private: libavutil >= 56.31.100
Conflicts:
Libs: -L${libdir}  -lswscale
Libs.private: -lm
Cflags: -I${includedir}
```

floppym:
My guess is that is that the pkgconfig check fails on your dev machine, but cmake is able to find the header/library in the default location via find_path / find_library anyway.

akallabeth:
@floppym thank you, you were right. Fixed the whole script and opened a new pr.

PR #5798, opened 2019-12-24
settings.c: fix memory leak in freerdp_settings_new

ckelsel:


kubistika:
@akallabeth @hardening wouldn't it be better to call `freerdp_settings_free` instead of just writing the free() calls again and again? In that way if the settings struct changes the only function that needs to be changed regarding freeing the data is `freerdp_settings_free`

akallabeth:
@kubistika you´re right, better to call `freerdp_settings_free` and ensure it does proper cleanup.

akallabeth:
@kubistika thx for the hint, added this to to #5813

PR #5801, opened 2019-12-26
server: proxy: improve logging

kubistika:
When more than one client is connected to the proxy, it gets hard to understand which log is related to which client. This PR solves this issue by creating a unique session ID to each proxy session. Also, this PR makes sure that all proxy logs are in the same format ([session_id][func_name]: message) so it will be easy to parse the logs and forward them to splunk/elastic/whatever.

### DIFF
--- a/winpr/libwinpr/thread/thread.c
+++ b/winpr/libwinpr/thread/thread.c
@@ -558,7 +558,7 @@ BOOL ThreadCloseHandle(HANDLE handle)
 
 		if ((thread->started) && (WaitForSingleObject(thread, 0) != WAIT_OBJECT_0))
 		{
-			WLog_ERR(TAG, "Thread running, setting to detached state!");
+			WLog_DBG(TAG, "Thread running, setting to detached state!");
 			thread->detached = TRUE;
 			pthread_detach(thread->thread);
 		}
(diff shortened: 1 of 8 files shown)


winpr/libwinpr/thread/thread.c
```diff
@@ -558,7 +558,7 @@ BOOL ThreadCloseHandle(HANDLE handle)
 
 		if ((thread->started) && (WaitForSingleObject(thread, 0) != WAIT_OBJECT_0))
 		{
-			WLog_ERR(TAG, "Thread running, setting to detached state!");
+			WLog_DBG(TAG, "Thread running, setting to detached state!");
 			thread->detached = TRUE;
 			pthread_detach(thread->thread);
```
hardening: Not related to this PR, but I'm wondering if there's a possibility for a race condition here if the thread terminates just after the `WaitForSingleObject` call (what's the behaviour of `pthread_detach` on a terminated thread ?)

kubistika: According to [this](https://stackoverflow.com/questions/51882300/calling-pthread-detach-for-an-already-exited-thread?answertab=votes#tab-top) I don't think that there is a problem with calling pthread_detach, but I do think that the other branch that calls `cleanup_handle` won't be executed (?)

akallabeth: @hardening @kubistika no, the `cleanup_handle` is also called for threads exiting prematurely (see `thread_launcher`)

kubistika:
@akallabeth good to see you back ;)
lets merge please ^^

akallabeth:
@kubistika sometimes holidays are required ;)
some rebasing is in order though before merging ^^

PR #5808, opened 2019-12-31
server: proxy: rewrite modules api

kubistika:


include/freerdp/server/rail.h
```diff
@@ -130,7 +130,7 @@ struct _rail_server_context
 	psRailServerGetAppidRespEx ServerGetAppidRespEx;
 
 	RailServerPrivate* priv;
-	rdpContext* rdpContext;
+	rdpContext* rdpcontext;
 };
```
hardening: This renaming sounds like a bad idea

kubistika: because?
it is a must because it causes a compilation error when mixing c and c++ code (typedef changes meaning of... blabla warning)

kubistika:
@akallabeth @hardening BTW, the CI failed because the compiler uses an old version of the c++ standard. can you please update it to support c++ 17?

channels/rail/server/rail_main.c
```diff
@@ -1360,8 +1360,8 @@ static DWORD WINAPI rail_server_thread(LPVOID arg)
 		}
 	}
 
-	if (error && context->rdpContext)
```
akallabeth: No good idea to rename existing context variables

include/freerdp/server/rail.h
akallabeth: agree with @hardening

server/proxy/freerdp_proxy.c
```diff
@@ -24,32 +24,56 @@
...
-
-	if (!config)
-		return -1;
 
 	if (argc > 1)
 		cfg = argv[1];
 
-	if (!pf_modules_init())
+	/* Register SIGINT handler for graceful termination */
+	signal(SIGINT, sigint_handler);
```
akallabeth: sure this is the only signal to listen to?

PR #5815, opened 2020-01-07
proxy: Add received response in case of an error

mfleisz:
Adds detailed information if the received proxy http reply is not as expected.

### DIFF
--- a/libfreerdp/core/proxy.c
+++ b/libfreerdp/core/proxy.c
@@ -388,6 +388,18 @@ BOOL proxy_connect(rdpSettings* settings, BIO* bufferedBio, const char* proxyUse
 	}
 }
 
+static const char* get_response_header(char* response)
+{
+	char* current_pos = strchr(response, '\r');
+	if (!current_pos)
+		current_pos = strchr(response, '\n');
+
+	if (current_pos)
+		*current_pos = '\0';
+
+	return response;
+}
+
 static BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 port)
 {
 	int status;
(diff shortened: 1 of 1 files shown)


libfreerdp/core/proxy.c
```diff
@@ -425,7 +437,7 @@ static BOOL http_proxy_connect(BIO* bufferedBio, const char* hostname, UINT16 po
 	{
 		if (resultsize >= sizeof(recv_buf) - 1)
 		{
-			WLog_ERR(TAG, "HTTP Reply headers too long.");
+			WLog_ERR(TAG, "HTTP Reply headers too long: %s", remove_crlf_in_string(recv_buf));
```
akallabeth: not really sure it is a good idea to dump whole requests/responses in production code, it might contain session keys / ...

libfreerdp/core/proxy.c
```diff
@@ -388,6 +388,18 @@ BOOL proxy_connect(rdpSettings* settings, BIO* bufferedBio, const char* proxyUse
 	}
 }
 
+static const char* remove_crlf_in_string(char* str)
+{
+	char* current_pos = strstr(str, CRLF);
+	while (current_pos)
+	{
+		*current_pos = '\\';
+		*(current_pos + 1) = 'n';
```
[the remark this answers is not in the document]
mfleisz: No - this will replace the line break with a a string representation of \n (the character '\' and the character 'n')

libfreerdp/core/proxy.c
mfleisz: I don't think that the proxy response will contain any sensible data. This will only contain the HTTP response of the connect request to the proxy and we only output the response in case of an error (so there can't be any active session).

mfleisz: Proposal: I could change the code so that only the first line (the status code) will be printed (i.e. replace the first occurance of \n or \r with a \0 in recv_buf and print that).

PR #5835, opened 2020-01-14
Implemented chroma subsampling for planar non RLE

akallabeth:
Allow planar data with chroma subsampling to be decoded.

### DIFF
--- a/libfreerdp/codec/planar.c
+++ b/libfreerdp/codec/planar.c
@@ -87,6 +87,126 @@ static INLINE INT32 planar_skip_plane_rle(const BYTE* pSrcData, UINT32 SrcSize, 
 	return (INT32)(pRLE - pSrcData);
 }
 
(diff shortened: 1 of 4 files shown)


akallabeth:
@hardening thx. works basically but colors are still a bit distorted (need to find out what MS means with the `DRAW_ALLOW_DYNAMIC_COLOR_FIDELITY` bug.

akallabeth:
@hardening ok, figured it out. Have now working code paths for `RGB` and `YCoCg` with `chroma subsampling` but did not manage to get a version that has no subsampling.
Do you know some kind of test for that?

akallabeth:
@hardening additionally, do you know how to provoke a server to send planar over gfx? I´m currently only testing with `/bpp:32` and `settings->DrawAllowDynamicColorFidelity = TRUE;` for `YCoCg` and `settings->DrawAllowDynamicColorFidelity = FALSE;` for `RGB`

PR #5836, opened 2020-01-15
fixes: rail and update issues

kubistika:


channels/rdpsnd/server/rdpsnd_main.c
```diff
@@ -867,6 +867,9 @@ void rdpsnd_server_context_free(RdpsndServerContext* context)
 	if (context->priv->input_stream)
 		Stream_Free(context->priv->input_stream, TRUE);
 
+	if (context->server_formats)
```
akallabeth: if () for free is unnecessary.

kubistika: the API of the rdpsnd server is quite bad. server->formats is being set by the caller (for example in shadow_channels.c), and it might be unknown if the caller of `server_rdpsnd_get_formats` should free the formats and mark it NULL / they should be freed by rdpsnd_server_context_free. that's why I've added the if statement. I'll remove it if you think there's really no reason for it ^^

akallabeth:
@freerdp-bot test

Issue #5804, opened 2019-12-27
Using sound in windows leads to client crash after some time.


vscfreire:
**Description**
Using sound in windows leads to client crash after some time.

**To Reproduce**
1. Connect to a server with sound support using the /sound switch.
2. Open a really long youtube video
3. Wait for some time. I can consistently reproduce the crash after about 1 hour, sometimes less, sometimes more.

** Application details
* Nightly build 710, win32, vs2013
* Windows Server 2016

**Desktop:**
 - OS: Windows 10

**Additional context**
Came across these bug while testing for issue #5514

akallabeth:
#5818 should fix the issues (and your sound problems in general).
Looks like a buffer was freed up prematurely

vscfreire:
Testing nightly build 718: The sound is much improved but the client freezes after a while.

akallabeth:
@vscfreire well, some log would be interesting ;)

vscfreire:
@akallabeth unfortunately the log isn't very helpful as the client freezes, requiring to force close it. This looks like something between the client and the sound backend because until I force close it, I can't even open Windows volume mixer.

akallabeth:
@vscfreire I´m closing this as resolved, should you experience another sound related hang feel free to create a new issue.

PR #5840, opened 2020-01-16
Fixes in windows sound backends

mfleisz:
This PR fixes following issues:

- Possible crash in audin_winmm if no audio device is available (log instance not initialized)
- Possible crash in rdpsnd_winmm if no audio device is available (log instance not initialized)
- Fix possible deadlock in rdpsdn_winmm caused by calling waveOut* functions from within the provided callback (see remarks: https://docs.microsoft.com/en-us/previous-versions/dd743869(v=vs.85))


mfleisz:
This should also fix #5804

PR #5842, opened 2020-01-16
server.c: fix leak while freeing rdpPeerChannel->queue

kubistika:
in server.c:92 a message is malloced and posted to the queue, but if won't be processed, a leak will occur when MessageQueue_Free will be called.

### DIFF
--- a/libfreerdp/core/server.c
+++ b/libfreerdp/core/server.c
@@ -990,6 +990,47 @@ static void peer_channel_queue_free_message(void* obj)
 	free(obj);
 }
 
+static void channel_free(rdpPeerChannel* channel)
+{
+	if (!channel)
+		return;
+
+	MessageQueue_Free(channel->queue);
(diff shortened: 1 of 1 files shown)


libfreerdp/core/server.c
```diff
@@ -1033,6 +1038,7 @@ HANDLE WINAPI FreeRDP_WTSVirtualChannelOpen(HANDLE hServer, DWORD SessionId, LPS
 
 	if (!channel)
 	{
+		wObject queueCallbacks;
```
akallabeth: Missing initialization `wObject queueCallbacks = { 0 };`

kubistika: fixed it, thank you :))

kubistika:
@akallabeth nice!

kubistika:
@akallabeth should we merge this? ^^

PR #5844, opened 2020-01-16
Gfx fix disconnect reorder

akallabeth:


akallabeth:
@kubistika I´ve reverted your changes in gfx channel but kept the cleanup for your use case too.
could you check it still works?

kubistika:
@akallabeth I can confirm that it still works as expected.

PR #5838, opened 2020-01-16
core: rdp: reset rdp->nla in rdp_reset

kubistika:
This fixes a memory leak when freerdp_connect fails, and then called again with NlaSecurity = FALSE. Found this while testing the proxy (AllowFallbackToTls feature) compiled with address sanitizer. Output was:

Indirect leak of 72 byte(s) in 1 object(s) allocated from:
    #0 0x7fba283e0ce6 in calloc (/lib/x86_64-linux-gnu/libasan.so.5+0x10dce6)
    #1 0x7fba27892d59 in sspi_CredentialsNew /home/blabla/git/kubistika/FreeRDP/winpr/libwinpr/sspi/sspi_winpr.c:217
    #2 0x7fba2788e8f1 in negotiate_AcquireCredentialsHandleA /home/blabla/git/kubistika/FreeRDP/winpr/libwinpr/sspi/Negotiate/negotiate.c:449
    #3 0x7fba27896392 in winpr_AcquireCredentialsHandleA /home/blabla/git/kubistika/FreeRDP/winpr/libwinpr/sspi/sspi_winpr.c:823
    #4 0x7fba27bac575 in nla_client_init /home/blabla/git/kubistika/FreeRDP/libfreerdp/core/nla.c:401
    #5 0x7fba27bac824 in nla_client_begin /home/blabla/git/kubistika/FreeRDP/libfreerdp/core/nla.c:431
    #6 0x7fba27c7a648 in transport_connect_nla /home/blabla/git/kubistika/FreeRDP/libfreerdp/core/transport.c:342
    #7 0x7fba27bbd46c in nego_security_connect /home/blabla/git/kubistika/FreeRDP/libfreerdp/core/nego.c:248
    #8 0x7fba27bbd15d in nego_connect /home/blabla/git/kubistika/FreeRDP/libfreerdp/core/nego.c:220
    #9 0x7fba27c45df0 in rdp_client_connect /home/blabla/git/kubistika/FreeRDP/libfreerdp/core/connection.c:332
    #10 0x7fba27c19963 in freerdp_connect /home/blabla/git/kubistika/FreeRDP/libfreerdp/core/freerdp.c:197
    #11 0x56054439f559 in pf_client_connect /home/blabla/git/kubistika/FreeRDP/server/proxy/pf_client.c:342
    #12 0x56054439f9bc in pf_client_thread_proc /home/blabla/git/kubistika/FreeRDP/server/proxy/pf_client.c:392
[...]

PR #5847, opened 2020-01-19
Improve proxy shutdown

kubistika:
Reverted #5838, simply because it was a mistake in understanding how rdpNla*, rdpTls* instances are managed. After digging, I found that it is allocated in transport_connect_nla and freed by freerdp_disconnect. Furthermore, I found that the output from ASAN, that pointed to this leak, was actually caused because the proxy **itself** didn't close all active connections gracefully.

### DIFF
--- a/libfreerdp/core/rdp.c
+++ b/libfreerdp/core/rdp.c
@@ -1857,13 +1857,11 @@ void rdp_reset(rdpRdp* rdp)
 	license_free(rdp->license);
 	transport_free(rdp->transport);
 	fastpath_free(rdp->fastpath);
-	nla_free(rdp->nla);
 	rdp->transport = transport_new(context);
 	rdp->license = license_new(rdp);
 	rdp->nego = nego_new(rdp->transport);
 	rdp->mcs = mcs_new(rdp->transport);
 	rdp->fastpath = fastpath_new(rdp);
-	rdp->nla = nla_new(rdp->instance, rdp->transport, rdp->settings);
 	rdp->transport->layer = TRANSPORT_LAYER_TCP;
 	rdp->errorInfo = 0;
 	rdp->deactivation_reactivation = 0;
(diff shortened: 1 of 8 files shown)


libfreerdp/core/nla.c
```diff
@@ -2451,6 +2451,7 @@ void nla_free(rdpNla* nla)
 	sspi_SecBufferFree(&nla->ClientNonce);
 	sspi_SecBufferFree(&nla->PublicKey);
 	sspi_SecBufferFree(&nla->tsCredentials);
+	sspi_SecBufferFree(&nla->negoToken);
```
akallabeth: ouch, this will break redirect/gateway related stuff

libfreerdp/core/rdp.c
```diff
@@ -1857,13 +1857,11 @@ void rdp_reset(rdpRdp* rdp)
 	license_free(rdp->license);
 	transport_free(rdp->transport);
 	fastpath_free(rdp->fastpath);
-	nla_free(rdp->nla);
```
akallabeth: Same here, this is recreated to reset to initial state.

kubistika: no, this is an addition i added in  #5838. is was a mistake, i explained why. this struct is created in transport_connect_nla and does not need to be recreated here

libfreerdp/core/nla.c
kubistika: can you explain?
this buffer shouldn't be freed when the whole rdpNla* struct is freed?

kubistika: BTW I checked and xfreerdp ables to be redirected after this change, i didn't see anything wrong

libfreerdp/core/rdp.c
akallabeth: @kubistika oops, forgot ^^

libfreerdp/core/nla.c
akallabeth: @kubistika no, this is most of the time just a pointer to another buffer and this will be a double free

kubistika: okay, reverted

PR #5857, opened 2020-01-23
Scanbuild fixes

akallabeth:
Fix a few warnings detected with clang scanbuild.


server/proxy/pf_modules.c
```diff
@@ -74,7 +74,7 @@ static const char* pf_modules_get_hook_type_string(PF_HOOK_TYPE result)
 BOOL pf_modules_run_hook(PF_HOOK_TYPE type, proxyData* pdata)
 {
 	BOOL ok = TRUE;
-	size_t index;
+	int index;
```
kubistika: Don't you think it will be better to cast the return value from `ArrayList_Count` to `size_t` instead of using `int` as indexer type? Would require less changes (and this is not considered as an API change because proxy is the only one uses this mcaro).

PR #5858, opened 2020-01-23
[WIP] Updated dependency and channel defaults for detection

akallabeth:
Update defaults for channels and dependency detection.


akallabeth:
@freerdp-bot test

PR #5859, opened 2020-01-23
core: Always send CS_MULTITRANSPORT PDU to server

mfleisz:
This PDU is required by Microsoft servers in order for bandwidth
management to work correctly. Even if we do not support multi-transport
for now we should just send a PDU with flags set to 0 to enable correct
handing of bandwidth measurement PDUs.


hardening:
You mean, we don't support multitransport yet ;)

mfleisz:
@hardening exactly ;-)

PR #5868, opened 2020-02-03
proxy: server: support static vc proxy

kubistika:
This PR adds couple of features to the proxy:
* Proxy static virtual channel without the need to implement them in freerdp/proxy
* New API for proxy plugins that a plugin to filter/log the passthrough vc data:
  * ClientChannelData - client received data from server
  * ServerChannelData - server received data from peer

Also moved up PluginUnload callback so it won't break the ABI next time when the proxy plugins API will be extended.


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

kubistika:
@akallabeth sorry, was on vacation. nope, it can be merged.

server/proxy/modules/modules_api.h
```diff
@@ -44,6 +44,8 @@ typedef struct proxy_plugin
 	const char* name;        /* unique module name */
 	const char* description; /* module description */
 
+	BOOL (*PluginUnload)();
```
akallabeth: should this really be a function pointer with variable arguments?

PR #5906, opened 2020-02-25
Made CHANNEL_TSMF optional and deprecated.

akallabeth:
* Disable by default to avoid accidental use by people compiling.
* Added deprecation warnings everywhere to communicate that using this channel is not a good idea
* Moved dependency detection to channel as `gstreamer` is only used by that single channel and should therefore not be used if not compiled in


server/proxy/pf_channels.c
```diff
@@ -26,7 +26,9 @@
 #include <freerdp/gdi/gfx.h>
 
 #include <freerdp/client/rdpei.h>
+#if defined(CHANNEL_TSMF_CLIENT)
```
kubistika: I'd just remove the include, there's no use of TSMF in proxy and I don't see a reason there will ever be.

mfleisz:
LGTM except the include mentioned by @kubistika

akallabeth:
@freerdp-bot test

PR #5911, opened 2020-02-26
Do not advertise /usb in help output if the channel is not built

ondrejholy:
The help output advertise an option and an example for usb redirection
regardless of the fact whether the urbdrc channel was built or not.
This is confusing for people. Let's do not show /usb in help if it is
not built.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -432,7 +432,9 @@ BOOL freerdp_client_print_command_line_help_ex(int argc, char** argv,
 	printf("\n");
 	printf("Multimedia Redirection: /multimedia:sys:oss,dev:/dev/dsp1,decoder:ffmpeg\n");
 	printf("Multimedia Redirection: /multimedia:sys:alsa\n");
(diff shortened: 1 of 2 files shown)


hendwolt:
I know, it's a little bit late but, is it possible, that this works only if I compile with "-DBUILTIN_CHANNELS=OFF" ?

akallabeth:
@hendwolt what do you mean? The variable is defined anyway and has nothing to do with that. There was an issue with a missing `config.h` include that has been fixed though.

hendwolt:
I mean:
`xfreerdp /u:user /d:domain /f -clipboard -wallpaper /v:destination_w7 /rfx  /network:auto +fonts /gdi:hw /usb:id,dev:0d8c:000c     
[15:44:27:696] [18352:18352] [ERROR][com.winpr.commandline] - Failed at index 11 [/usb:id,dev:0d8c:000c]: Unexpected keyword`

But I'll try again with the next build.

PR #5914, opened 2020-02-26
Fix #5910: USB hotplug added device wrong

### LINKED ISSUE #5910: USB Redirection
When using RemoteFX usb redirection the first time a usb device is plugged in it redirects fine. You take the device out and it is removed from the device manager on the remote server as expected. However, if you go to plug the device in again it no longer redirects but is picked up on the OS of the computer. Running both as a normal user and as root causes this to happen.

akallabeth:
When a USB device was added via hotplug add_device was called
with wrong arguments.


AzureusNation:
If this could be merged today I would greatly appreciate this so I can continue rolling out thin clients that require this functionality.

akallabeth:
@AzureusNation good point with the device classes you mentioned in #5910. I´ve added a device class filter to the hotplug function which will prevent known classes that should never be redirected to be filtered from hotplug.

AzureusNation:
Hate to say this but checkout your branch for this fix and built FreeRDP but the issues still persist. Ive attached the output of the issue with plugging in a remotefx usb device.

The first entry adds the device correctly and the second one stops seeming half way through.
[RemoteFX Log.txt](https://github.com/FreeRDP/FreeRDP/files/4260581/RemoteFX.Log.txt)

akallabeth:
@AzureusNation ok, you attach 2 devices at the same time, correct?
Looks like a call to the second one is stuck. Could you add some log messages in `urbdrc_process_query_device_text` and see if it exits early or is stuck in `libusb_udev_control_query_device_text`?

AzureusNation:
Just noticed that myself. Going to get a log for you with nothing plugged in to start and then add and remove devices.

akallabeth:
@AzureusNation thank you. Added a fix for the failure case, it the removes the device from redirection.

AzureusNation:
[log.txt](https://github.com/FreeRDP/FreeRDP/files/4260709/log.txt)

This is a better log. Nothing plugged in to start with. Plugged in first time. Appears, second and third tie nothing (Same port). Changed port and still nothing on forth.

akallabeth:
@AzureusNation as for the storage device, are you sure? do you use the current pr with all commits? added the device class filter and that should exclude storage devices from this.

AzureusNation:
I believe so, i check out your branch for this fix.

AzureusNation:
Its really weird how it will register the first time but then after pulling and getting the release_interface error 0 = -5 and attach_kerne_driver 0 = -4. And any subsequent plugins of that same device will no longer work. I i plug in another device after removing this one that registers fine... odd.

akallabeth:
@AzureusNation ok, that is really weired.
Does the updated mass storage filter work though?

AzureusNation:
Just building this now to test since latest commit

AzureusNation:
Okay, you'll be glad to know that the usb drives are now being filtered :)

akallabeth:
@AzureusNation nice! Added a commit that makes the log messages a bit more useful.

AzureusNation:
@akallabeth let me know when you want me to test. Got all day today and tomorrow to do this so no rush.

Issue #5908, opened 2020-02-25
X11 event handling crash


user-72632:
**Describe the bug**
Copying a large file (200M) between a Debian 10 box and a Windows 10 VM, I received a crash at 93%. Reconnected, started transfer again and it worked the second time with no problems. On the Win10 VM, I was copying the file from a redirected folder on my Debian 10 box.

**To Reproduce**
Could not reproduce.

**Expected behavior**
File copy with no errors - second time worked

**Screenshots**
N/A - provided dump file.

**Application details**
* FreeRDP Version
This is FreeRDP version 2.0.0-dev5 (dc89923f4

* Command line used
$ /opt/freerdp-nightly/bin/xfreerdp +old-license /log-level:info /u:Administrator /p:Password /size:1600x800 +clipboard /drive:Temp,/home/user/Temp /admin /sec:nla /cert-ignore /v:192.168.0.XX

* output of `/buildconfig`
$ /opt/freerdp-nightly/bin/xfreerdp /buildconfig
This is FreeRDP version 2.0.0-dev5 (dc89923f4)
[...]

PR #5918, opened 2020-02-27
X11 event lock

akallabeth:
* Fixes X11 event lock issues (see #5908 )
* Use const arguments in event handling code
* Use appropriate type in event handling code to avoid misuse

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -496,17 +496,16 @@ static BOOL xf_process_x_events(freerdp* instance)
 	{
 		xf_lock_x11(xfc, FALSE);
 		pending_status = XPending(xfc->display);
-		xf_unlock_x11(xfc, FALSE);
 
 		if (pending_status)
 		{
 			ZeroMemory(&xevent, sizeof(xevent));
 			XNextEvent(xfc->display, &xevent);
 			status = xf_event_process(instance, &xevent);
-
-			if (!status)
-				return status;
 		}
+		xf_unlock_x11(xfc, FALSE);
+		if (!status)
+			break;
 	}
 
 	return status;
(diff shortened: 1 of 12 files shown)


akallabeth:
@user-72632 I´ve refactored the event handling code and think to have fixed the issue, could you give this branch a try and confirm?

user-72632:
@akallabeth sorry, I can't seem to build your branch. Lately, I have been having trouble just building master due to multiple definition of `rail_send_pdu' in proxy. I have been just waiting for the nightlies to check on things because I have not had time to look into all the problems. Thanks for looking into this.

akallabeth:
@user-72632 how do you build? this sounds strange as we build server & proxy quite often ourselfes.

user-72632:
I don't want to bother @akallabeth with this but here is my problems. I use a custom bash script to build and have been for years so I know it works. All builds when using WITH_SERVER=ON do not build to completion. I have attached build logs for Ubuntu 16.04 and Debian 10. When I do not use that directive it builds fine except for the many pragma warnings in all Ubuntu 16.04 builds which I assume is from an older gcc, but should still be fixed. From the errors, it looks to be in the proxy code/headers.

I could never get your branch to compile, don't know why sorry.

Ubuntu 16.04 build log with errors and pragma warnings:

[ubuntu-16.04-out-of-source-freerdp-build-with-errors-log.txt](https://github.com/FreeRDP/FreeRDP/files/4267935/ubuntu-16.04-out-of-source-freerdp-build-with-errors-log.txt)

Debian 10 build log with errors:

[debian-10-out-of-source-freerdp-build-with-errors-log.txt](https://github.com/FreeRDP/FreeRDP/files/4267938/debian-10-out-of-source-freerdp-build-with-errors-log.txt)

HTH

user-72632:
Good news! I did manage to compile the downloaded zip file of your branch and test it. Using my test case it did make it through the entire transfer the first time! I noticed a couple things during the transfer; although the reported transfer speed looked about the same as the previous times, the actual wall clock of the transfer seemed a little longer than the previous transfers. I also noticed that the bitrate during the transfer stayed even and orderly as opposed to the bursty nature of the previous transfers, which I believe is a good thong.

I never did get your git branch to compile. Looking into it a little deeper it appears there are some dependencies that I may not have that you are using. Not sure what the difference is in the git clone of your branch and the downloaded zip file is though.

In any event, thanks for all your work on this problem which appears to be fixed!

akallabeth:
@user-72632 interesting, just built current master with your cmake command line and do not get such errors, debian 10 here

user-72632:
Yes, that is. How are you syncing up with the latest dependencies required? For example, are you grepping from a Jenkins workspace or something else in order to stay current with the latest dependencies. The documentation for that has been out of date for a while now.

akallabeth:
@user-72632 well, gstreamer is only used by tsmf.
Rest depends on options you activate, have a look at the nightly debian package in the repo.

PR #5923, opened 2020-02-27
fix FindFFmpeg.cmake: locating avcodec not using pkgconfig variable

MagicalTux:
Fixing a simple typo that prevents avcodec from working if installed in a non-standard path returned by pkg-config


akallabeth:
@freerdp-bot test

akallabeth:
@MagicalTux thank you for the fix, looks like a nasty typo ^^

PR #5932, opened 2020-03-02
Fix RDPEGFX bitmap cache indexing

nfedera:
---

Please **don't merge** until Microsoft answers https://github.com/microsoft/WindowsProtocolTestSuites/issues/227

---

Microsoft seems to use 1-based indexing in the RDPGFX bitmap cache PDU's.
This does not seem to be documented but can be deducted from the RDP client test code in Microsoft's "Windows Protocol Test Suites" [RDP client test code](https://github.com/microsoft/WindowsProtocolTestSuites/tree/staging/TestSuites/RDP/Client/src/TestSuite) and the observation that mstsc aborts with a protocol error if the cacheSlot index value 0 is used in a GFX surface to cache PDU.

I've also renamed the RDPGFX_PLUGIN struct member `MaxCacheSlot` to `MaxCacheSlots` which is less confusing, since FreeRDP of course uses 0-based numbering internally.

### DIFF
--- a/channels/rdpgfx/client/rdpgfx_main.c
+++ b/channels/rdpgfx/client/rdpgfx_main.c
@@ -75,11 +75,11 @@ static void free_surfaces(RdpgfxClientContext* context, wHashTable* SurfaceTable
 	free(pKeys);
 }
 
-static void evict_cache_slots(RdpgfxClientContext* context, UINT16 MaxCacheSlot, void** CacheSlots)
+static void evict_cache_slots(RdpgfxClientContext* context, UINT16 MaxCacheSlots, void** CacheSlots)
 {
 	UINT16 index;
 
-	for (index = 0; index < MaxCacheSlot; index++)
+	for (index = 0; index < MaxCacheSlots; index++)
 	{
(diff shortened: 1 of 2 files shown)


nfedera:
Please don't merge until Microsoft answers https://github.com/microsoft/WindowsProtocolTestSuites/issues/227

hardening:
Does it mean that sometime MS servers use cache index `MaxCacheSlots` as it is a 1 based index (so index goes from 1 to `MaxCacheSlots` instead of 0 to `MaxCacheSlots - 1`) ?

nfedera:
@hardening wrote
> Does it mean that sometime MS servers use cache index `MaxCacheSlots` as it is a 1 based index (so index goes from 1 to `MaxCacheSlots` instead of 0 to `MaxCacheSlots - 1`) ?

Exactly, but it seems not "sometimes" but rather always.

akallabeth:
@mfleisz that was a bit too soon ;)

mfleisz:
@akallabeth Yes - realized so after merging it ... ^^

PR #5939, opened 2020-03-04
Reverted minimum scaling diff

akallabeth:
Revert to the diff value from @volth
@ikeboy does that fix the issue?


akallabeth:
@freerdp-bot test

PR #5950, opened 2020-03-06
[RDPEGFX] checks for cache import, prevent 0 size allocations

akallabeth:


channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -1104,9 +1104,9 @@ static UINT rdpgfx_recv_cache_import_offer_pdu(RdpgfxServerContext* context, wSt
 	/* cacheEntriesCount (2 bytes) */
 	Stream_Read_UINT16(s, pdu.cacheEntriesCount);
 
-	if (pdu.cacheEntriesCount <= 0)
+	/* 2.2.2.16 RDPGFX_CACHE_IMPORT_OFFER_PDU */
+	if (pdu.cacheEntriesCount > 5462)
```
mfleisz: 2.2.2.16 says "MUST be less than 5462" so this check should be `>=`

akallabeth: good catch

akallabeth:
@freerdp-bot test

akallabeth:
@hardening 
* `0` size allocations are not valid and will trigger compiler / sanitizer warnings
* The `cacheEntriesCount` value was not checked for validity according to spec. as it is client supplied information in a server side function this should be checked.

PR #5971, opened 2020-03-13
X11 selection improvements

sgtatham:
Here's a collection of changes that all improve handling of X11 selections. They're more or less independent as patches, but they all contribute to the same overall goal, which is that if I run an `xfreerdp` side by side with an `xtightvncviewer` on the same X display, then I can transfer clipboard data from the FreeRDP to the VNC viewer quickly and easily without anything going wrong.

### DIFF
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -68,6 +68,9 @@ struct xf_clipboard
 	Atom clipboard_atom;
 	Atom property_atom;
 
+	Atom timestamp_property_atom;
+	Time selection_ownership_timestamp;
+
 	Atom raw_transfer_atom;
 	Atom raw_format_list_atom;
 
@@ -111,6 +114,7 @@ struct xf_clipboard
 };
 
 static UINT xf_cliprdr_send_client_format_list(xfClipboard* clipboard);
+static void xf_cliprdr_set_selection_owner(xfContext* xfc, xfClipboard* clipboard, Time timestamp);
 
 static void xf_cliprdr_check_owner(xfClipboard* clipboard)
 {
@@ -758,6 +762,17 @@ static void xf_cliprdr_provide_targets(xfClipboard* clipboard, const XSelectionE
 	}
 }
 
+static void xf_cliprdr_provide_timestamp(xfClipboard* clipboard, const XSelectionEvent* respond)
+{
+	xfContext* xfc = clipboard->xfc;
+
+	if (respond->property != None)
+	{
+		XChangeProperty(xfc->display, respond->requestor, respond->property, XA_INTEGER, 32,
+		                PropModeReplace, (BYTE*)&clipboard->selection_ownership_timestamp, 1);
+	}
+}
+
 static void xf_cliprdr_provide_data(xfClipboard* clipboard, const XSelectionEvent* respond,
                                     const BYTE* data, UINT32 size)
 {
@@ -849,7 +864,9 @@ static BOOL xf_cliprdr_process_selection_request(xfClipboard* clipboard,
 
 	if (xevent->target == clipboard->targets[0]) /* TIMESTAMP */
 	{
-		/* TODO */
+		/* Someone else requests the selection's timestamp */
+		respond->property = xevent->property;
+		xf_cliprdr_provide_timestamp(clipboard, respond);
 	}
 	else if (xevent->target == clipboard->targets[1]) /* TARGETS */
 	{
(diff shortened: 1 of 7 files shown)


akallabeth:
@freerdp-bot test

client/common/cmdline.c
```diff
@@ -3173,6 +3173,11 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 			if (!settings->SmartcardLogon)
 				activate_smartcard_logon_rdp(settings);
 		}
+		CommandLineSwitchCase(arg, "x-selection-atom")
```
akallabeth: I´m against a new command line option for this, Could be added to `/clipboard:use-selection`

sgtatham: I did wonder about that myself, because I couldn't see a way to make this X11-specific CLI option only appear in the X11 version of the code. I'm happy to change the CLI syntax.

Could you clarify your alternative suggestion, please? I'm looking at the command-line code, and it doesn't look to me as if I can add `COMMAND_LINE_VALUE_OPTIONAL` to the existing clipboard option without removing the `COMMAND_LINE_VALUE_BOOL` it already has, which will break existing command lines that use the `+clipboard` or `-clipboard` syntax.

Are you suggesting doing that anyway, or have I misunderstood?

sgtatham:
(That force-push is just the clang-format fixes. I haven't tried to change the CLI syntax yet.)

akallabeth:
@freerdp-bot test

akallabeth:
@sgtatham indeed, this is an issue. and since clipboard is on by default the change is not that easy :/
will need to check how that can be done.

sgtatham:
How about that? It turns out that `COMMAND_LINE_VALUE_OPTIONAL` and  `COMMAND_LINE_VALUE_BOOL` can be made to coexist with only a one-line change in `winpr`. If that's acceptable, then your suggested CLI syntax can be implemented on top of that with no trouble.

sgtatham:
@akallabeth Sorry to nag, but is this still on the radar? I'm happy to do it a different way if you're not convinced by what I've done here.

sgtatham:
Sorry – shouldn't have pressed that button before checking the current status of the merge. Actually I need to fix conflicts before asking you to take another look.

sgtatham:
Right, _now_ I'd like a re-review, please. Sorry for pressing all the buttons in the wrong order.

I think the change I've made in generate_argument_docbook isn't quite right – for an option that can be false, true, or a string, I should be handling all of those possibilities sensibly in the default slot. But before I go to the effort, I'd like to know if the basic idea of allowing options to take that range of values is acceptable to you?

akallabeth:
@freerdp-bot test

client/X11/xf_cliprdr.c
```diff
@@ -531,6 +535,7 @@ static void xf_cliprdr_get_requested_targets(xfClipboard* clipboard)
 	CLIPRDR_FORMAT_LIST formatList = { 0 };
 	formats = xf_cliprdr_get_client_formats(clipboard, &numFormats);
 	formatList.msgFlags = CB_RESPONSE_OK;
+	formatList.msgType = CB_FORMAT_LIST;
```
akallabeth, line 538: unnecessary, already set 3 lines below

client/X11/xf_cliprdr.c
```diff
@@ -1099,6 +1127,7 @@ static UINT xf_cliprdr_send_client_format_list(xfClipboard* clipboard)
 	}
 
 	formatList.msgFlags = CB_RESPONSE_OK;
+	formatList.msgType = CB_FORMAT_LIST;
```
akallabeth, line 1130: duplicate, look 3 lines below

akallabeth:
@sgtatham the clang format issue still persists

client/X11/xf_cliprdr.c
sgtatham: Oh yes, your cmmit cbee45aace1bb4fd485633224db5433443a201fe has obsoleted one of mine.

sgtatham:
(Sorry for the double push – I spotted a typo in one of the commit messages at the last minute.)

This version is re-clang-formatted, and also I've added the missing case in `generate_argument_docbook.c` to print any possible default sensibly for a hybrid bool/string option.

akallabeth:
@freerdp-bot test

sgtatham:
@akallabeth Hmm, what's happened to this PR now? It's been silent for two weeks, and the last activity was that you told freerdp-bot to test the patch, and apparently it never reported any results. Is it possible to tell it to try again?

akallabeth:
@sgtatham sorry, did have a few other things to take care of first.

PR #5973, opened 2020-03-14
fix not accepted mask: after /gfx-h264:

ifsdk:

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -2549,8 +2549,6 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings, 
 						{
 							settings->GfxAVC444 = TRUE;
 						}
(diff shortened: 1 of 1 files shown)


ifsdk:
this should make work mask: not only after /gfx: but also there as the code suggests. if useless then the code cheking for mask should be removed there.

akallabeth:
@ifsdk no, it just moves a check down... This is done in a loop, order does not matter...

ifsdk:
check the command line, it doesn't work as the check point for avc420 there aborts

ifsdk:
so xfreerdp /gfx-h264 works only with :avc{444,420} but not with :mask in any place there

akallabeth:
Ah, I see, the check is wrong. but so is your pr ;)
Closing and creating a new one.

ifsdk:
just for clarity my pr isn't wrong, thou it would be better to delete the mask code there

PR #5974, opened 2020-03-15
Cmd avc420 fix

akallabeth:


akallabeth:
@ifsdk should fix your issue.

ifsdk:
@akallabeth good fix. but it's 6 more lines then my pr #5973;) also in the line 2531 there yet is
settings->GfxH264 = TRUE;
Also if it is deprecated as marked in the addition then the mask code for /gfx-h264 should be removed as it didn't work at all yet, and that would get reduction of useless code lines.

akallabeth:
The path did contain a few issues, it is now on par with /gfx.
Only because it is not the preferred method it should still work properly ;)

ifsdk:
@akallabeth but it didn't work properly, i.e. it didn't accept mask at all so far. so there is nothing to deprecate;)

akallabeth:
@ifsdk well, that was a bug and has nothing to do with the thing being deprecated ;)
one is a bug, the other information for users to start using the (unified) `/gfx` syntax

ifsdk:
@akallabeth imho the information then shuld look like
{ "gfx-h264", COMMAND_LINE_VALUE_OPTIONAL,
	  "[AVC420|AVC444] [DEPRECATED] use /gfx: instead", NULL, NULL, -1, NULL,
	  "RDP8.1 graphics pipeline using H264 codec" }

client/common/cmdline.c
```diff
@@ -2547,10 +2547,14 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 
 						if (_strnicmp("AVC444", val, 7) == 0)
 						{
+							settings->GfxH264 = TRUE;
```
mfleisz, line 2550: Not sure if we should enable GfxH264 here as well. IIRC the ms client only sends AVC444 supported flags but sets AVC420 to unsupported during capability exchange.

akallabeth: @mfleisz since we support it and it is only used with servers that actually fall back to these old caps it does not hurt.

mfleisz: > ... it is only used with servers that actually fall back to these old caps ...
That's the part I am actually not sure about ... ^^
Anyways everything else lgtm

PR #5979, opened 2020-03-18
Fixed `update_write_surfcmd_surface_bits` to use `CMDTYPE_SET_SURFACE_BITS`

viniciusjarina:
Although 2.2.9.2.1 Set Surface Bits Command (TS_SURFCMD_SET_SURF_BITS) and 2.2.9.2.2 Stream Surface Bits Command (TS_SURFCMD_STREAM_SURF_BITS) are almost the same and handled by the same codepath (https://github.com/FreeRDP/FreeRDP/blob/72ca88f49c1b7f6678a66c09fa5fe5506623ed35/libfreerdp/core/surface.c#L150) when mstsc connect to FreeRDP server (shadow) it doesn't update the screen using `CMDTYPE_STREAM_SURFACE_BITS` but it does using `CMDTYPE_SET_SURFACE_BITS`

### DIFF
--- a/libfreerdp/core/surface.c
+++ b/libfreerdp/core/surface.c
@@ -226,7 +226,7 @@ BOOL update_write_surfcmd_surface_bits(wStream* s, const SURFACE_BITS_COMMAND* c
 	if (!Stream_EnsureRemainingCapacity(s, SURFCMD_SURFACE_BITS_HEADER_LENGTH))
 		return FALSE;
 
-	Stream_Write_UINT16(s, CMDTYPE_STREAM_SURFACE_BITS);
+	Stream_Write_UINT16(s, CMDTYPE_SET_SURFACE_BITS);
 	Stream_Write_UINT16(s, cmd->destLeft);
 	Stream_Write_UINT16(s, cmd->destTop);
 	Stream_Write_UINT16(s, cmd->destRight);


hardening:
LGTM

hardening:

viniciusjarina:
@hardening it seems green ✅

akallabeth:
@viniciusjarina opened #5982 that should address the issue you´re trying to fix. That one lets the server implementation decide which command should be send.

nfedera:
Note: it depends !!  

[MSRDPRFX 3.1.8.3.1 Encode Message Sequencing](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdprfx/028ed301-385c-4b17-84ec-0a158a22c1b2)

>The sequence of encode messages is encapsulated inside an Extended Bitmap Data structure ([MS-RDPBCGR] section 2.2.9.2.1.1), which is encapsulated in a **Stream Surface Bits Surface Command**

[MS-RDPNSC 1.3 Protocol Overview](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpnsc/a2e57e6f-879f-4084-91a6-56dfac4fa2ea)

>NSCodec-compressed bitmaps that cannot be cached are sent encapsulated in **Set Surface Bits Surface Commands**

nfedera:
@viniciusjarina you may want to confirm that the final fix https://github.com/FreeRDP/FreeRDP/pull/6021 (already merged into master) works for you

viniciusjarina:
Thank you @nfedera I appreciate.

I've just updated to tip of `master` I will do some tests I'll let you know.

PR #5982, opened 2020-03-19
Fix sending/receiving surface bits command.

akallabeth:
* Pass on proper command type to application
* On send let the server implementation decide to send
   2.2.9.2.1 Set Surface Bits Command (TS_SURFCMD_SET_SURF_BITS) or
   2.2.9.2.2 Stream Surface Bits Command (TS_SURFCMD_STREAM_SURF_BITS)
Thank you @viniciusjarina for pointing out the issue.


viniciusjarina:
@akallabeth any ETA to merge this one? 😊

akallabeth:
@viniciusjarina whenever other devs approve ;)
Currently a bit busy as you can imagine.

viniciusjarina:
No worries. Thank you @akallabeth

akallabeth:
@bmiklautz @mfleisz @hardening anyone time for review/merge?

hardening:
I feel like it's gonna break compatibility, isn't it ?

akallabeth:
@hardening if it was used broken like we did then yes. can add a fallback to current behaviour if the wrong command value is set?

akallabeth:
@hardening ok, now it is just like it was except the correct command is supplied.

libfreerdp/core/surface.c
```diff
@@ -221,17 +222,31 @@ static BOOL update_write_surfcmd_bitmap_ex(wStream* s, const TS_BITMAP_DATA_EX*
 	return TRUE;
 }
 
-BOOL update_write_surfcmd_surface_bits(wStream* s, const SURFACE_BITS_COMMAND* cmd)
+BOOL update_write_surfcmd_surface_bits(wStream* s, const SURFACE_BITS_COMMAND* ccmd)
 {
+	SURFACE_BITS_COMMAND cmd = *ccmd;
```
mfleisz, line 227: Why copy the whole struct if we only every override cmdType?

akallabeth: true, should not be necessary.

akallabeth: @mfleisz done, only `cmdType` copied.

hardening:
I guess that updated version will output a lot of warnings in weston and in ogon, but it can be easily fixed.

libfreerdp/core/surface.c
```diff
@@ -223,10 +224,25 @@ static BOOL update_write_surfcmd_bitmap_ex(wStream* s, const TS_BITMAP_DATA_EX*
...
+	switch (cmdType)
+	{
+		case CMDTYPE_SET_SURFACE_BITS:
+		case CMDTYPE_STREAM_SURFACE_BITS:
+			break;
+		default:
+			WLog_WARN(TAG,
+			          "%s SURFACE_BITS_COMMAND->cmdType 0x%08" PRIx32
+			          " not allowed, correcting to 0x%08X",
+			          cmdType, CMDTYPE_STREAM_SURFACE_BITS);
```
nfedera, line 241: The `%s` and `PRIx32` are wrong here.

nfedera: I guess this should be something like:
```
WLog_WARN(TAG, "SURFACE_BITS_COMMAND->cmdType 0x%04" PRIx16 ""
          " not allowed, correcting to 0x%04" PRIx16 "",
          cmdType, CMDTYPE_STREAM_SURFACE_BITS);

```

akallabeth: @nfedera good catch, fixed.

nfedera:
damn, merged to early:

See:

[MSRDPRFX 3.1.8.3.1 Encode Message Sequencing](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdprfx/028ed301-385c-4b17-84ec-0a158a22c1b2)

>The sequence of encode messages is encapsulated inside an Extended Bitmap Data structure ([MS-RDPBCGR] section 2.2.9.2.1.1), which is encapsulated in a **Stream Surface Bits Surface Command**

[MS-RDPNSC 1.3 Protocol Overview](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpnsc/a2e57e6f-879f-4084-91a6-56dfac4fa2ea)

>NSCodec-compressed bitmaps that cannot be cached are sent encapsulated in **Set Surface Bits Surface Commands**


We need another PR that changes the shadow sample to use `CMDTYPE_SET_SURFACE_BITS`for nscodec and `CMDTYPE_STREAM_SURFACE_BITS` for rfx

PR #5987, opened 2020-03-23
Add option to adjust the tcp ack timeout

akallabeth:
On high latency links the default of 9 second timeout might be too
strict. Adjusting this for all users will result in a long time
for connections to fail, so let these with high latency links
adjust the value manually. @CeeMac-git this should address the problem you face.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -2434,6 +2434,13 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings, 
 		{
 			settings->DisableThemes = !enable;
 		}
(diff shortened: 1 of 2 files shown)


CeeMac-git:
pulled down latest master tonight and compiled with cmake, works great, thanks again for adding this feature!

seamus-45:
So many years I have been forced to patch it :laughing:

akallabeth:
@seamus-45 creating a pull request can help avoid that ;)

seamus-45:
@akallabeth not all of us are like C-jedi, but yes, you are right :)

PR #6019, opened 2020-04-02
Fix issues with boundary access.

akallabeth:


libfreerdp/core/rdp.c
```diff
@@ -145,29 +145,33 @@ void rdp_write_security_header(wStream* s, UINT16 flags)
...
-		rdp_read_flow_control_pdu(s, type);
+		if (!rdp_read_flow_control_pdu(s, type))
+			return FALSE;
 		*channel_id = 0;
 		*length = 8; /* Flow control PDU is 8 bytes */
 		return TRUE;
 	}
 
-	if (((size_t)*length - 2) > Stream_GetRemainingLength(s))
+	if ((len < 4) || ((len - 2) > Stream_GetRemainingLength(s)))
```
bmiklautz, line 168: What's the reason for the len < 4?

libfreerdp/core/orders.c
```diff
@@ -2161,7 +2161,7 @@ static CACHE_BITMAP_V3_ORDER* update_read_cache_bitmap_v3_order(rdpUpdate* updat
 	Stream_Read_UINT16(s, bitmapData->height); /* height (2 bytes) */
 	Stream_Read_UINT32(s, new_len);            /* length (4 bytes) */
 
-	if (Stream_GetRemainingLength(s) < new_len)
+	if ((new_len == 0) || (Stream_GetRemainingLength(s) < new_len))
```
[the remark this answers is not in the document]
akallabeth: maybe in theory, but an empty bitmap could break a lot later on and does not make much sense.

libfreerdp/core/rdp.c
akallabeth: true, that should be 2

libfreerdp/core/capabilities.c
```diff
@@ -3353,193 +3356,191 @@ static BOOL rdp_print_capability_sets(wStream* s, UINT16 numberCapabilities, BOO
...
+		size_t rest;
+		wStream sub;
+		if (!rdp_read_capability_set_header(s, &length, &type))
+			return FALSE;
 
-		if (Stream_GetRemainingLength(s) < (size_t)(length - 4))
-		{
-			WLog_ERR(TAG, "error processing stream");
+		WLog_INFO(TAG, "%s ", receiving ? "Receiving" : "Sending");
+		Stream_StaticInit(&sub, Stream_Pointer(s), length - 4);
```
hardening, line 3368: length should be checked (if it's less than 4)

libfreerdp/core/capabilities.c
```diff
@@ -3549,14 +3550,14 @@ static BOOL rdp_print_capability_sets(wStream* s, UINT16 numberCapabilities, BOO
...
 
-		if (Stream_Pointer(s) != em)
+		rest = Stream_GetRemainingLength(&sub);
+		if (rest > 0)
 		{
 			WLog_ERR(TAG,
 			         "incorrect offset, type:0x%04" PRIX16 " actual:%" PRIuz " expected:%" PRIuz "",
-			         type, Stream_Pointer(s) - bm, em - bm);
+			         type, length + rest, length);
 		}
```
hardening, line 3559: suggestion: here instead of printing the offset which is useless we could just show the remaining bytes...

libfreerdp/core/capabilities.c
hardening: Sorry I've just noticed that  `rdp_read_capability_set_header` now does this check

libfreerdp/core/rdp.c
```diff
@@ -1384,7 +1393,7 @@ static int rdp_recv_tpkt_pdu(rdpRdp* rdp, wStream* s)
 				          "pduType %s not properly parsed, %" PRIdz
```
hardening, line 1393: Perhaps you could include my patch (gist sent on IRC), as it fixes the fact that you can announce a small `pduLength` but decoding functions will be able to read more than `pduLength` (because they do checks with `Stream_GetRemainingLength(mainStream)`)

akallabeth:
@hardening will address your comments in a separate PR, I think better merge this now as it is already big enough.

libfreerdp/core/capabilities.c
nfedera: what @hardening wrote, just show how many bytes were not handled

libfreerdp/core/capabilities.c
```diff
@@ -3806,12 +3805,13 @@ static BOOL rdp_read_capability_sets(wStream* s, rdpSettings* settings, UINT16 n
...
+		rest = Stream_GetRemainingLength(&sub);
+		if (rest > 0)
 		{
 			WLog_ERR(TAG,
-			         "incorrect offset, type:0x%04" PRIX16 " actual:%" PRIuz " expected:%" PRIuz "",
-			         type, Stream_Pointer(s) - bm, em - bm);
-			Stream_SetPointer(s, em);
+			         "incorrect offset, type:0x%04" PRIX16 " actual:%" PRIuz " expected:%" PRIu16
+			         "",
+			         type, rest + length, length);
```
nfedera, line 3814: probably that shoud have been `length - 4 - rest, length - 4`  but like above, just print the `rest` bytes not handled.

akallabeth: No, `lenght` was already read and `rest` are the bytes still available.

libfreerdp/core/capabilities.c
akallabeth: already done in latest commit.

libfreerdp/core/capabilities.c
akallabeth: Sorry, correct, my fault.

PR #6026, opened 2020-04-07
Allow /size to work with /multimon

nloewen:
Factor percent scaling from the /size parameter into the resolution that's reported to the RDP server.
Together with /smart-sizing this allows multi monitor connections at non-native resolutions.

run with /multimon /smart-sizing /size:xx%wh

Outstanding issues. (These don't bother me, and I may or may not fix them)
- Cursor is the wrong size when showing drag handles and other alternate shapes
- /multimon /size should probably imply /smart-sizing since it already implies /fullscreen

### DIFF
--- a/client/X11/xf_monitor.c
+++ b/client/X11/xf_monitor.c
@@ -367,12 +367,22 @@ BOOL xf_detect_monitors(xfContext* xfc, UINT32* pMaxWidth, UINT32* pMaxHeight)
 			if (!vscreen->monitors)
 				return FALSE;
 
-			settings->MonitorDefArray[nmonitors].x = vscreen->monitors[i].area.left;
-			settings->MonitorDefArray[nmonitors].y = vscreen->monitors[i].area.top;
+			settings->MonitorDefArray[nmonitors].x =
+			    (vscreen->monitors[i].area.left *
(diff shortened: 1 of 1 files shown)


bmiklautz:
@freerdp-bot test

bmiklautz:
@freerdp-bot test.

bmiklautz:
@freerdp-bot test

nloewen:
@akallabeth When I set MonitorDefArray, smart scaling doesn't work in fullscreen. Any idea where to poke the code to make that happen?

akallabeth:
@nloewen `xf_disp_sendLayout`, `freerdp_display_send_monitor_layout`, `gcc_write_client_monitor_data`, `gcc_write_client_monitor_extended_data` are used for informing the server of the current monitor layout in various states/versions/... of RDP (yes, that is historical protocol creep)
The `MonitorDefArray` is also used in `xf_SetWindowFullscreen` and to populate `xfFullscreenMonitors fullscreenMonitors;` which then is responsible for fullscreen setting.

nloewen:
@akallabeth Moved to MonitorDefArray

akallabeth:
@freerdp-bot test

PR #6049, opened 2020-04-10
tls: support non-RSA keys

lheckemann:
This allows using e.g. EC private keys, which may have better performance characteristics than RSA.


akallabeth:
@freerdp-bot test

PR #6050, opened 2020-04-10
shadow_server: allow specifying IP addresses to listen on

lheckemann:
To test, run one of the shadow servers:
 - With no extra arguments — behaviour should be same as before
 - With `/ipc-socket:some-socket` option — behaviour should be same as before
 - With `/listen-address:127.0.0.1` — should listen only on IPv4 localhost
 - With `/listen-address:[::]` — should listen on all IPv6 addresses
 - With `/listen-address:0.0.0.0 /ipc-socket:some-socket` — aborts for incompatible arguments

### DIFF
--- a/include/freerdp/server/shadow.h
+++ b/include/freerdp/server/shadow.h
@@ -144,6 +144,7 @@ struct rdp_shadow_server
 	UINT32 h264QP;
 
 	char* ipcSocket;
+	char* bindAddress;
 	char* ConfigPath;
 	char* CertificateFile;
 	char* PrivateKeyFile;
--- a/include/freerdp/server/shadow.h
+++ b/include/freerdp/server/shadow.h
@@ -144,7 +144,6 @@ struct rdp_shadow_server
 	UINT32 h264QP;
 
 	char* ipcSocket;
-	char* bindAddress;
 	char* ConfigPath;
 	char* CertificateFile;
 	char* PrivateKeyFile;
--- a/server/shadow/shadow_client.c
+++ b/server/shadow/shadow_client.c
@@ -128,6 +128,7 @@ static INLINE void shadow_client_free_queued_message(void* obj)
 
 static BOOL shadow_client_context_new(freerdp_peer* peer, rdpShadowClient* client)
 {
+	const char bind_address[] = "bind-address,";
 	rdpSettings* settings;
 	rdpShadowServer* server;
 	const wObject cb = { NULL, NULL, NULL, shadow_client_free_queued_message, NULL };
@@ -157,7 +158,8 @@ static BOOL shadow_client_context_new(freerdp_peer* peer, rdpShadowClient* clien
 	if (!(settings->RdpKeyFile = _strdup(settings->PrivateKeyFile)))
 		goto fail_rdpkey_file;
 
-	if (server->ipcSocket)
+	if (server->ipcSocket && (strncmp(bind_address, server->ipcSocket,
+	                                  strnlen(bind_address, sizeof(bind_address))) != 0))
 	{
 		settings->LyncRdpMode = TRUE;
 		settings->CompressionEnabled = FALSE;
--- a/server/shadow/shadow_server.c
+++ b/server/shadow/shadow_server.c
@@ -52,7 +52,7 @@ static const COMMAND_LINE_ARGUMENT_A shadow_args[] = {
 	{ "ipc-socket", COMMAND_LINE_VALUE_REQUIRED, "<ipc-socket>", NULL, NULL, -1, NULL,
 	  "Server IPC socket" },
 	{ "bind-address", COMMAND_LINE_VALUE_REQUIRED, "<bind-address>", NULL, NULL, -1, NULL,
-	  "An address to bind to. Use '[<ipv6>]' for IPv6 addresses, e.g. '[::]' for "
+	  "An address to bind to. Use '[<ipv6>]' for IPv6 addresses, e.g. '[::1]' for "
 	  "localhost" },
 	{ "monitors", COMMAND_LINE_VALUE_OPTIONAL, "<0,1,2...>", NULL, NULL, -1, NULL,
 	  "Select or list monitors" },
@@ -536,7 +536,12 @@ int shadow_server_start(rdpShadowServer* server)
 		return -1;
 	}
 
-	/* Bind to TCP port if nothing is specified, otherwise bind to everything specified */
+	/* Bind magic:
+	 *
+	 * emtpy                 ... bind TCP all
+	 * <local path>          ... bind local (IPC)
+	 * bind-socket,<address> ... bind TCP to specified interface
+	 */
 	ipc = server->ipcSocket && (strncmp(bind_address, server->ipcSocket,
 	                                    strnlen(bind_address, sizeof(bind_address))) != 0);
 	if (!ipc)
@@ -557,7 +562,12 @@ int shadow_server_start(rdpShadowServer* server)
 					WLog_ERR(TAG, "Could not parse bind-address %s", address);
 					return -1;
 				}
-				*end = '\0';
+				*end++ = '\0';
+				if (strlen(end) > 0)
+				{
+					WLog_ERR(TAG, "Excess data after IPv6 address: '%s'", end);
+					return -1;
+				}
 				modaddr++;
 			}
 		}
--- a/server/shadow/shadow_server.c
+++ b/server/shadow/shadow_server.c
@@ -51,7 +51,8 @@ static const COMMAND_LINE_ARGUMENT_A shadow_args[] = {
 	{ "port", COMMAND_LINE_VALUE_REQUIRED, "<number>", NULL, NULL, -1, NULL, "Server port" },
 	{ "ipc-socket", COMMAND_LINE_VALUE_REQUIRED, "<ipc-socket>", NULL, NULL, -1, NULL,
 	  "Server IPC socket" },
-	{ "bind-address", COMMAND_LINE_VALUE_REQUIRED, "<bind-address>", NULL, NULL, -1, NULL,
+	{ "bind-address", COMMAND_LINE_VALUE_REQUIRED, "<bind-address>[,<another address>, ...]", NULL,
+	  NULL, -1, NULL,
 	  "An address to bind to. Use '[<ipv6>]' for IPv6 addresses, e.g. '[::1]' for "
 	  "localhost" },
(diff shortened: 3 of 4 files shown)


akallabeth:
@freerdp-bot test

server/shadow/shadow_server.c
```diff
@@ -510,10 +519,12 @@ int shadow_server_start(rdpShadowServer* server)
 		return -1;
 	}
 
-	if (!server->ipcSocket)
-		status = server->listener->Open(server->listener, NULL, (UINT16)server->port);
-	else
-		status = server->listener->OpenLocal(server->listener, server->ipcSocket);
+	// Bind to TCP port if nothing is specified, otherwise bind to everything specified
+	if (!server->ipcSocket || server->bindAddress)
+		status &=
```
akallabeth, line 524: Better do a check right after assignment here.

server/shadow/shadow_server.c
```diff
@@ -510,10 +519,12 @@ int shadow_server_start(rdpShadowServer* server)
...
+		    server->listener->Open(server->listener, server->bindAddress, (UINT16)server->port);
```
akallabeth, line 525: Sure you want to change the `if(){ else {}` this way? the whole change is not necessary, `server->bindAddress` should simply be initialized as `NULL`

lheckemann: I think so—since we want to bind both sockets if both options are given, and default to binding just TCP if neither of the options are given.

server/shadow/shadow_server.c
lheckemann: Done

akallabeth:
@freerdp-bot test

include/freerdp/server/shadow.h
```diff
@@ -144,6 +144,7 @@ struct rdp_shadow_server
 	UINT32 h264QP;
 
 	char* ipcSocket;
+	char* bindAddress;
```
akallabeth, line 147: API breakage, not eligible for 2.0. Easy fix could be to reuse `ipcSocket` and set `port` to something `> 65525`

hardening:
General remark, when we're on that subject, probably that it would be a good idea that it would be on multiple addresses...

akallabeth:
@hardening true. We could build a compatible extension and reuse `ipcSocket`.
That and we should use the IPv6 syntax just like with the client, the one with the braces `[::]`

akallabeth:
@lheckemann I´ve pushed a commit refactoring your changes so we don´t need a new field (which would make this addition incompatible with the 2.0 series)
@hardening do you think that is a good solution?

akallabeth:
@freerdp-bot test

server/shadow/shadow_server.c
```diff
@@ -510,16 +536,50 @@ int shadow_server_start(rdpShadowServer* server)
 		return -1;
 	}
 
-	if (!server->ipcSocket)
-		status = server->listener->Open(server->listener, NULL, (UINT16)server->port);
-	else
-		status = server->listener->OpenLocal(server->listener, server->ipcSocket);
+	/* Bind to TCP port if nothing is specified, otherwise bind to everything specified */
```
lheckemann, line 539: With your refactoring, this comment is no longer valid, since it will only bind a TCP socket _or_ a local socket

server/shadow/shadow_server.c
```diff
@@ -510,16 +536,50 @@ int shadow_server_start(rdpShadowServer* server)
...
+		char* address =
+		    (server->ipcSocket)
+		        ? _strdup(server->ipcSocket + strnlen(bind_address, sizeof(bind_address)))
+		        : NULL;
+		char* modaddr = address;
 
-	if (!status)
+		if (modaddr)
+		{
+			if (modaddr[0] == '[')
```
lheckemann, line 552: So this requires addresses to be in square brackets, including IPv4?

server/shadow/shadow_server.c
```diff
@@ -510,16 +536,50 @@ int shadow_server_start(rdpShadowServer* server)
...
+			{
+				char* end = strchr(address, ']');
```
lheckemann, line 554: This will completely discard any trailing characters after a ], right? I don't think that's desirable.

server/shadow/shadow_server.c
```diff
@@ -510,16 +536,50 @@ int shadow_server_start(rdpShadowServer* server)
...
+					WLog_ERR(TAG, "Could not parse bind-address %s", address);
+					return -1;
+				}
+				*end = '\0';
+				modaddr++;
+			}
+		}
+		status = server->listener->Open(server->listener, modaddr, (UINT16)server->port);
+		free(address);
+		if (!status)
```
lheckemann, line 566: If we're only doing TCP XOR local, the error handling could be moved out a level like in the original code to avoid repetition?

server/shadow/shadow_server.c
```diff
@@ -45,10 +45,15 @@
...
 
+static const char bind_address[] = "bind-address,";
+
 static const COMMAND_LINE_ARGUMENT_A shadow_args[] = {
 	{ "port", COMMAND_LINE_VALUE_REQUIRED, "<number>", NULL, NULL, -1, NULL, "Server port" },
 	{ "ipc-socket", COMMAND_LINE_VALUE_REQUIRED, "<ipc-socket>", NULL, NULL, -1, NULL,
 	  "Server IPC socket" },
+	{ "bind-address", COMMAND_LINE_VALUE_REQUIRED, "<bind-address>", NULL, NULL, -1, NULL,
+	  "An address to bind to. Use '[<ipv6>]' for IPv6 addresses, e.g. '[::]' for "
+	  "localhost" },
```
lheckemann, line 56: This is incorrect, `::` is all addresses and `::1` is localhost

server/shadow/shadow_client.c
```diff
@@ -128,6 +128,7 @@ static INLINE void shadow_client_free_queued_message(void* obj)
 
 static BOOL shadow_client_context_new(freerdp_peer* peer, rdpShadowClient* client)
 {
+	const char bind_address[] = "bind-address,";
```
lheckemann, line 131: Should this maybe be all-uppercase to make its constantness and role as a sentinel more obvious? Not sure about coding style here.

server/shadow/shadow_server.c
lheckemann: Maybe this parsing code shouldn't be here but in `freerdp_tcp_resolve_host`?

server/shadow/shadow_server.c
```diff
@@ -220,11 +225,31 @@ int shadow_server_parse_command_line(rdpShadowServer* server, int argc, char** a
...
 			server->ipcSocket = _strdup(arg->Value);
 
 			if (!server->ipcSocket)
 				return -1;
 		}
+		CommandLineSwitchCase(arg, "bind-address")
+		{
+			int rc;
+			size_t len = strlen(arg->Value) + sizeof(bind_address);
+			/* /ipc-socket is incompatible */
```
lheckemann, line 241: The usage message doesn't make this clear, I think a specific error message would be nice to have here

server/shadow/shadow_server.c
akallabeth: good catch

akallabeth:
@freerdp-bot test

server/shadow/shadow_server.c
```diff
@@ -510,16 +536,60 @@ int shadow_server_start(rdpShadowServer* server)
 		return -1;
 	}
 
-	if (!server->ipcSocket)
-		status = server->listener->Open(server->listener, NULL, (UINT16)server->port);
-	else
-		status = server->listener->OpenLocal(server->listener, server->ipcSocket);
+	/* Bind magic:
+	 *
+	 * emtpy                 ... bind TCP all
```
lheckemann, line 541: TCP/IPv4 only, at least with the current implementation on Linux

server/shadow/shadow_server.c
akallabeth: not really, as that is command line escape (we already use `:` for something else).
libfreerdp expects a valid DNS name or ip address

server/shadow/shadow_server.c
akallabeth: no, why?

akallabeth: only limitation is you can only bind 1 interface with the current API

lheckemann: I'm not exactly sure why — I think _something_ is causing the shadow server to only bind the first address returned by `getaddrinfo`, but I'm not sure why — my interpretation of `listener.c` https://github.com/FreeRDP/FreeRDP/blob/ce03f5eac59a26a667a578532ab783699b65bc08/libfreerdp/core/listener.c#L74 suggests that it would bind all of them.


```
[linus@bringtnix:~/projects/freerdp]$ ./server/shadow/freerdp-shadow-cli
[09:02:02:439] [30228:30228] [INFO][com.freerdp.server.shadow.x11] - X11 Extensions: XFixes: 1 Xinerama: 1 XDamage: 0 XShm: 0
[09:02:02:442] [30228:30228] [INFO][com.freerdp.core.listener] - Listening on [0.0.0.0]:3389

<in another terminal>
[linus@bringtnix:~]$ nc -v6 localhost 3389
nc: connect to localhost port 3389 (tcp) failed: Connection refused

[linus@bringtnix:~]$ nc -v localhost 3389
nc: connect to localhost port 3389 (tcp) failed: Connection refused
Connection to localhost 3389 port [tcp/ms-wbt-server] succeeded!
```

akallabeth: @lheckemann check out `MAX_LISTENER_HANDLES` which only allows 5 listeners.

lheckemann: Sure, but it's only listening on the one address, so I don't think that limit is what's stopping it from listening on IPv6.

akallabeth: @lheckemann well, if I start with `/bind-address:[::]` or `/bind-address:[::1]` it listens ;)

akallabeth:
@bmiklautz @hardening @mfleisz anyone want to merge this one?

server/shadow/shadow_server.c
lheckemann: Right, but "TCP on all IPv4 addresses" isn't the same thing as "TCP on all IP addresses", and the latter is what I'd expect given no arguments :/ that said, I wouldn't consider it critical.

akallabeth:
@freerdp-bot test

server/shadow/shadow_server.c
akallabeth: @lheckemann added some code for the tcp path to allow multiple bind addresses.

PR #6062, opened 2020-04-11
Optimize function xcrush_copy_bytes()

rgfernandes:
Use memcpy to copy the bytes when we can assure that the memory areas does not overlap.
With this change, the copy is ~30x faster.

### DIFF
--- a/libfreerdp/codec/xcrush.c
+++ b/libfreerdp/codec/xcrush.c
@@ -732,13 +732,27 @@ static int xcrush_generate_output(XCRUSH_CONTEXT* xcrush, BYTE* OutputBuffer, UI
 	return 1;
 }
 
-static size_t xcrush_copy_bytes(BYTE* dst, const BYTE* src, size_t num)
+static INLINE size_t xcrush_copy_bytes(BYTE* dst, const BYTE* src, size_t num)
 {
-	size_t index;
-
-	for (index = 0; index < num; index++)
+	size_t diff, rest, end, a;
+	if (src + num < dst || src > dst + num)
 	{
-		dst[index] = src[index];
+		memcpy(dst, src, num);
+	}
+	else
+	{
+		// src and dst overlaps
+		// we should copy the area that doesn't overlap repeatly
+		diff = (dst > src) ? dst - src : src - dst;
+		rest = num % diff;
+		end = num - rest;
+		for (a = 0; a < end; a += diff)
+		{
+			memcpy(&dst[a], &src[a], diff);
+		}
+
+		if (rest != 0)
+			memcpy(&dst[end], &src[end], rest);
 	}
 
 	return num;


akallabeth:
@freerdp-bot test

akallabeth:
@rgfernandes thant you for the pull request.
There are a few things to note though:
1. Please run `clang-format` on your changes (or apply the patch from the pull request tester output)
1. Could you give some improvement statistics, these changes look like a modern compiler can do them automatically without having `#ifdef` (which we try to avoid)

rgfernandes:
>     1. Please run `clang-format` on your changes (or apply the patch from the pull request tester output)

Done.

>     2. Could you give some improvement statistics, these changes look like a modern compiler can do them automatically without having `#ifdef` (which we try to avoid)

I have applied another patch that mantains the original code and the optimized version and call the 2 one after another. The patch is:
https://github.com/rgfernandes/FreeRDP/commit/0c1075e5484ff97f5a174e2b5168f894018bc657
Profiled with valgrind, acessed a RDP and checked the results.
xcrush_copy_bytes
memcpy appears at distance 1-2 because it is called directly in xcrush_decompress_l1() and xcrush_copy_bytes_opt().
The valgrind's output:
[xcrush_copy_bytes.zip](https://github.com/FreeRDP/FreeRDP/files/4465001/xcrush_copy_bytes.zip)

akallabeth:
@rgfernandes thank you. did you try with `memmove` instead of the manual loop unroll too?

rgfernandes:
> @rgfernandes thank you. did you try with `memmove` instead of the manual loop unroll too?

Yes. Using memmove leads to a crash. Possibly a buggy implementation of memmove. I have tried the -fno-builtin-memmove option in gcc to avoid using the (buggy) builtin memmove implementation, but it leads to artifacts in image.

rgfernandes:
>     these changes look like a modern compiler can do them automatically without having `#ifdef` (which we try to avoid)

Another shot from valgrind (kcachegrind):
xcrush_copy_bytes1
My gcc is 8.3.0 from debian 10.
Look at the dissassembled function. It gets the value with a movzbl (move one byte to a 4 bytes register) and add only one to counter.
The best is use memmove, but as I comment, it crashes.
I can remove the code inside #if. This path (areas that overlay) is the least used anyway.

akallabeth:
Yes, but you could do classic loop unroll and yield a bit more (sse et al can use 16 or 32 bytes at once)
You just need a loop counter with x+16 to hint the compiler.

As for memmove, that would be a serious bug, do you have a reference to a debian issue or similar?

akallabeth:
Btw, how dou you test (xfreerdp commandline)?

akallabeth:
@rgfernandes ok, managed to reprodue a testcase (connection to win7, `/v:host /u:user /p:xxxx /bpp:16 +compression` 

long story short, your patch is buggy.
```
#define min(a,b) \
   ({ __typeof__ (a) _a = (a); \
       __typeof__ (b) _b = (b); \
     _a < _b ? _a : _b; })

static size_t xcrush_copy_bytes(BYTE* dst, const BYTE* src, size_t num)
{
    size_t offset = 0;
    size_t diff = (dst > src) ? dst - src : src - dst;
    size_t block = min(num, diff);

    while(offset < num) {
        size_t add = min(block, num - offset);
        memmove(&dst[offset], &src[offset], add);
        offset += add;
    }

    return num;
}
```
will work. The reason you have issues with `memmove` is that it behaves as if you copy n bytes to a temporary buffer and then to dst. there the blocks not only overlap but you actually move block wise.

rgfernandes:
After exautive trial and error, now I have understood what was happening.
The non trivial behaviour is from protocol.
The "crashes" I found is not really crashes. It is protocol errors leading to program exit.
AFAIK, it should be like this:

- If the areas don't overlap, it should copy with memcpy as usual.
- If the areas overlap, you have to copy the portion that does not overlap from src until it completes the dst.

I have not found one time that the areas overlap and dst is before src.
I have tested this code:

```C
static size_t xcrush_copy_bytes(BYTE* dst, const BYTE* src, size_t num)
{
    if (src + num > dst)
    {
        // src and dst overlaps
        // we should copy the area from src that doesn't overlap repeatly
        // until completes the dst
        size_t diff = dst - src;
        size_t rest = num % diff;
        size_t end = num - rest;
        for (size_t a = 0; a < end; a += diff)
        {
            memcpy(&dst[a], &src[a], diff);
        }
        if (rest != 0)
            memcpy(&dst[end], &src[end], rest);
    }
    else
    {
        memcpy(dst, src, num);
        if (dst < src && dst + num > src)
        {
            printf("Error\n");
            return num;
        }
    }
    return num;
}
```
I worked flawlessly.
What do you think about this code?

akallabeth:
As the destination buffer is from outside (`xcrush_decompress` is public API) I´d not optimize anything that can not hold under every possible input.

rgfernandes:
> As the destination buffer is from outside (`xcrush_decompress` is public API) I´d not optimize anything that can not hold under every possible input.

It is easy to resolve.

```C
static size_t xcrush_copy_bytes(BYTE* dst, const BYTE* src, size_t num)
{
    if (src + num < dst || src > dst + num)
    {
        memcpy(dst, src, num);
    }
    else
    {
        // src and dst overlaps
        // we should copy the area that doesn't overlap repeatly
        size_t diff = (dst > src) ? dst - src : src - dst;
        size_t rest = num % diff;
        size_t end = num - rest;
        for (size_t a = 0; a < end; a += diff)
        {
            memcpy(&dst[a], &src[a], diff);
        }
        if (rest != 0)
            memcpy(&dst[end], &src[end], rest);
    }
    return num;
}
```

rgfernandes:
Is it ok now?

libfreerdp/codec/xcrush.c
```diff
@@ -734,11 +734,31 @@ static int xcrush_generate_output(XCRUSH_CONTEXT* xcrush, BYTE* OutputBuffer, UI
...
+	if (src + num < dst || dst + num < src)
+	{
+		memcpy(dst, src, num);
+	}
+	else
 	{
-		dst[index] = src[index];
+		size_t index = 0;
+		// copy 4 or 8 bytes at once
+#if __LP64__
```
akallabeth, line 745: Removed this, the compilers know how to handle vectorization, you just disabled this path for a lot of compilers/architectures

libfreerdp/codec/xcrush.c
```diff
@@ -734,11 +734,31 @@ static int xcrush_generate_output(XCRUSH_CONTEXT* xcrush, BYTE* OutputBuffer, UI
...
+		size_t end = num / 8;
```
akallabeth, line 746: this condition is wrong, should be more like `num - num % 8`

libfreerdp/codec/xcrush.c
```diff
@@ -734,11 +734,31 @@ static int xcrush_generate_output(XCRUSH_CONTEXT* xcrush, BYTE* OutputBuffer, UI
...
+		for (; index < end; index += 8)
+		{
+			*((UINT64*)(dst + index)) = *((UINT64*)(src + index));
+		}
+#else
+		size_t end = num / 4;
```
akallabeth, line 752: This condition is wrong, should be more like `num - num % 4`

libfreerdp/codec/xcrush.c
rgfernandes: I will use the last code I have shown

rgfernandes:
Please, rewiew the changes I've just commited.
I think that this new way the code is more clear and uses memcpy everytime.

libfreerdp/codec/xcrush.c
```diff
@@ -821,7 +834,7 @@ static int xcrush_decompress_l1(XCRUSH_CONTEXT* xcrush, const BYTE* pSrcData, UI
 				    (&Literals[OutputLength] > pSrcEnd))
 					return -1009;
 
-				xcrush_copy_bytes(HistoryPtr, Literals, OutputLength);
```
akallabeth, line 824: Revert this, your assumption that source and destination are different might not hold.

libfreerdp/codec/xcrush.c
```diff
@@ -849,7 +862,7 @@ static int xcrush_decompress_l1(XCRUSH_CONTEXT* xcrush, const BYTE* pSrcData, UI
 		if ((&HistoryPtr[OutputLength] >= HistoryBufferEnd) || (&Literals[OutputLength] > pSrcEnd))
 			return -1012;
 
-		xcrush_copy_bytes(HistoryPtr, Literals, OutputLength);
```
akallabeth, line 852: Revert this, your assumption that source and destination are different might not hold.

libfreerdp/codec/xcrush.c
```diff
@@ -734,11 +734,24 @@ static int xcrush_generate_output(XCRUSH_CONTEXT* xcrush, BYTE* OutputBuffer, UI
 
 static size_t xcrush_copy_bytes(BYTE* dst, const BYTE* src, size_t num)
```
akallabeth, line 735: Declaring this function `INLINE` might also help

rgfernandes: Yes

libfreerdp/codec/xcrush.c
rgfernandes: No problem to me

libfreerdp/codec/xcrush.c
rgfernandes: No problem too

akallabeth:
@freerdp-bot test

libfreerdp/codec/xcrush.c
```diff
@@ -732,13 +732,26 @@ static int xcrush_generate_output(XCRUSH_CONTEXT* xcrush, BYTE* OutputBuffer, UI
...
+	}
+	else
 	{
-		dst[index] = src[index];
+		// src and dst overlaps
+		// we should copy the area that doesn't overlap repeatly
+		size_t diff = (dst > src) ? dst - src : src - dst;
+		size_t rest = num % diff;
+		size_t end = num - rest;
+		for (size_t a = 0; a < end; a += diff)
```
akallabeth, line 748: sorry, that is `C++` style and breaks with older compilers.

rgfernandes: I have put the declarations on the beginning of the function.

akallabeth:
@freerdp-bot test

akallabeth:
@rgfernandes thank you for your contribution!

PR #6070, opened 2020-04-13
Mark unused functions to avoid compiler warnings

rgfernandes:
Only adds ``__attribute__((unused))`` to unused functions to avoid compiler warnings.


akallabeth:
Sorry, not portable.

AnonymousRetard:
I think this can easily be made portable by also adding these defines:

#ifdef _GNUC
  #define ATTR_UNUSED __attribute__((unused))
#else
  #define ATTR_UNUSED
#endif

PR #6071, opened 2020-04-14
Adds the title option to wlfreerdp

chakuzo:
Adds the /t: option to wlfreerdp

If no title is set "FreeRDP" is used.

This time the formatting should also be correct


hardening:

PR #6073, opened 2020-04-14
Language fixes for aFreeRDP

comradekingu:
I am in #freeRDP
Edit: Adding this project on Hosted Weblate and merging these after the fact
would be preferable https://weblate.org/hosting/


akallabeth:
@freerdp-bot test

PR #6077, opened 2020-04-14
Progressive limit stack

akallabeth:
Fix for #6045


akallabeth:
@lemonmojo I´ve successfully tested this on our build machine, can you confirm? Looks like the culprit are the stack limitations on mac os and optimizer exceeding it in favour of faster execution

lemonmojo:
@akallabeth Success! Tested all the crashing cases in https://github.com/FreeRDP/FreeRDP/issues/6045 and have not been able to reproduce after applying your changes.

PR #6080, opened 2020-04-14
Fixi #6086: smartcard filter early abort.

akallabeth:
Do not abort before incrementing read position.


akallabeth:
@hardening problem is the abort condition, it then does not advance the read position and the final `\0` is not added after the loop.

PR #6091, opened 2020-04-16
Spelling fixes for iOS strings

comradekingu:
Why is the formatting UTF-16LE?


akallabeth:
@freerdp-bot test
@comradekingu no idea, ios is currently unmaintained, would be great if someone could take a look at the client from time to time

PR #6095, opened 2020-04-17
uwac, wlfreerdp: corrected API for UwacOutput and added monitor listing

hardening:
`/monitor-list` was not implemented for `wlfreerdp`. This PR also fixes bugs in UWAC when retrieving a specific output.

### DIFF
--- a/client/Wayland/wlf_disp.c
+++ b/client/Wayland/wlf_disp.c
@@ -377,3 +377,25 @@ BOOL wlf_disp_uninit(wlfDispContext* wlfDisp, DispClientContext* disp)
 	wlfDisp->disp = NULL;
 	return TRUE;
 }
+
+int wlf_list_monitors(wlfContext* wlc)
+{
+	uint32_t i, nmonitors = UwacDisplayGetNbOutputs(wlc->display);
+
+	for (i = 0; i < nmonitors; i++)
+	{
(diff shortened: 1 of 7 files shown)


client/Wayland/wlf_disp.c
```diff
@@ -377,3 +377,23 @@ BOOL wlf_disp_uninit(wlfDispContext* wlfDisp, DispClientContext* disp)
...
 	return TRUE;
 }
+
+int wlf_list_monitors(wlfContext* wlc)
+{
+	uint32_t i, nmonitors = UwacDisplayGetNbOutputs(wlc->display);
+
+	for (i = 0; i < nmonitors; i++)
+	{
+		const UwacOutput* monitor = UwacDisplayGetOutput(wlc->display, i);
```
akallabeth, line 387: is this guaranteed to be `!= NULL`?

hardening: yes the number of monitors can't change in our back, but you're right I should add a check.

akallabeth: @hardening you are adding a check or not? do want to merge this ;)

hardening:
Ok, added the check.
There would still need to announce and detect multi-monitors, but for this the change is not trivial (I gave a quick look at xf_monitors).

akallabeth:
@hardening nice, one step at a time ;)

PR #6097, opened 2020-04-20
Refactored the parsing of the dev: command line option in the urbdrc subsystem

lukasfink1:
The parsing wasn’t very well written, e.g. there was an unsafe strcpy leading to buffer overflows when the device id string is too long.

### DIFF
--- a/channels/urbdrc/client/libusb/libusb_udevman.c
+++ b/channels/urbdrc/client/libusb/libusb_udevman.c
@@ -601,26 +601,28 @@ static void udevman_load_interface(UDEVMAN* udevman)
 static BOOL udevman_parse_device_id_addr(const char** str, UINT16* id1, UINT16* id2, UINT16 max,
                                          char split_sign, char delimiter)
 {
-	const char* mid;
+	char* mid;
+	char* end;
 	unsigned long rc;
 
-	rc = strtoul(*str, (char**)&mid, 16);
+	rc = strtoul(*str, &mid, 16);
 	/* These casts are safe, because strtoul only uses it to return a pointer */
 
 	if ((mid == *str) || (*mid != split_sign) || (rc > max))
 		return FALSE;
 
 	*id1 = (UINT16)rc;
-	rc = strtoul(++mid, (char**)str, 16);
+	rc = strtoul(++mid, &end, 16);
 
-	if ((*str == mid) || (rc > max))
+	if ((end == mid) || (rc > max))
 		return FALSE;
 
 	*id2 = (UINT16)rc;
 
-	if (**str == '\0')
+	*str += end - *str;
+	if (*end == '\0')
 		return TRUE;
-	if (**str == delimiter)
+	if (*end == delimiter)
 	{
 		(*str)++;
 		return TRUE;
--- a/channels/urbdrc/client/libusb/libusb_udevman.c
+++ b/channels/urbdrc/client/libusb/libusb_udevman.c
@@ -347,75 +347,6 @@ static BOOL udevman_unregister_all_udevices(IUDEVMAN* idevman)
 	return TRUE;
 }
 
-static BOOL udevman_parse_device_addr(const char* str, size_t maxLen, UINT8* id1, UINT8* id2,
-                                      char sign)
-{
-	unsigned long rc;
-	char s1[8] = { 0 };
-	char* s2 = strchr(str, sign);
-	size_t len = strnlen(str, maxLen);
-	size_t cpLen;
-
-	if (!s2)
-		return FALSE;
-	s2++;
-
-	cpLen = len - (strnlen(s2, len) + 1);
-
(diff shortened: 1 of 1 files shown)


channels/urbdrc/client/libusb/libusb_udevman.c
```diff
@@ -347,73 +347,30 @@ static BOOL udevman_unregister_all_udevices(IUDEVMAN* idevman)
...
-
-	if (cpLen >= sizeof(s1))
-		cpLen = sizeof(s1) - 1;
-
-	strncpy(s1, str, cpLen);
-	rc = strtoul(s1, NULL, 16);
+	*id1 = strtoul(*str, &mid, 16);
 
-	if ((rc > UINT8_MAX) || (errno != 0))
+	if (mid == *str || *mid != split_sign)
```
akallabeth, line 357: We prefer braces around each block

channels/urbdrc/client/libusb/libusb_udevman.c
```diff
@@ -347,73 +347,30 @@ static BOOL udevman_unregister_all_udevices(IUDEVMAN* idevman)
 	return TRUE;
 }
 
-static BOOL udevman_parse_device_addr(const char* str, size_t maxLen, UINT8* id1, UINT8* id2,
-                                      char sign)
+static BOOL udevman_parse_device_id_addr(char** str, unsigned long* id1, unsigned long* id2,
```
akallabeth, line 350: Would suggest making the `id1` and `id2` of type `UINT16` and use an internal variable for `strtoul` (add a range arguement too, so too large values fail the parsing)

akallabeth:
@freerdp-bot test

akallabeth:
@lukasfink1 Thank you for this nice enhancement in command line parsing, makes the whole thing a lot nicer to read.
If you could address the (small) remarks and squash the changes in a single commit that would be really great.

lukasfink1:
Also I wanted to add that this implementation is more strict in that it doesn’t allow multiple hashes without ids in between like in `/usb:id,dev:1234:5678##9abcd:efef###abab:cdcd`. If you want the old behaviour, I can change that.

channels/urbdrc/client/libusb/libusb_udevman.c
```diff
@@ -667,56 +598,62 @@ static void udevman_load_interface(UDEVMAN* udevman)
 	udevman->iface.initialize = udevman_initialize;
 }
 
-static BOOL urbdrc_udevman_register_devices(UDEVMAN* udevman, const char* devices)
+static BOOL udevman_parse_device_id_addr(char** str, UINT16* id1, UINT16* id2, UINT16 max,
```
akallabeth, line 601: It might be a nice addition to make this `const char** str` to indicate that we don´t modify the content.

channels/urbdrc/client/libusb/libusb_udevman.c
```diff
@@ -667,56 +598,62 @@ static void udevman_load_interface(UDEVMAN* udevman)
...
-		strcpy(hardware_id, token);
-		token = strtok(NULL, "#");
+		(*str)++;
+		return TRUE;
+	}
+
+	return FALSE;
+}
 
+static BOOL urbdrc_udevman_register_devices(UDEVMAN* udevman, char* devices)
```
akallabeth, line 631: It might be a nice addition to make this `const char* device` to indicate that we don´t modify the content.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test
@lukasfink1 didn´t want to annoy you again and fixed the casts away. thanks for this fine improvement!

channels/urbdrc/client/libusb/libusb_udevman.c
```diff
@@ -601,26 +601,28 @@ static void udevman_load_interface(UDEVMAN* udevman)
...
+	rc = strtoul(++mid, &end, 16);
 
-	if ((*str == mid) || (rc > max))
+	if ((end == mid) || (rc > max))
 		return FALSE;
 
 	*id2 = (UINT16)rc;
 
-	if (**str == '\0')
+	*str += end - *str;
```
lukasfink1, line 622: I think `*str = end;` should work too

lukasfink1:
But I actually went with the casts, because without them, strtoul spits out pointers with write-access to our string that should be read-only, while with the casts only strtoul could change the string and strtoul is defined in the C standard to not do that.

channels/urbdrc/client/libusb/libusb_udevman.c
akallabeth: @lukasfink1 I don´t really like casts if not absolutely required, they tend to trick the compiler. Using just `end` will spit out a warning (`const` and non const)

PR #6098, opened 2020-04-20
SCardReadCache/SCardWriteCache should actually cache data

arekinath:
In the PCSC SCard code, currently since the `hash`/`keyCompare`/`keyClone` members on the `context->cache` were never being set, we were using the `HashTable_Pointer*` variants, meaning that lookup always failed (since we never ask for the same *pointer* twice, just the same *strings*).

This also revealed that the logic for autoallocate on these ops was a bit backwards, and some error codes and support for the "freshness" counter were missing.

In Win10 (at least with some card minidrivers) the freshness counter is load-bearing and smartcard login won't work without implementing at least a very basic version of it. In this patch I just return the `STALE` error if the requested freshness counter `!=` the stored value. In [MS-RDPESC] it seems to imply that this should be fine for implementors, but the MSDN docs for `SCardReadCache` seem to imply instead that we should only return `STALE` if our stored counter is *older*. Go figure. I went with `!=` for now, it's easy to change though.

I also silenced the error/warn level logs on these two new error codes (`SCARD_W_CACHE_ITEM_NOT_FOUND`, `SCARD_W_CACHE_ITEM_STALE`) since they're routine return values that don't really indicate that something has necessarily gone wrong (and every time I log in I see about half a dozen of these or more).

I'm testing this mostly against latest Win10 at the moment with the YubiKey Minidriver installed on the remote side, which after this patch seems to be working for Smartcard logon finally without changing the `/client-build-number`.

### DIFF
--- a/channels/smartcard/client/smartcard_operations.c
+++ b/channels/smartcard/client/smartcard_operations.c
@@ -959,7 +959,7 @@ static LONG smartcard_ReadCacheA_Call(SMARTCARD_DEVICE* smartcard, SMARTCARD_OPE
 	if (!call->Common.fPbDataIsNULL)
 	{
 		ret.cbDataLen = call->Common.cbDataLen;
-		if (autoalloc)
+		if (!autoalloc)
 		{
 			ret.pbData = malloc(ret.cbDataLen);
 			if (!ret.pbData)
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@arekinath could you please run `clang-format` on your changes? that should enforce coding guidelines and make @freerdp-bot happy

arekinath:
Woops, sorry about the format issues. Hopefully that's a bit better.

akallabeth:
@freerdp-bot test

akallabeth:
Thank you for this really nice improvement!
Looking forward to this working (to be honest, I´m the one that missed this, damn :/)

PR #6100, opened 2020-04-21
Fixed #6099: Add a flag for legacy hash entries

### LINKED ISSUE #6099: Replace Misleading 'certificate changed' warning and help user to compare fingerprints
Commit 9c999b7135401b15d6fe81bceb7fec269aa72bdb changed fingerprint generation from SHA1 to SHA256. As the changelog already states, this invalidated all entries in the known_hosts2 file.
As these entries are not deleted, but still persist with SHA1 fingerprints, a warning is presented to the user at the next connection attempt, stating that _"the certificate of the server **changed** (...) This may indicate that the certificate has been tampered with. Please contact the administrator (...)"_ implying that this may be caused by a man-in-the-middle attack. 
**This is misleading and terrifying the user for no reason**, because not the certificate itself changed, but only the local fingerprint hashing algorithm changed.
In addition, only the SHA256 fingerprint of the server certificate is presented to the user.
Unfortunately **Windows 10's Certificate Manager is only capable of presenting SHA1 fingerprints to the user**. so it is not obvious in the first place, whether the certificate still matches the one of the RDP server or not.
 
If a (SHA256) fingerprint mismatch is detected, FreeRDP could check whether the SHA1 fingerprint of the server certificate still matches the entry in known_hosts2. If so, and _only_ in this case, the terrifying warning could be replaced by some smoother warning like
_"the fingerprint entry in our database of known hosts needs to be updated from SHA1 to SHA256, please check whether the server certificate is still correct"_.
In addition, both the SHA256 **and** SHA1 fingerprints of the certificate could be printed. This simplifies checking whether the server certificate may still be correct. Of course, the user would still have to check whether the SHA256 fingerprint also matches to be really sure.

An alternative would be to remove all SHA1 entries from known_hosts2 or ignore them as if the host was completey unknown.
But this would also make the user feel unsure at the next connection attempt when the user still believes that he is connecting to a well-known host.

**Additional context**
Plase consider that the freerdp2 update may be automatic, so the user never read the changelog and is not aware of the change in fingerprint representation from SHA1 to SHA256 and its consequences.

akallabeth:
If a legacy entry is found in certificate hash store print
additional information to the user informing about the change
with FreeRDP 2.0

### DIFF
--- a/client/common/client.c
+++ b/client/common/client.c
@@ -549,6 +549,7 @@ DWORD client_cli_verify_certificate_ex(freerdp* instance, const char* host, UINT
 	printf("\tSubject:     %s\n", subject);
 	printf("\tIssuer:      %s\n", issuer);
 	printf("\tThumbprint:  %s\n", fingerprint);
+
 	printf("The above X.509 certificate could not be verified, possibly because you do not have\n"
 	       "the CA certificate in your certificate store, or the certificate has expired.\n"
 	       "Please look at the OpenSSL documentation on how to add a private CA to the store.\n");
@@ -644,6 +645,14 @@ DWORD client_cli_verify_changed_certificate_ex(freerdp* instance, const char* ho
 	printf("\tIssuer:      %s\n", old_issuer);
 	printf("\tThumbprint:  %s\n", old_fingerprint);
 	printf("\n");
+	if (flags & VERIFY_CERT_FLAG_MATCH_LEGACY_SHA1)
+	{
+		printf("\tA matching entry with legacy SHA1 was found in local known_hosts2 store.\n");
+		printf("\tIf you just upgraded from a FreeRDP version before 2.0 this is expected.\n");
+		printf("\tThe hashing algorithm has been upgraded from SHA1 to SHA256.\n");
+		printf("\tAll manually accepted certificates must be reconfirmed!\n");
+		printf("\n");
+	}
 	printf("The above X.509 certificate does not match the certificate used for previous "
 	       "connections.\n"
 	       "This may indicate that the certificate has been tampered with.\n"
--- a/include/freerdp/freerdp.h
+++ b/include/freerdp/freerdp.h
@@ -67,6 +67,7 @@ extern "C"
 #define VERIFY_CERT_FLAG_GATEWAY 0x20
 #define VERIFY_CERT_FLAG_CHANGED 0x40
 #define VERIFY_CERT_FLAG_MISMATCH 0x80
+#define VERIFY_CERT_FLAG_MATCH_LEGACY_SHA1 0x100
 
 	typedef BOOL (*pContextNew)(freerdp* instance, rdpContext* context);
 	typedef void (*pContextFree)(freerdp* instance, rdpContext* context);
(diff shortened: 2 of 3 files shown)


stsichler:
Hm. For some reason that fix doesn't work for me. But I'm not sure whether I did something wrong.
I cloned stable-2.0 branch, applied your fix as local patch and built.

`ldd` confirmed that the freshly built xfreerdp client links agains the patched libraries in the source tree:
`(...)
libfreerdp-client2.so.2 => /home/stefan/Entwicklung/FreeRDP/client/common/libfreerdp-client2.so.2 (0x00007ff7d7738000)
	libfreerdp2.so.2 => /home/stefan/Entwicklung/FreeRDP/libfreerdp/libfreerdp2.so.2 (0x00007ff7d7384000)
(...)
libwinpr2.so.2 => /home/stefan/Entwicklung/FreeRDP/winpr/libwinpr/libwinpr2.so.2 (0x00007ff7d6aa5000)
`

but still, I get the message

`[12:36:34:599] [7165:7166] [ERROR][com.freerdp.crypto] - The host key for xx.xx.xx.xx:3389 has changed
[12:36:34:599] [7165:7166] [ERROR][com.freerdp.crypto] - @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
[12:36:34:599] [7165:7166] [ERROR][com.freerdp.crypto] - @    WARNING: REMOTE HOST IDENTIFICATION HAS CHANGED!     @
[12:36:34:599] [7165:7166] [ERROR][com.freerdp.crypto] - @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
[12:36:34:599] [7165:7166] [ERROR][com.freerdp.crypto] - IT IS POSSIBLE THAT SOMEONE IS DOING SOMETHING NASTY!
[12:36:34:599] [7165:7166] [ERROR][com.freerdp.crypto] - Someone could be eavesdropping on you right now (man-in-the-middle attack)!
[12:36:34:600] [7165:7166] [ERROR][com.freerdp.crypto] - It is also possible that a host key has just been changed.
[12:36:34:600] [7165:7166] [ERROR][com.freerdp.crypto] - The fingerprint for the host key sent by the remote host is xx:xx:xx:xx:xx:xx.....:xx
[12:36:34:600] [7165:7166] [ERROR][com.freerdp.crypto] - Please contact your system administrator.
[12:36:34:600] [7165:7166] [ERROR][com.freerdp.crypto] - Add correct host key in /home/stefan/.config/freerdp/known_hosts2 to get rid of this message.
[12:36:34:600] [7165:7166] [ERROR][com.freerdp.crypto] - Host key for xx.xx.xx.xx has changed and you have requested strict checking.
[12:36:34:600] [7165:7166] [ERROR][com.freerdp.crypto] - Host key verification failed.
`
What is wrong?

The SHA1 fingerprint matches, so your code should have been triggered.

akallabeth:
@stsichler true, did have the wrong variable in the second match check, should be working now.

stsichler:
Ok, the check triggers now! But the new message is now printed **in addition** to the terrifying messages about man-in-the-middle-attack and all further warnings that were there before.
Was this your intention?

akallabeth:
@stsichler thank you for the confirmation, and yes, that is intended.
Since (just as the message tells) this is only valid for the case that an upgrade was performed, which we can not really detect.

akallabeth:
@hardening @bmiklautz what do you think, could reduce upgrade pain for some.

PR #6106, opened 2020-04-22
Pointer fixes

akallabeth:
* Input validation according to capabilities exchanged
* Fix capability exchange, set flags to what was agreed upon


client/common/cmdline.c
```diff
@@ -3265,8 +3265,6 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 	if (settings->RemoteFxCodec || settings->NSCodec || settings->SupportGraphicsPipeline)
 	{
 		settings->FastPathOutput = TRUE;
-		settings->LargePointerFlag =
```
akallabeth, line 3268: Removed as this overrides pointer support settings of client.

PR #6119, opened 2020-04-23
Optimization: avoid calling repeatedly GetBytesPerPixel() and GetBitsPerPixel()

rgfernandes:
This PR has 2 commits.
The first commit caches the return value of GetBytesPerPixel when used inside loops.
This is already used in code as seen in:
- libfreerdp/primitives/prim_YUV.c
- libfreerdp/primitives/prim_colors.c
- libfreerdp/primitives/prim_YCoCg.c
- libfreerdp/primitives/prim_YCoCg_opt.c
- libfreerdp/codec/rfx.c
- libfreerdp/gdi/gfx.c
- libfreerdp/gdi/shape.c
- server/shadow/Mac/mac_shadow.c

I have only added the same logic to other places.
The second commit removes the call to GetBitsPerPixel from ReadColor() and WriteColor().
These functions (ReadColor() and WriteColor()) are called several times inside loops leading to repetitive calls to GetBitsPerPixel() with always the same result.
This image show the cost of GetBitsPerPixel().
cost_of_getbitsperpixel
To verify the optimization, I create 4 new functions: ReadColor_orig, ReadColor_opt, WriteColor_orig and WriteColor_opt.
The patch with this code is https://github.com/rgfernandes/FreeRDP/commit/1f895172c0a776e7164dcec306beb0b9e428f6bb
The result is shown in this picture.
read_writecolor

The 2 optimized versions don't call GetBitsPerPixel() repeatedly anymore and represents a speedup of ~20%.

### DIFF
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -627,7 +627,7 @@ static BOOL convert_rail_icon(ICON_INFO* iconInfo, xfRailIcon* railIcon)
 
 	for (i = 2; i < nelements; i++)
 	{
-		pixels[i] = ReadColor(nextPixel, PIXEL_FORMAT_BGRA32);
+		pixels[i] = ReadColor(nextPixel, PIXEL_FORMAT_BGRA32, 32 /*GetBitsPerPixel(PIXEL_FORMAT_BGRA32)*/);
 		nextPixel += 4;
 	}
 
--- a/include/freerdp/codec/color.h
+++ b/include/freerdp/codec/color.h
@@ -711,15 +711,15 @@ extern "C"
 	 * Read a pixel from memory to internal representation
 	 *
 	 * @param src    The source buffer
-	 * @param format The PIXEL_FORMAT_* define the source buffer uses for encoding
+	 * @param bitsPerPixel The bits per pixel of PIXEL_FORMAT_* define the source buffer uses for encoding
 	 *
 	 * @return The pixel color in internal representation
 	 */
-	static INLINE UINT32 ReadColor(const BYTE* src, UINT32 format)
+	static INLINE UINT32 ReadColor(const BYTE* src, UINT32 format, UINT32 bitsPerPixel)
 	{
 		UINT32 color;
 
-		switch (GetBitsPerPixel(format))
+		switch (bitsPerPixel)
 		{
 			case 32:
 				color = ((UINT32)src[0] << 24) | ((UINT32)src[1] << 16) | ((UINT32)src[2] << 8) |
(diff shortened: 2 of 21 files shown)


akallabeth:
@rgfernandes 
1. Can you add your test case somewhere to run it on other hardware too?
1. Which compiler flags did you use? All the functions you optimized are `INLINE`, so this changes results considerably
1. The changes to `ReadColor` and `WriteColor` are not eligible, we need to keep some API compatibility for the 2.0 series.

akallabeth:
@rgfernandes I´ve run valgrind function profiler on a build with:
1.  `cmake -DCMAKE_BUILD_TYPE=Release .` 
1. Connect with `xfreerdp /v:<win7> /u:<user> /p:<pwd>` 
1. Opened a fullscreen youtube video

The only function not inlined (and optimized away) in `freerdp_image_copy` was `SplitColor`

akallabeth:
You might also check out https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html for compiler options first to check if that already does the job (similar options exist for clang and msvc)

akallabeth:
@freerdp-bot test

akallabeth:
I´ve tried with `-DCMAKE_C_FLAGS_RELEASE="-Ofast -DNDEBUG -finline-limit=1000 -fmerge-all-constants"` and nearly all the optimizations you did manually were done by the compiler without changes.

hardening:
@akallabeth does it mean it should be the default c flags when compiling in Release mode ?

akallabeth:
@hardening means if we want to optimize we should start with default `C_FLAGS`, yes. Doing inline that the compiler already knows how to handle does not make much sense, but everything compiler is not able to handle we can, of course, optimize manually.
@rgfernandes as for the `ReadColor` and `WriteColor` changes, if they are still worth the effort, please don´t change existing function signatures in public headers, but add a new one with the desired options. That way the API stays compatible.

akallabeth:
@rgfernandes oh, and let `clang-format` run over the code, @freerdp-bot does not like the formatting.

akallabeth:
@rgfernandes and if you are at optimizing, you might want to replace the `INLINE` definition with
```
#ifdef _WIN32
#define INLINE __forceinline
#else
#define INLINE __attribute__((always_inline)) inline
#endif
``` 

and run some tests.

akallabeth:
@rgfernandes you´re still working on this one?

rgfernandes:
> @rgfernandes you´re still working on this one?

Sorry for delay.
I'm working in a new patch that make this change obsolete.
I'm closing this PR and will open another later.

PR #6126, opened 2020-04-25
Android client version bump

beersheba:
- Update Android library build script for Mac
- Correct android client libraries version

### DIFF
--- a/scripts/android-build-common.sh
+++ b/scripts/android-build-common.sh
@@ -12,7 +12,7 @@ if [ -z $NDK_TARGET ]; then
 fi
 
 if [ -z $CMAKE_PROGRAM ]; then
-  	CMAKE_PROGRAM=$(find $ANDROID_SDK/cmake -name cmake -type f -executable -print -quit)
+  	CMAKE_PROGRAM=$(find $ANDROID_SDK/cmake -name cmake -type f -print -quit)
 fi
 
 if [ -z $CCACHE ]; then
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

scripts/android-build-common.sh
```diff
@@ -12,7 +12,7 @@ if [ -z $NDK_TARGET ]; then
 fi
 
 if [ -z $CMAKE_PROGRAM ]; then
-  	CMAKE_PROGRAM=$(find $ANDROID_SDK/cmake -name cmake -type f -executable -print -quit)
+  	CMAKE_PROGRAM=$(find $ANDROID_SDK/cmake -name cmake -type f -print -quit)
```
akallabeth, line 15: sorry, this breaks the sdk/ndk stuff on our build machines :/

beersheba:
```
Command /opt/android-sdk-linux/cmake/3.10.2.4988404/share/cmake-3.10/completions/cmake not found. Install and add it to the PATH
```
@akallabeth did my change caused it?

akallabeth: yes. it finds the first file matching `cmake` but that is not necessarily an `executable`

beersheba: reverted. can you review it again?

akallabeth:
@freerdp-bot test

PR #6129, opened 2020-04-26
libuwac/input: Fix memory leak

SaschaWessel:
`create_pointer_buffer()` allocates memory with `mmap` and never frees it.

Adding a corresponding `munmap` fixes this issue.

### DIFF
--- a/uwac/libuwac/uwac-input.c
+++ b/uwac/libuwac/uwac-input.c
@@ -26,6 +26,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <assert.h>
+#include <errno.h>
 #include <time.h>
 #include <unistd.h>
 #include <sys/mman.h>
@@ -68,6 +69,10 @@ static struct wl_buffer* create_pointer_buffer(UwacSeat* seat, const void* src, 
 	    wl_shm_pool_create_buffer(pool, 0, seat->pointer_image->width, seat->pointer_image->height,
 	                              seat->pointer_image->width * 4, WL_SHM_FORMAT_ARGB8888);
 	wl_shm_pool_destroy(pool);
+
+	if (munmap(data, size) < 0)
+		fprintf(stderr, "%s: munmap(%p, %" PRIuz ") failed with [%d] %s\n", __FUNCTION__, data,
+		        size, errno, strerror(errno));
 
 error_mmap:
 	close(fd);


akallabeth:
@freerdp-bot test

uwac/libuwac/uwac-input.c
```diff
@@ -69,6 +69,8 @@ static struct wl_buffer* create_pointer_buffer(UwacSeat* seat, const void* src,
 	                              seat->pointer_image->width * 4, WL_SHM_FORMAT_ARGB8888);
 	wl_shm_pool_destroy(pool);
 
+	munmap(data, size);
```
akallabeth, line 72: return value check and log message if failed might be a good idea

hardening:
@SaschaWessel could you add the lo message as suggested by @akallabeth ?

SaschaWessel:
Done. For data++, the output looks as follows:
```
[ERROR][com.freerdp.uwac.input] - munmap(0x00007fe150477001, 9216) failed with [22] Invalid argument
```

akallabeth:
@freerdp-bot test

uwac/libuwac/uwac-input.c
```diff
@@ -69,6 +73,10 @@ static struct wl_buffer* create_pointer_buffer(UwacSeat* seat, const void* src,
 	                              seat->pointer_image->width * 4, WL_SHM_FORMAT_ARGB8888);
 	wl_shm_pool_destroy(pool);
 
+	if (munmap(data, size) < 0)
```
akallabeth, line 76: Sorry, did forget, here we only have `stdlib`, no `winpr` functions available.
use `fprintf`

SaschaWessel: OK. Sure? I have not looked into the CMake files yet, but:
```
$ ldd uwac/libuwac/libuwac0.so.0.0.1 | grep local
	libfreerdp3.so.3 => /home/sascha/.local/lib/libfreerdp3.so.3 (0x00007f0883c3a000)
	libwinpr3.so.3 => /home/sascha/.local/lib/libwinpr3.so.3 (0x00007f0881c44000)
```

akallabeth: yes. the linking against `libfreerdp` is not desired and will be removed, `libwinpr` is only linked privately to `libfreerdp` and not to `libuwac` (`objdump -x libuwac.so` explains it better)

akallabeth:
@freerdp-bot test

Issue #6072, opened 2020-04-14
FFMPEG AAC  Input contains (near) NaN/+-Inf


Clorden:
When using the /microphone option, xfreerdp "crashes" within a few minutes.

This seems to be a new issue in 2.0.0 as it does not happen in 2.0.0 RC4.
I collected the information below before having to roll back to RC4 (this is production machine).

Troubleshooting done:

- Tried with `/audio` and `/microphone:sys:alsa` - same issue
- Tried with `/audio` only - _seems_ stable
- Rolled back to 2.0.0 RC4 - no issue

**To Reproduce**
Run xfreerdp with the /microphone option. See command line below.
The target machine is a Windows 10 VMware VM.

**Expected behavior**
The connection to the target box should be stable.

**Application details**
* Version of FreeRDP : 2.0.0
* Command line used 
`xfreerdp /u:XXX@ZZZ.com /p:YYY /audio /microphone /drive:Shared,/home/xxxxxx/Shared /dynamic-resolution /v:10.125.81.172 /log-level:DEBUG
`
* output of `/buildconfig`

```
[...]

PR #6131, opened 2020-04-28
Workaround #6072: FFMPEG AAC encoding graded experimental

akallabeth:
Due to many reporing issues with different AAC encoder configurations
deactivate support by default. Can be enabled by compiling with
experimental codec support.


kubistika:
LGTM

akallabeth:
@kubistika you need to press the review button in the `Files change` tab ;)

PR #6132, opened 2020-04-30
wlfreerdp/input: Fix mouse wheel / touchpad scroll behavior

SaschaWessel:
Mouse wheel / touchpad scroll speed was too slow / too fast on Wayland.

This pr fixes the conversion of pointer axis scroll event values.

### DIFF
--- a/client/Wayland/wlf_input.c
+++ b/client/Wayland/wlf_input.c
@@ -138,7 +138,6 @@ BOOL wlf_handle_pointer_axis(freerdp* instance, const UwacPointerAxisEvent* ev)
 	rdpInput* input;
 	UINT16 flags = 0;
 	int direction;
-	uint32_t step;
 	uint32_t x, y;
 
 	if (!instance || !ev || !instance->input)
@@ -152,13 +151,14 @@ BOOL wlf_handle_pointer_axis(freerdp* instance, const UwacPointerAxisEvent* ev)
 
 	input = instance->input;
 
-	direction = wl_fixed_to_int(ev->value);
+	direction = wl_fixed_to_double(ev->value) * 10.0;
 	switch (ev->axis)
 	{
 		case WL_POINTER_AXIS_VERTICAL_SCROLL:
 			flags |= PTR_FLAGS_WHEEL;
 			if (direction > 0)
 				flags |= PTR_FLAGS_WHEEL_NEGATIVE;
+			direction = -direction;
 			break;
 
 		case WL_POINTER_AXIS_HORIZONTAL_SCROLL:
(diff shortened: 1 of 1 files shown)


hardening:

client/Wayland/wlf_input.c
```diff
@@ -152,13 +151,14 @@ BOOL wlf_handle_pointer_axis(freerdp* instance, const UwacPointerAxisEvent* ev)
 
 	input = instance->input;
 
-	direction = wl_fixed_to_int(ev->value);
+	direction = wl_fixed_to_double(ev->value) * 10.0;
```
akallabeth, line 154: This is the only relevant change in the pr.
you just multiply the default value by 10.

Everything else down can be removed, as it does the same as the old code. (get an absolute value `step` and add it to the mask)

akallabeth:
@SaschaWessel thank you for the pull request, it gave me the right idea where the bug was hidden.
I´ve created #6133 with a (smaller) fix. you got credits though ;)

client/Wayland/wlf_input.c
SaschaWessel: No, the code changes are required. Please see the following comparisson:

```
axis: 0 | value: -320 | direction/new:  255 | flags/new: 0x2ff | flags/old: 0x340
axis: 0 | value: -256 | direction/new:  255 | flags/new: 0x2ff | flags/old: 0x300
axis: 0 | value: -192 | direction/new:  192 | flags/new: 0x2c0 | flags/old: 0x2c0
axis: 0 | value: -128 | direction/new:  128 | flags/new: 0x280 | flags/old: 0x280
axis: 0 | value:  -64 | direction/new:   64 | flags/new: 0x240 | flags/old: 0x240
axis: 0 | value:    0 | direction/new:    0 | flags/new: 0x200 | flags/old: 0x200
axis: 0 | value:   64 | direction/new:  -64 | flags/new: 0x3c0 | flags/old: 0x340
axis: 0 | value:  128 | direction/new: -128 | flags/new: 0x380 | flags/old: 0x380
axis: 0 | value:  192 | direction/new: -192 | flags/new: 0x340 | flags/old: 0x3c0
axis: 0 | value:  256 | direction/new: -256 | flags/new: 0x300 | flags/old: 0x300
axis: 0 | value:  320 | direction/new: -256 | flags/new: 0x300 | flags/old: 0x340
```

SaschaWessel:
> The pr looks good from the idea, you just use the wrong value range.
> `WheelRotationMask` gives you the maximum value allowed, which is larger than a `uint8_t`

No, there is one sign bit and 8 bits for the value.

PR #6133, opened 2020-05-01
Fixed #6087: Inconsistend scroll on wayland

### LINKED ISSUE #6087: Inconsistent mouse wheel scroll behavior on Wayland
**Describe the bug**
Mouse wheel scrolling is wrong on Wayland. Scrolling down works perfectly, but scrolling up only scrolls about a quarter of the way it would scroll down. Very annoying :-) Everything is fine on X11 (even with Xwayland)

**To Reproduce**
1. Connect to RDP server with wlfreerdp
2. Open a document
3. Use the mouse wheel to scroll down a few times
4. Use the mouse wheel to scroll up a few times

**Expected behavior**
Scrolling down and up scrolls by the same amount of lines.

**Screenshots**
-

**Application details**
* FreeRDP git 971be4fe9
* `wlfreerdp /u:<user> /v:<host>`
* `This is FreeRDP version 3.0.0-dev (971be4fe9)
Build configuration: BUILD_TESTING=OFF BUILTIN_CHANNELS=ON HAVE_AIO_H=1 HAVE_EXECINFO_H=1 HAVE_FCNTL_H=1 HAVE_INTTYPES_H=1 HAVE_JOURNALD_H=TRUE HAVE_MATH_C99_LONG_DOUBLE=1 HAVE_POLL_H=1 HAVE_PTHREAD_MUTEX_TIMEDLOCK=ON HAVE_PTHREAD_MUTEX_TIMEDLOCK_LIB=1 HAVE_PTHREAD_MUTEX_TIMEDLOCK_SYMBOL= HAVE_SYSLOG_H=1 HAVE_SYS_EVENTFD_H=1 HAVE_SYS_FILIO_H= HAVE_SYS_MODEM_H= HAVE_SYS_SELECT_H=1 HAVE_SYS_SOCKIO_H= HAVE_SYS_STRTIO_H= HAVE_SYS_TIMERFD_H=1 HAVE_TM_GMTOFF=1 HAVE_UNISTD_H=1 HAVE_XI_TOUCH_CLASS=1 WITH_ALSA=ON WITH_CAIRO=OFF WITH_CCACHE=ON WITH_CHANNELS=ON WITH_CLANG_FORMAT=ON WITH_CLIENT=ON WITH_CLIENT_AVAILABLE=1 WITH_CLIENT_CHANNELS=ON WITH_CLIENT_CHANNELS_AVAILABLE=1 WITH_CLIENT_COMMON=ON WITH_CLIENT_INTERFACE=OFF WITH_CUPS=ON WITH_DEBUG_ALL=OFF WITH_DEBUG_CAPABILITIES=OFF WITH_DEBUG_CERTIFICATE=OFF WITH_DEBUG_CHANNELS=OFF WITH_DEBUG_CLIPRDR=OFF WITH_DEBUG_DVC=OFF WITH_DEBUG_KBD=OFF WITH_DEBUG_LICENSE=OFF WITH_DEBUG_MUTEX=OFF WITH_DEBUG_NEGO=OFF WITH_DEBUG_NLA=OFF WITH_DEBUG_NTLM=OFF WITH_DEBUG_RAIL=OFF WITH_DEBUG_RDP=OFF WITH_DEBUG_RDPDR=OFF WITH_DEBUG_RDPEI=OFF WITH_DEBUG_RDPGFX=OFF WITH_DEBUG_REDIR=OFF WITH_DEBUG_RFX=OFF WITH_DEBUG_RINGBUFFER=OFF WITH_DEBUG_SCARD=OFF WITH_DEBUG_SND=OFF WITH_DEBUG_SVC=OFF WITH_DEBUG_SYMBOLS=OFF WITH_DEBUG_THREADS=OFF WITH_DEBUG_TIMEZONE=OFF WITH_DEBUG_TRANSPORT=OFF WITH_DEBUG_TSG=OFF WITH_DEBUG_TSMF=OFF WITH_DEBUG_TSMF=OFF WITH_DEBUG_TSMF_AVAILABLE=0 WITH_DEBUG_WND=OFF WITH_DEBUG_X11=OFF WITH_DEBUG_X11_CLIPRDR=OFF WITH_DEBUG_X11_LOCAL_MOVESIZE=OFF WITH_DEBUG_XV=OFF WITH_DSP_EXPERIMENTAL=OFF WITH_DSP_FFMPEG=ON WITH_EVENTFD_READ_WRITE=1 WITH_FAAC=ON WITH_FAAD2=ON WITH_FFMPEG=TRUE WITH_FFMPEG=TRUE WITH_GFX_H264=ON WITH_GPROF=OFF WITH_GSM=ON WITH_GSSAPI=ON WITH_ICU=OFF WITH_IPP=OFF WITH_JPEG=ON WITH_LAME=OFF WITH_LIBRARY_VERSIONING=ON WITH_LIBSYSTEMD=ON WITH_MACAUDIO=OFF WITH_MACAUDIO=OFF WITH_MACAUDIO_AVAILABLE=0 WITH_MANPAGES=ON WITH_MBEDTLS=OFF WITH_OPENCL=ON WITH_OPENH264=OFF WITH_OPENSLES=OFF WITH_OPENSSL=ON WITH_OSS=OFF WITH_PCSC=ON WITH_PROFILER=OFF WITH_PULSE=ON WITH_SAMPLE=OFF WITH_SANITIZE_ADDRESS=OFF WITH_SANITIZE_ADDRESS_AVAILABLE=1 WITH_SANITIZE_MEMORY=OFF WITH_SANITIZE_MEMORY_AVAILABLE=1 WITH_SANITIZE_THREAD=OFF WITH_SANITIZE_THREAD_AVAILABLE=1 WITH_SERVER=OFF WITH_SERVER_INTERFACE=ON WITH_SMARTCARD_INSPECT=OFF WITH_SOXR=OFF WITH_SSE2=ON WITH_SWSCALE=ON WITH_THIRD_PARTY=OFF WITH_VAAPI=OFF WITH_VALGRIND_MEMCHECK=OFF WITH_VALGRIND_MEMCHECK_AVAILABLE=1 WITH_WAYLAND=ON WITH_WINPR_TOOLS=ON WITH_X11=ON WITH_X264=OFF WITH_XCURSOR=ON WITH_XEXT=ON WITH_XFIXES=ON WITH_XI=ON WITH_XINERAMA=ON WITH_XKBFILE=ON WITH_XRANDR=ON WITH_XRENDER=ON WITH_XSHM=ON WITH_XV=ON WITH_ZLIB=ON
Build type:          Release
CFLAGS:               -fPIC -Wall -Wno-unused-result -Wno-unused-but-set-variable -Wno-deprecated-declarations -fvisibility=hidden -Wimplicit-function-declaration -Wredundant-decls -DWINPR_DLL
Compiler:            GNU, 9.3.0
Target architecture: x64
`
* Server is Windows 10 Version 1903
* [wlfreerdp-log.txt](https://github.com/FreeRDP/FreeRDP/files/4485221/wlfreerdp-log.txt)

**Desktop (please complete the following information):**
 - OS: Linux Manjaro

akallabeth:
Thanks to @yol and @SaschaWessel a bug in scroll step conversion
was uncovered. The RDP value ranges are inverted when scrolling
in negative direction.

### DIFF
--- a/client/Wayland/wlf_input.c
+++ b/client/Wayland/wlf_input.c
@@ -171,9 +171,19 @@ BOOL wlf_handle_pointer_axis(freerdp* instance, const UwacPointerAxisEvent* ev)
 			return FALSE;
 	}
 
-	step = (uint32_t)abs(direction);
-	if (step > WheelRotationMask)
-		step = WheelRotationMask;
+	/* Wheel rotation steps:
+	 *
+	 * positive: 0 ... 0xFF  -> slow ... fast
+	 * negative: 0 ... 0xFF  -> fast ... slow
+	 */
+	step = abs(direction);
+	if (step > 0xFF)
+		step = 0xFF;
+
+	/* Negative rotation, so count down steps from top */
+	if (flags & PTR_FLAGS_WHEEL_NEGATIVE)
+		step = 0xFF - step;
+
 	flags |= step;
 
 	return freerdp_input_send_mouse_event(input, flags, (UINT16)x, (UINT16)y);


SaschaWessel:
Sorry, I should have explained why the multiplication by 10 in #6132 was intended.

AFAIK, `ev->value` for one scroll event is 10 or 15 depending on the compositor.
For Windows the intended value seems to be 0x78 = 120. So we have to multiply `ev->value`
by (120/10) = 12 or (120/15) = 8. Mean value is 10.0.

client/Wayland/wlf_input.c
```diff
@@ -171,9 +173,12 @@ BOOL wlf_handle_pointer_axis(freerdp* instance, const UwacPointerAxisEvent* ev)
...
 
-	step = (uint32_t)abs(direction);
-	if (step > WheelRotationMask)
-		step = WheelRotationMask;
+	/* Wheel rotation steps:
+	 *
+	 * positive: 0 ... 0xFF  -> slow ... fast
+	 * negative: 0 ... 0xFF  -> fast ... slow
+	 */
+	step = (uint32_t)direction & 0xFF;
```
SaschaWessel, line 181: This will not work as expected, for `ev->value == 256`, `step` should be `255` not `0`.

akallabeth:
@SaschaWessel ok, the value 0x78 was just a value picked at random for x11.
The 0xff mask worked fine here, while the original did scroll up slow it works as fast as down with these changes.

akallabeth:
@SaschaWessel but true, missed the saturation step, will add when back at my machine.

yol:
I'll try it out tomorrow. I assume though that without scaling, behavior when using touchpads might be suboptimal

yol:
I have trouble believing that it is a random choice to be honest, 0x78 is exactly WHEEL_DELTA in WM_MOUSEWHEEL

akallabeth:
@yol you might be right, but I'd like to have a bit of information about the value range of wayland first. There are lots of devices out there with different sensitivity and I don't want to code something like this in there without some good reason.

akallabeth:
found the source for wayland: https://wayland.freedesktop.org/libinput/doc/latest/api/group__event__pointer.html#ga2116f4bbedb61532e71d16c4f87bd4ca

akallabeth:
@hardening looks like we need to patch `uwac` for this to properly work.
I´d suggest merging this as is and adding a way to expose `axis_source`, `axis_stop` and `axis_discrete` events.
That way we could distinguish between different scrolling sources (touch, mouse wheel, and continuous)

yol:
@akallabeth If you meant that link as source for the value of "15", it might not be what you think it is. libinput (which to the extent of my knowledge every major Wayland compositor is using *internally*) reports wheel clicks in degrees, with 15 being the default for most mice. However, that is not what the Wayland axis event is about.

See: https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_pointer-event-axis - the `value` property is "length of vector in surface-local coordinate space", so it has nothing to do with degrees and some kind of conversion has to be done already in the compositor. Those that do this conversion (weston - the reference, mutter - GNOME) usually normalize one click to the value of 10, but I do not think this value is specified anywhere, as it is the "freedom" of the compositor to decide how far it wants one wheel click to scroll in terms of surface coordinates. Others (e.g., sway) pass through the libinput value unmodified, which may or may not make much sense (to me it does not - you can argue though that it is also a scaling/normalization, but to the value of 15 instead of 10).

Besides, you will always need some kind of scaling in FreeRDP, even if you use the supplementary axis* events. Touchpads do not send discrete events and the values of their continuous `axis` events will be useless unless scaled, since Windows/RDP is clearly using a different value range.

The cleanest approach IMO would be to map 1 `axis_discrete` step to 1 `WHEEL_DELTA` and use scaling with either 10 or 15 = 1 `WHEEL_DELTA` for continuous scroll events. I'm not sure it's worth the trouble though, since you will get almost identical behavior by just using the regular `axis` events and scaling.

akallabeth:
@yol agree with you, but see previous comment.
We need to expose the events first, and this commit fixes the inconsistent scrolling.
When the discrete/non discrete events are exposed we can then start mapping as you suggested.

yol:
Yep, if you want to go the 100% clean route :+1: Thanks for fixing! At least scroll wheel should work now.

PR #6135, opened 2020-05-01
Fixed copy paste error in MessagePipe.c

qarmin:
`!pipe->In` was checked twice but `!pipe->Out` zero times


akallabeth:
@freerdp-bot test

PR #6138, opened 2020-05-04
Fixed warning messages for RDP file options

akallabeth:
networkautodetect and bandwidthautodetect are both boolean options
which must match each other. Added adequate warnings.

### DIFF
--- a/client/common/file.c
+++ b/client/common/file.c
@@ -1614,30 +1614,14 @@ BOOL freerdp_client_populate_settings_from_rdp_file(rdpFile* file, rdpSettings* 
 	{
 		if (file->BandwidthAutoDetect != 0)
 		{
-			if (!freerdp_set_connection_type(settings, CONNECTION_TYPE_AUTODETECT))
-				return FALSE;
-			setDefaultConnectionType = FALSE;
-		}
-		if (!freerdp_settings_set_bool(settings, FreeRDP_NetworkAutoDetect,
-		                               file->BandwidthAutoDetect != 0))
-			return FALSE;
-	}
-
-	if (~file->NetworkAutoDetect)
-	{
-		if (file->NetworkAutoDetect == 0)
-		{
-			if (file->BandwidthAutoDetect != CONNECTION_TYPE_AUTODETECT)
+			if ((~file->NetworkAutoDetect) && (file->NetworkAutoDetect != 0))
 			{
 				WLog_WARN(TAG,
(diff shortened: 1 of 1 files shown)


client/common/file.c
```diff
@@ -1614,38 +1614,48 @@ BOOL freerdp_client_populate_settings_from_rdp_file(rdpFile* file, rdpSettings*
 	{
 		if (file->BandwidthAutoDetect != 0)
 		{
+			if (file->NetworkAutoDetect != 0)
```
mfleisz, line 1617: This check is ambiguous as it is also true if NetworkAutoDetect is not set in the file. We should only output a warning if it is explicitly wrong in the rdp file.

client/common/file.c
```diff
@@ -1614,38 +1614,48 @@ BOOL freerdp_client_populate_settings_from_rdp_file(rdpFile* file, rdpSettings*
...
-		                               file->BandwidthAutoDetect != 0))
+		                               (file->BandwidthAutoDetect != 0) ||
+		                                   (file->NetworkAutoDetect == 0)))
 			return FALSE;
 	}
 
 	if (~file->NetworkAutoDetect)
 	{
-		if (file->NetworkAutoDetect == 0)
+		if (file->NetworkAutoDetect != 0)
```
mfleisz, line 1639: I think this should be `== 0` (is enabled)

client/common/file.c
```diff
@@ -1614,38 +1614,48 @@ BOOL freerdp_client_populate_settings_from_rdp_file(rdpFile* file, rdpSettings*
...
-				          file->NetworkAutoDetect, file->BandwidthAutoDetect,
-				          CONNECTION_TYPE_AUTODETECT);
-				WLog_WARN(TAG,
-				          "Add bandwidthautodetect:i:%" PRIu32
-				          " to your RDP file to eliminate this warning.",
-				          CONNECTION_TYPE_AUTODETECT);
+				          ". Correcting to bandwidthautodetect:i:0",
+				          file->NetworkAutoDetect, file->BandwidthAutoDetect);
+				WLog_WARN(
+				    TAG, "Add bandwidthautodetect:i:0 to your RDP file to eliminate this warning.");
```
mfleisz, line 1648: Should be `i:1`

akallabeth: @mfleisz sure? this is the stuff inverted thing again

PR #6144, opened 2020-05-05
Fixed #6136: Cleaned up Stack API

### LINKED ISSUE #6136: Never exited critical section when error in realloc occurs
```
if (!new_arr)
	return;
```
should look probably
```
if (!new_arr){
	if (stack->synchronized)
		LeaveCriticalSection(&stack->lock);
	return;
}
```

https://github.com/FreeRDP/FreeRDP/blob/02614cce49ce6ff120a1306962c5d3ff6b05ea75/winpr/libwinpr/utils/collections/Stack.c#L125-L137

akallabeth:


winpr/libwinpr/utils/collections/Stack.c
```diff
@@ -127,20 +144,19 @@ void Stack_Push(wStack* stack, void* obj)
...
-		int new_cap;
-		void** new_arr;
-		new_cap = stack->capacity * 2;
-		new_arr = (void**)realloc(stack->array, sizeof(void*) * new_cap);
+		const size_t new_cap = stack->capacity * 2;
+		void** new_arr = (void**)realloc(stack->array, sizeof(void*) * new_cap);
 
-		if (!new_arr)
-			return;
+		if (new_arr)
```
kubistika, line 150: Wrong condition?

akallabeth: @kubistika good catch

PR #6145, opened 2020-05-05
avoid swamping console with SCARD_E_TIMEOUT errors

frispete:
Here's my proposal to deal with excessive smartcard timeout logs.


akallabeth:
duplicate #6140

frispete:
#6140 is much better, indeed. Thanks again.

PR #6173, opened 2020-05-12
Drdynvc cleanup

akallabeth:


hardening:
Shouldn't this also got to stable ?

PR #6179, opened 2020-05-13
RFX improvement

Fotosmile:
1. Fix typo in GFX logs. The log output had `left` word while the actual value was `bottom`;
2. Improved RFX DWT vertical direction inverse. I combined multiple loops into one since the calculated value of the `even` coefficients can be used in the `odd` coefficients right away without affecting `even` coefficients. I ran the client and everything works fine, moreover, the performance of the inverse was improved about 4x (by quick measurements).


akallabeth:
@freerdp-bot test

hardening:
@Fotosmile out of curiosity were you compiling in release or debug mode when you did your performance test ?

Fotosmile:
@hardening, just a second ago retested in release mode, and for RFX test and real session, the average improvement in the optimized section of code is x2.

Issue #455, opened 2012-02-22
Can't send CTRL-ALT-DEL to remote desktop


dbrandesky:
I see that commands like "Alt-F4" have been fixed in the past, but I can't seem to send a ctrl-alt-del to my RDP clients through freerdp. If the RDP window is active the command does seem to be grabbed by that window (Ubuntu doesn't offer me the shutdown dialog box) but the client definitely doesn't receive it; taskmgr doesn't appear in windows XP, and I can't log into my Windows Server 2003 machine because it requires that key sequence at login. I searched around and didn't see anyone else with this issue, so if I missed something obvious please let me know. Thanks!

antiflu:
I'm on Linux Mint and the Ctrl-Alt-Del as well as the Ctrl-Alt-End buttons are taken by the OS as shortcuts. Only after you remove them in Preferences-Keyboard, FreeRDP will pick them up.

kefir-:
This still appears to be an issue on Ubuntu 16.04. It'd be nice if there were some way to define and send special key combos. Most other similar software I've seen have that, like VirtualBox, other VNC clients, and other remote console applications.

amtlib-dot-dll:
It seems that Wayland deliberately forbids global key grabbing due to [keyloggers](https://lists.fedoraproject.org/archives/list/desktop@lists.fedoraproject.org/thread/BBPOOC4GVBJXHLKSSFLGERG6GYAU6M6U/) and this issue may not be fixed in a [long time](https://bugzilla.redhat.com/show_bug.cgi?id=1285770), so I suggest that FreeRDP to provide an extra console receiving commands, for example users can type commands like `send ^~del` to send Ctrl-Alt-Del to the remote computer when `xfreerdp` is launched from a terminal window.

Issue #3494, opened 2016-09-09
Feature Request: Map Ctrl-Alt-End to Ctrl-Alt-Del


HenryJacques:
Windows users are used to use to keyboard combination Ctrl+Alt+End when using mstsc in order to change their password for example/

When using FreeRDP, they can directly use Ctrl+Alt+Del.

rdesktop has this feature to remap Ctrl+Alt+End to Ctrl+Alt+Del
https://sourceforge.net/p/rdesktop/patches/187/

Is this also possible with FreeRDP

thanks in advance!

PR #6186, opened 2020-05-17
Map Ctrl+Alt+Del to Ctrl+Alt+End

tcheneau:
On Microsoft Windows systems, when connecting to a remote desktop with
mstsc.exe, Ctrl+Alt+Del does not open the task manager on the remote
host. It will be intercepted by the current system and opens the task
manager locally. A different key combo is used to pass Ctrl+Alt+Del to
the remote host: Ctrl+Alt+End.

Similarly with xfreerdp, the Ctrl+Alt+Del combo is often interpreted by
the Window Manager. This patch maps Ctrl+Alt+End to pass the
Ctrl+Alt+Del combo.

Fixes issue #455 and #3494

Regards,
Tony

P.-S.: this patch was provided by a colleague of mine, hence the different author name

### DIFF
--- a/client/X11/xf_keyboard.c
+++ b/client/X11/xf_keyboard.c
@@ -43,6 +43,9 @@
 #include <freerdp/log.h>
 #define TAG CLIENT_TAG("x11")
 
+static BOOL controlDown = FALSE;
+static BOOL laltDown = FALSE;
+
 static BOOL firstPressRightCtrl = TRUE;
 static BOOL ungrabKeyboardWithRightCtrl = TRUE;
 
@@ -191,6 +194,28 @@ void xf_keyboard_send_key(xfContext* xfc, BOOL down, BYTE keycode)
 	DWORD rdp_scancode;
 	rdpInput* input;
 	input = xfc->context.input;
+
+	if (keycode == 0x25)
+	{
+		if (down)
+			controlDown = TRUE;
+		else
+			controlDown = FALSE;
+	}
+
+	if (keycode == 0x40)
+	{
+		if (down)
+			laltDown = TRUE;
+		else
+			laltDown = FALSE;
+	}
+
+	if (laltDown && controlDown && keycode == 0x73 && down == 1)
+	{
+		keycode = 0x77; // Del key
+	}
+
 	rdp_scancode = freerdp_keyboard_get_rdp_scancode_from_x11_keycode(keycode);
 
 	if (rdp_scancode == RDP_SCANCODE_UNKNOWN)


client/X11/xf_keyboard.c
```diff
@@ -43,6 +43,9 @@
 #include <freerdp/log.h>
 #define TAG CLIENT_TAG("x11")
 
+static BOOL controlDown = FALSE;
```
akallabeth, line 46: These should be part of `xfContext` (same is true for the existing ones, looks like a bug)

client/X11/xf_keyboard.c
```diff
@@ -191,6 +194,28 @@ void xf_keyboard_send_key(xfContext* xfc, BOOL down, BYTE keycode)
 	DWORD rdp_scancode;
 	rdpInput* input;
 	input = xfc->context.input;
+
+	if (keycode == 0x25)
```
akallabeth, line 198: Better check against `rdp_scancode` and the `RDP_SCANCODE*` macros instead of these magic numbers

client/X11/xf_keyboard.c
```diff
@@ -191,6 +194,28 @@ void xf_keyboard_send_key(xfContext* xfc, BOOL down, BYTE keycode)
...
+	{
+		if (down)
+			controlDown = TRUE;
+		else
+			controlDown = FALSE;
+	}
+
+	if (keycode == 0x40)
```
akallabeth, line 206: Better check against `rdp_scancode` and the `RDP_SCANCODE*` macros instead of these magic numbers

client/X11/xf_keyboard.c
```diff
@@ -191,6 +194,28 @@ void xf_keyboard_send_key(xfContext* xfc, BOOL down, BYTE keycode)
...
+	{
+		if (down)
+			laltDown = TRUE;
+		else
+			laltDown = FALSE;
+	}
+
+	if (laltDown && controlDown && keycode == 0x73 && down == 1)
```
akallabeth, line 214: Better check against `rdp_scancode` and the `RDP_SCANCODE*` macros instead of these magic numbers

ckujau:
Workaround, until this is merged (and because I could not find this documented anywhere else): disable the `CTRL`-`ALT`-`DEL` keyboard shortcut in your window manager, that way `xfreerdp` should be able catch that shortcut and will be able to forward it to the remote system. For example, in the Gnome3 [Settings](https://wiki.gnome.org/Design/SystemSettings/Keyboard "Gnome Common keyboard related settings.") program:

gnome_3

jdranchman:
ctrl-alt-end works in the W10 rdp client. It should also work for freerdp (IMHO) for compatibility

Neustradamus:
@tcheneau: Have you progressed on this PR?

tcheneau:
@Neustradamus No, I didn't have time to get back to it despite all the very valid comments offered by @akallabeth. I left my employer since I first submitted this PR and no longer have a working setup to try the patch. I'm OK if this PR is dropped.

PR #6187, opened 2020-05-18
change use of strtok to strtok_s

kubistika:
This should also fix #6139.

PR #6188, opened 2020-05-18
Replaced strtok with strtok_s

akallabeth:
Follow up to #6187

### DIFF
--- a/client/X11/xf_keyboard.c
+++ b/client/X11/xf_keyboard.c
@@ -74,7 +74,8 @@ static BOOL xf_keyboard_action_script_init(xfContext* xfc)
 
 	while (fgets(buffer, sizeof(buffer), keyScript) != NULL)
 	{
-		strtok(buffer, "\n");
+		char* context = NULL;
+		strtok_s(buffer, "\n", &context);
 		keyCombination = _strdup(buffer);
 
 		if (!keyCombination || ArrayList_Add(xfc->keyCombinations, keyCombination) < 0)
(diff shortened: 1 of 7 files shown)


client/X11/xf_keyboard.c
```diff
@@ -74,7 +74,8 @@ static BOOL xf_keyboard_action_script_init(xfContext* xfc)
 
 	while (fgets(buffer, sizeof(buffer), keyScript) != NULL)
 	{
-		strtok(buffer, "\n");
+		char* context = NULL;
```
kubistika, line 77: Shouldn’t this be outside of the loop?

libfreerdp/crypto/tls.c
```diff
@@ -1198,9 +1198,10 @@ static BOOL is_accepted_fingerprint(CryptoCert cert, const char* CertificateAcce
 		char* cur = strtok_s(copy, ",", &context);
 		while (cur)
 		{
+			char* subcontext = NULL;
```
kubistika, line 1201: Same

client/X11/xf_keyboard.c
akallabeth: no, `strtok` is never called with a `NULL` argument, so each tokenize is a new context

libfreerdp/crypto/tls.c
akallabeth: no, the context must be reset for each new `cur`

PR #6193, opened 2020-05-19
server: proxy: code refactor 

kubistika:


server/proxy/modules/modules_api.h
```diff
@@ -48,10 +48,13 @@ typedef struct proxy_plugin
 
 	/* proxy hooks. a module can set these function pointers to register hooks */
 	proxyHookFn ClientPreConnect;
+	proxyHookFn ClientPostConnect;
```
akallabeth, line 51: we have already released a version, so better add this to the end of the struct to avoid issues

akallabeth:
@freerdp-bot test

PR #6207, opened 2020-05-20
Abort on first possible certificate validation error

akallabeth:
Only retry certificate validation if the purpose was wrong.


akallabeth:
@freerdp-bot test

PR #6211, opened 2020-05-20
winpr: utils: sam: fix strtok_s call

kubistika:


akallabeth:
Already part of #6210

Issue #6236, opened 2020-05-27
Loading of video channel fails with "undefined symbol: mappedGeometryRef"


ondrejholy:
**Describe the bug**
Usage of `/video` option cause the following failure if freerdp is built with  `-DBUILTIN_CHANNELS=OFF`:
```
$ xfreerdp /u:user /p:password /v:server /video
[08:51:08:567] [387041:387042] [INFO][com.freerdp.core] - freerdp_connect:freerdp_set_last_error_ex resetting error state
[08:51:08:567] [387041:387042] [INFO][com.freerdp.client.common.cmdline] - loading channelEx rdpdr
[08:51:08:568] [387041:387042] [INFO][com.freerdp.client.common.cmdline] - loading channelEx rdpsnd
[08:51:08:568] [387041:387042] [INFO][com.freerdp.client.common.cmdline] - loading channelEx cliprdr
[08:51:08:568] [387041:387042] [INFO][com.freerdp.client.common.cmdline] - loading channelEx drdynvc
[08:51:09:946] [387041:387042] [INFO][com.freerdp.primitives] - primitives autodetect, using optimized
[08:51:09:947] [387041:387042] [INFO][com.freerdp.core] - freerdp_tcp_is_hostname_resolvable:freerdp_set_last_error_ex resetting error state
[08:51:09:947] [387041:387042] [INFO][com.freerdp.core] - freerdp_tcp_connect:freerdp_set_last_error_ex resetting error state
[08:51:09:096] [387041:387042] [WARN][com.freerdp.crypto] - Certificate verification failure 'unable to get local issuer certificate (20)' at stack position 0
[08:51:09:096] [387041:387042] [WARN][com.freerdp.crypto] - CN = DEMOWIN
[08:51:10:301] [387041:387042] [INFO][com.freerdp.gdi] - Local framebuffer format  PIXEL_FORMAT_BGRX32
[08:51:10:301] [387041:387042] [INFO][com.freerdp.gdi] - Remote framebuffer format PIXEL_FORMAT_RGB16
[08:51:10:341] [387041:387042] [INFO][com.winpr.clipboard] - initialized POSIX local file subsystem
[08:51:10:345] [387041:387042] [INFO][com.freerdp.channels.rdpsnd.client] - [static] Loaded fake backend for rdpsnd
[08:51:10:345] [387041:387042] [INFO][com.freerdp.channels.drdynvc.client] - Loading Dynamic Virtual Channel geometry
[...]

PR #6238, opened 2020-05-27
winpr/library: Use RTLD_GLOBAL for dlopen

ondrejholy:
LoadLibraryA implementation uses the RTLD_LOCAL flag for dlopen currently.
This flag doesn't allow the symbols to be used by the subsequently loaded
libraries. This is a problem for the video channel when -DBUILTIN_CHANNELS=OFF
is used as it uses functions from the geometry channel. Let's use RTLD_GLOBAL
instead to prevent "undefined symbol" errors in such cases.

Fixes: https://github.com/FreeRDP/FreeRDP/issues/6236

### DIFF
--- a/winpr/libwinpr/library/library.c
+++ b/winpr/libwinpr/library/library.c
@@ -125,7 +125,7 @@ HMODULE LoadLibraryA(LPCSTR lpLibFileName)
 	return hModule;
 #else
 	HMODULE library;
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #6240, opened 2020-05-28
Configure USB interface only when necessary

Raven24:
This PR contributes to solving #6150 - USB interrupt events were not triggered, since the device was apparently set to an invalid config/state, when initialized. The code around the libusb function was adapted to only run when actually called with a changed value. It appears that was already the intended behaviour, I just inverted the logic around it.

I tested with the HID device Olympus Optical Co., Ltd Foot Switch RS-28 (ID 07b4:0218) - without this change, none of the button presses trigger interrupts. With the patch applied, buttons are triggering as expected and forwarded in the RDP session.

Let me know, if this needs more work before it can be merged.

/re @akallabeth


akallabeth:
@freerdp-bot test

akallabeth:
@Raven24 please run `clang-format` on the changes, `git clang-format HEAD~1` should do it if installed.

Raven24:
sure, updated the branch

kubistika:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #6244, opened 2020-05-29
USB pass actual length of bytes transferred to callback function

Raven24:
The libusb transfer reports the actual number of bytes written, use that to pass to the callback function instead of the size of the allocated buffer. (contributes to solving #6150)
This distinction is important, since a regular interrupt event will produce valid, usable data, but a timeout would just repeat stale data of the last interrupt or whatever happens to be in the buffer at that time.

/re @akallabeth


akallabeth:
@freerdp-bot test

PR #6247, opened 2020-05-29
Fix memory leaks in client/encomsp_main.c and client/remdesk_main.c

patrickchin:
Similar fix to #2278. Not sure how to trigger encomsp PDUs with xfreerdp but the leak was seen while using freerdp library


akallabeth:
@freerdp-bot test

PR #6248, opened 2020-05-30
[generate_argument_docbook] Fix typo in XML entity.

sgtatham:
The character '>' was being rendered as &lt; instead of &gt;.


akallabeth:
@freerdp-bot test

PR #6256, opened 2020-06-03
Do not remove transfer data on usb cancel transfer

akallabeth:
Fixes a double free in case of cancelled USB transfers.
libusb will call `func_bulk_transfer_cb` which removes the data.
Same for `urb_isoch_transfer_cb`

### DIFF
--- a/channels/rdpei/rdpei_common.c
+++ b/channels/rdpei/rdpei_common.c
@@ -306,7 +306,7 @@ BOOL rdpei_write_4byte_signed(wStream* s, INT32 value)
 		value *= -1;
 	}
 
-	if (value <= 0x1FUL)
+	if (value <= 0x1FL)
 	{
 		byte = value & 0x1F;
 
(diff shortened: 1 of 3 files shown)


akallabeth:
@Raven24 @trentasis found a few issues with USB channel and some double free conditions, which should be fixed with this pr.

trentasis:
Ok, I can try, but in where branch is this change, on stable is not available?

trentasis:
Hi, sorry I ahve found, I think taht is https://github.com/akallabeth/FreeRDP/tree/usb_cancel_fix

trentasis:
Hi restarted windows OS, then compiled and tried with https://github.com/akallabeth/FreeRDP/commits/usb_cancel_fix and during some second webcam device hasbeen detected on windwos, but then dissappeared... Attached logs 
[20200603.txt](https://github.com/FreeRDP/FreeRDP/files/4723198/20200603.txt)

Raven24:
I tested this branch with my device, but I don't know if that changes anything in my use-case - I don't think any requests are cancelled usually. But I'm always a fan of proper memory management ;)

PR #6257, opened 2020-06-03
Fixes for some defects found by covscan

ondrejholy:
See commit descriptions for more details about each fix.


akallabeth:
@ondrejholy please run `git clang-format` on the source, looks like some formatting is off.

ondrejholy:
> @ondrejholy please run `git clang-format` on the source, looks like some formatting is off.

Sorry about that and thanks for the fix. It would be good to mention `git clang-format` in `PULL_REQUEST_TEMPLATE.md` because it mentions just the `cmake` target, which reformates the whole codebase and there are many unrelated style issues, so it is easy to miss those relevant...

akallabeth:
@ondrejholy oh, yes, good point.

PR #6268, opened 2020-06-08
uwac: don't try to use O_TMPFILE on FreeBSD

kevans91:
Currently, this sets an invalid open flag and attempts to open(), which
will fail.  Instead of doing that, don't try to define O_TMPFILE where
such a definition can't exist and force the fallback rather than making
an always-fail open() call.


akallabeth:
@freerdp-bot test

Issue #6267, opened 2020-06-07
[bisected] 2.1.1: fails 169 - TestFreeRDPCodecInterleaved on ppc (32bit, Big Endian)


ernsteiswuerfel:
2.1.0 passes the testsuite whereas 2.1.1 fails 169 - TestFreeRDPCodecInterleaved on ppc (32bit, Big Endian). Found that out during running tests on Gentoo Linux (downstream [bug 727446](https://bugs.gentoo.org/727446)).

The test also fails in git master.

I did a git bisect replicating Gentoo's build parameters:
```
mkdir build && cd build
cmake -DCMAKE_INSTALL_PREFIX=/usr -DBUILD_TESTING=ON -DCHANNEL_URBDRC=ON -DWITH_ALSA=ON -DWITH_CCACHE=OFF -DWITH_CUPS=OFF -DWITH_DEBUG_ALL=OFF -DWITH_MANPAGES=OFF -DWITH_FFMPEG=ON -DWITH_SWSCALE=ON -DWITH_CAIRO=OFF -DWITH_DSP_FFMPEG=ON -DWITH_GSTREAMER_1_0=ON -DWITH_JPEG=ON -DWITH_NEON=OFF -DWITH_OPENH264=OFF -DWITH_PULSE=ON -DWITH_SERVER=OFF -DWITH_PCSC=OFF -DWITH_LIBSYSTEMD=ON -DWITH_X11=ON -DWITH_XINERAMA=OFF -DWITH_XV=ON -DWITH_WAYLAND=OFF -DCMAKE_BUILD_TYPE=Gentoo ..
cmake --build . && ctest -j 24 --test-load 24
```
and got the following commit as result of the bisect:
```
 # git bisect bad | tee -a ~/build/bisect01.log
96621d4f459dd2b3f8b5dbe812c7aa2f1a5859f3 is the first bad commit
commit 96621d4f459dd2b3f8b5dbe812c7aa2f1a5859f3
Author: akallabeth <akallabeth@posteo.net>
Date:   Fri May 15 11:16:18 2020 +0200

    Always require aligned memory for interleaved codec.
    
    (cherry picked from commit 20c11fb7178462a8ba36b4fdbfb3e2a03197356f)

 CMakeLists.txt                 | 4 ----
 config.h.in                    | 1 -
 libfreerdp/codec/interleaved.c | 4 ----
 3 files changed, 9 deletions(-)
```
[...]

PR #6269, opened 2020-06-08
Write pixel value as little-endian in write_pixel_16

floppym:
Fixes: 722790f4cadc3f2d551bf42702a99b3a3a69e5a3
Closes: https://github.com/FreeRDP/FreeRDP/issues/6267


akallabeth:
@floppym ok, looks like we came to the same conclusion (see #6270 ) at about the same time ^^

PR #6273, opened 2020-06-08
MessageQueue write time to current message not the next

patrickchin:
Is this a bug?
`time` is not accessed anywhere so has no effect however shouldn't the time be set for the current message rather than the next?


akallabeth:
@freerdp-bot test

PR #6297, opened 2020-06-22
create_release_tarball: run: redirect status messages to stderr

floppym:
Fixes .source_version in release tarballs.

```
winpr/include/winpr/version.h:30:22: warning: missing terminating " character
   30 | #define GIT_REVISION "[RUN] echo 2.1.2
      |                      ^
winpr/include/winpr/version.h:31:1: error: too many decimal points in number
   31 | 2.1.2"
      | ^~~~~
```


akallabeth:
@floppym thanks for the fix. I´ve created #6301 instead as I was not completely happy with cluttering `stderr` with run messages.

PR #6304, opened 2020-06-23
[WIP] 2.1.2 fixes

akallabeth:


joakim-tjernlund:
Could you merge to stable as you go? Merging and releasing in one go does not get any testing before release.

akallabeth:
@joakim-tjernlund stable only receives cherry-picks from master, so feedback for master is the way to go ;)

joakim-tjernlund:
> @joakim-tjernlund stable only receives cherry-picks from master, so feedback for master is the way to go ;)

master is not ABI compatible and new features I don't need. I do use stable , to test and possible get a bug fix I need ASAP.
The current method makes stable virtually untested before release, how do you know it is any good?
This bug could have been detected if you had pushed as you go.

akallabeth:
@joakim-tjernlund this change is in there for at least a year now and this was the first feedback regarding that.
As for ABI compatible, ok, so you don´t want to test, that is fine.

akallabeth:
Further investigation exposed the changes are not required.

PR #6319, opened 2020-06-26
Stable 2 backports from master

akallabeth:


pnowack:
@akallabeth Could you also consider including the commit from https://github.com/FreeRDP/FreeRDP/pull/6320?

PR #6331, opened 2020-06-29
Add support for -grab-mouse, to prevent remote mouse movement

b01t:
freerdp2 introduced some annoying behaviour for me: When switching virtual desktops to one where freerdp2 is running and focused, the pointer would be moved to the last position it had inside of the freerdp2 window. I.e. if it was moved out at the upper right corner, that's where it would go. This happens regardless of whether or not the pointer was positioned on top of the freerdp2 window at the time of switching.

The main annoyance caused by this is that when I used the scroll wheel on my dock to flip through my virtual desktops, the mouse would suddenly pop into the freerdp2 window, and continued scrolling would scroll whichever document I happened to have open on the remote Windows desktop.

This was tested on Debian 10 (Buster), with Openbox WM and a tint2 dock. It apparently happens only when logged into a Windows 10 machine (tested with 2 different ones), and not when loggied into a Windows Server 2016 (or 2019) machine (tested with 3 different ones). I also tested freerdp (1) on the same setup, which did not exhibit this behaviour.

With the included patch, -grab-mouse can be specified on the command line to prevent freerdp2 from moving the user's mouse.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -2896,6 +2896,10 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings, 
 		{
 			settings->GrabKeyboard = enable;
 		}
+		CommandLineSwitchCase(arg, "grab-mouse")
+		{
+			settings->GrabMouse = enable;
+		}
 		CommandLineSwitchCase(arg, "unmap-buttons")
 		{
 			settings->UnmapButtons = enable;
--- a/client/common/cmdline.h
+++ b/client/common/cmdline.h
@@ -180,6 +180,7 @@ static const COMMAND_LINE_ARGUMENT_A args[] = {
 	{ "gp", COMMAND_LINE_VALUE_REQUIRED, "<password>", NULL, NULL, -1, NULL, "Gateway password" },
 	{ "grab-keyboard", COMMAND_LINE_VALUE_BOOL, NULL, BoolValueTrue, NULL, -1, NULL,
 	  "Grab keyboard" },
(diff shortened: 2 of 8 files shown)


akallabeth:
@freerdp-bot test

akallabeth (dismissed):
LGTM
nice addition!

akallabeth:
@b01t please run `clang-format` (or `git clang-format`) on the source

bmiklautz:
@b01t first of thank you for your PR. Well done! If you could update the formatting issue wie can merge it.


@b01t  @akallabeth for completeness: As far as I remember the update_pointer is only sent from the server if the RDP window lost the focus by moving the mouse outside (at least for Win7 and Win10 - didn't test the other systems).

If you for example switch between windows with some key combination (assuming -grab-keyboard was used) this doesn't happen and the mouse stays where it was when the window was left.  So for me it looks like that in the combination used by @b01t using the mouse wheel is interpreted the same way as if the mouse was moved out the window (and therefore the server updates the pointer position on focus).

bmiklautz:
A side-effect to mention: if -grab-mouse is used remote assistant (screen sharing/mirroring) will not update the pointer.

bmiklautz:
@b01t please run clang-format (or git clang-format) on the source then our tester is happy and we can merge the PR.
Thanks.

akallabeth (dismissed):

akallabeth:
@freerdp-bot test

akallabeth (approved):

b01t:
@bmiklautz Hey. Terribly sorry about not responding to your comments. Turns out I don't get notified about mentions anymore. I did however get notified about the merge. I'll fix my notifications and do better in the future :)

PR #6333, opened 2020-06-30
Removed use of empty_region

akallabeth:
* Simplify region handling
* Remove hacks that require freeing a const pointer

### DIFF
--- a/libfreerdp/codec/region.c
+++ b/libfreerdp/codec/region.c
@@ -74,19 +74,17 @@ struct _REGION16_DATA
 	long nbRects;
 };
 
-static REGION16_DATA empty_region = { 0, 0 };
-
 void region16_init(REGION16* region)
 {
 	assert(region);
 	ZeroMemory(region, sizeof(REGION16));
-	region->data = &empty_region;
 }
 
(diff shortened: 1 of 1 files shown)


hardening:
Well I like the empty region thing, because it allowed to completely remove tests on `data`. I'd rather propose to add a ref counter in `REGION16_DATA`, and if the empty region starts with a 1 counter, then it will never be freed and we can have a standard case in every places.
Not sure we will use it but it will also allow to do some region aliasing (2 region pointing on the same data).

akallabeth:
@hardening that would break the API as `REGION16_DATA` is exposed.
Restructured it this way to avoid the hack (free on const data) and simplify tests while keeping compatibility with 2.1.2

hardening:
I disagree `REGION16_DATA` is not exposed, it's presented as an opaque type.

hardening:
@akallabeth what about https://github.com/hardening/FreeRDP/tree/region_ref ?

bmiklautz:
@akallabeth I remove consider for stable (for now) as it's not necessary for stable.

akallabeth:
@freerdp-bot test

PR #6338, opened 2020-07-02
region: add a ref counter on REGION16_DATA to harden frees of empty_region

hardening:


akallabeth:
@hardening don´t really see the point in keeping a reference counter that is not used at all.
Nothing against adding one once it is possible to share region data, but until then better simply remove the data if not needed.

hardening:
@akallabeth well it's used to not free the empty region data

bmiklautz:
@hardening did you hit any issue with region?

hardening:
> @hardening did you hit any issue with region?

nope it's more an alternative to https://github.com/FreeRDP/FreeRDP/pull/6333

PR #6344, opened 2020-07-06
rail: server: mark rail_send_pdu as static function

kubistika:


kubistika:
@akallabeth no, this is a different function (used only in server code). It's confusing because they have the same name.

PR #6357, opened 2020-07-07
server: proxy: support server heartbeats

kubistika:
I noticed that `rdpRdp` struct contains a pointer to another *empty* struct called `rdpHeartbeat`. So, I just added a callback called `ServerHeartbeat` and called it whenever the client receives a server heartbeat. I also added server functionality to send an heartbeat. I'm not sure I like the way the `rdpHeartbeat` struct is currently used, because I don't think it's place is under `rdpRdp`, as only clients (freerdp* structs) really receive heartbeats (from servers). I'd maybe remove this struct from `rdpRdp` and maintain it only in `freerdp` struct. What do you guys think?

### DIFF
--- a/include/freerdp/heartbeat.h
+++ b/include/freerdp/heartbeat.h
@@ -17,23 +17,30 @@
  * limitations under the License.
  */
 
-#ifndef FREERDP_LIB_CORE_HEARTBEET_H
-#define FREERDP_LIB_CORE_HEARTBEET_H
+#ifndef FREERDP_HEARTBEAT_H
+#define FREERDP_HEARTBEAT_H
 
-#include "rdp.h"
+#include <freerdp/types.h>
 
-#include <freerdp/heartbeat.h>
-#include <freerdp/freerdp.h>
-#include <freerdp/log.h>
-#include <freerdp/api.h>
+typedef struct rdp_heartbeat rdpHeartbeat;
 
-#include <winpr/stream.h>
+typedef BOOL (*pServerHeartbeat)(freerdp* instance, BYTE period, BYTE count1, BYTE count2);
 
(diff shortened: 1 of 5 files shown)


include/freerdp/heartbeat.h
```diff
@@ -0,0 +1,46 @@
...
+{
+	pServerHeartbeat ServerHeartbeat;
+};
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+	FREERDP_API int freerdp_heartbeat_send_heartbeat_pdu(freerdp_peer* peer, BYTE period,
```
akallabeth, line 39: I´d opt for `BOOL` return as the only return is `-1` and 0 anyway

libfreerdp/core/heartbeat.c
```diff
@@ -44,6 +44,40 @@ int rdp_recv_heartbeat_packet(rdpRdp* rdp, wStream* s)
 	         "received Heartbeat PDU -> period=%" PRIu8 ", count1=%" PRIu8 ", count2=%" PRIu8 "",
 	         period, count1, count2);
 
+	if (rdp->heartbeat)
+	{
+		BOOL rc = rdp->heartbeat->ServerHeartbeat(rdp->instance, period, count1, count2);
```
akallabeth, line 49: Screams for `IFCALL` 
I think most of the time the callback will be `NULL` crashing here? Except in proxy, of course.

kubistika: You're right, thanks

include/freerdp/heartbeat.h
kubistika: K

PR #6360, opened 2020-07-07
server: proxy: rail: handshake ex flags sync

kubistika:
Also made `rail_server_handle_messages` a static method, as it should be.

### DIFF
--- a/channels/rail/server/rail_main.c
+++ b/channels/rail/server/rail_main.c
@@ -297,8 +297,6 @@ static UINT rail_send_server_handshake_ex(RailServerContext* context,
 	if (!context || !handshakeEx || !context->priv)
 		return ERROR_INVALID_PARAMETER;
 
-	/* Save channel flags to context */
-	context->priv->channelFlags = handshakeEx->railHandshakeFlags;
 	s = rail_pdu_init(RAIL_HANDSHAKE_EX_ORDER_LENGTH);
 
 	if (!s)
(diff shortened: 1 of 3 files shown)


include/freerdp/server/rail.h
```diff
@@ -140,7 +140,7 @@ extern "C"
 
 	FREERDP_API RailServerContext* rail_server_context_new(HANDLE vcm);
 	FREERDP_API void rail_server_context_free(RailServerContext* context);
-	FREERDP_API UINT rail_server_handle_messages(RailServerContext* context);
```
akallabeth, line 143: Not a good idea if you want a backport of this ;)

kubistika: Hmm okay, although I don't like that this is exported :/ it really shouldn't be, I'll create an issue for that and we'll address it in the next major?

channels/rail/server/rail_main.c
```diff
@@ -297,8 +297,6 @@ static UINT rail_send_server_handshake_ex(RailServerContext* context,
 	if (!context || !handshakeEx || !context->priv)
 		return ERROR_INVALID_PARAMETER;
 
-	/* Save channel flags to context */
-	context->priv->channelFlags = handshakeEx->railHandshakeFlags;
```
akallabeth, line 301: Maybe I´m missing something, but here we should update the flags as these are the ones the client is allowed to use?

server/proxy/pf_rail.c
```diff
@@ -66,6 +75,10 @@ static UINT pf_rail_server_handshake_ex(RailClientContext* client,
 	proxyData* pdata = (proxyData*)client->custom;
 	RailServerContext* server = (RailServerContext*)pdata->ps->rail;
 	WLog_DBG(TAG, __FUNCTION__);
+
+	/* update rail handshake flags */
+	rail_server_set_handshake_ex_flags(server, handshakeEx->railHandshakeFlags);
```
akallabeth, line 80: Remove this and add it where it was before (in `rail_send_server_handshake_ex`)

Issue #6362, opened 2020-07-08
[URBDRC] ISO transfer double free


akallabeth:
Follow up to #6337 :
@LordOfTheRats reported a double free with ISO transfers due to missing checks for duplicate `streamID`

LordOfTheRats:
Ah, thank you for creating this issue.

Here is a [trace with stacktrace](https://gist.github.com/LordOfTheRats/a1761872e485d6c6ec48c6872dc36967) showing the use-after-free.

I'm still setting up a debug environment to have a closer look.

akallabeth:
@LordOfTheRats the referenced pr should already abort if the condition causing the double free occurs.
If I´m right the server sends a duplicated `streamID` and we get into trouble because the pending transfer buffers are freed up before completion.

PR #6363, opened 2020-07-08
[URBDRC] Added return checks for replaced HashTable with ArrayList

akallabeth:
Fixes for #6362

### DIFF
--- a/channels/urbdrc/client/data_transfer.c
+++ b/channels/urbdrc/client/data_transfer.c
@@ -763,9 +763,7 @@ static void urb_isoch_transfer_cb(IUDEVICE* pdev, URBDRC_CHANNEL_CALLBACK* callb
 		Stream_Write_UINT32(out, OutputBufferSize); /** OutputBufferSize */
 		Stream_Seek(out, OutputBufferSize);
 
-		if (!pdev->isChannelClosed(pdev))
-			callback->channel->Write(callback->channel, Stream_GetPosition(out), Stream_Buffer(out),
-			                         NULL);
+		stream_write_and_free(callback->plugin, callback->channel, out);
 	}
 }
 
--- a/channels/urbdrc/client/libusb/libusb_udevice.c
+++ b/channels/urbdrc/client/libusb/libusb_udevice.c
@@ -222,7 +222,6 @@ static ASYNC_TRANSFER_USER_DATA* async_transfer_user_data_new(IUDEVICE* idev, UI
 
 static void async_transfer_user_data_free(ASYNC_TRANSFER_USER_DATA* user_data)
 {
-
 	if (user_data)
 	{
 		Stream_Free(user_data->data, TRUE);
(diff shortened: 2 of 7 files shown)


akallabeth:
@LordOfTheRats the latest commit should fix your issue, can you confirm?

LordOfTheRats:
Working on it. Restarting my tests using 666dc4a :)

akallabeth:
@LordOfTheRats did find another issue remaining, the latest should work now, did here with address sanitizers enabled.

LordOfTheRats:
On it. I can already tell there is another use-after-free when the device (webcam or headset) is unplugged while the session is running. I already had this in 9155f04.

[Trace of webcam being unplugged in 9155f04](https://gist.github.com/LordOfTheRats/3115e176810524b20226943687c71339)

I'll get a new trace showing this in 3ecfe74 as well.

LordOfTheRats:
I'm doing these tests using the debug version. I disabled data dumps since they tend to slow USB down to a crawl.

Webcam works in general.

[Use-after-free when device is unplugged](https://gist.github.com/LordOfTheRats/3a7d06bb27191c11cb975792587f1865)

[Use-after-free when hotplugging the webcam](https://gist.github.com/LordOfTheRats/1e222bc4b06b71b0c427edcbbc5835a7)
This happened only one out of five tests.

[Webcam test app reports device busy](https://gist.github.com/LordOfTheRats/1e222bc4b06b71b0c427edcbbc5835a7)
This only happened once. Unfortunately we don't have a data dump for this. Might be impractical to debug. Maybe I'll catch this with data dump on someday.


I'll repeat these tests using the headset.

LordOfTheRats:
Headset works quite well. Crash on hotplug did not happen again.

For completeness: [Use-after-free when unplugging the headset](https://gist.github.com/LordOfTheRats/98a5c1858b7cca57c1ffe41e961d4b16)

akallabeth:
@LordOfTheRats ok, found the double free issue. Should be fixed with the latest commit.

LordOfTheRats:
FreeRDP is not crashing anymore. :+1:

However, it also does not notice that the device is gone and consequently does not notify the server. Trying to access the unplugged on the server (obviously) leads to a freeze of the webcam app and FreeRDP reporting

`
Failed to find a USB device mapped to deviceId
`

Trying to log off in this state leads to FreeRDP remain with a black screen after logoff for a minute or two, until finally timing out.

[Trace](https://gist.github.com/LordOfTheRats/1ce90fa843b26d21a3b7772452082a31)

akallabeth:
@LordOfTheRats ok, finally had the time to analyze the problem with the double free properly.
The latest commit (rebased everyting on current master, so don´t worry, it is only these changes) fixes the recursive cleanup on device removal.

LordOfTheRats:
Unfortunately I'm still getting heap-use-after-free when unplugging a device in current master (2d715e). Also hotplugging stopped working again.

I will collect the usual logs and open a new issue for this.

akallabeth:
@LordOfTheRats ok, you´ re lucky today, just had (hopefully) the same crash with current nightlies during testing and have created https://github.com/FreeRDP/FreeRDP/pull/6435 as a fix.
Damn, that was not even part of the usb channel ^^

PR #6379, opened 2020-07-20
core: fix invalid inttype and reference

nfedera:
The error message in `tpdu_read_header()` printed the pointer address and used `PRIu16` for an `UINT8` type


bmiklautz:
@freerdp-bot test

Issue #6105, opened 2020-04-22
OpenIndiana version 2.0.0 build fails (2.0.0-rc4 had no problems)


3eka:
Hi all,

I was able to compile (w/o any intervention) version 2.0.0-rc4 on OpenIndiana, with GCC 9 compiler suite (only 32 bit one).
Now, I have tried 2.0.0 - andfaced with compilation problem....

**Describe the bug**
C program channels/rdpdr/client/rdpdr_main.c, line 583 contains
```
#include <mntent.h>
```
So, first failure was missing header file. On OpenIndiana it could be found with:
```
#include <sys/mntent.h>
```
Then (with that changed), compilation fails with:
```
[ 75%] Building C object channels/rdpdr/client/CMakeFiles/rdpdr-client.dir/rdpdr_main.c.o
/nobackup/SFW_build/freerdp-2.0.0/channels/rdpdr/client/rdpdr_main.c: In function 'handle_hotplug':
/nobackup/SFW_build/freerdp-2.0.0/channels/rdpdr/client/rdpdr_main.c:666:16: warning: implicit declaration of function 'getmntent' [-Wimplicit-function-declaration]
  666 |  while ((ent = getmntent(f)) != NULL)
      |                ^~~~~~~~~
/nobackup/SFW_build/freerdp-2.0.0/channels/rdpdr/client/rdpdr_main.c:666:14: warning: assignment to 'struct mntent *' from 'int' makes pointer from integer without a cast [-Wint-conversion]
  666 |  while ((ent = getmntent(f)) != NULL)
      |              ^
/nobackup/SFW_build/freerdp-2.0.0/channels/rdpdr/client/rdpdr_main.c:669:25: error: dereferencing pointer to incomplete type 'struct mntent'
  669 |   const char* path = ent->mnt_dir;
      |                         ^~
gmake[2]: *** [channels/rdpdr/client/CMakeFiles/rdpdr-client.dir/build.make:89: channels/rdpdr/client/CMakeFiles/rdpdr-client.dir/rdpdr_main.c.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:2727: channels/rdpdr/client/CMakeFiles/rdpdr-client.dir/all] Error 2
gmake: *** [Makefile:152: all] Error 2
```

**Application details**
* Version of FreeRDP
2.0.0
* Command line used
```
export CC=/usr/gcc/9/bin/gcc
export CXX=/usr/gcc/9/bin/g++
[...]

Issue #6377, opened 2020-07-16
borken build on freebsd in 9f3adf10f85f0ab51c5ed67dc2a8cb3b0d519ebb


sss123next:
borken build on freebsd in 9f3adf10f85f0ab51c5ed67dc2a8cb3b0d519ebb

<mntent.h> does not exists on freebsd, even with gnulib installed.

bmiklautz:
@sss123next thanks for the report. We don't have any *BSD builders for our CI so this one slipped through.

akallabeth:
@sss123next if you know the equivalent of getting current mount points on `FreeBSD` we could fix this easily (there was a similar report for solaris) but we lack the test environments.

sss123next:
unfortunately i don't, but i can investigate, we have freebsd people, so i think someone should know something, also here is another problem in client channels on FreeBDSM, related to libusb, i will create separate ticket later.

rozhuk-im:
```
#include <sys/mount.h>

	int mntsize;
	size_t idx;
	struct statfs *mntbuf = NULL;

	mntsize = getmntinfo(&mntbuf, MNT_NOWAIT);
	for (idx = 0; idx < (size_t)mntsize; idx ++) {
		printf("%s - %s, local = %i, f_type = %i",
		    mntbuf[idx].f_mntonname, mntbuf[idx].f_fstypename,
		    (int)(0 != (MNT_LOCAL & mntbuf[idx].f_flags)),
		    mntbuf[idx].f_type);
	}
	free(mntbuf);
```

akallabeth:
@rozhuk-im looks promising, are you able to create a pull request for that (sorry, lacking a test env here)

rozhuk-im:
@akallabeth sorry, I'm going to trip in next day and have no time, @sss123next ask me for help and will handle it.

sss123next:
ah.., me again ( ok, ok .... i will handle this...

PR #6387, opened 2020-07-23
Add support for getmntent support on Illumos

mvwieringen:
This implements reading mount entries on Solarish operating systems like
OpenIndiana etc using the somewhat different getmntent support available
there.

This fixes issue #6105

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -580,7 +580,11 @@ static DWORD WINAPI drive_hotplug_thread_func(LPVOID arg)
 
 #else
 
+#ifdef __sun
+#include <sys/mnttab.h>
+#else
 #include <mntent.h>
+#endif
 #define MAX_USB_DEVICES 100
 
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@mvwieringen thank you for this patch.
@sss123next could this patch work for freebsd as mentioned in #6377 as well?

mvwieringen:
@akallabeth Regarding FreeBSD I checked yesterday on my test VM and it doesn't have the getmntent() library call so I don't think it will fix anything there. For that you probably need to use the create a new case analog for how things are done for windows and macOS earlier in the file.

sss123next:
yes, this works on freebsd as well, at least builds, but should be checked like this `#if defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__sun)`
instead of `#ifdef __sun`

sss123next:
sorry, i was wrong, this solution does not work on freebsd, `sys/mnttab.h` is also absent here.

PR #6392, opened 2020-07-24
cliprdr_common.c: fix memory leak in certain error condition

qunying:
Under some error condition, wStream *s is leaked in cliprdr_common.c


akallabeth:
@freerdp-bot test

Issue #6349, opened 2020-07-06
wlfreerdp, fast resizing breaks the session


chakuzo:
I use a simple Qt-Wayland Compositor for testing, when I resize the wlfreerdp window the session can follow only if the resize events comming not to fast (3-5 seconds delay is need). Otherwise the screen get white and the session will run in background.

I use the nightly build.

Maybe someone has an idea how to analyze the problem more precisely. For me everything looks fine in the debug output. 

Debug output:

[13:43:02:884] [28530:28530] [DEBUG][com.freerdp.channels.disp.client] - disp_send_display_control_monitor_layout_pdu: NumMonitors=1
[13:43:02:884] [28530:28530] [DEBUG][com.freerdp.channels.disp.client] - 	0 : Flags: 0x00000001 Left/Top: (0,0) W/H=744x827)
[13:43:02:884] [28530:28530] [DEBUG][com.freerdp.channels.disp.client] - 	   PhysicalWidth: 744 PhysicalHeight: 827 Orientation: 0
[13:43:02:884] [28530:28530] [DEBUG][com.freerdp.channels.drdynvc.client] - write_data: ChannelId=10 size=56
[13:43:02:895] [28530:28530] [DEBUG][com.freerdp.core.rdp] - rdp_send_data_pdu: sending data (type=0x1f size=37 channelId=1009)
[13:43:02:895] [28530:28530] [DEBUG][com.freerdp.core.rdp] - rdp_send_data_pdu: sending data (type=0x14 size=41 channelId=1009)
[13:43:02:895] [28530:28530] [DEBUG][com.freerdp.core.rdp] - rdp_send_data_pdu: sending data (type=0x14 size=41 channelId=1009)
[13:43:02:895] [28530:28530] [DEBUG][com.freerdp.core.rdp] - rdp_send_data_pdu: sending data (type=0x27 size=41 channelId=1009)
[13:43:02:896] [28530:28530] [DEBUG][com.freerdp.core.rdp] - recv Synchronize Data PDU (0x1F), length: 22
[13:43:02:896] [28530:28530] [DEBUG][com.freerdp.core.rdp] - recv Control Data PDU (0x14), length: 26
[13:43:02:996] [28530:28530] [DEBUG][com.freerdp.core.rdp] - recv Control Data PDU (0x14), length: 26
[13:43:02:996] [28530:28530] [DEBUG][com.freerdp.core.rdp] - recv Font Map Data PDU (0x28), length: 26
[...]

PR #6394, opened 2020-07-26
Check for pending buffer

chakuzo:
I had some problems with resizing of windows. After some time of debugging I found maybe the problem. I am little bit unsure if this the correct way to handle it.

Issue: https://github.com/FreeRDP/FreeRDP/issues/6349

### DIFF
--- a/uwac/libuwac/uwac-window.c
+++ b/uwac/libuwac/uwac-window.c
@@ -121,7 +121,7 @@ static void xdg_handle_toplevel_configure(void* data, struct xdg_toplevel* xdg_t
 	event->window = window;
 	event->states = surfaceState;
 
-	if (width && height)
+	if (width && height && window->pendingBufferIdx == -1)
 	{
 		event->width = width;
 		event->height = height;


uwac/libuwac/uwac-window.c
```diff
@@ -121,7 +121,7 @@ static void xdg_handle_toplevel_configure(void* data, struct xdg_toplevel* xdg_t
 	event->window = window;
 	event->states = surfaceState;
 
-	if (width && height)
```
akallabeth, line 124: Shouldn´t this just check if the `width` and `height` actually changed?

chakuzo: I think this should check if the width and height is valid. But it is maybe an improvement to check (window->width != width || window->height != height) also. 

but the problem still exists that the request was not sucessful, the window size hangs on the state before and the compositor is not informed about it.

hardening (commented):
Looking at the current code it seems like this has been addressed and the PR could be closed ?

chakuzo:
I am not quite sure if the problem is solved. But I think the PR can be closed.

PR #6397, opened 2020-07-28
remove wrong initialization value

qunying:
Initialization value is wrong, remove it to avoid confusion.


akallabeth:
@freerdp-bot test

PR #6411, opened 2020-08-03
Http gateway message support

mfleisz:
This PR implements parsing of the consent and service messages in the http RD gateway implementation. The `PresentGatewayMessage` callback has only been called by the legacy gateway code.
Additionally the PR adds callback handlers for the Windows, Wayland and X11 clients.

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -638,6 +638,33 @@ fail:
 	}
 }
 
+static BOOL wf_present_gateway_message(freerdp* instance, UINT32 type, BOOL isDisplayMandatory,
+                                       BOOL isConsentMandatory, size_t length, const WCHAR* message)
+{
+	if (!isDisplayMandatory && !isConsentMandatory)
+		return TRUE;
+
+	/* special handling for consent messages (show modal dialog) */
+	if (type == 1 && isConsentMandatory)
+	{
+		int mbRes;
+		WCHAR* msg;
+
+		msg = wf_format_text(L"%.*s\n\nI understand and agree to the terms of this policy", length,
+		                     message);
+		mbRes = MessageBoxW(NULL, msg, L"Consent Message", MB_YESNO);
+		free(msg);
+
+		if (mbRes != IDYES)
+			return FALSE;
+	}
+	else
+		return client_cli_present_gateway_message(instance, type, isDisplayMandatory,
+		                                          isConsentMandatory, length, message);
+
+	return TRUE;
+}
+
 static DWORD WINAPI wf_input_thread(LPVOID arg)
 {
 	int status;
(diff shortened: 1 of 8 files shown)


client/Windows/wf_client.c
```diff
@@ -638,6 +638,33 @@ static DWORD wf_verify_changed_certificate_ex(freerdp* instance, const char* hos
 	}
 }
 
+BOOL wf_present_gateway_message(freerdp* instance, UINT32 type, BOOL isDisplayMandatory,
```
akallabeth, line 641: maybe `static`?

libfreerdp/core/gateway/rdg.c
```diff
@@ -669,6 +711,76 @@ static BOOL rdg_process_handshake_response(rdpRdg* rdg, wStream* s)
...
+	{
+		WCHAR* msg;
+		UINT16 msgLenBytes;
+		rdpContext* context = rdg->context;
+
+		assert(context);
+		assert(context->instance);
+
+		/* Read message string and invoke callback */
+		if (!rdg_read_http_unicode_string(s, &msg, &msgLenBytes))
```
akallabeth, line 771: Is this string really `\0` terminated? (unchecked data from server)

client/Windows/wf_client.c
mfleisz: Good point

libfreerdp/core/gateway/rdg.c
mfleisz: The documentation does not say anything about \0 termination but I check if there are at least msgLenBytes of string data available and the both parameters (buffer and buffer size) are forwarded to the callback

client/Windows/wf_client.c
```diff
@@ -638,6 +638,33 @@ static DWORD wf_verify_changed_certificate_ex(freerdp* instance, const char* hos
 	}
 }
 
+static BOOL wf_present_gateway_message(freerdp* instance, UINT32 type, BOOL isDisplayMandatory,
```
akallabeth, line 641: Just noticed, is there a documentation of what values `type` can have? An enum here would greatly improve readability (remove the magic numbers)

mfleisz: @akallabeth Yes - they are actually already there (in core/gateway/tsg.h) so I will just export them to the public API. 👍

akallabeth: @mfleisz great, then please also change the type for the callback accordingly.

akallabeth, line 641: Wrong function signature, update to enum.

client/common/client.c
```diff
@@ -663,6 +663,66 @@ DWORD client_cli_verify_changed_certificate_ex(freerdp* instance, const char* ho
 	return client_cli_accept_certificate(instance->settings);
 }
 
+BOOL client_cli_present_gateway_message(freerdp* instance, UINT32 type, BOOL isDisplayMandatory,
```
akallabeth, line 666: Wrong function signature, update to enum.

PR #6429, opened 2020-08-10
Added duplicate initialization check for plugins.

akallabeth:
Fixes the issue in #6413: 
1. Under certain conditions a channel could be initialized twice leading to double free issues.
1. Prevent this by checking the device has not been initialized yet.


akallabeth:
@mfleisz I don´t think so (at least not in a manner we can backport to 2.0)

mfleisz:
@akallabeth That is true ...

PR #6435, opened 2020-08-12
Fixed double free on channel close in channel write.

akallabeth:
The channel write function might close the channel.
Delay that close until the channel lock was lifted, otherwise this is an access of invalid data.

### DIFF
--- a/channels/drdynvc/client/drdynvc_main.c
+++ b/channels/drdynvc/client/drdynvc_main.c
@@ -30,10 +30,12 @@
 
 #define TAG CHANNELS_TAG("drdynvc.client")
 
+static UINT dvcman_close_channel(IWTSVirtualChannelManager* pChannelMgr, UINT32 ChannelId,
+                                 BOOL bSendClosePDU);
 static void dvcman_free(drdynvcPlugin* drdynvc, IWTSVirtualChannelManager* pChannelMgr);
 static void dvcman_channel_free(void* channel);
 static UINT drdynvc_write_data(drdynvcPlugin* drdynvc, UINT32 ChannelId, const BYTE* data,
-                               UINT32 dataSize);
+                               UINT32 dataSize, BOOL* close);
 static UINT drdynvc_send(drdynvcPlugin* drdynvc, wStream* s);
 
 static void dvcman_wtslistener_free(DVCMAN_LISTENER* listener)
@@ -446,6 +448,7 @@ fail:
 static UINT dvcman_write_channel(IWTSVirtualChannel* pChannel, ULONG cbSize, const BYTE* pBuffer,
                                  void* pReserved)
 {
+	BOOL close = FALSE;
 	UINT status;
 	DVCMAN_CHANNEL* channel = (DVCMAN_CHANNEL*)pChannel;
 
@@ -454,8 +457,12 @@ static UINT dvcman_write_channel(IWTSVirtualChannel* pChannel, ULONG cbSize, con
 		return CHANNEL_RC_BAD_CHANNEL;
 
 	EnterCriticalSection(&(channel->lock));
-	status = drdynvc_write_data(channel->dvcman->drdynvc, channel->channel_id, pBuffer, cbSize);
+	status =
+	    drdynvc_write_data(channel->dvcman->drdynvc, channel->channel_id, pBuffer, cbSize, &close);
 	LeaveCriticalSection(&(channel->lock));
+	/* Close delayed, it removes the channel struct */
+	if (close)
+		dvcman_close_channel(channel->dvcman->drdynvc->channel_mgr, channel->channel_id, TRUE);
 	return status;
 }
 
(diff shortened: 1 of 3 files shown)


LordOfTheRats:
Awesome! Most importantly: No more crashes!

Hotplug seems to work, although it can take a few seconds for FreeRDP to notice the new device and sometimes up to a minute after that for the server to actually see the device.

Hotremove worked well so far.


Having the device already plugged in when the session starts gets some different results, though. Quite often the webcam cannot be opened first try and tends to crash the channel. It comes back after about a minute, courtesy of hotplug I guess. 

```
[11:15:55:351] [17401:17424] [DEBUG][com.freerdp.ch[11:15:55:439] [17401:17423] [ERROR][com.freerdp.channels.drdynvc.client] - VirtualChannelWriteEx failed with CHANNEL_RC_BAD_INIT_HANDLE [00000009]                                         
[11:15:55:439] [17401:17423] [ERROR][com.freerdp.channels.urbdrc.client] - libusb_reset_device: error LIBUSB_ERROR_NO_DEVICE[-4]                                                              
```

I see USB errors in dmesg: 
```
[ 2381.472870] usb 1-4: New USB device found, idVendor=046d, idProduct=081b
[ 2381.472876] usb 1-4: New USB device strings: Mfr=0, Product=0, SerialNumber=2
[ 2381.472881] usb 1-4: SerialNumber: 34601DA0
[ 2381.473497] uvcvideo: Found UVC 1.00 device <unnamed> (046d:081b)
[ 2381.487854] uvcvideo 1-4:1.0: Entity type for entity Extension 4 was not initialized!
[ 2381.487861] uvcvideo 1-4:1.0: Entity type for entity Extension 6 was not initialized!
[ 2381.487866] uvcvideo 1-4:1.0: Entity type for entity Extension 7 was not initialized!
[ 2381.487870] uvcvideo 1-4:1.0: Entity type for entity Processing 2 was not initialized!
[ 2381.487875] uvcvideo 1-4:1.0: Entity type for entity Extension 3 was not initialized!
[ 2381.487879] uvcvideo 1-4:1.0: Entity type for entity Camera 1 was not initialized!
[ 2381.488328] input: UVC Camera (046d:081b) as /devices/pci0000:00/0000:00:14.0/usb1/1-4/1-4:1.0/input/input104
[ 2381.492188] usb 1-4: 3:1: cannot get freq at ep 0x86
[ 2381.492788] usb 1-4: 3:2: cannot set freq 24000 to ep 0x86
[ 2381.493331] usb 1-4: 3:3: cannot set freq 32000 to ep 0x86
[ 2381.493862] usb 1-4: 3:4: cannot set freq 48000 to ep 0x86
[ 2381.494710] usb 1-4: 5:2: cannot get min/max values for control 2 (id 5)
[ 2381.496131] usb 1-4: USB disconnect, device number 81
```

Maybe uvcvideo is interfering, but I also see the set freq errors without FreeRDP running, so I will need to check my setup first for unrelated USB hardware problems before continuing testing.


I have two traces in case it helps:
coldplugged, webcam vanished on open, then came back: https://gist.github.com/LordOfTheRats/20bf8b3dca4174f1969b2cb66e53b72c
coldplugged, took 3 tries to open the webcam: https://gist.github.com/LordOfTheRats/d52abc20752e4ac0d8226499ec4439aa

LordOfTheRats:
6ce0517 actually works even better! Hotplug feels much quicker, and so far no more failed to open device when coldplugged, no more vanishing.

I still see USB errors in dmesg, but they don't seem to make a difference.

Will test using a different setup to verify anyway.

-----

Alright, one more crash: https://gist.github.com/LordOfTheRats/6b2e456226133739d0170a998dfa691f
I was mean, plugged and then immediately pulled the webcam just as FreeRDP started to accessing it. :>

LordOfTheRats:
Switched from Server 2012R2 to Server 2016.

Got a heap buffer overflow when opening the webcam: https://gist.github.com/LordOfTheRats/811297efd1d9feb2a48d49d656f3d1a3

I tried to catch this with data dump enabled. Instead of a crash I just get device busy on the server side (I'd guess a timing issue because of the dump).

akallabeth:
@LordOfTheRats ok, that one was new. Fixed with latest commit.

LordOfTheRats:
Sorry I somehow missed your last comment. I will test the last commit immediately.

Just a thought: Would it help if we provided you with a Logitech C310 webcam? That's what I (arbitrarily) use for testing. We could ship one for you to keep.

LordOfTheRats:
Alright. So far I had no more crashes and the redirection has been quite reliable for the most part. Also so far I was not able to provoke a crash by being mean :>

There is a hickup on hotplug. Often it takes about a minute for the device to show up on the server. See https://gist.github.com/LordOfTheRats/3cda2aed5d6368653c8bb91cc68dbc3b#file-freerdp-usb-trace-hotplug4-L1366 at [16:07:09:606] (line 1366). 
The log is as seen on the terminal, including the out of band log line at the end of line 1361.

There seems to be exactly 60seconds between line 1366 and the out of band print in line 1361. I suppose the first attempt accessing the device on hotplug gets stuck somehow and runs into a 60s timeout, after which FreeRDP successfully tries to access the device again and continues on.

mfleisz (approved):

PR #6324, opened 2020-06-29
API to offload io operations to external implementation

sss123next:

PR #6448, opened 2020-08-25
Implemented switchable transport layer

akallabeth:
Superseeds #6324

### DIFF
--- a/include/freerdp/freerdp.h
+++ b/include/freerdp/freerdp.h
@@ -31,7 +31,6 @@ typedef struct rdp_graphics rdpGraphics;
 typedef struct rdp_metrics rdpMetrics;
 typedef struct rdp_codecs rdpCodecs;
 typedef struct rdp_transport rdpTransport; /* Opaque */
-typedef struct rdp_transport_io rdpTransportIo;
 
 typedef struct rdp_freerdp freerdp;
 typedef struct rdp_context rdpContext;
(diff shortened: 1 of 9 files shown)


akallabeth:
@sss123next I´ve cleaned up your pr a bit and implemented something that passes the usual connection tests.
I´ve removed the proxy callback as that is not required (just ensure that `FreeRDP_ProxyType == PROXY_TYPE_IGNORE` in settings when overriding the backend)

This is a first draft, still need to implement the chain to allow setting the backend dynamically (usually more than 1 transport is in use, so add some cache in the upper layers to set that on creation)

sss123next:
do you need any assistance from me ?, it is blocker right now in my project, so i can work on full time  for this task.

sss123next:
you dropped `DataHandler` which is essential for async io implementation.

akallabeth:
@sss123next I´ve added `transport_parse_pdu` to check if a whole PDU is available to the exported functions.
This way you can async read a whole PDU and notify the freerdp main loop if it is available (which then reads it with your own `ReadBytes` function)

sss123next:
yes something like this, thx.
also ii cannot push to your branch.
```
ANG= git push akallabeth-repo external_io
ERROR: Permission to akallabeth/FreeRDP.git denied to sss123next.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

i have dropped connect/accept nla/rdp, also i have eliminated tls code duplication which is removed requirement to hook connect/accept nla

but i cannot push it ...

PR #6455, opened 2020-09-01
cleanup of https://github.com/FreeRDP/FreeRDP/pull/6448

sss123next:
* eliminated code duplication, so it's not required to export rdp protocol specific api
* dropped some protocol related functions which is not needed for external io backend.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #6462, opened 2020-09-07
Scale cursor so it matches the scale of everything else

nloewen:
Makes use of the percent scale setting to scale the cursor in the X11 client. This ensures that the cursor scale matches the scale of the rest of the RDP session. In my particular use case, the cursor is not tiny when viewing a 1080p session on a 4k monitor.

### DIFF
--- a/client/X11/xf_graphics.c
+++ b/client/X11/xf_graphics.c
@@ -235,8 +235,16 @@ static BOOL xf_Pointer_New(rdpContext* context, rdpPointer* pointer)
 	XcursorImage ci;
 	xfContext* xfc = (xfContext*)context;
 	xfPointer* xpointer = (xfPointer*)pointer;
+	rdpSettings* settings;
+	XcursorPixel* pixelBuffer;
+	size_t bufferSize;
 
 	if (!context || !pointer || !context->gdi)
+		return FALSE;
+
+	settings = xfc->context.settings;
+
+	if (!settings)
 		return FALSE;
 
 	if (!xfc->invert)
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@nloewen I think you misunderstood something there. The settings you check have nothing to do with scaling, they just set the current resolution. `smart-sizing` is the only option that enforces local scaling, the other options activate server side scaling, which already adjusts the cursor sizes.

nloewen:
The way I've scaled it, the cursor is always an appropriate size for the monitors native resolution, when using percent scale.
I can change it to use `settings->DesktopWidth / (double)xfc->scaledWidth` as the scale, which would have the benefit of matching the scale of the remote content, but the cursor would have to be resized any time the window was resized. That seemed like it would add a lot of complexity, and the simpler solution seemed like an improvement over no scaling at all.

nloewen:
I updated my branch to use the window scale. It generates new cursor sizes as needed, and selects the correct size for the current scale when setting the cursor. The scaled cursors are cached since I figure it's common to go between a couple different window sizes. Unscaled windowed, and fullscreen for example. I suppose the latest commit isn't showing up here since the PR is closed already. I can create a new PR if needed.

akallabeth:
@nloewen there is no percent scale, you use a convenience setting `settings->PercentScreenUseWidth` et al, which has nothing to do with scaling, as your factor.
Check `settings->SmartSizing` for local scaling enabled and compare local and remote resolutions to get a factor.

akallabeth:
`DesktopScaleFactor` and `DeviceScaleFactor` are the settings enabling high dpi support server side (if the server supports that)

PR #6467, opened 2020-09-07
Fixes #5747: Only activate sound output device

### LINKED ISSUE #5747: Explicitly set output device in rdpsnd Mac backend
A Royal TSX customer recently alerted me of the fact that [Micro Snitch](https://www.obdev.at/products/microsnitch), a microphone monitoring tool for macOS reported that his microphone was active when playing (NOT recording) audio in an RDP session. That happens even when the audin channel is not even loaded so it's definitely a false positive.

I did some debugging and was able to reproduce the behavior. I then found the culprit in [this commit](https://github.com/FreeRDP/FreeRDP/commit/52af56d956e4e725f62030b6c990683158de5733#diff-81bf742aadc5d6ed432a42c812d1af2fL157).

The problem basically is this: If the output device is NOT explicitly set, AVAudioEngine activates both, input and output channels leading to the behavior that CoreAudio sends notifications of the activation of both devices, even if no audio is recorded through the input channel.

A demo project demonstrating the issue is [available here](https://github.com/lemonmojo/AVTest).

The fix is quite easy: Bring back the code that explicitly sets the output device to the default audio output device. Code for this can be found either in the FreeRDP commit I mentioned above or my [sample project](https://github.com/lemonmojo/AVTest/blob/master/AVTest/AppDelegate.m#L171).

lemonmojo:
Fixes #5747 by only activating the sound output device instead of both, input and output devices.
This works around the problem that some microphone monitoring tools (like "Micro Snitch") were activated by just playing a sound inside a remote session.
This also fixes the problem that starting with macOS 10.15 sometimes the audio device wasn't ready when sound was starting to play. The fix is to check if the audio engine is running, and if it isn't, starting it inside `rdpsnd_mac_start`.

### DIFF
--- a/channels/rdpsnd/client/mac/rdpsnd_mac.m
+++ b/channels/rdpsnd/client/mac/rdpsnd_mac.m
@@ -167,22 +167,14 @@ static BOOL rdpsnd_mac_open(rdpsndDevicePlugin *device, const AUDIO_FORMAT *form
 	if (!mac->engine)
 		return FALSE;
 
-	if (@available(macOS 10.15, *))
+	err = AudioUnitSetProperty(mac->engine.outputNode.audioUnit,
+							   kAudioOutputUnitProperty_CurrentDevice, kAudioUnitScope_Global,
+							   0, &outputDeviceID, sizeof(outputDeviceID));
+	if (err)
 	{
-		/* Setting the output audio device on 10.15 or later breaks sound playback. Do not set for
-		 * now until we find a proper fix for #5747 */
-	}
-	else
-	{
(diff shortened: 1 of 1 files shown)


akallabeth (dismissed):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@lemonmojo could you run `clang-format` on your patch?

lemonmojo:
@akallabeth done!

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #6482, opened 2020-09-18
memory grows over time in wlfreerdp with /gfx:AVC444 fixed.

xsumirx:
- When running wlfreerdp with /gfx:AVC444 parameter. Resident size of the FreeRDP process gradually grows overtime and never shrinks even if the video paused or stoped.

- Sometimes wlfreerdp crashes when there are a lot of movement on the screen (like, dragging window fast in windows).

**Compilation**
`cmake -GNinja  -DWITH_VAAPI=ON -DWITH_GFX_H264=ON -DWITH_FFMPEG=ON -DWITH_DSP_FFMPEG=ON  WITH_SSE2=ON -DWITH_CAIRO=ON -DWITH_OPENH264=OFF -DCHANNEL_URBDRC=ON -DWITH_CUPS=ON -DWITH_PULSE=ON -DWITH_FAAC=OFF -DWITH_FAAD2=OFF -DWITH_GSM=OFF -DWITH_WAYLAND=ON -DCMAKE_INSTALL_PREFIX:PATH=/usr/ -DCMAKE_BUILD_TYPE=Release .`

**Run Command**
`/wlfreerdp /v:<ip> /u:<username> /p:<password> /relax-order-checks +glyph-cache /cert-ignore /audio-mode:0 +mic /size:1920x1080 /network:LAN /gfx:avc444 `

### DIFF
--- a/client/Wayland/wlfreerdp.c
+++ b/client/Wayland/wlfreerdp.c
@@ -44,22 +44,6 @@
 
 #define TAG CLIENT_TAG("wayland")
 
-static BOOL wl_begin_paint(rdpContext* context)
-{
-	rdpGdi* gdi;
-
-	if (!context || !context->gdi)
-		return FALSE;
-
-	gdi = context->gdi;
-
-	if (!gdi->primary)
-		return FALSE;
-
-	gdi->primary->hdc->hwnd->invalid->null = TRUE;
-	return TRUE;
-}
-
 static BOOL wl_update_buffer(wlfContext* context_w, INT32 ix, INT32 iy, INT32 iw, INT32 ih)
 {
 	BOOL res = FALSE;
@@ -148,7 +132,13 @@ static BOOL wl_end_paint(rdpContext* context)
 	w = gdi->primary->hdc->hwnd->invalid->w;
 	h = gdi->primary->hdc->hwnd->invalid->h;
 	context_w = (wlfContext*)context;
-	return wl_update_buffer(context_w, x, y, w, h);
+	if(!wl_update_buffer(context_w, x, y, w, h)){
+		return FALSE;
+	}
+
+	gdi->primary->hdc->hwnd->invalid->null = TRUE;
+	gdi->primary->hdc->hwnd->ninvalid = 0;
+	return TRUE;
 }
 
 static BOOL wl_refresh_display(wlfContext* context)
@@ -267,7 +257,6 @@ static BOOL wl_post_connect(freerdp* instance)
 	UwacWindowSetFullscreenState(window, NULL, instance->context->settings->Fullscreen);
 	UwacWindowSetTitle(window, title);
 	UwacWindowSetOpaqueRegion(context->window, 0, 0, w, h);
-	instance->update->BeginPaint = wl_begin_paint;
 	instance->update->EndPaint = wl_end_paint;
 	instance->update->DesktopResize = wl_resize_display;
 	freerdp_keyboard_init(instance->context->settings->KeyboardLayout);
--- a/uwac/libuwac/uwac-window.c
+++ b/uwac/libuwac/uwac-window.c
@@ -121,7 +121,8 @@ static void xdg_handle_toplevel_configure(void* data, struct xdg_toplevel* xdg_t
 	event->window = window;
 	event->states = surfaceState;
 
-	if (width && height)
+	if ((width > 0 && height > 0) &&
+		(width != window->width ||  height != window->height))
 	{
 		event->width = width;
 		event->height = height;


akallabeth (changes requested):
Thank you for this pr,
But could you please remove the wlf_gfx.c implementation? It just duplicates the default fallback.
This is reserved for hardware accelerated backends (gl, vulkan, ximage)

xsumirx:
> Thank you for this pr,
> But could you please remove the wlf_gfx.c implementation? It just duplicates the default fallback.
> This is reserved for hardware accelerated backends (gl, vulkan, ximage)

Sure.

hardening:
Could you clean the log history so that we would have only 1 commit ?

xsumirx:
> Could you clean the log history so that we would have only 1 commit ?

Do you want me to squash both the commits into one commit ? Is that all ?

hardening:
> > Could you clean the log history so that we would have only 1 commit ?
> 
> Do you want me to squash both the commits into one commit ? Is that all ?

Yeah sure

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

akallabeth:
@xsumirx thank you for the pr. Just a question to understand: with this changes there is no more leak or it is just hidden?
I can not find a place where there are buffers handled differently here than before (just most of the events filtered)

xsumirx:
> @xsumirx thank you for the pr. Just a question to understand: with this changes there is no more leak or it is just hidden?
> I can not find a place where there are buffers handled differently here than before (just most of the events filtered)

@akallabeth From what I have observed after making those changes, that the resident size of the process doesn't rises beyond a certain memory size (105MB in my case), but earlier before these changes,  a very long test of about 9-10 hours of continues video play can result the resident size of above 800MB in ram, which was not normal.

As far as my narrow understanding about the internals of FreeRDP goes, It seems that, marking invalid flag of primary gdi in begin_paint function even before actually rendering the frame, somehow leads FreeRDP to allocate more memory for frames overtimes.

The other change in uwac-window.c is a very crucial check, I have encountered that in many instances, I got negative numbers for either width or heigh. which lead to whole process crash.

akallabeth:
@xsumirx that sounds strange. many of our clients use the same pattern with `begin` and `end_paint`, so I doubt that has something to do with your ram usage.
the second fix with the `width>0` check on the other hand seems legit.

xsumirx:
If you can run some test with above command, and watch over res size in ps command. I sure will see the difference. You have to run the test for minimum of 2 hours.

akallabeth:
@xsumirx ok, please run that. I´ve run a test last month with a custom client that utilizes the same logic, test ran for a week with automated tests and the memory usage did not increase.
So, maybe we have some other issue with the wayland client hidden.

xsumirx:
> @xsumirx ok, please run that. I´ve run a test last month with a custom client that utilizes the same logic, test ran for a week with automated tests and the memory usage did not increase.
> So, maybe we have some other issue with the wayland client hidden.

Ok. I don't think, I am gonna run anymore test now. Because I already did so very rigorously. I am not sure why you haven't got this issue with Wayland considering that you had run the test with the same parameter as I had mentioned above and played fullscreen video consistently for whole test time.

Let me know, If this pr not relevant from your understanding and not suitable to be proceeded further. I will close it.

akallabeth:
@xsumirx no, your pr is relevant ;)
What I did was not a test with the wayland client, but a custom one.
FreeRDP uses the same core routines to do most of the stuff and only what you do in the functions for the client differs, so if you have a memory leak with the `begin_paint` and `end_paint` cycle, it must be in all clients or there is something else wrong with the client.
That is why I´m curious. Being content with hiding bugs never is a good strategy in the long run. (that is the reason for the inquiry)

xsumirx:
> @xsumirx no, your pr is relevant ;)
> 
> What I did was not a test with the wayland client, but a custom one.
> 
> FreeRDP uses the same core routines to do most of the stuff and only what you do in the functions for the client differs, so if you have a memory leak with the `begin_paint` and `end_paint` cycle, it must be in all clients or there is something else wrong with the client.
> 
> That is why I´m curious. Being content with hiding bugs never is a good strategy in the long run. (that is the reason for the inquiry)

That sounds good.

So, Can you guide me how can i identify the exact cause of the issue. 

Any, strategy that i need to follow ?

akallabeth:
@xsumirx No easy strategy I can think of now. I´ll check the wayland client as soon as I have time for some tests. Sorry to keep you waiting :/

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #6484, opened 2020-09-21
improve proxy logs

kubistika:


libfreerdp/core/server.c
```diff
@@ -670,6 +670,28 @@ const char* WTSChannelGetName(freerdp_peer* client, UINT16 channel_id)
...
+	UINT32 index;
+
+	if (!client || !client->context || !count)
+		return NULL;
+
+	mcs = client->context->rdp->mcs;
+
+	*count = mcs->channelCount;
+
+	names = (char**)malloc(sizeof(char*) * mcs->channelCount);
```
akallabeth, line 686: Better use `calloc` to have initialized memory, add `NULL` checks too.

akallabeth: you can use a slightly larger buffer to keep the whole thing allocatable by `malloc/free`:
1. Allocate `(sizeof(char*) + strlen() + 1) * mcs->channelCount  ` bytes
1. the first `sizeof(char*) * mcs->channelCount` are your pointers
1. the rest is your data buffer containing the data

akallabeth: Another way (if you don´t need a copy), just allocate the pointers, remove the `strdup`

akallabeth (changes requested):
Idea is sound, some code cleanups necessary.

akallabeth (approved):
LGTM

PR #6500, opened 2020-10-05
core/nla: fix order of operations

peterh:
The order of evaluation of the two sides of addition is undefined in C.
Since there is no sequence point between ber_write_contextual_tag and
ber_write_octet_string, these two functions can be called in any order.

Force the correct order by breaking the two function calls into two
separate statements.

### DIFF
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -159,9 +159,14 @@ static void nla_identity_free(SEC_WINNT_AUTH_IDENTITY* identity);
 
 #define ber_sizeof_sequence_octet_string(length) \
 	ber_sizeof_contextual_tag(ber_sizeof_octet_string(length)) + ber_sizeof_octet_string(length)
-#define ber_write_sequence_octet_string(stream, context, value, length)                \
-	ber_write_contextual_tag(stream, context, ber_sizeof_octet_string(length), TRUE) + \
-	    ber_write_octet_string(stream, value, length)
+
+static INLINE size_t ber_write_sequence_octet_string(wStream* stream, BYTE context,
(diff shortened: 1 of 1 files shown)


libfreerdp/core/nla.c
```diff
@@ -159,9 +159,14 @@ static void nla_identity_free(SEC_WINNT_AUTH_IDENTITY* identity);
 
 #define ber_sizeof_sequence_octet_string(length) \
 	ber_sizeof_contextual_tag(ber_sizeof_octet_string(length)) + ber_sizeof_octet_string(length)
-#define ber_write_sequence_octet_string(stream, context, value, length)                \
-	ber_write_contextual_tag(stream, context, ber_sizeof_octet_string(length), TRUE) + \
-	    ber_write_octet_string(stream, value, length)
+
+static inline size_t ber_write_sequence_octet_string(wStream* stream, BYTE context, void* value,
```
akallabeth, line 163: Please use the `INLINE` macro to support older compilers

akallabeth: `ber_write_octet_string` uses `const BYTE*` for `value`, please also use this here.

peterh: Thanks for the review. Changes made.

peterh:
For reference, this fixes NLA login when compiled with MSVC 2019 with optimizations enabled. Tested with both O2 and LTCG (aka LTO) enabled. I don't know which of O2, LTCG, or both is necessary to reproduce the issue; I didn't try them separately.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM
thank you for the fix

PR #6501, opened 2020-10-06
server: add WTSVirtualChannelManagerCheckFileDescriptorEx API

kubistika:
This PR makes it possible to turn off drdynvc on server side, but still allow static channels.
Also added missing `extern "C"` for rdpgfx public API.

### DIFF
--- a/include/freerdp/channels/wtsvc.h
+++ b/include/freerdp/channels/wtsvc.h
@@ -59,6 +59,7 @@ extern "C"
 	FREERDP_API void WTSVirtualChannelManagerGetFileDescriptor(HANDLE hServer, void** fds,
 	                                                           int* fds_count);
 	FREERDP_API BOOL WTSVirtualChannelManagerCheckFileDescriptor(HANDLE hServer);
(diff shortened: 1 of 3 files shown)


akallabeth (commented):
The RdpGFX changes look sound,
but I don´t like the WTS API changes.
@retro @mfleisz @bmiklautz you know more about that, what do you think?

kubistika:
can you elaborate why you don't like this "change"? it's not changing anything (the API was not changed, just extended).
I need this because I want a way to disable drdynvc in server mode (sounds to me like a reasonable thing to do).

bmiklautz:
@kubistika @akallabeth does look ok for me as it's a only new call and the API is still compatible.

akallabeth:
@kubistika I´m just cautious here. WTS api was source for many headaches before, so I might have stated my suspicion badly, I just want feedback that the change does nothing unwanted.

kubistika:
@kubistika got you, no problem. @bmiklautz already approved but lets wait for other opinions if you want :)

kubistika:
@mfleisz can you please review the changes?

@akallabeth

mfleisz (approved):

PR #6509, opened 2020-10-12
Fixed parsing of FastGlyph order.

akallabeth:
Fixed parsing of fastglyph according to
[MS-RDPEGDI] 2.2.2.2.1.2.6.1 Cache Glyph Data - Revision 2 (TS_CACHE_GLYPH_DATA_REV2)

### DIFF
--- a/libfreerdp/core/orders.c
+++ b/libfreerdp/core/orders.c
@@ -1061,76 +1061,94 @@ static INLINE BOOL update_read_delta_points(wStream* s, DELTA_POINT* points, int
 
 	return TRUE;
 }
-#define ORDER_FIELD_BYTE(NO, TARGET)                        \
-	do                                                      \
-	{                                                       \
-		if (orderInfo->fieldFlags & (1 << (NO - 1)))        \
-		{                                                   \
-			if (Stream_GetRemainingLength(s) < 1)           \
-			{                                               \
-				WLog_ERR(TAG, "error reading %s", #TARGET); \
-				return FALSE;                               \
-			}                                               \
-			Stream_Read_UINT8(s, TARGET);                   \
-		}                                                   \
-	} while (0)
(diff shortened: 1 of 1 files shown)


libfreerdp/core/orders.c
```diff
@@ -1153,11 +1171,12 @@ static INLINE BOOL FIELD_SKIP_BUFFER16(wStream* s, UINT32 TARGET_LEN)
...
-	ORDER_FIELD_COORD(3, dstblt->nWidth);
-	ORDER_FIELD_COORD(4, dstblt->nHeight);
-	ORDER_FIELD_BYTE(5, dstblt->bRop);
+	if (!read_order_field_coord(orderInfo, s, 1, &dstblt->nLeftRect, FALSE) ||
+	    !read_order_field_coord(orderInfo, s, 2, &dstblt->nTopRect, FALSE) ||
+	    !read_order_field_coord(orderInfo, s, 3, &dstblt->nWidth, FALSE) ||
+	    !read_order_field_coord(orderInfo, s, 4, &dstblt->nHeight, FALSE) ||
+	    !read_order_field_byte(orderInfo, s, 5, &dstblt->bRop, TRUE))
+		return FALSE;
 	return TRUE;
```
hardening, line 1180: So I guess that could be replaced by 
```C
return read_order_field_coord(orderInfo, s, 1, &dstblt->nLeftRect, FALSE) &&
 read_order_field_coord(orderInfo, s, 2, &dstblt->nTopRect, FALSE) &&
...
```

akallabeth: true, but I prefer to be able to set a breakpoint.

mfleisz: I think @hardening's main point is that `!a || !b || ... || !z` can be replaced with `a && b && ... && z` which makes the statement clearer

mfleisz (approved):

PR #6517, opened 2020-10-17
Re-add cursor updates when using RemoteApp and add outlining for XOR cursors in place of the previous black/white alternation

troyrollo:
At some point in the past cursor updates when using RemoteApp got broken. This change adds them back in. It also generates outlined cursors where part of the Windows cursor has an XOR (inversion) portion. This is in place of the existing checkerboarding effect which makes the cursor harder to see.

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -944,6 +944,22 @@ static int _xf_error_handler(Display* d, XErrorEvent* ev)
 	return xf_error_handler(d, ev);
 }
 
+static int xf_error_printer(Display* d, XErrorEvent *ev)
+{
+	char buf[256];
+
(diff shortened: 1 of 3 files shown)


akallabeth (changes requested):
The pr looks promising, but a few things are wired:
The `freerdp_image_inverted_pointer_color` did paint the original pointer with inverted colors (black/white) so it is visible with all background colors.
Your `freerdp_add_cursor_shadow` uses another approach, but does basically the same. Why not just place that code in `freerdp_image_inverted_pointer_color`? 
Also, you might want to make that optional as it is not desirable for all use cases (add a CMake option to enable it instead of the `inverted` painting mode maybe?

akallabeth:
Ok, fixed the window bug with #6642 
the questions regarding changes to the pointer have not been answered, so closing.

PR #6539, opened 2020-10-29
Cleanup internal channel management

mfleisz:
This PR gets rid of some unneeded/unused fields and functions in the channel handling code. It also makes it possible to call `VirtualChannelWrite` from any thread like the Windows API allows. The only restriction is that the channel must be initialized (`VirtualChannelInit`) from the same thread that called `freerdp_connect`.


akallabeth (approved):
LGTM

PR #6548, opened 2020-11-03
core: Fix race when deleting channel manager

mfleisz:
This PR fixes a possible crash when the channel manager was freed and there were pending messages in the message queue.
The problem was that even though the message queue already received the WMQ_QUIT message, it was still possible to enqueue messages after this point. This resulted in unprocessed messages in the queue when it was deleted. The delete handler then called into channel handlers which where aleady freed/deleted.
With this PR adding messages after WMQ_QUIT was posted to the message queue returns an error and all channel messages are now processed before the channels are closed/terminated.


akallabeth (approved):
LGTM

PR #6550, opened 2020-11-05
Bugfix: when `freerdp_assistance_parse_file_buffer` was called from external code, it did not copy the access password to the `rdpAssistanceFile` struct. 

odednoam:
`freerdp_assistance_parse_file_buffer` may be called directly, not necessarily from `freerdp_assistance_parse_file`, so password should be saved to the `rdpAssistanceFile` in `freerdp_assistance_parse_file_buffer`.


akallabeth (commented):
Please also move the `free(file->password)` call accordingly.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #6578, opened 2020-11-13
Fixed display channel anounce race

akallabeth:
* The display resolution change message was prone to a race condition
* Check for actual fullscreen state instead of settings
* Assume 75dpi for display resolution to mm conversion


hardening (approved):
Hum good catch for the `lastSentDate`.

PR #6584, opened 2020-11-17
Updated settings access

akallabeth:


akallabeth:
@freerdp-bot test

hardening:
hum not sure I really like the new settings style

hardening:
Especially, using getters in the internal code of FreeRDP really makes the code horrible.

akallabeth:
@hardening well, problem is that we are lacking a central point to manage memory and have no (easy) way to add additional checks on settings.
The opaque part can be disabled for the time being, but direct access is always painful (you need the same allocators as the client, ...)

client/Mac/MRDPView.m
```diff
@@ -1339,8 +1346,9 @@ BOOL mac_desktop_resize(rdpContext *context)
 	CGContextRef old_context = view->bitmap_context;
 	view->bitmap_context = NULL;
 	CGContextRelease(old_context);
-	mfc->width = settings->DesktopWidth;
-	mfc->height = settings->DesktopHeight;
+	mfc->width = freerdp_settings_get_uint32(settings, FreeRDP_DesktopWidth);
+	mfc->height = freerdp_settings_get_uint32(settings, FreeRDP_DesktopHeight);
+	;
```
kubistika, line 1351: this should be removed

kubistika (commented):
sorry, mistake

PR #6587, opened 2020-11-18
Added const to function arguments

akallabeth:
Do some refactoring, make everything possible `const`, add some return values and checks
Split from #6584 as that one is getting too big on its own.


channels/drdynvc/client/drdynvc_main.c
```diff
@@ -1507,9 +1506,10 @@ static UINT drdynvc_virtual_channel_event_connected(drdynvcPlugin* drdynvc, LPVO
 
 	settings = (rdpSettings*)drdynvc->channelEntryPoints.pExtendedData;
 
-	for (index = 0; index < settings->DynamicChannelCount; index++)
+	for (index = 0; index < freerdp_settings_get_uint32(settings, FreeRDP_DynamicChannelCount);
+	     index++)
```
hardening, line 1510: `freerdp_settings_get_uint32(settings, FreeRDP_DynamicChannelCount);` should be part of another PR

channels/rdp2tcp/client/rdp2tcp_main.c
```diff
@@ -84,9 +84,9 @@ static int init_external_addin(Plugin* plugin)
 	}
 
 	// Execute plugin
-	if (!CreateProcess(NULL,
-	                   plugin->channelEntryPoints.pExtendedData, // command line
-	                   NULL,                                     // process security attributes
+	if (!CreateProcess(plugin->channelEntryPoints.pExtendedData,
```
mfleisz, line 87: Not an expert with this code but is this switch intended? With the new code it is not possible to provide a command line with arguments to CreateProcess (as was before)

akallabeth: https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa

Both are valid and the first argument has the correct const parameters we require.

mfleisz: Just tested on Win10 20H2
```
	WCHAR buf[] = { L"notepad.exe test.txt" };
	auto res = CreateProcessW(buf, nullptr, nullptr, nullptr, FALSE, 0, nullptr, nullptr, &suInfo, &procInfo);
```
Result: `FALSE` (GetLastError: `ERROR_FILE_NOT_FOUND`)

```
	WCHAR buf[] = { L"notepad.exe test.txt" };
	auto res = CreateProcessW(nullptr, buf, nullptr, nullptr, FALSE, 0, nullptr, nullptr, &suInfo, &procInfo);
```
Result: `TRUE` - notepad is started with test.txt as an argument

mfleisz (commented):
The PR looks good but I would revert one change:

The `data` parameter (and `pExtendedData` in the channel entry points structs) in the `freerdp_channels_client_load` (and similar) functions should not be const. This is an opaque user data pointer and assuming it points to `const` data is not correct imo.

akallabeth:
@mfleisz ok, done. rebased to master and reverted that part.

hardening:
I feel like there's still some pieces of #6584 in misc places, it's not only a `make every possible const adjustments` PR.

akallabeth:
@bmiklautz @hardening @mfleisz anyone time to review this one?

libfreerdp/core/certificate.c
```diff
@@ -671,7 +671,7 @@ BOOL certificate_read_server_certificate(rdpCertificate* certificate, BYTE* serv
 	if (length < 4) /* NULL certificate is not an error see #1795 */
 		return TRUE;
 
-	s = Stream_New(server_cert, length);
+	s = Stream_New((BYTE*)server_cert, length);
```
hardening, line 674: Not your code, but I guess we could use `Stream_Static_Init` and avoid the the `Stream_Free`in `certificate_read_server_proprietary_certificate` and `certificate_read_server_x509_certificate_chain`.
That could be part of another PR.

hardening (approved):
LGTM, lots of nice changes.
Anyway I'm wondering if we really want the plugin args to be const, and so no possibility to change them.

akallabeth:
@hardening thx. Changing the arguments after having passed them in is not really required as we only parse it and then work with the extracted data. Removing sensitive data is only required for the `main` function arguments as the command line is visible from process data.

PR #6590, opened 2020-11-19
Added data_pdu_type_to_string

akallabeth:


hardening:
I feel like some calls should be `Stream_EnsureRemainingCapacity` instead of `Stream_GetRemainingCapacity`

akallabeth:
@hardening both are possible, because currently the stream is preallocated. Size is written to `tpkt` header so resizing the stream afterward might break the client

mfleisz (approved):
LGTM

PR #6596, opened 2020-11-23
Create codeql-analysis.yml

akallabeth:


kubistika:
Great work 😀

bmiklautz:
@akallabeth looks good overall but I think this might generate some noise once it's really active and the first results are detected.

bmiklautz (approved):

PR #6607, opened 2020-11-26
Sound fix buffering

akallabeth:
1. Some `CMake` warnings for excessive debug flags (performance and security issues might arise if enabled)
1. Disable sound buffer overrun checks for all but known formats with issues.


nfedera (approved):
lgtm

PR #6609, opened 2020-11-26
Fixed sending of clipboard format request response

akallabeth:
If a request was pending and new formats were announced the response
to a ServerFormatDataRequest was never sent.


mfleisz (approved):
LGTM

Issue #6592, opened 2020-11-20
RFC: improving packet forging


hardening:
I feel like in quite a lot of places we creates packets by doing this kind of job:
```c
function layer4() {
  wStream *payload = createPayloadLayer5(...)
  size_t headerSize = computeHeaderSz(payload)

  wStream *fullPacket = Stream_New(NULL, headerSize + Stream_GetPosition(payload));

  writeHeaders(fullPacket, payload);
  Stream_Write(fullPacket, Stream_GetBuffer(payload), Stream_GetPosition(payload));
return fullPacket;
}
```

And it happens that the caller of ̀`layer4` will do the same kind of job when prepending the layer3 and so on, we end up with tons of memory copies of growing chunks.

So I'm thinking of 2 solutions (not exclusive):

1. we could do it the `libevent` way: layer 1 would create a chunkList object, would reserve a chunk to store its headers, would pass the chunkList to layer 2 that would reserve a chunk for layer 2 headers, and so on, with the last layer that would reserve a chunk for the effective payload. When returning from layer2 treatment, we would have a chunk list that we would "linearize" (assemble all the small chunks) to send it over the transport;
1. another solution is to allocate a stream with some extra space, and start working on the stream after that headers reserved space (we proceed that way in some places). In this solution we need some careful checks to be sure that we don't overflow that reserved space;

PR #6614, opened 2020-11-29
freerdp-utils: add the chunkstream component

hardening:
The idea of a ChunkStream is to accumulate chunks of data, and to do the assembly at the last moment when emitting the packet. Doing so we avoid multiple copies of memory blocks (especially with packets that have variable length headers).

May be a good start to implement solution 1 of #6592.

### DIFF
--- a/include/freerdp/utils/chunkstream.h
+++ b/include/freerdp/utils/chunkstream.h
@@ -1,0 +1,215 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ *
+ * Copyright 2020 Hardening <contact@hardening-consulting.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef FREERDP_UTILS_CHUNKSTREAM_H
+#define FREERDP_UTILS_CHUNKSTREAM_H
+
+#include <winpr/wtypes.h>
+#include <winpr/stream.h>
+#include <freerdp/api.h>
+
+typedef struct _ChunkStream ChunkStream;
+typedef struct _ChunkStreamSlot ChunkStreamSlot;
+
+#define CHUNKSTREAM_MAX_SLOTS 50
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+	/**
+	 * A ChunkStream is a component to prevent multiple memory copies when forging a packet from
+	 * headers to the payload. The idea is to create a list of chunks (slots) and to do the
+	 * assembly of the full content only when needed (usually when we want to send the content over
+	 * the network).
+	 *
+	 * The chunks can come from static blobs (chunkstream_getStaticStringSlot or
+	 * chunkstream_getStaticMemSlot), from memory allocated by the user (chunkstream_getMallocSlot)
+	 * or from a limited pool allocated by the chunkstream (chunkstream_getPoolSlot).
+	 *
+	 * chunkstream_sizeAfterSlot is useful to compute the size of chunks after a certain slot,
+	 * typically when you want to write some headers that inform of the size of the following
+	 * payload.
+	 *
+	 * A full example usage with a protocol with at least 2 layers:
+	 * 		void layer2(ChunkStream *cs)
+	 * 		{
+	 * 			... allocate and write some more chunks ...
+	 * 		}
+	 *
+	 * 		void layer1(ChunkStream *cs)
+	 * 		{
+	 * 			wStream layer1Headers;
+	 * 			// headers of layers 1 are at max 20 bytes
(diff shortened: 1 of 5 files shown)


hardening:
@freerdp-bot seems like you should re test for windows host

hardening:
@freerdp-bot can you test again the windows part ?

include/freerdp/utils/chunkstream.h
```diff
@@ -0,0 +1,215 @@
...
+#define FREERDP_UTILS_CHUNKSTREAM_H
+
+#include <winpr/wtypes.h>
+#include <winpr/stream.h>
+#include <freerdp/api.h>
+
+typedef struct _ChunkStream ChunkStream;
+typedef struct _ChunkStreamSlot ChunkStreamSlot;
+
+#define CHUNKSTREAM_MAX_SLOTS 50
```
akallabeth, line 29: Might better be private, in the `.c` file

include/freerdp/utils/chunkstream.h
```diff
@@ -0,0 +1,215 @@
...
+
+	/**
+	 * Retrieves a slot containing a memory blob alloced by the user, the chunkstream takes
+	 * ownership of the provided memory, so it will be freed when the chunkstream is destroyed.
+	 * @param chunkstream the chunkstream
+	 * @param data the memory blob to add
+	 * @param sz size of the memory blob
+	 * @return the corresponding ChunkStreamSlot, NULL if it failed (no available slots)
+	 */
+	FREERDP_API ChunkStreamSlot* chunkstream_getMallocSlot(ChunkStream* chunkstream, BYTE* data,
```
akallabeth, line 120: so, you assume the memory is allocated with `malloc`? (there are also the `aligned_malloc` et al which require a custom deallocator)

include/freerdp/utils/chunkstream.h
```diff
@@ -0,0 +1,215 @@
...
+	 */
+	FREERDP_API BOOL chunkstream_linearizeInStream(ChunkStream* chunkstream, wStream* s);
+
+	/**
+	 * Retrieves the size stored in the chunkstream after a given slot
+	 * @param chunkstream the chunkstream
+	 * @param slot the slot to start to compute
+	 * @return the stored size, -1 if the slot was not found
+	 */
+	FREERDP_API int chunkstream_sizeAfterSlot(const ChunkStream* chunkstream,
```
akallabeth, line 170: wrong data type, if you require `-1` use `SSIZE_T`

akallabeth (commented):
Maybe I´m missing something, but where is the benefit of all this?
There is additional allocation overhead and more data needs to be copied.

To be precise: I´d appreciate a good wrapper for reoccurring patterns, just not convinced here.

include/freerdp/utils/chunkstream.h
hardening: Well, that's what I originally did, but I figured that a caller might be interested to know how many slots/chunks it can create.
could also be a getter on the `ChunkStream` though.

include/freerdp/utils/chunkstream.h
hardening: Hum, `ChunkStream` is sending packets oriented, so I don't know if we're really using memory allocated by `aligned_malloc` when forging packets, but I can add a function in that sense if needed.

include/freerdp/utils/chunkstream.h
hardening: ok will change it. Anyway `int` is not problematic, is it ?

akallabeth: no, not problematic, but better to use the dedicated size types.

hardening:
I may not have explained enough what `ChunkStream` tries to solve. On the path to create a full RDP packet we have places where headers have variable length, and the size of the headers usually depend on the size and content of the payload (so we can't really pre-allocate a fixed place for headers in a `wStream`). So for a layer `n` the code usually looks like:

```C
wStream *createPacketLayerN(...) {
   wStream *payload = createLayerNPayload(...); // will call createPacketLayerN+1
   size_t len = computeLayerN_size(payload);

   wStream *ret = Stream_New(NULL, len);
   writeHeadersLayer(ret);
   Stream_Write(ret, Stream_Buffer(payload), Stream_Position(payload));
   return ret;
}
```
If you stack some layers then you have these memory copies done at each layer.

In other places we use `Stream_EnsureRemainingLength()` to grow the memory block as much as we need, but when a `realloc` happens that also means a possible memory copy in our back.

The idea with `chunkStream` is that layer 0 (on top of transport for us) creates a `ChunkStream`, it reserves the slots for its headers and ask layer 1 to do the same, and the `ChunkStream` is passed across different layers that add more chunks:

* when a layer just put static bits they aren't copied during all the layer traversal;
* when it's just small bits of data like headers you can use the `ChunkStream` pool, so no `malloc`;
* when data have bigger size then `malloced` style slots allows to put dynamic data and have the cleanup done once used

I'm figuring that for fastpath fragmentation we may need a split function that would take an input `ChunkStream` and would split it in 2 with the first one bound to the given size.

akallabeth:
@hardening ok, I might be mistaken, but:

1. the header size is known for the packets, only the payload length is unknown, so this will not differ (much) from what we currently do (skip header length, write payload data, write header)
1. The lower level functions already get/provide a copy anyway (encryption/compression/...)
the fragmented packets must be assembled too, which happens currently when each fragment is received, so no big change there.
1. Reallocations can easily be avoided using best guess overallocation (and a stream pool?)

so, yes the pr is interesting, but I currently fail to see where the copies you mention happen.

hardening:
About your point 2, `ChunkStream` is mainly to help sending packet, not receiving.

But ok, I get your point let's drop that PR for now.

PR #6620, opened 2020-12-01
reset codecs in gdi_pipeline_init

kubistika:
Fixes issues with `clear codec` with rdp proxy:
In case of GFX channel disconnect and connect done during session runtime the codec state was not reset


akallabeth (approved):
LGTM

PR #6636, opened 2020-12-04
Check for backtrace symbol and link to execinfo if needed

DanySpin97:
I assume that backtrace symbol is needed for libwinpr (as my build failed to link without it), so cmake now errors when it cannot find `backtrace` function.

[libexecinfo](https://pkgs.alpinelinux.org/package/edge/main/x86/libexecinfo) is a library that provides backtrace header and symbol on systems that don't have it (like musl based ones). This PR should be safe to add for glibc users.

### DIFF
--- a/winpr/libwinpr/utils/CMakeLists.txt
+++ b/winpr/libwinpr/utils/CMakeLists.txt
@@ -15,6 +15,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+include(CheckFunctionExists)
+
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
 set(${MODULE_PREFIX}_COLLECTIONS_SRCS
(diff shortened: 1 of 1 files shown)


winpr/libwinpr/CMakeLists.txt
```diff
@@ -87,6 +87,17 @@ if (TIMER_CREATE AND TIMER_DELETE AND TIMER_SETTIME AND TIMER_GETTIME)
 	winpr_library_add_private(rt)
 endif()
 
+set(CMAKE_REQUIRED_INCLUDES backtrace.h)
+check_function_exists(backtrace BACKTRACE)
+if (NOT BACKTRACE)
+    set(CMAKE_REQUIRED_LIBRARIES execinfo)
+    check_function_exists(backtrace EXECINFO)
+    if (NOT EXECINFO)
+        message(FATAL_ERROR "Could not find backtrace function")
```
akallabeth, line 96: Pleas no fatal errors, linux is not the only platform we support.

akallabeth (changes requested):

winpr/libwinpr/CMakeLists.txt
DanySpin97: Fixed.

akallabeth:
@freerdp-bot test

akallabeth (changes requested):
Patch now looks good, but there is already a check in 
`winpr/libwinpr/utils/CMakeLists.txt`
- so better move it there. (sorry, did not check that before)

DanySpin97:
> Patch now looks good, but there is already a check in
> `winpr/libwinpr/utils/CMakeLists.txt`
> 
>     * so better move it there. (sorry, did not check that before)

Oh, I've completely missed that. I have moved the check there and restricted it to unix systems only. I've also updated the commit message.

akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #6649, opened 2020-12-08
Update README.android in regards to CMake, NDK and ABIs

Redirion:
- CMake should be used from Android SDK. Personally I had issues using CMake from my OS (Ubuntu)
- minimum version of Android NDK must be 15c, as build target API level 26 is not supported on earlier releases
- removed ABIs that are unsupported in recent NDK and also not build by default

@akallabeth thank you for pointing out that I was using the wrong CMake version.

### DIFF
--- a/docs/README.android
+++ b/docs/README.android
@@ -13,8 +13,8 @@ Build requirements
 For the Android port some additional dependencies need to be fulfilled:
 
 * for JNI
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):

Issue #6664, opened 2020-12-13
server side external io api logic broken


sss123next:
currently impossible to override `io.TransportAttach` because `->context->rdp->transport` is available only after `freerdp_peer_context_new` call, but inside of `freerdp_peer_context_new_patched` `transport_attach` is called before any chance to override it, it is possible to create own implementation of `freerdp_peer_context_new`, but it will also add requirement for implementation of lot freerdp internals which is undesired and should be avoided if possible.

sss123next:
also looks like some  in some other places io descriptor used directly in server side (peer) code

sss123next:
looks like another task for me (.
ok, i am on it.

sss123next:
https://github.com/FreeRDP/FreeRDP/pull/6668 - started work here

PR #6668, opened 2020-12-15
WIP: #6664 fixes for server side external io api

sss123next:
this will contain necessary fixes in order server side part of external io api to work properly.

### DIFF
--- a/libfreerdp/core/peer.c
+++ b/libfreerdp/core/peer.c
@@ -823,11 +823,11 @@ BOOL freerdp_peer_context_new(freerdp_peer* client)
 	{
 		if (!transport_attach(rdp->transport, client->sockfd))
 			goto fail_transport_attach;
+		transport_set_blocking_mode(rdp->transport, FALSE);
 	}
 
 	rdp->transport->ReceiveCallback = peer_recv_callback;
 	rdp->transport->ReceiveExtra = client;
-	transport_set_blocking_mode(rdp->transport, FALSE);
 	client->IsWriteBlocked = freerdp_peer_is_write_blocked;
 	client->DrainOutputBuffer = freerdp_peer_drain_output_buffer;
 	client->HasMoreToRead = freerdp_peer_has_more_to_read;
--- a/libfreerdp/core/peer.c
+++ b/libfreerdp/core/peer.c
@@ -819,8 +819,11 @@ BOOL freerdp_peer_context_new(freerdp_peer* client)
 		goto fail_error_description;
 	}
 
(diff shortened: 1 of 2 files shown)


libfreerdp/core/peer.c
```diff
@@ -819,8 +819,11 @@ BOOL freerdp_peer_context_new(freerdp_peer* client)
 		goto fail_error_description;
 	}
 
-	if (!transport_attach(rdp->transport, client->sockfd))
-		goto fail_transport_attach;
+	if (client->sockfd != -1)
```
akallabeth, line 822: Why not simply add a `client->ContextConfigure` callback to `struct rdp_freerdp_peer` and call it here?
If the backend desires to configure something, implement it, otherwise skip it.
(just add the new function pointer at the end of the struct)

akallabeth (changes requested):
Thank you for the patch.
Could you address the suggested changes?
Thank you.

libfreerdp/core/peer.c
sss123next: because `freerdp_peer_context_new` is not internal only call and used by existing implementations,  but i think suggested solution will also work.

sss123next: additionally it will require to implement duplicated `freerdp_peer_context_new` in implementation, i am not sure what this is a good idea, or maybe i misunderstand you ?

sss123next:
i am just started, this is not ready to be merged yet, it's work in progress as suggested by prefix in PR name.
looks like in `peer.c` a lot of io done directly and not using api from `transport.c`

libfreerdp/core/peer.c
sss123next: one more reason to not add one more callback i what context creation api already have one callback for additional context initialization from user.

sss123next: i think this api can be changed to more convenient and flexible one, but this will break compatibility with existing implementations using this api's

akallabeth: Ok, the suggestion I tried to explain was add a simple call `IFCALLRET(client->ContextConfigure, ret, client, client->context);` or similar before any `transport_attach` et al are called.
That will allow you to set your own transport functions and you can remove these socket checks (that dereference deep into the lower layers)

sss123next: ok, but we need somehow to disable socket related calls anyway, i mean calls which is not in switchable api, just avoiding call `transport_attach` maybe not most elegant solution, but it will do the job without requirement of additional changes

PR #6670, opened 2020-12-15
progressive codec simple tile fixes

akallabeth:
Fixes issues with `ogon` and `xfreerdp /v:<host> /gfx` which forces progressive GFX codec

1. Adds working `progressive_compress` function producing simple tiles
1. Adds unit test to (lossy) compare image data
1. Fixes the decoding process for missing `subbandiff`, which requires fallback to RFX 
1. Cleans up RFX API (const correctness, signed/unsigned)

@bmiklautz @nfedera @hardening I´ve taken the liberty to copy `progressive_rfx_write_message_progressive_simple` serialization code mostly from ogon, please add your approval for the relicensing, otherwise I´ll remove that.


libfreerdp/codec/progressive.c
```diff
@@ -2324,22 +2556,38 @@ BOOL progressive_context_reset(PROGRESSIVE_CONTEXT* progressive)
...
-		progressive->SurfaceContexts = HashTable_New(TRUE);
-		progressive_context_reset(progressive);
-		progressive->log = WLog_Get(TAG);
-	}
+	progressive->Compressor = Compressor;
+	progressive->quantProgValFull.quality = 100;
+	progressive->log = WLog_Get(TAG);
+	if (!progressive->log)
+		goto fail;
+		progressive->rfx_context = rfx_context_new(Compressor);
```
hardening, line 2569: indentation problem here

hardening:
I'll try it as soon as I can, because I had also noticed that bug when in `rfx3` mode.

mfleisz (approved):
LGTM - @hardening if this is working for you as well we can merge

libfreerdp/codec/progressive.c
```diff
@@ -2308,10 +2318,253 @@ INT32 progressive_decompress(PROGRESSIVE_CONTEXT* progressive, const BYTE* pSrcD
...
+	RFX_TILE* tile;
+	UINT32 tilesDataSize;
+
+	if (context->mode != RLGR1)
+	{
+		WLog_ERR(TAG, "%s: error, RLGR1 mode is required!", __FUNCTION__);
+		return FALSE;
+	}
+
+	/* RFX_PROGRESSIVE_SYNC */
```
pnowack, line 2336: Both `RFX_PROGRESSIVE_SYNC` and `RFX_PROGRESSIVE_CONTEXT` should only be sent once. This is what the RDS in MS Windows does.

pnowack (commented):
The quantization values for RFX Progressive (plain quants) in MS Windows are different from the ones that plain RFX uses here in FreeRDP: `{6, 6, 6, 7, 8, 8, 9, 9, 9, 10}` (Order is LL3, HL3, LH3, HH3, HL2, LH2, HH2, HL1, LH1, HH1).  
These should be used here too in my opinion (for RFX Progressive).

libfreerdp/codec/progressive.c
```diff
@@ -2308,10 +2318,253 @@ INT32 progressive_decompress(PROGRESSIVE_CONTEXT* progressive, const BYTE* pSrcD
...
+	{
+		return FALSE;
+	}
+	Stream_Write_UINT16(s, 0xCCC2);   /* blockType (2 bytes) */
+	Stream_Write_UINT32(s, blockLen); /* blockLen (4 bytes) */
+
+	return TRUE;
+}
+
 int progressive_compress(PROGRESSIVE_CONTEXT* progressive, const BYTE* pSrcData, UINT32 SrcSize,
```
pnowack, line 2457: Maybe this could be renamed to `progressive_compress_simple`? This would leave space for a potential `progressive_compress_first` and `progressive_compress_upgrade` function for progressive encoding.

akallabeth: plan is to let the compression routine decide that at a later point in time according to input data, so no manual switch is required.

libfreerdp/codec/progressive.c
akallabeth: Yes, it is a possible improvement, but the current flow is within spec, so I´d prefer fixing the issue at hand first and improve upon that later on.

akallabeth:
@pnowack the (default) quantization values might have changed. I´d prefer to change that part in a follow up pr.

libfreerdp/codec/progressive.c
pnowack: Ah ok. That sounds great.

libfreerdp/codec/progressive.c
pnowack: Alright.

PR #6672, opened 2020-12-15
reset NDK_TARGET back to 21

Redirion:
this was changed in this commit here: https://github.com/FreeRDP/FreeRDP/commit/2ff19597dcd4abc486dbd311b551f0c22369177c#diff-e3c55acefd4c51dba191184fbb0d7e03724736f813c9a81bc95a1dbb2aff0d0b

Tested with NDK r21d and NDK r15c for OPENH264 without issues.


akallabeth:
sorry, this is the debug configuration, which benefits greatly from the higher API level.

PR #6688, opened 2020-12-23
Fix parsing of current keyboard locale

mfleisz:
The layout specifier might contain a comma separated list of keyboard layouts i.e. `"at,us"`. The current parsing code fails and `find_keyboard_layout_in_xorg_rules` returns 0. The PR fixes this issue by taking the first country code found and returning the correct layout id.
The PR also adds a missing keyboard layout for Austrian keyboards.


akallabeth (approved):
LGTM

Issue #5688, opened 2019-10-28
Copy and Paste files via clipboard on RDP sessions between Host and Clients or Clients and Clients


reddwarf666:
**Describe the solution you'd like**
The ability to copy files via clipboard between Host and Client or Client to Client sessions

**Describe alternatives you've considered**
Currently sharing folders between Host and Client and move files like that

akallabeth:
@reddwarf666 copy in one direction (client -> server) is already supported by `xfreerdp` (#3905). For the other direction some work is necessary.
pull requests welcome.

reddwarf666:
@akallabeth thanks! I am kinda dependent on a tool using FreeRDP, namely Royal TSX. I will inform them about the xfreerdp library and hope they can work with that (or even contribute).

akallabeth:
@reddwarf666 ah, ok. could get challenging as some kind of async file data request is necessary (on linux it may be possible to use fuse) but mac could get challenging. (macfuse et al need kernel modules)

lemonmojo:
Hi there! This is Felix, main dev of Royal TSX. We've had many requests for this particular feature over the last couple of years and I'd be willing to finance the implementation if someone's available for the task.

lemonmojo:
That being said, FUSE isn't an option. If any dependencies are necessary, I need to be able to statically link them in.

byteboon:
FWIW, I did a deep dive on this and came up with FUSE really being the only option.

See the details on https://stackoverflow.com/questions/58886232/how-to-use-nspasteboard-with-kpasteboardtypefileurlpromise-for-copy-paste

chucker:
@byteboon I'm a bit confused by the answer there:

> It appears that the key piece here is that Finder requires that the file actually be present on disk for the paste action to be enabled for a file URL.

The whole point of NSFile**Promise**Provider, i.e. file _promises_, is that the files _don't_ have to be present. (Maybe I'm misunderstanding what they're saying here.) For example, when you use the Photos app, you can drag photos to the Finder (and other apps that support file promises), even if those photos haven't actually been downloaded in full quality. The actually file data gets downloaded on _drop_. That matches with [Apple's documentation](https://developer.apple.com/documentation/appkit/nsfilepromiseprovider?language=objc), too:

>When you're working with drag and drop, use promises to indicate intent for future action. **Avoid loading or performing any actions on the file until the promise completes.**

[...]

Issue #5291, opened 2019-03-01
Copy file from remote to local all file is


vicentwu:
**Found a bug? - We would like to help you and smash the bug away.**
1. __Please don't "report" questions as bugs. For these (questions/build instructions/...) please use one of the following means of contact:__
   * We are reachable via IRC _#freerdp on freenode_
   * We are reachable via mailing list <freerdp-devel@lists.sourceforge.net>
   * Try our mailing list for discussions/questions
1. Before reporting a bug have a look into our issue tracker to see if the bug was already reported and you can add some additional information.
1. If it's a __new__ bug - create a new issue.
1. For more details see https://github.com/FreeRDP/FreeRDP/wiki/BugReporting

**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

** Application details
* Version of FreeRDP
* Command line used
* output of `/buildconfig`
* OS version connecting to
* If available the log output from a run with `/log-level:trace` 

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

** Please remove this text before submitting your issue!

_Thank you for reporting a bug!_

akallabeth:
Could you please be a bit more precise with your description?
* `Latest` is a very vague term (there are several sources you can get freerdp from...)
* Are you talking about clipboard, drive redirection, ...?

akallabeth:
Ok, looks like what was meant was the clipboard file copy feature.

vicentwu:
Yes,  problem of copying files！

jkfoong:
I'm having the same issue. Has this been resolved?

akallabeth:
@jkfoong still waiting for a usable problem description, don't have it on my test machines and nobody told what todo and how to reproduce ;)

jkfoong:
@akallabeth I'm having this issue when i'm trying to copy files from server to local, just simple copy and paste.
Could you please guide me to reproduce the debug log to further investigate this issue?

akallabeth:
@jkfoong so far so good, but any versions, command line, server os, ...?

akallabeth:
Ok, messed the last word, sorry 😓

jkfoong:
This is FreeRDP version 2.0.0-dev5 (8cf19fd94)
And here's the debug log
[](https://jpst.it/1XrdP)

akallabeth:
@jkfoong thank you. your remote machine is also windows 10?

jkfoong:
[...]

Issue #4703, opened 2018-06-07
Guest to host file copy paste not working


varunatarcon:
**Describe the bug**
Copying file from guest to host is not working. But working other way. Bidirectional copy paste is supported for text though

**To Reproduce**
1. Copy a file from guest machine.
2. Paste it on host machine

**Expected behavior**
File must be copied to host machine.

** Application details
* Version of FreeRDP: FreeRDP v2.0.0 dev3
* Command line used: 
xfreerdp /v:<someip> /u:<someusername> /p:<password> +wallpaper /rfx /size:60% /smart-sizing:640x480 +clipboard
* output of `/buildconfig`

This is FreeRDP version 2.0.0-dev3 (n/a)
[...]

PR #6699, opened 2020-12-30
x11: add support for remote to local clipboard file copy

jackyzy823:
## Idea:

Thanks @ilammy for his previous excellent work (#3905)  and  vmware's `open-vm-tools` project for inspiring me when i'm copying files from host to guest with nautilus and letting me know TARGET and format required by nautilus and other file managers.

## It should fix:
#5688  #6355  #5291  #4703  and others.

## How:

TL:DR: **make use of `Delayed Rendering` of X11**

1. remote rdp server Ctrl+C => remote send FormatList PDU
2. local rdp client => get  FormatList PDU
3. local rdp client => send FormatData Request PDU
4. remote rdp server => send FormatData Response PDU with prepared filedescriptors list
5. local rdp client => get  FormatData Response PDU with prepared filedescriptors list
6. local  rdp client  make a fuse which represents filedescriptors
7. local rdp client  tell  X Server  own  CLIPBOARD
8. filemanager know from x server that someone own CLIPBOARD 
9. filemanager (via X server) ask local rdp client if  local rdp client can response to target  like `x-special/gnome-copied-files` (xfce/unity asks) or `UTF8_STRING` (for nautilus) `text/uri-list` (for kde) and etc..
10. local rdp client  get required target ,  convert filedescriptors to fuse files list ,  add  prefix/suffix  of required target  to list and  response (via X Server)  to filemanager
11. filemanager get a list of files in fuse  to copy 
12. filemanager trigger filesystem releated syscalls to local rdp client's fuse 
13. when read syscall, local rdp client send Format Data Request PDU to remote rdp server
14. remote rdp server  send Format Data Response PDU 
15. local rdp client get Format Data Response PDU and return data to read syscall via callback

## Usage:

1. in filemanager with  Ctrl+V or right click menu-> paste
2. direct `cp -r <fuse_mount_path>/somefile dst`  (which means you can copy files which you're intersted without copying all files)

## About speed:

Speed depends on the  size per read call (max 32 pages 128kb , for fuse3 max 1M see FUSE_MAX_PAGE ) and time interval between send file contents request and receive file contents response. 

On my two vms in the same machine , speed can be 5~8 MB/s which i think is adorable.

However on another two vms in another machine , speed is 160KB/s, and i find time interval between send file contents request and receive file contents response is about 1 second which is weird.

## Test Passed (simple test):

1. Arch/Gnome/nautilus 3.38.2-1 pass
2. Manjaro/KDE/dolphin pass
3. Mint20/MATE/caja :no local->remote FormatData Request triggered.  -> fix by adding `x-special/mate-copied-files`
4. Xubuntu latest/XFCE/thunar pass
5. elemetaryos/Pantheon/Pantheon files
6. Lubuntu20.10/LXQT/pcmanfm-qt pass
7. Lubuntu18.04/LXDE/pcmanfm 1.2.5 (libfm 1.2.5)

## Test Failed:

1.  Deepin : need TARGET=UTF8_STRING but format in `x-special/gnome-copied-files`  which conflicts with nautilus 
2. Mint20/Cinnamon/nemo : it do not need last \n therefore it conflicts nautilus. it shows a error dialog but user can just skip it and continue.
3. Arch/LXDE/pcmanfm 1.3.1 (libfm 1.3.1)  treat some file as folders 

## Test cases should be considered:
1. a lot of files
2. large files
3. file name with CJK or other non-english
4. file name without extension to test determine file/folder
5. nested folder
6. test cases from ilammy/freerdp#4

## Internal Implementation:

1. ArrayList `stream_list` : contains all (stream_id , fuse_req_t) for response to `read` call  asynchronously
2. ArrayList `inode_table`: cotains file name,size,time and parent ino number and child inode numbers (if it is a folder)
3. Hashtable `mapDir`: use when constructing  `inode_table` to  get parent ino_num quickly

## Problems and TODOs:
1. when generating uri-list , different filemangers have different requirement for last \n and null-terminated char.
2. More tests required. 
    1. under bad network.
    2.  **What if another remote Ctrl+C action when transmission**
    3. performance of large amount of FileDescriptor
3. stat time nsecs need check_struct_has_member. currently it is ignored.
    see `libfuse/lib/fuse_misc.h` and AC_CHECK_MEMBERS([struct stat.st_atimespec]),
4. optimize data struct of inode table and performance test
5. direct cp may have data loss issue. (to be investigated) (maybe fixed by fuse  direct_io)
6. fuse related 
    1. **as a optional/required dependence , if optional ,define a macro. if required it's a breaking change.**
    2. fuse2 or fuse3 (fuse 3 has FUSE_MAX_PAGES and better lowlevel api)
    3. better FindFUSE.cmake
7. split fuse code to xf_cliprdr_fuse.c to make code struct more properly?
8. make this exposeable to remmina or others or just a reference implementaion.
9. basepath configaurable? currently GetKnownPath(TMP)/.xfreerdp.cliprdr.<pid>
10. fuse multi thread or single thread (current choice).
11. HUGE FILE flag support
12. Do we need send File Contents Request (the FILECONTENTS_SIZE one) PDU  to get filesize? since FormatData Response PDU contains filesize too.
13. Do we need send LockData Request which is described 'optional' in documents
14. port to wlfreerdp
15. transmission may be canceled when lockscreen
16. all other problems.

## Others:

1. After reading MacOS' NSPasteBoard document , i think it's impossbile to do this  since it do not support `Delayed Rendering` feature described in `MS-RDPECLIP` 1.3.1.4 Delayed Rendering of Clipboard Data
2. However we can do direct cp -r <fuse_mount_path>/somefile dst , with support of osxfuse 
4. However why just not use /drive: and Microsoft RDP client for MacOS  can only do this too


jackyzy823:
Well . `libfuse-dev` is required. :disappointed:

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+		fuse_reply_err(req, ENOTDIR);
+	}
+	else
+	{
+		fuse_reply_open(req, fi);
+	}
+}
+
+static struct fuse_lowlevel_ops xf_cliprdr_fuse_oper = {
+	.lookup = xf_cliprdr_fuse_lookup,
```
pnowack, line 2194: You don't need an implementation for `lookup` and `opendir` (not necessary, can be removed).

client/X11/xf_cliprdr.c
```diff
@@ -1388,6 +1515,19 @@ xf_cliprdr_server_format_data_request(CliprdrClientContext* context,
 	return CHANNEL_RC_OK;
 }
 
+static char* xf_cliprdr_fuse_split_basename(char* name, int len)
```
pnowack, line 1518: You can use `strrchr` instead of this function.

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+	/* TODO: set up a filesystem base path for local URI */
+	/* TODO get basePath from config or use default*/
+	UINT32 basePathSize;
+	char* basePath;
+	char* tmpPath;
+	tmpPath = GetKnownPath(KNOWN_PATH_TEMP);
+	/* 10 is max length of DWORD string and 1 for \0 */
+	basePathSize = strlen(tmpPath) + strlen("/.xfreerdp.cliprdr.") + 11;
+	basePath = calloc(basePathSize, sizeof(char));
+	_snprintf(&basePath[0], basePathSize, "%s/.xfreerdp.cliprdr.%d", tmpPath,
```
pnowack, line 2215: Using `mkdtemp` for the mount path should be safer here, since that will create a directory with a name that definitely won't collide with an existing one.

pnowack (commented):
> fuse2 or fuse3 (fuse 3 has FUSE_MAX_PAGES and better lowlevel api)

Always use the newer one for new implementations (especially fuse3).

> basepath configaurable? currently GetKnownPath(TMP)/.xfreerdp.cliprdr.

Should be fine in my opinion. `XDG_RUNTIME_DIR` would also be an option.

> Do we need send File Contents Request (the FILECONTENTS_SIZE one) PDU to get filesize? since FormatData Response PDU contains filesize too.

You first check for the `FD_FILESIZE` flag for a file. If it is present, you already have the size and you don't need to request the size any more. Otherwise you need to request the size when you receive a FUSE request where the size is necessary.

> Do we need send LockData Request which is described 'optional' in documents

No, its optional. However, when advertising the CB_CAN_LOCK_CLIPDATA flag and the server supports file locking too, you need to expect that the server can request such lock.  
I don't know how this needs to be implemented. The easiest way is to simply not advertise the flag to not expect such request from the server, but at the same time you won't be able to request a lock from the server, as it is disabled.

winpr/libwinpr/clipboard/posix.c
```diff
@@ -627,20 +638,18 @@ static void* convert_filedescriptors_to_uri_list(wClipboard* clipboard, UINT32 f
...
 
-		if (rc != (int)curLen)
-		{
-			free(curName);
-			free(dst);
-			return NULL;
-		}
-
-		rc = _snprintf(&dst[pos], alloc - pos, "%s/%s\r\n", clipboard->delegate.basePath, curName);
+		rc = _snprintf(&dst[pos], alloc - pos, "file://%s/%s\n", clipboard->delegate.basePath,
```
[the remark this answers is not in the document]
jackyzy823: Ok , i'll fix this in a more proper way.

winpr/libwinpr/clipboard/posix.c
```diff
@@ -657,15 +666,104 @@ static void* convert_filedescriptors_to_uri_list(wClipboard* clipboard, UINT32 f
...
+	size_t header_len = strlen(header);
+	*pSize = *pSize + header_len;
+	char* pNewDstData = calloc(*pSize, sizeof(char));
+	_snprintf(pNewDstData, *pSize, "%s%s", header, (char*)pDstData);
+	free(pDstData);
+	return pNewDstData;
+}
+
+/* Prepend header of nautilus based filemanager's format to file list*/
+static void* convert_filedescriptors_to_nautilus_clipboard(wClipboard* clipboard, UINT32 formatId,
```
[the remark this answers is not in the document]
jackyzy823: I think this should be reserved for the compatibility. Even new version of nautilus including these fixes released, there're many distros will not update nautilus.

winpr/libwinpr/clipboard/posix.c
```diff
@@ -627,20 +638,18 @@ static void* convert_filedescriptors_to_uri_list(wClipboard* clipboard, UINT32 f
...
-		if (rc != (int)curLen)
-		{
-			free(curName);
-			free(dst);
-			return NULL;
-		}
-
-		rc = _snprintf(&dst[pos], alloc - pos, "%s/%s\r\n", clipboard->delegate.basePath, curName);
+		rc = _snprintf(&dst[pos], alloc - pos, "file://%s/%s\n", clipboard->delegate.basePath,
+		               curName);
```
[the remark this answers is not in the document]
jackyzy823: I don't know why won't be able to . I've tested with  filenames with ä, ö, ü chars. You can copy them properly.

client/X11/xf_cliprdr.c
```diff
@@ -1825,6 +2461,26 @@ void xf_clipboard_free(xfClipboard* clipboard)
 			free(clipboard->clientFormats[i].formatName);
 	}
 
+	if (clipboard->fuse_thread)
+	{
+		if (clipboard->fuse_sess)
+		{
+			fuse_session_exit(clipboard->fuse_sess);
```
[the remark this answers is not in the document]
jackyzy823: see comments and lines after this line.

client/X11/xf_cliprdr.c
jackyzy823: i think no. otherwise fuse do not know relationship between  ino number with filename.

jackyzy823:
> Always use the newer one for new implementations (especially fuse3).

Then it may not be backported to stable.  since some old distros (but not end of life)  do not have fuse3.

> You first check for the FD_FILESIZE flag for a file. If it is present, you already have the size and you don't need to request the size any more. Otherwise you need to request the size when you receive a FUSE request where the size is necessary.

will work on this.

client/X11/xf_cliprdr.c
```diff
@@ -1434,9 +1575,133 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
...
+
+						inode->st_size = (((UINT64)descriptors[lindex].nFileSizeHigh) << 32) |
+						                 ((UINT64)descriptors[lindex].nFileSizeLow);
+						ticks =
+						    (((UINT64)descriptors[lindex].ftLastWriteTime.dwHighDateTime << 32) |
+						     ((UINT64)descriptors[lindex].ftLastWriteTime.dwLowDateTime)) -
+						    116444736000000000;
+						inode->st_mtim.tv_sec = ticks / 10000000ULL;
+						/* tv_nsec Not used for now */
+						inode->st_mtim.tv_nsec = ticks % 10000000ULL;
```
ilammy, line 1669: I believe these magic constants need to be named and commented on.

Also, I'm not sure that those values are always set in FILEDESCRIPTOR structure.

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+{
+	double timeout = 0;
+	struct stat stbuf;
+
+	xfCliprdrFuseInode* node;
+	xfClipboard* clipboard = (xfClipboard*)fuse_req_userdata(req);
+
+	ArrayList_Lock(clipboard->ino_list);
+	node = (xfCliprdrFuseInode*)ArrayList_GetItem(clipboard->ino_list, ino - 1);
+	ArrayList_Unlock(clipboard->ino_list);
```
ilammy, line 1941: I'm not sure it's okay to release the lock right after the item has been found. It can be removed from the list and freed while we're keeping a pointer to it here, if the user is really anxious and keeps slamming Ctrl-C. Ditto for other accesses to `clipboard->ino_list`.

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+				fuse_remove_signal_handlers(clipboard->fuse_sess);
+				fuse_session_remove_chan(ch);
+			}
+			fuse_session_destroy(clipboard->fuse_sess);
+		}
+		fuse_unmount(clipboard->delegate->basePath, ch);
+	}
+	fuse_opt_free_args(&args);
+
+	if (PathFileExistsA(clipboard->delegate->basePath))
```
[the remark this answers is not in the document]
ilammy: > This is always true

Not if someone goes and drops FreeRDP's stuff manually from the terminal. But I agree, it is less prone to race conditions to just remove the directory directly.

ilammy (commented):
Oh well, I'm really happy that someone has finally came up with the missing link! Thank you, @jackyzy823! ❤️

I have not read through the whole changeset yet, but here are some drive-by nitpicks. Please don't consider them blocking, I'm not really a maintainer here so it's all just a random advice on the Internet.

client/X11/xf_cliprdr.c
```diff
@@ -31,10 +31,18 @@
 #include <X11/extensions/Xfixes.h>
 #endif
 
+#define FUSE_USE_VERSION 26
```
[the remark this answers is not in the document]
akallabeth: could be a `CMake` option so that the default is `35` but every distro maintainer can set that to whatever they require

akallabeth (approved):
Thank you for this, awsome :)
Looks good overall, some minor cleanups may be required, but that can be done later on.

akallabeth:
@freerdp-bot test
@jackyzy823 it may be nice to move the fuse part from `X11` to core library wherever possible, so that other clients (wayland, mac, ...) can easily reuse this.

akallabeth:
@jackyzy823 you might also add `libfuse-dev` or `libfuse3-dev` to `.github/workflows/codeql-analysis.yml` `apt` install line

client/X11/xf_cliprdr.c
```diff
@@ -1157,6 +1224,66 @@ static UINT xf_cliprdr_send_client_format_list_response(xfClipboard* clipboard,
...
+			break;
+		}
+	}
+	ArrayList_Unlock(clipboard->stream_list);
+	if (found)
+	{
+		fuse_reply_buf(stream->req, (const char*)fileContentsResponse->requestedData,
+		               fileContentsResponse->cbRequested);
+		ArrayList_RemoveAt(clipboard->stream_list, index);
+	}
```
ilammy, line 1283: I believe the lock needs to be held here until after the found item has been removed from the list.

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+		stbuf.st_mode = node->st_mode;
+		stbuf.st_mtime = node->st_mtim.tv_sec;
+		if (ino == FUSE_ROOT_ID)
+		{
+			stbuf.st_nlink = 2;
+		}
+		else
+		{
+			stbuf.st_nlink = 1;
+		}
```
ilammy, line 1959: Ternary operator makes this a bit more compact:

```c
stbuf.st_nlink = (ino == FUSE_ROOT_ID) ? 2 : 1;
```

Though, is this required for correctness? Why can't we just say "1 link" for everything?

I mean, strictly speaking, each regular file should have 1 link while for directories it depends on the number of subdirectories in them. Getting that number accurate is a bother and is not particularly important, I think.

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+}
+
+static void xf_cliprdr_fuse_readdir(fuse_req_t req, fuse_ino_t ino, size_t size, off_t off,
+                                    struct fuse_file_info* fi)
+{
+	size_t count;
+	size_t index;
+	size_t child_ino;
+	size_t direntry_len;
+	char* buf = (char*)calloc(size, sizeof(char));
```
ilammy, line 1971: It would be nice to check that this allocation succeeds.

Also, this buffer currently leaks if we exit early with ENOENT or ENOTDIR, or return an empty file list. I believe the buffer may be allocated only once we're sure that there is some non-empty data that we want to return.

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+
+				struct stat stbuf;
+				stbuf.st_ino = child_node->ino;
+				fuse_add_direntry(req, buf + pos, size - pos, child_node->name, &stbuf, index);
+				pos += direntry_len;
+			}
+		}
+
+		fuse_reply_buf(req, buf, pos);
+		free(buf);
```
ilammy, line 2047: I guess this better be done after the lock has been released. Who knows how long a FUSE call will take.

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+		fuse_reply_buf(req, NULL, 0);
+	}
+	else
+	{
+		for (index = off; index < count + 2; index++)
+		{
+			if (index == 0)
+			{
+				struct stat stbuf;
+				stbuf.st_ino = ino;
```
ilammy, line 2004: At first I wanted to note that these `struct stat` values should be properly initialized. `fuse_add_direntry()` docs suggest that it ignores everything except for `st_ino` and a part of `st_mode`. Inode number is handled here, but the mode is left with garbage value still.

However, I still think it would be safer to memset() them to zero, then fill in the required fields. Ditto for other call sites.

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+				direntry_len = fuse_add_direntry(req, NULL, 0, ".", NULL, index);
+				if (pos + direntry_len > size)
+				{
+					break;
+				}
+				fuse_add_direntry(req, buf + pos, size - pos, ".", &stbuf, index);
```
ilammy, line 2010: I think it is enough to do only one call here:
```c
direntry_len = fuse_add_direntry(req, buf + pos, size - pos, ".", &stbuf, index);
if (direntry_len > size - pos)
	break;
```
since `fuse_add_direntry()` promises to not modify the buffer if it cannot fit an entry there, but will still return the expected length.

Ditto for other call sites.

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+	size_t child_ino;
+	size_t direntry_len;
+	char* buf = (char*)calloc(size, sizeof(char));
+	size_t pos = 0;
+	xfCliprdrFuseInode* child_node;
+	xfCliprdrFuseInode* node;
+	xfClipboard* clipboard = (xfClipboard*)fuse_req_userdata(req);
+
+	ArrayList_Lock(clipboard->ino_list);
+	node = (xfCliprdrFuseInode*)ArrayList_GetItem(clipboard->ino_list, ino - 1);
```
ilammy, line 1978: This snippet with `- 1` correction and a cast is so ubiquitous that I think it might be useful to introduce a little utility:

```c
static inline xfCliprdrFuseInode* xf_cliprdr_get_inode(xfClipboard* clipboard, fuse_ino_t ino)
{
	return (xfCliprdrFuseInode*)ArrayList_GetItem(clipboard->ino_list, ino - 1);
}
```

This makes it less obvious that you need to hold a lock, but at least it removes some copypasta as well as all those 'magical' adjustments by one. (Come to think of it, might it be easier to just allocate one extra entry for the root inode which will never be used otherwise?)

client/X11/xf_cliprdr.c
```diff
@@ -1663,6 +1928,334 @@ static UINT xf_cliprdr_clipboard_file_range_failure(wClipboardDelegate* delegate
...
+	stream->stream_id = clipboard->current_stream_id;
+	clipboard->current_stream_id++;
+	ArrayList_Add(clipboard->stream_list, stream);
+	ArrayList_Unlock(clipboard->stream_list);
+
+	UINT32 nPositionLow = (off >> 0) & 0xFFFFFFFF;
+	UINT32 nPositionHigh = (off >> 32) & 0xFFFFFFFF;
+
+	xf_cliprdr_send_client_file_contents(clipboard, stream->stream_id, node->lindex, nPositionLow,
+	                                     nPositionHigh, size);
```
ilammy, line 2114: Can it happen that we have several overlapping read() requests in flight for a single file? 🤔

ilammy (commented):
Here's a next batch of comments.

Also, I think that all FUSE-related code lacks logging. It would be super-helpful during initial testing to have FreeRDP verbosely describe what the hell it's doing when debug logging is enabled.

client/X11/xf_cliprdr.c
jackyzy823: After a double check , i found that  calling `ArrayList_RemoveAt` before `ArrayList_Unlock` will cause deadlock if  ArrayList_New with `BOOL synchronized=TRUE`. :disappointed:

client/X11/xf_cliprdr.c
jackyzy823: i'll assign all needed  pointer's member to local var  between the lock to avoid missing pointer.

client/X11/xf_cliprdr.c
ilammy: You're not wrong in that it will cause a double lock if the collection is synchronized. However, given that almost everywhere you are locking it manually and inspecting it via ArrayList_GetItem(), I think it could be instantiated with `synchronized` set to false. Then ArrayList_RemoveAt() won't add more locks on top of whatever is done manually with ArrayList_Lock().

winpr/libwinpr/clipboard/posix.c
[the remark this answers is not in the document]
pnowack: I just looked up my test case for this (I implemented a server side cliprdr solution where I ran into this issue too):

On a MS Windows machine create directory called `#ä` and into that directory another directory, called `äö#`. When copying the `#ä` directory and pasting it locally, you won't get the expected result (don't remember what went wrong there, only that it didn't work).  
If the uris on the uri-list are escaped, everything worked fine.

client/X11/xf_cliprdr.c
pnowack: Ah, I just see that the function signatures for `fuse_operations` and `fuse_lowlevel_ops` are not the same (`fuse_operations` works on the filepath and not on the inode while `fuse_lowlevel_ops` uses the inode), so nvm.

client/X11/xf_cliprdr.c
```diff
@@ -1434,9 +1575,133 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
 
 		if (strcmp(clipboard->data_format_name, "FileGroupDescriptorW") == 0)
 		{
+			/* Build inode table for FILEDESCRIPTORW*/
+			/* Avoid Crash on empty message*/
+			if (size >= 4)
```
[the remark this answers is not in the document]
pnowack: There is also a function that already does this (`cliprdr_parse_file_list`), so that can be used here too

client/X11/xf_cliprdr.c
pnowack: > Also, I'm not sure that those values are always set in FILEDESCRIPTOR structure.

They are not. For example, if there is a server implementation that uses FreeRDPs current implementation for "current-peer to other-peer" file transfer, then they are not set (fixed by https://github.com/FreeRDP/FreeRDP/pull/6704).
The last write time should only be used if the `FD_WRITESTIME` flag is set, otherwise just fall back to `time (NULL)`.

winpr/libwinpr/clipboard/posix.c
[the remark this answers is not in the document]
ilammy: > directory called `#ä`

Oh... I believe the issue here is not “ä” but “#”, which is special in URLs but allowed in Windows filenames just fine. Trying to use “&” in for example should have similar issues if that's the case.

I guess “ä” are fine without escaping because the parser does not particularly care about them and skips them as is, but for “#” it does try to interpret the rest of the string as anchor, and that breaks filenames.

jackyzy823:
thanks for your all's reviews and comments.

Now make fuse `recommended`. users can build without fuse by -DWITH_FUSE=off. 
And if user specify -DWITH_FUSE2=on or -DWITH_FUSE3=on , fuse2/fuse3 will be forcely chose and if specficied lib is  not presented , cmake will fail. 


Still some issues found:
1. With `fuse2`,  manually umounting via  `fusermount -u path`  before closing client  will crash h at `fuse_session_exit `   when exiting . (do not know why and how ....)
2. In KDE Dolphin, required TARGET order seem random  ( or freerdp recvieve issue?).  so sometimes Dolphin requires UTF8_STRING first then freerdp will return a nautilus format file list. however Dolphin just complain about it ,users can skip it.

winpr/libwinpr/clipboard/posix.c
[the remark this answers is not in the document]
jackyzy823: Oh i haven't seem this . still more works to be done. like `#` `?`  :disappointed: 

ref:https://en.wikipedia.org/wiki/File_URI_scheme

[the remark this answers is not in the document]
jackyzy823: Maybe fixed.  Haven't had enough testing.

client/X11/xf_cliprdr.c
```diff
@@ -1583,6 +1598,247 @@ static char* xf_cliprdr_fuse_split_basename(char* name, int len)
...
+error4:
+	free(descriptor);
+error3:
+	HashTable_Free(mapDir);
+error2:
+	ArrayList_Clear(clipboard->ino_list);
+error:
+	ArrayList_Unlock(clipboard->ino_list);
+	Stream_Free(s, FALSE);
+	return FALSE;
```
ilammy, line 1852: This is a sign that there are too many things going on this function. It's very easy to mess up something here. Even if it's correct now, someone will edit it later and they could break something down: jump to an incorrect error label, forget to free or unlock something, etc.

I believe this could be improved by splitting this function into several parts where each is responsible for one or maybe a couple of resources and associated error handling. A function that allocates and frees `wStream`. A function that keeps a track of `ino_list`, its lock, and maybe the `HashTable`. A function that reads through the `FILEDESCRIPTORW` list and deals with filenames. A function that finally inserts `xfCliprdrFuseInode` objects. Or something like that.

client/X11/xf_cliprdr.c
```diff
@@ -37,6 +37,7 @@
 #include <sys/stat.h>
 #include <errno.h>
 #include <time.h>
+#define WIN32_FILETIME_TO_UNIX_EPOCH_USEC UINT64_C(116444736000000000)
```
ilammy, line 45: I guess this is better be placed after includes, together with other defines.

client/X11/xf_cliprdr.c
```diff
@@ -1886,12 +1900,14 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
...
 		if (strcmp(clipboard->data_format_name, "FileGroupDescriptorW") == 0)
 		{
+#ifdef WITH_FUSE
 			/* Build inode table for FILEDESCRIPTORW*/
 			if (xf_cliprdr_fuse_generate_list(clipboard, data, size) == FALSE)
 			{
 				/* just continue */
 				WLog_WARN(TAG, "fail to generate list for FILEDESCRIPTOR");
 			}
+#endif
```
ilammy: Is there a point in handling `FileGroupDescriptorW` at all if there is no FUSE support? Why not wrap this entire
```c
if (strcmp(clipboard->data_format_name, "FileGroupDescriptorW") == 0)
```
condition in an `#ifdef`?

winpr/libwinpr/clipboard/posix.c
```diff
@@ -629,6 +659,7 @@ static void* convert_filedescriptors_to_file_list(wClipboard* clipboard, UINT32
 			size_t curLen = _wcsnlen(descriptors[x].cFileName, ARRAYSIZE(descriptors[x].cFileName));
 			alloc += WideCharToMultiByte(CP_UTF8, 0, descriptors[x].cFileName, (int)curLen, NULL, 0,
 			                             NULL, NULL);
+			alloc += count_special_chars(descriptors[x].cFileName) * 2;
```
ilammy, line 662: I guess this warrants a comment as to why 2, not 3, or something else.

(Each special character is escaped as `%FF` – 3 characters – but one of them is already included in `cFileName` length, previously added.)

winpr/libwinpr/clipboard/posix.c
```diff
@@ -655,21 +686,62 @@ static void* convert_filedescriptors_to_file_list(wClipboard* clipboard, UINT32
...
+		previous_at = curName;
+		while ((stop_at = stop_at_special_chars(previous_at)) != NULL)
+		{
+			char* tmp = strndup(previous_at, stop_at - previous_at);
+			if (!tmp)
+			{
+				free(dst);
+				free(curName);
+				return NULL;
+			}
```
ilammy, line 711: This whole escaping thing gets really hard to read and reason about... I wonder if it can be improved, but I don't have any concrete ideas about it. Tough, as this does not seem to be like a very performance-sensitive piece of code, I guess it might be more simple to just read character by character and escape them as necessary, instead of using all those fancy C string functions.

Or maybe it will be better to first construct a string with printf() in a buffer – that makes it clear what the string actually contains. Then that buffer can be passed over to some escape_uri() function which will make a quick pass to check whether there *is* something to escape. If not then we're on a fast path, return the buffer unmodified. If there is, well then process it.

akallabeth:
@freerdp-bot test

akallabeth:
@jackyzy823 please run `clang-format` on your changes

jackyzy823:
@akallabeth  
CI says `https://ci.freerdp.com/job/code-quality-checker/1194/console`. However my `git clang-format` ask me to do this change.

```diff --git a/client/X11/xf_cliprdr.c b/client/X11/xf_cliprdr.c
index e8e2b0b3f..67f903831 100644
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -1034,8 +1034,7 @@ static BOOL xf_cliprdr_process_selection_request(xfClipboard* clipboard,
 
 	if (!delayRespond)
 	{
-		union
-		{
+		union {
 			XEvent* ev;
 			XSelectionEvent* sev;
 		} conv;
@@ -2023,8 +2022,7 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
 
 	xf_cliprdr_provide_data(clipboard, clipboard->respond, pDstData, DstSize);
 	{
-		union
-		{
+		union {
 			XEvent* ev;
 			XSelectionEvent* sev;
 		} conv;

```
And `clang-format  --version ` outputs `clang-format version 11.0.0`

client/X11/CMakeLists.txt
```diff
@@ -227,7 +227,7 @@ if(WITH_XFIXES)
 endif()
 
 find_package(FUSE REQUIRED)
-include_directories(${FUSE_INCLUDE_DIRS})
+include_directories(${FUSE_INCLUDE_DIR})
```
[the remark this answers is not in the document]
jackyzy823: will do this before these could be merged.

client/X11/xf_cliprdr.c
```diff
@@ -1302,20 +1312,24 @@ xf_cliprdr_server_file_contents_response(CliprdrClientContext* context,
...
+				/** ino must be exists and fileContentsResponse->cbRequested should be 64bit */
+				if (!ino || data_len != sizeof(UINT64))
 				{
+					fuse_reply_err(stream->req, EIO);
 					break;
 				}
+				UINT64 size;
+				wStream* s = Stream_New((BYTE*)data, data_len);
+				Stream_Read_UINT64(s, size);
+				Stream_Free(s, FALSE);
```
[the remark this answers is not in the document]
jackyzy823: then https://github.com/FreeRDP/FreeRDP/pull/6699#discussion_r550834787

client/X11/xf_cliprdr.c
```diff
@@ -66,7 +66,11 @@ typedef struct xf_cliprdr_format xfCliprdrFormat;
 struct xf_cliprdr_fuse_stream
 {
 	UINT32 stream_id;
+	/* must be one of FILECONTENTS_SIZE or FILECONTENTS_RANGE*/
+	UINT32 req_type;
 	fuse_req_t req;
+	/*for FILECONTENTS_SIZE must be xfCliprdrFuseInode* */
+	void* req_data;
```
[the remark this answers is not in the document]
jackyzy823: i'll rethink about it .

client/X11/xf_cliprdr.c
```diff
@@ -2017,10 +2164,10 @@ static void xf_cliprdr_fuse_getattr(fuse_req_t req, fuse_ino_t ino, struct fuse_
...
 
 	ArrayList_Lock(clipboard->ino_list);
-	node = (xfCliprdrFuseInode*)ArrayList_GetItem(clipboard->ino_list, ino - 1);
-	ArrayList_Unlock(clipboard->ino_list);
+	node = xf_cliprdr_get_inode(clipboard->ino_list, ino);
 	if (!node)
 	{
+		ArrayList_Unlock(clipboard->ino_list);
 		fuse_reply_err(req, ENOENT);
 	}
```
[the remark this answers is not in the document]
jackyzy823: thanks for the advice.

akallabeth:
@jackyzy823 some changes require multiple runs. But this case looks like correct? your posted changes and the ones from https://ci.freerdp.com/job/code-quality-checker/1194/console are equal, are they not?

jackyzy823:
@akallabeth  i mean CI requires ```union {``` but my clang-format changes it to
 ```
union
{
```

akallabeth:
@jackyzy823 true, we still had `clang-format-6.0` on the builder, updated to `clang-format-11` now.
@freerdp-bot test

jackyzy823:
More works to do (but not now)
1. cannot copy text while pasting file  (windows rdp can do this)
2. another remote Ctrl+C action when  pasting file  will break the transmission. (windows rdp will not)

These may be related to x11 clipboard mechanism or just xf_cliprdr incomplete implement.

akallabeth:
@freerdp-bot test
@jackyzy823 some hints for your remaining issues:
1. File copy should be done in a separate thread
2. I think `CLIPRDR_LOCK_CLIPDATA` and `CLIPRDR_UNLOCK_CLIPDATA` are responsible for that (informing the server that the data in the clipboard should not be discarded

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth (changes requested):
Overall looking good, but I get constant crashes on client close in `StreamPool_Free`

akallabeth:
@freerdp-bot test

akallabeth (approved):
Ok, basically working.
Some rogh edges and needs some refactoring (the fuse components can be reused for wayland and osx/fuse) 

Overall nice work, thank you very much!

a10155212722:
Excuse me，I have a problem when I use cliboard to copy file from remote host to my local host.
I have build FreeRDP-master with fuse.
My local host's system is Ubuntu20.02.
My remote host's system is Windows7.
I used "xfreerdp /v:172.17.7.143 /u:ypl /p :123tfyypl" 
to connect my remote host.
when I use the right mouse button to copy files from remote host, it's normal and no error. Next, I pasted to my local directory ,  a window appeared，it said "/tmp/.xfreerdp.cliprdr.595517/testfile.txt not a file or a directory", I couldn't find /tmp/.xfreerdp.cliprdr.595517 on my local host.
What may be the reason for this? 
Thanks.

a10155212722:
> Excuse me，I have a problem when I use cliboard to copy file from remote host to my local host.
> I have build FreeRDP-master with fuse.
> My local host's system is Ubuntu20.02.
> My remote host's system is Windows7.
> I used "xfreerdp /v:172.17.7.143 /u:ypl /p :123tfyypl"
> to connect my remote host.
> when I use the right mouse button to copy files from remote host, it's normal and no error. Next, I pasted to my local directory , a window appeared，it said "/tmp/.xfreerdp.cliprdr.595517/testfile.txt not a file or a directory", I couldn't find /tmp/.xfreerdp.cliprdr.595517 on my local host.
> What may be the reason for this?
> Thanks.

@akallabeth

jackyzy823:
hello @a10155212722  

Can you run xfreerdp with debug log?

if xfreerdp can not create a tmp folder , there must be error logs like "Failed to create directory ...."

PR #6704, opened 2021-01-03
winpr/clipboard: Also save lastWriteTime for FILEDESCRIPTORW

pnowack:

```
Currently, when a local uri-list is converted into a FILEDESCRIPTORW
list, WinPR doesn't submit the last write time for each file.
The result of this is that the last write time of each file on the
other peer will have the current time and not the actual last write
time that is present on the peer, where the files were copied from.

Fix this by also writing the last write time in addition to the
FD_WRITESTIME flag.
```

Tested with the stable-2.0 branch, but should also work fine with current master.

### DIFF
--- a/winpr/libwinpr/clipboard/posix.c
+++ b/winpr/libwinpr/clipboard/posix.c
@@ -22,6 +22,7 @@
 #endif
 
 #define _FILE_OFFSET_BITS 64
+#define WIN32_FILETIME_TO_UNIX_EPOCH UINT64_C(11644473600)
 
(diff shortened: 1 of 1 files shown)


pnowack:
test

pnowack:
Seems that the `FD_SHOWPROGRESSUI` flag has been renamed on the master branch to `FD_PROGRESSUI`.

@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #6707, opened 2021-01-07
core: Fix gateway error code mapping

mfleisz:
Some of the TSG error codes are "cut down" using the HRESULT_CODE macro.
While the macro was correctly used in the definition of RPC_TSG_FAULT_CODES it was not used in the comparison of the error
codes.
Not all codes are using HRESULT_CODE so we need to check for both the real code value and the code value using the macro to get all the errors.

The PR also removes the error code from the error string returned by rpc_error_to_string. The error code is passed into the function so it is not necessary to append it to the returned string as well.

The screwed formatting of the error code tables is also fixed.


mfleisz:
Supersedes #6698

akallabeth (approved):
LGTM

libfreerdp/core/gateway/rpc_fault.c
```diff
@@ -391,8 +374,16 @@ const char* rpc_error_to_string(UINT32 code)
...
 		if (RPC_TSG_FAULT_CODES[index].code == code)
 		{
-			sprintf_s(buffer, ARRAYSIZE(buffer), "%s [0x%08" PRIX32 "]",
-			          RPC_TSG_FAULT_CODES[index].name, code);
+			sprintf_s(buffer, ARRAYSIZE(buffer), "%s", RPC_TSG_FAULT_CODES[index].name);
+			goto out;
+		}
+	}
+
+	for (index = 0; RPC_TSG_FAULT_CODES[index].name != NULL; index++)
```
akallabeth, line 382: Can this only happen with `RPC_TSG_FAULT_CODES` and not with `RPC_FAULT_CODES`?

akallabeth (approved):

PR #6709, opened 2021-01-11
primitives: Respect alpha values in YUV444 SSE pixel routine

mfleisz:
Currently ssse3_YUV444Pixel overrides the destination alpha value and resets it to solid (0xFF). This caused graphical issues if previously the alpha channel was set (i.e. using `RDPGFX_CODECID_ALPHA`).


akallabeth (approved):
LGTM

PR #6711, opened 2021-01-12
Ignore alpha channel during color conversion.

akallabeth:
Most GFX graphics do not carry alpha data. Keep the original
alpha value unless specified otherwise.


mfleisz:
@freerdp-bot test

PR #6721, opened 2021-01-15
Check drive to hotplug for already being redirected

akallabeth:
some hotplug implementations report the same drive multiple times.
to avoid redirecting the same drive multiple times check if it is
already in the list before adding.


mfleisz (approved):
LGTM

PR #6724, opened 2021-01-18
locale: Fix keyboard detection on MacOS

mfleisz:
This PR fixes keyboard layout detection on MacOS.
The `LANG` environment variable might be empty on MacOS and causes `freerdp_get_system_language_and_country_codes` to fail. As this is not a reliable way to determine the current keyboard layout we now use CF APIs in order to obtain the current keyboard layout.
`freerdp_get_system_language_and_country_codes` is also fixed using the `CFLocale` API now.


hardening (approved):
Far from being a mac expert, but LGTM

akallabeth (approved):

PR #6739, opened 2021-01-21
channels/smartcard: Fix race in channel cleanup

mfleisz:
Sometimes the client crashes on MacOS when redirecting smartcard due to _xpc_api_misuse.
The issue seems to be a synchronization problem with apples pc/sc implementation.
With these changes we give the pc/sc service a bit of time to handle the cancel request before closing the context.


hardening (approved):
LGTM

akallabeth (approved):

Issue #6686, opened 2020-12-22
Non-standard exit codes under UNIX


ondrejholy:
**Describe the bug**
The exit code for `xfreerdp /version` is currently -2003. One problem is that as per the UNIX conventions, the programs should return 0 in the case of success, and options like `/version`, `/help`, or `/buildconfig` are usually considered as a success. Another problem is that those exit codes overflows in UNIX. So, it returns -2003 in the mentioned case, however, UNIX limits this to 256, so the result is 45 (= 256 - 2003 % 256). Given the fact that exit codes overflow the range, it may also happen that some errors overflow to 0, which is also unwanted. The xfreerdp man page doesn't provide info about return codes, so this is not necessarily a bug. But there is a convention, that programs should return 0 in the case of success and non-zero otherwise, so it would be nice to honor it. However, I am not sure what would be the right fix here as the error codes are defined in winpr library, which makes me think that the same codes are returned from `mstsc`...

**To Reproduce**
1. `$ xfreerdp /version`
2. `$ echo $?`
3. `45`

**Expected behavior**
According to the conventions it is expected that 0 is returned in case of success and non-zero otherwise.

**Application details**
FreeRDP stable-2.0 branch, but after looking into the codes, I suppose that the master branch has the same issue.

**Environment:**
 - OS: Linux
 - Version/Distribution: Fedora/RHEL

**Additional context**
This was initially reported downstream on https://bugzilla.redhat.com/show_bug.cgi?id=1910029.

ondrejholy:
Given the fact that I can't find some list of `mstsc` exit codes, I assume that this is just freerdp implementation detail and it can be fixed. I've found only https://social.technet.microsoft.com/wiki/contents/articles/37870.remote-desktop-client-troubleshooting-disconnect-codes-and-reasons.aspx, but the mentioned codes are probably unrelated to the exit value and also the codes mentioned in the description are not there.

PR #6741, opened 2021-01-21
client: Fix exit codes for /help and similar option

ondrejholy:
Currently, non-zero exit code is returned for /version, /buildconfig, /help,
/monitor-list, /kbd-list and /kbd-lang-list commandline options for several
clients. This is against conventions, because zero is usually returned in
such cases. Also, there is potentionally another problem that the returned
codes overflow on UNIX systems (where the exit code is a number between 0
and 255). Let's fix the clients to return 0 in the mentioned cases to honor
conventions and 1 for the command-line parsing errors (or -1 for clients
which already use that value).

Fixes: https://github.com/FreeRDP/FreeRDP/issues/6686


client/Sample/tf_freerdp.c
```diff
@@ -338,12 +338,13 @@ int main(int argc, char* argv[])
 		goto fail;
 
 	status = freerdp_client_settings_parse_command_line(context->settings, argc, argv, FALSE);
-	status =
-	    freerdp_client_settings_command_line_status_print(context->settings, status, argc, argv);
-
 	if (status)
 	{
-		rc = 0;
+		freerdp_client_settings_command_line_status_print(context->settings, status, argc, argv);
```
akallabeth, line 343: Since this is always the same code, it might be a good idea to integrate this check in `freerdp_client_settings_command_line_status_print` and return the desired exit code, so you just need `rc = freerdp_client_settings_command_line_status_print(context->settings, status, argc, argv);`

akallabeth (dismissed):
LGTM, but could be made smaller

akallabeth:
@ondrejholy thank you for the pr. I´ve taken the liberty to refactor it a bit so that the changes are consistent for all clients.

akallabeth (approved):

client/Sample/tf_freerdp.c
ondrejholy: I initially made this the way you suggests and then I have changed it to the current version to not break unofficial clients...

ondrejholy:
> @ondrejholy thank you for the pr. I´ve taken the liberty to refactor it a bit so that the changes are consistent for all clients.

But now you should change https://github.com/FreeRDP/FreeRDP/blob/master/client/Mac/cli/AppDelegate.m as well to properly exit in the affected cases.

client/common/cmdline.c
```diff
@@ -1465,15 +1465,19 @@ int freerdp_client_settings_command_line_status_print_ex(rdpSettings* settings,
...
 		}
 
-		return COMMAND_LINE_STATUS_PRINT;
+		goto out;
 	}
 	else if (status < 0)
 	{
 		freerdp_client_print_command_line_help_ex(argc, argv, custom);
-		return COMMAND_LINE_STATUS_PRINT_HELP;
+		status = COMMAND_LINE_STATUS_PRINT_HELP;
```
ondrejholy, line 1473: This is not right as it consequently returns 0 also for command-line parsing errors. It should return 0 only if `status == COMMAND_LINE_STATUS_PRINT_HELP`, shouldn't it?

akallabeth: this is exactly the same as it was before, but will check.

akallabeth:
@ondrejholy the other clients are already handled. `freerdp_client_settings_command_line_status_print_ex` already ignores if the `status` argument is `>=0` and it is called in the mac version as well.

akallabeth (approved):

ondrejholy:
> @ondrejholy the other clients are already handled. `freerdp_client_settings_command_line_status_print_ex` already ignores if the `status` argument is `>=0` and it is called in the mac version as well.

But there is an opposite problem. In Mac client, there is `status = freerdp_client_settings_command_line_status_print(...)` which overwrites the status e.g. from `COMMAND_LINE_STATUS_PRINT_HELP` to `0`, so consequently,  the client won't terminate in `applicationDidFinishLaunching`...

akallabeth:
@ondrejholy ah, that is what you mean. will create a follow up for this :/

PR #6742, opened 2021-01-22
Fixed missing check for fastpath input messages

akallabeth:
Input events are only allowed after a connection was established
(connection state is active)
This check aborts input sending when done before that.

### DIFF
--- a/libfreerdp/core/fastpath.c
+++ b/libfreerdp/core/fastpath.c
@@ -934,8 +934,9 @@ wStream* fastpath_input_pdu_init(rdpFastPath* fastpath, BYTE eventFlags, BYTE ev
 	return s;
 }
 
-BOOL fastpath_send_multiple_input_pdu(rdpFastPath* fastpath, wStream* s, int iNumEvents)
+BOOL fastpath_send_multiple_input_pdu(rdpFastPath* fastpath, wStream* s, size_t iNumEvents)
 {
+	int state;
 	BOOL rc = FALSE;
 	rdpRdp* rdp;
 	UINT16 length;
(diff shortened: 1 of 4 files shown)


libfreerdp/core/fastpath.c
```diff
@@ -944,8 +945,16 @@ BOOL fastpath_send_multiple_input_pdu(rdpFastPath* fastpath, wStream* s, int iNu
 	if (!s)
 		return FALSE;
 
-	if (!fastpath || !fastpath->rdp)
+	if (!fastpath)
+		goto fail;
+
+	state = rdp_client_get_state(rdp);
```
mfleisz, line 951: `rdp` seems to be uninitialized here

mfleisz (changes requested):

mfleisz (approved):
LGTM

krajorama:
Tested OK with Guacamole 1.3.0 + FreeRDP 2.2.0 + this fix backported.

akallabeth:
@krajorama good to know. but be aware, this is just fixing (or hiding) bad application behavior, guacamole needs to fix this properly, as there are other messages as well that are not allowed before a connection is properly established.

krajorama:
@akallabeth Sounds reasonable, but how would Guacamole know when a redirect starts and it needs to stop sending messages? I was trying to find a solution, but all I could find was that I can subscribe to the end of the redirect: PubSub_SubscribeActivated(context->pubSub, my_callback);

Thanks in advance, krajo

akallabeth:
@krajorama that is a reasonable approach. The session state is `ready for use` when that happened (it is basically the same check we do with this patch)

krajorama:
@akallabeth , right I've tested it, but it is NOT the full solution, since I don't know when to stop sending the messages. The scenario is this: I connect to the server, it shows me a welcome screen, I interact with it - so I have to send keyboard/mouse events. Then there is a redirect. The problem before your patch was that if I move my mouse during the redirect, the connection is failed by the server because of protocol violation.

So the question is : how do I know in Guac that I should stop sending messages?

akallabeth:
@krajorama Indeed, there seems to be a missing event that propagates the current connection state. I´ll create a patch for that.

PR #6744, opened 2021-01-22
transport: Also ask for password when connecting using TLS

mfleisz:
This PR adds the possibility to ask the user for his username/password when using TLS secured connections instead of NLA.
Currently the client would not allow the user to enter a username/password upon connection which results in the greeter/login dialog to be shown on the server.

Tested with and without gateway

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -260,6 +260,39 @@ fail:
 	return FALSE;
 }
 
+static BOOL is_empty(const char* str)
+{
+	if (!str)
+		return TRUE;
+	if (strlen(str) == 0)
+		return TRUE;
+	return FALSE;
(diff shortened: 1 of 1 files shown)


libfreerdp/core/transport.c
```diff
@@ -272,6 +305,14 @@ BOOL transport_connect_tls(rdpTransport* transport)
 {
 	if (!transport)
 		return FALSE;
+
+	/* Only prompt for password if we use TLS (NLA also calls this function) */
+	if (transport->settings->SelectedProtocol == PROTOCOL_SSL)
```
akallabeth, line 310: @mfleisz what about `RDP` security?

akallabeth (approved):
LGTM for tls

PR #6748, opened 2021-01-25
Fixed CodeQL warnings

akallabeth:
Mostly integer conversion warnings (according to C standard)


hardening:
I'm always sceptical about changes that make the code uglier just to please code quality checker. It makes me think that the code quality checker has wrong assertions.

akallabeth:
@hardening the assertions are correct here, we rely on undefined behaviour for the casts. this pr ensures the size of the calculation matches the result type.

mfleisz (approved):
Not a big fan of the implicit `* 1ULL` casting but otherwise LGTM

PR #6750, opened 2021-01-25
Only update destination rect for SurfaceToSurface and SurfaceToCache

akallabeth:


mfleisz (approved):
LGTM but can't we just ditch `invalidRect` and simply use `rect`

PR #6752, opened 2021-01-25
rdg websocket support

akarl10:
websocket support for  remotedesktop gateway RDG_OUT_DATA

It automatically falls back to classic RDG_OUT_DATA/RDG_IN_DATA if it is not supported.

This pull request is relatively big because this time client->server flow has to distinguish between chunked encoding and websocket framing.

currently server->client masking is not supported. if the server does this reading does not work anymore (because it is not impemented, but the header is parsed, so security wise this should not be exploitable)
On the other hand rfc 6455 mandates masking from client to server, the other way around is not specified. In fact it is defined that the client has to generate a new masking key per frame (not for the server) so I think it is practically prohibited for a server to do that.
Anyhow, I don't know any server that does this. If one should exists, it will always be possible to implement it.

### DIFF
--- a/libfreerdp/core/gateway/http.c
+++ b/libfreerdp/core/gateway/http.c
@@ -30,6 +30,9 @@
 
 #include <freerdp/log.h>
 
+/* websocket need sha1 for Sec-Websocket-Accept */
+#include <winpr/crypto.h>
+
 #ifdef HAVE_VALGRIND_MEMCHECK_H
 #include <valgrind/memcheck.h>
 #endif
@@ -39,6 +42,8 @@
 #define TAG FREERDP_TAG("core.gateway.http")
 
 #define RESPONSE_SIZE_LIMIT 64 * 1024 * 1024
+
+#define WEBSOCKET_MAGIC_GUID "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"
 
 struct _http_context
 {
@@ -52,6 +57,8 @@ struct _http_context
 	char* Pragma;
 	char* RdgConnectionId;
 	char* RdgAuthScheme;
+	BOOL websocketUpgrade;
+	char SecWebsocketKey[16];
 };
 
 struct _http_request
@@ -77,6 +84,8 @@ struct _http_response
 	size_t ContentLength;
 	const char* ContentType;
 	TRANSFER_ENCODING TransferEncoding;
+	const char* SecWebsocketVersion;
+	const char* SecWebsocketAccept;
 
 	size_t BodyLength;
 	BYTE* BodyContent;
@@ -259,6 +268,30 @@ BOOL http_context_set_rdg_connection_id(HttpContext* context, const char* RdgCon
 	return TRUE;
 }
 
+BOOL http_context_enable_websocket_upgrade(HttpContext* context, BOOL enable)
+{
+	if (!context)
+		return FALSE;
+
+	if (enable)
+	{
+		int i;
+		winpr_RAND((BYTE*)context->SecWebsocketKey, 15);
+		for (i = 0; i < 16; i++)
+			context->SecWebsocketKey[i] = (context->SecWebsocketKey[i] | 0x40) & 0x5f;
+		context->SecWebsocketKey[15] = '\0';
+	}
+	else
+		context->SecWebsocketKey[0] = '\0';
+	context->websocketUpgrade = enable;
+	return TRUE;
+}
+
+BOOL http_context_is_websocket_upgrade_enabled(HttpContext* context)
+{
+	return context->websocketUpgrade;
+}
+
 BOOL http_context_set_rdg_auth_scheme(HttpContext* context, const char* RdgAuthScheme)
 {
 	if (!context || !RdgAuthScheme)
@@ -426,12 +459,25 @@ wStream* http_request_write(HttpContext* context, HttpRequest* request)
 
 	if (!http_encode_header_line(s, request->Method, request->URI) ||
 	    !http_encode_body_line(s, "Cache-Control", context->CacheControl) ||
-	    !http_encode_body_line(s, "Connection", context->Connection) ||
 	    !http_encode_body_line(s, "Pragma", context->Pragma) ||
 	    !http_encode_body_line(s, "Accept", context->Accept) ||
 	    !http_encode_body_line(s, "User-Agent", context->UserAgent) ||
 	    !http_encode_body_line(s, "Host", context->Host))
 		goto fail;
+
+	if (!context->websocketUpgrade)
+	{
+		if (!http_encode_body_line(s, "Connection", context->Connection))
+			goto fail;
+	}
+	else
+	{
+		if (!http_encode_body_line(s, "Connection", "Upgrade") ||
+		    !http_encode_body_line(s, "Upgrade", "websocket") ||
+		    !http_encode_body_line(s, "Sec-Websocket-Version", "13") ||
+		    !http_encode_body_line(s, "Sec-Websocket-Key", context->SecWebsocketKey))
+			goto fail;
+	}
 
 	if (context->RdgConnectionId)
 	{
@@ -556,7 +602,6 @@ static BOOL http_response_parse_header_field(HttpResponse* response, const char*
                                              const char* value)
 {
 	BOOL status = TRUE;
-
 	if (!response || !name)
 		return FALSE;
 
@@ -586,6 +631,20 @@ static BOOL http_response_parse_header_field(HttpResponse* response, const char*
 			response->TransferEncoding = TransferEncodingChunked;
 		else
 			response->TransferEncoding = TransferEncodingUnknown;
+	}
+	else if (_stricmp(name, "Sec-WebSocket-Version") == 0)
+	{
+		response->SecWebsocketVersion = value;
+
+		if (!response->SecWebsocketVersion)
+			return FALSE;
+	}
+	else if (_stricmp(name, "Sec-WebSocket-Accept") == 0)
+	{
+		response->SecWebsocketAccept = value;
+
+		if (!response->SecWebsocketAccept)
+			return FALSE;
 	}
 	else if (_stricmp(name, "WWW-Authenticate") == 0)
 	{
@@ -1040,4 +1099,57 @@ TRANSFER_ENCODING http_response_get_transfer_encoding(HttpResponse* response)
 		return TransferEncodingUnknown;
 
 	return response->TransferEncoding;
+}
+
+BOOL http_response_is_websocket(HttpContext* http, HttpResponse* response)
+{
+	BOOL isWebsocket = FALSE;
+	WINPR_DIGEST_CTX* sha1 = NULL;
+	char* base64accept = NULL;
+	BYTE sha1_digest[WINPR_SHA1_DIGEST_LENGTH];
+
+	if (!http || !response)
+		return FALSE;
+
+	if (!http->websocketUpgrade || response->StatusCode != HTTP_STATUS_SWITCH_PROTOCOLS)
+		return FALSE;
+
+	if (response->SecWebsocketVersion && _stricmp(response->SecWebsocketVersion, "13") != 0)
+		return FALSE;
+
+	if (!response->SecWebsocketAccept)
+		return FALSE;
+
+	/* now check if Sec-Websocket-Accept is correct */
+
+	sha1 = winpr_Digest_New();
+	if (!sha1)
+		goto out;
+
+	if (!winpr_Digest_Init(sha1, WINPR_MD_SHA1))
+		goto out;
+
+	if (!winpr_Digest_Update(sha1, (const BYTE*)http->SecWebsocketKey,
+	                         strlen(http->SecWebsocketKey)))
+		goto out;
+	if (!winpr_Digest_Update(sha1, (const BYTE*)WEBSOCKET_MAGIC_GUID, strlen(WEBSOCKET_MAGIC_GUID)))
+		goto out;
+
+	if (!winpr_Digest_Final(sha1, sha1_digest, sizeof(sha1_digest)))
+		goto out;
+
+	base64accept = crypto_base64_encode(sha1_digest, WINPR_SHA1_DIGEST_LENGTH);
+	if (!base64accept)
+		goto out;
+
+	if (_stricmp(response->SecWebsocketAccept, base64accept) != 0)
+	{
+		WLog_WARN(TAG, "Webserver gave Websocket Upgrade response but sanity check failed");
+		goto out;
+	}
+	isWebsocket = TRUE;
+out:
+	winpr_Digest_Free(sha1);
+	free(base64accept);
+	return isWebsocket;
 }
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -104,6 +104,42 @@
 #define HTTP_CAPABILITY_REAUTH 0x10
 #define HTTP_CAPABILITY_UDP_TRANSPORT 0x20
 
+#define WEBSOCKET_MASK_BIT 0x80
+#define WEBSOCKET_FIN_BIT 0x80
+
+typedef enum _WEBSOCKET_OPCODE
+{
+	WebsocketContinuationOpcode = 0x0,
+	WebsocketTextOpcode = 0x1,
+	WebsocketBinaryOpcode = 0x2,
+	WebsocketCloseOpcode = 0x8,
+	WebsocketPingOpcode = 0x9,
+	WebsocketPongOpcode = 0xa,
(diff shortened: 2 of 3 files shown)


libfreerdp/core/gateway/http.c
```diff
@@ -52,6 +55,8 @@ struct _http_context
 	char* Pragma;
 	char* RdgConnectionId;
 	char* RdgAuthScheme;
+	BOOL websocketUpgrade;
+	char* SecWebsocketKey;
```
akallabeth, line 59: Use a static size here as the length is fixed 16 char

libfreerdp/core/gateway/http.c
```diff
@@ -1041,3 +1099,54 @@ TRANSFER_ENCODING http_response_get_transfer_encoding(HttpResponse* response)
 
 	return response->TransferEncoding;
 }
+
+BOOL http_response_is_websocket(HttpContext* http, HttpResponse* response)
+{
+	if (!http || !response)
+		return FALSE;
+	if (http->websocketUpgrade == TRUE && response->StatusCode == HTTP_STATUS_SWITCH_PROTOCOLS &&
```
akallabeth, line 1107: Better use braces around the conditions, otherwise this is hard to read.

libfreerdp/core/gateway/http.c
```diff
@@ -1041,3 +1099,54 @@ TRANSFER_ENCODING http_response_get_transfer_encoding(HttpResponse* response)
...
+		base64accept = crypto_base64_encode(sha1_digest, WINPR_SHA1_DIGEST_LENGTH);
+		if (!base64accept)
+			goto out;
+
+		if (_stricmp(response->SecWebsocketAccept, base64accept) != 0)
+		{
+			WLog_WARN(TAG, "Webserver gave Websocket Upgrade response but sanity check failed");
+			goto out;
+		}
+		winpr_Digest_Free(sha1);
```
akallabeth, line 1142: Duplicated code, use a return variable defaulting to `FALSE` set to `TRUE` here instead

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,299 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+	int streamPos;
+
+	s = Stream_Length(sPacket);
+	inStream = Stream_Buffer(sPacket);
+	if (s > INT_MAX)
+		return FALSE;
+
+	if (s < 126)
+		sWS = Stream_New(NULL,
+		                 Stream_Length(sPacket) + 6); // 2 byte "mini header" + 4 byte masking key
```
akallabeth, line 380: Duplicated, you already have the size in `s`

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,299 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+	else if (s < 0x10000)
+		sWS = Stream_New(NULL, Stream_Length(sPacket) +
```
akallabeth, line 382: Duplicated, you already have the size in `s`

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,299 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+		                           8); // 2 byte "mini header" + 2 byte length + 4 byte masking key
+	else
+		sWS = Stream_New(NULL, Stream_Length(sPacket) +
```
akallabeth, line 385: Duplicated, you already have the size in `s`

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,299 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+		Stream_Write_UINT8(sWS, 127 | WEBSOCKET_MASK_BIT);
+		Stream_Write_UINT32(sWS, 0); // payload is limited to INT_MAX
+		Stream_Write_UINT32(sWS, htonl(s));
+	}
+	Stream_Write_UINT32(sWS, maskingKey);
+
+	// mask as much as possible with 32bit access
+	for (streamPos = 0; streamPos + 4 <= s; streamPos += 4)
+	{
+		uint32_t masked = *((uint32_t*)(inStream + streamPos)) ^ maskingKey;
```
akallabeth, line 412: Never cast data from network directly, use `Stream_Read_*`

akallabeth (changes requested):

libfreerdp/core/gateway/http.c
```diff
@@ -1041,3 +1097,54 @@ TRANSFER_ENCODING http_response_get_transfer_encoding(HttpResponse* response)
 
 	return response->TransferEncoding;
 }
+
+BOOL http_response_is_websocket(HttpContext* http, HttpResponse* response)
+{
+	if (!http || !response)
+		return FALSE;
+	if ((http->websocketUpgrade == TRUE) &&
```
hardening, line 1105: I'd suggest to invert the condition to do some early exits, that would make the code easier to read.
```C
if (!http->websocketUpgrade || response->StatusCode != HTTP_STATUS_SWITCH_PROTOCOLS)
   return FALSE;
```
and so on

libfreerdp/core/gateway/http.c
```diff
@@ -1041,3 +1097,54 @@ TRANSFER_ENCODING http_response_get_transfer_encoding(HttpResponse* response)
...
+		if (!sha1)
+			goto out;
+
+		if (!winpr_Digest_Init(sha1, WINPR_MD_SHA1))
+			goto out;
+
+		if (!winpr_Digest_Update(sha1, (const BYTE*)http->SecWebsocketKey,
+		                         strlen(http->SecWebsocketKey)))
+			goto out;
+		if (!winpr_Digest_Update(sha1, (const BYTE*)"258EAFA5-E914-47DA-95CA-C5AB0DC85B11",
```
hardening, line 1126: A define for the GUID would be nice

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,313 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+{
+	size_t s;
+	int status;
+	wStream* sWS;
+
+	uint32_t maskingKey;
+
+	int streamPos;
+
+	s = Stream_Length(sPacket);
```
hardening, line 372: would be nice to have a variable with another name, we usually use `s` for `stream`, not for `size` (len ?).

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,313 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+	if (s > INT_MAX)
+		return FALSE;
+
+	if (s < 126)
+		sWS = Stream_New(NULL, s + 6); // 2 byte "mini header" + 4 byte masking key
+	else if (s < 0x10000)
+		sWS = Stream_New(NULL, s + 8); // 2 byte "mini header" + 2 byte length + 4 byte masking key
+	else
+		sWS = Stream_New(NULL, s + 14); // 2 byte "mini header" + 8 byte length + 4 byte masking key
+
```
hardening, line 384: Just nitpicking, but perhaps the line above could just compute an allocation size and the effective `Stream_New` would be done here.

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,313 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+	else
+	{
+		Stream_Write_UINT8(sWS, 127 | WEBSOCKET_MASK_BIT);
+		Stream_Write_UINT32_BE(sWS, 0); // payload is limited to INT_MAX
+		Stream_Write_UINT32_BE(sWS, s);
+	}
+	Stream_Write_UINT32(sWS, maskingKey);
+
+	// mask as much as possible with 32bit access
+	for (streamPos = 0; streamPos + 4 <= s; streamPos += 4)
```
hardening, line 407: Doing the comment once just here, but we avoid C++ comments because some older compilers don't like them (so if you could drop them in the rest of the code)

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,313 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+		return FALSE;
+
+	return TRUE;
+}
+
+static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
+{
+	if (rdg->transferEncoding.isWebsocketTransport)
+		return rdg_write_websocket(rdg, sPacket);
+	else
```
hardening, line 438: `else` not needed

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,313 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+			case WebSocketStateMaskingKey:
+			{
+				WLog_WARN(
+				    TAG, "Websocket Server sends data with masking key. This is against RFC 6455.");
+				return -1;
+			}
+			break;
+			case WebSocketStatePayload:
+			{
+				uint8_t effectiveOpcode =
```
hardening, line 519: I'd suggest this state to have its own treatment function as it's quite big right now

akallabeth:
@freerdp-bot test

akarl10:
sorry, don't know what exactly the oldest toolchain can do...
hope this one is compatible (uint8_t to BYTE, variable declaration removed in for loop)

akallabeth:
@freerdp-bot test
@akarl10 best stick to C99 rules (except for `stdint.h` and variable declaration at beginning of scope)

akallabeth:
@freerdp-bot test

akallabeth:
@akarl10 is the pr now ready for testing?

akarl10:
if you are ok with the code quality I would say yes.
Personally I have tested it against  RDG 2016 directly, RDG behind nginx* with websocket upgrade, without websocket upgrade with and without chunked transfer encoding, but only running on a recent os (Ubuntu 20.10).

* the nginx tests used only PAA authentication, but in that regard it should not make any difference.

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,346 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+		}
+		break;
+		case WebsocketPingOpcode:
+		{
+			status = rdg_websocket_read_discard(bio, encodingContext);
+			if (status < 0)
+				return status;
+
+			if (encodingContext->payloadLength == 0)
+				rdg_websocket_reply_pong(bio);
```
MartinHaimberger, line 557: According to the rfc section  5.5.3 (https://tools.ietf.org/html/rfc6455#section-5.5.3) the pong answer must have identical "Application data" if there is "Application data" included in the ping frame.

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,346 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+static BOOL rdg_write_websocket(rdpRdg* rdg, wStream* sPacket)
+{
+	size_t len;
+	size_t fullLen;
+	int status;
+	wStream* sWS;
+
+	uint32_t maskingKey;
+
+	int streamPos;
```
akallabeth, line 371: type error, should be `size_t`

akallabeth (changes requested):

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,346 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
+			break;
+			case WebSocketStatePayload:
+			{
+				status = rdg_websocket_handle_payload(bio, pBuffer, size, encodingContext);
+				if (status < 0)
+					return (effectiveDataLen > 0 ? effectiveDataLen : status);
+
+				effectiveDataLen += status;
+
+				if (status == size)
```
akallabeth, line 664: this requires a cast to `(size_t)status`

libfreerdp/core/gateway/rdg.c
```diff
@@ -323,34 +359,346 @@ static BOOL rdg_write_packet(rdpRdg* rdg, wStream* sPacket)
...
 				{
-					encodingContext->context.chunked.state = ChunkStateFooter;
-					encodingContext->context.chunked.headerFooterPos = 0;
+					encodingContext->state = ChunkStateFooter;
+					encodingContext->headerFooterPos = 0;
 				}
 				effectiveDataLen += status;
 
-				if ((size_t)status == size)
+				if (status == size)
```
akallabeth, line 701: this requires a cast to `(size_t) status`

libfreerdp/core/gateway/rdg.c
```diff
@@ -409,16 +406,18 @@ static BOOL rdg_write_websocket(rdpRdg* rdg, wStream* sPacket)
 	// mask as much as possible with 32bit access
 	for (streamPos = 0; streamPos + 4 <= s; streamPos += 4)
 	{
-		uint32_t masked = *((uint32_t*)(inStream + streamPos)) ^ maskingKey;
-		Stream_Write_UINT32(sWS, masked);
+		uint32_t data;
+		Stream_Read_UINT32(sPacket, data);
```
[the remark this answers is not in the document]
akarl10: it must be "same endian" as where the masking key is transmitted a few lines above:
Stream_Write_UINT32(sWS, maskingKey);
since maskingKey itself is random it does not really matter for the actual "number" itself

If you want I can change both to UNINT32_BE

libfreerdp/core/gateway/rdg.c
akarl10: I will look into this. Do you have a way to test this? Afaik MS does not send pings, so introducing a buffer for this requires someone to test it or at least confirm that it will work (probably more that 4 eyes looking on this).
https://tools.ietf.org/html/rfc6455#section-5.5.1 (close) does not mandate relaying back application data.

libfreerdp/core/gateway/rdg.c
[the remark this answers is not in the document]
akallabeth: if it is by intention it is ok, I was just curious (did not want to dig too deep) if network byte order is required or not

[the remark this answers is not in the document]
akarl10: side note: to me it looks more readably in little endian because a few lines down there is this:

BYTE* partialMask = ((BYTE*)&maskingKey) + (streamPos % 4);

this is essentially litte endian.

libfreerdp/core/gateway/rdg.c
MartinHaimberger: Hi, the rfc just states that it would be typically to echo the status code of the close:
`(When
   sending a Close frame in response, the endpoint typically echos the
   status code it received.) `

akallabeth:
@freerdp-bot test

libfreerdp/core/gateway/rdg.c
akarl10: I did a Test with a quick and got it to echo the ping/pong back (by patching a simple websocket server/gateway to send a ping immediately after the websocket upgrade). Hope you like the way I implemented the buffer handling.
I even implemented relaying back the status code of the close frame and enforced that no data is sent after that.

MartinHaimberger: @akarl10 very nice job!

akallabeth (approved):
LGTM

MartinHaimberger:
LGTM

akarl10:
@akallabeth  should I squash commit or is it ok like this?

akarl10:
I have rebased and squashed commited.
The view on github shows a lot more changes. Is this normal?
checked with git diff b05fea134e0244d0a9f67941e2385229b267ab5c (current head) and it shows only my changes..

akallabeth:
@akarl10 ok, sorry, you did something wrong. rebase to current master and check the diff.
Please also note that now we need to review again :/

akarl10:
now something got really wrong..
https://github.com/FreeRDP/FreeRDP/compare/master...akarl10:websocket
shows up ok, but this page now sometimes gives me a 500 server error.
additionally the update does not show up here...
What should I do now? close this an do a new pull request?

akallabeth:
@akarl10 just fetch and rebase to master, I´ve tried that and it was not on based on current master.
github being a bit unreliable is something that happens sometimes ;)

akarl10:
ok, should be fixed now.
My apologies about that.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #6755, opened 2021-01-26
Added OrderInfo callback for primary, secondary and altsec orders

akallabeth:
This allows gathering statistics about which orders have been
received


hardening (approved):
LGTM

Issue #6388, opened 2020-07-23
Scrolling and keyholding behaviour 


kayfoe:
We use Gentoo and the Wayland version of FreeRDP and there are two bug that we found.
They only appear in Wayland and not in X.
1. Scrolling:
Even it got a lot better so far, the scrolling still is not like in regular Windows or with the X version of FreeRDP.
It is just too slow compared to Windows scrolling.

2. Keyholding:
When you press a key and hold it then normally it will repeat the letter over and over again, but with wlfreerdp they just don't.
It is just one single letter for a single press, no matter how long you hold the key.

We really like Your product so far and hope it will get even better.
Sincerly Kay

Edit:
We also tried it with Ubuntu to make sure that it's not a Gentoo problem but the bugs still occurred.

ReubenM:
Adding myself to follow changes to this.

Interacting with a text terminal on remote windows box using wayland is awful compared to X11.

compilenix:
Same here on Fedora 33 (`freerdp-2.2.0-4.fc33`).

manio:
Only my two cents:
I also had two described problems: "slow" scrolling and lack of key repeat on wayland+swaywm, but after recently update on my Arch distro to package:
```freerdp 2:2.3.0-1```
then those two problems are gone, so maybe this issue is now obsolete?

PR #6758, opened 2021-01-26
Fix key repeat in Wayland

0140454:
## Description

Without this commit, key repeat doesn't work.
When I press space key and hold it, there is only one space character in editor.

After applying this patch, it will repeat the characters as the expected.

## Test Environment

* Arch Linux
* KDE Plasma 5.20.5

## Related Issues

* Scrolling and keyholding behaviour (#6388)


hardening (approved):
LGTM

hardening:

Issue #6759, opened 2021-01-26
Running on Windows, can't connect to a Windows server which uses per-device RDS licensing/CALs


vscfreire:
**Describe the bug**
When using FreeRDP on Windows (Windows 10), and connecting to a Windows Server (2016) with per-device RDS licensing/CAL, the connection fails. The only error is `ERRCONNECT_CONNECT_TRANSPORT_FAILED`.

On the Windows server side, it even crashes some components ("The Windows logon process has unexpectedly terminated.").

On macOS and Linux, this behavior is not reproducible, everything is working fine.

The problem is primarily located in the `saveCal()` function on the `libfreerdp/core/license.c` file (see details below).

**To Reproduce**
1. Configure per-device CALs on the Windows Server RDS;
2. Connect to the server using a client which uses FreeRDP on Windows end-user device.
3. Can't login.

**Expected behavior**
When using per-device CALs, FreeRDP is supposed to receive the license/CAL information and store it as a file on the `ConfigDir/licenses` directory (on Windows: `%APPDATA%\freerdp\licenses`).

It first creates a file called `[long hex string].cal.new` where it stores the data it receives from the server, and then renames it to `[long hex string].cal` (removing the `.new` suffix).

It later uses this file to retrieve licensing information to use in the login process somehow (haven't investigated this part of the process).

**Screenshots**
n/a

**Application details**
* FreeRDP version: from the repository, branch stable-2.0
* OS version connecting to: Windows Server 2016

**Environment (please complete the following information):**
 - OS: Windows only (not on macOS/Linux, not tested on Android)
 - Version/Distribution: Windows 10
 - Architecture: amd64

**Additional context**
As stated above, the culprit is the `saveCal()` function on the `libfreerdp/core/license.c` file.

This function builds the path for the temporary and final file names by calling `GetCombinedPath()` on several occasions.

It then writes to the temporary file (with the `.new` suffix) and finally calls `MoveFileEx()` to rename the file from the temporary name to the final one.

On Linux/macOS this works fine, because `MoveFileEx()` is actually defined to be `MoveFileExA()` (ANSI version of the function).

[...]

PR #6761, opened 2021-01-27
Fixed #6760: Use wide char file operations on windows.

### LINKED ISSUE #6760: Move cliprdr_parse_file_list() outside of freerdp-client3 library
The function `cliprdr_parse_file_list()` allows the conversion of a packet file list into an array of `FILEDESCRIPTORW`s and is useful for both server- and client- implementations of the `CLIPRDR` channel.

However, the function is only available when the implementation links against the `freerdp-client` library, which is inconvenient for server implementations, as linking against the `freerdp-client` library should only be needed for client implementations.

To solve this situation, this function should be moved outside of `freerdp-client` into the `freerdp` library for FreeRDP 3.0 (current master), so neither the linkage against `freerdp-client` or `freerdp-server` library is necessary (, as this function is useful for both server side and client side `CLIPRDR` implementations).

akallabeth:


mfleisz:
#6760 references a clipboard channel issue and looks unrelated to this PR - I guess this is supposed to fix #6759

mfleisz (commented):
Fix LGTM - but what is that `fopen_wrap` function actually for?

akallabeth:
@mfleisz to open files with unicode names in it (windows only)

mfleisz:
Just wondering why this is necessary:
- `PathFileExistsA` is used to check `licenseStorePath` for existance
- The appended file name (`"hash_of_hostname.cal"`) won't contain any chars that require Unicode.

hardening (dismissed):

akallabeth:
@mfleisz ok, fixed the remaining non wide char calls.
the end of the path may not contain unicode, but the username et al might (which is part of the path)

pnowack:
> Fixed #6760: Use wide char file operations on windows.

Typo: This should probably be https://github.com/FreeRDP/FreeRDP/issues/6759, not https://github.com/FreeRDP/FreeRDP/issues/6760

mfleisz (approved):
LGTM

PR #6763, opened 2021-01-27
gateway: Fix consent/service message in legacy gateway code

mfleisz:
Missed a place for handling messages in the legacy gateway code when introducing the `PresentGatewayMessage` callback.
Also fixed an ignored return value of `rpc_client_recv_pdu`.


akallabeth (approved):
LGTM

PR #6765, opened 2021-01-27
Multiply axis value by 8 in Wayland

0140454:
## Description

When using wlfreerdp, the number of scrolling rows is less than either
xfreerdp or other window.

In KDE Wayland session, I get a default axis value, which is 15.

And according to the comment in issue #6087, we have a default step
value in X11, which is 120.

To make wlfreerdp and xfreerdp behave identically, we should multiply
axis value by 8.

## Test Environment

* Arch Linux
* KDE Plasma 5.20.5

## Related Issues

* Inconsistent mouse wheel scroll behavior on Wayland (#6087)
* Scrolling and keyholding behaviour (#6388)

### DIFF
--- a/client/Wayland/wlf_input.c
+++ b/client/Wayland/wlf_input.c
@@ -176,7 +176,7 @@ BOOL wlf_handle_pointer_axis(freerdp* instance, const UwacPointerAxisEvent* ev)
 	 * positive: 0 ... 0xFF  -> slow ... fast
 	 * negative: 0 ... 0xFF  -> fast ... slow
 	 */
-	step = abs(direction);
+	step = 8 * abs(direction);
 	if (step > 0xFF)
 		step = 0xFF;
 


client/Wayland/wlf_input.c
```diff
@@ -176,7 +176,7 @@ BOOL wlf_handle_pointer_axis(freerdp* instance, const UwacPointerAxisEvent* ev)
 	 * positive: 0 ... 0xFF  -> slow ... fast
 	 * negative: 0 ... 0xFF  -> fast ... slow
 	 */
-	step = abs(direction);
+	step = 8 * abs(direction);
```
akallabeth, line 179: If you do this it would be a good idea to replace `direction = wl_fixed_to_int(ev->value);` with `direction = wl_fixed_to_double(ev->value);` in code above.

akallabeth:
@0140454 thank you for the pull request.
I´ve checked wayland documentation and found the reason for the inconsistent behaviour and fixed it in #6766 
(that one matches the behaviour of the `xfreerdp` version exactly as it listens to discrete events and uses the same conversion value)

PR #6766, opened 2021-01-27
Added wayland discrete axis events for mouse wheel

akallabeth:
the discrete axis event gives changes in steps just like the
xfreerdp version uses. This way scrolling can be implemented
consistent with the behaviour of xfreerdp


hardening (approved):
LGTM

0140454:
When only using discrete value, the problem I encountered is scrolling will not work if I decrease `Scrolling speed` in KDE system settings.

With level 5, discrete value is 1.
Screenshot_20210128_090537

However, discrete event will not be emitted if level is 4.
Screenshot_20210128_090647

akallabeth:
@0140454 which values do you get with the axis event in these cases? this might actually also be an issue with kde (did they already finish their wayland support?)

0140454:
@akallabeth 

| Level | Axis | Discrete |
| :-----: | :----: | :---------: |
| 1 | 384 | No event |
| 2 | 1152 | No event |
| 3 | 1920 | No event |
| 4 | 2880 | No event |
| 5 | 3840 | 1 |
| 6 | 5760 | 1 |
| 7 | 7680 | 2 |
| 8 | 11520 | 3 |
| 9 | 15360 | 4 |
| 10 | 19200 | 5 |
| 11 | 26880 | 7 |
| 12 | 34560 | 9 |
| 13 | 46080 | 12 |
| 14 | 57600 | 15 |
| 15 | 76800 | 20 |

According to [news](https://www.omgubuntu.co.uk/2021/01/kde-plans-for-2021-wayland-fingerprint) about KDE Wayland support, I think they are still working on it.

PR #6767, opened 2021-01-28
Split netchar autodetect from autodetect setting

akallabeth:
it can be used independently from the connection autodetect setting
and not setting it might trigger bugs in windows RDP servers


mfleisz:
I think we should just set `RNS_UD_CS_SUPPORT_NETCHAR_AUTODETECT` when `NetworkAutoDetect` is true (and also do the protocol checks against that setting).
Also we should get rid of the connection type overrides in `gcc_write_client_core_data` in order to behave like mstsc.
mstsc always announces support for `RNS_UD_CS_SUPPORT_NETCHAR_AUTODETECT` unless it is disabled in an rdp file (using the `networkautodetect:i:1` setting). It does not modify the connection type - no matter what setting is chosen for `NetworkAutoDetect` (unlike freerdp does).

PR #6779, opened 2021-02-03
libfreerdp: core: add mem checks in gcc_write_client_data_blocks

kubistika:
Fixes a heap overflow that may occur when connecting with 3+ screens (which makes the packet larger than the memory allocated for it). There were some missing memory checks.

There might still be more missing checks there, as that part of code usually uses void functions without checks/error handling to perform these write operations. So if someone has time to dig in and see if there are some more missing checks, it's better.


mfleisz (approved):
LGTM

hardening (approved):
LGTM, nice job.

PR #6787, opened 2021-02-04
Synchronize NumLock, CapsLock state in Wayland

0140454:
According to [Wayland Procotol Spec](https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_keyboard-event-enter), `enter` event is followed by `modifiers` event.

Therefore, I make wlfreerdp send FocusIn event with toggle keys state to remote in the first `modifiers` event after `enter` event.
And the subsequent `modifiers` events will do nothing.


akallabeth (dismissed):
LGTM

akallabeth:
@freerdp-bot test

hardening (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #6789, opened 2021-02-05
Use separate codec context for GDI/SurfaceBits and GFX channel

akallabeth:
GFX and SurfaceBits commands share a common codec backend.
Recreating all codec context on GFX initialization will break
clients that are decoding SurfaceBits codec while having the GFX
channel initialized.

#6620 was the cause of this, probably a fix for the proxy is required.

### DIFF
--- a/client/X11/xf_gfx.c
+++ b/client/X11/xf_gfx.c
@@ -253,7 +253,7 @@ static UINT xf_CreateSurface(RdpgfxClientContext* context,
 	if (!surface)
 		return CHANNEL_RC_NO_MEMORY;
 
-	surface->gdi.codecs = gdi->context->codecs;
+	surface->gdi.codecs = context->codecs;
 
 	if (!surface->gdi.codecs)
 	{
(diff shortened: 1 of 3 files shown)


akallabeth:
@kubistika you committed the change that did always reset the codecs, do you remember why?

kubistika:
Yes I did it because that's the behavior of MS clients. If you're reverting this commit, I'll be glad if you at least add public api to reset the codecs.

akallabeth:
@kubistika is this true for all codecs? the problem arises when connecting with e.g. `xfreerdp /v:host /rfx /gfx` where there might be a gfx channel open but still `SurfaceBits` data sent.

kubistika:
I will need to check that. Currently on vacation :/

akallabeth:
@kubistika ok, but vacation is no reason to be sad ;)

akallabeth:
@kubistika ok, did find a way to satisfy both requirements.
Just use a separate codec context for GFX so we have no issues with interferance

mfleisz (approved):
LGTM

PR #6792, opened 2021-02-05
Smartcard tests and fixes

akallabeth:
Changed:
1. Fixed broken call decoding (`GetAttr`, ...)
2. Refactored decoding, split structs off to own header
3. Simplified handle/context decoding, done for all calls the same way now (all structs use a common base)
4. Initialized return structs properly
5. Allow empty NDR pointers for calls 

With this pull request smartcard:
1. Runs https://github.com/akallabeth/scard-tester without crashes and proper return codes
2. works with a yubikey neo `certutil -v -scinfo`
3. Tested with a `Datev` smartcard and run `Gesamtbasistest` (thank you @nfedera )


mfleisz (approved):
No smartcard to try this but code wise lgtm

PR #6797, opened 2021-02-08
Added ConnectionStateChangeEvent for clients

akallabeth:
Clients can now subscribe to connection state change events to
prevend data from being transmitted on sessions being redirected.
See #6742

### DIFF
--- a/include/freerdp/event.h
+++ b/include/freerdp/event.h
@@ -71,6 +71,11 @@ extern "C"
 	BOOL firstActivation;
 	DEFINE_EVENT_END(Activated)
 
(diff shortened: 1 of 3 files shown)


krajorama:
Thanks Armin, original reporter here: the workaround in PR #6742 is good enough at the moment , due to other work I'll get back to trying this #6797 probably in April.

mfleisz (approved):
LGTM
The `active` event parameter seems redundant as it is indicates if `state == CONNECTION_STATE_ACTIVE`

akallabeth:
@mfleisz yes, but it is an easy check without magic numbers as currently the state enum is not exposed.

PR #6804, opened 2021-02-10
Fixed #6801: Delay sending of resolution change

### LINKED ISSUE #6801: Testing FreeRDP /dynamic-resolution compatibility with an XRDP prototype, need debugging assistance.
Yes. I've been working on implementing resize-on-the-fly (/dynamic-resolution) compatibility in XRDP: https://github.com/neutrinolabs/xrdp/issues/448

I have tested my branch here: https://github.com/Nexarian/xrdp/commits/devel with the new Microsoft Remote Desktop clients on Windows 10 (Note, not MSTSC, the version installed from the App Store) and the Mac OS client, and it works great! Dynamic resizing is close at hand with XRDP!

Unfortunately, it does not seem to work with the latest versions of FreeRDP. I've tested this on both Mac OS ->Ubuntu as well as pure Ubuntu (Two 20.04.2 VMs connecting to each other hosted on VirtualBox on Windows 10 Pro)

The first problem is a weird "oscillation" of FreeRDP -- The window will resize to what I resized the screen to, then flip back to what it was before I resized the window, and then back and forth. It'll either do this forever, or eventually disconnect.

Otherwise, after login, the screen may resize a couple times, or it may oscillate. Either way, you only get a handful of resize attempts before the client disconnects after a resize attempt. Sometimes, however, it does succeed, though it's not common.

I'd like to work with the maintainers of FreeRDP to diagnose whether or not this is something that needs to be corrected on the XRDP side or if it needs to be fixed on the FreeRDP side before XRDP commits this feature.

Another option, since this works with two different Microsoft clients reliably, is to merge this into XRDP and then leave the debugging of the compatibility of `/dynamic-resolution`

**Additional context**
I've noticed that this issue is a lot less stable when I use `/sec-rdp` -- So I suspect this has something to do with the reliability of the network socket connections here. Possibly a threading issue

Also note that the XRDP session is undamaged by this. I can reconnect with either FreeRDP or the Microsoft Remote Desktop clients.

Here is a log of a recent attempt: [FreeRDP debug log.txt](https://github.com/FreeRDP/FreeRDP/files/5949044/FreeRDP.debug.log.txt)

akallabeth:
Changed the logic of the disp channel to wait for 800ms after a
ConfigureNotify before sending the new resolution.
The problem fixed with this patch is the following:
1.  Resize the window with the mouse
2.  ConfigureNotify triggers a resize notification
3.  The server resizes to the desired resolution
3a. More ConfigureNotify events are generated
4.  The local window resize to the new resolution triggers another
    ConfigureNotify

a. Depending on the timing (sending is already rate limited) the
   events from 3a and 4 will make the size of the window jump
b. Very fast resizing will pick a random resolution from the
   sequence of ConfigureNotify events as the final resolution

### DIFF
--- a/client/X11/xf_disp.c
+++ b/client/X11/xf_disp.c
@@ -46,7 +46,6 @@ struct _xfDispContext
 	UINT64 lastSentDate;
 	int targetWidth, targetHeight;
 	BOOL activated;
-	BOOL waitingResize;
 	BOOL fullscreen;
 	UINT16 lastSentDesktopOrientation;
 	UINT32 lastSentDesktopScaleFactor;
@@ -125,7 +124,6 @@ static BOOL xf_disp_sendResize(xfDispContext* xfDisp)
 	}
 	else
 	{
-		xfDisp->waitingResize = TRUE;
 		layout.Flags = DISPLAY_CONTROL_MONITOR_PRIMARY;
 		layout.Top = layout.Left = 0;
 		layout.Width = xfDisp->targetWidth;
@@ -142,6 +140,16 @@ static BOOL xf_disp_sendResize(xfDispContext* xfDisp)
 	}
 
 	return xf_update_last_sent(xfDisp);
+}
+
+static BOOL xf_disp_queueResize(xfDispContext* xfDisp, UINT32 width, UINT32 height)
+{
+	if ((xfDisp->targetWidth == width) && (xfDisp->targetHeight == height))
+		return TRUE;
+	xfDisp->targetWidth = width;
+	xfDisp->targetHeight = height;
+	xfDisp->lastSentDate = GetTickCount64();
+	return xf_disp_sendResize(xfDisp);
 }
 
 static BOOL xf_disp_set_window_resizable(xfDispContext* xfDisp)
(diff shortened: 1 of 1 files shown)


akallabeth:
@Nexarian I´ve found the issue it seems. While this is no way a perfect fix (just keep the mouse button pressed and observe some resizing every ~800ms) it mitigates the resize loop by waiting until the window size is stable

Nexarian (approved):

Nexarian:
I don't see the change to make it around 800 ms? The refactor looks good, but curious where the timing change is.

akallabeth:
@Nexarian the `xfDisp->lastSentDate = GetTickCount64();` line when setting a new resolution to send resets the timestamp to the current one and the logic needs to wait until the `RESIZE_MIN_DELAY` has passed. Just also recoginized, that is `200ms` not the `800ms` I assumed, need to check where I got that number from

Nexarian:
@akallabeth This change is overall a massive improvement. Without the change of RESIZE_MIN_DELAY to 800 MS I still got some oscillations and some disconnects under heavy resizing conditions. However, it's much, much more stable.

With the resize to 800 MS it's not perfect but it's very very close. Check this in with my blessing, either with or without the 800 MS change, and this will definitely help.

I would also encourage you to take it for a spin with my XRDP branch yourself, but up to you!

mfleisz (approved):
LGTM

PR #6805, opened 2021-02-10
client/X11: Use the correct wheel rotation value

pnowack:
For the negative scrolling direction, RDP uses the two's complement,
instead of the positive wheel value with the negative flag.
xfreerdp currently uses the positive wheel value in addition to the
negative flag, which results in a wrong wheel value on the server side
(136 instead of 120).

Fix this, by using the correct wheel rotation value, which is in the
two's complement.


pnowack:
(This should be backported to the stable-2.0 branch.)

akallabeth:
@pnowack thank you, I´ve updated your commit for the other clients too, so closing this.

PR #6806, opened 2021-02-11
clients: Use the correct wheel rotation value

akallabeth:
For the negative scrolling direction, RDP uses the two's complement,
instead of the positive wheel value with the negative flag.
xfreerdp currently uses the positive wheel value in addition to the
negative flag, which results in a wrong wheel value on the server side
(136 instead of 120).

Fix this, by using the correct wheel rotation value, which is in the
two's complement.

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -1076,8 +1076,9 @@ static const button_map xf_button_flags[NUM_BUTTONS_MAPPED] = {
 	{ Button2, PTR_FLAGS_BUTTON3 },
 	{ Button3, PTR_FLAGS_BUTTON2 },
 	{ Button4, PTR_FLAGS_WHEEL | 0x78 },
-	{ Button5, PTR_FLAGS_WHEEL | PTR_FLAGS_WHEEL_NEGATIVE | 0x78 },
-	{ 6, PTR_FLAGS_HWHEEL | PTR_FLAGS_WHEEL_NEGATIVE | 0x78 },
+	/* Negative value is 9bit twos complement */
+	{ Button5, PTR_FLAGS_WHEEL | PTR_FLAGS_WHEEL_NEGATIVE | (0x100 - 0x78) },
+	{ 6, PTR_FLAGS_HWHEEL | PTR_FLAGS_WHEEL_NEGATIVE | (0x100 - 0x78) },
 	{ 7, PTR_FLAGS_HWHEEL | 0x78 },
 	{ 8, PTR_XFLAGS_BUTTON1 },
 	{ 9, PTR_XFLAGS_BUTTON2 },
--- a/client/Mac/MRDPView.h
+++ b/client/Mac/MRDPView.h
@@ -72,16 +72,6 @@
 
 @end
 
-/* Pointer Flags */
-#define PTR_FLAGS_WHEEL 0x0200
-#define PTR_FLAGS_WHEEL_NEGATIVE 0x0100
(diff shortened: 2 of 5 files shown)


client/X11/xf_client.c
```diff
@@ -1076,8 +1076,9 @@ static const button_map xf_button_flags[NUM_BUTTONS_MAPPED] = {
 	{ Button2, PTR_FLAGS_BUTTON3 },
 	{ Button3, PTR_FLAGS_BUTTON2 },
 	{ Button4, PTR_FLAGS_WHEEL | 0x78 },
-	{ Button5, PTR_FLAGS_WHEEL | PTR_FLAGS_WHEEL_NEGATIVE | 0x78 },
-	{ 6, PTR_FLAGS_HWHEEL | PTR_FLAGS_WHEEL_NEGATIVE | 0x78 },
+	{ Button5, PTR_FLAGS_WHEEL | PTR_FLAGS_WHEEL | 0x78 },
```
pnowack, line 1079: This is now wrong. It should either be `{ Button5, PTR_FLAGS_WHEEL |  0x188 }` or `{ Button5, PTR_FLAGS_WHEEL | 0x100 - 0x78 }`.
What you did here is duplicating the `PTR_FLAGS_WHEEL` flag and ensuring that Button5 now behaves the same as Button 4, which should not be the case (it should be the opposite direction).

client/X11/xf_client.c
```diff
@@ -1076,8 +1076,9 @@ static const button_map xf_button_flags[NUM_BUTTONS_MAPPED] = {
...
+	/* Negative value is 9bit twos complement */
+	{ 6, PTR_FLAGS_HWHEEL | PTR_FLAGS_WHEEL | (0x100 - 0x78) },
```
pnowack, line 1081: This is now also wrong: The documentation says:
```
The only valid flags in a horizontal wheel rotation event are PTRFLAGS_WHEEL_NEGATIVE and the
WheelRotationMask
```
It also mentions:
```
If both PTRFLAGS_WHEEL and PTRFLAGS_HWHEEL are specified, then PTRFLAGS_WHEEL takes
precedence.
```

So, you need to remove the `PTR_FLAGS_WHEEL` here.

pnowack (changes requested):

pnowack:
Just these two things need to be fixed, rest looks good to me.

client/X11/xf_client.c
akallabeth: @pnowack thx, sorry, pushed the wrong commit.

client/X11/xf_client.c
```diff
@@ -1077,7 +1077,8 @@ static const button_map xf_button_flags[NUM_BUTTONS_MAPPED] = {
 	{ Button3, PTR_FLAGS_BUTTON2 },
 	{ Button4, PTR_FLAGS_WHEEL | 0x78 },
 	{ Button5, PTR_FLAGS_WHEEL | PTR_FLAGS_WHEEL_NEGATIVE | 0x78 },
```
pnowack, line 1079: Seems like that you need to force push to this branch again :D You reverted now the complete change for Button5. `0x78` needs to be replaced here with `(0x100 - 0x78)`.

akallabeth: looks like I am blind today :/

pnowack:
Looks good to me now.

mfleisz (approved):
LGTM

PR #6816, opened 2021-02-13
Allow parallel listen to IPv4 and v6 addresses

linuxrrze:
While testing freerdp based ogon RDP service I noticed, that the rdp server component does listen to IPv4 only.

Further analysis showed, that both IPv4 and IPv6 bind() calls were done, however the second one failed (resource busy).

Details s.
https://github.com/ogon-project/ogon-project/issues/16

Attached patch should fix that.


akallabeth:
@linuxrrze thank you for the pull request, but could you run `clang-format` on your changes please?

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #6818, opened 2021-02-15
server: ignore unknown channelId received from client.

llyzs:


akallabeth (approved):
LGTM

hardening (approved):
LGTM, I'm just curious to know when such thing happens ?

llyzs:
> LGTM, I'm just curious to know when such thing happens ?

It happens if custom client has non-standard channel registration and connects to another custom server implementation which has no such channel support.

PR #6824, opened 2021-02-17
Fixed #6821: _stricmp does not like NULL strings

### LINKED ISSUE #6821: macOS smartcard support still has issues
**Describe the bug**
Recent pull request #6792 considerably improved smart card usability when using FreeRDP on Linux to connect to a Windows server. It improved performance (no hard data, but things seem to be faster) and [scard-tester](https://github.com/akallabeth/scard-tester), which previously crashed before getting to the end of tests now runs until the end. Overall, the improvements were very positive.

On macOS, which also uses the PCSC backend, there are still some issues. Generally speaking, the pull request seems to have improved performance (smart card access seems quicker). I can identify, however, two odd situations which are easily reproducible. As stated before, these scenarios present no problems when running the client on Linux.

**To Reproduce**
Steps to reproduce the behavior (scenario 1):
1. Connect a smart card reader and insert smart card
2. Connect to a remote windows server (tested with WS 2016) with smart card redirection
3. Run scard-tester on the remote server
4. After a couple of seconds, the RDP session disconnects (before the PR changes, the scard-tester application crashed)

Steps to reproduce the behavior (scenario 2):
1. Connect a smart card reader and insert smart card
2. Connect to a remote windows server (tested with WS 2016) with smart card redirection
3. On the remote server, use the Microsoft Remote Desktop Client (mstsc.exe) to connect to a second Windows Server
4. On the second remote server, use any smart card application
5. The second (mstsc.exe) RDP session hangs/freezes
6. Remove the card from the reader (no need to disconnect the reader)
7. The second (mstsc.exe) RDP resumes and is now working normally

Note that both these scenarios work just fine when running the client on Linux.

**Expected behavior**
Scenario 1: the scard-tester application should run normally.
Scenario 2: the second RDP session (through mstsc.exe) should be able to use the redirected smart card normally.

I have limited access to the macOS device to test the client, but I'm able to obtain additional logs (FreeRDP or OS), if necessary.

Thanks!

akallabeth:


hardening (approved):
LGTM

hardening:
Should it be backported ?

PR #6826, opened 2021-02-18
Fixed smartcard crash on error return (invalid length)

akallabeth:
`SCardStatusW` does not set `cBytes` to `0` on error return


hardening (approved):
LGTM

PR #6836, opened 2021-02-24
Fix /monitor-list flag being ignored on X11/Wayland

caseif:
This PR fixes a bug introduced by commit d7566f5f5af8778cf1ed647be3897e3a43213375 wherein the `/monitor-list` flag was being ignored due to the flag variable being grabbed _before_ the call to `freerdp_client_settings_command_line_status_print`, which is responsible for actually setting it when a status command has been issued (`freerdp_client_settings_parse_command_line` exits early in this scenario and does not see the flag).


akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #6851, opened 2021-03-02
Wayland pointer frame

akallabeth:


winpr/libwinpr/utils/wlog/ConsoleAppender.c
```diff
@@ -138,6 +138,7 @@ static BOOL WLog_ConsoleAppender_WriteMessage(wLog* log, wLogAppender* appender,
 
 	if (message->Level != WLOG_OFF)
 		fprintf(fp, "%s%s\n", message->PrefixString, message->TextString);
+	fflush(fp);
```
hardening, line 141: Hum this one seems unrelated.
Doesn't it have the possibility to kill the performances of the console appender ?

akallabeth: oops ^^

akallabeth:
@hardening ok, cleaned up the commits, now there should be no debug leftovers.

PR #6860, opened 2021-03-04
Parse on a copy of the argument string for printer

akallabeth:
If done on original the wrong backend is loaded on autoreconnect

### DIFF
--- a/channels/printer/client/printer_main.c
+++ b/channels/printer/client/printer_main.c
@@ -980,7 +980,7 @@ printer_DeviceServiceEntry
 
 	device = (RDPDR_PRINTER*)pEntryPoints->device;
 	name = device->Name;
-	driver_name = device->DriverName;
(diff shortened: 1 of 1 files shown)


channels/printer/client/printer_main.c
```diff
@@ -1017,7 +1017,7 @@ printer_DeviceServiceEntry
 	if (!driver)
 	{
 		WLog_ERR(TAG, "Could not get a printer driver!");
-		return CHANNEL_RC_INITIALIZATION_ERROR;
+		goto fail;
```
mfleisz, line 1020: Intentional that this will return `CHANNEL_RC_OK`?

akallabeth: no, fixed.

mfleisz (approved):

PR #6864, opened 2021-03-05
Revert "allow to use in single threaded mode"

akallabeth:
This reverts commit bee2e1526d158e7cd9f7b626fab1456407e7d32d.
This reverts #5267
@sss123next we´ve had too many issues with this, I have to revert it for now (e.g. rails is no longer working).


mfleisz (approved):

sss123next:
where can i find description of problems introduced by this ?

akallabeth:
@sss123next sorry, did not create an issue for that. The `RAILS` channel is completely broken with this (that is, remote app mode)

sss123next:
understood.
i will check it.

sss123next:
i have fixed this bug (code lost probably during last few years if reviews for this patch), how can i apply this fix now ?

PR #6867, opened 2021-03-05
use tlsOut BIO when using websocket in rdg_bio_ctrl

akarl10:
Reviewing rdg.c while inspecting possible causes of https://github.com/FreeRDP/FreeRDP/issues/6838 I discovered that some socket operations where done on the wrong non existing tlsIn socket.
It did not fix the issue mentioned but espescially BIO_C_WAIT_WRITE may cause something unexpected if essentially turning into a nop


akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #6875, opened 2021-03-08
Fix xf_Pointer_SetPosition with smart-sizing

akallabeth:


akallabeth:
@pnowack already tested this

PR #6885, opened 2021-03-11
client: Do not override connection type

mfleisz:
The RDP file parser was overriding the connection type setting if either, networkautodetect or bandwidthautodetect was enabled


hardening (approved):
Not the expert on that subject but LGTM

akallabeth (approved):

PR #6924, opened 2021-03-26
Clipboard close fd

akallabeth:
Do not cache open file descriptors:
1. We might run out of file descriptors available when copying many files
2. the server might not read the whole file, so we keep cached file descriptors unnecessarily

### DIFF
--- a/winpr/libwinpr/clipboard/posix.c
+++ b/winpr/libwinpr/clipboard/posix.c
@@ -1016,7 +1016,9 @@ UINT posix_file_read_close(struct posix_file* file, BOOL force)
 	if (file->fd < 0)
 		return NO_ERROR;
 
-	if ((file->offset >= file->size) || force)
+	/* Always force close the file. Clipboard might open hundreds of files
+	 * so avoid caching to prevent running out of available file descriptors */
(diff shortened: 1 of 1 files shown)


winpr/libwinpr/clipboard/posix.c
```diff
@@ -1017,12 +1011,14 @@ static UINT posix_file_read_perform(struct posix_file* file, UINT32 size, BYTE**
...
-static UINT posix_file_read_close(struct posix_file* file)
+UINT posix_file_read_close(struct posix_file* file, BOOL force)
 {
 	if (file->fd < 0)
 		return NO_ERROR;
 
-	if (file->offset == file->size)
+	/* Always force close the file. Clipboard might open hundreds of files
+	 * so avoid caching to prevent running out of available file descriptors */
+	if ((file->offset >= file->size) || force || TRUE)
```
kubistika, line 1021: I think the `|| TRUE` is redundant (you already check for `force`).
Other than that looks good.

akallabeth: @kubistika no, I´ve force disabled this until we have determined such a fd cache will not always keep (lots of) handles open.

PR #6939, opened 2021-04-07
Wayland client: Disable shortcut inhibition on rctrl press

caseif:
This PR adds code to the Wayland client to disable shortcut inhibition when the right control key is pressed, which is the same behavior as the X11 client. While certain DEs provide their own key sequence for disabling the feature, to the best of my knowledge KWin does not, making it difficult to escape from a fullscreen RDP session when shortcut inhibition is enabled.

Let me know if any changes need to be made regarding the exact behavior when intercepting the key press.

### DIFF
--- a/client/Wayland/wlf_input.c
+++ b/client/Wayland/wlf_input.c
@@ -301,6 +301,9 @@ BOOL wlf_handle_key(freerdp* instance, const UwacKeyEvent* ev)
 	if (!instance || !ev || !instance->input)
 		return FALSE;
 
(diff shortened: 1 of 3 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

caseif:
I just rebased the commit on master, and apart from that, I also removed an erroneous TODO comment that I meant to delete before initially pushing. No actual code has changed, though.

akallabeth:
@freerdp-bot test

akallabeth:
@caseif thank you for this nice contribution!

PR #6944, opened 2021-04-08
core: Clear message queue to reset its closed state

mfleisz:
This fixes an issue where channels are broken when you call `freerdp_connect` again on a freerdp instance that failed to connect previously.


akallabeth (approved):
LGTM

PR #6949, opened 2021-04-09
[WIP] Implement overlapped operations for files

hardening:
This PR contains a `clangformat` pass, plus an implementation of overlapped operations for files


akallabeth:
@hardening why a separate `clang-format` run and not just format your changes with `git clang-format -f` before commit?

hardening:
@akallabeth to treat other files that are marked as modified when running `make clangformat`

Issue #6981, opened 2021-04-27
Coverity tool reports some possible defects


ondrejholy:
**Describe the bug**
Coverity tool reports some possible defects in the freerdp 2.2.0 version. There is a list of the important ones: [scan-results-imp.err](https://github.com/FreeRDP/FreeRDP/files/6384241/freerdp-2.2.0-scan-results-imp.err.txt). There is a list of all of them: [scan-results.err](https://github.com/FreeRDP/FreeRDP/files/6384244/freerdp-2.2.0-scan-results.err.txt). Few of them are already fixed in master. I am going to propose fixes for the important ones...

ondrejholy:
Some more details for the [scan-results-imp.err](https://github.com/FreeRDP/FreeRDP/files/6384241/freerdp-2.2.0-scan-results-imp.err.txt):
- The `USE_AFTER_FREE` defects look like false-positives, but I modified the code to make covscan happy.
- The `BAD_FREE` defect looks like a false-positive as well, but I added a comment to suppress it as it is not obvious.
- The `CPPCHECK_WARNING` defects also look like false-positives, but they are obvious, so I kept the code untouched.
- The `RESOURCE_LEAK` are mostly valid, so I provided fixes for most of them in https://github.com/FreeRDP/FreeRDP/pull/6982.

akallabeth:
@ondrejholy thank you for the effort.
If possible, can you add a `.github/workflow/coverity.yml` to automatically check in the future?

ondrejholy:
Hmm, it would be ideal to make Coverity tool part of GitHub CI, but I don't have totally any experience with it and I can't spend more time on it right now. I got those reports from our infrastructure when submitting new builds...

akallabeth:
@ondrejholy ok. Thought you had played with `coverity` and already know how to set it up. Also busy currently, but would be a nice to have thing in addition to already existing checkers.
@hardening @mfleisz @bmiklautz anyone time for playing with the `CI`?

bmiklautz:
Just a note: Coverity is in the meanwhile fully integrated and runs daily see https://github.com/FreeRDP/FreeRDP/blob/master/.github/workflows/coverity.yml for details

PR #6982, opened 2021-04-27
Fix various issues reported by Coverity

ondrejholy:
Covscan report for freerdp 2.2.0 contains various defects which were marked as important. I have spent some time analyzing them and although they were marked as important, most of them are in error cases, so probably nothing serious. Let's fix most of them anyway. The rest are false positives, or too complicated to fix, or already fixed in master, or simply I am unsure about them.

See: https://github.com/FreeRDP/FreeRDP/issues/6981

### DIFF
--- a/libfreerdp/core/gateway/ncacn_http.c
+++ b/libfreerdp/core/gateway/ncacn_http.c
@@ -121,6 +121,7 @@ BOOL rpc_ncacn_http_recv_in_channel_response(RpcChannel* inChannel, HttpResponse
 	if (ntlmTokenData && ntlmTokenLength)
 		return ntlm_client_set_input_buffer(ntlm, FALSE, ntlmTokenData, ntlmTokenLength);
 
+	free(ntlmTokenData);
 	return TRUE;
 }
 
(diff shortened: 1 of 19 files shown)


libfreerdp/core/gateway/ncacn_http.c
```diff
@@ -121,6 +121,7 @@ BOOL rpc_ncacn_http_recv_in_channel_response(RpcChannel* inChannel, HttpResponse
 	if (ntlmTokenData && ntlmTokenLength)
 		return ntlm_client_set_input_buffer(ntlm, FALSE, ntlmTokenData, ntlmTokenLength);
 
+	free(ntlmTokenData);
```
akallabeth, line 124: Possible duplicate free (see `ntlm_client_set_input_buffer`)

libfreerdp/core/gateway/ncacn_http.c
```diff
@@ -274,5 +275,6 @@ BOOL rpc_ncacn_http_recv_out_channel_response(RpcChannel* outChannel, HttpRespon
...
```
akallabeth, line 278: possible double free, wee `ntlm_client_set_input_buffer`

akallabeth (changes requested):
Thank you for the effort, but it looks like this could introduce some double free, see comments.

libfreerdp/core/gateway/ncacn_http.c
akallabeth: ok, forget it, need coffee :/

libfreerdp/core/gateway/ncacn_http.c
akallabeth: ok, forget it, need coffee :/

akallabeth (approved):

PR #6984, opened 2021-04-28
Fix uninitialized variable warnings.

akallabeth:


hardening:
Hum this warning was a false positive, the initialization was not needed.

PR #7006, opened 2021-05-11
Fixed AAC encoding for microphone

akallabeth:
* Fixed b-frame interval for FFMPEG
* Fixed sample size for FAAC

### DIFF
--- a/libfreerdp/CMakeLists.txt
+++ b/libfreerdp/CMakeLists.txt
@@ -115,6 +115,10 @@ foreach(${MODULE_PREFIX}_SUBMODULE ${${MODULE_PREFIX}_SUBMODULES})
     add_subdirectory(${${MODULE_PREFIX}_SUBMODULE})
 endforeach()
 
+if (NOT WITH_DSP_FFMPEG AND NOT WITH_FAAC)
+    message(WARNING "Compiling without WITH_DSP_FFMPEG and WITH_FAAC, AAC encoder support disabled")
+endif ()
+
 ## cmake source properties are only seen by targets in the same CMakeLists.txt
 ## therefore primitives and codecs need to be defined here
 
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

hardening (approved):
Not the best expert but LGTM

vscfreire:
Hi @akallabeth, as you know, I'm always interested in sound subsystem fixes, optimizations and improvements, as I use lots of video/sound/mic over FreeRDP. Can you confirm that you tested this fixes with Windows? I'm asking because I've done several tests with this and AAC and the mic sound is still garbled. PCM works perfect, but AAC has lots of distortion. The first few tries it seemed a little better than before this fix, but to be sincere, I reverted the fixes and it sounds the same.

akallabeth:
@vscfreire yes, tested against windows 8 to 10 (and server 2016 and 2019)
no distortions (and especially no increasing delay) with these patches.
tested with integrated sound recorder app.
* Which kind of distortions do you have? (maybe attach some sample record)?
* Which backend do you use (ffmpeg or faac)
* Which os do you use on the client side?

vscfreire:
- [AACRecordingTest.zip](https://github.com/FreeRDP/FreeRDP/files/6842295/AACRecordingTest.zip)
- ffmpeg - tried the latest minors of 4.2.4, 4.3.2 and 4.4.0, all had the same kind of distortions
- Windows 10 19043.1110 on the client, Windows Server 2016 on the server

Used Audacity to record the sample, if I pressed a second time on the Record, the session would also disconnect. With PCM, not only the recorded sound is perfect, but the session doesn't disconnect if I record again.

akallabeth:
@vscfreire ok, weired. did you try with `faac` instead of `ffmpeg` too? (different encoder, so you can rule out issues there)
will do a recheck of the windows client, mac and linux do work fine with `faac` and `ffmpeg` (but there is some format conversion issue with some `ffmpeg` builds)

PR #7008, opened 2021-05-12
Fix FIPS mode support and build with OpenSSL 3.0

ondrejholy:
FreeRDP fails to build with OpenSSL 3.0 because of usage of the `FIPS_mode`
and `FIPS_mode_set` functions, which were removed. However, `FIPS_mode_set`
seems to have done nothing since OpenSSL 1.1.1 anyway. This is tentative fix
to make FreeRDP build with OpenSSL 3.0 and to fix the FIPS mode support.

See: https://bugzilla.redhat.com/show_bug.cgi?id=1952937

### DIFF
--- a/winpr/libwinpr/utils/ssl.c
+++ b/winpr/libwinpr/utils/ssl.c
@@ -244,9 +244,17 @@ static BOOL winpr_enable_fips(DWORD flags)
 #else
 		WLog_DBG(TAG, "Ensuring openssl fips mode is ENabled");
 
+#if defined(OPENSSL_VERSION_MAJOR) && (OPENSSL_VERSION_MAJOR >= 3)
+		if (!EVP_default_properties_is_fips_enabled(NULL))
+#else
 		if (FIPS_mode() != 1)
+#endif
 		{
+#if defined(OPENSSL_VERSION_MAJOR) && (OPENSSL_VERSION_MAJOR >= 3)
+			if (EVP_set_default_properties(NULL, "fips=yes"))
+#else
 			if (FIPS_mode_set(1))
+#endif
 				WLog_INFO(TAG, "Openssl fips mode ENabled!");
 			else
 			{


ondrejholy:
I wrote a tentative fix because I've only verified that it fixes FreeRDP build with OpenSSL 3.0, but I don't have currently an environment to test the FIPS mode changes...

ondrejholy:
Dammit, obviously 1.1.1 doesn't contain `EVP_default_properties_is_fips_enabled` and `EVP_set_default_properties` yet. So maybe 1.1.1 needs to be fixed in a different way...

winpr/libwinpr/utils/ssl.c
```diff
@@ -244,9 +244,17 @@ static BOOL winpr_enable_fips(DWORD flags)
 #else
 		WLog_DBG(TAG, "Ensuring openssl fips mode is ENabled");
 
+#if (OPENSSL_VERSION_NUMBER > 0x10101000L)
```
akallabeth, line 247: you might want to change the version to be `>1.1.2a`  instead of `1.1.1a`

akallabeth: the 3.0 alpha versions did introduce `OPENSSL_VERSION_MAJOR` `OPENSSL_VERSION_MINOR` and `OPENSSL_VERSION_PATCH` so that might be a good way to fix the test

ondrejholy: So I have read some more documents (e.g. https://wiki.openssl.org/index.php/FIPS_modules) and my understanding is that FIPS mode was supported by 1.0.x versions, but it is not supported in 1.1.x versions at all  (although `FIPS_mode` and `FIPS_mode_set` exists) and going to be supported by 3.x. So `OPENSSL_VERSION_MAJOR >= 3` is probably the right way to go here...

ondrejholy:
@akallabeth, it is probably safe to merge this now as it builds fine with OpenSSL 1.1.x/3.x and given the fact that the FIPS support was "broken" with OpenSSL 1.1.x anyway (as noted in https://github.com/FreeRDP/FreeRDP/pull/7008#discussion_r631662679). But still, it would be nice to prove that it actually fixes the FIPS support for FreeRDP. @akallabeth, are you able to verify that somehow?

akallabeth (approved):
LGTM

PR #7010, opened 2021-05-12
winpr: add a HashTable_Foreach function and associated tests

hardening:
This useful functions allows to browse all value pairs of a hashtable without having to allocate some memory for keys and then retrieving each element.
It may also make sense with synchronized hashtables because before you were forced to `HashTable_GetKeys()` and then `HashTable_GetItemValue()` which is locking again for each element of the table.


akallabeth (approved):
LGTM

PR #7012, opened 2021-05-17
Hash table opaque

akallabeth:
* Refactored HashTable API to be opaque
* Fixed `ArrayList_Add` to conform to other function return types

### DIFF
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -1051,7 +1051,7 @@ static UINT xf_rail_server_get_appid_response(RailClientContext* context,
 	return CHANNEL_RC_OK;
 }
 
-static BOOL rail_window_key_equals(void* key1, void* key2)
+static BOOL rail_window_key_equals(const void* key1, const void* key2)
 {
 	const UINT64* k1 = (const UINT64*)key1;
 	const UINT64* k2 = (const UINT64*)key2;
@@ -1062,7 +1062,7 @@ static BOOL rail_window_key_equals(void* key1, void* key2)
 	return *k1 == *k2;
 }
 
-static UINT32 rail_window_key_hash(void* key)
+static UINT32 rail_window_key_hash(const void* key)
 {
 	const UINT64* k1 = (const UINT64*)key;
 	return (UINT32)*k1;
@@ -1101,18 +1101,30 @@ int xf_rail_init(xfContext* xfc, RailClientContext* rail)
 	if (!xfc->railWindows)
 		return 0;
 
-	xfc->railWindows->keyCompare = rail_window_key_equals;
-	xfc->railWindows->hash = rail_window_key_hash;
-	xfc->railWindows->valueFree = rail_window_free;
+	if (!HashTable_SetHashFunction(xfc->railWindows, rail_window_key_hash))
+		goto fail;
+	{
+		wObject* obj = HashTable_KeyObject(xfc->railWindows);
+		if (!obj)
+			goto fail;
+		obj->fnObjectEquals = rail_window_key_equals;
+	}
+	{
+		wObject* obj = HashTable_ValueObject(xfc->railWindows);
+		if (!obj)
+			goto fail;
+		obj->fnObjectFree = rail_window_free;
+	}
 	xfc->railIconCache = RailIconCache_New(xfc->context.settings);
 
 	if (!xfc->railIconCache)
 	{
-		HashTable_Free(xfc->railWindows);
-		return 0;
 	}
 
 	return 1;
+fail:
+	HashTable_Free(xfc->railWindows);
+	return 0;
 }
 
 int xf_rail_uninit(xfContext* xfc, RailClientContext* rail)
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -1173,7 +1173,7 @@ xfAppWindow* xf_rail_add_window(xfContext* xfc, UINT64 id, UINT32 x, UINT32 y, U
 	appWindow->width = width;
 	appWindow->height = height;
 	xf_AppWindowCreate(xfc, appWindow);
-	HashTable_Add(xfc->railWindows, &appWindow->windowId, (void*)appWindow);
+	HashTable_Insert(xfc->railWindows, &appWindow->windowId, (void*)appWindow);
 	return appWindow;
 }
 
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -1173,7 +1173,11 @@ xfAppWindow* xf_rail_add_window(xfContext* xfc, UINT64 id, UINT32 x, UINT32 y, U
 	appWindow->width = width;
 	appWindow->height = height;
 	xf_AppWindowCreate(xfc, appWindow);
-	HashTable_Insert(xfc->railWindows, &appWindow->windowId, (void*)appWindow);
+	if (!HashTable_Insert(xfc->railWindows, &appWindow->windowId, (void*)appWindow))
+	{
+		rail_window_free(appWindow);
+		return NULL;
+	}
 	return appWindow;
 }
 
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -1105,14 +1105,10 @@ int xf_rail_init(xfContext* xfc, RailClientContext* rail)
 		goto fail;
 	{
 		wObject* obj = HashTable_KeyObject(xfc->railWindows);
-		if (!obj)
-			goto fail;
 		obj->fnObjectEquals = rail_window_key_equals;
 	}
 	{
 		wObject* obj = HashTable_ValueObject(xfc->railWindows);
-		if (!obj)
-			goto fail;
 		obj->fnObjectFree = rail_window_free;
 	}
 	xfc->railIconCache = RailIconCache_New(xfc->context.settings);
--- a/libfreerdp/codec/progressive.c
+++ b/libfreerdp/codec/progressive.c
@@ -357,7 +357,7 @@ static INLINE BOOL progressive_set_surface_data(PROGRESSIVE_CONTEXT* progressive
 	key = ((ULONG_PTR)surfaceId) + 1;
 
 	if (pData)
-		return HashTable_Add(progressive->SurfaceContexts, (void*)key, pData) >= 0;
+		return HashTable_Insert(progressive->SurfaceContexts, (void*)key, pData) >= 0;
 
 	HashTable_Remove(progressive->SurfaceContexts, (void*)key);
 	return TRUE;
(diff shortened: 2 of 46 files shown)


hardening:
Wow that's a big change with massive API breakage. That's a so massive break, that it would probably be a good idea to name it HashTable2 or we're gonna have failing software forever that will `HashTable_Add()` and will test for a negative number while we should test FALSE.

May I suggest a utility function that wires hash/copy/compare for strings, I've seen quite a lot of occurrence of this in your changes.

akallabeth:
@hardening true. Renamed the functions that changed return types in an (incompatible) way so it does not fail silently

akallabeth:
@freerdp-bot test

winpr/libwinpr/utils/collections/HashTable.c
```diff
@@ -148,19 +179,69 @@ static void HashTable_Rehash(wHashTable* table, int numOfBuckets)
 	table->numOfBuckets = numOfBuckets;
 }
 
-wKeyValuePair* HashTable_Get(wHashTable* table, void* key)
+static BOOL HashTable_Equals(wHashTable* table, const wKeyValuePair* pair, const void* key)
+{
+	if (!key || !pair || !table)
+		return FALSE;
```
hardening, line 185: I feel like testing `table` here is not a good idea as for example in `HashTable_Get` we're gonna test `table` at each value of a bucket. So most probably we shall remove that test and instead test `table` at the beginning of every public `HashTable` function (I guess static private function don't need that).

winpr/libwinpr/utils/collections/HashTable.c
```diff
@@ -74,24 +105,24 @@ void HashTable_StringFree(void* str)
...
-			return 0;
+			return FALSE;
 	}
 
-	return 1; /* maybe */
+	return TRUE; /* maybe */
 }
 
-static long HashTable_CalculateIdealNumOfBuckets(wHashTable* table)
+static size_t HashTable_CalculateIdealNumOfBuckets(wHashTable* table)
```
hardening, line 123: Perhaps we can have a better way to compute ideal sizes as it's done here: https://github.com/microsoft/weston-mirror/blob/working/libweston/backend-rdp/hash.c#L67 (no computations, just a table scan)

winpr/include/winpr/collections.h
```diff
@@ -302,66 +289,67 @@ extern "C"
...
-		HASH_TABLE_KEY_FREE_FN keyFree;
-		HASH_TABLE_VALUE_FREE_FN valueFree;
-
-		DWORD foreachRecursionLevel;
-		DWORD pendingRemoves;
-	};
 	typedef struct _wHashTable wHashTable;
 
-	typedef BOOL (*HASH_TABLE_FOREACH_FN)(const void* key, void* value, void* arg);
+	typedef BOOL (*HASH_TABLE_FOREACH_FN)(const void* key, const void* value, void* arg);
```
hardening, line 296: IMHO setting value as `const` is a bad idea, as in many cases we will want to modify `value`, and so we will have to break the const cast.

hardening:
You test too much in your static private functions:

* there's the case with the `table` argument in HashTable functions that gets rested multiple times, sometime even when  scanning a bucket;
* but also a pair can't be null. And in a pair, key and value can't be null either as we forbid such an insert, so these tests aren't necessary;
* tests should concentrate on input values given to public functions

winpr/libwinpr/utils/collections/HashTable.c
akallabeth: this is static, so the compiler will eliminate duplicate checks. Also updated to `INLINE`

winpr/include/winpr/collections.h
akallabeth: You´re right, removed.

akallabeth:
@hardening added input parameter checks as suggested.
As for the checks in the `static INLINE` functions, these get eliminated by the compiler anyway, so better safe than sorry.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

hardening:
@akallabeth that's always our disagreement point, I've started a conversation on this to no pollute this PR => https://github.com/FreeRDP/FreeRDP/discussions/7019 and it would be super cool if we could end up with a document giving guidelines.

hardening (changes requested):
Let's agree on checks in the discussion before treating this PR.

akallabeth:
@freerdp-bot test

akallabeth:
@hardening I´ve added `WINPR_ASSERT` checks instead of checking `list`/`table`. Anything else to change to get this pr in?

akallabeth:
@freerdp-bot test

client/X11/xf_rail.c
```diff
@@ -1101,18 +1101,30 @@ int xf_rail_init(xfContext* xfc, RailClientContext* rail)
...
 		return 0;
 
-	xfc->railWindows->keyCompare = rail_window_key_equals;
-	xfc->railWindows->hash = rail_window_key_hash;
-	xfc->railWindows->valueFree = rail_window_free;
+	if (!HashTable_SetHashFunction(xfc->railWindows, rail_window_key_hash))
+		goto fail;
+	{
+		wObject* obj = HashTable_KeyObject(xfc->railWindows);
+		if (!obj)
```
hardening, line 1108: That should be an assert, if `HashTable_KeyObject` returns `NULL` it means we're in big trouble.

libfreerdp/codec/progressive.c
```diff
@@ -357,7 +357,7 @@ static INLINE BOOL progressive_set_surface_data(PROGRESSIVE_CONTEXT* progressive
 	key = ((ULONG_PTR)surfaceId) + 1;
 
 	if (pData)
-		return HashTable_Add(progressive->SurfaceContexts, (void*)key, pData) >= 0;
+		return HashTable_Insert(progressive->SurfaceContexts, (void*)key, pData);
```
hardening, line 360: Not your code, but the indexing of surfaces shall be changed to be a pointer on surfaceId and not the surfaceId itself, to remove the +1 trick for non null keys.

libfreerdp/core/client.c
```diff
@@ -830,8 +830,8 @@ static UINT VCAPITYPE FreeRDP_VirtualChannelInitEx(
 		pChannelOpenData->OpenHandle = InterlockedIncrement(&g_OpenHandleSeq);
 		pChannelOpenData->channels = channels;
 		pChannelOpenData->lpUserParam = lpUserParam;
-		HashTable_Add(g_ChannelHandles, (void*)(UINT_PTR)pChannelOpenData->OpenHandle,
-		              (void*)pChannelOpenData);
+		HashTable_Insert(g_ChannelHandles, (void*)(UINT_PTR)pChannelOpenData->OpenHandle,
```
hardening, line 833: missing check of return value ?

libfreerdp/core/client.c
```diff
@@ -920,8 +920,8 @@ static UINT VCAPITYPE FreeRDP_VirtualChannelInit(LPVOID* ppInitHandle, PCHANNEL_
 		pChannelOpenData = &channels->openDataList[channels->openDataCount];
 		pChannelOpenData->OpenHandle = InterlockedIncrement(&g_OpenHandleSeq);
 		pChannelOpenData->channels = channels;
-		HashTable_Add(g_ChannelHandles, (void*)(UINT_PTR)pChannelOpenData->OpenHandle,
-		              (void*)pChannelOpenData);
+		HashTable_Insert(g_ChannelHandles, (void*)(UINT_PTR)pChannelOpenData->OpenHandle,
```
hardening, line 923: missing check of return value ?

winpr/include/winpr/collections.h
```diff
@@ -124,7 +124,8 @@ extern "C"
 	WINPR_API void ArrayList_Clear(wArrayList* arrayList);
 	WINPR_API BOOL ArrayList_Contains(wArrayList* arrayList, const void* obj);
 
-	WINPR_API int ArrayList_Add(wArrayList* arrayList, void* obj);
+	WINPR_API WINPR_DEPRECATED(int ArrayList_Add(wArrayList* arrayList, const void* obj));
+	WINPR_API BOOL ArrayList_Append(wArrayList* arrayList, const void* obj);
```
hardening, line 128: I don't see any advantage to have the obj as `const void *` as we wanna use the items in the list, so well they're not `const` at all.

akallabeth: the advantage is that with the new API you can use `const` data without compiler warnings.

akallabeth: The relevant function for manipulation is `ArrayList_GetItem` and that still returns a non `const` value.

winpr/include/winpr/collections.h
```diff
@@ -124,7 +124,8 @@ extern "C"
 	WINPR_API void ArrayList_Clear(wArrayList* arrayList);
 	WINPR_API BOOL ArrayList_Contains(wArrayList* arrayList, const void* obj);
 
-	WINPR_API int ArrayList_Add(wArrayList* arrayList, void* obj);
+	WINPR_API WINPR_DEPRECATED(int ArrayList_Add(wArrayList* arrayList, const void* obj));
```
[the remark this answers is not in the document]
akallabeth: Will reduce compiler warnings as the old prototype does not support adding `const` data at all.
We can keep the old one of course, but it does not have any benefit

client/X11/xf_rail.c
akallabeth: since this is external of `winpr` I´d keep the error check. will modify the internals for `HashTable_SetupForStringData` though

winpr/libwinpr/utils/collections/HashTable.c
```diff
@@ -714,19 +799,58 @@ void HashTable_Free(wHashTable* table)
...
+	return &table->key;
+}
+
+wObject* HashTable_ValueObject(wHashTable* table)
+{
+	WINPR_ASSERT(table);
+	return &table->value;
+}
+
+BOOL HashTable_SetHashFunction(wHashTable* table, HASH_TABLE_HASH_FN fn)
```
hardening, line 825: Then `HashTable_SetHashFunction` shall probably return void as it can only assert

hardening:
Calls to `HashTable_KeyObject` and `HashTable_ValueObject` can never return NULL (it will only assert), so I guess the checks for the return value should be removed. It will simplify code in many places;

akallabeth:
@hardening done. Also added `WINPR_ASSERT` to a few more collections

winpr/libwinpr/utils/collections/HashTable.c
hardening: True. 
The limit between input parameters checks and API violation that gives an assert is tight. We assert on `table == NULL` but we return FALSE on NULL `fn`. Should it be also an assert ?

PR #7034, opened 2021-05-24
Fix build errors in MinGW

Biswa96:


Biswa96:
There is a conflict with NTLM_AV_ID and MSV1_0_AVID enumerations. Both have same members. I can not figure out a non-destructive conclusion.

hardening (commented):
I feel like some of these changes suggest that some struct definitions are in the wrong place.
Isn't it a problem if a public header in the freerdp include directory needs some `#ifdef _WIN32` ? I mean isn't it the sign that we have missed a winPR abstraction somewhere ?

Biswa96:
> Isn't it a problem if a public header in the freerdp include directory needs some #ifdef _WIN32 ?

Yeah... Agree. But that may need a huge code reformat which has a greater probability of destroying everything. Doing `grep -rl WIN32 | wc -l` command inside install folder reveals that there are 54 files with `#ifdef _WIN32`.

hardening:

hardening:
@freerdp-bot I feel like you failed the OsX test for no reason !

include/freerdp/codec/audio.h
```diff
@@ -69,6 +69,7 @@ typedef struct AUDIO_FORMAT AUDIO_FORMAT;
  * http://tools.ietf.org/html/rfc2361
  */
 
+#ifndef __MINGW32__
```
akallabeth, line 72: Might be a good idea to prefix the `WAVE_FORMAT_*` symbols with `FREERDP_WAVE_FORMAT_*` if there is a collision

include/freerdp/channels/rdpdr.h
```diff
@@ -298,6 +298,7 @@ enum FILE_FS_DEVICE_FLAG
 	FILE_DEVICE_SECURE_OPEN = 0x00000100
 };
 
+#ifndef __MINGW32__
```
akallabeth, line 301: Might be a good idea to rename the struct as it might not be equal to what we expect in freerdp

server/Windows/wf_wasapi.c
```diff
@@ -18,6 +18,7 @@
 //#define REFTIMES_PER_SEC  50000
 //#define REFTIMES_PER_MILLISEC  50
 
+#ifndef __MINGW32__
```
akallabeth, line 21: Is there maybe a windows header we forgot to include so we can remove this define?

akallabeth (commented):
Thank you for the (updated) pull request for mingw.
looks like you uncovered some naming collisions within freerdp, other than that this pr looks good to me.

include/freerdp/codec/audio.h
Biswa96: I was wondering why not remove these. Changing symbols will effect whole project.

server/Windows/wf_wasapi.c
Biswa96: These GUIDs are in mmdeviceapi.h which is already included. Not sure why these were added.

akallabeth:
@freerdp-bot I gave you more time, would you now be so kind and run the test through?

server/Windows/wf_wasapi.c
akallabeth: might be something like https://stackoverflow.com/questions/31666420/undefined-reference-to-clsid-mmdeviceenumerator-and-iid-immdeviceenumerator

include/freerdp/channels/rdpdr.h
Biswa96: `FILE_FS_INFORMATION_CLASS` is defined in mingw-w64 but not in Microsoft Windows SDK. So, this may be the right choice.

include/freerdp/codec/audio.h
akallabeth: because we need them, windows is not the only target ;)
changing at this point is no issue, as the next version will be `FreeRDP 3` as we already have broken too much to stay compatible with `2.0`

akallabeth (approved):

akallabeth:
I´d say we merge and fix the duplicate definitions later on (they are not `mingw` specific anyway)

Biswa96:
Is it possible to move/copy the CI builds for all platforms (Win, Mac, *nix) in GitHub Actions so that one can do tests in forked repository?

akallabeth:
no, but you can run the tests locally with:

1. Add  `-DBUILD_TESTING=ON` to `CMake` configure step
1. Run `cmake --build .`
2. Run `cmake --build . --target test`

PR #7036, opened 2021-05-24
codec/progressive: Allow the usage of multithreading for decoding

pnowack:
Quoting the main commit here:

```
While decoding RemoteFX encoded frames is multithreaded, decoding
RemoteFX Progressive frames is not, although both codecs work
relatively similarly.
This is especially noticeable with frames, that have a resolution
larger than 1920x1080 pixels.

decompress_tile_first() and decompress_tile_upgrade() can both run in
different threads at the same time for different tiles without necessary
adjustments.

So, do exactly that using the ThreadPool that already exists in the
RFX_CONTEXT to decrease the decoding time and therefore increase the
performance.
On a 3K display (2880x1620 pixels) this makes out of a choppy
experience a fluid experience.
```

This fixes a heavy performance regression with the RemoteFX Progressive codec, which the plain RemoteFX codec (when using the graphics pipeline) does not have.
I have tested this on top of the stable-2.0 branch using gnome-remote-desktop (uses TILE_SIMPLE) and Win10 (use TILE_FIRST + TILE_UPGRADE)

### DIFF
--- a/libfreerdp/codec/progressive.c
+++ b/libfreerdp/codec/progressive.c
@@ -37,6 +37,7 @@
 #include "rfx_quantization.h"
 #include "rfx_dwt.h"
 #include "rfx_rlgr.h"
+#include "rfx_types.h"
 #include "progressive.h"
 
 #define TAG FREERDP_TAG("codec.progressive")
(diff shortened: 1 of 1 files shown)


libfreerdp/codec/progressive.c
```diff
@@ -1712,24 +1752,62 @@ static INLINE int progressive_process_tiles(PROGRESSIVE_CONTEXT* progressive, wS
...
-			default:
-				WLog_Print(progressive->log, WLOG_ERROR, "Invalid block type %04 (%s)" PRIx16,
-				           tile->blockType, progressive_get_block_type_string(tile->blockType));
-				return -42;
+			SubmitThreadpoolWork(work_objects[index]);
+			close_cnt = index + 1;
+		}
+		else
+		{
+			switch (tile->blockType)
```
akallabeth, line 1795: Maybe just call the thread callback here too to remove duplicate code?

akallabeth (approved):
Overall LGTM, just maybe remove the code duplication for the single threaded path.

libfreerdp/codec/progressive.c
pnowack: Done.

akallabeth (approved):

akallabeth:
@freerdp-bot test

akallabeth:
thanks for this fix, really appreciated :)

PR #7038, opened 2021-05-25
Fix Windows Unicode file path handling issues

awakecoding:
Fix multiple Windows Unicode path handling issues caused by the usage of ANSI Windows APIs. Since we use UTF-8 strings and convert to UTF-16 only when needed in FreeRDP, I added a few functions that take the UTF-8 string and handle the required conversion only when needed on Windows.

One such function is fopen() which unfortunately does not handle UTF-8 correctly. I added the winpr_fopen() wrapper that returns a usage FILE* pointer on all platform, and does the necessary adaptation under the hood only on Windows. Simply calling _wfopen() is not enough: if you open the file in text mode, you'll get a BOM, so the workaround is to open text files in binary mode to get a FILE* pointer to which we can correctly write in UTF-8 for all platforms with no BOM.

We've been using these modifications for years on our side, but I am currently cleaning up our patches so we can have them merged upstream. Expect a few more pull requests this week.

### DIFF
--- a/winpr/include/winpr/file.h
+++ b/winpr/include/winpr/file.h
@@ -521,6 +521,8 @@ extern "C"
 	WINPR_API int GetNamePipeFileDescriptor(HANDLE hNamedPipe);
 	WINPR_API HANDLE GetFileHandleForFileDescriptor(int fd);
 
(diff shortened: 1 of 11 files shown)


akallabeth (approved):
LGTM

PR #7040, opened 2021-05-25
Add WINPR_ONLY build option to build WinPR separately from FreeRDP

awakecoding:
This pull request includes various cmake build system changes along with internal refactoring of OpenSSL/mbedTLS function structures to allow building WinPR against mbedTLS separately from FreeRDP (still using OpenSSL). While OpenSSL is still required for FreeRDP, we have been using WinPR mbedTLS builds for years on our side for multiple projects other than FreeRDP, and our FreeRDP builds are linked against the same prebuilt WinPR library that links against mbedTLS.

I've been very cautious with these changes, they should have no impact on the existing build options. Another option that was added is WITH_WINPR_TOOLS to turn off the WinPR tools like such as makecert that were never ported to mbedTLS.


akallabeth (changes requested):
Sorry, the whole crypto/ssl part should not be part of this.
We try to move this kind of dependency to winpr, not the other way around.

hardening:
I have another more general remark, isn't it the sign that winPR should move to its own repo, and FreeRDP would depend on it just like any other external dependency ?

awakecoding:
> Sorry, the whole crypto/ssl part should not be part of this.
> We try to move this kind of dependency to winpr, not the other way around.

That is simply not possible, because FreeRDP and WinPR currently share common OpenSSL crypto utils, and the only way to build FreeRDP and WinPR against a separate crypto library is to split those utils (one for FreeRDP, one for WinPR). In other words, I just cannot make a simple PR that does nothing more than edit the cmake files, the end result wouldn't even build and run.

FreeRDP still has a direct dependency on OpenSSL, and abstracting it all away to move it down to WinPR would be serious refactoring to the TLS and TS Gateway code. If we move it down to WinPR, it wouldn't feel much like a Windows API unless we make an SChannel interface, and that is probably not what we want, even if we already have the stubs in libwinpr-sspi.

awakecoding:
> I have another more general remark, isn't it the sign that winPR should move to its own repo, and FreeRDP would depend on it just like any other external dependency ?

I thought about it, and FreeRDP is already complex enough as it is. Moving WinPR to a separate repo would often mean making pull requests in both repos at the same time, a lot of the fixes being done have to touch both FreeRDP and WinPR sides. I am using WinPR in two other internal projects at Devolutions, but since I prebuild it separately, it is not much of an issue. Making it possible to build WinPR only inside the same FreeRDP build system is the solution that is the least painful for everyone.

libfreerdp/crypto/ssl.c
```diff
@@ -0,0 +1,388 @@
...
+
+#endif
+	g_freerdp_openssl_initialized_by_winpr = TRUE;
+
+	if (flags & WINPR_SSL_INIT_ENABLE_FIPS)
+	{
+#if (OPENSSL_VERSION_NUMBER < 0x10001000L) || defined(LIBRESSL_VERSION_NUMBER)
+		WLog_ERR(TAG,
+		         "Openssl fips mode ENable not available on openssl versions less than 1.0.1!");
+#else
```
hardening, line 279: Typo "ENable"

bmiklautz:
@freerdp-bot test

libfreerdp/crypto/ssl.c
awakecoding: I fixed those typos and made the debug logging casing more consistent, this was basically already there.

awakecoding:
@akallabeth @hardening as discussed on IRC, we want to take baby steps here, and the last thing I want is to make this initial FreeRDP/WinPR separation more than what it should be. We can consider removing the OpenSSL dependency in FreeRDP later, and I don't think we need to plan to move WinPR outside of the FreeRDP repository, because it would make FreeRDP development much harder by having to make a lot of commits in two separate repositories.

However, I would appreciate if we could get this PR moving promptly this week, as it is a very important change that we need on side to get back in sync with the FreeRDP upstream sources.

akallabeth:
@freerdp-bot test

akallabeth (changes requested):
Sorry, why not just call `cmake` from `winpr` subfolder? There are a few missing definitions currently, but that will build winpr only and you just need to fix the missing includes/find_library

awakecoding:
@akallabeth that is what I had initially attempted, only to realize the sheer amount of cmake duplication it would require to build correctly. Since I didn't intend to split repositories (and still don't) the path of least disturbance was to keep using the FreeRDP parent cmake build system but add a WINPR_ONLY option in it. If this is what you truly want anyway, I would still recommend we do it in a second PR after this one to reduce the size of individual PRs.

akallabeth:
@awakecoding there is not that much duplication (`config.h`, `build-flags.h` files maybe) but the build procedure must change. See #7082 which is not ready yet.

PR #7041, opened 2021-05-26
Remove strnlen(INT_MAX) trick to fix Android x86/ARM emulation

awakecoding:
This pull request removes the odd strnlen(INT_MAX) weird trick that unfortunately doesn't play well with some Android x86 devices doing ARM emulation, or Android ARM64 devices emulating ARM binaries. I suspect it is caused by the usage of limit values and the way the emulation is done, and rather than tell a few customers to "buy another Android device", I'd rather get it fixed here.

I understand the intent, but I'm not sure this trick is really any safer than a plain strlen anyway, and it is definitely causing me to have to patch my builds just for this. Are there other techniques that don't involve using the upper bound of INT_MAX in strnlen? If we really have to check for this one, I suspect there are a lot of other places we'd rather check for safety in FreeRDP.

### DIFF
--- a/winpr/libwinpr/crt/unicode.c
+++ b/winpr/libwinpr/crt/unicode.c
@@ -170,7 +170,7 @@ int MultiByteToWideChar(UINT CodePage, DWORD dwFlags, LPCSTR lpMultiByteStr, int
 
 	if (cbMultiByte == -1)
 	{
-		size_t len = strnlen((const char*)lpMultiByteStr, INT32_MAX);
+		size_t len = strlen((const char*)lpMultiByteStr);
 		if (len >= INT32_MAX)
 			return 0;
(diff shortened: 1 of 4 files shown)


bmiklautz:
Hey @awakecoding long time ;) - What was the problem your customer was  hit with? Did the build or the runtime fail?

awakecoding:
@bmiklautz long time no see ;) this is not a compile time issue, it is a runtime issue which we've seen happen on some Android x86 and ARM64 devices. The ARM64 problem may no longer be there because we're making ARM64 builds now, but the Android x86 devices out there will emulate the ARM builds and this is where it can break.

bmiklautz:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #7042, opened 2021-05-26
Fix possible WSAEventSelect() file descriptor leak

awakecoding:
This pull request adds robustness to the WinPR event handle deallocator to ensure that all file descriptors that should be closed, are closed. We originally noticed a file descriptor leak on macOS related to WSAEventSelect() calls, and this is how we got it fixed.


akallabeth (approved):
LGTM

PR #7043, opened 2021-05-26
Replace fopen and path functions with wrappers

akallabeth:
Functions like fopen, PathFileExists, PathMakePath need to call
the wide character versions on windows for utf-8 support.
Follow up to https://github.com/FreeRDP/FreeRDP/pull/7038


mfleisz (approved):
LGTM

PR #7044, opened 2021-05-26
Gracefully handle edge case where clipboard format name is NULL

awakecoding:
The ClipboardFindFormat function does not gracefully handle a null format name in the format list - this is an edge case we've managed to hit on our side. This pull request simply skips over format entries that have a null format name when searching by name, instead of crashing in the strcmp call that follows.

### DIFF
--- a/winpr/libwinpr/clipboard/clipboard.c
+++ b/winpr/libwinpr/clipboard/clipboard.c
@@ -89,6 +89,9 @@ static wClipboardFormat* ClipboardFindFormat(wClipboard* clipboard, UINT32 forma
 	{
 		for (index = 0; index < clipboard->numFormats; index++)
 		{
+			if (!clipboard->formats[index].formatName)
+				continue;
+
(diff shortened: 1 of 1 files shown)


hardening:
Can you elaborate in which case we end up with a NULL format name ?

awakecoding:
@hardening honestly it's been a long time, but it did happen so we had to prevent the crash. It's definitely an edge case, and it might have happened outside of FreeRDP because we reuse those clipboard functions in Wayk. There's no harm in avoiding the crash anyway.

akallabeth (approved):

akallabeth:
@freerdp-bot test
@hardening I believe this is a possible case, nothing is more error ridden than clipboard (in every implementation) :/

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #7047, opened 2021-05-26
Avoid seeking when reading Android timezone identifier 'file'

awakecoding:
winpr_read_unix_timezone_identifier_from_file takes a file pointer to read a "file" by seeking to the end and allocating a buffer that fits. The only problem is that not all code paths return a file handle that you can seek, resulting in deadlocks on some Android devices. The Android timezone detection code on Android tries using JNI and fallbacks to calling `popen("getprop persist.sys.timezone", "r");` to obtain a file pointer, but that file pointer cannot be seeked, leading to the issue. A long time ago I think the original code was doing a loop + realloc to read the buffer by chunks and it was changed to use seek calls, this pull request basically restores that old code to read by chunks and avoid seeking.

### DIFF
--- a/winpr/libwinpr/timezone/timezone.c
+++ b/winpr/libwinpr/timezone/timezone.c
@@ -53,35 +53,42 @@ static UINT64 winpr_windows_gmtime(void)
 
 static char* winpr_read_unix_timezone_identifier_from_file(FILE* fp)
 {
-	INT64 length;
-	char* tzid = NULL;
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #7054, opened 2021-05-27
Disable poll usage on macOS/iOS because it is unreliable

awakecoding:
Long story short, poll on macOS/iOS sort of "works", but it is known to be unreliable. [The issue has been reported](https://github.com/curl/curl/issues/1057) [by the curl developers](https://daniel.haxx.se/blog/2016/10/11/poll-on-mac-10-12-is-broken/). Even if the blog post and original issue in the curl project are a few years old, it appears that poll on macOS has a long history of being unreliable, and there is no reason to believe it works any better on iOS. Because it can return too early without an actual timeout, we've seen threads taking too much CPU in some of our customer sites. The only way around it is unfortunately to stay away from poll on macOS/iOS an disable it.


akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #7055, opened 2021-05-27
iOS build system fixes

awakecoding:
The FreeRDP cmake build system for iOS makes a few dangerous hardcoded assumptions that should normally be handled in the cmake toolchain file. This pull request fixes many unconditional variable overrides that make it harder to use a proper iOS toolchain file.

- Don't override CMAKE_OSX_SYSROOT if it is already defined
- Don't override CMAKE_OSX_DEPLOYMENT_TARGET if it is already defined
- Don't override IOS_TARGET_SDK if it is already defined
- Don't append -isysroot -g flags (that should be in the toolchain file)
- Don't check for posix timer on iOS, it's not supported yet the test can return true


akallabeth:
@freerdp-bot test

akallabeth:
@awakecoding looks like the sysroot is not set in `cmake/iOSToolchain.cmake` or `ci/cmake-preloads/config-ios.txt` so the test is failing

awakecoding:
@akallabeth I will update cmake/iOSToolchain.cmake and ci/cmake-preloads/config-ios.txt then. I suppose the original iOS toolchain file didn't get much improving over the years, I've got a better one I maintain here: https://github.com/Devolutions/cbake

awakecoding:
@freerdp-bot test

awakecoding:
@akallabeth I just managed to make it build on my side by passing -DIOS_PLATFORM=SIMULATOR like what can be seen in the cmake cache, and using the same cmake toolchain file, and it works. I've triggered a new build because the logs from the last one that failed are no longer available, I want to what the failure is.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #7060, opened 2021-05-30
Fix JNI_OnLoad duplicate definition for Android

awakecoding:
libwinpr/timezone/timezone.c defines JNI_OnLoad() in order to get a JavaVM* handle on Android. This function can only be defined once per shared library, and it is also defined in the Android FreeRDP library, breaking the build against a static WinPR library.

In order to support both static and shared libraries, I have added a function to register the JavaVM* handle in WinPR such that it can be accessed from the function that requires it, but without using JNI_OnLoad().  The single JNI_OnLoad function in the Android library now calls winpr_set_java_vm to register the handle, and the conflict is now resolved.

### DIFF
--- a/client/Android/android_freerdp.c
+++ b/client/Android/android_freerdp.c
@@ -1093,7 +1093,10 @@ jint JNI_OnLoad(JavaVM* vm, void* reserved)
 	(*env)->RegisterNatives(env, activityClass, methods, sizeof(methods) / sizeof(methods[0]));
 	/* create global reference for class */
 	gJavaActivityClass = (*env)->NewGlobalRef(env, activityClass);
+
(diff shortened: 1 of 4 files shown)


akallabeth (changes requested):
This fix may work nicely for static builds, but it breaks if the assumption `winpr_get_unix_timezone_identifier_from_file` is called after `libfreerdp` has been loaded is broken.
I´d stay with the `JNI_OnLoad` for shared builds and add a `CMake` foo to wrap all necessary function calls (e.g. `winpr_JNI_OnLoad`, `freerdp_JNI_OnLoad` in a generated `JNI_OnLoad` ) similar to the channels that can be build statically or dynamically

PR #7061, opened 2021-05-30
Fix OpenSSL link library order on Android

awakecoding:
The order of link libraries passed to the compiler can be tricky to get right, especially when compiling static libraries into a larger shared library. OpenSSL has two libraries, and for some reason the order was incorrect on Android, breaking the build when using -DBUILD_SHARED_LIBS=OFF. It only happens on the first shared library linked against the static libraries, in this case, the Android library but also our own custom shared library suffered the same problem. This simple one-liner fixes the issue that was probably obscured by the current default compilation options used in the CI.


akallabeth (approved):

akallabeth:
@freerdp-bot be nice, test and osx please do run properly!

PR #7062, opened 2021-05-30
Fix Android ARM64 NEON enabling

awakecoding:
Enabling Android ARM64 NEON has always been a bit tricky, unfortunately the fix I did some time ago still appears required, otherwise `__ARM_NEON__` doesn't get defined and breaks rfx_neon.c with WITH_NEON=ON.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -598,6 +598,12 @@ if(ANDROID)
 		set (WITH_NEON OFF)
 	endif()
 
+	if(ANDROID_ABI STREQUAL arm64-v8a)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -598,6 +598,12 @@ if(ANDROID)
 		set (WITH_NEON OFF)
 	endif()
 
+	if(ANDROID_ABI STREQUAL arm64-v8a)
```
akallabeth, line 601: Referenced issue tells that `NDK r21` fixed this?

akallabeth (approved):

PR #7063, opened 2021-05-30
Fix broken channel addin.h relative include path

awakecoding:
This is one leaves me a bit perplexed: when building FreeRDP and WinPR separately, I got errors with the relative addin.h include done in two virtual channels. When I looked closely at the relative path, they were off by one parent directory (one missing ".."). I believe it may have worked by accident due to some other directory being added to the include path, but the real relative path is the one I've changed it to, and it fixes the build.

### DIFF
--- a/channels/cliprdr/client/cliprdr_main.c
+++ b/channels/cliprdr/client/cliprdr_main.c
@@ -32,7 +32,7 @@
 #include <freerdp/constants.h>
 #include <freerdp/client/cliprdr.h>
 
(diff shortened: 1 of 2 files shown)


akallabeth (approved):

akallabeth:
@freerdp-bot test

akallabeth:
@awakecoding nice one, that is really nasty.

akallabeth:
@freerdp-bot test, osx come on :/

PR #7064, opened 2021-05-30
Windows ARM64 build fixes

awakecoding:
Good news: FreeRDP/WinPR builds for Windows on ARM (Surface Pro X)!

There are two tiny little things to fix:
- Avoid defining `_AMD64_` and `_X86_` based on pointer size. It's not even needed and breaks Windows ARM64.
- The va_start macro doesn't accept the type cast in parameters, but it turns out we don't need it: the ArrayList_ForEachFkt callback type definition didn't use the `(*MY_CALLBACK)` notation used everywhere else.

That's it, it builds just fine now for Windows on ARM.


akallabeth:
@freerdp-bot test

akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #7073, opened 2021-06-02
Fixed issues with planar codec buffer alignment

akallabeth:
Align width and height to next multiple of 4 to avoid issues with
internal buffer sizes.
@antenore @pnowack please confirm I understood https://gitlab.com/Remmina/Remmina/-/issues/2507 correctly.

### DIFF
--- a/libfreerdp/codec/planar.c
+++ b/libfreerdp/codec/planar.c
@@ -33,6 +33,8 @@
 
 #define TAG FREERDP_TAG("codec")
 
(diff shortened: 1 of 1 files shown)


CruzR:
Original reporter of the linked Remmina issue here.

From a cursory glance at the code, this should work for me. Tomorrow is a public holiday where I am, so I can take some time to compile this commit and see if it fixes my issue.

akallabeth:
@CruzR ok. I´ve successfully reproduced the issue with `xfreerdp /v:server /size:642x480 /bpp:32` to a windows 7 remote. Gone with this pr

PR #7076, opened 2021-06-02
Win dyn resize

akallabeth:


hardening (approved):
LGTM, does it also work when you're in fullscreen and monitor resolution changes ?

akallabeth:
@hardening no idea, just did a fast check for windowed mode.

277hz:
@hardening As the window size should change in that case I suppose an adequate event will be emitted. However I did not test that specific case.

PR #7077, opened 2021-06-02
Test speedup

akallabeth:
Improves runtime speed of unit tests:

1. Set shorter timeouts
2. Reduce dataset sizes


akallabeth:
@mfleisz @bmiklautz @hardening please have a look, I´ve modified the unit tests to be a lot faster bringing down the time of each unit test `<1second` (except `TestConnect` which is still a lot faster now `~2.5 seconds`)

PR #7081, opened 2021-06-09
Added WINPR_ASSERT

akallabeth:
Added a WinPR macro that can be used for WinPR and FreeRDP to
replace standard C assert.
It is more verbose and logs issues to WLog which makes it easier
for us to trace problems in the field.
@hardening what do you think about this?

### DIFF
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -25,7 +25,7 @@
 #endif
 
 #include <errno.h>
-#include <assert.h>
+#include <winpr/assert.h>
 #include <stdio.h>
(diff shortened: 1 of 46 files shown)


hardening:
Adding a cmake option to expand WINPR_ASSERT to void would make it fantastic

akallabeth:
@hardening can do that, but I´d keep the log entry

akallabeth:
@hardening merging as is, we can improve what we find lacking with another pr.

akallabeth:
@hardening merged as is, we can improve the implementation with another pr if we find it lacking somewhere.

PR #7082, opened 2021-06-10
Cmake upgrade

akallabeth:
* Upgrade minimum required CMake version
* Fix `WinPR` standalone build
*  Fix `uwac` standalone build
* Fix `rdtk` standalone build


hardening:
do we still need rdtk ? I can't tell which piece of software is using this...

akallabeth:
@hardening shadow seems to use it

akallabeth:
@awakecoding @hardening I´ve completed the PR as far as it is now possible to compile each library standalone.
Build now works like this:

For each library listed below do: `cmake -B<builddir> -H<srcdir> -DCMAKE_PREFIX_PATH=<install path> <cmake options> && cmake --build <builddir> --target install`

1.  `WinPR`
2. `uwac`
3. `rdtk`
4. `FreeRDP`

be sure that no other `FreeRDP` headers/libraries are around, otherwise the seach path might need adjustment.

akallabeth:
@freerdp-bot test

awakecoding:
@akallabeth it is indeed possible to make the split, you just need to move a lot of code around. The tricky stuff is everything that goes in the generated config.h files. The vast majority of feature detection is currently in the root CMakeLists.txt, so a good half of it needs to be moved "down" to WinPR, or possibly checked twice in case the same header check is needed in both FreeRDP/WinPR. I'm not against making such a cleaner split, it's a lot more moving parts and things to check to make sure they aren't broken in the process.

awakecoding:
@akallabeth regarding RDTK: I made this project originally to support FreeRDS, as an equivalent of XRDP's in-memory login window rendering. It is officially dead weight because FreeRDS has been abandoned for a long time now. If nobody else used it in all those years, we should probably remove it entirely.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #7090, opened 2021-06-17
Warnings fixed

akallabeth:


winpr/libwinpr/utils/collections/HashTable.c
```diff
@@ -570,7 +570,7 @@ size_t HashTable_GetKeys(wHashTable* table, ULONG_PTR** ppKeys)
 		if (table->synchronized)
 			LeaveCriticalSection(&table->lock);
 
-		return -1;
+		return 0;
 	}
```
hardening, line 574: is it me or this doesn't allow to detect errors anymore ?

akallabeth: leftover, should be `WINPR_ASSERT`, true.

hardening: Hum I don't agree, OOM are normal conditions.

akallabeth: ok. still, this change was part of the previous #7012 pr, this one just fixes the return values.

PR #7096, opened 2021-06-19
Make able to be used as a cmake subproject

Pollux42:
This commit replaces CMAKE_SOURCE_DIR/CMAKE_BINARY_DIR occurrences with PROJECT_SOURCE_DIR/PROJECT_BINARY_DIR.

It should not change anything for those who are working directly on FreeRDP, but it enables using FreeRDP as a CMake subproject (which seems relevant as FreeRDP is a library)


akallabeth (changes requested):
#7082 should be what you want or not?

Pollux42:
No, it's not. (But my changes could be integrated to #7082 as it is almost the same topic.

Anyway, it's different, it's only about PROJECT_{BINARY,SOURCE}_DIR vs CMAKE_{BINARY,SOURCE}_DIR.
Let say I have a project MyProject that include FreeRDP inside (so we have the path MyProject/FreeRDP), and MyProject/CMakeLists.txt does _add_subdirectory(FreeRDP)_.

Then if you look at your CMakeLists.txt line 42: set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/)
CMAKE_MODULE_PATH will equal MyProject/cmake (which is wrong)
By replacing CMAKE_SOURCE_DIR by PROJECT_SOURCE_DIR, CMAKE_MODULE_PATH will be MyProject/FreeRDP/cmake which is what we want.

Pollux42:
I precise that if we use FreeRDP as top-level project, CMAKE_SOURCE_DIR and PROJECT_SOURCE_DIR will always be the same.

akallabeth:
@Pollux42 ah, ok. Will add that to the pr.

PR #7099, opened 2021-06-21
Fixed audin codec negociation

Pollux42:
Currently, there is a bug when try to open an RDP session to a server that has multiple DSP versions, and if the first one is not supported by the client: That first one will anyway be chosen because of the bug, so the audio in won't be correctly redirected.

This patch really check that the codec is supported before to choose it.

### DIFF
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -251,7 +251,7 @@ static UINT audin_process_formats(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* c
 			continue;
 		}
 
-		if (freerdp_dsp_supports_format(&format, TRUE) ||
+		if (freerdp_dsp_supports_format(&format, TRUE) &&
 		    audin->device->FormatSupported(audin->device, &format))
 		{
 			/* Store the agreed format in the corresponding index */


akallabeth (changes requested):
Sorry, that patch is wrong.
1. `freerdp_dsp_supports_format` checks if the format is supported by some encoder
1. `audin->device->FormatSupported` checks, if the format is supported natively by the audio backend.

Please describe your issue in a bit more detail, maybe we can find why you experience the problem.

Pollux42:
In practice, my issue was that I was not able to make my microphone works remotely in a certain configuration.
It was working with an old version of FreeRDP, and the issue appears when we upgrade FreeRDP to the last stable version.

I have realized that the chosen codec between the old and the recent version of FreeRDP was different, so I tried to figure out why.
Then I saw that during the codec negotiation, the `audin->device->FormatSupported` was never called on a DSP codec (as `freerdp_dsp_supports_format` always returns true for such a codec). So the client has finally chosen a DSP codec that was not supported by the audio backend (winmm, `the waveInOpen` did not return MMSYSERR_NOERROR).

So I don't understand your remark. Why the client would be able to choose a codec that is not supported by the audio backend?

akallabeth:
@Pollux42 the way the channel works is the following:
1. Check if we have a native format support from the backend for the chosen format (e.g. winmm produces `PCM` data, do not encode but just sent the data as is)
2. Check if we support the format with some encoder (e.g. `aac` works by encoding `PCM` to `aac` and then sending the data)

so, if you have an issue there, then your build is broken. (there are some configuration options regarding audio marked experimental)

PR #7113, opened 2021-06-23
Added extended shadow support for mouse and unicode

akallabeth:


hardening:
Does the extended input needs unicode ?

akallabeth:
@hardening no, but it enables the possibility for the client to send as unicode.

Issue #7067, opened 2021-05-31
Various warnings reported on the mingw CI that look pertinent 


hardening:
```
C:/_/mingw-w64-freerdp/src/freerdp-2.3.2/winpr/libwinpr/wtsapi/wtsapi_win32.c:63:13: warning: 'g_Initialized' defined but not used [-Wunused-variable]
     63 | static BOOL g_Initialized = FALSE;
        |             ^~~~~~~~~~~~~
```
True, this is never used in any build

```
[222/507] Building C object libfreerdp/CMakeFiles/freerdp.dir/core/license.c.obj
  C:/_/mingw-w64-freerdp/src/freerdp-2.3.2/libfreerdp/core/license.c: In function 'saveCal':
  C:/_/mingw-w64-freerdp/src/freerdp-2.3.2/libfreerdp/core/license.c:244:14: warning: passing argument 1 of 'DeleteFileW' from incompatible pointer type [-Wincompatible-pointer-types]
    244 |   DeleteFile(filepathNew);
        |              ^~~~~~~~~~~
        |              |
        |              char *
  In file included from D:/a/_temp/msys/msys64/mingw64/x86_64-w64-mingw32/include/winbase.h:18,
                   from D:/a/_temp/msys/msys64/mingw64/x86_64-w64-mingw32/include/windows.h:70,
                   from D:/a/_temp/msys/msys64/mingw64/x86_64-w64-mingw32/include/winsock2.h:23,
                   from C:/_/mingw-w64-freerdp/src/freerdp-2.3.2/winpr/include/winpr/windows.h:31,
                   from winpr/include/winpr/wtypes.h:36,
                   from C:/_/mingw-w64-freerdp/src/freerdp-2.3.2/winpr/include/winpr/platform.h:25,
                   from C:/_/mingw-w64-freerdp/src/freerdp-2.3.2/winpr/include/winpr/winpr.h:22,
                   from C:/_/mingw-w64-freerdp/src/freerdp-2.3.2/winpr/include/winpr/crt.h:27,
                   from C:/_/mingw-w64-freerdp/src/freerdp-2.3.2/libfreerdp/core/license.c:26:
  D:/a/_temp/msys/msys64/mingw64/x86_64-w64-mingw32/include/fileapi.h:166:50: note: expected 'LPCWSTR' {aka 'const short unsigned int *'} but argument is of type 'char *'
[...]

PR #7124, opened 2021-06-27
Cherry pick MinGW build fixes to stable-2.0

Biswa96:
This tries to fix some mingw build warnings as stated here https://github.com/FreeRDP/FreeRDP/issues/7067


akallabeth:
@freerdp-bot test

include/freerdp/server/rdpdr.h
```diff
@@ -35,20 +35,22 @@
...
 {
 	UINT32 NextEntryOffset;
 	UINT32 FileIndex;
-	UINT64 CreationTime;
-	UINT64 LastAccessTime;
-	UINT64 LastWriteTime;
-	UINT64 ChangeTime;
-	UINT64 EndOfFile;
-	UINT64 AllocationSize;
+	LARGE_INTEGER CreationTime;
```
akallabeth, line 43: This change is problematic for stable branch as the API breaks with it.

akallabeth (changes requested):
Overall LGTM, only the change in the public header (although correct) needs some discussion as it breaks public API

include/freerdp/server/rdpdr.h
Biswa96: According to MS docs [FILE_DIRECTORY_INFORMATION structure][1], those members are of LARGE_INTEGER type.

[1]: https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/ns-ntifs-_file_directory_information

libfreerdp/core/nla.h
```diff
@@ -58,7 +58,7 @@ FREERDP_LOCAL SEC_WINNT_AUTH_IDENTITY* nla_get_identity(rdpNla* nla);
 FREERDP_LOCAL NLA_STATE nla_get_state(rdpNla* nla);
 FREERDP_LOCAL BOOL nla_set_state(rdpNla* nla, NLA_STATE state);
 
-FREERDP_LOCAL BOOL nla_set_service_principal(rdpNla* nla, LPSTR principal);
+FREERDP_LOCAL BOOL nla_set_service_principal(rdpNla* nla, LPTSTR principal);
```
hardening, line 61: We avoid using `LPTSTR` that have different behaviours on msvc and gcc

PR #7145, opened 2021-07-02
Monitor coordinates are exclusive

akallabeth:


channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -1233,7 +1233,7 @@ static UINT rdpgfx_recv_surface_to_surface_pdu(RDPGFX_CHANNEL_CALLBACK* callback
 
 	Stream_Read_UINT16(s, pdu.destPtsCount); /* destPtsCount (2 bytes) */
 
-	if (Stream_GetRemainingLength(s) < (size_t)(pdu.destPtsCount * 4))
+	if (Stream_GetRemainingLength(s) < (pdu.destPtsCount * 4ULL))
 	{
```
hardening, line 1237: `Stream_GetRemainingLength(s)/4 < pdu.destPtsCount` so that there's no integer overflow ?

akallabeth: true, that also works. but since the left is already `size_t` it does not matter if the right is also promoted to such a type

mfleisz (approved):

Issue #7123, opened 2021-06-26
Proxy USB redirect


alexandrubagu-bss:
I am looking into allowing users to connect with remote usb. From what I see it looks like the `urbdrc` channel has only the client side implemented. 

Is there a way to pass-through the data that comes from either side without having to parse it or is it necessary to actually process all data and then repackage and send it to the other side?

akallabeth:
@alexandrubagu-bss yes, the usb channel has currently only the client sided implemented.
For the rest, please clarify. You´re talking about the proxy, right?
Channel pass through currently requires some parsing, but could be reworked to be pass through.

alexandrubagu-bss:
Yes, I am talking about the proxy. 
I want to be able to allow usb redirection in the proxy server. Can you give me some pointers as to how this can be implemented? I've been going through the proxy code related to dynamic channels and I still have yet to understand what exactly is going on there.

alexandrubagu-bss:
I've started working on a channel that would allow dynamic passthrough from one side to the other however I don't know exactly how I can tell each side to open a specific channel.
Am I supposed to send a DYNVC_CREATE_REQ to each side?

akallabeth:
@alexandrubagu-bss I think you are going in the wrong track here.

1. `proxy` actually is a `RDP-client` and a `RDP server` (on the proxy)
1. The `proxy` then copies the data it receives from the proxy-RDP-client and sents it (via the proxy=RDP-Server component) to the actual client and the other way around.
1. There are some additional things done (filter out some insecure modes, record a session if desired, ...)

So, what you want is not your own channel but low level access to the channels so you don´t need to decode -> reencode the whole data

hardening:
well from what I've read, the need is for a pass through channel implementation, so with minimal decoding/encoding (only the dynamic channel layer).

alexandrubagu-bss:
> @alexandrubagu-bss I think you are going in the wrong track here.
> 
> 1. `proxy` actually is a `RDP-client` and a `RDP server` (on the proxy)
> 2. The `proxy` then copies the data it receives from the proxy-RDP-client and sents it (via the proxy=RDP-Server component) to the actual client and the other way around.
> 3. There are some additional things done (filter out some insecure modes, record a session if desired, ...)
> 
> So, what you want is not your own channel but low level access to the channels so you don´t need to decode -> reencode the whole data

[...]

PR #7154, opened 2021-07-05
proxy dynamic channel passthrough support

alexandru-bagu:
this PR adds support for dynamic channel passthrough
the proxy config now allows you to set which channels you want proxied both for static as well as dynamic and it gives some hints as to what you may want to proxy
additionally this PR adds a log to inform you whether a channel is being actively ignored or not

### DIFF
--- a/channels/drdynvc/client/drdynvc_main.c
+++ b/channels/drdynvc/client/drdynvc_main.c
@@ -26,10 +26,12 @@
 #include <winpr/crt.h>
 #include <winpr/stream.h>
 
+#include <freerdp/client/dynamic_passthrough.h>
 #include "drdynvc_main.h"
 
 #define TAG CHANNELS_TAG("drdynvc.client")
 
+static UINT dvcman_close_passthrough_channel(DYNAMIC_PASSTHROUGH_DVCMAN_CHANNEL* channel);
 static UINT dvcman_close_channel(IWTSVirtualChannelManager* pChannelMgr, UINT32 ChannelId,
                                  BOOL bSendClosePDU);
 static void dvcman_free(drdynvcPlugin* drdynvc, IWTSVirtualChannelManager* pChannelMgr);
@@ -505,6 +507,7 @@ static UINT dvcman_create_channel(drdynvcPlugin* drdynvc, IWTSVirtualChannelMana
 	if (!ArrayList_Append(dvcman->channels, channel))
 		return ERROR_INTERNAL_ERROR;
 
+	BOOL found = FALSE;
 	ArrayList_Lock(dvcman->listeners);
 	for (i = 0; i < ArrayList_Count(dvcman->listeners); i++)
 	{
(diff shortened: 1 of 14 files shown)


alexandru-bagu:
PR for #7123

hardening:
I'm wondering about `freerdp/client/dynamic_passthrough.h` and `freerdp/server/dynamic_passthrough.h`, I'm not sure it has its place in public headers.

hardening:
I'm also wondering about all these extra threads, 2 new threads per passed-through dynamic channel looks like overkill.

kubistika:
There's no need for all of these changes.
The only thing is that the (already implemented) static vc passthrough needs to be fixed to buffer the VC messages and only then call the callback to forward them to the other side of the connection (currently it is not done so long messages like in drdynvc won't passthrough). After fixing this, you could just add "drdynvc" and "urbdrc" to the static passthrough channels list in proxy config and everything will work.

I already did fix it but I can't find the branch I did it in. Currently have no time for that, but I could elaborate if you need more info.

alexandru-bagu:
> I'm wondering about `freerdp/client/dynamic_passthrough.h` and `freerdp/server/dynamic_passthrough.h`, I'm not sure it has its place in public headers.
 `freerdp/server/dynamic_passthrough.h` shouldn't be public true. issue with `client` one is that is needed by both drydvnc client plugin and proxy.

> I'm also wondering about all these extra threads, 2 new threads per passed-through dynamic channel looks like overkill.
I just followed the 'standard' I saw in the code. Though I suppose there's only 1 new thread, since one channel is handled by drydvnc plugin.

> There's no need for all of these changes.
> The only thing is that the (already implemented) static vc passthrough needs to be fixed to buffer the VC messages and only then call the callback to forward them to the other side of the connection (currently it is not done so long messages like in drdynvc won't passthrough). After fixing this, you could just add "drdynvc" and "urbdrc" to the static passthrough channels list in proxy config and everything will work.
> 
> I already did fix it but I can't find the branch I did it in. Currently have no time for that, but I could elaborate if you need more info.
I welcome another PR that does this better. What you are describing is maybe the client side, but for the server-side you still have to join the dynamic channel and handle it somehow.

kubistika:
No, what I wrote is relevant to both server & client side.
Currently the passthrough callbacks just pass on the data to the other side of the connection (server -> client, client -> server), instead of first buffering the data (vc messages can be sent in chunks, and currently chunks are not handled by the proxy). That's why proxying the `drydnvc` isn't working now. But that could be easily fixed by properly handling the VC flags (CHANNEL_FIRST, CHANNEL_LAST, etc). Unfortunately I'm too busy now for creating a PR for that, but for sure, all of these changes are not needed for supporting `drdynvc` passthrough.

akallabeth:
Refactoring proxy in https://github.com/FreeRDP/FreeRDP/tree/proxy
@kubistika if you are not too busy, would love to have your feedback to that.

PR #7155, opened 2021-07-05
proxy - allow a way to have different configs for each session

alexandru-bagu:
this PR will allow proxy plugins to overwrite the config for the current session; split the config load function into 3 to allow either loading from a buffer or a file (a buffer might be more convenient if you don't want to keep any session information on disk). the idea is to allow each session to have their own settings (different clipboard/passthrough options etc). 
this functionality is meant to be used from by a plugin.

if a plugin does create a config and assigns it to the proxyData then on disconnect it will be freed (must be different from server->config).

### DIFF
--- a/server/proxy/pf_config.c
+++ b/server/proxy/pf_config.c
@@ -239,7 +239,68 @@ static BOOL pf_config_load_gfx_settings(wIniFile* ini, proxyConfig* config)
 	return TRUE;
 }
 
-proxyConfig* pf_server_config_load(const char* path)
+proxyConfig* server_config_load_ini(wIniFile* ini)
+{
+	proxyConfig* config = NULL;
+
+	config = calloc(1, sizeof(proxyConfig));
+	if (config)
+	{
+		if (!pf_config_load_server(ini, config))
+			goto out;
+
+		if (!pf_config_load_target(ini, config))
+			goto out;
+
+		if (!pf_config_load_channels(ini, config))
+			goto out;
+
+		if (!pf_config_load_input(ini, config))
+			goto out;
+
+		if (!pf_config_load_security(ini, config))
+			goto out;
+
+		if (!pf_config_load_modules(ini, config))
+			goto out;
+
+		if (!pf_config_load_clipboard(ini, config))
+			goto out;
+
+		if (!pf_config_load_gfx_settings(ini, config))
+			goto out;
+	}
+	return config;
+out:
+	pf_server_config_free(config);
+	return NULL;
+}
+
+proxyConfig* pf_server_config_load_buffer(const char* buffer)
+{
+	proxyConfig* config = NULL;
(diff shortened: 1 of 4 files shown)


alexandru-bagu:
this PR might lead to some confusions as to how you should overwrite the config from proxyData I image. should I add a comment as to how this should be done? maybe something along the lines of: 
```
typedef struct proxy_server
{
	/* config by default is the one used by the server. if you want to override it you will have to
	 load a new config from a file or a buffer and replace the pointer. do not change the values of
	 the config, just the pointer of the config */
	proxyConfig* config;
...
```

akallabeth:
@freerdp-bot test

akallabeth (commented):
Looks good to me,
@kubistika if time permits, could you add your opinion?

server/proxy/pf_config.c
```diff
@@ -239,7 +239,44 @@ static BOOL pf_config_load_gfx_settings(wIniFile* ini, proxyConfig* config)
 	return TRUE;
 }
 
-proxyConfig* pf_server_config_load(const char* path)
+proxyConfig* server_config_load_ini(wIniFile* ini)
```
akallabeth, line 242: should be `static proxyConfig* server_config_load_ini(wIniFile* ini)`

server/proxy/pf_server.c
```diff
@@ -470,6 +470,8 @@ static DWORD WINAPI pf_server_handle_peer(LPVOID arg)
 	pf_server_channels_free(ps);
 	LOG_INFO(TAG, ps, "freeing proxy data");
 	ArrayList_Remove(server->clients, pdata);
+	if (server->config != pdata->config)
```
akallabeth, line 473: I don´t see how this might happen with current code. Is this maybe part of some unrelated/follow up change?

akallabeth (changes requested):
Sorry it took so long, but our proxy expert seems to be busy.
Overall LGTM, just 2 small remarks to address, then we can merge this.

server/proxy/pf_server.c
alexandru-bagu: The way I thought this would work is a plugin could overwrite the pdata->config pointer with its own configuration (parsed using `server_config_load_ini` or `server_config_load_buffer`). 
There's a few issues that might have to be resolved like 
* having access to `server_config_load_*` from the plugins (I just  added pf_config.h and pf_config.c to plugin's CMakelists though I guess the header should have been enough if I also linked against the proxy)
* possibly having multiple plugins overwrite the pdata->config pointer (one fix would be a clone of the server->config when the connection is being established, that way every plugin can just change the settings in the config, not the config itself; this would remove the `if` and leave the `pf_server_config_free`)

akallabeth:
refactoring proxy at https://github.com/FreeRDP/FreeRDP/tree/proxy

PR #7156, opened 2021-07-06
winpr: Fix parsing of comma separated values if list is an empty string

mfleisz:
This PR fixes `CommandLineParseCommaSeparatedValuesEx` that incorrectly returned a count of 1 when an empty string `""` was passed in the list parameter.


hardening:
Good catch, any chance you would also update cmd line unitary test ?

mfleisz:
@hardening updated 👍

akallabeth (approved):

PR #7157, opened 2021-07-06
Fixes in rdpdr and smartcard channel

mfleisz:
This PR fixes two issues:
- The client name length was incorrectly calculated in `rdpdr_send_client_name_request`. Instead of the byte count the character count was used as the length causing scrambled strings in the session.
- The output buffer lenght calculation in `smartcard_irp_device_control_call` was wrong causing the whole smartcard channel to fail. Additionally this error also caused at least drive redirection to fail as well when smartcard redirection was activated (and probably others using rdpdr as well).

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -1089,9 +1089,9 @@ static UINT rdpdr_send_client_name_request(rdpdrPlugin* rdpdr)
 	if (!rdpdr->computerName[0])
 		gethostname(rdpdr->computerName, sizeof(rdpdr->computerName) - 1);
 
-	computerNameLenW = ConvertToUnicode(CP_UTF8, 0, rdpdr->computerName, -1, &computerNameW, 0);
+	computerNameLenW = ConvertToUnicode(CP_UTF8, 0, rdpdr->computerName, -1, &computerNameW, 0) * 2;
 	WINPR_ASSERT(computerNameLenW >= 0);
-	s = Stream_New(NULL, 16U + (size_t)computerNameLenW * 2U + 2U);
+	s = Stream_New(NULL, 16U + (size_t)computerNameLenW);
 
 	if (!s)
 	{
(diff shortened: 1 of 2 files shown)


channels/rdpdr/client/rdpdr_main.c
```diff
@@ -1254,7 +1253,7 @@ static UINT dummy_irp_response(rdpdrPlugin* rdpdr, wStream* s)
 	Stream_Write_UINT16(output, PAKID_CORE_DEVICE_IOCOMPLETION); /* PacketId (2 bytes) */
 	Stream_Write_UINT32(output, DeviceId);                       /* DeviceId (4 bytes) */
 	Stream_Write_UINT32(output, CompletionId);                   /* CompletionId (4 bytes) */
-	Stream_Write_INT32(output, STATUS_UNSUCCESSFUL);             /* IoStatus (4 bytes) */
+	Stream_Write_UINT32(output, STATUS_UNSUCCESSFUL);            /* IoStatus (4 bytes) */
```
akallabeth, line 1256: `STATUS_UNSUCCESSFUL` is a `NTSTATUS` so this needs a cast when converting to unsigned.

channels/rdpdr/client/rdpdr_main.c
```diff
@@ -1104,10 +1104,9 @@ static UINT rdpdr_send_client_name_request(rdpdrPlugin* rdpdr)
 	Stream_Write_UINT16(s, PAKID_CORE_CLIENT_NAME); /* PacketId (2 bytes) */
 	Stream_Write_UINT32(s, 1);                      /* unicodeFlag, 0 for ASCII and 1 for Unicode */
 	Stream_Write_UINT32(s, 0);                      /* codePage, must be set to zero */
-	Stream_Write_UINT32(s, (UINT32)computerNameLenW +
-	                           2U); /* computerNameLen, including null terminator */
+	Stream_Write_UINT32(s,
+	                    (UINT32)computerNameLenW); /* computerNameLen, including null terminator */
 	Stream_Write(s, computerNameW, (size_t)computerNameLenW);
-	Stream_Write_UINT16(s, 0); /* null terminator */
```
akallabeth, line 1110: Sure we do not need to write the `\0` terminator?

mfleisz: `computerNameW` is 0-terminated so the terminator is already written in the line above.

channels/rdpdr/client/rdpdr_main.c
mfleisz: Fixed

akallabeth (approved):

Issue #7192, opened 2021-07-23
White border when app is maximized


jobs-git:
**Summary**
White border appears when app is maximized (see https://github.com/Fmstrat/winapps/issues/223)

**Reproduction**
1. Open Excel App
2. Maximize it

![White border in maximized app](https://user-images.githubusercontent.com/11503879/119261232-cb1cf480-bbd6-11eb-9896-0795cc10972b.png)

xfreerdp --version
```
This is FreeRDP version 2.3.2 (2.3.2)
```
uname -srm
```
Linux 4.18.0-305.7.1.el8_4.x86_64 x86_64
```
lsb_release -a
```
LSB Version:	:core-4.1-amd64:core-4.1-noarch
Distributor ID:	CentOS
Description:	CentOS Linux release 8.4.2105
Release:	8.4.2105
Codename:	n/a
```

jobs-git:
what this fixed already?

jobs-git:
any progress?

jobs-git:
Thank you for looking into this, however when tested on xfreerdp 3.0.0 beta 4 but its not still fixed, there is still this white border

PR #7193, opened 2021-07-25
Fixes white border when maximizing x11 decoration

jobs-git:
Fixes https://github.com/FreeRDP/FreeRDP/issues/7192

A pseudo maximize seems to be implemented already by xfreerdp window when x11 decoration is dragged to top.

Note that this is more of a work around rather than a fix, because manually maximizing the internal app will force the white border to get displayed


akallabeth (changes requested):
Sorry, can´t merge this.
You´re `fixing` something that can not be fixed (at least not without quite a bunch of side effects) with a dirty hack. (e.g. you still can simply press the `maximize` button of the app)

jobs-git:
that is unfortunate, what is the proposed solution?

akallabeth:
@jobs-git that is indeed a good question (without proper answer so far)

Issue #7214, opened 2021-08-04
smartcard: invalid report of no card readers


hardening:
When running `certutil -v -scinfo` in my windows session with no yubikey inserted, I get the following message that seems to indicate that the answer to the `listReaders` call is invalid:

```console
C:\Users\Administrateur>certutil -v -scinfo
Le gestionnaire de ressource des cartes à puce est en cours d’exécution.
État de la carte/lecteur actuel :
SCardListReaders: Le relais a reçu des données incorrectes. 0x6f7 (WIN32: 1783 RPC_X_BAD_STUB_DATA)
Échec de SCardListReaders pour SCARD_ALL_READERS
Aucune liste de lecteurs de carte à puce n’a pu être déterminée.

Terminé.
CertUtil : -SCInfo ÉCHEC de la commande : 0x6f7 (WIN32: 1783 RPC_X_BAD_STUB_DATA)
CertUtil: Le relais a reçu des données incorrectes.

C:\Users\Administrateur>
```
I'd expect a message saying "no card reader".

Looking at the code it seems like it is related to a bad encoding of NDR pointer when the pointer is empty: there should have a pointer record with a 0 length, instead of nothing like now. Patch to come (that will also fix other smartcard related stuff).

hardening:
Applying #7215 I have the expected answer:
```
C:\Users\Administrateur>certutil -v -scinfo
Le gestionnaire de ressource des cartes à puce est en cours d’exécution.
État de la carte/lecteur actuel :
SCardListReaders: Impossible de trouver un lecteur de carte à puce. 0x8010002e (-2146435026 SCARD_E_NO_READERS_AVAILABLE)
Échec de SCardListReaders pour SCARD_ALL_READERS
Aucun lecteur de carte à puce n’est disponible actuellement.

Terminé.
CertUtil : -SCInfo ÉCHEC de la commande : 0x8010002e (-2146435026 SCARD_E_NO_READERS_AVAILABLE)
CertUtil: Impossible de trouver un lecteur de carte à puce.

C:\Users\Administrateur>
```

PR #7215, opened 2021-08-04
[WIP] smartcard: fix listing when no reader is available

hardening:
This patch fixes enumeration of readers when nothing is connected as in #7214.

### DIFF
--- a/channels/smartcard/client/smartcard_pack.c
+++ b/channels/smartcard/client/smartcard_pack.c
@@ -205,16 +205,13 @@ static BOOL smartcard_ndr_pointer_write(wStream* s, UINT32* index, DWORD length)
 	return TRUE;
 }
 
-static LONG smartcard_ndr_write(wStream* s, const BYTE* data, UINT32 size, UINT32 elementSize,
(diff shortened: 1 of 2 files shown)


channels/smartcard/client/smartcard_pack.c
```diff
@@ -205,17 +205,14 @@ static BOOL smartcard_ndr_pointer_write(wStream* s, UINT32* index, DWORD length)
 	return TRUE;
 }
 
-static LONG smartcard_ndr_write(wStream* s, const BYTE* data, UINT32 size, UINT32 elementSize,
+static LONG smartcard_ndr_write(wStream* s, const BYTE* data, UINT32 count, UINT32 elementSize,
```
akallabeth, line 208: unrelated changes

akallabeth (changes requested):
* The changes in `smartcard_ListReadersA_Call` and `smartcard_ListReadersW_Call` differ but should not (e.g. cchReaders = 0;)
* `smartcard_ndr_write` changes are unrelated

hardening:
Ok, comparing with packets sent by mstsc I figured that there's differences, from what I see it's a problem with padding. So I'm gonna clean up this PR to only fix the empty list of readers and then I'll push more fixes.

Issue #6968, opened 2021-04-20
TestUnicodeConversion fails on big endian targets


smaeul:
**Describe the bug**
The test added in https://github.com/FreeRDP/FreeRDP/pull/6803 is not endian-safe. The wide character literals are native endianness, however ConvertToUnicode always generates UTF16-LE.

**To Reproduce**
1. Run the test suite on a big endian target.
2. TestUnicodeConversion fails.

**Expected behavior**
The test suite passes.

**Screenshots**
```
(gdb) frame 0
#0  test_ConvertToUnicode_wrapper () at TestUnicodeConversion.c:416
416                     if (memcmp(wname, cmp, sizeof(cmp)) != 0)
(gdb) x/30c cmp
0x3fffffffebe2: 0 '\000'        115 's' 0 '\000'        111 'o' 0 '\000'        109 'm' 0 '\000'        101 'e'
0x3fffffffebea: 0 '\000'        116 't' 0 '\000'        101 'e' 0 '\000'        115 's' 0 '\000'        116 't'
0x3fffffffebf2: 0 '\000'        115 's' 0 '\000'        116 't' 0 '\000'        114 'r' 0 '\000'        105 'i'
0x3fffffffebfa: 0 '\000'        110 'n' 0 '\000'        103 'g' 0 '\000'        0 '\000'
(gdb) x/30c wname
0x3fffffffec00: 115 's' 0 '\000'        111 'o' 0 '\000'        109 'm' 0 '\000'        101 'e' 0 '\000'
0x3fffffffec08: 116 't' 0 '\000'        101 'e' 0 '\000'        115 's' 0 '\000'        116 't' 0 '\000'
0x3fffffffec10: 115 's' 0 '\000'        116 't' 0 '\000'        114 'r' 0 '\000'        105 'i' 0 '\000'
0x3fffffffec18: 110 'n' 0 '\000'        103 'g' 0 '\000'        0 '\000'        0 '\000'
```

**Application details**
* FreeRDP version: `master`

**Environment (please complete the following information):**
 - OS: Gentoo Linux
 - Architecture: `powerpc64-gentoo-linux-musl`

akallabeth:
@smaeul we´re lacking big endian test targets, could you adjust the test and create a pull request for that?

PR #7219, opened 2021-08-08
Fix TestUnicodeConversion on big endian machines

floppym:
Wide character literals are stored in native byte order.
Use an array of bytes as a reference instead.

Fixes: https://github.com/FreeRDP/FreeRDP/issues/6968


bmiklautz (approved):

bmiklautz:
@floppym thank you.

PR #7222, opened 2021-08-09
winpr/clipboard: Omit last line ending for gnome-copied-files

pnowack:
Quoting the commit here:
```
Every uri in the x-special/gnome-copied-files mime type is separated by
a '\n' char.
However, not every line ends with '\n'. The last line MUST not include
the new line character.
```

Without this commit, nautilus just crashes upon retrieving the content for the `x-special/gnome-copied-files` mime type.
Although the crash is actually a bug in nautilus, it is still the wrong format for the `x-special/gnome-copied-files` mime type.


hardening:

bmiklautz:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #7231, opened 2021-08-20
Some fixes for audin redirection

Pollux42:
This PR includes:

* Fixed DVI_ADPCM codec (in DSP_EXPERIMENTAL mode)
* Fixed channel mix and resampling
* Improve native codec research (by trying mono and stereo)
* Fix audin cli
* ...

### DIFF
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -426,6 +426,11 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
 		for (x = 0; x < ARRAYSIZE(samplerates); x++)
 		{
 			format.nSamplesPerSec = samplerates[x];
+			format.nChannels = audin->format->nChannels;
+			test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
+			if (test)
+				break;
+			format.nChannels = 3 - format.nChannels;
 			test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
 			if (test)
 				break;
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -944,7 +944,7 @@ BOOL audin_process_addin_args(AUDIN_PLUGIN* audin, const ADDIN_ARGV* args)
 		{
 			unsigned long val = strtoul(arg->Value, NULL, 0);
 
-			if ((errno != 0) || (val > UINT16_MAX))
+			if ((errno != 0) || (val < UINT16_MAX))
 				audin->fixed_format->nChannels = val;
 		}
 		CommandLineSwitchDefault(arg)
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -421,6 +421,7 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
 
 		format.wFormatTag = WAVE_FORMAT_PCM;
 		format.wBitsPerSample = 16;
+		format.cbSize = 0;
 		test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
 		if (!test)
 		{
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -448,7 +448,7 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
 
 	if (!supported)
 	{
-		if (!freerdp_dsp_context_reset(audin->dsp_context, audin->format))
+		if (!freerdp_dsp_context_reset(audin->dsp_context, audin->format, audin->FramesPerPacket))
 			return FALSE;
 	}
 
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -416,23 +416,19 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
 	if (!supported)
 	{
 		/* Default sample rates supported by most backends. */
-		const UINT32 samplerates[] = { 96000, 48000, 44100, 22050 };
+		const UINT32 samplerates[] = { format.nSamplesPerSec, 96000, 48000, 44100, 22050 };
 		BOOL test = FALSE;
+		size_t x;
 
 		format.wFormatTag = WAVE_FORMAT_PCM;
 		format.wBitsPerSample = 16;
 		format.cbSize = 0;
-		test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
-		if (!test)
+		for (x = 0; x < ARRAYSIZE(samplerates); x++)
 		{
-			size_t x;
-			for (x = 0; x < ARRAYSIZE(samplerates); x++)
-			{
-				format.nSamplesPerSec = samplerates[x];
-				test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
-				if (test)
-					break;
-			}
+			format.nSamplesPerSec = samplerates[x];
+			test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
+			if (test)
+				break;
 		}
 		if (!test)
 			return FALSE;
--- a/libfreerdp/codec/dsp.c
+++ b/libfreerdp/codec/dsp.c
@@ -138,7 +138,7 @@ static BOOL freerdp_dsp_channel_mix(FREERDP_DSP_CONTEXT* context, const BYTE* sr
 		return FALSE;
 
 	bpp = srcFormat->wBitsPerSample > 8 ? 2 : 1;
-	samples = size / bpp;
+	samples = size / bpp / srcFormat->nChannels;
 
 	if (context->format.nChannels == srcFormat->nChannels)
 	{
--- a/libfreerdp/codec/dsp.c
+++ b/libfreerdp/codec/dsp.c
@@ -1305,7 +1305,8 @@ BOOL freerdp_dsp_supports_format(const AUDIO_FORMAT* format, BOOL encode)
 #endif
 }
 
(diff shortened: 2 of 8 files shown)


libfreerdp/codec/dsp.c
```diff
@@ -238,9 +240,14 @@ static BOOL freerdp_dsp_resample(FREERDP_DSP_CONTEXT* context, const BYTE* src,
 	/* We want to ignore differences of source and destination format. */
 	format = *srcFormat;
 	format.wFormatTag = WAVE_FORMAT_UNKNOWN;
+	format.wBitsPerSample = 0;
```
akallabeth, line 243: Maybe just initialize like this: `AUDIO_FORMAT format = { 0 };`

libfreerdp/codec/dsp.c
```diff
@@ -261,10 +268,6 @@ static BOOL freerdp_dsp_resample(FREERDP_DSP_CONTEXT* context, const BYTE* src,
 	*length = Stream_Length(context->resample);
 	return (error == 0) ? TRUE : FALSE;
 #else
-	WINPR_UNUSED(src);
-	WINPR_UNUSED(size);
-	WINPR_UNUSED(data);
-	WINPR_UNUSED(length);
```
akallabeth, line 267: Why remove these?

libfreerdp/codec/dsp.c
```diff
@@ -1307,6 +1325,22 @@ BOOL freerdp_dsp_context_reset(FREERDP_DSP_CONTEXT* context, const AUDIO_FORMAT*
 		return FALSE;
 
 	context->format = *targetFormat;
+
+	if (context->format.wFormatTag == WAVE_FORMAT_DVI_ADPCM)
+	{
+		size_t min_frame_data =
+		    context->format.wBitsPerSample * context->format.nChannels * FramesPerPacket;
```
akallabeth, line 1332: like the codeql error tells. simply add `*1ULL` or cast all to `size_t` to fix.

libfreerdp/codec/dsp.c
```diff
@@ -1254,6 +1270,7 @@ BOOL freerdp_dsp_supports_format(const AUDIO_FORMAT* format, BOOL encode)
 #if defined(WITH_DSP_EXPERIMENTAL)
 
 		case WAVE_FORMAT_ADPCM:
+			return FALSE;
 		case WAVE_FORMAT_DVI_ADPCM:
```
akallabeth, line 1274: Is this still experimental after this fixes?

libfreerdp/codec/dsp.c
```diff
@@ -768,9 +772,15 @@ static BOOL freerdp_dsp_encode_ima_adpcm(FREERDP_DSP_CONTEXT* context, const BYT
 			*dst++ = encoded;
 			size -= 4;
 		}
+
+		if (dst - start == context->adpcm.ima.packet_size)
+		{
+			Stream_Write(out, start, context->adpcm.ima.packet_size);
+			dst = Stream_Buffer(context->buffer);
```
akallabeth, line 779: sure you want to ignore all data after `packet_size`?

libfreerdp/codec/dsp.c
```diff
@@ -768,9 +772,15 @@ static BOOL freerdp_dsp_encode_ima_adpcm(FREERDP_DSP_CONTEXT* context, const BYT
...
+		}
 	}
 
-	Stream_SetPointer(out, dst);
+	Stream_SetPointer(context->buffer, dst);
```
akallabeth, line 783: not really sure what you try to do with this line.

akallabeth (changes requested):
Overall LGTM,
some (small) refinements are necessary though.
Anyway, thank you for this nice addition!

libfreerdp/codec/dsp.c
Pollux42: If we go deeper in details, the audio_format_compatible checks four things:
* the format tag, but we ignore it here
* the bits per sample, now the same
* the number of channel, but it's always the same because of line 1166 (later instruction just before the call to freerdp_dsp_resample).
* the number of sample per seconds, the only one, finally

So finally, we could directly replace the audio_format_compatible call by comparing only the number of sample per seconds.

But If I'm wrong, it means that the audio_format_compatible still need to compare the number of channel, and so you ask me to replace:

```
	format = *srcFormat;
	format.wFormatTag = WAVE_FORMAT_UNKNOWN;
	format.wBitsPerSample = 0;
```
by
```
	AUDIO_FORMAT format = { 0 };
	[...]
	format.nChannels = srcFormat->nChannels;
	format.nSamplesPerSec = srcFormat->nSamplePerSec;
```

What do you choose?

libfreerdp/codec/dsp.c
Pollux42: Because they are now always used. I have added:

```
		*data = src;
		*length = size;
```

...in the "Fix dsp resemple" commit.

libfreerdp/codec/dsp.c
akallabeth: @Pollux42 looks like I overlooked `format = *srcFormat;` here, so the original is actually quite fine. (Just reset what you want to ignore in the comparison), sorry for the noise.

channels/audin/client/audin_main.c
```diff
@@ -416,22 +416,24 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
...
-				test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
-				if (test)
-					break;
-			}
+			format.nSamplesPerSec = samplerates[x];
+			format.nChannels = audin->format->nChannels;
+			test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
+			if (test)
+				break;
+			format.nChannels = 3 - format.nChannels;
```
akallabeth, line 433: looks strange, so you want to check if the format is compatible with different channel count? better run a loop instead of magic (e.g. `3 - 1` or `3 - 2` in this case here)

libfreerdp/codec/dsp.c
Pollux42: For me it works, but...
Do you mean that "experimental" means "does not work"? :-)

Anyway I let you judge. I understood that part is marked experimental because not well tested (as the most tested part is the FFMPEG part). I just make it work for my usage, I don't guarantee that it works for all use cases compatible with DVI_ADPCM.

About the diff above, I just wanted to flag that the WAVE_FORMAT_ADPCM was not working, and so I wanted to prevent its usage, even in WITH_DSP_EXPERIMENTAL. (The issue with that codec is probably the same as the one I fixed about DVI_ADPCM, but I just needed to make DVI_ADPCM work, so the ADPCM was out of my scope)

akallabeth: ok, got it. well, then lets keep it this way and change once we have some field data from test installations :)

libfreerdp/codec/dsp.c
Pollux42: No, I don't ignore the data. The encoded data is still in context->buffer and will be recovered in the next call to `freerdp_dsp_encode_ima_adpcm`

libfreerdp/codec/dsp.c
Pollux42: Maybe I need to better explain what the issue was, and what I did to fix it.
Before, this encoding function was in JIT production without buffering. It was working before the deep audin refactoring (those who decouple backend and dsp) because the input of this function had already been properly sized.
But now, this function gets data directly from the backend, and the backend does not guarantee that the data has always the same and the good size.
Since the protocol needs an alignment and a minimal frame number per packet, I have added some buffering here to ensure we output properly sized packets.

So here at the end of the function, I do not modify the output buffer, but only the intermediate buffer (that contain well-encoded data, but does not contain enough data to be sent yet)

channels/audin/client/audin_main.c
Pollux42: I have found my method more elegant than a loop because I did not want to simply iterate over 1 and 2 channels. I want that iteration always begins with the same number of channels as the currently chosen format.
I don't think that using a loop with that kind of behavior will be simpler. Actually, I don't see how to express that particular behavior without doing something "strange" or excessively verbose.

libfreerdp/codec/dsp.c
Pollux42: Done

channels/audin/client/audin_main.c
akallabeth: well, `3 ` is a magic number here, that is why I did ask. Not that important though.

libfreerdp/codec/dsp.c
akallabeth: Ok, I see that now. The way you did it confused me a bit. Maybe better use `Stream_GetPosition` and just do a `Stream_SetPosition(buffer, 0)` on write?

akallabeth: Ok, see why you did it this way now, ok, we can leave it as is.

libfreerdp/codec/dsp.c
akallabeth: See comment above

akallabeth:
@freerdp-bot test

channels/audin/client/audin_main.c
```diff
@@ -416,22 +416,24 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
...
-		const UINT32 samplerates[] = { 96000, 48000, 44100, 22050 };
+		const UINT32 samplerates[] = { format.nSamplesPerSec, 96000, 48000, 44100, 22050 };
 		BOOL test = FALSE;
 
 		format.wFormatTag = WAVE_FORMAT_PCM;
 		format.wBitsPerSample = 16;
-		test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
-		if (!test)
+		format.cbSize = 0;
+		size_t x;
```
akallabeth, line 425: Variables are only allowed at beginning of scope (older compilers do not like mid code declarations, e.g. VS2010)

Pollux42: Done

akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #7233, opened 2021-08-23
Fixed check of orderLength for secondary orders

akallabeth:
The orderLength field of [MS-RDPEGDI] 2.2.2.2.1.2.1.1 might be
negative if not adjusted for the 13byes of the header.


mfleisz (approved):
lgtm

PR #7241, opened 2021-08-24
Marked deprecated symbols

akallabeth:


akallabeth:
merging to get some (soft) breakage in dependent projects. (they can complain and enable the deprecated symbols without too much pain for the time being)

PR #7242, opened 2021-08-24
Proxy refactoring

akallabeth:
* Allows using proxy as a library
* Refactored configuration/proxy/... API to be opaque
* Moved module and proxy API related headers to public headers
* Refactored module API to be instance aware


include/freerdp/server/proxy/proxy_modules_api.h
```diff
@@ -0,0 +1,189 @@
...
+ * Main API for use by external modules.
+ * Supports:
+ *  - Registering a plugin.
+ *  - Setting/getting plugin's per-session specific data.
+ *  - Aborting a session.
+ */
+struct proxy_plugins_manager
+{
+	/* 0 used for registering a fresh new proxy plugin. */
+	BOOL(*RegisterPlugin)
```
hardening, line 114: missing space after `BOOL` ?

server/proxy/pf_client.c
```diff
@@ -216,73 +241,211 @@ static BOOL pf_client_pre_connect(freerdp* instance)
...
+			if ((flags & CHANNEL_FLAG_FIRST) &&
+			    (strncmp(channel_name, DRDYNVC_SVC_CHANNEL_NAME, CHANNEL_NAME_LEN + 1) == 0))
+			{
+				BYTE cmd;
+				if (size < 1)
+					return FALSE;
+
+				cmd = data[0] >> 4;
+				if (cmd == 0x01)
+				{
```
hardening, line 332: Should use defines here instead if `0x01`

server/proxy/pf_client.c
```diff
@@ -216,73 +241,211 @@ static BOOL pf_client_pre_connect(freerdp* instance)
...
+					proxyChannelDataEventInfo dev;
+					size_t len, nameLen;
+					const char* name;
+					UINT32 dynChannelId;
+					BYTE cbId = data[0] & 0x03;
+					switch (cbId)
+					{
```
hardening, line 339: Here I don't like the way wire data are parsed, should use `Stream_Read_XXX` functions instead (using a `Stream_Static_Init` on `data`).

server/proxy/pf_client.c
```diff
@@ -336,22 +507,19 @@ static BOOL pf_client_post_connect(freerdp* instance)
...
-			HashTable_Insert(pc->vc_ids, (void*)channel_name, (void*)channel_id);
-		}
-	}
-
 	instance->heartbeat->ServerHeartbeat = pf_client_on_server_heartbeat;
 
+	pc->connected = TRUE;
+
+	/* Send cached channel data */
+	ArrayList_Lock(pc->cached_server_channel_data);
```
hardening, line 518: I was wondering, if this `ArrayList_Lock/Unlock` is required as `ArrayList_ForEach` does the locking internally  (although there's the `ArrayList_Clear` after ) ?

server/proxy/pf_client.c
```diff
@@ -483,22 +684,27 @@ static BOOL pf_client_connect(freerdp* instance)
...
 	freerdp* instance = (freerdp*)arg;
-	pClientContext* pc = (pClientContext*)instance->context;
-	proxyData* pdata = pc->pdata;
-	DWORD nCount;
+	pClientContext* pc;
+	proxyData* pdata;
+	DWORD nCount = 0;
 	DWORD status;
-	HANDLE handles[65];
+	HANDLE handles[65] = { 0 };
```
hardening, line 691: Should not be `65` but the magic constant for the maximum wait handles (can't remember the name)

server/proxy/pf_client.c
```diff
@@ -560,91 +774,126 @@ static int pf_logon_error_info(freerdp* instance, UINT32 data, UINT32 type)
...
-	pClientContext* pc = (pClientContext*)context;
+	const proxyChannelDataEventInfo* src = obj;
+	proxyChannelDataEventInfo* dst;
 
-	if (!pc)
-		return;
+	WINPR_ASSERT(src);
+
+	dst = calloc(1, sizeof(proxyChannelDataEventInfo));
+	WINPR_ASSERT(dst);
```
hardening, line 833: I don't think an assert makes sense here

hardening (approved):
Need some fixes in some places, but apart from that looks good.

server/proxy/pf_client.c
akallabeth: too much overhead here, we just look at the first byte

server/proxy/pf_client.c
akallabeth: yes, as otherwise a new entry could be added before the clear resulting in duplicate data sent.

server/proxy/pf_client.c
akallabeth: true

server/proxy/pf_client.c
akallabeth: true, should abort.

server/proxy/pf_client.c
hardening: Well that's not entirely true as the name is also read, and the channel id. `Stream_Static_Init` is really light, that would remove the reading of channel id by hand, and the name would be retrieved by a simple `Stream_Pointer()`. I can't remember but I think we also have functions that read a null terminated string from a stream...
Except if I missed something and this function is the proxy's bottleneck, I don't see why we would do everything "by hand" here and use regular Stream functions everywhere else.

akallabeth:
@hardening addressed your issues, added a few commits fixing problems with the proxy (e.g. client resource cleanup did some nasty `assert` on out of memory)

PR #7267, opened 2021-09-03
Core cleanup

akallabeth:


server/Sample/sfreerdp.c
```diff
@@ -823,15 +925,23 @@ static DWORD WINAPI test_peer_mainloop(LPVOID arg)
...
 	WLog_INFO(TAG, "We've got a client %s", client->local ? "(local)" : client->hostname);
 
 	while (error == CHANNEL_RC_OK)
 	{
 		count = 0;
 		{
-			DWORD tmp = client->GetEventHandles(client, &handles[count], 32 - count);
+			DWORD tmp;
+
+			WINPR_ASSERT(client->GetEventHandles);
```
mfleisz, line 943: assert for `GetEventHandles` and `CheckFileDescriptor` in every loop iteration doesn't make sense - assert before the loop is enough (same in `test_server_mainloop`)

mfleisz (approved):
overall lgtm

server/Sample/sfreerdp.c
akallabeth: I think since internals might change function pointers due to bugs at any time it is (at least) for the sample server not a bad idea to always check.

PR #7278, opened 2021-09-09
Warnings fixed, sfreerdp allow to specify certificate

akallabeth:


libfreerdp/core/transport.h
```diff
@@ -113,8 +113,8 @@ FREERDP_LOCAL BOOL transport_set_connected_event(rdpTransport* transport);
 FREERDP_LOCAL BOOL transport_set_recv_callbacks(rdpTransport* transport, TransportRecv recv,
                                                 void* extra);
 
-FREERDP_LOCAL transport_tcp_connect(rdpTransport* transport, const char* hostname, int port,
-                                    DWORD timeout);
+FREERDP_LOCAL int transport_tcp_connect(rdpTransport* transport, const char* hostname, int port,
+                                        DWORD timeout);
```
hardening, line 117: good catch

winpr/include/winpr/synch.h
```diff
@@ -293,9 +293,18 @@ extern "C"
 	InitializeCriticalSectionAndSpinCount(lpCriticalSection, dwSpinCount)
 #endif
 
+#if defined(__clang__)
```
hardening, line 296: what about putting this enter block and exit block in a specific header and just include it ?

mfleisz (approved):
LGTM - agree that putting the `#pragma`s in a separate header file would be a nice idea

winpr/include/winpr/synch.h
akallabeth: good idea, so we could catch compiler differences better.

akallabeth:
I´ll merge this and add the silence warning macro in another one.

PR #7282, opened 2021-09-10
Added proxy input state sync

akallabeth:
The proxy server component might receive input related events
before the proxy client has established the connection to the
target machine.
With this change, the current keyboard state is cached and sent
to the target when it is ready. All input events received before
the target is ready are discarded.


mfleisz (approved):
lgtm

PR #7285, opened 2021-09-10
Replace backslashes with underscores in printer names

DavBfr:
On a Windows workstation, the network printers look like this: `\\DOMAIN\PRINTER` but Windows Server refuses to redirect a printer having a `\` in its name.

This fix just replaces the `\` with `_`.

### DIFF
--- a/channels/printer/client/printer_main.c
+++ b/channels/printer/client/printer_main.c
@@ -300,6 +300,7 @@ static BOOL printer_load_from_config(const rdpSettings* settings, rdpPrinter* pr
 	void* CachedPrinterConfigData = NULL;
 	UINT32 CachedFieldsLen = 0;
 	UINT32 PrinterNameLen = 0;
+	WCHAR* wptr = NULL;
 
 	if (!settings || !printer)
 		return FALSE;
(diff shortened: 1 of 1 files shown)


hardening:

hardening:
LGTM, but shouldn't we have the same treatment for the save config treatment ?

akallabeth:
Agree with @hardening 
Should not only check this in the `RDPDR_ADD_PRINTER_EVENT` enum value settings.
Manually changed configurations should be discarded entirely.

What way can this be exposed?

hardening:
@DavBfr any news regarding our comments ?

akallabeth:
@DavBfr your changes are in `printer_load_from_config` but it should also be handled in `printer_save_to_config`

DavBfr:
@akallabeth This name is changed only for the server, it doesn't matter what the local configuration filename is as it's hashed anyway so no filename issue here.
I moved the renaming just before sending it to the server.

akallabeth:
@freerdp-bot test please

akallabeth (approved):
LGTM

PR #7290, opened 2021-09-13
Add Findx264.cmake

frispete:
when building FreeRDP with x264 option enabled, I noticed, that this
isn't doing well if this package installed in a standard linux environment,
because of missing cmake instructions.

### DIFF
--- a/cmake/Findx264.cmake
+++ b/cmake/Findx264.cmake
@@ -1,0 +1,14 @@
+                
+find_path(X264_INCLUDE_DIR x264.h)
+                                
+find_library(X264_LIBRARY x264) 
+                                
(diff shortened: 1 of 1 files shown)


akallabeth:
Thanks for the effort, but without a `x264` backend this pr is quite useless ;)

frispete:
Hmm, what does -DWITH_X264=ON then?

hardening:
@frispete we don't have any implementation for x264, could you change your patch to remove completely x264 related items ? Or if you prefer, you can also ship the corresponding implementation doing encoding and decoding using x264 ;)

frispete:
Will do, thanks.

PR #7302, opened 2021-09-19
Fix X11 floatbar: XGetWindowAttributes returns zero on error

frispete:
This effectively disables the floatbar.

See https://www.x.org/releases/X11R7.7/doc/libX11/libX11/libX11.html


hardening (approved):
LGTM, good catch

hardening:
@freerdp-bot can you test that

PR #7303, opened 2021-09-19
rdpei/server: Fix PDU length for RDPINPUT_PROTOCOL_V300

pnowack:

```
When the server supports the protocol version RDPINPUT_PROTOCOL_V300,
the additional supportedFeatures field will be present.
The pduLength in the RDPINPUT_HEADER should, however, reflect this.

So, fix this error by writing the correct PDU length when the
supportedFeatures field is present.
```

**Note:** This is compile tested only. It should however be correct. Found this, while reading through the source code.

See also `2.2.3.1 RDPINPUT_SC_READY_PDU` in `[MS-RDPEI]`


akallabeth:
@freerdp-bot test

akallabeth (approved):

akallabeth:
@pnowack nice catch!

PR #7307, opened 2021-09-21
Prefer constant division over multiplication for length checks

akallabeth:
@hardening thanks for pointing out we´ve still had this.

### DIFF
--- a/channels/urbdrc/client/data_transfer.c
+++ b/channels/urbdrc/client/data_transfer.c
@@ -801,7 +801,7 @@ static UINT urb_isoch_transfer(IUDEVICE* pdev, URBDRC_CHANNEL_CALLBACK* callback
 	Stream_Read_UINT32(s, NumberOfPackets); /** NumberOfPackets */
 	Stream_Read_UINT32(s, ErrorCount);      /** ErrorCount */
 
-	if (Stream_GetRemainingLength(s) < NumberOfPackets * 12 + 4)
+	if (Stream_GetRemainingLength(s) < NumberOfPackets * 12ULL + 4ULL)
 		return ERROR_INVALID_DATA;
 
 	packetDescriptorData = Stream_Pointer(s);
--- a/libfreerdp/core/orders.c
+++ b/libfreerdp/core/orders.c
@@ -2409,7 +2409,7 @@ static CACHE_COLOR_TABLE_ORDER* update_read_cache_color_table_order(rdpUpdate* u
 		goto fail;
 	}
 
-	if (Stream_GetRemainingLength(s) < cache_color_table->numberColors * 4)
+	if (Stream_GetRemainingLength(s) / 4 < cache_color_table->numberColors)
 		goto fail;
 
 	colorTable = (UINT32*)&cache_color_table->colorTable;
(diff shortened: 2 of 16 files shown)


channels/urbdrc/client/data_transfer.c
```diff
@@ -801,7 +801,7 @@ static UINT urb_isoch_transfer(IUDEVICE* pdev, URBDRC_CHANNEL_CALLBACK* callback
 	Stream_Read_UINT32(s, NumberOfPackets); /** NumberOfPackets */
 	Stream_Read_UINT32(s, ErrorCount);      /** ErrorCount */
 
-	if (Stream_GetRemainingLength(s) < NumberOfPackets * 12 + 4)
+	if (Stream_GetRemainingLength(s) < NumberOfPackets * 12ULL + 4ULL)
 		return ERROR_INVALID_DATA;
```
hardening, line 805: AFAICT this one does not address the possible overflow problem

libfreerdp/core/update.c
```diff
@@ -2122,7 +2122,7 @@ BOOL update_read_refresh_rect(rdpUpdate* update, wStream* s)
 	Stream_Read_UINT8(s, numberOfAreas);
 	Stream_Seek(s, 3); /* pad3Octects */
 
-	if (Stream_GetRemainingLength(s) < ((size_t)numberOfAreas * 4 * 2))
+	if (Stream_GetRemainingLength(s) / 2 / 4 < numberOfAreas)
 		return FALSE;
```
hardening, line 2126: `/ 8` would be more explicit, I had to re-read it twice to be sure it was correct

akallabeth: kept it the same as it was before (e.g. 4 coordinates, 2 bytes size for each)

channels/urbdrc/client/data_transfer.c
akallabeth: it does, as it forces the data type to be `uint64`. `NumberOfPackets` is `UINT32`

libfreerdp/core/orders.c
```diff
@@ -2664,7 +2664,7 @@ static BOOL update_decompress_brush(wStream* s, BYTE* output, size_t outSize, BY
 	const BYTE* palette = Stream_Pointer(s) + 16;
 	const size_t bytesPerPixel = ((bpp + 1) / 8);
 
-	if (Stream_GetRemainingLength(s) < 16 + bytesPerPixel * 4)
+	if (Stream_GetRemainingLength(s) < 16ULL + bytesPerPixel * 4ULL)
 		return FALSE;
```
hardening, line 2668: Like above to address all the overflow cases, we should probably do a 
`if ((Stream_GetRemainingLength(s) < 16ULL) || ((Stream_GetRemainingLength(s) -16) / 4 < bytesPerPixel)`

hardening (changes requested):
Except my remarks LGTM

libfreerdp/core/orders.c
akallabeth: not required due to range of input data type. `bpp` is (at most) `UINT32` and we operate on `UINT64` here.

akallabeth: sorry, even more narrow, at most `UINT8`

libfreerdp/core/update.c
akallabeth: done.

PR #7312, opened 2021-09-23
Refactor proxy

akallabeth:
* Add simple RDP stream dump/replay option to library. (not exposed as it is currently meant for testing only)
* Remove obsolete proxy channel files, we do no longer decode channel data and do not need these.
* Added new filter options for static channels
* Added configuration more settings 
* Apply configuration settings via filter


mfleisz (approved):
LGTM

PR #7315, opened 2021-09-29
Audin winmm

Pollux42:
Small fixes about audin and winmm.

### DIFF
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -431,11 +431,14 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
 			test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
 			if (test)
 				break;
-			format.nChannels = 3 - format.nChannels;
+			/* Try with different number of channels */
+			format.nChannels = format.nChannels == 1 ? 2 : 1;
 			format.nBlockAlign = 2 * format.nChannels;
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test
@Pollux42 thanks you for the pr, but could you please add a description which issue this fixes (and how to test)?

Pollux42:
@akallabeth Indeed.

In a recent commit, in audin_winmm_set_format, you have forced the number of channels to one. But this should also be done in audin_winmm_format_supported so the dsp is properly initialized to do a proper channel mix.
This is the purpose of the first commit.

Unfortunately, this still does not work because waveInOpen refuse the format if the channel number and the block alignment are not consistent.
So this is the purpose of the second commit.

The last commit makes sure we always first try the same number of channels for each sample rate attempt. (plus some refactoring)

akallabeth (changes requested):
The changes in `channels/audin/client/audin_main.c` might work for you, but you change these windows backend limitations for every supported platform.

akallabeth:
Ok, refactored your pr slightly and created a new pr.
Thank you for your contribution, indeed a nice bugfix!

Pollux42:
> The changes in channels/audin/client/audin_main.c might work for you, but you change these windows backend limitations for every supported platform.

I did not report those windows limitations for every supported platform. All changes I did in `channels/audin/client/audin_main.c` should have no negative impact. Which change in this file makes you suspicious?

akallabeth:
@Pollux42 I should stop reviewing before coffee, formulated the issue badly.

PR #7316, opened 2021-09-30
Force one channel input fallback for audin

akallabeth:
Windows winmm backend does not support stereo microphone properly.
Fall back to mono and ensure that fallback is used.
Replaces #7315

### DIFF
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -425,13 +425,16 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
 		format.cbSize = 0;
 		for (x = 0; x < ARRAYSIZE(samplerates); x++)
 		{
+			size_t y;
 			format.nSamplesPerSec = samplerates[x];
-			format.nChannels = audin->format->nChannels;
-			test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
-			if (test)
-				break;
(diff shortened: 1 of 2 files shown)


channels/audin/client/audin_main.c
```diff
@@ -425,13 +425,16 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
...
 			format.nSamplesPerSec = samplerates[x];
-			format.nChannels = audin->format->nChannels;
-			test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
-			if (test)
-				break;
-			format.nChannels = 3 - format.nChannels;
-			test = IFCALLRESULT(FALSE, audin->device->FormatSupported, audin->device, &format);
+			for (y = audin->format->nChannels; y > 0; y--)
+			{
+				format.nChannels = audin->format->nChannels;
```
Pollux42, line 432: If I understand what you wanted to do, you should write 
`format.nChannels = y` instead of `format.nChannels = audin->format->nChannels`
...otherwise, I don't see the point of the iterator y.

Anyway, this won't work well, because if `audin->format->nChannels` is 1, you won't test with 2 channels, and this was the purpose of the code you replace.

akallabeth: ok, never commit before coffee, thank you @Pollux42 :)

Pollux42: No problem, but have you got your coffee now?
Because I still don't see the point of this diff. :-)

PR #7320, opened 2021-10-03
Fix double-free of rdpCertificateStore members

smaeul:
certificate_store_uninit is always called from certificate_store_free
before freeing the certificate store. Since the certificate store is
freed if certificate_store_init fails, calling certificate_store_uninit
from certificate_store_init means it gets called twice, and all of the
struct members get double-freed.


akallabeth (changes requested):
You´re right about the problem, but your solution touches the wrong part of code.
`certificate_store_uninit` should reset the pointers to `NULL` to fix this.

PR #7322, opened 2021-10-04
Cert store fix && Stream API refinements

akallabeth:
Replaces #7320 and #7321 

* Fix a missing `NULL` reset after free in certificate store
* Adds validity checks for `wStream API`
* Removes problematic `wStream` setter functions that are not strictly necessary.

### DIFF
--- a/channels/tsmf/client/tsmf_codec.c
+++ b/channels/tsmf/client/tsmf_codec.c
@@ -569,7 +569,7 @@ BOOL tsmf_codec_parse_media_type(TS_AM_MEDIA_TYPE* mediatype, wStream* s)
 
 BOOL tsmf_codec_check_media_type(const char* decoder_name, wStream* s)
 {
-	BYTE* m;
+	size_t pos;
 	BOOL ret = FALSE;
 	TS_AM_MEDIA_TYPE mediatype;
 
@@ -583,10 +583,10 @@ BOOL tsmf_codec_check_media_type(const char* decoder_name, wStream* s)
 			decoderAvailable = TRUE;
 	}
 
-	Stream_GetPointer(s, m);
+	pos = Stream_GetPosition(s);
 	if (decoderAvailable)
 		ret = tsmf_codec_parse_media_type(&mediatype, s);
-	Stream_SetPointer(s, m);
+	Stream_SetPosition(s, pos);
 
 	if (ret)
 	{
--- a/libfreerdp/codec/clear.c
+++ b/libfreerdp/codec/clear.c
@@ -1045,8 +1045,6 @@ INT32 clear_decompress(CLEAR_CONTEXT* clear, const BYTE* pSrcData, UINT32 SrcSiz
 	if (!s)
 		return -2005;
 
-	Stream_SetLength(s, SrcSize);
-
 	if (Stream_GetRemainingLength(s) < 2)
 	{
 		WLog_ERR(TAG, "stream short %" PRIuz " [2 expected]", Stream_GetRemainingLength(s));
--- a/libfreerdp/codec/dsp.c
+++ b/libfreerdp/codec/dsp.c
@@ -117,12 +117,6 @@ static INT16 read_int16(const BYTE* src)
 	return (INT16)(src[0] | (src[1] << 8));
 }
 
-static void write_int16(BYTE* dst, INT32 val)
-{
-	dst[1] = (val >> 8) & 0xFF;
-	dst[0] = val & 0xFF;
-}
-
 static BOOL freerdp_dsp_channel_mix(FREERDP_DSP_CONTEXT* context, const BYTE* src, size_t size,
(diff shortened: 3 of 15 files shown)


akallabeth:
@eyalitki So, I´ve modified the `wStream API` with more thorough checks and added return values to functions that modify the internal state.
Functions modifying the internal state that are not essentially necessary have been removed.
This way the streams can now no longer be in an inconsistent state and internal `WINPR_ASSERT` (which might abort and terminate the program) are able to (hard) enforce these limits

eyalitki:
Looks very good. I didn't think you'll go through the trouble of adding such checks per API function.

Just a question regarding the `WINPR_ASSERT`: Are the checks are done in runtime in the default/recommended compilation? Or they only used in development? We need these checks to detect bugs in the code, but mainly to block attacks in runtime.

akallabeth:
@eyalitki the `WINPR_ASSERT` is only done when compiled in.
But due to the newly introduced `Stream_Set*` checks the internal state of the internal stream can not be invalid. (e.g. `s->buffer`, `s->pointer` and `s->length` are in a valid within the stream, not necessarily correct for the application though and `s->capacity` is internal only, except when utilizing external buffers)

Only thing that can still happen is a missing `Stream_GetRemainingLength` or `Stream_GetRemainingCapacity` check but that is something that the developer has to take care of anyway.

akallabeth:
@eyalitki oh, and our nightly builds have that enabled, of course.

eyalitki:
The thing is that most of the vulnerabilities that I encountered thus far in FreeRDP were exactly in cases that the developer forgot / miscalculated the size of `Stream_GetRemainingLength`. And once a single such check was missing/wrong, all the rest of the checks are pointless because the return value of `size_t` will return a huge value instead of a negative one. While I respect the  responsibility of a developer to write secure / correct code, the current API is fragile, and year-after-year it proved itself as prone to vulnerabilities. This is the reason I submitted #7321, and I still stand behind my original pull request.

As long as the `WINPR_ASSERT` calls are only used during development, they will be very handy in catching bugs, but they won't suffice to block attacks against your users at runtime. And recent years proved there are numerous such vulns in your code, most of which are a result of the need for every developer to count every byte, and remember to update every check if some line was added later on in the code. Changing `size_t` to `ssize_t` will change the current state drastically, as it will allow later code checks to catch earlier errors, instead of doing nothing. 

If this is the case, I request you reconsider again your decision regarding #7321, as at least the 2nd line of defense there could be added as part of this pull request and it will be useful to block attacks (after they got triggered). I still think that the entire pull request of mine would greatly contribute to the project, especially if most of the checks you added are absent at runtime.

My pull request aimed to make FreeRDP more robust against human-errors. Counting on every developer, of every feature in the data-path, to always count all bytes correctly, isn't the best approach, and it didn't really prove itself as a secure methodology up until now.

akallabeth:
@eyalitki yes, the problems you faced were mostly due to broken input validation.
The situation there has been addressed in various cleanups.

But regarding your pr:

The case for `Stream_GetRemainingCapacity` is no longer possible, as `capacity` is handled internally. (The case with `Stream_New(buffer, capacity)` might still give a problem, but there is nothing we can do if a developer lies to us about the buffer size)

In case of `Stream_GetRemainingLength` this might happen, true. But that does not require a change in return, but simply returning nothing larger that there actually is so that parsing is aborted when a check is actually done. (this does not avoid missing checks but can at least tell `nothing left´ instead of `take whatever you want` when a check is done.

akallabeth:
@eyalitki I´ve updated the pr to address the issues you raised:

1. It is now impossible to have an invalid large return value at runtime from `Stream_GetRemainingCapacity` and `Stream_GetRemainingLength`
2. Whenever a violation of a (maybe not compiled in) `WINPR_ASSERT` for the length checks is found, a big fat warning is logged (with a complete stack trace)

this way it is:
1. Easy to spot (even at runtime) that there is something wrong
2. No invalid read might occur after a (later) check as the value returned is never larger than the actual data available.

eyalitki:
Looks good, it indeed resolves the issue with `WINPR_ASSERT` being present only during development. If we can pay the performance impact of these checks, the current commits are indeed a great step forward (security-wise).

As for #7321 I added the relevant comments over there.

akallabeth:
@eyalitki Hope the pr is now in a shape we can both agree on.
Found some (minor) issues in `guacamole` while testing, see https://github.com/apache/guacamole-server/pull/347

eyalitki:
Looks very good, thanks a lot for your fast cooperation on this :)

At the current state it indeed contains all features I wanted to add as part of #7321.

akallabeth:
@eyalitki you´re welcome, always trying to get this more reliable.
Did move the check function to internal linkage though, don´t want to expose it (there is always someone who will start using and depending on it :))

akallabeth:
@freerdp-bot test

mfleisz (approved):
LGTM

PR #7323, opened 2021-10-04
add missing filter type strings

alexandru-bagu:
* fixes crash that happens while debugging the proxy


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #7328, opened 2021-10-06
Fix pf_server_new: pass own copy of proxyConfig

akallabeth:
The lifecycle of proxyConfig passed to pf_server_new is unknown,
so pass proxyServer->config copy to modules.


mfleisz (approved):
LGTM

PR #7331, opened 2021-10-06
Add missing HOOK_TYPE enums

qarmin:
There is 13 values
https://github.com/FreeRDP/FreeRDP/blob/03a9ca80c848912adc0e85a049886203a1d5c236/server/proxy/proxy_modules.h#L44-L62
and some were not handled in 
https://github.com/FreeRDP/FreeRDP/blob/1f4ae26a6f31d2eafae8a80af396759deb58cb40/server/proxy/pf_modules.c#L67-L73
and could cause out of bounds read e.g. - "HOOK_TYPE_SERVER_SESSION_END" would cause this


akallabeth:
@qarmin thank you for the pull request.
I´ve taken the liberty to create my own pull to fix this issue with https://github.com/FreeRDP/FreeRDP/pull/7333
(the array access version looks too error prone, so I´ve replaced it with a switch)

PR #7334, opened 2021-10-07
Cleaned up RDPDR channel interhitance

akallabeth:
* All channels inheriting from RDPDR_DRIVE base struct
* Add functions to create/free a new device of a type
* Fixed settings array resize code, prevent 0 sizes


akallabeth:
@freerdp-bot test

PR #7345, opened 2021-10-11
add optional support for win32 console

alexandru-bagu:
* add option to build as console by using WITH_WIN_CONSOLE
* add passphrase read from stdin for Win32

Basically WITH_WIN_CONSOLE makes it so the application is built with the subsystem console instead of windows which makes it work correctly as a console application. the obvious issue with this is that if you use a shortcut to start the application with the necessary parameters to connect to a host you will also get a console window which is why WITH_WIN_CONSOLE is optional. 
The console can be closed automatically (with FreeConsole) but that will have to be another PR as it will require additional logic as to when FreeConsole should be called (maybe when a flag is included). This would make it so a console window open and then immediately close - a flicker. 
This PR is however essential for allowing FreeRDP to be used properly from command line on Windows, essentially to allow /from-stdin.

*WIP because freerdp_passphrase_read is not working properly, will make another commit the following day
**comments are welcome**

### DIFF
--- a/libfreerdp/utils/passphrase.c
+++ b/libfreerdp/utils/passphrase.c
@@ -28,12 +28,13 @@
 
 #include <stdio.h>
 #include <io.h>
+#include <conio.h>
 
 char read_chr(int isTty)
 {
+	char chr;
 	if (isTty)
 		return _getch();
-	char chr;
 	if (scanf_s("%c", &chr, (UINT32)sizeof(char)) && !feof(stdin))
 		return chr;
 	return 0;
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

alexandru-bagu:
Apologies for the many force pushes. I am now happy with this PR.
Since the initial commit I have added the following:
* allow redirected stdin to be used for stdin input
* flush stdout to help with automation
* use stdin instead of gui when /from-stdin is present
* add error log for stdin flag but stdin redirected

I don't quite understand why the freerdp-bot tests fail in Windows for this PR though, there's no build log.

akallabeth:
@freerdp-bot test

PR #7352, opened 2021-10-13
fix hashtable insert

alexandru-bagu:
for case when a new pair is being added but the newPair has a value of 0 and is being disposed by the dispose hook (specifically found in mappedGeometryUnref_void)

Issue begins at this line:
https://github.com/FreeRDP/FreeRDP/blob/834fdf0046d8a76a12a66937ff92293b5ac75d6d/winpr/libwinpr/utils/collections/HashTable.c#L344


akallabeth:
@alexandru-bagu which issue does this fix?

PR #7381, opened 2021-10-19
Wslg Support

spronovo:
Hey folks,

The purpose of this PR is to share back the changes we did to FreeRDP in the context of developing the support for Linux GUI applications in the Windows Subsystem for Linux (aka WSLg: https://github.com/microsoft/wslg).

First I wanted to say a big thank you to the FreeRDP community and for all the efforts that have been put into building it over the years. FreeRDP is working great for us and it would have been a challenge to write a new RDP server from scratch. FreeRDP has been solid in powering our support for Linux GUI applications and we hope we can continue contributing going forward.

We weren’t sure it would be best to split the changes into multiple PR or a single PR broken down into specific commit. We opted for the later for our initial PR to give an holistic view of the changes that were needed to support WSLg. Please let us know what you think and if you would prefer we submit this as several PR with a single commit each.

We’ve organized the commit in this PR for easy consumption. They are self-contained changes which could be taken independently or in different order. Each commit has details of the change and the reason behind it and we would be happy to provide additional context. The commit roughly do the following:

1)	Fix connection negotiation with mstsc/msrdc
2)	Add support for HyperV socket connection
3)	Fix protocol violation when uploading large ICON
4)	Update header with missing flag
5)	Added support for AUDIO_PLAYBACK_DVC
6)	Add support for graphics redirection protocol
7)	Fix protocol violation in Display Control Virtual Channel Extension

#3 in particular is something we would like to discuss further and get folks take on it. This is an interesting problem that occurs when doing large update orders, such a loading large icon (more details in the commit description). We made a simple/safe fix but we believe the better and more appropriate fix would be to change Stream_EnsureCapacity as not to change the sealed length of a stream being grown. But since that simpler alternative change would have a much broader impact, we opted to wait for a discussion before going into that direction.

For #6, we’re working on official documentation for the graphics redirection protocol that will be released in the near future. We had hope to have this ready before pushing changes upstream, but it is taking longer than anticipated. Because this protocol deals specifically with RDP being used over a VM boundaries (rather than over a network), we will post this documentation to the WSLg project sites once it is ready rather than the protocol documentation site to avoid confusion as the later is meant specifically for communication that are going over a network.

### DIFF
--- a/include/freerdp/channels/gfxredir.h
+++ b/include/freerdp/channels/gfxredir.h
@@ -82,7 +82,6 @@ struct _GFXREDIR_CAPS_HEADER
 	UINT32 signature;       // GFXREDIR_CAPS_SIGNATURE
 	UINT32 version;         // GFXREDIR_CAPS_VERSION
 	UINT32 length;          // GFXREDIR_CAPS_HEADER_SIZE + size of capsData
-	const BYTE capsData[0]; // GFXREDIR_CAPS_HEADER.length - GFXREDIR_CAPS_HEADER_SIZE
 };
 
 typedef struct _GFXREDIR_CAPS_HEADER GFXREDIR_CAPS_HEADER;
--- a/include/freerdp/update.h
+++ b/include/freerdp/update.h
@@ -254,7 +254,7 @@ struct rdp_update
 
 	wStream* us;
 	UINT16 numberOrders;
-	UINT16 offsetOrders; /* the offset to patch numberOrders in the stream */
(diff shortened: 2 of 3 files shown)


awakecoding:
@hardening @bmiklautz @akallabeth this pull request is of high importance, it is required to upstream the WSLg changes, the first contribution from Microsoft to FreeRDP!

akallabeth:
@freerdp-bot test

awakecoding:
@freerdp-bot test

PR #7384, opened 2021-10-20
Color profile check YUV2RGB matrix

akallabeth:
thanks to @alexandru-bagu for doing such intensive testing :)

### DIFF
--- a/libfreerdp/primitives/prim_YUV.c
+++ b/libfreerdp/primitives/prim_YUV.c
@@ -460,9 +460,9 @@ static pstatus_t general_YUV444ToRGB_8u_P3AC4R(const BYTE* const pSrc[3], const 
 	}
 }
 /**
- * | R |   ( | 256     0    403 | |    Y    | )
- * | G | = ( | 256   -48   -120 | | U - 128 | ) >> 8
- * | B |   ( | 256   475      0 | | V - 128 | )
+ * | R |   ( | 298     0    409 | | Y - 16  | )
+ * | G | = ( | 298   -100  -208 | | U - 128 | ) >> 8
+ * | B |   ( | 298    516    0  | | V - 128 | )
  */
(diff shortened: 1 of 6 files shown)


akallabeth:
@alexandru-bagu damn, looks like the different implementations need some more attention, the unit test fails (generic and sse produce differing results)

alexandru-bagu:
Unfortunately I don't get these failures on Windows and the best I can do is run the tests in Ubuntu docker, which I did and it seems that for *width = 80; *height = 72; (among other combinations) fails. Issue is I can't test properly because I don't have a proper dev environment for linux.

alexandru-bagu:
You are missing my last 3 commits. 
https://github.com/alexandru-bagu/FreeRDP/commits/color_profile_check

akallabeth:
@alexandru-bagu ok, strange, I've merged your pr. well, cherry-picked them now, lets see if that works out.

akallabeth:
@llyzs maybe you can have a look at this too?

llyzs:
@akallabeth I don't know what was the original issue that the PR is trying to solve, but we should not use different color matrix than specified in the MS spec. The problem could be elsewhere.

akallabeth:
@llyzs the issue is only visible with `[RDPEVOR]` video redirection.
When the h264 data is sent, then the colors do not match the background.
To test, play a video on youtube and wait until the video redirection kicks in.
When that happens, the white area surrounding the video will have a slightly different color from what it previously was. (the redirection does usually contain a few lines around the video as well)

alexandru-bagu:
Maybe it would be worth considering whether we can have the matrix change only for decoding when we know we are connected to a Windows terminal server implementation, since that's when the issue presents itself.

akallabeth:
@alexandru-bagu you don´t know that, there is no way to (properly) identify the remote os

PR #7387, opened 2021-10-20
Windows taskbar progress bar and some miscellaneous changes

alexandru-bagu:
* fixed build for -DWITH_CLIENT_INTERFACE=ON for windows
* added -DWITH_PROGRESS_BAR=ON (by default ON shows window minimised with marquee progress bar until video is ready, OFF results in no window (not even taskbar) until video is ready)
* removed the width/height limit as wide screens go past 4096


akallabeth:
@freerdp-bot test

Issue #7396, opened 2021-10-23
Compiler warnings building 2.4.1


floppym:
When building FreeRDP 2.4.1, GCC 11.2.0 outputs the following warnings, which may indicate some code that will fail at runtime under certain conditions.
```
/x/portage/net-misc/freerdp-2.4.1/work/freerdp-2.4.1/winpr/libwinpr/asn1/asn1.c:102:40: warning: array subscript 1 is outside array bounds of ‘struct ASN1encoding_s[1]’ [-Warray-bounds]
/x/portage/net-misc/freerdp-2.4.1/work/freerdp-2.4.1/winpr/libwinpr/asn1/asn1.c:97:40: warning: array subscript 1 is outside array bounds of ‘struct ASN1encoding_s[1]’ [-Warray-bounds]
/x/portage/net-misc/freerdp-2.4.1/work/freerdp-2.4.1/channels/rdpei/client/rdpei_main.c:1389:9: warning: ‘free’ called on pointer ‘rdpei’ with nonzero offset 112 [-Wfree-nonheap-object]
```

floppym:
In winpr/libwinpr/asn1/asn1.c, the code calls `calloc` to allocate memory for 1 `ASN1encoding_s` struct, but then goes on to assign a value to `encoder[1].magic` as if there are 2 `ASN1encoding_s` structs in an array. This makes no sense to me. This code is not actually called from anywhere within the FreeRDP project, but it is exposed in the winpr library.

In, channels/rdpei/client/rdpei_main.c,  the code calls `free(rdpei->contactPoints)`, but `rdppei->contactPoints` is an array within the struct, not a pointer generated via `malloc`.

floppym:
> In, channels/rdpei/client/rdpei_main.c, the code calls free(rdpei->contactPoints), but rdppei->contactPoints is an array within the struct, not a pointer generated via malloc.

It looks like this was fixed as part of this commit on master. Should this commit get backported to stable-2.0, or do we just want to remove the bad `free` call?

https://github.com/FreeRDP/FreeRDP/commit/268bc2e8ef7b76887d14d50702bb8446ce15a4c1#diff-f97fd0ac6b09c0b042fd5a53c2edaf36911dbb99c0c7387864e644d2cae0f3abL1396

floppym:
Resolved by 68eb6e4afce800cca6227ea8bee761c63f5045bd.

PR #7398, opened 2021-10-24
winpr: fix out-of-bounds access in ASN1_CreateEncoder

floppym:
Fixes: https://github.com/FreeRDP/FreeRDP/issues/7396


akallabeth:
@floppym thank you for the effort, but the whole thing is only a stub, so drop it until someone really implements that. (#7422 )

PR #7404, opened 2021-10-25
log failure to load dynamic channel without returning error

alexandru-bagu:
Instead of returning error when we don't manage to load a dynamic channel we should log it as an error and let it continue, which works fine.
I propose this change because if you start FreeRDP with a rdp file with "Video capture devices" enabled, freerdp will be stuck in reconnect loop (connect, get dynamic channel list, try to open `rdpecam`, fail to load channel, reconnect, etc).
Obviously one fix would be to disable the dynamic channel but I believe having an error instead will be just as effective, especially since new dynamic channels might continue to be added by Microsoft.


akallabeth:
I agree with @hardening 
Anything not supported should not be compiled in (e.g. use define guards for the channel)

alexandru-bagu:
Very well, I will submit another PR taking that into account.

PR #7412, opened 2021-10-28
Temporary trust certificate when helped by local cert store

alexandru-bagu:
Valid certificates are valid only for a specific amount of time and afterwards they must be changed. Any new certificate will have a different fingerprint and as such freerdp will reject the new certificate automatically even though it is (or may be) valid.
Since this trust (the one provided via local CA) is optional by default it should not interfere with manually accepted certificates nor should it require manual intervention to accept a new certificate.


akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #7416, opened 2021-10-30
fix outPackets counter

eronnen:


hardening:
Need to check that, but I feel like the patch will not fix packet counting, but will count sent chunks. And a chunk on top of the transport may contain multiple packets. I guess it depends what we expect `outPackets` to count.

akallabeth (changes requested):
This only works for the default transport and not breaks if another one is used (e.g. accelerated ssl hardware, ...) as they override the default transport.

eronnen:
@akallabeth So is there any good place to increment the packet counter? right now it's almost never updated

PR #7438, opened 2021-11-12
Fix compilation with mingw

hardening:
This patch addresses some issues when compiling against mingw. With these changes FreeRDP can be compiled out-of-the box on a Ubuntu 20.04.


winpr/include/winpr/smartcard.h
```diff
@@ -115,6 +115,14 @@
 
 #endif
 
+/* missing definition with mingw */
+#ifndef SCARD_E_PIN_CACHE_EXPIRED
+#define SCARD_E_PIN_CACHE_EXPIRED (HRESULT)(0x80100032L)
```
akallabeth, line 120: shouldn´t each of these be guarded by a `#ifndef`?

akallabeth (approved):
LGTM

winpr/include/winpr/smartcard.h
hardening: Well I think it comes as a full set, but I can split it for more safety.

winpr/include/winpr/spec.h
```diff
@@ -965,6 +965,7 @@ extern "C++"
 
 #endif
 
+#if !defined(__MINGW32__)
```
akallabeth, line 968: Does mingw already define these macros?
Would be no good idea to export all functions for mingw builds

hardening: yes they are defined in mingw headers.

PR #7443, opened 2021-11-16
Fixed issues with libusb device unref

akallabeth:
Fixed some unref issues and cleaned up warnings.


akallabeth:
@alexandru-bagu ok, found a few issues with usb redirection. (mostly the unreference of the device holding the current handle)

PR #7448, opened 2021-11-18
Load legacy provider when initializing OpenSSL 3.0 

ondrejholy:
With OpenSSL 3.0, FreeRDP log contains errors like: `4036740A4C7F0000:error:0308010C:digital envelope routines: inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:346: Global default library context, Algorithm (MD4 : 85), Properties () `. This leads to connection failures in some cases, e.g. https://github.com/FreeRDP/FreeRDP/issues/7449. This is because algorithms like MD4 are now part of the legacy provider, which is not loaded by default. Let's explicitly load that provider.

This MR also contains some other semi-relevant changes, please see the commit descriptions for more details.

### DIFF
--- a/winpr/libwinpr/crypto/cipher.c
+++ b/winpr/libwinpr/crypto/cipher.c
@@ -29,9 +29,6 @@
 #include <openssl/rc4.h>
 #include <openssl/des.h>
 #include <openssl/evp.h>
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):

ondrejholy:
It would be nice to backport this to stable as the other OpenSSL 3.0.0 fixes...

PR #7465, opened 2021-11-28
fixing android pointer not initialized when calling realloc

eronnen:
fixing bug where `update_pointer_new` calls `realloc` on uninitialized pointer

### DIFF
--- a/client/Android/android_freerdp.c
+++ b/client/Android/android_freerdp.c
@@ -292,6 +292,7 @@ static BOOL android_register_pointer(rdpGraphics* graphics)
 	if (!graphics)
 		return FALSE;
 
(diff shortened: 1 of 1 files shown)


client/Android/android_freerdp.c
```diff
@@ -292,6 +292,7 @@ static BOOL android_register_pointer(rdpGraphics* graphics)
 	if (!graphics)
 		return FALSE;
 
+	memset(&pointer, 0, sizeof(rdpPointer));
```
akallabeth, line 295: nice find, thank you for the pr.
We´d prefer the `rdpPointer pointer = { 0 };` syntax though, I´ll do that for you.

PR #7472, opened 2021-11-30
More warning fixes

akallabeth:


include/freerdp/channels/rdpdr.h
```diff
@@ -126,12 +126,14 @@ enum IRP_MN
 #define FILE_OVERWRITTEN 0x00000003
 
 /* DR_CORE_CLIENT_ANNOUNCE_RSP.VersionMinor */
+#define RDPDR_MAJOR_RDP_VERSION 1
```
mfleisz, line 129: Should make this an `enum` too like `RDPDR_MINOR_RDP_VERSION`

akallabeth: can do that as soon as we have a second value to check for.

PR #7479, opened 2021-12-03
Fixed freerdp_disconnect

akallabeth:
Added a missing call to freerdp_abort_connect to ensure all
threads terminate in an orderly fashion


mfleisz (approved):
LGTM

PR #7480, opened 2021-12-03
allow ffmpeg usage on Windows without pkgconfig

eronnen:
Allow compiling FreeRDP with Ffmpeg h264 subsystem without pkgconfig (like openh264)

### DIFF
--- a/cmake/FindFFmpeg.cmake
+++ b/cmake/FindFFmpeg.cmake
@@ -8,14 +8,15 @@
 set(REQUIRED_AVCODEC_VERSION 0.8)
 set(REQUIRED_AVCODEC_API_VERSION 53.25.0)
 
-include(FindPkgConfig)
-
-if (PKG_CONFIG_FOUND)
-	pkg_check_modules(AVCODEC libavcodec)
-	pkg_check_modules(AVUTIL libavutil)
-	pkg_check_modules(AVRESAMPLE libavresample)
-	pkg_check_modules(SWRESAMPLE libswresample)
-endif(PKG_CONFIG_FOUND)
+if (UNIX AND NOT ANDROID)
+  find_package(PkgConfig QUIET)
+  if (PKG_CONFIG_FOUND)
+    pkg_check_modules(AVCODEC libavcodec)
+    pkg_check_modules(AVUTIL libavutil)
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@eronnen could you clarify what this pr should do?
```
find_package(PkgConfig QUIET)
  if (PKG_CONFIG_FOUND)
...
endif()
```
already takes care of not using `pkg-config` if it is not found.

eronnen:
@akallabeth When I'm building on Windows or Android I have `pkg-config` installed, but specifically for `ffmpeg` it doesn't work, for example the binaries in https://github.com/BtbN/FFmpeg-Builds/releases don't contain `pkg-config` files, so I just copied the logic from https://github.com/FreeRDP/FreeRDP/blob/master/cmake/FindOpenH264.cmake#L10 which does work on Windows and Android.

akallabeth:
@eronnen then you can always set the paths manually.

eronnen:
@akallabeth why set the paths manually? it's not a trivial task on windows. Is there any necessity for the `pkg-config` files in `FindFFmpeg.cmake`? it's an equivalent usage to `FindOpenH264.cmake` which doesn't require `pkg-config`...

akallabeth:
@eronnen because you installed `pkg-config` and do not want to use it.

eronnen:
@akallabeth  I have `pkg-config` installed since it's required by other things, but unfortunately `FFmpeg` public builds on windows don't support it.

I changed the diff to 

```
if (UNIX AND NOT ANDROID)
  find_package(PkgConfig QUIET)
  pkg_check_modules(AVCODEC QUIET libavcodec)
  pkg_check_modules(AVUTIL QUIET libavutil)
  pkg_check_modules(AVRESAMPLE QUIET libavresample)
  pkg_check_modules(SWRESAMPLE QUIET libswresample)
endif (UNIX AND NOT ANDROID)
```

does that make more sense? (it's completely identical to https://github.com/FreeRDP/FreeRDP/blob/master/cmake/FindOpenH264.cmake#L10)

akallabeth:
actually no, because `openh264` works differently. (we use `pkg-config` for `ffmpeg` on windows too)

PR #7481, opened 2021-12-06
winpr: implement ncrypt API

hardening:
This implements the emulation for the ncrypt API and more specifically the smartcard provider so that we can list available keys and certificates using `pkcs11-helper`.


hardening:
If you know other paths for `opensc` modules on different distro and archs, I'd be happy to add them.

hardening:
note: mingw changes allows to compile tests under mingw

winpr/CMakeLists.txt
```diff
@@ -227,6 +227,17 @@ set(OPENSSL_FEATURE_TYPE "RECOMMENDED")
 		set(TESTING_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/Testing")
 endif()
 
+if (NOT WIN32)
```
akallabeth, line 230: shouldn´t this move to `WinPR` as this module is there and should be buildable without `FreeRDP`?

akallabeth (commented):
Overall LGTM,
will do a test run later

winpr/CMakeLists.txt
hardening: Not sure to understand your comment, I've put that in `winPR` precisely because I wanted FreeRDP to not care about `libpkcs11` and have winpr hiding this.

akallabeth: ok, strange, did show under `CMakeLists.txt` instead of `winpr/CMakeLists.txt` last time? anyway, here it is perfekt.

akallabeth (approved):
LGTM

PR #7483, opened 2021-12-10
Re-enable poll usage on Apple systems (revert PR7074)

thenextman:
Using `poll` on macOS was problematic for us at Devolutions and we raised [PR7074](https://github.com/FreeRDP/FreeRDP/pull/7054) accordingly.

Note that on our side, we originally changed this in our fork in [2018](https://github.com/Devolutions/FreeRDP/pull/5). PR7074 is an artifact of us catching up to upstream.

In our specific case, we saw that `poll` was somewhat unreliable on certain macOS versions and could cause high CPU usage. See [POLL ON MAC 10.12 IS BROKEN](https://daniel.haxx.se/blog/2016/10/11/poll-on-mac-10-12-is-broken/).

That said, the fallback to `select` is now causing other problems. We often see **kernel panics** when using RFX on macOS 12 (Monterey).

This can be fixed by reverting to `poll`. Since Apple's `poll` implementation was [fixed on at least one OS update](https://github.com/curl/curl/issues/1057#issuecomment-271866997), we think the most pragmatic step is to re-enable `poll` (rather than troubleshoot the interaction between Monterey and the `select`-based implementation in WinPRs sync functions).


hardening:
@freerdp-bot can you give a try for this PR

akallabeth:
@freerdp-bot come on, please do some test!

akallabeth:
@thenextman since this seems to toggle every few releases maybe add a `CMake` variable that allows the default to be overrideen?

thenextman:
@akallabeth I enabled `poll` by default; with the option to disable it with `-DWITH_POLL="OFF"`. Let me know what you think.

akallabeth (approved):
LGTM

PR #7490, opened 2021-12-14
Code cleanup

akallabeth:


winpr/include/winpr/collections.h
```diff
@@ -246,13 +246,13 @@ extern "C"
 	WINPR_API void* LinkedList_First(wLinkedList* list);
 	WINPR_API void* LinkedList_Last(wLinkedList* list);
 
-	WINPR_API BOOL LinkedList_Contains(wLinkedList* list, void* value);
+	WINPR_API BOOL LinkedList_Contains(wLinkedList* list, const void* value);
 	WINPR_API void LinkedList_Clear(wLinkedList* list);
 
-	WINPR_API BOOL LinkedList_AddFirst(wLinkedList* list, void* value);
-	WINPR_API BOOL LinkedList_AddLast(wLinkedList* list, void* value);
+	WINPR_API BOOL LinkedList_AddFirst(wLinkedList* list, const void* value);
```
hardening, line 252: Not sure I agree with adding `const` modifiers to add functions, because it doesn't make sense. I mean when we add something to the list it is meant to be retrieved and modified later, so it's not const at all. Although I'm ok for the other changes on `Contains` and `Remove`.

akallabeth: @hardening well, the issue here is that our lists can hold `const` and `mutable` elements in the list.
the `Add` functions do only add the pointer.
And adding it as `const` here only tells us that `LinkedList_AddFirst` does not change the value.

hardening: hum ok fair enough

PR #7495, opened 2021-12-16
Fix build with iOS audio redirection

thenextman:
Building with `-DWITH_IOSAUDIO="ON"` fails because:

> rdpsnd_ios.c:288:12: error: no member named 'SetFormat' in 'struct rdpsnd_device_plugin'
>        p->device.SetFormat = rdpsnd_ios_set_format;

Since `SetFormat` doesn't seem to be part of the rdpsnd subsystem device interface, this PR removes the function and fixes the build. It also corrects 3 warnings caused by incorrect method signatures.

Finally, "iOSToolchain.cmake" is corrected to "ios.toolchain.cmake" in the docs.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #7496, opened 2021-12-16
Fix freerdp_channels_process_message cleanup

akallabeth:
Not all code paths did clean up the message, specifically WMQ_QUIT

### DIFF
--- a/libfreerdp/core/client.c
+++ b/libfreerdp/core/client.c
@@ -23,6 +23,8 @@
 #include "config.h"
 #endif
 
+#include <winpr/assert.h>
+
 #include <freerdp/log.h>
 #include <freerdp/channels/drdynvc.h>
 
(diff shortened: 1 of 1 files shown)


libfreerdp/core/client.c
```diff
@@ -570,25 +572,27 @@ BOOL freerdp_channels_process_message_free(wMessage* message, DWORD type)
...
-		return FALSE;
-	}
+	BOOL rc = FALSE;
 
-	if (message->id == 0)
+	WINPR_ASSERT(instance);
+	WINPR_ASSERT(message);
+
+	if (message->id == WMQ_QUIT)
+		rc = FALSE;
```
mfleisz, line 581: This is confusing - should this actually be `goto fail;` (and return `FALSE`) or really `rc = FALSE;` which gets overriden to `TRUE` later on (and finally return `TRUE`)?

akallabeth: good catch ^^

PR #7505, opened 2021-12-20
Winpr remove unimplemented wnd

akallabeth:


hardening:
I'm rather ok with this patch except for the NDR part. 
I think our smartcard impl should be coded using NDR code instead of doing the stuff ourself, that generates the alignment problem we have with some calls. If I read correctly it's done this way under windows: IDL files compiled to code using NDR to marshal/unmarshal smartcard calls.

akallabeth:
@hardening as for `NDR` I think it is overkill for smartcard (the whole implementation that is, the only thing really used is the struct with pointer/buffer serialization) but we can add it at any time when it is implemented.
Having a broken stub just does not seem right.

hardening (approved):
Ok let's remove unused bits

PR #7530, opened 2022-01-09
rail: client: fix memory leak

kubistika:


akallabeth (approved):
LGTM
nice catch, really nasty one.

PR #7544, opened 2022-01-12
Reject invalid android build configuration

ccpp:
Without this patch the default build configuration will be picked silently and continuous integration does not detect any problem in this case:

```
./scripts/android-build-freerdp.sh --config android-build-release.conf  # forgot to give the full path of the config file
```


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #7545, opened 2022-01-12
Fix checks for input flags:

akallabeth:
The settings have double meaning:
* Setting the default in client preconnect callback can disable
  announcement of a flag to the server regardless of support
* If set it is overridden with the value the server supports

### DIFF
--- a/libfreerdp/core/capabilities.c
+++ b/libfreerdp/core/capabilities.c
@@ -1294,23 +1294,16 @@ static BOOL rdp_read_input_capability_set(wStream* s, rdpSettings* settings)
 			settings->FastPathInput = FALSE;
 		}
 
+		/* Note: These settings have split functionality:
+		 * 1. If disabled in client pre_connect, it can disable announcing the feature
+		 * 2. If enabled in client pre_connect, override it with the server announced support flag.
+		 */
 		if (settings->HasHorizontalWheel)
-		{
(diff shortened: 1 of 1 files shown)


libfreerdp/core/capabilities.c
```diff
@@ -1294,23 +1294,16 @@ static BOOL rdp_read_input_capability_set(wStream* s, rdpSettings* settings)
...
+		 * 1. If disabled in client pre_connect, it can disable announcing the feature
+		 * 2. If enabled in client pre_connect, override it with the server announced support flag.
+		 */
 		if (settings->HasHorizontalWheel)
-		{
-			if (inputFlags & TS_INPUT_FLAG_MOUSE_HWHEEL)
-				settings->HasHorizontalWheel = TRUE;
-		}
-
+			settings->HasHorizontalWheel = (inputFlags & TS_INPUT_FLAG_MOUSE_HWHEEL) ? TRUE : FALSE;
```
hardening, line 1302: `settings->HasHorizontalWheel = !!(inputFlags & TS_INPUT_FLAG_MOUSE_HWHEEL)` ?

akallabeth: what do you mean?
only set this to `TRUE` if it was already requested in `preconnect`, otherwise let it stay `FALSE` and send that back?

hardening: oh sorry I was not clear, I was just suggesting to have the boolean conversion done using `!!`

PR #7547, opened 2022-01-13
Stable 2.0 android ffmpeg

akallabeth:


akallabeth:
@ccpp I think that might be of interest for the F-Droid android build.
I´ve backported the FFMPEG build scripts and updated the configurations accordingly.

Since this did not make it into the last release, can you use this or should I create a separate tag?

ccpp:
> 



> 

@akallabeth if we want to have this already now on F-Droid, I think it would make sense to not only a new tag but before that have a new raw version string of "2.5.0-ffmpeg" with its own version-code of 27.

ccpp:
@akallabeth Is any further action required to build this, apart from calling the android-build-freerdp script? I could successfully compile FFMpeg but then FreeRDP CMake configuration fails with `FFMPEG support requested but not detected`

akallabeth:
@ccpp did some fixes/improvements with #7550 which still requires some care. Will backport that when ready.

PR #7561, opened 2022-01-17
Do not reset scard context cancel

akallabeth:


hardening (approved):
Fixes the slow disconnection for me

mfleisz (approved):

Issue #7560, opened 2022-01-17
Advanced Mouse Input channel


akallabeth:
During a discussion with @pnowack we´ve found something missing from the RDP core spec.
There is no (proper) way to send relative mouse input.

While the perfect approach would be to extend `[MS-RDPBCGR]` that might take too long to get accepted.

For the time being I´d propose a dynamic channel instead.
That would have the following properties:

1. Automatically loaded by `FreeRDP` based clients and used instead of mouse PDU if the server supports and opens it.
2. A module for windows RDP servers could be written reading the channel data and inserting it as mouse event using `SendInput`
3. `xrdp`, `ogon` and `gnome remote desktop` could implement it as well. Ogon might even be able to utilize the protocol parsing part of `FreeRDP` and only implement the actual backend mouse event injection

So, todo:
1. Implement a sample channel
2. Inform upstream of https://github.com/neutrinolabs/xrdp , https://github.com/ogon-project/ogon-project , https://gitlab.gnome.org/GNOME/gnome-remote-desktop , TODO: Who else? to also implement this

@pnowack @mfleisz @bmiklautz @hardening  what do you think, anything I missed?

hardening:
Just a hint: if you plan to implement the channel for windows (server side), do not use winPR for WTS API as it's totally buggy for dynamic channels.

akallabeth:
@hardening you can be sure we´re not trying to use that. Maybe it would even be better to implement that in `rust` (as MS is supporting that quite well and it is definitely nicer than the old C API)

pnowack:
My general approach here would be to still have a capability exchange, as that would allow us to do changes in the future, if needed.

What I currently have in mind is the following approach (for capability exchange):
- The server side initiates the channel (unlike the graphics pipeline, where the client side starts with CapsAdvertise)
- The server sends, like the client side in the graphics pipeline, capability sets. These ones can have a variable length, where the length is specified in the header. (Unsupported capability sets will be skipped by a peer (via its length)).
- The capability sets don't have their own PDU type, but a PDU subtype (Type is capability set, subtype is the version of the capability set).
Reason is that we don't pollute the PDU types themselves with capability sets, in case of we add other PDUs for functionality later (and with that further capability sets).
- PDUs are versioned (have a suffix of e.g. _v1 for version 1)

For the PDUs, there will be one RelativeMouseInput_v1 PDU:
- Argument 1: `dx`: relative mouse motion in the x component
[...]

PR #7562, opened 2022-01-17
Advanced input channel

akallabeth:
A first draft for implementing #7560 

How to test:
1. start `sfreerdp-server --cert=<path> --key=<path>` from `server/Sample` source directory (path required to find mouse image)
2. start `xfreerdp /v:host /u:foo /p:bar`
3. When connected type `i`, this will toggle the channel.
4. Move the mouse

When the channel is active, the mouse icon will not move, but the server will print warning log messages.
If the channel is disabled, the mouse icon will follow the mouse.

Protocol messages (after dynamic channel is established):
A `MSG_AINPUT_VERSION` from server to client to announce which channel version the server supports.
`MSG_AINPUT_MOUSE` messages from client to server when a mouse event happens.

### DIFF
--- a/channels/ainput/server/ainput_main.c
+++ b/channels/ainput/server/ainput_main.c
@@ -54,6 +54,14 @@ typedef struct ainput_server_
 
 } ainput_server;
 
+static BOOL ainput_server_is_open(ainput_server_context* context)
+{
+	ainput_server* ainput = (ainput_server*)context;
+
+	WINPR_ASSERT(ainput);
+	return ainput->thread != NULL;
+}
+
 /**
  * Function description
  *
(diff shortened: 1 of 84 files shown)


channels/ainput/server/ainput_main.c
```diff
@@ -0,0 +1,360 @@
...
+		WLog_ERR(TAG, "WTSQuerySessionInformationA failed!");
+		return ERROR_INTERNAL_ERROR;
+	}
+
+	ainput->SessionId = (DWORD)*pSessionId;
+	WTSFreeMemory(pSessionId);
+	hEvent = WTSVirtualChannelManagerGetEventHandle(ainput->context.vcm);
+	StartTick = GetTickCount();
+
+	while (ainput->ainput_channel == NULL)
```
hardening, line 84: Why is there that strange waiting loop ? Aren't we guaranteed to have the channel if it exists on the client side ?

akallabeth: did it similar to `echo`, but that is still todo

client/Wayland/wlf_disp.c
```diff
@@ -250,9 +250,9 @@ void wlf_disp_free(wlfDispContext* disp)
 
 	if (disp->wlc)
 	{
-		PubSub_UnsubscribeActivated(disp->wlc->context.pubSub, wlf_disp_OnActivated);
-		PubSub_UnsubscribeGraphicsReset(disp->wlc->context.pubSub, wlf_disp_OnGraphicsReset);
-		PubSub_UnsubscribeTimer(disp->wlc->context.pubSub, wlf_disp_OnTimer);
+		PubSub_UnsubscribeActivated(disp->wlc->common.context.pubSub, wlf_disp_OnActivated);
```
hardening, line 253: Perhaps a local variable aliasing `disp->wlc->common.pubSub` would make the code nicer...

client/Windows/wf_event.c
```diff
@@ -746,7 +745,7 @@ BOOL wf_scale_blt(wfContext* wfc, HDC hdc, int x, int y, int w, int h, HDC hdcSr
 	if (!wh)
 		wh = dh;
 
-	if (wfc->fullscreen || !wfc->context.settings->SmartSizing || (ww == dw && wh == dh))
+	if (wfc->fullscreen || !wfc->common.context.settings->SmartSizing || (ww == dw && wh == dh))
```
hardening, line 748: couldn't it be`if (wfc->fullscreen || !settings->SmartSizing || (ww == dw && wh == dh))` ?

akallabeth:
@pnowack ok, first working draft here.
Protocol is the simplest one possible now, just consisting of a server announced version and the mouse events sent back.
Should we want to introduce new messages we just increment the version and add the new message.

pnowack:
@akallabeth Can you create a version for the stable-2.0 branch? That would allow me to implement this for g-r-d.

Did a brief look at the changes:
For the client side: I don't think that just sending the mouse position diff changes is enough.
For something like a shooter the mouse pointer needs to be confined to the client window, as otherwise, when looking around, the mouse will fall out of the window and no input is submitted any more.  
So, when relative mouse input is used, the mouse pointer would need to be confined, like games usually do. This would imply that the mouse pointer is grabbed.

Another thing is with smart sizing: Do the x and y diff changes need to be scaled with smart sizing?

akallabeth:
@pnowack yes, did not do mouse grabbing related changes.
As for smart-sizing, these are already scaled.
A backport of the channel itself can easily be done, the whole changes for clients are a different matter. (but that should not stop from adoption in grd right?)

pnowack:
> As for smart-sizing, these are already scaled.

Okay

> (but that should not stop from adoption in grd right?)

It shouldn't, as FreeRDP3 can be installed in parallel to FreeRDP2 (not entirely parallel, since the uwac versions are the same in both versions, so I need to strip them out of one of the builds)

akallabeth:
@pnowack which unit do the values in https://gitlab.gnome.org/GNOME/mutter/-/blob/75f263ddb8beaaad0ea982ca0e9c37d81a42049b/data/dbus-interfaces/org.gnome.Mutter.RemoteDesktop.xml#L158 have?
relative pixel change to last position?

Also, do you know how to get relative changes from the `X11 API`? I have a small patch allowing to grab the mouse pointer in the `xfreerdp` window, but the events  only have the position within the window  or within the screen as arguments.

pnowack:
@akallabeth Forwarding the answer from Carlos (mutter dev) (forwarded the question):

`pnowack: you want XI_RawMotion to get relative motion info on X11 clients. Answering the other question, dx/dy is pixel based, e.g. passed as-is to XTestFakeRelativeMotionEvent`

PR #7564, opened 2022-01-19
smartcard_emulate: change the announced ATR

hardening:
This other ATR seems to be more recognized by new windows versions.


mfleisz (approved):
LGTM

PR #7565, opened 2022-01-19
Advanced Input Channel

akallabeth:
* Implements a custom FreeRDP channel to circumvent limitations of `[MS-RDPBCGR]` mouse events (allow sending relative position updates, more than 5 buttons, ...)
* Refactor clients to move generic channel code to `client/common`
* Automatically load the custom channel by all clients and use it, if the server supports it.
* Sample implementation in sample server. (Press `i` to toggle, just prints received events to `WLog`)

### DIFF
--- a/client/common/client.c
+++ b/client/common/client.c
@@ -1029,19 +1029,25 @@ BOOL freerdp_client_send_wheel_event(rdpClientContext* cctx, UINT16 mflags)
 		UINT rc;
 		UINT64 flags = 0;
 		INT32 x = 0, y = 0;
+		INT32 value = mflags & 0xFF;
+
+		if (mflags & PTR_FLAGS_WHEEL_NEGATIVE)
+			value = -1 * (0x100 - value);
+
 		if (mflags & PTR_FLAGS_WHEEL)
 		{
 			flags |= AINPUT_FLAGS_WHEEL;
-			x = flags & 0xff;
-			if (mflags & PTR_FLAGS_WHEEL_NEGATIVE)
-				x = -x;
+			y = value;
 		}
+
+		/* We have discrete steps, scale this so we can also support high
+		 * resolution wheels. */
+		value *= 0x10000;
+
 		if (mflags & PTR_FLAGS_HWHEEL)
 		{
 			flags |= AINPUT_FLAGS_WHEEL;
-			y = flags & 0xff;
-			if (mflags & PTR_FLAGS_WHEEL_NEGATIVE)
-				y = -y;
+			x = value;
 		}
 
 		WINPR_ASSERT(cctx->ainput->AInputSendInputEvent);
--- a/client/common/client.c
+++ b/client/common/client.c
@@ -1058,6 +1058,7 @@ BOOL freerdp_client_send_wheel_event(rdpClientContext* cctx, UINT16 mflags)
 #endif
 	if (!handled)
 		freerdp_input_send_mouse_event(cctx->context.input, mflags, 0, 0);
+
 	return TRUE;
 }
 
@@ -1106,13 +1107,20 @@ BOOL freerdp_client_send_button_event(rdpClientContext* cctx, BOOL relative, UIN
 
 	if (!handled)
 	{
-		const rdpSettings* settings = cctx->context.settings;
-		static INT64 lastX = 0;
-		static INT64 lastY = 0;
-
-		lastX = MIN(MAX(lastX + x, 0), settings->DesktopWidth);
-		lastY = MIN(MAX(lastY + y, 0), settings->DesktopHeight);
-		freerdp_input_send_mouse_event(cctx->context.input, mflags, (UINT16)lastX, (UINT16)lastY);
+		if (relative)
+		{
+			cctx->lastX += x;
+			cctx->lastY += y;
+			WLog_WARN(TAG, "Relative mouse input but channel % not available, sending absolute!",
+			          AINPUT_DVC_CHANNEL_NAME);
+		}
+		else
+		{
+			cctx->lastX = x;
+			cctx->lastY = y;
+		}
+		freerdp_input_send_mouse_event(cctx->context.input, mflags, (UINT16)cctx->lastX,
+		                               (UINT16)cctx->lastY);
 	}
 	return TRUE;
 }
@@ -1120,6 +1128,7 @@ BOOL freerdp_client_send_button_event(rdpClientContext* cctx, BOOL relative, UIN
 BOOL freerdp_client_send_extended_button_event(rdpClientContext* cctx, BOOL relative, UINT16 mflags,
                                                INT32 x, INT32 y)
 {
+	BOOL handled = FALSE;
 	WINPR_ASSERT(cctx);
 
 #if defined(CHANNEL_AINPUT_CLIENT)
@@ -1137,10 +1146,27 @@ BOOL freerdp_client_send_extended_button_event(rdpClientContext* cctx, BOOL rela
 			flags |= AINPUT_XFLAGS_BUTTON2;
 
 		ainput_send_diff_event(cctx, flags, x, y);
+		handled = TRUE;
 	}
-	else
 #endif
-		freerdp_input_send_extended_mouse_event(cctx->context.input, mflags, x, y);
+
+	if (!handled)
+	{
+		if (relative)
+		{
+			cctx->lastX += x;
+			cctx->lastY += y;
+			WLog_WARN(TAG, "Relative mouse input but channel % not available, sending absolute!",
+			          AINPUT_DVC_CHANNEL_NAME);
+		}
+		else
+		{
+			cctx->lastX = x;
+			cctx->lastY = y;
+		}
+		freerdp_input_send_extended_mouse_event(cctx->context.input, mflags, (UINT16)cctx->lastX,
+		                                        (UINT16)cctx->lastY);
+	}
 
 	return TRUE;
 }
--- a/client/common/client.c
+++ b/client/common/client.c
@@ -886,3 +886,20 @@ BOOL client_auto_reconnect_ex(freerdp* instance, BOOL (*window_events)(freerdp* 
 	WLog_ERR(TAG, "Maximum reconnect retries exceeded");
 	return FALSE;
 }
+
+int freerdp_client_common_stop(rdpContext* context)
+{
+	rdpClientContext* cctx = (rdpClientContext*)context;
+	WINPR_ASSERT(cctx);
+
+	freerdp_abort_connect(cctx->context.instance);
+
+	if (cctx->thread)
+	{
+		WaitForSingleObject(cctx->thread, INFINITE);
+		CloseHandle(cctx->thread);
+		cctx->thread = NULL;
+	}
+
+	return 0;
+}
--- a/client/common/client.c
+++ b/client/common/client.c
@@ -1016,3 +1016,125 @@ void freerdp_client_OnChannelDisconnectedEventHandler(void* context,
 	}
 #endif
 }
+
+BOOL freerdp_client_send_wheel_event(rdpClientContext* cctx, UINT16 mflags)
+{
+	BOOL handled = FALSE;
+
+	WINPR_ASSERT(cctx);
+
+#if defined(CHANNEL_AINPUT_CLIENT)
+	if (cctx->ainput)
+	{
+		UINT rc;
+		UINT64 flags = 0;
+		INT32 x = 0, y = 0;
+		if (mflags & PTR_FLAGS_WHEEL)
+		{
+			flags |= AINPUT_FLAGS_WHEEL;
+			x = flags & 0xff;
+			if (mflags & PTR_FLAGS_WHEEL_NEGATIVE)
+				x = -x;
+		}
+		if (mflags & PTR_FLAGS_HWHEEL)
+		{
+			flags |= AINPUT_FLAGS_WHEEL;
+			y = flags & 0xff;
+			if (mflags & PTR_FLAGS_WHEEL_NEGATIVE)
+				y = -y;
+		}
+
+		WINPR_ASSERT(cctx->ainput->AInputSendInputEvent);
+		rc = cctx->ainput->AInputSendInputEvent(cctx->ainput, flags, x, y);
+		if (rc == CHANNEL_RC_OK)
+			handled = TRUE;
+	}
+#endif
+	if (!handled)
+		freerdp_input_send_mouse_event(cctx->context.input, mflags, 0, 0);
+	return TRUE;
+}
+
+#if defined(CHANNEL_AINPUT_CLIENT)
+static INLINE BOOL ainput_send_diff_event(rdpClientContext* cctx, UINT64 flags, INT32 x, INT32 y)
+{
+	UINT rc;
+
+	WINPR_ASSERT(cctx);
+	WINPR_ASSERT(cctx->ainput);
+	WINPR_ASSERT(cctx->ainput->AInputSendInputEvent);
+
+	rc = cctx->ainput->AInputSendInputEvent(cctx->ainput, flags, x, y);
+
+	return rc == CHANNEL_RC_OK;
+}
+#endif
+
+BOOL freerdp_client_send_button_event(rdpClientContext* cctx, BOOL relative, UINT16 mflags, INT32 x,
+                                      INT32 y)
+{
+	BOOL handled = FALSE;
+
+	WINPR_ASSERT(cctx);
+
+#if defined(CHANNEL_AINPUT_CLIENT)
+	if (cctx->ainput)
+	{
+		UINT64 flags = 0;
+
+		if (relative)
+			flags |= AINPUT_FLAGS_REL;
+		if (mflags & PTR_FLAGS_DOWN)
+			flags |= AINPUT_FLAGS_DOWN;
+		if (mflags & PTR_FLAGS_BUTTON1)
+			flags |= AINPUT_FLAGS_BUTTON1;
+		if (mflags & PTR_FLAGS_BUTTON2)
+			flags |= AINPUT_FLAGS_BUTTON2;
+		if (mflags & PTR_FLAGS_BUTTON3)
+			flags |= AINPUT_FLAGS_BUTTON3;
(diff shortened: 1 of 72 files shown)


pnowack:
Argh, I hate github. I am not allowed to add review comments to codelines (anymore). Not sure why.
So I have to add comments this way:

Just a note: Wasn't able to (fully) implement the server side yet. Some things should be addressed for this:

1. The `AInputEventFlags` enum should be moved from ` include/freerdp/client/ainput.h` to ` include/freerdp/channels/ainput.h`, so that the server side can use these values too.
2. The `AINPUT_FLAGS_MOVE` enum value is currently unused. It should be used for absolute and relative pointer events.
3. The `AInputEventFlags` doesn't expose an enum value for `PTR_FLAGS_WHEEL_NEGATIVE` + Could we use for wheel events the same behaviour like BCGR defines it for the fastpath and slowpath (two's complement)?
This would preserve the behaviour regarding high resolution mouse wheels (and I can also copy-paste the code from the fastpath handling :P)
4. In ` client/common/client.c`:
```
if (mflags & PTR_FLAGS_BUTTON2)
    flags |= AINPUT_FLAGS_BUTTON3;
```
should be `if (mflags & PTR_FLAGS_BUTTON3)` in the condition

akallabeth:
>     1. The `AInputEventFlags` enum should be moved from ` include/freerdp/client/ainput.h` to ` include/freerdp/channels/ainput.h`, so that the server side can use these values too.
> 
Ok, will do.

>     2. The `AINPUT_FLAGS_MOVE` enum value is currently unused. It should be used for absolute and relative pointer events.
>
True.
 
>     3. The `AInputEventFlags` doesn't expose an enum value for `PTR_FLAGS_WHEEL_NEGATIVE` + Could we use for wheel events the same behaviour like BCGR defines it for the fastpath and slowpath (two's complement)?

This flag is not necessary. With this API we can just send the wheel values as signed integer as `x` or `y` (depending on which wheel triggered)

>        This would preserve the behaviour regarding high resolution mouse wheels (and I can also copy-paste the code from the fastpath handling :P)
> 
>     4. In ` client/common/client.c`:
> 
> 
> ```
> if (mflags & PTR_FLAGS_BUTTON2)
>     flags |= AINPUT_FLAGS_BUTTON3;
> ```
> 
good catch

akallabeth:
@pnowack yes, the value conversion issue will stay the same we already had, but we can at least work with a ready to use integer instead of masking out data from the flags. (the value range did not change, just the representation as integer instead of part of a bitmask)

pnowack:
> yes, the value conversion issue will stay the same we already had, but we can at least work with a ready to use integer instead of masking out data from the flags. (the value range did not change, just the representation as integer instead of part of a bitmask)

In such case, the conversion from using the two's complement needs to happen in `freerdp_client_send_wheel_event`, as the feeded flags use the two's complement, i.e. something like this would need to happen: https://gitlab.gnome.org/GNOME/gnome-remote-desktop/-/blob/9d138b47e9cc62bcf7a54863a0db9823560511ed/src/grd-session-rdp.c#L1305

As otherwise scrolling in the opposite direction would have a different absolute value.

pnowack:
@akallabeth Status report: I have a working branch with g-r-d: https://gitlab.gnome.org/pnowack/gnome-remote-desktop/-/commits/disp-advanced-input

The important part: Relative mouse input works! And that really well.

Some issues that I found (haven't investigated all of them yet):
- xfreerdp also forwards the input (at least with relative mouse input), even when it is not focused (xfreerdp can here even be on a different workspace).
- Haven't completed checked yet, but: It seems that input now only works with the advanced input channel. If the channel is not opened, no mouse input is forwarded.
- Wheel events don't work (yet). Not sure yet, where the problem is. Vertical scroll events don't work at all. Horizontal scroll events only very slow.
Not sure yet, where the issue is. My wheel handling looks currently like this: https://gitlab.gnome.org/pnowack/gnome-remote-desktop/-/commit/c3a869c7bf12319d2dae3ef53de076da0c1f22c3#9f219500e83b691da346f692aa73b8bb5ab03761_0_99

client/common/client.c
```diff
@@ -893,3 +916,232 @@ BOOL client_auto_reconnect_ex(freerdp* instance, BOOL (*window_events)(freerdp*
...
+			x = flags & 0xff;
+			if (mflags & PTR_FLAGS_WHEEL_NEGATIVE)
+				x = -x;
+		}
+		if (mflags & PTR_FLAGS_HWHEEL)
+		{
+			flags |= AINPUT_FLAGS_WHEEL;
+			y = flags & 0xff;
+			if (mflags & PTR_FLAGS_WHEEL_NEGATIVE)
+				y = -y;
```
pnowack, line 1051: Seems that I can now add comments to codelines. Anyway, for horizontal wheel events (`PTR_FLAGS_HWHEEL`) the x value should be `!= 0`, not the `y` value.
Similar situation for the vertical mouse wheel event.

pnowack (changes requested):

client/common/client.c
pnowack: Regarding the `x` and `y` value themselves: Can we scale them (i.e. multiply them with `120`). In my testing, I found out that for normal discrete steps `1` and `-1` are submitted.
In the case of high resolution mouse wheels lower and more scrolling values are submitted. Since `x` and `y` are not floating point numbers, but integers, this would make high resolution mouse wheel input impossible.

akallabeth: ok, fixed the `x` and `y` mixup and also scaled the input value (take discrete value * 0x10000 so we have enough space for more fine grained steps)

akallabeth:
> @akallabeth Status report: I have a working branch with g-r-d: https://gitlab.gnome.org/pnowack/gnome-remote-desktop/-/commits/disp-advanced-input
> 
> The important part: Relative mouse input works! And that really well.
> 
great! Fixed some fallback handling issues with the latest commits.

> Some issues that I found (haven't investigated all of them yet):
> 
>     * xfreerdp also forwards the input (at least with relative mouse input), even when it is not focused (xfreerdp can here even be on a different workspace).
>
Does this always happen or only if the mouse is still grabbed?
 
>     * Haven't completed checked yet, but: It seems that input now only works with the advanced input channel. If the channel is not opened, no mouse input is forwarded.
> 
Tested against a windows 10, no. The channel is only loaded if the server supports it, otherwise the default input method is used.
Of course, if the server opens the channel all input is sent over it.

> Edit: Another thing (just tiny nit): For the other channels, the client side variant and the server side variant also include `#include <freerdp/channels/CHANNEL_NAME.h>`, the ainput channel does not do it yet. So, I think that `#include <freerdp/channels/ainput.h>` should be included in `#include <freerdp/server/ainput.h>` too
> 
done, thx for the hint.

pnowack:
> Does this always happen or only if the mouse is still grabbed?

Not sure. When it happened, it only happened with `/mouse-relative`. I just tried to reproduce it again (no rebuild) and this time I was not able to reproduce the issue. My cmdline is: `xfreerdp3 /v:<ip> /network:auto /u:<user> /p:<pw> +async-input /size:1920x1080 /smart-sizing:1920x1080 /f /mouse-relative`

> Tested against a windows 10, no. The channel is only loaded if the server supports it, otherwise the default input method is used.
Of course, if the server opens the channel all input is sent over it.

It is the case though with gnome-remote-desktop. Just retested this. As soon as I revert the commits that open the channel, or all commits regarding the advanced input channel, no mouse input is possible any more with xfreerdp from master + this MR, while xfreerdp from stable-2.0 works.
Without reverting the advanced input channel commits, xfreerdp from master + this MR works fine again.
The RDP backend in weston might encounter the same situation, as its server backend is FreeRDP-based too (maybe @hardening can check that).

akallabeth:
@pnowack ok, that is strange. I´ve added the channel to `sfreerdp-server` (the sample server) and there I can toggle the channel and the client will send via channel or default fastpath input.

akallabeth:
@pnowack even weirder is that this might only happen, if `OnChannelConnected` is triggered with `FreeRDP::Advanced::Input`  although the server does not support the channel?

pnowack:
> >Does this always happen or only if the mouse is still grabbed?

> Not sure. When it happened, it only happened with /mouse-relative. I just tried to reproduce it again (no rebuild) and this time I was not able to reproduce the issue. My cmdline is: xfreerdp3 /v:<ip> /network:auto /u:<user> /p:<pw> +async-input /size:1920x1080 /smart-sizing:1920x1080 /f /mouse-relative

Ok, I know now why I could not reproduce the issue any more. While trying to reproduce the issue, I did not realize, that I hovered over wayland clients.
The issue obviously can only happen, when I hover XWayland client (or any client an X11 session).
Visual Studio Code is runs via XWayand, that is why I reproduced the issue with that client.

> even weirder is that this might only happen, if OnChannelConnected is triggered with FreeRDP::Advanced::Input although the server does not support the channel?

Apparently, yet. I actually ran into this issue today at 2am (before writing my first comment here), because at that time I did not have an implementation for the Advanced input channel in g-r-d, and input events did not work, except for xfreerdp from the stable-2.0 branch.
So, somehow this is triggered, despite not being opened.
In case of the issue origin is not found. We could a "ready" PDU to the channel, which the server emits, when it supports the channel. This will workaround the issue, but not solve it. But it is one way to circumvent, in case the cause cannot be found.

With the last push to this MR, the scrolling values are now different for horizontal and vertical scrolling:

```
0, -120
0, -120
0, -120
0, 120
0, 120
0, 120
7864320, 0
7864320, 0
7864320, 0
7864320, 0
-7864320, 0
-7864320, 0
-7864320, 0
```

akallabeth:
@pnowack ok, the server needs to open the channel or we will not have a `OnChannelConnected` event. If we do not have such an event we send input the old way.
If you have had such an issue this seems to be a problem with handling dynamic channels? I´d not want to add a workaround if we don´t have to ;)

As for the relative mouse stuff, added a window check so only events originating on the session window will update the mouse positions.

Mouse wheel, oops, conversion of raw value too late, fixed.

akallabeth:
@pnowack also tested against `freerdp-shadow`, also not trying to use `ainput` channel...

pnowack:
> If you have had such an issue this seems to be a problem with handling dynamic channels? I´d not want to add a workaround if we don´t have to ;)
> also tested against freerdp-shadow, also not trying to use ainput channel...

Just retested. It works now perfectly fine, even when I build g-r-d without advance input support. Don't know why this was not the case in the older versions.

> Mouse wheel, oops, conversion of raw value too late, fixed.

Yep, works fine now.

> As for the relative mouse stuff, added a window check so only events originating on the session window will update the mouse positions.

Yep, that issue is now also solved.

---

Other thing: Can you make a define for the `0x10000` value? In g-r-d we use `double` for the scroll wheel, so I can use the value there.

pnowack:
Actually, something is wrong with the last push: Despite using `/mouse-relative`, the mouse input is not relative any more, it now always sends absolute mouse motion

akallabeth:
@pnowack do you get log warn messages like `Relative mouse input but channel %s not available, sending absolute!`? Then you´re not using the channel.

pnowack:
> do you get log warn messages like Relative mouse input but channel %s not available, sending absolute!?

No, I don't. I see `[INFO][com.freerdp.channels.drdynvc.client] - Loading Dynamic Virtual Channel ainput` on the client side. And with a debug message, I see that the input events come from via the channel.
The mouse pointer is moved, but `AINPUT_FLAGS_REL` is not set any more.

My cmdline is (still): `xfreerdp3 /v:<ip> /network:auto /u:<user> /p:<pw> +async-input /size:1920x1080 /smart-sizing:1920x1080 /f /mouse-relative`

akallabeth:
@pnowack ok, got the issue. Now I can properly use the mouse in any mode (against `sfreerdp-server`):
1. Channel disabled -> use `[MS-RDPBCGR]` events
2. Channel enabled -> use it for all mouse events
3. Only use relative mouse move, if `+mouse-relative +grab-mouse` are set: `XI_RawMotion` et al only work on the root window, so we can not restrict the pointer to any window if the mouse is not grabbed.

akallabeth:
@pnowack additionally, added `WLOG_TRACE` messages to dump channel events (client and server side), grep for `com.freerdp.channels.ainput.client` or `com.freerdp.channels.ainput.server`

pnowack:
@akallabeth Checked the last push, the issue is unfortunately still there, no relative mouse events are emitted (only absolute mouse motion) (`/mouse-relative` is used):

```
[com.freerdp.channels.ainput.client] - [ainput_send_input_event] sending flags=AINPUT_FLAGS_MOVE [0x00000004], 1393x342
[com.freerdp.channels.drdynvc.client] - write_data: ChannelId=2 size=18
[com.freerdp.client.x11] - xf_unlock_x11_:	[0] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[1] from xf_process_x_events
[com.freerdp.client.x11] - xf_unlock_x11_:	[0] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[1] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[2] from xf_input_show_cursor
[com.freerdp.client.x11] - xf_unlock_x11_:	[1] from xf_input_show_cursor
[com.freerdp.client.x11] - xf_unlock_x11_:	[0] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[1] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[2] from xf_input_show_cursor
[com.freerdp.client.x11] - xf_unlock_x11_:	[1] from xf_input_show_cursor
[com.freerdp.channels.ainput.client] - [ainput_send_input_event] sending flags=AINPUT_FLAGS_MOVE [0x00000004], 1393x342
[com.freerdp.channels.drdynvc.client] - write_data: ChannelId=2 size=18
[com.freerdp.client.x11] - xf_unlock_x11_:	[0] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[1] from xf_process_x_events
[com.freerdp.client.x11] - xf_unlock_x11_:	[0] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[1] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[2] from xf_input_show_cursor
[com.freerdp.client.x11] - xf_unlock_x11_:	[1] from xf_input_show_cursor
[com.freerdp.client.x11] - xf_unlock_x11_:	[0] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[1] from xf_process_x_events
[com.freerdp.client.x11] - xf_lock_x11_:[2] from xf_input_show_cursor
[com.freerdp.client.x11] - xf_unlock_x11_:	[1] from xf_input_show_cursor
[com.freerdp.channels.ainput.client] - [ainput_send_input_event] sending flags=AINPUT_FLAGS_MOVE [0x00000004], 1392x342
[com.freerdp.channels.drdynvc.client] - write_data: ChannelId=2 size=18
```

I actually still have an older build of this branch, that I made before the fix of the mouse event capture, that happened, even when xfreerdp is not focused (or even on a different workspace), and relative mouse events work there (just with the bugs, that are fixed in the current version, but relative mouse motion work there fine).

In the old(er) build, sometimes both relative and absolute mouse events were emitted (things though worked fine on the server side) (the following are messages, where the server side, checks for the `AINPUT_FLAGS_REL` flag):

```
DEBUG: 10:52:59.007: relative_input: 1
DEBUG: 10:52:59.007: relative_input: 0
DEBUG: 10:52:59.007: relative_input: 1
DEBUG: 10:52:59.007: relative_input: 0
DEBUG: 10:52:59.007: relative_input: 1
DEBUG: 10:52:59.008: relative_input: 0
DEBUG: 10:52:59.009: relative_input: 1
DEBUG: 10:52:59.009: relative_input: 0
DEBUG: 10:52:59.010: relative_input: 1
DEBUG: 10:52:59.010: relative_input: 0
DEBUG: 10:52:59.011: relative_input: 1
DEBUG: 10:52:59.012: relative_input: 0
DEBUG: 10:52:59.012: relative_input: 1
DEBUG: 10:52:59.012: relative_input: 0
DEBUG: 10:52:59.013: relative_input: 1
DEBUG: 10:52:59.013: relative_input: 0
DEBUG: 10:52:59.014: relative_input: 1
DEBUG: 10:52:59.014: relative_input: 0
```

In the current version (since the first push, which fixed the event capture only capturing events, when xfreerdp is focused), `AINPUT_FLAGS_REL` isn't set any more.

I also tried the wayland variant this time (did disable it before, due to uwac version conflict), but also only absolute mouse motion events.

akallabeth:
@pnowack maybe I did not mention that prominently enough, with the latest push relative mouse only works, if mouse grabbing is also enabled. Try `+grab-mouse +mouse-relative`

akallabeth:
@pnowack as far as my tests have shown, `XI_RawEvent` are only emitted on the root window of a display, so I can´t properly limit that to only the `xfreerdp` window. Since this is too confusing of a behaviour, I´ve chosen to limit this to only be supported when the mouse is grabbed.

pnowack:
@akallabeth Still the same result. cmdline is now `xfreerdp3 /v:<ip> /network:auto /u:<user> /p:<pw> +async-input /size:1920x1080 /smart-sizing:1920x1080 /f /mouse-relative +grab-mouse +mouse-relative`, but no change.

Not sure if the behaviour is different with plain X11 sessions (I'm in a wayland session, xfreerdp runs via XWayland, but that does not stop mouse pointer grabbing).

akallabeth:
@pnowack I´ve just tested again (using debian bullseye, gnome-wayland):
1. Start `sfreerdp-server --port=3333` (check `test_icon.ppm`, `server.key` and `server.crt` are in the same folder)
2. Connect with `xfreerdp /v:localhost:3333 +mouse-relative +grab-mouse /p:`
3. Press `i` in `xfreerdp` to enable the `ainput` channel
4. See `sfreerdp-server` log with `[14:47:02:778] [22411:22724] [WARN][com.freerdp.server.sample.ainput] - sf_peer_ainput_mouse_event not implemented: 0x00000014, -2x3`

akallabeth:
@hardening added the option for external thread, anything else to address?

PR #7566, opened 2022-01-19
Added Advanced Input Channel

akallabeth:
Backport https://github.com/FreeRDP/FreeRDP/pull/7565


akallabeth:
@freerdp-bot test

hardening:
Perhaps the thread should be made optional like we've done for the sound channel, as possibly someone will want to integrate the ainput stuff in its own, external, event loop...

PR #7569, opened 2022-01-21
Cache Emulate_IsConfigured result

akallabeth:
The check is used very often, so this caching reduces necessary
allocations.
(tested with proxy emulated smartcard logon against windows server 2019:
* before ~180000 allocations with heaptrack
* after ~20000)


akallabeth:
@hardening this might explain some lag for `FreeRDP`, `winpr_HexLogDump` (and therefore `winpr_HexDump`) did always allocate/format a log message even if it was discarded.

mfleisz (approved):

PR #7571, opened 2022-01-22
cmdline: Always reinitialize NetworkAutoDetect

marex:
Always reinitialize NetworkAutoDetect, otherwise it might end set to
previous value and not the desired one. This happens when the client
sets /network:lan , in which case NetworkAutoDetect remains set to
TRUE because freerdp_set_connection_type() was previously called in
prepare_default_settings() and NetworkAutoDetect is never updated by
freerdp_set_connection_type() afterward.


hardening (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@hardening do yout want to create a backport pr?

PR #7582, opened 2022-01-25
Replace getlogin_r with getpwuid_r

Chingyat:
`getlogin_r()` doesn't work on my system.

And the manpage sais,

```
BUGS
       Unfortunately, it is often rather easy to fool getlogin().  Sometimes it does not work at all, because some program  messed  up
       the utmp file.  Often, it gives only the first 8 characters of the login name.  The user currently logged in on the controlling
       terminal of our program need not be the user who started it.  Avoid getlogin() for security-related purposes.

       Note that glibc does not follow the POSIX specification and uses stdin instead of /dev/tty.  A  bug.   (Other  recent  systems,
       like SunOS 5.8 and HP-UX 11.11 and FreeBSD 4.8 all return the login name also when stdin is redirected.)

       Nobody knows precisely what cuserid() does; avoid it in portable programs.  Or avoid it altogether: use getpwuid(geteuid()) in‐
       stead, if that is what you meant.  Do not use cuserid().
```

It is recommended to use `getpwuid` or environment variable `LOGNAME` to get the loggied in user name.

https://stackoverflow.com/questions/49936099/get-logged-in-user-in-ubuntu-16-04-lts-desktop?noredirect=1&lq=1
https://stackoverflow.com/questions/35068592/getlogin-r-and-getlogin-do-only-work-on-tty

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -49,6 +49,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <pwd.h>
 #endif
 
 #ifdef __MACOSX__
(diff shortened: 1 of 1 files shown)


akallabeth:
@Chingyat thank you for this addition. I´ve taken the hint and created #7583 which does some additional cleanups.

akallabeth:
@Chingyat thank you for the pull request. The issue seems valid, so I´ve taken the hint and created #7583 to fix this in the whole project (and not only for rdpdr)

PR #7595, opened 2022-01-27
Added pduLength check in rdpgfx_read_header function

zbstao:
if pduLength is zero will lead to an infinite loop in rdpgfx_recv_pdu function, suppose the rdp server returns illegal data.

### DIFF
--- a/channels/rdpgfx/rdpgfx_common.c
+++ b/channels/rdpgfx/rdpgfx_common.c
@@ -1,4 +1,5 @@
 /**
+﻿/**
  * FreeRDP: A Remote Desktop Protocol Implementation
  * Graphics Pipeline Extension
  *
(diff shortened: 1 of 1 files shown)


channels/rdpgfx/rdpgfx_common.c
```diff
@@ -119,6 +120,13 @@ UINT rdpgfx_read_header(wStream* s, RDPGFX_HEADER* header)
 	Stream_Read_UINT16(s, header->cmdId);     /* cmdId (2 bytes) */
 	Stream_Read_UINT16(s, header->flags);     /* flags (2 bytes) */
 	Stream_Read_UINT32(s, header->pduLength); /* pduLength (4 bytes) */
+
+    if (header->pduLength < 8)
```
akallabeth, line 124: Might be a good idea to check, if there is enough data too here?

akallabeth:
@zbstao while your assumption is correct (there should be a plausibility check there) how did you end up with an infinite loop?
While the length field is wrong, all further processing requires data to be present in the stream and each header read consumes 8 bytes, so an abort condition will be reached after all input data is consumed.
Additionally, each `header.cmdId` must be valid (and have enough data in the stream) to not abort with an error.

akallabeth:
@freerdp-bot test

akallabeth:
@zbstao so, thank you for this pull request, really appreciated!

zbstao:
@akallabeth thank you too, this is my first time contributing to the open source community.

PR #7604, opened 2022-01-29
Fix mediacodec resolution handling

eronnen:
* Fixes problem with specific codecs where configuring initial width and height is required
* Fixes the handling of scenarios where the codec changes the output width and height

### DIFF
--- a/libfreerdp/codec/h264_mediacodec.c
+++ b/libfreerdp/codec/h264_mediacodec.c
@@ -84,11 +84,12 @@ typedef media_status_t (*AMediaCodec_getName_t)(AMediaCodec*, char**);          
 typedef void (*AMediaCodec_releaseName_t)(AMediaCodec*, char*);                           // 28
 typedef AMediaFormat* (*AMediaCodec_getInputFormat_t)(AMediaCodec*);                      // 28
 
-const int COLOR_FormatYUV420Planar = 19;
-const int COLOR_FormatYUV420Flexible = 0x7f420888;
+static const int COLOR_FormatYUV420Planar = 19;
+static const int COLOR_FormatYUV420Flexible = 0x7f420888;
 
-const int MEDIACODEC_MINIMUM_WIDTH = 320;
-const int MEDIACODEC_MINIMUM_HEIGHT = 240;
+/* https://developer.android.com/reference/android/media/MediaCodec#qualityFloor */
+static const int MEDIACODEC_MINIMUM_WIDTH = 320;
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

libfreerdp/codec/h264_mediacodec.c
```diff
@@ -86,13 +87,18 @@ typedef AMediaFormat* (*AMediaCodec_getInputFormat_t)(AMediaCodec*);
 const int COLOR_FormatYUV420Planar = 19;
 const int COLOR_FormatYUV420Flexible = 0x7f420888;
 
+const int MEDIACODEC_MINIMUM_WIDTH = 320;
```
akallabeth, line 90: if not used outside this file, make these `static` (same for the existing ones above)

libfreerdp/codec/h264_mediacodec.c
```diff
@@ -86,13 +87,18 @@ typedef AMediaFormat* (*AMediaCodec_getInputFormat_t)(AMediaCodec*);
...
+const int MEDIACODEC_MINIMUM_HEIGHT = 240;
```
akallabeth, line 91: As for the `320x240` is there a specific reason for this resolution or is the requirement just `!= 0` ? If the latter, better use `16` or similar as `FreeRDP` is used embedded where displays might be really small (h264 alignment is 16 if I recall correctly?)

akallabeth (commented):
Overall LGTM, just not sure about the resolution (see comment)

akallabeth:
@freerdp-bot test

libfreerdp/codec/h264_mediacodec.c
eronnen: According to MediaCodec documentation smaller resolutions than 320*240 are not supported:

https://developer.android.com/reference/android/media/MediaCodec#qualityFloor

With Java it's possible to use `MediaCodecInfo.VideoCapabilities` class (https://developer.android.com/reference/android/media/MediaCodecInfo.VideoCapabilities) in order to get the actual minimum/maximum resolutions supported by the codec but unfortunately this API is not exported to the NDK yet

akallabeth: Ok, then we´ll just take this. Might be a good idea to reference the link with the constants.

akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #7606, opened 2022-01-31
Added stereo to supported backend formats for audin winmm

akallabeth:


akallabeth:
@vscfreire ok, this does fix it on my windows test machine too.
checking why it was added (there were some complaints for other devices)

akallabeth:
Ok, looks like I´ve found the real culprit for broken sound under windows.
The FFMPEG resampler path does something wrong, compiling with `-DWITH_DSP_FFMPEG=OFF` and the channel / samplerate encoding works.

akallabeth:
@vscfreire might be of interest for you.

PR #7615, opened 2022-02-01
uwac-window: Change protocol selection order and make ivi surface-id configurable

kadamkshitij:
**uwac-window: Change protocol selection order** 

The current order of protocol selection is xdg_base followed by
ivi_application followed by fullscreen_shell. As a result xdg_base
protocol will be selected even if ivi_application protocol is needed.

The order is now changed to ivi_application followed by fullscreen_shell
followed by xdg_base. It is now possible to select any one of the three
protocols by appropriately defining BUILD_IVI and BUILD_FULLSCREEN_SHELL
macro.

Signed-off-by: Kshitij Kadam <Kshitij.Kadam@ifm.com>

**uwac-window: Make ivi surface-id configurable** 

The ivi surface-id is made configurable by fetching it from the
environment variables. An environment variable IVI_SURFACE_ID needs to
be set to the required surface-id. In case it is not set then the code
will take 1 as the default surface-id.

Signed-off-by: Kshitij Kadam <Kshitij.Kadam@ifm.com>

### DIFF
--- a/uwac/libuwac/uwac-window.c
+++ b/uwac/libuwac/uwac-window.c
@@ -485,6 +485,22 @@ UwacWindow* UwacCreateWindowShm(UwacDisplay* display, uint32_t width, uint32_t h
 
 	wl_surface_set_user_data(w->surface, w);
 
+#if BUILD_IVI
+	if (display->ivi_application)
+	{
+		w->ivi_surface = ivi_application_surface_create(display->ivi_application, 1, w->surface);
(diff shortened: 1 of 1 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@kadamkshitij thank you for this fix!

PR #7623, opened 2022-02-03
proxy: improve channel treatment

hardening:
This PR introduces per channel context so that we can speed up operations like retrieving the channel name from its id, or knowing what shall be done for a packet (no config ACL recomputation at each packet).

### DIFF
--- a/server/proxy/pf_context.c
+++ b/server/proxy/pf_context.c
@@ -24,12 +24,58 @@
 #include <winpr/crypto.h>
 #include <winpr/print.h>
 
+#include <freerdp/server/proxy/proxy_log.h>
 #include <freerdp/server/proxy/proxy_server.h>
 
 #include "pf_client.h"
+#include "pf_utils.h"
 #include <freerdp/server/proxy/proxy_context.h>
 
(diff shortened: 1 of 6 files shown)


server/proxy/pf_context.c
```diff
@@ -24,13 +24,59 @@
...
 
+#define TAG PROXY_TAG("server")
+
+static UINT32 ChannelId_Hash(const void* key)
+{
+	const UINT32* v = (const UINT32*)key;
+	return *v;
+}
+
+static BOOL ChannelId_Compare(const UINT32* v1, const UINT32* v2)
```
akallabeth, line 44: Why not just use the pointer value? that would eliminate the requirement for custom compare and hash functions.

hardening: If I'm correct without the custom function it prevents from using `id==0` because the code in hashtable checks for null pointer for the key.

akallabeth: true. is `channelID == 0` allowed?

akallabeth (approved):

PR #7642, opened 2022-02-14
Fixed SSPI fallback to NTLM

akallabeth:
* Fix fallback if no kerberos credentials are found
* Move library detection to `WinPR`
* Add option to disable NTLM fallback
* Fix client NLA abort on unexpected message (no need to wait for activation timeout here)


akallabeth:
@freerdp-bot test

akallabeth:
@hardening agree with that. but I think that would be something bigger that should be addressed in another pr.

mfleisz (approved):
LGTM

PR #7644, opened 2022-02-15
Added runtime configuration option for kerberos/NTLM fallback

akallabeth:
* Adds `NTLM` and `kerberos` configuration options to registry
* Defaults to `WITH_GSS_NO_NTLM_FALLBACK=ON` now
* Fix and cleanup registry parser a bit


akallabeth:
@freerdp-bot test

PR #7647, opened 2022-02-16
Removed BUILTIN_CHANNELS

akallabeth:
Building client channels out of tree does not work reliable as the
dependencies can not be easily split. (exceptions may be simple ones
as echo channel)
It does complicate the build system and code though, so remove this
for maintainability reasons.


akallabeth:
@ondrejholy I think redhat/fedora uses `-DWITH_BUILTIN_CHANNELS=OFF` any problems with removing this and building all channels into the main library?

ondrejholy:
Yes, Fedora uses that, but I have no idea why that option was added, but perhaps @scaronni recalls it...

hardening (approved):

hardening:
Rebase needed but looks ok.

scaronni:
> @ondrejholy I think redhat/fedora uses `-DWITH_BUILTIN_CHANNELS=OFF` any problems with removing this and building all channels into the main library?

Can't remember to be honest, but after many years of maintaining the package there's a new maintainer. I will pull him in.

kathenas:
> > @ondrejholy I think redhat/fedora uses `-DWITH_BUILTIN_CHANNELS=OFF` any problems with removing this and building all channels into the main library?
> 
> Can't remember to be honest, but after many years of maintaining the package there's a new maintainer. I will pull him in.

I am the maintainer of remmina and not freerdp, but @scaronni emailed re this and I thak him for that. @ondrejholy seems to be most active with this package for fedora/red hat.

The commit that brought in -DWITH_BUILTIN_CHANNELS=OFF is:

https://src.fedoraproject.org/rpms/freerdp/c/c70363f61bcbd4e12da192cd7030984893f9c9de?branch=rawhide

7 years of way back when...

This change is for 3 and will not make it into the 2 stable tree I assume, so changes will only need to be made when 3 becomes stable.

akallabeth:
@kathenas ok, thank you. yes, it is for 3 (which we hopefully stabilize soon®)

PR #7658, opened 2022-02-22
server/disp: Sanitize optional physical monitor size values

pnowack:
```
The server sided handling of the display control channel currently
treats the physical monitor size values as obligatory values.
When the physical monitor size is invalid according to the
documentation, it is supposed to be ignored.
However, this is currently not done, leading to an abort of the display
control channel handling, when using Remmina, as Remmina does not submit
physical-monitor- widths and -heights.

Fix this issue by sanitizing the physical monitor size values, instead
of handling them as obligatory values, where both the physical with and
the physical height must be constrained to the values in the
documentation.
```

This is compile tested-only, since only Remmina is a reproducer of the issue and compiling Remmina is currently pain (apparently it does not use the system FreeRDP library).

### DIFF
--- a/channels/disp/server/disp_main.c
+++ b/channels/disp/server/disp_main.c
@@ -72,6 +72,24 @@ error:
 	return NULL;
 }
 
+static void disp_server_sanitize_monitor_layout(DISPLAY_CONTROL_MONITOR_LAYOUT* monitor)
+{
+	if (monitor->PhysicalWidth < DISPLAY_CONTROL_MIN_PHYSICAL_MONITOR_WIDTH ||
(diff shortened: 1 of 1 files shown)


pnowack:
I should probably sanitize the values, before the debug output

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

pnowack:
> This is compile tested-only, since only Remmina is a reproducer of the issue and compiling Remmina is currently pain (apparently it does not use the system FreeRDP library).

Tested now the changes too and it works as expected. (Did accidentally build FreeRDP with this MR on the client side and not on the server side. Remmina does use the system FreeRDP library)

PR #7667, opened 2022-02-23
Added quoted argument support to CommandLineParseCommaSeparatedValuesEx

akallabeth:
* Argument quoting support
* Empty list element detection
* Unit test for argument parser

Now allows a syntax of `/smartcard-logon:"a,b,c"` or `/smartcard-logon:a,b,"something special: , !!",c`


hardening (approved):
That LGTM, would be nice if the quotes could also be `'`

PR #7671, opened 2022-02-24
Android mediacodec

akallabeth:


hardening:
Just wondering: what happens if you try to load freerdp jar on a system that has an old ndk ? I guess we have link error and that we can't initialize FreeRDP or I'm missing something ?

akallabeth:
@hardening API 21 is the minimum we test with and that already has the library.

PR #7677, opened 2022-02-28
Openh264 fix graphical artifacts

akallabeth:
With OpenH264 2.0 `FlushFrame` was introduced. This is necessary to flush all buffers so that the YUV data is consistent after calling `DecodeFrame2`

See https://github.com/cisco/openh264/issues/3476


mfleisz (approved):
LGTM

PR #7679, opened 2022-02-28
Use _vsnwprintf instead of vswprintf_s

akallabeth:


hardening (approved):
LGTM, even if the code `wf_format_text` could be a little more obvious the 2 calls to `_vsnwprintf` unrolled

PR #7700, opened 2022-03-07
Fixed #7696: Abort freerdp_connect if manually canceled

### LINKED ISSUE #7696: Aborting connection in the middle of creating a connection doesn't work
When calling `freerdp_abort_connect` it sets the `abortEvent` event, but there is a possible flow in the `freerdp_connect` function where it calls `freerdp_reconnect` which resets the `abortEvent` event. 
This means that if I call `freerdp_abort_connect` before the `freerdp_reconnect` function is being called inside `freerdp_connect` - then the connection will not be aborted.

akallabeth:
If freerdp_abort_connect is called, set FREERDP_ERROR_CONNECT_CANCELLED
This way freerdp_reconnect can distinguish between network issues and
user interaction and abort a retry attempt.

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -37,6 +37,8 @@
 #include <freerdp/listener.h>
 #include <freerdp/cache/pointer.h>
 
+#include "utils.h"
+
 #define TAG FREERDP_TAG("core.connection")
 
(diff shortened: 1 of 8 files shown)


mfleisz (approved):
LGTM

PR #7701, opened 2022-03-07
Improve and document version detection

bmiklautz:


hardening:
Very nice documentation effort.

hardening (approved):

PR #7708, opened 2022-03-08
winpr/{libwinpr/thread/thread.c,include/winpr/thread.h}: Keep DumpThr…

sunweaver:
…eadHandles as a symbol even if WITH_DEBUG_THREADS is OFF.

Without this patch, one would have to introduce an SOVERSION change in cases where FreeRDP was provided with debugging enabled first and then disabled.

To avoid such awkwardnesses, keep the DumpTheadHandles symbols around even if debugging is disabled.


akallabeth:
@freerdp-bot test

bmiklautz (approved):

PR #7709, opened 2022-03-08
Smartcard tls logon fix

akallabeth:
* Skip authentication if we have TLS and smartcard logon or just ask for `Smartcard-Pin` if we did not enter a password
* Remove duplicates, e.g. `settings->Password` is the same as `settings->SmartcardPin` and having both just leads to confusion
* Removes `password-is-pin` as the setting is only of interest server side and in client side code dependent on `smartcard-logon` anyway


libfreerdp/core/nla.c
```diff
@@ -636,32 +636,21 @@ static BOOL nla_adjust_settings_from_smartcard(rdpNla* nla)
...
-			}
-
-			if (!instance->AuthenticateEx(instance, &settings->Username, &settings->SmartcardPin,
-			                              &settings->Domain, AUTH_SMARTCARD_PIN))
-			{
-				WLog_ERR(TAG, "no pin code entered");
-				return TRUE;
-			}
-		}
+		    auth_status reason = utils_authenticate(instance, AUTH_NLA, TRUE);
```
mfleisz, line 642: Shouldn't this be `AUTH_SMARTCARD_PIN`?

akallabeth: @hardening might shed some light on this. Is `user` and `domain` required for `smartcard-nla` then no, if not then yes.

hardening: That should definitely be `AUTH_SMARTCARD_PIN` as the pin is required to logon as the server do some checks once the NLA logon has passed using the smartcard channel.

client/common/client.c
```diff
@@ -424,20 +393,22 @@ static BOOL client_cli_authenticate_raw(freerdp* instance, rdp_auth_reason reaso
 {
 	static const size_t password_size = 512;
 	const char* auth[] = { "Username: ", "Domain:   ", "Password: " };
-	const char* authPin[] = { "Username: ", "Domain:   ", "Pin:      " };
+	const char* authPin[] = { "Username: ", "Domain:   ", "Smartcard-Pin:      " };
```
hardening, line 396: The spaces in the messages were arranged so that all prompts are vertically aligned, could you adjust that ?

libfreerdp/core/nla.c
akallabeth: @hardening right. I´ve checked, the `user` and `domain` field should already be prefilled from the certificate, right?

mfleisz (approved):

PR #7714, opened 2022-03-09
Workaround for [MS-RDPBCGR] 2.2.9.2.3 Frame Marker Command (TS_FRAME_…

akallabeth:
…MARKER)

Connections with  windows 2016 and 2019 sometimes receive short
frame marker. Ignore these to prevent disconnects

### DIFF
--- a/libfreerdp/core/surface.c
+++ b/libfreerdp/core/surface.c
@@ -171,14 +171,21 @@ static BOOL update_recv_surfcmd_frame_marker(rdpUpdate* update, wStream* s)
 
 	WINPR_ASSERT(s);
 
-	if (Stream_GetRemainingLength(s) < 6)
+	if (Stream_GetRemainingLength(s) < 2)
 	{
 		WLog_ERR(TAG, "got %" PRIuz ", expected %" PRIuz " bytes", Stream_GetRemainingLength(s), 6);
 		return FALSE;
 	}
 
 	Stream_Read_UINT16(s, marker.frameAction);
-	Stream_Read_UINT32(s, marker.frameId);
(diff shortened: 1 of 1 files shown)


akallabeth:
@Fantu please try this out and check, if the disconnections are gone. It should print a very specific warning whenever this issue happens.

Fantu:
@akallabeth big thanks, applied this to 2.6.1, reproduced one time:
`[12:38:27:588] [1153:1168] [WARN][com.freerdp.core.surface] - [SERVER-BUG]: got 0, expected 4 bytes. [MS-RDPBCGR] 2.2.9.2.3 Frame Marker Command (TS_FRAME_MARKER) is missing frameId, ignoring`
but didn't disconnect this time, so this seems solved
now I started to update some users thin client to check if there are different cases

Fantu:
I reproduced it for example connecting to:
- windows server 2019 standard build 17763.2565 (I wait users confirm that after update disconnect is not reproducible anymore) (on one customers we have 2 twins virtual server both with build 17763.256 one have disconnect issue one not, tried to find different settings windows side but don't found any) other 2 customers have disconnect issue with W2019 but I not had debug on them to be sure is this issue the cause, I'll update some thin clients and ask to users to tell me if the disconnect persist (and in that case I'll take a debug) and if not persist I'll have the confirm was this
- Windows 10 Pro build 19043.1526 (seems solved, don't disconnected anymore for now and I still testing it keep connecting from thin client)
- Windows 10 Pro build 19041.1415 (where I checked the log, saw the warning of problem reproduce but didn't disconnect)

sorry for my bad english

PR #7726, opened 2022-03-15
Refactored WinPR::NTLM

akallabeth:
* added assertions
* added log messages
* removed code duplication


akallabeth:
@pnowack I´ve started to refactor the `NTLM` code so that logging is more usable and the negotiation flags are respected. Could you give a try and check if something changes for you?

pnowack:
> Could you give a try and check if something changes for you?

I need a version for stable-2.0 for this.

akallabeth:
@pnowack well, backports will most likely happen once this works. Is it not possible to compile against `master` to test?
Or even better, if you could provide a full log (with the user credentials/sam file used), I can add a unit test to check this.

Issue #4660, opened 2018-05-17
Screen artefacts and disappearing dropdown menus bug are still present [rail plugin]


yoklmn:
Hello. I use actual nightly version 2.0.0-dev3 (9460f4292) on Ubuntu 18.04 on Gnome Flashback session. Bugs, described here are still present:
https://github.com/FreeRDP/FreeRDP/issues/3997
https://github.com/FreeRDP/FreeRDP/issues/3656

Dropdown menus show briefly and then disappear. only on 10-20 prompts menu displayed, then starts disapear again. Sometimes I see background apps or desktop. I shot the video where shown disapearing menus and on the end of video shown artifact in right bottom corner

[record-16.23.38[17.05.2018].zip](https://github.com/FreeRDP/FreeRDP/files/2013301/record-16.23.38.17.05.2018.zip)

I try run by there commands:
```
/opt/freerdp-nightly/bin/xfreerdp /u:username /p:password /app:"||myprogram" /v:myserver:3389
/opt/freerdp-nightly/bin/xfreerdp /u:username /p:password /app:"||myprogram" /v:myserver:3389 +decorations +clipboard +fonts +window-drag +aero +menu-anims /gdi:hw
/opt/freerdp-nightly/bin/xfreerdp /u:username /p:password /app:"||myprogram" /v:myserver:3389 +decorations +clipboard +fonts +window-drag +aero +menu-anims /gdi:sw
/opt/freerdp-nightly/bin/xfreerdp -u username -p password --app --plugin rail --data "||myprogram" -- myserver:3389

```

akallabeth:
@yoklmn does it work more reliable when adding `/rfx /gfx`?

yoklmn:
If I run with /rfx and /gfx, or only /rfx problem still present. But if run only with /gfx key it work finely, but not enough:

[record-00.09.34[18.05.2018].zip](https://github.com/FreeRDP/FreeRDP/files/2014763/record-00.09.34.18.05.2018.zip)

gnshkl1:
@yoklmn screen recording does not showing any graphics artifacts issue.

yoklmn:
@gnshkl1 artifact shows on 01:04 on the tape. Here is screenshot from that screen recording http://prntscr.com/jqlt6w

akallabeth:
@yoklmn Could you run with `/lob-level:trace` and add the log? Would be interesting which operation is failing there.

nornic:
[...]

PR #7729, opened 2022-03-16
Added patch from #4660 @mokyle

akallabeth:
do not deactivate rail windows, just keep them always on


hardening (approved):
Not the expert but LGTM

PR #7730, opened 2022-03-16
Fix buffer overflow in MRDPView

thenextman:
`NSData` is not a C string; if the pasteboard holds data for "public.utf8-plain-text" it's not null-terminated. It's definitely not appropriate to increment the size before calling `ClipboardSetData`; the subsequent `memcpy` will cause an out-of-bounds read.

Since we know the pasteboard contains a UTF8 string, it's clearer to just handle it that way.

Additionally, the result of `dataForType:` must be checked against `nil`, as per the [documentation](https://developer.apple.com/documentation/appkit/nspasteboard/1531810-datafortype?language=objc).

*Note* this does not resolve the issue where, when connected to a Windows 11 server, every _second_ call to `ClientFormatList` fails (server returns `CB_RESPONSE_FAIL`, even on identical inputs).

### DIFF
--- a/client/Mac/MRDPView.m
+++ b/client/Mac/MRDPView.m
@@ -660,12 +660,13 @@ DWORD fixKeyCode(DWORD keyCode, unichar keyChar, enum APPLE_KEYBOARD_TYPE type)
 
 - (void)onPasteboardTimerFired:(NSTimer *)timer
 {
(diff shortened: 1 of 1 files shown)


akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #7734, opened 2022-03-21
Ntlm verbose

akallabeth:
* Add some more checks for `NTLM` module
* Add more log messages to debug errors in `NTLM` state machine

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -29,6 +29,7 @@
 #include <string.h>
 
 #include <winpr/crt.h>
+#include <winpr/sysinfo.h>
 #include <winpr/assert.h>
 #include <winpr/stream.h>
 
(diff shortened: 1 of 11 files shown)


channels/rdpdr/client/rdpdr_main.c
```diff
@@ -1170,7 +1171,10 @@ static UINT rdpdr_send_client_name_request(rdpdrPlugin* rdpdr)
 	WINPR_ASSERT(rdpdr);
 
 	if (!rdpdr->computerName[0])
-		gethostname(rdpdr->computerName, sizeof(rdpdr->computerName) - 1);
+	{
+		DWORD size = sizeof(rdpdr->computerName) - 1;
+		GetComputerNameA(rdpdr->computerName, &size);
```
hardening, line 1176: Not your code (and not the scope of this PR), but I know that some companies working on proxies would be interested if this `GetComputerNameA` could be abstracted so that the returned name can be controlled (doesn't it make some problems with our proxy impl ?)

akallabeth: @hardening not if you just use `passthrough` mode for `rdpdr` (then this path is not used), also with `intercept` this function is not used (see `rdpdr_send_client_name_request` in `pf_channels_rdpdr.c`

winpr/libwinpr/sspi/NTLM/ntlm_compute.c
```diff
@@ -545,7 +653,11 @@ void ntlm_generate_random_session_key(NTLM_CONTEXT* context)
 
 void ntlm_generate_exported_session_key(NTLM_CONTEXT* context)
 {
-	CopyMemory(context->ExportedSessionKey, context->RandomSessionKey, 16);
+	WINPR_ASSERT(context);
+	WINPR_ASSERT(sizeof(context->ExportedSessionKey) == sizeof(context->RandomSessionKey));
```
hardening, line 657: I don't see any reason for this assert, this is the design `NTLM_CONTEXT`, we're not going to check at every call the design of this struct !

PR #7742, opened 2022-03-25
nla: Fix handling of NULL identity

mfleisz:
While the identity got correctly reset if no username was set, identityPtr was dangling and caused AcquireCredentialsHandle to fail.


hardening (approved):
LGTM good catch

akallabeth:
@freerdp-bot test

PR #7743, opened 2022-03-25
Nla server cleanup && server auth fix

akallabeth:
* Some refacroring cleaning up `NLA` server side code (encapsulate some logic blocks in functions)
* Remove duplicate `instance`, `settings` pointer in `NLA`, replace with `context` (`instance` might be `NULL` in server side code)
* Fix `nla_server_authenticate` return, now `shadow` can authenticate with NLA again (if a SAM file is provided)


mfleisz (approved):
LGTM

PR #7746, opened 2022-03-25
Shadow leak fixes

akallabeth:


akallabeth:
@pnowack this and #7750 should fix your `NTLM` issue.

PR #7752, opened 2022-03-28
Moved clipboard utils to core library, fixes #6760

### LINKED ISSUE #6760: Move cliprdr_parse_file_list() outside of freerdp-client3 library

akallabeth:
(cherry picked from commit 8d5509573aafa811e55d19557908761bd48aa6e4)


akallabeth:
@pnowack `freerdp-client` links in `freerdp`

pnowack:
> @pnowack `freerdp-client` links in `freerdp`

Okay

mfleisz (approved):

PR #7758, opened 2022-03-28
Fixed #7753: Step calculation in YUV decoder

### LINKED ISSUE #7753: AVC sometimes leaves artifact on master but not stable 2.0
image

reported by @pnowack

akallabeth:


libfreerdp/codec/yuv.c
```diff
@@ -290,48 +363,57 @@ static BOOL pool_decode(YUV_CONTEXT* context, PTP_WORK_CALLBACK cb, const BYTE*
...
+				y.top += TILE_SIZE;
+			}
+
+			r.left += TILE_SIZE;
 		}
 	}
 	rc = TRUE;
 fail:
-	free_objects(work_objects, params, nobjects);
+	free_objects(context->work_objects, waitCount);
```
mfleisz, line 407: Doesn't this result in a dangling `work_objects` pointer in the context after the first `pool_decode` resulting in freed memory access when `pool_decode` is called again (unless the yuv context has been reset meanwhile)?

akallabeth: no, it just cleans up the work objects internals, not the memory storing the struct

mfleisz (approved):
LGTM

PR #7760, opened 2022-03-29
freerdp/constants: Conform osMinorType values to documentation

pnowack:

```
The OSMINORTYPE_NATIVE_WAYLAND value does not officially exist in
[MS-RDPBCGR].
Its value also is the one for Windows RT.

Fix this situation, by introducing the os minor type
OSMINORTYPE_WINDOWS_RT with its correct value and by incrementing the
value for OSMINORTYPE_NATIVE_WAYLAND, as that value does not exist
officially.
```


include/freerdp/constants.h
```diff
@@ -58,6 +58,8 @@ enum RDP_CODEC_ID
 #define OSMINORTYPE_MACINTOSH 0x0006
 #define OSMINORTYPE_NATIVE_XSERVER 0x0007
 #define OSMINORTYPE_PSEUDO_XSERVER 0x0008
-#define OSMINORTYPE_NATIVE_WAYLAND 0x0009
+#define OSMINORTYPE_WINDOWS_RT 0x0009
+/* As of 2022-03-29 the following does not exist officially in [MS-RDPBCGR] */
+#define OSMINORTYPE_NATIVE_WAYLAND 0x000A
```
akallabeth, line 63: If we start redefining this, it may be a good idea to count down as the range up will (most likely) be successively defined

pnowack: Sure. Choose the value 0xFFFE (if that is okay) (left 0xFFFF as reserved here, in case this would become useful in the future)

akallabeth:
@freerdp-bot test

PR #7762, opened 2022-03-29
X11/cliprdr: Clear selection requests, when they become invalid

pnowack:
```
When a FormatDataRequest by xfreerdp is answered with the
CB_RESPONSE_FAIL flag, then the request was answered, but xfreerdp does
not mark it as answered, by leaving the request as pending.
This results in non-functional copy-paste actions, when copying
something from the server side, after the failed request.
A similar situation can happen, when a new FormatList is received,
while there is a pending FormatDataRequest.

Fix these situations by clearing the pending request, when a form of an
answer, either via CB_RESPONSE_FAIL flag or via new FormatList, is
received.

Fixes: https://github.com/FreeRDP/FreeRDP/issues/7757
```

### DIFF
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -1616,6 +1616,10 @@ static UINT xf_cliprdr_server_format_list(CliprdrClientContext* context,
 	xfc = clipboard->xfc;
 	WINPR_ASSERT(xfc);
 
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #7771, opened 2022-04-02
Fixed parsing of https_proxy environment variable

AdrianVollmer:
Hi, my C skills are a little rusty, so please bear with me.

For testing, consider these two sets of commands.

```
$ unset https_proxy
$ https_proxy=http://user:pass@127.0.0.1:1234
$ export https_proxy
$ xfreerdp /v:somehost /log-level:DEBUG|grep 'connecting to peer'
```

and

```
$ unset https_proxy
$ xfreerdp /v:somehost /log-level:DEBUG /proxy:http://user:pass@127.0.0.1:1234|grep 'connecting to peer'
```

You'd expect both to behave the same, but the first interprets `user:pass@hostname` as the hostname. This is clear from the [code](https://github.com/FreeRDP/FreeRDP/blob/edeea86dd1ebed7364b0a43d1ef6c6495148f1c7/libfreerdp/core/proxy.c#L295) but can also be seen by the error message, which is only seen in the second case:

```
[14:14:07:034] [1298730:1298731] [DEBUG][com.freerdp.core] - connecting to peer 127.0.0.1
```

Both cases where the proxy URI is parsed should share the same code, so I moved it to `settings.c`, because both affected files are already including it.

My remaining reasoning is in the commit message (copied below).

Please make double sure I didn't accidentally introduce any memory bugs...

## Commit Message

A proxy URI was parsed both in the function `proxy_read_environment()` in
`libfreerdp/core/proxy.c` and in the function
`freerdp_client_settings_parse_command_line_arguments()` in
`client/common/cmdline.c`. But only the latter was taking into account
that the URI can contain a user and password, which made it impossible
to use a proxy with authentication by just using an environment
variable like `https_proxy`.

The function `proxy_parse_uri()` in `libfreerdp/core/proxy.c` was
clearly not parsing the optional username and password.

This commit moves the code from `client/common/cmdline.c` to
`libfreerdp/common/settings.c` into a function `proxy_parse_uri()` so
redundant code is avoided. This function is then called in both
instances where a proxy URI is parsed.

Some minor adjustment were made to the code, so the `WLog_INFO()`
statement at the end works in both cases, where credentials were either
provided or not. Also, some error messages were added or clarified.

If the parsing of the environment variable fails, we ignore it entirely
(with a warning) and proceed.


libfreerdp/common/settings.c
```diff
@@ -1415,3 +1415,151 @@ BOOL freerdp_target_net_addresses_copy(rdpSettings* settings, char** addresses,
...
+
+BOOL proxy_parse_uri(rdpSettings* settings, const char* uri_in)
+{
+	BOOL rc = FALSE;
+	const char* protocol = "";
+	char* p;
+	char* atPtr;
+	char* uri_copy = _strdup(uri_in);
+	char* uri = uri_copy;
+	if (!uri)
```
akallabeth, line 1427: Memory leak (`uri_copy`)

libfreerdp/common/settings.c
```diff
@@ -1415,3 +1415,151 @@ BOOL freerdp_target_net_addresses_copy(rdpSettings* settings, char** addresses,
 
 	return TRUE;
 }
+
+BOOL proxy_parse_uri(rdpSettings* settings, const char* uri_in)
```
akallabeth, line 1419: Function should stay in `core/proxy.c`, it is not really something `settings` specific but a part of `proxy`

include/freerdp/settings.h
```diff
@@ -1697,6 +1697,7 @@ extern "C"
 	FREERDP_API void freerdp_target_net_addresses_free(rdpSettings* settings);
 	FREERDP_API BOOL freerdp_target_net_addresses_copy(rdpSettings* settings, char** addresses,
 	                                                   UINT32 count);
+	FREERDP_API BOOL proxy_parse_uri(rdpSettings* settings, const char* uri_in);
```
akallabeth, line 1700: Should be in `include/freerdp/utils/proxy_utils.h` or similar , not really settings

akallabeth (changes requested):
Overall LGTM.
Fix the memory leak (and the header exposing the new API function) and this is ready to merge.

AdrianVollmer:
Please take a look at the new commit; I'll squash after you approve.

Btw. the memory leak is already present in `master` as far as I can tell.

akallabeth:
@freerdp-bot test

libfreerdp/core/proxy.c
```diff
@@ -727,3 +797,23 @@ static BOOL socks_proxy_connect(BIO* bufferedBio, const char* proxyUsername,
 
 	return FALSE;
 }
+
+static BOOL value_to_int(const char* value, LONGLONG* result, LONGLONG min, LONGLONG max)
```
akallabeth, line 801: Must be declared before first use

AdrianVollmer:
Fixed. Also, I noticed that the default port used to be set to 80 in case of http and we don't want to break existing setups.

PR #7774, opened 2022-04-03
Switch to official OpenSSL Download location and away from github.

iiordanov:
One of the benefits is that this adds support for building with OpenSSL 3.0 versions without breaking support for building with 1.1.1.

As part of the work, the assumption that there is an /archive/ in the download path is removed.

Furthermore, cmake modules for finding OpenSSL are updated in order to support 3.0.

## This is how are pull requests handled by FreeRDP
1. Every new pull request needs to build and pass the unit tests at https://ci.freerdp.com
1. At least 1 (better two) people need to review and test a pull request and agree to accept

## Preparations before creating a pull
* Rebase your branch to current master, no merges allowed!
* Try to clean up your commit history, group changes to commits
* Check your formatting! A _clang-format_ script can be found at ```.clang-format```
  * The cmake target ```clangformat``` reformats the whole codebase
* Optional (but higly recommended)
  * Run a clang scanbuild before and after your changes to avoid introducing new bugs
  * Run your compiler at pedantic level to check for new warnings

## To ease accepting your contribution
* Give the pull request a proper name so people looking at it have an basic idea what it is for
* Add at least a brief description what it does (or should do :) and what it's good for
* Give instructions on how to test your changes
* Ideally add unit tests if adding new features

## What you should be prepared for
* fix issues found during the review phase
* Joining our chat to talk to other developers or help them test your pull might accelerate acceptance
  * Matrix room : #FreeRDP:matrix.org (main)
  * XMPP channel: #FreeRDP#matrix.org@matrix.org (bridged)
  * IRC channel : #freerdp @ irc.oftc.net (bridged)
* Joining our mailing list <freerdp-devel@lists.sourceforge.net> may be helpful too.


cmake/FindPackageMessage.cmake
```diff
@@ -0,0 +1,48 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
akallabeth, line 1: Why did you add this, it is part of your `CMake` installation....

cmake/FindPackageHandleStandardArgs.cmake
```diff
@@ -0,0 +1,605 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
akallabeth, line 1: Why did you add this, this is part of your `CMake` installation...

cmake/FindOpenSSL.cmake
```diff
@@ -1,101 +1,214 @@
...
-#
-# This software is distributed WITHOUT ANY WARRANTY; without even the
-# implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
-# See the License for more information.
-#=============================================================================
-# (To distribute this file outside of CMake, substitute the full
-#  License text for the above reference.)
-
-if (UNIX AND NOT ANDROID)
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
akallabeth, line 1: I´d just remove this file, it was previously necessary for older `CMake` installations without this file. (FreeRDP 2.0 still supports `CMake 2.8`)

cmake/FindPackageMessage.cmake
iiordanov: Good question, I got an error about it, so cmake didn't find it with the current way it is executed in the freerdp build script. Let me look into what is wrong.

cmake/FindPackageHandleStandardArgs.cmake
iiordanov: Same as above. Will do more testing and see what needs fixing.

cmake/FindPackageMessage.cmake
iiordanov: Fixed.

akallabeth:
@iiordanov please rebase to current master, your pull seems to be based on an older branch (b73ce8a9f6a630419b00397243207ce2fac6e16c from 2021-11-25) which has some conflicts with current master

akallabeth:
@iiordanov please also check the `OPENSSL_TAG` in all supplied build configurations, see `scripts/*.conf`

akallabeth (approved):

akallabeth (approved):
LGTM
Nice improvement, thank you!

iiordanov:
Any time, @akallabeth!

akallabeth:
@freerdp-bot test please

akallabeth:
@iiordanov ok, spoke too soon. looks like you need to adjust the `SCM_URL` of the other dependencies and revert the change in `common_update`

iiordanov:
I think it's good now. Will the tests relaunch by themselves?

akallabeth:
@freerdp-bot test
@iiordanov no, only freerdp members are able to talk to it (ok, no, that is not true, but we get a reply... sometimes)

Issue #7773, opened 2022-04-03
TestBacktrace fails with musl libc


floppym:
When built against musl libc on Gentoo Linux, TestBacktrace fails:
```
----------------------------------------------------------

75/188 Testing: TestBacktrace
75/188 Test: TestBacktrace
Command: "/var/tmp/portage/net-misc/freerdp-2.6.1/work/freerdp-2.6.1_build/Testing/TestWinPRUtils" "TestBacktrace"
Directory: /var/tmp/portage/net-misc/freerdp-2.6.1/work/freerdp-2.6.1_build/winpr/libwinpr/utils/test
"TestBacktrace" start time: Apr 03 05:12 UTC
Output:
----------------------------------------------------------
[05:12:52:399] [22101:22101] [FATAL][com.winpr.utils.debug] - Invalid stacktrace buffer! check if platform is supported!
winpr_backtrace failed!
<end of output>
Test time =   0.03 sec
----------------------------------------------------------
Test Failed.
"TestBacktrace" end time: Apr 03 05:12 UTC
"TestBacktrace" time elapsed: 00:00:00
----------------------------------------------------------
```
musl does not provide execinfo.h.

Downstream bug report: https://bugs.gentoo.org/836716

akallabeth:
pull requests welcome.

floppym:
> pull requests welcome.

Would it be acceptable to skip the test when there is no viable way to produce a backtrace?

floppym:
musl doesn't provide a `backtrace()` function. A possible alternative would be to depend on [libunwind](https://www.nongnu.org/libunwind/) and make use of its `unw_backtrace()` function.

akallabeth:
@floppym `libunwind` support is already there, but currently only for android.

akallabeth:
Ok, is now enabled by default.
As long as `libunwind` is linked in and `dladdr` is available the backtrace backend will work.

PR #7776, opened 2022-04-05
winpr: skip TestBacktrace when there is no viable backtrace implement…

floppym:
Fixes: https://github.com/FreeRDP/FreeRDP/issues/7773

### DIFF
--- a/winpr/libwinpr/utils/test/CMakeLists.txt
+++ b/winpr/libwinpr/utils/test/CMakeLists.txt
@@ -44,5 +44,7 @@ foreach(test ${${MODULE_PREFIX}_TESTS})
 	add_test(${TestName} ${TESTING_OUTPUT_DIRECTORY}/${MODULE_NAME} ${TestName})
 endforeach()
 
+set_property(TEST TestBacktrace PROPERTY SKIP_RETURN_CODE 77)
+
 set_property(TARGET ${MODULE_NAME} PROPERTY FOLDER "WinPR/Test")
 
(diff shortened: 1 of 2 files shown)


akallabeth (approved):
LGTM

akallabeth:
@floppym does `MUSL` provide some kind of stack trace API or none at all?
If it does not provide any, it may be the better option to implement the API and return success (although empty stack trace) so that  it will not fail during runtime. (guarded with a `MUSL` specific define if possible)

The intention of the test was to fail, if new platforms are added and the existing implementation does not support it. Default is to return error codes from the API (which is the case here)
If we are comfortable with the API not returning meaningful data, we can just add an `empty` implementation

floppym:
> @floppym does `MUSL` provide some kind of stack trace API or none at all?

As far as I can tell, it does not provide any such API. There are third party libraries that could fill in, but I'm not sure this is worth adding additional dependencies.

floppym:
> (guarded with a MUSL specific define if possible)

musl actually makes it difficult to detect itself via macros.

https://wiki.musl-libc.org/faq.html

akallabeth:
@floppym well, then it might be a good idea to instead add a build option `I_DONT_WANT_BACKTRACE_SUPPORT_AND_WILL_NOT_COMPLAIN` (slightly exaggerated name, please feel free to use something less verbose ;)) which does the same as a `MUSL` specific check (e.g. provide a `empty` backtrace implementation).
I just feel it is not right to take the edge off the test by default, just because something we assume is there is missing.
Having to use some additional build option to disable it seems like the more sensible approach, as there will always be new backends (or however you call the new implementation of dependencies)

akallabeth:
as for other backtrace `API` support, feel free to add any you like, nice to have a choice

floppym:
Yeah, adding a build option seems reasonable.

floppym:
As for adding an empty backtrace implementation: that pointless to me. The existing behavior is to simply return `NULL` from `winpr_backtrace`, and that seems like the most appropriate thing to do when we can't produce something useful.

PR #7788, opened 2022-04-07
winpr: fix cache assertion before it's been created

StefanKoell:
Cache assertion fails because the call below (gdi_init_ex) creates the cache. I just changed the order and put the assertion after gdi_init_ex has been called.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

hardening (approved):

PR #7792, opened 2022-04-09
cmake: add TEST_WINPR_BACKTRACE option

floppym:
This allows TestBacktrace to be disabled when building on platforms that
are known to have no working winpr_backtrace implementation.

Fixes: https://github.com/FreeRDP/FreeRDP/issues/7773


floppym:
Actually, I can just skip the test downstream.

PR #7794, opened 2022-04-11
Fix #7793: Do not expose internal input API

### LINKED ISSUE #7793: [API] rdp input scancodes mess
in current master df3737b085de8e907020faf7b9d6f0eee303f419  function freerdp_input_send_keyboard_event can't handle input code achieved from RDP_SCANCODE_LWIN, which returning 347 which failing  at WINPR_ASSERT(code <= UINT8_MAX); ( https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/core/input.c#L301 ), this change introduced in 5fb59a23a9f25f5996d7b015e20dcb5ce31208fb

before this assert i have succesfully used this macros (RDP_SCANCODE_LWIN) and it worked properly.

akallabeth:
Slow-Path input uses UINT16 for scancodes on wire, but only the
lower byte is actually used. (the extended fields are sent in
keyboardFlags field)
Hide this implementation detail and adjust the API to use UINT8
for the code instead just like the corresponding Fast-Path PDU


mfleisz (approved):
LGTM

PR #7807, opened 2022-04-19
fix restore a maximized window that is minimized back to the maximized state

2fly2:
fix restore a maximized window that is minimized back to the maximized state


akallabeth:
@2fly2 thank you for the pull request, but

* Could you please clean up the commit history so that there is  a single commit with a commit message representing what has been changed
* Add some description on why you changed the line (including how you tested the change)

2fly2:
@akallabeth OK I will try another pull request later.

PR #7811, opened 2022-04-20
[stable-2.0] Backport rail define + AUDIO_PLAYBACK_DVC

pnowack:
Split out of https://github.com/FreeRDP/FreeRDP/pull/7381

Won't break existing builds. `AUDIO_PLAYBACK_DVC` is only used, when `use_dynamic_virtual_channel` was explicitly set.  
I put the `use_dynamic_virtual_channel` attribute at the end of the public struct, like it was done with the huge file support boolean for the CLIPRDR channel.


akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #7812, opened 2022-04-20
Added FDK AAC support

akallabeth:


akallabeth:
@pnowack I´ve now implemented the core procedures to use FDK-AAC, so the encoder/decoder loop should work.
What is still missing is proper configuration of the encoder/decoder according to format requirements.
Do you have any idea where to get good documentation for that?

pnowack:
@akallabeth I found this: https://github.com/mstorsjo/fdk-aac/blob/master/documentation/aacEncoder.pdf, but only took a very brief look into the document.

akallabeth:
Ok, this seems blocked. I can not get `fdk-aac` to decode the streams sent by windows RDP servers.
While this works with `FFMPEG` and the internal `AAC` decoder it fails as soon as switching to `fdk-aac` (even when still using ffmpeg with the same configuration)

Same problem with direct decoding attempts.

akallabeth:
abandoned due to the encoder not being compatible (or me lacking the documentation) with the ms server implementation.

PR #7814, opened 2022-04-20
fix parntheses syntax error

eronnen:


akallabeth (approved):

akallabeth:
@freerdp-bot test
@eronnen thak you, looks like I missed that one.

PR #7815, opened 2022-04-22
fix x11 black RAIL window when connect to server 2019

2fly2:
problem: xfreerdp  connect server 2019 app, sometimes get a black window.
root cause:  the log print "Error Info SESSION_ID [LOGON_MSG_SESSION_CONTINUE]",  when the app window created later, the paint event is suppressed, so the window is black. but LOGON_MSG_SESSION_CONTINUE is normal state, we should not disable app mode at this time.


akallabeth (approved):

akallabeth:
@freerdp-bot test
@2fly2 nice find, thank you for the fix!

Issue #7449, opened 2021-11-18
Connection over gateway doesn't work with OpenSSL 3.0


ondrejholy:
**Describe the bug**
Connection over gateway doesn't work currently with up-to-date `master`/`stable-2.0` branches when build with OpenSSL 3.0. Individual connections to both machines work properly. It also works with older OpenSSL versions. Any idea what might be wrong?

**To Reproduce**
1. Build freerdp with OpenSSL 3.0, or use e.g. `xfreerdp` from Fedora rawhide:
1. `xfreerdp /v:IP /g:GATEWAYIP /u:USER /p:PASS /redirect-prefer:ip`
2. It fails with:
```
[11:41:14:369] [335533:335534] [ERROR][com.freerdp.core.gateway.http] - http_response_recv: Retries exceeded
[11:41:14:369] [335533:335534] [ERROR][com.freerdp.core.gateway.http] - 404629DD7E7F0000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:346:Global default library context, Algorithm (MD4 : 84), Properties ()
[11:41:14:369] [335533:335534] [ERROR][com.freerdp.core.gateway.http] - 404629DD7E7F0000:error:03000086:digital envelope routines:evp_md_init_internal:initialization error:crypto/evp/digest.c:234:
[11:41:14:369] [335533:335534] [ERROR][com.freerdp.core.gateway.http] - 404629DD7E7F0000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:346:Global default library context, Algorithm (MD4 : 84), Properties ()
[11:41:14:369] [335533:335534] [ERROR][com.freerdp.core.gateway.http] - 404629DD7E7F0000:error:03000086:digital envelope routines:evp_md_init_internal:initialization error:crypto/evp/digest.c:234:
[11:41:14:370] [335533:335534] [ERROR][com.freerdp.core.nego] - Protocol Security Negotiation Failure
[11:41:14:370] [335533:335534] [ERROR][com.freerdp.core] - rdp_client_connect:freerdp_set_last_error_ex ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED [0x0002000C]
[11:41:14:370] [335533:335534] [ERROR][com.freerdp.core.connection] - Error: protocol security negotiation or connection failure
```
[...]

PR #7822, opened 2022-04-24
Implement BIO_CTRL_GET_KTLS_SEND and BIO_CTRL_GET_KTLS_SEND

akarl10:
Openssl 3.0 requires to respond to this controls. According to there
documentation it should not need them, but in practice openssl's own source
is full of places where negative return values are not checked.

This is a fix for #7449

### DIFF
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -2407,7 +2407,23 @@ static long rdg_bio_ctrl(BIO* in_bio, int cmd, long arg1, void* arg2)
 		 */
 		status = BIO_ctrl(tlsOut->bio, cmd, arg1, arg2);
 	}
-
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@akarl10 thank you, was planning to look into this, you saved me a lot of debugging and reading into changes :)

akallabeth (approved):
LGTM
does indeed fix the gateway connection issue with current master and openssl3

akallabeth:
@akarl10 since this looks like a bug in openssl3, did you report upstream or is it already known?

bmiklautz:
@akarl10 thank you!

akarl10:
> @akarl10 since this looks like a bug in openssl3, did you report upstream or is it already known?

@akallabeth I just opened https://github.com/openssl/openssl/issues/18176
I did not find any bug in this direction, so maybe we were the first to hit this issue

akallabeth:
@akarl10 thank you.
Maybe we´re using it wrong, but it might still be a good thing to have some (better) docs for return values ^^

PR #7832, opened 2022-04-27
fix restore a maximized window when it is minimized

2fly2:
fix problem:
1. connect remote app, then show maximized, click taskbar icon, the window show minimized, now click taskbar icon again, the RAIL window will show restore in normal state, not maximized.
2. connect remote app, then show maximized, right click taskbar icon,select show restore, current will not send restore message to server.

os ubuntu 18.04.3 LTS

### DIFF
--- a/client/X11/xf_event.c
+++ b/client/X11/xf_event.c
@@ -839,7 +839,7 @@ static BOOL xf_event_MapNotify(xfContext* xfc, const XMapEvent* event, BOOL app)
 			 * Doing this here would inhibit the ability to restore a maximized window
 			 * that is minimized back to the maximized state
 			 */
-			xf_rail_send_client_system_command(xfc, appWindow->windowId, SC_RESTORE);
+			// xf_rail_send_client_system_command(xfc, appWindow->windowId, SC_RESTORE);
 			appWindow->is_mapped = TRUE;
 		}
 	}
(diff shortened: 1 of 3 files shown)


client/X11/xf_event.c
```diff
@@ -839,7 +839,7 @@ static BOOL xf_event_MapNotify(xfContext* xfc, const XMapEvent* event, BOOL app)
 			 * Doing this here would inhibit the ability to restore a maximized window
 			 * that is minimized back to the maximized state
 			 */
-			xf_rail_send_client_system_command(xfc, appWindow->windowId, SC_RESTORE);
+			// xf_rail_send_client_system_command(xfc, appWindow->windowId, SC_RESTORE);
```
akallabeth, line 842: Please add an explanation for this change

akallabeth (changes requested):
Overall LGTM, but the commented `SC_RESTORE` needs some explanation.

client/X11/xf_event.c
2fly2: it is redundancy, all window state change (maximized, minimized, restore) should be handled in xf_event_PropertyNotify. xf_event_MapNotify should stay the same style with xf_event_UnmapNotify in app mode, just update is_mapped member.

a window is mapped does not mean it should be shown restore or other state, so i think (appWindow->rail_state == WINDOW_SHOW)  is also redundancy.

akallabeth (approved):
LGTM
if a backport is desired, please notify me (or do a backport pr against stable-2.0 branch with `git cherry-pick -x` )

akallabeth:
@freerdp-bot test

PR #7844, opened 2022-04-27
Fixed cast-quality warnings

akallabeth:
Eliminate all `const-qual` warnings that we are responsible for (`OpenSSL` related ones stay)


akallabeth:
@freerdp-bot test

PR #7871, opened 2022-05-02
Fixed settings pointer functions

akallabeth:
* Bugfixes for offset gettern/setter
* Additional unit tests

### DIFF
--- a/libfreerdp/common/settings.c
+++ b/libfreerdp/common/settings.c
@@ -1242,7 +1242,7 @@ BOOL freerdp_settings_set_pointer_len(rdpSettings* settings, size_t id, const vo
 				freerdp_target_net_addresses_free(settings);
 			return freerdp_settings_set_pointer_len_(settings, FreeRDP_TargetNetAddresses,
 			                                         FreeRDP_TargetNetAddressCount, data, len,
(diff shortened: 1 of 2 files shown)


libfreerdp/common/settings.c
```diff
@@ -1333,62 +1333,144 @@ BOOL freerdp_settings_set_pointer_len(rdpSettings* settings, size_t id, const vo
...
+			if (offset >= max)
+				goto fail;
+			return settings->TargetNetPorts[offset];
+		case FreeRDP_ClientTimeZone:
+			max = 1;
+			if (offset >= max)
+				goto fail;
+			return settings->ClientTimeZone;
+		case FreeRDP_RdpServerCertificate:
+			max = 1;
```
mfleisz, line 1455: Shouldn't all the parameters that are checked against 1 (`FreeRDP_RdpServerCertificate`, `FreeRDP_RdpServerRsaKey`, and so on) return NULL and an invalid id log message since they are not arrays and should be queried using `freerdp_settings_get_pointer_writable`?

akallabeth: depends on how you view it, did allow these as `array[1]` types instead.

PR #7879, opened 2022-05-05
shadow: convert incoming key scancode to DWORD before |= KBDEXT

arekinath:
KBDEXT is 0x100, so if we |= it onto a UINT8 it actually does nothing, making us interpret all scancodes as if they're non-extended.


akallabeth (approved):
LGTM, nice catch!

akallabeth:
@freerdp-bot test

PR #7880, opened 2022-05-05
winpr: fix unitary test when CreateProcess is CreateProcessW

hardening:
As the second argument of CreateProcessW is modified to split with spaces and so this can not be a const arg.

### DIFF
--- a/winpr/include/winpr/crt.h
+++ b/winpr/include/winpr/crt.h
@@ -39,39 +39,39 @@
 
 #ifndef _strtoui64
 #define _strtoui64 strtoull
-#endif
+#endif /* _strtoui64 */
 
 #ifndef _strtoi64
 #define _strtoi64 strtoll
-#endif
+#endif /* _strtoi64 */
 
 #ifndef _rotl
 static INLINE UINT32 _rotl(UINT32 value, int shift)
 {
 	return (value << shift) | (value >> (32 - shift));
 }
-#endif
+#endif /* _rotl */
 
 #ifndef _rotl64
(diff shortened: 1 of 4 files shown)


winpr/include/winpr/crt.h
```diff
@@ -23,6 +23,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <malloc.h>
```
akallabeth, line 26: non standard header, should already be included with `stdlib.h`, breaks build on CI

akallabeth (changes requested):
Overall LGTM, but you need to get rid of the header

winpr/libwinpr/thread/test/TestThreadCreateProcess.c
```diff
@@ -15,7 +15,14 @@ int TestThreadCreateProcess(int argc, char* argv[])
 	BOOL status;
 	DWORD exitCode;
 	LPCTSTR lpApplicationName;
-	LPTSTR lpCommandLine;
+
+#ifdef _WIN32
+	TCHAR lpCommandLine[200] = _T("cmd /C set");
```
akallabeth, line 20: Sure this will copy the string to the buffer and not just set a pointer?

hardening: I think so, at least the unit test passes correctly.

hardening:
Changes done

akallabeth (approved):

PR #7883, opened 2022-05-05
transport: Set server socket as keepalive in transport_default_attach()

marex:
Set the same keepalive properties to accepted server socket as are set
to client sockets. This prevents the server from keeping sockets open
for tens of minutes in case the client disappears.

For details, see "Busy ESTAB socket is not forever" in
https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/


akallabeth:
@freerdp-bot test

libfreerdp/core/transport.c
```diff
@@ -160,8 +160,13 @@ static BOOL transport_default_attach(rdpTransport* transport, int sockfd)
...
 	BIO* socketBio = NULL;
 	BIO* bufferedBio;
+	const rdpSettings* settings;
+	rdpContext* context = transport_get_context(transport);
 
-	WINPR_ASSERT(transport);
+	settings = context->settings;
+	WINPR_ASSERT(settings);
+
+	freerdp_tcp_set_keep_alive_mode(settings, sockfd);
```
akallabeth, line 169: function returns a value, better check it (yes, currently always `TRUE` but that might change)

akallabeth (approved):
LGTM

akallabeth (approved):

akallabeth:
@freerdp-bot test
@marex thank you for the pull request!

PR #7885, opened 2022-05-07
fix moving a window shrinks it 14x7 when connect to server 2019

2fly2:
problem: connect to server 2019, every time we move the RAIL window, it shrinks.
root cause: Window positions sent using Client Window Move PDU SHOULD include hit-testable margins(Window Resize Margins).

reference doc: 
MS-RDPERP 1.3.4
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdperp/a7a1793b-090c-4599-b12d-ca7bba00281c

MS-RDPERP 2.2.2.7.4
https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-rdperp/495acc38-df83-42ec-8ed1-b8d58c047262

for lower version server,  server will not send resize margins, so all resize margins is 0, this fix work as before.

TODO:
this pull request only fix window that is resizable(mspaint), because server only send resize margins for resizable window(MS-RDPERP 3.2.5.1.6). other window(calc) still shrinks when move, we can caculate resize margins from window size change, because the only reason of size change is client window move pdu, i will open another pull request when test ok.

### DIFF
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -138,10 +138,10 @@ void xf_rail_adjust_position(xfContext* xfc, xfAppWindow* appWindow)
 		 * Calculate new size/position for the rail window(new values for
 		 * windowOffsetX/windowOffsetY/windowWidth/windowHeight) on the server
 		 */
-		windowMove.left = appWindow->x;
-		windowMove.top = appWindow->y;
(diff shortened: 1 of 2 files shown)


akallabeth (approved):
LGTM, nice addition!

akallabeth:
@freerdp-bot test

PR #7889, opened 2022-05-11
Mingw fix

hardening:


akallabeth (approved):
LGTM

winpr/include/winpr/crt.h
```diff
@@ -146,19 +148,7 @@ extern "C"
...
-	WINPR_API void* _aligned_offset_malloc(size_t size, size_t alignment, size_t offset);
-	WINPR_API void* _aligned_offset_realloc(void* memblock, size_t size, size_t alignment,
-	                                        size_t offset);
-	WINPR_API void* _aligned_offset_recalloc(void* memblock, size_t num, size_t size,
-	                                         size_t alignment, size_t offset);
-
-	WINPR_API size_t _aligned_msize(void* memblock, size_t alignment, size_t offset);
-
-	WINPR_API void _aligned_free(void* memblock);
+#ifndef _WIN32
```
akallabeth, line 151: This looks wrong, should be before the `#ifdef __cplusplus` above, shouldn´t it?

akallabeth (changes requested):
Ok, did miss that line before, but the `#ifndef _WIN32` thing closes the `extern "C"` block defined outside

winpr/include/winpr/crt.h
```diff
@@ -169,11 +159,69 @@ extern "C"
...
 
 #endif /* _WIN32 */
 
+#if !defined(_WIN32) || (defined(__MINGW32__) && ! defined(_UCRT))
+/* note: we use our own implementation of _aligned_XXX function when:
+ * - it's not win32
+ * - it's mingw with native libs (not ucrt64) because we didn't managed to have it working
+ *   and not have C runtime deadly mixes
+ */
+#if WINPR_MSVCR_ALIGNMENT_EMULATE
```
akallabeth, line 174: Leftovers? This does not make any sense in this branch as we do not use the `_aligned_malloc` et al in `FreeRDP` and `WinPR` any more.

akallabeth: If you want to have this so that we export the old symbols, I´d suggest that you add a `CMake option` for `WINPR_MSVCR_ALIGNMENT_EMULATE`

hardening: Well I've added these if someone still wants to use WinPR to compile win32 apps under linux

winpr/include/winpr/crt.h
hardening: good catch.

winpr/include/winpr/crt.h
akallabeth: ah, ok. but shouldn´t `#ifdef` or `#if defined()` be used instead? this way you need to do `#define WINPR_MSVCR_ALIGNMENT_EMULATE 1` instead of only `#define WINPR_MSVCR_ALIGNMENT_EMULATE`

hardening: oh yes good catch

akallabeth (approved):
Ok, pushed a fix for the define guards (and missing  definition of `WITH_PKCS11`)

PR #7893, opened 2022-05-12
Fix for rdpsnd mac memory leaks

lemonmojo:
The macOS implementation of the rdpsnd channel had some serious memory leaks which this PR fixes.
* In `rdpsnd_mac_play`, `AVAudioFormat` was only released when the creation of `AVAudioPCMBuffer` failed and the buffer was actually never(!) released.
* Autorelease Pools were missing in `rdpsnd_mac_open`, `rdpsnd_mac_close`, `rdpsnd_mac_set_volume`, `rdpsnd_mac_start` and `rdpsnd_mac_play` where ObjC methods were called which caused autoreleased objects to be created behind the scenes.

Just FYI: To check for missing autorelease pools, just run xfreerdp or any other client with the environment variable `OBJC_DEBUG_MISSING_POOLS` set to `YES`. On my machine, this yielded log output like this as soon as I started sound playback in the remote session:

```
objc[3094]: MISSING POOLS: (0x700003bbd000) Object 0x600001eacb70 of class __NSCFString autoreleased with no pool in place - just leaking - break on objc_autoreleaseNoPool() to debug
objc[3094]: MISSING POOLS: (0x700003bbd000) Object 0x6000010b5fc0 of class NSTaggedPointerStringCStringContainer autoreleased with no pool in place - just leaking - break on objc_autoreleaseNoPool() to debug
objc[3094]: MISSING POOLS: (0x700003bbd000) Object 0x600001eacab0 of class __NSCFString autoreleased with no pool in place - just leaking - break on objc_autoreleaseNoPool() to debug
objc[3094]: MISSING POOLS: (0x700003bbd000) Object 0x6000010ac920 of class __NSCFString autoreleased with no pool in place - just leaking - break on objc_autoreleaseNoPool() to debug
objc[3094]: MISSING POOLS: (0x700003bbd000) Object 0x6000010ac940 of class __NSCFString autoreleased with no pool in place - just leaking - break on objc_autoreleaseNoPool() to debug
objc[3094]: MISSING POOLS: (0x700003bbd000) Object 0x6000010acc80 of class NSTaggedPointerStringCStringContainer autoreleased with no pool in place - just leaking - break on objc_autoreleaseNoPool() to debug
objc[3094]: MISSING POOLS: (0x700003bbd000) Object 0x6000010acd80 of class NSTaggedPointerStringCStringContainer autoreleased with no pool in place - just leaking - break on objc_autoreleaseNoPool() to debug
[...]

PR #7895, opened 2022-05-12
Fix for audin mac memory leak

lemonmojo:
Like discussed in https://github.com/FreeRDP/FreeRDP/pull/7893#issuecomment-1124962245, here's a fix for a small memory leak in the audin mac channel.


lemonmojo:
Once merged, this should be backported too.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM
thank you for the fix! :beers:

PR #7901, opened 2022-05-13
gids: Fix broken check in GIDS emulation

mfleisz:


akallabeth (approved):
Ouch, that one is on me.
LGTM

PR #7903, opened 2022-05-16
fix RAIL abort when reconnect

2fly2:
bug: xfreerdp abort when reconnect
rootcause: when network break, RAIL windows will be free after channel disconnect, but during attempting reconnect, RAIL window event process may still running, and try to get RAIL window in hash table.


hardening:

akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #7905, opened 2022-05-16
Fix proxy rdpdr/smartcard handling

mfleisz:
This PR fixes two issues with proxy device/smartcard redirection:
- rdpdr client might receive a server capabilities request while in running state (i.e. seen this when server redirects).
- stream in smartcard redirection was checked for remaining length before its position was reset


hardening:
General remark: don't we have to do a global reset of all proxy channel states during a resize or a redirection from the backend ? And so that introduce the question to forward that reset to the front client (like in the resize case)...

mfleisz:
@hardening Yes - that's what I also noted ... as far as I can tell (at least regarding the rdpdr code) we only reset the proxy's client side channel, but the server backend currently won't even get notified in case of such an event.

hardening (approved):

PR #7913, opened 2022-05-22
utils: add implementation of POD(primitive object data) arrays

hardening:


include/freerdp/utils/pod_arrays.h
```diff
@@ -0,0 +1,146 @@
...
+	FREERDP_API const T* array_##TLOWER##_cdata(const Array##T* a); \
+	FREERDP_API T array_##TLOWER##_get(const Array##T* a, size_t idx); \
+	FREERDP_API void array_##TLOWER##_set(Array##T* a, size_t idx, T v); \
+	FREERDP_API BOOL array_##TLOWER##_append(Array##T* a, T v); \
+	FREERDP_API BOOL array_##TLOWER##_contains(const Array##T* a, T v); \
+	FREERDP_API BOOL array_##TLOWER##_foreach(const Array##T* a, Array##T##Cb cb, void* data); \
+	FREERDP_API void array_##TLOWER##_reset(Array##T* a); \
+	FREERDP_API void array_##TLOWER##_uninit(Array##T* a)
+
+#define POD_ARRAYS_IMPL(T, TLOWER) \
```
akallabeth, line 45: Better move to `.c` file as the implementation should not be public.

include/freerdp/utils/pod_arrays.h
```diff
@@ -0,0 +1,146 @@
...
+		WINPR_ASSERT(a); \
+		free(a->values); \
+		\
+		a->values = NULL; \
+		a->nvalues = 0; \
+	}
+
+
+POD_ARRAYS_DECL(UINT16, uint16);
+POD_ARRAYS_DECL(UINT32, uint32);
```
akallabeth, line 142: why only unsigned types, also why no `UINT64`?

include/freerdp/utils/pod_arrays.h
```diff
@@ -0,0 +1,146 @@
...
+
+#define POD_ARRAYS_DECL(T, TLOWER) \
+	typedef struct \
+	{ \
+		T* values; \
+		size_t nvalues; \
+	} Array##T; \
+	typedef BOOL Array##T##Cb(T v, void* data); \
+	\
+	FREERDP_API void array_##TLOWER##_init(Array##T* a); \
```
akallabeth, line 33: I´d change this to `Array##T array_##TLOWER##_new();` making it easier to declare and initialize the value at the same time

include/freerdp/utils/pod_arrays.h
```diff
@@ -0,0 +1,146 @@
...
+ * limitations under the License.
+ */
+#ifndef FREERDP_UTILS_POD_ARRAYS_H_
+#define FREERDP_UTILS_POD_ARRAYS_H_
+
+#include <winpr/wtypes.h>
+#include <freerdp/freerdp.h>
+
+#define POD_ARRAYS_DECL(T, TLOWER) \
+	typedef struct \
```
akallabeth, line 26: I´d hide the actual implementation of the struct from a public header to avoid people start using direct dereference

include/freerdp/utils/pod_arrays.h
```diff
@@ -0,0 +1,146 @@
...
+	FREERDP_API size_t array_##TLOWER##_size(const Array##T* a); \
+	FREERDP_API T* array_##TLOWER##_data(const Array##T* a); \
+	FREERDP_API const T* array_##TLOWER##_cdata(const Array##T* a); \
+	FREERDP_API T array_##TLOWER##_get(const Array##T* a, size_t idx); \
+	FREERDP_API void array_##TLOWER##_set(Array##T* a, size_t idx, T v); \
+	FREERDP_API BOOL array_##TLOWER##_append(Array##T* a, T v); \
+	FREERDP_API BOOL array_##TLOWER##_contains(const Array##T* a, T v); \
+	FREERDP_API BOOL array_##TLOWER##_foreach(const Array##T* a, Array##T##Cb cb, void* data); \
+	FREERDP_API void array_##TLOWER##_reset(Array##T* a); \
+	FREERDP_API void array_##TLOWER##_uninit(Array##T* a)
```
akallabeth, line 43: Better use the `_new` and `_free` thing here, we don´t only initialize, we also allocate

akallabeth (changes requested):
Overall a nice idea.

I´d prefer to hide implementation details as much as possible though. (we did weaken this with `wStream` but I still don´t really like this)
So maybe use a public `typedef struct` reserving enough space and use a `union` / internal `struct` to hide the actual fields.
Or use allocated (internal) memory instead, we already require `malloc` and `free`, so this should not really impact much.

include/freerdp/utils/pod_arrays.h
```diff
@@ -0,0 +1,146 @@
+/**
```
akallabeth, line 1: Shouldn´t this be part of `WinPR` just like all the other array/map/... types?

hardening: Yes could be, but as WinPR tends to mimic existing windows API...

include/freerdp/utils/pod_arrays.h
hardening: `_new` and `_free` sound like the ArrayUINT32 itself is allocated and I want to avoid that

include/freerdp/utils/pod_arrays.h
hardening: Setting it public allows to define your own arrays of basic objects (like a struct).

include/freerdp/utils/pod_arrays.h
hardening: Sure we can add as much as needed, but I'd rather add a new one as we make use of it.

include/freerdp/utils/pod_arrays.h
akallabeth: hmm, then it might be better to create this as header only type?
e.g. make all functions `static INLINE`

hardening: Could do that, but it will multiply the implementations in all C files that are using them.

akallabeth: well, as far as I understand it this is just a convenience wrapper for dynamic arrays, so it should be sufficient to have a generic (typed) template we can use, the implementation itself will not really change much.

You still get the type safety and unified (internal) consistency checks over separate implementations and `one place to change them all`

hardening:
Hum ok will have a look to have a mutualized implementation

hardening:
@akallabeth I did various tries with a generic implementation and hiding the internal details (using a public space holder to hide the details of the array impl). The result is useless complexity, that goes against the KISS paradigm with no obvious gains.

I'm totally for hiding the details of implementations when the component is rather complex and manipulate its own fields, but with POD arrays it's not the case: just 2 fields, obvious behaviour, setters and getters. I think we should take it as is, and check in further PR that people don't touch the fields directly.

If you think about it that's more or less what we do with `wStream`, and if you ask me, I think we should not make the internals of streams private. From today's situation we only have things to loose (as it's gonna prevent all possible compiler optimisations) and in practice we ask for changes when people try to manipulate stream fields directly (I'm not even sure that I've seen someone doing this).

akallabeth:
@hardening well, there are obvious gains (you can change the internals without breaking every use or even need to recompile every use of the module/class/struct however you call it), you can also change the internal layout without API/ABI breaks (if the instance is heap allocated), but you´re right for this lightweight wrapper it would be best to have it header only (with only `static INLINE` functions)
there is a lack of support in `C` that you can not hide such things from public headers, but that is something we have to live with.

hardening:
@akallabeth ok I'll switch to `static inline` functions then.

hardening:
Ok, I've done the changes to have `static inlines`

akallabeth (approved):

PR #7916, opened 2022-05-24
Updated RDPGFX_CAPSET_VERSION106 code to conform to spec

akallabeth:


hardening (approved):
LGTM

include/freerdp/channels/rdpgfx.h
```diff
@@ -96,11 +96,10 @@ typedef struct
...
 #define RDPGFX_CAPVERSION_105 0x000A0502 /** [MS-RDPEGFX] 2.2.3.8 */
-#define RDPGFX_CAPVERSION_106                                           \
-	0x000A0600 /** [MS-RDPEGFX] 2.2.3.9, [MS-RDPEGFX-errata] 2018-12-10 \
-	            */
+#define RDPGFX_CAPVERSION_106 0x000A0601 /** [MS-RDPEGFX] 2.2.3.9 */
+#define RDPGFX_CAPVERSION_107 0x000A0701 /** [MS-RDPEGFX] 2.2.3.10 */
 
-#define RDPGFX_NUMBER_CAPSETS 9
+#define RDPGFX_NUMBER_CAPSETS 10
 #define RDPGFX_CAPSET_BASE_SIZE 8
```
hardening, line 103: I was about to suggest to have this number computed, but I'm also wondering if that shall be in a public header...

akallabeth: Well, I don´t know a easy way to count defines.
Also, we need this in a public header as server and/or client implementations need this number

PR #7929, opened 2022-05-26
Fix ClearCodec context reset

awakecoding:
Fix ClearCodec context resetting, this issue was originally observed in Lync RDP traffic while doing offline pcap replaying.

### DIFF
--- a/libfreerdp/codec/clear.c
+++ b/libfreerdp/codec/clear.c
@@ -82,6 +82,39 @@ static const UINT32 CLEAR_LOG2_FLOOR[256] = {
 
 static const BYTE CLEAR_8BIT_MASKS[9] = { 0x00, 0x01, 0x03, 0x07, 0x0F, 0x1F, 0x3F, 0x7F, 0xFF };
 
+void clear_reset_vbar_storage(CLEAR_CONTEXT* clear, BOOL zero)
+{
+	int i;
+
+	if (zero) {
+		for (i = 0; i < ARRAYSIZE(clear->VBarStorage); i++)
+			free(clear->VBarStorage[i].pixels);
+
+		ZeroMemory(clear->VBarStorage, sizeof(clear->VBarStorage));
+	}
+
+	clear->VBarStorageCursor = 0;
+
+	if (zero) {
+		for (i = 0; i < ARRAYSIZE(clear->ShortVBarStorage); i++)
+			free(clear->ShortVBarStorage[i].pixels);
+
+		ZeroMemory(clear->ShortVBarStorage, sizeof(clear->ShortVBarStorage));
+	}
+
+	clear->ShortVBarStorageCursor = 0;
+}
+
+void clear_reset_glyph_cache(CLEAR_CONTEXT* clear)
+{
+	int i;
+
+	for (i = 0; i < ARRAYSIZE(clear->GlyphCache); i++)
+		free(clear->GlyphCache[i].pixels);
+
+	ZeroMemory(clear->GlyphCache, sizeof(clear->GlyphCache));
+}
+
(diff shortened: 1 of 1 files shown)


libfreerdp/codec/clear.c
```diff
@@ -82,6 +82,35 @@ static const UINT32 CLEAR_LOG2_FLOOR[256] = {
 
 static const BYTE CLEAR_8BIT_MASKS[9] = { 0x00, 0x01, 0x03, 0x07, 0x0F, 0x1F, 0x3F, 0x7F, 0xFF };
 
+void clear_reset_vbar_storage(CLEAR_CONTEXT* clear)
+{
+	int i;
+
+	for (i = 0; i < 32768; i++)
```
akallabeth, line 89: `ARRAYSIZE(clear->VBarStorage)` might be better here instead of the magic number

libfreerdp/codec/clear.c
```diff
@@ -82,6 +82,35 @@ static const UINT32 CLEAR_LOG2_FLOOR[256] = {
...
+		free(clear->VBarStorage[i].pixels);
+
+	ZeroMemory(clear->VBarStorage, sizeof(clear->VBarStorage));
+
+	clear->VBarStorageCursor = 0;
+
+	for (i = 0; i < 16384; i++)
```
akallabeth, line 96: `ARRAYSIZE()` might be better here instead of a magic number

libfreerdp/codec/clear.c
```diff
@@ -82,6 +82,35 @@ static const UINT32 CLEAR_LOG2_FLOOR[256] = {
...
+	ZeroMemory(clear->ShortVBarStorage, sizeof(clear->ShortVBarStorage));
+
+	clear->ShortVBarStorageCursor = 0;
+}
+
+void clear_reset_glyph_cache(CLEAR_CONTEXT* clear)
+{
+	int i;
+
+	for (i = 0; i < 4000; i++)
```
akallabeth, line 108: `ARRAYSIZE()` might be better here instead of a magic number

akallabeth (commented):
Overall LGTM

akallabeth (commented):
Overall LGTM,
just need to check if there are regressions with this (clearcodec does have quite a few rough edges)

libfreerdp/codec/clear.c
```diff
@@ -1007,8 +1036,7 @@ INT32 clear_decompress(CLEAR_CONTEXT* clear, const BYTE* pSrcData, UINT32 SrcSiz
 
 	if (glyphFlags & CLEARCODEC_FLAG_CACHE_RESET)
 	{
-		clear->VBarStorageCursor = 0;
-		clear->ShortVBarStorageCursor = 0;
+		clear_reset_vbar_storage(clear);
```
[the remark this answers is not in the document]
awakecoding: @akallabeth agreed, this should not be merged unless we can hit this particular code path and ensure it doesn't break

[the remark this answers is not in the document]
awakecoding: to avoid causing another possible regression like with the xcrush context reset, I will modify the current PR to keep the current reset behaviour, just with cleaned up code. I'll add a bool parameter to do a "full" internal reset, and make sure this call site only does a "shallow" reset that only touches the indices. I know there was an issue years ago with pcap replaying, but it would be safer to revisit it at the very end.

awakecoding:
@akallabeth I modified this PR to avoid resetting the context more than it previously did, but it's still cleaner now. I have a bunch of other PRs to split and merge individually until I've got enough to try and get my pcap replayer working on top of the latest master, after which I would revisit what may be broken with regards to context resetting.

akallabeth (approved):

PR #7930, opened 2022-05-26
Fix XCRUSH context reset

awakecoding:
Fix XCRUSH context resetting. This issue was originally observed with Lync RDP traffic while replaying a recorded pcap file offline.


hardening (approved):
LGTM

akallabeth (approved):

PR #7931, opened 2022-05-26
Fix processing of RemoteFX and NSCodec codec ids

awakecoding:
Fix edge cases with RemoteFX and NSCodec IDs (RDP_CODEC_ID_REMOTEFX, RDP_CODEC_ID_IMAGE_REMOTEFX, RDP_CODEC_ID_NSCODEC). These were originally observed with Lync RDP traffic but could happen with anything, so it doesn't hurt to correctly process the codec ids to determine how to decode the images.


hardening (approved):
LGTM

akallabeth (approved):

PR #7934, opened 2022-05-29
implement proper SPNEGO negotiation

fifthdegree:
- Implement SPNEGO client- and server-side
- Implement Make and Verify Signature functions for NTLM and kerberos
- Avoid using SEC_I_COMPLETE_NEEDED in NTLM (spnego may need to use signature functions before the last call)
- Move acquiring kerberos credentials to its proper function
- Use the default kerberos cache only if initialized to the right principal
- Use one structure for pAuthData


akallabeth:
@freerdp-bot test

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -121,119 +110,12 @@ static void kerberos_ContextFree(KRB_CONTEXT* context)
...
 	}
 
-	sspi_gss_release_cred(&minor_status, &context->cred);
-
-	if (context->gss_ctx)
-	{
-		sspi_gss_delete_sec_context(&minor_status, &context->gss_ctx, SSPI_GSS_C_NO_BUFFER);
-		context->gss_ctx = SSPI_GSS_C_NO_CONTEXT;
-	}
+	// if (context->gss_ctx)
```
hardening, line 113: Should remove that before merge

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -244,7 +126,7 @@ static krb5_error_code krb5_prompter(krb5_context context, void* data, const cha
 
 	for (i = 0; i < num_prompts; i++)
 	{
-		if (ptypes && ptypes[i] == KRB5_PROMPT_TYPE_PREAUTH)
+		if (ptypes && ptypes[i] == KRB5_PROMPT_TYPE_PREAUTH && data)
```
hardening, line 129: Just curious, have you ever hit the case where `data` is NULL ?

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -253,458 +135,377 @@ static krb5_error_code krb5_prompter(krb5_context context, void* data, const cha
...
-		}
+	sspi_SecureHandleInvalidate(phCredential);
+	return SEC_E_OK;
+}
 
+static SECURITY_STATUS SEC_ENTRY kerberos_QueryCredentialsAttributesW(PCredHandle phCredential,
+                                                                      ULONG ulAttribute,
+                                                                      void* pBuffer)
+{
+	if (ulAttribute == SECPKG_CRED_ATTR_NAMES)
```
hardening, line 417: Can we change that to a `switch()` statement ?

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -253,458 +135,377 @@ static krb5_error_code krb5_prompter(krb5_context context, void* data, const cha
...
+		if (p)
+			*p = '@';
 
-		if (kerberos_SetContextServicePrincipalNameA(context, pszTargetName) < 0)
-		{
-			kerberos_ContextFree(context);
-			return SEC_E_INTERNAL_ERROR;
-		}
+		major = sspi_gss_import_name(&minor, &target_name, SSPI_GSS_C_NT_HOSTBASED_SERVICE,
+		                             &context->target_name);
```
hardening, line 469: We shall check `major` value for errors ?

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -713,16 +514,43 @@ static SECURITY_STATUS SEC_ENTRY kerberos_InitializeSecurityContextA(
...
-	if (SSPI_GSS_ERROR(context->major_status))
+	if (SSPI_GSS_ERROR(major))
+	{
+		gss_log_status_messages(major, minor);
 		return SEC_E_INTERNAL_ERROR;
+	}
 
 	return SEC_E_OK;
+#else
+	return SEC_E_INTERNAL_ERROR;
```
hardening, line 531: Not your code, but don't we have a `not implemented` return code (something like function not implemented could do the job)

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -72,45 +99,171 @@ const SecPkgInfoW NEGOTIATE_SecPkgInfoW = {
 	NEGOTIATE_SecPkgInfoW_Comment /* Comment */
 };
 
-static void negotiate_SetSubPackage(NEGOTIATE_CONTEXT* context, const TCHAR* name)
+static const sspi_gss_OID_desc spnego_OID = { 6, "\x2b\x06\x01\x05\x05\x02" };
```
hardening, line 102: Not a big deal, but IIRC these are already defined in libkrb5.
Another remark for kerberos there's 2 OID: the official one and the microsoft's one (the official one with endianess problem in the middle of the OID). You can check in libkrb5's code to see how they adapt to that.

winpr/libwinpr/sspi/Kerberos/kerberos.c
fifthdegree: It should actually be uncommented; I'll do that

winpr/libwinpr/sspi/Kerberos/kerberos.c
fifthdegree: I don't think so... I don't remember why I put that in although I suppose it doesn't hurt

winpr/libwinpr/sspi/Kerberos/kerberos.c
fifthdegree: As it stands the function doesn't actually do anything in any case, but I could do that (not my code by the way)

winpr/libwinpr/sspi/Kerberos/kerberos.c
fifthdegree: I did in fact write that (diffs can get confusing) but you're right; I'll change that

winpr/libwinpr/sspi/Negotiate/negotiate.c
fifthdegree: Couldn't find them in the krb5 headers, but either way I think it's a bit much to include them just for that (especially since they may not be available in every build). Although it might be better to define them in one of the sspi headers
Regarding the old microsoft OID, I'll try to put in support for it

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -72,45 +99,171 @@ const SecPkgInfoW NEGOTIATE_SecPkgInfoW = {
...
+	return buf;
+}
+
+static BYTE* negotiate_read_tlv(BYTE* buf, BYTE* tag, ULONG* len, ULONG* bytes_remain)
+{
+	int len_bytes = 0;
+	*len = 0;
+
+	*bytes_remain -= 2;
+	if (*bytes_remain < 0)
```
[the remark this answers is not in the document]
fifthdegree: Yup, an unsigned number will never be less than zero. Oh well, I was trying to accomplish two things at once (callers need to check anyway that values don't exceed the end of the buffer)

hardening:
@freerdp-bot can you give another test please ?

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -36,17 +36,44 @@
 static const char NEGO_REG_KEY[] =
     "Software\\" WINPR_VENDOR_STRING "\\" WINPR_PRODUCT_STRING "\\SSPI\\Negotiate";
 
+#define ASN_LENGTH_BYTES(len)  \
```
hardening, line 39: That's so unfortunate because we have these computations elsewhere in ber modules (so in FreeRDP). Does it suggest that this should be moved to winPR (any opinion @akallabeth ?)
I'd probably prefer a `static INLINE` function instead of the define with cascading tri-operator.

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -72,45 +99,184 @@ const SecPkgInfoW NEGOTIATE_SecPkgInfoW = {
...
+		MoveMemory(buf, value, len);
+		buf += len;
 	}
+
+	return buf;
+}
+
+static BYTE* negotiate_write_contextual_tlv(BYTE* buf, BYTE contextual, BYTE tag, ULONG len,
+                                            BYTE* value)
+{
```
hardening, line 167: `value` could be `const BYTE*`
I'd also suggest to swap the order of `len` and `value` (most functions have it in this order: data, len)

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -72,45 +99,184 @@ const SecPkgInfoW NEGOTIATE_SecPkgInfoW = {
...
+	enum NegState negState;
+	BOOL init;
+	sspi_gss_OID_desc supportedMech;
+	SecBuffer mechTypes;
+	SecBuffer mechToken;
+	SecBuffer mic;
+} NegToken;
+
+static BYTE* negotiate_write_tlv(BYTE* buf, BYTE tag, ULONG len, BYTE* value)
```
hardening, line 138: `value` could be `const BYTE*`
We don't use `ULONG` (`len` parameter) often for that kind of stuff `size_t` would be better

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -72,45 +99,184 @@ const SecPkgInfoW NEGOTIATE_SecPkgInfoW = {
...
+		return "Kerberos [wrong OID] (1.2.840.48018.1.2.2)";
+	else if (negotiate_oid_compare(oid, &ntlm_OID))
+		return "NTLM (1.3.6.1.4.1.311.2.2.10)";
+	else
+		return "Unknown mechanism";
+}
+
+static const Mech* negotiate_GetMechByOID(sspi_gss_OID_desc oid)
+{
+	if (negotiate_oid_compare(&oid, &kerberos_wrong_OID))
```
hardening, line 251: If you add an entry in `MechTable` for `kerberos_wrong_OID` (pointing to regular kerberos) that may do the job instead of having that special case

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -72,45 +99,184 @@ const SecPkgInfoW NEGOTIATE_SecPkgInfoW = {
...
+
+static PSecHandle negotiate_FindCredential(MechCred* creds, const Mech* mech)
+{
+	for (int i = 0; i < MECH_COUNT; i++)
+	{
+		if (creds[i].mech == mech)
+		{
+			if (creds[i].valid)
+				return &creds[i].cred;
+			break;
```
hardening, line 273: perhaps `return NULL;` would be clearer ?

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -160,143 +326,544 @@ static BOOL negotiate_get_config(BOOL* kerberos, BOOL* ntlm)
 	return TRUE;
 }
 
-static SECURITY_STATUS SEC_ENTRY negotiate_InitializeSecurityContextW(
-    PCredHandle phCredential, PCtxtHandle phContext, SEC_WCHAR* pszTargetName, ULONG fContextReq,
-    ULONG Reserved1, ULONG TargetDataRep, PSecBufferDesc pInput, ULONG Reserved2,
-    PCtxtHandle phNewContext, PSecBufferDesc pOutput, PULONG pfContextAttr, PTimeStamp ptsExpiry)
+static BOOL negotiate_write_neg_token(PSecBuffer output_buffer, NegToken token)
```
hardening, line 329: Perhaps passing token as `const NegToken *` would be nice (just an idea)

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -160,143 +326,544 @@ static BOOL negotiate_get_config(BOOL* kerberos, BOOL* ntlm)
...
-		sspi_SecureHandleSetUpperPointer(phNewContext, (void*)NEGO_SSP_NAME);
-	}
 
-	/* if Kerberos has previously failed or WITH_GSSAPI is not defined, we use NTLM directly */
-	if (kerberos && !ErrorInitContextKerberos)
+		if (!context->spnego)
+			return status;
+
+		/* Write the SEQUENCE tag */
+		p = negotiate_write_tlv(context->mechTypes.pvBuffer, 0x30, inner_mech_list_len, NULL);
```
hardening, line 713: It's not obvious how the check is done that `context->mechTypes.pvBuffer` has enough space to hold what we have to write in it...

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -160,143 +326,544 @@ static BOOL negotiate_get_config(BOOL* kerberos, BOOL* ntlm)
...
+			else
+				context->state = NEGOTIATE_STATE_FINAL;
+		}
+
+		/* Check if the acceptor sent its final token without a mic */
+		if (context->state == NEGOTIATE_STATE_FINAL && input_token.mic.cbBuffer == 0)
+		{
+			if (context->mic || input_token.negState != ACCEPT_COMPLETED)
+				return SEC_E_INVALID_TOKEN;
+			else
```
hardening, line 816: Could drop the `else` tha would make the code nicer

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -305,32 +872,263 @@ static SECURITY_STATUS SEC_ENTRY negotiate_AcceptSecurityContext(
...
+	creds = sspi_SecureHandleGetLowerPointer(phCredential);
+	input_buffer = sspi_FindSecBuffer(pInput, SECBUFFER_TOKEN);
+	if (pOutput)
+		output_buffer = sspi_FindSecBuffer(pOutput, SECBUFFER_TOKEN);
 
 	if (!context)
 	{
-		context = negotiate_ContextNew();
+		/* Check for NTLM token */
+		if (strncmp(input_buffer->pvBuffer, "NTLMSSP", 8) == 0)
```
hardening, line 905: We shall check that `input_buffer->cbBuffer` is at least 8 before doing this comparison

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -355,189 +1153,125 @@ static SECURITY_STATUS SEC_ENTRY negotiate_DeleteSecurityContext(PCtxtHandle phC
...
-
-	if (context->sspiW->QueryContextAttributesW)
-		status =
-		    context->sspiW->QueryContextAttributesW(&(context->SubContext), ulAttribute, pBuffer);
-
-	return status;
+	if (context->mech->pkg->table_w->QueryContextAttributesW)
+		return context->mech->pkg->table_w->QueryContextAttributesW(&context->sub_context,
+		                                                            ulAttribute, pBuffer);
+	else
```
hardening, line 1191: Drop the else

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -625,83 +1331,92 @@ static SECURITY_STATUS SEC_ENTRY negotiate_QueryCredentialsAttributesA(PCredHand
...
+	NEGOTIATE_CONTEXT* context = (NEGOTIATE_CONTEXT*)sspi_SecureHandleGetLowerPointer(phContext);
 
-	if (context->sspiW->EncryptMessage)
-		status =
-		    context->sspiW->EncryptMessage(&(context->SubContext), fQOP, pMessage, MessageSeqNo);
+	if (!context)
+		return SEC_E_INVALID_HANDLE;
+
+	if (context->mic)
+		MessageSeqNo++;
```
hardening, line 1356: Hum that's very strange that `MessageSeqNo` is passed by value

hardening:
There's a few security show stoppers, but overall it's a very good job.
Next step will be to add Kerberos server-side and we'll be able to accept rdesktop and FreeRDP with NLA + Kerberos.
For mstsc we will need user2user, but the custom SPNego impl is interesting to treat this case: almost impossible to do with libkrb5 because of their SPNego impl that trusts the announced mechs instead of looking at the real OID of the optimistic token.

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -72,45 +99,184 @@ const SecPkgInfoW NEGOTIATE_SecPkgInfoW = {
...
+static const sspi_gss_OID_desc ntlm_OID = { 10, "\x2b\x06\x01\x04\x01\x82\x37\x02\x02\x0a" };
+
+static const SecPkg SecPkgTable[] = {
+	{ KERBEROS_SSP_NAME, &KERBEROS_SecurityFunctionTableA, &KERBEROS_SecurityFunctionTableW },
+	{ NTLM_SSP_NAME, &NTLM_SecurityFunctionTableA, &NTLM_SecurityFunctionTableW }
+};
+
+static const Mech MechTable[] = {
+	{ &kerberos_OID, &SecPkgTable[0], 0, 0 },
+	{ &ntlm_OID, &SecPkgTable[1], 0, 0 },
```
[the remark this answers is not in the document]
fifthdegree: No, because we use that table to iterate through the available mechanisms and for the order of preferred mechanisms

winpr/libwinpr/sspi/Negotiate/negotiate.c
fifthdegree: The reason for that order is because it's the order used by ASN (Tag, Length, Value) but no objection if you want to change that

winpr/libwinpr/sspi/Negotiate/negotiate.c
fifthdegree: Yes you're absolutely right

winpr/libwinpr/sspi/Negotiate/negotiate.c
fifthdegree: It's set in `mechTypes.cbBuffer` and allocated in `negotiate_ContextNew()`

winpr/libwinpr/sspi/Negotiate/negotiate.c
fifthdegree: Why's that?

winpr/libwinpr/sspi/Negotiate/negotiate.c
fifthdegree: Ah, I see

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -72,45 +99,184 @@ const SecPkgInfoW NEGOTIATE_SecPkgInfoW = {
...
+static BYTE* negotiate_write_contextual_tlv(BYTE* buf, BYTE contextual, BYTE tag, ULONG len,
+                                            BYTE* value)
+{
+	buf = negotiate_write_tlv(buf, contextual, ASN_TLV_LENGTH(len), NULL);
+	buf = negotiate_write_tlv(buf, tag, len, value);
+
+	return buf;
+}
+
+static BYTE* negotiate_read_tlv(BYTE* buf, BYTE* tag, ULONG* len, ULONG* bytes_remain)
```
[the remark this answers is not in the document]
fifthdegree: That may not accomplish much since buffers are sometimes assigned to `void *` elements of structures (i.e. `SecBuffer->pvBuffer`)

winpr/libwinpr/sspi/Negotiate/negotiate.c
hardening: Ok I think I understand the thing, it's that when we have a MIC the real sequence number is always shifted by 1

libfreerdp/core/nla.c
```diff
@@ -1634,8 +1600,7 @@ static int nla_server_authenticate(rdpNla* nla)
 				}
 			}
 
-			if (!nla_complete_auth(nla, &outputBufferDesc))
-				goto fail;
+			hash_set = TRUE;
 		}
```
hardening, line 1604: Ok for not setting multiple time the hash callback and data, but why removing the `nla_complete_auth()` call. Does server-side still work after your changes ?

fifthdegree: So it seems that the return value `SEC_I_COMPLETE_NEEDED` was being used to signal that certain information (SamFile / HashCallback) was needed to continue setting up the context. They don't need to be set at any particular stage; but it needs to be after at least one call to `AcceptSecurityContext()` so that there is a context to set it in (in fact, fwiw, I think it would be better to have them passed into `AquireCredentialHandle()` and stored in the credential). Anyway, it's difficult to use `COMPLETE_NEEDED` in this way with spnego because `AcceptSecurityContext()` needs the context set up to use the signature functions before returning the last call

fifthdegree: After looking this over I realized this won't work when NTLM is not used directly, because there still won't be an ntlm context after the first call. I still think the best option is to store the information in the credential

fifthdegree: Should we move the callback function's definition to sspi.h, do you think? Also, do you think we should add elements to `SSPI_CREDENTIALS` or define a new `NTLM_CREDENTIALS` structure?

include/freerdp/peer.h
```diff
@@ -148,7 +147,7 @@ struct rdp_freerdp_peer
 	ALIGN64 psPeerGetEventHandles GetEventHandles;
 	ALIGN64 psPeerAdjustMonitorsLayout AdjustMonitorsLayout;
 	ALIGN64 psPeerClientCapabilities ClientCapabilities;
-	ALIGN64 psPeerComputeNtlmHash ComputeNtlmHash;
```
hardening, line 151: It would be propably better to deprecate that field, and put the new one at then end of the struct (look at FREERDP_DEPRECATED in other places)

fifthdegree: Sorry no... I'm just not quite sure I understand how that would work. Should we have a second field in `NTLM_CONTEXT` as well? Which would be used when compiling with the DEPRECATED option?

hardening: IMHO we shall deprecate `ComputeNtlmHash` and have a new field (some "sspi" in the name could make sense) with the right type. And of course have the rest of the code use the new field.

fifthdegree: OK, got it

include/freerdp/peer.h
```diff
@@ -147,7 +148,13 @@ struct rdp_freerdp_peer
...
 	ALIGN64 psPeerClientCapabilities ClientCapabilities;
-	ALIGN64 psSspiComputeNtlmHash ComputeNtlmHash;
+#if defined(WITH_FREERDP_DEPRECATED)
+	WINPR_DEPRECATED_VAR("Use freerdp_peer::SspiNtlmHashCallback instead",
+	                     ALIGN64 psPeerComputeNtlmHash ComputeNtlmHash;)
+#else
+	UINT64 reserved2;
+#endif
+	ALIGN64 psSspiNtlmHashCallback SspiNtlmHashCallback;
 	ALIGN64 psPeerLicenseCallback LicenseCallback;
```
hardening, line 158: Please put that new field at the end of the struct to not break ABI

hardening:

winpr/include/winpr/sspi.h
```diff
@@ -598,19 +598,7 @@ typedef struct
...
-	char* armorCache;
-	char* pkinitX509Anchors;
-	char* pkinitX509Identity;
-	BOOL withPac;
-	INT32 startTime;
-	INT32 renewLifeTime;
-	INT32 lifeTime;
-	BYTE certSha1[20];
-} SEC_WINPR_KERBEROS_SETTINGS;
+#define SEC_WINNT_AUTH_IDENTITY_EXTENDED 0x4
```
hardening, line 601: I'd suggest to choose another value as it seems like it is already used in `SEC_WINNT_AUTH_IDENTITY_EXW.flags`for `SEC_WINNT_AUTH_IDENTITY_MARSHALLED`

winpr/libwinpr/sspi/NTLM/ntlm.c
```diff
@@ -313,10 +314,30 @@ static SECURITY_STATUS SEC_ENTRY ntlm_AcquireCredentialsHandleW(
...
 	credentials->pvGetKeyArgument = pvGetKeyArgument;
+
 	identity = (SEC_WINNT_AUTH_IDENTITY*)pAuthData;
 
 	if (identity)
+	{
 		sspi_CopyAuthIdentity(&(credentials->identity), identity);
+		if (identity->Flags & SEC_WINNT_AUTH_IDENTITY_EXTENDED)
+			settings = &((SEC_WINNT_AUTH_IDENTITY_WINPR*)pAuthData)->ntlmSettings;
+	}
```
hardening, line 325: As we're in the unicode variant of the function another possibility (than adding a new flag) would be that `pAuthData` would be a kind of `SEC_WINNT_AUTH_IDENTITY_EXW` (a NTLM specialized one) where you can distinguish the various kind by looking at the `Version` and `Length` fields (that's what they do in win32 implementations). Anyway that has the drawback that the ascii version of the function have to build a unicode version of the parameters.
Your design is probably the best, I just wanted to share my thoughts.

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -72,45 +90,203 @@ const SecPkgInfoW NEGOTIATE_SecPkgInfoW = {
...
+{
+	enum NegState negState;
+	BOOL init;
+	sspi_gss_OID_desc supportedMech;
+	SecBuffer mechTypes;
+	SecBuffer mechToken;
+	SecBuffer mic;
+} NegToken;
+
+static inline size_t asn_tlv_length(size_t len)
```
hardening, line 129: s/inline/INLINE/

hardening:
A few fixes, make the CI happy (do we have to compile Negociate SSPI under win32 ?) and we should be good for a merge.

winpr/libwinpr/sspi/NTLM/ntlm.c
fifthdegree: True. Since there's a flags field anyway we may as well use it... makes for cleaner code IMHO

hardening:
@freerdp-bot can you test the thing ?

include/freerdp/peer.h
```diff
@@ -147,7 +148,13 @@ struct rdp_freerdp_peer
 	ALIGN64 psPeerGetEventHandles GetEventHandles;
 	ALIGN64 psPeerAdjustMonitorsLayout AdjustMonitorsLayout;
 	ALIGN64 psPeerClientCapabilities ClientCapabilities;
-	ALIGN64 psSspiComputeNtlmHash ComputeNtlmHash;
+#if defined(WITH_FREERDP_DEPRECATED)
+	WINPR_DEPRECATED_VAR("Use freerdp_peer::SspiNtlmHashCallback instead",
+	                     ALIGN64 psPeerComputeNtlmHash ComputeNtlmHash;)
+#else
+	UINT64 reserved2;
```
akallabeth, line 155: If you are concerned about API/ABI compatility, then you need to add the new callback at the end of the struct, otherwise this `reserved2` field is useless.

server/shadow/shadow_server.c
```diff
@@ -392,6 +392,10 @@ int shadow_server_parse_command_line(rdpShadowServer* server, int argc, char** a
 			if (!freerdp_settings_set_bool(settings, FreeRDP_GfxAVC444, arg->Value ? TRUE : FALSE))
 				return COMMAND_LINE_ERROR;
 		}
+		CommandLineSwitchCase(arg, "keytab")
+		{
+			freerdp_settings_set_string(settings, FreeRDP_KerberosKeytab, arg->Value);
```
akallabeth, line 397: missing return value check

hardening:
I did some tests and at least with MIT KRB5 the client side SPNEGO doesn't work when you already have a cache with a non matching principal. The main reason is that in `kerberos_AcquireCredentialsHandle`, when we try to acquire creds we only catch `SSPI_GSS_S_NO_CRED` and it seems like the return value is `GSS_S_CRED_UNAVAIL` (which equals to `GSS_S_FAILURE`). That completely disables Kerberos in the nego. I'll push a fix for that on your branch.

fifthdegree:
That's strange... we re-initialize the cache if it's not initialized with the right principle. Are you providing a cache name as an option or are you talking about the default cache?

fifthdegree:
@hardening, can you verify that the reason it's failing is because the principle doesn't match? I'm not getting the same results

hardening:
> That's strange... we re-initialize the cache if it's not initialized with the right principle. Are you providing a cache name as an option or are you talking about the default cache?

I'm using the default cache. In that cache, the principal doesn't match. I've checked the code of libkrb5 and in the header there's a `
/* XXXX This is a necessary evil until the spec is fixed */
#define GSS_S_CRED_UNAVAIL GSS_S_FAILURE
`
and if the cache is empty (as we have with the freshly created `MEMORY:` one) the return result is `GSS_S_CRED_UNAVAIL`, so `GSS_S_FAILURE` and we only catch `SSPI_GSS_S_NO_CRED`...

fifthdegree:
Interesting, that doesn't happen on my setup. Regardless, it's not such a big deal to try to get a ticket on any failure. I'll patch that

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -253,458 +135,416 @@ static krb5_error_code krb5_prompter(krb5_context context, void* data, const cha
...
-	if (options)
-		krb5_get_init_creds_opt_free(ctx, options);
+	krb5_cc_get_full_name(ctx, ccache, &ccache_name);
+	krb5_kt_get_name(ctx, keytab, keytab_name, PATH_MAX);
 
-#endif
-
-	if (init_ctx)
-		krb5_init_creds_free(ctx, init_ctx);
+	cred_store_opts[0].key = "ccache";
```
hardening, line 335: I'm not sure that it's a so good idea to provide the keytab option because if `keytab_name` is set but the file not accessible for some reason, it stops the call to `gss_acquire_cred_from()` (I've looked at the source code).
The options given in `cred_store_opts` are a logical AND not a OR (all given options must be retrieved, unless the value is NULL).

fifthdegree: I suppose we can just provide it only if acceptor credentials are required

fifthdegree: Scratch that, it's only provided server side. But isn't it necessary to have the keytab anyway?

hardening: Well the client could wish to log in with a keytab, even if it doesn't fit very well with RDP. 
I was wondering if the key name is `client_keytab` or `keytab` the doc is not very clear.

hardening:
@fifthdegree I've just pushed my fixes, tell me if it still work for you after that

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -598,8 +598,7 @@ static SECURITY_STATUS negotiate_mic_exchange(NEGOTIATE_CONTEXT* context, NegTok
 	/* If peer expects a MIC then generate it */
 	if (input_token->negState != ACCEPT_COMPLETED)
 	{
-		if (context->mech->pkg->table->QueryContextAttributesA(
-		        &context->sub_context, SECPKG_ATTR_SIZES, &sizes) != SEC_E_OK)
+		if (table->QueryContextAttributesA(&context->sub_context, SECPKG_ATTR_SIZES, &sizes) != SEC_E_OK)
```
fifthdegree, line 601: Looking at this, I realize we don't actually do anything with `sizes`...

fifthdegree:
Yep, it's good, although I would make that function `inline` (sorry microsoft, _`__inline`_)

winpr/libwinpr/sspi/Kerberos/kerberos.c
fifthdegree: I'm pretty sure it's `client_keytab`. We could add a keytab option client side, but then we'd run into the issue you raised

hardening:

hardening:
I did some server-side tests and it works correctly (except for mstsc but that's rather out-of-scope as we'd need user to user and that can be a next PR).

winpr/libwinpr/sspi/Negotiate/negotiate.c
hardening: After some thoughts and some reverse inquiries, I think that probably means that we must reintroduce `msasn1` in winPR...

hardening:
In the client impl, if the ccache contains expired credentials we fail, I guess we shall either update them in the existing ccache or use a new empty `MEMORY:` ccache.

fifthdegree:
@hardening can you check if it works correctly now?

hardening:
@freerdp-bot can you do the test before commit ?

hardening:
@freerdp-bot can you test the conflict resolution ?

hardening (approved):

PR #7937, opened 2022-05-30
Revert "* libfreerdp-codec: fix XCRUSH context reset"

awakecoding:
Revert the recent change to call xcrush_context_reset(xcrush, FALSE); when Level2ComprFlags & PACKET_FLUSHED is true. I know this was once required to fix xcrush-related issues while replaying pcap files, but it apparently causes random decompression corruption failures for regular connections. Let's just revert the change for now, and revisit it later.


hardening:
Would be fantastic if you had a dump of failing traffic...

PR #7942, opened 2022-06-01
Implement RDP persistent bitmap cache

awakecoding:
Implement the RDP persistent bitmap cache using the same file format as mstsc.exe on Windows. This is code I wrote years ago in order to process persisted bitmap caches and reuse them between sequential packet captures for offline packet replaying.

A good portion of the pull request deals with the bitmap cache file itself, the rest are improvements to load, use and save the persistent bitmap files in RDPGFX. The bitmap cache format has V2/V3 file format, we support both. Some PDUs during the connection sequence had to be modified to correctly support the cache offer/reply.


channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -668,6 +598,291 @@ static UINT rdpgfx_recv_evict_cache_entry_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
...
+
+fail:
+	Stream_Free(s, TRUE);
+	return error;
+}
+
+static int rdpgfx_send_cache_offer(RDPGFX_PLUGIN* gfx)
+{
+	int index, count;
+	PERSISTENT_CACHE_ENTRY entry;
```
hardening, line 776: Can we rename `index` to `idx` to avoid the clash with the `index` function (pure cosmetic change, no build implication) ?

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -689,19 +904,22 @@ static UINT rdpgfx_recv_cache_import_reply_pdu(RDPGFX_CHANNEL_CALLBACK* callback
 	if (!Stream_CheckAndLogRequiredLength(TAG, s, 2ull * pdu.importedEntriesCount))
 		return ERROR_INVALID_DATA;
 
-	pdu.cacheSlots = (UINT16*)calloc(pdu.importedEntriesCount, sizeof(UINT16));
+	if (pdu.importedEntriesCount > RDPGFX_CACHE_ENTRY_MAX_COUNT)
+		return ERROR_INVALID_DATA;
 
-	if (!pdu.cacheSlots)
+	if (Stream_GetRemainingLength(s) < (size_t)(pdu.importedEntriesCount * 2))
 	{
```
hardening, line 911: `if (Stream_GetRemainingLength(s) / 2 < (size_t)pdu.importedEntriesCount)` would be better for int overflow

libfreerdp/cache/persistent.c
```diff
@@ -0,0 +1,327 @@
...
+	entry->key64 = entry3.key64;
+	entry->width = entry3.width;
+	entry->height = entry3.height;
+	entry->size = entry3.width * entry3.height * 4;
+	entry->flags = 0;
+
+	if (entry->size > persistent->bmpSize)
+	{
+		persistent->bmpSize = entry->size;
+		persistent->bmpData = (BYTE*)realloc(persistent->bmpData, persistent->bmpSize);
```
hardening, line 133: Shall use a temporary variable here or we leak memory if realloc fails

libfreerdp/core/activation.c
```diff
@@ -247,32 +247,140 @@ static BOOL rdp_write_persistent_list_entry(wStream* s, UINT32 key1, UINT32 key2
...
+
+	keyCount = rdp_load_persistent_key_list(rdp, &keyList);
+
+	info.totalEntriesCache0 = settings->BitmapCacheV2CellInfo[0].numEntries;
+	info.totalEntriesCache1 = settings->BitmapCacheV2CellInfo[1].numEntries;
+	info.totalEntriesCache2 = settings->BitmapCacheV2CellInfo[2].numEntries;
+	info.totalEntriesCache3 = settings->BitmapCacheV2CellInfo[3].numEntries;
+	info.totalEntriesCache4 = settings->BitmapCacheV2CellInfo[4].numEntries;
+
+	info.numEntriesCache0 =
```
hardening, line 356: all these are `MIN()` calls, aren't they ?

awakecoding:
@hardening I made the recommended changes, and also fixed a few warnings. I found a bug that prevented the persistent key list PDU from being sent, but oddly enough when sending it for real I hit a problem where the server won't accept too many keys at once. MS-RDPCGR says the client SHOULD not send more than 169 entries at once, but from experimentation, the server rejects anything more than 2042 entries at once. Rather than send a list of persistent key list PDUs, I decided just to advertise the maximum bitmap cache entries that can be sent at once. We can revisit this at a later time

awakecoding:
@akallabeth this PR is ready to go, and I don't want to let it go stale :)

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -668,14 +598,299 @@ static UINT rdpgfx_recv_evict_cache_entry_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
...
+static int rdpgfx_save_persistent_cache(RDPGFX_PLUGIN* gfx)
+{
+	int index;
+	UINT16 cacheSlot;
+	PERSISTENT_CACHE_ENTRY cacheEntry;
+	rdpPersistentCache* persistent = NULL;
+	rdpSettings* settings = gfx->settings;
+	RdpgfxClientContext* context = (RdpgfxClientContext*)gfx->iface.pInterface;
+
+	if (!context || !context->ExportCacheEntry)
```
hardening, line 665: I'm wondering if it should be `WINPR_ASSERT(context);` ?

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -668,14 +598,299 @@ static UINT rdpgfx_recv_evict_cache_entry_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
...
+		count = RDPGFX_CACHE_ENTRY_MAX_COUNT - 1;
+
+	if (count > gfx->MaxCacheSlots)
+		count = gfx->MaxCacheSlots;
+
+	offer->cacheEntriesCount = (UINT16)count;
+
+	for (index = 0; index < count; index++)
+	{
+		persistent_cache_read_entry(persistent, &entry);
```
hardening, line 643: return value is not checked

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -668,14 +598,299 @@ static UINT rdpgfx_recv_evict_cache_entry_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
...
+	if (count > gfx->MaxCacheSlots)
+		count = gfx->MaxCacheSlots;
+
+	offer.cacheEntriesCount = (UINT16) count;
+
+	WLog_DBG(TAG, "Sending Cache Import Offer: %d", count);
+
+	for (idx = 0; idx < count; idx++)
+	{
+		persistent_cache_read_entry(persistent, &entry);
```
hardening, line 815: return value is not checked

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -668,14 +598,299 @@ static UINT rdpgfx_recv_evict_cache_entry_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
...
+
+	count = persistent_cache_get_count(persistent);
+
+	count = (count < reply->importedEntriesCount) ? count : reply->importedEntriesCount;
+
+	WLog_DBG(TAG, "Receiving Cache Import Reply: %d", count);
+
+	for (idx = 0; idx < count; idx++)
+	{
+		persistent_cache_read_entry(persistent, &entry);
```
hardening, line 871: return value is not checked

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -1643,6 +1867,8 @@ static UINT rdpgfx_recv_pdu(RDPGFX_CHANNEL_CALLBACK* callback, wStream* s)
 				WLog_Print(gfx->log, WLOG_ERROR,
 				           "rdpgfx_recv_caps_confirm_pdu failed with error %" PRIu32 "!", error);
 
+			rdpgfx_send_cache_offer(gfx);
```
hardening, line 1870: return value shall be propagated ?

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -689,19 +904,22 @@ static UINT rdpgfx_recv_cache_import_reply_pdu(RDPGFX_CHANNEL_CALLBACK* callback
...
 
-	for (index = 0; index < pdu.importedEntriesCount; index++)
+	for (idx = 0; idx < (int) pdu.importedEntriesCount; idx++)
 	{
-		Stream_Read_UINT16(s, pdu.cacheSlots[index]); /* cacheSlot (2 bytes) */
+		Stream_Read_UINT16(s, pdu.cacheSlots[idx]); /* cacheSlot (2 bytes) */
 	}
 
+	rdpgfx_load_cache_import_reply(gfx, &pdu);
+
```
hardening, line 922: The result shall be checked, I'm wondering what we must do if the import fails (disconnect or continue ?)

libfreerdp/cache/bitmap.c
```diff
@@ -278,6 +282,108 @@ void bitmap_cache_register_callbacks(rdpUpdate* update)
 	}
 }
 
+int bitmap_cache_load_persistent(rdpBitmapCache* bitmapCache)
+{
```
hardening, line 286: AFAICT this function is never called and does not much (its only function would be to check that the cache file "loadable"), shall we remove it ?

libfreerdp/cache/persistent.c
```diff
@@ -0,0 +1,333 @@
...
+	entry->key64 = entry2.key64;
+	entry->width = entry2.width;
+	entry->height = entry2.height;
+	entry->size = entry2.width * entry2.height * 4;
+	entry->flags = entry2.flags;
+
+	entry->data = persistent->bmpData;
+
+	if (fread((void*)entry->data, 0x4000, 1, persistent->fp) != 0)
+		return -1;
```
hardening, line 61: Looks like the test of `fread()` result is inverted

libfreerdp/cache/persistent.c
```diff
@@ -0,0 +1,333 @@
...
+
+	return 1;
+}
+
+int persistent_cache_write_entry_v2(rdpPersistentCache* persistent, PERSISTENT_CACHE_ENTRY* entry)
```
hardening, line 66: `entry` could be `const`

hardening:
Sorry when reviewing the fixes, I found some more things to fix.
There's possibly functions that could be made static in persistent.c.

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -668,14 +598,299 @@ static UINT rdpgfx_recv_evict_cache_entry_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
...
+		return -1;
+
+	if (persistent_cache_open(persistent, settings->BitmapCachePersistFile, TRUE, 3) < 1)
+		goto error;
+
+	for (index = 0; index < gfx->MaxCacheSlots; index++)
+	{
+		if (gfx->CacheSlots[index])
+		{
+			cacheSlot = (UINT16)index;
```
akallabeth, line 686: Better declare variables in scope

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -668,14 +598,299 @@ static UINT rdpgfx_recv_evict_cache_entry_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
...
+{
+	int idx;
+	int count;
+	UINT16 cacheSlot;
+	PERSISTENT_CACHE_ENTRY entry;
+	rdpPersistentCache* persistent = NULL;
+	rdpSettings* settings = gfx->settings;
+	RdpgfxClientContext* context = (RdpgfxClientContext*)gfx->iface.pInterface;
+
+	if (!settings->BitmapCachePersistEnabled)
```
akallabeth, line 846: Repeated checks (`RDPGFX_PLUGIN* gfx` valid, `settings` proper, ...)
maybe unify in a static function doing all these checks?

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -948,6 +1165,9 @@ static UINT rdpgfx_recv_wire_to_surface_1_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
 	pdu.bitmapData = Stream_Pointer(s);
 	Stream_Seek(s, pdu.bitmapDataLength);
 
+	if (pdu.codecId == 5)
```
akallabeth, line 1168: Magic number, better use a define or at least a comment what this value is

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -1026,6 +1246,10 @@ static UINT rdpgfx_recv_wire_to_surface_2_pdu(RDPGFX_CHANNEL_CALLBACK* callback,
 	Stream_Read_UINT32(s, pdu.bitmapDataLength); /* bitmapDataLength (4 bytes) */
 	pdu.bitmapData = Stream_Pointer(s);
 	Stream_Seek(s, pdu.bitmapDataLength);
+
+	if (pdu.codecId == 5)
```
akallabeth, line 1250: Repeated magic number

client/common/cmdline.c
```diff
@@ -3082,6 +3082,17 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
 		{
 			settings->BitmapCacheEnabled = enable;
 		}
+		CommandLineSwitchCase(arg, "persist-cache")
+		{
+			settings->BitmapCachePersistEnabled = enable;
+		}
+		CommandLineSwitchCase(arg, "persist-cache-file")
```
akallabeth, line 3089: For the file we might set a default in application data dir? We have `GetKnownPath` so that should be good enough to get the proper base directory on all supporte os?

include/freerdp/cache/persistent.h
```diff
@@ -0,0 +1,114 @@
...
+	UINT32 flags;
+	BYTE* data;
+};
+typedef struct _PERSISTENT_CACHE_ENTRY PERSISTENT_CACHE_ENTRY;
+
+struct rdp_persistent_cache
+{
+	FILE* fp;
+	BOOL write;
+	int version;
```
akallabeth, line 83: better use unsigned (e.g. `UINT32`) for versions

include/freerdp/cache/persistent.h
```diff
@@ -0,0 +1,114 @@
...
+	UINT64 key64;
+	UINT16 width;
+	UINT16 height;
+	UINT32 size;
+	UINT32 flags;
+	BYTE* data;
+};
+typedef struct _PERSISTENT_CACHE_ENTRY PERSISTENT_CACHE_ENTRY;
+
+struct rdp_persistent_cache
```
akallabeth, line 79: Is this struct required to be in public API header? Maybe the `opaque` version is enough here.

libfreerdp/cache/bitmap.c
```diff
@@ -278,6 +282,108 @@ void bitmap_cache_register_callbacks(rdpUpdate* update)
...
+
+	count = persistent_cache_get_count(persistent);
+
+	for (index = 0; index < count; index++)
+	{
+		if (persistent_cache_read_entry(persistent, &cacheEntry) < 1)
+			continue;
+	}
+
+	persistent_cache_free(persistent);
```
akallabeth, line 319: Maybe use a return variable, so the `success` and `error` case here can be eliminated?

libfreerdp/gdi/gfx.c
```diff
@@ -1445,9 +1446,111 @@ static UINT gdi_CacheToSurface(RdpgfxClientContext* context,
...
+	cacheEntry->scanline = (cacheEntry->width + (cacheEntry->width % 4)) * 4;
+	cacheEntry->data = (BYTE*)calloc(cacheEntry->height, cacheEntry->scanline);
+
+	if (!cacheEntry->data)
+	{
+		free(cacheEntry);
+		return ERROR_INTERNAL_ERROR;
+	}
+
+	freerdp_image_copy(cacheEntry->data, cacheEntry->format, cacheEntry->scanline, 0, 0,
```
akallabeth, line 1507: Return value checks?

libfreerdp/gdi/gfx.c
```diff
@@ -1445,9 +1446,111 @@ static UINT gdi_CacheToSurface(RdpgfxClientContext* context,
...
+	                   cacheEntry->width, cacheEntry->height, importCacheEntry->data,
+	                   PIXEL_FORMAT_BGRX32, 0, 0, 0, NULL, FREERDP_FLIP_NONE);
+
+#if 0
```
akallabeth, line 1511: Please no dead code

akallabeth (changes requested):
Overall LGTM.

Nice to have though:
* `WINPR_ASSERT` function arguments that must not be `NULL` (but do not return an error code)
* Declare variables only in the scope they are used in
* All return values should be checked whenever possible

channels/rdpgfx/client/rdpgfx_main.c
awakecoding: do we keep asserts in release builds? I see we are using asserts a lot more now, which is fine, but I'd rather keep the null checks in place to avoid null pointer exceptions, even if unlikely to happen in practice.

akallabeth: `WINPR_ASSERT`  is currently also defined for (normal) release builds, if you don´t want them (performance critical systems maybe?) you need to define `WITH_VERBOSE_WINPR_ASSERT=OFF`

awakecoding: @akallabeth I am not really concerned about the performance cost of asserts in release builds, the issue is that they abort the entire process, which is really a nasty way to report an error condition in production. I understand the asserts are supposed to check for conditions that really shouldn't happen, but sometimes they still do happen anyway.

awakecoding: It's not like you can guarantee that stuff that must not happen cannot happen, especially if we're talking about cases where someone would be deliberately sending malicious payloads in hope of triggering asserts in production, resulting in a denial-of-service attack. As for the "you're in trouble anyway", yes, but I'd rather have FreeRDP gracefully terminate the faulty connection instead of terminating the entire process, which in many cases is a lot more than a single-connection RDP client. Think about connection managers hosting multiple RDP client connections at the same time, proxy servers, and RDP servers. Aborting the entire process instead of bubbling up an improbable error condition followed by graceful termination of the faulty connection is not a good idea.

I don't want to this fight this issue, as I said, I'm not so much concerned about the performance cost of the asserts. However, if it's possible to disable the asserts, then I still want those null pointer checks to avoid null pointer exceptions. I don't mind adding asserts, but I want the code to still be properly checking null pointers if the asserts are disabled.

channels/rdpgfx/client/rdpgfx_main.c
awakecoding: I added the proper error check, I'm not sure if we should hard fail on this one

libfreerdp/cache/persistent.c
awakecoding: fixed

libfreerdp/cache/persistent.c
awakecoding: good catch, I also modified all the fread/fwrite calls in persistent.c to swap the "number" and "count" fields such that the return value is the number of bytes read/written.

channels/rdpgfx/client/rdpgfx_main.c
awakecoding: I modified rdpgfx_send_cache_offer to return an error code that can be propagated like the rest

channels/rdpgfx/client/rdpgfx_main.c
akallabeth: @awakecoding the `asserts` are meant for internal stuff (ensure function arguments are valid), of course you should not use that for checks regarding external controlled entities/data.

client/common/cmdline.c
awakecoding: That would also require creating a directory to make sure the file can be created inside of it, and which directory should we use by default? Right now you have to explicitly set the bitmap cache file path, but this is an improvement that could come later

akallabeth: Most likely a subdirectory of `GetKnownPath(KNOWN_PATH_XDG_CACHE_HOME)`, but true, can be done later on.

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -689,22 +953,31 @@ static UINT rdpgfx_recv_cache_import_reply_pdu(RDPGFX_CHANNEL_CALLBACK* callback
 	if (!Stream_CheckAndLogRequiredLength(TAG, s, 2ull * pdu.importedEntriesCount))
 		return ERROR_INVALID_DATA;
 
-	pdu.cacheSlots = (UINT16*)calloc(pdu.importedEntriesCount, sizeof(UINT16));
+	if (pdu.importedEntriesCount > RDPGFX_CACHE_ENTRY_MAX_COUNT)
+		return ERROR_INVALID_DATA;
 
-	if (!pdu.cacheSlots)
+	if (Stream_GetRemainingLength(s) / 2 < (size_t)pdu.importedEntriesCount)
```
akallabeth, line 959: Better use `Stream_CheckAndLogRequiredLength` so we are notified about protocol violations in log messages

include/freerdp/cache/bitmap.h
```diff
@@ -33,6 +33,8 @@ typedef struct
 	rdpBitmap** entries;
 } BITMAP_V2_CELL;
 
+#include <freerdp/cache/persistent.h>
```
akallabeth, line 36: Includes at start of file please

include/freerdp/cache/persistent.h
```diff
@@ -0,0 +1,103 @@
...
+	FREERDP_API int persistent_cache_read_entry(rdpPersistentCache* persistent,
+	                                            PERSISTENT_CACHE_ENTRY* entry);
+	FREERDP_API int persistent_cache_write_entry(rdpPersistentCache* persistent,
+	                                             const PERSISTENT_CACHE_ENTRY* entry);
+
+	FREERDP_API int persistent_cache_open(rdpPersistentCache* persistent, const char* filename,
+	                                      BOOL write, UINT32 version);
+	FREERDP_API int persistent_cache_close(rdpPersistentCache* persistent);
+
+	FREERDP_API rdpPersistentCache* persistent_cache_new();
```
akallabeth, line 96: should be `FREERDP_API rdpPersistentCache* persistent_cache_new(void);`

libfreerdp/cache/persistent.c
```diff
@@ -0,0 +1,344 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
```
akallabeth, line 21: Replace with ` #include <freerdp/config.h>`, the illusion of `HAVE_CONFIG_H` has been removed on master (compile does not work without the configuration header anyway)

channels/rdpgfx/client/rdpgfx_main.c
awakecoding: hum... there's already such a check a few lines above, should we just remove this check then?

akallabeth: oh, yes, true.
looks like this is a duplicate, indeed.

akallabeth (approved):
LGTM

hardening (approved):

PR #7943, opened 2022-06-01
fixed winpr_event_init for non-linux unix-like systems

sss123next:


bmiklautz:
@freerdp-bot test

hardening:
@freerdp-bot can you give a test at this ?

hardening:
The fix is that both fd gets the `O_NONBLOCK` option ? And reading the manual it says `pipe2() is Linux-specific.`, so I'm confused by your commit title...

hardening:
Looks like `FreeBSD` is not the all Unix world, and that `pipe2` doesn't exist at least on MacOS X !

sss123next:
so, what do you suggest ?, current implementation hangs on freebsd (at least)

sss123next:
`BOOL winpr_event_reset(WINPR_EVENT_IMPL* event)` always deadlocks on ret = read(event->fds[1], &value, 1);

sss123next:
also, looks like it should read from 0 not 1 ?
```
pipe() creates a pipe, a unidirectional data channel that can be used
       for interprocess communication.  The array pipefd is used to return two
       file descriptors referring to the ends of the pipe.  pipefd[0] refers
       to the read end of the pipe.  pipefd[1] refers to the write end of the
       pipe.  Data written to the write end of the pipe is buffered by the
       kernel until it is read from the read end of the pipe.  For further
       details, see pipe(7).
```

sss123next:
it's already fixed in upstream, sorry to trouble.

PR #7941, opened 2022-06-01
Add WTSChannelGetOptions (for retrieving peer channel option flags)

skoolkid:
Along with the existing `WTSChannelGetName` function, `WTSChannelGetOptions` enables the names and option flags of a peer's connected channels to be retrieved. This is useful for populating a client's channel definitions before it connects, for example in an RDP proxy where all channel data is simply passed through.

Backport of #7945

PR #7945, opened 2022-06-02
Add WTSChannelGetOptions

skoolkid:
Along with the existing `WTSChannelGetName` function, `WTSChannelGetOptions` enables the names and option flags of a peer's connected channels to be retrieved. This is useful for populating a client's channel definitions before it connects, for example in an RDP proxy where all channel data is simply passed through.

Could this change also be backported to `stable-2.0`? (See #7941 for reference.)


akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #7955, opened 2022-06-07
Fix logic inversion introduced by scanbuild fixes

ManMower:
commit ec699f6c75b179409221cbc215a97eaa930bf2a1 seems to have
accidentally inverted this check.

### DIFF
--- a/libfreerdp/core/activation.c
+++ b/libfreerdp/core/activation.c
@@ -572,17 +572,11 @@ BOOL rdp_server_accept_client_font_list_pdu(rdpRdp* rdp, wStream* s)
 	    peer->AdjustMonitorsLayout(peer))
 	{
 		/* client supports the monitorLayout PDU, let's send him the monitors if any */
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth (changes requested):
Looks like you try to hide a memory leak ;)
There is an issue, but it is that the place it is called the argument is a already allocated buffer

ManMower:
Thanks for taking  a look.

Sorry, I've re-read the "scanbuild fixes" change more closely, I hadn't realize it moved the responsibility for the array allocation into the callee without updating the caller. The new changes should address that.

akallabeth:
@ManMower looks like I just missed the caller, the function was not properly defined before that.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #7960, opened 2022-06-09
proxy: add fixed domain/passwd

hardening:
This adds a `Domain` and `Password` parameter in the `Target` section of the configuration to specify and use a fixed backend domain or password.


akallabeth (changes requested):
Are you not missing a fixed user or what should these options be used for?

hardening:
@akallabeth good idea for the fixed user, pushing that.

hardening:
@akallabeth I've added the fixed user

akallabeth:
@hardening ok, so, this just authenticates against the destination target, right? Which credentials are checked from the source?

akallabeth (approved):
Ok, looks good.

PR #7966, opened 2022-06-13
client channels: mutualize as much as possible dynamic channels

hardening:
Most dynamic channels share the same code copied and pasted code. This PR tries to mutualize as much as possible these parts (types and code).


akallabeth (commented):
Overall LGTM and a nice idea.
Not sure about removing the specific typedef though.

akallabeth (approved):

PR #7970, opened 2022-06-15
proxy: correctly handle dynamic channel fragmentation

hardening:
This big patch fixes fragmentation handling in the dynamic channel. We used to have a single state to handle fragmentation at the main dynamic channel level, but in fact packets can be fragmented per sub channel. So we have to maintain a fragmentation
state per sub channel, this involve treating dynamic and static channels differently (so the size of the patch that has to implement state tracking per dynamic channels).


akallabeth (approved):
LGTM

hardening:
Just fixed a leak.

PR #7987, opened 2022-06-23
Add a new command line arg to enforce tls1.2

sivagudivadaz:
I'm facing a TLS_CONNECT_ERROR issue where a customer has both TLS1.3 and TLS1.2 enabled on their windows 2019 server - which I believe has experimental support for TLS1.3. And because the server doesn't have full TLS1.3 support, calling TLS_client_method in tls.c - which negotiates the highest available TLS version, fails to establish a proper connection as the server sends back a RST after ClientHello.

Note that this is not an issue with Windows 2022 server - which has full support for TLS1.3. With both TLS1.2 and TLS1.3 enabled on a win 2022 server, FreeRDP can connect just fine to the target VM.

This PR adds a new command line argument to enforce TLS1.2 that is set through our application that consumes FreeRDP.

Please advice if there is a less intrusive way to enforce TLS1.2 during the initial TLS handshake.


hardening:
For the changes in settings.h, just change the `rdp_settings` struct and then run the `update-rdpSettings` and `update-settings-tests` scripts in `tools`, it will re-generate setters/getters and the ids.

libfreerdp/crypto/tls.c
```diff
@@ -872,6 +872,18 @@ int tls_connect(rdpTls* tls, BIO* underlying)
 	return tls_do_handshake(tls, TRUE);
 }
 
+BOOL tls_prep(rdpTls* tls,BIO* underlying, int options, BOOL clientMode)
```
hardening, line 875: This function shouldn't exist, you should do the method selection stuff in `tls_connect()`. If we have this `EnforceTLSv1_2` option it should probably also apply for server-side connections

akallabeth (approved):
LGTM,
except running 
```
./tools/update-rdpSettings
./tools/update-rdpSettings
git clang-format -f
git commit . -m"updated rdpSettings for tls1.2"
```

(and maybe use the `freerdp_settings_get_bool` and `freerdp_settings_set_bool` instead of direct access to the `rdpSettings` variable

libfreerdp/crypto/tls.c
akallabeth: @hardening as far as I see it this is just a `hack` to fix broken behavior of windows servers?

sivagudivadaz: @hardening  
> If we have this EnforceTLSv1_2 option it should probably also apply for server-side connections
Not sure if you are suggesting a code change here in FreeRDP but the target server I'm trying to connect to already has TLS1.2 enabled. The issue is very specific to the win 2019 server and only occurs exclusively when both TLS1.2 and TLS1.3 are enabled - almost as if this machine is misleading FreeRDP during the handshake and then resulting in a `RST` response. 

FWIW, both win2019 and win2022 server machines can't be accessed even from the official Microsoft RDP client when TLS1.3 is the only TLS version enabled - maybe RDP can't solely function on TLS1.3 ? idk 🤷‍♂️ at least we are on parity with the official RDP client in this regard :)

akallabeth:
@sivagudivadaz ok, I´ve done the changes required, I´m force pushing to your branch, just you know ;)

sivagudivadaz:
> @sivagudivadaz ok, I´ve done the changes required, I´m force pushing to your branch, just you know ;)

Oh ok 👍, I was just about to make those changes and push too, but if you are ready already go ahead. 

Sorry about missing the clang formatting though, will keep that in mind next time.

akallabeth:
@freerdp-bot test

akallabeth:
@sivagudivadaz thank you for investigating (and fixing) this.

PR #7997, opened 2022-06-24
Winpr asserts, streampool fixes, authentication leak fixes

akallabeth:


akallabeth:
@hardening do you agree to this pr or should I split the `rdpdr` `StreamPool` stuff from this into a separate PR?

hardening (approved):

PR #7998, opened 2022-06-27
Gfx refactor / assert

akallabeth:


channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -159,7 +164,9 @@ static UINT rdpgfx_send_caps_advertise_pdu(RdpgfxClientContext* context,
 
 static BOOL rdpgfx_is_capability_filtered(RDPGFX_PLUGIN* gfx, UINT32 caps)
 {
-	const UINT32 filter = gfx->capsFilter;
+	WINPR_ASSERT(gfx);
+	const UINT32 filter =
+	    freerdp_settings_get_uint32(gfx->rdpcontext->settings, FreeRDP_GfxCapsFilter);
```
hardening, line 169: Don't we have a `gfx->settings` field shortcut that we could use here and below ?

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -491,7 +503,10 @@ static UINT rdpgfx_recv_reset_graphics_pdu(GENERIC_CHANNEL_CALLBACK* callback, w
 	UINT32 index;
 	MONITOR_DEF* monitor;
 	RDPGFX_RESET_GRAPHICS_PDU pdu = { 0 };
+	WINPR_ASSERT(callback);
+	;
```
hardening, line 507: shall drop this line

channels/rdpgfx/client/rdpgfx_main.c
akallabeth: @hardening removed that here, as I try to unify (and eliminate) the duplicated fields (makes changes to `rdpContext` very error prone if we store the pointers everywhere but never update them)

channels/rdpgfx/client/rdpgfx_main.c
akallabeth: good catch

hardening (approved):
LGTM.

hardening:
The egfx probably needs 2 other cleanups (at least):

* the surface data hashtable indexed by surfaceId has the drawback that we must do the id+1 trick to correctly register data for surface id 0. Possible solution: index with a UINT32* (instead of UINT32), update hashtable functions accordingly and always have first field of opaque surfaceData being the surfaceId (perhaps we can change the `void*` in `[Set|Get]SurfaceData` to a `COMMON_SURFACE_DATA*` that would have this);
* `EGFX_PLUGIN` could have a direct pointer to the context instead of passing through the opaque (and error prone) `iface` field (BTW IMHO that should be the case of all channels)

akallabeth:
@hardening yes, most likely. sounds like something for your unify pr?

PR #7999, opened 2022-06-27
Fix building with LibreSSL 2.7.0 or newer

aperezdc:
With LibreSSL 2.7.0 (or newer versions) some more structs have made opaque, which requires a few changes:

- `BIO_meth_new()` and related functions are now defined by LibreSSL, the versions from `opensslcompat.{h,c}` does not need to be used anymore.
- `HMAC_CTX` is now opaque, `HMAC_CTX_new()`, `EVP_MD_CTX_new()`, and related functions should be used instead in winpr's `hash.c`.

### DIFF
--- a/libfreerdp/crypto/opensslcompat.c
+++ b/libfreerdp/crypto/opensslcompat.c
@@ -19,7 +19,8 @@
 
 #include "opensslcompat.h"
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
+#if OPENSSL_VERSION_NUMBER < 0x10100000L || \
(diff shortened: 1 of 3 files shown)


aperezdc:
I have successfully built and tested Weston's RDP support in a RISC-V image made with Buildroot using this patch. I'll be happy to share the build configuration and any other information that may be helpful for reviewing this fix 😄

akallabeth (approved):

Zirias:
I don't think this is fully correct. Trying to run FreeRDP 2.8.0 on FreeBSD 13.1 with LibreSSL 3.4.3, I got this error:

    ld-elf.so.1: /usr/local/lib/libfreerdp2.so.2: Undefined symbol "BIO_get_init"

I could "fix" it by changing the version check in `opensslcompat.h` to `LIBRESSL_VERSION_NUMBER < 0x3050000fL` (with 3.5, the opaquing is AFAIK complete). I'm not sure whether this is fully correct either....

PR #8000, opened 2022-06-27
Redirect channel fix

akallabeth:
* Fixes a client problem on redirect (reset `rdpSettings` on redirect to allow new negotiation of channels/settings/...)
* Fixes a regression (`Stream_New` instead of `StreamPool_Take` in `rdpdr`)
* Fixes a regression (missing `LoadChannels` callback)

### DIFF
--- a/libfreerdp/core/client.c
+++ b/libfreerdp/core/client.c
@@ -304,7 +304,8 @@ UINT freerdp_channels_attach(freerdp* instance)
 
 	channels = instance->context->channels;
 	hostname = freerdp_settings_get_string(instance->context->settings, FreeRDP_ServerHostname);
-	hostnameLength = strlen(hostname);
+	WINPR_ASSERT(hostname);
+	hostnameLength = strnlen(hostname, MAX_PATH);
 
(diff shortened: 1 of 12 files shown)


libfreerdp/core/client.c
```diff
@@ -303,7 +303,7 @@ UINT freerdp_channels_attach(freerdp* instance)
 	WINPR_ASSERT(instance->context->settings);
 
 	channels = instance->context->channels;
-	hostname = instance->context->settings->ServerHostname;
+	hostname = freerdp_settings_get_string(instance->context->settings, FreeRDP_ServerHostname);
 	hostnameLength = strlen(hostname);
```
hardening, line 307: Hum perhaps we can add a WINPR_ASSERT here or `strlen`  will segfault or replace the call with `_strlen`

akallabeth: @hardening true, added a `WINPR_ASSERT` and limit the hostname length to `MAX_PATH` (`URL` are less and file paths most likely too)

akallabeth:
@freerdp-bot test

PR #8001, opened 2022-06-27
Fix building with LibreSSL 2.7.0 or newer

aperezdc:
With LibreSSL 2.7.0 (or newer versions) some more structs have made opaque, which requires a few changes:

- `BIO_meth_new()` and related functions are now defined by LibreSSL, the versions from `opensslcompat.{h,c}` does not need to be used anymore.
- `HMAC_CTX` is now opaque, `HMAC_CTX_new()`, `EVP_MD_CTX_new()`, and related functions should be used instead in winpr's `hash.c`.

----

This is a backport to the `stable-2.0` branch of #7999

### DIFF
--- a/libfreerdp/crypto/opensslcompat.c
+++ b/libfreerdp/crypto/opensslcompat.c
@@ -19,7 +19,8 @@
 
 #include "opensslcompat.h"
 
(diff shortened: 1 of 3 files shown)


akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #8019, opened 2022-06-30
Fixed copyright headers and stream length

akallabeth:


include/freerdp/channels/rdpecam.h
```diff
@@ -1,9 +1,8 @@
 /**
  * FreeRDP: A Remote Desktop Protocol Implementation
- * Video Optimized Remoting Virtual Channel Extension
+ * Camera Redirection Virtual Channel Extension
```
pnowack, line 3: Just a nit, feel free to ignore: While camera redirection is the accurate name of that extension, I would though go with `Video Capture Virtual Channel Extension`, as that is the name of the spec.

pnowack:
Just noticing one more thing: The `Stream_SetLength(s, BytesReturned);` call in the enumerator source needs to be moved too.

pnowack:
LGTM

PR #8024, opened 2022-07-01
Clear openssl error queue after connecting with nla

fifthdegree:
This is necessary when using pkcs11 due to a bug in openssl where errors emitted by unrelated operations can cause BIO_should_retry() to report false negatives (see [Issue #11889](https://github.com/openssl/openssl/issues/11889))

### DIFF
--- a/client/iOS/Additions/TSXAdditions.m
+++ b/client/iOS/Additions/TSXAdditions.m
@@ -220,6 +220,7 @@
 	BIO_set_flags(context, BIO_FLAGS_BASE64_NO_NL);
 
 	// Encode all the data
+	ERR_clear_error();
 	BIO_write(context, [self bytes], [self length]);
 	(void)BIO_flush(context);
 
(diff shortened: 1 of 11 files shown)


akallabeth:
@freerdp-bot test

hardening:
Just wondering if the correct fix would not be to do a `ERR_clear_error()` before all `SSL_Read` / `SSL_Write` calls ? That's what suggest the openSSL post...

akallabeth:
@fifthdegree could you run `git clang-format` on your commit? looks like formatting is not in required style.
as for what @hardening said, that might actually be a good idea

fifthdegree:
I think I got all of them; let me know if I missed anything

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@fifthdegree you did miss a few, but I´ve pushed these changes too.
left out the ones in a `BIO` method as they call recursivly

akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #8028, opened 2022-07-01
Rem malloc

akallabeth:


include/freerdp/channels/cliprdr.h
```diff
@@ -32,26 +32,32 @@
...
+typedef enum
+{
+	CB_FORMAT_HTML = 0xD010,
+	CB_FORMAT_PNG = 0xD011,
+	CB_FORMAT_JPEG = 0xD012,
+	CB_FORMAT_GIF = 0xD013,
+	CB_FORMAT_TEXTURILIST = 0xD014,
+	CB_FORMAT_GNOMECOPIEDFILES = 0xD015,
+	CB_FORMAT_MATECOPIEDFILES = 0xD016
+} CliprdrFormatType;
```
pnowack, line 44: I am not so sure about turning this one into an enum. The thing is that client and server implementations may not (want to) support all kinds of clipboard formats and in addition to that some formats here are dynamic.

For the former argument here, implementations would have to use only their supported formats in a switch statement, in addition to a default path, because without the default path, the compiler will warn about unhandled enum values (which is usually the point of an enum in a switch statement).

For the latter one here (dynamic formats), I am not so sure someone would want to write `CliprdrFormatType type = my_own_clipboard_format`.
Dynamic formats are identified via its name and have a variable id and thus values may clash.

As a result of this, this `CliprdrFormatType` enum would mostly not be a subset or superset of the actual implemented clipboard formats by the client or server implementation.

akallabeth: @pnowack only if the type we switch on is of the enum, not if the enum is only used for checking no duplicates exist.

pnowack: Indeed, though at least I prefer the actual enum type here for switches (which is why I'm not so fond of the enum in this situation here).

PR #8037, opened 2022-07-05
Remove xmlto

StephDC:
As the use of xmlto to generate manpage is replaced by xsltproc in https://github.com/FreeRDP/FreeRDP/commit/9b95c555742a70b9e81be4ee5622e6d743403972 , I believe it would be better to remove xmlto related codes, so
unnecessary dependencies would not be installed for building this package.

This shall not introduce any issues as no other code is using xmlto right now.


akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #8041, opened 2022-07-05
Removed ALAW/ULAW from supported backend formats

akallabeth:
pulse/alsa/oss do not reliably work with ALAW/ULAW codecs with all
hardware available.
Since these formats are poor quality anyway, deactivate them for
audin channel.


hardening:
LGTM.
Looking at the changes, perhaps we can add ALAW/ULAW with `default:` statements in switch/case and add a comment saying that we don't want support these (could help in the future if people want to contribute back support for alaw/ulaw).

hardening (approved):

PR #8046, opened 2022-07-06
Proxy fixes

akallabeth:


hardening:
Shall we document somewhere the removal of `RdpKeyContent` and `RdpKeyFile` so that migration to FreeRDP3 is made easier ? (these fields are used by many project that are FreeRDP2 based)
Perhaps we could start a Wiki page..

akallabeth:
@hardening yes, we should really start documenting `FreeRDP 3` somewhere.

hardening:
I'm wondering about the removal of `RdpKeyContent` and `RdpKeyFile`, IIRC on the RDP side there was some limitation on the crypto strength (limited to 1024 bits ?) that could be used. So if we have only one common crypto source (for TLS and RDP), then we may introduce some incompatibility with old client, or be forced to use weak crypto (and so possibly be incompatible with modern clients).

hardening:
@akallabeth FYI I've created a page here: https://github.com/FreeRDP/FreeRDP/wiki/FreeRDP3-migration-notes

hardening:
> > I'm wondering about the removal of `RdpKeyContent` and `RdpKeyFile`, IIRC on the RDP side there was some limitation on the crypto strength (limited to 1024 bits ?) that could be used. So if we have only one common crypto source (for TLS and RDP), then we may introduce some incompatibility with old client, or be forced to use weak crypto (and so possibly be incompatible with modern clients).
> 
> that is true for all certificates with certain RDP servers (2048 RSA key size limit until recently if I´m not mistaken)

Yes but then we loose some compatibility, because with `RdpKeyFile` you can give a key that will be used for RDP security, and `PrivateKeyFile` is used for TLS. So we must really check that the limit was not only on the RDP security part but was also applying on TLS security.

PR #8048, opened 2022-07-06
TLS version control

DVeron-RC:
Added settings to control the bounds of the TLS versions allowed.

This removes the setting "enforce_TLSv1.2",
but the behaviour of the corresponding CLI argument is unchanged.

There is no CLI argument to control the bounds.
If needed one can be added.

To test it, you can client side inspect the packets with wireshark,
and server side you can use a tool like sslscan.


hardening:

hardening:
Promoting my own agenda, would be so nice to also have a command-line option to force weak crypto (TLS1.0 without PFS) so that you can analyze wireshark captures with the server's private key (with last wireshark that can decode most of the RDP traffic)...

hardening:
@freerdp-bot can you test again, looks like the MacOsX build was at fault

akallabeth (approved):
LGTM

PR #8053, opened 2022-07-07
gateway: Base-64 encode websocket key in request header

mfleisz:
According to the RFC the websocket key in the request header should be base-64 encoded:

> The request MUST include a header field with the name |Sec-WebSocket-Key|. The value of this header field MUST be a nonce consisting of a randomly selected 16-byte value that has been base64-encoded (see Section 4 of [RFC4648]). The nonce MUST be selected randomly for each connection.

If we just send a random key this might cause problems with gateways/proxies that try to decode the key, resulting in an error (i.e. HAProxy returns 400 Bad Request).


MartinHaimberger (approved):
LGTM

hardening (approved):

PR #8055, opened 2022-07-07
[stable-2.0] Relieve CLIPRDR filename restriction when possible backport

pnowack:
Backport of https://github.com/FreeRDP/FreeRDP/pull/8051


pnowack:
> this does not seem to be based on current stable-2.0 branch

It is though, rebased the patches on top of stable-2.0 and tested them again (relieving the restriction only when the server is not on MS Windows platform). No sure what you mean or what I missed.

akallabeth (approved):
LGTM
sorry, the changlog looked broken (missing entries)

PR #8059, opened 2022-07-08
Drdynvc needs love

hardening:
Some cleanups and rework of the client dynamic channel.

### DIFF
--- a/libfreerdp/core/info.c
+++ b/libfreerdp/core/info.c
@@ -1098,13 +1098,6 @@ static BOOL rdp_recv_logon_info_v2(rdpRdp* rdp, wStream* s, logon_info* info)
 			          logonInfoV2TotalSize, Size);
 			return FALSE;
 		}
-		else
-		{
-			WLog_WARN(TAG,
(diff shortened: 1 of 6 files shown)


libfreerdp/core/info.c
```diff
@@ -1098,13 +1098,6 @@ static BOOL rdp_recv_logon_info_v2(rdpRdp* rdp, wStream* s, logon_info* info)
 			          logonInfoV2TotalSize, Size);
 			return FALSE;
 		}
-		else
```
akallabeth, line 1101: protocol violation warning, please do not remove. reducing priority if you must.

akallabeth (approved):
LGTM

libfreerdp/core/info.c
hardening: Well as it happens all the time with almost all my windows servers, it's not that important, we can safely ignore that protocol trick (and that's exactly what we do).
And moreover the message is kinda misleading, it gives the impression that we've dropped the logonInfoV2 packet while we treat it anyway.

mfleisz (approved):

PR #8074, opened 2022-07-14
Switch to using the WinPR ASN1 library in nla.c

fifthdegree:


mfleisz:
@freerdp-bot test

mfleisz:
@freerdp-bot test

libfreerdp/core/nla.c
```diff
@@ -2049,91 +1968,142 @@ static BOOL nla_read_ts_credentials(rdpNla* nla, SecBuffer* data, size_t offset)
...
+		if (!WinPrAsn1EncSeqContainer(enc))
+			goto out;
 
-		smartcardCreds.pin = Password ? Password : "";
+		/* pin [0] OCTET STRING */
+		str = freerdp_settings_get_string_writable(settings, FreeRDP_Password);
+		octet_string.len =
+		    ConvertToUnicode(CP_UTF8, 0, str, -1, (LPWSTR*)&octet_string.data, 0) * sizeof(WCHAR);
+		ret = WinPrAsn1EncContextualOctetString(enc, 0, &octet_string);
+		if (octet_string.data)
```
mfleisz, line 2024: Don't check for NULL before `free`

libfreerdp/core/nla.c
```diff
@@ -2049,91 +1968,142 @@ static BOOL nla_read_ts_credentials(rdpNla* nla, SecBuffer* data, size_t offset)
...
-		if (!credsContentStream)
-			return FALSE;
+		for (int i = 0; i < ARRAYSIZE(cspData_fields); i++)
+		{
+			str = freerdp_settings_get_string_writable(settings, cspData_fields[i].setting_id);
+			octet_string.len =
+			    ConvertToUnicode(CP_UTF8, 0, str, -1, (LPWSTR*)&octet_string.data, 0) *
+			    sizeof(WCHAR);
+			ret = WinPrAsn1EncContextualOctetString(enc, cspData_fields[i].tag, &octet_string);
+			if (octet_string.len)
```
mfleisz, line 2045: Don't check for NULL before `free`

libfreerdp/core/nla.c
```diff
@@ -2049,91 +1968,142 @@ static BOOL nla_read_ts_credentials(rdpNla* nla, SecBuffer* data, size_t offset)
...
+				goto out;
+		}
 
-		if (ber_write_nla_TSSmartCardCreds(credsContentStream, &smartcardCreds) == 0)
-			return FALSE;
+		if (!WinPrAsn1EncEndContainer(enc))
+			goto out;
 
-		cr.credType = 2;
+		if (!WinPrAsn1EncEndContainer(enc))
```
mfleisz, line 2054: Would be nice to have some comment explaining why this is necessary twice (looks suspicious on first sight)

libfreerdp/core/nla.c
```diff
@@ -2049,91 +1968,142 @@ static BOOL nla_read_ts_credentials(rdpNla* nla, SecBuffer* data, size_t offset)
...
-			passCreds.password = (BYTE*)nla->identity->Password;
+			password.len = nla->identity->PasswordLength * 2;
+			password.data = (BYTE*)nla->identity->Password;
 		}
 
-		length = ber_sizeof_nla_TSPasswordCreds(&passCreds);
-		credsContentStream = Stream_New(NULL, length);
-		if (!credsContentStream)
+		if (!WinPrAsn1EncContextualOctetString(enc, 0, &domain))
+			return FALSE;
```
mfleisz, line 2079: These returns will all leak `enc`

winpr/libwinpr/utils/asn1/asn1.c
```diff
@@ -1377,6 +1377,32 @@ size_t WinPrAsn1DecReadContextualOID(WinPrAsn1Decoder* dec, WinPrAsn1_tagId tagI
...
+{
+	size_t ret, ret2;
+	WinPrAsn1_tag ftag;
+	WinPrAsn1Decoder content;
+
+	WINPR_ASSERT(error);
+
+	ret = WinPrAsn1DecPeekContextualTag(dec, &ftag, &content);
+	if (!ret || ftag != tagId)
+		return 0;
```
hardening, line 1392: I think the bug is present in other `WinPrAsn1DecReadContextualXXXXX` functions: `*error` is not set in that case.

hardening: Would be nice to complement asn1 unitary tests to check for that case.

libfreerdp/core/nla.c
hardening: Yeah could be nice to have in comment which container is closed (just an idea).

libfreerdp/core/nla.c
```diff
@@ -1936,110 +1899,66 @@ SECURITY_STATUS nla_decrypt_public_key_hash(rdpNla* nla)
...
-
-	/* The sequence is empty, return early,
-	 * TSPasswordCreds (SEQUENCE) is optional. */
-	if (length == 0)
-		return TRUE;
-
-	/* [0] domainName (OCTET STRING) */
-	if (!ber_read_contextual_tag(s, 0, &length, TRUE) || !ber_read_octet_string_tag(s, &length))
-	{
+	if (!data || data->cbBuffer < offset)
```
hardening, line 1910: Hum, is it asserts here ? I guess that if we receive such input it's a problem from the caller or I'm missing something ?

libfreerdp/core/nla.c
```diff
@@ -1936,110 +1899,66 @@ SECURITY_STATUS nla_decrypt_public_key_hash(rdpNla* nla)
...
-		return FALSE;
+	WinPrAsn1Decoder_InitMem(&dec, WINPR_ASN1_DER, credentials.data, credentials.len);
 
-	passwordLen = length / sizeof(WCHAR);
-	if (length > 0)
-		password = Stream_PointerAs(s, const WCHAR);
-
-	if (!Stream_SafeSeek(s, length))
-		return FALSE;
+	if (credType == 1)
```
hardening, line 1931: If we receive some `TSSmartCardCreds` or some `TSRemoteGuardCreds` we return TRUE even if we've read nothing, perhaps you can change the `if (credType==1) {` by a `switch(credType) { case 1: <code>... }` and have a default statement that returns FALSE until we have implemented the other kind of credentials

hardening:
General remark: `tscredentials.[ch]` were the only users of the `tools/asn_parser_generator.py` script, so perhaps that if we remove these files, the script shall be removed too (or at least updated to use WinPRAsn1). This can also be part of another PR, when we'll have removed all instance of BER/DER functions...
`libfreerdp/core/credssp.asn1` shall be removed too as it was the source to generate `tscredentials.[ch]`.

libfreerdp/core/nla.c
fifthdegree: I don't know, it was that way before but I agree that `data` should probably be asserted; I guess `data->cbBuffer < offset` shouldn't either be able to happen because decryption would fail

libfreerdp/core/nla.c
fifthdegree: True; they should have been `goto out`

libfreerdp/core/nla.c
hardening: nla has no "auto-login", either the transport is established (and authenticated), or it fails (no login greeter in NLA).

libfreerdp/core/nla.c
fifthdegree: Yes but `octet_string.data` wouldn't be NULL. If `ConvertToUnicode` didn't do the conversion for whatever reason, it would be the previously freed pointer from the last conversion.

mfleisz: Actually `octet_string.data` should be initialized with `NULL` before the `ConvertToUnicode` call and then just `free`d.

You could just do a `WinPrAsn1_OctetString octet_string = { 0 };` to have the struct correctly initialized.

libfreerdp/core/nla.c
hardening: Hum you're perfectly correct.

winpr/libwinpr/utils/asn1/asn1.c
fifthdegree: Do you want to do that in a separate PR?

hardening: I will

hardening:
@freerdp-bot can you test the new version please ?

hardening:
I've complemented the checks for ContextualReads in #8085

hardening:

hardening (approved):

PR #8078, opened 2022-07-16
 RAIL Fixed the defect that the window will shrink abnormally when moving

vesystem2022:
RAIL Fixed the defect that the window will shrink abnormally when moving
RemoteApp model  Defects of window movement

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,3 +154,6 @@ packaging/deb/freerdp-nightly/freerdp-nightly-dbg
 # VisualStudio Code
 .vscode
 cache/
+
+#build
+build
--- a/.gitignore
+++ b/.gitignore
@@ -157,3 +157,5 @@ cache/
 
 #build
 build
+cmake-build-debug
+
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -499,6 +499,12 @@ static BOOL xf_rail_window_common(rdpContext* context, const WINDOW_ORDER_INFO* 
 		    (appWindow->visibleOffsetY -
 		     (appWindow->clientOffsetY - appWindow->windowClientDeltaY));
 
+		if( appWindow->windowOffsetX == 0  && appWindow->windowOffsetY == 0 )
+		{         
+		    appWindow->windowOffsetX = appWindow->x;
+		    appWindow->windowOffsetY = appWindow->y;
+		}
+		
 		/*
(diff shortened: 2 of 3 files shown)


vesystem2022 (commented):
2022-07-16 16-06-19 的屏幕截图


TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE
这个属性并不被 Win7 以上有操作系统良好支持，会导致以 0000000 模式连接 WIIN8 /WIN10 系统，应用程序窗口出现画面缺陷 和 窗口失控漂动
TS_ RAIL_ CLIENTSTATUS_ ALLOWLOCALMOVESIZE This attribute is not well supported by the operating system above win7, which will lead to the connection of wiin8 /win10 system in RAIL mode, and the application window will have picture defects and out of control drift

vesystem2022:

Yes, thank you very much

mfleisz (changes requested):
Removing `TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE` support from the channel is not the right way to fix this. Something must be wrong with the way `xf_rail_server_local_move_size` is handling things.

vesystem2022:
> Removing `TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE` support from the channel is not the right way to fix this. Something must be wrong with the way `xf_rail_server_local_move_size` is handling things.

Removing the attribute TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE is the simplest and effective method. It has been verified on multiple operating systems, and can solve various defects in the display and movement of the application window in RAIL mode. This defect has existed for a long time. In RAIL mode, the application window will appear a splash screen when it is superimposed, and the screen will appear fake death when it is moved, Especially when Linux is connected to RemoteAPP of win8 / win8.1 / win10, this kind of phenomenon is very prominent.

The only problem is that when dragging the window, only the frame of the window is displayed, and the content of the window is not displayed. This is also a normal display method on windows, which saves the consumption of real-time painting.

mfleisz:
While this might looks like the simplest way to solve this issue for your use case we have to be sure this won't affect clients on other platforms in combination with other servers that you might have not tested. Additionally, with your change you might affect other, 3rd party clients that use our library and rely on this callback to be called.

Instead of removing the flag/callback completely, rather have a look at how to fix the callback in xfreerdp and have the least impact on the library.

Issue #7592, opened 2022-01-26
[Windows] Microphone with audio format 0x0001 (WAVE_FORMAT_PCM) broken on 2.5.0


vscfreire:
FreeRDP 2.5.0 includes some fixes to the "audin" (microphone) channel code for Windows (on the [audin_winmm.c](https://github.com/FreeRDP/FreeRDP/blob/stable-2.0/channels/audin/client/winmm/audin_winmm.c) file, commit [d4f5003bc922bcd0e53e8abcfaee6eb7ad70e2fb](https://github.com/FreeRDP/FreeRDP/commit/d4f5003bc922bcd0e53e8abcfaee6eb7ad70e2fb)).

These changes cause the `wfreerdp.exe` client to crash when the `/microphone:format:0x0001` flag is used (corresponding to the `WAVE_FORMAT_PCM` format) and the microphone is used remotely. The connection establishes correctly and it only crashes the client on microphone access. It crashes with Windows Server 2016 and Linux/xrdp servers.

When no format parameter is passed (just `/microphone`), or when using the `0x0006` (`WAVE_FORMAT_ALAW`) or `0xA106` (`WAVE_FORMAT_AAC`), microphone works correctly. Tested with other formats, and the microphone doesn't work, but the client doesn't crash.

Reverting the [d4f5003bc922bcd0e53e8abcfaee6eb7ad70e2fb](https://github.com/FreeRDP/FreeRDP/commit/d4f5003bc922bcd0e53e8abcfaee6eb7ad70e2fb) commit fixes the problem, and the microphone works again with `0x0001`/`WAVE_FORMAT_PCM`.

This format is very important, because it's the only format supported by `xrdp` (ie. when connecting to Linux servers). Using any other format, the microphone doesn't work.

Command line: `wfreerdp.exe /v:XXXXXXXX /u:YYYYYYYY /cert-tofu /microphone:format:0x0001`
Buildconfig:
```
This is FreeRDP version 2.5.0 (n/a)
[...]

PR #8080, opened 2022-07-18
audin (client): Do not bother the dsp if the format is supported

Pollux42:
This fix an issue that only happen in very few cases:
* FFMPEG is enabled
* the chosen codec is PCM

But in this case (that should work), freerdp_dsp_context_reset is called, fails, resulting in no sound.

### DIFF
--- a/channels/audin/client/audin_main.c
+++ b/channels/audin/client/audin_main.c
@@ -438,8 +438,11 @@ static BOOL audin_open_device(AUDIN_PLUGIN* audin, AUDIN_CHANNEL_CALLBACK* callb
 		return FALSE;
 	}
 
-	if (!freerdp_dsp_context_reset(audin->dsp_context, audin->format, audin->FramesPerPacket))
-		return FALSE;
+	if (!supported)
+	{
+		if (!freerdp_dsp_context_reset(audin->dsp_context, audin->format, audin->FramesPerPacket))
+			return FALSE;
+	}
 
 	IFCALLRET(audin->device->Open, error, audin->device, audin_receive_wave_data, callback);
 


mfleisz:
This reverts the changes from https://github.com/FreeRDP/FreeRDP/pull/7342

Unfortunately there is no description given why this change was done back then ...

@Pollux42 can you determine where in `ffmpeg_open_context` the failure happens in your case ?

Pollux42:
@mfleisz It fail at `context->codec = avcodec_find_encoder(context->id)`, `context->id` being `AV_CODEC_ID_PCM_U16LE`

Anyway, in further processing (ie in audin_receive_wave_data), we don't call any freerdp_dsp_* functions so I think we should not need to call `freerdp_dsp_context_reset` at this point.
Besides, I mimicked what was already done in rdpsnd, so I was pretty confident...
Seems to be also linked to #7592, but I don't get why the original diff fixes an issue...

mfleisz:
@Pollux42 thanks a lot for the additional info! 
I'll loop in @akallabeth to have a look at this since he's done the last change (removing the `if`) and maybe knows what was the reason for the initial change.

akallabeth:
@freerdp-bot test

akallabeth:
@Pollux42 sorry for the late loop in, but I can´t reproduce your issue and not resetting the dsp might lead to problems if the sample rate needs adjustment.
What I´ve tried:

* debian 12 and Fedora 38
* FFMPEG as audio backend
* `xfreerdp /sound /mic:format:1 ...` 

everything works normally.
so, I´m closing this for now, if you have any objections, just ping and we can repopen/create a new fix

PR #8082, opened 2022-07-19
Fixed X11 RAIL the defect that the window will shrink abnormally when moving

vesystem2022:
Fixed X11 RAIL the defect that the window will shrink abnormally when moving

Before repair
https://user-images.githubusercontent.com/109400746/179733947-6bb8c32f-4c5d-45a7-bd4b-d0e3c20920e9.mp4

After repair
https://user-images.githubusercontent.com/109400746/179734006-bc51e33b-f85d-481f-a0ad-5dc62e45fbcd.mp4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -154,3 +154,6 @@ packaging/deb/freerdp-nightly/freerdp-nightly-dbg
 # VisualStudio Code
 .vscode
 cache/
+
+#build
+build
--- a/.gitignore
+++ b/.gitignore
@@ -157,3 +157,5 @@ cache/
 
 #build
 build
+cmake-build-debug
+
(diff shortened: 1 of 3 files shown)


vesystem2022:

OK

akallabeth (changes requested):
* Removing `TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE` is not the correct approach, you just create different problems (e.g. out of sync windows if a local move was done)
* Copying the whole code of `client_rail_server_start_cmd` is also not recommended.

akallabeth:
fixed with `HiDefRail` by RDP-spec

PR #8086, opened 2022-07-21
Enumerate keys in PKCS11 by finding public keys instead of private keys

fifthdegree:
This is important because according to the spec (see [Sec. 5.7.7](https://docs.oasis-open.org/pkcs11/pkcs11-spec/v3.1/csd01/pkcs11-spec-v3.1-csd01.html#_Toc98177135)) private keys (or any object) that are PIN protected (e.g PIV slot 9a) will explicitly not be found with `C_FindObjects`


hardening (approved):
LGTM.
Tested and works here.

hardening:
@freerdp-bot cna you test please.

Issue #7965, opened 2022-06-12
Long smartcard reader names get truncated


fifthdegree:
Since cryptoki defines `slotDescription` (which is used to retrieve the reader name) as a 64 byte buffer, readers with names longer than 64 bytes are truncated and (at least in the opensc implementation) the last three characters are replaced with ellipses. This raises several issues:

1. If a user sets the 'reader:' option to the full name, the reader will not be found
2. Retrieving the ATR for a card fails, since pcsc won't find the reader by its truncated name (I'm not sure why this step is required)
3. The reader name in `TSSmartCardCreds` would be incorrect, preventing automatic selection of the right key

akallabeth:
That sounds like it needs to be documented somewhere, but nothing we can fix in this project (as we depend on `pcsc` and others to give us the correct names)
but I think (do not have concrete evidence at hand) that the `64 byte` limitation is part of the standard (at least it sounds like a reasonable limit for microcontroller provided strings)

fifthdegree:
The 64 byte limit is in PKCS11; pcsc limits names to 128 bytes. I think we should at least be able to solve this in the case where there's only one reader connected (or if other readers names don't match the first 61 bytes). I don't know if it's possible to link pkcs11 tokens with the readers pcsc returns, but if it is that could be a solution

PR #8087, opened 2022-07-21
Smartcard should be usable even when unable to get its ATR

fifthdegree:
Retrieving a card's ATR can fail when its reader name as given by pcsc is longer than 64 bytes (see #7965 - reader names can be up to 128 bytes). We don't seem to use the ATR for anything, so perhaps it can be removed entirely, but at least failing to retrieve it should not make it unusable for login


mfleisz:
@freerdp-bot test

mfleisz (approved):
LGTM

PR #8111, opened 2022-08-04
kerberos: renew the cache when credentials are expired

hardening:
This PR address the case with:
* missing ccache file
* a ccache with expired TGT


fifthdegree:
The same logic used for making sure the default cache is initialized with the right principal can be used for the case of a custom cache by just setting the default cache to the user-specified one (so krb5_cc_cache_match searches that) and initializing that cache instead of "MEMORY:" (suggestion)

akallabeth:
@hardening is this ready for review or still wip?

hardening:
I need to retest all that after U2U merge

PR #8120, opened 2022-08-12
tls: add an option to dump tls secrets for wireshark decoding

hardening:
This new option /tls-secret-file:<file> allows to dump TLS secrets in a file with the SSLKEYLOGFILE format. So this way you can setup the TLS dissector of wireshark (Pre-Master-Secret log filename) and see the traffic in clear in wireshark.
It also add some more PFS ciphers to remove for netmon captures.


hardening:
This PR is worth giving a try, with last wireshark from git you'll basically see all traffic in clear when you have configured the "Pre-Master-Secret log filename".

mfleisz (approved):

PR #8135, opened 2022-08-18
Fix memory leak in tls.c::tls_accept with openssl reference count

DVeron-RC:
There was an issue in the reference count management of the private
key and the X509 certificate.

The local reference will be lost at the end of the scope anyway
so the reference count should be decremented whether the
subsequent call succeeds or fails.

To test, you can run a RDP server using FreeRDP under valgrind.


mfleisz:
@freerdp-bot test

mfleisz (approved):
LGTM

PR #8143, opened 2022-08-20
remote assistance allow ip addresses of minimum length 7 (x.x.x.x)

garbb:
Minimum valid IP address is x.x.x.x (length 7)?


hardening:
@freerdp-bot would do the test of this PR please ?

hardening (approved):

PR #8147, opened 2022-08-22
fix freerdp_assistance_parse_address_list parsing

garbb:
Fails to parse when connection string only has one host:port because there is no ";" character. Also when multiple host:port;host:port it skips first host:port and parses remaining host:port as ";host:port...end" of connection string:
eg:
;192.168.93.138:49626;192.168.93.139:49627;192.168.93.140:49628
;192.168.93.139:49627;192.168.93.140:49628
;192.168.93.140:49628

### DIFF
--- a/libfreerdp/common/assistance.c
+++ b/libfreerdp/common/assistance.c
@@ -207,7 +207,7 @@ static BOOL freerdp_assistance_parse_address_list(rdpAssistanceFile* file, char*
 	}
 	rc = TRUE;
 out:
-	return rc;	
+	return rc;
 }
 
 static BOOL freerdp_assistance_parse_connection_string1(rdpAssistanceFile* file)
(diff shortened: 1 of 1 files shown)


hardening:

libfreerdp/common/assistance.c
```diff
@@ -180,33 +180,34 @@ static BOOL append_address(rdpAssistanceFile* file, const char* host, const char
...
 
-		p = q;
+		token = strtok(NULL, s);
 	}
-
 	rc = TRUE;
 out:
-	return rc;
+	return rc;	
 }
```
hardening, line 211: extra spaces à EOL

hardening:
@freerdp-bot test please

hardening (approved):
LGTM

PR #8148, opened 2022-08-23
drdynvc: code cleanup of the client dynamic channel

hardening:
This patch does various cleanups in the client dynamic channel. The main goal of the cleanup was to add the sending of Close messages to the server when a channel is locally closed. The refcounter is there to ensures that the DVC_CHANNEL is not freed while some pieces of code are still holding a reference on it.

I did some tests by using a custom server-side echo channel at https://github.com/hardening/echoChannel, it allows to send a given amount of packets and then close (to test server-side initiated closes). It compiles with mingw (so under linux) and so it can be easily deployed (no deps).

### DIFF
--- a/channels/drdynvc/client/drdynvc_main.c
+++ b/channels/drdynvc/client/drdynvc_main.c
@@ -23,6 +23,7 @@
 
 #include <winpr/crt.h>
 #include <winpr/stream.h>
+#include <winpr/interlocked.h>
 
 #include <freerdp/channels/drdynvc.h>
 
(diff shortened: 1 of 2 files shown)


channels/drdynvc/client/drdynvc_main.c
```diff
@@ -336,83 +343,128 @@ static UINT dvcman_load_addin(drdynvcPlugin* drdynvc, IWTSVirtualChannelManager*
...
-	DVCMAN_CHANNEL* channel;
-
-	if (dvcman_find_channel_by_id(pChannelMgr, ChannelId))
-	{
-		WLog_Print(drdynvc->log, WLOG_ERROR,
-		           "Protocol error: Duplicated ChannelId %" PRIu32 " (%s)!", ChannelId,
-		           ChannelName);
-		return NULL;
-	}
+	if (!channel)
```
mfleisz, line 348: Think this should be an `assert`

mfleisz (approved):
Minor comment - otherwise LGTM

channels/drdynvc/client/drdynvc_main.c
hardening: Hum yes correct, that's the rest of the previous implementation, `dvcman_channel_free` was used as dtor of the `channelById` Hashtable, and it happens that the dtor is called with NULL. Will change that and see if that's the only case.

hardening: I did the changes.

mfleisz (approved):
LGTM

PR #8160, opened 2022-08-29
smartcard: also filter certificate by domain name

hardening:
Command line username is used to filter the smartcard certificates during enumeration, this patch also add the domain as a filter.


mfleisz (approved):
LGTM

PR #8163, opened 2022-08-30
proxy: Fix smartcard emulation with clients with active rdpdr channel

mfleisz:
When a client connects to a proxy that has smartcard emulation active, both the emulated smartcard PDUs and the original client PDUs are sent to the server's rdpdr channel. This is because `pf_rdpdr_front_data` and `pf_rdpdr_back_data` always return `PF_CHANNEL_RESULT_PASS` which means the original client/server rdpdr PDU is always forwarded.

This PR fixes smartcard emulation by checking if smartcard emulation is enabled (using `pf_channel_smartcard_client_emulate`) and if this is the case `PF_CHANNEL_RESULT_DROP` will be returned.

The `Stream_CheckAndLogRequiredLength` was changed because clients might send an empty device list announcement which only consists of the header (two `UINT16`) and the device count (`UINT32` which is 0).

The PR also contains a small logging fix where an invalid format specifier was specified.


hardening (approved):
LGTM

PR #8167, opened 2022-08-31
gateway: Fix Sec-WebSocket-Key accept handling (#8166)

mfleisz:
Fixes the issue reported in #8166

### DIFF
--- a/libfreerdp/core/gateway/http.c
+++ b/libfreerdp/core/gateway/http.c
@@ -56,7 +56,7 @@ struct s_http_context
 	char* RdgConnectionId;
 	char* RdgAuthScheme;
 	BOOL websocketUpgrade;
-	BYTE SecWebsocketKey[16];
+	char* SecWebsocketKey;
 };
 
 struct s_http_request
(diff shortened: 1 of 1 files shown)


libfreerdp/core/gateway/http.c
```diff
@@ -1128,7 +1127,7 @@ BOOL http_response_is_websocket(HttpContext* http, HttpResponse* response)
 	if (!winpr_Digest_Init(sha1, WINPR_MD_SHA1))
 		goto out;
 
-	if (!winpr_Digest_Update(sha1, http->SecWebsocketKey, sizeof(http->SecWebsocketKey)))
+	if (!winpr_Digest_Update(sha1, http->SecWebsocketKey, strlen(http->SecWebsocketKey)))
```
akarl10, line 1130: you already fixed that previously by making http->SecWebsocketKey of type BYTE, but now gcc complains about the second argument being a char*. I suggest casting that to BYTE. strlen should be correct.

akarl10:
this patch works for me.

libfreerdp/core/gateway/http.c
mfleisz: You'r right - thx for pointing out

akarl10:
LGTM

akallabeth (approved):

PR #8173, opened 2022-09-05
x11: Fix disconnects if WITH_XCURSOR is set to OFF

mfleisz:
If `WITH_XCURSOR` is set to `OFF`, `xf_Pointer_New` will currently return `FALSE` causing a disconnect.


pnowack:
From my perspective, what is written in the description here should actually be in the body of the commit message, as otherwise it gets lost (when backporting, or viewing commits locally (not on github)).

mfleisz:
@pnowack updated the description 👍

pnowack:
Change itself looks good to me.

> @pnowack updated the description +1

I think you forgot to force push your changes.

pnowack (approved):
LGTM

PR #8177, opened 2022-09-06
Manage default printer for the win backend

Pollux42:
Currently, when printer are redirected, the declared default printer is always the first redirected printer.
With this PR, the declared default printer is now the local default printer.

### DIFF
--- a/channels/printer/client/win/printer_win.c
+++ b/channels/printer/client/win/printer_win.c
@@ -234,7 +234,7 @@ static void printer_win_release_ref_printer(rdpPrinter* printer)
 }
 
 static rdpPrinter* printer_win_new_printer(rdpWinPrinterDriver* win_driver, const WCHAR* name,
-                                           const WCHAR* drivername, BOOL is_default)
+                                           const WCHAR* drivername)
 {
 	rdpWinPrinter* win_printer;
 	DWORD needed = 0;
@@ -252,7 +252,7 @@ static rdpPrinter* printer_win_new_printer(rdpWinPrinterDriver* win_driver, cons
 
 	if (!win_printer->printer.name)
 		goto fail;
-	win_printer->printer.is_default = is_default;
+	win_printer->printer.is_default = _wcscmp(name, win_driver->defaultPrinter) == 0;
 
 	win_printer->printer.CreatePrintJob = printer_win_create_printjob;
 	win_printer->printer.FindPrintJob = printer_win_find_printjob;
@@ -346,7 +346,7 @@ static rdpPrinter** printer_win_enum_printers(rdpPrinterDriver* driver)
 	{
 		rdpPrinter* current = printers[num_printers];
 		current = printer_win_new_printer((rdpWinPrinterDriver*)driver, prninfo[i].pPrinterName,
-		                                  prninfo[i].pDriverName, 0);
+		                                  prninfo[i].pDriverName);
 		if (!current)
 		{
 			printer_win_release_enum_printers(printers);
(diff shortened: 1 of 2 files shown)


mfleisz:
@freerdp-bot test

channels/printer/client/win/printer_win.c
```diff
@@ -432,6 +440,15 @@ rdpPrinterDriver* win_freerdp_printer_client_subsystem_entry(void)
 		win_driver->driver.ReleaseRef = printer_win_release_ref_driver;
 
 		win_driver->id_sequence = 1;
+		win_driver->defaultPrinter = NULL;
+
+		GetDefaultPrinter(NULL, &size);
+		if (size)
+		{
+			win_driver->defaultPrinter = (LPWSTR)calloc(size, sizeof(WCHAR));
+			if (!GetDefaultPrinter(win_driver->defaultPrinter, &size))
```
hardening, line 449: missing `calloc` return check

hardening:
Question: by setting the default printer at the driver level doesn't it mean that if the default printer is changed by any mean, then we'll continue to use the printer that was set as default at the beginning of the program ?

channels/printer/client/win/printer_win.c
```diff
@@ -408,6 +408,12 @@ static void printer_win_release_ref_driver(rdpPrinterDriver* driver)
 	rdpWinPrinterDriver* win = (rdpWinPrinterDriver*)driver;
 	if (win->references <= 1)
 	{
+		if (win->defaultPrinter)
```
hardening, line 411: no need to check for `win->defaultPrinter != NULL` as `free(NULL)` is safe (and does the test internally)

channels/printer/client/win/printer_win.c
Pollux42: Done

channels/printer/client/win/printer_win.c
Pollux42: Done

Pollux42:
Yes, the default printer name is set at the beginning of the program and cannot be changed after.

Pollux42:
I have just realized that the cups backend had the same issue, so I did the same code change for it.

channels/printer/client/win/printer_win.c
```diff
@@ -432,6 +435,19 @@ rdpPrinterDriver* win_freerdp_printer_client_subsystem_entry(void)
...
 
 		win_driver->id_sequence = 1;
+		win_driver->defaultPrinter = NULL;
+
+		GetDefaultPrinter(NULL, &size);
+		if (size)
+		{
+			win_driver->defaultPrinter = (LPWSTR)calloc(size, sizeof(WCHAR));
+
+			if (win_driver->defaultPrinter)
```
hardening, line 445: If we can't allocate memory for the default printer's name, I'd rather exit with an error and log a message, than continuing like if everything is ok..

Pollux42: So what about:

		win_driver->defaultPrinter = NULL;

		GetDefaultPrinter(NULL, &size);
		if (size)
		{
			win_driver->defaultPrinter = (LPWSTR)calloc(size, sizeof(WCHAR));

			if (!win_driver->defaultPrinter)
			{
				win_driver->driver.ReleaseRef(&win_driver->driver);
				return NULL;
			}

			if (!GetDefaultPrinter(win_driver->defaultPrinter, &size))
				win_driver->defaultPrinter[0] = "\0";
		}

(And I did the same kind of error message as for if the win_driver allocation fails ;-) )

Pollux42: Or directly `free(win_driver)` instead of `ReleaseRef`?

hardening: What ever works, probably that the direct free is better as this pointer can't have any other references

Pollux42: Done

hardening:
@freerdp-bot test

hardening:
@Pollux42 run `clang-format` on your changes to pass the code quality checker, and I'll merge it

hardening:
@freerdp-bot can you check and test again please ?

PR #8181, opened 2022-09-07
FIX: error checking

KangLin:
Error checking buffer in wf_format_text.


mfleisz:
@freerdp-bot test

mfleisz (approved):
Nice catch - thanks

PR #8194, opened 2022-09-12
Fix #8176: Check fullscreen state and not setting

### LINKED ISSUE #8176: /dynamic-resolution does not working.
**Describe the bug**
After running with /f, /dynamic-resolution does not work when using shortcut keys or float bar to cancel full screen.

**To Reproduce**
1. Run xfreerdp command.

        $ xfreerdp /log-level:DEBUG /f /floatbar:sticky:off,default:hidden,show:always /dynamic-resolution /v:xxx.xxx.xx.x /u:xxxxx

2. Remove the full screen display by manipulating the floatbar (or press ctrl + alt + enter)
3. Resize the window.
4. Automatic resizing does not work.The window size can be adjusted, but the resolution is not changed.

**Expected behavior**
When the window size is adjusted, I would like the resolution to automatically change to match the window size.

**Screenshots**
screenshot

**Additional context**
the problem did not occur in v2.2.0, but has been occurring since v2.3.0.
In addition to /f, the problem was also reproduced with /multimon.
I am connected to windows 10.

Specifically, the following commit was causing the problem:
d3ed42a7

akallabeth:
Timer based dynamic resolution updates are disabled in fullscreen state. Check current value and not (permanent) setting

### DIFF
--- a/client/X11/xf_disp.c
+++ b/client/X11/xf_disp.c
@@ -115,7 +115,7 @@ static BOOL xf_update_last_sent(xfDispContext* xfDisp)
 
 static BOOL xf_disp_sendResize(xfDispContext* xfDisp)
 {
-	DISPLAY_CONTROL_MONITOR_LAYOUT layout;
+	DISPLAY_CONTROL_MONITOR_LAYOUT layout = { 0 };
 	xfContext* xfc;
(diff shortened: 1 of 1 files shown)


hardening (approved):
LGTM

PR #8195, opened 2022-09-12
Unify encomsp request control code

akallabeth:


akallabeth:
@garbb please confirm windows client still works the way you intended

hardening (approved):
LGTM

garbb:
Yes it works as expected. Thanks.
Another thing I noticed if `/auto-request-control` parameter is used: The server has a button to disable control and when that is pressed control will be disabled but the client will call `client_encomsp_participant_created()` again and immediately request control again and popup prompt on server again. This is probably not intended and a bit annoying.


https://user-images.githubusercontent.com/4453789/189717483-c3b89db9-8bb9-4d6b-ab37-08d98a5594db.mp4



I would suggest to only run [this](https://github.com/FreeRDP/FreeRDP/pull/8195/files#diff-2d94332c2a112728e85ee57b6e35cee103fcdf0d3634af6e1b10f915d095ccdfR1017) code once upon first call to `client_encomsp_participant_created()`. Maybe this could be achieved by changing setting of `FreeRDP_RemoteAssistanceRequestControl` to FALSE after first call? Or maybe just set flag within `client_encomsp_participant_created()` to only request control on first call?
I tried to suggest code edit directly in this pull request but I couldn't figure out how.

akallabeth:
@garbb better make a new pr for that, I´ll merge this meanwhile.

PR #8197, opened 2022-09-12
Rdpei fix

akallabeth:
@hardening looks like we missed something with the generic channel code refactoring

### DIFF
--- a/channels/rdpei/client/rdpei_main.c
+++ b/channels/rdpei/client/rdpei_main.c
@@ -86,6 +86,7 @@ typedef struct
 	rdpContext* rdpcontext;
 	HANDLE thread;
 	HANDLE event;
+	BOOL running;
 } RDPEI_PLUGIN;
 
 /**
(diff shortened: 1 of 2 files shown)


channels/rdpei/client/rdpei_main.c
```diff
@@ -448,7 +449,7 @@ static DWORD WINAPI rdpei_periodic_update(LPVOID arg)
 		goto out;
 	}
 
-	while (rdpei->base.initialized)
+	while (rdpei->running)
```
hardening, line 452: Hum then we should probably remove the `base.initialized` field as `rdpei` was the only user

hardening (approved):

channels/rdpei/client/rdpei_main.c
akallabeth: Will check that later. Maybe we can move the init code in rdpei so that we can use that field again?

hardening: Well no other place use `base.initialized`, it was used in the previous code as the `running` bool that you have added for `rdpei`. But we can keep it in case some other channel would need it.

PR #8200, opened 2022-09-13
Fixed gnome and mate clipboard format synthesizers

akallabeth:


akallabeth:
@pnowack found the culprit. half done support for older clipboard formats

winpr/libwinpr/clipboard/posix.c
```diff
@@ -572,28 +584,118 @@ static FILEDESCRIPTORW* convert_local_file_list_to_filedescriptors(wArrayList* f
...
+}
+
+static BOOL process_mate_copied_files(wClipboard* clipboard, const char* data, UINT32 pSize)
+{
+	return process_files(clipboard, data, pSize, "copy\n");
+}
+
+static BOOL process_nautilus_clipboard(wClipboard* clipboard, const char* data, UINT32 pSize)
+{
+	return process_files(clipboard, data, pSize, "x-special/nautilus-clipboard\copy\n");
```
pnowack, line 661: I wonder, when we will remove the old nautilus-clipboard type. Next week GNOME 43 is released (meaning only 42 and 43 are supported as stable releases). Then old nautilus-clipboard type was removed in 40.
Maybe we can remove it here in FreeRDP with the next Debian release next year (which will use GNOME 43 for debian stable).

---

In any case, `"x-special/nautilus-clipboard\copy\n"` should probably be `"x-special/nautilus-clipboard\\copy\n"` (missing escaping).

pnowack (changes requested):
Did a quick test: It seems, that not the copied file is used for the file list, but its parent instead. E.g. if I copy a lone file, which is in a folder containing 100.000 files, then the resulting file contains that parent folder, with its files.

In addition to that, not only the parent is copied, but the file instead. What I mean with this, is that a dialog is shown on the server, whether to replace the file to copy (because the file is two times on the file list (one via parent, one itself).
(Used here g-r-d on the server side).

Haven't looked (further) yet at the code.

akallabeth:
@pnowack ok, interesting. which format do you get?
I´ve tested with a few files from my download folder and only got the ones I did request.
Nice catch with the typo ;)

pnowack:
@akallabeth The `x-special/gnome-copied-files` format is used here (nautilus 43 is the producer). I just reconnected to the server and it did not happen this time (when copying one file).
However, then I copied the file again (same connection), then copied the file again, deleted the file on the server side.
Since it is a new file list, the cached data on the server is also deleted, then I pasted it again and the file was copied + I got a dialog for replacing the file.

So, it seems, that the old list is not cleared, when the selection on the client side changes.

winpr/libwinpr/clipboard/posix.c
```diff
@@ -572,28 +583,121 @@ static FILEDESCRIPTORW* convert_local_file_list_to_filedescriptors(wArrayList* f
...
+	data += prefix_len;
+	pSize -= prefix_len;
+
+	BOOL rc = FALSE;
+	char* copy = strndup(data, pSize);
+	if (!copy)
+		goto fail;
+
+	char* endptr = NULL;
+	char* tok = strtok_s(copy, "\n", &endptr);
```
pnowack, line 635: `strtok_s()` is not thread-safe, so this should use the thread safe variant here (`strtok()`) (especially since the posix clipboard can be used outside of FreeRDP (as it is part of the API)).

winpr/libwinpr/clipboard/posix.c
```diff
@@ -572,28 +583,121 @@ static FILEDESCRIPTORW* convert_local_file_list_to_filedescriptors(wArrayList* f
...
+
+	if (!data || (pSize < prefix_len))
+		return FALSE;
+	if (strncmp(data, prefix, prefix_len) != 0)
+		return FALSE;
+	data += prefix_len;
+	pSize -= prefix_len;
+
+	BOOL rc = FALSE;
+	char* copy = strndup(data, pSize);
```
pnowack, line 630: I wonder, if we really need that copy here.

winpr/libwinpr/clipboard/posix.c
pnowack: Seems that `strtok()` is not the thread safe variant one. `strtok_r()` is apparently the thread safe one (https://linux.die.net/man/3/strtok).

pnowack: Looks like `strtok_r()` and `strtok_s()` are both fine. Resolving.

pnowack:
> Did a quick test: It seems, that not the copied file is used for the file list, but also its parent instead. E.g. if I copy a lone file, which is in a folder containing 100.000 files, then the resulting file contains that parent folder, with its files.

Can't reproduce that bug any more and I cannot find any indicator for that as bug in FreeRDP for this.

---

LGTM

pnowack (approved):

PR #8201, opened 2022-09-13
if /auto-request-control, only request control once upon connect

garbb:
if auto-request-control setting is enabled then only request control once upon connect, otherwise it will auto request control again every time server turns off control which is a bit annoying.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #8205, opened 2022-09-14
Improve string version extraction regex

akallabeth:
Allow (git) tags of format 'some-important-text_v1.2.3-something_else' Any character is now allowed, as long as there is a version of format 1.2.3 somewhere in the tag name.


hardening (approved):
LGTM

PR #8210, opened 2022-09-15
Disable execinfo support if only header detected

akallabeth:
Some implementations (e.g. Android with API < 33) provide the execinfo.h header file, but do not define the backtrace functions expected. Disable the support for execinfo backtrace in such a case


hardening (approved):
Looks ok

akallabeth:
@freerdp-bot test

PR #8214, opened 2022-09-18
remember participantId for use with assistance set control request

garbb:
In freerdp_client_encomsp_set_control() participantId was set to 0. Usually this works fine but in some cases host ignores set control request unless it is sent with same participantId that was sent from host in participant_created_pdu. So save participantId and send it for control request.
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpemc/1d6d3484-0eff-44fa-9a4e-2ad6e1391c45


hardening (approved):

hardening:
@freerdp-bot give it a test please

akallabeth (approved):

PR #8216, opened 2022-09-19
WinPR::Clipboard::Posix: Support WINDOWS

KangLin:
- WinPR::Clipboard::Posix: Support WINDOWS
- [WinPR::Clipboard: rename posix.c to synthetic_file.c](https://github.com/FreeRDP/FreeRDP/commit/b893f45519f94c2fb89472f0beaade241aec2e6c)
Rename posix to synthetic for more precise meaning.


KangLin:
@freerdp-bot test!

winpr/libwinpr/clipboard/dirent.c
```diff
@@ -0,0 +1,85 @@
...
+
+DIR* opendir(const char* name)
+{
+	DIR* dir;
+	WIN32_FIND_DATA FindData;
+	char namebuf[512];
+
+	sprintf(namebuf, "%s\\*.*", name);
+
+	HANDLE hFind = FindFirstFileA(namebuf, &FindData);
```
akallabeth, line 16: Better use `FindFirstFileW` on windows, or unicode file names will not work.

akallabeth:
@freerdp-bot test

winpr/libwinpr/clipboard/dirent.c
KangLin: Yes. But the input parameter is const char* local_name

static BOOL add_directory_contents_to_list(wClipboard* clipboard, const char* local_name,
                                           const WCHAR* remote_name, wArrayList* files)

Need to modify all functions?

akallabeth: @KangLin no, expect the input to use `utf-8`.

winpr/libwinpr/clipboard/dirent.c
```diff
@@ -0,0 +1,91 @@
...
+	WIN32_FIND_DATAW FindData;
+	char namebuf[MAX_PATH];
+	WCHAR* file = NULL;
+	int nLen = 0;
+	int nRet = 0;
+
+	sprintf(namebuf, "%s\\*.*", name);
+	nLen = strlen(namebuf);
+
+	do
```
akallabeth, line 20: Hint: You can just do ` { code }` instead of `do { code } while(0);` for scoping

winpr/libwinpr/clipboard/dirent.c
```diff
@@ -0,0 +1,91 @@
...
+		if (!dir)
+		{
+			WLog_ERR(TAG, "DIR memory allocate fail\n");
+			break;
+		}
+
+		dir->dd_fd = hFind; // simulate return
+	} while (0);
+
+	if (file)
```
akallabeth, line 43: `free(ptr)` does work with `NULL` pointers, no need to check.

winpr/libwinpr/clipboard/dirent.c
```diff
@@ -0,0 +1,91 @@
...
+	// fail or end
+	if (!bf)
+		return 0;
+
+	struct dirent* dirent =
+	    (struct dirent*)calloc(1, sizeof(struct dirent) + sizeof(FileData.cFileName));
+
+	for (i = 0; i < 256; i++)
+	{
+		dirent->d_name[i] = FileData.cFileName[i];
```
akallabeth, line 68: This should be converted from 16bit unicode to `utf-8` or am I missing something?

winpr/libwinpr/clipboard/dirent.c
```diff
@@ -0,0 +1,91 @@
...
+
+	bf = FindNextFileW(d->dd_fd, &FileData);
+	// fail or end
+	if (!bf)
+		return 0;
+
+	struct dirent* dirent =
+	    (struct dirent*)calloc(1, sizeof(struct dirent) + sizeof(FileData.cFileName));
+
+	for (i = 0; i < 256; i++)
```
akallabeth, line 66: you might use `ARRAYSIZE(FileData.cFileName)` instead

winpr/libwinpr/clipboard/dirent.c
```diff
@@ -0,0 +1,91 @@
...
+
+	if (file)
+		free(file);
+
+	return dir;
+}
+
+struct dirent* readdir(DIR* d)
+{
+	int i;
```
akallabeth, line 51: Index variable should here be of type `size_t`

winpr/libwinpr/clipboard/dirent.c
KangLin: Here's a trick to make sure you don't forget to free memory.

winpr/libwinpr/clipboard/dirent.c
KangLin: Yes. Do a pointer check, keep it a good habit.

winpr/libwinpr/clipboard/dirent.c
KangLin: OK. It has been modified. but it is only work in utf-8.

My OS:
 Microsoft Windows [Version 10.0.19044.2006]
 Simplified Chinese Windows is GB2313 - CodePage 936

It is not work.

KangLin:
@freerdp-bot test

winpr/libwinpr/clipboard/dirent.c
KangLin: After testing, it turns out that using FindFirstFileA is correct. Pelease see TestDirent.c

winpr/libwinpr/clipboard/dirent.c
mfleisz: @KangLin actually checking for `NULL` before calling `free` is bad practice as `free(NULL)` is well defined by the standard and considered safe.

KangLin: @mfleisz  This function does not need to check this pointer and will not make an error, but not checking the pointer before calling other functions may cause the program to crash.

akallabeth: @KangLin while this is true for many functions `free(NULL)` is `well defined` by the standard, so in case there is a crash fix your `libc` because it is broken ;)

winpr/libwinpr/clipboard/dirent.c
```diff
@@ -0,0 +1,171 @@
+
+#include <stdio.h>
+#include "dirent.h"
+
+#include "../log.h"
+#define TAG WINPR_TAG("clipboard.synthetic.file")
+
+DIR* opendirW(const char* name)
```
akallabeth, line 8: Having a `opendirW` and `opendirA` does not make sense. If you want to support `ANSI` builds on windows, do it internally here with the usual `define guards` as done for other windows builds

winpr/libwinpr/clipboard/synthetic_file.c
```diff
@@ -36,12 +36,22 @@
...
-#include <unistd.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 
+#ifdef _WIN32
+#include "dirent.h"
+#define S_ISDIR(x) (((x)&S_IFDIR))
+#if defined(_MSC_VER)
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
```
akallabeth, line 48: defined win `winpr/wtypes.h`

akallabeth (changes requested):
Exposing `opendirA` and `opendirW` is of no use and is just confusing

akallabeth (changes requested):
Ok, just did a dive into this, `dirent` does not need to be emulated on windows, we already do it the other way aryound.
Just replace the calls to `opendir` and `readdir` with `FindFirstFileW` and `FindNextFileW`

[edit] sorry for missing that before :/

winpr/libwinpr/clipboard/dirent.c
KangLin: @akallabeth  My understanding is that multi-byte parameters are char*, and wide-byte(UNICODE) parameters are wchar_t*.
FindFirstFileA handles multi-byte parameters (char*), include ASNI and UTF-8
FindFirstFileW handles wide-byte(UNICODE) parameters (wchar_t*). include UTF-16、GB2313 is 2byte, UTF-32 is 4byte etc.
Since the current function [DIR* opendir(const char* name), expect the input to use utf-8.] is passing a mult-ibyte parameter(char*), I think it is correct to use FindFirstFileA. 
If the function [DIR* opendir(const wchar_t* name), expect the input to use UTF-16、GB2313.] is passing a wide-byte parameter(wchar_t*), then use FindFirstFileW.

winpr/libwinpr/clipboard/test/TestDirent.c
```diff
@@ -0,0 +1,73 @@
...
+	if (nRet)
+	{
+		WLog_ERR(TAG, "failed to close directory");
+		return nRet;
+	}
+
+	return nRet;
+}
+
+int main(int argc, char* argv[])
```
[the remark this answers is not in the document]
KangLin: Yes. If is UNICODE, use wchar_t* argv[] arguments.

winpr/libwinpr/clipboard/dirent.c
akallabeth: @KangLin No, `FindFirstFileA` is `multi-byte`, yes, but not `utf-8` but current code page. we´ve fixed similar issues for freerdp on windows on multiple occasions.

winpr/libwinpr/clipboard/test/TestDirent.c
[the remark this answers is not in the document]
akallabeth: you do not understand. a `CTest` does not have a `main()` function but a `TestCase` (whatever name the test case has), e.g. `TestDirent()` entry point function.
check all the other unit tests.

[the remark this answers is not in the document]
KangLin: Where can I find CTest related documents?

akallabeth (changes requested):
There are more issues here:
the `open()`, `fstat()`, `stat()` and `close()` functions do not exist (or are not entirely compatible) under windows.

KangLin:
> There are more issues here: the `open()`, `fstat()`, `stat()` and `close()` functions do not exist (or are not entirely compatible) under windows.

I am aware of this issue and am making changes.

winpr/libwinpr/clipboard/test/TestDirent.c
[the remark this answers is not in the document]
KangLin: > https://cmake.org/cmake/help/latest/manual/ctest.1.html

I looked at this documentation but there is no description of the entry point function.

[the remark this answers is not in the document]
akallabeth: @KangLin just look at all the other unit tests in freerdp.

`winpr/libwinpr/clipboard/test/CMakeLists.txt` already contains a bunch of tests, just add anotherone there.
(the `add_test` loop expects the entry point function name to be the same as the filename)

KangLin:
See #8246

PR #8217, opened 2022-09-19
Improve length checks in parse_uri_to_local_file

akallabeth:


pnowack:
Changes themselves look good to me.

Nit: Would though use the word `Fix`, instead of `Improve`, in the commit message, because it would be in my opinion more precise (since previously we were able to run into out-of-bounds situations, and the word 'improve' itself does not imply in what direction things are 'improved' (or changed)).

pnowack:
> Nit: Would though use the word Fix, instead of Improve, in the commit message, because it would be in my opinion more precise (since previously we were able to run into out-of-bounds situations, and the word 'improve' itself does not imply in what direction things are 'improved' (or changed)).

`Fix out of bounds reads in parse_uri_to_local_file` should be precise enough here.

akallabeth:
@pnowack if you could add a positive review then we can merge ;)

pnowack (approved):
Thanks!

akallabeth:
@freerdp-bot test

PR #8218, opened 2022-09-19
fix connection abort when copy from remote wordpad, local windows 10,…

2fly2:
problem: 
on windows 10, with clipboard history opened, connect remote desktop, copy text from wordpad, connection abort.

root cause:
with local clipboard history opened, when remote copy text from wordpad, local system will request some format that server can not support. I think it is not an error, it means just can not copy the format from remote. 

fix:
treat server clipboard message flag CB_RESPONSE_FAIL as normal.

### DIFF
--- a/client/Windows/wf_cliprdr.c
+++ b/client/Windows/wf_cliprdr.c
@@ -2191,13 +2191,20 @@ wf_cliprdr_server_format_data_response(CliprdrClientContext* context,
 	if (!context || !formatDataResponse)
 		return ERROR_INTERNAL_ERROR;
 
-	if (formatDataResponse->common.msgFlags != CB_RESPONSE_OK)
-		return E_FAIL;
-
 	clipboard = (wfClipboard*)context->custom;
 
 	if (!clipboard)
 		return ERROR_INTERNAL_ERROR;
+
+	if (formatDataResponse->common.msgFlags != CB_RESPONSE_OK)
+	{
+		clipboard->hmem = NULL;
+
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):

client/Windows/wf_cliprdr.c
```diff
@@ -2191,14 +2191,21 @@ wf_cliprdr_server_format_data_response(CliprdrClientContext* context,
...
-		return E_FAIL;
-
 	clipboard = (wfClipboard*)context->custom;
 
 	if (!clipboard)
 		return ERROR_INTERNAL_ERROR;
 
+	if (formatDataResponse->common.msgFlags != CB_RESPONSE_OK)
+	{
+		clipboard->hmem = NULL;
```
mfleisz, line 2201: Hm - sure we don't leak any memory here?

akallabeth: @mfleisz no, this should already be handled. just look a few lines below, where it normally is allocated.

mfleisz: @akallabeth a few lines below the newly allocated memory is assigned to `clipboard->hmem` (line 2230) and if we just `NULL` it here that handle will be lost forever ... or do I miss something else?

akallabeth: @mfleisz yes, the `return CHANNEL_RC_OK` in the new path?

PR #8219, opened 2022-09-19
Support spnego/kerberos authentication for gateway

fifthdegree:
This allows for using kerberos to authenticate to a gateway. It also cleans up support code for authentication (there was some redundancy in nla and gateway)

The benefit is of course better security, but it also adds the potential for using smartcards

Note that for reasons beyond my ability to understand TS Gateway won't accept connections from rpc over http when authenticated using kerberos (schannel is still an option; I think that can be used for smartcard support)


hardening:
@fifthdegree  you're my hero, I was about to start working on that !

hardening:
@freerdp-bot test please

libfreerdp/core/nla.c
```diff
@@ -136,17 +118,15 @@ struct rdp_nla
...
+
+	rdpAuth* auth;
+	char* pkinitArgs;
+	BYTE certSha1[20];
 };
 
 static BOOL nla_send(rdpNla* nla);
 static int nla_server_recv(rdpNla* nla);
-static void nla_buffer_free(rdpNla* nla);
 static SECURITY_STATUS nla_encrypt_public_key_echo(rdpNla* nla);
```
akallabeth, line 130: Does not match implementation

akallabeth:
@freerdp-bot test

libfreerdp/core/nla.c
akallabeth: ok, fixed these (and a few other compiler warnings)

akallabeth:
@freerdp-bot test

akallabeth (approved):

akallabeth:
@freerdp-bot test

hardening:
@akallabeth sure I'll do a careful review

libfreerdp/core/gateway/rts.c
```diff
@@ -857,9 +857,9 @@ static BOOL rts_read_pdu_bind_nak(wStream* s, rpcconn_bind_nak_hdr_t* ctx)
 	WINPR_ASSERT(s);
 	WINPR_ASSERT(ctx);
 
-	if (!Stream_CheckAndLogRequiredLength(
-	        TAG, s, sizeof(rpcconn_bind_nak_hdr_t) - sizeof(rpcconn_common_hdr_t)))
-		return FALSE;
+	// if (!Stream_CheckAndLogRequiredLength(
```
hardening, line 860: There's something to do here

libfreerdp/core/gateway/rts.c
```diff
@@ -898,9 +898,9 @@ static BOOL rts_read_pdu_fault(wStream* s, rpcconn_fault_hdr_t* ctx)
 	WINPR_ASSERT(s);
 	WINPR_ASSERT(ctx);
 
-	if (!Stream_CheckAndLogRequiredLength(
-	        TAG, s, sizeof(rpcconn_fault_hdr_t) - sizeof(rpcconn_common_hdr_t)))
-		return FALSE;
+	// if (!Stream_CheckAndLogRequiredLength(
```
hardening, line 901: Same here

libfreerdp/core/nla.c
```diff
@@ -97,9 +94,7 @@
  *
  */
 
-#define NLA_PKG_NAME NEGO_SSP_NAME
-
-#define TERMSRV_SPN_PREFIX "TERMSRV/"
+#define NLA_PKG_NAME AUTH_PKG_SPNEGO
```
hardening, line 97: is it me or it SPNego the default for NLA ?

hardening:
Ok so first, this is a massive improvement and a very nice mutualization. Still I have general remarks:

* you have changed `nla_decrypt_public_key_echo` and friends to return a `BOOL` instead of a `SECURITY_STATUS`, I find it quite unfortunate because it was giving more informations than just "it worked", "it didn't worked" (especially for OOM cases). I'd be for keeping the original behaviour (except if I missed something important) ;
* I don't like the name `auth` like in `rdpAuth`, `auth.c` or `auth_new` it's way too generic, shall we call that sspiAuth or credSspAuth ? (20% of the time coding, the rest of the time finding names ;) )

libfreerdp/core/gateway/rts.c
fifthdegree: That wasn't supposed to be part of the commit...

libfreerdp/core/nla.c
fifthdegree: Wasn't it always the default?

fifthdegree:
So my reasoning for changing SECURITY_STATUS to BOOL is that `auth` is meant to be an abstraction to authentication. Users of it (like nla) shouldn't really need to be concerned about *why* authentication failed, only that it did. What it does upon failure won't change based on the reason. Of course *we* want to know why authentication failed and to that end `auth` will log the status codes returned.

Regarding the name: honestly, 'auth' seems fine to me, but if you want to change it to something else (whenever you settle on it ;) ) that's fine with me too

akallabeth:
@fifthdegree could you rebase your pr please?
@hardening any more points you wish to be addressed?

fifthdegree:
@akallabeth sure, soon as I get a chance

fifthdegree:
@akallabeth  something got messed up while rebasing I think it has to do with one of your commits wanna take a look?

akallabeth:
@fifthdegree yes, @hardening did some net stuff regarding kerberos and /me did some format strings hardening. looks like the `nla.c` parts need quite the thourough merge as quite a lot changed.

akallabeth:
@fifthdegree clashed with https://github.com/FreeRDP/FreeRDP/pull/8261 looks like you need to manually sort that :/

akallabeth:
@freerdp-bot test

hardening:
@freerdp-bot can you test the changes please ?

libfreerdp/core/gateway/rdg.c
```diff
@@ -40,6 +41,12 @@
 
 #define TAG FREERDP_TAG("core.gateway.rdg")
 
+#ifdef WITH_SPNEGO
```
hardening, line 44: Should we refine that `#ifdef` ? Under windows we always have Negotiate, so I feel like `WITH_SPNEGO` is not the right test...

hardening (approved):
Change the `#ifdef` and we're ready for merge

hardening (approved):

hardening:
@freerdp-bot can you do the final test please

PR #8220, opened 2022-09-20
fix copy 0 size file on windows

2fly2:
problem:
wfreerdp connect to windows 10, copy file from remote, paste local, get a 0 size file.

root cause:
rdp server(remote windows 10) tells wfreerdp client file size, but we not use it. see flag FD_FILESIZE.

fix:
when server send file size, then  init file size with it.

### DIFF
--- a/client/Windows/wf_cliprdr.c
+++ b/client/Windows/wf_cliprdr.c
@@ -1486,15 +1486,12 @@ static LRESULT CALLBACK cliprdr_proc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM 
 				case OLE_SETCLIPBOARD:
 					DEBUG_CLIPRDR("info: OLE_SETCLIPBOARD");
 
-					if (S_FALSE == OleIsCurrentClipboard(clipboard->data_obj))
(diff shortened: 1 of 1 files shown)


client/Windows/wf_cliprdr.c
```diff
@@ -1482,15 +1486,12 @@ static LRESULT CALLBACK cliprdr_proc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM
 				case OLE_SETCLIPBOARD:
 					DEBUG_CLIPRDR("info: OLE_SETCLIPBOARD");
 
-					if (S_FALSE == OleIsCurrentClipboard(clipboard->data_obj))
```
akallabeth, line 1485: Why is this line removed?

2fly2: when copy file A.txt from remote, then copy file B.txt, if add this line, will not place file B.txt in clipboard, because file A.txt clipboard data object is still in clipboard, so  OleIsCurrentClipboard(clipboard->data_obj) return S_OK.

akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #8221, opened 2022-09-20
Mingw clang

hardening:
Some changes to compile under mingw with clang


winpr/include/winpr/intrin.h
```diff
@@ -38,10 +38,12 @@ static INLINE UINT32 __lzcnt(UINT32 _val32)
 	return ((UINT32)__builtin_clz(_val32));
 }
 
+#if !defined(__clang__)
```
akallabeth, line 41: breaks mac os builds...

akallabeth: might need an explicit check for `__MINGW32__`

akallabeth (approved):
LGTM

PR #8223, opened 2022-09-20
Improve management of default printer

Pollux42:
This PR is an improvement of #8177.

Printer are shared using two ways:
- ENUM to automatically share all the local printer
- GET to selectively share what printer to share

Since #8177, any printer shared (ENUM or GET) will be declared as the default printer if and only if it is the local default printer and this cannot be changed during execution.
Now, this behavior is sticked to the ENUM method. When you declare printers using GET, you can explicitely choose if the printer will be the default of not, and if you use freerdp as a library, you can even change that during a session.

### DIFF
--- a/channels/printer/client/cups/printer_cups.c
+++ b/channels/printer/client/cups/printer_cups.c
@@ -46,7 +46,6 @@ typedef struct
 
 	int id_sequence;
 	size_t references;
-	char* defaultPrinter;
 } rdpCupsPrinterDriver;
 
 typedef struct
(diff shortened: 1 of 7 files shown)


akallabeth (approved):
LGTM

channels/printer/client/cups/printer_cups.c
```diff
@@ -431,13 +429,6 @@ rdpPrinterDriver* cups_freerdp_printer_client_subsystem_entry(void)
 		uniq_cups_driver->driver.ReleaseRef = printer_cups_release_ref_driver;
 
 		uniq_cups_driver->id_sequence = 1;
-		uniq_cups_driver->defaultPrinter = _strdup(cupsGetDefault());
```
akallabeth, line 434: While you moved this on the windows printer implementation to the `enum_printers` function, you failed to do the same for `cups`

Pollux42: No, there is the same for cups, but it's very more concise. (No need to get the name of the default printer here)

PR #8224, opened 2022-09-20
Cleaned up android input event processing

akallabeth:
* Unified event cleanup code
* Add proper support for unicode input
* Add proper return code for event process failures


hardening (approved):
LGTM

PR #8229, opened 2022-09-20
fix mouse button "stuck" down when dragged out of window

garbb:
If user presses mouse button down within window and then moves it outside of window before releasing, host will still see mouse button as "stuck down" because mouse up event is not sent. So use `SetCapture()`/`ReleaseCapture()` to capture mouse events until button is released.
Also, since coordinates from WM_* events are signed values and will be negative if mouse is above or to the left of the window, need to prevent negative values from being sent or else they will be interpreted as large positive values and send the cursor flying off to the opposite end of the screen.

### DIFF
--- a/client/Windows/wf_event.c
+++ b/client/Windows/wf_event.c
@@ -214,7 +214,7 @@ void wf_event_focus_in(wfContext* wfc)
 }
 
 static BOOL wf_event_process_WM_MOUSEWHEEL(wfContext* wfc, HWND hWnd, UINT Msg, WPARAM wParam,
-                                           LPARAM lParam, BOOL horizontal, UINT16 x, UINT16 y)
+                                           LPARAM lParam, BOOL horizontal, INT32 x, INT32 y)
 {
 	int delta;
 	UINT16 flags = 0;
@@ -483,11 +483,13 @@ LRESULT CALLBACK wf_event_proc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM lParam
 			case WM_LBUTTONDOWN:
 				wf_scale_mouse_event(wfc, PTR_FLAGS_DOWN | PTR_FLAGS_BUTTON1,
 				                     X_POS(lParam) - wfc->offset_x, Y_POS(lParam) - wfc->offset_y);
+				SetCapture(wfc->hwnd);
 				break;
 
 			case WM_LBUTTONUP:
 				wf_scale_mouse_event(wfc, PTR_FLAGS_BUTTON1, X_POS(lParam) - wfc->offset_x,
 				                     Y_POS(lParam) - wfc->offset_y);
+				ReleaseCapture();
 				break;
 
 			case WM_RBUTTONDOWN:
@@ -816,12 +818,12 @@ BOOL wf_scale_blt(wfContext* wfc, HDC hdc, int x, int y, int w, int h, HDC hdcSr
 	return TRUE;
 }
 
-static BOOL wf_scale_mouse_pos(wfContext* wfc, UINT16* x, UINT16* y)
+static BOOL wf_scale_mouse_pos(wfContext* wfc, INT32 x, INT32 y, UINT16* px, UINT16* py)
 {
 	int ww, wh, dw, dh;
 	rdpSettings* settings;
 
-	if (!wfc || !x || !y)
+	if (!wfc || !px || !py)
 		return FALSE;
 
 	settings = wfc->common.context.settings;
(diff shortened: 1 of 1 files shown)


garbb:
@freerdp-bot test

akallabeth:
@freerdp-bot test
@garbb it does not listen to everyone (would be a nice ddos)

client/Windows/wf_event.c
```diff
@@ -829,6 +831,11 @@ static BOOL wf_scale_mouse_pos(wfContext* wfc, UINT16* x, UINT16* y)
 	if (!settings)
 		return FALSE;
 
+	if (*(INT16*)x < 0)
```
akallabeth, line 834: What is this for? `UINT16` can never be negative, so you effectively check for `x > INT16_MAX`

akallabeth (changes requested):
Remove the code in `wf_scale_mouse_pos`, it is definitely wrong.

[edit] looks like you just forgot to change the function signature of `wf_scale_mouse_pos`, I´ll do that

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth (approved):

garbb:
Seems to still have the problem with the mouse position moving it to the opposite side of the screen when dragged out of left side of window.

https://user-images.githubusercontent.com/4453789/191563255-cddc8e54-6cd3-4ad8-b7ed-b43bbc017eef.mp4

I added debugging
```
if (!wf_scale_mouse_pos(wfc, x, y, &px, &py))
return FALSE;

WLog_DBG("", "wf_scale_mouse_event %d, %d", px, py);

if (freerdp_client_send_button_event(&wfc->common, FALSE, flags, px, py))
return FALSE;
```

x value is signed int being interpreted as unsigned int?
```
[09:40:38:550] [9540:00000458] [DEBUG][] - wf_scale_mouse_event 65370, 262
[09:40:38:558] [9540:00000458] [DEBUG][] - wf_scale_mouse_event 65355, 262
[09:40:38:566] [9540:00000458] [DEBUG][] - wf_scale_mouse_event 65342, 263
[09:40:38:574] [9540:00000458] [DEBUG][] - wf_scale_mouse_event 65331, 265
[09:40:38:582] [9540:00000458] [DEBUG][] - wf_scale_mouse_event 65324, 266
```

But when dragged out of top of window y value is being kept at zero.
```
[09:40:27:456] [9540:00000458] [DEBUG][] - wf_scale_mouse_event 452, 0
[09:40:27:464] [9540:00000458] [DEBUG][] - wf_scale_mouse_event 452, 0
[09:40:27:472] [9540:00000458] [DEBUG][] - wf_scale_mouse_event 452, 0
[09:40:27:480] [9540:00000458] [DEBUG][] - wf_scale_mouse_event 452, 0

```
I don't understand why...

akallabeth:
hmm, you only lock for left button, maybe that is the reason?
so you get events outside the window (negative coordinates?)

garbb:
I think `SetCapture()` captures all mouse inputs to the window until `ReleaseCapture()` ?
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setcapture

> Sets the mouse capture to the specified window belonging to the current thread. SetCapture captures mouse input either when the mouse is over the capturing window, or when the mouse button was pressed while the mouse was over the capturing window and the button is still down. Only one window at a time can capture the mouse.

With more debugging for WM_LBUTTONDOWN, WM_LBUTTONUP, WM_MOUSEMOVE:
```
case WM_LBUTTONDOWN:
	WLog_DBG("WM_LBUTTONDOWN", "%d, %d", X_POS(lParam) - wfc->offset_x, Y_POS(lParam) - wfc->offset_y);
	wf_scale_mouse_event(wfc, PTR_FLAGS_DOWN | PTR_FLAGS_BUTTON1,
						 X_POS(lParam) - wfc->offset_x, Y_POS(lParam) - wfc->offset_y);
	SetCapture(wfc->hwnd);
	break;

case WM_LBUTTONUP:
	WLog_DBG("WM_LBUTTONUP", "%d, %d", X_POS(lParam) - wfc->offset_x, Y_POS(lParam) - wfc->offset_y);
	wf_scale_mouse_event(wfc, PTR_FLAGS_BUTTON1, X_POS(lParam) - wfc->offset_x,
						 Y_POS(lParam) - wfc->offset_y);
	ReleaseCapture();
	break;

case WM_MOUSEMOVE:
	WLog_DBG("WM_MOUSEMOVE", "%d, %d", X_POS(lParam) - wfc->offset_x, Y_POS(lParam) - wfc->offset_y);
	wf_scale_mouse_event(wfc, PTR_FLAGS_MOVE, X_POS(lParam) - wfc->offset_x,
						 Y_POS(lParam) - wfc->offset_y);
```


https://user-images.githubusercontent.com/4453789/191780912-8cee612f-dd52-4751-8513-46a41baa6af7.mp4

So it is getting WM_LBUTTONDOWN, WM_LBUTTONUP, and WM_MOUSEMOVE. But for some reason the Y coord is getting set to zero by wf_scale_mouse_pos() but the X coord is not.

garbb:
@akallabeth
I think X_POS/YPOS macros:
```
#define X_POS(lParam) ((UINT16)(lParam & 0xFFFF))
#define Y_POS(lParam) ((UINT16)((lParam >> 16) & 0xFFFF))
```
are wrong because x and y coordinates are signed values:
https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-mousemove#remarks

Also I think x and y params for wf_scale_mouse_pos, wf_scale_mouse_event, wf_scale_mouse_event_ex should be INT16 instead of INT32 because x and y coordinate values are of type short which is signed 16bit int? (even when compiled as x64 as far as I can tell)

akallabeth:
@garbb Should be fixed with #8247

PR #8232, opened 2022-09-21
Test uri fix

akallabeth:


winpr/include/winpr/winpr.h
```diff
@@ -61,6 +61,20 @@
 #define WINPR_NORETURN(obj) obj
 #endif
 
+#ifdef BUILD_TESTING
```
hardening, line 64: Not sure that everyone will realize that building tests also make local symbols exportable...

hardening: Perhaps we must find a way to have some FreeRDP internal unitary tests that don't rely on the public API...

akallabeth: @hardening that is the reason we export all symbols with debug builds. otherwise we can not access most of the functions we test (maybe we can change the windows build to default to export without a macro?)

winpr/libwinpr/clipboard/clipboard.c
```diff
@@ -611,3 +611,145 @@ void ClipboardDestroy(wClipboard* clipboard)
 	DeleteCriticalSection(&(clipboard->lock));
 	free(clipboard);
 }
+
+static BOOL is_dos_driver(const char* path, size_t len)
```
mfleisz, line 615: Typo? (should be `is_dos_drive` I guess)

channels/client/addin.c
```diff
@@ -212,7 +212,13 @@ static FREERDP_ADDIN** freerdp_channels_list_dynamic_addins(LPCSTR pszName, LPCS
 	NativePathCchAppendA(pszSearchPath, cchSearchPath + 1, pszAddinPath);
 	NativePathCchAppendA(pszSearchPath, cchSearchPath + 1, pszPattern);
 	free(pszPattern);
-	hFind = FindFirstFileA(pszSearchPath, &FindData);
+	{
+		WCHAR* wpath = NULL;
```
mfleisz, line 216: Moving this up (where `FindData` already is defined) and avoiding the separate scope would make the code better readable imo

winpr/include/winpr/winpr.h
```diff
@@ -61,6 +61,20 @@
...
+#define WINPR_LOCAL WINPR_API
+#else
+#if defined _WIN32 || defined __CYGWIN__
```
mfleisz, line 67: Can't this just be
```
#if __GNUC__ >= 4
#define WINPR_LOCAL __attribute__((visibility("hidden")))
#else
#define WINPR_LOCAL
#endif
```
(no need to check for `WIN32` or `CYGWIN`)

mfleisz (commented):
Some smaller comments, functionality lgtm

winpr/include/winpr/winpr.h
akallabeth: actually we only need this for windows as it is not as easy to export everything there. (with gcc/clang you just set global symbol visibility), but on windows you need the dll `__declspec` stuff.

winpr/libwinpr/clipboard/clipboard.c
akallabeth: true.

channels/client/addin.c
akallabeth: ok, moved to a separate function.

winpr/include/winpr/winpr.h
akallabeth: @hardening ok, did update the code to now only use `EXPORT_ALL_SYMBOLS` instead of `BUILD_TESTING`. this way it should be more understandable

winpr/include/winpr/winpr.h
akallabeth: and second, `gcc` is also available on windows, where this does not apply (there you need the `dllexport` stuff with `gcc` too

akallabeth:
@hardening @mfleisz anything left here to fix? I´d like to merge this if possible.

mfleisz (approved):
LGTM

PR #8233, opened 2022-09-21
Load channels and winpr checks

hardening:
Fix the loading of the rdp2tcp channel.
Adds checks for the kind of object for events and threads.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -3629,7 +3629,6 @@ BOOL freerdp_client_load_addins(rdpChannels* channels, rdpSettings* settings)
 		{ FreeRDP_EncomspVirtualChannel, ENCOMSP_SVC_CHANNEL_NAME, settings },
 #endif
 		{ FreeRDP_RemdeskVirtualChannel, REMDESK_SVC_CHANNEL_NAME, settings },
-		{ FreeRDP_RDP2TCPArgs, RDP2TCP_DVC_CHANNEL_NAME, settings->RDP2TCPArgs },
 		{ FreeRDP_RemoteApplicationMode, RAIL_SVC_CHANNEL_NAME, settings }
 	};
 	size_t i;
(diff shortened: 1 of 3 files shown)


client/common/cmdline.c
```diff
@@ -3856,6 +3855,13 @@ BOOL freerdp_client_load_addins(rdpChannels* channels, rdpSettings* settings)
 		}
 	}
 
+	if (settings->RDP2TCPArgs)
+	{
+		if (!freerdp_client_load_static_channel_addin(channels, settings, "rdp2tcp",
```
akallabeth, line 3860: Better use the define `RDP2TCP_DVC_CHANNEL_NAME` for the name?

client/common/cmdline.c
```diff
@@ -3629,7 +3629,6 @@ BOOL freerdp_client_load_addins(rdpChannels* channels, rdpSettings* settings)
 		{ FreeRDP_EncomspVirtualChannel, ENCOMSP_SVC_CHANNEL_NAME, settings },
 #endif
 		{ FreeRDP_RemdeskVirtualChannel, REMDESK_SVC_CHANNEL_NAME, settings },
-		{ FreeRDP_RDP2TCPArgs, RDP2TCP_DVC_CHANNEL_NAME, settings->RDP2TCPArgs },
```
akallabeth, line 3632: Why not just fix this here? `{ 0, RDP2TCP_DVC_CHANNEL_NAME, settings->RDP2TCPArgs },` should work or not?

client/common/cmdline.c
hardening: Oups good catch

client/common/cmdline.c
hardening: Well if I do this the rdp2tcp channel will always be loaded even with NULL settings->RDP2TCPArgs

akallabeth (approved):

PR #8236, opened 2022-09-22
Send resize on window state change

fujames1:
Currently if you connect to a server with `/dynamic-resolution`, resize the window so that it is smaller than the screen, then press `Ctrl+Alt+Enter` to enable full screen mode, the resolution won't be updated.

This PR should fix that issue.  After the window state changes to full screen, it will call `xf_disp_sendResize` to update the resolution.


hardening:

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #8238, opened 2022-09-22
Fixed missing encomsp check

akallabeth:
The channel might not have been initialized and the pointer NULL


pnowack (approved):
Thanks! Fix looks good to me.

PR #8243, opened 2022-09-23
winpr: add a function to export timer file descriptor

hardening:
The equivalent exist for Events, and it happens that sometime you also need to access the internal file descriptor associated with a timer.


akallabeth (approved):
LGTM

PR #8244, opened 2022-09-23
winpr: improve WaitForSingleObject and WaitForMultipleObject with processes

hardening:
This patch improves waiting on process handles:

* under linux we take advantage of pidfd if available, when it can be used it allows to have one or more process in the objects that are WaitForMultipleObject-ed.
* the patch also make the code honor the timeout parameter during a WaitForSingleObject and possibly run completion when WaitForSingleObjectEx is executed with alertable set to true

### DIFF
--- a/winpr/libwinpr/synch/wait.c
+++ b/winpr/libwinpr/synch/wait.c
@@ -190,19 +190,42 @@ DWORD WaitForSingleObjectEx(HANDLE hHandle, DWORD dwMilliseconds, BOOL bAlertabl
 		return WAIT_FAILED;
 	}
 
-	if (Type == HANDLE_TYPE_PROCESS)
+	if (Type == HANDLE_TYPE_PROCESS && winpr_Handle_getFd(hHandle) == -1)
 	{
+		/* note: if we have pidfd support (under linux and we have managed to associate a
+		 * 		pidfd with our process), we use the regular method with pollset below.
+		 * 		If not (on other platforms) we do a waitpid */
 		WINPR_PROCESS* process = (WINPR_PROCESS*)Object;
 
-		if (process->pid != waitpid(process->pid, &(process->status), 0))
+		do
 		{
-			WLog_ERR(TAG, "waitpid failure [%d] %s", errno, strerror(errno));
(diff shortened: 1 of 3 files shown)


winpr/libwinpr/thread/process.c
```diff
@@ -478,6 +487,11 @@ BOOL TerminateProcess(HANDLE hProcess, UINT uExitCode)
 static BOOL ProcessHandleCloseHandle(HANDLE handle)
 {
 	WINPR_PROCESS* process = (WINPR_PROCESS*)handle;
+	if (process->fd >= 0)
```
akallabeth, line 490: I think you should first check if the handle is valid?

hardening: I can add an assert, but when we're here the handle has been dereferenced multiple time by caller code

winpr/libwinpr/synch/wait.c
```diff
@@ -190,19 +190,42 @@ DWORD WaitForSingleObjectEx(HANDLE hHandle, DWORD dwMilliseconds, BOOL bAlertabl
...
+			{
+				WLog_ERR(TAG, "waitpid failure [%d] %s", errno, strerror(errno));
+				SetLastError(ERROR_INTERNAL_ERROR);
+				return WAIT_FAILED;
+			}
 
-		process->dwExitCode = (DWORD)process->status;
-		return WAIT_OBJECT_0;
+			/* sleep by slices of 50ms */
+			status = SleepEx(50, bAlertable);
```
akallabeth, line 217: small issue, but what if the `dwMilliseconds` starts with a value `<50`?

akallabeth (approved):
Overall LGTM.

just the time slots for sleeping might be nicer if the actual value is checked before just waiting 50ms

winpr/libwinpr/synch/wait.c
hardening: hum yeah will change that

hardening:
Note: the same could be achieved on BSD systems by calling `pdfork` instead of `fork`.

akallabeth (approved):

PR #8251, opened 2022-09-27
Support for windows credential manager when username and password omitted

StefanKoell:
* added wf_defaults.c (also in CMakeLists.txt) which provides the capability to read credential manager entries and set username and password in the rdpSettings accordingly
* the same will be done if a gateway host is specified without credentials
* the behavior will be similar to mstsc.exe on windows

To test the implementation, create a new "Generic Credential" in the Windows Credential Manager and use the following values:
**Internet or network address:** TERMSRV/[your RDP server name or IP]
**User name:** the username to log on to the RDP server
**Password:** the password to log on to the RDP server

Feature PR discussion:
https://matrix.to/#/!mnXPZoPpWIGqYeWkOf:matrix.org/$Is6wfkc_qDECcMvvUydVkIm21I6MxIBbkrZkJxL_Z1A?via=matrix.org&via=tchncs.de&via=windcorp.ru

### DIFF
--- a/client/Windows/wf_defaults.c
+++ b/client/Windows/wf_defaults.c
@@ -2,6 +2,7 @@
 #include <wincred.h>
 #include <stdio.h>
 #include <malloc.h>
+#include <freerdp/settings.h>
 
 #pragma warning(disable : 4706 4094) // assignment within conditional expression
 
@@ -14,21 +15,18 @@
 
 typedef struct rdp_settings rdpSettings;
 
-EXTERN_C_START
-
-const char* freerdp_settings_get_string(const rdpSettings* settings, size_t id);
-BOOL freerdp_settings_set_string(rdpSettings* settings, size_t id, const char* param);
-
-EXTERN_C_END
-
 PCWSTR ValidateString(const BYTE* pb, ULONG cb)
 {
-	return pb &&                                       // present ?
-	               cb &&                               // size not 0 ?
-	               !(cb & (sizeof(WCHAR) - 1)) &&      // cb == n * sizeof(WCHAR) ?
-	               !*(WCHAR*)(pb + cb - sizeof(WCHAR)) // 0 terminated ?
-	           ? (PCWSTR)pb
-	           : 0;
+	if (!pb || !cb)
+		return 0;
+
+	if (cb & (sizeof(WCHAR) - 1))
+		return 0;
+
+	if (*(WCHAR*)(pb + cb - sizeof(WCHAR)))
+		return 0;
+
+	return (PCWSTR)pb;
 }
 
 void AddDefaultSettings_I(rdpSettings* settings, size_t idHostname, size_t idUsername,
@@ -37,17 +35,13 @@ void AddDefaultSettings_I(rdpSettings* settings, size_t idHostname, size_t idUse
 	PCSTR ServerHostname = freerdp_settings_get_string(settings, idHostname);
 
 	if (!ServerHostname)
-	{
 		return;
-	}
 
-	BOOL bExistUserName = 0 != freerdp_settings_get_string(settings, idUsername);
-	BOOL bExistPassword = 0 != freerdp_settings_get_string(settings, idPassword);
+	BOOL bExistUserName = freerdp_settings_get_string(settings, idUsername) != 0;
+	BOOL bExistPassword = freerdp_settings_get_string(settings, idPassword) != 0;
 
 	if (bExistUserName && bExistPassword)
-	{
 		return;
-	}
 
 	PCREDENTIALA Credential;
 	PCWSTR lpWideCharStr;
@@ -56,57 +50,51 @@ void AddDefaultSettings_I(rdpSettings* settings, size_t idHostname, size_t idUse
 	PSTR UserName;
 
 	PSTR TargetName = 0;
-	int len = 0;
+	int len = _vsnprintf(TargetName, 0, "TERMSRV/%s", (va_list)&ServerHostname);
+	len++;
+	TargetName = (PSTR)malloc(len);
+	_vsnprintf(TargetName, len, "TERMSRV/%s", (va_list)&ServerHostname);
 
-	while (0 < (len = _vsnprintf(TargetName, len, "TERMSRV/%s", (va_list)&ServerHostname)))
+	if (!TargetName)
+		return;
+
+	TargetName[len - 1] = 0;
+
+	if (!CredReadA(TargetName, CRED_TYPE_GENERIC, 0, &Credential))
+		return;
+
+	if (!bExistPassword)
 	{
-		if (TargetName)
+		ULONG cch = Credential->CredentialBlobSize;
+
+		if (lpWideCharStr = ValidateString(Credential->CredentialBlob, cch))
 		{
-			if (CredReadA(TargetName, CRED_TYPE_GENERIC, 0, &Credential))
+			Password = 0, len = 0, cch /= sizeof(WCHAR);
+
+			len = WideCharToMultiByte(CP_UTF8, 0, lpWideCharStr, cch, Password, len, 0, 0);
+			Password = (PSTR)malloc(len);
+			WideCharToMultiByte(CP_UTF8, 0, lpWideCharStr, cch, Password, len, 0, 0);
+			
+			if (Password)
 			{
-				if (!bExistPassword)
-				{
-					ULONG cch = Credential->CredentialBlobSize;
-
-					if (lpWideCharStr = ValidateString(Credential->CredentialBlob, cch))
-					{
-						Password = 0, len = 0, cch /= sizeof(WCHAR);
-
-						while (len = WideCharToMultiByte(CP_UTF8, 0, lpWideCharStr, cch, Password,
-						                                 len, 0, 0))
-						{
-							if (Password)
-							{
-								freerdp_settings_set_string(settings, idPassword, Password);
-								break;
-							}
-
-							Password = (PSTR)alloca(len);
-						}
-					}
-				}
-
-				if (!bExistUserName)
-				{
-					if (UserName = Credential->UserName)
-					{
-						freerdp_settings_set_string(settings, idUsername, UserName);
-					}
-				}
-
-				CredFree(Credential);
+				freerdp_settings_set_string(settings, idPassword, Password);
 			}
-
-			break;
 		}
-
-		TargetName = (PSTR)alloca(++len);
 	}
+
+	if (!bExistUserName)
+	{
+		if (UserName = Credential->UserName)
+		{
+			freerdp_settings_set_string(settings, idUsername, UserName);
+		}
+	}
+
+	CredFree(Credential);
 }
 
 EXTERN_C void WINAPI AddDefaultSettings(_Inout_ rdpSettings* settings)
 {
 	AddDefaultSettings_I(settings, FreeRDP_ServerHostname, FreeRDP_Username, FreeRDP_Password);
-	AddDefaultSettings_I(settings, FreeRDP_GatewayHostname, FreeRDP_GatewayUsername,
-	                     FreeRDP_GatewayPassword);
+	AddDefaultSettings_I(settings, FreeRDP_GatewayHostname, FreeRDP_GatewayUsername, FreeRDP_GatewayPassword);
 }
--- a/client/Windows/wf_defaults.c
+++ b/client/Windows/wf_defaults.c
@@ -1,0 +1,112 @@
+#include <Windows.h>
+#include <wincred.h>
+#include <stdio.h>
+#include <malloc.h>
+
+#pragma warning(disable : 4706 4094) // assignment within conditional expression
+
+#define FreeRDP_ServerHostname (20)
+#define FreeRDP_Username (21)
+#define FreeRDP_Password (22)
+#define FreeRDP_GatewayHostname (1986)
+#define FreeRDP_GatewayUsername (1987)
+#define FreeRDP_GatewayPassword (1988)
+
+typedef struct rdp_settings rdpSettings;
+
+EXTERN_C_START
+
+const char* freerdp_settings_get_string(const rdpSettings* settings, size_t id);
+BOOL freerdp_settings_set_string(rdpSettings* settings, size_t id, const char* param);
+
+EXTERN_C_END
+
+PCWSTR ValidateString(const BYTE* pb, ULONG cb)
+{
+	return pb &&                                       // present ?
+	               cb &&                               // size not 0 ?
+	               !(cb & (sizeof(WCHAR) - 1)) &&      // cb == n * sizeof(WCHAR) ?
+	               !*(WCHAR*)(pb + cb - sizeof(WCHAR)) // 0 terminated ?
+	           ? (PCWSTR)pb
+	           : 0;
+}
+
+void AddDefaultSettings_I(rdpSettings* settings, size_t idHostname, size_t idUsername,
+                          size_t idPassword)
+{
+	PCSTR ServerHostname = freerdp_settings_get_string(settings, idHostname);
+
+	if (!ServerHostname)
+	{
+		return;
+	}
+
+	BOOL bExistUserName = 0 != freerdp_settings_get_string(settings, idUsername);
+	BOOL bExistPassword = 0 != freerdp_settings_get_string(settings, idPassword);
+
+	if (bExistUserName && bExistPassword)
+	{
+		return;
+	}
+
+	PCREDENTIALA Credential;
+	PCWSTR lpWideCharStr;
+
+	PSTR Password;
+	PSTR UserName;
+
+	PSTR TargetName = 0;
+	int len = 0;
+
+	while (0 < (len = _vsnprintf(TargetName, len, "TERMSRV/%s", (va_list)&ServerHostname)))
+	{
+		if (TargetName)
+		{
+			if (CredReadA(TargetName, CRED_TYPE_GENERIC, 0, &Credential))
+			{
+				if (!bExistPassword)
+				{
+					ULONG cch = Credential->CredentialBlobSize;
+
+					if (lpWideCharStr = ValidateString(Credential->CredentialBlob, cch))
+					{
+						Password = 0, len = 0, cch /= sizeof(WCHAR);
+
+						while (len = WideCharToMultiByte(CP_UTF8, 0, lpWideCharStr, cch, Password,
+						                                 len, 0, 0))
+						{
+							if (Password)
+							{
+								freerdp_settings_set_string(settings, idPassword, Password);
+								break;
+							}
+
+							Password = (PSTR)alloca(len);
+						}
+					}
+				}
+
+				if (!bExistUserName)
+				{
+					if (UserName = Credential->UserName)
+					{
+						freerdp_settings_set_string(settings, idUsername, UserName);
+					}
+				}
+
+				CredFree(Credential);
+			}
+
+			break;
+		}
+
+		TargetName = (PSTR)alloca(++len);
+	}
+}
+
+EXTERN_C void WINAPI AddDefaultSettings(_Inout_ rdpSettings* settings)
+{
+	AddDefaultSettings_I(settings, FreeRDP_ServerHostname, FreeRDP_Username, FreeRDP_Password);
+	AddDefaultSettings_I(settings, FreeRDP_GatewayHostname, FreeRDP_GatewayUsername,
+	                     FreeRDP_GatewayPassword);
+}
--- a/client/Windows/wf_defaults.c
+++ b/client/Windows/wf_defaults.c
@@ -4,18 +4,7 @@
 #include <malloc.h>
 #include <freerdp/settings.h>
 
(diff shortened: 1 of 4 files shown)


client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,112 @@
...
+#define FreeRDP_Password (22)
+#define FreeRDP_GatewayHostname (1986)
+#define FreeRDP_GatewayUsername (1987)
+#define FreeRDP_GatewayPassword (1988)
+
+typedef struct rdp_settings rdpSettings;
+
+EXTERN_C_START
+
+const char* freerdp_settings_get_string(const rdpSettings* settings, size_t id);
```
akallabeth, line 19: why not simply `#include <freerdp/settings.h>`?

akallabeth:
@freerdp-bot test

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,112 @@
...
+	PCREDENTIALA Credential;
+	PCWSTR lpWideCharStr;
+
+	PSTR Password;
+	PSTR UserName;
+
+	PSTR TargetName = 0;
+	int len = 0;
+
+	while (0 < (len = _vsnprintf(TargetName, len, "TERMSRV/%s", (va_list)&ServerHostname)))
```
akallabeth, line 61: shouldn´t this simply be `_vsnprintf(TargetName, len, "TERMSRV/%s", ServerHostname)` ?

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,112 @@
...
+					}
+				}
+
+				CredFree(Credential);
+			}
+
+			break;
+		}
+
+		TargetName = (PSTR)alloca(++len);
```
akallabeth, line 103: `alloca` is discouraged. 
urls have limited length anyway, so just throw it on the stack?
https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

client/Windows/wf_defaults.c
hardening: Yeah I had to re-read it twice to see what was checked

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,112 @@
...
+
+	PSTR Password;
+	PSTR UserName;
+
+	PSTR TargetName = 0;
+	int len = 0;
+
+	while (0 < (len = _vsnprintf(TargetName, len, "TERMSRV/%s", (va_list)&ServerHostname)))
+	{
+		if (TargetName)
```
hardening, line 63: Can we avoid the if / then /else staircase by inverting conditions / use break/continue ?

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,112 @@
...
+BOOL freerdp_settings_set_string(rdpSettings* settings, size_t id, const char* param);
+
+EXTERN_C_END
+
+PCWSTR ValidateString(const BYTE* pb, ULONG cb)
+{
+	return pb &&                                       // present ?
+	               cb &&                               // size not 0 ?
+	               !(cb & (sizeof(WCHAR) - 1)) &&      // cb == n * sizeof(WCHAR) ?
+	               !*(WCHAR*)(pb + cb - sizeof(WCHAR)) // 0 terminated ?
```
hardening, line 29: Can you change that to a if / then / else instead, it's way too complicated to read...

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,112 @@
...
+                          size_t idPassword)
+{
+	PCSTR ServerHostname = freerdp_settings_get_string(settings, idHostname);
+
+	if (!ServerHostname)
+	{
+		return;
+	}
+
+	BOOL bExistUserName = 0 != freerdp_settings_get_string(settings, idUsername);
```
hardening, line 44: No variable declaration in the middle of the block

hardening:
Sounds promising.
General remark: AFAICT we don't use the JAVA code style like `0 != <expression>`, we always use  `<expression> != 0`, would be nice to keep the style coherent

client/Windows/wf_defaults.c
StefanKoell: good point!

client/Windows/wf_defaults.c
StefanKoell: hmm, doesn't work or maybe I'm doing something wrong? how would you rewrite that code without the while?

StefanKoell: I should mention that I'm not really a C dev. Any help would be appreciated.

StefanKoell: removing the while I end up with something like this:

```C
	PSTR TargetName = 0;
	int len = _vsnprintf(TargetName, 0, "TERMSRV/%s", (va_list)&ServerHostname);
	TargetName = (PSTR)malloc(++len);
	_vsnprintf(TargetName, len, "TERMSRV/%s", (va_list)&ServerHostname);

```

is this the preferred way? (btw, I failed using _snprintf)

client/Windows/wf_defaults.c
akallabeth: @hardening no longer strict here, changed the `C` standard recently ;)

hardening: nice to know !

client/Windows/wf_defaults.c
StefanKoell: something like that?

```C
PCWSTR ValidateString(const BYTE* pb, ULONG cb)
{
	if (!pb || !cb)
		return 0;

	if (!(cb & (sizeof(WCHAR) - 1)))
		return 0;

	if (!*(WCHAR*)(pb + cb - sizeof(WCHAR)))
		return 0;

	return (PCWSTR)pb;
}
```

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,100 @@
+#include <Windows.h>
+#include <wincred.h>
+#include <stdio.h>
+#include <malloc.h>
+#include <freerdp/settings.h>
+
+#pragma warning(disable : 4706 4094) // assignment within conditional expression
```
akallabeth, line 7: Don´t disable the warning, initialize the value.

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,100 @@
...
+
+#define FreeRDP_ServerHostname (20)
```
akallabeth, line 9: already part of the included header?

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,100 @@
...
+#define FreeRDP_Username (21)
+#define FreeRDP_Password (22)
+#define FreeRDP_GatewayHostname (1986)
+#define FreeRDP_GatewayUsername (1987)
+#define FreeRDP_GatewayPassword (1988)
+
+typedef struct rdp_settings rdpSettings;
```
akallabeth, line 16: already part of the included header?

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,100 @@
...
+
+PCWSTR ValidateString(const BYTE* pb, ULONG cb)
```
akallabeth, line 18: should be `static`

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,100 @@
...
+	if (cb & (sizeof(WCHAR) - 1))
+		return 0;
+
+	if (*(WCHAR*)(pb + cb - sizeof(WCHAR)))
+		return 0;
+
+	return (PCWSTR)pb;
+}
+
+void AddDefaultSettings_I(rdpSettings* settings, size_t idHostname, size_t idUsername,
```
akallabeth, line 32: should be `static`

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,100 @@
...
+		if (UserName = Credential->UserName)
+		{
+			freerdp_settings_set_string(settings, idUsername, UserName);
+		}
+	}
+
+	CredFree(Credential);
+}
+
+EXTERN_C void WINAPI AddDefaultSettings(_Inout_ rdpSettings* settings)
```
akallabeth, line 96: No need to export the function, only used in the binary itself (not in an external API)
but creating a header file for that (which is included where it is used) is highly recommended. (that way the compiler knows properly of which type arguments and return are)

akallabeth (changes requested):
Some changes still required

StefanKoell:
Thanks for your patience and all the comments. Learning a lot here. I'm not sure about the header file rules, so I did my best by looking at other files ;)

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,89 @@
...
+	if (!CredReadA(TargetName, CRED_TYPE_GENERIC, 0, &Credential))
+		return;
+
+	if (!bExistPassword)
+	{
+		ULONG cch = Credential->CredentialBlobSize;
+
+		if (lpWideCharStr = ValidateString(Credential->CredentialBlob, cch))
+		{
+			Password = 0, len = 0, cch /= sizeof(WCHAR);
```
hardening, line 61: We never use that code style in FreeRDP, can you change that to one statement per line

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,89 @@
...
+
+			len = WideCharToMultiByte(CP_UTF8, 0, lpWideCharStr, cch, Password, len, 0, 0);
+			Password = (PSTR)malloc(len);
```
hardening, line 64: the result of `malloc` shall be checked

client/Windows/wf_defaults.h
```diff
@@ -0,0 +1,8 @@
+#ifndef FREERDP_CLIENT_WIN_DEFAULTS_H
+#define FREERDP_CLIENT_WIN_DEFAULTS_H
+
+#include <freerdp/settings.h>
+
+void WINAPI AddDefaultSettings(_Inout_ rdpSettings* settings);
```
hardening, line 6: In FreeRDP we don't use the windows `_Inout_`  modifiers for parameters

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,89 @@
...
+		return;
+
+	PCREDENTIALA Credential;
+	PCWSTR lpWideCharStr;
+
+	PSTR Password;
+	PSTR UserName;
+
+	PSTR TargetName = 0;
+	int len = _vsnprintf(TargetName, 0, "TERMSRV/%s", (va_list)&ServerHostname);
```
akallabeth, line 42: this is wrong, use `_snprintf`  (or what is current on windows) if you want variable arguments or create a proper `va_list`

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,89 @@
...
+	len++;
+	TargetName = (PSTR)malloc(len);
+	_vsnprintf(TargetName, len, "TERMSRV/%s", (va_list)&ServerHostname);
+
+	if (!TargetName)
+		return;
+
+	TargetName[len - 1] = 0;
+
+	if (!CredReadA(TargetName, CRED_TYPE_GENERIC, 0, &Credential))
```
akallabeth, line 52: Might be a good idea to use `CredReadW` to get unicode strings for `user` et al properly

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,89 @@
...
+
+	if (!bExistPassword)
+	{
+		ULONG cch = Credential->CredentialBlobSize;
+
+		if (lpWideCharStr = ValidateString(Credential->CredentialBlob, cch))
+		{
+			Password = 0, len = 0, cch /= sizeof(WCHAR);
+
+			len = WideCharToMultiByte(CP_UTF8, 0, lpWideCharStr, cch, Password, len, 0, 0);
```
akallabeth, line 63: We have `ConvertFromUnicode` wrapper function in `WinPR` for doing this allocating conversion.

client/Windows/wf_defaults.h
```diff
@@ -0,0 +1,8 @@
+#ifndef FREERDP_CLIENT_WIN_DEFAULTS_H
```
akallabeth, line 1: You might add the copyright and license header with your name here ;)
see other files for examples

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,89 @@
+#include <Windows.h>
```
akallabeth, line 1: you might want to add the license and copyright headers here

client/Windows/wf_defaults.c
StefanKoell: @akallabeth can you help me out here? when I change it to _snprintf, it will not work. The resulting string is completely messed up:

```C
	PSTR TargetName = 0;
	int len = _snprintf(TargetName, 0, "TERMSRV/%s", &ServerHostname);
	len++;
	TargetName = (PSTR)malloc(len);

	if (!TargetName)
		return;

	_snprintf(TargetName, len, "TERMSRV/%s", &ServerHostname);
```

TargetName is then `TERMSRV/ðŽEÎ`

mfleisz: @StefanKoell you must not specify the address of the string (`&`) but the string itself in the argument list:

`int len = _snprintf(TargetName, 0, "TERMSRV/%s", ServerHostname);`

StefanKoell: Just tried that but I don't get a valid/correct string in this case. I'm surely missing something.

StefanKoell: Please disregard my last comment. It is working and I'm an idiot ;)

StefanKoell:
@akallabeth @hardening @mfleisz with a little help from a friend I was able to rewrite it to CredReadW. Please check if this is acceptable this way. Thanks again!

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ *
+ * Copyright 2014 Marc-Andre Moreau <marcandre.moreau@gmail.com>
```
akallabeth, line 4: well, I think you don´t need to have Marc-Andre in here, you´ve written that yourself, didn´t you ;)

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
...
+	static const PSTR TERMSRV = "TERMSRV/%s";
+
+	PSTR TargetName = 0;
+	PSTR UserName = 0;
+	PSTR Password = 0;
+	PWSTR UserNameW = 0;
+	PWSTR PasswordW = 0;
+	PWSTR TargetNameW = 0;
+	PWSTR ServerHostnameW = 0;
+	PCREDENTIALW Credential;
```
akallabeth, line 52: best have that initialized with `PCREDENTIALW Credential = { 0 };`

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
...
+
+	return (PCWSTR)pb;
+}
+
+static void AddDefaultSettings_I(rdpSettings* settings, size_t idHostname, size_t idUsername,
+                                 size_t idPassword)
+{
+	static const PSTR TERMSRV = "TERMSRV/%s";
+
+	PSTR TargetName = 0;
```
akallabeth, line 45: While actually the same, `PSTR` is a pointer type, so better initialize with `NULL` (not blocking, just a remark)

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
...
+	int len = _snprintf(TargetName, 0, TERMSRV, ServerHostname);
+	len++;
+	TargetName = (PSTR)malloc(len);
+
+	if (!TargetName)
+		return;
+
+	_snprintf(TargetName, len, TERMSRV, ServerHostname);
+
+	if (!TargetName)
```
akallabeth, line 74: check here no longer required, you already checked allocation success above.

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
...
+
+	BOOL bExistUserName = freerdp_settings_get_string(settings, idUsername) != 0;
+	BOOL bExistPassword = freerdp_settings_get_string(settings, idPassword) != 0;
+
+	if (bExistUserName && bExistPassword)
+		return;
+
+	int len = _snprintf(TargetName, 0, TERMSRV, ServerHostname);
+	len++;
+	TargetName = (PSTR)malloc(len);
```
akallabeth, line 67: `calloc(len, sizeof(char))` will do the same, but initialize memory to `\0`

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
...
+	if (!TargetName)
+		return;
+
+	TargetName[len - 1] = 0;
+
+	int result = ConvertToUnicode(CP_UTF8, 0, TargetName, -1, &TargetNameW, 0);
+	if (!result)
+		return;
+
+	if (!CredReadW(TargetNameW, CRED_TYPE_GENERIC, 0, &Credential))
```
akallabeth, line 83: you need to handle the allocated resources on return (e.g. call `free` on them)

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
...
+
+	if (!bExistUserName)
+	{
+		UserNameW = Credential->UserName;
+
+		if (UserNameW)
+		{
+			result = ConvertFromUnicode(CP_UTF8, 0, UserNameW, -1, &UserName, 0, NULL, NULL);
+			if (result)
+				freerdp_settings_set_string(settings, idUsername, UserName);
```
akallabeth, line 106: you need to free up `UserName`

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
...
+	if (!CredReadW(TargetNameW, CRED_TYPE_GENERIC, 0, &Credential))
+		return;
+
+	if (!bExistPassword)
+	{
+		PasswordW = ValidateString(Credential->CredentialBlob, Credential->CredentialBlobSize);
+
+		if (PasswordW)
+		{
+			result = ConvertFromUnicode(CP_UTF8, 0, PasswordW, -1, &Password, 0, NULL, NULL);
```
akallabeth, line 92: You need to free up `Password`

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
...
+
+	_snprintf(TargetName, len, TERMSRV, ServerHostname);
+
+	if (!TargetName)
+		return;
+
+	TargetName[len - 1] = 0;
+
+	int result = ConvertToUnicode(CP_UTF8, 0, TargetName, -1, &TargetNameW, 0);
+	if (!result)
```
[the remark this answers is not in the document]
akallabeth: you also need to free `TargetNameW` later on

client/Windows/wf_defaults.c
```diff
@@ -0,0 +1,118 @@
...
+	if (!ServerHostname)
+		return;
+
+	BOOL bExistUserName = freerdp_settings_get_string(settings, idUsername) != 0;
+	BOOL bExistPassword = freerdp_settings_get_string(settings, idPassword) != 0;
+
+	if (bExistUserName && bExistPassword)
+		return;
+
+	int len = _snprintf(TargetName, 0, TERMSRV, ServerHostname);
```
akallabeth, line 65: you need to free `TargetName` later on

client/Windows/wf_defaults.h
```diff
@@ -0,0 +1,27 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ *
+ * Copyright 2014 Marc-Andre Moreau <marcandre.moreau@gmail.com>
```
akallabeth, line 4: I think this was not written by Marc-Andre?

akallabeth (changes requested):
You lack some resource cleanups.
easiest is the pattern we use for that:
```
char* str = NULL;

... somecode

str = malloc(42);
if (!str)
  goto fail;

... some code

fail:
free(str);
return;
}

client/Windows/wf_defaults.c
StefanKoell: right!

client/Windows/wf_defaults.c
akallabeth: `PSTR TargetName = NULL;` (`NULL` is for pointers, value is (most of the time) same as 0)

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@StefanKoell did take some time, but now it is ready. Thank you for the hard work!

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

StefanKoell:
Learned a lot. Thanks for all the feedback and help. I will create another PR with the suggestion from @awakecoding to use the environment variable for the TERMSRV/ prefix. Let's see if my lessons learned will make this quicker ;)

PR #8255, opened 2022-09-28
More winpr fixes

hardening:


mfleisz (approved):
LGTM

client/Windows/wf_event.c
```diff
@@ -34,7 +34,7 @@
 
 #include <freerdp/event.h>
 
-#include <Windowsx.h>
+#include <windowsx.h>
```
akallabeth, line 37: @hardening ok, looks like mingw has different header (capitialization) than windows itself?

PR #8260, opened 2022-09-29
proxy: Fix crash due to format specifier/type mismatch in debug logging

mfleisz:
The `cmd` parameter has been logged using the `%s` format specifier causing an access violation.

Fixed by using `get_packet_type` to convert `cmd` to a string.


akallabeth (commented):
LGTM,
good catch, looks like I missed that :/

akallabeth (approved):

PR #8261, opened 2022-09-29
Add SSPI SetCredentialsAttributes missing definitions + Kerberos KDC URL custom attribute

awakecoding:
This pull request adds the required SSPI plumbing for SetCredentialsAttributes calls along with missing SECPKG_CRED_ATTR IDs and structure definitions. It also adds a custom SECPKG_CRED_ATTR_KDC_URL extended attribute meant to pass the KDC URL from FreeRDP settings down to the Kerberos SSPI module (--kdc-url). Windows SSPI only supports an explicit KDC proxy server configuration through SECPKG_CRED_ATTR_KDC_PROXY_SETTINGS, and it uses a weird, error-prone format. The URL format can encode all KDC server "types" nicely, where we can assume a default protocol (TCP) and port (88), like this:

- IT-HELP-DC.ad.it-help.ninja
- IT-HELP-DC.ad.it-help.ninja:88
- tcp://IT-HELP-DC.ad.it-help.ninja:88
- udp://IT-HELP-DC.ad.it-help.ninja:88
- http://kdc.ad.it-help.ninja/KdcProxy
- https://kdc.ad.it-help.ninja:4343/KdcProxy

This KDC URL can then be parsed and injected into Kerberos SSPI modules that support it. For now, the WinPR Kerberos SSPI module just throws a warning because it doesn't yet support an explicit KDC server configuration, but we can implement it in a separate pull request using [krb5_init_context_profile](https://web.mit.edu/kerberos/krb5-devel/doc/plugindev/profile.html) to create an in-memory profile without a krb5.conf file on disk.

The KDC proxy settings injection through SECPKG_CRED_ATTR_KDC_PROXY_SETTINGS is also left up to a future pull request for support on Windows, alongside support for parsing KDCProxyName from .RDP file settings. It will also requires URL to "Name" format conversions helpers (https://kdc.ad.it-help.ninja:4343/KdcProxy -> kdc.ad.it-help.ninja:4343:KdcProxy, yes, that's really a ':' instead of '/', it's a cursed format).

This pull request is of high importance to me because it unblocks proper FreeRDP settings to inject a KDC URL when used with my Rust SSPI modules. [sspi-rs progresses well](https://github.com/Devolutions/sspi-rs/pull/21), and we've got it built for Windows, macOS, Linux, Android and iOS, and we're basically just improving KDC detection methods now + adding ways to explicit inject the KDC information on a per-connection basis, within the same process.

### DIFF
--- a/client/common/cmdline.h
+++ b/client/common/cmdline.h
@@ -224,6 +224,7 @@ static const COMMAND_LINE_ARGUMENT_A global_cmd_args[] = {
 	{ "kbd-unicode", COMMAND_LINE_VALUE_FLAG, "", NULL, NULL, -1, NULL,
 	  "Send unicode symbols, e.g. use the local keyboard map. ATTENTION: Does not work with every "
 	  "RDP server!" },
+	{ "kdc-url", COMMAND_LINE_VALUE_REQUIRED, "<url>", NULL, NULL, -1, NULL, "KDC server URL" },
 	{ "kerberos", COMMAND_LINE_VALUE_REQUIRED,
 	  "[lifetime:<time>,start-time:<time>,renewable-lifetime:<time>,cache:<path>,armor:<path>,"
 	  "pkinit-anchors:<path>,pkcs11-module:<name>]",
--- a/client/common/cmdline.h
+++ b/client/common/cmdline.h
@@ -224,7 +224,6 @@ static const COMMAND_LINE_ARGUMENT_A global_cmd_args[] = {
 	{ "kbd-unicode", COMMAND_LINE_VALUE_FLAG, "", NULL, NULL, -1, NULL,
 	  "Send unicode symbols, e.g. use the local keyboard map. ATTENTION: Does not work with every "
 	  "RDP server!" },
-	{ "kdc-url", COMMAND_LINE_VALUE_REQUIRED, "<url>", NULL, NULL, -1, NULL, "KDC server URL" },
 	{ "kerberos", COMMAND_LINE_VALUE_REQUIRED,
 	  "[lifetime:<time>,start-time:<time>,renewable-lifetime:<time>,cache:<path>,armor:<path>,"
 	  "pkinit-anchors:<path>,pkcs11-module:<name>]",
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -946,6 +946,7 @@ static BOOL nla_setup_kerberos(rdpNla* nla)
 
 static BOOL nla_client_init_cred_handle(rdpNla* nla)
 {
+	SECURITY_STATUS secStatus;
 	SEC_WINPR_KERBEROS_SETTINGS* kerbSettings;
 
 	WINPR_ASSERT(nla);
@@ -962,16 +963,21 @@ static BOOL nla_client_init_cred_handle(rdpNla* nla)
 		if (!secAttr.KdcUrl)
 			return FALSE;
 
-		nla->table->SetCredentialsAttributesW(&nla->credentials, SECPKG_CRED_ATTR_KDC_URL,
-		                                      (void*)&secAttr, sizeof(secAttr));
+		secStatus = nla->table->SetCredentialsAttributesW(
+		    &nla->credentials, SECPKG_CRED_ATTR_KDC_URL, (void*)&secAttr, sizeof(secAttr));
 
 		free(secAttr.KdcUrl);
 #else
 		SecPkgCredentials_KdcUrlA secAttr = { NULL };
 		secAttr.KdcUrl = kerbSettings->kdcUrl;
-		nla->table->SetCredentialsAttributesA(&nla->credentials, SECPKG_CRED_ATTR_KDC_URL,
-		                                      (void*)&secAttr, sizeof(secAttr));
+		secStatus = nla->table->SetCredentialsAttributesA(
+		    &nla->credentials, SECPKG_CRED_ATTR_KDC_URL, (void*)&secAttr, sizeof(secAttr));
 #endif
+		if (secStatus != SEC_E_OK)
+		{
+			WLog_WARN(TAG, "Explicit Kerberos KDC URL (%s) injection is not supported",
+			          kerbSettings->kdcUrl);
+		}
 	}
 
 	return TRUE;
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -958,12 +958,14 @@ static BOOL nla_client_init_cred_handle(rdpNla* nla)
 #ifdef UNICODE
 		SecPkgCredentials_KdcUrlW secAttr = { NULL };
 		ConvertToUnicode(CP_UTF8, 0, kerbSettings->kdcUrl, -1, &secAttr.KdcUrl, 0);
-		nla->table->SetCredentialsAttributesW(&nla->credentials, SECPKG_CRED_ATTR_KDC_URL, (void*) &secAttr, sizeof(secAttr));
+		nla->table->SetCredentialsAttributesW(&nla->credentials, SECPKG_CRED_ATTR_KDC_URL,
+		                                      (void*)&secAttr, sizeof(secAttr));
 		free(secAttr.KdcUrl);
 #else
 		SecPkgCredentials_KdcUrlA secAttr = { NULL };
 		secAttr.KdcUrl = kerbSettings->kdcUrl;
-		nla->table->SetCredentialsAttributesA(&nla->credentials, SECPKG_CRED_ATTR_KDC_URL, (void*) &secAttr, sizeof(secAttr));
+		nla->table->SetCredentialsAttributesA(&nla->credentials, SECPKG_CRED_ATTR_KDC_URL,
+		                                      (void*)&secAttr, sizeof(secAttr));
 #endif
 	}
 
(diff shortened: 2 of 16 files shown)


libfreerdp/core/nla.c
```diff
@@ -1010,6 +1053,8 @@ static int nla_client_init(rdpNla* nla)
 		return -1;
 	}
 
+	nla_client_init_cred_handle(nla);
```
hardening, line 1056: Shouldn't we fail if it returns False ?

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -1224,6 +1227,91 @@ static SECURITY_STATUS SEC_ENTRY kerberos_QueryContextAttributesW(PCtxtHandle ph
...
+			credentials->kdc_url = NULL;
+		}
+
+		if (unicode)
+		{
+			SEC_WCHAR* KdcUrl = ((SecPkgCredentials_KdcUrlW*)pBuffer)->KdcUrl;
+
+			if (KdcUrl)
+			{
+				ConvertFromUnicode(CP_UTF8, 0, (WCHAR*)KdcUrl, -1, &credentials->kdc_url, 0, NULL,
```
hardening, line 1277: result shall be checked in case of OOM

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -1202,6 +1202,64 @@ static SECURITY_STATUS SEC_ENTRY negotiate_SetContextAttributesA(PCtxtHandle phC
...
+		MechCred* cred = &creds[i];
+
+		if (!cred->valid)
+			continue;
+
+		WINPR_ASSERT(cred->mech);
+		WINPR_ASSERT(cred->mech->pkg);
+		WINPR_ASSERT(cred->mech->pkg->table);
+		WINPR_ASSERT(cred->mech->pkg->table_w->SetCredentialsAttributesW);
+		cred->mech->pkg->table_w->SetCredentialsAttributesW(&cred->cred, ulAttribute, pBuffer,
```
hardening, line 1227: `negotiate_SetCredentialsAttributesW` and `negotiate_SetCredentialsAttributesA` return Success whatever happened when calling `creds->SetCredentialsAttributesW`, perhaps we should return at the first error detected ?

hardening (approved):
A few remarks, but overall looks good to me. 
I think the hash functions for NTLM could be migrated as cred attributes, that's exactly what I was looking for when I've implemented them

client/common/cmdline.h
```diff
@@ -224,6 +224,7 @@ static const COMMAND_LINE_ARGUMENT_A global_cmd_args[] = {
 	{ "kbd-unicode", COMMAND_LINE_VALUE_FLAG, "", NULL, NULL, -1, NULL,
 	  "Send unicode symbols, e.g. use the local keyboard map. ATTENTION: Does not work with every "
 	  "RDP server!" },
+	{ "kdc-url", COMMAND_LINE_VALUE_REQUIRED, "<url>", NULL, NULL, -1, NULL, "KDC server URL" },
```
akallabeth, line 227: I think this should be part of the `/kerberos` options?

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -1202,6 +1202,64 @@ static SECURITY_STATUS SEC_ENTRY negotiate_SetContextAttributesA(PCtxtHandle phC
...
+		MechCred* cred = &creds[i];
+
+		if (!cred->valid)
+			continue;
+
+		WINPR_ASSERT(cred->mech);
+		WINPR_ASSERT(cred->mech->pkg);
+		WINPR_ASSERT(cred->mech->pkg->table);
+		WINPR_ASSERT(cred->mech->pkg->table->SetCredentialsAttributesA);
+		cred->mech->pkg->table->SetCredentialsAttributesA(&cred->cred, ulAttribute, pBuffer,
```
akallabeth, line 1256: Missing return value check

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -1224,6 +1227,91 @@ static SECURITY_STATUS SEC_ENTRY kerberos_QueryContextAttributesW(PCtxtHandle ph
...
+				                   NULL);
+			}
+		}
+		else
+		{
+			SEC_CHAR* KdcUrl = ((SecPkgCredentials_KdcUrlA*)pBuffer)->KdcUrl;
+
+			if (KdcUrl)
+			{
+				credentials->kdc_url = _strdup(KdcUrl);
```
akallabeth, line 1287: Missing `NULL` check

libfreerdp/core/nla.c
akallabeth, line 1056: Missing return value check

libfreerdp/core/nla.c
```diff
@@ -929,6 +939,39 @@ static BOOL nla_setup_kerberos(rdpNla* nla)
...
+#ifdef UNICODE
+		SecPkgCredentials_KdcUrlW secAttr = { NULL };
+		ConvertToUnicode(CP_UTF8, 0, kerbSettings->kdcUrl, -1, &secAttr.KdcUrl, 0);
+		nla->table->SetCredentialsAttributesW(&nla->credentials, SECPKG_CRED_ATTR_KDC_URL,
+		                                      (void*)&secAttr, sizeof(secAttr));
+		free(secAttr.KdcUrl);
+#else
+		SecPkgCredentials_KdcUrlA secAttr = { NULL };
+		secAttr.KdcUrl = kerbSettings->kdcUrl;
+		nla->table->SetCredentialsAttributesA(&nla->credentials, SECPKG_CRED_ATTR_KDC_URL,
```
akallabeth, line 967: Missing return value check

libfreerdp/core/nla.c
```diff
@@ -929,6 +939,39 @@ static BOOL nla_setup_kerberos(rdpNla* nla)
...
+	kerbSettings = nla->kerberosSettings;
+	WINPR_ASSERT(kerbSettings);
+	WINPR_ASSERT(nla->table->SetCredentialsAttributes);
+
+	if (kerbSettings->kdcUrl)
+	{
+#ifdef UNICODE
+		SecPkgCredentials_KdcUrlW secAttr = { NULL };
+		ConvertToUnicode(CP_UTF8, 0, kerbSettings->kdcUrl, -1, &secAttr.KdcUrl, 0);
+		nla->table->SetCredentialsAttributesW(&nla->credentials, SECPKG_CRED_ATTR_KDC_URL,
```
akallabeth, line 961: Missing return value check

libfreerdp/core/nla.c
```diff
@@ -929,6 +939,39 @@ static BOOL nla_setup_kerberos(rdpNla* nla)
...
+	WINPR_ASSERT(nla);
+	kerbSettings = nla->kerberosSettings;
+	WINPR_ASSERT(kerbSettings);
+	WINPR_ASSERT(nla->table->SetCredentialsAttributes);
+
+	if (kerbSettings->kdcUrl)
+	{
+#ifdef UNICODE
+		SecPkgCredentials_KdcUrlW secAttr = { NULL };
+		ConvertToUnicode(CP_UTF8, 0, kerbSettings->kdcUrl, -1, &secAttr.KdcUrl, 0);
```
akallabeth, line 960: Missing return value check

akallabeth (dismissed):

client/common/cmdline.h
hardening: @awakecoding sorry I've changed my mind about this option, I agree with @akallabeth, that would be nice to add this to `/kerberos`, and much simpler as the treatment already does all the checks (then it's just a matter of adding an entry in the array of options and set the target settings id).

We do the split first on coma so that `/kerberos:kdc-url:udp://IT-HELP-DC.ad.it-help.ninja:88,lifetime:2d,cache:/tmp/mycache` would work perfectly without the need for quoting the URL.

Could you also update the various help messages, so that we have an extensive documentation of all options

libfreerdp/core/nla.c
awakecoding: I'll add a check, but I'll really need to make sure that whatever happens in it, it's considered a soft failure and will return true after throwing a warning.

client/common/cmdline.h
awakecoding: I've moved the kdc-url option as a kerberos suboption instead

awakecoding:
All the requested modifications have been made. As for checking return values from SetCredentialsAttributes call, and the difficulty of handling it properly in negotiate_SetCredentialsAttributes + the possibility that the underlying SSPI module does not accept specific attributes, here's what I did:

negotiate_SetCredentialsAttributes loops over the submodules to call SetCredentialsAttributes - we don't know which one should go with which, and there's also the possibility that options may need to be set on multiple submodules. I check the return value for individual calls, but I return SEC_E_OK if at least one of the submodules accepted the attribute. This logic is sufficient to iterate over the list and give each submodule a chance to accept the attribute, and we can return a meaningful status code up in the chain (at least one module accepted the attribute).

As for nla_client_init_cred_handle, it checks the return value from SetCredentialsAttribute(SECPKG_CRED_ATTR_KDC_URL), but it only throws a warning if it gets rejected. I really want to avoid breaking the entire connection if the underlying submodule fails to accept it, especially since there's a good chance it may work without it through KDC auto-detection.

hardening (approved):

PR #8270, opened 2022-10-03
Floatbar fix

akallabeth:


client/X11/xf_floatbar.c
```diff
@@ -474,6 +503,8 @@ static void xf_floatbar_button_update_positon(xfFloatbar* floatbar)
 				break;
 		}
 
+		WINPR_ASSERT(xfc);
```
mfleisz, line 506: Does it really make sense to ASSERT this in every loop iteration?

mfleisz (approved):
Minor comment - not a X11 expert otherwise but LGTM

client/X11/xf_floatbar.c
akallabeth: dereferencing here, and it is not a high performance loop, just over the buttons

PR #8277, opened 2022-10-06
Refactored licensing module

akallabeth:
* Make the whole module opaque for easier testing


libfreerdp/core/license.c
```diff
@@ -542,14 +910,115 @@ int license_recv(rdpLicense* license, wStream* s)
...
+			break;
+
+		case PLATFORM_CHALLENGE_RESPONSE:
+			if (!license_ensure_state(license, LICENSE_STATE_PLATFORM_CHALLENGE, bMsgType))
+				goto fail;
+			if (!license_read_client_platform_challenge_response(license, s))
+				goto fail;
+
+			// TODO: validate challenge response
+			if (FALSE)
```
hardening, line 965: Shall we merge the PR and you'll work that in a second step ?

libfreerdp/core/license.c
```diff
@@ -1703,7 +2646,173 @@ void license_free(rdpLicense* license)
...
+
+	if (!certificate_read_server_certificate(license->certificate, settings->ServerCertificate,
+	                                         settings->ServerCertificateLength))
+		return FALSE;
+
+	s = Stream_New(NULL, 1024);
+	if (!s)
+		return FALSE;
+	else
+	{
```
hardening, line 2779: we can probably remove the `else` (not that we have upgraded C standard, declaring `r` right after should be ok)

hardening:
Overall looks good to me.
There's some TODOs in the code, and for some of them they seem like rest of coding work, could check these ?

libfreerdp/core/license.c
akallabeth: true, but I think the scope makes it more readable, but can change that.

libfreerdp/core/license.c
```diff
@@ -1492,56 +2374,38 @@ BOOL license_answer_license_request(rdpLicense* license)
...
 {
-	wStream* s;
+	wStream* s = license_send_stream_init(license);
 	wStream* challengeRespData;
 	int length;
 	BYTE* buffer;
-	BYTE mac_data[16];
 	BOOL status;
 
+	WINPR_ASSERT(license);
```
MartinHaimberger, line 2391: already checked in license_send_stream_init

MartinHaimberger (changes requested):

libfreerdp/core/connection.c
```diff
@@ -1053,23 +1053,40 @@ BOOL rdp_client_connect_auto_detect(rdpRdp* rdp, wStream* s)
 int rdp_client_connect_license(rdpRdp* rdp, wStream* s)
 {
 	int status;
-	status = license_recv(rdp->license, s);
+	LICENSE_STATE state;
+	UINT16 length, channelId, securityFlags;
 
-	if (status < 0)
-		return status;
+	if (!rdp_read_header(rdp, s, &length, &channelId))
```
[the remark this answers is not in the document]
akallabeth: not required, the pointers are not dereferenced and the function called do the assertions

akallabeth:
@MartinHaimberger ok, added/removed assertions, only kept the ones that are required as the pointers are dereferenced and removed the ones that are only passed on

MartinHaimberger (approved):
still one not needed ASSERT but otherwise it looks good

PR #8278, opened 2022-10-06
Adjust smartcard listing

hardening:
When no CSP is provided, we were listing smartcard materials by querying the `MS_SCARD_PROV_A` CSP, unfortunately on some windows hosts, the smartcards aren't listed in that CSP. So this patch does the key listing by browsing all CSPs instead of just a default one. You can still force a CSP and you'll get keys only from this one.

This patch also address cases where the certificate on the smartcard doesn't have a UPN attribute, if that happen we try to get a UPN from the email address.


akallabeth (approved):
LGTM
@mfleisz what do you say?

mfleisz (approved):

PR #8287, opened 2022-10-07
fixes for NLA under win32

hardening:
That does not enterely fix NLA under win32, but at least fixes low hanging fruits

### DIFF
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -51,6 +51,8 @@
 
 #define SERVER_KEY "Software\\" FREERDP_VENDOR_STRING "\\" FREERDP_PRODUCT_STRING "\\Server"
 
+#define NLA_AUTH_PKG "Negotiate"
+
 /**
  * TSRequest ::= SEQUENCE {
  * 	version    [0] INTEGER,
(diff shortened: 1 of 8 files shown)


akallabeth (approved):
Overall LGTM

just a general question: Why is `AcquireCredentialsHandleA` used here all over the place, this will not work with unicode usernames under windows or am I missing something?

awakecoding:
> Overall LGTM
> 
> just a general question: Why is `AcquireCredentialsHandleA` used here all over the place, this will not work with unicode usernames under windows or am I missing something?

I second this, FreeRDP has always used the "W" APIs on Windows, and the "A" APIs on non-Windows, where WinPR replacement APIs support UTF-8 properly. Calling AcquireCredentialsHandleA will indeed break Unicode support with Windows SSPI.

fifthdegree:
@akallabeth, @awakecoding the authData structure can still be UTF16 even when passed to `AcquireCredentialsHandleA`; there's a flag in the structure that indicates whether or not it's UTF16

awakecoding:
> @akallabeth, @awakecoding the authData structure can still be UTF16 even when passed to `AcquireCredentialsHandleA`; there's a flag in the structure that indicates whether or not it's UTF16

yes, that's the only exception. In fact, we always use SEC_WINNT_AUTH_IDENTITY_UNICODE in nla.c even when building without UNICODE defined + calling the "A" APIs, and WinPR handles it nicely by checking the flag in the structure.

libfreerdp/core/nla.c
```diff
@@ -229,6 +222,22 @@ static BOOL nla_adjust_settings_from_smartcard(rdpNla* nla)
...
+
+	if (!settings->CspName)
+	{
+		if (info->csp &&
+		    ConvertFromUnicode(CP_UTF8, 0, info->csp, -1, &settings->CspName, 0, NULL, FALSE) <= 0)
+		{
+			WLog_ERR(TAG, "unable to set CSP name");
+			goto out;
+		}
+		else if (!(settings->CspName = _strdup(MS_SCARD_PROV_A)))
```
akallabeth, line 234: Please use `freerdp_settings_set_string`

PR #8290, opened 2022-10-09
Fixed ownership of negoToken

akallabeth:
* Ensure negoToken is cleaned up in nla_free
* Renamed function credssp_auth_take_input_buffer now invalidates input buffer an takes ownership of that buffer


hardening (approved):
LGTM

hardening (approved):

PR #8292, opened 2022-10-10
(mostly) Server side  protocol checks improved

akallabeth:


libfreerdp/core/gcc.c
```diff
@@ -681,6 +778,90 @@ BOOL gcc_write_user_data_header(wStream* s, UINT16 type, UINT16 length)
...
+}
+
+static BOOL updateEarlyServerCaps(rdpSettings* settings, UINT32 earlyCapabilityFlags,
+                                  UINT32 connectionType)
+{
+	WINPR_ASSERT(settings);
+
+	settings->SupportDynamicTimeZone =
+	    settings->SupportDynamicTimeZone && (earlyCapabilityFlags & RNS_UD_SC_DYNAMIC_DST_SUPPORTED)
+	        ? TRUE
```
hardening, line 859: The tri-operator is not needed, as the result is already a bool. Or `settings->SupportDynamicTimeZone && !!(earlyCapabilityFlags & RNS_UD_SC_DYNAMIC_DST_SUPPORTED)` does the job ?

libfreerdp/core/gcc.c
```diff
@@ -681,6 +778,90 @@ BOOL gcc_write_user_data_header(wStream* s, UINT16 type, UINT16 length)
...
+{
+	WINPR_ASSERT(settings);
+
+	if (settings->SupportMonitorLayoutPdu)
+		settings->SupportMonitorLayoutPdu =
+		    earlyCapabilityFlags & RNS_UD_CS_SUPPORT_MONITOR_LAYOUT_PDU;
+
+	if (settings->SupportHeartbeatPdu)
+		settings->SupportHeartbeatPdu =
+		    (earlyCapabilityFlags & RNS_UD_CS_SUPPORT_HEARTBEAT_PDU) ? TRUE : FALSE;
```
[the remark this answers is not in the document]
akallabeth: @hardening why `!!`? the `TRUE : FALSE` is there for readability, the datatype is already `BOOL` a `typedef` to `stdbool` where possible) and `!!` is quite cryptic (double invert to get from `int` to `bool` values)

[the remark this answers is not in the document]
hardening: Well quite a lot of projects use `!!`

libfreerdp/core/server.h
```diff
@@ -66,6 +66,8 @@ struct rdp_peer_channel
 	BYTE dvc_open_state;
 	UINT32 dvc_total_length;
 	rdpMcsChannel* mcsChannel;
+
+	char name[128];
```
hardening, line 70: Perhaps it should be `channelName` to match other fields and we can probably move it with other `channelX` fields as I'm not sure we have much to care about ABI here.

akallabeth: good idea

libfreerdp/core/server.h
```diff
@@ -66,6 +66,8 @@ struct rdp_peer_channel
 	BYTE dvc_open_state;
 	UINT32 dvc_total_length;
 	rdpMcsChannel* mcsChannel;
+
+	char channelName[128];
```
hardening, line 70: Just wondering, if I understand correctly this is a static channel, perhaps we can reduce the array size as the name is limited to 8 or 16 (can't remember) ?

winpr/libwinpr/utils/collections/PubSub.c
```diff
@@ -185,7 +185,9 @@ int PubSub_OnEvent(wPubSub* pubSub, const char* EventName, void* context, const
 	size_t index;
 	wEventType* event;
 	int status = -1;
-	WINPR_ASSERT(pubSub);
+
+	if (!pubSub)
```
hardening, line 189: Do you have a case in mind where the assert was not appropriate (it's public API so yeah we shall not assert, but just wondering) ?

hardening (approved):

libfreerdp/core/server.h
akallabeth: no, this is also the dynamic channel struct. did not want to add fancy `malloc` / `free` handling just to have a debug name

PR #8293, opened 2022-10-10
More fixes

akallabeth:
* Adds some more assertions/checks
* Adds basic multitransport message parsing

### DIFF
--- a/libfreerdp/core/multitransport.c
+++ b/libfreerdp/core/multitransport.c
@@ -17,35 +17,155 @@
  * limitations under the License.
  */
 
+#include <winpr/assert.h>
 #include <freerdp/config.h>
-
-#include "multitransport.h"
-
 #include <freerdp/log.h>
+
+#include "rdp.h"
+#include "multitransport.h"
 
 #define TAG FREERDP_TAG("core.multitransport")
 
-int rdp_recv_multitransport_packet(rdpRdp* rdp, wStream* s)
+struct rdp_multitransport
 {
-	UINT32 requestId;
-	UINT16 requestedProtocol;
-	UINT16 reserved;
-	BYTE securityCookie[16];
+	rdpRdp* rdp;
+
+	UINT32 requestId;         /* requestId (4 bytes) */
+	UINT16 requestedProtocol; /* requestedProtocol (2 bytes) */
+	UINT16 reserved;          /* reserved (2 bytes) */
+	BYTE securityCookie[16];  /* securityCookie (16 bytes) */
+};
+
+static BOOL multitransport_compare(const rdpMultitransport* srv, const rdpMultitransport* client)
+{
+	BOOL abortOnError = !freerdp_settings_get_bool(srv->rdp->settings, FreeRDP_TransportDumpReplay);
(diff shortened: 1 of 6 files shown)


libfreerdp/core/rdp.c
```diff
@@ -1971,7 +1986,7 @@ rdpRdp* rdp_new(rdpContext* context)
 	if (!rdp->heartbeat)
 		goto fail;
 
-	rdp->multitransport = multitransport_new();
+	rdp->multitransport = multitransport_new(rdp, INITIATE_REQUEST_PROTOCOL_UDPFECL);
```
hardening, line 1990: Lossy is the old version not used in first intention by the servers, I'd suggest `INITIATE_REQUEST_PROTOCOL_UDPFECR | INITIATE_REQUEST_PROTOCOL_UDPFECR`

libfreerdp/core/multitransport.c
```diff
@@ -17,35 +17,155 @@
...
+{
+	rdpRdp* rdp;
+
+	UINT32 requestId;         /* requestId (4 bytes) */
+	UINT16 requestedProtocol; /* requestedProtocol (2 bytes) */
+	UINT16 reserved;          /* reserved (2 bytes) */
+	BYTE securityCookie[16];  /* securityCookie (16 bytes) */
+};
+
+static BOOL compare(const rdpMultitransport* srv, const rdpMultitransport* client)
```
hardening, line 39: Perhaps we can have a less generic name than `compare` ...

akallabeth: `true`, did not think about that as the function is `static`

libfreerdp/core/rdp.c
akallabeth: @hardening typo? you meant `INITIATE_REQUEST_PROTOCOL_UDPFECL | INITIATE_REQUEST_PROTOCOL_UDPFECR` right?

hardening:
Can we wait before merging this one, I'd like to influence the multi-transport parsing so that it matches my own changes with UDP and multi-transport...

libfreerdp/core/multitransport.c
```diff
@@ -17,35 +17,155 @@
...
 #define TAG FREERDP_TAG("core.multitransport")
 
-int rdp_recv_multitransport_packet(rdpRdp* rdp, wStream* s)
+struct rdp_multitransport
+{
+	rdpRdp* rdp;
+
+	UINT32 requestId;         /* requestId (4 bytes) */
+	UINT16 requestedProtocol; /* requestedProtocol (2 bytes) */
+	UINT16 reserved;          /* reserved (2 bytes) */
```
hardening, line 35: We can avoid storing reserved that is not used at all.

akallabeth:
@hardening ok, then we merge this first? I´ll mark it ready then.

libfreerdp/core/multitransport.c
akallabeth: true, was just for completeness

MartinHaimberger (approved):
LGTM

PR #8303, opened 2022-10-12
Fixed FindXi.cmake

akallabeth:
XITouchClass is a define, XITouchClassInfo a struct, that is not properly detected by CMake check_symbol_exists. Ignore this as all recently enough versions support touch.


hardening:
@freerdp-bot can you re test please ? (mysterious unit test fail)

hardening (approved):

PR #8306, opened 2022-10-12
fix smartcard logon with smartcard emulation

hardening:
When smartcard emulation was enabled we were dumping the key and cert to temporary files for PKINIT call, but they were deleted before we have actually done the PKINIT. This patch fixes it.

It also add debug statement for the listing of smartcard keys / certs


akallabeth (commented):
Wouldn´t it be easier to just add `SmartcardCerts* certs; size_t cert_count;` to `rdpNla` and move the clearing to `nla_free`?
No need for a new API call and duplicated file handling, everything already there?

hardening:
> Wouldn´t it be easier to just add `SmartcardCerts* certs; size_t cert_count;` to `rdpNla` and move the clearing to `nla_free`? No need for a new API call and duplicated file handling, everything already there?

Ok good idea

winpr/libwinpr/ncrypt/ncrypt.c
```diff
@@ -271,4 +271,64 @@ SECURITY_STATUS winpr_NCryptOpenStorageProviderEx(NCRYPT_PROV_HANDLE* phProvider
...
+		NTE_CASE(NTE_BAD_KEYSET_PARAM);
+		NTE_CASE(NTE_FAIL);
+		NTE_CASE(NTE_SYS_ERR);
+		NTE_CASE(NTE_SILENT_CONTEXT);
+		NTE_CASE(NTE_TOKEN_KEYSET_STORAGE_FULL);
+		NTE_CASE(NTE_TEMPORARY_PROFILE);
+		NTE_CASE(NTE_FIXEDPARAMETER);
+
+		default:
+			sprintf_s(unknown, sizeof(unknown), "<unknown 0x%" PRIx32 ">", (UINT32)status);
```
akallabeth, line 329: this might be problematic if called from different threads simultaneously

libfreerdp/core/nla.c
```diff
@@ -200,22 +200,22 @@ static BOOL nla_adjust_settings_from_smartcard(rdpNla* nla)
 	if (!settings->SmartcardLogon)
 		return TRUE;
 
-	if (!smartcard_enumerateCerts(settings, &certs, &count))
+	if (!smartcard_enumerateCerts(settings, &nla->smartcardCerts, &nla->nsmartcardCerts))
```
akallabeth, line 203: Might be a good idea to clean up with `smartcardCerts_Free` before starting the enumeration?

akallabeth (approved):
Overall LGTM,
still some nitpicking (but nothing blocking a merge)

libfreerdp/core/smartcardlogon.c
```diff
@@ -280,10 +283,19 @@ static BOOL list_provider_keys(const rdpSettings* settings, NCRYPT_PROV_HANDLE p
 		                       NULL) <= 0)
 			goto endofloop;
 
+		WLog_DBG(TAG, "opening key %s", cert->info.containerName);
+
 		status = NCryptOpenKey(provider, &phKey, keyName->pszName, keyName->dwLegacyKeySpec,
-		                       keyName->dwFlags);
+		                       NCRYPT_SILENT_FLAG);
```
akallabeth, line 289: Any reason to change `keyName->dwFlags` to `NCRYPT_SILENT_FLAG`?

hardening: Yeah, the return flags of `NCryptEnumKeys` gives information, they aren't intented to be use later for `NCryptOpenKey`. I have  a windows system where it breaks the key listing, `NCRYPT_SILENT_FLAG` is enough.

winpr/libwinpr/ncrypt/ncrypt.c
hardening: Yeah sure, it's not the only place where we do that, and it would not be a big deal (just a wrong message returned). I can change it to just `<unknown>` without the error code (don't want to do some allocations).

libfreerdp/core/smartcardlogon.c
akallabeth: ok, so not by accident :+1:

akallabeth (approved):
LGTM

PR #8309, opened 2022-10-13
Compile fix

akallabeth:


client/Wayland/wlf_input.c
```diff
@@ -441,7 +444,11 @@ BOOL wlf_handle_touch_up(freerdp* instance, const UwacTouchUp* ev)
 		return FALSE;
 
 	WINPR_ASSERT(rdpei->TouchEnd);
+	int contactId;
```
bmiklautz, line 447: Coding style. Even if a another if defined is required this should be at the beginning of the function.

client/Wayland/wlf_input.c
```diff
@@ -441,7 +444,11 @@ BOOL wlf_handle_touch_up(freerdp* instance, const UwacTouchUp* ev)
...
 	rdpei->TouchEnd(rdpei, touchId, x, y, &contactId);
+#else
+	WLog_WARN(TAG, "Touch event detected but RDPEI support not compiled in");
+#endif
```
bmiklautz, line 451: Possibly we should add hint how to disable rdpei.

akallabeth: you mean enable? added that to the log.

hardening (approved):

PR #8311, opened 2022-10-13
nla: Fix some issues with server-side NLA authentication

mfleisz:
This PR fixes following issues with server-side NLA authentication:

- The client nonce should only be sent by the client
- The final stage in the nego token exchange checked the negoToken buffer for data. Instead the corresponding credssp API is now used which checks the correct buffer (output_buffer).
- The negoToken buffer needs to be cleared before sending the public key echo. In some cases the buffer was not empty and incorrectly was part of the response to the client.

### DIFF
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -825,7 +825,8 @@ static int nla_server_authenticate(rdpNla* nla)
 
 		if (res == 1)
 		{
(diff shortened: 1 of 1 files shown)


mfleisz:
@fifthdegree maybe yo can have a look as well?

fifthdegree:
Good catches!

akallabeth (approved):

PR #8312, opened 2022-10-13
Fix CredSSP extended credential attributes on Windows (SECPKG_CRED_ATTR_KDC_URL)

awakecoding:
Fix a regression introduced by https://github.com/FreeRDP/FreeRDP/pull/8219 caused by incorrect usage of SecPkgCredentials_KdcUrlA with "W" SSPI APIs on Windows, when SecPkgCredentials_KdcUrlW should be used instead. I adapted the older code that correctly handled it.

### DIFF
--- a/libfreerdp/core/credssp_auth.c
+++ b/libfreerdp/core/credssp_auth.c
@@ -151,6 +151,48 @@ BOOL credssp_auth_setup_auth_data(rdpCredsspAuth* auth, const SEC_WINNT_AUTH_IDE
 	return TRUE;
 }
 
+static BOOL credssp_auth_client_init_cred_attributes(rdpCredsspAuth* auth)
+{
+	SECURITY_STATUS status;
+
+	WINPR_ASSERT(auth);
+
+	if (auth->kerberosSettings.kdcUrl)
+	{
+#ifdef UNICODE
+		SecPkgCredentials_KdcUrlW secAttr = { NULL };
+		ConvertToUnicode(CP_UTF8, 0, auth->kerberosSettings.kdcUrl, -1, &secAttr.KdcUrl, 0);
+
+		if (!secAttr.KdcUrl)
+			return FALSE;
+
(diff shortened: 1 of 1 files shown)


awakecoding:
@fifthdegree FYI

libfreerdp/core/credssp_auth.c
```diff
@@ -151,6 +151,48 @@ BOOL credssp_auth_setup_auth_data(rdpCredsspAuth* auth, const SEC_WINNT_AUTH_IDE
...
+{
+	SECURITY_STATUS status;
+
+	WINPR_ASSERT(auth);
+
+	if (auth->kerberosSettings.kdcUrl)
+	{
+#ifdef UNICODE
+		SecPkgCredentials_KdcUrlW secAttr = { NULL };
+		ConvertToUnicode(CP_UTF8, 0, auth->kerberosSettings.kdcUrl, -1, &secAttr.KdcUrl, 0);
```
akallabeth, line 164: Missing return check

awakecoding: see right below, it checks if the output pointer is NULL, and it was initialized to NULL before the call. it doesn't check the return value from ConvertToUnicode, but it doesn't matter because it checks if ConvertToUnicode set a pointer value. We also only call this code if auth->kerberosSettings.kdcUrl is non-null

fifthdegree:
See, if we didn't have this strange desire to use W functions just because UNICODE is defined, we wouldn't need this...

awakecoding:
> See, if we didn't have this strange desire to use W functions just because UNICODE is defined, we wouldn't need this...

I told you the other day, I hate those W functions just as much as you do, but for Windows SSPI Unicode support, it's the only way. As for extended attributes, using UTF-8 strings in W APIs would just break the convention, so it's just better to go along with it. I know it's a pain, it's just the best option we can come up with that will work with everything.

akallabeth (approved):
LGTM

PR #8313, opened 2022-10-13
Eliminate memset

akallabeth:


channels/tsmf/client/alsa/tsmf_alsa.c
```diff
@@ -226,14 +226,15 @@ static void tsmf_alsa_free(ITSMFAudioDevice* audio)
 
 ITSMFAudioDevice* alsa_freerdp_tsmf_client_audio_subsystem_entry(void)
 {
-	TSMFAlsaAudioDevice* alsa;
-	alsa = (TSMFAlsaAudioDevice*)malloc(sizeof(TSMFAlsaAudioDevice));
-	ZeroMemory(alsa, sizeof(TSMFAlsaAudioDevice));
+	TSMFAlsaAudioDevice* alsa = (TSMFAlsaAudioDevice*)calloc(1, sizeof(TSMFAlsaAudioDevice));
```
pnowack, line 229: This cast should not be needed (the `(TSMFAlsaAudioDevice*)`), applies to the other `calloc`s here in this MR too.

libfreerdp/primitives/test/TestPrimitivesSet.c
```diff
@@ -62,7 +61,7 @@ static BOOL test_set8u_func(void)
 	for (off = 0; off < 16; ++off)
 	{
 		UINT32 len;
-		memset(dest, 3, sizeof(dest));
+		BYTE dest[1024] = { 3 };
```
pnowack, line 64: Same here. You change the test here. `{ 3 }` sets the first element to `3` and the rest (which is not specified) is automatically set to `0`, while before everything was set to `3`.

server/shadow/shadow_capture.c
```diff
@@ -87,13 +87,12 @@ int shadow_capture_compare(BYTE* pData1, UINT32 nStep1, UINT32 nWidth, UINT32 nH
 	UINT32 nrow, ncol;
 	UINT32 l, t, r, b;
 	BYTE *p1, *p2;
-	BOOL rows[1024];
+	BOOL rows[1024] = { 0xff };
```
pnowack, line 90: Same here. `{ 0xFF }` sets the first element to `0xFF` and the rest (which is not specified) is automatically set to `0`, while before everything was set to `0xFF`.

libfreerdp/primitives/test/TestPrimitivesAlphaComp.c
```diff
@@ -179,8 +178,6 @@ static int test_alphaComp_speed(void)
 	for (i = 0; i < sizeof(src2) / 4; ++i)
 		*ptr++ |= 0xFF000000U;
 
-	memset(dst1, 0, sizeof(dst1));
```
pnowack, line 182: The first `memset` removal was fine. But does this apply to this one too? I haven't specifically looked into the function calls above, whether they change `dst1`, but if they do, the behaviour of this test changes here.

channels/cliprdr/server/cliprdr_main.c
```diff
@@ -631,8 +631,9 @@ static UINT cliprdr_server_receive_temporary_directory(CliprdrServerContext* con
 
 	memset(cliprdr->temporaryDirectory, 0, ARRAYSIZE(cliprdr->temporaryDirectory));
 
-	if (ConvertFromUnicode(CP_UTF8, 0, wszTempDir, -1, &(cliprdr->temporaryDirectory),
-	                       ARRAYSIZE(cliprdr->temporaryDirectory), NULL, NULL) < 1)
+	char* ptr = &cliprdr->temporaryDirectory[0];
+	if (ConvertFromUnicode(CP_UTF8, 0, wszTempDir, -1, &ptr, ARRAYSIZE(cliprdr->temporaryDirectory),
```
pnowack, line 635: Oh, this won't work here. Not due to this change here, but it did not work before. This was actually something, that was mentioned afair while you were on holidays.
Basically this convert call works on `stable-2.0`, while on `master`, it will (almost) always fail. Reason is the `ARRAYSIZE(cliprdr->temporaryDirectory)`, as the actual string is likely shorter than that length here.

On `stable-2.0`, `ConvertFromUnicode()` tries to find the null terminator by itself (I think there is a manual out-of-bounds check before that) and that way does not fail.

pnowack (changes requested):
I think my eyes are rectangles now. I hope I haven't missed anything else here.

akallabeth:
@pnowack thanks for the review, addressed your remarks

channels/cliprdr/server/cliprdr_main.c
akallabeth: ok, then leave this out of here.

libfreerdp/primitives/test/TestPrimitivesAlphaComp.c
akallabeth: yes, the loop uses `ptr = (UINT32*)src2;` so `dst1` is never touched before

libfreerdp/primitives/test/TestPrimitivesSet.c
akallabeth: reverted, good catch

channels/tsmf/client/oss/tsmf_oss.c
```diff
@@ -232,15 +232,16 @@ static void tsmf_oss_free(ITSMFAudioDevice* audio)
 
 ITSMFAudioDevice* oss_freerdp_tsmf_client_audio_subsystem_entry(void)
 {
-	TSMFOssAudioDevice* oss;
-	oss = (TSMFOssAudioDevice*)malloc(sizeof(TSMFOssAudioDevice));
-	ZeroMemory(oss, sizeof(TSMFOssAudioDevice));
+	TSMFOssAudioDevice* oss = (TSMFOssAudioDevice*)calloc(1, sizeof(TSMFOssAudioDevice));
```
pnowack, line 235: Unneeded cast (can be removed).

libfreerdp/cache/persistent.c
```diff
@@ -319,15 +346,13 @@ int persistent_cache_close(rdpPersistentCache* persistent)
 
 rdpPersistentCache* persistent_cache_new(void)
 {
-	rdpPersistentCache* persistent;
-
-	persistent = (rdpPersistentCache*)calloc(1, sizeof(rdpPersistentCache));
+	rdpPersistentCache* persistent = (rdpPersistentCache*)calloc(1, sizeof(rdpPersistentCache));
```
pnowack, line 349: Unneeded cast.

libfreerdp/cache/persistent.c
```diff
@@ -319,15 +346,13 @@ int persistent_cache_close(rdpPersistentCache* persistent)
...
 
 	if (!persistent)
 		return NULL;
 
 	persistent->bmpSize = 0x4000;
-	persistent->bmpData = (BYTE*)malloc(persistent->bmpSize);
+	persistent->bmpData = (BYTE*)calloc(1, persistent->bmpSize);
```
pnowack, line 355: Unneeded cast.

pnowack (commented):
With https://github.com/FreeRDP/FreeRDP/pull/8313#discussion_r995386447, I meant the other casts too in the changes. Added them in this round of review here.

Otherwise, lgtm.

PR #8315, opened 2022-10-14
fix smartcard listing

hardening:
This commit fixes various bugs that I've noticed on some windows systems with smartcards that contains multiple certificates:

* With some drivers if you retrieve the ATR while enumerating the NCrypt keys, it seems to confuse the NCrypt key context (and you're unable to retrieve certificate property). As we don't use the ATR, let's remove the ATR retrieval.
* if don't give any user or domain on the command line, in settings you get User=Domain=NULL, but if you pass /u:user, you get User="user" and Domain = ""(empty string not NULL). The smartcard filtering by user/domain was not ready for that.

### DIFF
--- a/include/freerdp/utils/smartcardlogon.h
+++ b/include/freerdp/utils/smartcardlogon.h
@@ -37,8 +37,6 @@ typedef struct
 	char* domainHint;
 	char* subject;
 	char* issuer;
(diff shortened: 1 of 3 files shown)


akallabeth (approved):
LGTM
@mfleisz what do you say?

PR #8316, opened 2022-10-14
Add UserSpecifiedServerName setting, /server-name command-line parameter

awakecoding:
This pull request makes it possible to specify a server name different from the network hostname to be used for server validation in TLS and Kerberos. The internal setting is UserSpecifiedServerName, which is the name used for the same internal property in the Microsoft RDP client. The command-line parameter /server-name can be used to set UserSpecifiedServerName. A helper function freerdp_settings_get_server_name() returns UserSpecifiedServerName if it is set, otherwise it returns ServerHostname (the previous default).

For instance, if DNS is not configured, it is possible to use the IP address of the destination host + specify the real server name to be used for validation, which is important for Kerberos. As for TLS, well, /cert-ignore can always ignore it, but if automatic validation should work, then this can be used to set the expected hostname for the certificate validation as well:

`/server-name:IT-HELP-DVLS.ad.it-help.ninja /v:10.10.0.6`

In cases where DNS is properly configured, but the connection is done through a relatively simple TCP proxy like the Devolutions Gateway, or simple TCP port forwarding, then the RDP client unfortunately thinks that the proxy/relay host is the destination RDP server name. This is actually the primary reason for adding such an option, because otherwise it would be impossible to make Kerberos work properly through the Devolutions Gateway:

`/server-name:IT-HELP-DVLS.ad.it-help.ninja /v:gateway.ad.it-help.ninja`

In this example, the RDP client connects to "gateway.ad.it-help.ninja" but traffic is forwarded as-is to "IT-HELP-DVLS.ad.it-help.ninja", so both TLS and Kerberos server validation should be done with "IT-HELP-DVLS.ad.it-help.ninja" rather than "gateway.ad.it-help.ninja"

The same problem already exists for RD Gateway connections, but we already have a lot of code in place to handle the two different hostnames (RD Gateway and RDP server). The RD Gateway protocol is also much higher level than simple TCP forwarding, and is a full tunneling protocol with its own messages to handle this. I didn't touch the RD Gateway code paths and they should work the same way they always did, except we have the option to force-override the server name to be used for validation if for whatever reason the RD Gateway can only accept a specific hostname that doesn't match the one we want to use for server validation.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -2463,6 +2463,11 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings, 
 					return rc;
 			}
 		}
+		CommandLineSwitchCase(arg, "server-name")
+		{
+			if (!freerdp_settings_set_string(settings, FreeRDP_UserSpecifiedServerName, arg->Value))
+				return COMMAND_LINE_ERROR_MEMORY;
+		}
 		CommandLineSwitchCase(arg, "shell")
 		{
 			if (!freerdp_settings_set_string(settings, FreeRDP_AlternateShell, arg->Value))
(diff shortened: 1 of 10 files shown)


hardening (approved):
LGTM

PR #8319, opened 2022-10-16
Smartcard support for gateway authentication

fifthdegree:
This brings smartcard logon support to gateway authentication. In addition, it does the following:

- Restructures smartcardlogon.c a little bit to return exactly one certificate to use (and fails otherwise). 
- Defines a callback to facilitate choosing a certificate when there's more than one possibility, and includes an implementation for terminal clients.
- Uses the correct key container name for PIV and GIDS compatible cards.


hardening:
@freerdp-bot give a test to this nice PR please

akallabeth (commented):
Overall LGTM

hardening:
@fifthdegree you're my hero for doing the smartcard calls to compute the containerId

winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
```diff
@@ -793,6 +813,117 @@ static SECURITY_STATUS NCryptP11EnumKeys(NCRYPT_PROV_HANDLE hProvider, LPCWSTR p
...
+
+	if (SCardEstablishContext(SCARD_SCOPE_USER, NULL, NULL, &context) != SCARD_S_SUCCESS)
+		return NTE_BAD_KEY;
+
+	if (SCardConnectW(context, reader, SCARD_SHARE_SHARED, SCARD_PROTOCOL_Tx, &card, &proto) !=
+	    SCARD_S_SUCCESS)
+		goto out;
+	pci = (proto == SCARD_PROTOCOL_T0) ? SCARD_PCI_T0 : SCARD_PCI_T1;
+
+	buf_len = 2;
```
hardening, line 864: Should be `buf_len = sizeof(buf)` some smartcard returns more than 2 bytes. See my fix in https://github.com/hardening/FreeRDP/blob/gw_smartcard/winpr/libwinpr/ncrypt/ncrypt_pkcs11.c#L866

winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
```diff
@@ -793,6 +813,117 @@ static SECURITY_STATUS NCryptP11EnumKeys(NCRYPT_PROV_HANDLE hProvider, LPCWSTR p
...
+	if (SCardTransmit(card, pci, APDU_PIV_SELECT_AID, sizeof(APDU_PIV_SELECT_AID), NULL, buf,
+	                  &buf_len) != SCARD_S_SUCCESS)
+		goto out;
+	if ((buf[0] != 0x90 || buf[1] != 0) && buf[0] != 0x61)
```
hardening, line 868: This is not entirely correct, if the the result is `61XX` the spec says that you're supposed to issue a `GET_RESPONSE` command like in https://github.com/hardening/FreeRDP/blob/gw_smartcard/winpr/libwinpr/ncrypt/ncrypt_pkcs11.c#L870 and SW2 contains the remaining bytes

hardening:
This PR works correctly except when websocket is enabled, in this case the protocol is `101 switched` to websocket in the middle of the negotiation

hardening:
@freerdp-bot can you test buddy ?

winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
fifthdegree: Since we skip the Le byte in the SELECT command, cards will not return any data, just the status bytes. This is intentional, because we really don't need the data it returns, we just want to make sure that we do in fact have a PIV card and that it's the selected app

hardening: Here with just `buf_len=2`, `SCardTransmit` returns `not enough memory`

hardening:

fifthdegree:
> This PR works correctly except when websocket is enabled, in this case the protocol is `101 switched` to websocket in the middle of the negotiation

This now works as expected after enabling mutual authentication

hardening (approved):

PR #8320, opened 2022-10-17
Do not terminate on error info

akallabeth:


pnowack (changes requested):
The change itself looks fine to me, except for two points:

1. I think, the option should be enabled by default. Otherwise, it would have bad synergy with the `auto-reconnection` feature.
In g-r-d, I use the error info, to ensure (except for graphics subsystem failed), that the client does not reconnect. Afaik, MS Windows RDS does the same.
2. The commit message should never exceed 72 characters, otherwise ugly line breaks happen (like here).

akallabeth:
@pnowack good point, in that case I´ve changed to just not terminate in that case (as the setting was just there to keep current behaviour without a change)

pnowack (approved):
LGTM

PR #8321, opened 2022-10-17
Fix Format List PDU is not sent when the Clipboard Redirection Virtual Channel is initialized for the second time. (X11 client)

XiaopengZHOU:


hardening:

KangLin:
If you modify it this way, you will not be able to send the format list the first time.

akallabeth (approved):
LGTM

akallabeth:
@Ka

> If you modify it this way, you will not be able to send the format list the first time.

Why should that be the case?
`xf_clipboard_formats_free` just resets `lastSentFormats` so the current clipboard content is detected as change to the state the server has.
There might be an issue, if the local clipboard is empty though...

KangLin:
Sorry. I was misled by the function name of the xf_clipboard_formats_free.
It is recommended to rename to xf_clipboard_last_sent_formats_free.

PR #8323, opened 2022-10-18
Increase yuv decoder worker count

akallabeth:
The yuv decoder might run out of workers if the rectangles are smaller than 64x64. Assume 16x16 tiles for the decoder


hardening (approved):
Looks OK

PR #8326, opened 2022-10-18
uwac: fix wl_buffer_destroy of wl_cursor buffer

gschwind:
Following the wayland-cursor documentation [1] the buffer created by wl_cursor_image_get_buffer should not be destroyed. The new code avoid to destroy it by avoiding to connect the release callback to this buffer.

Note that I did the fix based on my understanding of the code and the documentation. I do not have a test case that trigger the bug. I did tested the regular path.

[1] https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/cursor/wayland-cursor.c#L147-L154


hardening:

akallabeth (approved):
LGTM

PR #8329, opened 2022-10-19
Cmdline tls unified option

akallabeth:


client/common/cmdline.c
```diff
@@ -1609,6 +1609,107 @@ static BOOL setSmartcardEmulation(const char* value, rdpSettings* settings)
...
+	{
+		WLog_WARN(TAG, "Option /tls-ciphers is deprecated, use /tls:ciphers instead");
+		rc = parse_tls_ciphers(settings, arg->Value);
+	}
+	CommandLineSwitchCase(arg, "tls-seclevel")
+	{
+		WLog_WARN(TAG, "Option /tls-seclevel is deprecated, use /tls:seclevel instead");
+		rc = parse_tls_seclevel(settings, arg->Value);
+	}
+	CommandLineSwitchCase(arg, "tls-secrets-file")
```
hardening, line 1694: this option is really new (1 month ?), perhaps we're not forced to provide retro-compatibility...

hardening (approved):

client/common/cmdline.c
akallabeth: they are now not build in anyway (except you define `-DWITH_FREERDP_DEPRECATED=ON`)

PR #8333, opened 2022-10-20
Cmdline simplify

akallabeth:
* Groups a lot of options logically connected
* Adds a new `CMake` option to enable/disable deprecated command line separate from deprecated functions
* Added/Unified logging messages for deprecated options

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -148,6 +148,11 @@ if (WITH_FREERDP_DEPRECATED)
 	add_definitions(-DWITH_FREERDP_DEPRECATED)
 endif()
 
+option(WITH_FREERDP_DEPRECATED_COMMANDLINE "Build FreeRDP deprecated command line options" OFF)
+if (WITH_FREERDP_DEPRECATED_COMMANDLINE)
+    add_definitions(-DWITH_FREERDP_DEPRECATED_COMMANDLINE)
+endif()
+
 # Make paths absolute
 if (CMAKE_INSTALL_PREFIX)
 	get_filename_component(CMAKE_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}" ABSOLUTE)
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -1873,7 +1873,7 @@ static int parse_gfx_options(rdpSettings* settings, const COMMAND_LINE_ARGUMENT_
 			{
 				const char* val = ptr[x];
 #ifdef WITH_GFX_H264
-				if (_strnicmp("AVC444", val, 7) == 0)
+				if (_strnicmp("AVC444", val, 6) == 0)
 				{
 					const int bval = parse_on_off_option(val);
 					if (bval < 0)
@@ -1882,7 +1882,7 @@ static int parse_gfx_options(rdpSettings* settings, const COMMAND_LINE_ARGUMENT_
 						GfxH264 = bval > 0;
 					codecSelected = TRUE;
 				}
-				else if (_strnicmp("AVC420", val, 7) == 0)
+				else if (_strnicmp("AVC420", val, 6) == 0)
 				{
 					const int bval = parse_on_off_option(val);
 					if (bval < 0)
@@ -1893,7 +1893,7 @@ static int parse_gfx_options(rdpSettings* settings, const COMMAND_LINE_ARGUMENT_
 				}
 				else
 #endif
-				    if (_strnicmp("RFX", val, 4) == 0)
+				    if (_strnicmp("RFX", val, 3) == 0)
 				{
 					const int bval = parse_on_off_option(val);
 					if (bval < 0)
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -1552,7 +1552,7 @@ int freerdp_client_settings_command_line_status_print_ex(rdpSettings* settings, 
 			else
 				return COMMAND_LINE_ERROR;
 		}
-#if defined(WITH_FREERDP_DEPRECATED)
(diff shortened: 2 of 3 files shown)


akallabeth:
@hardening this is the command line pr. I´ve added the new unified options and kept the old syntax active (but disabled for our nightly builds)

client/common/cmdline.c
```diff
@@ -1826,6 +1839,123 @@ static int parse_tls_options(rdpSettings* settings, const COMMAND_LINE_ARGUMENT_
...
+			BOOL GfxAVC444 = FALSE;
+			BOOL RemoteFxCodec = FALSE;
+			BOOL GfxProgressive = FALSE;
+			BOOL codecSelected = FALSE;
+
+			for (size_t x = 0; x < count; x++)
+			{
+				const char* val = ptr[x];
+#ifdef WITH_GFX_H264
+				if (_strnicmp("AVC444", val, 7) == 0)
```
pnowack, line 1876: Feel free to ignore:

Same for all other `_strnicmp` calls in this PR: Might be better to use `strlen() + 1` for the size to prevent likelihood of human error here of using the wrong size. I'm fine though with the current approach.

client/common/cmdline.c
```diff
@@ -1963,6 +1963,166 @@ static int parse_gfx_options(rdpSettings* settings, const COMMAND_LINE_ARGUMENT_
...
+				    freerdp_settings_get_string(settings, FreeRDP_KeyboardRemappingList);
+
+				/* Basic sanity test. Entries must be like <key>=<value>, e.g. 1=2 */
+				if (!check_kbd_remap_valid(now))
+					rc = COMMAND_LINE_ERROR_UNEXPECTED_VALUE;
+				else if (old)
+				{
+					const size_t olen = strlen(old);
+					const size_t alen = strlen(now);
+					char* tmp = calloc(olen + alen + 1, sizeof(char));
```
pnowack, line 2045: No space for null terminator (you have `"%s,%s"` as string)?

CMakeLists.txt
```diff
@@ -148,6 +148,11 @@ if (WITH_FREERDP_DEPRECATED)
 	add_definitions(-DWITH_FREERDP_DEPRECATED)
 endif()
 
+option(WITH_FREERDP_DEPRECATED_COMMANDLINE "Build FreeRDP deprecated command line options" OFF)
```
pnowack, line 151: Typo in commit message: `WITH_FREERDP_DEPRECATED_COMMANDLINE_COMMANDLINE` -> `WITH_FREERDP_DEPRECATED_COMMANDLINE`

client/common/cmdline.c
```diff
@@ -2903,29 +2990,24 @@ int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings,
...
 				return COMMAND_LINE_ERROR_MEMORY;
 
-			settings->RemoteApplicationMode = TRUE;
-			settings->RemoteAppLanguageBarSupported = TRUE;
-			settings->Workarea = TRUE;
-			settings->DisableWallpaper = TRUE;
-			settings->DisableFullWindowDrag = TRUE;
+			settings->LoadBalanceInfoLength = (UINT32)strlen((char*)settings->LoadBalanceInfo);
 		}
+#if defined(WITH_FREERDP_DEPRECATED_COMMANDLINE_COMMANDLINE)
```
pnowack: `WITH_FREERDP_DEPRECATED_COMMANDLINE_COMMANDLINE` -> `WITH_FREERDP_DEPRECATED_COMMANDLINE`

client/common/cmdline.c
```diff
@@ -2108,18 +2108,105 @@ static int parse_kbd_options(rdpSettings* settings, const COMMAND_LINE_ARGUMENT_
...
+				break;
+		}
+	}
+	return rc;
+}
+
+static int parse_app_option_program(rdpSettings* settings, const char* cmd)
+{
+	const size_t ids[] = { FreeRDP_RemoteApplicationMode, FreeRDP_RemoteAppLanguageBarSupported,
+		                   FreeRDP_Workarea, FreeRDP_DisableWallpaper,
```
pnowack, line 2131: Weird indentation, but if clang-format did it, well, then ignore this comment.

client/common/cmdline.c
akallabeth: this check has a bug, but another one as you imagined. `_strnicmp("AVC444", val, 6) == 0` is what is desired, as `AVC444:on` and `AVC444:off` is also valid syntax.

client/common/cmdline.c
akallabeth: fixed, good catch!

client/common/cmdline.c
akallabeth: yes, sometimes it is a bit strange, but better to have it this way than manual formatting all over the place.

client/common/cmdline.c
akallabeth: hmm, can not see this with grep? maybe a github artifact (did fix this with the last force push 5 days earlier?)

pnowack: > hmm, can not see this with grep? maybe a github artifact (did fix this with the last force push 5 days earlier?)

No, during review, I go from commit to commit. So, might be that you changed that in a follow-up commit. So, the overall change won't include that with a grep search.

pnowack:
I kinda tend to say that using a helper function for `_strnicmp("layout:", val, 7) == 0` and similar would be better (something like `has_string_prefix()`).  
Then we just use `strlen` in that helper function around that string to determine the length.
Then I also don't have to count the length of each string myself during the review.

PR #8273, opened 2022-10-05
Refactored capability parsing

akallabeth:
* Add new settings for previously ignored capability options
* Store raw capability data in settings for later use
* Add function to extract settings from raw capability data
* Split capability read/write from client/server logic (e.g. enforce limits, ...)

Issue #925, opened 2013-01-24
xfreerdp connect VBox error


Vitaly85:
Can't connect to my VBox(oracle 4.2.6) vrdp host, with error "rdp_recv_data_pdu failed".
xfreerdp crashes in rdp_recv_client_font_map_pdu(libfreerdp/core/activation.c 249 string) after stream_get_left(STREAM*)
change in code line 251 (if(stream_get_left(s) < 8) to if(stream_get_left(s) > 8))
now it work.
This is FreeRDP version 1.1.0-dev (git 1.0.1-1729-g81e1)
And i can connect to Ms RDP Windows with this changes.
connect line: ./xfreerdp /v:192.168.64.54:3389
Thanks for reading! Hope it will be fixed fast )

PR #8337, opened 2022-10-21
More checks, new functions for upcoming state machine PR

akallabeth:
This is the last pr in a series improving the return value checks and bringing protocol parsing to level to allow dissecting dumps of recorded sessions.
The next one will change the core RDP state machine to fix some missing states/checks that where previously not required for simple client/server operation.

Based on #8262 #8272 #8273 #8277 #8293 #8320 #8333

### DIFF
--- a/libfreerdp/core/activation.c
+++ b/libfreerdp/core/activation.c
@@ -34,13 +34,12 @@ static BOOL rdp_send_server_font_map_pdu(rdpRdp* rdp);
 
 static BOOL rdp_write_synchronize_pdu(wStream* s, const rdpSettings* settings)
 {
-	WINPR_ASSERT(s);
-	WINPR_ASSERT(settings);
+	const UINT32 PduSource = freerdp_settings_get_uint32(settings, FreeRDP_PduSource);
 
 	if (Stream_GetRemainingCapacity(s) < 4)
 		return FALSE;
-	Stream_Write_UINT16(s, SYNCMSGTYPE_SYNC);    /* messageType (2 bytes) */
-	Stream_Write_UINT16(s, settings->PduSource); /* targetUser (2 bytes) */
+	Stream_Write_UINT16(s, SYNCMSGTYPE_SYNC); /* messageType (2 bytes) */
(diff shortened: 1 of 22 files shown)


hardening:
Just wondering about the trauma that happened with last command line change. I don't have anything about packing options together but perhaps that should be a PR on its own ?

akallabeth:
@hardening they are, this pr is just based on that branch (will rebase once they are merged)

libfreerdp/core/activation.c
```diff
@@ -497,39 +514,17 @@ BOOL rdp_recv_font_map_pdu(rdpRdp* rdp, wStream* s)
...
-{
-	WINPR_ASSERT(rdp);
-	WINPR_ASSERT(s);
-
-	WLog_WARN(TAG, "Invalid PDU received: FONT_MAP only allowed client -> server");
-	return rdp_finalize_set_flag(rdp, FINALIZE_SC_FONT_MAP_PDU);
-}
-
-BOOL rdp_recv_client_font_map_pdu(rdpRdp* rdp, wStream* s)
-{
```
hardening, line 517: Here the behaviour of function is changed, it used to support a truncated PDU (I think it happens sometime) and with your changes it would be rejected with an error. (comment copied from #8361, sorry for the duplicate)

hardening (changes requested):
Except my remark that looks ok for me

libfreerdp/core/activation.c
akallabeth: @hardening the `[MS-RDPBCGR] 2.2.1.22 Server Font Map PDU` can only be sent from server to client, so the old code handled stuff that must not happen.

akallabeth: @hardening found the thing, https://github.com/FreeRDP/FreeRDP/issues/925 is what you meant? will readd the hack

akallabeth:
@hardening really nice catch (and long memory it seems ^^)

libfreerdp/core/activation.c
hardening: Hum there's this one (I didn't remember about it). But more recently I've also hit that with some other impl (can't remember which thought).

PR #8338, opened 2022-10-21
Rdp state machine update

akallabeth:
Big overhaul of the RDP state machine.
* Allows to properly attribute errors to states (e.g. which message failed)
* Removes recursive state machine calls

Based on #8337


libfreerdp/core/rdp.c
```diff
@@ -1617,37 +1617,39 @@ static int rdp_recv_pdu(rdpRdp* rdp, wStream* s)
 		return rc;
 }
 
-int rdp_recv_callback(rdpTransport* transport, wStream* s, void* extra)
+/* TODO: Need to properly return:
```
mfleisz, line 1620: Since we are already touching large parts of this code - can we replace these magic numbers with proper defines so everybody knows immediately what is going? (same in peer.c)

libfreerdp/core/rdp.c
```diff
@@ -1764,31 +1795,121 @@ int rdp_recv_callback(rdpTransport* transport, wStream* s, void* extra)
...
 			status = rdp_client_connect_demand_active(rdp, s);
 
 			if (status < 0)
 				WLog_DBG(TAG,
 				         "%s: %s - "
 				         "rdp_client_connect_demand_active() - %i",
 				         __FUNCTION__, rdp_get_state_string(rdp), status);
-
+			else if (status == 1)
+				status = 1;
```
mfleisz, line 1816: This looks awkward - maybe just `break;` to make the code more clear?

mfleisz (commented):
Two small comments

Tested server side and client side (including gateway and redirection) - works fine

libfreerdp/core/rdp.c
akallabeth: @mfleisz good idea, added an enum and a string function for it

libfreerdp/core/rdp.c
akallabeth: done

mfleisz (approved):
LGTM

PR #8342, opened 2022-10-23
Check smartcard certificates for correct EKU

fifthdegree:
Filter out certificates that don't have the Smart Card Logon EKU since they cannot be used for login


mfleisz:
@freerdp-bot test

PR #8346, opened 2022-10-25
Fix certificate and private key checks for smartcard logon

akallabeth:


mfleisz (approved):
LGTM

hardening (approved):
Looks ok even if I don't see what it fixes

akallabeth:
@hardening you set the temporary files to a temporary variable with 2 error aborts and only afterward you set the files to the info structure. so if a write fails the files will not be cleaned up

akallabeth:
@hardening also, current master crashes if you run `/smartcard-logon:cert:<certfile>` without a key (and vice versa)

PR #8347, opened 2022-10-25
Add setting name for invalid function access

akallabeth:
Try to be helpful and resolve the settings index with a readable name if possible to help diagnose wrong function use


hardening (approved):

hardening:
oh nice I had the need for that but was too lazy to code it

PR #8358, opened 2022-10-27
Filedescriptorw fix

akallabeth:


akallabeth:
@pnowack the timestamps were really not correct, but I can not find a bug with file overwriting.

winpr/libwinpr/clipboard/synthetic_file.c
```diff
@@ -330,7 +328,6 @@ static BOOL do_add_directory_contents_to_list(wClipboard* clipboard, const WCHAR
 		BOOL bRet = FindNextFileW(hFind, &FileData);
 		if (!bRet)
 		{
-			WLog_WARN(TAG, "FindNextFile failed (%" PRIu32 ")", GetLastError());
```
pnowack, line 333: Maybe, it would be sufficient to just move this warning a few lines below after the `if (ERROR_NO_MORE_FILES == GetLastError())` check.

libfreerdp/utils/cliprdr_utils.c
```diff
@@ -199,15 +197,24 @@ UINT cliprdr_serialize_file_list_ex(UINT32 flags, const FILEDESCRIPTORW* file_de
 		}
 
 		Stream_Write_UINT32(s, file->dwFlags);          /* flags (4 bytes) */
-		Stream_Zero(s, 32);                             /* reserved1 (32 bytes) */
+
+		Stream_Write_UINT32(s, file->clsid.Data1);
+		Stream_Write_UINT16(s, file->clsid.Data2);
+		Stream_Write_UINT16(s, file->clsid.Data3);
+		Stream_Write(s, &file->clsid.Data4, sizeof(file->clsid.Data4));
```
pnowack, line 204: Maybe we could add an assert here or (further) above that `sizeof(file->clsid.Data4) == 8`?

pnowack (approved):
Just two tiny things, otherwise LGTM.

libfreerdp/utils/cliprdr_utils.c
akallabeth: @pnowack ok, can do that.
@hardening well, it is serializing the data structure, of course the data sizes must match

winpr/libwinpr/clipboard/synthetic_file.c
```diff
@@ -330,9 +330,9 @@ static BOOL do_add_directory_contents_to_list(wClipboard* clipboard, const WCHAR
...
 			if (ERROR_NO_MORE_FILES == GetLastError())
 				return TRUE;
+			WLog_WARN(TAG, "FindNextFile failed (%" PRIu32 ")", GetLastError());
```
pnowack, line 340: One last nit after the last force push: Adjust the commit message (from `Removed unnecessary log message`), since it does not match its actual change anymore.

Maybe something like:

```
winpr/clipboard: Move a log message

ERROR_NO_MORE_FILES is the exit condition to complete a successful
directory traversal. So, don't warn about it.
```

akallabeth: ok, updated.

pnowack (approved):

PR #8370, opened 2022-11-03
[channel] Fixed broken length check

akallabeth:
The length check for channel chunk data was wrong. Not only was it checked twice, the second check expected the whole fragmented data to be available.


hardening (approved):

hardening:
Tested and working here

PR #8371, opened 2022-11-03
client: improve connection time

hardening:
The old code was looping with blindly checking for transport events, and then sleep for 100 ms. It was doing that until the connection is established or the timeout expired.
The new version polls the transport's events, potentially not having many 100 ms waits.


akallabeth (approved):
LGTM

PR #8372, opened 2022-11-04
fix HTTP proxy CONNECT header

akarl10:
https://github.com/FreeRDP/FreeRDP/commit/edeea86dd1ebed7364b0a43d1ef6c6495148f1c7 broke http proxy connect. This fixes that

### DIFF
--- a/libfreerdp/core/proxy.c
+++ b/libfreerdp/core/proxy.c
@@ -524,8 +524,8 @@ static BOOL http_proxy_connect(BIO* bufferedBio, const char* proxyUsername,
 	size_t reserveSize;
 	size_t portLen;
 	size_t hostLen;
-	const char connect[8] = "CONNECT";
-	const char httpheader[17] = " HTTP/1.1" CRLF "Host: ";
+	const char connect[] = "CONNECT ";
+	const char httpheader[] = " HTTP/1.1" CRLF "Host: ";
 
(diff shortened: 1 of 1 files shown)


hardening:

MartinHaimberger (changes requested):
Size of the connect char array is 1 short (including trailing '\0") should be 9.

Also the http header (not your fault) seems to be one char short to include the '\0' if CRLF are two chars ..

akarl10:
> Size of the connect char array is 1 short (including trailing '\0") should be 9.
> 
> Also the http header (not your fault) seems to be one char short to include the '\0' if CRLF are two chars ..

I think this was the reason why it was not obvious why it failed: a few lines later the char is written with
`Stream_Write(s, connect, ARRAYSIZE(connect));` The \0 would be written too. With [8] we hide that fact that on the "last" position there is a \0 (the string is not malloc'ed so there should not be any leak or any other issue.

akarl10:
actually this push is not ready.. I found the next \0 in the username transmission...

I will change to strlen since this is more readable

MartinHaimberger:
@akarl10 true, but this code is still ugly, would be better the string big enought for trailing '\0' and we use strlen or ARRAYSIZE - 1

hardening:

MartinHaimberger (approved):
Nice work, much better readable.
LGTM

PR #8382, opened 2022-11-09
proxy: Allow NLA when using proxy smartcard authentication

mfleisz:
Previously smartcard authentication was only supported with TLS connections so the proxy forced that security mode.

Since we now have a working NLA smartcard auth implementation we can remove that code and leave the correct configuration to the proxy user.


akallabeth (approved):
LGTM

PR #8387, opened 2022-11-11
[core] Unify RDP state machine

akallabeth:
Up to this commit the client and server state machine handling used different return values for state machine changes. This is fixed with this commit:
* Use common enum return values
* Use common helper functions

### DIFF
--- a/libfreerdp/core/peer.c
+++ b/libfreerdp/core/peer.c
@@ -36,7 +36,7 @@
 
 #define TAG FREERDP_TAG("core.peer")
 
-static int peer_recv_pdu(freerdp_peer* client, wStream* s);
+static state_run_t peer_recv_pdu(freerdp_peer* client, wStream* s);
 
 static HANDLE freerdp_peer_virtual_channel_open(freerdp_peer* client, const char* name,
                                                 UINT32 flags)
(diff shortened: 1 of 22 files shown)


mfleisz (commented):
Minor comment - otherwise lgtm

libfreerdp/core/state.c
```diff
@@ -0,0 +1,79 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "state.h"
+
+#include <winpr/string.h>
+
+BOOL state_run_failed(state_run_t status)
```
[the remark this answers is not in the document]
akallabeth: ok, done.

libfreerdp/core/peer.c
```diff
@@ -996,27 +1000,29 @@ static int peer_recv_callback_internal(rdpTransport* transport, wStream* s, void
...
 					                                        &primary))
-						ret = -2;
+						ret = STATE_RUN_FAILED;
 					else if (!freerdp_settings_set_uint32(settings, FreeRDP_MonitorCount, 1))
-						ret = -3;
+						ret = STATE_RUN_FAILED;
 				}
-				if (ret < -1)
+				if (state_run_failed(ret))
 				{
```
hardening, line 1008: Strange statement here, we do nothing if `state_run_failed(ret)`, perhaps the condition should be inverted with a `if (!state_run_failed(ret)) { if (!display_convert_rdp_m...` ?

hardening (approved):
Except my remark about the strange statement, that LGTM

libfreerdp/core/peer.c
akallabeth: tried to avoid another indentation level here, but you´re right.

Issue #8359, opened 2022-10-27
Problem against gateway presenting a sha512 certificate


devcon2012:
Hi,

I experience a problem as I try to use xfreerdp (V2.7.0 build on gentoo) against a gatway using http (rpc works ok). The problem is somehow triggered if the server uses a certificate signed with signature algorithm sha512WithRSAEncryption

To reproduce, try to connect using a gateway presenting a SHA512Hash signed certificate using http transport. Swapping to a sha256WithRSAEncryption certificate removes the problem for me.

I saw similar reports here and build the most recent xfreerdp version version from github, linked against openssl 3.0.2. Here is what happens in both cases:

(V2.7.0, openssl 1.1.1)

/usr/bin/xfreerdp /cert-ignore /bpp:'24' /kbd:'0x407' /t:'http gateway' /u:.... /compression-level:2 /p:.... +decorations /floatbar:sticky:off,default:hidden,show:fullscreen +fonts /network:lan +glyph-cache +offscreen-cache +bitmap-cache /jpeg /relax-order-checks -wallpaper /gfx:RFX +gfx-progressive /log-level:TRACE /v:'srvtestapp3' /g:srvtestapp4 /gt:http

[...]

PR #8388, opened 2022-11-13
rdg: Fallback to rpc if http status is not sent

akarl10:
It seems MS RDG 2016+ does not send a http status code if something with the request or configuration is not ok. It is worth retrying with rpc in that case

probably makes /gt:auto work in that case
https://github.com/FreeRDP/FreeRDP/issues/8359

### DIFF
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -1771,9 +1771,18 @@ static BOOL rdg_establish_data_connection(rdpRdg* rdg, rdpTls* tls, const char* 
 			return FALSE;
 
 		response = http_response_recv(tls, TRUE);
-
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test please

akallabeth (approved):
Overall LGTM

One (nitpicking) remark though: The cleanup code might be better done in `rdg_connect` as the `credssp_auth_free` and `http_context_enable_websocket_upgrade(rdg->http, FALSE);` should be done in any case when `*rpcFallback =TRUE`

akallabeth:
@akarl10 thank you for this nice pull request.
I´ve taken the liberty to add a commit on top unifying the `rpcFallback=TRUE` code cleanup in `rdg_connect`.

akallabeth:
@freerdp-bot test again please, and report if possible ;)

PR #8396, opened 2022-11-15
State transition fix

akallabeth:
* Check return codes of `*_transition_to_state` functions
* Add an event handle to `rdpTransportIo` (custom transport is broken since https://github.com/FreeRDP/FreeRDP/pull/8371 )

### DIFF
--- a/include/freerdp/transport_io.h
+++ b/include/freerdp/transport_io.h
@@ -55,6 +55,8 @@ extern "C"
 	};
 	typedef struct rdp_transport_io rdpTransportIo;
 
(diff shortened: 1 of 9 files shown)


akallabeth:
@freerdp-bot test please

hardening:
I understand why #8371 broke custom transport, if I'm correct that adds in all cases an extra event HANDLE to `transport` to poll in that already ineffective polling loop. Isn't there a way that the custom transport is just a special BIO wired in the normal transport that would return the event during the `BIO_get_event` call, and so the regular transport would have just 2 events to poll (reread event and the BIO handle) ?

hardening (approved):

PR #8400, opened 2022-11-16
fix use of return code from list_provider_keys()

pseudometric:
`list_provider_keys()` returns a Boolean, true == success. But `smartcard_hw_enumerateCerts()` expects the return value on success to be `ERROR_SUCCESS == 0`, and so inverts success/failure.

I ran into this while getting the new PKINIT-in-NLA support to work with a hardware token.


akallabeth:
@freerdp-bot test please

hardening (approved):

PR #8402, opened 2022-11-16
autodetect: prepare for multi-transport

hardening:
Autodetect packets can be transported either in TCP TPKT packets or be contained in multi-transport subheaders (transported on UDP). These changes do the appropriate modifications so that in further developments we can take the transport type in account when treating / writing these packets.

### DIFF
--- a/include/freerdp/autodetect.h
+++ b/include/freerdp/autodetect.h
@@ -30,18 +30,26 @@ extern "C"
 #endif
 	typedef struct rdp_autodetect rdpAutoDetect;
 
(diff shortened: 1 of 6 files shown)


hardening:
Sorry for that, but there is a lot of changes in `autodetect.h` because I passed it to `fromdos` as there was a nasty mix of windows and unix CR.

hardening:
Hum looks like `TestConnect` is not reliable, it fails sometime...

hardening:
@freerdp-bot can you retry the test as `TestConnect` fails mysteriously sometime...

akallabeth (approved):
LGTM

PR #8405, opened 2022-11-16
Release 2 9 0

akallabeth:


ChangeLog
```diff
@@ -1,6 +1,13 @@
-# 20xx-xx-xx Version 2.8.x
+# 2022-11-16 Version 2.9.0
 
 Notewhorth changes:
+* Backported #8252: Support server redirection callback
```
pnowack, line 4: Small nit: It is not a callback (for receiving that PDU). The method actively sends the Server Redirection PDU, so something like `Add support for sending server redirection PDUs`

akallabeth: good point

bmiklautz (approved):

PR #8410, opened 2022-11-16
[core] fixed CONNECTION_STATE_MCS_CREATE_REQUEST parsing

akallabeth:
* Fix various issues with state machine
* Fix some memory leaks
* Improve logging
* Add missing server side parsing code

### DIFF
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -1436,8 +1436,6 @@ static int nla_decode_ts_request(rdpNla* nla, wStream* s)
 	const size_t offset = WinPrAsn1DecReadSequence(&dec, &dec2);
 	if (offset == 0)
 		return -1;
-	if (Stream_GetRemainingLength(s) < offset)
-		return -1;
 	dec = dec2;
 
 	/* version [0] INTEGER */
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -1416,11 +1416,13 @@ fail:
 
 static int nla_decode_ts_request(rdpNla* nla, wStream* s)
 {
-	WinPrAsn1Decoder dec, dec2, dec3;
-	BOOL error;
-	WinPrAsn1_tagId tag;
-	WinPrAsn1_OctetString octet_string;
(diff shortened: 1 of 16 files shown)


libfreerdp/core/nla.c
```diff
@@ -1453,20 +1462,23 @@ static int nla_decode_ts_request(rdpNla* nla, wStream* s)
 		return -1;
 	}
```
hardening, line 1464: From my PoV changes before are ok, but the next changes that save the result of decoding in `rc` are useless and make the more complex to read.

akallabeth: didn´t like the handling of the return as boolean value (`!value`) so I used the `rc=fkt(); if (rc == 0) return;` style

libfreerdp/core/nla.c
```diff
@@ -1432,13 +1433,21 @@ static int nla_decode_ts_request(rdpNla* nla, wStream* s)
 	WLog_DBG(TAG, "<<----- receiving...");
 
 	/* TSRequest */
-	if (!WinPrAsn1DecReadSequence(&dec, &dec2))
+	const size_t offset = WinPrAsn1DecReadSequence(&dec, &dec2);
+	if (offset == 0)
+		return -1;
+	if (Stream_GetRemainingLength(s) < offset)
```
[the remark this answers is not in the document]
akallabeth: removed

mfleisz (commented):
Overall LGTM

I agree with @hardening - the changes in `nla.c` bloats the code, makes it harder to read (best example is the `while` below) and introduces a mixed API usage style compared to how the ASN API is used i.e. in the SSPI kerberos/negotiate implementations.

libfreerdp/core/nla.c
akallabeth: ok, changed to use similar style as before but check for `xxx() == 0` instead of using a return variable.

akallabeth:
@hardening @mfleisz ok, update the pr accordingly.

hardening (approved):

PR #8411, opened 2022-11-16
Fix wlfreerdp pointer cursors

jadahl:
Two commits:

    uwac/input: Set the right serial when setting cursor
    
    The serial in wl_pointer.set_cursor must exactly match the one from
    wl_pointer.enter, it should not use whatever serial for any input class

and

    uwac/input: Don't pass hotspot as offset when attaching cursor buffer
    
    Passing a non-zero offset to cursor buffer and then calling
    wl_pointer.set_cursor doesn't make much sense, as any offset will
    immediately be reset. The protocol specifies the cursor set by
    wl_pointer.set_cursor to be
    
    > The parameters hotspot_x and hotspot_y define the position of
    > the pointer surface relative to the pointer location. Its
    > top-left corner is always at (x, y) - (hotspot_x, hotspot_y),
    > where (x, y) are the coordinates of the pointer location, in
    > surface-local coordinates.
    
    This leaves no room available for any previously set offsets to be kept

### DIFF
--- a/uwac/libuwac/uwac-input.c
+++ b/uwac/libuwac/uwac-input.c
@@ -134,7 +134,7 @@ static UwacReturnCode set_cursor_image(UwacSeat* seat, uint32_t serial)
 
 	if (surface && buffer)
 	{
-		wl_surface_attach(surface, buffer, -x, -y);
+		wl_surface_attach(surface, buffer, 0, 0);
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot please test this nice pr!

hardening (approved):
LGTM

PR #8417, opened 2022-11-18
core: Preserve user provided credentials when being redirected

mfleisz:
This PR fixes an issue where user credentials were lost when connecting to a farm that redirects the client.

During a redirect the connection settings are overridden by the settings stored in `rdp->originalSettings`. However these settings miss any credentials the user provides during the connection phase, thus causing another authentication prompt to appear.


akallabeth (approved):
LGTM

PR #8419, opened 2022-11-19
Use Stream_CheckAndLogRequiredLength, fix WLog TAG

akallabeth:
* Log stream length requirement violations
* Use proper defines for WLog tags


mfleisz (approved):
LGTM

PR #8428, opened 2022-11-21
[ntlm]: use the hash algorithm as mandated by rfc5929

akarl10:
This is for making ntlm (and probably Negotiate/Kerberos too) with gateways that use sha384 or sha512 certificate hashes.


libfreerdp/crypto/crypto.c
```diff
@@ -289,6 +289,30 @@ BYTE* crypto_cert_hash(X509* xcert, const char* hash, UINT32* length)
 	return fp;
 }
 
+char* cryto_cert_hash_use_signature_alg(X509* xcert)
```
akallabeth, line 292: @akarl10 ok, if you want to use it this way better use a `char* crypto_cert_get_signature_alg(X509* xcert, char* buffer, size_t buffersize)` function signature

akallabeth (approved):
Basically LGTM, I´d just rename the function and use a supplied buffer here.

akallabeth:
@freerdp-bot test this nice pr please!

libfreerdp/crypto/tls.c
```diff
@@ -616,7 +616,21 @@ static SecPkgContext_Bindings* tls_get_channel_bindings(X509* cert)
 	SEC_CHANNEL_BINDINGS* ChannelBindings;
 	SecPkgContext_Bindings* ContextBindings;
 	const size_t PrefixLength = strnlen(TLS_SERVER_END_POINT, ARRAYSIZE(TLS_SERVER_END_POINT));
-	BYTE* CertificateHash = crypto_cert_hash(cert, "sha256", &CertificateHashLength);
+
```
akallabeth, line 619: Maybe add a link to the RFX in a comment here that we know what is going on in the future

libfreerdp/crypto/crypto.c
akallabeth: another option would be to return a `WINPR_MD_TYPE`

akarl10: I think maybe the second option would be better. The buffer could be hidden inside the function and at the end of the day WINPR_MD_TYPE is what we need. `winpr_md_type_to_string` returns a pointer to a const char*, so we don't need to bother about the allocation

libfreerdp/crypto/tls.c
akarl10: definitly makes sense

libfreerdp/crypto/crypto.c
```diff
@@ -289,6 +289,30 @@ BYTE* crypto_cert_hash(X509* xcert, const char* hash, UINT32* length)
...
+	hashName = calloc(50, sizeof(char));
+	if (hashName == NULL)
+		return NULL;
+
+	X509_ALGOR_get0(&sigAlgObj, NULL, NULL, tsig_alg);
+	OBJ_obj2txt(hashName, 50, sigAlgObj, 0);
+
+	/* OBJ_obj2txt returns hashWithEncryption, ex. sha256WithRSAEncrpytion;
+	 * we need only the part before With */
+	char* w = strstr(hashName, "With");
```
akallabeth, line 309: this check is not sufficient, it only covers `RSA` certificates.
Might be better to use `X509_get_signature_nid` and do a big switch for all (currently) supported algorighm combinations (maybe there is something simpler?)

libfreerdp/crypto/crypto.c
akallabeth: good point

akarl10: is it allowed to include something from winpr in inlcude/freerdp/crypto? I get `crypto.h:60:21: error: unknown type name ‘WINPR_MD_TYPE’`

akallabeth: @akarl10 yes, just not the other way around

akallabeth:
@akarl10 https://stackoverflow.com/questions/36076998/how-to-get-the-type-of-signature-hash-algorithm-from-certificate might be of interest, that way you get the hash algorithm `NID` directly

akallabeth:
@freerdp-bot test

akallabeth:

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test

akallabeth:

akallabeth:
@freerdp-bot test please

akallabeth:

PR #8442, opened 2022-11-22
Pointer cache split

akallabeth:
Properly handle the different pointer caches client and server side

### DIFF
--- a/include/freerdp/settings.h
+++ b/include/freerdp/settings.h
@@ -857,6 +857,7 @@ typedef struct
 #define FreeRDP_BitmapCachePersistFile (2503)
 #define FreeRDP_ColorPointerFlag (2560)
 #define FreeRDP_PointerCacheSize (2561)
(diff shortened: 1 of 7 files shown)


mfleisz:
I don't think we really need to keep separate track of these caches.

Reading the spec it seems the color pointer cache is the legacy cache (fixed to 24 bpp pointers) and the pointer cache is its generic replacement (supporting arbitrary bit depths).

So I guess handling of the pointer cache size like it is done now in `rdp_apply_pointer_capability_set` should be good enough (no need to keep two different cache sizes around).

Also noted that `FreeRDP_GatewayHttpExtAuthSspiNtln` is misspelled.

akallabeth:
@mfleisz these caches are not separate but one big cache according to spec. Only their size might differ (e.g. you can turn of one or the other by setting the size to 0)

mfleisz (approved):

PR #8453, opened 2022-11-24
Multitransport plus other fixes / improvements

hardening:


libfreerdp/core/gcc.c
```diff
@@ -2337,7 +2337,8 @@ BOOL gcc_read_client_multitransport_channel_data(wStream* s, rdpMcs* mcs, UINT16
 	if (blockLength < 4)
 		return FALSE;
 
-	Stream_Read_UINT32(s, settings->MultitransportFlags);
+	Stream_Read_UINT32(s, remoteFlags);
+	settings->MultitransportFlags &= remoteFlags; /* merge local and remote flags */
```
akallabeth, line 2341: should be read as is, the merge happens later on when settings are merged

libfreerdp/core/gcc.c
```diff
@@ -2386,15 +2388,31 @@ BOOL gcc_read_server_multitransport_channel_data(wStream* s, rdpMcs* mcs)
 	if (!Stream_CheckAndLogRequiredLength(TAG, s, 4))
 		return FALSE;
 
-	Stream_Read_UINT32(s, settings->MultitransportFlags); /* flags, merge with client setting */
+	Stream_Read_UINT32(s, remoteFlags);
+	settings->MultitransportFlags &= remoteFlags; /* merge with client setting */
```
akallabeth, line 2392: should be read as is, merge later on when settings are merged

akallabeth (changes requested):

akallabeth:
@hardening settings have been split, there are:
1. Initial settings the client/server is configured with
2. A copy of these for each connection/peer
3. Settings of the other end that have been sent (capabilities)

the reading of capabilities is now split from the settings merge, so you always have the remote end settings readily available for debugging/checking....

hardening:
> @hardening settings have been split, there are:
> 
>     1. Initial settings the client/server is configured with
> 
>     2. A copy of these for each connection/peer
> 
>     3. Settings of the other end that have been sent (capabilities)
> 
> 
> the reading of capabilities is now split from the settings merge, so you always have the remote end settings readily available for debugging/checking....

Ok will have a look

libfreerdp/core/multitransport.c
```diff
@@ -101,20 +74,37 @@ BOOL multitransport_server_send_request(rdpMultitransport* multitransport)
...
+	Stream_Write_UINT16(s, reqProto); /* requestedProtocol (2 bytes) */
+	Stream_Zero(s, 2);                /* reserved (2 bytes) */
+	Stream_Write(s, cookie, 16);      /* securityCookie (16 bytes) */
 
-	return rdp_send_message_channel_pdu(multitransport->rdp, s, SEC_TRANSPORT_REQ);
+	return rdp_send_message_channel_pdu(multi->rdp, s, SEC_TRANSPORT_REQ);
 }
 
-BOOL multitransport_client_send_response(rdpMultitransport* multitransport, HRESULT hr)
+state_run_t multitransport_server_request(rdpMultitransport* multi, UINT32 reqId, UINT16 reqProto,
```
akallabeth, line 85: any reason the `cookie` is not part of `rdpMultitransport`?

libfreerdp/core/gcc.c
```diff
@@ -2387,15 +2389,31 @@ BOOL gcc_read_server_multitransport_channel_data(wStream* s, rdpMcs* mcs)
...
+	rdpContext* context;
+	rdpSettings* settings;
+
+	context = transport_get_context(mcs->transport);
+	WINPR_ASSERT(context);
+
+	settings = context->settings;
+	WINPR_ASSERT(settings);
+
+	if (settings->SupportMultitransport && settings->MultitransportFlags)
```
akallabeth, line 2411: Might be a good idea to move this check out of this function (e.g. keep the `write` and don´t do an `maybe write`)

libfreerdp/core/multitransport.c
```diff
@@ -24,74 +24,47 @@
...
-	UINT32 requestId;         /* requestId (4 bytes) */
-	UINT16 requestedProtocol; /* requestedProtocol (2 bytes) */
-	UINT16 reserved;          /* reserved (2 bytes) */
-	BYTE securityCookie[16];  /* securityCookie (16 bytes) */
+	RDPTUNNEL_ACTION_CREATEREQUEST = 0x00,
+	RDPTUNNEL_ACTION_CREATERESPONSE = 0x01,
+	RDPTUNNEL_ACTION_DATA = 0x02
 };
 
-static BOOL multitransport_compare(const rdpMultitransport* srv, const rdpMultitransport* client)
```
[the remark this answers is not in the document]
hardening: Couldn't make it match with real code handling UDP connections, sorry

hardening:
Ok fixed everything

akallabeth (commented):
Overall LGTM

libfreerdp/core/multitransport.h
```diff
@@ -36,16 +36,37 @@ typedef enum
...
+
+struct rdp_multitransport
+{
+	rdpRdp* rdp;
+
+	MultiTransportRequestCb MtRequest;
+	MultiTransportResponseCb MtResponse;
+
+	/* server-side data */
+	UINT32 reliableReqId;
```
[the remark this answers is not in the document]
hardening: I don't clearly understand the point, we're already in a private header. I don't see why we shall make this struct more opaque inside of FreeRDP itself (I must be be missing something)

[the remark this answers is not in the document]
akallabeth: The point being, if it is not necessary don´t make it visible anywhere.
Many issues we had was the cross referencing of data in structs stored in structs without a proper way to ensure that is correct.
by keeping everything accessible from outside (be it from within freerdp or outsied) small (e.g. what is actually meant to be read/written) it is easier to trace problems and prevent accidental read/write of wrong data (we are not strong with writing documentation as well, so this is a way to keep it less error prone)

libfreerdp/core/peer.c
```diff
@@ -960,13 +961,26 @@ static state_run_t peer_recv_callback_internal(rdpTransport* transport, wStream*
 		case CONNECTION_STATE_MULTITRANSPORT_BOOTSTRAPPING_REQUEST:
 			if (settings->SupportMultitransport)
 			{
-				if (!multitransport_server_send_request(rdp->multitransport))
-					ret = STATE_RUN_FAILED;
-				else
+				/* TODO: this counter shall end up in a listener or server instance */
+				static UINT32 reqId = 0;
```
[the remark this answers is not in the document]
hardening: Well this counter is something kinda global, we could say bound to the listener component. 

The global scenario is an accepted TCP peer will grab one of these requestId, will send the packet with id and cookie over TCP. 
After a while we're gonna have an incoming UDP connection (on a UDP listener or the UDP part of the listener), establishing the UDP transport, and in the final packet it will send to the corresponding requestId and cookie. When we receive that, we must inquiry current TCP connections (that mean that the listener or another component must keep track of these), and if we have a reqId/cookie match, then the incoming UDP transport will be set as UDP part of the corresponding RDP session.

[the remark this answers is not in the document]
hardening: So to answer the comment, the cookie could/should be generated in `multitransport_server_request`, but for the request counter we must have a larger view.

libfreerdp/core/multitransport.h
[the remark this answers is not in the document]
hardening: Well that's our big disagreement on boundaries: clear rules associated with the kind of boundaries.
Anyway I'm not sure `rdpMultitransport` really needs to be visible outside of multitransport.c.

libfreerdp/core/peer.c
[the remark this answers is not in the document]
hardening: Ok moved the logic in multitransport_server_request

libfreerdp/core/multitransport.h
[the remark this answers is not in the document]
hardening: Made private again.

libfreerdp/core/multitransport.c
```diff
@@ -124,47 +133,72 @@ BOOL multitransport_client_send_response(rdpMultitransport* multitransport, HRES
...
+state_run_t multitransport_recv_response(rdpMultitransport* multi, wStream* s)
 {
-	rdpMultitransport multi = { 0 };
+	WINPR_ASSERT(multi && multi->rdp);
+	WINPR_ASSERT(s);
 
-	WINPR_ASSERT(multitransport);
-	WINPR_ASSERT(hr);
+	rdpSettings* settings = multi->rdp->settings;
+	if (!settings->ServerMode)
```
akallabeth, line 147: `WINPR_ASSERT` the settings

libfreerdp/core/multitransport.c
```diff
@@ -124,47 +133,72 @@ BOOL multitransport_client_send_response(rdpMultitransport* multitransport, HRES
...
+                                         const BYTE* cookie)
+{
+	return multitransport_client_send_response(multi, reqId, E_ABORT) ? STATE_RUN_SUCCESS
+	                                                                  : STATE_RUN_FAILED;
+}
+
+static state_run_t multitransport_server_handle_response(rdpMultitransport* multi, UINT32 reqId,
+                                                         UINT32 hrResponse)
+{
+	rdpRdp* rdp = multi->rdp;
```
akallabeth, line 175: `WINPR_ASSERT` input variables (used as function pointer, so `static` does not apply)

libfreerdp/core/multitransport.c
```diff
@@ -124,47 +133,72 @@ BOOL multitransport_client_send_response(rdpMultitransport* multitransport, HRES
...
+		return STATE_RUN_FAILED;
+
+	return STATE_RUN_CONTINUE;
 }
 
 rdpMultitransport* multitransport_new(rdpRdp* rdp, UINT16 protocol)
 {
+	WINPR_ASSERT(rdp);
+
+	rdpSettings* settings = rdp->settings;
```
akallabeth, line 187: `WINPR_ASSERT` settings

libfreerdp/core/multitransport.h
```diff
@@ -36,16 +36,22 @@ typedef enum
 	INITIATE_REQUEST_PROTOCOL_UDPFECL = 0x02
 } MultitransportRequestProtocol;
 
-FREERDP_LOCAL state_run_t multitransport_client_recv_request(rdpMultitransport* multitransport,
-                                                             wStream* s);
-FREERDP_LOCAL BOOL multitransport_server_send_request(rdpMultitransport* multitransport);
+typedef state_run_t (*MultiTransportRequestCb)(rdpMultitransport* multi, UINT32 reqId,
```
akallabeth, line 39: Since `cookie` is not a string it might be a good idea to add a `cookieLen` argument as well (and not implicitly assume a certain size)

akallabeth (approved):
Overall LGTM

libfreerdp/core/multitransport.h
hardening: Cookie is always 16 bytes, fixed size.

hardening: no it can't change change it's hardcoded in the protocol, there's no way to have it to anything else

akallabeth: or is it `\0` terminated?

hardening: The cookie is just 16 bytes of random data that will have to be re-given on the UDP channel at the end of the negotiation, with the request id it make it match an incoming UDP connection with its TCP counterpart.

PR #8455, opened 2022-11-25
Rdp security checks fixed

akallabeth:
* Unify RDP security rc4 encryption/decryption key handling in wrapper functions
* Add logging and proper checks for states that should not happen


akallabeth:
@PF4Public this should address the problem you mentioned (at least it will properly disconnect the session in that case)

PF4Public:
I see you're targeting master. Will it apply over 2.9 branch?

hardening (approved):
LGTM

PR #8459, opened 2022-11-28
[core,client] Added logging for finalize flags

akallabeth:
Log an error if expected finalize flags did not match what we got.


hardening (approved):
Looks ok for me, just wondering about the log...

akallabeth:
@hardening well, the order of messages is as suggested in `[MS-RDPBCGR]`, the log will tell us if that is actually in use by servers (and which ones)

akallabeth:
@akarl10 I´ve added a simple state machine modification, now there should only be log messages for unexpected events but the connection should not be disconnected.

PR #8461, opened 2022-11-29
[winpr,smartcard] fix PCSC_SCardListReaders

akallabeth:


winpr/libwinpr/smartcard/smartcard_pcsc.c
```diff
@@ -981,11 +980,13 @@ static LONG WINAPI PCSC_SCardListReadersW(SCARDCONTEXT hContext, LPCWSTR mszGrou
 			return SCARD_E_NO_MEMORY;
 	}
 
+	status =
+	    PCSC_SCardListReaders_Internal(hContext, mszGroupsA, (LPSTR*)&mszReadersA, pcchReaders);
 	if (status == SCARD_S_SUCCESS)
```
hardening, line 985: I must test, but I'd bet that it doesn't work as the caller expect a list of wstrings

akallabeth: @hardening yes, and the `UTF8` list is converted to a `WCHAR` list a few lines below

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -946,20 +963,29 @@ static SECURITY_STATUS SEC_ENTRY kerberos_InitializeSecurityContextA(
...
-	sspi_SecureHandleSetLowerPointer(phNewContext, context);
-	sspi_SecureHandleSetUpperPointer(phNewContext, KERBEROS_SSP_NAME);
+		sspi_SecureHandleSetLowerPointer(phNewContext, context);
+		sspi_SecureHandleSetUpperPointer(phNewContext, KERBEROS_SSP_NAME);
+	}
 
 cleanup:
 
+{
+	/* second_ticket is not allocated */
```
fifthdegree, line 984: Maybe my brain's still asleep, but I don't see the point of setting memory about to be de-allocated to zero

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -946,20 +963,29 @@ static SECURITY_STATUS SEC_ENTRY kerberos_InitializeSecurityContextA(
...
+	krb5_data edata = { 0 };
+	in_creds.second_ticket = edata;
+	krb5_free_cred_contents(context->ctx, &in_creds);
```
fifthdegree, line 987: Will this also free `.server` and `.client`?

fifthdegree: Never mind, of course it will. I don't know what I was thinking

winpr/libwinpr/sspi/Kerberos/kerberos.c
fifthdegree: Because the `krb5_cred` structure in this case is not used as a "credential"; rather it's just used to organize different pieces of information (the client and server principals for the new cred)

hardening (approved):

PR #8463, opened 2022-11-29
[winpr,crt] fix unicode conversion functions

akallabeth:
There are subtle differences between MultiByteToWideChar, WideCharToMultibyte and the ICU equivalents


akallabeth:
@mfleisz @hardening fixed issues with the unicode conversion backends, now it respects NULL termination or not and returns the number of characters converted.
It also ignores any NULL termination once the input buffer length is set.

Had to fix a few test cases, as the converted character length might differ from the input length.

hardening (approved):

PR #8467, opened 2022-11-30
rdp: notify the Activate state as soon as it happens

hardening:
Without the patch, we parse more packets and the calling code doesn't have the opportunity to invoke PostConnect callback (make the connection not work in the proxy)


akallabeth (approved):
LGTM

PR #8470, opened 2022-11-30
proxy: fix channel id shift between front and back

hardening:
When some channels are filtered, some misalignment of channel ids could happen. This patch keeps track of the back and front channel ids to correctly identify a channel and send packets with the correct channel id.


akallabeth (approved):
LGTM

PR #8472, opened 2022-11-30
State machine relax

akallabeth:


bmiklautz (approved):

libfreerdp/core/connection.c
```diff
@@ -226,6 +225,59 @@ static BOOL rdp_client_reset_codecs(rdpContext* context)
...
+			WLog_ERR(TAG, "error retrieving connection events");
+			return FALSE;
+		}
+
+		wstatus = WaitForMultipleObjectsEx(nevents, events, FALSE, (dueDate - now), TRUE);
+		switch (wstatus)
+		{
+			case WAIT_TIMEOUT:
+				/* will make us quit with a timeout */
+				now = dueDate + 1;
```
mfleisz, line 254: I know this is not new - but how will this make us quit if `now` will be assigned `GetTickCount64()` at the end of the `for`-loop?

mfleisz (commented):
Apart from that one remark LGTM

libfreerdp/core/connection.c
akallabeth: the new assignment happens in the `for (; now < dueDate; now = GetTickCount64())`, so it is first checked against `dueDate`

mfleisz: @akallabeth this is not how a `for` loop works in C: https://en.cppreference.com/w/c/language/for

`cond-expression`: is evaluated **before** the loop body. If the result of the expression is zero, the loop statement is exited immediately.
`iteration-expression`: is evaluated **after** the loop body and its result is discarded. After evaluating iteration-expression, control is transferred to cond-expression.

So this definitely never checks the above assignment in the condition

PR #8475, opened 2022-12-01
[rdg] fix FAILED detection

akarl10:
FAILED is a check for errorCode < 0, but in this case errorCode is a UINT

also include a fix for the PAA coockie length calculation (only the packet length was updated, but the last byte was not written and so the connection stalled)

### DIFF
--- a/libfreerdp/core/gateway/rdg.c
+++ b/libfreerdp/core/gateway/rdg.c
@@ -1052,7 +1052,8 @@ static BOOL rdg_send_tunnel_request(rdpRdg* rdg)
 			return FALSE;
 		}
 
-		packetSize += 2 + (UINT32)(PAACookieLen + 1) * sizeof(WCHAR);
+		PAACookieLen += 1; /* include \0 */
+		packetSize += 2 + (UINT32)(PAACookieLen) * sizeof(WCHAR);
 		fieldsPresent = HTTP_TUNNEL_PACKET_FIELD_PAA_COOKIE;
 	}
 
@@ -1335,7 +1336,7 @@ static BOOL rdg_process_handshake_response(rdpRdg* rdg, wStream* s)
 	         ", extendedAuth=%s",
 	         error, verMajor, verMinor, serverVersion, extended_auth_to_string(extendedAuth));
 
-	if (FAILED(errorCode))
+	if (FAILED((HRESULT)errorCode))
 	{
 		WLog_ERR(TAG, "Handshake error %s", error);
 		freerdp_set_last_error_log(rdg->context, errorCode);
@@ -1435,7 +1436,7 @@ static BOOL rdg_process_tunnel_response(rdpRdg* rdg, wStream* s)
 	WLog_DBG(TAG, "serverVersion=%" PRId16 ", errorCode=%s, fieldsPresent=%s", serverVersion, error,
 	         tunnel_response_fields_present_to_string(fieldsPresent));
 
-	if (FAILED(errorCode))
+	if (FAILED((HRESULT)errorCode))
 	{
 		WLog_ERR(TAG, "Tunnel creation error %s", error);
 		freerdp_set_last_error_log(rdg->context, errorCode);
@@ -1470,7 +1471,8 @@ static BOOL rdg_process_tunnel_authorization_response(rdpRdg* rdg, wStream* s)
 	WLog_DBG(TAG, "errorCode=%s, fieldsPresent=%s", error,
 	         tunnel_authorization_response_fields_present_to_string(fieldsPresent));
 
-	if (FAILED(errorCode))
+	/* [MS-TSGU] 3.7.5.2.7 */
+	if (errorCode != S_OK && errorCode != E_PROXY_QUARANTINE_ACCESSDENIED)
 	{
 		WLog_ERR(TAG, "Tunnel authorization error %s", error);
 		freerdp_set_last_error_log(rdg->context, errorCode);
@@ -1549,7 +1551,7 @@ static BOOL rdg_process_channel_response(rdpRdg* rdg, wStream* s)
 	WLog_DBG(TAG, "channel response errorCode=%s, fieldsPresent=%s", error,
 	         channel_response_fields_present_to_string(fieldsPresent));
 
-	if (FAILED(errorCode))
+	if (FAILED((HRESULT)errorCode))
 	{
 		WLog_ERR(TAG, "channel response errorCode=%s, fieldsPresent=%s", error,
 		         channel_response_fields_present_to_string(fieldsPresent));


libfreerdp/core/gateway/rdg.c
```diff
@@ -1052,7 +1052,8 @@ static BOOL rdg_send_tunnel_request(rdpRdg* rdg)
 			return FALSE;
 		}
 
-		packetSize += 2 + (UINT32)(PAACookieLen + 1) * sizeof(WCHAR);
+		PAACookieLen += 1; /* include \0 */
```
akallabeth, line 1055: this looks wrong, if the content is `WCHAR` then the `\0` must be `sizeof(WCHAR)`

libfreerdp/core/gateway/rdg.c
```diff
@@ -1335,7 +1336,8 @@ static BOOL rdg_process_handshake_response(rdpRdg* rdg, wStream* s)
 	         ", extendedAuth=%s",
 	         error, verMajor, verMinor, serverVersion, extended_auth_to_string(extendedAuth));
 
-	if (FAILED(errorCode))
+	/* [MS-TSGU] 3.7.5.2.3 */
+	if (errorCode != 0)
```
akallabeth, line 1340: according to `[MS-RSGU]` this is an `UNSIGNED` but a `HRESULT`, go figure...

akallabeth (changes requested):
the size calculation looks wrong.

libfreerdp/core/gateway/rdg.c
akallabeth: sorry, looks like I need some coffee

akallabeth (approved):
LGTM,
sorry, misread the line before.

akallabeth:
@freerdp-bot test

libfreerdp/core/gateway/rdg.c
mfleisz: I think the old code was correct - `FAILED` will do the correct casting (see https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-failed).

akarl10:
```
grep -r '#define FAILED'
winpr/include/winpr/error.h:#define FAILED(hr) (((hr)) < 0)
```
I noted that FAILED never triggered because it did not disconnect when getting a errorCode=E_PROXY_COOKIE_AUTHENTICATION_ACCESS_DENIED

mfleisz: @akarl10 Yes - obviously our `FAILED` implementation is not correct

akarl10: > according to `[MS-RSGU]` this is an `UNSIGNED` but a `HRESULT`, go figure...

i just looked what is used in that specific case and then copied the allowed values from [MS-TSGU]
This is why sometimes I have written != 0 and sometime != S_OK

mfleisz: Here is a description of the `HRESULT` type: https://learn.microsoft.com/en-us/office/client-developer/outlook/mapi/hresult

Basically it is a 32-bit value that encodes various kinds of return values (errors, success, warnings, ...)

`HRESULT` is defined as a signed type so you can easily check for failure (most-significant bit is set - so the value is negative) or success (value is `>= 0`). Checking for `!= 0` is not 100% correct therefore.

akarl10: > I think the old code was correct - `FAILED` will do the correct casting (see https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-failed).

well, I can change the define in winpr and leave only the `errorCode != S_OK && errorCode != E_PROXY_QUARANTINE_ACCESSDENIED` case

mfleisz: > > I think the old code was correct - `FAILED` will do the correct casting (see https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-failed).
> 
> well, I can change the define in winpr and leave only the `errorCode != S_OK && errorCode != E_PROXY_QUARANTINE_ACCESSDENIED` case

I think that's the proper fix - but the check should remain `if (FAILED(errorCode))` or do you need any special handling for `E_PROXY_QUARANTINE_ACCESSDENIED`?

akarl10: > 

generally speaking, yes, negative values are errors, positive successes.
in this case tough (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tsgu/390deecf-4c75-4838-8be5-64be3083d3eb) the exact values that are success are defined

akallabeth: @akarl10 no, do not touch the define, that is correct. you need to read the value into a `HRESULT`, so just change the definition of the variable and you should be fine.

> > I think the old code was correct - `FAILED` will do the correct casting (see https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-failed).
> 
> well, I can change the define in winpr and leave only the `errorCode != S_OK && errorCode != E_PROXY_QUARANTINE_ACCESSDENIED` case

winpr/include/winpr/error.h
```diff
@@ -158,8 +158,8 @@ static INLINE HRESULT HRESULT_FROM_WIN32(unsigned long x)
 
 #define HRESULT_SEVERITY(hr) (((hr) >> 31) & 0x1)
 
-#define SUCCEEDED(hr) (((hr)) >= 0)
```
akallabeth, line 161: sorry, this is not really good.
the define should check if a `HRESULT`, if you supply something else then you´re at fault.

Just change the variable you want to check to a `HRESULT` and do not touch the header.

akallabeth: ok, sorry for the noise, looks like `MSDN` also proposes this (error hiding) way of defining `FAILED` :/

akallabeth (approved):

akallabeth:
@freerdp-bot test

winpr/include/winpr/error.h
akarl10: ok. in reality I 1:1 copied that over from https://learn.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-succeeded

now I casted to HRESULT, hope this is how you meant.
As you said, the statusCode is defined as `statusCode (4 bytes): An unsigned integer representing errors that are detected by the RDG server`

Don't know if other protocol definitions are like this one, but MS-TSGU feels like it was written after they implemented this protocol and a manager forced them to document it.

libfreerdp/core/gateway/rdg.c
```diff
@@ -1335,7 +1336,7 @@ static BOOL rdg_process_handshake_response(rdpRdg* rdg, wStream* s)
 	         ", extendedAuth=%s",
 	         error, verMajor, verMinor, serverVersion, extended_auth_to_string(extendedAuth));
 
-	if (FAILED(errorCode))
+	if (FAILED((HRESULT)errorCode))
```
akallabeth, line 1339: ok, now I think we have seen all versions except the obvious (`HRESULT errorCode;`  instead of `UINT32 errorCode;` as definition :))
joking aside, sorry for the misleading last comments, the define in winpr is really faulty, seen the (very annoying) documentation in msdn. (don´t really like it as you can just throw in anything without the compiler complaining)

anyway, just leave it, well merge and do the rest from here on (to avoid further confusion, as mentioned before, need coffee ;))

akallabeth:
@freerdp-bot test

PR #8490, opened 2022-12-06
Xfreerdp cleanup

akallabeth:
* Allow `xfreerdp` to run without `DISPLAY` variable for `/auth-only`
* Adds support for transparent windows (`RAILS`)
* Adds a `PostFinalDisconnect` client callback that is called after everything shut down (`PostDisconnect` is called before channel shutdown, this one is similar to `PreConnect`)
* Some `X11` related code cleanups, specifically drawing path and modifier map updates
* RAIL transparency in this PR currently only works with `AVC444`, other codecs still need some follow up bug fixing


akallabeth:
@freerdp-bot test this again please

akallabeth:
@freerdp-bot test

channels/rdp2tcp/client/rdp2tcp_main.c
```diff
@@ -168,7 +171,9 @@ static DWORD WINAPI copyThread(void* data)
 			goto fail;
 		}
 
-		WaitForSingleObject(plugin->writeComplete, INFINITE);
+		handles[0] = plugin->writeComplete;
+		handles[1] = freerdp_abort_event(plugin->channelEntryPoints.context);
+		status = WaitForMultipleObjects(2, handles, FALSE, INFINITE);
 		ResetEvent(plugin->writeComplete);
```
hardening, line 177: Here we would ResetEvent `plugin->writeComplete` unconditionnally even if status is an error, that seems fishy

client/X11/xf_rail.c
```diff
@@ -209,63 +210,61 @@ void xf_rail_end_local_move(xfContext* xfc, xfAppWindow* appWindow)
...
-	xf_rail_invalidate_region(xfc, &invalidRegion);
-	region16_uninit(&invalidRegion);
+	BOOL rc = TRUE;
+	ULONG_PTR* pKeys = NULL;
+	size_t count = 0;
+
+	WINPR_ASSERT(xfc);
+	WINPR_ASSERT(rect);
+
+	if (xfc->railWindows)
```
hardening, line 256: Perhaps we could use `HashTable_Foreach` ?

hardening (approved):
A few remarks but globally LGTM

channels/rdp2tcp/client/rdp2tcp_main.c
akallabeth: @hardening well, as the other condition is `terminate the session` I didn´t think the check would be useful, but yes, could be added.

client/X11/xf_rail.c
akallabeth: done

channels/rdp2tcp/client/rdp2tcp_main.c
akallabeth: but fixed nevertheless.

PR #8495, opened 2022-12-06
[drdynvc] assert and proper cleanup

akallabeth:
@hardening there was a recursive cleanup loop in the latest drdynvc code, fixed that with a hash table cleanup function.

### DIFF
--- a/channels/drdynvc/client/drdynvc_main.c
+++ b/channels/drdynvc/client/drdynvc_main.c
@@ -31,6 +31,8 @@
 
 #define TAG CHANNELS_TAG("drdynvc.client")
 
+static void dvcman_channel_free(DVCMAN_CHANNEL* channel);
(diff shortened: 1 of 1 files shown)


channels/drdynvc/client/drdynvc_main.c
```diff
@@ -242,19 +254,31 @@ static void wts_listener_free(void* arg)
 
 static BOOL channelIdMatch(const void* k1, const void* k2)
 {
+	WINPR_ASSERT(k1);
```
hardening, line 257: These checks are too paranoïd, it's a function that will  be called on every item

akallabeth: @hardening that is what `assert` is for ;)
just build with `-DWITH_VERBOSE_WINPR_ASSERT=OFF` and they are gone in your build ;)

PR #8509, opened 2022-12-09
[build] make OpenSSL and mbedtls mutually exclusive

akallabeth:


hardening:
I guess this can only be a temporary workaround, as IIRC some people (devolution) do build with mbed and openssl, but in the meantime it would do the job.

awakecoding:
We build WinPR with mbedTLS, and then FreeRDP with OpenSSL against the prebuilt WinPR that uses mbedTLS. Please do not break this build scenario 🙏

akallabeth:
@awakecoding should not be an issue with this, it just ensures that not both `mbedTLS` and `OpenSSL` can be defined at once.

awakecoding:
I think that should be fine, we build WinPR in a first step with mbedTLS only, and then FreeRDP with OpenSSL only, against WinPR which is linked against mbedTLS. You can see our build recipes here, there's one for freerdp separate from winpr: https://github.com/Devolutions/conan-public/tree/master/recipes

bmiklautz (approved):

bmiklautz:
Also think that this shouldn't cause any problems as only openssl or mbedTLS at a time makes sense (and works ;)

PR #8512, opened 2022-12-09
[winpr,utils] pretty print stacktrace

akallabeth:
do not add an additional newline in the messages.


hardening (approved):
Perfect, you're my hero :D

PR #8513, opened 2022-12-09
[winpr,sspi] fix kerberos resource cleanup

akallabeth:
kerberos cache and keytab were not properly cleaned up for cases where the credentials were not available in kerberos


hardening (approved):
LGTM

PR #8521, opened 2022-12-12
Kbd autorepeat api fix

akallabeth:


hardening:
That looks ok for me, anyway I suspect that the change of behaviour might impact project that would like to migrate to 3.0. Perhaps it's the good timing to open a "migrate to 3.0" page in the WIKI or have a specific doc file in the repo ?

akallabeth:
@hardening https://github.com/FreeRDP/FreeRDP/wiki/Migration-from-2.x-to-3.0
if you would be so kind to add your review we can merge ;)

hardening (approved):

PR #8524, opened 2022-12-12
Gdi bitmap update fail log

akallabeth:


libfreerdp/gdi/graphics.c
```diff
@@ -214,12 +223,19 @@ static BOOL gdi_Bitmap_Decompress(rdpContext* context, rdpBitmap* bitmap, const
 			const size_t dstSize = SrcSize * dbpp / sbpp;
 
 			if (dstSize < bitmap->length)
+			{
+				WLog_ERR(TAG, "[%s] dstSize %" PRIu32 " < bitmap->length %" PRIu32, __FUNCTION__,
```
mfleisz, line 227: For `dstSize` we should use `PRIuz`

mfleisz (commented):
Just a minor comment - otherwise lgtm

libfreerdp/codec/include/bitmap.c
```diff
@@ -130,8 +145,11 @@ static INLINE BOOL RLEDECOMPRESS(const BYTE* pbSrcBuffer, UINT32 cbSrcBuffer, BY
 		*/
 		code = ExtractCodeId(*pbSrc);
 
+		WLog_VRB(TAG, "[%s] pbSrc=%p code=%s, rem=%" PRIuz, __FUNCTION__, pbSrc,
```
mfleisz, line 148: I am not sure we want to have this inside the loop of a decoding function. This will be very very noisy if using verbose logging while offering not a lot of useful information and it will waste quite a couple of cycles even in release builds on every iteration.

mfleisz (commented):
I haven't done any tests but I have a feeling that after this PR we will have quite some performance hit with all the pedantic `WINPR_ASSERT`ing (functions like `write_pixel_` will get called many many times) - especially since `WITH_VERBOSE_WINPR_ASSERT` is still a default ON even in release build iirc ...

libfreerdp/codec/include/bitmap.c
akallabeth: @mfleisz the `WINPR_ASSERT` stuff shouldn´t matter, as we will deactivate the default once the `master` branch stabilizes. (then these checks are no longer compiled in). (so, anyone interested can already deactivate it manually, but this way we get feedback from the lazy ones :grin: )
as for the log message, true, I´ll add a define like the other extended debug stuff optionally compiled.

akallabeth: @mfleisz ok, added `WITH_DEBUG_CODECS` and deactivated all non `ERROR`, `INFO` or `WARN` messages in interleaved/progressive/...

mfleisz (approved):
lgtm - maybe someone else wants to also have a look as well since quite some parts of the codecs got touched so we don't miss anyhing

akallabeth:
@mfleisz ran extensive tests for comparison, every result so far identical with and without the debug option activated. (that is, aside from the verbose/less verbose logging)

PR #8527, opened 2022-12-13
[channels,rdpsnd] fix rdpsnd channel cleanup

akallabeth:
The rdpsnd channel needs to keep the resources allocated until the last user (static, dynamic channel) is terminated.


mfleisz (approved):
lgtm

PR #8531, opened 2022-12-13
[codec,xcrush] follow up to #8529

akallabeth:
There was still an issue left, if diff == 0 then the loop counter did never increment.
Skip this case now completely as the memory does not need to be copied anyway.

### DIFF
--- a/libfreerdp/codec/xcrush.c
+++ b/libfreerdp/codec/xcrush.c
@@ -759,12 +759,12 @@ static INLINE size_t xcrush_copy_bytes(BYTE* dst, const BYTE* src, size_t num)
 
 	if (src + num < dst || src > dst + num)
 		memcpy(dst, src, num);
-	else
+	else if (src != dst)
 	{
 		// src and dst overlaps
 		// we should copy the area that doesn't overlap repeatly
 		const size_t diff = (dst > src) ? dst - src : src - dst;
-		const size_t rest = (diff > 0) ? num % diff : 0;
+		const size_t rest = num % diff;
 		const size_t end = num - rest;
 
 		for (size_t a = 0; a < end; a += diff)


mfleisz (approved):

mihainradu:
I remember I left a similar message, but I cannot find it so I write it again, and I apologies if it looks like a spam.
@akallabeth 
There is a situation like:
num = 7
at src:[1,2,3,4,5,6,7]
dst = src + 1
after the call to this method: at dst:[1,1,1,1,1,1,1]
this does not sound right to me, 
so if it does not sound right to you maybe the actual test is faulted

https://cpp.sh/?source=%2F*+memmove+example+*%2F%0D%0A%23include+%3Cstdio.h%3E%0D%0A%23include+%3Cstring.h%3E%0D%0A%0D%0A%0D%0Astatic+size_t+xcrush_copy_bytes(char*+dst%2C+const+char*+src%2C+size_t+num)%0D%0A%7B%0D%0A%0D%0A%09if+(src+%2B+num+%3C+dst+%7C%7C+src+%3E+dst+%2B+num)%0D%0A%09%09memcpy(dst%2C+src%2C+num)%3B%0D%0A%09else+if+(src+!%3D+dst)%0D%0A%09%7B%0D%0A%09%09%2F%2F+src+and+dst+overlaps%0D%0A%09%09%2F%2F+we+should+copy+the+area+that+doesn%27t+overlap+repeatly%0D%0A%09%09const+size_t+diff+%3D+(dst+%3E+src)+%3F+dst+-+src+%3A+src+-+dst%3B%0D%0A%09%09const+size_t+rest+%3D+num+%25+diff%3B%0D%0A%09%09const+size_t+end+%3D+num+-+rest%3B%0D%0A%0D%0A%09%09for+(size_t+a+%3D+0%3B+a+%3C+end%3B+a+%2B%3D+diff)%0D%0A%09%09%09memcpy(%26dst%5Ba%5D%2C+%26src%5Ba%5D%2C+diff)%3B%0D%0A%09%09if+(rest+!%3D+0)%0D%0A%09%09%09memcpy(%26dst%5Bend%5D%2C+%26src%5Bend%5D%2C+rest)%3B%0D%0A%09%7D%0D%0A%09return+num%3B%0D%0A%7D%0D%0A%0D%0Aint+main+()%0D%0A%7B%0D%0A++char+str%5B%5D+%3D+%221234567890%22%3B%0D%0A++memmove+(str%2B1%2Cstr%2C9)%3B%0D%0A++puts+(str)%3B%0D%0A%0D%0A++char+str2%5B%5D+%3D+%221234567890%22%3B%0D%0A%0D%0A++xcrush_copy_bytes(str2%2B1%2C+str2%2C+9)%3B%0D%0A++puts+(str2)%3B%0D%0A++%0D%0A++return+0%3B%0D%0A%7D%0D%0A

akallabeth:
@mihainradu is is indeed a strange part of the code, but this function operates on the history buffer of the bulk decompressor. so using `memmove` breaks the history buffer and the decompression produces invalid data.
the original code did use a simple loop to just copy byte by byte from `dst[x] = src[x]` which the current code also does.

akallabeth:
so, you can check yourself, if you go through the git history and use the original function, which was here db328a1e93

mihainradu:
@akallabeth thanks for the reply
Important thing is that it is the intention of the code to behave like that.
I would also update the code so that it would also indicate better the intention to the reader
(instead of &src[a] use just src)
eg: 
```
memcpy(&dst[a], src, diff);
...
memcpy(&dst[a], src, diff);
```

PR #8534, opened 2022-12-13
tls: cleanup and add some methods to do handshakes asynchronously

hardening:
This patch does a few cleanups to allow creating TLS and DTLS contexts. It also introduces tls_accept_ex and tls_connect_ex that can start the SSL handshake, and it can be finished by calling tls_handshake


mfleisz (approved):
LGTM

PR #8537, opened 2022-12-14
[core,info] enforce extended info limits

akallabeth:
the cbClientAddress field limits defined in [MS-RDPBCGR] 2.2.1.11.1.1.1 Extended Info Packet (TS_EXTENDED_INFO_PACKET) are now properly enforced.


pnowack:
Since github does not allow adding comments lines, that are too much outside of the actual diff (but are still relevant), adding here:

`Stream_Write_UINT16(s, cbClientAddress + 2); /* cbClientAddress (2 bytes) */` should become `Stream_Write_UINT16(s, cbClientAddress); /* cbClientAddress (2 bytes) */`, few lines below, since the null terminator is now included there.

pnowack:
Change looks good to me. Seems though the CI is not happy yet.

pnowack (approved):

PR #8539, opened 2022-12-14
[core,info] ensure stream length

akallabeth:
ensure the stream is large enough for all info packets


pnowack (approved):
Couldn't spot anything wrong here.

pnowack (approved):

PR #8545, opened 2022-12-15
core: Fix handling of alternate shell if set to an empty string

mfleisz:
In case alternate shell was set to an empty string ("") the old code would try to convert it to a WChar string. This resulted in a NULL pointer being returned by `ConvertUtf8ToWCharAlloc` that was interpreted as an error and the connection failed.


mfleisz:
@freerdp-bot test

hardening (approved):
LGTM

akallabeth (approved):
nice catch

PR #8547, opened 2022-12-15
new [tls/server]: disable client side renegotiation

bmiklautz:
Client side renegotiation could lead to CPU intensive recalculations. Regeneration is disabled per default in openssl 3+ and not available anymore TLS1.3. Therefore disable it for older TLS versions if possible.


MartinHaimberger (approved):
LGTM

hardening:
Is it something we want only for server-side ?

hardening (approved):

PR #8555, opened 2022-12-16
[winpr,image] add winpr_bitmap_write_ex

akallabeth:
Added a helper function writing a bitmap from an image that has a stride that is not width * <bytes per pixel>

### DIFF
--- a/winpr/libwinpr/utils/image.c
+++ b/winpr/libwinpr/utils/image.c
@@ -152,9 +152,18 @@ fail:
 int winpr_bitmap_write(const char* filename, const BYTE* data, size_t width, size_t height,
                        size_t bpp)
 {
+	return winpr_bitmap_write_ex(filename, data, 0, width, height, bpp);
+}
+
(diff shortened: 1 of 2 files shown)


winpr/libwinpr/utils/image.c
```diff
@@ -169,10 +178,16 @@ int winpr_bitmap_write(const char* filename, const BYTE* data, size_t width, siz
...
 
-	if (fwrite(bmp_header, WINPR_IMAGE_BMP_HEADER_LEN, 1, fp) != 1 ||
-	    fwrite((const void*)data, img_size, 1, fp) != 1)
+	if (fwrite(bmp_header, WINPR_IMAGE_BMP_HEADER_LEN, 1, fp) != 1)
 		goto fail;
 
+	for (size_t y = 0; y < height; y++)
+	{
+		const void* line = &data[stride * y];
+		if (fwrite(line, bpp_stride, 1, fp) != 1)
```
mfleisz, line 187: Should be `width` as we are not saving stride anywhere in the bmp and we will load garbage when reading the file

mfleisz (approved):

PR #8564, opened 2022-12-18
Shadow resize

akallabeth:
fixed 2 issues:
1. `rdp_server_reactivate` must not block
2. the shadow server should discard a peer that rejects the resize request


hardening (approved):
LGTM

PR #8571, opened 2022-12-22
revert changes so that the osmajortype/osminortype is not overwritten

DVeron-RC:
2.8.0 introduced a regression overwriting the client capabilies.
The code cherry picked was intended in 2.8.0 was intended to work with a new setting struct.
The changes that made this new struct were not backported to 2.0-stable.


mfleisz:
@freerdp-bot test

PR #8572, opened 2022-12-22
Proxy fixes

akallabeth:


server/proxy/pf_client.c
```diff
@@ -466,23 +466,22 @@ static BOOL pf_client_receive_channel_data_hook(freerdp* instance, UINT16 channe
 	switch (channel->onBackData(pdata, channel, xdata, xsize, flags, totalSize))
 	{
 		case PF_CHANNEL_RESULT_PASS:
-			break;
+			server_channel_id = WTSChannelGetId(ps->context.peer, channel->channel_name);
```
hardening, line 469: I know it's refactoring (not your code) but we'd better use `channel->front_channel_id` which is the cached value...

akallabeth: @hardening done. you´ve given me quite the headache with the `return PF_CHANNEL_RESULT_PASS;` above ;)

hardening (approved):

PR #8576, opened 2022-12-26
Wtsapi changes

hardening:
A few changes with WTSAPI

### DIFF
--- a/libfreerdp/core/server.c
+++ b/libfreerdp/core/server.c
@@ -1693,9 +1693,9 @@ BOOL WINAPI FreeRDP_WTSVirtualChannelQuery(HANDLE hChannelHandle, WTS_VIRTUAL_CL
 						break;
 
 					default:
-						bval = FALSE;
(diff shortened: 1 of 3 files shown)


libfreerdp/core/server.c
```diff
@@ -1693,9 +1697,9 @@ BOOL WINAPI FreeRDP_WTSVirtualChannelQuery(HANDLE hChannelHandle, WTS_VIRTUAL_CL
 						break;
 
 					default:
-						bval = FALSE;
-						status = FALSE;
-						break;
+						ppBuffer = NULL;
+						pBytesReturned = 0;
```
mfleisz, line 1701: Shouldn't this be `*ppBuffer = NULL;` and `*pBytesReturned = 0;`?

hardening: oh good catch !

mfleisz (approved):

PR #8591, opened 2023-01-03
cliprdr: Fix sending of server-side PDUs

mfleisz:
The old implementation used `Stream_GetLength` to obtain the amount of data to send. However some PDUs (i.e. CLIPRDR_FILECONTENTS_REQUEST) are overallocated as they contain optional data (clipDataId in this case).

We now use the position to determine the actual data written to the PDU instead of the amount of bytes allocated, avoid sending garbage data to the client.


akallabeth (approved):
LGTM

sidenote: `Stream_Length` requires `Stream_SealLength` to be called before that the length is fixed to the capacity of the stream (the bug you have found here). your fix is, of course, perfectly fine too.

PR #8595, opened 2023-01-09
Alpha fix

akallabeth:
fixes alpha channel handling for rails mode with progressive and clear codec:
These codecs do not have alpha, so it is required to keep the destination alpha channel that has been set with the alpha codec


hardening (approved):
LGTM

mfleisz (approved):

akallabeth:
@freerdp-bot test

PR #8607, opened 2023-01-13
hide password for mac client

amazingfate:
Hide password like xfreerdp on mac so that people can't see password from command like `ps`.


akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #8611, opened 2023-01-13
[core] add missing redirection fields

akallabeth:
* Add read/write of redirection guid, target certificate


pnowack:
Regarding the `psPeerSendServerRedirection` callback, would be good to have @jotolo2 review on that part.

include/freerdp/peer.h
```diff
@@ -54,13 +54,12 @@ typedef BOOL (*psPeerPostConnect)(freerdp_peer* peer);
...
-typedef BOOL (*psPeerSendServerRedirection)(freerdp_peer* peer, UINT32 sessionId,
-                                            const char* targetNetAddress, const char* routingToken,
-                                            const char* userName, const char* domain,
-                                            const char* password, const char* targetFQDN,
-                                            const char* targetNetBiosName, DWORD tsvUrlLength,
-                                            const BYTE* tsvUrl, UINT32 targetNetAddressesCount,
-                                            const char** targetNetAddresses);
+typedef BOOL (*psPeerSendServerRedirection)(
+    freerdp_peer* peer, UINT32 sessionId, const char* targetNetAddress, const char* routingToken,
+    const char* userName, const char* domain, const char* password, const char* targetFQDN,
```
pnowack, line 59: This should allow setting the `LB_PASSWORD_IS_PK_ENCRYPTED` flag.

libfreerdp/core/redirection.c
```diff
@@ -339,32 +368,17 @@ static state_run_t rdp_recv_server_redirection_pdu(rdpRdp* rdp, wStream* s)
...
-
 		/* [MS-RDPBCGR] specifies 512 bytes as the upper limit for the password length
 		 * including the null terminatior(s). This should also be enough for the unknown
 		 * password cookie format (see previous comment).
 		 */
 
-		if (!Stream_CheckAndLogRequiredLength(TAG, s, redirection->PasswordLength))
-			return STATE_RUN_FAILED;
-
 		if (redirection->PasswordLength > LB_PASSWORD_MAX_LENGTH)
```
pnowack, line 380: This does not apply (anymore), when `LB_PASSWORD_IS_PK_ENCRYPTED` is set.

pnowack:
Regarding the writing part, Joan already did some changes: https://github.com/jotolo2/FreeRDP/commit/8f6c066734ffa0d51cb8473bf91815ddc32f9709.  
I don't know, whether there are any unpushed (local) changes though, but they should already work (see the FreeRDP discussion about mstsc refusing to connect, when a wrong targetCertificate is specified).

libfreerdp/core/peer.c
```diff
@@ -1225,146 +1225,78 @@ static BOOL freerdp_peer_send_channel_data(freerdp_peer* client, UINT16 channelI
 	return rdp_send_channel_data(client->context->rdp, channelId, data, size);
 }
 
+static BOOL redir_write_string(UINT32 flag, wStream* s, const char* str)
+{
+	const size_t length = (strlen(str) + 1);
+	if (!Stream_EnsureRemainingCapacity(s, 4ull + length * sizeof(WCHAR)))
+		return FALSE;
+
+	Stream_Write_UINT32(s, length);
```
jotolo2, line 1234: This field is the length in bytes so if we are writing in UTF16 that should be `length * sizeof(WCHAR)` isn't it?

libfreerdp/core/peer.c
```diff
@@ -1391,81 +1325,87 @@ static BOOL freerdp_peer_send_server_redirection_pdu(
...
+			if (!redir_write_string(LB_TARGET_NET_ADDRESSES, s, targetNetAddresses[i]))
+				goto fail;
 		}
-		free(targetNetAddressesW);
-		free(targetNetAddressesWLength);
+
+		/* Write length field */
+		const size_t end = Stream_GetPosition(s);
+		Stream_SetPosition(s, start);
+		Stream_Write_UINT32(s, end - start - 8);
```
jotolo2, line 1393: When writing the length of `TartetNetAddresses`, it just shouldn't count `TargetNetAddressesLentght` field which is 4 bytes, not 8, so assuming we are counting bytes I think that should be: `end - start - 4`

include/freerdp/settings.h
```diff
@@ -1214,7 +1221,11 @@ struct rdp_settings
 	ALIGN64 char* RedirectionAcceptedCert;        /* 1231 */
 	ALIGN64 UINT32 RedirectionAcceptedCertLength; /* 1232 */
 	ALIGN64 UINT32 RedirectionPreferType;         /* 1233 */
-	UINT64 padding1280[1280 - 1234];              /* 1234 */
+	ALIGN64 BYTE* RedirectionGuid;                /* 1234 */
```
jotolo2, line 1224: Could these fields be just after `RedirectionTsvUrl` to follow the same order as the documentation?
Down there, the `TargetNetAddressesCount` and `TargetNetAddresses` might make sense to add the `Redirection` prefix.
And I think `RedirectionAcceptedCert`, `RedirectionAcceptedCertLength` and  `RedirectionPreferType` are orphans fields that don't have relation with the Server Redirection PDU.

jotolo2:
@pnowack @akallabeth Yes, the changes from https://github.com/jotolo2/FreeRDP/commit/8f6c066734ffa0d51cb8473bf91815ddc32f9709 already work. The main difference from these changes is the encoding of `redirectionGUID` and `targetCertificate`.

I encode them as defined in the documentation to base64 and then Unicode (also receive `targetCertificate` as a string in PEM format and previously encode it in DER).

Also, I added an encryptedPassword field that adds LB_PASSWORD_IS_PK_ENCRYPTED and sends that as a string without encoding it to Unicode. I'm not sure how should be the right format.

pnowack:
> I'm not sure how should be the right format.

I think that is correct due to: `If the LB_PASSWORD_IS_PK_ENCRYPTED (0x00004000) flag is specified in the RedirFlags field, then the password MUST be treated as an opaque encrypted blob` (i.e. raw data, that should not be touched (changed)).
The unicode part is only for `LB_PASSWORD_IS_PK_ENCRYPTED` specified.

include/freerdp/peer.h
akallabeth: working on it, will push that later. (will make the `redirection_new` stuff public and use that as argument)

akallabeth:
@pnowack @jotolo2 ok, updated and tested (client side) the pr. so far so good, any remarks?

libfreerdp/core/redirection.c
```diff
@@ -481,27 +593,354 @@ void redirection_free(rdpRedirection* redirection)
...
+			    redir_write_string(LB_TARGET_NET_ADDRESSES, s, redirection->TargetNetAddresses[i]);
+			if (rcc < 0)
+				goto fail;
+			length += (UINT32)rcc;
+		}
+
+		/* Write length field */
+		const size_t end = Stream_GetPosition(s);
+		Stream_SetPosition(s, start);
+		Stream_Write_UINT32(s, length);
```
jotolo2, line 751: Here the length field should count also TargetNetAddressesCount which is 4 bytes so: `length + 4`

libfreerdp/core/redirection.c
```diff
@@ -45,58 +47,119 @@ struct rdp_redirection
 	char* TargetNetAddress;
 	UINT32 TargetNetAddressesCount;
 	char** TargetNetAddresses;
+	UINT32 RedirectionGuidLength;
```
jotolo2, line 50: Should all length fields be defined either with `UINT32` or with `DWORD`?

jotolo2:
LGTM appart from my two remarks :+1:

libfreerdp/core/redirection.c
```diff
@@ -481,27 +593,354 @@ void redirection_free(rdpRedirection* redirection)
...
+
+static SSIZE_T redir_write_string(UINT32 flag, wStream* s, const char* str)
+{
+	const size_t length = (strlen(str) + 1);
+	if (!Stream_EnsureRemainingCapacity(s, 4ull + length * sizeof(WCHAR)))
+		return -1;
+
+	const size_t pos = Stream_GetPosition(s);
+	Stream_Write_UINT32(s, (UINT32)length * sizeof(WCHAR));
+	Stream_Write_UTF16_String_From_UTF8(s, length, str, length, TRUE);
```
pnowack, line 622: The return value is not checked.

libfreerdp/core/redirection.c
```diff
@@ -481,27 +593,354 @@ void redirection_free(rdpRedirection* redirection)
...
+		/* Write length field */
+		const size_t end = Stream_GetPosition(s);
+		Stream_SetPosition(s, start);
+		Stream_Write_UINT32(s, length);
+		Stream_SetPosition(s, end);
+	}
+
+	if (!Stream_EnsureRemainingCapacity(s, 1))
+		goto fail;
+	Stream_Write_UINT8(s, 0);
```
pnowack, line 757: I assume you mean the `Pad` here? Should these be 8Bytes instead of 8Bits?

```
Pad (8 bytes): An optional 8-element array of 8-bit unsigned integers. Padding. Values in this field
MUST be ignored.
```

libfreerdp/core/redirection.c
pnowack: It's pretty much a style thing, but usually we use `UINT32` for this.

pnowack (approved):

libfreerdp/core/redirection.c
akallabeth: updated

libfreerdp/core/redirection.c
akallabeth: updated

libfreerdp/core/redirection.c
akallabeth: the types in `winpr` are mimicking `windows` types, so there is some ambiguity.
preferring the expressive types over the more obscure ones.

libfreerdp/core/redirection.c
akallabeth: done

PR #8618, opened 2023-01-16
[core,gcc] fix earlyCapablityFlag handling

akallabeth:
Only reply with flags that are currently activated on server


akallabeth:
Replaced by #8629

PR #8628, opened 2023-01-19
Cert cleanups

akallabeth:


libfreerdp/core/nego.c
```diff
@@ -1303,31 +1321,49 @@ BOOL nego_send_negotiation_response(rdpNego* nego)
...
 		}
-		else if (settings->SelectedProtocol == PROTOCOL_HYBRID)
+		else if (nego->SelectedProtocol == PROTOCOL_HYBRID)
 		{
-			settings->TlsSecurity = TRUE;
-			settings->NlaSecurity = TRUE;
-			settings->RdpSecurity = FALSE;
-			settings->UseRdpSecurityLayer = FALSE;
-			settings->EncryptionLevel = ENCRYPTION_LEVEL_NONE;
+			if (!freerdp_settings_set_bool(settings, FreeRDP_TlsSecurity, TRUE))
```
hardening, line 1355: In that kind of case the usage of `freerdp_settings_set_XXX` makes it so heavy and useless (lots of unnecessary checks) compared to the original code...

PR #8629, opened 2023-01-19
Skip channel join v2

akallabeth:
* Fix `EarlyCapabilityFlags` usage (extract flags and generate on the fly, client and server have different ones in use)
* Add new flags


akallabeth:
@freerdp-bot test

libfreerdp/core/gcc.c
```diff
@@ -829,49 +853,139 @@ BOOL gcc_write_user_data_header(wStream* s, UINT16 type, UINT16 length)
...
+	const UINT32 filtered = flags & mask;
+	const UINT32 unknown = flags & ~mask;
+	if (unknown != 0)
+	{
+		char buffer[256] = { 0 };
+		WLog_WARN(TAG,
+		          "(TS_UD_CS_CORE)::EarlyCapabilityFlags [0x%08" PRIx32 " & 0x%08" PRIx32
+		          " --> 0x%08" PRIx32 "] filtering %s, feature not implemented",
+		          flags, ~mask, unknown,
+		          rdp_early_client_caps_string(unknown, buffer, sizeof(buffer)));
```
mfleisz, line 910: Don't think it makes sense to call `rdp_early_client_caps_string` since we already know it is an unknown flag and we already log all the flags in this message

libfreerdp/core/gcc.c
```diff
@@ -829,49 +853,139 @@ BOOL gcc_write_user_data_header(wStream* s, UINT16 type, UINT16 length)
...
 static BOOL updateEarlyClientCaps(rdpSettings* settings, UINT32 earlyCapabilityFlags,
                                   UINT32 connectionType)
 {
 	WINPR_ASSERT(settings);
 
+	if (settings->SupportErrorInfoPdu)
+		settings->SupportErrorInfoPdu =
+		    (earlyCapabilityFlags & RNS_UD_CS_SUPPORT_ERRINFO_PDU) ? TRUE : FALSE;
+
+#if 0
```
mfleisz, line 965: If not needed the code should be removed (or even better replaced with a comment why we do not set this flag here)

libfreerdp/core/gcc.c
```diff
@@ -829,49 +853,139 @@ BOOL gcc_write_user_data_header(wStream* s, UINT16 type, UINT16 length)
...
+		EarlyCapabilityFlags |= RNS_UD_SC_DYNAMIC_DST_SUPPORTED;
+	if (settings->SupportEdgeActionV2)
+		EarlyCapabilityFlags |= RNS_UD_SC_EDGE_ACTIONS_SUPPORTED_V2;
+	if (settings->SupportSkipChannelJoin)
+		EarlyCapabilityFlags |= RNS_UD_SC_SKIP_CHANNELJOIN_SUPPORTED;
+
+	return filterAndLogEarlyServerCapabilityFlags(EarlyCapabilityFlags);
+}
+
+static UINT16 filterAndLogEarlyClientCapabilityFlags(UINT32 flags)
```
mfleisz, line 892: Should return `UINT32`

libfreerdp/core/gcc.c
```diff
@@ -885,18 +999,11 @@ static BOOL updateEarlyClientCaps(rdpSettings* settings, UINT32 earlyCapabilityF
...
-	if (settings->SupportErrorInfoPdu)
-		settings->SupportErrorInfoPdu =
-		    earlyCapabilityFlags & RNS_UD_CS_SUPPORT_ERRINFO_PDU ? TRUE : FALSE;
-
-	if (!(earlyCapabilityFlags & RNS_UD_CS_VALID_CONNECTION_TYPE))
+	if ((earlyCapabilityFlags & RNS_UD_CS_VALID_CONNECTION_TYPE) == 0)
 		connectionType = 0;
 	settings->ConnectionType = connectionType;
 
+	filterAndLogEarlyClientCapabilityFlags(earlyCapabilityFlags);
```
mfleisz, line 1006: This call seems redundant (the returned, filtered value is not used and logging of the flags is already done by the caller)

libfreerdp/core/gcc.c
akallabeth: no, the flags for the client are `UINT16`, but we use `UINT32` internally

libfreerdp/core/gcc.c
akallabeth: there is the `RNS_UD_CS_SUPPORT_SKIP_CHANNELJOIN` flag that is not implemented yet.

libfreerdp/core/gcc.c
akallabeth: yes, the returned value is intendendly unused, but the logging here is different. while the caller just does some debug level logging the filter warns about the flags, which is the intention of this line.

libfreerdp/core/gcc.c
akallabeth: ok, replaced with a comment why that flag is not handled in this function.

mfleisz (approved):
LGTM

PR #8630, opened 2023-01-20
Improved logging

akallabeth:


libfreerdp/common/settings.c
```diff
@@ -2048,7 +2048,11 @@ BOOL freerdp_settings_set_string_from_utf16N(rdpSettings* settings, size_t id, c
 
 	char* str = ConvertWCharNToUtf8Alloc(param, length, &len);
 	if (!str && (length != 0))
-		return FALSE;
+	{
+		const size_t wlen = _wcsnlen(param, length);
```
mfleisz, line 2052: Don't quite get this additional check ... is it for checking empty strings?
Shouldn't this be correctly handled by `ConvertWCharNToUtf8Alloc`?

libfreerdp/core/tpdu.c
```diff
@@ -246,14 +246,17 @@ BOOL tpdu_write_data(wStream* s)
...
-	BYTE li;
+	BYTE code = 0;
+	BYTE li = 0;
 
 	if (!tpdu_read_header(s, &code, &li, tpktlength))
 		return FALSE;
 
 	if (code != X224_TPDU_DATA)
+	{
+		WLog_ERR(TAG, "tpdu got code 0x%02" PRIx8 " expected X224_TPDU_DATA [0x%02x]", code);
```
mfleisz, line 257: Missing format parameter

libfreerdp/common/settings.c
akallabeth: this case handles the return value for the following problem:
if `length > 0` but the string contained in `param` is empty then `ConvertWCharNToUtf8Alloc` returns `NULL` with `len=0`

`ConvertWCharNToUtf8Alloc` does correctly convert the string, but we do not want to fail here as the conversion was successful and just clear the (maybe) existing value from settings.

libfreerdp/core/tpdu.c
akallabeth: fixed

client/SDL/sdl_freerdp.c
```diff
@@ -948,11 +948,13 @@ static int WINAPI sdl_client_thread_proc(LPVOID arg)
...
 	}
 
+	freerdp_disconnect(instance);
+
 disconnect:
 	if (freerdp_settings_get_bool(instance->context->settings, FreeRDP_AuthenticationOnly))
 		WLog_Print(sdl->log, WLOG_INFO, "Authentication only, exit status %s [%" PRId32 "]",
 		           sdl_map_to_code_tag(exit_code), exit_code);
-	freerdp_disconnect(instance);
+
```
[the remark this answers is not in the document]
akallabeth: fixed

libfreerdp/common/settings.c
mfleisz: That's what I mean - if we pass an empty string to `ConvertWCharNToUtf8Alloc` shouldn't it also return an empty (non-null) string? IIRC there are some cases with string settings that differentiate between `NULL` and `""` - like you said in this case we would clear the value (which should be done in case `NULL` is passed) and not set it to an empty string.

akallabeth: ok, removed for the time being, the change might result in another pr.

mfleisz (approved):

PR #8632, opened 2023-01-20
Fix handling of RAIL HandshakeEx flag

mfleisz:
When using Enhanced RAIL the HandshakeEx flag must also be set. However in the current code it was always overwritten by the server flags (which might lack the flag).


akallabeth (approved):
LGTM

PR #8635, opened 2023-01-21
Redir fix

akallabeth:
* Revert to `REDIRECTION_VERSION5` as `REDIRECTION_VERSION6`does not work
* Add some debug logging
* Fix wrong flag `REDIRECTED_SMARTCARD` (only set if we are doing smartcard logon)


libfreerdp/core/settings.c
```diff
@@ -374,7 +374,7 @@ rdpSettings* freerdp_settings_new(DWORD flags)
 	                               (flags & FREERDP_SETTINGS_SERVER_MODE) ? TRUE : FALSE) ||
 	    !freerdp_settings_set_bool(settings, FreeRDP_WaitForOutputBufferFlush, TRUE) ||
 	    !freerdp_settings_set_uint32(settings, FreeRDP_ClusterInfoFlags,
-	                                 REDIRECTION_SUPPORTED | (REDIRECTION_VERSION6 << 2)) ||
+	                                 REDIRECTION_SUPPORTED | (REDIRECTION_VERSION5 << 2)) ||
```
mfleisz, line 377: Just checking if this is really wanted (going back from 6 to 5)?

mfleisz (approved):
Just checking on the version thing - otherwise LGTM

PR #8639, opened 2023-01-24
Log stream capacity

akallabeth:
* Split length argument of `Stream_CheckAndLogRemainingLength*` functions to have better logging
* Similar to `Stream_CheckAndLogRemainingLength` add functions to check and log remaining capacity.


hardening (approved):
LGTM

PR #8641, opened 2023-01-25
[client,sdl] Add WSAStartup

akallabeth:
Windows requires WSAStartup to be called for the client to work.


hardening:
`WSAStartup` and `WSACleanup` also exists in winpr, so no need for the `#ifdef _WIN32`.

PR #8645, opened 2023-01-26
Log and leak

akallabeth:


libfreerdp/core/credssp_auth.c
```diff
@@ -104,7 +104,7 @@ rdpCredsspAuth* credssp_auth_new(const rdpContext* rdp_ctx)
 
 BOOL credssp_auth_init(rdpCredsspAuth* auth, TCHAR* pkg_name, SecPkgContext_Bindings* bindings)
 {
-	SECURITY_STATUS status;
+	SECURITY_STATUS status = ERROR_INTERNAL_ERROR;
```
hardening, line 107: This is not needed, `status` is unconditionality set

hardening:
IMHO the changes in libfreerdp/core/credssp_auth.c just had noise in the code by initializing status with values even when it is unconditionally set after. One way to workaround this could be to move the variable declaration at the first affectation.

hardening:
And the same applies to nla.c, over-initializing variables just had noise. Especially when you declare the variable initialized to NULL at the beginning of the function, and then affect it 3 lines later. As we've changed the rules for variables declared in the middle of a block, let's take advantage of that.

libfreerdp/core/credssp_auth.c
akallabeth: will change to declare where used where possible.

mfleisz (approved):
LGTM

PR #8654, opened 2023-01-30
[core,settings] update getter/setter generation

akallabeth:
fix issues with const and non const string pointer update

### DIFF
--- a/libfreerdp/common/settings.c
+++ b/libfreerdp/common/settings.c
@@ -1570,7 +1570,7 @@ void* freerdp_settings_get_pointer_array_writable(const rdpSettings* settings, s
 			max = freerdp_settings_get_uint32(settings, FreeRDP_TargetNetAddressCount);
 			if (offset >= max)
 				goto fail;
-			return settings->TargetNetPorts[offset];
+			return (void*)(uintptr_t)settings->TargetNetPorts[offset];
 		case FreeRDP_ClientTimeZone:
 			max = 1;
 			if (offset >= max)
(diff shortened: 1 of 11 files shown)


libfreerdp/common/settings.c
```diff
@@ -1570,7 +1570,7 @@ void* freerdp_settings_get_pointer_array_writable(const rdpSettings* settings, s
 			max = freerdp_settings_get_uint32(settings, FreeRDP_TargetNetAddressCount);
 			if (offset >= max)
 				goto fail;
-			return settings->TargetNetPorts[offset];
+			return (void*)(uintptr_t)settings->TargetNetPorts[offset];
```
mfleisz, line 1573: Is this really what we want?
`TargetNetPorts` is a `UINT32*` so we would return the Port value as an address the way the code is currently

akallabeth: @mfleisz no, `TargetNetPorts[x]` is a `UINT32`. we return the value in a `void*` (see function return). To avoid any issues with integer/pointer sizes, it is first cast to `uintptr_t` so that no invalid data is there if e.g. `uintptr_t` or `size_t ` integer variables are used on the other end.

mfleisz: @akallabeth from `include/freerdp/settings.h`:

` ALIGN64 UINT32* TargetNetPorts;`

mfleisz: @akallabeth right - but why do we return the port value in a `void*`? The function `freerdp_settings_get_pointer_array_writable` suggests that I get a pointer that I am able to write to. Seems dangerous if i.e. we return `(void*)3389` from that function ...

akallabeth: @mfleisz but good point, this function is not used with `FreeRDP_TargetNetPorts` currently, so change the return to a `UINT32*` value

mfleisz (approved):

PR #8659, opened 2023-02-01
refactor gateway command line

akallabeth:


client/common/cmdline.h
```diff
@@ -166,11 +166,20 @@ static const COMMAND_LINE_ARGUMENT_A global_cmd_args[] = {
 	{ "from-stdin", COMMAND_LINE_VALUE_OPTIONAL, "force", NULL, NULL, -1, NULL,
 	  "Read credentials from stdin. With <force> the prompt is done before connection, otherwise "
 	  "on server request." },
+	{ "gateway", COMMAND_LINE_VALUE_REQUIRED,
+	  "g:<gateway>[:<port>],u:<user>,d:<domain>,p:<password>,auto-consent:[on|off],usage-method:["
+	  "direct|detect],access-token:<"
+	  "token>,type:[rpc|http[,no-websockets][,extauth-sspi-ntlm]|auto[,no-websockets][,extauth-"
+	  "sspi-ntlm]],",
```
hardening, line 173: `extauth-sspi-ntlm` option is duplicated, isn't it ?

hardening: hum in fact the help message seems to have something wrong (if it's right it is so complicated to interpret that it makes it useless)

hardening (approved):
Except my remark seems ok

client/common/cmdline.c
```diff
@@ -2349,11 +2349,292 @@ static int parse_cache_options(rdpSettings* settings, const COMMAND_LINE_ARGUMEN
...
+			goto fail;
+	}
+
+	rc = TRUE;
+fail:
+	free(args);
+	return rc;
+}
+
+static void fill_credentail_string(COMMAND_LINE_ARGUMENT_A* args, const char* value)
```
mfleisz, line 2580: Typo "credentail"

akallabeth: oops, fixed

client/common/cmdline.h
akallabeth: @hardening once for `http` and once for `auto` type

PR #8642, opened 2023-01-25
Cert cleanup

akallabeth:

PR #8662, opened 2023-02-01
Cert unify

akallabeth:
based on #8659 and #8642 

* Refactor cert store:
   *  store certificates as files instead of using `known_hosts2`
   * use `rdpCertificate` internally
* Merge `CrytoCert` into `rdpCertificate`
* Use `rdpCertificate` and `rdpRsaKey` for smartcard emulation
* Remove `crypto_cert_*` form public API


mfleisz (approved):
LGTM (also tested with NLA smartcard auth)

PR #8675, opened 2023-02-07
[core,fastpath] fix too verbose log

akallabeth:
for fastpath_recv_update_synchronize only skip the available bytes as older servers tend to send short packets. This avoids (too) verbose logging.


mfleisz (approved):
LGTM but might be good to add a comment or this could get "fixed" again

akallabeth:
@mfleisz expand the code above, there is already one.

PR #8682, opened 2023-02-09
[autodetect] add pointer for custom user data

akallabeth:


include/freerdp/autodetect.h
```diff
@@ -28,6 +28,16 @@
...
 {
 #endif
+
+	typedef enum
+	{
+		AUTODETECT_STATE_INITIAL,
+		AUTODETECT_STATE_REQUEST,
+		AUTODETECT_STATE_RESPONSE,
+		AUTODETECT_STATE_COMPLETE,
+		AUTODETECT_STATE_FAIL
```
pnowack, line 38: Nit: Could also end with a `,` here. Then, in case of state additions, we won't have change unrelated existing lines.

pnowack (approved):

include/freerdp/autodetect.h
akallabeth: that would produce quite a few warnings during build as gcc/clang complain about that when set to pedantic/everything ;)

mfleisz (approved):

PR #8683, opened 2023-02-09
proxy: Fix NLA to TLS fallback connection

mfleisz:
Currently the proxy's TLS fallback if an NLA connection attempt failed
is broken. There are two issues with the current code that this PR
fixes:

- freerdp_reconnect is used which requires an already established
  connection to work correctly. This is not the case since the NLA
  connectin attempt failed. This resulted in a seemingly working TLS
  connection but i.e. channels where missing/not working.
- The fallback connection attempt just altered the NLA security setting
  in the instance's settings. However these settings have been already
  modified by the NLA connection attempt so we need to create a copy of
  the original connection settings before doing the first connect.

The PR also introduces freerdp_reset_context which restores the initial
connection settings for the given instance.


akallabeth (approved):
LGTM
nice catches!

PR #8687, opened 2023-02-13
core: Add possibility to distinguish between auth cancelled and no creds

mfleisz:
Currently if the authentication callback returns `FALSE` the utils function handle this as scenario as no credentials provided (returns `AUTH_NO_CREDENTIALS)`.

This PR introduces a new `auth_status` called `AUTH_CANCELLED` that is returned if the authentication callback returns `FALSE`. If the callback returns `TRUE` and username or password are empty the util function will continue to return `AUTH_NO_CREDENTIALS`.

The PR also fixes some incorrect returns in RPC over HTTP gateway code.

### DIFF
--- a/libfreerdp/core/gateway/ncacn_http.c
+++ b/libfreerdp/core/gateway/ncacn_http.c
@@ -161,21 +161,24 @@ BOOL rpc_ncacn_http_auth_init(rdpContext* context, RpcChannel* channel)
 		case AUTH_SUCCESS:
 		case AUTH_SKIP:
 			break;
+		case AUTH_CANCELLED:
+			freerdp_set_last_error_log(instance->context, FREERDP_ERROR_CONNECT_CANCELLED);
+			return FALSE;
 		case AUTH_NO_CREDENTIALS:
 			freerdp_set_last_error_log(instance->context,
 			                           FREERDP_ERROR_CONNECT_NO_OR_MISSING_CREDENTIALS);
-			return TRUE;
+			return FALSE;
 		case AUTH_FAILED:
 		default:
 			return FALSE;
 	}
 
 	if (!credssp_auth_init(auth, AUTH_PKG, tls->Bindings))
-		return TRUE;
+		return FALSE;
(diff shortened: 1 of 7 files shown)


mfleisz:
@freerdp-bot test

libfreerdp/core/gateway/rpc_bind.c
```diff
@@ -136,6 +136,9 @@ static int rpc_bind_setup(rdpRpc* rpc)
 		case AUTH_SUCCESS:
 		case AUTH_SKIP:
 			break;
+		case AUTH_CANCELLED:
+			freerdp_set_last_error_log(instance->context, FREERDP_ERROR_CONNECT_CANCELLED);
+			return FALSE;
```
akallabeth, line 141: wrong return type, function returns `int`

libfreerdp/core/gateway/ncacn_http.c
```diff
@@ -161,21 +161,24 @@ BOOL rpc_ncacn_http_auth_init(rdpContext* context, RpcChannel* channel)
...
 	if (!credssp_auth_init(auth, AUTH_PKG, tls->Bindings))
-		return TRUE;
+		return FALSE;
 
 	if (sspi_SetAuthIdentityA(&identity, settings->GatewayUsername, settings->GatewayDomain,
 	                          settings->GatewayPassword) < 0)
-		return TRUE;
+		return FALSE;
 
 	credssp_auth_setup_client(auth, "HTTP", settings->GatewayHostname, &identity, NULL);
```
akallabeth, line 183: not your code, but missing return check

akallabeth (changes requested):
Overall LGTM except the wrong return value

akallabeth (approved):

libfreerdp/core/gateway/ncacn_http.c
mfleisz: @akallabeth thx for fixing 👍

PR #8692, opened 2023-02-14
core: Fix pointer corruption with d2i_X509

mfleisz:
The `d2i_X509` function manipulates the passed pointer on success. This resulted in a corrupted `rdpCertBlob` struct, crashing later on free.


akallabeth (approved):
LGTM
nice catch.

PR #8694, opened 2023-02-15
Client callback defaults and argument refactoring

akallabeth:
* Add `instance` argument to smartcard listing callback (required for any gui tool to have some context information)
* automatically initialize the client callbacks to the `cli` versions, clients can always override with custom implementations


akallabeth:

mfleisz (approved):

PR #8698, opened 2023-02-16
client/win: Fix invalid ShowWindow call from gfx thread

mfleisz:
ShowWindow must be called from the thread that owns the window. Currently it might be called from the gfx channel thread which causes random app hangs/window not showing up issues.


akallabeth (approved):
LGTM

PR #8699, opened 2023-02-16
[core] add log for experimental settings

akallabeth:
Too often experimental flags had been used without the user noticing that. As bug reports are hard to analyze without proper information take this approach and inform about experimental flags in use by logging these.

### DIFF
--- a/libfreerdp/core/freerdp.c
+++ b/libfreerdp/core/freerdp.c
@@ -21,6 +21,7 @@
 
 #include <freerdp/config.h>
 
+#include <string.h>
 #include <stdarg.h>
 
 #include "rdp.h"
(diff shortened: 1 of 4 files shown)


mfleisz (commented):
Overall LGTM

In `rdp_log_build_warnings` i'd favor using a `BOOL` and just display the output once since this cannot be turned off (currently it will print the warnings up to 3 times).

I wonder if the code is not better placed in `client/common` in order to reuse `starts_with` and `option_equals` code.

akallabeth:
@mfleisz might be able to move it to `core/utils` but not `client/common` as these things are logged for proxy/server as well

mfleisz (approved):

PR #8705, opened 2023-02-20
[client,common] working REDIRECTION_VERSION6

akallabeth:
* REDIRECTION_VERSION6 requires enabled multitransport, enable it
* Add a fallback if multitransport was disabled


akallabeth:
@jotolo2 ok, updated multitransport default and the redirection version

hardening (approved):
Except that I'm not sure enabling multi-transport by default is a good idea, that looks good to me.

PR #8707, opened 2023-02-20
core: Allow change to smart card logon in Authentication callbacks

mfleisz:
This PR adds a few changes so that a client is able to change the authentication/logon type in the Authentication callback. I.e. if the client was started without user/domain the authentication callback is now able to activate smart card logon by setting the SmartcardLogon setting along with csp/container/reader name.


akallabeth (approved):
LGTM

PR #8711, opened 2023-02-21
Clip fix once more

akallabeth:


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

libfreerdp/core/settings.c
```diff
@@ -460,8 +460,7 @@ rdpSettings* freerdp_settings_new(DWORD flags)
 		goto out_fail;
 
 	if (!freerdp_settings_set_uint32(settings, FreeRDP_MultitransportFlags,
-	                                 (TRANSPORT_TYPE_UDP_FECR | TRANSPORT_TYPE_UDP_FECL |
-	                                  TRANSPORT_TYPE_UDP_PREFERRED | SOFTSYNC_TCP_TO_UDP)))
+	                                 TRANSPORT_TYPE_UDP_FECR))
```
pnowack, line 463: Actually, this flag is for now not necessary too. Or will we run into issues, when we have no flags set here?

client/X11/xf_cliprdr.c
```diff
@@ -31,13 +31,15 @@
...
 #endif
 
-#ifdef WITH_FUSE
-#define FUSE_USE_VERSION FUSE_API_VERSION
-#if FUSE_USE_VERSION >= 30
-#include <fuse3/fuse_lowlevel.h>
-#else
-#include <fuse/fuse_lowlevel.h>
+#ifdef WITH_FUSE3
+#define FUSE_USE_VERSION 30
```
pnowack, line 35: We can directly go with version `35` here.

client/X11/xf_cliprdr.c
```diff
@@ -3072,17 +3072,15 @@ static DWORD WINAPI xf_cliprdr_fuse_thread(LPVOID arg)
...
+	struct fuse_chan* ch = fuse_mount(clipboard->delegate->basePath, &args);
+	if (ch != NULL)
 	{
 		clipboard->fuse_sess = fuse_lowlevel_new(&args, &xf_cliprdr_fuse_oper,
 		                                         sizeof(xf_cliprdr_fuse_oper), (void*)clipboard);
 		if (clipboard->fuse_sess != NULL)
 		{
 			fuse_session_add_chan(clipboard->fuse_sess, ch);
-			err = fuse_session_loop(clipboard->fuse_sess);
+			const int err = fuse_session_loop(clipboard->fuse_sess);
```
pnowack, line 3113: I honestly don't see a reason to make this constant here. I mean, we won't change the value here afterwards, sure, but this here is a bit unusual, we normally never do that for errors.  
In pure theory, `fuse_session_loop` could always fail, but with different error value on each run. Having this a constant here would be a bit misleading.

client/X11/xf_cliprdr.c
```diff
@@ -1097,6 +1104,44 @@ static BOOL xf_cliprdr_process_selection_notify(xfClipboard* clipboard,
...
+static xfCliprdrFuseInode* xf_cliprdr_fuse_create_root_node(void)
+{
+	xfCliprdrFuseInode* rootNode = (xfCliprdrFuseInode*)calloc(1, sizeof(xfCliprdrFuseInode));
+	if (!rootNode)
+		return NULL;
+
+	rootNode->ino = 1;
+	rootNode->parent_ino = 1;
+	rootNode->st_mode = S_IFDIR | 0755;
+	rootNode->name = _strdup("/");
```
pnowack, line 1117: Can't this return null on allocation failure?

pnowack (approved):
Just some relatively minor things, where I commented on.

libfreerdp/core/settings.c
akallabeth: @pnowack if no flags are set multitransport is automatically disabled. (we need to announce something, we can always tell the server it does not work later on)

client/X11/xf_cliprdr.c
akallabeth: no need for that, we just need the minimum version we support here.

client/X11/xf_cliprdr.c
akallabeth: just my preference (add `const` by default) to avoid overriding variables by accident. but yes, we currently do not check the return, fixing that.

client/X11/xf_cliprdr.c
akallabeth: yes, it could. and therefore it is checked a few lines below.

libfreerdp/core/settings.c
pnowack: > we need to announce something, we can always tell the server it does not work later on

alright.

client/X11/xf_cliprdr.c
pnowack: Ah, right. Didn't realize that, sry.

PR #8712, opened 2023-02-21
[core,gcc] Fix applying RedirectionVersionMask

jotolo2:
The RedirectionVersionMask is 0x3c i.e. 00111100.
So the left shift operation to set RedirectionVersion is of 2 places.


hardening (approved):
Good catch

PR #8713, opened 2023-02-22
Xclip refactor

akallabeth:


winpr/libwinpr/clipboard/clipboard.c
```diff
@@ -63,6 +63,13 @@ static const char* CF_STANDARD_STRINGS[CF_MAX] = {
 	"CF_DIBV5"         /* 17 */
 };
 
+const char* ClipboardGetFormatIdString(UINT32 formatId)
+{
+	if (formatId < ARRAYSIZE(CF_STANDARD_STRINGS))
+		return CF_STANDARD_STRINGS[formatId];
+	return "CF_REGISTERED_FORMAT";
```
pnowack, line 70: Wouldn't `CF_UNREGISTERED_FORMAT` be better (since the string to look for could not be found)?

akallabeth: the formats not predefined must be (manually) registered, so that is the reason for the name

pnowack: ah ok

pnowack (approved):
Could not spot something wrong here.

PR #8714, opened 2023-02-22
core: Add CAPI support for enumerating smart card key containers

mfleisz:
Windows seems to favor using the legacy Crypto API (CAPI) for enumerating RSA key containers and only relies on the newer CNG APIs for ECC keys.

This PR adds support for CAPI key container enumeration on Windows.

The PR also fixes an issue where the CSP was always set to the MS Base Smart Card Provider during NLA authentication.


akallabeth (approved):
LGTM

PR #8719, opened 2023-02-24
Implemented clipboard redirection for iOS.

iiordanov:
Implemented clipboard redirection for iOS inspired by the android implementation.

Also:
- Fix for missing iopenssl/err.h include in TSXAdditions.m.

### DIFF
--- a/client/iOS/FreeRDP/ios_cliprdr.m
+++ b/client/iOS/FreeRDP/ios_cliprdr.m
@@ -1,0 +1,503 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ * Android Clipboard Redirection
+ *
+ * Copyright 2013 Felix Long
+ * Copyright 2015 Thincast Technologies GmbH
+ * Copyright 2015 DI (FH) Martin Haimberger <martin.haimberger@thincast.com>
+ * Copyright 2023 Iordan Iordanov
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <freerdp/config.h>
+
+#include <winpr/crt.h>
+#include <winpr/stream.h>
+#include <winpr/clipboard.h>
+
+#include <freerdp/client/channels.h>
+#include <freerdp/client/cliprdr.h>
+
+#include "ios_cliprdr.h"
+
+UINT ios_cliprdr_send_client_format_list(CliprdrClientContext *cliprdr)
+{
+	UINT rc = ERROR_INTERNAL_ERROR;
+	UINT32 index;
+	UINT32 formatId;
+	UINT32 numFormats;
+	UINT32 *pFormatIds;
+	const char *formatName;
+	CLIPRDR_FORMAT *formats;
+	CLIPRDR_FORMAT_LIST formatList = { 0 };
+
+	if (!cliprdr)
+		return ERROR_INVALID_PARAMETER;
+
+	mfContext *afc = (mfContext *)cliprdr->custom;
+
+	if (!afc || !afc->cliprdr)
+		return ERROR_INVALID_PARAMETER;
+
+	pFormatIds = NULL;
+	numFormats = ClipboardGetFormatIds(afc->clipboard, &pFormatIds);
+	formats = (CLIPRDR_FORMAT *)calloc(numFormats, sizeof(CLIPRDR_FORMAT));
+
+	if (!formats)
+		goto fail;
+
+	for (index = 0; index < numFormats; index++)
+	{
+		formatId = pFormatIds[index];
+		formatName = ClipboardGetFormatName(afc->clipboard, formatId);
+		formats[index].formatId = formatId;
+		formats[index].formatName = NULL;
+
+		if ((formatId > CF_MAX) && formatName)
+		{
+			formats[index].formatName = _strdup(formatName);
+
+			if (!formats[index].formatName)
+				goto fail;
+		}
+	}
+
+	formatList.common.msgFlags = CB_RESPONSE_OK;
+	formatList.numFormats = numFormats;
+	formatList.formats = formats;
+	formatList.common.msgType = CB_FORMAT_LIST;
+
+	if (!afc->cliprdr->ClientFormatList)
+		goto fail;
+
+	rc = afc->cliprdr->ClientFormatList(afc->cliprdr, &formatList);
+fail:
+	free(pFormatIds);
+	free(formats);
+	return rc;
+}
+
+static UINT ios_cliprdr_send_client_format_data_request(CliprdrClientContext *cliprdr,
+                                                        UINT32 formatId)
+{
+	UINT rc = ERROR_INVALID_PARAMETER;
+	CLIPRDR_FORMAT_DATA_REQUEST formatDataRequest = { 0 };
+	mfContext *afc;
+
+	if (!cliprdr)
+		goto fail;
+
+	afc = (mfContext *)cliprdr->custom;
+
+	if (!afc || !afc->clipboardRequestEvent || !cliprdr->ClientFormatDataRequest)
+		goto fail;
+
+	formatDataRequest.common.msgType = CB_FORMAT_DATA_REQUEST;
+	formatDataRequest.common.msgFlags = 0;
+	formatDataRequest.requestedFormatId = formatId;
+	afc->requestedFormatId = formatId;
+	ResetEvent(afc->clipboardRequestEvent);
+	rc = cliprdr->ClientFormatDataRequest(cliprdr, &formatDataRequest);
+fail:
+	return rc;
+}
+
+static UINT ios_cliprdr_send_client_capabilities(CliprdrClientContext *cliprdr)
+{
+	CLIPRDR_CAPABILITIES capabilities;
+	CLIPRDR_GENERAL_CAPABILITY_SET generalCapabilitySet;
+
+	if (!cliprdr || !cliprdr->ClientCapabilities)
+		return ERROR_INVALID_PARAMETER;
+
+	capabilities.cCapabilitiesSets = 1;
+	capabilities.capabilitySets = (CLIPRDR_CAPABILITY_SET *)&(generalCapabilitySet);
+	generalCapabilitySet.capabilitySetType = CB_CAPSTYPE_GENERAL;
+	generalCapabilitySet.capabilitySetLength = 12;
+	generalCapabilitySet.version = CB_CAPS_VERSION_2;
+	generalCapabilitySet.generalFlags = CB_USE_LONG_FORMAT_NAMES;
+	return cliprdr->ClientCapabilities(cliprdr, &capabilities);
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_monitor_ready(CliprdrClientContext *cliprdr,
+                                      const CLIPRDR_MONITOR_READY *monitorReady)
+{
+	UINT rc;
+	mfContext *afc;
+
+	if (!cliprdr || !monitorReady)
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext *)cliprdr->custom;
+
+	if (!afc)
+		return ERROR_INVALID_PARAMETER;
+
+	if ((rc = ios_cliprdr_send_client_capabilities(cliprdr)) != CHANNEL_RC_OK)
+		return rc;
+
+	if ((rc = ios_cliprdr_send_client_format_list(cliprdr)) != CHANNEL_RC_OK)
+		return rc;
+
+	afc->clipboardSync = TRUE;
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_server_capabilities(CliprdrClientContext *cliprdr,
+                                            const CLIPRDR_CAPABILITIES *capabilities)
+{
+	UINT32 index;
+	CLIPRDR_CAPABILITY_SET *capabilitySet;
+	mfContext *afc;
+
+	if (!cliprdr || !capabilities)
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext *)cliprdr->custom;
+
+	if (!afc)
+		return ERROR_INVALID_PARAMETER;
+
+	for (index = 0; index < capabilities->cCapabilitiesSets; index++)
+	{
+		capabilitySet = &(capabilities->capabilitySets[index]);
+
+		if ((capabilitySet->capabilitySetType == CB_CAPSTYPE_GENERAL) &&
+		    (capabilitySet->capabilitySetLength >= CB_CAPSTYPE_GENERAL_LEN))
+		{
+			CLIPRDR_GENERAL_CAPABILITY_SET *generalCapabilitySet =
+			    (CLIPRDR_GENERAL_CAPABILITY_SET *)capabilitySet;
+			afc->clipboardCapabilities = generalCapabilitySet->generalFlags;
+			break;
+		}
+	}
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_server_format_list(CliprdrClientContext *cliprdr,
+                                           const CLIPRDR_FORMAT_LIST *formatList)
+{
+	UINT rc;
+	UINT32 index;
+	CLIPRDR_FORMAT *format;
+	mfContext *afc;
+
+	if (!cliprdr || !formatList)
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext *)cliprdr->custom;
+
+	if (!afc)
+		return ERROR_INVALID_PARAMETER;
+
+	if (afc->serverFormats)
+	{
+		for (index = 0; index < afc->numServerFormats; index++)
+			free(afc->serverFormats[index].formatName);
+
+		free(afc->serverFormats);
+		afc->serverFormats = NULL;
+		afc->numServerFormats = 0;
+	}
+
+	if (formatList->numFormats < 1)
+		return CHANNEL_RC_OK;
+
+	afc->numServerFormats = formatList->numFormats;
+	afc->serverFormats = (CLIPRDR_FORMAT *)calloc(afc->numServerFormats, sizeof(CLIPRDR_FORMAT));
+
+	if (!afc->serverFormats)
+		return CHANNEL_RC_NO_MEMORY;
+
+	for (index = 0; index < afc->numServerFormats; index++)
+	{
+		afc->serverFormats[index].formatId = formatList->formats[index].formatId;
+		afc->serverFormats[index].formatName = NULL;
+
+		if (formatList->formats[index].formatName)
+		{
+			afc->serverFormats[index].formatName = _strdup(formatList->formats[index].formatName);
+
+			if (!afc->serverFormats[index].formatName)
+				return CHANNEL_RC_NO_MEMORY;
+		}
+	}
+
+	BOOL unicode = FALSE;
+	BOOL text = FALSE;
+	for (index = 0; index < afc->numServerFormats; index++)
+	{
+		format = &(afc->serverFormats[index]);
+
+		if (format->formatId == CF_UNICODETEXT)
+			unicode = TRUE;
+
+		else if (format->formatId == CF_TEXT)
+			text = TRUE;
+	}
+
+	if (unicode)
+		return ios_cliprdr_send_client_format_data_request(cliprdr, CF_UNICODETEXT);
+	if (text)
+		return ios_cliprdr_send_client_format_data_request(cliprdr, CF_TEXT);
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT
+ios_cliprdr_server_format_list_response(CliprdrClientContext *cliprdr,
+                                        const CLIPRDR_FORMAT_LIST_RESPONSE *formatListResponse)
+{
+	if (!cliprdr || !formatListResponse)
+		return ERROR_INVALID_PARAMETER;
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT
+ios_cliprdr_server_lock_clipboard_data(CliprdrClientContext *cliprdr,
+                                       const CLIPRDR_LOCK_CLIPBOARD_DATA *lockClipboardData)
+{
+	if (!cliprdr || !lockClipboardData)
+		return ERROR_INVALID_PARAMETER;
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT
+ios_cliprdr_server_unlock_clipboard_data(CliprdrClientContext *cliprdr,
+                                         const CLIPRDR_UNLOCK_CLIPBOARD_DATA *unlockClipboardData)
+{
+	if (!cliprdr || !unlockClipboardData)
+		return ERROR_INVALID_PARAMETER;
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT
+ios_cliprdr_server_format_data_request(CliprdrClientContext *cliprdr,
+                                       const CLIPRDR_FORMAT_DATA_REQUEST *formatDataRequest)
+{
+	UINT rc;
+	BYTE *data;
+	UINT32 size;
+	UINT32 formatId;
+	CLIPRDR_FORMAT_DATA_RESPONSE response = { 0 };
+	mfContext *afc;
+
+	if (!cliprdr || !formatDataRequest || !cliprdr->ClientFormatDataResponse)
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext *)cliprdr->custom;
+
+	if (!afc)
+		return ERROR_INVALID_PARAMETER;
+
+	formatId = formatDataRequest->requestedFormatId;
+	data = (BYTE *)ClipboardGetData(afc->clipboard, formatId, &size);
+	response.common.msgFlags = CB_RESPONSE_OK;
+	response.common.dataLen = size;
+	response.requestedFormatData = data;
+
+	if (!data)
+	{
+		response.common.msgFlags = CB_RESPONSE_FAIL;
+		response.common.dataLen = 0;
+		response.requestedFormatData = NULL;
+	}
+
+	rc = cliprdr->ClientFormatDataResponse(cliprdr, &response);
+	free(data);
+	return rc;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT
+ios_cliprdr_server_format_data_response(CliprdrClientContext *cliprdr,
+                                        const CLIPRDR_FORMAT_DATA_RESPONSE *formatDataResponse)
+{
+	BYTE *data;
+	UINT32 size;
+	UINT32 index;
+	UINT32 formatId;
+	CLIPRDR_FORMAT *format = NULL;
+	mfContext *afc;
+	freerdp *instance;
+
+	if (!cliprdr || !formatDataResponse)
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext *)cliprdr->custom;
+
+	if (!afc)
+		return ERROR_INVALID_PARAMETER;
+
+	instance = ((rdpContext *)afc)->instance;
+
+	if (!instance)
+		return ERROR_INVALID_PARAMETER;
+
+	for (index = 0; index < afc->numServerFormats; index++)
+	{
+		if (afc->requestedFormatId == afc->serverFormats[index].formatId)
+			format = &(afc->serverFormats[index]);
+	}
+
+	if (!format)
+	{
+		SetEvent(afc->clipboardRequestEvent);
+		return ERROR_INTERNAL_ERROR;
+	}
+
+	if (format->formatName)
+		formatId = ClipboardRegisterFormat(afc->clipboard, format->formatName);
+	else
+		formatId = format->formatId;
+
+	size = formatDataResponse->common.dataLen;
+
+	ClipboardLock(afc->clipboard);
+	if (!ClipboardSetData(afc->clipboard, formatId, formatDataResponse->requestedFormatData, size))
+		return ERROR_INTERNAL_ERROR;
+
+	SetEvent(afc->clipboardRequestEvent);
+
+	if ((formatId == CF_TEXT) || (formatId == CF_UNICODETEXT))
+	{
+		formatId = ClipboardRegisterFormat(afc->clipboard, "UTF8_STRING");
+		data = (BYTE *)ClipboardGetData(afc->clipboard, formatId, &size);
+		size = (UINT32)strnlen(data, size);
+		if (afc->ServerCutText != NULL)
+		{
+			afc->ServerCutText((rdpContext *)afc, (uint8_t *)data, size);
+		}
+	}
+	ClipboardUnlock(afc->clipboard);
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT
+ios_cliprdr_server_file_contents_request(CliprdrClientContext *cliprdr,
+                                         const CLIPRDR_FILE_CONTENTS_REQUEST *fileContentsRequest)
+{
+	if (!cliprdr || !fileContentsRequest)
+		return ERROR_INVALID_PARAMETER;
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_server_file_contents_response(
+    CliprdrClientContext *cliprdr, const CLIPRDR_FILE_CONTENTS_RESPONSE *fileContentsResponse)
+{
+	if (!cliprdr || !fileContentsResponse)
+		return ERROR_INVALID_PARAMETER;
+
+	return CHANNEL_RC_OK;
+}
+
+BOOL ios_cliprdr_init(mfContext *afc, CliprdrClientContext *cliprdr)
+{
+	wClipboard *clipboard;
+	HANDLE hevent;
+
+	if (!afc || !cliprdr)
+		return FALSE;
+
+	if (!(hevent = CreateEvent(NULL, TRUE, FALSE, NULL)))
+		return FALSE;
+
+	if (!(clipboard = ClipboardCreate()))
+	{
+		CloseHandle(hevent);
+		return FALSE;
+	}
+
+	afc->cliprdr = cliprdr;
+	afc->clipboard = clipboard;
+	afc->clipboardRequestEvent = hevent;
+	cliprdr->custom = (void *)afc;
+	cliprdr->MonitorReady = ios_cliprdr_monitor_ready;
+	cliprdr->ServerCapabilities = ios_cliprdr_server_capabilities;
+	cliprdr->ServerFormatList = ios_cliprdr_server_format_list;
+	cliprdr->ServerFormatListResponse = ios_cliprdr_server_format_list_response;
+	cliprdr->ServerLockClipboardData = ios_cliprdr_server_lock_clipboard_data;
+	cliprdr->ServerUnlockClipboardData = ios_cliprdr_server_unlock_clipboard_data;
+	cliprdr->ServerFormatDataRequest = ios_cliprdr_server_format_data_request;
+	cliprdr->ServerFormatDataResponse = ios_cliprdr_server_format_data_response;
+	cliprdr->ServerFileContentsRequest = ios_cliprdr_server_file_contents_request;
+	cliprdr->ServerFileContentsResponse = ios_cliprdr_server_file_contents_response;
+	return TRUE;
+}
+
+BOOL ios_cliprdr_uninit(mfContext *afc, CliprdrClientContext *cliprdr)
+{
+	if (!afc || !cliprdr)
+		return FALSE;
+
+	cliprdr->custom = NULL;
+	afc->cliprdr = NULL;
+	ClipboardDestroy(afc->clipboard);
+	CloseHandle(afc->clipboardRequestEvent);
+	return TRUE;
+}
--- a/client/iOS/FreeRDP/ios_freerdp.h
+++ b/client/iOS/FreeRDP/ios_freerdp.h
@@ -2,6 +2,7 @@
  RDP run-loop
 
  Copyright 2013 Thincast Technologies GmbH, Authors: Martin Fleisz, Dorian Johnson
+ Copyright 2023 Iordan Iordanov
 
  This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0.
  If a copy of the MPL was not distributed with this file, You can obtain one at
@@ -13,8 +14,12 @@
 #import <freerdp/freerdp.h>
 #import <freerdp/channels/channels.h>
 #import "TSXTypes.h"
+#import <winpr/clipboard.h>
+#import <freerdp/client/cliprdr.h>
 
 @class RDPSession, RDPSessionView;
+
+typedef BOOL (*pServerCutText)(rdpContext *context, UINT8 *data, UINT32 size);
 
 // FreeRDP extended structs
 typedef struct mf_info mfInfo;
@@ -25,6 +30,16 @@ typedef struct mf_context
 
 	mfInfo *mfi;
 	rdpSettings *settings;
+
+	BOOL clipboardSync;
+	wClipboard *clipboard;
+	UINT32 numServerFormats;
+	UINT32 requestedFormatId;
+	HANDLE clipboardRequestEvent;
+	CLIPRDR_FORMAT *serverFormats;
+	CliprdrClientContext *cliprdr;
+	UINT32 clipboardCapabilities;
+	pServerCutText ServerCutText;
 } mfContext;
 
 struct mf_info
@@ -69,3 +84,4 @@ void ios_uninit_freerdp(void);
 freerdp *ios_freerdp_new(void);
 int ios_run_freerdp(freerdp *instance);
 void ios_freerdp_free(freerdp *instance);
+void ios_send_clipboard_data(void *context, const void *data, UINT32 size);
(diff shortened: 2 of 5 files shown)


akallabeth:
@freerdp-bot test

include/freerdp/update.h
```diff
@@ -190,6 +190,7 @@ typedef BOOL (*pSaveSessionInfo)(rdpContext* context, UINT32 type, void* data);
 typedef BOOL (*pSetKeyboardImeStatus)(rdpContext* context, UINT16 imeId, UINT32 imeState,
                                       UINT32 imeConvMode);
 typedef BOOL (*pServerStatusInfo)(rdpContext* context, UINT32 status);
+typedef BOOL (*pServerCutText)(rdpContext* context, UINT8* data, UINT32 size);
```
akallabeth, line 193: why is this necessary in the public API?

include/freerdp/update.h
```diff
@@ -231,7 +232,8 @@ struct rdp_update
 	 * fills BITMAP_DATA struct members: flags, cbCompMainBodySize and cbCompFirstRowSize.
 	 */
 	BOOL autoCalculateBitmapData; /* 71 */
-	UINT32 paddingE[80 - 72];     /* 72 */
+	pServerCutText ServerCutText; /* 72 */
```
akallabeth, line 235: why is this necessary in the public API?

akallabeth (changes requested):
Overall LGTM, but having the `ServerCutText` in public API is not good.
better move that to some client specific context. (`mfContext` or if it could be reused by other clients `rdpClientContext`)

So enough nitpicking, nice work, thank you :)

akallabeth:
@freerdp-bot test

include/freerdp/update.h
iiordanov: Hey @akallabeth thanks for reviewing. The reason it's there is the client needs to be able to set it as a call-back for the server to send clipboard updates. I thought it would be a good fit alongside

    instance->update->DesktopResize
    instance->update->BitmapUpdate
    instance->update->BeginPaint
    instance->update->EndPaint

It seemed logical to place it in an equally accessible place.
    instance->update->ServerCutText

That being said, if you think it's not the right place I'm happy to move it as long as i can access rdpClientContext directly from the instance created with freerdp_new() (that returns a freerdp struct). It's not immediately obvious how I can get ahold of rdpClientContext from freerdp.

include/freerdp/update.h
iiordanov: See the previous answer.

include/freerdp/update.h
akallabeth: @iiordanov sorry, your answer does not make any sense.
1. what do you mean with `client needs to be able to set it as a call-back for the server to send clipboard updates`?
2. What do you think this callback is for and how is it supposed to work?

akallabeth: @iiordanov also, you do not reference that callback anywhere, is this maybe supposed to be part of something else?

iiordanov: @akallabeth It's referenced in an app that uses FreeRDP:

https://gitlab.com/iiordanov/remote-desktop-clients-ios/-/blob/master/bVNC/RdpBridge.m#L226

akallabeth: @iiordanov ok, did check that. why do you patch this into the core library? it is something that should not be in there as that callback is only used to be called when the clipboard data is available in a desired format?
you are already in the client implementation where this is called and it is very specific for your client, so no, this just looks wrong there.

akallabeth: @iiordanov the android implementation you mention also has such stuff encapsulated to the client implementation (e.g. in `androidContext`)

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+	UINT rc = ERROR_INTERNAL_ERROR;
+	UINT32 index;
+	UINT32 formatId;
+	UINT32 numFormats;
+	UINT32* pFormatIds;
+	const char* formatName;
+	CLIPRDR_FORMAT* formats;
+	CLIPRDR_FORMAT_LIST formatList = { 0 };
+
+	if (!cliprdr)
```
pnowack, line 44: It would not be an error due to invalid input, but wrong API usage. Use an assertion (`WINPR_ASSERT()`) instead.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+		if ((formatId > CF_MAX) && formatName)
+		{
+			formats[index].formatName = _strdup(formatName);
+
+			if (!formats[index].formatName)
+				goto fail;
+		}
+	}
+
+	formatList.common.msgFlags = CB_RESPONSE_OK;
```
pnowack, line 75: This flag is wrong and does not conform to `[MS-RDPECLIP]`. Only allowed flag is the assci_names flag.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+	formatList.numFormats = numFormats;
+	formatList.formats = formats;
+	formatList.common.msgType = CB_FORMAT_LIST;
+
+	if (!afc->cliprdr->ClientFormatList)
```
pnowack, line 80: Remove this check. It won't ever happen (would also be an actual error in FreeRDP).

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+
+	rc = afc->cliprdr->ClientFormatList(afc->cliprdr, &formatList);
+fail:
+	free(pFormatIds);
+	free(formats);
+	return rc;
+}
+
+static UINT ios_cliprdr_send_client_format_data_request(CliprdrClientContext* cliprdr,
+                                                            UINT32 formatId)
```
pnowack, line 91: Run `clang-format` on the changed/added files, it will take care of the formatting.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+	UINT rc = ERROR_INVALID_PARAMETER;
+	CLIPRDR_FORMAT_DATA_REQUEST formatDataRequest = { 0 };
+	mfContext* afc;
+
+	if (!cliprdr)
+		goto fail;
+
+	afc = (mfContext*)cliprdr->custom;
+
+	if (!afc || !afc->clipboardRequestEvent || !cliprdr->ClientFormatDataRequest)
```
pnowack, line 102: The `!afc` and `!cliprdr->ClientFormatDataRequest` checks won't ever trigger here. Remove those two, since they are just dead code.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+{
+	UINT rc;
+	mfContext* afc;
+
+	if (!cliprdr || !monitorReady)
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext*)cliprdr->custom;
+
+	if (!afc)
```
pnowack, line 148: Like above, dead code.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_monitor_ready(CliprdrClientContext* cliprdr,
+                                          const CLIPRDR_MONITOR_READY* monitorReady)
+{
+	UINT rc;
+	mfContext* afc;
+
+	if (!cliprdr || !monitorReady)
```
pnowack, line 143: Remove this check. It won't ever happen (would also be an actual error in FreeRDP).

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_server_capabilities(CliprdrClientContext* cliprdr,
+                                                const CLIPRDR_CAPABILITIES* capabilities)
+{
+	UINT32 index;
+	CLIPRDR_CAPABILITY_SET* capabilitySet;
+	mfContext* afc;
+
+	if (!cliprdr || !capabilities)
```
pnowack, line 173: Remove this check. It won't ever happen (would also be an actual error in FreeRDP).

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext*)cliprdr->custom;
+
+	if (!afc)
```
pnowack, line 178: Like above, dead code.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+ */
+static UINT ios_cliprdr_server_format_list(CliprdrClientContext* cliprdr,
+                                               const CLIPRDR_FORMAT_LIST* formatList)
+{
+	UINT rc;
+	UINT32 index;
+	CLIPRDR_FORMAT* format;
+	mfContext* afc;
+
+	if (!cliprdr || !formatList)
```
pnowack, line 211: Remove this check. It won't ever happen (would also be an actual error in FreeRDP).

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext*)cliprdr->custom;
+
+	if (!afc)
```
pnowack, line 216: dead code

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT
+ios_cliprdr_server_format_list_response(CliprdrClientContext* cliprdr,
+                                            const CLIPRDR_FORMAT_LIST_RESPONSE* formatListResponse)
+{
+	if (!cliprdr || !formatListResponse)
```
pnowack, line 286: Remove this check. It won't ever happen (would also be an actual error in FreeRDP).

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT
+ios_cliprdr_server_lock_clipboard_data(CliprdrClientContext* cliprdr,
```
pnowack, line 298: No need for that, you did not advertise the locking capability.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_server_unlock_clipboard_data(
```
pnowack, line 312: No need for that, you did not advertise the locking capability.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+                                           const CLIPRDR_FORMAT_DATA_REQUEST* formatDataRequest)
+{
+	UINT rc;
+	BYTE* data;
+	UINT32 size;
+	UINT32 formatId;
+	CLIPRDR_FORMAT_DATA_RESPONSE response = { 0 };
+	mfContext* afc;
+
+	if (!cliprdr || !formatDataRequest || !cliprdr->ClientFormatDataResponse)
```
pnowack, line 337: `!cliprdr` and `!cliprdr->ClientFormatDataResponse` are guaranteed to be non-NULL.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext*)cliprdr->custom;
+
+	if (!afc)
```
pnowack, line 342: dead code (never hit).

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+{
+	BYTE* data;
+	UINT32 size;
+	UINT32 index;
+	UINT32 formatId;
+	CLIPRDR_FORMAT* format = NULL;
+	mfContext* afc;
+	freerdp* instance;
+
+	if (!cliprdr || !formatDataResponse)
```
pnowack, line 380: Remove this check. It won't ever happen (would also be an actual error in FreeRDP).

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext*)cliprdr->custom;
+
+	if (!afc)
```
pnowack, line 385: dead code (never hit), remove this.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+	SetEvent(afc->clipboardRequestEvent);
+
+	if ((formatId == CF_TEXT) || (formatId == CF_UNICODETEXT))
+	{
+		formatId = ClipboardRegisterFormat(afc->clipboard, "UTF8_STRING");
+		data = (BYTE*)ClipboardGetData(afc->clipboard, formatId, &size);
+		size = (UINT32)strnlen(data, size);
+		if (instance->context->update->ServerCutText != NULL)
+		{
+			instance->context->update->ServerCutText(afc, (uint8_t*)data, size);
```
pnowack, line 424: No, this should directly handle that situation. Code of virtual channels should not be leaked outside into the core part of FreeRDP.
What you need to keep in mind is that the iOS client is an actual client implementation here, not some abstraction layer.
The protocol parsing of `[MS-RDPECLIP]` is already done by the FreeRDP library, but this here is actual client implementation.

(This comment here links to akallabeths comment above in this MR.)

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_server_file_contents_request(
```
pnowack, line 436: You did not advertise the file capabilities, so, no need to have the callbacks implemented.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_server_file_contents_response(
```
pnowack, line 450: You did not advertise the file capabilities, so, no need to have the callbacks implemented.

client/iOS/FreeRDP/ios_freerdp.h
```diff
@@ -69,3 +81,4 @@ void ios_uninit_freerdp(void);
 freerdp *ios_freerdp_new(void);
 int ios_run_freerdp(freerdp *instance);
 void ios_freerdp_free(freerdp *instance);
+void ios_send_clipboard_data(void *context, const void* data, UINT32 size);
```
pnowack, line 84: Please add the missing new line after that added line.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+
+	return CHANNEL_RC_OK;
+}
+
+/**
+ * Function description
+ *
+ * @return 0 on success, otherwise a Win32 error code
+ */
+static UINT ios_cliprdr_server_format_list(CliprdrClientContext* cliprdr,
```
pnowack, line 203: When receiving a `FormatList` from the server side, you cannot expect to get a `FormatDataResponse` for a pending request.
While all requests should be answered, this is one thing were Microsoft messed up in the protocol, that they did not assign ids to new `FormatList`s and `FormatDataRequests`.

The following situation can happen:
Server sends `FormatList` with some text
The client side sends a `FormatListResponse` with the OK flag.
The client side sends a `FormatDataRequest` for the advertised text.
The server side sends in the same moment a new `FormatList`.

With MS Windows RDS, you won't get a response for your pending `FormatDataRequest`. Regardless of what is inside in this PDU, you first need to cancel your pending request and handle it in the way as it would have failed.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,503 @@
...
+
+#include <winpr/crt.h>
+#include <winpr/stream.h>
+
+#include <freerdp/client/channels.h>
+#include <freerdp/client/cliprdr.h>
+
+#include "ios_cliprdr.h"
+
+UINT ios_cliprdr_send_client_format_list(CliprdrClientContext* cliprdr)
```
pnowack, line 33: Similar to my comment to `ios_cliprdr_server_format_list`: You need to cancel any pending `FormatDataRequest` from here.

pnowack (changes requested):
There is one thing, that I don't know how it is handled on the client side, and that is whether your calls (like `ClientFormatList`, etc.) are handled in the same thread as the callback calls, where the PDUs from the server side are handled.

If these two things don't happen in the same thread, then extra synchronization needs to be taken.
`wClipboard` for example is not thread safe. If you use it on two different threads to convert data at the same time, the result is undefined.

There are (at least) two ways to handle such situations:

1. Ensure all things are handled in the same thread. In that case, the callbacks for the PDUs from the server side would push some task to your "handling thread".
2. Use mutexes and co for synchronization.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Hey Pascal, i am not sure I am familiar enough with FreeRDP to make this kind of change and actually ensure that it works. The majority of this implementation was simply taken from android_cliprdr.c. I think your comments are valuable but maybe we should take them up in a separate PR that refactors both ios_cliprdr.c and android_cliprdr.c. What do you think?

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_freerdp.h
iiordanov: Will do!

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as my first response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as my first comment.

include/freerdp/update.h
iiordanov: Let me try what you suggest @akallabeth .

client/iOS/FreeRDP/ios_cliprdr.m
pnowack: No, that's not how it works. You introduce a protocol violation here.

iiordanov: @pnowack formatList.common.msgFlags = CB_RESPONSE_OK; is exactly what is called in client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c.

What shall we do about android_cliprdr?

Are you referring to:

#define CB_ASCII_NAMES 0x0004

Shall we change both files then?

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: Same as the previous response.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: This code is actually inside client/iOS/FreeRDP, so it is indeed actual client implementation. Also, that's exactly the same spot where the android client in FreeRDP implements the same functionality except for Android.

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: @iiordanov not a blocking thing, but yes, the clients do some unnecessary checks. we moved to using assertions for required arguments some time ago, but did not have the time to go through the whole code base.
so, if you´d do that that would be great (and to do it in another pr most likely the best way to go)

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: that line can be removed, the `cliprdr_packet_format_list_new` ignores this field anyway.
the `CB_ASCII_NAMES` state from capability exchange is stored internally and used to determine if short or long names can be used.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@iiordanov run a `git clang-format origin/master -f` on your changes, the formatting looks broken and the CI complains.

iiordanov:
> @iiordanov run a `git clang-format origin/master -f` on your changes, the formatting looks broken and the CI complains.

Hi @akallabeth I reformatted earlier today. It also seems to have fixed an existing issue in a file I had changed.

akallabeth:
@freerdp-bot test

akallabeth:
@iiordanov ok, pr looks good so far. do you want to change the `assert` suggestions for the ios client in this pr? (my previous comment was a bit ambigous, I meant that the changes for android you suggested might be better of in another pr)

iiordanov:
> @iiordanov ok, pr looks good so far. do you want to change the `assert` suggestions for the ios client in this pr? (my previous comment was a bit ambigous, I meant that the changes for android you suggested might be better of in another pr)

@akallabeth I'd rather tackle the dead-code cleanup and assertions in a new (set of) PR(s).

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: moved to follow up pr

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: moved to follow up pr

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: moved to follow up pr

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: moved to follow up pr

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: moved to follow up pr

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: moved to follow up pr

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: moved to follow up pr

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: moved to follow up pr

akallabeth:
@iiordanov then the only remaining thing is the remark from @pnowack with thread safety, client and server messages happen on different threads.

akallabeth:
@iiordanov any updates? we´re waiting for the thread locking to be fixed before a merge.

iiordanov:
> @iiordanov any updates? we´re waiting for the thread locking to be fixed before a merge.

Apologies for the delay @akallabeth I got really busy and wasn't able to look at this until today. Thanks for putting it back up on my agenda, however :).

I've added calls to ClipboardLock and ClipboardUnlock where it seemed appropriate. Also added a missing cast that was giving me a warning.

Let me know if you and @pnowack think this will work.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,508 @@
...
+	if (!cliprdr)
+		return ERROR_INVALID_PARAMETER;
+
+	mfContext *afc = (mfContext *)cliprdr->custom;
+
+	if (!afc || !afc->cliprdr)
+		return ERROR_INVALID_PARAMETER;
+
+	pFormatIds = NULL;
+	numFormats = ClipboardGetFormatIds(afc->clipboard, &pFormatIds);
```
akallabeth, line 54: missing locking

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,508 @@
...
+	formats = (CLIPRDR_FORMAT *)calloc(numFormats, sizeof(CLIPRDR_FORMAT));
+
+	if (!formats)
+		goto fail;
+
+	for (index = 0; index < numFormats; index++)
+	{
+		formatId = pFormatIds[index];
+		formatName = ClipboardGetFormatName(afc->clipboard, formatId);
```
akallabeth, line 63: missing locking

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,508 @@
...
+	if (!cliprdr || !formatDataRequest || !cliprdr->ClientFormatDataResponse)
+		return ERROR_INVALID_PARAMETER;
+
+	afc = (mfContext *)cliprdr->custom;
+
+	if (!afc)
+		return ERROR_INVALID_PARAMETER;
+
+	formatId = formatDataRequest->requestedFormatId;
+	data = (BYTE *)ClipboardGetData(afc->clipboard, formatId, &size);
```
akallabeth, line 348: missing locking

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,508 @@
...
+	}
+
+	if (!format)
+	{
+		SetEvent(afc->clipboardRequestEvent);
+		return ERROR_INTERNAL_ERROR;
+	}
+
+	if (format->formatName)
+		formatId = ClipboardRegisterFormat(afc->clipboard, format->formatName);
```
akallabeth, line 408: missing locking

akallabeth (changes requested):
you´ve missed a few places to lock (format names and ids depend on data that might change depending on registered/announced formats)

akallabeth:
> > @iiordanov any updates? we´re waiting for the thread locking to be fixed before a merge.
> 
> Apologies for the delay @akallabeth I got really busy and wasn't able to look at this until today. Thanks for putting it back up on my agenda, however :).
> 
> I've added calls to ClipboardLock and ClipboardUnlock where it seemed appropriate. Also added a missing cast that was giving me a warning.
> 
> Let me know if you and @pnowack think this will work.

that is what I thought :)
(know that issue with time too well :/)

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: @akallabeth if this is the case then why aren't we locking and unlocking the keyboard before calls to ClipboardGetFormatIds in wlf_cliprdr.c and xf_cliprdr.c? Is that a bug?

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: @akallabeth if this is the case then why aren't we locking and unlocking the keyboard before calls to ClipboardGetFormatName in wlf_cliprdr.c and xf_cliprdr.c?

iiordanov:
I've added locking to all remaining places.

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: there are no calls to these functions there? do you mix that up with calls to `ClipboardGetFormatIdString`?

client/iOS/FreeRDP/ios_cliprdr.m
akallabeth: that might actually be the case, checking this.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@pnowack anything else left or can we merge this?

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,524 @@
...
+		}
+	}
+
+	for (index = 0; index < afc->numServerFormats; index++)
+	{
+		format = &(afc->serverFormats[index]);
+
+		if (format->formatId == CF_UNICODETEXT)
+		{
+			if ((rc = ios_cliprdr_send_client_format_data_request(cliprdr, CF_UNICODETEXT)) !=
```
pnowack, line 263: Since you are only interested in text formats here, change just this to  `return ios_cliprdr_send_client_format_data_request(cliprdr, CF_UNICODETEXT)`.

Currently, you are only doing this, when that call fails. But if it does not (regular case), then you return `CHANNEL_RC_OK` in the end anyway. Since in this situation you already request the data, there is no point in further actions here.
In fact, it actually causes issues: If (and that is not unlikely) the server side advertises both `CF_UNICODETEXT` and `CF_TEXT` in the same format list, you will ask for data twice.
As format data requests are not coupled with ids, and since you set `afc->requestedFormatId` before sending the id, you may end up sending a request for `CF_UNICODETEXT` (where `afc->requestedFormatId` is set too), and sending a request for `CF_TEXT` (where `afc->requestedFormatId` is also set again).
Upon receiving the format data response, you obviously need `afc->requestedFormatId` again, but very likely, the server sent data for `CF_UNICODETEXT`, but for conversion, you use `CF_TEXT` here.

This is especially relevant for copying non-ASCII content (like non-english locale text). `CF_TEXT` is also dependent on the locale, `CF_UNICODETEXT` is not.

So, overall, in addition to the direct return, which I mentioned in this comment at the beginning. Do your search in two separate for loops.
In the first one, you look for `CF_UNICODETEXT`. In the second one, you look for `CF_TEXT`.
With the direct return, you will upon finding what you wanted, send the request, and just return with that.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,524 @@
...
+	else
+	{
+		formatId = format->formatId;
+	}
+
+	size = formatDataResponse->common.dataLen;
+
+	ClipboardLock(afc->clipboard);
+	res = ClipboardSetData(afc->clipboard, formatId, formatDataResponse->requestedFormatData, size);
+	ClipboardUnlock(afc->clipboard);
```
pnowack, line 429: Between setting the data and getting the converted one, the clipboard should not be unlocked. Of course, you must then not forget to unlock the clipboard before returning the error.

client/iOS/FreeRDP/ios_freerdp.h
```diff
@@ -69,3 +84,4 @@ void ios_uninit_freerdp(void);
 freerdp *ios_freerdp_new(void);
 int ios_run_freerdp(freerdp *instance);
 void ios_freerdp_free(freerdp *instance);
+void ios_send_clipboard_data(void *context, const void *data, UINT32 size);
```
pnowack, line 87: The new line after this line is missing.

client/iOS/FreeRDP/ios_cliprdr.m
```diff
@@ -0,0 +1,524 @@
...
+	afc = (mfContext *)cliprdr->custom;
+
+	if (!afc)
+		return ERROR_INVALID_PARAMETER;
+
+	formatId = formatDataRequest->requestedFormatId;
+	ClipboardLock(afc->clipboard);
+	data = (BYTE *)ClipboardGetData(afc->clipboard, formatId, &size);
+	ClipboardUnlock(afc->clipboard);
+	response.common.msgFlags = CB_RESPONSE_OK;
```
pnowack, line 355: You don't call `ClipboardSetData`, in the same lock situation, as when you get the data here. `ClipboardGetData` may fail here. So set the response flag according to that, i.e. if `size && data`, then `CB_RESPONSE_OK`, otherwise `CB_RESPONSE_FAIL`.

pnowack:
Added a review comment here about a line regarding the OK/FAIL flag for a format data response. Deleted that one, since it does not apply (did not see the handling after that line in the first glance). So, just in case you are wondering why you may got a mail about that comment, but don't see it here.

> @pnowack anything else left or can we merge this?

Just the three little things I added here above.

wanhongjun:
Hello, brother.
I have used this branch alone, because I am not professional enough, so I met some problems, please give me some advice, the following is the code I used:

wanhongjun:
Hello, brother.
I have used this branch alone, because I am not professional enough, so I met some problems, please give me some advice, the following is the code I used:

wanhongjun:
Hello, brother.
I have used this branch alone, because I am not professional enough, so I met some problems and failed to paste it successfully. Please give me some advice. The following is the code I used:

- (void)sendClipbord:(NSString *)value {
    if (value.length > 0) {
        if ([self mfi]->connection_state == TSXConnectionConnected) {
            mfInfo *mfi = [self mfi];
            ios_send_clipboard_data(mfi->context, [value cStringUsingEncoding:NSUTF8StringEncoding], 150);
        }
    }
}

Here is the printable when I called paste to linux:

[10:55:50:051] [25871:012a8580] [DEBUG][com.freerdp.client.common] - [freerdp_client_settings_parse_command_line]: This is 3.0.0-dev Build configuration: BUILD_TESTING=OFF WINPR_HAVE_POLL_H=1 WITH_CAIRO=OFF WITH_CCACHE=ON WITH_CHANNELS=ON WITH_CJSON=OFF WITH_CLANG_FORMAT=ON WITH_CLIENT=ON WITH_CLIENT_AVAILABLE=1 WITH_CLIENT_CHANNELS=ON WITH_CLIENT_CHANNELS_AVAILABLE=1 WITH_CLIENT_COMMON=ON WITH_CLIENT_INTERFACE=OFF WITH_CLIENT_SDL=ON WITH_CLIENT_SDL_AVAILABLE=1 WITH_DEBUG_ALL=OFF WITH_DEBUG_CAPABILITIES=OFF WITH_DEBUG_CERTIFICATE=OFF WITH_DEBUG_CHANNELS=OFF WITH_DEBUG_CLIPRDR=OFF WITH_DEBUG_CODECS=OFF WITH_DEBUG_DVC=OFF WITH_DEBUG_EVENTS=OFF WITH_DEBUG_KBD=OFF WITH_DEBUG_LICENSE=OFF WITH_DEBUG_MUTEX=OFF WITH_DEBUG_NEGO=OFF WITH_DEBUG_NLA=OFF WITH_DEBUG_NTLM=OFF WITH_DEBUG_RAIL=OFF WITH_DEBUG_RDP=OFF WITH_DEBUG_RDPDR=OFF WITH_DEBUG_RDPEI=OFF WITH_DEBUG_RDPGFX=OFF WITH_DEBUG_REDIR=OFF WITH_DEBUG_RFX=OFF WITH_DEBUG_RINGBUFFER=OFF WITH_DEBUG_SCARD=OFF WITH_DEBUG_SDL_EVENTS=OFF WITH_DEBUG_SDL_KBD_EVENTS=OFF WITH_DEBUG_SND=OFF WITH_DEBUG_SVC=OFF WITH_DEBUG_SYMBOLS=OFF WITH_DEBUG_THREADS=OFF WITH_DEBUG_TIMEZONE=OFF WITH_DEBUG_TRANSPORT=OFF WITH_DEBUG_TSG=OFF WITH_DEBUG_TSMF=OFF WITH_DEBUG_TSMF=OFF WITH_DEBUG_TSMF_AVAILABLE=0 WITH_DEBUG_WND=OFF WITH_DEBUG_X11=OFF WITH_DEBUG_X11_LOCAL_MOVESIZE=OFF WITH_DEBUG_XV=OFF WITH_DSP_EXPERIMENTAL=OFF WITH_DSP_FFMPEG=OFF WITH_DSP_FFMPEG=OFF WITH_DSP_FFMPEG_AVAILABLE=0 WITH_FAAC=OFF WITH_FAAD2=OFF WITH_FFMPEG=OFF WITH_FREERDP_DEPRECATED=OFF WITH_FREERDP_DEPRECATED_COMMANDLINE=OFF WITH_FUSE=OFF WITH_GFX_H264=OFF WITH_GPROF=OFF WITH_GSM=OFF WITH_INTERNAL_MD4=OFF WITH_INTERNAL_MD5=OFF WITH_JPEG=OFF WITH_KRB5=OFF WITH_LAME=OFF WITH_LIBRARY_VERSIONING=ON WITH_LIBSYSTEMD=OFF WITH_MACAUDIO=OFF WITH_MACAUDIO=OFF WITH_MACAUDIO_AVAILABLE=0 WITH_MANPAGES=ON WITH_MBEDTLS=OFF WITH_NATIVE_SSPI=OFF WITH_OPENCL=OFF WITH_OPENH264=OFF WITH_OPENSSL=ON WITH_PKCS11=OFF WITH_POLL=ON WITH_PROFILER=OFF WITH_SAMPLE=OFF WITH_SANITIZE_ADDRESS=OFF WITH_SANITIZE_ADDRESS_AVAILABLE=1 WITH_SANITIZE_MEMORY=OFF WITH_SANITIZE_MEMORY_AVAILABLE=1 WITH_SANITIZE_THREAD=OFF WITH_SANITIZE_THREAD_AVAILABLE=1 WITH_SERVER=OFF WITH_SERVER_INTERFACE=ON WITH_SMARTCARD_EMULATE=OFF WITH_SMARTCARD_INSPECT=OFF WITH_SMARTCARD_PCSC=ON WITH_SOXR=OFF WITH_SSE2=OFF WITH_SWSCALE=OFF WITH_THIRD_PARTY=OFF WITH_VAAPI=OFF WITH_VAAPI=OFF WITH_VAAPI_AVAILABLE=0 WITH_VALGRIND_MEMCHECK=OFF WITH_VALGRIND_MEMCHECK_AVAILABLE=1 WITH_VERBOSE_WINPR_ASSERT=ON WITH_VIDEO_FFMPEG=OFF WITH_VIDEO_FFMPEG=OFF WITH_VIDEO_FFMPEG_AVAILABLE=0 WITH_WINPR_DEPRECATED=OFF WITH_WINPR_TOOLS=ON WITH_ZLIB=ON
Build type:          Release
CFLAGS:              
Compiler:            AppleClang, 14.0.0.14000029

[10:55:51:634] [25871:6fa2f000] [DEBUG][com.freerdp.client.common.cmdline] - [freerdp_client_load_static_channel_addin]: loading channelEx rdpdr

[10:57:43:104] [25871:012a8580] [DEBUG][com.freerdp.channels.channels.cliprdr.client] - [cliprdr_client_format_list]: ClientFormatList: numFormats: 1
[10:57:43:104] [25871:012a8580] [DEBUG][com.freerdp.channels.cliprdr.client] - [cliprdr_packet_send]: Cliprdr Sending (36 bytes)
[10:57:43:136] [25871:6fceb000] [DEBUG][com.freerdp.channels.cliprdr.client] - [cliprdr_order_recv]: msgType: CB_FORMAT_LIST_RESPONSE (3), msgFlags: 1 dataLen: 0
[10:57:43:136] [25871:6fceb000] [DEBUG][com.freerdp.channels.channels.cliprdr.client] - [cliprdr_process_format_list_response]: ServerFormatListResponse

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: I changed the implementation according to your recommendations, thanks @pnowack! Please take a look.

client/iOS/FreeRDP/ios_cliprdr.m
iiordanov: I changed the implementation according to your recommendations, thanks @pnowack! Please take a look.

akallabeth (approved):
LGTM

client/iOS/FreeRDP/ios_cliprdr.m
pnowack: This wasn't addressed in your push (https://github.com/FreeRDP/FreeRDP/compare/5938d557f59c64fc49bcab1385a193f0aa011491..fc1bf8ff6d115a8a9783d87f17bc5160f5ec72bc). The situation is still the same here.

akallabeth:
@freerdp-bot test

pnowack (approved):

akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #8722, opened 2023-02-24
expose last NLA/CredSSP SSPI error code (freerdp_get_nla_sspi_error)

awakecoding:
Through our recent work with [Kerberos in sspi-rs](https://github.com/Devolutions/sspi-rs/blob/master/src/utils.rs#L51), we've had to diagnose issues where the last SSPI error code would be extremely useful to report better error messages in the user interface, such as a connection failure caused by a clock skew (happens all the time). Until now, the SSPI error code in credssp_auth_authenticate was simply discarded in case of authentication failure. This pull request saves that last error code into the credssp context, and exposes it to the nla context, which is then exposed externally through freerdp_get_nla_sspi_error.

Ideally, the internal Kerberos error code could also be propagated, but it's more difficult to do since we'd need to use custom SSPI context attributes. For now, we've managed to map most Kerberos error codes to meaningful SSPI error codes, the only thing missing was a way to get that SSPI error code propagated all the way to the application calling the FreeRDP APIs.

This way, we can warn users about a clock skew being the cause for their Kerberos authentication failures, which will save us tons of hours of debugging trying to figure it from logs, especially on a mobile device where it's very tricky.

### DIFF
--- a/libfreerdp/core/freerdp.c
+++ b/libfreerdp/core/freerdp.c
@@ -1131,6 +1131,19 @@ BOOL freerdp_nla_revert_to_self(rdpContext* context)
 	return nla_revert_to_self(nla);
 }
 
+UINT32 freerdp_get_nla_sspi_error(rdpContext* context)
+{
+	rdpNla* nla;
+
+	WINPR_ASSERT(context);
+	WINPR_ASSERT(context->rdp);
+	WINPR_ASSERT(context->rdp->transport);
+
+	nla = transport_get_nla(context->rdp->transport);
+
+	return nla_get_sspi_error(nla);
+}
+
 HANDLE getChannelErrorEventHandle(rdpContext* context)
 {
 	WINPR_ASSERT(context);
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -1790,3 +1790,9 @@ DWORD nla_get_error(rdpNla* nla)
 		return ERROR_INTERNAL_ERROR;
 	return nla->errorCode;
 }
+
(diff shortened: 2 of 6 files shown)


libfreerdp/core/nla.c
```diff
@@ -1790,3 +1790,11 @@ DWORD nla_get_error(rdpNla* nla)
 		return ERROR_INTERNAL_ERROR;
 	return nla->errorCode;
 }
+
+UINT32 nla_get_sspi_error(rdpNla* nla)
+{
+	if (!nla || !nla->auth)
```
akallabeth, line 1796: better `WINPR_ASSERT` these, at least `nla` must not be `NULL`

libfreerdp/core/freerdp.c
```diff
@@ -1131,6 +1131,24 @@ BOOL freerdp_nla_revert_to_self(rdpContext* context)
 	return nla_revert_to_self(nla);
 }
 
+UINT32 freerdp_get_nla_sspi_error(rdpContext* context)
+{
+	rdpNla* nla;
+
+	if (!context)
```
akallabeth, line 1138: better `WINPR_ASSERT` these, at least `context`, `context->rdp` and `context->rdp->transport` must not be `NULL`

libfreerdp/core/nla.c
awakecoding: fixed

libfreerdp/core/freerdp.c
awakecoding: fixed

akallabeth (approved):
LGTM

PR #8720, opened 2023-02-24
Clip data len fix

akallabeth:
* Fixed some issues with the uri list generation
* Add some logging for selection events
* make cli functions use abortable `fgetc` et al

PR #8727, opened 2023-02-27
Clip file split

akallabeth:
based on #8720 

splits away the file clipboard to `client/common` to be reused by `wlfreerdp`, `sdl-freerdp` and `xfreerdp` (and maybe `mfreerdp` too)

What works:
1. clients only announce local paste if the file clipboard supports fuse
2. copy client --> server

what is todo:
1. the clipboard lock functionality is not supported (requires some additional changes to keep the local/remote file lists in memory)
3. there are still some missing improvements on client --> server copy (we should cache open `FILE*` but need to keep track of the amount of open files somehow)

what is still buggy:
* ~~fuse file list generation is not reliable, working on that in a follow up pr.~~ fixed for single non locking clipboard use case

these issues are now contained to the file clipboard implementation, so I´d appreciate if I could get a review of these (major) changes and fix the issues in a separate pr.


KangLin:
QT is also needed.

[WinFsp · Windows File System Proxy](https://github.com/winfsp/winfsp)
[osxfuse](https://osxfuse.github.io/): macFUSE allows you to extend macOS's native file handling capabilities via third-party file systems.

akallabeth:
@KangLin what do you mean with `QT`?

KangLin:
[My program](https://github.com/KangLin/RabbitRemoteControl) uses qt.

splits away the file clipboard to client/common to be reused by wlfreerdp, sdl-freerdp and xfreerdp (and maybe mfreerdp too) and qt.

Also consider that [WinFsp · Windows File System Proxy](https://github.com/winfsp/winfsp) may be used under Windows

libfreerdp/utils/cliprdr_utils.c
```diff
@@ -131,6 +111,66 @@ UINT cliprdr_parse_file_list(const BYTE* format_data, UINT32 format_data_length,
...
+
+	if (!Stream_EnsureRemainingCapacity(s, sizeof(FILEDESCRIPTORW)))
+		return FALSE;
+
+	Stream_Write_UINT32(s, file->dwFlags); /* flags (4 bytes) */
+
+	Stream_Write_UINT32(s, file->clsid.Data1);
+	Stream_Write_UINT16(s, file->clsid.Data2);
+	Stream_Write_UINT16(s, file->clsid.Data3);
+	WINPR_ASSERT(sizeof(file->clsid.Data4) == 8);
```
hardening, line 157: strange assert !

hardening (approved):

client/common/client_cliprdr_file.c
```diff
@@ -0,0 +1,2371 @@
...
+
+	log(cur->context->log, WLOG_WARN, __FILE__, __FUNCTION__, __LINE__,
+	    "[key %" PRIu32 "] lockID %" PRIu32 ", count %" PRIuz ", locked %d", *ukey, cur->lockId,
+	    cur->count, cur->locked);
+	for (size_t x = 0; x < cur->count; x++)
+	{
+		const CliprdrLocalFile* file = &cur->files[x];
+		log(cur->context->log, WLOG_WARN, __FILE__, __FUNCTION__, __LINE__, "file [%" PRIuz "] ", x,
+		    file->name, file->size);
+	}
```
hendwolt, line 1472: Shouldn't this function return a value?

akallabeth: good catch

libfreerdp/utils/cliprdr_utils.c
akallabeth: indeed, that one was a copy & paste one, so will update.

PR #8732, opened 2023-03-01
Deprecation warn

akallabeth:
like we talked about, lets inform the user that clients are deprecated.

### DIFF
--- a/client/Mac/cli/main.m
+++ b/client/Mac/cli/main.m
@@ -7,10 +7,13 @@
 //
 
 #import <Cocoa/Cocoa.h>
+#include <freerdp/client/cmdline.h>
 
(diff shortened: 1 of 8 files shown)


akallabeth:
@freerdp-bot test please

hardening:
Hum isn't there a contradiction to deprecate most clients and point to the SDL client, and at the same time print some experimental warning for the SDL client ?

akallabeth:
@hardening `SDL` client improvement is on the todo list. It has already more features than mac/wayland clients. (only exception being `clipboard` which is still todo)
anyway, this should inform users about the deprecation and (hoepfully) point someone our way for maintaining the clients if there is interest.

hardening (approved):

PR #8737, opened 2023-03-01
[client,x11] wrap XChangeProperty

akallabeth:
to debug bug reports occuring only with certain setups wrap the function call with a logging edition so we have the arguments available in the log for debugging.


hardening (approved):
Looks ok, I don't know if this PR was meant to be merged, or just to help debugging (maybe both)

PR #8738, opened 2023-03-01
[client,common] fix /gateway option parsing

akallabeth:
due to a bug the first letter of the gateway options was ignored.


akallabeth:
@freerdp-bot test

hardening (approved):

PR #8742, opened 2023-03-03
FIX: install export target

KangLin:


KangLin:
add_channel_client_library
add_channel_client_subsystem_library
add_channel_server_library

The parameter of the function _dynamic, but it is not used. Just generate a static library.
Are these functions not completed?

akallabeth:
ok, I think the issue you had was fixed with #8777

PR #8749, opened 2023-03-06
[core,peer] fix multitransport request checks

akallabeth:
If the client does not announced support for UDPFECR then do not send a multitransport request and just skip the state.


hardening (approved):
Oh good catch

hardening:
That will do the trick, but I guess that the final fix would have a server and client multi-transport flags, so that we can take decisions on what we do locally server-side (only reliable) and what the remote client supports...

akallabeth:
@hardening that is already the case here, the server initializes the settings with the flags it supports and when receiving the client flags a `serverFlag & clientFlag` is done, so only agreed upon flags are available.

PR #8751, opened 2023-03-06
common: Fix parsing of rdp file domain info

mfleisz:
`freerdp_parse_username` always returns non-NULL domain information when called. This currently results in `settings->domain` being overridden in every case, even though we might have read domain information from the file before.

This PR fixes this issue by checking if domain information was present in the file and if not use the parsed domain informatin.


akallabeth (approved):
LGTM

PR #8755, opened 2023-03-06
[client,sdl] lock updates

akallabeth:
use a critical section to lock SDL related data updates


client/SDL/sdl_freerdp.c
```diff
@@ -611,16 +612,25 @@ static BOOL sdl_create_windows(sdlContext* sdl)
 
 	rc = TRUE;
 fail:
+
 	if (!SetEvent(sdl->windows_created))
```
hardening, line 616: hum strange we notify the event even in the failure case...

akallabeth: we don´t want to lock up here, just return an error. (some thread is waiting on the other end)

PR #8757, opened 2023-03-07
common: Minor code cleanup of rdp file handling

mfleisz:
This PR contains various changes to rdp file handling:

The old code had a strange mixture of handling settings. When loading a file every line of the file was cached in `rdpFile::lines`. Sometimes functions would operate on these cached lines, sometimes they would operate on the actual values in the `rdpFile` instance.

On the other hand if an `rdpFile` instance was created from `rdpSettings`, this line cache simply did not exist, causing functions to behave differently, depending on whether the instance was created by reading a file or by populating it from `rdpSettings`.

The new implementation has now a single way of accessing values ( `find_integer_entry`/`find_string_entry`) and the `rdpFile::lines` data is used to handle unknown settings.

The PR also adds some argument checking and assertions.


akallabeth (approved):
Overall LGTM
need to run a few tests though.

PR #8758, opened 2023-03-07
Add RDSTLS security protocol

jotolo2:
Hi!
I've finally implemented RDSTLS security protocol.

I've tested and works in:
- FreeRDP server - mstsc client
- RDP Windows Server - FreeRDP client
- FreeRDP server - FreeRDP client

I should check what other clients support RDSLTS.


mfleisz:
@freerdp-bot test

akallabeth:
@jotolo2 looks like you should run `git clang-format` on your changes, the pr tester complains about that

libfreerdp/core/transport.c
```diff
@@ -773,6 +838,82 @@ SSIZE_T transport_parse_pdu(rdpTransport* transport, wStream* s, BOOL* incomplet
 			}
 		}
 	}
+	else if (transport->RdstlsMode)
+	{
+		UINT16 version = (header[1] << 8) | header[0];
```
akallabeth, line 843: Might be a good idea to use the `Stream_Read_UINT16` et al functions here. old code uses this style, but for new code it might be the better choice.

libfreerdp/core/rdstls.c
```diff
@@ -0,0 +1,784 @@
...
+{
+	const size_t pos = Stream_GetPosition(s);
+
+	if (!Stream_EnsureRemainingCapacity(s, 2))
+		return -1;
+
+	if (!str)
+	{
+    /* Write unicode null */
+		Stream_Write_UINT16(s, 2);
```
akallabeth, line 140: unchecked writes, you only ensure 2 bytes available here.

libfreerdp/core/rdstls.c
```diff
@@ -0,0 +1,784 @@
...
+}
+
+static BOOL rdstls_write_data(wStream* s, UINT32 length, const BYTE* data)
+{
+	if (!Stream_EnsureRemainingCapacity(s, 2))
+		return -1;
+
+	if (!data)
+	{
+    /* Write unicode null */
```
akallabeth, line 165: unchecked write, you only ensure 2 bytes available here.

libfreerdp/core/rdstls.c
```diff
@@ -0,0 +1,784 @@
...
+		return FALSE;
+
+	if (length <= 2)
+  {
+    Stream_Seek(s, length);
+    return TRUE;
+  }
+
+	free(*pData);
+	*pData = (BYTE*)malloc(length);
```
akallabeth, line 301: no need to allocate, your only use just needs a pointer to the ` wStream` and a size, then you can remove the free in `rdstls_process_authentication_request_with_password`

akallabeth (changes requested):
Overall LGTM, nice work!
some small changes are required though (and a reformat)

libfreerdp/core/transport.c
```diff
@@ -773,6 +838,82 @@ SSIZE_T transport_parse_pdu(rdpTransport* transport, wStream* s, BOOL* incomplet
...
+		UINT16 pduType;
+		UINT16 dataType;
+
+		if (version != RDSTLS_VERSION_1)
```
hardening, line 847: General remark about `transport_parse_pdu` that goes beyond your changes (but that your changes expose): it really look like we do the job twice.
We have that function that will compute the PDU size, but to do so it partially does the parsing of the PDU. So I guess that we always do the computation of the size that does a partial parsing, and then we do the parsing itself when treating the PDU. One could legitimately ask why not doing all that in a single step: having a smart parsing function that would parse the block and be able to say if there was enough bytes (complete parsing) and how many were expected if not...

libfreerdp/core/transport.c
```diff
@@ -773,6 +838,82 @@ SSIZE_T transport_parse_pdu(rdpTransport* transport, wStream* s, BOOL* incomplet
...
+		switch (pduType)
+		{
+			case RDSTLS_TYPE_CAPABILITIES:
+				pduLength = 8;
+				break;
+			case RDSTLS_TYPE_AUTHREQ:
+				if (position < 6)
+					return 0;
+
+				dataType = (header[5] << 8 | header[4]);
```
hardening, line 866: Definitely use a static stream and `Stream_Read_XXX` functions to do all that stuff.

libfreerdp/core/rdstls.c
akallabeth: pushed a fix.

libfreerdp/core/rdstls.c
akallabeth: pushed a fix.

akallabeth:
@freerdp-bot test this please

libfreerdp/core/rdstls.c
```diff
@@ -0,0 +1,787 @@
...
+{
+	WINPR_ASSERT(rdstls);
+	return rdstls->state;
+}
+
+static BOOL rdstls_set_state(rdpRdstls* rdstls, RDSTLS_STATE state)
+{
+	WINPR_ASSERT(rdstls);
+
+	WLog_DBG(TAG, "-- %s\t--> %s", rdstls_get_state_str(rdstls->state),
```
akallabeth, line 112: Might be a good idea to check allowed state transitions here and bail out if an invalid one is detected.
Also might be a good idea to check the return of that function where it is used.

jotolo2:
@hardening regarding the state machine and the callstack, what's the part you don't like and how could I improve it?

akallabeth:
@jotolo2 looks like we did both check the remaining issues at the same time. did do a few improvements here:  https://github.com/akallabeth/freerdp/tree/rdstls that does a little refactoring and state transition checks along with better logging. If you like it you can merge it here.

akallabeth:
@freerdp-bot test please

akallabeth:
@jotolo2 I´ve rebased my changes on top of your pr, will be pushing the commits to this. @hardening then it should be ready to merge.

akallabeth:
@freerdp-bot test

libfreerdp/core/transport.c
akallabeth: @jotolo2 did rewrite this to use `wStream` and I´ve pushed 66f657877dd910a611a3007106b0ee04c8842857 that does the `rdpRdstls` parsing in another function

akallabeth (approved):

PR #8759, opened 2023-03-07
Redirection fixes

jotolo2:
Hello.
Some little fixes.

The most important one is the one that makes `NLA` use the redirection password:
I've checked and mstsc client and Windows Remote Desktop client use the old password when reconnecting on a redirection process.
Though, other RDP clients use the new password.
If we have `RDSTLS`, we could rely on it for redirection with windows clients and use `NLA` for redirection with other clients to keep compatibility because I think most of the clients use the new password on `NLA` but don't support `RDSTLS`.

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -618,16 +618,7 @@ BOOL rdp_client_redirect(rdpRdp* rdp)
 	settings = rdp->settings;
 	WINPR_ASSERT(settings);
 
-	if (settings->RedirectionFlags & LB_LOAD_BALANCE_INFO)
-	{
-		if (settings->LoadBalanceInfo && (settings->LoadBalanceInfoLength > 0))
-		{
-			if (!nego_set_routing_token(rdp->nego, settings->LoadBalanceInfo,
-			                            settings->LoadBalanceInfoLength))
-				return FALSE;
-		}
-	}
-	else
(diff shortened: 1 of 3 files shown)


mfleisz:
@freerdp-bot test

libfreerdp/core/connection.c
```diff
@@ -618,16 +618,7 @@ BOOL rdp_client_redirect(rdpRdp* rdp)
...
-	{
-		if (settings->LoadBalanceInfo && (settings->LoadBalanceInfoLength > 0))
-		{
-			if (!nego_set_routing_token(rdp->nego, settings->LoadBalanceInfo,
-			                            settings->LoadBalanceInfoLength))
-				return FALSE;
-		}
-	}
-	else
+	if (!(settings->RedirectionFlags & LB_LOAD_BALANCE_INFO))
```
akallabeth, line 621: better us a `(flags & flag) == 0` syntax, `!` works from the standard, but I think it is more readable that way.

akallabeth (approved):
Overall LGTM,
one nitpicking remark

libfreerdp/core/connection.c
akallabeth: ok, rebased your pr and changed the line.

akallabeth:
@freerdp-bot test this please

akallabeth:
@jotolo2 thank you for your contribution :)
nice to see these parts of freerdp finally getting enough attention!

PR #8760, opened 2023-03-08
Support for RDS AAD Security

fifthdegree:
Implements the new Azure AD authentication mechanism


akallabeth:
@freerdp-bot test

include/freerdp/crypto/crypto.h
```diff
@@ -42,6 +42,10 @@ extern "C"
 	FREERDP_API void crypto_base64_decode(const char* enc_data, size_t length, BYTE** dec_data,
 	                                      size_t* res_length);
 
+	FREERDP_API char* crypto_base64url_encode(const BYTE* data, size_t length);
+	FREERDP_API void crypto_base64url_decode(const char* enc_data, size_t length, BYTE** dec_data,
```
akallabeth, line 46: I´d keep the old return `char*` so you can check for failure (e.g. `NULL` return), just add the return size argument

libfreerdp/core/transport.c
```diff
@@ -839,39 +884,61 @@ static int transport_default_read_pdu(rdpTransport* transport, wStream* s)
...
 	{
+		BYTE c;
 		int rc;
-		if (!Stream_EnsureRemainingCapacity(s, 1))
-			return -1;
-		rc = transport_read_layer_bytes(transport, s, 1);
-		if (rc != 1)
-			return rc;
-		status = transport_parse_pdu(transport, s, &incomplete);
+		while (1)
```
akallabeth, line 893: blocking loops are very problematic here as the user might cancel the connection, so check the `abortEvent` for being signalled

akallabeth (commented):
Wow, that looks really nice!
Got only minor remarks, so good work :)

I have a question though, is it necessary to have our own `JSON` parser or could we reuse existing parser libraries?
(with more and more dependencies it is hard to keep them all up to date, so that might be something to take into consideration)

libfreerdp/core/transport.c
```diff
@@ -839,39 +884,61 @@ static int transport_default_read_pdu(rdpTransport* transport, wStream* s)
 	WINPR_ASSERT(transport);
 	WINPR_ASSERT(s);
 
-	/* Read in pdu length */
-	status = transport_parse_pdu(transport, s, &incomplete);
-	while ((status == 0) && incomplete)
+	/* RDS AAD Auth PDUs have no length indicator. We need to determine the end of the PDU by
```
hardening, line 887: General comment: I think all the stuff around `transport_parse_pdu` needs some rework as I stated in the RDSTLS PR (#8758) about the double treatment (determining the size that already does some parsing, and then doing the effective parsing and partially redo the same job)

libfreerdp/crypto/base64.c
```diff
@@ -24,8 +24,9 @@
 #include <freerdp/crypto/crypto.h>
 
 static const char base64[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+static const char base64url[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
```
hardening, line 27: I'm fine with the changes could you add some tests in `libfreerdp/crypto/test/TestBase64.c` for these new functions ?

hardening:
I'd tend to agree with @akallabeth about the JSON parser, he suggested https://github.com/DaveGamble/cJSON that seems to be packaged under ubuntu as `libcjson-dev`. I'm very sorry about that because I imagine that writing that minimal JSON parser took half of the time of this PR...

hardening:
Oh and @fifthdegree thanks for the contribution, as usual you're our hero ;)

client/common/client.c
```diff
@@ -934,6 +937,35 @@ BOOL client_cli_present_gateway_message(freerdp* instance, UINT32 type, BOOL isD
 	return TRUE;
 }
 
+BOOL client_cli_get_aad_auth_code(const char* hostname, char** code)
```
akallabeth, line 940: Might be a good idea to add `rdpContext` as argument so you can get/set data for the session in the callback

akallabeth:
@fifthdegree might be a good idea to write down the steps to test this somewhere (`docs/xxx`?)

include/freerdp/crypto/crypto.h
akallabeth: ok, sorry, did mix up the lines, looks like a case of `ENOTENOUGHCOFFEE`

libfreerdp/core/transport.c
fifthdegree: It isn't blocking per se, it'll only read what's available. The only time it would be an issue is if the server sends a really really (really) long packet that takes time to consume

fifthdegree:
Closing in favor of #8768. (I'm thinking I might fully build out the JSON code I wrote and turn it into a library)

PR #8767, opened 2023-03-08
[core,transport] check for rdstls  == NULL

akallabeth:


jotolo2:
Hi!
I've forked this branch and added two commits on top of it fixing some bits on parsing rdstls pdu and checking rdstls == NULL on accept funct where it is also created a rdstls object.

My PR was merged so fast that I couldn't share the little fixes I found.

Here is my branch: https://github.com/jotolo2/FreeRDP/tree/transport_null_check
I'm not sure if you prefer to add them here or I create a new PR.

akallabeth:
@jotolo2 added your commits to this pr and also cleaned up the parser function.

mfleisz (approved):

PR #8772, opened 2023-03-09
core: Fix sending incorrect GUID in RDSTLS auth request

mfleisz:
The spec states that the GUID must be sent as a Base64-encoded GUID in Unicode format. However in the redirection code we read the (correctly formatted) GUID and convert it to a binary BLOB.

This PR removes the unnecessary conversion which now results in a correct RDSTLS auth request.

It also removes some dead code in `rdstls_write_data`.


akallabeth (approved):
LGTM

PR #8775, opened 2023-03-09
Fix rdstls write data

jotolo2:
It is possible to send an empty `REDIRECTION_GUID`, this PR allows it.
Also, the configured password was got wrongly as a pointer so it wasn't used to authenticate.

### DIFF
--- a/libfreerdp/core/rdstls.c
+++ b/libfreerdp/core/rdstls.c
@@ -456,7 +456,7 @@ static BOOL rdstls_process_authentication_request_with_password(rdpRdstls* rdstl
 	    freerdp_settings_get_uint32(settings, FreeRDP_RedirectionGuidLength);
 	serverUsername = freerdp_settings_get_string(settings, FreeRDP_Username);
 	serverDomain = freerdp_settings_get_string(settings, FreeRDP_Domain);
-	serverPassword = freerdp_settings_get_pointer(settings, FreeRDP_Password);
+	serverPassword = freerdp_settings_get_string(settings, FreeRDP_Password);
 
 	rdstls->resultCode = ERROR_SUCCESS;
 
--- a/libfreerdp/core/rdstls.c
+++ b/libfreerdp/core/rdstls.c
@@ -224,11 +224,19 @@ static SSIZE_T rdstls_write_string(wStream* s, const char* str)
 
 static BOOL rdstls_write_data(wStream* s, UINT32 length, const BYTE* data)
 {
-	if (!data)
-		return FALSE;
-
 	if (!Stream_EnsureRemainingCapacity(s, 2))
 		return FALSE;
+
+	if (!data)
+	{
+		/* Write unicode null */
+		Stream_Write_UINT16(s, 2);
+		if (!Stream_EnsureRemainingCapacity(s, 2))
+			return FALSE;
+
+		Stream_Write_UINT16(s, 0);
+		return TRUE;
+	}
 
 	Stream_Write_UINT16(s, length);
 


libfreerdp/core/rdstls.c
```diff
@@ -224,12 +224,20 @@ static SSIZE_T rdstls_write_string(wStream* s, const char* str)
...
-	if (!data)
-		return FALSE;
-
 	if (!Stream_EnsureRemainingCapacity(s, 2))
 		return FALSE;
 
+	if (!data)
+	{
+		/* Write unicode null */
+		Stream_Write_UINT16(s, 2);
```
akallabeth, line 233: why this special treatment? shouldn´t a `length = 0` empty string also work?

libfreerdp/core/rdstls.c
```diff
@@ -448,7 +456,7 @@ static BOOL rdstls_process_authentication_request_with_password(rdpRdstls* rdstl
 	    freerdp_settings_get_uint32(settings, FreeRDP_RedirectionGuidLength);
 	serverUsername = freerdp_settings_get_string(settings, FreeRDP_Username);
 	serverDomain = freerdp_settings_get_string(settings, FreeRDP_Domain);
-	serverPassword = freerdp_settings_get_pointer(settings, FreeRDP_Password);
+	serverPassword = freerdp_settings_get_string(settings, FreeRDP_Password);
```
akallabeth, line 459: good catch!

libfreerdp/core/rdstls.c
```diff
@@ -224,12 +224,20 @@ static SSIZE_T rdstls_write_string(wStream* s, const char* str)
 
 static BOOL rdstls_write_data(wStream* s, UINT32 length, const BYTE* data)
 {
-	if (!data)
```
akallabeth, line 227: good catch, should have been `if (!data && (length != 0))`

libfreerdp/core/rdstls.c
mfleisz: I think this is definitely wrong. If an empty string shall be written use `rdstls_write_string`.
If empty data shall be written with `rdstls_write_data` then we should just write the 2-byte length specifier and that's it. Binary data is not necessarily terminated with a unicode NULL.

jotolo2: I've set the unicode NULL on writing/reading data because when reading these fields they aren't directly BLOB, they are BLOB encoded in Unicode base 64. 

I think the way RedirectionGUID and password are treated should be changed.

Some info I discovered testing (helpful due to lack on documentation):
- If the mstsc client gets an empty RedirectionGUID on the Server Redirection PDU it will send on RDSTLS a NULL Unicode, this is also true on Username and Domain.
- The windows RDP server sends the encrypted password (`LB_PASSWORD_IS_ENCRYPTED`) on the Server Redirection PDU as a BLOB encoded in Unicode base 64.
- If the mstsc client gets directly a password BLOB from the Server Redirection PDU without being encoded in Unicode base 64, it will send the RDSTLS password using that BLOB and adding at the and a NULL unicode (this would fail authentication).

I think RedirectionGUID and password (when encrypted) should be, on read, parsed to get their plain BLOB value and, on write, encoded to base 64 Unicode. This should also affect redirection funcs.

If you agree I would proceed with those changes in this PR.

akallabeth: @jotolo2 I think that is not a good idea. we´ve removed the parsing of GUID and passwords as we do not need to know their contents (we do nothing with it locally in contrast to the redirection certificate, which we parse)
we actually also do not care if the data is `Unicode base 64` data as we do not need to touch it anyway. (and the specs are (as often) a bit vague on the details of that string, so better not touch it

akallabeth: @jotolo2 also, in your explanation you mean `length == 0` with `NULL` or the `length == 2` empty string case? there is a difference between `NULL` and `\0` and I´m not entirely sure I got all that right from your explanation.

jotolo2: Okey, not touching these two fields.

> If the mstsc client gets an empty RedirectionGUID on the Server Redirection PDU it will send on RDSTLS a NULL Unicode, this is also true on Username and Domain

Sorry, I meant the client doesn't get any RedirectionGUID (no `LB_REDIRECTION_GUID` flag) and sends a length == 2 empty string (02 00 00 00).


Then in the case where no RedirectionGUID has been received from Server Redirection PDU we could send:
- `length == 0 NULL`
or 
- `length == 2 empty Unicode`

Because it is expected to be in `Unicode base 64` and mstsc client does that way I would vote for length == 2 empty Unicode. But I think there would be no problem if it's sent just a length == 0 NULL. 

The read_data function should still consider the length == 2 as an empty value to keep compatibility with mstsc client.

akallabeth: @jotolo2 I´d keep logic mostly away from low level read/write routines. `length == 2` is only one special case, what if a client sends a short `GUID` or a longer one?

jotolo2: > I´d keep logic mostly away from low level read/write routines

Makes sense.

We could check after reading if values are valid (if they are not an empty Unicode).

> what if a client sends a short GUID or a longer one?

From documentation, GUID must be 16 bytes of length, that transformed to base 64 Unicode NULL terminated makes it of a fixed length of 50 bytes. This is something that could be checked on sending Server Redirection PDU
and receiving Server Redirection PDU. 

If mstsc client receives on Server Redirection PDU a redirectionGUID of a different size it sends it anyway.
In this case it's not following the documentation though the functionality keeps the same.

akallabeth: so they do the smart thing and let the client out of the loop regarding what a `GUID` actually is and let the `RDP` server decide what it is. we should do the same.
So only check the `GUID` where it matters (e.g. the server side implementation that actually sends/receives this and needs to check it) and not in the client side code here.

jotolo2: So you are saying that when sending a Server Redirection PDU we must check that the fields needed by `RDSTLS` will be valid (`RedirectionGUID`, `Username`, `Domain` and `Password`, and also `TargetCertificate`). 

Valid meaning not string empty nor NULL so on `RDSTLS` we won't need to check if those values are valid. 

From my testings only the `Password` and `TargetCertificate` were completely required to make `RDSTLS` work. So that's why was trying to make these other fields the possibility of being null. 

Do I forget about this null thing then?

akallabeth: well, you have the `RedirectionFlags`, so when the server tells the value is there don´t question it or do you have a reason to mistrust the server sending you the data?

akallabeth: also, for your issue with `mstsc` appending a `\0` does that fail the authentication against windows RDP or your `g-r-d` implementation?

akallabeth: anyway, I´d opt for not touching the data client side and passing it on as is.
your observation with the appended `\0` does not contradict that, as `mstsc` also changes behavior (e.g. bugs fixed)
it also more or less just does not matter.
if the contents is string (`WCHAR`) data the server will only evaluate until the first `\0` is reached anyway.
yes, server side you can not count on the data passed back being unmodified (as in length matches) but that should be easy enough to handle.

akallabeth:
pushed the commits to #8780 except the disputed `rdstls_write_data` one. closing this.

PR #8781, opened 2023-03-10
fix and improve cJSON detection

awakecoding:
With the latest commit, CMake generation fails if cJSON is missing (it uses get_filename_component on empty variables). I refactored the detection code to be more robust and less verbose as it tries to detect cJSON using both CMake export scripts and manual detection that does not rely on CMake export scripts or pkg-config (that's actually what I'm using myself in my conan builds)


awakecoding:
@fifthdegree FYI

libfreerdp/CMakeLists.txt
```diff
@@ -82,37 +82,30 @@ macro (freerdp_definition_add)
...
-
-	find_library(CJSON_LIBRARIES
-		NAMES cjson
-	)
-
-	if (CJSON_INCLUDE_DIRS AND CJSON_LIBRARIES)
-		set(CJSON_FOUND ON)
-	endif()
+    find_path(CJSON_INCLUDE_DIR
+        NAMES csjon/cJSON.h
```
akallabeth, line 90: this path might not be detected :)

akallabeth (changes requested):
nice one, except the little typo

awakecoding:
@akallabeth good catch - I ended up rewriting it to make it a recommended feature that gets turned on automatically if detected, and I used WITH_CJSON in the code rather than CJSON_FOUND (allows building without it even if it's detected)

akallabeth (approved):
LGTM

Issue #8774, opened 2023-03-09
Failed to connect with RDP security (breaking change)


Dr-Hex:
**Describe the bug**
FreeRDP failed to connect to local machine using RDP protocol, however the previous build (07/20/22) is connecting nomally.

**To Reproduce**
1. Install fresh Windows 10 x64 on server
2. Use default RDP settings, or use the one from this screenshot:
![изображение](https://user-images.githubusercontent.com/94066056/224077331-1994a71c-e5cd-4ec6-8257-56472f9176b4.png)
3. Run the latest FreeRDP client on OS Windows 10 x64 with the command line provided in *Logs* section below.
4. The error "Failed to connect with RDP security" is appear in console log.

**Expected behavior**
1. NLA authorization window is appearing:
![изображение](https://user-images.githubusercontent.com/94066056/224078263-ad480ac9-bc27-4aea-9a32-c2b3e30cb105.png)
2. After pressing the "Cancel" button, the Windows Logon Screen is appearing.

**Application details**
* FreeRDP version: 3.0.0-dev (093bf7983)
* Command line used: see logs section.
* Output of `xfreerdp /buildconfig`: see logs section.
* OS version connecting to (server side): Windows 10 x64 Enterprise
* If you built it yourself ...: used official pre-built binaries

**Environment (please complete the following information):**
 - OS: Windows
 - Version/Distribution: Windows 10 x64 Enterprise
 - Architecture: x64

**Additional context:**
 - If I include "username" & "password" ( /u: /p: ) fields in command line, the connection is success.

**Logs**

Below 3 logs has been provided:
 - (**latest build** v3.0.0-dev (093bf7983) ) with RDP + TLS + NLA protocols enabled (connection **failed**)
 - (**latest build** v3.0.0-dev (093bf7983) ) with RDP protocols enabled only (connection **failed**)
 - (**old build** v3.0.0-dev (e3cfe7300) ) with RDP + TLS + NLA protocols enabled (connection **success**)
All tests done at the same time with the same server and its settings.

1) (**latest build** v3.0.0-dev (093bf7983) )
With Command line:
```
set IP=192.168.60.129:3389
> "dump.txt" 2>&1 wfreerdp.exe /v:%IP% /size:4000x1000 /cert:ignore /bpp:32 -wallpaper -themes -clipboard +async-update /log-level:TRACE
```
Connection failed:
```
[...]

PR #8782, opened 2023-03-11
Auth callback fix

akallabeth:
* reported in #8774 

1. With recent library changes and default callbacks for authentication it is required that we prefer the legacy callbacks if set. 
2. Improve logging with `freerdp_set_last_error` : always log the location it was called instead of the function itself
3. some cleanups with `SelectedProtocol` in `nego`


akallabeth:
@mfleisz this is a nasty one, can you have a look?
the `Authenticate` and `AuthenticateEx` callbacks return `FALSE` if the authentication dialog was aborted, but for `RDP` and `TLS` security this just means we don´t have credentials (as there is no `skip` option) and the session should not abort.

mfleisz (approved):

PR #7306, opened 2021-09-20
early bail from update_read_window_state_order breaks protocol

byteboon:
When handling WINDOW_ORDER_FIELD_WND_RECTS in update_read_window_state_order, the early bail breaks subsequent protocol parsing.

Log looks like this:
Alternate Secondary Drawing Order [0x0b] Windowing
WindowUpdate windowId=0x66190 [ex]style=<0x86080044, 0x10000 popup> clientOffset=(-5000,-5000) windowOffset=(-5000,-5000) windowClientDelta=(0,0) windowWidth=0 windowHeight=0 windowRects=() visibleOffset=(0,0) visibilityRects=()
pf_client_window_update
Alternate Secondary Drawing Order [0x1e] UNKNOWN
[0x1e] UNKNOWN - Alternate Secondary Drawing Order UNKNOWN
[0x1e] UNKNOWN - SERVER BUG: The support for this feature was not announced!
order flags 78 failed
Fastpath update Orders [0] failed, status 0
fastpath_recv_update() - -1
fastpath_recv_update_data() fail
pf_client_end_paint

PR #8798, opened 2023-03-14
Update read window state order

akallabeth:
Backport #7306

### DIFF
--- a/libfreerdp/core/window.c
+++ b/libfreerdp/core/window.c
@@ -359,33 +359,31 @@ static BOOL update_read_window_state_order(wStream* s, WINDOW_ORDER_INFO* orderI
 
 		Stream_Read_UINT16(s, windowState->numWindowRects); /* numWindowRects (2 bytes) */
 
-		if (windowState->numWindowRects == 0)
+		if (windowState->numWindowRects > 0)
 		{
-			return TRUE;
-		}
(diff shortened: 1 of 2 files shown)


libfreerdp/core/window.c
```diff
@@ -359,33 +359,31 @@ static BOOL update_read_window_state_order(wStream* s, WINDOW_ORDER_INFO* orderI
...
+				windowState->windowRects = NULL;
+				return FALSE;
+			}
 
-		windowState->windowRects = newRect;
+			windowState->windowRects = newRect;
 
-		if (Stream_GetRemainingLength(s) < 8 * windowState->numWindowRects)
-			return FALSE;
+			if (Stream_GetRemainingLength(s) < 8 * windowState->numWindowRects)
```
jayaddison, line 376: Sorry, nitpicking on already-merged code, but I wonder whether:

```suggestion
			if (Stream_GetRemainingLength(s) < * windowState->numWindowRects * 8)
```

...could be more consistent with a [similar line](https://github.com/akallabeth/FreeRDP/blob/c4869357dee79cac856738066d2fa1312909d246/libfreerdp/core/window.c#L420) later in this file: 

```c
			if (Stream_GetRemainingLength(s) < windowState->numVisibilityRects * 8)
```

(and potentially safer, in the sense of looking less like a bitshift, and with clearer operator precedence)

PR #8800, opened 2023-03-15
[utils] fix windows console read

akallabeth:
On windows freerdp_interruptible_getc did not work at all. Fix this so that reading characters works. This still does not hide input for passwords or is actually interruptible, but at least it reads the data correctly.


akallabeth:

hardening (approved):
LGTM

PR #8801, opened 2023-03-15
use cmake object libraries for built-in virtual channel plugins

awakecoding:
Built-in virtual channel plugins have always been built as individual static libraries that freerdp-client links to, but recent changes [introduced a cyclic dependency](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#cyclic-dependencies-of-static-libraries) where some virtual channel plugins would also need to link to freerdp-client. Linkers don't behave consistently with cyclic dependencies, and clang from the Android NDK is stricter than on other platforms. Since Android CI builds use shared libraries, the issue is avoided, but we make a single shared library with everything statically linked into it, so we've hit the issue.

As a permanent fix for the problem, I've just switched to CMake object libraries, which is the recommended solution that avoids having to fix the cyclic dependency in the code. Back when we first built this part of the CMake build system, object libraries required a version of CMake we could not assume was available, but that was so long ago that this is no longer an issue.


akallabeth (approved):
LGTM

PR #8805, opened 2023-03-16
packaging/rpm: Enable FFmpeg support on Fedora 36+ and RHEL 9+

Conan-Kudo:
Since Fedora Linux 36, an FFmpeg implementation is provided in the base distribution as "ffmpeg-free". This implementation is sufficient for building FreeRDP with FFmpeg support.

This implementation is also available in Fedora Extra Packages for Enterprise Linux (EPEL) for Red Hat Enterprise Linux 9 and derivatives.

Thus, enable FFmpeg support when building for those distributions.

### DIFF
--- a/packaging/rpm/freerdp-nightly.spec
+++ b/packaging/rpm/freerdp-nightly.spec
@@ -89,6 +89,10 @@ BuildRequires: wayland-devel
 BuildRequires: libwayland-client-devel
 %endif
 
+%if 0%{?fedora} >= 36 || 0%{?rhel} >= 9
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@hardening it looks like your Jenkins configuration doesn't test the packaging at all?

Conan-Kudo:
This should be okay to backport to `stable-2.0` as well, what's the process for that?

akallabeth:
@Conan-Kudo create a pr against `stable-2.0` with `git cherry-pick -x <commin>` and assign me as reviewer

PR #8817, opened 2023-03-20
[codecs] make progressive honor threading flags

hardening:
Progressive codec was not honoring the threading flags in settings like does remoteFX, so even when no multiple threads decoding was asked, progressive was using multiple threads anyway. This patch fixes it.


akallabeth (approved):
LGTM

I´d remove the old function altogether as it is not very likely that it was used previously outside the project. (if someone complains we can add it back anyway)

PR #8819, opened 2023-03-22
[clients] fix clipboard locking

akallabeth:
In #8719 we found a few locking issues, turned out these were also in wayland and X11 clients.


pnowack (approved):
The message title of the commit message is ambigiuous. It is not clear from the commit message, whether locking is meant with respect to CLIPRDR, or whether it is regarding the mutex of a `wClipboard` instance.

akallabeth:
@pnowack good point, changed `clipboard` to `wClipboard` in the commit message.

PR #8824, opened 2023-03-22
[android] fixed build and runtime issues

akallabeth:
* fixed a return value bug in event handling
* fixed some android manifest issues with newer API
* fixed version parser in java to properly load libraries


akallabeth:

mfleisz (approved):

Issue #8816, opened 2023-03-20
Bug in progressive decoding


hardening:
I'm seeing that valgrind complains when trying to replay progressive traffic, seems like there's a bug when reallocating internal cache items:

```
==372600== Invalid read of size 2
==372600==    at 0x4B936E8: progressive_process_tiles_tile_work_callback (progressive.c:1657)
==372600==    by 0x4B945A2: progressive_process_tiles (progressive.c:1811)
==372600==    by 0x4B9936C: progressive_wb_region (progressive.c:2395)
==372600==    by 0x4B99B50: progressive_decompress (progressive.c:2484)
==372600==    by 0x4A1131B: gdi_SurfaceCommand_Progressive (gfx.c:950)
==372600==    by 0x4A11740: gdi_SurfaceCommand (gfx.c:1040)
==372600==    by 0x516B8EB: rdpgfx_recv_wire_to_surface_2_pdu (rdpgfx_main.c:1395)
==372600==    by 0x51717A6: rdpgfx_recv_pdu (rdpgfx_main.c:1891)
==372600==    by 0x5172672: rdpgfx_on_data_received (rdpgfx_main.c:2095)
==372600==    by 0x20FE8B: CaptureEgfx::inject(bool, wStream*) (egfx.cpp:249)
==372600==    by 0x23BF72: handleContent(RingBuffer*, PerChannelData*, ScreenshooterConfig const&) (screenshooter-main.cpp:381)
==372600==    by 0x23C5E9: handleReplay(ScreenshooterConfig const&) (screenshooter-main.cpp:484)
==372600==  Address 0x114883e0 is 13,440 bytes inside a block of size 347,760 free'd
==372600==    at 0x484DCD3: realloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==372600==    by 0x4B8C81F: progressive_allocate_tile_cache (progressive.c:461)
==372600==    by 0x4B8CE7E: progressive_surface_tile_replace (progressive.c:583)
==372600==    by 0x4B91FF4: progressive_tile_read_upgrade (progressive.c:1576)
==372600==    by 0x4B94022: progressive_process_tiles (progressive.c:1734)
==372600==    by 0x4B9936C: progressive_wb_region (progressive.c:2395)
==372600==    by 0x4B99B50: progressive_decompress (progressive.c:2484)
==372600==    by 0x4A1131B: gdi_SurfaceCommand_Progressive (gfx.c:950)
[...]

PR #8826, opened 2023-03-23
[codec] fix tile usage in progressive

hardening:
This is a fix for #8816. 

When PROGRESSIVE_SURFACE_CONTEXT.tiles were reallocated, we were ending up with wrong tiles in PROGRESSIVE_BLOCK_REGION.tiles when the memory block was moved.

### DIFF
--- a/libfreerdp/codec/progressive.c
+++ b/libfreerdp/codec/progressive.c
@@ -382,6 +382,7 @@ static void progressive_tile_free(RFX_PROGRESSIVE_TILE* tile)
 		winpr_aligned_free(tile->sign);
 		winpr_aligned_free(tile->current);
 		winpr_aligned_free(tile->data);
+		free(tile);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


libfreerdp/codec/progressive.c
```diff
@@ -460,15 +461,25 @@ static BOOL progressive_allocate_tile_cache(PROGRESSIVE_SURFACE_CONTEXT* surface
...
 		if (!tmp)
 			return FALSE;
 		surface->tilesSize = surface->gridSize;
 		surface->tiles = tmp;
 
+		BOOL allocFailed = FALSE;
 		for (x = oldIndex; x < surface->tilesSize; x++)
-			surface->tiles[x] = empty;
+		{
+			surface->tiles[x] = malloc(sizeof(RFX_PROGRESSIVE_TILE));
```
akallabeth, line 473: Maybe use `calloc`? or even better, initialize with`NULL` and let the allocation be done by `progressive_tile_allocate`?

akallabeth (approved):
LGTM

PR #8831, opened 2023-03-24
[client,windows]fix: keyboard focus error when minimize window by click taskbar icon

407219125:
[windows client]When click the taskbar icon to minimize the window，then keyboard focus still in wfreerdp.exe. So use wParam‘s
high word to judge whether the window is minimized. If window is minimized, then set `g_focus_hWnd = NULL`.


akallabeth (approved):
LGTM,

akallabeth:
@freerdp-bot test please

PR #8833, opened 2023-03-24
[codec,progressive] simplify tile cache allocation

akallabeth:
I just don´t like having `_new` and `_free` functions and doing allocations outside of these.


libfreerdp/codec/progressive.c
```diff
@@ -456,34 +451,29 @@ static BOOL progressive_allocate_tile_cache(PROGRESSIVE_SURFACE_CONTEXT* surface
 	if (surface->tiles)
 	{
 		oldIndex = surface->gridSize;
-		surface->gridSize *= 2;
+		surface->gridSize += 1024;
 	}
```
hardening, line 455: Yeah I like that, doubling the size was too much.

hardening (approved):
LGTM

PR #8840, opened 2023-03-28
[cli, gateway] replace gateway type delimiter from "," to ":"

akarl10:
the new gateway parameter delimits its sub-parameters with , so it is not possible to use this character for sub-sub parameters

example: /gateway:type:http,no-websockets,g:gateway

with the patch this would change to

/gateway:type:http:no-websockets,g:gateway


akallabeth (changes requested):
Good catch, but wrong fix.

should be caught outside this function as top level option (e.g. you can keep the `,` as delimiter)

PR #8841, opened 2023-03-28
[client,cmdline] fix gateway option parser

akallabeth:
Options like /gateway:type:http,no-websockets,port:1234 were not properly parsed. The `type:http,no-websockets` must be considered as separate options with weak dependency.
Thanks to @akarl10 for pointing out this problem.

Replaces #8840


hardening (approved):
lgtm

PR #8849, opened 2023-04-02
Replace __FUNCTION__ with __func__

JohannesKauffmann:
Use `__func__` instead of `__FUNCTION__`, which is a GCC extension.

The `__func__` identifier is standard since C99, however Visual Studio supports it since VS2015.

### DIFF
--- a/winpr/include/winpr/assert.h
+++ b/winpr/include/winpr/assert.h
@@ -35,7 +35,7 @@
 		{                                                                                     \
 			wLog* _log_cached_ptr = WLog_Get("com.freerdp.winpr.assert");                     \
 			WLog_Print(_log_cached_ptr, WLOG_FATAL, "%s [%s:%s:%" PRIuz "]", #cond, __FILE__, \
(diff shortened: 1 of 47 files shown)


hardening (approved):

hardening:

JohannesKauffmann:
Ah, it looks like VS2013 is still a thing huh. That's unfortunate.

Main reason for this PR is that using __FUNCTION__ in public headers generates warnings when using these headers in other projects with -Wpedantic.

What would you suggest moving forward? Is VS2013 support a hard requirement? Would using an extra preprocessor define, to make an exception for VS2013, be an acceptable solution?

akallabeth:
@JohannesKauffmann thank you for the pull request, but we´ll have to update the ci first. might take some time until done.
Perfectly fine with the change (these annoying warnings I´d like to get rid of too)

akallabeth:
@JohannesKauffmann I´ve done a new pr now that our ci has been updated. Thank you for your try anyway, appreciated!

PR #8855, opened 2023-04-04
[core,credssp] fix logging with unicode

akallabeth:
convert argument to proper type before logging

### DIFF
--- a/libfreerdp/core/credssp_auth.c
+++ b/libfreerdp/core/credssp_auth.c
@@ -111,6 +111,13 @@ BOOL credssp_auth_init(rdpCredsspAuth* auth, TCHAR* pkg_name, SecPkgContext_Bind
 	const rdpSettings* settings = auth->rdp_ctx->settings;
 	WINPR_ASSERT(settings);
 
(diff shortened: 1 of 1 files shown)


libfreerdp/core/credssp_auth.c
```diff
@@ -111,6 +111,13 @@ BOOL credssp_auth_init(rdpCredsspAuth* auth, TCHAR* pkg_name, SecPkgContext_Bind
 	const rdpSettings* settings = auth->rdp_ctx->settings;
 	WINPR_ASSERT(settings);
 
+	char name[64] = { 0 };
+#if defined(UNICODE)
+	ConvertWCharToUtf8(pkg_name, name, ARRAYSIZE(name));
+#else
+	strncpy(name, pkg_name, ARRAYSIZE(name));
```
mfleisz, line 118: This might produce a non-null terminated string for security packages with long names (>= 64)

akallabeth: @mfleisz good catch, fixed.

mfleisz (approved):

PR #8862, opened 2023-04-06
Sspi and log

akallabeth:


pnowack (approved):
The log commit looks good to me. The SSPI commit too (password would just contain the null terminator in empty password scenarios). But I have to say, that I don't know much about the SSPI handling.

PR #8867, opened 2023-04-06
[core,redirection] Fix sending redirection certificate

joantolo:
This is a little thing missing in the `Server Redirection`.

This PR is based on the previous TODO comment it changes.

Another and simpler way would be to just set the certificate in `DER` format, save it as `rdpCertificate` and only build the `TARGET_CERTIFICATE_CONTAINER` when sending.

### DIFF
--- a/include/freerdp/redirection.h
+++ b/include/freerdp/redirection.h
@@ -43,6 +43,9 @@
 
 #define LB_PASSWORD_MAX_LENGTH 512
 
+#define ELEMENT_TYPE_CERTIFICATE 32
+#define ENCODING_TYPE_ASN1_DER 1
+
 #ifdef __cplusplus
 extern "C"
 {
--- a/libfreerdp/core/redirection.c
+++ b/libfreerdp/core/redirection.c
@@ -56,9 +56,6 @@ struct rdp_redirection
 	rdpCertificate* TargetCertificate;
 };
 
-#define ELEMENT_TYPE_CERTIFICATE 32
-#define ENCODING_TYPE_ASN1_DER 1
-
 static void redirection_free_array(char*** what, UINT32* count)
 {
 	WINPR_ASSERT(what);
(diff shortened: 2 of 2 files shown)


libfreerdp/core/redirection.c
```diff
@@ -413,16 +410,15 @@ static BOOL rdp_redirection_write_target_cert(wStream* s, const rdpRedirection*
...
 	if (!serialized)
 		goto fail;
 
 	if (!rdp_redirection_write_target_cert(serialized, redirection))
 		goto fail;
 
-	if (!rdp_redirection_write_base64_wchar(
-	        LB_TARGET_CERTIFICATE, s, Stream_GetPosition(serialized), Stream_Buffer(serialized)))
-		return FALSE;
+	rc = rdp_redirection_write_base64_wchar(
```
hardening, line 420: `rdp_redirection_write_base64_wchar` should take a `wStream` as argument that would allow check for `serialized`'s free space (otherwise how can we ensure that 2048 is enough ?)

akallabeth:
@freerdp-bot test

libfreerdp/core/redirection.c
akallabeth: @hardening `rdp_redirection_write_target_cert` writes to `serialized`, `rdp_redierction_write_base64_wchar` writes to `s` (and takes the data from `serialized`)

akallabeth:

akallabeth (approved):
LGTM
you´re missing a `git clang-format` on your changes though.

joantolo:
Rebased and done!

akallabeth:
@freerdp-bot test

include/freerdp/redirection.h
```diff
@@ -43,6 +43,9 @@
 
 #define LB_PASSWORD_MAX_LENGTH 512
 
+#define ELEMENT_TYPE_CERTIFICATE 32
```
akallabeth, line 46: Curious, why did you make these public definitions?

joantolo: Because when the certificate is set with `redirection_set_byte_option` the certificate is in `TARGET_CERTIFICATE_CONTAINER` format, which uses these two values, and to store it, these values are checked in `rdp_redirection_read_target_cert`.

So making these values public helps build the certificate externally in `TARGET_CERTIFICATE_CONTAINER` format and keep compatibility.

If we set the certificate in plain DER format, this would be simpler and these values wouldn't need to be public.

PR #8868, opened 2023-04-07
Unique kerberos ccaches for parallel connections.

ggirol-rc:
The "MEMORY" ccache is shared in a process.
If a client uses it to make parallel connections,
the same ccache may be used for several clients with distinct credentials.
To prevent such sharing we create a unique, dedicated ccache when necessary with krb5_cc_new_unique.

To destroy the ccaches we created, the struct KRB_CREDENTIALS is extended to express the ccache ownership.

### DIFF
--- a/winpr/libwinpr/sspi/Kerberos/kerberos.c
+++ b/winpr/libwinpr/sspi/Kerberos/kerberos.c
@@ -39,6 +39,7 @@
 #include <winpr/endian.h>
 #include <winpr/crypto.h>
 #include <winpr/path.h>
+#include <winpr/wtypes.h>
 
 #include "kerberos.h"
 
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -28,6 +28,7 @@
 #include <fcntl.h>
 #include <string.h>
 #include <ctype.h>
+#include <krb5/krb5.h>
```
akallabeth, line 31: This include needs to be guarded for builds without working kerberos (looking at you apple)

akallabeth (changes requested):

winpr/libwinpr/sspi/Kerberos/kerberos.c
ggirol-rc: I removed the include altogether as it seems to be brought by the later `#include "krb5glue.h"`

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #8869, opened 2023-04-07
don't log an error when running GetCurrentThread from non winpr thread

ggirol-rc:
Since https://github.com/FreeRDP/FreeRDP/commit/7c4a774e4eb9dfdf9271b6bbef2bb1b5191db398 this can legitimately happen in WaitForMultipleObjectsEx and is not actually an error.


akallabeth:
@freerdp-bot test please

akallabeth (approved):
LGTM

PR #8866, opened 2023-04-06
[codecs] make h264 and planar opaque

akallabeth:

PR #8870, opened 2023-04-07
fix warnings in static inline functions in installed headers

ggirol-rc:
= { 0 } has different meaning in C and C++: in C it initialises all members to 0, in C++ only the first one so the compiler warns that other members are not initialised. Use an empty initialiser list instead.

It's not very pretty; if you prefer a different style let me know :)

### DIFF
--- a/winpr/include/winpr/pool.h
+++ b/winpr/include/winpr/pool.h
@@ -249,7 +249,11 @@ extern "C"
 
 	static INLINE VOID InitializeThreadpoolEnvironment(PTP_CALLBACK_ENVIRON pcbe)
 	{
-		const TP_CALLBACK_ENVIRON empty = { 0 };
+		const TP_CALLBACK_ENVIRON empty = {
+#ifndef __cplusplus
+			0
+#endif
+		};
 		*pcbe = empty;
 		pcbe->Version = 1;
 	}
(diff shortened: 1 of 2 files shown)


hardening (approved):
Thank you for that, I've been bothered by it too.

hardening:
@freerdp-bot can you test that my friend ?

akallabeth (changes requested):
Already changed in #8866

winpr/include/winpr/pool.h
```diff
@@ -249,14 +249,19 @@ extern "C"
...
+#endif
+		};
 		*pcbe = empty;
 		pcbe->Version = 1;
 	}
 
 	static INLINE VOID DestroyThreadpoolEnvironment(PTP_CALLBACK_ENVIRON pcbe)
 	{
 		/* no actions, this may change in a future release. */
+		(void)pcbe;
```
akallabeth, line 264: better use `WINPR_UNUSED`

akallabeth:
Looking at these functions, it might be the better option to make them normal functions.
We´d run into trouble if we change something in that code as it will always be `ABI` breaking and these functions are exposed via `API` to potential 3rd party software.

PR #8876, opened 2023-04-12
core/autodetect: Rework autodetect handling + allow overriding connect-time autodetection handling

pnowack:
See the commit messages for the individual changes.

I also renamed `AUTODETECT_STATE` to `FREERDP_AUTODETECT_STATE` in the rework commit, since that name is too general (can easily conflict with API users).

For the override handling, I removed the `CONNECTION_STATE_CONNECT_TIME_AUTO_DETECT_RESPONSE` state, as it became unused.
I did not remove the `CONNECTION_STATE_CONNECT_TIME_AUTO_DETECT_REQUEST` state, since it is used in the client side and was not sure about replacing it.

Something that looked weird to me initially was that the size of all incoming PDUs is counted, but the payload size of payload PDUs is added to the byteCount counter too.
I thought, that we would count the payload PDUs here 2 times (on the client side). But after testing this does not seem to be the case.

The maximum payload size for payload PDUs is not `UINT16_MAX`, but actually `15 * 1024 + 512 + 256 + 128 + 64 + 32 + 16 + 2 + 1` (`16371`). One byte more and mstsc or MS' iOS client disconnect with a "protocol" error.
It might be related to the TLS segment size of 16KiB.

I did not remove the `rttMeasureStartTime` variable, although it is wrong, when multiple RTT requests are emitted at the same time, since that variable is more or less used for the default connect-time autodetection handling.

Tested with https://gitlab.gnome.org/pnowack/gnome-remote-desktop/-/commits/layout-freerdp3


pnowack:
Was a bit thinking about whether `RDP_AUTODETECT_STATE` is better as a less common name for `AUTODETECT_STATE` than `FREERDP_AUTODETECT_STATE`, but not sure, since `[MS-RDPBCGR]` does not mention such state names here.

akallabeth (approved):
LGTM

PR #8880, opened 2023-04-13
[winpr,assert] eliminate c++ warnings

akallabeth:
Eliminate cast warnings when used in C++ code


hardening (commented):
Looks ok

akallabeth:
@hardening an approving review would be great, then we can merge this ;)

hardening (approved):

PR #8882, opened 2023-04-13
Color depth split

akallabeth:
* Expose `SupportedColorDepths` via `rdpSettings`
* Do not insist on `32bpp` for `gfx` or `rfx` modes


pnowack (approved):
Changes look good to me. Tested by connecting with xfreerdp3 to g-r-d (compiled against FreeRDP3). Also tested with mstsc connecting to g-r-d (compiled against FreeRDP3).

When now xfreerdp3 is started with e.g. `/gfx /bpp:16`, or in mstsc the colour depth field is set to 16-bit, then `ColorDepth` will be 16, but `SupportedColorDepths` will still contain 32-bit support.
So, working as expected.

PR #8886, opened 2023-04-14
pointer: Rename position to hotspot position in Color Pointer Update

pnowack:

```
The attributes xPos and yPos for a Color Pointer Update are confusing,
as they may be confused with the xPos and yPos of the pointer bitmap on
the actual screen.
Rename these attributes to what they actually represent, and that is the
hotspot position.
xPos and yPos are still members of the hotspot. However, hotSpotX and
hotSpotY are much more clearer.
In addition to that, the Large Pointer Update uses the same names for
the hotspot coordinates.
```

### DIFF
--- a/include/freerdp/pointer.h
+++ b/include/freerdp/pointer.h
@@ -50,13 +50,13 @@ extern "C"
 
 	typedef struct
 	{
-		UINT32 cacheIndex;
-		UINT32 xPos;
-		UINT32 yPos;
-		UINT32 width;
-		UINT32 height;
-		UINT32 lengthAndMask;
-		UINT32 lengthXorMask;
+		UINT16 cacheIndex;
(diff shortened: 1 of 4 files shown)


hardening (approved):
This will break ABI, but appart from that LGTM

pnowack:
Oh, did not mention one thing in the commit message and that is that I changed some attributes in the `POINTER_COLOR_UPDATE` struct from `UINT32` to `UINT16` (like in the docs).
The Large Pointer Update does not appear to need this (looks like it is already there like in the docs).

akallabeth (approved):
LGTM

akallabeth:
since the API/ABI breakage only affects server implementations (and ogon/gnome remote desktop are the only ones I know so far apart from our shadow/sample servers) it should not be a big issue.

PR #8888, opened 2023-04-14
[winpr,wtypes] fix BOOL definition for APPLE

akallabeth:
* Add TargetConditionals.h include to ensure the symbols are defined
* Check for symbol existence and value != 0


hardening (approved):
looks ok

Issue #8846, opened 2023-04-01
client_cliprdr_file handling for remote files violates [MS-RDPECLIP] specification


pnowack:
Affected version: xfreerdp3/master

Reproduction steps:
1. Connect with xfreerdp3 to g-r-d
2. Copy some large folder with lots of files (e.g. some git repo) on the server side
3. Paste them on the client side (e.g. in nautilus)

Observation:
Only a few files inside the remote folder are copied, then the copy operation aborts due to an IO error.

When watching the log of g-r-d, it is noticeable, that xfreerdp3 tries to lock the file list with an id, which it already used before to lock a file list, but did not unlock, or xfreerdp3 tries to lock the same file list multiple times at the same time.

This violates `3.1.5.3.1 Sending a Lock Clipboard Data PDU`:

```
The clipDataId field MUST contain an unsigned integer value that will serve as an identifier to
uniquely tag any File Stream data (section 1.3.1.1.5) on the clipboard of the Shared Clipboard Owner
(section 1.3.2.1).
```

The "uniquely tag" is crucial here.

A weird part is also, that despite having a lock PDU sent, the `clipDataId` is not used in the subsequent file content requests.

The observation about those IO errors can also be observed, when disabling the locking behaviour (by adding a `&& FALSE` to the `cliprdr_file_context_remote_get_flags(stream->context) & CB_CAN_LOCK_CLIPDATA`).

These IO errors don't happen on lone files, but on larger folders (folder with 100+ files might already be enough).

akallabeth:
@pnowack ok, found an issue. there was a lock missing on the hash table when discarding streams.
can you check if that fixes your duplicate locking/unlocking issues?

pnowack:
@akallabeth I may have an idea, why the IO errors happen (not the protocol violations though, since I didn't dig that much into `client_cliprdr_file`):
Close to the end of the log, where the protocol violation happens, the `FileGroupDescriptorW` data is fetched again.
xfreerdp3 caches the data of the last request. While apparently there is no fetch for a different format of the format list, that mechanism still exposes a problem:
A refetch of `FileGroupDescriptorW` might mean, that the file list becomes invalid. And xfreerdp3 might try to replace the contents of the file list, which changes the contents of the FUSE filesystem, leading to those IO errors.

The fix might be simple, and that is by using a hash table for the cached data: Each entry in the hash table would be `formatId` -> data for the `formatId`.

---

[...]

PR #8892, opened 2023-04-15
client/X11: Fix IO errors during server to client file transfer

pnowack:

```
When receiving a file list, xfreerdp3 rebuilds the content of the FUSE
filesystem.
Since fetching uri-lists can happen during a paste action too, xfreerdp3
caches the content of the last fetched mime type.
However, uri-lists exists in different variations, e.g. nautilus uses a
different mime type, than gnome-terminal does.
Furthermore, FormatLists can also contain other formats in addition to
file lists.
Fetching those contents during a paste operation leads to IO errors in
the paste operation.

In order to fix those errors, cache every mimetype in two hash tables:
One for the raw (unconverted) data, and one for the converted data.
When a content request is received, xfreerdp3 can with the already
cached converted data, directly serve the data.
If the content is not available as cached data, but its source data was
already fetched, use that source data to create the converted data.
Then serve the converted data.
Only if no cached converted data or cached raw data is available, issue
a new FormatDataRequest.
```

This fixes the IO errors mentioned in https://github.com/FreeRDP/FreeRDP/issues/8846, but is not enough to fix the protocol violations.
In addition to that, this PR uncovers another issue in the FUSE filesystem: There are entries in the FUSE filesystem, that are duplicate (user receives replace-file dialogs for some files during pasting) or missing (certain file not copied).

### DIFF
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -72,6 +72,12 @@ typedef struct
 	char* formatName;
 } xfCliprdrFormat;
 
+typedef struct
+{
+	BYTE* data;
+	UINT32 data_length;
+} xfCachedData;
+
 struct xf_clipboard
 {
 	xfContext* xfc;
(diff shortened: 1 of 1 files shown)


akallabeth (approved):
LGTM

PR #8897, opened 2023-04-18
Add WITH_UNICODE_BUILTIN option to use built-in UTF conversion functions

awakecoding:
Add WITH_UNICODE_BUILTIN cmake option (OFF by default) to use built-in UTF conversion functions, rather than system-provided Unicode conversion libraries. It's not a big deal on most platforms, but on Linux, the libicu link dependency has caused us a few headaches and we'd rather not link to it, even if there's a chance of degraded Unicode support for our builds.

I have imported a cleaned up code snippet from Unicode.org similar to the one we were using originally - it turns out that once re-integrated, all Unicode conversion tests pass with ctest. @akallabeth said that one of the reasons for switching to libicu was because of failing edge cases, and that the unit tests had been updated to cover those, but now that everything works, maybe it's good enough?

I won't argue for changing the default build option, I'm fine with leaving it off by default, but it looks like there may not be known failing cases with the built-in UTF conversion anymore.


hardening (approved):
I have nothing against it, @akallabeth ?

cmake/ConfigOptions.cmake
```diff
@@ -164,6 +164,8 @@ option(USE_VERSION_FROM_GIT_TAG "Extract FreeRDP version from git tag." ON)
 option(WITH_CAIRO    "Use CAIRO image library for screen resizing" OFF)
 option(WITH_SWSCALE  "Use SWScale image library for screen resizing" OFF)
 
+option(WITH_UNICODE_BUILTIN  "Use built-in Unicode conversion (don't use system-provided libraries)" OFF)
```
akallabeth, line 167: This option only affects `WinPR`, so it might be better to have that somewhere in `winpr` subdirectory instead of here.

akallabeth (commented):
LGTM

PR #8898, opened 2023-04-18
[core,rdstls] Avoid Stream_SafeSeek on pdu parsing

joantolo:
Because it is being parsed the pdu it is expected that not all bytes are read yet.

Using `Stream_SafeSeek` prints a lot of warnings that shouldn't be printed in this case.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #8900, opened 2023-04-19
winpr/input: Add actual evdev keyboard handling

pnowack:

```
WinPR provides APIs to convert between keycodes between virtual
keycodes.
These keycodes can currently be evdev keycodes or Apple keycodes.
The evdev handling, however, handles XKB keycodes and not evdev ones.
The main difference between these is that XKB keycodes are shifted by
the value 8, compared to evdev keycodes.

In order to fix this situation, rename the evdev keycodes to XKB ones,
and introduce additionally a new keycode evdev, including its handling
for this keycode type.
```

This is an API breaking change.

### DIFF
--- a/winpr/libwinpr/input/keycode.c
+++ b/winpr/libwinpr/input/keycode.c
@@ -32,14 +32,6 @@
  */
 
 static DWORD KEYCODE_TO_VKCODE_APPLE[256] = {
-	0,                    /* 0 */
-	0,                    /* 1 */
-	0,                    /* 2 */
-	0,                    /* 3 */
-	0,                    /* 4 */
-	0,                    /* 5 */
-	0,                    /* 6 */
-	0,                    /* 7 */
 	VK_KEY_A,             /* APPLE_VK_ANSI_A (0x00) */
 	VK_KEY_S,             /* APPLE_VK_ANSI_S (0x01) */
 	VK_KEY_D,             /* APPLE_VK_ANSI_D (0x02) */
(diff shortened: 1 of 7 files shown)


pnowack:
Did split up a few lines to have both a consistent indentation for the comments and to make clang-format happy.

winpr/libwinpr/input/keycode.c
```diff
@@ -623,6 +902,17 @@ DWORD GetKeycodeFromVirtualKeyCode(DWORD vkcode, DWORD dwFlags)
 			}
 		}
 	}
+	else if (dwFlags & KEYCODE_TYPE_XKB)
+	{
+		for (index = 0; index < 256; index++)
```
akallabeth, line 907: Maybe use `ARRAYSIZE(KEYCODE_TO_VKCODE_XKB)` here?

pnowack: Yep, did also the same for `KEYCODE_TO_VKCODE_APPLE`, and `KEYCODE_TO_VKCODE_EVDEV`.

pnowack:
@freerdp-bot test

hardening:
Suggestion: if that breaks API/API we shall start a file to document all these changes, and give some hints of how to migrate to 3.0

pnowack:
> Suggestion: if that breaks API/API we shall start a file to document all these changes, and give some hints of how to migrate to 3.0

Indeed. Do we already have something like an empty wiki page, where markdown works (something which we members can all edit)?

akallabeth (approved):
LGTM

PR #8906, opened 2023-04-20
gateway: Do not encrypt message during RPC NTLM auth

mfleisz:
Commit 2de7a4c2498b6d2405cbf1209a58a42957698e44 introduced major changes in the gateway authentication code. One of these changes was to decouple NTLM specific authentication from the gateway code.

However with these changes, gateway authenciation with the old RPC code stopped working and returned an authentication error. The problem is that currently `credssp_auth_encrypt` encrypts the given message along creating a signature.

The old code prevented encryption of the message by specifying `SECBUFFER_READONLY` on the message buffer. The native Windows SSPI then leaves this buffer as-is and gateway authentication works again.

This fix only applies to Windows platforms using the native SSPI API. Interestingly this works on other platforms using the WinPR SSPI so there seems to be a difference between the implementations (but that's a topic for another PR).


hardening (approved):
looks ok

Issue #8909, opened 2023-04-20
Client may receive an IRP for an invalid device ID and the connection is aborted


thenextman:
Client may receive an IRP for an invalid device ID and the connection is aborted.

**Application details**
* 3.0.0-dev (current master)
* Windows Server 2k19

**Environment (please complete the following information):**
 - OS: macOS
 - Version/Distribution: 13 (Ventura)

**Additional context**
Customer reports that connecting to an existing login session works, but logging into a new session always fails. All rdprdr redirection options are _disabled_; however the client receives an IRP request for an invalid device ID. Because of [df4c076](https://github.com/devolutions/freerdp/commit/df4c076946bc8716b8bf3fb9e512374b57cca1cb), this scenario causes the connection to fail.

I expect this is an issue with the server-side configuration or environment and am unable to reproduce the issue myself. 

Debug log:

> [04:41:40:690] [30980:6d723000] [DEBUG][com.freerdp.client.common.cmdline] - [freerdp_client_load_static_channel_addin]: loading channelEx rdpdr
> [04:41:40:690] [30980:6d723000] [DEBUG][com.freerdp.client.common.cmdline] - [freerdp_client_load_static_channel_addin]: loading channelEx rdpsnd
> [04:41:40:690] [30980:6d723000] [DEBUG][com.freerdp.channels.drdynvc.client] - [drdynvc_VirtualChannelEntryEx]: VirtualChannelEntryEx
> [04:41:40:690] [30980:6d723000] [DEBUG][com.freerdp.client.common.cmdline] - [freerdp_client_load_static_channel_addin]: loading channelEx drdynvc
> [04:41:40:690] [30980:6d723000] [WARN][com.freerdp.core.client] - [freerdp_channels_client_load_ex]: Skipping, channel already loaded
> [04:41:40:690] [30980:6d723000] [DEBUG][com.freerdp.client.common.cmdline] - [freerdp_client_load_static_channel_addin]: loading channelEx rdpdr
> [04:41:40:690] [30980:6d723000] [WARN][com.freerdp.core.client] - [freerdp_channels_client_load_ex]: Skipping, channel already loaded
> [04:41:40:690] [30980:6d723000] [DEBUG][com.freerdp.client.common.cmdline] - [freerdp_client_load_static_channel_addin]: loading channelEx rdpsnd
[...]

PR #8911, opened 2023-04-21
[channels,settings] add a setting to ignore invalid devices

thenextman:
Add a setting `IgnoreInvalidDevices`. This is an escape hatch to optionally revert the behaviour of [df4c076](https://github.com/devolutions/freerdp/commit/df4c076946bc8716b8bf3fb9e512374b57cca1cb) - in #8909 we see a case where a server sends an IRP for an non-existent device, despite all rdpdr forwarding options being disabled.

If `IgnoreInvalidDevices` is set to `TRUE`, `rdpdr` will send a dummy response instead of returning an error (and preventing the connection from succeeding).

### DIFF
--- a/channels/rdpdr/client/irp.c
+++ b/channels/rdpdr/client/irp.c
@@ -107,7 +107,7 @@ IRP* irp_new(DEVMAN* devman, wStreamPool* pool, wStream* s, wLog* log, UINT* err
 	if (!device)
 	{
 		if (error)
-			*error = ERROR_INVALID_PARAMETER;
+			*error = ERROR_DEV_NOT_EXIST;
 
 		return NULL;
 	}
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -1067,6 +1067,8 @@ static UINT rdpdr_process_connect(rdpdrPlugin* rdpdr)
 	settings = rdpdr->rdpcontext->settings;
 	WINPR_ASSERT(settings);
 
+	rdpdr->ignoreInvalidDevices = settings->IgnoreInvalidDevices;
+
 	if (settings->ClientHostname)
 		strncpy(rdpdr->computerName, settings->ClientHostname, sizeof(rdpdr->computerName) - 1);
 	else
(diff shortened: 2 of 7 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

hedwigz:
I just stumbled this error also. Might worth making this a default?

akallabeth:
@hedwigz no, definitely not.
if you have any issue please provide some details, that is a bug!

Sebastian-Roth:
@akallabeth We might be able to provide some more information in this topic.

What I see is the following log output when running from the command line: `xfreerdp3 /u:username /v:hostname`:

```
[15:26:53:900] [2487:000009ca] [WARN][com.freerdp.channels.rdpdr.client] - [devman_get_device_by_id]: could not find device ID 0xffffffff
[15:26:53:900] [2487:000009ca] [ERROR][com.freerdp.channels.rdpdr.client] - [rdpdr_process_irp]: irp_new failed with 55!
[15:26:53:900] [2487:000009ca] [ERROR][com.freerdp.channels.rdpdr.client] - [rdpdr_process_receive]: rdpdr_process_irp failed with error 55
[15:26:53:900] [2487:000009ca] [ERROR][com.freerdp.channels.rdpdr.client] - [rdpdr_virtual_channel_client_thread]: rdpdr_process_receive failed with error 55!
[15:26:53:900] [2487:000009b8] [ERROR][com.freerdp.core] - [checkChannelErrorEvent]: rdpdr_virtual_channel_client_thread reported an error. Error was 55
[15:26:53:900] [2487:000009b8] [ERROR][com.freerdp.core] - [freerdp_check_event_handles]: checkChannelErrorEvent() failed - 0
[15:26:53:900] [2487:000009b8] [INFO][com.freerdp.client.common] - [client_auto_reconnect_ex]: Network disconnect!
```

We don't have resources forwarded - actually using xfreerdp from the command line to test without resource options set and non is activated by default as far as I see in the manual.

One very interesting finding a colleague made is that this error only starts to arise when the system we try connect to is joined to a AD domain. But on the other hand I have read through other topics like https://github.com/FreeRDP/FreeRDP/issues/10138 and none of those things (disabling kerberos, host name resolution fix in krb5.conf) made any difference in our case.

As well this only happens with the recent version 3.x but not with the 2.x versions of XFreeRDP.

Find a diff between logs before and after joining an AD domain here: https://www.diffchecker.com/lD26xB6L

Also see my post in another forum (https://forum.devolutions.net/topics/39314/rdp-is-not-able-to-connect--2023171) where the responder pointed me to this topic here.

Thanks in advance!

akallabeth:
@Sebastian-Roth this looks like a different issue, could you please open a new issue for that?

* What I can think of here is that in your environment `smartcard` redirection is enforced (to log in with one) and something is not handled correctly.
* the devolutions topic is correct as far as kerberos is concerned, it has nothing to do with this.

gbhauck:
I opened a new issue - I also have this problem (for about 1 year) ...

[issue 10807](https://github.com/FreeRDP/FreeRDP/issues/10807)

PR #8915, opened 2023-04-24
[server,proxy] allow PEM in config file

akallabeth:
CertificateContent and PrivateKeyContent now have two valid formats: It can be in format PEM (multiple lines) or a single line base64 encoded PEM.
The first format is preferrable in case the pf_config* API is used to set the certificate/key,  the latter in case an actual config file is in use where multiline configuration data can not be directly entered.


mfleisz (approved):
LGTM

PR #8918, opened 2023-04-24
[core,autodetect] revert to [MS-RDPBCGR] states

akallabeth:
remove 'convenience' states in main RDP state machine for autodetect

@pnowack got issues with a few tools that require state sync between client  and server RDP instances.
I´ve reverted your state changes for the main RDP state machine.
Does this way work for you too or are there issues on your end this way?


pnowack (approved):
LGTM (also works in testing here).

PR #8921, opened 2023-04-24
rail: add missing server-side calls

hardening:
This patch adds support for text_scale and caret_blink messages on the server-side.


akallabeth (approved):
LGTM

PR #8922, opened 2023-04-25
Openssl3 aad

akallabeth:
Follow up to #8920


akallabeth:
@fifthdegree I´ve modified the code so that we do not directly use `OpenSSL` but our wrappers. Can you confirm the code still works as you expect it?

libfreerdp/crypto/privatekey.c
```diff
@@ -357,3 +357,151 @@ size_t freerdp_key_get_bits(const rdpPrivateKey* key)
...
+		}
+		if (!bn)
+			return NULL;
+		bn = BN_dup(bn);
+		if (!bn)
+			return NULL;
+	}
+#endif
+
+	const int length = BN_num_bytes(bn);
```
fifthdegree, line 479: Remember to set the `plength` argument

libfreerdp/core/aad.c
```diff
@@ -607,30 +609,30 @@ static char* aad_create_jws_payload(rdpAad* aad, const char* ts_nonce)
...
+static char* aad_final_digest(rdpAad* aad, WINPR_DIGEST_CTX* ctx)
 {
 	char* jws_signature = NULL;
 
 	WINPR_ASSERT(aad);
 	WINPR_ASSERT(ctx);
 
 	size_t siglen = 0;
-	const int dsf = EVP_DigestSignFinal(ctx, NULL, &siglen);
+	const int dsf = winpr_DigestSign_Final(ctx, NULL, &siglen);
```
fifthdegree, line 635: Also change the call to `EVP_DigestSignFinal` below on line 650

libfreerdp/core/aad.c
```diff
@@ -905,44 +900,8 @@ static BOOL read_http_message(rdpAad* aad, BIO* bio, long* status_code, char** c
 
 static BOOL generate_rsa_2048(rdpAad* aad)
 {
-	BOOL rc = FALSE;
 	WINPR_ASSERT(aad);
-
-	EVP_PKEY_CTX* ctx = EVP_PKEY_CTX_new_id(EVP_PKEY_RSA, NULL);
```
fifthdegree, line 911: You need to allocate the `rdpPrivateKey` struct at some point before calling `freerdp_key_generate`

fifthdegree (commented):
Just a few issues; works well otherwise

libfreerdp/core/aad.c
akallabeth: done

libfreerdp/core/aad.c
akallabeth: done

libfreerdp/crypto/privatekey.c
akallabeth: done

hardening (approved):
LGTM

PR #8929, opened 2023-04-26
mcs: drop a server-side warning on FreeRDP_ChannelDefArray::len

hardening:
Server side we often see "FreeRDP_ChannelDefArray::len expected to be >= 31, but have XXX", where XXX is lower than 31.

This patche fixes that, the old code was setting the size of ChannelDefArray to the number of ChannelCount, which is usually not what we want. We want to keep it to 31 and have ChannelCount indicate how many of these channels are used.

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -1858,7 +1858,7 @@ BOOL rdp_channels_from_mcs(rdpSettings* settings, const rdpRdp* rdp)
 	WINPR_ASSERT(mcs);
 
 	if (!freerdp_settings_set_pointer_len(settings, FreeRDP_ChannelDefArray, NULL,
-	                                      mcs->channelCount))
+	                                      CHANNEL_MAX_COUNT))
 		return FALSE;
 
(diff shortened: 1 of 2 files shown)


akallabeth (changes requested):
Looks buggy, what if `FreeRDP_ChannelCount` is larger than `FreeRDP_ChannelDefArraySize`?

hardening:
> Looks buggy, what if `FreeRDP_ChannelCount` is larger than `FreeRDP_ChannelDefArraySize`?

This cannot happen, as the treatment in mcs is limited to what is possible in the packet: 31 channels.
Just to clarify, `FreeRDP_ChannelDefArraySize` is the capacity of `FreeRDP_ChannelDefArray`, while `FreeRDP_ChannelCount` is the number used items. So there's no real real reason to change the capacity of `FreeRDP_ChannelDefArray`.

akallabeth:
@hardening so, `freerdp_settings_clone`  et al work correctly in that case? I think I remember some issues during copy operations, so maybe have a look there too

akallabeth (approved):
LGTM

PR #8948, opened 2023-05-08
[winpr,scard] clean up smartcard wrapper

akallabeth:
there have been name collissions, remove all not required from API


hardening:
Moving private all the items related to `SCardApiFunctionTable` is probably a bad idea, as it's what could make it possible to have both smartcard emulation and regular support at the same time (smartcard emulation not a build time option)

akallabeth:
@hardening issue is that `WinPR` exposes functions that might clash with `pcsc-lite`.
so, imagine a channel that links in `WinPR` and `libfoo` where `libfoo` links `pcsc-lite`. you might end up `libfoo` accessing variables (or calling functions) from `WinPR` instead due to naming conflicts

hardening:
> @hardening issue is that `WinPR` exposes functions that might clash with `pcsc-lite`. so, imagine a channel that links in `WinPR` and `libfoo` where `libfoo` links `pcsc-lite`. you might end up `libfoo` accessing variables (or calling functions) from `WinPR` instead due to naming conflicts

Ok, I'm probably missing something but I don't see how the defines around `SCardApiFunctionTable` would hurt, I'd be to rename the problematic typedefs, not make the whole thing private.

awakecoding:
@akallabeth @hardening given that we are actually moving forward with runtime loading/switching of the WinSCard API function table, I vote not to do this suggested change. Where did you hit a conflict, and why would making the entire thing private fix it? We still want that API function table in a public header no matter what

PR #8958, opened 2023-05-10
Progressive fix

hardening:
Various progressive fixes / improvements, plus typo fix.


akallabeth (approved):
Overall LGTM,
just don´t get the `dirty` flag, it is set when we increment `numUpdatedTiles` and reset when we set `numUpdatedTiles=0`, so kind of redundant?

PR #8959, opened 2023-05-11
[core,settigs] default to 32bpp session color depth

akallabeth:
there is no real reason to default to `16bpp` nowadays.


akallabeth:
@freerdp-bot test again please!

PR #8960, opened 2023-05-11
refactor [locale/keyboard_x11]: cleanup and remove calls to xprop

bmiklautz:
* remove unused function freerdp_detect_keymap_from_xkb
* instead of querying the x keyboard rule properties (_XKB_RULES_NAMES_BACKUP and _XKB_RULES_NAMES) with xprop as external program use xlib directly


akallabeth (approved):
LGTM

PR #8961, opened 2023-05-11
File copy fix

akallabeth:
`url unescape` file paths, otherwise space and other symbols break file copy.


client/common/client_cliprdr_file.c
```diff
@@ -152,6 +152,55 @@ struct cliprdr_file_context
...
+		{
+			*px += 2;
+			return *pair->what;
+		}
+	}
+
+	return *what;
+}
+
+static char* url_unescape(const char* str, size_t len)
```
pnowack, line 184: Is it maybe worth it to put this function under winPR utils, so it can be reused?

client/common/client_cliprdr_file.c
```diff
@@ -152,6 +152,55 @@ struct cliprdr_file_context
...
 };
 
+struct escape_pair
+{
+	const char* what;
+	const char* escape;
+};
+
+static const struct escape_pair url_escape_codes[] = {
+	{ " ", "\%20" }, { "<", "\%3C" }, { ">", "\%3E" }, { "#", "\%23" },  { "%", "\%25" },
```
hardening, line 162: I don't get why you need \ in`\%` ?

client/common/client_cliprdr_file.c
akallabeth: might be worth a look.

client/common/client_cliprdr_file.c
akallabeth: @hardening true, not really required here.

hardening:
May I suggest to add a unitary test for these new public functions ?

akallabeth:
@hardening done. also rewritten the encoding/decoding to work with all unicode characters.

hardening (approved):
LGTM

PR #8963, opened 2023-05-11
Static linking fix

akallabeth:


hardening (approved):
No opinions on cmake things, looks ok for me.

PR #8964, opened 2023-05-11
[winpr,input] fix GetKeycodeFromVirtualKeyCode(code, KEYCODE_TYPE_XKB)

hardening:
As KEYCODE_TYPE_XKB is 3, in the previous code we were doing some "and masking" and so when calling GetKeycodeFromVirtualKeyCode(code, KEYCODE_TYPE_XKB), the function was always interpreting the virtual key code with the apple layout. This patch fixes that and also mutualize the search in the code array.

### DIFF
--- a/winpr/libwinpr/input/keycode.c
+++ b/winpr/libwinpr/input/keycode.c
@@ -878,41 +878,32 @@ DWORD GetVirtualKeyCodeFromKeycode(DWORD keycode, DWORD dwFlags)
 DWORD GetKeycodeFromVirtualKeyCode(DWORD vkcode, DWORD dwFlags)
 {
 	DWORD index;
-	DWORD keycode = 0;
+	DWORD* targetArray;
+	size_t targetSize;
 
-	if (dwFlags & KEYCODE_TYPE_APPLE)
+	switch (dwFlags)
 	{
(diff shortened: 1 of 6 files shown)


pnowack (approved):
Oh my god, sry. I did not realize that.

The `switch` looks btw pretty neat.

winpr/libwinpr/input/keycode.c
```diff
@@ -878,41 +878,32 @@ DWORD GetVirtualKeyCodeFromKeycode(DWORD keycode, DWORD dwFlags)
...
-		{
-			if (vkcode == KEYCODE_TO_VKCODE_APPLE[index])
-			{
-				keycode = index;
-				break;
-			}
-		}
-	}
-	else if (dwFlags & KEYCODE_TYPE_EVDEV)
+	switch (dwFlags)
```
pnowack, line 884: `dwFlags` could maybe be renamed too.

akallabeth (approved):

winpr/libwinpr/input/keycode.c
akallabeth: @pnowack pushed a commit for that and the invalid `+ 8` offset for mac keyboards

PR #8966, opened 2023-05-12
[smartcard] fix smartcard listing with /kerberos:pkcs11-module:<path>

hardening:
When a PKCS11 module was provided, the CSP could not be set by command line arguments, leading to an error when loading the ncrypt module, and an empty smartcard list.

### DIFF
--- a/libfreerdp/core/smartcardlogon.c
+++ b/libfreerdp/core/smartcardlogon.c
@@ -603,6 +603,9 @@ static BOOL smartcard_hw_enumerateCerts(const rdpSettings* settings, LPCWSTR csp
 		/* load a unique CSP by pkcs11 module path */
 		LPCSTR paths[] = { Pkcs11Module, NULL };
 
+		if (!csp)
+			csp = MS_SCARD_PROV;
+
 		status = winpr_NCryptOpenStorageProviderEx(&provider, csp, 0, paths);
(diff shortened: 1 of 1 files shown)


libfreerdp/core/smartcardlogon.c
```diff
@@ -603,6 +603,16 @@ static BOOL smartcard_hw_enumerateCerts(const rdpSettings* settings, LPCWSTR csp
 		/* load a unique CSP by pkcs11 module path */
 		LPCSTR paths[] = { Pkcs11Module, NULL };
 
+		if (!csp)
+		{
+			csp = _wcsdup(MS_SCARD_PROV);
```
akallabeth, line 608: @hardening memory leak?

akallabeth (changes requested):
1. memory leak (why allocate the `WCHAR` string?
2. the settings split for `smartcard-logon` and `/kerberos` looks problematic, might be a good idea to rethink that

libfreerdp/core/smartcardlogon.c
hardening: Ok fixed, I thought that it would make the callees segfault but in fact they all do copies which is fine.

hardening:
Code quality checker is totally dumb !

akallabeth (approved):
LGTM

PR #8971, opened 2023-05-15
channels/audin: Rework API to be closer to documentation

pnowack:

```
The current server side channel handling of AUDIO_INPUT is currently
very constrained:

- Server implementations cannot measure the clients uplink, since the
  Incoming Data PDU is currently unhandled and FreeRDPs DSP handling
  delays the callback call of ReceiveSamples
- Servers currently cannot prefer a different protocol version
- Servers currently cannot change the used format

To solve these issues without running into the risk that some
simplifications constraint certain API usage, rework the current channel
handling to be very close to the documentation.
This means, that all documented API calls can be made by server
implementations and all documented PDUs, that the server side is
expected to receive are just parsed inside FreeRDP and then forwarded to
the API implementation.
```

Overall the channel handling is now done in a similar way to the server side handling of RDPECAM.
Tested with the following g-r-d branch: https://gitlab.gnome.org/pnowack/gnome-remote-desktop/-/commits/audio-input-lf3


pnowack:
I also tested the `shadow_audin` part (except for checking the retrieved PCM samples themselves, because there was nothing hooked up to it in the first place), but did not realize, that there is also dummy audin handling in `mf_audin.c` and `sf_audin.c` (because I don't include them in my build)).

include/freerdp/channels/audin.h
```diff
@@ -21,9 +22,116 @@
 #define FREERDP_CHANNEL_AUDIN_H
 
 #include <freerdp/api.h>
+#include <freerdp/codec/audio.h>
 #include <freerdp/dvc.h>
 #include <freerdp/types.h>
 
 #define AUDIN_DVC_CHANNEL_NAME "AUDIO_INPUT"
 
+typedef enum
```
akallabeth, line 31: why expose this internal detail?
better keep types and structs not necessary for external use away from public headers

include/freerdp/server/shadow.h
```diff
@@ -99,7 +99,7 @@ extern "C"
 	typedef BOOL (*pfnShadowChannelAudinServerReceiveSamples)(rdpShadowSubsystem* subsystem,
 	                                                          rdpShadowClient* client,
 	                                                          const AUDIO_FORMAT* format,
-	                                                          wStream* buf, size_t nframes);
```
akallabeth, line 102: Any reason for not keeping the `wStream*` here?

server/Mac/mf_audin.c
```diff
@@ -29,47 +29,139 @@
...
+	open.captureFormat.nAvgBytesPerSec = 44100 * 2 * 2;
+	open.captureFormat.nBlockAlign = 4;
+	open.captureFormat.wBitsPerSample = 16;
+
+	return audin->SendOpen(audin, &open);
+}
+
+static UINT mf_peer_audin_receive_formats(audin_server_context* audin, const SNDIN_FORMATS* formats)
+{
+	mfPeerContext* context = audin->userdata;
```
akallabeth, line 70: better `WINPR_ASSERT` parameters

server/Mac/mf_audin.c
```diff
@@ -29,47 +29,139 @@
...
 	return CHANNEL_RC_OK;
 }
 
-/**
- * Function description
- *
- * @return 0 on success, otherwise a Win32 error code
- */
-static UINT mf_peer_audin_open_result(audin_server_context* context, UINT32 result)
+static UINT mf_peer_audin_incoming_data(audin_server_context* audin,
```
akallabeth, line 105: better `WINPR_ASSERT` parameters

server/Mac/mf_audin.c
```diff
@@ -29,47 +29,139 @@
...
+			}
+		}
+	}
+
+	WLog_ERR(TAG, "Could not agree on a audio format with the server");
+
+	return ERROR_INVALID_DATA;
+}
+
+static UINT mf_peer_audin_open_reply(audin_server_context* audin,
```
akallabeth, line 97: better `WINPR_ASSERT` parameters

server/Mac/mf_audin.c
```diff
@@ -29,47 +29,139 @@
...
+
+	WLog_DBG(TAG, "AUDIO_INPUT version of client: %u", version->Version);
+
+	formats.NumFormats = context->audin_n_server_formats;
+	formats.SoundFormats = context->audin_server_formats;
+
+	return audin->SendFormats(audin, &formats);
+}
+
+static UINT send_open(audin_server_context* audin)
```
akallabeth, line 51: better `WINPR_ASSERT` parameters

server/Mac/mf_audin.c
```diff
@@ -29,47 +29,139 @@
 #include <freerdp/log.h>
 #define TAG SERVER_TAG("mac")
 
-/**
- * Function description
- *
- * @return 0 on success, otherwise a Win32 error code
- */
-static UINT mf_peer_audin_opening(audin_server_context* context)
+static UINT mf_peer_audin_receive_version(audin_server_context* audin, const SNDIN_VERSION* version)
```
akallabeth, line 32: better `WINPR_ASSERT` parameters

server/Mac/mf_audin.c
```diff
@@ -29,47 +29,139 @@
...
-/**
- * Function description
- *
- * @return 0 on success, otherwise a Win32 error code
- */
-static UINT mf_peer_audin_receive_samples(audin_server_context* context, const AUDIO_FORMAT* format,
-                                          wStream* buf, size_t nframes)
+static UINT mf_peer_audin_data(audin_server_context* audin, const SNDIN_DATA* data)
 {
+	/* TODO: Implement */
```
akallabeth, line 115: better `WINPR_ASSERT` parameters

server/Mac/mf_audin.c
```diff
@@ -29,47 +29,139 @@
...
+	WINPR_ASSERT(data);
+
+	WLog_WARN(TAG, "not implemented");
+	WLog_DBG(TAG, "receive %" PRIdz " bytes.", data->DataSize);
+	return CHANNEL_RC_OK;
+}
+
+static UINT mf_peer_audin_receive_format_change(audin_server_context* audin,
+                                                const SNDIN_FORMATCHANGE* format_change)
+{
```
akallabeth, line 126: better `WINPR_ASSERT` parameters

server/Mac/mf_audin.c
```diff
@@ -29,47 +29,139 @@
...
+		WLog_ERR(TAG, "NewFormat in FormatChange differs from requested format");
+		return ERROR_INVALID_DATA;
+	}
+
+	WLog_DBG(TAG, "Received Format Change PDU: %u", format_change->NewFormat);
+
 	return CHANNEL_RC_OK;
 }
 
 void mf_peer_audin_init(mfPeerContext* context)
```
akallabeth, line 140: better `WINPR_ASSERT` parameters

server/Mac/mf_audin.c
```diff
@@ -29,47 +29,139 @@
...
+	context->audin->ReceiveFormats = mf_peer_audin_receive_formats;
+	context->audin->OpenReply = mf_peer_audin_open_reply;
+	context->audin->IncomingData = mf_peer_audin_incoming_data;
+	context->audin->Data = mf_peer_audin_data;
+	context->audin->ReceiveFormatChange = mf_peer_audin_receive_format_change;
+
+	context->audin_n_server_formats = server_audin_get_formats(&context->audin_server_formats);
+}
+
+void mf_peer_audin_uninit(mfPeerContext* context)
```
akallabeth, line 158: better `WINPR_ASSERT` parameters

server/Sample/sf_audin.c
```diff
@@ -32,51 +32,111 @@
...
 
 #if defined(CHANNEL_AUDIN_SERVER)
-/**
- * Function description
- *
- * @return 0 on success, otherwise a Win32 error code
- */
-static UINT sf_peer_audin_opening(audin_server_context* context)
+static UINT sf_peer_audin_receive_version(audin_server_context* audin, const SNDIN_VERSION* version)
 {
```
akallabeth, line 36: better `WINPR_ASSERT` parameters

server/Sample/sf_audin.c
```diff
@@ -32,51 +32,111 @@
...
+	formats.NumFormats = context->audin_n_server_formats;
+	formats.SoundFormats = context->audin_server_formats;
+
+	return audin->SendFormats(audin, &formats);
+}
 
-	WLog_DBG(TAG, "AUDIN opening.");
-	/* Simply choose the first format supported by the client. */
-	context->SelectFormat(context, 0);
+static UINT send_open(audin_server_context* audin)
```
akallabeth, line 54: better `WINPR_ASSERT` parameters

server/Sample/sf_audin.c
```diff
@@ -32,51 +32,111 @@
...
+	open.captureFormat.nChannels = 2;
+	open.captureFormat.nSamplesPerSec = 44100;
+	open.captureFormat.nAvgBytesPerSec = 44100 * 2 * 2;
+	open.captureFormat.nBlockAlign = 4;
+	open.captureFormat.wBitsPerSample = 16;
+
+	return audin->SendOpen(audin, &open);
+}
+
+static UINT sf_peer_audin_receive_formats(audin_server_context* audin, const SNDIN_FORMATS* formats)
```
akallabeth, line 71: better `WINPR_ASSERT` parameters

server/Sample/sf_audin.c
```diff
@@ -32,51 +32,111 @@
...
+			}
+		}
+	}
+
+	WLog_ERR(TAG, "Could not agree on a audio format with the server");
+
+	return ERROR_INVALID_DATA;
+}
+
+static UINT sf_peer_audin_open_reply(audin_server_context* audin,
```
akallabeth, line 100: better `WINPR_ASSERT` parameters

server/Sample/sf_audin.c
```diff
@@ -32,51 +32,111 @@
...
 }
 
-/**
- * Function description
- *
- * @return 0 on success, otherwise a Win32 error code
- */
-static UINT sf_peer_audin_open_result(audin_server_context* context, UINT32 result)
+static UINT sf_peer_audin_incoming_data(audin_server_context* audin,
+                                        const SNDIN_DATA_INCOMING* data_incoming)
```
akallabeth, line 109: better `WINPR_ASSERT` parameters

server/Sample/sf_audin.c
```diff
@@ -32,51 +32,111 @@
...
-static UINT sf_peer_audin_receive_samples(audin_server_context* context, const AUDIO_FORMAT* format,
-                                          wStream* buf, size_t nframes)
+static UINT sf_peer_audin_receive_format_change(audin_server_context* audin,
+                                                const SNDIN_FORMATCHANGE* format_change)
 {
-	/* TODO: Implement */
-	WINPR_ASSERT(context);
-	WINPR_ASSERT(format);
-	WINPR_ASSERT(buf);
+	testPeerContext* context = audin->userdata;
```
akallabeth, line 130: better `WINPR_ASSERT` parameters

server/Sample/sf_audin.c
```diff
@@ -140,6 +203,12 @@ BOOL sf_peer_audin_running(testPeerContext* context)
 void sf_peer_audin_uninit(testPeerContext* context)
 {
 #if defined(CHANNEL_AUDIN_SERVER)
-	audin_server_context_free(context->audin);
+	if (context->audin)
```
akallabeth, line 206: better `WINPR_ASSERT` parameters

server/shadow/shadow_audin.c
```diff
@@ -32,71 +32,117 @@
...
- *
- * @return 0 on success, otherwise a Win32 error code
- */
-static UINT AudinServerOpening(audin_server_context* context)
+
+static UINT AudinServerReceiveVersion(audin_server_context* audin, const SNDIN_VERSION* version)
 {
-	AUDIO_FORMAT* agreed_format = NULL;
-	size_t i = 0, j = 0;
+	rdpShadowClient* client = audin->userdata;
```
akallabeth, line 38: better `WINPR_ASSERT` parameters

server/shadow/shadow_audin.c
```diff
@@ -32,71 +32,117 @@
...
+
+	formats.NumFormats = client->audin_n_server_formats;
+	formats.SoundFormats = client->audin_server_formats;
+
+	return audin->SendFormats(audin, &formats);
+}
+
+static UINT send_open(audin_server_context* audin)
+{
+	rdpShadowClient* client = audin->userdata;
```
akallabeth, line 57: better `WINPR_ASSERT` parameters

server/shadow/shadow_audin.c
```diff
@@ -32,71 +32,117 @@
...
+	open.captureFormat.nAvgBytesPerSec = 44100 * 2 * 2;
+	open.captureFormat.nBlockAlign = 4;
+	open.captureFormat.wBitsPerSample = 16;
+
+	return audin->SendOpen(audin, &open);
+}
+
+static UINT AudinServerReceiveFormats(audin_server_context* audin, const SNDIN_FORMATS* formats)
+{
+	rdpShadowClient* client = audin->userdata;
```
akallabeth, line 74: better `WINPR_ASSERT` parameters

server/shadow/shadow_audin.c
```diff
@@ -32,71 +32,117 @@
...
- * @return 0 on success, otherwise a Win32 error code
- */
-static UINT AudinServerOpenResult(audin_server_context* context, UINT32 result)
+
+static UINT AudinServerOpenReply(audin_server_context* audin, const SNDIN_OPEN_REPLY* open_reply)
 {
-	/* TODO: Implement */
-	WLog_WARN(TAG, "not implemented");
-	WLog_INFO(TAG, "AUDIN open result %" PRIu32 ".\n", result);
+	/* TODO: Implement failure handling */
```
akallabeth, line 103: better `WINPR_ASSERT` parameters

server/shadow/shadow_audin.c
```diff
@@ -32,71 +32,117 @@
...
- * Function description
- *
- * @return 0 on success, otherwise a Win32 error code
- */
-static UINT AudinServerReceiveSamples(audin_server_context* context, const AUDIO_FORMAT* format,
-                                      wStream* buf, size_t nframes)
+
+static UINT AudinServerIncomingData(audin_server_context* audin,
+                                    const SNDIN_DATA_INCOMING* data_incoming)
+{
```
akallabeth, line 110: better `WINPR_ASSERT` parameters

server/shadow/shadow_audin.c
```diff
@@ -32,71 +32,117 @@
...
+	/* TODO: Implement bandwidth measure of clients uplink */
+	WLog_DBG(TAG, "Received Incoming Data PDU");
+	return CHANNEL_RC_OK;
+}
+
+static UINT AudinServerData(audin_server_context* audin, const SNDIN_DATA* data)
 {
-	rdpShadowClient* client = (rdpShadowClient*)context->data;
+	rdpShadowClient* client = audin->userdata;
```
akallabeth, line 118: better `WINPR_ASSERT` parameters

server/shadow/shadow_audin.c
```diff
@@ -32,71 +32,117 @@
...
+	                  client->audin_negotiated_format, data->Data, data->DataSize))
 		return ERROR_INTERNAL_ERROR;
 
 	return CHANNEL_RC_OK;
 }
+
+static UINT AudinServerReceiveFormatChange(audin_server_context* audin,
+                                           const SNDIN_FORMATCHANGE* format_change)
+{
+	rdpShadowClient* client = audin->userdata;
```
akallabeth, line 134: better `WINPR_ASSERT` parameters

channels/audin/server/audin.c
```diff
@@ -21,356 +22,206 @@
...
-		return ERROR_INVALID_DATA;
-	}
+	WINPR_ASSERT(context);
+	WINPR_ASSERT(header);
+
+	pdu.Header = *header;
 
-	Stream_Read_UINT32(s, Result);
-	IFCALLRET(audin->context.OpenResult, success, &audin->context, Result);
+	pdu.DataSize = Stream_GetRemainingLength(s);
```
akallabeth, line 187: Might be a good idea to usw a `wStream*` in the `SNDIN_DATA` pdu and initialize it as a static stream?

include/freerdp/channels/audin.h
```diff
@@ -21,9 +22,116 @@
...
+
+typedef struct
+{
+	SNDIN_PDU Header;
+} SNDIN_DATA_INCOMING;
+
+typedef struct
+{
+	SNDIN_PDU Header;
+	size_t DataSize;
```
akallabeth, line 127: better use a `wStream*` here

akallabeth (approved):
Overall LGTM, nice work!

some nitpicking though, hope you don´ t mind :)

server/Mac/mf_audin.c
pnowack: Done (same with the other `WINPR_ASSERT` parts).

include/freerdp/channels/audin.h
pnowack: They are actually part of the spec. I moved those now though to `channels/audin/server/audin.c` and `channels/audin/client/audin_main.c`.

server/Sample/sfreerdp.h
```diff
@@ -60,6 +60,10 @@ struct test_peer_context
 	HANDLE debug_channel_thread;
 #if defined(CHANNEL_AUDIN_SERVER)
 	audin_server_context* audin;
+	AUDIO_FORMAT* audin_server_formats;
```
akallabeth, line 63: on a second look, might these be moved to `audin_server_context`? they should be required by every server implementation and their scope is the same, so why not unify?

pnowack: Their allocation can be static (I do that in g-r-d). And in that case, they have to be always set to `NULL` before clearing the `audin_server_context` to prevent a crash.

The dummy handling in the test servers here in FreeRDP unfortunately use the same struct for the audin handling, that is also used for anything else, so this can be a bit perceived to be annoying here in the test servers.
However, normally the audin handling (or any other channel) is usually handled in its own source file with its own struct, where only the related settings are (e.g. PipeWire or Pulseaudio hook-up, etc.).

akallabeth: @pnowack I was more thinking along having some of the stuff moved to the channel. (you need to copy the formats initially, correct)
it would then be possible to have `AudinServerReceiveVersion`, `AudinServerReceiveFormats`, `AudinServerReceiveFormatChange` to have a common implementation.
(`	audin->serverVersion = SNDIN_VERSION_Version_2;` should also be moved to init code, so you are ready to go with just the remaining callbacks you need)

pnowack: > (you need to copy the formats initially, correct)

Not really, I just assign the pointer to the struct, containing my formats: https://gitlab.gnome.org/pnowack/gnome-remote-desktop/-/blob/78e2ef9194a3cad196a0f6b4920a7e0606157b83/src/grd-rdp-audio-input.c#L257

> (	audin->serverVersion = SNDIN_VERSION_Version_2; should also be moved to init code, so you are ready to go with just the remaining callbacks you need)

We don't know how the audin protocol changes in the future. I would like to preserve the behaviour, where the server can choose what protocol version it wants to support.
Would it be okay for you, if we set a default value in `audin_server_context_new()`? Then the server implementation can override that value, if it wants to?

`AudinServerReceiveFormatChange` cannot be in a common implementation. In version 2 of the protocol, the server side can tell the client to use a different format.
That format change does not apply immediately. The client has to confirm that change by sending that PDU itself to server side too.
After receiving that PDU, the server can then (and must) expect the (encoded) samples to be in the new format, but not before that.

akallabeth: > > (you need to copy the formats initially, correct)
> 
> Not really, I just assign the pointer to the struct, containing my formats: https://gitlab.gnome.org/pnowack/gnome-remote-desktop/-/blob/78e2ef9194a3cad196a0f6b4920a7e0606157b83/src/grd-rdp-audio-input.c#L257
> 
yes, that is what you do now. if the behaviour is changed to you provide this pointer (and size) to the `audin` context and it creates a copy on `init` then my suggestion would work. (more to do, but only once for each server implementation)

> > (	audin->serverVersion = SNDIN_VERSION_Version_2; should also be moved to init code, so you are ready to go with just the remaining callbacks you need)
> 
> We don't know how the audin protocol changes in the future. I would like to preserve the behaviour, where the server can choose what protocol version it wants to support. Would it be okay for you, if we set a default value in `audin_server_context_new()`? Then the server implementation can override that value, if it wants to?
> 
you can always check for the version, talking about defaults here.

> > AudinServerReceiveFormatChange to have a common implementation
> 
> `AudinServerReceiveFormatChange` cannot be in a common implementation. In version 2 of the protocol, the server side can tell the client to use a different format. That format change does not apply immediately. The client has to confirm that change by sending that PDU itself to server side too. After receiving that PDU, the server can then (and must) expect the (encoded) samples to be in the new format, but not before that.
if all the required variables are part of the `audin` context, then no, you can just let it be handled there. (that would be true for the shadow and sample server, you can always provide your own specific implementation if you want more)

akallabeth: If I think a bit more about it, `AudinServerIncomingData` could also get a `default` implementation (with some simple basic measuring algorithm)

channels/audin/server/audin.c
pnowack: Did that in the last push.

include/freerdp/server/shadow.h
pnowack: Uses now a `wStream *` since the last push.

akallabeth:
@pnowack I´ve pushed a commit  to your branch:
1. It moved format handling to the server channel
1. It provides a default implementation for most of the server side `audin` callbacks
1. It uses a `wLog` instance from `audin_server` context

with this it should be possible to have minimal template code to set up a working server channel, only the actual `Data` callback is required.

pnowack:
@akallabeth Seems to work! Also with overriding the callbacks (did not need to change my g-r-d handling here).

Some small things though:
- There is now a build warning about `server_audin_get_formats()` being implicitly declared (doesn't abort building for some reason though)
- Maybe the default callbacks should end with the suffix `_default` (or similar) to indicate, that these are just default handlings?
- `sf_peer_audin_init` only implicitly returns `FALSE`, when `CHANNEL_AUDIN_SERVER` is not defined.

pnowack:
@akallabeth Looks good now from my perspective. I think the refactor commit could be squashed with your previous commit (since its pretty much a fixup commit) and that should be it.

include/freerdp/server/audin.h
```diff
@@ -38,89 +39,141 @@ extern "C"
 
 	typedef struct s_audin_server_context audin_server_context;
 
-	typedef BOOL (*psAudinServerChannelIdAssigned)(audin_server_context* context, UINT32 channelId);
+	typedef BOOL (*psAudinServerChannelOpen)(audin_server_context* context);
+	typedef BOOL (*psAudinServerChannelIsOpen)(audin_server_context* context);
```
hardening, line 43: Shall we name it `psAudinServerChannelIsOpened` instead ?

pnowack: hm..I was thinking about that name too:

`Close()` closes the DVC and shut downs the channel thread handling.
`Open()` sets up the channel thread handling and opens the DVC by asking the client side to create the respective DVC.

But `IsOpen()` just mentions, that the channel thread handling is set up. To actually know, that the DVC is open, it needs to wait for the client response of the DVC creation request.
But `IsOpen()` can only be `TRUE`, in the time period between `Open()` and `Close()`.

PR #8972, opened 2023-05-15
[server,sample] move assert in SuppressOutput

akallabeth:
if allowDisplayUpdates == 0 no area rectangle is available.


pnowack (approved):
LGTM. Also works here in a short test.

PR #8976, opened 2023-05-16
[client,common] add command line argument /from

akallabeth:
This new argument allows reading all command line options from a file or from stdin. It is a standalone argument and can not be combined with any other.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -2605,8 +2605,8 @@ static void fill_credential_strings(COMMAND_LINE_ARGUMENT_A* args)
 	}
 }
 
-int freerdp_client_settings_parse_command_line_arguments(rdpSettings* settings, int argc,
-                                                         char** argv, BOOL allowUnknown)
+int freerdp_client_settings_parse_command_line_arguments_int(rdpSettings* settings, int argc,
+                                                             char* argv[], BOOL allowUnknown)
 {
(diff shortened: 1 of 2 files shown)


hardening:
Perhaps naming the new arg `/from` is not the best idea as we already have `/from-stdin` which does something mostly unrelated. Perhaps `/args-from` would be less misleading...

akallabeth:

mfleisz (approved):
LGTM

bernd-wechner:
https://github.com/FreeRDP/FreeRDP/pull/8976/commits/5bb2e76178f1820113d9c69fc3704e898b98c121 was merged in May 2023. But the /args-from argument remains undocumented: 

https://github.com/awakecoding/FreeRDP-Manuals/blob/master/User/FreeRDP-User-Manual.markdown 

and I just tried it with:

```
xfreerdp --version
This is FreeRDP version 2.11.5 (2.11.5)
```
Released in Jan 2024:

https://github.com/FreeRDP/FreeRDP/releases/tag/2.11.5

and not there.

I notice a 3.* line in parallel but ma using the Ubuntu noble packaged version which happens to be 2.11.5.

akallabeth:
@bernd-wechner the documentation for your specific version is `/help`
as for the link, I´ll remove that from the wiki.

PR #8989, opened 2023-05-17
[cmake] Enable static building

ligurio:
OSS Fuzz strongly recommends static linking for tests [1]. Patch enables static linking for a fuzzing test added in commit
2ad14696124b5a7689a323f56859efb329a3d25c ("Add fuzzer for certificate_data_set_pem()").

1. https://google.github.io/oss-fuzz/further-reading/fuzzer-environment/#runtime-dependencies

### DIFF
--- a/cmake/ConfigOptions.cmake
+++ b/cmake/ConfigOptions.cmake
@@ -199,6 +199,11 @@ if (BUILD_FUZZERS)
 
     set(BUILD_TESTING ON)
 
+    if (BUILD_SHARED_LIBS STREQUAL "OFF")
+        set(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
+        set(CMAKE_CXX_FLAGS "-static ${CMAKE_CXX_FLAGS}")
+    endif()
+
     # A special target with fuzzer and sanitizer flags.
     add_library(fuzzer_config INTERFACE)
 
--- a/packaging/deb/freerdp-nightly/rules
+++ b/packaging/deb/freerdp-nightly/rules
@@ -4,7 +4,7 @@ NULL =
 
 DEB_HOST_ARCH=$(shell dpkg-architecture -qDEB_HOST_ARCH)
 
-SANATIZE_ADDRESS = -DWITH_SANITIZE_ADDRESS=ON
+SANITIZE_ADDRESS = -DWITH_SANITIZE_ADDRESS=ON
 
 DEB_CMAKE_EXTRA_FLAGS :=  -GNinja \
                           -DCMAKE_SKIP_RPATH=FALSE \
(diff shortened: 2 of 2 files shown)


ligurio:
With proposed patch building in OSS Fuzz finished successfully [^1].

[^1] https://github.com/google/oss-fuzz/pull/4748

CMakeLists.txt
```diff
@@ -196,6 +196,11 @@ endif()
 option(BUILD_SHARED_LIBS "Build shared libraries" ${LIB_DEFAULT})
 option(EXPORT_ALL_SYMBOLS "Export all symbols form library" OFF)
 
+if (BUILD_SHARED_LIBS STREQUAL "OFF")
```
akallabeth, line 199: I don´t like this, should be moved to some `toolchain` file you use for your specific case and not part of the main `CMakeLists.txt`

libfreerdp/core/test/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ if (BUILD_FUZZERS)
 		target_link_libraries(${TestName} freerdp winpr freerdp-client fuzzer_config)
 		add_test(${TestName} ${TESTING_OUTPUT_DIRECTORY}/${MODULE_NAME} ${TestName})
 		set_target_properties(${TestName} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${TESTING_OUTPUT_DIRECTORY}")
+		set_target_properties(${TestName} PROPERTIES POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS})
```
akallabeth, line 41: why do you need to disable `POSITION_INDEPENDENT_CODE` ? if so, do you want it globally disabled or really only for this target?
might be a good idea to change the setting in the main `CMakeLists.txt` to an `option` so you can override it from outside

akallabeth (changes requested):

libfreerdp/core/test/CMakeLists.txt
ligurio: > why do you need to disable POSITION_INDEPENDENT_CODE ? if so, do you want it globally disabled or really only for this target?

I thought it is required for building static binary, but after double-checking it is not so.

Removed this hunk. If I'll need it, there is a `CMAKE_POSITION_INDEPENDENT_CODE` that can be disabled from outside.

CMakeLists.txt
ligurio: Moved:

```diff
--- a/cmake/ConfigOptions.cmake
+++ b/cmake/ConfigOptions.cmake
@@ -199,6 +199,11 @@ if (BUILD_FUZZERS)
 
     set(BUILD_TESTING ON)
 
+    if (BUILD_SHARED_LIBS STREQUAL "OFF")
+        set(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
+        set(CMAKE_CXX_FLAGS "-static ${CMAKE_CXX_FLAGS}")
+    endif()
+
     # A special target with fuzzer and sanitizer flags.
     add_library(fuzzer_config INTERFACE)
 
```

akallabeth (changes requested):

cmake/ConfigOptions.cmake
```diff
@@ -199,6 +199,11 @@ if (BUILD_FUZZERS)
 
     set(BUILD_TESTING ON)
 
+    if (BUILD_SHARED_LIBS STREQUAL "OFF")
```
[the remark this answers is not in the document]
akallabeth: sorry, missed the scope of the change moved, so change is indeed ok.

akallabeth (approved):

PR #8990, opened 2023-05-17
Use an embedded browser window to obtain an authorization code

fifthdegree:
This allows us to actually capture the authorization code instead of having to manually paste it in.

### DIFF
--- a/client/SDL/CMakeLists.txt
+++ b/client/SDL/CMakeLists.txt
@@ -98,9 +98,24 @@ set(LIBS
         freerdp-client
     )
 
+option(WITH_WEBVIEW "Build with QtWebEngine support for AAD login popup browser" OFF)
+if (WITH_WEBVIEW)
+	include(FindPkgConfig)
+	pkg_check_modules(QT REQUIRED Qt5WebEngineWidgets)
+
+  list(APPEND SRCS sdl_webview.cpp)
+  add_definitions(-DWITH_WEBVIEW)
+  include_directories(${QT_INCLUDE_DIRS})
+  list(APPEND LIBS ${QT_LIBRARIES})
+endif()
+
 add_executable(${PROJECT_NAME}
 	${SRCS}
     )
+
+if (QT_FOUND)
+  target_compile_options(${PROJECT_NAME} PRIVATE -fPIC)
+endif()
 
 set_target_properties(${PROJECT_NAME} PROPERTIES OUTPUT_NAME "sdl-freerdp")
 target_link_libraries(${PROJECT_NAME} PRIVATE ${LIBS})
--- a/client/SDL/CMakeLists.txt
+++ b/client/SDL/CMakeLists.txt
@@ -100,24 +100,30 @@ set(LIBS
 
 option(WITH_WEBVIEW "Build with QtWebEngine support for AAD login popup browser" OFF)
 if (WITH_WEBVIEW)
-	include(FindPkgConfig)
-	pkg_check_modules(QT REQUIRED Qt5WebEngineWidgets)
+  find_package(Qt5 COMPONENTS WebEngineWidgets REQUIRED)
 
-  list(APPEND SRCS sdl_webview.cpp)
-  add_definitions(-DWITH_WEBVIEW)
-  include_directories(${QT_INCLUDE_DIRS})
-  list(APPEND LIBS ${QT_LIBRARIES})
+  list(APPEND SRCS
+      sdl_webview.cpp
+      sdl_webview.hpp
+      )
+  list(APPEND LIBS Qt5::WebEngineWidgets)
 endif()
 
 add_executable(${PROJECT_NAME}
 	${SRCS}
     )
 
-if (QT_FOUND)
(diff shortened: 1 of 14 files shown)


akallabeth (commented):
not really sure to pull in qt as a dependency is a good idea.

maybe something like https://github.com/babelouest/iddawc/blob/v1.1.8/examples/microsoft_example.c is sufficient?

akallabeth:
not entirely sure we want a whole qt as dependency, but it might be ok for a client to (optionally) have this. (e.g. add it only to `SDL` client as that is already `C++` without changing the defaults for `FreeRDP` library itself)

client/common/webview.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+			if (pair.size() != 2 || pair[0] != "code")
+				continue;
+
+			QByteArray code = pair[1].toUtf8();
+			*codeptr = (char*)calloc(1, code.size() + 1);
+			strcpy(*codeptr, code.constData());
+			break;
+		}
+		qApp->exit();
```
hardening, line 54: Perhaps we could pass the QApplication as parameter in the ctor instead.

hardening:
I'd bet that creating the QApplication in FreeRDP will bother any program that uses FreeRDP and Qt...

fifthdegree:
> I'd bet that creating the QApplication in FreeRDP will bother any program that uses FreeRDP and Qt...

We can just check if `qApp` is already initialized. But if you guys decide that you want to scrap this idea, I do have another (not as good) idea. Turns out the client id used by client.wvd.microsoft.com has "https://login.microsoftonline.com/common/oauth2/nativeclient" as one of its 90 valid redirect URIs. That way, at least the url shows up in the browser bar, so it's easier to copy/paste

akallabeth:
@fifthdegree I´ve been looking into this a bit and found https://github.com/webview/webview which is a wrapper around platform specific webview stuff.
might be exactly what we need here (open a webview, get result code) and does have c bindings (so you don´t need a c++ compiler for the `FreeRDP` core project (the SDL client and the proxy modules are separate `CMake` projects you don´t have to build)
the previous mentioned library seems to be only able to handle the parsing of the oauth-2 parts and does not contain a webview

fifthdegree:
The issue with that wrapper is that it doesn't wrap the functionality needed to intercept requests in any way

fifthdegree:
@akallabeth, @hardening I've moved things around a bit.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot please test this nice thing!

client/SDL/CMakeLists.txt
```diff
@@ -98,10 +98,29 @@ set(LIBS
         freerdp-client
     )
 
+option(WITH_WEBVIEW "Build with QtWebEngine support for AAD login popup browser" OFF)
+if (WITH_WEBVIEW)
+	include(FindPkgConfig)
+	pkg_check_modules(QT Qt5WebEngineWidgets)
```
akallabeth, line 104: might be a good idea to just add `REQUIRED` here and remove the found check below. If we already request `WITH_WEBVIEW=ON` manually then expect it is available and if not fail.

akallabeth:
looks good (except a `clang-format` missing).
with a follow up pr we can split out `sdl_webview_get_aad_auth_code` to a separate library and wrap up different implementations (gtk, qt, windows, apple, ...)

client/SDL/CMakeLists.txt
```diff
@@ -98,10 +98,25 @@ set(LIBS
...
+  include_directories(${QT_INCLUDE_DIRS})
+  list(APPEND LIBS ${QT_LIBRARIES})
+endif()
+
 add_executable(${PROJECT_NAME}
 	${SRCS}
     )
 
+if (QT_FOUND)
+  target_compile_options(${PROJECT_NAME} PRIVATE -fPIC)
```
akallabeth, line 117: that is already handled with `CMAKE_POSITION_INDEPENDENT_CODE` (or the corresponding target option)

akallabeth (approved):

hardening (approved):

client/SDL/CMakeLists.txt
fifthdegree: Apparently, cmake will only enable the -fPIE flag which is not good enough for qt

akallabeth: `-v`? 
* I can compile perfectly without this line (and start `sdl-freerdp`)
* this option will only work with `gcc` and `clang` but fail with ms compilers
* https://stackoverflow.com/questions/15978608/qt-compile-your-code-with-fpic-or-fpie

akallabeth:
@freerdp-bot test

client/SDL/CMakeLists.txt
akallabeth: @fifthdegree but I´ve found the issue, using `find_package` with the `Qt::WebEngineWidgets` target properly sets all required options. When using `pkg-config` you have to manually set all of these, so I´ve pushed a fix here.

PR #9002, opened 2023-05-23
Compile and cmake fixes

akallabeth:
a cleanup sweep again to reduce warnings


client/SDL/sdl_disp.cpp
```diff
@@ -138,7 +138,7 @@ static BOOL sdl_disp_check_context(void* context, sdlContext** ppsdl, sdlDispCon
 	if (!context)
 		return FALSE;
 
-	sdl = (sdlContext*)context;
+	sdl = reinterpret_cast<sdlContext*>(context);
```
mfleisz, line 141: Should be a `static_cast`

client/SDL/sdl_freerdp.cpp
```diff
@@ -982,7 +975,7 @@ static void sdl_post_final_disconnect(freerdp* instance)
  * after the connection ends. */
 static DWORD WINAPI sdl_client_thread_proc(void* arg)
 {
-	sdlContext* sdl = (sdlContext*)arg;
+	auto sdl = reinterpret_cast<sdlContext*>(arg);
```
mfleisz, line 978: Should be `static_cast`

client/SDL/sdl_webview.cpp
```diff
@@ -47,7 +50,7 @@ class SchemeHandler : public QWebEngineUrlSchemeHandler
 				continue;
 
 			QByteArray code = pair[1].toUtf8();
-			*codeptr = (char*)calloc(1, code.size() + 1);
+			*codeptr = reinterpret_cast<char*>(calloc(1, code.size() + 1));
```
mfleisz, line 53: Should be `static_cast` and return check is missing. This `char** codeptr` should be replaced with a `std::string` anyways.

libfreerdp/crypto/privatekey.c
```diff
@@ -422,6 +422,8 @@ BOOL freerdp_key_generate(rdpPrivateKey* key, size_t key_length)
 
 char* freerdp_key_get_param(const rdpPrivateKey* key, enum FREERDP_KEY_PARAM param, size_t* plength)
 {
+	BYTE* buf = NULL;
```
mfleisz, line 425: This function still might return an "uninitialized" (zeroed) buffer (i.e. if `BN_bn2bin` fails)

client/SDL/sdl_disp.cpp
akallabeth: no, as the types are pointers to different structs. (just try it ;))

mfleisz: `context` is typeless (`void*`) so a `static_cast` is perfectly fine - just try it :-)

akallabeth: ah ^^
yes, indeed this instance the parameter is `void*` (did have the issues with `reinterpret_cast` with the ones that have `rdpContext*` as argument

mfleisz (approved):

PR #9008, opened 2023-05-26
[core,licensing] allow all SEC_AUTODETECT_REQ messages

akallabeth:
follow up to fdcc36dd8d9cb7b3569a3814f70dad5e3671aa35 autodetect messages also seem to be sent over global channel id instead of message channel id


akallabeth:

hardening (approved):

PR #9009, opened 2023-05-26
[cmake] build xfreerdp standalone project

akallabeth:
Follow up to #9005 for `xfreerdp`


mfleisz (approved):
Not a cmake expert but LGTM

PR #9012, opened 2023-05-29
fix:clipboard direction remote->local, need send empty format list

cumt2010:
## This is how are pull requests handled by FreeRDP
1. Every new pull request needs to build and pass the unit tests at https://ci.freerdp.com
1. At least 1 (better two) people need to review and test a pull request and agree to accept

## Preparations before creating a pull
* Rebase your branch to current master, no merges allowed!
* Try to clean up your commit history, group changes to commits
* Check your formatting! A _clang-format_ script can be found at ```.clang-format```
  * The cmake target ```clangformat``` reformats the whole codebase
* Optional (but higly recommended)
  * Run a clang scanbuild before and after your changes to avoid introducing new bugs
  * Run your compiler at pedantic level to check for new warnings

## To ease accepting your contribution
* Give the pull request a proper name so people looking at it have an basic idea what it is for
* Add at least a brief description what it does (or should do :) and what it's good for
* Give instructions on how to test your changes
* Ideally add unit tests if adding new features

## What you should be prepared for
* fix issues found during the review phase
* Joining our chat to talk to other developers or help them test your pull might accelerate acceptance
  * Matrix room : #FreeRDP:matrix.org (main)
  * XMPP channel: #FreeRDP#matrix.org@matrix.org (bridged)
  * IRC channel : #freerdp @ irc.oftc.net (bridged)
* Joining our mailing list <freerdp-devel@lists.sourceforge.net> may be helpful too.

### DIFF
--- a/channels/cliprdr/client/cliprdr_main.c
+++ b/channels/cliprdr/client/cliprdr_main.c
@@ -740,8 +740,16 @@ static UINT cliprdr_client_format_list(CliprdrClientContext* context,
 	    freerdp_settings_get_uint32(context->rdpcontext->settings, FreeRDP_ClipboardFeatureMask);
 	CLIPRDR_FORMAT_LIST filterList = cliprdr_filter_local_format_list(formatList, mask);
 	if (filterList.numFormats == 0)
-		return CHANNEL_RC_OK;
-
+	{
+		if (filterList.formats)
+		{
+			free(filterList.formats);
+			filterList.formats = NULL;
+		}
+		
+		filterList.common.msgType = CB_FORMAT_LIST;
+	}
+		
 	s = cliprdr_packet_format_list_new(&filterList, cliprdr->useLongFormatNames);
 	cliprdr_free_format_list(&filterList);
 


channels/cliprdr/client/cliprdr_main.c
```diff
@@ -740,8 +740,18 @@ static UINT cliprdr_client_format_list(CliprdrClientContext* context,
 	    freerdp_settings_get_uint32(context->rdpcontext->settings, FreeRDP_ClipboardFeatureMask);
 	CLIPRDR_FORMAT_LIST filterList = cliprdr_filter_local_format_list(formatList, mask);
 	if (filterList.numFormats == 0)
-		return CHANNEL_RC_OK;
+	{
+		WLog_DBG(TAG, "forbidden direction local->remote, also need send empty file list");
```
pnowack, line 744: It is right, that it needs to be possible to send empty format lists (e.g. on the initialization sequence of cliprdr, where the client needs to send a format list as part of that sequence, but also when the selection did indeed change and no supported formats are on the selection list).

However, this message is not necessarily true. There are situations, where it matches the situation, but not always (see the aforementioned situations).
So, just remote this message

akallabeth:
@cumt2010 can you please add a short description why you need this? If only `remote -> local` redirection is enabled, clearing the remote clipboard on local clipboard changes is not what is to be expected

akallabeth:
replaced by #9015

PR #9015, opened 2023-05-30
Clip dir fix

akallabeth:
* Improved version of #9012
* Initially client must send a format list (we use an empty one) to enable server side clipboard updates.
* `xfreerdp` should only send formats that are actually in the clipboard instead of every supported format


mfleisz (approved):
LGTM

PR #9021, opened 2023-06-01
[winpr] fix some msan warnings

akallabeth:
initialize local stack variables

### DIFF
--- a/winpr/libwinpr/synch/wait.c
+++ b/winpr/libwinpr/synch/wait.c
@@ -77,10 +77,10 @@ int _mach_clock_gettime(int clk_id, struct timespec* t);
 
 int _mach_clock_gettime(int clk_id, struct timespec* t)
 {
-	UINT64 time;
-	double seconds;
-	double nseconds;
-	mach_timebase_info_data_t timebase;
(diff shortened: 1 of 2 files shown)


winpr/libwinpr/synch/wait.c
```diff
@@ -179,9 +179,9 @@ static void ts_add_ms(struct timespec* ts, DWORD dwMilliseconds)
 
 DWORD WaitForSingleObjectEx(HANDLE hHandle, DWORD dwMilliseconds, BOOL bAlertable)
 {
-	ULONG Type;
-	WINPR_HANDLE* Object;
-	WINPR_POLL_SET pollset;
+	ULONG Type = 0;
+	WINPR_HANDLE* Object = NULL;
+	WINPR_POLL_SET pollset = { 0 };
```
hardening, line 184: ok for pollset init (call of `pollset_uninit` at the end of the function), but not for `Type` and `Object` that are set just below by `winpr_Handle_GetInfo`

akallabeth: @hardening `winpr_Handle_GetInfo` is a function that might change.

PR #9026, opened 2023-06-05
[core,rdp] log rdp_read_share_control_header

akallabeth:
in rdp_read_share_control_header read and log all available header data before checking remaining length.


akallabeth:

PR #9046, opened 2023-06-12
[common,assistance] clean up assistance file parser

akallabeth:
* ensure the input buffer is always `\0` terminated
* unify string allocations


hardening (approved):
Looks ok for me

PR #9048, opened 2023-06-12
Fix usage of explicit server name when different from connection host

awakecoding:
Fix usage of explicit server name (UserSpecifiedServerName) when it is set and different from the connection hostname (ServerHostname). This optional setting makes it possible to make a network connection to a given hostname yet perform TLS and Kerberos validation in function of a user-provided server name. This can come in handy when DNS isn't really possible, but also when doing TCP relaying of RDP connections with the PCB like we do in the Devolutions Gateway. The RD Gateway protocol does something similar under the hood - this is where UserSpecifiedServerName comes from, it's the internal option that sets the server name different from the connection hostname (RD Gateway). It can also come in handy to perform a connect with the IP address where you manually set the server name to be used for validation in cases where DNS isn't properly set up, and you don't feel like editing the hosts file locally. The introduction of UserSpecifiedServerName /server-name:<server-name> was done months prior to this, but obviously the work had only been done halfway, this is what's required to make it work properly.


akallabeth (approved):
LGTM,
works for normal connections.

PR #9049, opened 2023-06-12
[shadow,subsystem] refactor builtin loader

akallabeth:
follow up to #9043
@pnowack you´re right, forgot to commit the latest change :/

### DIFF
--- a/server/shadow/shadow_subsystem_builtin.c
+++ b/server/shadow/shadow_subsystem_builtin.c
@@ -22,60 +22,43 @@
 
 typedef struct
 {
-	const char* name;
+	const char* (*name)(void);
 	pfnShadowSubsystemEntry entry;
 } RDP_SHADOW_SUBSYSTEM;
 
-#ifdef WITH_SHADOW_X11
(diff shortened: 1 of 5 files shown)


server/shadow/shadow_subsystem_builtin.c
```diff
@@ -22,70 +22,51 @@
...
+static size_t g_SubsystemCount = ARRAYSIZE(g_Subsystems);
 
 static pfnShadowSubsystemEntry shadow_subsystem_load_static_entry(const char* name)
 {
-	int index;
-
 	if (!name)
 	{
-		for (index = 0; index < g_SubsystemCount; index++)
+		for (size_t index = 0; index < g_SubsystemCount; index++)
```
pnowack, line 43: Doesn't this for loop become now superfluous (it does now always pick the first entry)?

pnowack (approved):
Tested the change here, the shadow server works now again.

server/shadow/shadow_subsystem_builtin.c
akallabeth: @pnowack well, it will return `NULL` if `g_Subsystem` is empty. but yes, not really required.

PR #9052, opened 2023-06-13
Fixed some incorrect key mappings with.  I found that the "+/=" key o…

DanAtIntegrateIT:
I found that the "+/=" key on a UK keyboard didn't do anything.  I also found that the "+" key on the keypad acted as if it was the "+/=" key.  I've corrected some of the keycodes and confirmed that they keyboard functions as expected.

However I am not sure if I should do this when the keyboard layout is detected as United Kingdom for example with "/kbd:layout:"United Kingdom"".

Anyway, all works as expected for me on a standard UK key board.


akallabeth:
@freerdp-bot test please

PR #9053, opened 2023-06-13
Fixed some incorrect key mappings and added a keyboard shortcut for disconnect.

DanAtIntegrateIT:
I found that the "+/=" key on a UK keyboard didn't do anything. I also found that the "+" key on the keypad acted as if it was the "+/=" key. I've corrected some of the keycodes and confirmed that they keyboard functions as expected.
However I am not sure if I should do this when the keyboard layout is detected as United Kingdom for example with "/kbd:layout:"United Kingdom"".
Anyway, all works as expected for me on a standard UK key board.


akallabeth (approved):

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test please

hardening (approved):

hardening:
Not your code, but giving a look at `sdlInput::keyboard_handle_event` there's place for code factorization (instead of multiple tests for key down, that could be a if before the switch). Thanks for the contribution.

DanAtIntegrateIT:
Thanks David, I see what you mean, I'll take a look and push a change.

akallabeth:
@hardening @DanAtIntegrateIT yes, could be unified, but prefer this way as we might want to react to key press *or* key release events for specific keys (there are still things to be done ;))

akallabeth:
@DanAtIntegrateIT and thank you for this nice fix!

PR #9056, opened 2023-06-14
Server connection limit

akallabeth:


hardening (approved):
I was not that much for adding the feature at the listener level, but given the design of listener I don't see any obvious way to do it differently.
So LGTM.

PR #9086, opened 2023-06-22
Multitouch common

akallabeth:
* move x11 code to client common (to be reused by other clients)
* add support for rotation and tilt flags
* add raw event functions to call
* add functions with `va_list` as well
* simplify internal handling of new contact detection

@digitalsignalperson could you have a look at this? only have a touch input device, no pen to test.
might also be a good idea to check how the `SDL` client handles the pen input.

### DIFF
--- a/channels/rdpei/client/rdpei_main.c
+++ b/channels/rdpei/client/rdpei_main.c
@@ -1184,7 +1184,7 @@ static UINT rdpei_add_pen(RdpeiClientContext* context, INT32 externalId,
 }
 
 static UINT rdpei_pen_process(RdpeiClientContext* context, INT32 externalId, UINT32 contactFlags,
-                              UINT32 fieldFlags, INT32 x, INT32 y, BOOL begin, va_list ap)
+                              UINT32 fieldFlags, INT32 x, INT32 y, va_list ap)
 {
 	RDPINPUT_PEN_CONTACT_POINT* contactPoint;
 	RDPEI_PLUGIN* rdpei;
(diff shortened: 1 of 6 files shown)


hardening (approved):
Looks ok

channels/rdpei/client/rdpei_main.c
```diff
@@ -1310,7 +1330,7 @@ static UINT rdpei_pen_hover_update(RdpeiClientContext* context, INT32 externalId
 	va_start(ap, y);
 	error = rdpei_pen_process(context, externalId,
 	                          RDPINPUT_CONTACT_FLAG_UPDATE | RDPINPUT_CONTACT_FLAG_INRANGE,
-	                          fieldFlags, x, y, FALSE, ap);
+	                          fieldFlags, x, y, ap);
```
digitalsignalperson, line 1333: with the begin logic added to rdpei_pen_process rdpei_pen_hover_update is now the same as rdpei_pen_hover_begin so could delete one

digitalsignalperson:
Doing some testing:

In the SDL client:
- pen hover and touch is interpreted as mouse (cursor is mouse icon), for both tip and eraser
- maybe touch is too, since in ms paint can only draw with 1 finger
- and pinch zoom doesn't work (e.g. in powerpoint)

In the X11 client:
- pen also seems to be interpreted as mouse
- touch not working properly. If I try touch in my paint:
    - clicking doesn't work
    - dragging finger from one point to another doesn't show motion
    - touching finger and then lift for 3 different points: A, B, C
        - when B is touched, I see the cursor move to point A
        - when C is touched I see the cursor move to point B

channels/rdpei/client/rdpei_main.c
akallabeth: true, but lets keep it the way it is.

akallabeth:
@digitalsignalperson ok, but does the behaviour between before this pr and after change? It is just meant to unify the client pen & touch handling, fixing the whole thing will be part of another pr if that was/is not working.

PR #9088, opened 2023-06-22
Adding CredSSP with Early User Auth support (sec:ext)

hlefebvre:
Recently I wasn't able to connect to my work servers using FreeRDP.

After digging I found that without using /sec:ext the server was closing the connection.

But with the command line switch it wasn't working, I found out that HYBRID_EX wasn't handled in nego.c.

Afterward I found out that we were receiving 4 bytes between the end of NLA and the MCS connect, I found out that it was the Early User Auth Result PDU (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/d0e560a3-25cb-4563-8bdc-6c4cc625bbfc).

I don't know RDP protocol neither the FreeRDP source code but I attempted to add support the best I can, trying to respect the way the code is architectured and using clang_format on the modified sources.

With this patch I can now connect to my work servers again, if it can help other people then I'm happy to share the patch.

### DIFF
--- a/libfreerdp/core/nla.c
+++ b/libfreerdp/core/nla.c
@@ -54,6 +54,12 @@
 
 #define NLA_AUTH_PKG NEGO_SSP_NAME
 
+typedef enum
+{
+	AUTHZ_SUCCESS = 0x00000000,
+	AUTHZ_ACCESS_DENIED = 0x00000005,
+} AUTHZ_RESULT;
+
 /**
  * TSRequest ::= SEQUENCE {
  * 	version    [0] INTEGER,
@@ -127,6 +133,7 @@ struct rdp_nla
 	char* pkinitArgs;
 	SmartcardCertInfo* smartcardCert;
 	BYTE certSha1[20];
+	BOOL earlyUserAuth;
 };
 
 static BOOL nla_send(rdpNla* nla);
@@ -137,6 +144,13 @@ static BOOL nla_decrypt_public_key_echo(rdpNla* nla);
 static BOOL nla_decrypt_public_key_hash(rdpNla* nla);
 static BOOL nla_encrypt_ts_credentials(rdpNla* nla);
 static BOOL nla_decrypt_ts_credentials(rdpNla* nla);
+
+void nla_set_early_user_auth(rdpNla* nla, BOOL earlyUserAuth)
+{
+	WINPR_ASSERT(nla);
+	WLog_DBG(TAG, "Early User Auth active: %s", nla->earlyUserAuth ? "true" : "false");
+	nla->earlyUserAuth = earlyUserAuth;
+}
 
 static void nla_buffer_free(rdpNla* nla)
 {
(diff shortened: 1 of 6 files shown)


akallabeth:
@freerdp-bot test please

libfreerdp/core/nla.c
```diff
@@ -1552,63 +1580,80 @@ int nla_recv_pdu(rdpNla* nla, wStream* s)
...
 
-	if (nla_decode_ts_request(nla, s) < 1)
-		return -1;
-
-	if (nla->errorCode)
+	if (nla_get_state(nla) == NLA_STATE_EARLY_USER_AUTH)
 	{
 		UINT32 code;
+		Stream_Read_UINT32(s, code);
+		if (code != 0x00000000)
```
akallabeth, line 1587: better use the `AUTHZ_SUCCESS` and `AUTHZ_ACCESS_DENIED` defines here

libfreerdp/core/nla.c
```diff
@@ -1552,63 +1580,80 @@ int nla_recv_pdu(rdpNla* nla, wStream* s)
...
+		{
+			WLog_DBG(TAG, "Early User Auth active: FAILURE");
```
akallabeth, line 1589: maybe log the result code too (should only be 2 allowed, but you never know when that will change)

akallabeth (approved):
LGTM,
nice patch, thank you for your contribution!

libfreerdp/core/transport.c
```diff
@@ -374,6 +375,59 @@ BOOL transport_connect_nla(rdpTransport* transport)
 	return rdp_client_transition_to_state(rdp, CONNECTION_STATE_NLA);
 }
 
+BOOL transport_connect_nla_ext(rdpTransport* transport)
```
hardening, line 378: Hum this function is a copy'n paste of `transport_connect_nla`, instead I'd suggest to have a `transport_connect_nla_ex(rdpTransport* transport, BOOL isExt)` and `transport_connect_nla_ext` and `transport_connect_nla` would call this one setting the correct value for `isExt`.

libfreerdp/core/nla.c
hlefebvre: ok changed

libfreerdp/core/nla.c
hlefebvre: I agree with you, code updated

libfreerdp/core/transport.c
hlefebvre: I agree, I remove the nla_ext function and add a BOOL param to connect_nla should be better

hlefebvre:
> LGTM, nice patch, thank you for your contribution!
> 
> if you could squash the commits into one and fix the (minor) remarks it would be really great!

ok I fixed everything and squashed it

libfreerdp/core/nla.c
```diff
@@ -54,6 +54,9 @@
 
 #define NLA_AUTH_PKG NEGO_SSP_NAME
 
+static const UINT32 AUTHZ_SUCCESS = 0x00000000;
```
hardening, line 57: nitpicking but in FreeRDP we use enums for these defines

hardening:
Fix my remark and we're ok (we'll ask for test bots after that)

libfreerdp/core/nla.c
hlefebvre: ok I never used them in C, I've changed it to an enum

hardening (approved):

hardening:
@freerdp-bot can you test buddy ?

hlefebvre:
I've fixed the wrong indentation for the enum

libfreerdp/core/nla.c
```diff
@@ -54,6 +54,12 @@
 
 #define NLA_AUTH_PKG NEGO_SSP_NAME
 
+typedef enum
+{
+	AUTHZ_SUCCESS = 0x00000000,
```
akallabeth, line 59: redefinition, already defined in `nego.h`

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test please

libfreerdp/core/nla.c
akallabeth: ok, pushed a fix.

PR #9093, opened 2023-06-26
[multitransport] Ignore unknown data

akarl10:
If the reserved filed is not 0 the request PDU seems to contain some extra data. Two bytes of 0 (probably a version field) followed by a JSON payload (not null terminated, until the end of the packet. There seems to be no dedicated length field)

This PDU was observed when connecting to a AVD over a gateway. The Json is probably metadata needed for a STUN/TURN setup of the UDP tunnels that are used by multitransport.

### DIFF
--- a/libfreerdp/core/multitransport.c
+++ b/libfreerdp/core/multitransport.c
@@ -63,13 +63,29 @@ state_run_t multitransport_recv_request(rdpMultitransport* multi, wStream* s)
 
 	UINT32 requestId;
 	UINT16 requestedProto;
+	UINT16 reserved;
 	const BYTE* cookie;
 
 	Stream_Read_UINT32(s, requestId);      /* requestId (4 bytes) */
 	Stream_Read_UINT16(s, requestedProto); /* requestedProtocol (2 bytes) */
-	Stream_Seek(s, 2);                     /* reserved (2 bytes) */
+	Stream_Read_UINT16(s, reserved);       /* reserved (2 bytes) */
 	cookie = Stream_ConstPointer(s);
 	Stream_Seek(s, RDPUDP_COOKIE_LEN); /* securityCookie (16 bytes) */
+	if (reserved != 0)
+	{
+		/*
+		 * If the reserved filed is not 0 the request PDU seems to contain some extra data.
+		 * If the reserved value is 1, then two bytes of 0 (probably a version field)
+		 * are followed by a JSON payload (not null terminated, until the end of the packet.
+		 * There seems to be no dedicated length field)
+		 *
+		 * for now just ignore all that
+		 */
+		WLog_WARN(TAG,
+		          "reserved is %" PRIu16 " instead of 0, skipping %" PRIuz "bytes of unknown data",
+		          reserved, Stream_GetRemainingLength(s));
+		Stream_SafeSeek(s, Stream_GetRemainingLength(s));
+	}
 
 	WINPR_ASSERT(multi->MtRequest);
 	return multi->MtRequest(multi, requestId, requestedProto, cookie);


hardening (approved):
Looks ok for me (would need to have a look at what's going on in MS dlls regarding this)

hardening:

libfreerdp/core/multitransport.c
```diff
@@ -63,13 +63,20 @@ state_run_t multitransport_recv_request(rdpMultitransport* multi, wStream* s)
...
 
 	Stream_Read_UINT32(s, requestId);      /* requestId (4 bytes) */
 	Stream_Read_UINT16(s, requestedProto); /* requestedProtocol (2 bytes) */
-	Stream_Seek(s, 2);                     /* reserved (2 bytes) */
+	Stream_Read_UINT16(s, reserved);       /* reserved (2 bytes) */
 	cookie = Stream_ConstPointer(s);
 	Stream_Seek(s, RDPUDP_COOKIE_LEN); /* securityCookie (16 bytes) */
+	if (reserved != 0)
+	{
+		WLog_WARN(TAG, "reserved is %d instead of 0, skipping %d bytes of unknown data", reserved,
```
akallabeth, line 76: illegal format string, `size_t` requires `PRIuz`

akallabeth (changes requested):

libfreerdp/core/multitransport.c
```diff
@@ -63,13 +63,20 @@ state_run_t multitransport_recv_request(rdpMultitransport* multi, wStream* s)
...
+		          Stream_GetRemainingLength(s));
+		Stream_Seek(s, Stream_GetRemainingLength(s));
```
akallabeth, line 78: `Stream_SafeSeek` might be appropriate, as we don´t know if the stream actually contains all that data

akallabeth:
@akarl10 thank you, nice patch, but the comment for this pr might be better added in code so we know why it was added (or what is missing)

libfreerdp/core/multitransport.c
```diff
@@ -63,13 +63,20 @@ state_run_t multitransport_recv_request(rdpMultitransport* multi, wStream* s)
...
 
 	Stream_Read_UINT32(s, requestId);      /* requestId (4 bytes) */
 	Stream_Read_UINT16(s, requestedProto); /* requestedProtocol (2 bytes) */
-	Stream_Seek(s, 2);                     /* reserved (2 bytes) */
+	Stream_Read_UINT16(s, reserved);       /* reserved (2 bytes) */
 	cookie = Stream_ConstPointer(s);
 	Stream_Seek(s, RDPUDP_COOKIE_LEN); /* securityCookie (16 bytes) */
+	if (reserved != 0)
+	{
+		WLog_WARN(TAG, "reserved is %d instead of 0, skipping %" PRIuz "bytes of unknown data",
```
akallabeth, line 76: same is true for  `UINT16` (`unsigned short`) which requires `PRIu16`

libfreerdp/core/multitransport.c
hardening: I disagree except if I miss something `Stream_Seek(s, Stream_GetRemainingLength(s));` is always safe as Safe_Seek just checks `Stream_GetRemainingLength(s)`

akallabeth: @hardening true, forgot to add that with the log above `reserved` should contain the amount of remaining data?

libfreerdp/core/multitransport.c
akarl10: ok

akallabeth (approved):
LGTM,
thank you for this contribution!

libfreerdp/core/multitransport.c
akarl10: reserved does actually not contain the length but 1. So probably reserved is now a version field.

akallabeth:

PR #9099, opened 2023-06-26
Mouse move restrict

akallabeth:


client/X11/xf_graphics.c
```diff
@@ -575,6 +575,11 @@ static BOOL xf_Pointer_SetPosition(rdpContext* context, UINT32 x, UINT32 y)
 	WLog_DBG(TAG, "%" PRIu32 "x%" PRIu32, x, y);
 	if (xfc->remote_app && !xfc->focused)
 		return TRUE;
+	else if (!xfc->remote_app)
+	{
+		if (!xfc->focused)
```
pnowack, line 580: If we take the already existing `if ()` part here, couldn't we simplify the condition to `if (!xfc->focused && (xfc->remote_app ||  !xfc->remote_app))`, and that to `if (!xfc->focused && (TRUE))`, and that to `if (!xfc->focused)`?

akallabeth: good point

client/common/cmdline.c
```diff
@@ -4001,6 +4001,42 @@ static int freerdp_client_settings_parse_command_line_arguments_int(rdpSettings*
...
+					const PARSE_ON_OFF_RESULT bval = parse_on_off_option(cur);
+					if (bval == PARSE_FAIL)
+						rc = COMMAND_LINE_ERROR_UNEXPECTED_VALUE;
+					else
+					{
+						const BOOL val = bval != PARSE_OFF;
+						size_t key = 0;
+						if (option_starts_with("relative", cur))
+							key = FreeRDP_MouseUseRelativeMove;
+						else if (option_starts_with("grab", cur))
```
pnowack, line 4024: Looks like the `local-move` setting is missing

pnowack (approved):
Just one thing, otherwise lgtm

client/common/cmdline.c
akallabeth: oops, should not have been in header, fixed.

akallabeth:

PR #9102, opened 2023-06-26
[gateway,arm] Azure Virtual Desktop Gateway support

akarl10:
This PR creates a new /gateway:type:arm transport. It depends on CJSON

The arm transport is not a transport by itself but is responsible for getting the websocket endpoint from a configuration url derived from the configured gateway host in the rdpw file or cmdline


akallabeth:
@freerdp-bot test please

akarl10:
I think I found the issue with mac os compile
Also fixed a typo in the message if compiled without cjson

akallabeth:
@freerdp-bot test please

libfreerdp/core/transport.c
```diff
@@ -470,6 +471,10 @@ BOOL transport_connect(rdpTransport* transport, const char* hostname, UINT16 por
 
 	if (transport->GatewayEnabled)
 	{
+		if (settings->GatewayArmTransport)
```
akallabeth, line 474: for new code better use `freerdp_settings_get_bool` (and no braces for one liners)

libfreerdp/core/gateway/wst.c
```diff
@@ -337,7 +358,15 @@ BOOL wst_connect(rdpWst* wst, DWORD timeout)
 	WINPR_ASSERT(wst);
 	if (!wst_tls_connect(wst, wst->tls, timeout))
 		return FALSE;
-
+	if (wst->settings->GatewayArmTransport)
```
akallabeth, line 361: for new code better use `freerdp_settings_get_bool`

libfreerdp/core/gateway/wst.c
```diff
@@ -351,11 +380,55 @@ BOOL wst_connect(rdpWst* wst, DWORD timeout)
...
-		case HTTP_STATUS_NOT_FOUND:
-		case HTTP_STATUS_BAD_METHOD:
+		case HTTP_STATUS_FORBIDDEN:
+		case HTTP_STATUS_OK:
 		{
-			http_response_free(response);
-			return FALSE;
+			/* AVD returns a 403 response with a ARRAffinity cookie set. retry with that cookie */
+			const char* affinity = http_response_get_setcookie(response, "ARRAffinity");
+			if (affinity && wst->settings->GatewayArmTransport)
```
akallabeth, line 388: for new code better use `freerdp_settings_get_bool`

libfreerdp/core/gateway/websocket.c
```diff
@@ -99,6 +100,42 @@ BOOL websocket_write_wstream(BIO* bio, wStream* sPacket, WEBSOCKET_OPCODE opcode
...
+{
+	WINPR_ASSERT(bio);
+	WINPR_ASSERT(data);
+	int status;
+	int offset = 0;
+
+	while (offset < length)
+	{
+		ERR_clear_error();
+		status = BIO_write(bio, &data[offset], length - offset);
```
akallabeth, line 113: variable could be declared here

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+static BOOL arm_tls_connect(rdpArm* arm, rdpTls* tls, int timeout)
+{
+	WINPR_ASSERT(arm);
+	WINPR_ASSERT(tls);
+	int sockfd = 0;
+	long status = 0;
+	BIO* socketBio = NULL;
+	BIO* bufferedBio = NULL;
+	rdpSettings* settings = arm->settings;
+	const char* peerHostname = arm->settings->GatewayHostname;
```
akallabeth, line 72: * better `WINPR_ASSERT` all elements to be accessed
* for settings better use `freerdp_settings_get_*` for new code

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+	UINT16 peerPort = arm->settings->GatewayPort;
+	const char *proxyUsername, *proxyPassword;
+	BOOL isProxyConnection =
+	    proxy_prepare(settings, &peerHostname, &peerPort, &proxyUsername, &proxyPassword);
+
+	sockfd = freerdp_tcp_connect(arm->context, peerHostname, peerPort, timeout);
+
+	WLog_DBG(TAG, "connecting to %s %d", peerHostname, peerPort);
+	if (sockfd < 0)
+	{
```
akallabeth, line 82: no braces for one liners

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+	BIO_set_fd(socketBio, sockfd, BIO_CLOSE);
+	bufferedBio = BIO_new(BIO_s_buffered_socket());
+
+	if (!bufferedBio)
+	{
+		BIO_free_all(socketBio);
+		return FALSE;
+	}
+
+	bufferedBio = BIO_push(bufferedBio, socketBio);
```
akallabeth, line 103: missing return check

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+		}
+	}
+
+	if (!status)
+	{
+		BIO_free_all(bufferedBio);
+		return FALSE;
+	}
+
+	tls->hostname = arm->settings->GatewayHostname;
```
akallabeth, line 122: for new code use `freerdp_settings_get_*`

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+	tls->port = arm->settings->GatewayPort;
+	tls->isGatewayTransport = TRUE;
+	status = freerdp_tls_connect(tls, bufferedBio);
+	if (status < 1)
+	{
+		rdpContext* context = arm->context;
+		if (status < 0)
+		{
```
akallabeth, line 130: no braces for one liners

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+			freerdp_set_last_error_if_not(context, FREERDP_ERROR_TLS_CONNECT_FAILED);
+		}
+		else
+		{
```
akallabeth, line 134: no braces for one liners

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+	uri = http_context_get_uri(arm->http);
+	request = http_request_new();
+
+	if (!request)
+		return NULL;
+
+	if (!http_request_set_method(request, method) || !http_request_set_uri(request, uri))
+		goto out;
+
+	else if (arm->settings->GatewayHttpExtAuthBearer)
```
akallabeth, line 163: use `freerdp_settings_get_*` for new code

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+	{
+		http_request_set_auth_scheme(request, "Bearer");
+		http_request_set_auth_param(request, arm->settings->GatewayHttpExtAuthBearer);
+	}
+
+	http_request_set_transfer_encoding(request, transferEncoding);
```
akallabeth, line 169: return checks

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+	return s;
+}
+
+static BOOL arm_send_http_request(rdpArm* arm, rdpTls* tls, const char* method,
+                                  const char* content_type, const char* data, size_t content_length)
+{
+	size_t sz;
+	wStream* s = NULL;
+	int status = -1;
+	s = arm_build_http_request(arm, method, TransferEncodingIdentity, content_type, content_length);
```
akallabeth, line 189: `wStream* s = ...`? (same for other variables)

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+		return FALSE;
+
+	sz = Stream_Length(s);
+
+	if (sz <= INT_MAX)
+		status = freerdp_tls_write_all(tls, Stream_Buffer(s), (int)sz);
+
+	Stream_Free(s, TRUE);
+	if (status >= 0 && content_length > 0)
+	{
```
akallabeth, line 201: no braces for one liners

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+
+	rdpArm* arm = (rdpArm*)calloc(1, sizeof(rdpArm));
+	char* lbi = NULL;
+	cJSON* json = NULL;
+	char* message = NULL;
+
+	if (arm)
+	{
+		arm->context = context;
+		arm->settings = arm->context->settings;
```
akallabeth, line 239: `NULL` check or `assert`

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+		    !http_context_set_connection(arm->http, "Keep-Alive") ||
+		    !http_context_set_user_agent(arm->http, "FreeRDP/3.0") ||
+		    !http_context_set_x_ms_user_agent(arm->http, "FreeRDP/3.0") ||
+		    !http_context_set_host(arm->http, arm->settings->GatewayHostname))
+			goto arm_error;
+
+		if (!arm_tls_connect(arm, arm->tls, timeout))
+			goto arm_error;
+
+		json = cJSON_CreateObject();
```
akallabeth, line 265: will not compile without `cJSON`, better move the guards so everything is covered.

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+		if (!json)
+			goto arm_error;
+		cJSON_AddStringToObject(json, "application", arm->settings->RemoteApplicationProgram);
+
+		lbi = calloc(arm->settings->LoadBalanceInfoLength + 1, sizeof(char));
```
akallabeth, line 270: better use the `freerdp_settings_set_*` functions for new code

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,348 @@
...
+		message = NULL;
+		cJSON_Delete(json);
+		json = NULL;
+
+		response = http_response_recv(arm->tls, TRUE);
+		if (!response)
+			goto arm_error;
+
+		StatusCode = http_response_get_status_code(response);
+		if (StatusCode == HTTP_STATUS_OK)
```
akallabeth, line 301: might be a good idea to split this function into smaller ones, one per scope (easier to read/understand)

libfreerdp/core/gateway/http.c
```diff
@@ -302,12 +340,52 @@ BOOL http_context_set_rdg_auth_scheme(HttpContext* context, const char* RdgAuthS
...
+	if (!context || !CookieName || !CookieValue)
+		return FALSE;
+	if (ListDictionary_Contains(context->cookies, CookieName))
+	{
+		char* dupCookieValue = strdup(CookieValue);
+		if (dupCookieValue)
+		{
+			status = ListDictionary_SetItemValue(context->cookies, CookieName, dupCookieValue);
+			if (!status)
+			{
```
akallabeth, line 355: no braces for one liners

libfreerdp/core/gateway/http.c
```diff
@@ -302,12 +340,52 @@ BOOL http_context_set_rdg_auth_scheme(HttpContext* context, const char* RdgAuthS
...
+			}
+		}
+		else
+			return FALSE;
+	}
+	else
+	{
+		char* dupCookieName;
+		char* dupCookieValue;
+		dupCookieName = strdup(CookieName);
```
akallabeth, line 366: `char* dupCookieName = ...`

libfreerdp/core/gateway/http.c
```diff
@@ -302,12 +340,52 @@ BOOL http_context_set_rdg_auth_scheme(HttpContext* context, const char* RdgAuthS
...
+		if (!dupCookieName)
+			return FALSE;
+		dupCookieValue = strdup(CookieValue);
```
akallabeth, line 369: `char* dupCookieValue = ...`

libfreerdp/core/gateway/http.c
```diff
@@ -130,7 +133,14 @@ static BOOL strings_equals_nocase(const void* obj1, const void* obj2)
 
 HttpContext* http_context_new(void)
 {
-	return (HttpContext*)calloc(1, sizeof(HttpContext));
+	HttpContext* context = (HttpContext*)calloc(1, sizeof(HttpContext));
+	if (context)
+	{
+		context->cookies = ListDictionary_New(FALSE);
+		ListDictionary_KeyObject(context->cookies)->fnObjectFree = free;
```
akallabeth, line 140: might be a good idea to also set a new function, so you can remove the `strdup` below

libfreerdp/core/gateway/http.c
```diff
@@ -446,6 +525,47 @@ static BOOL http_encode_authorization_line(wStream* s, const char* AuthScheme,
...
+{
+	ULONG_PTR* keys;
+	int count;
+	BOOL status;
+
+	if (!s && !cookies)
+		return FALSE;
+
+	ListDictionary_Lock(cookies);
+	count = ListDictionary_GetKeys(cookies, &keys);
```
akallabeth, line 538: `const int count = ...`

libfreerdp/core/gateway/http.c
```diff
@@ -823,6 +1010,110 @@ static int print_bio_error(const char* str, size_t len, void* bp)
 	return len;
 }
 
+int http_chuncked_read(BIO* bio, BYTE* pBuffer, size_t size,
+                       http_encoding_chunked_context* encodingContext)
+{
+	int status;
+	int effectiveDataLen = 0;
+	WINPR_ASSERT(encodingContext != NULL);
```
akallabeth, line 1018: `WINPR_ASSERT(bio);`?

libfreerdp/core/gateway/http.c
```diff
@@ -950,60 +1244,112 @@ HttpResponse* http_response_recv(rdpTls* tls, BOOL readContentLength)
 		}
 
 		/* Fetch remaining body! */
-		while (response->BodyLength < bodyLength)
+		if (response->TransferEncoding == TransferEncodingChunked && readContentLength)
```
akallabeth, line 1247: braces around each (non simple bool) condition

libfreerdp/core/gateway/http.c
```diff
@@ -950,60 +1244,112 @@ HttpResponse* http_response_recv(rdpTls* tls, BOOL readContentLength)
...
 		{
-			int status;
-
-			if (!Stream_EnsureRemainingCapacity(response->data, bodyLength - response->BodyLength))
-				goto out_error;
+			char buffer[2048];
```
akallabeth, line 1249: `={0};` to initialize

libfreerdp/core/gateway/http.c
```diff
@@ -950,60 +1244,112 @@ HttpResponse* http_response_recv(rdpTls* tls, BOOL readContentLength)
...
+			http_encoding_chunked_context ctx;
```
akallabeth, line 1250: `={0};` to initialize

libfreerdp/core/gateway/http.c
```diff
@@ -950,60 +1244,112 @@ HttpResponse* http_response_recv(rdpTls* tls, BOOL readContentLength)
...
+					memcpy(Stream_Pointer(response->data), buffer, status);
+					Stream_Seek(response->data, (size_t)status);
+					full_len += status;
+				}
 
-			if (status <= 0)
+			} while (ctx.state != ChunkStateEnd);
+			response->BodyLength = full_len;
+			if (response->BodyLength > 0)
+				response->BodyContent = &(Stream_Buffer(response->data))[payloadOffset];
```
akallabeth, line 1284: is the buffer checked for `payloadOffset` bytes available?

libfreerdp/core/gateway/http.c
```diff
@@ -950,60 +1244,112 @@ HttpResponse* http_response_recv(rdpTls* tls, BOOL readContentLength)
...
+					WLog_ERR(TAG, "Request body too large! (%" PRIdz " bytes) Aborting!",
+					         response->BodyLength);
+					goto out_error;
+				}
 			}
-		}
 
-		if (response->BodyLength > 0)
-			response->BodyContent = &(Stream_Buffer(response->data))[payloadOffset];
+			if (response->BodyLength > 0)
```
akallabeth, line 1324: is the buffer checked for `payloadOffset` bytes available?

libfreerdp/core/gateway/http.c
```diff
@@ -1016,12 +1362,18 @@ HttpResponse* http_response_new(void)
 	if (!response->Authenticates)
 		goto fail;
 
+	response->SetCookie = ListDictionary_New(FALSE);
+
+	if (!response->SetCookie)
```
akallabeth, line 1367: better add a new/free function to this. otherwise you need to guarantee that the cookie string is not stack allocated and the lifecycle the same as this.

akallabeth (changes requested):
Overall LGTM.

some style and check related nitpicking, that's all folks ;)

really a great piece of work, thank you!

libfreerdp/core/gateway/http.c
akarl10: as far as I understood every string in HttpResponse is only valid as long as HttpResponse is around (every string essentially points to the response buffer read over tls/response->data). This is why i allocate/deallocate in the http_context_set_cookie case (because there I need to guarantee this string is valid for the context's lifetime)

libfreerdp/core/gateway/http.c
akarl10: this code is not really mine, but payload offset is taken from a Stream_GetPosition

It may make sense adding a WINPR_ASSERT

akallabeth: @akarl10 ah, ok. did miss that part, then it is already `checked`.

libfreerdp/core/gateway/arm.c
akarl10: actually the "whole" function body is `#ifdef CJSON`
else it will print
`WLog_ERR(TAG, "arm gateway support not compiled in");`
and
immediatly return FALSE

see line 220

akallabeth: oh, indeed, `#else`.
still, might be a good idea to just not compile in the whole code for this if `cJSON` is not available, not just fail here. (but can be added in a separate pr, this one is already huge enough)

libfreerdp/core/gateway/http.c
akarl10: ok, I added `ListDictionary_KeyObject(context->cookies)->fnObjectNew = (OBJECT_NEW_FN)_strdup`

If I undestood it right ListDictionary_SetItemValue/ListDictionary_Add should make a copy of the string.

The compiler does however say `expected ‘void *’ but argument is of type ‘const char *’`
is it safe that I remove the const or are there dedicated function for that case?

akarl10: actually I don't see any use of fnObjectNew in ListDictionary.c.
are you sure this works? only fnObjectFree is used

akallabeth:
@freerdp-bot test please

akarl10:
> @freerdp-bot test please

sorry, too early, found something I broke while refactoring, just preparing a new push (got it working again :) )

akallabeth:
@akarl10 well, that is my luck, always bad timing

libfreerdp/core/gateway/http.c
```diff
@@ -302,12 +340,50 @@ BOOL http_context_set_rdg_auth_scheme(HttpContext* context, const char* RdgAuthS
...
 }
 
+BOOL http_context_set_cookie(HttpContext* context, const char* CookieName, const char* CookieValue)
+{
+	BOOL status;
+	if (!context || !CookieName || !CookieValue)
+		return FALSE;
+	if (ListDictionary_Contains(context->cookies, CookieName))
+	{
+		char* dupCookieValue = strdup(CookieValue);
```
akallabeth, line 350: you can eliminate all these `strdup` things by just providing a `fnObjectNew` for the list key/value objects

akarl10:
regarding the ListDictionary fnObjectNew:

the function does the following:
```c
BOOL ListDictionary_SetItemValue(wListDictionary* listDictionary, const void* key, void* value)
{
	BOOL status = FALSE;
	wListDictionaryItem* item;
	OBJECT_EQUALS_FN keyEquals;

	if (!listDictionary)
		return FALSE;

	if (listDictionary->synchronized)
		EnterCriticalSection(&listDictionary->lock);

	keyEquals = listDictionary->objectKey.fnObjectEquals;

	if (listDictionary->head)
	{
		item = listDictionary->head;

		while (item)
		{
			if (keyEquals(item->key, key))
				break;

			item = item->next;
		}

		if (item)
		{
			if (listDictionary->objectValue.fnObjectFree)
				listDictionary->objectValue.fnObjectFree(item->value);

			item->value = value;
		}

		status = (item) ? TRUE : FALSE;
	}

	if (listDictionary->synchronized)
		LeaveCriticalSection(&listDictionary->lock);

	return status;
}
```
if compared to ArrayList_SetItem
```c
BOOL ArrayList_SetItem(wArrayList* arrayList, size_t index, const void* obj)
{
	WINPR_ASSERT(arrayList);
	if (index >= arrayList->size)
		return FALSE;

	if (arrayList->object.fnObjectNew)
	{
		arrayList->array[index] = arrayList->object.fnObjectNew(obj);
		if (obj && !arrayList->array[index])
			return FALSE;
	}
....
```

In short: ListDictionary is missing the copy operation

akallabeth:
> regarding the ListDictionary fnObjectNew:
> 
> the function does the following:
> 
> ```c
> BOOL ListDictionary_SetItemValue(wListDictionary* listDictionary, const void* key, void* value)
> {
> 	BOOL status = FALSE;
> 	wListDictionaryItem* item;
> 	OBJECT_EQUALS_FN keyEquals;
> 
> 	if (!listDictionary)
> 		return FALSE;
> 
> 	if (listDictionary->synchronized)
> 		EnterCriticalSection(&listDictionary->lock);
> 
> 	keyEquals = listDictionary->objectKey.fnObjectEquals;
> 
> 	if (listDictionary->head)
> 	{
> 		item = listDictionary->head;
> 
> 		while (item)
> 		{
> 			if (keyEquals(item->key, key))
> 				break;
> 
> 			item = item->next;
> 		}
> 
> 		if (item)
> 		{
> 			if (listDictionary->objectValue.fnObjectFree)
> 				listDictionary->objectValue.fnObjectFree(item->value);
> 
> 			item->value = value;
> 		}
> 
> 		status = (item) ? TRUE : FALSE;
> 	}
> 
> 	if (listDictionary->synchronized)
> 		LeaveCriticalSection(&listDictionary->lock);
> 
> 	return status;
> }
> ```
> 
> if compared to ArrayList_SetItem
> 
> ```c
> BOOL ArrayList_SetItem(wArrayList* arrayList, size_t index, const void* obj)
> {
> 	WINPR_ASSERT(arrayList);
> 	if (index >= arrayList->size)
> 		return FALSE;
> 
> 	if (arrayList->object.fnObjectNew)
> 	{
> 		arrayList->array[index] = arrayList->object.fnObjectNew(obj);
> 		if (obj && !arrayList->array[index])
> 			return FALSE;
> 	}
> ....
> ```
> 
> In short: ListDictionary is missing the new operation

yes, but fixed in https://github.com/FreeRDP/FreeRDP/pull/9104

akallabeth:
@akarl10  you can base your branch on top of the pr and already prepare your commits (no need to wait)

akallabeth:
@akarl10 ok, #9104 is merged :)

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,392 @@
...
+	rdpSettings* settings = arm->settings;
+	if (!settings)
+		return FALSE;
+
+	const char* peerHostname = freerdp_settings_get_string(settings, FreeRDP_GatewayHostname);
+	if (!peerHostname)
+		return FALSE;
+
+	UINT16 peerPort = (UINT16)freerdp_settings_get_uint32(
+	    settings, FreeRDP_GatewayPort); // for some reason GatewayPort is a UINT32
```
hardening, line 81: can we remove the comment, not so informative and that would also make clang-format reformat that nicely

libfreerdp/core/gateway/arm.c
```diff
@@ -0,0 +1,392 @@
...
+
+	if ((freerdp_settings_get_uint32(context->settings, FreeRDP_LoadBalanceInfoLength) == 0) ||
+	    (freerdp_settings_get_string(context->settings, FreeRDP_RemoteApplicationProgram) == NULL))
+	{
+		WLog_ERR(TAG, "loadBalanceInfo and RemoteApplicationProgram needed");
+		return FALSE;
+	}
+
+	rdpArm* arm = (rdpArm*)calloc(1, sizeof(rdpArm));
+	char* message = NULL;
```
hardening, line 309: if calloc fails the function still returns true

libfreerdp/core/gateway/arm.c
akarl10: was more for myself..

libfreerdp/core/gateway/http.c
```diff
@@ -823,6 +1007,112 @@ static int print_bio_error(const char* str, size_t len, void* bp)
...
+					encodingContext->nextOffset = 0;
+					encodingContext->state = ChunkStateEnd;
+					return -1;
+				}
+				encodingContext->nextOffset = tmp;
+				encodingContext->state = ChunkStateData;
+
+				if (encodingContext->nextOffset == 0)
+				{ /* end of stream */
+					WLog_DBG(TAG, "cunked encoding end of stream received");
```
hardening, line 1102: typo from our code moved here: `s/cunked/chunked/`

libfreerdp/core/gateway/websocket.c
```diff
@@ -99,6 +100,41 @@ BOOL websocket_write_wstream(BIO* bio, wStream* sPacket, WEBSOCKET_OPCODE opcode
 	return TRUE;
 }
 
+static int websocket_write_all(BIO* bio, const BYTE* data, int length)
```
hardening, line 103: Perhaps I'm wrong but this function looks a lot like `tls_write_all`, in fact it is very generic and has nothing to do with websockets (but IIRC that's also the case for `tls_write_all`).
Also don't we tend to use `size_t` for length parameters ?

akarl10: this is because it is the same function, but with BIO* instead of tls (since I don't have tls around here)
I can use size_t

libfreerdp/core/gateway/wst.c
```diff
@@ -351,18 +381,60 @@ BOOL wst_connect(rdpWst* wst, DWORD timeout)
 
 	switch (StatusCode)
 	{
-		case HTTP_STATUS_NOT_FOUND:
-		case HTTP_STATUS_BAD_METHOD:
+		case HTTP_STATUS_FORBIDDEN:
```
hardening, line 384: Not your code, but IIRC we had questions about "302 Moved" response...

libfreerdp/core/gateway/wst.c
```diff
@@ -641,11 +715,17 @@ static BOOL wst_parse_url(rdpWst* wst, const char* url)
 
 	if (strncmp("wss://", url, 6) != 0)
 	{
-		WLog_ERR(TAG, "Websocket URL is invalid. Only wss:// URLs are supported");
-		return FALSE;
+		if (strncmp("https://", url, 8) != 0)
```
hardening, line 718: Shall we always relax the check for `wss://` or is it AMD specific ?

akarl10: I'll update the message.

hardening:
A few remarks, but very nice contribution. Thank you.

libfreerdp/core/gateway/wst.c
akarl10: really?
If that's really needed I would suggest making a different PR. At least AVD seems to not need that

libfreerdp/core/gateway/wst.c
hardening: My remark was not for the message but for the check itself. Do we want to always relax the check for all websocket things, or just for AMD ?

akallabeth:
@freerdp-bot test please

akallabeth (approved):
LGTM,
really nice work, thank you!

akallabeth:
@freerdp-bot test
@akarl10 pushed  a few fixes I´ve found during (another) review. Also removed the casts you did (the pending PR resolved these automatically). Hope it still works?

libfreerdp/core/gateway/arm.c
akallabeth: no more

akarl10:
hmm, now I get
`[FATAL][com.freerdp.winpr.assert] - [winpr_int_assert]: auth && auth->info [build/libfreerdp/CMakeFiles/freerdp.dir/compiler_depend.ts:credssp_auth_pkg_name:683]` (checking against my fake api endpoint).

will have a look why this happens

akallabeth:
@akarl10 I´ll merge this now, so if anything is remaining we can handle that in a separate pr.

akallabeth:
@akarl10 I´ve moved the `auth` initialization, so maybe that is the reason?

akarl10:
I think the reason is because now you always create a auth context, not like before where it was created just in the 401 case

akarl10:
ok, moving `credssp_auth_new` into wst_handle_denied fixes it.

should I make a new push to this or do you want to merge and make  a new pr (note that this pr will crash with the message above)

akallabeth:
@akarl10 did push a fix for that.

akallabeth:
@freerdp-bot test please

akarl10:
@akallabeth ok, this version works

PR #9104, opened 2023-06-27
Collections refactor

akallabeth:
make some more collections utilize opaque context


mfleisz (commented):
Overall LGTM but I don't think it is a good idea to remove all the `NULL` checks and replace them with `ASSERTS` in the collection functions.
`ASSERT` should verify internal state and these are generic, public APIs which should not crash if someone from the outside passes invalid (`NULL`) arguments into them.

akallabeth:
@mfleisz the only asserts are for required arguments (e.g. the `instance` of the collection) which also crashes most public API of other projects. Calling functions with such arguments is a bug and that should fail early.

mfleisz (approved):
@akallabeth I get your point but IMO either make an APIs safe or document how to use it - here we have neither. Anyways like I said everything else LGTM

akallabeth:
@freerdp-bot test please

PR #9109, opened 2023-06-28
[core,gateway] only encode a cookie line if there is a cookie

akarl10:
if cookie length was 0 a empty line was added and this would cut off the http header and move following headers in the body


akallabeth:
@freerdp-bot test please

akallabeth (approved):
LGTM

akarl10:
guess libpkcs11-helper1-dev is missing on the "fuzzer"

PR #9111, opened 2023-06-28
[core,gateway] make http authentication work again for websocket transport

akarl10:
the function wst_handle_denied never returned TRUE, even if credssp_auth_is_complete finished


akallabeth (approved):
LGTM

akallabeth:
@akarl10 ouch, that one is on me. nice catch!

PR #9112, opened 2023-06-28
Ini file fixes

akallabeth:


winpr/include/winpr/ini.h
```diff
@@ -30,27 +30,123 @@ extern "C"
...
+	 *
+	 *  @return > 0 for success, < 0 for failure
+	 */
 	WINPR_API int IniFile_ReadFile(wIniFile* ini, const char* filename);
 
+	/** @brief write an ini instance to a buffer
+	 *
+	 *  @param ini The instance to use, must not be \b NULL
+	 *
+	 *  @return A newly allocates string, use \b free after use. \b NULL in case of failure
```
mfleisz, line 55: allocate~~s~~d string

winpr/libwinpr/utils/ini.c
```diff
@@ -304,35 +322,45 @@ static wIniFileSection* IniFile_GetSection(wIniFile* ini, const char* name)
 	return section;
 }
 
+static BOOL IniFile_SectionResize(wIniFile* ini, size_t count)
+{
+	WINPR_ASSERT(ini);
+
+	if (ini->nSections + count >= ini->cSections)
+	{
+		const size_t new_size = ini->cSections + count + 1024;
```
mfleisz, line 331: Why is the `+ 1024` (same in `IniFile_SectionKeysResize`)?
It seems we are always carrying 1024 additional empty slots around with us

mfleisz (approved):
2 minor comments - otherwise lgtm

winpr/libwinpr/utils/ini.c
akallabeth: block allocation, always extend by `1024` if we reach the current limit

PR #9113, opened 2023-06-28
Proxy module fixes

akallabeth:
* Fix missing return value checks in proxy (abort if something fails)
* Fix missing `NULL` checks in `dyn-channel-dump`
* change logging, dump each packet into a separate file (use a counter per channel for both, incoming and outgoing to have the original order of traffic)
* add session handling to dump module
* fix an issue with the `drdynvc` state tracker

### DIFF
--- a/server/proxy/channels/pf_channel_drdynvc.c
+++ b/server/proxy/channels/pf_channel_drdynvc.c
@@ -87,6 +87,21 @@ typedef enum
 	DYNCVC_READ_INCOMPLETE /*!< missing bytes to read the complete packet */
 } DynvcReadResult;
 
+static const char* openstatus2str(PfDynChannelOpenStatus status)
+{
+	switch (status)
+	{
+		case CHANNEL_OPENSTATE_WAITING_OPEN_STATUS:
+			return "CHANNEL_OPENSTATE_WAITING_OPEN_STATUS";
+		case CHANNEL_OPENSTATE_CLOSED:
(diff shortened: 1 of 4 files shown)


server/proxy/channels/pf_channel_drdynvc.c
```diff
@@ -409,6 +439,7 @@ static PfChannelResult DynvcTrackerPeekFn(ChannelStateTracker* tracker, BOOL fir
 			         direction);
 			channelTracker_setMode(tracker, CHANNEL_TRACKER_PASS);
 			dynChannel->openStatus = CHANNEL_OPENSTATE_CLOSED;
+			HashTable_Remove(dynChannelContext->channels, &dynChannel->channelId);
```
hardening, line 442: With this the CLOSE_ACK (CLOSE response) will be dropped by the proxy. Shall we remove the channel only when we receive the CLOSE ack, or after a timeout ?

akallabeth: @hardening changed the behaviour:
1. `HashTable_Remove` is now not called on close or open failure (the tracker state remains in the list)
2. If a channel open encounters a already allocated state tracker, a message is printed and the previous state removed

this way we `CLOSE_ACK` will work and the overhead is just some allocated memory until we reuse the channel.

hardening (approved):
Looks ok

PR #9114, opened 2023-06-28
[core,gcc] gcc_read_client_data_blocks update checks

akallabeth:
* Update checks for stream consumed
* Update log messages, use stringified version of block type


hardening (approved):
Looks ok

PR #9115, opened 2023-06-29
[winpr,wlog] extend wLog to allow logging custom data

akallabeth:
A logger allocated with WLog_Get now can have a customized context by setting it with WLog_SetContext. Along with the newly introduced format specifier %ctx an additional field is printed for the context.

### DIFF
--- a/winpr/libwinpr/utils/wlog/Layout.c
+++ b/winpr/libwinpr/utils/wlog/Layout.c
@@ -38,6 +38,16 @@
 #include <sys/syscall.h>
 #endif
 
+struct format_option
+{
+	const char* fmt;
+	size_t fmtlen;
+	const char* replace;
+	size_t replacelen;
+	const char* (*fkt)(void*);
+	void* arg;
+};
+
 /**
  * Log Layout
  */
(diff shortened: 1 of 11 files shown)


winpr/libwinpr/utils/wlog/Layout.c
```diff
@@ -57,201 +57,137 @@ static void WLog_PrintMessagePrefix(wLog* log, wLogMessage* message, const char*
...
+	struct format_option
+	{
+		const char* fmt;
+		const char* replace;
+		const char* (*fkt)(void*);
+		const void* arg;
+	};
+
+	const struct format_option options[] = {
+		{ "%ctx", "%s", log->custom, log->context },                    /* log context */
```
hardening, line 131: In this workflow, what about searching for `%` in `p`, and then trying to match options. Doing this we avoid scanning the beginning of the string. 
For example if prefix is "start of string %ml", then we will browse the _whole_ prefix numerous times (passing over `start of string ` with `strstr()`) until we match the `%ml` option which is the last one.
We may also rearrange the options so that the most commonly used are present first (if that's not already the case).

hardening (approved):
Except my remark about perfs, looks ok for me

winpr/libwinpr/utils/wlog/Layout.c
akallabeth: @hardening ok, implemented `bsearch` for option comparison. this should reduce the number of required comparisons drastically.

akallabeth:

PR #9117, opened 2023-06-29
[channels] modify types for plugin entries

akallabeth:
* Change return type to uintptr_t to ensure it can hold all return values possible (e.g. UINT and data pointer)


hardening:
Can you elaborate on the need for such change ? Do you want to be able to return a pointer to some error context ?

akallabeth:
@hardening we did return data pointers for the printer subsystem, see `win_freerdp_printer_client_subsystem_entry` for example.
so, while using a return code of `UINT` works for most, it is not guaranteed that it has the necessary space to hold a pointer

pnowack:
Why not directly using a pointer (even if it is a `void *`)?

akallabeth:
@pnowack `uintptr_t` gives less compiler warnings and can represent both. with `void*` you get `clang` warnings (`-Weverything`) if you cast to an integer. 

for `void* a=0x1234; uintptr_t b=(uintptr_t)a;` you get
`warning: cast from function call of type 'void *' to non-matching type 'uintptr_t' (aka 'unsigned long') [-Wbad-function-cast]`

whereas `uintptr_t a=0x1234; void* b = a; ` you don´t get warnings.

pnowack:
> for void* a=0x1234; uintptr_t b=(uintptr_t)a; you get
warning: cast from function call of type 'void *' to non-matching type 'uintptr_t' (aka 'unsigned long') [-Wbad-function-cast]

hmm...okay

akallabeth:
fixed by changing usage in printer channel, see #9216

PR #9126, opened 2023-07-01
[core,gateway,wst] fix use after free in wst_connect

akarl10:
http_response_is_websocket was called after http_response_free


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test please
@akarl10 nice catch!

PR #9073, opened 2023-06-20
[client,sdl] move webview to static lib

akallabeth:
* move to static lib aad-view
* add support for https://github.com/webview/webview (e.g. use native webview if available)

Remaining issues tracked in #9174

PR #9127, opened 2023-07-02
Improve Azure AD communication

fifthdegree:


CMakeLists.txt
```diff
@@ -741,7 +745,11 @@ find_feature(Pulse ${PULSE_FEATURE_TYPE} ${PULSE_FEATURE_PURPOSE} ${PULSE_FEATUR
 find_feature(Cups ${CUPS_FEATURE_TYPE} ${CUPS_FEATURE_PURPOSE} ${CUPS_FEATURE_DESCRIPTION})
 find_feature(PCSC ${PCSC_FEATURE_TYPE} ${PCSC_FEATURE_PURPOSE} ${PCSC_FEATURE_DESCRIPTION})
 
-find_feature(CJSON ${CJSON_FEATURE_TYPE} ${CJSON_FEATURE_PURPOSE} ${CJSON_FEATURE_DESCRIPTION})
+option(WITH_AAD "Compile with support for Azure AD authentication" ON)
+if (WITH_AAD)
+	find_feature(CJSON REQUIRED ${CJSON_FEATURE_PURPOSE} ${CJSON_FEATURE_DESCRIPTION})
```
akallabeth, line 750: better switch to [find_package](https://cmake.org/cmake/help/latest/command/find_package.html)

CMakeLists.txt
```diff
@@ -741,7 +745,11 @@ find_feature(Pulse ${PULSE_FEATURE_TYPE} ${PULSE_FEATURE_PURPOSE} ${PULSE_FEATUR
...
+	find_feature(Curl REQUIRED ${CURL_FEATURE_PURPOSE} ${CURL_FEATURE_DESCRIPTION})
```
akallabeth, line 751: better switch to [find_package](https://cmake.org/cmake/help/latest/command/find_package.html)

client/SDL/sdl_webview.cpp
```diff
@@ -66,35 +66,14 @@ class SchemeHandler : public QWebEngineUrlSchemeHandler
 	std::string m_code;
 };
 
-BOOL sdl_webview_get_aad_auth_code(freerdp* instance, const char* hostname, char** code,
-                                   const char** client_id, const char** redirect_uri)
+std::string sdl_webview_get_auth_code(QString url)
```
akallabeth, line 69: shouldn´ t this function be `static`?

akallabeth (commented):
* I´ve been working on getting webview to support more backends aside from `Qt` with https://github.com/FreeRDP/FreeRDP/pull/9073 (not quite finished yet for platforms other than linux)
* introducing the `curl` dependency might better be optional (the suggestion I made in matrix was to replace the whole http parser in freerdp, not just here ;))

hardening:
>     * introducing the `curl` dependency might better be optional (the suggestion I made in matrix was to replace the whole http parser in freerdp, not just here ;))

I guess that `libcurl` is able to handle the specific things of rdg protocol (like the almost never ending request) when not using the "easy" interface.
I'm seeing that libcurl-ws could also do websocket for us...

akarl10:
> > ```
> > * introducing the `curl` dependency might better be optional (the suggestion I made in matrix was to replace the whole http parser in freerdp, not just here ;))
> > ```
> 
> I guess that `libcurl` is able to handle the specific things of rdg protocol (like the almost never ending request) when not using the "easy" interface. I'm seeing that libcurl-ws could also do websocket for us...

we would first make sure one thing:
How a underlying socket can be forced for a specific request.
This will be needed for Negotiate/NTLM authentication since this type of requests are not (like everthing else in HTTP) stateless. libcurl would support NTLM and Negotiate, but I think FreeRDP's impementation is better suited and I think we don't want different behaviour in this case.

Also for example RDG_OUT_DATA first needs to get a 200 OK with a empty body on the underlying connection before the actual RDG_OUT_DATA is sent on the same connection (except a 101 is returned, then a websocket connection is used -> this is opportunistic, don't know if that is easy to do in libcurl-ws)

akallabeth:
@hardening @akarl10 that sounds like we should split this pr.
* the changes regarding `libcurl` might be ok or not, but it looks like there needs to be some agreement on that first
* the callbacks look good (although it could be argued to refactor the initial AAD callback so it can handle both types)

akarl10:
@akallabeth just to make that clear: I'm not against libcurl in the gateway case, I just wanted to say that it might be harder then we expect in that case. For the cases where we can use the "easy" api, like token fetching and the arm/connections/ endpoint libcurl for sure is the better option.

The idea @fifthdegree had that if you use AAD you need cJSON and libcurl is not that bad I think (that is porting aad.c and probably arm.c to libcurl).

The second step then might really be porting other things in gateway/ to libcurl, and when all is ported remove http.c/h, and websocket.c/h

one thing that we might not forget: /tls:seclevel:0 might be required for gateway server < 2012, so we probably need to set such a option also for libcurl

akallabeth:
@akarl10 I got that, but it might take some time to implement. adding `libcurl` in between (while still using the custom `http` and `websockets` parsing) might not be the best approach.
so, I´m just suggesting that we postpone this until we have resolved the questions surrounding `libcurl` before adding it as a dependency.
(side note, `seclecel` related stuff is mostly `ssl` related, but you might be right that `curl` also tries to do stuff there.

hardening:
We might find equivalent for SecLevel in libcurl

fifthdegree:
I initially intended to change the gateway over to libcurl too, but haven't found the time for that yet. We will need to provide a callback function to openssl for certificate verification and to compute SecBindings. The curl docs label their websocket api as experimental; I'm not sure if that means the api might still change

akarl10:
just a quick note on libcurl ws support: I think ubuntu (maybe debian too) have not enabled that in the delivered libcurl. Sure, we might compile against a non system libcurl, but I think this would complicate things considerably

akallabeth:
@fifthdegree where are we at for this pr?

akallabeth:
@fifthdegree hmm, we are thinking about releasing a 3.0-beta and it might be a nice thing to have in there and introduce the `curl` rewrite later on?

fifthdegree:
OK, took out the curl stuff and added a very simple function under utils for doing standard http requests. Then we can just use curl there for all standard requests.

client/SDL/sdl_webview.hpp
```diff
@@ -26,8 +26,9 @@ extern "C"
 {
 #endif
 
-	BOOL sdl_webview_get_aad_auth_code(freerdp* instance, const char* hostname, char** code,
-	                                   const char** client_id, const char** redirect_uri);
+	BOOL sdl_webview_get_avd_access_token(freerdp* instance, char** token);
```
akallabeth, line 29: Maybe a good idea to unify these.
something like `GetAccessToken(freerdp* instance, int tokenType, char** token, ...);` might do the trick.

libfreerdp/utils/http.c
```diff
@@ -0,0 +1,231 @@
...
+	BIO* bio = NULL;
+	SSL_CTX* ssl_ctx = NULL;
+
+	WINPR_ASSERT(status_code);
+	WINPR_ASSERT(response);
+	WINPR_ASSERT(response_length);
+
+	*response = NULL;
+
+	if (!url || strlen(url) < 8 || strncmp(url, "https://", 8) != 0 ||
```
akallabeth, line 69: might use `strnlen(url, 8)` here as we are not interested in full length.

libfreerdp/utils/http.c
```diff
@@ -0,0 +1,231 @@
...
+	    !(path = strchr(url + 8, '/')))
+	{
+		WLog_ERR(TAG, "invalid url provided");
+		goto out;
+	}
+
+	hostname = calloc(1, path - (url + 8) + 1);
```
akallabeth, line 76: `strdup(path)`?

libfreerdp/utils/http.c
```diff
@@ -0,0 +1,231 @@
...
+				WLog_ERR(TAG, "could not parse content length (%s): %s [%d]", val, strerror(errno),
+				         errno);
+				goto out;
+			}
+		}
+	} while (strcmp(buffer, "\r\n") != 0);
+
+	if (*response_length > 0)
+	{
+		*response = calloc(1, *response_length + 1);
```
akallabeth, line 192: might move the allocation after checking validity of length

akallabeth (changes requested):
nicely done.
A few small remarks still, but overall LGTM

akallabeth:
@freerdp-bot test please

akallabeth:
@fifthdegree will fix these issues myself, so don´t bother ;)

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test please

akallabeth:

akallabeth:
@freerdp-bot test please

akallabeth:

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot and again you test please

akallabeth:

akallabeth:

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test

akallabeth:

PR #9129, opened 2023-07-03
core: Fix flag check to enable/disable RDSTLS security

mfleisz:
On some platforms (i.e. Mac) the `BOOL` type might be narrower than the `RedirectionFlags` type leading to an overflow when just assigning the logical AND result to the setting. This fix makes sure that the value is correctly set to either `TRUE` or `FALSE`.

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -661,7 +661,8 @@ BOOL rdp_client_redirect(rdpRdp* rdp)
 			return FALSE;
 	}
 
-	settings->RdstlsSecurity = settings->RedirectionFlags & LB_PASSWORD_IS_PK_ENCRYPTED;
+	settings->RdstlsSecurity =
+	    (settings->RedirectionFlags & LB_PASSWORD_IS_PK_ENCRYPTED) != 0 ? TRUE : FALSE;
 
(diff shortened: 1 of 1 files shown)


libfreerdp/core/connection.c
```diff
@@ -661,7 +661,8 @@ BOOL rdp_client_redirect(rdpRdp* rdp)
 			return FALSE;
 	}
 
-	settings->RdstlsSecurity = settings->RedirectionFlags & LB_PASSWORD_IS_PK_ENCRYPTED;
+	settings->RdstlsSecurity =
+	    (settings->RedirectionFlags & LB_PASSWORD_IS_PK_ENCRYPTED) != 0 ? TRUE : FALSE;
```
hardening, line 665: Could also be `!!(settings->RedirectionFlags & LB_PASSWORD_IS_PK_ENCRYPTED)` like they do in `glib` and related projects...

hardening (approved):

hardening:
Don't we have a pattern and we probably have some other similar bugs ?

mfleisz:
@hardening I checked this file and `redirection.c` but will have a look at the others too - could be that there are similar errors lurking somewhere ...

PR #9132, opened 2023-07-04
[client,file] refactor RDP file parser

akallabeth:
* Only print warnings for settings parsed but support not compiled in
* Use constants for key values


akallabeth:

mfleisz (approved):

PR #9133, opened 2023-07-04
misc: More int to BOOL conversion fixes

mfleisz:
This is a follow up to #9129.

This PR fixes some problematic `int` to `BOOL` conversions that might cause overflows when checking for bit flags.


akallabeth (approved):
LGTM

PR #9134, opened 2023-07-04
Proxy mod

akallabeth:


akallabeth:

mfleisz:
@freerdp-bot test

mfleisz:
@freerdp-bot test

winpr/libwinpr/utils/collections/ListDictionary.c
```diff
@@ -209,7 +209,15 @@ static void item_set(wListDictionary* listDictionary, wListDictionaryItem* item,
 	if (listDictionary->objectValue.fnObjectNew)
 		item->value = listDictionary->objectValue.fnObjectNew(value);
 	else
-		item->value = value;
+	{
+		union
```
mfleisz, line 213: Why are we using this `union` stuff again when it's sufficient to use a simple cast?

akallabeth: these cast `const` away still gives warnings `-Wcast_qual` if only done like `item->value = (void*)value;`

mfleisz: Reading https://gcc.gnu.org/onlinedocs/gcc/gcc-command-options/options-to-request-or-suppress-warnings.html#cmdoption-Wcast-qual this switch warns in exactly the above case. The `union` just tricks the compiler and actually allows the same "unsafe" code as outlined in the example there.

Bottom line:
Do we want to have "safe" code - then we need to fix the typing (which will be difficult for generic containers in C).
Do we want to use these "unsafe" casts - use a simple cast i.e. `(void*)(uintptr_t)value` and/or tell the compiler not to bugger us (`-Wno-cast-qual`).

akallabeth: @mfleisz yes, and instead of using compiler dependent pragma to disable the warning for this part of code (which we know is ok) the union does the same. disabling compiler warnings altogether limits usefulness of output.

mfleisz (approved):

PR #9138, opened 2023-07-05
[core,gateway] cleanup http response

akarl10:
also if http_context_enable_websocket_upgrade was called multiple times it could memory leak previos SecWebsocketKey values


mfleisz:
@freerdp-bot test

akallabeth (approved):

PR #9143, opened 2023-07-05
[utils] use select instead of poll to read passphrases on macOS

bmiklautz:
According to the man page on macOS poll currently doesn't support devices. This includes `/dev/tty`.
In this use case poll will returned immediately indicating that something can be read but the following read returned `-1` (with errno set to EAGAIN).

Using select on macOS for passphrase reading prevents this problem.


hardening (approved):

akallabeth:

PR #9144, opened 2023-07-06
[winpr,wlog] added %% formatter and skip NULL %ctx

akallabeth:
* New formatter %%xxx%% which conditionally prints xxx if the context is not NULL
* Skip %ctx formatter if context is NULL
* add logger context for `rdpRdp::log`


winpr/libwinpr/utils/wlog/Layout.c
```diff
@@ -123,166 +129,162 @@ static int opt_compare_fn(const void* a, const void* b)
...
+	char tid[32] = { 0 };
+	SYSTEMTIME localTime = { 0 };
+	GetLocalTime(&localTime);
+
+	struct format_option_recurse recurse = {
+		.options = NULL, .nroptions = 0, .log = log, .layout = layout, .message = message
+	};
+
+#define ENTRY(x) x, sizeof(x) - 1
+	const struct format_option options[] = {
```
hardening, line 249: may I suggest to put the array out of the function, this way we have 1 ident less and possibly most of the entries on one line even after the `clang-format` pass ?

hardening:
Seems ok for me (IMHO the `%{` syntax is much better).
It's quite a long time that this part of wLog has gained a lot of functionalities, but no unitary tests. Isn't it time to add unitary tests for all these fancy features ?

winpr/libwinpr/utils/wlog/Layout.c
```diff
@@ -38,6 +38,12 @@
 #include <sys/syscall.h>
 #endif
 
+#ifndef MIN
```
hardening, line 41: Remark that is probably out of the scope of this PR: `MAX` is defined in the FreeRDP part (not winPR unfortunately). But I noticed that we have various occurrence of `MIN` definitions in winPR.

hardening (approved):
LGTM

winpr/libwinpr/utils/wlog/Layout.c
akallabeth: no as the struct contains a by call context

winpr/libwinpr/utils/wlog/Layout.c
akallabeth: yes, that is a todo.

PR #9145, opened 2023-07-06
core: Fix issues in settings copy code

mfleisz:
This code fixes an issue where some settings were lost when copying settings around. I.e. the device count was first set but then calling `freerdp_settings_set_pointer_len` caused the device count to be reset to 0 because we passed `NULL` for the data value.

This means the count got lost upon the first copy and further copies also resulted in losing the array data (because the count was set to 0).

This PR fixes this issue by first resetting the array and afterwards setting the correct count value.


akallabeth (approved):
LGTM

PR #9156, opened 2023-07-12
[core,nla] print name of error code in log

akallabeth:


akallabeth:
@freerdp-bot test properly!

PR #9167, opened 2023-07-18
fix the issue of incorrect display of remote desktop cursor

2fly2:
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setcursor

"If your application must set the cursor while it is in a window, make sure the class cursor for the specified window's class is set to NULL. If the class cursor is not NULL, the system restores the class cursor each time the mouse is moved."


akallabeth (approved):
LGTM

PR #9172, opened 2023-07-20
X11/cliprdr: Rework server to client clipboard handling

pnowack:
This gets rid of the multiple lock issue reported in https://github.com/FreeRDP/FreeRDP/issues/8846.

Quoting the main commit here:
```
The purpose of clipboard data locking is to make the other peer
retaining the current file list until a pending paste operation is done,
even though the clipboard selection changed.
As it may be difficult to determine, when a lock is needed, imitate the
same behaviour as mstsc:
When the server side supports clipboard data locking, always attempt to
lock the file list on the server regardless of what is advertised in a
FormatList PDU.
The Lock Clipboard Data PDU can even be already sent, before the
Format List Response PDU is sent.
This is also what mstsc, does: First, lock the new (potential) file
list, then unlock the file list, when the pending paste operation is
done.
So, rework the current clipboard implementation in that direction.

Since the implementation for timeouts for old file lists is a bit hard,
for now always force unlock pending locks, when the selection changes.
However, timeouts for old file lists can still be added in the future.

The reworked clipboard handling is done with the help of three hash
tables:

1. The inode table: This hash table manages all inodes for each file.
   The keys in this table are the inodes themselves, while the values
   the files and directories and their attributes (file size, last write
   time, etc.).
2. The clipdata table: This table manages the locks for each file list.
   The keys in this table represent the clip data id and the values the
   clip data entries, which have a reference to the clip data dir, a
   directory containing the whole selection, and some helper attributes,
   like the clip data id itself.
3. The request table: Every file size or file range request is managed
   here. When a FileContentsRequest is made, its stream id with the
   respective details are added to this table. When a response is
   received, these details can then be easily looked up here.
```

Tested with g-r-d and MS Windows RDS on Win 11.

Closes: https://github.com/FreeRDP/FreeRDP/issues/8846

### DIFF
--- a/channels/cliprdr/cliprdr_common.c
+++ b/channels/cliprdr/cliprdr_common.c
@@ -138,7 +138,7 @@ cliprdr_packet_unlock_clipdata_new(const CLIPRDR_UNLOCK_CLIPBOARD_DATA* unlockCl
 	if (!unlockClipboardData)
 		return NULL;
 
-	s = cliprdr_packet_new(CB_LOCK_CLIPDATA, 0, 4);
+	s = cliprdr_packet_new(CB_UNLOCK_CLIPDATA, 0, 4);
 
 	if (!s)
 		return NULL;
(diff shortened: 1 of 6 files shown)


pnowack:
Overall, the structure is very similar to what I did in g-r-d (minus the missing timeout part (can be added later)).

akallabeth (approved):
LGTM, nice work!

PR #9173, opened 2023-07-20
[util,http] set sni on the ssl connection

akarl10:
the new http util does not set the tls sni header.

This is not a big issue, but the firewall at our company actually does sni inspection and will not allow connections without anymore


akallabeth:
@freerdp-bot test please

akallabeth (approved):
LGTM
thank you for the quick fix.

akallabeth:
pushed the commit to #9176 as without the dep the ci is failing.

PR #9192, opened 2023-07-24
[client,aad] restore client_id to 5177bc73-fd99-4c77-a90c-76844c9b6999

akarl10:
somehow the rdsaad client_id was changed to a85cf173-4192-42f8-81fa-777a763e6e2c. Webviews still use 5177bc73-fd99-4c77-a90c-76844c9b6999

@fifthdegree could you confirm if this can cause INVALID_TOKEN errors (Authentication Error -2146893048)

### DIFF
--- a/client/common/client.c
+++ b/client/common/client.c
@@ -991,7 +991,7 @@ static BOOL client_cli_get_rdsaad_access_token(freerdp* instance, const char* sc
 	size_t size = 0;
 	char* url = NULL;
 	char* token_request = NULL;
-	const char* client_id = "a85cf173-4192-42f8-81fa-777a763e6e2c";
+	const char* client_id = "5177bc73-fd99-4c77-a90c-76844c9b6999";
 	const char* redirect_uri =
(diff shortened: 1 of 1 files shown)


fifthdegree:
No, don't do that. The reason for the different client_ids is because one has the "ms-appx-web://" redirect and the other has the "native_client" redirect; neither has both.

akarl10:
@fifthdegree in SDL/sdl_webview.cpp are used both and both have a ms-appx-web redirect. Is that wrong? Do you have a way to check if the sdl and cli variants work?

Do you remember if you have seen Authentication Error -2146893048 and when it happened?
What also changed is that the server name is taken from "redirectedServerName" in the arm case. What error do you get if the name is wrong?

akallabeth:
@akarl10 with https://github.com/FreeRDP/FreeRDP/pull/9194 you can switch between console and webview with `+force-console-callbacks`

fifthdegree:
Oh, you know what, I take that back. The a8... client does list both types of redirects. Either way, if the client id was incorrect, azure would refuse to grant an access token at all

PR #9198, opened 2023-07-25
aad: Convert JSON number to integer before checking/logging

mfleisz:
floating compares against 0.0 might fail due to conversion/rounding errors. Also, rather log the 32-bit hex return value than a floating point value.


fifthdegree:
You beat me to it!

akallabeth:
closed in favour of #9199 (I like strings in error messages ;))

PR #9205, opened 2023-07-26
gateway: Allow authentication using a NULL (current user) identity

mfleisz:
The PR also fixes a wrong return value in `utils_authenticate_gateway`


akallabeth (approved):
LGTM

PR #9207, opened 2023-07-26
[cmake] build using external lodepng

akallabeth:
* remove shipped `lodepng` sources
* deactivate support by default (currently unused anyway)


akallabeth:
@freerdp-bot test please

PR #9209, opened 2023-07-26
[client,common] fix leak in rdp parser

akallabeth:


akallabeth:
@freerdp-bot test please

PR #9210, opened 2023-07-26
[client,sdl] set gui consent message

akallabeth:


akallabeth:
@freerdp-bot test please

PR #9211, opened 2023-07-27
[standard] replace __FUNCTION__ with __func__

akallabeth:
we finally updated our ci, so now we can change this.

* replaces #8849
* replaces every use of `__FUNCTION__`


akallabeth:

akallabeth:

channels/audin/server/audin.c
```diff
@@ -142,16 +142,30 @@ static UINT audin_server_recv_formats(audin_server_context* context, wStream* s,
 		WLog_Print(audin->log, WLOG_WARN,
 		           "cbSizeFormatsPacket is invalid! Expected: %u Got: %zu. Fixing size",
 		           pdu.cbSizeFormatsPacket, Stream_GetPosition(s));
-		pdu.cbSizeFormatsPacket = Stream_GetPosition(s);
+		const size_t pos = Stream_GetPosition(s);
+		if (pos > UINT32_MAX)
+		{
+			error = ERROR_INVALID_PARAMETER;
```
pnowack, line 148: This looks weird, because the same is not applied to the previous condition here (`pdu.cbSizeFormatsPacket != Stream_GetPosition(s)`).

In any case, it is not a fatal error to hit the `pdu.cbSizeFormatsPacket != Stream_GetPosition(s)` condition. The `cbSizeFormatsPacket` value is more for convenience purpose, since we can derive this value here too.
So, this should probably incorporate the warning message too.

channels/audin/server/audin.c
```diff
@@ -142,16 +142,30 @@ static UINT audin_server_recv_formats(audin_server_context* context, wStream* s,
...
+			goto fail;
+		}
+		pdu.cbSizeFormatsPacket = (UINT32)pos;
 	}
 
-	pdu.ExtraDataSize = Stream_GetRemainingLength(s);
+	const size_t rem = Stream_GetRemainingLength(s);
+
+	if (rem > UINT32_MAX)
```
pnowack, line 156: That is not correct. The size of the `ExtraData` field is arbitrary. It is not capped by some limit.

channels/audin/server/audin.c
```diff
@@ -559,8 +573,12 @@ static UINT audin_server_packet_send(audin_server_context* context, wStream* s)
 	WINPR_ASSERT(context);
 	WINPR_ASSERT(s);
 
-	if (!WTSVirtualChannelWrite(audin->audin_channel, (PCHAR)Stream_Buffer(s),
-	                            Stream_GetPosition(s), &written))
+	const size_t pos = Stream_GetPosition(s);
+	if (pos > UINT32_MAX)
+		return ERROR_INVALID_PARAMETER;
```
pnowack, line 578: This should probably output a warning message like with the other cases below (something like "Oversized DVC Data message").

libfreerdp/codec/audio.c
```diff
@@ -38,7 +38,7 @@ UINT32 audio_format_compute_time_length(const AUDIO_FORMAT* format, size_t size)
 
 	if (format->wBitsPerSample)
 	{
-		const size_t samples = = (size * 8) / format->wBitsPerSample;
+		const size_t samples = (size * 8) / format->wBitsPerSample;
```
pnowack, line 41: This fixup should be in its previous commit.

pnowack (approved):
Just some minor things

libfreerdp/codec/audio.c
akallabeth: done

channels/audin/server/audin.c
akallabeth: done. also for smartcard the same.

akallabeth:
@freerdp-bot test please

channels/audin/server/audin.c
akallabeth: added a log

PR #9216, opened 2023-07-28
[channels,printer] fix backend load function

akallabeth:
changed the signature of the backend loader function. the previous version casted the allocated pointer to UINT which might not be able to hold a pointer.


akallabeth:

mfleisz (approved):

PR #9218, opened 2023-07-28
[winpr,stream] replace defines with inline functions

akallabeth:


akallabeth:
@freerdp-bot test please

PR #9230, opened 2023-07-31
Shadow error check

akallabeth:
this pr fixes various issues with shadow server

1. allow building `rdtk` without `lodepng`  (buildin copy was removed with #9207 )
2. fix missing return value checks in shadow server 
3. fix thread handling for server side rdpgfx channel


pnowack:
> If done within the channel race conditions might occur with
capability announcement and graphics updates.

When would such situation be the case? The client can potentially send the `CapsAdvertise` PDU any time (again) (when the initial caps exchange was with version >= 103 that is also protocol conform). I think this mixes up here actual implementation and protocol parsing, since this way (receiving all messages in a channel thread and then calling the respective callbacks in that same thread) is also done in the other virtual channels here (cliprdr, echo, rdpsnd, audin, rdpecam, etc.).

pnowack:
Actually, don't we have an extra Initialize method in some channels, where an external thread could be used, but does not have to? Maybe we can use the same behaviour here too.

akallabeth:
@pnowack it is the case for the first capability confirm.
the state machine is already running and the messages for caps confirm and reset graphics might get reordered as the messages are handled on different threads in shadow server.

as for using the external thread, that might be possible, but as long as the internal thread is broken it does not make sense to do so

pnowack:
> the state machine is already running and the messages for caps confirm and reset graphics might get reordered as the messages are handled on different threads in shadow server.

But that would be a thing in the shadow server. It needs to wait for the first `CapsAdvertise` message first (I do the same in g-r-d). I think a better solution would be to use the same handling (in this PR), as in `camera_device_enumerator_main.c`. That way, we can keep the current handling and also allow the shadow server to integrate the rdpgfx handling into another thread.

While the internal thread seems to be problematic for the shadow server here, it is not problematic with every implementation. I would like to keep the channel implementations in g-r-d similar in the way of the message handling.

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -1361,16 +1361,25 @@ static UINT rdpgfx_server_receive_pdu(RdpgfxServerContext* context, wStream* s)
...
 	DWORD nCount = 0;
 	HANDLE events[8] = { 0 };
 	UINT error = CHANNEL_RC_OK;
 
-	events[nCount++] = priv->stopEvent;
+	WINPR_ASSERT(priv);
+	WINPR_ASSERT(priv->stopEvent);
+	WINPR_ASSERT(priv->channelEvent);
+
+	if (priv->ownThread)
```
pnowack, line 1381: Maybe this could be an assertion here (since `rdpgfx_server_thread_func` would normally not run if `ownThread` is `FALSE` in the first place)?

akallabeth: ok, updated. also refactored the thread setting to be compatible with other channels.

akallabeth:

pnowack:
I cannot tell much about the rdtk parts, but the rest looks good to me. Also did a quick test (g-r-d against FreeRDP3 with this PR) and things worked fine.

The shadow server seems to skip updates until the graphics pipeline was initialized. What I wonder is whether this has an effect of whether updates may be missing. To describe this better, assume the following situation:
1. The shadow server has a frame ready. It will skip the update, because gfx is not ready yet.
2. gfx is initialized (`CapsConfirm` is sent)
3. The shadow server has no new frame ready.

In this situation here (point 3), will the screen remain black until something new happens on the screen, or will it pick up the previous skipped updated?

akallabeth:
@pnowack ok, good. as for shadow, a newly connected client will trigger updates. I´ve tested with `Xephyr && xterm && metacity`  (e.g. only a screen update when mouse is in `Xephyr`) and I get a proper first screen with `mstsc` as well as `xfreerdp`

akallabeth:
@freerdp-bot test again...

pnowack:
> I´ve tested with Xephyr && xterm && metacity (e.g. only a screen update when mouse is in Xephyr) and I get a proper first screen with mstsc as well as xfreerdp

Alright

akallabeth:
@freerdp-bot test properly!

PR #9239, opened 2023-08-01
Export interface include directories

dg0yt:
Another backport from vcpkg packaging: Make the imported target carry proper `INTERFACE_INCLUDE_DIRECTORIES`.

Disclaimer: This isn't much tested. It is guided by  
- occurences of `install(TARGETS ...)`
- for shared or static library targets
- inserted in front of `target_link_libraries` 
  (or at the end of the `if` branch where the target might alternatively be an executable)
- with values which should match the pc files (i.e. with winpr and uwac being different from freerdp).

It might make sense to have end-to-end tests for building a user project with cmake config, but I won't spent more work in this direction.


hardening:

akallabeth (approved):
LGTM, nice, thx

PR #9246, opened 2023-08-01
client/cliprdr_file: Move some debug messages under DEBUG_CLIPRDR

pnowack:
Quoting the main commit here:
```
Do no expose filenames of copied files, unless it is explicitly wanted.
```


akallabeth (approved):
LGTM

PR #9250, opened 2023-08-02
rdpdr: Fix potential null pointer access

mfleisz:
`_wcsrchr` might return `NULL` if no separator is in the path


akallabeth (approved):

akallabeth:
@freerdp-bot test please

PR #9251, opened 2023-08-02
Update error message when update_read* function failed

ehdgks0627:


akallabeth (approved):

akallabeth:
@freerdp-bot test

libfreerdp/core/orders.c
```diff
@@ -3487,7 +3487,7 @@ static BOOL read_primary_order(wLog* log, const char* orderName, wStream* s,
 
 	if (!rc)
 	{
-		WLog_Print(log, WLOG_ERROR, "%s - update_read_dstblt_order() failed", orderName);
+		WLog_Print(log, WLOG_WARN, "Read Primary Drawing Order %s failed", orderName);
```
akallabeth, line 3490: this should stay an error, as the return is `FALSE` and the session terminated.

akallabeth (changes requested):
wlog level revert to error please

libfreerdp/core/orders.c
ehdgks0627: https://github.com/FreeRDP/FreeRDP/blob/45ee30da5e5cb94a0335100c8e5f11304c1be6ff/libfreerdp/core/orders.c#L3762-L3766

I have set WLOG_WARN referring to this code. Should the above code also be set to WLOG_ERROR?

ehdgks0627: https://github.com/FreeRDP/FreeRDP/blob/45ee30da5e5cb94a0335100c8e5f11304c1be6ff/libfreerdp/core/orders.c#L3762-L3766
https://github.com/FreeRDP/FreeRDP/blob/45ee30da5e5cb94a0335100c8e5f11304c1be6ff/libfreerdp/core/orders.c#L3941-L3944
https://github.com/FreeRDP/FreeRDP/blob/45ee30da5e5cb94a0335100c8e5f11304c1be6ff/libfreerdp/core/orders.c#L4132-L4135

It seems that the logs and log levels are different when the rc is false in the functions `update_recv_altsec_order`, `update_recv_secondary_order`, and `update_recv_primary_order`. How about standardizing them?

akallabeth: good idea. log entries which return `FALSE` here abort the session (and are therefore an error)

akallabeth (approved):

akallabeth:
@freerdp-bot test please

akallabeth:

PR #9254, opened 2023-08-02
Fix build WITH_AAD

dmitigr:


dmitigr:
> Looks wrong. `libfreerdp` already links `public` to `cjson` so the symbols are already visible.

I don't know, but without that fix it doesn't works for me on Ubuntu 22.04.

akallabeth:
@dmitigr no. just compiled on a 22.04 vm and current master compiles just fine.

dmitigr:
@akallabeth Yep, this PR looks wrong. cJSON was not installed on my machine at the configure time and I've got link errors upon of build time. And then I've installed cJSON, but linkage still not worked because it wasn't configured properly.

Instead of this PR I would fix `FindCJSON.cmake` because `find_package(CJSON REQUIRED)` is called with `REQUIRED` but it's not handled in `FindCJSON.cmake`, which causes troubles with linkage if cJSON is not installed at configure time.

PR #9255, opened 2023-08-02
client/cliprdr_file: Do not destroy FUSE session while using it

pnowack:
```
When invalidating inodes, it is obligatory, that the session was not
destroyed yet.
So, in case of the FUSE loop stops before the session stops wait with
the destroyal of the session, until it is clear, that it is not used
anymore.
```


akallabeth (approved):
LGTM

pnowack:
@freerdp-bot Please do another test!

akallabeth:
@freerdp-bot test

akallabeth:

PR #9258, opened 2023-08-03
[client,common] Fix minor code errors

hodol-han:
Pass null device id rather than `deviceid` to find an empty pen device slot.

Hello! Please let me know if there is any misunderstanding.

Regards,
Hodol Han

### DIFF
--- a/client/common/client.c
+++ b/client/common/client.c
@@ -1887,6 +1887,8 @@ static FreeRDP_PenDevice* freerdp_client_get_pen(rdpClientContext* cctx, INT32 d
 static BOOL freerdp_client_register_pen(rdpClientContext* cctx, UINT32 flags, INT32 deviceid,
                                         double pressure)
 {
+	static const INT32 null_deviceid = 0;
+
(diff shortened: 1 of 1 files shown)


client/common/client.c
```diff
@@ -1896,7 +1898,7 @@ static BOOL freerdp_client_register_pen(rdpClientContext* cctx, UINT32 flags, IN
 	}
 
 	size_t pos = 0;
-	FreeRDP_PenDevice* pen = freerdp_client_get_pen(cctx, deviceid, &pos);
+	FreeRDP_PenDevice* pen = freerdp_client_get_pen(cctx, null_deviceid, &pos);
```
akallabeth, line 1901: ok, this is a bug, but the solution does not look correct.
1. should check for existing device with same id -> fail
2. should find first free slot and register there

akallabeth:
@freerdp-bot test

akallabeth (approved):

akallabeth:

PR #9259, opened 2023-08-03
fix unicode conversion functions with allocation for empty string

DVeron-RC:
A forti client was giving me cbClientDir = 2 and empty string clientDir in the info packet.
This is because the call in libfreerdp/core/info.c:360 is returning NULL when an empty string would be expected.
Empty strings are valid inputs for those unicode conversion functions.

Unicode conversion functions with allocation returned NULL when the input was a string of len = 0. 
Now those functions allocate buffers with the corresponding empty string that the caller has to free.

I manage to trigger the bug with shadow, and this PR fixes it.

I looked at the tests but was quite lost, will ask on matrix to add some coverage before this is merge.

### DIFF
--- a/winpr/libwinpr/crt/unicode.c
+++ b/winpr/libwinpr/crt/unicode.c
@@ -517,13 +517,13 @@ char* ConvertWCharToUtf8Alloc(const WCHAR* wstr, size_t* pUtfCharLength)
 	const SSIZE_T rc = ConvertWCharToUtf8(wstr, NULL, 0);
 	if (pUtfCharLength)
 		*pUtfCharLength = 0;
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot  test please

akallabeth:
@freerdp-bot test please

akallabeth (commented):
LGTM,
but for consistency sake it might be a good idea to also check the `WCHAR->UTF8` conversion functions for the same behaviour.

DVeron-RC:
Where are those functions ? I changed all the convert*Alloc functions of the unicode.c file.

PR #9261, opened 2023-08-03
[client,Android] Fix clipboard server-to-client

weoses:
Fix android server-to-client clipboard
Reproduce - connect to win7/win10 computer, and try to copy some text in connection, then paste it in another android app
Tested on win7/winserver2022

And few questions...
* What is "UTF8_STRING" constant (android_cliprdr.c:425)? Why we trying to register this clipboard format, but requesting format is CF_TEXT/CF_UNICODETEXT (android_cliprdr.c:255)?
* Is windows/xrdp servers always send CF_UNICODETEXT? I thing it is defined by application, which paste data into clipboard, and in this case it is worth doing some conversion using CF_LOCALE
* Is CF_UNICODETEXT is always UTF-16? All unicode in winapi is utf-16-le, but, if something went wrong.....

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c
+++ b/client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c
@@ -33,6 +33,7 @@
 #include "android_jni_utils.h"
 #include "android_jni_callback.h"
 
+#define TAG CLIENT_TAG("android.cliprdr")
 UINT android_cliprdr_send_client_format_list(CliprdrClientContext* cliprdr)
 {
 	UINT rc = ERROR_INTERNAL_ERROR;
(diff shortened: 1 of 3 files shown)


client/Android/Studio/freeRDPCore/src/main/cpp/android_jni_utils.c
```diff
@@ -107,6 +107,19 @@ char* get_string_from_string_builder(JNIEnv* env, jobject strBuilder)
 	return result;
 }
 
+/* Because wchar_t is 4-byte in android */
```
akallabeth, line 110: delete, use `_wcslen` from winpr

client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c
```diff
@@ -252,26 +253,28 @@ static UINT android_cliprdr_server_format_list(CliprdrClientContext* cliprdr,
 		}
 	}
 
+	bool hasUnicodeText = false;
```
akallabeth, line 256: windows always has `CF_UNICODETEXT`, `CF_TEXT` (or `CF_OEM_TEXT`) are not utf-8 but some local encoding, so better ignore these

client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c
```diff
@@ -422,11 +425,22 @@ android_cliprdr_server_format_data_response(CliprdrClientContext* cliprdr,
 		JNIEnv* env;
 		jstring jdata;
 		jboolean attached;
-		formatId = ClipboardRegisterFormat(afc->clipboard, "UTF8_STRING");
 		data = (void*)ClipboardGetData(afc->clipboard, formatId, &size);
+
```
akallabeth, line 429: should have been sufficient to replace `formatId` with `CF_TEXT` `UTF8_STRING` (that one is a leftover from 2.0)
`ClipboardInitSynthesizers` (winpr) does some magic on text formats to always have these available and on linux/mac os we have only utf-8 (at least nowadays) so local encoding issues do not apply

akallabeth (changes requested):

client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c
akallabeth: and yes, `CF_UNICODE` is always UTF16-le

akallabeth:
added the fix to #9248 (just use `text/plain` as format)

PR #9268, opened 2023-08-04
[various] fix integer conversions

akallabeth:


akallabeth:
@freerdp-bot test please

channels/cliprdr/server/cliprdr_main.c
```diff
@@ -532,7 +531,7 @@ static UINT cliprdr_server_receive_capabilities(CliprdrServerContext* context, w
 	Stream_Read_UINT16(s, capabilities.cCapabilitiesSets); /* cCapabilitiesSets (2 bytes) */
 	Stream_Seek_UINT16(s);                                 /* pad1 (2 bytes) */
 
-	for (index = 0; index < capabilities.cCapabilitiesSets; index++)
+	for (size_t index = 0; index < capabilities.cCapabilitiesSets; index++)
```
pnowack, line 534: A bit late now, it looks like `cCapabilitiesSets` is UINT32. In addition to that `[MS-RDPECLIP]` defines this value as UINT16. Maybe `index` should better be UINT32 (like `cCapabilitiesSets`)?
Alternative, would be to change API to fix the type of `cCapabilitiesSets` in the first place.

akallabeth: @pnowack not really an issue, as `size_t` does fit both (uint16 and uint32) and it is used to index an array.

PR #9276, opened 2023-08-08
Do not fail on certificates without subject

ggirol-rc:
It is possible to implement an rdp client that accepts certificates by fingerprint by using VerifyCertificateEx. In case the server uses a certificate without subject (which, apparently, is not mandated by X509) freerdp_certificate_data_load_cache fails and the certificate is refused even before calling VerifyCertificateEx. This commit changes freerdp_certificate_data_load_cache to consider that missing subject is the same as an empty string.

Also downgrade the log message complaining about missing subject and issuer to a warning.

To generate a certificate without these fields, you can run:
```
openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -sha256 -days 3650 -nodes -subj "/"
```

### DIFF
--- a/libfreerdp/crypto/certificate_data.c
+++ b/libfreerdp/crypto/certificate_data.c
@@ -70,7 +70,7 @@ static BOOL freerdp_certificate_data_load_cache(rdpCertificateData* data)
 
 	data->cached_subject = freerdp_certificate_get_subject(data->cert);
 	if (!data->cached_subject)
-		goto fail;
+		data->cached_subject = calloc(1, 1);
 
 	size_t pemlen = 0;
 	data->cached_pem = freerdp_certificate_get_pem(data->cert, &pemlen);
@@ -83,7 +83,7 @@ static BOOL freerdp_certificate_data_load_cache(rdpCertificateData* data)
 
 	data->cached_issuer = freerdp_certificate_get_issuer(data->cert);
 	if (!data->cached_issuer)
-		goto fail;
+		data->cached_issuer = calloc(1, 1);
 
 	rc = TRUE;
 fail:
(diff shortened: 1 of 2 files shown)


hardening (approved):
Seems ok for me

hardening:

hardening:
@ggirol-rc please rebase that PR so that the CI is happy

ggirol-rc:
done

hardening:
@freerdp-bot can you test that correctly now please ?

akallabeth:
@ggirol-rc you are correct that is not mandated by `X509` but could you please provide some case this makes any sense?

the implementation in `RDP` do require some fields that are not mandated by `X509` and I consider the subject one of these (as it is more or less the only human readable identifier for a server, just like a website certificate without the URL does not make much sense)

akallabeth (changes requested):
need an argument to convince me this is actually a good idea.

hardening:
I'm also hitting the case with a server that exposes a certificate with no subject, and I did no fancy configuration with it...

hardening:
The fix is either you have empty issuer/subject (this PR) or you have certs callback robust to NULL issuer / subject.

libfreerdp/crypto/certificate_data.c
```diff
@@ -83,7 +83,7 @@ static BOOL freerdp_certificate_data_load_cache(rdpCertificateData* data)
 
 	data->cached_issuer = freerdp_certificate_get_issuer(data->cert);
 	if (!data->cached_issuer)
-		goto fail;
+		data->cached_issuer = calloc(1, 1);
```
akallabeth, line 86: just remove the check, `NULL` should stay `NULL`

libfreerdp/crypto/certificate_data.c
```diff
@@ -70,7 +70,7 @@ static BOOL freerdp_certificate_data_load_cache(rdpCertificateData* data)
 
 	data->cached_subject = freerdp_certificate_get_subject(data->cert);
 	if (!data->cached_subject)
-		goto fail;
+		data->cached_subject = calloc(1, 1);
```
akallabeth, line 73: just remove the check, `NULL` should stay `NULL`

akallabeth (approved):

PR #9278, opened 2023-08-09
[cmake] Locate PkgConfig module with find_package() instead of include(FindPkgConfig)

KimonHoffmann:
This is the contemporary way of including find modules and including the find module this way avoids cmake errors in conjunction with find_package_handle_standard_args().

## How To Test

Perform a `--fresh` cmake run on an OS type that uses `pkg-config` and observe that there are no more warnings of the type:

```
CMake Warning (dev) at /usr/share/cmake-3.XX/Modules/FindPackageHandleStandardArgs.cmake:LLL (message):
  The package name passed to `find_package_handle_standard_args` (PkgConfig)
  does not match the name of the calling package (RealPackageName).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  [...]
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,7 +80,6 @@ include(CheckIncludeFiles)
 include(CheckLibraryExists)
 include(CheckSymbolExists)
 include(CheckStructHasMember)
-include(FindPkgConfig)
 include(TestBigEndian)
 
 include(FindFeature)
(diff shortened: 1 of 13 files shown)


hardening:
@freerdp-bot can you give a test on that ?

hardening:
Ok TestThreadExitThread is buggy, and it is waiting 100ms instead of 1 second

akallabeth:
@freerdp-bot test please

akallabeth (approved):
LGTM

PR #9280, opened 2023-08-10
rail: fix window freeze when moving window after minimizing and restoring

fujames1:
When restoring from minimized state, sometimes MapNotify is received before PropertyNotify, and the rail_state is still WINDOW_SHOW_MINIMIZED.


akallabeth:
@freerdp-bot test please

akallabeth (approved):
LGTM, thank you for the contribution!

PR #9282, opened 2023-08-11
fix freerdp_key_generate for some versions of OpenSSL (null key->evp)

awakecoding:
freerdp_key_generate is an internal helper that generates an RSA private key with OpenSSL, but one of the ifdefs for older versions of OpenSSL failed to account for initializing the key->evp parameter. I've looked [at a code sample](https://riptutorial.com/openssl/example/16737/generate-rsa-key) that is very similar to what we were doing, and found we needed to manually initialize the evp parameter ourselves, and that was all it took to make it work.


hardening:
@freerdp-bot can you re test please ?

awakecoding:
@hardening rebased

awakecoding:
@freerdp-bot test please

PR #9283, opened 2023-08-11
fix IFCALLRESULT expand to wrong expression; fix show black window when connect transparent app in windows server 2012r2

2fly2:
1、
if (!IFCALLRESULT(TRUE, pointer->New, context, pointer))
		goto out_fail;

2、
cmd:  wfreerdp.exe ...  /gfx:rfx ... 
when connect to windows server 2012r2, there is no alpha command.

### DIFF
--- a/include/freerdp/api.h
+++ b/include/freerdp/api.h
@@ -92,7 +92,7 @@
 		});                                                                \
 	})
 #else
-#define IFCALLRESULT(_default_return, _cb, ...) (_cb != NULL) ? _cb(__VA_ARGS__) : (_default_return)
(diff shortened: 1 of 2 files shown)


hardening:
Seems ok, could you cleanup the git history log: remove Merge pull request from 2fly2/2fly2-patch-1 and also rework your commit messages (check git log for pnovack commits for good examples). Thanks for the contribution.

2fly2:
> Seems ok, could you cleanup the git history log: remove Merge pull request from 2fly2/2fly2-patch-1 and also rework your commit messages (check git log for pnovack commits for good examples). Thanks for the contribution.
ok. When modifying these two issues, I will separate them in my Git commits. Thank you for your advice.

PR #9285, opened 2023-08-12
Fix incorrect definition IFCALLRESULT; fix rdp gfx caps flags

2fly2:
Fix the issue of incorrect macro definition IFCALLRESULT:
Correct code that looks similar to this:
	if (!IFCALLRESULT(TRUE, pointer->New, context, pointer))
		goto out_fail;
this will always run "goto out_fail".

fix show black window when connect transparent app in windows server 2012r2:
when specify RDPGFX_CAPS_FLAG_THINCLIENT and RDPGFX_CAPS_FLAG_SMALL_CACHE, windows server 2012 r2 will not send  
surface command with codec id RDPGFX_CODECID_ALPHA.

my command line is: wfreerdp.exe /v:xx /u:xx /p:xx /app:program:xx /gfx:rfx /f

### DIFF
--- a/winpr/include/winpr/custom-crypto.h
+++ b/winpr/include/winpr/custom-crypto.h
@@ -227,6 +227,7 @@ extern "C"
 
 	WINPR_API WINPR_CIPHER_CTX* winpr_Cipher_New(int cipher, int op, const void* key,
 	                                             const void* iv);
+	WINPR_API BOOL winpr_Cipher_SetPadding(WINPR_CIPHER_CTX* ctx, BOOL enabled);
 	WINPR_API BOOL winpr_Cipher_Update(WINPR_CIPHER_CTX* ctx, const void* input, size_t ilen,
 	                                   void* output, size_t* olen);
 	WINPR_API BOOL winpr_Cipher_Final(WINPR_CIPHER_CTX* ctx, void* output, size_t* olen);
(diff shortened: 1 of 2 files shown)


pnowack:
Please amend the second commit message to not extend over 72 characters. Use the body of the commit message for further details (also don't exceed the 72 char limit per line there too, just do new lines).
The title of the commit message must not exceed 72 characters. Second line must be empty. The body of the commit message starts at the third line.

> when specify RDPGFX_CAPS_FLAG_THINCLIENT and RDPGFX_CAPS_FLAG_SMALL_CACHE, windows server 2012 r2 will not send
surface command with codec id RDPGFX_CODECID_ALPHA.

Hmm.. do we actively do this somewhere? The `RDPGFX_CAPS_FLAG_THINCLIENT` and the `RDPGFX_CAPS_FLAG_SMALL_CACHE` flag should not be combined together (this is afaik also mentioned in the docs). So, if we currently do that, that would indeed be wrong.

Using the `RDPGFX_CAPS_FLAG_THINCLIENT` flag btw implicitly includes the `RDPGFX_CAPS_FLAG_SMALL_CACHE` effect.
I.e. both flags should not be used at the same time, but the small cache part is also part of the thinclient flag.

akallabeth:
@2fly2 the second change, you have an issue with `rails` there? might be a good idea to split this into two commits

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

akallabeth:
@2fly2 sorry, did accidentally close that, created a new pr with your changes (but `clang-formatted`)

PR #9286, opened 2023-08-12
TestThreadExitThread: fix messages and wait timeout for the windows CI

hardening:
Looks like for the windows x86_86 CI host 100ms is too small for a thread to terminate so let's wait for longer and decrease the number of loops.


hardening:
@akallabeth @mfleisz @awakecoding Please approve that PR so that we can unblock other pending PR failing in the CI because of TestThreadExitThread...

awakecoding (approved):

PR #9297, opened 2023-08-16
Refactor TestUnicodeConversion

DVeron-RC:
* added testcases when buffers reprensented empty string 
   * corrected the code to behave as the doc says
* added tests for the alloc versions

follow up of #9259

### DIFF
--- a/winpr/libwinpr/crt/unicode.c
+++ b/winpr/libwinpr/crt/unicode.c
@@ -374,12 +374,16 @@ void ByteSwapUnicode(WCHAR* wstr, size_t length)
 SSIZE_T ConvertWCharToUtf8(const WCHAR* wstr, char* str, size_t len)
 {
 	if (!wstr)
+	{
+		if (str && len)
+			str[0] = 0;
 		return 0;
+	}
 
 	const int rc =
 	    WideCharToMultiByte(CP_UTF8, 0, wstr, -1, str, (int)MIN(INT32_MAX, len), NULL, NULL);
 	if (rc <= 0)
-		return rc;
+		return -1;
 	else if ((size_t)rc == len)
 	{
 		if (str && (str[rc - 1] != '\0'))
@@ -398,7 +402,10 @@ SSIZE_T ConvertWCharNToUtf8(const WCHAR* wstr, size_t wlen, char* str, size_t le
 	size_t iwlen = _wcsnlen(wstr, wlen);
 
 	if (wlen > INT32_MAX)
+	{
+		SetLastError(ERROR_INVALID_PARAMETER);
 		return -1;
+	}
 
 	if (iwlen < wlen)
 	{
@@ -425,13 +432,16 @@ SSIZE_T ConvertWCharNToUtf8(const WCHAR* wstr, size_t wlen, char* str, size_t le
 
 SSIZE_T ConvertMszWCharNToUtf8(const WCHAR* wstr, size_t wlen, char* str, size_t len)
 {
-	if (len == 0)
+	if (wlen == 0)
 		return 0;
 
-	WINPR_ASSERT(str);
+	WINPR_ASSERT(wstr);
 
 	if (wlen > INT32_MAX)
+	{
+		SetLastError(ERROR_INVALID_PARAMETER);
 		return -1;
+	}
 
 	const int iwlen = MIN(INT32_MAX, len);
 	const int rc = WideCharToMultiByte(CP_UTF8, 0, wstr, (int)wlen, str, (int)iwlen, NULL, NULL);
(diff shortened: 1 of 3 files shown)


hardening:
@freerdp-bot test please

hardening:
That definitely needs a @akallabeth review !

winpr/libwinpr/crt/unicode_icu.c
```diff
@@ -61,7 +61,7 @@ int int_MultiByteToWideChar(UINT CodePage, DWORD dwFlags, LPCSTR lpMultiByteStr,
 		len = cbMultiByte;
 
 	if (len >= INT_MAX)
-		return -1;
+		return 0;
```
akallabeth, line 64: I´d consider an invalid input argument an error and not silently ignore that

winpr/libwinpr/crt/unicode.c
```diff
@@ -374,12 +374,12 @@ void ByteSwapUnicode(WCHAR* wstr, size_t length)
 SSIZE_T ConvertWCharToUtf8(const WCHAR* wstr, char* str, size_t len)
 {
 	if (!wstr)
-		return 0;
+		return -1;
```
akallabeth, line 377: this change is wrong, a `NULL` wstr` is allowed and should not return an error.

akallabeth (changes requested):
the changes to the tests look ok, but did not test yet.
the changes to the `API` are not good though.

akallabeth:
@freerdp-bot test

winpr/libwinpr/crt/unicode_icu.c
DVeron-RC: I did not find doc for this function (did not search a lot).
I aligned this with the other functions.
In  my oppinion, int_MultiByteToWideChar and int_WideCharToMultiByte, already tell they failed by returning 0.
0 should not be considered as a valid return value, as a buffer of size 0 does not exist.

winpr/libwinpr/crt/unicode.c
DVeron-RC: My bad did not tilt when reading the API doc, I'll revert this.

I disagree on the fact that NULL should be allowed as an input.
You won't be able to make a difference between NULL and an empty string.
So if you have a NULL WCHAR*, first call will return 0.
So you allocate a buffer of size 1 for the null termination.
Then you call the function again and nothing appens.
In the end you may have an uninitialiazed allocated buffer of size 1.
Not the worst thing in the world, but not the cleanest in my oppinion.

akallabeth: @DVeron-RC you are right about the uninitialized part, should initialize the buffer, so if you could add that would be great ;)

winpr/libwinpr/crt/unicode_icu.c
DVeron-RC: The windows counter part returns 0 on error
```The function returns 0 if it does not succeed. To get extended error information, the application can call GetLastError, which can return one of the following error codes:

ERROR_INSUFFICIENT_BUFFER. A supplied buffer size was not large enough, or it was incorrectly set to NULL.
ERROR_INVALID_FLAGS. The values supplied for flags were not valid.
ERROR_INVALID_PARAMETER. Any of the parameter values was invalid.
ERROR_NO_UNICODE_TRANSLATION. Invalid Unicode was found in a string.```
I think there is no legitimate case where this function should return 0 and not fail.
Every null terminated string are of size >=1.
And giving a buffer of size 0 as input is semantically dubious as you won't need to convert it.
It is considered as an invalid input in both windows API and WinPr:
```C
	if ((cbMultiByte == 0) || (cbMultiByte < -1))
		return 0;
```
There are no valid input of the function that should result in it returning 0 on success.

It is a good idea to set error on ERROR_INVALID_PARAMETER.
It is possible to arrive at len >= INT_MAX if the function is fed a null terminated string of 2G+, which I can see calling an invalid parameter.

akallabeth: ok, nice summary. sounds good to return `0` and set `ERROR_INVALID_PARAMETER`

winpr/libwinpr/crt/unicode.c
DVeron-RC: On it.

The more I think about it the more I still think NULL should not be allowed.
I don't understand why in ConvertWCharToUtf8 a NULL WCHAR* is valid,
whereas in ConvertWCharNToUtf8 it is not.

How does freerdp update its API ?

DVeron-RC: Ok, I'm dumb in the N version str can be NULL if len is 0. Coherent.

akallabeth:
@freerdp-bot test please

DVeron-RC:
I don't understand why the CI failed.

akallabeth (approved):
LGTM
nice work!

PR #9308, opened 2023-08-21
2fly2

akallabeth:


akallabeth:
reformatted #9285

pnowack:
I haven't checked the client side code here, but we should check the thinclient + small cache option first. If both are set, then the small cache flag should not be set in the first place (that is mentioned somewhere in `[MS-RDPEGFX]`.

akallabeth:
@pnowack reformated the commit message. as for the flag, I´ve checked and it should not be the default setting as that affects both, client and server. your suggestion should already be implemented, but I need to verify that.

libfreerdp/core/settings.c
```diff
@@ -648,7 +648,7 @@ rdpSettings* freerdp_settings_new(DWORD flags)
 	    !freerdp_settings_set_bool(settings, FreeRDP_NSCodecAllowDynamicColorFidelity, TRUE) ||
 	    !freerdp_settings_set_bool(settings, FreeRDP_AutoReconnectionEnabled, FALSE) ||
 	    !freerdp_settings_set_uint32(settings, FreeRDP_AutoReconnectMaxRetries, 20) ||
-	    !freerdp_settings_set_bool(settings, FreeRDP_GfxThinClient, TRUE) ||
+	    !freerdp_settings_set_bool(settings, FreeRDP_GfxThinClient, FALSE) ||
```
hardening, line 651: Shall we document that in the merging to FreeRDP 3.0 doc, as in some cases that can be a pretty important detail (those who have played with egfx proxying know what I'm talking about)

akallabeth: @hardening which document would that be? for the client command line that would now fit the doc as that was wrong.

hardening: Oh correct forget my remark

PR #9313, opened 2023-08-21
WinSCard API dynamic loading from custom library

awakecoding:
Add support for loading a custom WinSCard API library dynamically instead of the WinPR WinSCard API implementation that wraps either the original Windows API or adapts pcsc-lite to behave like it. A new WinSCardModule FreeRDP setting has been added alongside the /winscard-module command-line parameter. The goal is to do the same as the SSPI API with the SspiModule FreeRDP setting + /sspi-module command-line argument, and allow us to develop a drop-in WinSCard API replacement in Rust (winscard-rs) to load with our Rust SSPI implementation (sspi-rs) and emulate smartcards at the application layer without using any system libraries. Since Kerberos needs to talk to WinSCard for the smartcards, we'll likely just have both the SSPI API and WinSCard API inside the same shared library, bundled together and fully aware of each other, we'll just pass the path to the same library in both places.

### DIFF
--- a/libfreerdp/utils/smartcard_call.c
+++ b/libfreerdp/utils/smartcard_call.c
@@ -30,6 +30,7 @@
 #include <winpr/crt.h>
 #include <winpr/print.h>
 #include <winpr/stream.h>
+#include <winpr/library.h>
 #include <winpr/smartcard.h>
 
 #include <freerdp/freerdp.h>
@@ -47,10 +48,12 @@
 #include <freerdp/emulate/scard/smartcard_emulate.h>
 
 #define str(x) #x
-#define wrap(ctx, fkt, ...) \
-	ctx->useEmulatedCard ? Emulate_##fkt(ctx->emulation, ##__VA_ARGS__) : fkt(__VA_ARGS__)
+#define wrap(ctx, fkt, ...)                                             \
+	ctx->useEmulatedCard ? Emulate_##fkt(ctx->emulation, ##__VA_ARGS__) \
+	                     : ctx->pWinSCardApi->pfn##fkt(__VA_ARGS__)
 #else
-#define wrap(ctx, fkt, ...) ctx->useEmulatedCard ? SCARD_F_INTERNAL_ERROR : fkt(__VA_ARGS__)
+#define wrap(ctx, fkt, ...) \
+	ctx->useEmulatedCard ? SCARD_F_INTERNAL_ERROR : ctx->pWinSCardApi->pfn##fkt(__VA_ARGS__)
 #endif
 
 #define SCARD_MAX_TIMEOUT 60000
@@ -64,6 +67,9 @@ struct s_scard_call_context
 #if defined(WITH_SMARTCARD_EMULATE)
 	SmartcardEmulationContext* emulation;
 #endif
+	HANDLE hWinSCardLibrary;
+	SCardApiFunctionTable WinSCardApi;
+	SCardApiFunctionTable const* pWinSCardApi;
 	HANDLE stopEvent;
 	void* userdata;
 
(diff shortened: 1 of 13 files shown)


awakecoding:
@freerdp-bot test

libfreerdp/utils/smartcard_call.c
```diff
@@ -64,6 +67,9 @@ struct s_scard_call_context
 #if defined(WITH_SMARTCARD_EMULATE)
 	SmartcardEmulationContext* emulation;
 #endif
+	HANDLE hWinSCardLibrary;
+	SCardApiFunctionTable WinSCardApi;
+	PSCardApiFunctionTable pWinSCardApi;
```
akallabeth, line 72: should be `const`

libfreerdp/utils/smartcard_call.c
```diff
@@ -1855,6 +1861,31 @@ scard_call_context* smartcard_call_context_new(const rdpSettings* settings)
...
 	}
+	else
+	{
+		if (settings->WinSCardModule)
+		{
+			ctx->hWinSCardLibrary = LoadLibraryX(settings->WinSCardModule);
+
+			if (!ctx->hWinSCardLibrary)
+			{
+				WLog_ERR(TAG, "Failed to load WinSCard library: '%s'", settings->WinSCardModule);
```
akallabeth, line 1872: `goto fail;`?

akallabeth (changes requested):
overall LGTM, some changes required.

libfreerdp/utils/smartcard_call.c
awakecoding: maybe I'm holding it wrong... how do I make pWinSCardApi const without causing these kinds of problems:

```
FreeRDP/libfreerdp/utils/smartcard_call.c: In function ‘smartcard_call_context_new’:
FreeRDP/libfreerdp/utils/smartcard_call.c:1876:22: error: assignment of read-only member ‘pWinSCardApi’
 1876 |    ctx->pWinSCardApi = &ctx->WinSCardApi;
      |                      ^
FreeRDP/libfreerdp/utils/smartcard_call.c:1881:22: error: assignment of read-only member ‘pWinSCardApi’
 1881 |    ctx->pWinSCardApi = WinPR_GetSCardApiFunctionTable();
      |                      ^
FreeRDP/libfreerdp/utils/smartcard_call.c: In function ‘smartcard_call_context_free’:
FreeRDP/libfreerdp/utils/smartcard_call.c:1936:20: error: assignment of read-only member ‘pWinSCardApi’
 1936 |  ctx->pWinSCardApi = NULL;
```

awakecoding: I've used `SCardApiFunctionTable const*` instead of `const PSCardApiFunctionTable`, which I understand should give a pointer to a const SCardApiFunctionTable, rather than a const pointer to a non-const SCardApiFunctionTable. I'm hoping this is what you meant by 'const' here

awakecoding:
@freerdp-bot test

winpr/include/winpr/smartcard.h
```diff
@@ -1209,6 +1209,10 @@ extern "C"
 	WINSCARDAPI const char* WINAPI SCardGetCardStateString(DWORD dwCardState);
 	WINSCARDAPI char* WINAPI SCardGetReaderStateString(DWORD dwReaderState);
 
+	WINPR_API bool WinSCard_LoadApiTableFunctions(PSCardApiFunctionTable pWinSCardApiTable,
```
akallabeth: include `stdbool.h` or use `BOOL` instead

akallabeth (approved):
LGTM.

@awakecoding did rebase and fix a minor issue myself, hope you are ok with that.

PR #9319, opened 2023-08-22
[winpr,wtypes] add WINPR_RESTRICT macro

akallabeth:
preparation to allow code optimization.
the `restrict` keyword tells the compiler a buffer is not aliased, so it can better optimize (simdi, ...)


hardening (approved):
No the expert, but I have nothing against it

PR #9328, opened 2023-08-23
Turn BUILD_SHARED_LIBS off by default

awakecoding:
Turn CMake BUILD_SHARED_LIBS off by default, meaning a default build produces static libraries, instead of multiple shared libraries. We already set BUILD_SHARED_LIBS to OFF in CI builds, and as far as I'm concerned, I always set it off in my projects. If I want to make C# bindings, I just link everything into a shared library which an explicit SHARED property on the cmake target.

This does not mean we stop supporting shared libraries, just that it won't be the *default*, which is really what I'd expect developers to want. Let's just have the few places where it's desirable to have shared libraries set BUILD_SHARED_LIBS to ON in their builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,12 +200,7 @@ endif(CMAKE_CROSSCOMPILING)
 # Turn on solution folders (2.8.4+)
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
-
-set(LIB_DEFAULT ON)
-if (IOS)
-	set(LIB_DEFAULT OFF)
-endif()
-option(BUILD_SHARED_LIBS "Build shared libraries" ${LIB_DEFAULT})
+option(BUILD_SHARED_LIBS "Build shared libraries" OFF)
 option(EXPORT_ALL_SYMBOLS "Export all symbols form library" OFF)
 
 if(BUILD_TESTING)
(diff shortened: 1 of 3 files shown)


akallabeth (changes requested):
rejected.

we default to `ON` for a reason (mainly that we don´t ship applications but a library that can utilize RDP and is already used by a couple of applications)

bmiklautz:
Having BUILD_SHARED_LIBS  OFF might be handy for development or if building for a product to ship. However most (if not all) builds for distributions do have extra packages for the libraries as well as the client(s).
I also opt in for keeping it to `ON`.

However I your point @awakecoding and as dependencies and build options are getting more - what do you think about having some useful default `build` configurations/profiles/.. as we do ci builds in `ci/cmake-preloads`.

awakecoding:
> Having BUILD_SHARED_LIBS  OFF might be handy for development or if building for a product to ship. However most (if not all) builds for distributions do have extra packages for the libraries as well as the client(s).
> I also opt in for keeping it to `ON`.
> 
> However I your point @awakecoding and as dependencies and build options are getting more - what do you think about having some useful default `build` configurations/profiles/.. as we do ci builds in `ci/cmake-preloads`.

CMake preloads lack discoverability, and are cumbersome to use. Developers are going to try using CMake directly no matter what, and not figure out there might be a CMake preload file available. Even then, a CMake preload would explicitly set options, rather than detect and adapt based on what is available on the system. In the case of shared libraries, sure, we always want to statically link everything for developer builds, but what about the dependencies?

PR #9329, opened 2023-08-23
disable WITH_AAD if CJSON cannot be found

awakecoding:
In an effort to simplify cmake generation with fewer explicit options, I modified the CJSON detection to be optional and conditionally enable WITH_AAD that depends on it only if it is found. We have many options that require explicit disabling because of missing dependencies like this that make FreeRDP very difficult to approach by developers who waste a lot of time trying to generate projects repeatedly by disabling stuff until it finally works, and this is just one of those that was added recently.


akallabeth (commented):
note really a big fan of `works but does not what was told` kind of features. 

prefer the `blows up in my face` kind of approach where I actively have to check why something failed
and having a development branch change some defaults is to be expected. (that does not happen on the stable side and we´re working on 3.0 ;))

akallabeth:
implemented with #9455 
using standard `find_package(cJSON)` to detect `cJSON` and set the default for `option(WITH_AAD "sometext" ${cJSON_FOUND})`

PR #9331, opened 2023-08-24
[channels,printer] default CUPS to being enabled

akallabeth:
* printer support now enabled by default unless iOS or ANDROID is build.
* Move cups library detection to printer module


akallabeth:
@freerdp-bot test please

Issue #9363, opened 2023-09-04
[Microsoft AVD] Authentication result: SEC_E_INVALID_TOKEN


Brueggus:
**Describe the bug**
Connecting to my Microsoft Azure Virtual Desktop VM fails with the following error (full log below):

```
[16:22:06:441] [57043:0000ded4] [DEBUG][com.freerdp.core.rdp] - [rdp_client_transition_to_state][0x55d833ff91d0]: CONNECTION_STATE_NEGO --> CONNECTION_STATE_AAD
[16:22:06:448] [57043:0000ded4] [TRACE][com.freerdp.core.rdp] - [rdp_recv_callback][0x55d833ff91d0]: (client)[CONNECTION_STATE_AAD -> CONNECTION_STATE_AAD] current return STATE_RUN_SUCCESS [0] [feeding 158 bytes, 0 bytes not processed]
[16:22:06:473] [57043:0000ded4] [ERROR][com.freerdp.aad] - [aad_parse_state_auth]: Authentication result: SEC_E_INVALID_TOKEN (0x80090308)
[16:22:06:473] [57043:0000ded4] [ERROR][com.freerdp.core.rdp] - [rdp_recv_callback_int][0x55d833ff91d0]: CONNECTION_STATE_AAD - aad_recv() fail
[16:22:06:473] [57043:0000ded4] [ERROR][com.freerdp.core.rdp] - [rdp_recv_callback_int][0x55d833ff91d0]: CONNECTION_STATE_AAD status STATE_RUN_FAILED [-1]
```

**To Reproduce**
1. Go to https://client.wvd.microsoft.com/arm/webclient/index.html
2. Download .rdpw file
3. Launch `sdl-freerdp Downloads/My\ Personal\ AVD.rdpw /gateway:type:arm /u:my.user@domain.com /f`
4. Authenticate
5. Authenticate again 
6. sdl-freerdp quits: `Authentication result: SEC_E_INVALID_TOKEN (0x80090308)`

**Expected behavior**
The connection to the AVD should be successful.

**Screenshots**
n/a

**Log**

[freerdp.log](https://github.com/FreeRDP/FreeRDP/files/12514632/freerdp.log)

(I tried to redact any sensitive values. If important information is missing, please let me know.)

**Application details**
* FreeRDP version (`This is FreeRDP version 3.0.0-dev4 (5f8f69512)`)
[...]

PR #9370, opened 2023-09-07
[core,arm] Various fixes and adjustements for #9363

hardening:
Treat the case where the publicIpAddress is there but empty. 
Don't try RDSTLS if the password has not been been provided.
When RDSTLS is what we will do, disengage all the other security kinds.

### DIFF
--- a/libfreerdp/core/gateway/arm.c
+++ b/libfreerdp/core/gateway/arm.c
@@ -504,6 +504,10 @@ BOOL arm_encodeRedirectPasswd(rdpSettings* settings, const rdpCertificate* cert,
 	}
 
 	settings->RdstlsSecurity = TRUE;
+	settings->AadSecurity = FALSE;
+	settings->NlaSecurity = FALSE;
+	settings->RdpSecurity = FALSE;
+	settings->TlsSecurity = FALSE;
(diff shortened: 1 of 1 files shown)


libfreerdp/core/gateway/arm.c
```diff
@@ -699,54 +704,59 @@ static BOOL arm_fill_gateway_parameters(rdpArm* arm, const char* message, size_t
...
-		if (!redirectedAuthGuid)
-			break;
-
-		size_t wGUID_len = 0;
-		wGUID = ConvertUtf8ToWCharAlloc(redirectedAuthGuid, &wGUID_len);
-		if (!wGUID)
+		/* note: we retrieve some more fields for RDSTLS only if we have a password provided by the
+		 * user, otherwise these are useless: we will not be able to do RDSTLS
+		 */
+		do
```
akallabeth, line 712: maybe split this out to a `static` function? the `do {} while(FALSE)` grouping is probably easier to implement that way

akallabeth:
@freerdp-bot test
@hardening is this ready now?

hardening:
PR has been updated and is ready

akallabeth (approved):
LGTM

PR #9371, opened 2023-09-07
[codec,interleaved] fix type definition

akallabeth:


libfreerdp/codec/include/bitmap.c
```diff
@@ -291,7 +291,7 @@ static INLINE BOOL RLEDECOMPRESS(const BYTE* pbSrcBuffer, UINT32 cbSrcBuffer, BY
 					return FALSE;
 				pbSrc = pbSrc + advance;
 
-				if (!buffer_within_range(pbSrc + sizeof(fgPel), pbEnd))
+				if (!buffer_within_range(pbSrc + PIXEL_SIZE, pbEnd))
```
hardening, line 294: I also have investigated that bug, and here `fgPel` is present only in the case of "Set" commands (in my understanding of the specs). So either the check must be moved a few lines below when we test for `LITE_SET_FG_FGBG_IMAGE` and `MEGA_MEGA_SET_FGBG_IMAGE` or we compute an `extra` var depending on if it's a "set" command, and we check for `buffer_within_range(pbSrc + extra, pbEnd)`

libfreerdp/codec/interleaved.c
```diff
@@ -158,9 +160,9 @@ static INLINE BOOL buffer_within_range_(const void* pbSrc, const void* pbEnd, co
                                         const char* file, size_t line)
 {
 	WINPR_UNUSED(file);
-	if (pbSrc >= pbEnd)
+	if (pbSrc > pbEnd)
 	{
```
hardening, line 164: I think the previous version was the correct one.

libfreerdp/codec/include/bitmap.c
akallabeth: good point, checked all locations again.

libfreerdp/codec/interleaved.c
akallabeth: no.

Issue #9214, opened 2023-07-28
Wait until the AVD instance is running


progxaker:
**Is your feature request related to a problem? Please describe.**
I use AVD with auto-shutdown plugin and when the host is down,
I get the following error:

<details>
<summary>Error log</summary>

```
Browse to: https://login.microsoftonline.com/common/oauth2/v2.0/authorize?client_id=a85cf173-4192-42f8-81fa-777a763e6e2c&response_type=code&scope=https%3A%2F%2Fwww.wvd.microsoft.com%2F.default&redirect_uri=https%3A%2F%2Flogin.microsoftonline.com%2Fcommon%2Foauth2%2Fnativeclient  
Paste redirect URL here: 
https://login.microsoftonline.com/common/oauth2/nativeclient?code=<code>&session_state=<state>
[09:32:27:253] [35758:00008bb0] [DEBUG][com.freerdp.utils.http] - [freerdp_http_request]: headers:
POST /common/oauth2/v2.0/token HTTP/1.1
Host: login.microsoftonline.com
Content-Type: application/x-www-form-urlencoded
Content-Length: 1039


[09:32:28:453] [35758:00008bb0] [DEBUG][com.freerdp.utils.http] - [freerdp_http_request]: body:
grant_type=authorization_code&code=<code>&client_id=a85cf173-4192-42f8-81fa-777a763e6e2c&scope=https%3A%2F%2Fwww.wvd.microsoft.com%2F.default&redirect_uri=https%3A%2F%2Flogin.microsoftonline.com%2Fcommon%2Foauth2%2Fnativeclient
[09:32:28:30] [35758:00008bb0] [DEBUG][com.freerdp.utils.ringbuffer] - [ringbuffer_write]: ringbuffer_write(0x60600001e330): sz: 4841
[09:32:28:30] [35758:00008bb0] [DEBUG][com.freerdp.utils.ringbuffer] - [ringbuffer_peek]: ringbuffer_peek(0x60600001e330): sz: 4841
[09:32:28:30] [35758:00008bb0] [DEBUG][com.freerdp.utils.ringbuffer] - [ringbuffer_commit_read_bytes]: ringbuffer_commit_read_bytes(0x60600001e330): sz: 4841
[09:32:28:30] [35758:00008bb0] [DEBUG][com.freerdp.utils.ringbuffer] - [ringbuffer_write]: ringbuffer_write(0x60600001e330): sz: 238
[...]

PR #9383, opened 2023-09-13
[core,arm] Waiting until an AVD instance is up

progxaker:
## Preparations
* [x] The branch is re-based.
* [x] The commit history is cleaned up.
* [x] Formatted using clang-format
* [x] `scan-build` was run.
  * `master` branch - 155 bugs found.
  * `avd-auto-reconnect-poc` branch - (built `master` and the the current branch) 0 bugs found.
* [ ] Run a compiler at pedantic level.

## Accepting the contribution
* [x] Give a proper name.
* [x] Add a brief description.
* [x] Give instructions on how to test the changes.
* [ ] Add unit tests.

## Description

#9214 

In addition to the issue I would like to:
* prepare correct logging / function exits for cases where the response isn't JSON or there is no the `Code` property.
* move the functionality to a separate functions (maybe even outside `arm.c`).
* update the "running" message to match the Web-client version.
* no memory leaks on successful pass, but other ways are needed to be verified.

It would be ideal to add such functionality for the next case when we lost our internet connection or connected to a VPN:

```
[transport_default_write]: ERRCONNECT_CONNECT_TRANSPORT_FAILED [0x0002000D]
[transport_check_fds]: ERRCONNECT_CONNECT_TRANSPORT_FAILED [0x0002000D]
[transport_check_fds]: TODO: Trying to set error code ERRCONNECT_CONNECT_TRANSPORT_FAILED, but ERRCONNECT_CONNECT_TRANSPORT_FAILED already set!
[input_ensure_client_running]: [APPLICATION BUG] input functions called after the session terminated
```

## Instructions on how to test

1. Disable an AVD instance using Azure Portal or the `shudown /s` command in Command Prompt.
2. Connect to the instance.
3. The following messages appears:
```
Browse to: https://login.microsoftonline.com/common/oauth2/v2.0/authorize?client_id=a85cf173-4192-42f8-81fa-777a763e6e2c&response_type=code&scope=https%3A%2F%2Fwww.wvd.microsoft.com%2F.default&redirect_uri=https%3A%2F%2Flogin.microsoftonline.com%2Fcommon%2Foauth2%2Fnativeclient
Paste redirect URL here:
<redirect-URL>
[10:10:41:501] [49251:0000c064] [INFO][com.freerdp.core.gateway.arm] - [arm_resolve_endpoint]: Please wait... VM is starting running.
[10:11:42:259] [49251:0000c064] [INFO][com.freerdp.core.gateway.arm] - [arm_resolve_endpoint]: Please wait... VM is starting running.
```
4. After a few minutes, FreeRDP will connect to the instance.


progxaker:
I looked at my previous HAR dump and found [strings.json](https://client.wvd.microsoft.com/arm/webclient/assets/loc/en-US/strings.json). There is a original message: `"EVENT_BASED_SCALING_START": "Starting your VM. It may take up to 5 minutes"`. I'll update the message soon.

akallabeth:
@progxaker ok, nice pr!
did restructure your changes a bit, so now the memory leak should be gone (you called the http request before clearing up the last response)

akallabeth:
@freerdp-bot test

akallabeth:
@progxaker thank you for this pr!
@hardening what do you think about this?

hardening:
@akallabeth I think that #9370  should be merged before this one

libfreerdp/core/gateway/arm.c
```diff
@@ -56,18 +56,26 @@
...
+
+	UINT32 gateway_retry;
 };
 
 typedef struct rdp_arm rdpArm;
 
 #define TAG FREERDP_TAG("core.gateway.arm")
 
 #ifdef WITH_AAD
+static const UINT32 max_retries = 60;
```
hardening, line 76: Shouldn't these variable be in settings ?

libfreerdp/core/gateway/arm.c
```diff
@@ -56,18 +56,26 @@
 #include <cjson/cJSON.h>
 #endif
 
+#include <time.h>
```
hardening, line 59: Not sure where we now use functions from time.h

hardening:
The two static variables definitely need to be moved to settings (and probably have their command line switch). Apart from that looks ok.

libfreerdp/core/gateway/arm.c
akallabeth: not sure this needs to be configurable. the timeout is currently hard coded to 10 minutes, anything longer will most likely be useless anyway.

akallabeth:
> The two static variables definitely need to be moved to settings (and probably have their command line switch). Apart from that looks ok.
got already too many switches and this one looks kinda useless. (the timeout is something that does not really need configuration, if something is not there after 10 minutes it will never be there)

akallabeth:
@hardening but I´m thinking of something (without exposing to command line), e.g. have a setting and a configurable callback for displaying e.g. a dialog

libfreerdp/core/gateway/arm.c
akallabeth: no, removed.

akallabeth:
@hardening ok, refactored.
1. There is now a `RetryDialog` callback with a default implementation. If the function is `NULL` then no retries are done.
2. In the default callback the `AutoReconnectionEnabled`, `AutoReconnectMaxRetries` and `TcpConnectTimeout` are reused to decide if reconnections are attempted or aborted immediately.

akallabeth:
@freerdp-bot test

hardening (approved):

PR #9388, opened 2023-09-14
[drdynvc] improvements of server-side DRDYNVC

hardening:
This patch prepares the reading of the dynamic channel version so that next we can take in account this to take advantage of advanced features in last versions (compressions or priorities).
The patch also implement notifying the VCM event when the dynamic channel becomes ready so that users of FreeRDP can just do calls to `WTSVirtualChannelManagerGetDrdynvcState` when the channel event is set (no blind calls).

### DIFF
--- a/libfreerdp/core/server.c
+++ b/libfreerdp/core/server.c
@@ -167,8 +167,20 @@ static BOOL wts_read_drdynvc_capabilities_response(rdpPeerChannel* channel, UINT
 	Stream_Seek_UINT8(channel->receiveData); /* Pad (1 byte) */
 	Stream_Read_UINT16(channel->receiveData, Version);
 	DEBUG_DVC("Version: %" PRIu16 "", Version);
-	channel->vcm->drdynvc_state = DRDYNVC_STATE_READY;
-	return TRUE;
+
+	if (Version < 1)
+	{
+		WLog_ERR(TAG, "invalid version %" PRIu16 " for DRDYNVC", Version);
+		return FALSE;
+	}
+
+	WTSVirtualChannelManager* vcm = channel->vcm;
+	vcm->drdynvc_state = DRDYNVC_STATE_READY;
+
+	/* we only support version 1 for now (no compression yet) */
+	vcm->dvc_spoken_version = MAX(Version, 1);
+
+	return SetEvent(MessageQueue_Event(vcm->queue));
 }
 
 static BOOL wts_read_drdynvc_create_response(rdpPeerChannel* channel, wStream* s, UINT32 length)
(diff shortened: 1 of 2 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

hardening:

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test please

libfreerdp/core/server.c
```diff
@@ -167,8 +167,20 @@ static BOOL wts_read_drdynvc_capabilities_response(rdpPeerChannel* channel, UINT
...
+	{
+		WLog_ERR(TAG, "invalid version %" PRIu16 " for DRDYNVC", Version);
+		return FALSE;
+	}
+
+	WTSVirtualChannelManager* vcm = channel->vcm;
+	vcm->drdynvc_state = DRDYNVC_STATE_READY;
+
+	/* we only support version 1 for now (no compression yet) */
+	vcm->dvc_spoken_version = Version;
```
pnowack, line 181: It looks like we don't actually use the `dvc_spoken_version` variable. But even if we would do then the comment above would not apply, as `Version` can also be `3` here (which implies compression support).

libfreerdp/core/server.c
```diff
@@ -167,8 +167,20 @@ static BOOL wts_read_drdynvc_capabilities_response(rdpPeerChannel* channel, UINT
 	Stream_Seek_UINT8(channel->receiveData); /* Pad (1 byte) */
 	Stream_Read_UINT16(channel->receiveData, Version);
 	DEBUG_DVC("Version: %" PRIu16 "", Version);
-	channel->vcm->drdynvc_state = DRDYNVC_STATE_READY;
-	return TRUE;
+
+	if (Version < 1 || Version > 3)
```
pnowack, line 171: I think we should not filter out newer versions. In case MS adds version 4, it breaks our server implementations.

```suggestion
	if (Version < 1)
```

libfreerdp/core/server.c
hardening: True, I must admit that my main goal here was to remove the warning about `unused variable Version`. I will adjust that.

libfreerdp/core/server.c
hardening: Well in fact right now we only propose version 1 to the client, so we do not expect him to answer with something bigger.

pnowack: > Well in fact right now we only propose version 1 to the client, so we do not expect him to answer with something bigger.

In some MS protocols yes, but `[MS-RDPEDYC]` does exclude this situation here though:
```
The server indicates the maximum capability it supports, and the client responds with a Capabilities Response PDU that indicates the maximum capability that it can support.
```

pnowack (approved):

pnowack:
@freerdp-bot test

PR #9390, opened 2023-09-15
[gateway,arm] fix input string length check

akallabeth:


akallabeth:
@freerdp-bot test please

PR #9403, opened 2023-09-19
Leak fix

akallabeth:


libfreerdp/core/gateway/arm.c
```diff
@@ -956,6 +956,10 @@ BOOL arm_resolve_endpoint(rdpContext* context, DWORD timeout)
 				WLog_DBG(TAG, "Delay for %" PRIdz "ms before next attempt", delay);
 				Sleep(delay);
 			}
+			else
```
hardening, line 959: I'd suggest inverting the if condition 
```C
if (delay == 0)
   break;

WLog_DBG(TAG, "Delay for %" PRIdz "ms before next attempt", delay);
Sleep(delay);
```

hardening (approved):
nitpicking but LGTM

libfreerdp/core/gateway/arm.c
akallabeth: hmm, I´d say preference. and would be `delay <= 0` as `-1` is returned in error cases.

PR #9412, opened 2023-09-20
[winpr,cmdline] fix unquoting of value

akallabeth:
Comma separated lists of form 'a,b,\"b,c,d\"' were not correctly parsed as the trailing quote was not consumed.


mfleisz (approved):
lgtm

PR #9417, opened 2023-09-21
Arm leak fix

akallabeth:
reset `rdpTls` before doing a reconnect and leak already allocated memory


akallabeth:
@freerdp-bot test

mfleisz (approved):
LGTM - nice catch on the `ZeroMemory` bug

PR #9420, opened 2023-09-21
Option to suppress sending windows key to remote host.

bencollerson:
Currently implemented only for X11 and wayland clients

This patch provides an option that allows the "Windows Logo" key (aka Super or Mod4) to be used on the local host without opening the start menu on a remote windows machine.

This can be tested by connecting to a remote windows machine and confirming the start menu does not open when the Windows logo key is pressed.

### DIFF
--- a/client/Wayland/wlf_input.c
+++ b/client/Wayland/wlf_input.c
@@ -340,6 +340,10 @@ BOOL wlf_handle_key(freerdp* instance, const UwacKeyEvent* ev)
 	if (rdp_scancode == RDP_SCANCODE_UNKNOWN)
 		return TRUE;
 
+	if (( rdp_scancode == RDP_SCANCODE_LWIN || rdp_scancode == RDP_SCANCODE_RWIN ) &&
+		instance->context->settings->SuppressWinKey )
+		return TRUE;
+
 	return freerdp_input_send_keyboard_event_ex(input, ev->pressed, ev->repeated, rdp_scancode);
 }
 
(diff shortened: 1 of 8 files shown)


akallabeth (changes requested):
not really a good idea.
1. we already use `<ctrl>+<alt>` as ungrab sequence.
2. if you do not want this to be redirected you can use `-grab-keyboard`
3. implementation is too use case specific (why the `win` key and some other hot key?)

that said, if you provide something to allow filtering arbitrary keys that might be of interest

bencollerson:
No worries. I find this patch useful and thought other people might find it useful also.  If it is too obscure that is fine. I will just keep using it on my own. I really can't think of any key that is as annoying as a modifier that opens up a menu that steals all further keystrokes on a remote windows host. There is no point coming up with a generalised version.

bencollerson:
This pull request can be closed.

I finally worked out how remapping works and suppressing windows keys from being sent to the remote host can be done by appending the following command line option:

    /kbd:remap:0x15b=0xff,remap:0x15c=0xff

Scancodes can be found like this: 

    $ xfreerdp /list:kbd-scancode|grep '[LR]WIN'
    0x015b  --> VK_LWIN
    0x015c  --> VK_RWIN

(Note: The leading zeros in the scancodes above, between the 'x' and '1', must be omitted so the argument will parse.)

and "No Key" which these are both mapped to is defined as follows:

    winpr/include/winpr/input.h:#define VK_NONE 0xFF /* no key */

PR #9423, opened 2023-09-22
[warnings] fixed reserved-identifer warnings

akallabeth:


server/proxy/pf_context.c
```diff
@@ -89,6 +93,12 @@ void StaticChannelContext_free(pServerStaticChannelContext* ctx)
 	free(ctx);
 }
 
+static void HashStaticChannelContext_free(void* ptr)
+{
+	pServerStaticChannelContext* ctx = (pServerStaticChannelContext*)ptr;
```
pnowack, line 98: Casting void pointers is a superfluous action

```suggestion
	pServerStaticChannelContext* ctx = ptr;
```

server/proxy/pf_context.c
```diff
@@ -129,7 +139,7 @@ static BOOL client_to_proxy_context_new(freerdp_peer* client, rdpContext* ctx)
 	obj->fnObjectEquals = ChannelId_Compare;
 
 	obj = HashTable_ValueObject(context->channelsByFrontId);
-	obj->fnObjectFree = StaticChannelContext_free;
+	obj->fnObjectFree = HashStaticChannelContext_free;
```
pnowack, line 142: Nit: The name now reads to me as "hash of (a) static channel context", but this doesn't seem to be the case here. Maybe there is a different way to name this (I am actually fine with the previous `StaticChannelContext_free` (or the same written in snake case))?

pnowack (approved):
Just two nits. Aside from that, LGTM

server/proxy/pf_context.c
akallabeth: thx for the hint, will remember that for the next pr ;)

server/proxy/pf_context.c
akallabeth: hard to name it, as it only exists due to `C` lack of overloaded parameter support.

PR #9427, opened 2023-09-25
Implement connection to child session

hardening:
Only available on windows with `/connect-child-session` you can connect to your RDP child session under windows.

### DIFF
--- a/libfreerdp/core/childsession.c
+++ b/libfreerdp/core/childsession.c
@@ -1,0 +1,340 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ * Named pipe transport
+ *
+ * Copyright 2023 David Fort <contact@hardening-consulting.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "tcp.h"
+#include <winpr/library.h>
+#include <winpr/assert.h>
+#include "childsession.h"
+
+#define TAG FREERDP_TAG("childsession")
+
+typedef struct
+{
+	HANDLE hFile;
+} WINPR_BIO_NAMED;
+
+static int transport_bio_named_uninit(BIO* bio);
+
+static int transport_bio_named_write(BIO* bio, const char* buf, int size)
+{
+	WINPR_ASSERT(bio);
+	WINPR_ASSERT(buf);
+
+	WINPR_BIO_NAMED* ptr = (WINPR_BIO_NAMED*)BIO_get_data(bio);
+
+	if (!buf)
+		return 0;
+
+	BIO_clear_flags(bio, BIO_FLAGS_WRITE);
+	DWORD written = 0;
+
+	BOOL ret = WriteFile(ptr->hFile, buf, size, &written, NULL);
+	WLog_VRB(TAG, "transport_bio_named_write(%d)=%d written=%d", size, ret, written);
+
+	if (!ret)
+		return -1;
+
+	if (written == 0)
+		return -1;
+
+	return written;
+}
+
+static int transport_bio_named_read(BIO* bio, char* buf, int size)
+{
+	WINPR_ASSERT(bio);
+	WINPR_ASSERT(buf);
+
+	WINPR_BIO_NAMED* ptr = (WINPR_BIO_NAMED*)BIO_get_data(bio);
+
+	if (!buf)
+		return 0;
+
+	BIO_clear_flags(bio, BIO_FLAGS_READ);
+
+	DWORD readBytes;
+	BOOL ret = ReadFile(ptr->hFile, buf, size, &readBytes, NULL);
+	WLog_VRB(TAG, "transport_bio_named_read(%d)=%d read=%d", size, ret, readBytes);
+	if (!ret)
+	{
+		if (GetLastError() == ERROR_NO_DATA)
+			BIO_set_flags(bio, (BIO_FLAGS_SHOULD_RETRY | BIO_FLAGS_READ));
+		return -1;
+	}
+
+	if (readBytes == 0)
+	{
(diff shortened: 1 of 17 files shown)


libfreerdp/core/childsession.c
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	BIO_set_init(bio, 0);
+	BIO_set_flags(bio, 0);
+	return 1;
+}
+
+static int transport_bio_named_new(BIO* bio)
+{
+	BIO_set_flags(bio, BIO_FLAGS_SHOULD_RETRY);
```
akallabeth, line 189: `WINPR_ASSERT` all pointer arguments that must not be `NULL`

libfreerdp/core/childsession.c
```diff
@@ -0,0 +1,332 @@
...
+		BIO_set_data(bio, NULL);
+		free(ptr);
+	}
+
+	return 1;
+}
+
+BIO_METHOD* BIO_s_namedpipe(void)
+{
+	static BIO_METHOD* bio_methods = NULL;
```
akallabeth, line 221: sure it is a good idea to make this a single instance without thread sync?

libfreerdp/core/childsession.c
```diff
@@ -0,0 +1,332 @@
...
+		WLog_INFO(TAG, "child sessions aren't enabled");
+		if (!WTSEnableChildSessions(TRUE))
+		{
+			WLog_ERR(TAG, "error when calling WTSEnableChildSessions");
+			goto out;
+		}
+		WLog_INFO(TAG, "successfully enabled child sessions");
+	}
+
+	HANDLE hModule = LoadLibraryA("winsta.dll");
```
akallabeth, line 266: better scope this, `goto` does not like variables initialized after the first statement

libfreerdp/core/childsession.c
```diff
@@ -0,0 +1,332 @@
...
+	if (!hModule)
+		return FALSE;
+	WCHAR pipePath[0x80];
```
akallabeth, line 269: initialize with `={0};`

libfreerdp/core/childsession.c
```diff
@@ -0,0 +1,332 @@
...
+	char pipePathA[0x80];
```
akallabeth, line 270: initialize with `={0};`

libfreerdp/core/childsession.c
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	HRESULT hStatus = createChildSessionFn(pipePath, 0x80);
+	if (!SUCCEEDED(hStatus))
+	{
+		WLog_ERR(TAG, "error 0x%x when creating childSessionTransport", hStatus);
+		goto out;
+	}
+
+	ConvertWCharToUtf8(pipePath, pipePathA, sizeof(pipePathA));
```
akallabeth, line 288: better use `ConvertWCharNToUtf8` as the size of `pipePath` is known

libfreerdp/core/childsession.c
```diff
@@ -0,0 +1,332 @@
...
+	WLog_DBG(TAG, "child session is at '%s'", pipePathA);
+
+	HANDLE f = CreateFileW(pipePath, GENERIC_READ | GENERIC_WRITE, 0, (LPSECURITY_ATTRIBUTES)0x0,
```
akallabeth, line 291: useless casts, but use `NULL` instead of `0x0`

libfreerdp/core/nego.c
```diff
@@ -1661,6 +1648,11 @@ void nego_set_restricted_admin_mode_required(rdpNego* nego, BOOL RestrictedAdmin
 	nego->RestrictedAdminModeRequired = RestrictedAdminModeRequired;
 }
 
+void nego_set_childsession_enabled(rdpNego* nego, BOOL ChildSessionEnabled)
+{
+	nego->ConnectChildSession = ChildSessionEnabled;
```
akallabeth, line 1653: `WINPR_ASSERT` pointer arguments that must not be `NULL`

libfreerdp/core/transport.c
```diff
@@ -578,6 +579,16 @@ BOOL transport_connect(rdpTransport* transport, const char* hostname, UINT16 por
 	return status;
 }
 
+BOOL transport_connect_childsession(rdpTransport* transport)
+{
+	transport->frontBio = createChildSessionBio();
```
akallabeth, line 584: `WINPR_ASSERT` pointer arguments that must not be `NULL`

libfreerdp/core/utils.c
```diff
@@ -148,6 +148,9 @@ auth_status utils_authenticate(freerdp* instance, rdp_auth_reason reason, BOOL o
 	if (freerdp_shall_disconnect_context(instance->context))
 		return AUTH_FAILED;
 
+	if (settings->ConnectChildSession)
```
akallabeth, line 151: better use `freerdp_settings_get_bool`

akallabeth (commented):
Overall LGTM, but still need to check functionality.
some asserts/getters should be used here, otherwise looks good.

libfreerdp/core/childsession.c
hardening: Well do that for all the other BIOs...

akallabeth: https://www.openssl.org/docs/manmaster/man3/BIO_new.html also agrees

libfreerdp/core/utils.c
hardening: I'm not for using `freerdp_settings_get_bool` inside the FreeRDP library, accessing directly the struct member allows the compiler to do some type checks, while with `freerdp_settings_get_bool` we can have errors that we will discover only at runtime (if we ever hit that code path). And also the code is easier to read, which is also a big concern.
But I guess that we must have a discussion on the usage of these functions.

libfreerdp/core/childsession.c
hardening: I also agree, but we should probably do that for all the instances in a commit of its own.

hardening:
AFAICT I've fixed all the remarks.

libfreerdp/core/utils.c
akallabeth: well, we can have these checks with `freerdp_settings_get_bool` too (and many more static and runtime checks)
the direct access we have currently prevents implementing these as we never know the state of the fields as it might be changed in between outside the scope of such internal checks

akallabeth (approved):
LGTM

libfreerdp/core/childsession.c
hardening: Also fixed the NULL LPSECURITY_ATTRIBUTES

PR #9448, opened 2023-10-04
[server,drdynvc] use a hashtable for dynamic channels instead of an arraylist

hardening:
This speeds up all operations around dynamic channels.


akallabeth (approved):
LGTM

PR #9453, opened 2023-10-09
Missing api

akallabeth:
* make `pkg-config` files relocatable
* add missing `CMake` files for `winpr-tools`
* remove android `translatable=false` from translations
* fix android `release.properties` settings reading
* drop support for `fuse2` (unsupported since 2019)


akallabeth:

hardening (approved):

PR #9459, opened 2023-10-12
Rel mouse ptr

akallabeth:
* implements relative mouse events according to `[MS-RDPBCGR]`
* prefer standard relative events over custom channel
* also implement fastpath QOE callbacks


libfreerdp/core/capabilities.c
```diff
@@ -1351,6 +1351,10 @@ static BOOL rdp_apply_input_capability_set(rdpSettings* settings, const rdpSetti
 		}
 		if (settings->HasExtendedMouseEvent)
 			settings->HasExtendedMouseEvent = src->HasExtendedMouseEvent;
+		if (settings->HasRelativeMouseEvent)
+			settings->HasRelativeMouseEvent = src->HasRelativeMouseEvent;
```
pnowack, line 1355: Uhm, this condition + condition handling was duplicated from above

libfreerdp/core/fastpath.c
```diff
@@ -727,6 +727,39 @@ static BOOL fastpath_recv_input_event_mouse(rdpFastPath* fastpath, wStream* s, B
...
+	WINPR_ASSERT(s);
+
+	if (!Stream_CheckAndLogRequiredLength(TAG, s, 6))
+		return FALSE;
+
+	input = fastpath->rdp->input;
+
+	Stream_Read_UINT16(s, pointerFlags); /* pointerFlags (2 bytes) */
+	Stream_Read_INT16(s, xPos);          /* xPos (2 bytes) */
+	Stream_Read_INT16(s, yPos);          /* yPos (2 bytes) */
```
pnowack, line 749: There are not positions, but deltas. This should be named like in the docs: `xDelta, and `yDelta`.

include/freerdp/input.h
```diff
@@ -105,6 +107,8 @@ extern "C"
 	                                                           UINT16 code);
 	FREERDP_API BOOL freerdp_input_send_mouse_event(rdpInput* input, UINT16 flags, UINT16 x,
 	                                                UINT16 y);
+	FREERDP_API BOOL freerdp_input_send_rel_mouse_event(rdpInput* input, UINT16 flags, INT16 x,
+	                                                    INT16 y);
```
pnowack, line 113: `delta_x`, `delta_y` would fit better here (or camel case equivalent) to things more understandable upon reading

libfreerdp/core/input.c
```diff
@@ -222,6 +223,46 @@ static BOOL input_send_mouse_event(rdpInput* input, UINT16 flags, UINT16 x, UINT
 	return rdp_send_client_input_pdu(rdp, s);
 }
 
+static BOOL input_send_relmouse_event(rdpInput* input, UINT16 flags, INT16 x, INT16 y)
```
pnowack, line 226: `delta_x`, `delta_y` would fit better here in my opinion

libfreerdp/core/input.c
```diff
@@ -454,6 +495,48 @@ static BOOL input_send_fastpath_extended_mouse_event(rdpInput* input, UINT16 fla
 	return fastpath_send_input_pdu(rdp->fastpath, s);
 }
 
+static BOOL input_send_fastpath_relmouse_event(rdpInput* input, UINT16 flags, UINT16 x, UINT16 y)
```
pnowack, line 498: `delta_x`, `delta_y` would fit better here in my opinion

libfreerdp/core/input.c
```diff
@@ -222,6 +223,46 @@ static BOOL input_send_mouse_event(rdpInput* input, UINT16 flags, UINT16 x, UINT
...
+		WLog_ERR(TAG, "Sending relative mouse event, but no support for that");
+		return FALSE;
+	}
+
+	if (!freerdp_settings_get_bool(input->context->settings, FreeRDP_HasHorizontalWheel))
+	{
+		if (flags & PTR_FLAGS_HWHEEL)
+		{
+			WLog_WARN(TAG,
+			          "skip mouse event %" PRIu16 "x%" PRIu16 " flags=0x%04" PRIX16
```
pnowack, line 250: `skip mouse event` -> `skipping relative mouse event`

libfreerdp/core/input.c
```diff
@@ -454,6 +495,48 @@ static BOOL input_send_fastpath_extended_mouse_event(rdpInput* input, UINT16 fla
...
+		WLog_ERR(TAG, "Sending relative fastpath mouse event, but no support for that announced");
+		return FALSE;
+	}
+
+	if (!freerdp_settings_get_bool(input->context->settings, FreeRDP_HasHorizontalWheel))
+	{
+		if (flags & PTR_FLAGS_HWHEEL)
+		{
+			WLog_WARN(TAG,
+			          "skip mouse event %" PRIu16 "x%" PRIu16 " flags=0x%04" PRIX16
```
pnowack, line 524: `skip mouse event` -> `skipping relative mouse event`

pnowack (commented):
Did only do a very brief look on the changes so far

libfreerdp/core/input.c
pnowack: > Other thing: Isn't relative mouse input not a fastpath only event?

Ah, looks like it is slowpath too.

include/freerdp/input.h
```diff
@@ -105,6 +109,8 @@ extern "C"
 	                                                           UINT16 code);
 	FREERDP_API BOOL freerdp_input_send_mouse_event(rdpInput* input, UINT16 flags, UINT16 x,
 	                                                UINT16 y);
+	FREERDP_API BOOL freerdp_input_send_rel_mouse_event(rdpInput* input, UINT16 flags, INT16 x,
```
hardening, line 112: Perhaps this PR should also have the function for sending QoEEvent packets...

hardening:
AFAICT no consensus emerged from #9431, and I'm seeing some `freerdp_settings_get_bool` calls in "hot paths" like incoming packets treatments. These calls will eat more CPU cycles than what is done in the rest of the function, let's use direct access to the settings for these.

akallabeth:
> AFAICT no consensus emerged from #9431, and I'm seeing some `freerdp_settings_get_bool` calls in "hot paths" like incoming packets treatments. These calls will eat more CPU cycles than what is done in the rest of the function, let's use direct access to the settings for these.

no hot paths here

> AFAICT no consensus emerged from #9431, and I'm seeing some `freerdp_settings_get_bool` calls in "hot paths" like incoming packets treatments. These calls will eat more CPU cycles than what is done in the rest of the function, let's use direct access to the settings for these.

no hot paths here, we´re talking keyboard/mouse input and capabilities (once or twice called per connection)

libfreerdp/core/capabilities.c
akallabeth: no, this is the client - server ping pong.
only if locally enabled even consider accepting the value of the remote

pnowack: Oh, right indeed. Both setting names have the same length and I did not realize, that the first one was about the extended mouse event and the second about the relative one (for some reason it looked to me as line duplication).

> only if locally enabled even consider accepting the value of the remote

yeah, right.

include/freerdp/input.h
akallabeth: added

libfreerdp/core/input.c
akallabeth: copy & paste error, wheel events are not allowed

include/freerdp/settings.h
```diff
@@ -100,6 +100,7 @@ extern "C"
 #define CS_MONITOR 0xC005
 #define CS_MCS_MSGCHANNEL 0xC006
 #define CS_MONITOR_EX 0xC008
+#define CS_UNUSED1 0xC00C
 #define CS_MULTITRANSPORT 0xC00A
```
pnowack, line 104: Nit: Looks like these values are sorted, so `CS_UNUSED1` should be under `CS_MULTITRANSPORT` then

libfreerdp/core/fastpath.h
```diff
@@ -91,7 +91,9 @@ enum FASTPATH_INPUT_EVENT_CODE
 	FASTPATH_INPUT_EVENT_MOUSE = 0x1,
 	FASTPATH_INPUT_EVENT_MOUSEX = 0x2,
 	FASTPATH_INPUT_EVENT_SYNC = 0x3,
-	FASTPATH_INPUT_EVENT_UNICODE = 0x4
+	FASTPATH_INPUT_EVENT_UNICODE = 0x4,
+	TS_FP_RELPOINTER_EVENT = 0x5,
+	TS_FP_QOETIMESTAMP_EVENT = 0x6
```
pnowack, line 96: Nit: `2.2.8.1.2.2 Fast-Path Input Event (TS_FP_INPUT_EVENT)` defines these name-wise as `FASTPATH_INPUT_EVENT_MOUSEREL` and `FASTPATH_INPUT_EVENT_QOE_TIMESTAMP`.

Or was the intention here a shorter name? I think we should keep the convention consistent here (I'm fine with a further API break (we already break in FreeRDP3)).

libfreerdp/core/capabilities.c
```diff
@@ -1351,6 +1351,10 @@ static BOOL rdp_apply_input_capability_set(rdpSettings* settings, const rdpSetti
...
+		if (freerdp_settings_get_bool(settings, FreeRDP_HasQoeEvent))
```
pnowack, line 1356: Maybe this condition should be extended to also check, whether the fastpath is available (do we know this at this point in the connection sequence already?) (because according to `2.2.7.1.6 Input Capability Set (TS_INPUT_CAPABILITYSET)` there is no slowpath equivalent)?

pnowack (approved):
Don't know about the SDL client part here, and don't know yet how MS Windows handles relative input (since that event also seems to allow click events, but the resulting abs mouse pos is usually not the same after applying the relative motion). The other parts look fine to me.

libfreerdp/core/capabilities.c
akallabeth: that should be checked if used, not this part here.

include/freerdp/settings.h
akallabeth: true, but too lazy :yum:

libfreerdp/core/fastpath.h
pnowack: No, the event codes all start with `FASTPATH_INPUT_EVENT_`. See also: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/76c4dd59-7ba0-445d-a03c-885212ab80f6

pnowack: `TS_FP_INPUT_EVENT` is just the name of the fastpath PDU in general

PR #9468, opened 2023-10-16
fixes rdp_read_info_null_string api

DVeron-RC:
Removed buggy "isNullTerminated" parameter.
RDP specs tells that cbLen includes the nullTerminator if there is one.
There is no need to differenciate the behaviour if there is a nullTerminator.

To reproduce the bug, give a null terminated clientAddress of max size ie 80.

### DIFF
--- a/libfreerdp/core/info.c
+++ b/libfreerdp/core/info.c
@@ -74,8 +74,7 @@ static const struct info_flags_t info_flags[] = {
 };
 
 static BOOL rdp_read_info_null_string(rdpSettings* settings, size_t id, const char* what,
-                                      UINT32 flags, wStream* s, size_t cbLen, size_t max,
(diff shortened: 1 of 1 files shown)


akallabeth:
@DVeron-RC this code is there because older windows versions did not adhere to the spec.
so, aside from this being something a bit ugly, does it fix an actual bug you encountered?

DVeron-RC:
Yes, a client send us a cbClientAddress of size 80 (which is legal according to the spec) and the old code decreamented the max to 79 and refused it. This modification is more permissive ie the function will accept larger field in case they were null terminated ie 80 char for cbClientAddress and 512 for cbClientDir.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #9471, opened 2023-10-16
server: Add channel handling for location channel ([MS-RDPEL])

pnowack:
Quoting the main commit here:

```
With the location channel, an RDP server can redirect the location of
the user from the client side to the server side.
The PDUs are almost the same as in the documentation, except for the
encoded types, these ones are here already parsed.
Optional values are accessed via pointers. If the pointer of a value is
NULL, then that value was not supplied by the client side.
```

I was able to successfully test the following PDUs:
`PDUTYPE_SERVER_READY`
`PDUTYPE_CLIENT_READY`
`PDUTYPE_BASE_LOCATION3D`

I could not check the following PDUs, since I did not receive them:
`PDUTYPE_LOCATION2D_DELTA`
`PDUTYPE_LOCATION3D_DELTA`

mstsc only advertised version 1.0.0, so I got lesser data than wanted, but at least it worked.

As mentioned on the Matrix channel, there are quite some things wrongly documented, like the NON-optional flags value in the `PDUTYPE_SERVER_READY` PDU, or the non-optional speed value, that rather looks optional, etc..


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test please

PR #9479, opened 2023-10-18
codec/audio: Add missing format id for Opus codec

pnowack:
```
Id taken from [0].

[0]: https://learn.microsoft.com/en-us/windows/win32/medfound/audio-subtype-guids
```

### DIFF
--- a/include/freerdp/codec/audio.h
+++ b/include/freerdp/codec/audio.h
@@ -192,6 +192,8 @@ extern "C"
 #endif /* !WAVE_FORMAT_LUCENT_G723 */
 #define WAVE_FORMAT_AAC_MS 0xA106
 
+#define WAVE_FORMAT_OPUS 0x704F
+
 #define WAVE_FORMAT_EXTENSIBLE 0xFFFE
 
 	/**


akallabeth:
not sure that the referenced types are the correct ones, the ones here are somewhat similar to https://learn.microsoft.com/en-us/windows/win32/api/mmeapi/ns-mmeapi-waveformatex (although specifically mentioned in the rdpsdn and rdpai specs

pnowack:
> not sure that the referenced types are the correct ones, the ones here are somewhat similar to https://learn.microsoft.com/en-us/windows/win32/api/mmeapi/ns-mmeapi-waveformatex (although specifically mentioned in the rdpsdn and rdpai specs

There is unfortunately no 'one-site' in msdn where all format ids are referenced. In https://learn.microsoft.com/en-us/windows/win32/medfound/audio-subtype-guids, there are also some formats listed, that we have in `[MS-RDPEA]` referenced (`WAVE_FORMAT_PCM (1)`, `WAVE_FORMAT_MPEGLAYER3 (0x0055)`).

And some formats in `include/freerdp/codec/audio.h` are the same formats as in https://learn.microsoft.com/en-us/windows/win32/medfound/audio-subtype-guids (`WAVE_FORMAT_IEEE_FLOAT (0x0003)`, `WAVE_FORMAT_MPEG (0x0050)`, `WAVE_FORMAT_WMAUDIO2 (0x0161)`, `WAVE_FORMAT_WMAUDIO3 (0x0162)`, `WAVE_FORMAT_WMAUDIO_LOSSLESS (0x0163)`, `WAVE_FORMAT_DOLBY_AC3_SPDIF (0x0092)`)

akallabeth (approved):
well, then lets merge!

PR #9492, opened 2023-10-19
Xfreerdp drop gdi

akallabeth:
* `GDI` modes are obsolete
* the `GDI` to `X11` mapping is incomplete, better use the complete mapping in core library


hardening (approved):
Looks ok for me, although I don't know all the details on this part

mfleisz (approved):

PR #9496, opened 2023-10-20
[cmdline] fix alphabetical order

manio:


client/common/cmdline.h
```diff
@@ -364,6 +364,9 @@ static const COMMAND_LINE_ARGUMENT_A global_cmd_args[] = {
 	{ "port", COMMAND_LINE_VALUE_REQUIRED, "<number>", NULL, NULL, -1, NULL, "Server port" },
 	{ "suppress-output", COMMAND_LINE_VALUE_BOOL, NULL, BoolValueTrue, NULL, -1, NULL,
```
akallabeth, line 365: should reorder this too

manio:
I only fixed problem introduced by my change. But OK - now I hope I fixed other order problems too...

manio:
I also added a helper script for checking this - maybe you'll find it useful in the future...

akallabeth:
@manio thank you for the effort, but I´ve replaced it with https://github.com/FreeRDP/FreeRDP/pull/9499
I thing (as we already create a copy there) it is easier just to sort that list than keeping the struct definition sorted (you always need to keep that in mind whenever someone adds something)

PR #9499, opened 2023-10-21
[client,common] print command line sorted

akallabeth:
always sort the command line arguments before listing them.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -410,11 +410,28 @@ static size_t print_description(const char* text, size_t start_offset, size_t cu
 	return current;
 }
 
-static void freerdp_client_print_command_line_args(const COMMAND_LINE_ARGUMENT_A* arg)
+static int cmp_cmdline_args(const void* pva, const void* pvb)
 {
-	if (!arg)
+	const COMMAND_LINE_ARGUMENT_A* a = (const COMMAND_LINE_ARGUMENT_A*)pva;
+	const COMMAND_LINE_ARGUMENT_A* b = (const COMMAND_LINE_ARGUMENT_A*)pvb;
+
+	if (!a->Name && !b->Name)
+		return 0;
+	if (!a->Name)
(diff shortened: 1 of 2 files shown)


client/common/cmdline.c
```diff
@@ -410,11 +410,28 @@ static size_t print_description(const char* text, size_t start_offset, size_t cu
 	return current;
 }
 
-static void freerdp_client_print_command_line_args(const COMMAND_LINE_ARGUMENT_A* arg)
+static int sort_cmdline_args(const void* pva, const void* pvb)
```
pnowack, line 413: This function itself does not sort things, just compare (as cmp func for the sorting):
```suggestion
static int cmp_cmdline_args(const void* pva, const void* pvb)
```

akallabeth:
@freerdp-bot test please

client/common/cmdline.c
akallabeth: ok, updated.

client/common/cmdline.c
```diff
@@ -410,11 +410,28 @@ static size_t print_description(const char* text, size_t start_offset, size_t cu
...
 }
 
-static void freerdp_client_print_command_line_args(const COMMAND_LINE_ARGUMENT_A* arg)
+static int cmp_cmdline_args(const void* pva, const void* pvb)
 {
-	if (!arg)
+	const COMMAND_LINE_ARGUMENT_A* a = (const COMMAND_LINE_ARGUMENT_A*)pva;
+	const COMMAND_LINE_ARGUMENT_A* b = (const COMMAND_LINE_ARGUMENT_A*)pvb;
+
+	if (!a->Name && !b->Name)
```
mfleisz, line 418: Shouldn't this actually be an `assert(a->Name && b->Name)` (and the following name checks removed)?
Can't think of any reason this could be a valid case

mfleisz (approved):

PR #9500, opened 2023-10-23
[channel,rail] keep reference to rdpContext

akallabeth:
we need to access `rdpContext` from `client_rail_server_start_cmd` so move the function implementation to channel code.


akallabeth:
@freerdp-bot test please

mfleisz (approved):

PR #9513, opened 2023-10-24
server: Add channel handling for mouse cursor channel ([MS-RDPEMSC]) 

pnowack:
```
The mouse cursor channel enables remoting of the mouse cursor (bitmap)
over a DVC. The main use case is UDP, as only virtual channels can be
transported via UDP in RDP.
```

I was able to successfully test all three PDUs (`CapsAdvertise`, `CapsConfirm`, `MouseptrUpdate`).

Regarding the update types, I did not test the `TS_UPDATETYPE_MOUSEPTR_SYSTEM_DEFAULT` and `TS_UPDATETYPE_MOUSEPTR_LARGE_POINTER`. However, since I tested the `TS_UPDATETYPE_MOUSEPTR_SYSTEM_NULL` and `TS_UPDATETYPE_MOUSEPTR_POINTER` update types successfully, these (former two listed ones) two should also work, since they are very similar to the other two here (the latter listed ones here).


channels/rdpemsc/server/mouse_cursor_main.c
```diff
@@ -0,0 +1,727 @@
...
+		               "calling in this state is not possible!");
+		return ERROR_INVALID_STATE;
+	}
+
+	mouse_cursor->externalThread = externalThread;
+
+	return error;
+}
+
+static UINT mouse_cursor_server_open_channel(mouse_cursor_server* mouse_cursor)
```
akallabeth, line 73: missing `WINPR_ASSERT(context)` before dereferencing

channels/rdpemsc/server/mouse_cursor_main.c
```diff
@@ -0,0 +1,727 @@
...
+	{
+		WLog_ERR(TAG, "WTSQuerySessionInformationA failed!");
+		return ERROR_INTERNAL_ERROR;
+	}
+
+	mouse_cursor->SessionId = (DWORD)*pSessionId;
+	WTSFreeMemory(pSessionId);
+	hEvent = WTSVirtualChannelManagerGetEventHandle(mouse_cursor->context.vcm);
+
+	if (WaitForSingleObject(hEvent, 1000) == WAIT_FAILED)
```
akallabeth, line 96: What is the reason to use a timeout here (and a magic number at that)?

channels/rdpemsc/server/mouse_cursor_main.c
```diff
@@ -0,0 +1,727 @@
...
+		default:
+			WINPR_ASSERT(FALSE);
+			break;
+	}
+
+	return mouse_cursor_server_packet_send(context, s);
+}
+
+static void write_point16(wStream* s, const TS_POINT16* point16)
+{
```
akallabeth, line 587: missing `WINPR_ASSERT(point16)` before dereferencing

akallabeth (changes requested):
Overall LGTM, only some minor nitpicking

channels/rdpemsc/server/mouse_cursor_main.c
pnowack: Indeed, fixed.

channels/rdpemsc/server/mouse_cursor_main.c
pnowack: I actually copied that part from the other DVCs, not sure the overall origin (whether it was audin or some other DVC)

akallabeth: ok, I´m asking because with `FreeRDP 1` we had issues with events (e.g. not all relevant things were notified, so the hackish way was to use a polling loop or timeouts)

pnowack: not sure about the origin, but it is indeed unneeded, as the event handle for the channel queue is created, when the VCM is created. I removed it in the last push, it is unneeded.

akallabeth (approved):
LGTM

PR #9515, opened 2023-10-25
[crypto,tls] only print fingerprint in log

akallabeth:
printing the whole PEM to log is too verbose, just use the fingerprint instead.


pnowack (approved):
LGTM

PR #9520, opened 2023-10-29
[channels,rdpdr] fix user loggedon handling

akallabeth:
PAKID_CORE_USER_LOGGEDON might happen in quite a lot of combinations not obvious from the spec at first glance. Relax state checks so that all valid combinations are allowed.


mfleisz (approved):
LGTM

PR #9525, opened 2023-11-01
Externalize custom channels

akallabeth:
split custom channels from core library, prefer `/dvc:<channelname>` or `/svc:<channelname>` syntax


akallabeth:
@freerdp-bot test please

PR #9528, opened 2023-11-01
rdpsnd_pulse: optimize writes and allow reconnections

i-garrison:
- Eliminate synchronous write loop
- Allow reconnecting to pulseaudio server

### DIFF
--- a/channels/rdpsnd/client/pulse/rdpsnd_pulse.c
+++ b/channels/rdpsnd/client/pulse/rdpsnd_pulse.c
@@ -375,7 +375,7 @@ static BOOL rdpsnd_pulse_open(rdpsndDevicePlugin* device, const AUDIO_FORMAT* fo
 
 	if (pulse->latency > 0)
 	{
-		buffer_attr.maxlength = pa_usec_to_bytes(pulse->latency * 2 * 1000, &pulse->sample_spec);
+		buffer_attr.maxlength = (UINT32)-1;
 		buffer_attr.tlength = pa_usec_to_bytes(pulse->latency * 1000, &pulse->sample_spec);
 		buffer_attr.prebuf = (UINT32)-1;
 		buffer_attr.minreq = (UINT32)-1;
@@ -551,6 +551,7 @@ static BOOL rdpsnd_pulse_set_volume(rdpsndDevicePlugin* device, UINT32 value)
 static UINT rdpsnd_pulse_play(rdpsndDevicePlugin* device, const BYTE* data, size_t size)
 {
 	size_t length;
+	void* pa_data;
 	int status;
 	pa_usec_t latency;
 	int negative;
@@ -563,16 +564,16 @@ static UINT rdpsnd_pulse_play(rdpsndDevicePlugin* device, const BYTE* data, size
 
 	while (size > 0)
 	{
-		while ((length = pa_stream_writable_size(pulse->stream)) == 0)
-			pa_threaded_mainloop_wait(pulse->mainloop);
+		length = size;
 
-		if (length == (size_t)-1)
+		status = pa_stream_begin_write(pulse->stream, &pa_data, &length);
+
+		if (status < 0)
 			break;
 
-		if (length > size)
-			length = size;
+		memcpy(pa_data, data, length);
 
-		status = pa_stream_write(pulse->stream, data, length, NULL, 0LL, PA_SEEK_RELATIVE);
+		status = pa_stream_write(pulse->stream, pa_data, length, NULL, 0LL, PA_SEEK_RELATIVE);
 
 		if (status < 0)
 		{
--- a/channels/rdpsnd/client/pulse/rdpsnd_pulse.c
+++ b/channels/rdpsnd/client/pulse/rdpsnd_pulse.c
@@ -406,11 +406,11 @@ static BOOL rdpsnd_pulse_open_stream(rdpsndDevicePlugin* device)
 
 	if (pulse->latency > 0)
 	{
-		buffer_attr.maxlength = (UINT32)-1;
+		buffer_attr.maxlength = UINT32_MAX;
 		buffer_attr.tlength = pa_usec_to_bytes(pulse->latency * 1000, &pulse->sample_spec);
-		buffer_attr.prebuf = (UINT32)-1;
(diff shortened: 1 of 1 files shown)


akallabeth (commented):
:+1: for the async write, but for the reconnect I lack imagination for a use case.
can you describe why this is required?

akallabeth:
@freerdp-bot test

i-garrison:
> 👍 for the async write, but for the reconnect I lack imagination for a use case. can you describe why this is required?
Pulseaudio server connection can fail if pulseaudio server is restarted for some reason.
As far as I understand, currently freerdp will become silent until you restart it too. Guess this can also be handled on upper layer by reloading rdpsnd_pulse, but I do not see any API to notify the upper layer to reload plugin.

i-garrison:
> @freerdp-bot test
pushed style changes

akallabeth:
> > 👍 for the async write, but for the reconnect I lack imagination for a use case. can you describe why this is required?
> > Pulseaudio server connection can fail if pulseaudio server is restarted for some reason.
> > As far as I understand, currently freerdp will become silent until you restart it too. Guess this can also be handled on upper layer by reloading rdpsnd_pulse, but I do not see any API to notify the upper layer to reload plugin.

ok, but why should we handle this? Is there some reason I fail to see?

1. restarting the daemon requires restarting dependencies - see most audio applications
2. this does not happen during normal operation - we also do not handle stuff like package upgrades (changed files/plugins) and let the os handle that

i-garrison:
> > > 👍 for the async write, but for the reconnect I lack imagination for a use case. can you describe why this is required?
> > > Pulseaudio server connection can fail if pulseaudio server is restarted for some reason.
> > > As far as I understand, currently freerdp will become silent until you restart it too. Guess this can also be handled on upper layer by reloading rdpsnd_pulse, but I do not see any API to notify the upper layer to reload plugin.
> 
> ok, but why should we handle this? Is there some reason I fail to see?
> 
>     1. restarting the daemon requires restarting dependencies - see most audio applications

Well this is really optional feature, but it is not hard to have for `rdpsnd_pulse`.

I would disagree regarding most audio applications, best examples of the contrary are desktop sound plugins and sound control applets like `plasma-pa` and corresponding Gnome thing. Those do reattach to underlying audio subsystem on the fly, and I would expect freerdp to do the same if possible.

>     2. this does not happen during normal operation - we also do not handle stuff like package upgrades (changed files/plugins) and let the os handle that

No, not during normal operation - still it is very handy for sound to reappear should things break unexpectedly.

akallabeth:
@freerdp-bot test

channels/rdpsnd/client/pulse/rdpsnd_pulse.c
```diff
@@ -375,7 +406,7 @@ static BOOL rdpsnd_pulse_open(rdpsndDevicePlugin* device, const AUDIO_FORMAT* fo
 
 	if (pulse->latency > 0)
 	{
-		buffer_attr.maxlength = pa_usec_to_bytes(pulse->latency * 2 * 1000, &pulse->sample_spec);
+		buffer_attr.maxlength = (UINT32)-1;
```
akallabeth, line 409: looks wrong, why change this?

akallabeth (commented):
Ok, good point (there are some systems / hardware combos that have unstable audio)

a few more questions, but otherwise LGTM

channels/rdpsnd/client/pulse/rdpsnd_pulse.c
i-garrison: This is the absolute maximum number of bytes that can be stored in the buffer. We want it so large that incoming data always fits (therefore `maxlength=-1` here). If I understand correctly, remote side is feeding us appropriately sized chunks so we just need to put them into stream.

channels/rdpsnd/client/pulse/rdpsnd_pulse.c
```diff
@@ -48,6 +50,8 @@ typedef struct
 	pa_stream* stream;
 	UINT32 latency;
 	UINT32 volume;
+	time_t reconnect_delay_seconds;
```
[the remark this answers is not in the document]
i-garrison: This is to protect from unfortunate case where pulseaudio server went down and never recovers, limiting reconnection rate. This happens in practice in usual conditions, e.g. if you are using remote pulseaudio server over network to feed your speakers via rpi4.

Whether the default reconnect delay is good enough I do not know, so I added a parameter to configure that.

channels/rdpsnd/client/pulse/rdpsnd_pulse.c
i-garrison: Pulseaudio docs specify these values as `(uint32_t) -1` so that is how it is done here and in 3 (unmodified) lines below https://github.com/FreeRDP/FreeRDP/blob/master/channels/rdpsnd/client/pulse/rdpsnd_pulse.c#L380-L382

i-garrison: Here is the reference from pulseaudio docs https://freedesktop.org/software/pulseaudio/doxygen/structpa__buffer__attr.html#abef20d3a6cab53f716846125353e56a4

akallabeth: @i-garrison ok, but we´re supporting `C99` (and newer), so https://en.cppreference.com/w/c/types/integer is available and these casts are not required and the proper maximum of the type is used

i-garrison: Technically we can use proper `c99` max values here but I was looking to keep this closer to pulseaudio docs.

Do you want me to change all of these to `c99` constants?

i-garrison: I still believe this particular case could actually be an exception from general rule since it is 1) pretty well defined in `C` and 2) defined with corresponding integer cast in the 3rdparty API.

In any case, I now amended these (4) places to use `UINT32_MAX` constant, in extra commit.

akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #9551, opened 2023-11-08
(clipboard) Make sure there is always a Client Format List PUD after …

XiaopengZHOU:
After initializing the Clipboard virtual channel once, if the client (FreeRDP) receives a Server Monitor Ready PDU again, only the Client Clipboard Capabilities PDU message is sent back to the server. The Client Format List PDU to finalize the initialization sequence is missing.

### DIFF
--- a/channels/cliprdr/client/cliprdr_main.c
+++ b/channels/cliprdr/client/cliprdr_main.c
@@ -621,6 +621,9 @@ static UINT cliprdr_client_capabilities(CliprdrClientContext* context,
 
 	Stream_Write_UINT32(s, flags); /* generalFlags */
 	WLog_Print(cliprdr->log, WLOG_DEBUG, "ClientCapabilities");
+
+	cliprdr->initialFormatListSent = FALSE;
+
 	return cliprdr_packet_send(cliprdr, s);
 }
 


hardening:

akallabeth:
@XiaopengZHOU where do you get a reinitialization of the clipboard channel? would help to check if your changes are good ;)

XiaopengZHOU:
Hi akallabeth,

I am in charge of maintaining an RDP proxy which can connect a user to several RDP servers successfully in the same client RDP session. In this type of scenario, each of the servers will separately initialize the virtual clipboard channel. This behavior is supported by Microsoft Remote Desktop Connection.

FreeRDP is willing to reply to Server Monitor Ready PDU as many times as he receives it. It will also make more sense for the Client Clipboard Capabilities PDU (reply to Monitor Ready PDU) to be followed by the Client Format List PDU, so that the initialization sequence is complete.  :)

akallabeth (approved):
@XiaopengZHOU ok, I understand the issue with the incomplete (re) initialization, but was not sure (as the spec does not mention that) if that is allowed at all.
so, LGTM then

PR #9556, opened 2023-11-10
[ifdef] properly check for defined

akallabeth:


include/freerdp/api.h
```diff
@@ -52,7 +52,7 @@
 #endif
 #endif
 #else
-#if __GNUC__ >= 4
+#if defined(__GNUC__) && (__GNUC__ >= 4)
```
pnowack, line 55: Just a question: Can we actually fall into this here? I.e. that, when `__GNUC__` is not defined, that this condition is evaluated here as `true`?

pnowack (approved):

hardening (approved):

include/freerdp/api.h
akallabeth: @pnowack if you are not on windows and do not compile with `gcc`/clang` then yes

PR #9562, opened 2023-11-15
fix abnormal exit when remove the USB flash drive

LucasMou:
fix abnormal exit when remove the USB flash drive


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #9564, opened 2023-11-15
cmake: Revert warning compiler flag for MSVC

mfleisz:
MSVC does not implemented /Wall very well as it floods the output with a
plethora of warnings from system headers.

This greatly reduces compile performance so this change reverts /Wall
back to /W3.

### DIFF
--- a/cmake/CompilerFlags.cmake
+++ b/cmake/CompilerFlags.cmake
@@ -27,7 +27,7 @@ if (ENABLE_WARNING_VERBOSE)
 		endforeach()
 
 		set(C_WARNING_FLAGS
-			/Wall
+			/W3
 		)
 	else()
 		set(C_WARNING_FLAGS
--- a/cmake/CompilerFlags.cmake
+++ b/cmake/CompilerFlags.cmake
@@ -27,7 +27,8 @@ if (ENABLE_WARNING_VERBOSE)
 		endforeach()
 
 		set(C_WARNING_FLAGS
-			/W3
+			/W4
+			/wo4324
 		)
 	else()
 		set(C_WARNING_FLAGS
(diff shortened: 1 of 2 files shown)


cmake/CompilerFlags.cmake
```diff
@@ -27,7 +27,7 @@ if (ENABLE_WARNING_VERBOSE)
 		endforeach()
 
 		set(C_WARNING_FLAGS
-			/Wall
+			/W3
```
akallabeth, line 30: ok, are these there with `/W4` too? (tend to enable the most useful verbose warning level by default, disable with `-DENABLE_VERBOSE_WARNINGS=OFF`)

mfleisz: `/W4` does not cause any warnings for system headers but it causes a lot of `C4324` warnings with `rdpSettings`:

`'rdp_settings': structure was padded due to alignment specifier`

So if we want to enable `/W4` I think we should disable `C4324` - shall I make that change?

akallabeth: as we´ve done this for `gcc` and `clang` as well (`-Wno-padded` et al) yes, would be a good idea.

mfleisz: @akallabeth updated accordingly 👍

cmake/CompilerFlags.cmake
```diff
@@ -27,7 +27,8 @@ if (ENABLE_WARNING_VERBOSE)
 		endforeach()
 
 		set(C_WARNING_FLAGS
-			/Wall
+			/W4
+			/wo4324
```
akallabeth, line 31: shouldn´t that be `/wd4324`?

akallabeth (approved):
LGTM

PR #9575, opened 2023-11-20
codec/dsp: Add support for decoding Opus encoded streams

pnowack:
```
The Opus codec is a modern free audio codec, that is also royalty-free.
Adding support for it will allow clients and servers supporting it to
transfer audio in similar efficient way like with AAC.
So, add support it.
```


pnowack:
The CI probably needs an update (to include the `opus` pkg)

akallabeth (approved):
LGTM

PR #9589, opened 2023-11-25
[cmdline] add missing return statement

Duncaen:
I had this bug in my own code and was surprised it doesn't produce any compiler warnings. After a quick search for similar code patterns `if (...) NULL;` I was able to find this bug in FreeRDP.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@Duncaen nice catch, thank you!

PR #9593, opened 2023-11-27
Fix DPI issue when parent-window specified

StefanKoell:
2nd try (git hates me!)

This change should fix DPI scaling issue when wfreerdp.exe is started using parent-handle argument of a control/window with SystemAware or PerMonitorDpiAware mode.

This was tested on two machines but I'm not sure if the usage of these APIs is correct. Please let me know if there's a better way or if there are issues with my suggested implementation.

Thank you!

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -1386,6 +1386,28 @@ static int wfreerdp_client_start(rdpContext* context)
 
 	wfc->hWndParent = hWndParent;
 
+	if (freerdp_settings_get_bool(settings, FreeRDP_EmbeddedWindow))
+	{
+		typedef UINT(WINAPI * GetDpiForWindow_t)(HWND hwnd);
+		typedef BOOL(WINAPI * SetProcessDPIAware_t)(void);
+
+		HMODULE module = GetModuleHandle(_T("User32"));
+		if (module)
+		{
+			GetDpiForWindow_t pGetDpiForWindow =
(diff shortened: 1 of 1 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@StefanKoell looks like this needs some preprocessor conditional ( https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170 ) as the 32bit build  is running an older windows version.

StefanKoell:
Is it possible to just include the code branch (DPI check) on a certain Windows version? Or is there a Windows version I can use as preprocessor conditional which is suitable for wfreerdp.exe?

akallabeth:
@StefanKoell did push a fix for that on your branch.

akallabeth:
@freerdp-bot test

mfleisz:
Since the DPI functions are only available since a certain Win10 version (1607) run-time dynamic linking should be used instead of load-time dynamic linkage.

For an example refer to `client/Windows/wf_cliprdr.c` (search for user32.dll).

StefanKoell:
@akallabeth No, I didn't push anything else.

akallabeth:
@mfleisz good point.

akallabeth:
@StefanKoell I meant I did push ;)
but just as @mfleisz said, better way is to use `GetProcAddress` to get a function pointer to the library and only use it if it exists.

PR #9598, opened 2023-11-29
[client,common] enable gfx channel also on /gfx only

akallabeth:


pnowack (approved):
Change looks good to me. Also works fine.

hardening:
can you still disable egfx with these changes ?

pnowack:
> can you still disable egfx with these changes ?

@hardening Yes, `parse_gfx_options()` is only called, when a `gfx` option was specified

PR #9606, opened 2023-12-05
Server channel event

akallabeth:
* Remove timed wait after `WTSQuerySessionInformationA` before `WTSVirtualChannelOpenEx` (looks like some leftover from really long ago, anyone has a clue why that was added?)
* Do some cleanups / fixes in `server.c` core API
* Replace timed `WaitForMultipleObjects` with `INFINITE` (if there are issues we better fix them properly instead of using polling)

### DIFF
--- a/channels/ainput/server/ainput_main.c
+++ b/channels/ainput/server/ainput_main.c
@@ -89,9 +89,6 @@ static BOOL ainput_server_is_open(ainput_server_context* context)
  */
 static UINT ainput_server_open_channel(ainput_server* ainput)
 {
-	DWORD Error;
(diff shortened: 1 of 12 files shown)


pnowack (changes requested):
Found a bug: The audin server does not open anymore with this PR. g-r-d waits endlessly on the response for the AUDIO_INPUT DVC, but does not get an answer (and thus shuts that DVC down due to lack of response).
If I don't apply this PR here, then this bug does not happen

akallabeth:
@pnowack ok, removed the offending commit, but the rest I´ve kept.

PR #9620, opened 2023-12-08
Fix some compilation issues.

llyzs:


hardening:
Hum can FreeRDP really work without any timer support ?

llyzs:
> Hum can FreeRDP really work without any timer support ?

"CreateWaitableTimer" search result shows that it is used only in client implementation, and libfreerdp does not use this function. It looks like the timer implementation in winpr is just a convenience for client implementation but not a required component.

We are currently investigating the possibility to use freerdp in WebAssembly and the wasm environment provides no such implementation.

winpr/libwinpr/crypto/cipher.c
```diff
@@ -93,7 +93,7 @@ static WINPR_RC4_CTX* winpr_RC4_New_Internal(const BYTE* key, size_t keylen, BOO
 	if (!evp)
 		goto fail;
 
-	EVP_CIPHER_CTX_init(ctx->ctx);
+	EVP_CIPHER_CTX_reset(ctx->ctx);
```
akallabeth, line 96: this is openssl >= 3.0 only, needs a preprocessor switch

akallabeth (changes requested):

akallabeth (approved):
Sorry, did miss it was already part of 1.1.1
only 1.0 is lacking this, so LGTM

PR #9629, opened 2023-12-11
[winpr,crt] fix ICU string convert

akallabeth:
use ucnv_convert to make conversion endian safe


matoro:
> @matoro I´ve fixed `ICU` string conversions, now using `ucnv_convert` which allows to specify the exact locales to convert from/to. this way the `UTF-16LE` strings should now be properly converted. Could you please check?

Can you just cherry-pick the new CI flow onto this branch and see what happens?

PR #9630, opened 2023-12-11
[winpr,crt] fix ICU string convert

akallabeth:
use ucnv_convert to make conversion endian safe

(cherry picked from commit d47058011e7c5e628e47be0d968d0bb2ded63f99)

backport of #9629


akallabeth:
@lligurio looks like the fuzzer is no longer compiling? could you have a look?

PR #9632, opened 2023-12-11
Bigendian

akallabeth:
A first try to get proper test coverage for alternate architectures:
* Build the whole thing on these architectures and let the `ctests` run
* Fix string constants, make them always `UTF-16-LE`

What is not yet addressed:

* ~~`TestFreeRDPCodecProgressive` has some big endian issue~~ reverted commit 6ba4aad9ab0d3a039f982494dc40c7fc56a35376.
* ~~`TestBacktrace` and `TestASN1` fail on `ppc64le`~~ disabled backtrace support, not working in container
* `TestConnect` has an issue with all of the builds, most likely some restriction of the container, needs to be investigated

### DIFF
--- a/.github/workflows/alt-architectures.yml
+++ b/.github/workflows/alt-architectures.yml
@@ -79,6 +79,9 @@ jobs:
               libsdl2-ttf-dev \
               libwebkit2gtk-4.0-dev \
               libopus-dev \
+              libwebp-dev \
+              libpng-dev \
+              libjpeg-dev \
               liburiparser-dev \
               cmake \
               clang
--- a/.github/workflows/alt-architectures.yml
+++ b/.github/workflows/alt-architectures.yml
@@ -1,0 +1,97 @@
+name: '[arm,ppc,ricsv] architecture builds'
+on:
+  workflow_dispatch:
+    branches: [ master, stable* ]
+  schedule:
+    - cron: '30 5 * * SUN'
+
+jobs:
+  build_job:
+    runs-on: ubuntu-latest
+    name: "Test on ${{ matrix.distro }}/${{ matrix.arch }}"
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: armv6
+            distro: bullseye
+          - arch: armv7
+            distro: bullseye
+          - arch: aarch64
+            distro: bullseye
+          - arch: s390x
+            distro: bullseye
+          - arch: ppc64le
+            distro: bullseye
+          - arch: riscv64
+            distro: ubuntu22.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: uraimo/run-on-arch-action@master
+        name: "Run tests"
+        id: build
+        with:
+          arch: ${{ matrix.arch }}
+          distro: ${{ matrix.distro }}
+          githubToken: ${{ github.token }}
+          env: |
+            CTEST_OUTPUT_ON_FAILURE: 1
+            WLOG_LEVEL: 'trace'
+          install: |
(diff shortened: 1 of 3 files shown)


akallabeth:
@matoro do you know by chance why the asan stuff is failing? the library should be installed and I´ve already switched to clang as compiler, but still :/

matoro:
> @matoro do you know by chance why the asan stuff is failing? the library should be installed and I´ve already switched to clang as compiler, but still :/

No, I had to just chalk it up to QEMU issue.  Do you want to leave `TestConnect`?  This could be simply a race condition, there is a big perf hit due to QEMU so it could be that there just isn't enough delay for the server thread to spin up.

akallabeth:
> > @matoro do you know by chance why the asan stuff is failing? the library should be installed and I´ve already switched to clang as compiler, but still :/
> 
> No, I had to just chalk it up to QEMU issue. Do you want to leave `TestConnect`? This could be simply a race condition, there is a big perf hit due to QEMU so it could be that there just isn't enough delay for the server thread to spin up.
> 
would be nice to have that test working (test coverage around 60% of code, as a client and a server must be running and doing some protocol exchanges)
no other test does cover that much of the code base.
can try with extended timeouts though.

> I can also do a manual ASAN run on real hardware to make sure there aren't any hits specific to BE codepath.
no need, it would only be great to have that integrated to the tests.

akallabeth:
@matoro I can not find the issue the remaining architectures have, do you have any idea?

matoro:
> @matoro I can not find the issue the remaining architectures have, do you have any idea?

The ppc64le ones seem to be bogus qemu issues.  I just tested on real hardware and everything passes.  They might have to be excluded.

The big-endian one on s390 is legit though.  I reproduced on ppc64 BE and it says:

```
154/169 Testing: TestFreeRDPCodecProgressive
154/169 Test: TestFreeRDPCodecProgressive
Command: "/var/tmp/portage/net-misc/freerdp-9999/work/freerdp-9999_build/Testing/TestFreeRDPCodec" "TestFreeRDPCodecProgressive"
Directory: /var/tmp/portage/net-misc/freerdp-9999/work/freerdp-9999_build/libfreerdp/codec/test
"TestFreeRDPCodecProgressive" start time: Jan 24 16:00 UTC
Output:
xxxxxxx [19:16] [PIXEL_FORMAT_BGRX32] 1F1E1C8A != 881D1FFF
<end of output>
Test time =   0.07 sec
Test Failed.
"TestFreeRDPCodecProgressive" end time: Jan 24 16:00 UTC
"TestFreeRDPCodecProgressive" time elapsed: 00:00:00
```

Is that enough of a clue?

akallabeth:
@matoro ok, the progressive one was something I can work with, but the others were driving me nuts (aka I could not reproduce locally in my qemu)

akallabeth:
@matoro ok, found the issues :)
there is a problem with `unwind` or `execinfo` not working in the containers, so I´ve deactivated backtracing functions.

now only thing left is the trigger for this, I think running for every pull request is overkill, but once a week on current master would give a nice status report.

matoro:
> @matoro ok, found the issues :) there is a problem with `unwind` or `execinfo` not working in the containers, so I´ve deactivated backtracing functions.
> 
> now only thing left is the trigger for this, I think running for every pull request is overkill, but once a week on current master would give a nice status report.

Wow, thanks so much for this huge effort!  Please let me know if there's anything I can help with, or feel free to reach out for checking on unusual arches!

akallabeth:
split this: only the workflow files are part of this, code changes are in #9848

PR #9633, opened 2023-12-11
Fix Clang -Wincompatible-pointer-types warnings

JohannesKauffmann:
Fixes two cases of Clang warnings for conversions between incompatible pointer types.

In case of the smartcard code, change the (local) function to take a double pointer, instead of a single pointer combined with sometimes casting the argument to a double pointer.

For the crypto code, add a cast much like how OpenSSL does this internally.

For a more detailed description, please see the individual commits :)

### DIFF
--- a/winpr/libwinpr/crypto/hash.c
+++ b/winpr/libwinpr/crypto/hash.c
@@ -241,7 +241,7 @@ BOOL winpr_HMAC_Init(WINPR_HMAC_CTX* ctx, WINPR_MD_TYPE md, const void* key, siz
 		return FALSE;
 
 	const char* param_name = OSSL_MAC_PARAM_DIGEST;
-	const OSSL_PARAM param[] = { OSSL_PARAM_construct_utf8_string(param_name, hash, 0),
+	const OSSL_PARAM param[] = { OSSL_PARAM_construct_utf8_string(param_name, (char*)hash, 0),
 		                         OSSL_PARAM_construct_end() };
 
 	if (EVP_MAC_init(ctx->xhmac, key, keylen, param) == 1)
--- a/winpr/libwinpr/smartcard/smartcard_pcsc.c
+++ b/winpr/libwinpr/smartcard/smartcard_pcsc.c
@@ -872,7 +872,7 @@ static LONG WINAPI PCSC_SCardListReaderGroupsW(SCARDCONTEXT hContext, LPWSTR msz
 }
 
 static LONG WINAPI PCSC_SCardListReaders_Internal(SCARDCONTEXT hContext, LPCSTR mszGroups,
-                                                  LPSTR mszReaders, LPDWORD pcchReaders)
+                                                  LPSTR* mszReaders, LPDWORD pcchReaders)
 {
 	PCSC_LONG status = SCARD_S_SUCCESS;
 	BOOL pcchReadersAlloc = FALSE;
@@ -912,12 +912,12 @@ static LONG WINAPI PCSC_SCardListReaders_Internal(SCARDCONTEXT hContext, LPCSTR 
 			else
 				PCSC_AddMemoryBlock(hContext, tmp);
 
-			*(char**)mszReaders = tmp;
+			*mszReaders = tmp;
 		}
 	}
 	else
 	{
-		status = g_PCSC.pfnSCardListReaders(hContext, mszGroups, mszReaders, &pcsc_cchReaders);
+		status = g_PCSC.pfnSCardListReaders(hContext, mszGroups, *mszReaders, &pcsc_cchReaders);
 	}
 
 	*pcchReaders = (DWORD)pcsc_cchReaders;
@@ -946,7 +946,7 @@ static LONG WINAPI PCSC_SCardListReadersA(SCARDCONTEXT hContext, LPCSTR mszGroup
 	if (!PCSC_LockCardContext(hContext))
 		return SCARD_E_INVALID_HANDLE;
 
-	status = PCSC_SCardListReaders_Internal(hContext, mszGroups, mszReaders, pcchReaders);
+	status = PCSC_SCardListReaders_Internal(hContext, mszGroups, &mszReaders, pcchReaders);
 
 	if (!PCSC_UnlockCardContext(hContext))
 		return SCARD_E_INVALID_HANDLE;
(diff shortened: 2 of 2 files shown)


akallabeth:
@freerdp-bot test please

winpr/libwinpr/smartcard/smartcard_pcsc.c
```diff
@@ -872,7 +872,7 @@ static LONG WINAPI PCSC_SCardListReaderGroupsW(SCARDCONTEXT hContext, LPWSTR msz
 }
 
 static LONG WINAPI PCSC_SCardListReaders_Internal(SCARDCONTEXT hContext, LPCSTR mszGroups,
-                                                  LPSTR mszReaders, LPDWORD pcchReaders)
+                                                  LPSTR* mszReaders, LPDWORD pcchReaders)
```
akallabeth, line 875: at least the warnings here are no longer, nice fix!

winpr/libwinpr/crypto/hash.c
```diff
@@ -241,7 +241,7 @@ BOOL winpr_HMAC_Init(WINPR_HMAC_CTX* ctx, WINPR_MD_TYPE md, const void* key, siz
 		return FALSE;
 
 	const char* param_name = OSSL_MAC_PARAM_DIGEST;
-	const OSSL_PARAM param[] = { OSSL_PARAM_construct_utf8_string(param_name, hash, 0),
+	const OSSL_PARAM param[] = { OSSL_PARAM_construct_utf8_string(param_name, (char*)hash, 0),
```
akallabeth, line 244: will now complain with a `warning: cast from 'const char *' to 'char *' drops const qualifier [-Wcast-qual]`, you will not get this silent unless `OpenSSL` provides a `const` correct version of this function :/

akallabeth (approved):
LGTM, thank you!

winpr/libwinpr/smartcard/smartcard_pcsc.c
```diff
@@ -946,7 +946,7 @@ static LONG WINAPI PCSC_SCardListReadersA(SCARDCONTEXT hContext, LPCSTR mszGroup
 	if (!PCSC_LockCardContext(hContext))
 		return SCARD_E_INVALID_HANDLE;
 
-	status = PCSC_SCardListReaders_Internal(hContext, mszGroups, mszReaders, pcchReaders);
+	status = PCSC_SCardListReaders_Internal(hContext, mszGroups, &mszReaders, pcchReaders);
```
akallabeth, line 949: `mszReaders` might be `NULL` here

winpr/libwinpr/smartcard/smartcard_pcsc.c
```diff
@@ -912,12 +912,12 @@ static LONG WINAPI PCSC_SCardListReaders_Internal(SCARDCONTEXT hContext, LPCSTR
 			else
 				PCSC_AddMemoryBlock(hContext, tmp);
 
-			*(char**)mszReaders = tmp;
+			*mszReaders = tmp;
```
akallabeth, line 915: missing `NULL` checks

winpr/libwinpr/smartcard/smartcard_pcsc.c
```diff
@@ -912,12 +912,12 @@ static LONG WINAPI PCSC_SCardListReaders_Internal(SCARDCONTEXT hContext, LPCSTR
...
 		}
 	}
 	else
 	{
-		status = g_PCSC.pfnSCardListReaders(hContext, mszGroups, mszReaders, &pcsc_cchReaders);
+		status = g_PCSC.pfnSCardListReaders(hContext, mszGroups, *mszReaders, &pcsc_cchReaders);
```
akallabeth, line 920: missing `NULL`  checks

akallabeth (changes requested):
sorry, missed the missing `NULL` checks before (that is the reason mac os ci builder failed)

winpr/libwinpr/crypto/hash.c
JohannesKauffmann: It seems not only this function suffers from it: https://github.com/openssl/openssl/issues/22535

I've noticed that the Clang `-Wincompatible-pointer-types-discards-qualifiers` warning is emitted without any `-Wall` or `-Wextra`, while it's now gone with the cast in place. The `-Wcast-qual` also doesn't seem to be enabled on GCC with `-Wall -Wextra`. Not saying it isn't a valid warning, however I'd consider the situation with the cast better than without the cast.

Another option would be to use `strdup()` here, but that would make things really, really complex for solving a warning: not only do you have to remember to call free() on the early return before `OSSL_PARAM_construct_utf8_string`, but you also have to know what happens with the `strdup()`'ed string after you pass it to OpenSSL. Can it be safely freed after calling `EVP_MAC_init()`? I'm not sure. If you have some guidance there, maybe we could make that work, seeing as others (from the linked OpenSSL issue) have also solved this with `strdup()`.

Open to suggestions :)

JohannesKauffmann:
> that is the reason mac os ci builder failed

Do you have any concrete log output? For some reason I'm unable to view the warnings/errors or see why the macos CI failed, while I can for the builds that succeeded.

akallabeth:
@JohannesKauffmann only https://ci.freerdp.com/job/osx/10914/label=macosx/console and the only relevant output is `SEGFAULT` and which test crashed ;)

as for the cast, I´ve read openssl/openssl#22535 and it looks like everything except the `strdup` variant has this warning. (and I´m not eager to work around const correctness bugs in 3rd party libs)

JohannesKauffmann:
> and I´m not eager to work around const correctness bugs in 3rd party libs

I'll drop the change.

> https://ci.freerdp.com/job/osx/10914/label=macosx/console

Thanks, I couldn't find how to view the complete logs. I only found https://ci.freerdp.com/job/osx/10914/console through the GUI.

After some struggling I managed to build and run the tests (should probably have done this before committing!), but it turns out that just building the `winpr/` subdirectory isn't enough to build and run the tests.

In any case, I can't reproduce the error from the macos CI. The tests run fine on Ubuntu 23.10 WSL, on this branch and on master. I've build the project using:
```
CC=clang cmake -DBUILD_TESTING=ON -DWITH_X11=OFF -DWITH_AAD=OFF -DCMAKE_BUILD_TYPE=Debug -DWITH_SANITIZE_ADDRESS=ON ..
```

Before I go and theorize about what could possibly have gone wrong, I'd like to reproduce the test failure myself. Could you point me to the right direction? I noticed that the smartcard code was trying to load `libpcsclite.so.1`, so I've already  installed `libpcsclite-dev`, but the TestSmartCard prints:

```
SCardEstablishContext failure: SCARD_E_NO_SERVICE (0x8010001D)
```

akallabeth:
@JohannesKauffmann I have identified the issue: `LPSTR mszReaders` from `PCSC_SCardListReaders[A|W]` already contains the address of the string and you get the address of that ;)

PR #9636, opened 2023-12-13
Warnings fixed

akallabeth:


client/common/client.c
```diff
@@ -1142,7 +1142,7 @@ BOOL client_common_get_access_token(freerdp* instance, const char* request, char
 		goto cleanup;
 	}
 
-	*token = freerdp_utils_aad_get_access_token(log, response, response_length);
+	*token = (const char*)freerdp_utils_aad_get_access_token(log, response, response_length);
```
pnowack: Is this cast really necessary? `freerdp_utils_aad_get_access_token()` returns `char*` and `*token` also has type `char*`, so this cast looks superfluous.

client/common/client.c
```diff
@@ -1142,7 +1142,7 @@ BOOL client_common_get_access_token(freerdp* instance, const char* request, char
 		goto cleanup;
 	}
 
-	*token = (const char*)freerdp_utils_aad_get_access_token(log, response, response_length);
+	*token = freerdp_utils_aad_get_access_token(log, (const char*)response, response_length);
```
pnowack, line 1145: Same here with the token: `freerdp_utils_aad_get_access_token()` returns `char*` and `*token` also has type `char*`, so this cast looks superfluous.

pnowack (approved):
In the commit `utils/encoded_types: Get rid of implicit conversion warnings`, the `Get rid of these warnings by adding dummy shifting operations.` part can now be replaced with the following (since the content of the commit changed here regarding to what it does):
```
Get rid of these warnings by explicitly casting the respective values
before shifting them.
```

client/common/client.c
akallabeth: I think that comment is for something before a browser refresh?

client/common/client.c
akallabeth: looks like a comment for an old version?

client/common/client.c
pnowack: > I think that comment is for something before a browser refresh?

yes, indeed.

pnowack (approved):

PR #9646, opened 2023-12-14
[cmake] use uppercase in feature found to avoid warning

jackyzy823:
Warning:
```
CMake Warning (dev) at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (WAYLAND)
  does not match the name of the calling package (Wayland).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  cmake/FindWayland.cmake:69 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  cmake/FindFeature.cmake:26 (find_package)
  CMakeLists.txt:610 (find_feature)
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found WAYLAND: /usr/bin/wayland-scanner
-- Not detected recommended feature Wayland for Wayland (Wayland client), feature disabled
```


akallabeth (changes requested):
this change looks dangerous.
the pattern should be that the variables follow the pattern `<name>_FOUND` (e.g. `FooBar_FOUND`) but some findpackage files do not implement that correctly.

jackyzy823:
I think there should be only one kind https://github.com/FreeRDP/FreeRDP/blob/53cb5f8852258c8a4bb18715d34a51a31080e620/cmake/FindFeature.cmake#L49 

Either all `<name>_FOUND` or `<name_upper>_FOUND` ? Maybe i misunderstood?

jackyzy823:
@akallabeth  so maybe these lines should also changed to `${_feature}_FOUND}` ?

https://github.com/FreeRDP/FreeRDP/blob/53cb5f8852258c8a4bb18715d34a51a31080e620/cmake/FindFeature.cmake#L49  
https://github.com/FreeRDP/FreeRDP/blob/53cb5f8852258c8a4bb18715d34a51a31080e620/cmake/FindFeature.cmake#L51

PR #9655, opened 2023-12-14
[core] allow to specify the hostname used for AAD

hardening:
The previous code was assuming that the host name used for doing AAD was ServerHostname parameter. But when you connect directly to Azure hosts you most likely connect by IP and use short name for the AAD host, so you need to be able to give ServerHostname=<IP of host> and AadServerHostname=<shortname>.


akallabeth (approved):
LGTM,
how do you plan to expose this new setting?

hardening:
@akallabeth I've not though about it yet, it's more for software use right now (when you use freerdp as a lib)

PR #9658, opened 2023-12-15
exit after closing the auth window

st0ne77:
exit after closing the auth window


akallabeth (changes requested):
better push the error as result for the wait loop.

thanks for the fix though.
I´ll modify it and push to your branch, ok?

akallabeth:
@freerdp-bot test

akallabeth (approved):
Ok, changed to properly return an error to the calling thread.
Also unified all dialogs so all are handled the same way.
thank you for pointing me in the right direction, really appreciated ;)

PR #9662, opened 2023-12-15
migrate to mbedtls 3.x with mbedtls 2.x backward compatibility

awakecoding:
This pull request adds support for mbedtls 3.x while retaining mbedtls 2.x compatibility. mbedtls support is still limited to WinPR library builds, as we don't yet have full support for making FreeRDP hybrid openssl/mbedtls builds in one go. For our internal builds, we build WinPR+mbedtls first, then FreeRDP+openssl against WinPR+mbedtls. Fixing the build system for this would be better off done in a separate pull request, and start with mbedtls 3/2 compatibility.

Here is how I've built mbedtls 2 and 3 locally:

```
git clone git@github.com/Mbed-TLS/mbedtls.git
cd mbedtls

git checkout v3.5.1
mkdir build-mbedtls3 && cd build-mbedtls3
cmake -DENABLE_TESTING=OFF -DENABLE_PROGRAMS=OFF -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_INSTALL_PREFIX="/opt/deps/mbedtls3" ..
make
sudo make install
cd ..

git checkout v2.28.6
mkdir build-mbedtls2 && cd build-mbedtls2
cmake -DENABLE_TESTING=OFF -DENABLE_PROGRAMS=OFF -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_INSTALL_PREFIX="/opt/deps/mbedtls2" ..
make
sudo make install
cd ..
```

Here is how I have built WinPR against both versions of mbedtls locally (yes, the CMakeLists.txt in FreeRDP/WinPR can be used as the "root" one for WinPR-only builds):

```
cd FreeRDP/winpr
mkdir build-mbedtls2 && cd build-mbedtls2
cmake -DCMAKE_PREFIX_PATH="/opt/deps/mbedtls2" -DCMAKE_INSTALL_PREFIX="/opt/deps/winpr-mbedtls2" -DWITH_MBEDTLS=ON -DWITH_OPENSSL=OFF ..
make
sudo make install

cd FreeRDP/winpr
mkdir build-mbedtls3 && cd build-mbedtls3
cmake -DCMAKE_PREFIX_PATH="/opt/deps/mbedtls3" -DCMAKE_INSTALL_PREFIX="/opt/deps/winpr-mbedtls3" -DWITH_MBEDTLS=ON -DWITH_OPENSSL=OFF ..
make
sudo make install
```

And now, to build FreeRDP against a prebuilt WinPR (this requires a bit of fiddling in cmake/ConfigureFreeRDP.cmake that I'll look into a future pull request for the build system improvements):

```
cd FreeRDP
mkdir build-mbedtls3 && cd build-mbedtls3
cmake -DCMAKE_PREFIX_PATH="/opt/deps/mbedtls3;/opt/deps/winpr-mbedtls3" -DCMAKE_INSTALL_PREFIX="/opt/deps/winpr-mbedtls3" -DWITH_MBEDTLS=ON -DWITH_OPENSSL=OFF -DFREERDP_UNIFIED_BUILD=OFF -DWITH_SERVER=OFF ..
```

That's it! Most of the changes were to remove deprecated algorithms, and I also removed a bunch of definition mappings we were likely never going to use (I originally had replicated the entire set from mbedtls). Some struct members became private, requiring the usage of functions to access them. I added backward compatibility wrappers such that it can still build with mbedtls 2.

Since RC4 and MD4 were removed from mbedtls, I set the default to use internal RC4, MD4 for mbedtls, while leaving them on (previous default) for openssl builds. The old havege random number generated was also removed from mbedtls, so I added code for the newer APIs which aren't as straightforward to use, but should do the trick.


winpr/include/winpr/custom-crypto.h
```diff
@@ -185,38 +185,16 @@ extern "C"
 #define WINPR_CIPHER_AES_128_GCM 14
 #define WINPR_CIPHER_AES_192_GCM 15
 #define WINPR_CIPHER_AES_256_GCM 16
-#define WINPR_CIPHER_CAMELLIA_128_ECB 17
```
akallabeth, line 188: why remove these?

winpr/libwinpr/crypto/cipher.c
```diff
@@ -269,122 +252,6 @@ static const EVP_CIPHER* winpr_openssl_get_evp_cipher(int cipher)
 		case WINPR_CIPHER_AES_256_CCM:
 			evp = EVP_get_cipherbyname("aes-256-ccm");
 			break;
-
```
akallabeth, line 272: why remove these?

winpr/libwinpr/crypto/cipher.c
```diff
@@ -580,18 +343,6 @@ mbedtls_cipher_type_t winpr_mbedtls_get_cipher_type(int cipher)
 		case WINPR_CIPHER_AES_256_CCM:
 			type = MBEDTLS_CIPHER_AES_256_CCM;
 			break;
-
```
akallabeth, line 583: why remove these?

winpr/CMakeLists.txt
```diff
@@ -270,12 +280,10 @@ set(OPENSSL_FEATURE_TYPE "RECOMMENDED")
...
 	find_feature(MbedTLS ${MBEDTLS_FEATURE_TYPE} ${MBEDTLS_FEATURE_PURPOSE} ${MBEDTLS_FEATURE_DESCRIPTION})
 
-	if(OPENSSL_FOUND)
-		add_definitions("-DWITH_OPENSSL")
-	endif()
-
-	if(MBEDTLS_FOUND)
+	if(WITH_MBEDTLS AND MBEDTLS_FOUND)
 		add_definitions("-DWITH_MBEDTLS")
+	else()
```
akallabeth, line 285: better keep the `OPENSSL_FOUND` here, don´t assume it is there.

CMakeLists.txt
```diff
@@ -684,8 +684,10 @@ if(OPENSSL_FOUND)
 	add_definitions("-DWITH_OPENSSL")
 	message(STATUS "Using OpenSSL Version: ${OPENSSL_VERSION}")
 	include_directories(${OPENSSL_INCLUDE_DIR})
-elseif(MBEDTLS_FOUND)
-	add_definitions("-DWITH_MBEDTLS")
+endif()
+
+if(MBEDTLS_FOUND)
+	add_definitions("-DWITH_OPENSSL")
```
akallabeth, line 690: sure you want to define OpenSSL here?

winpr/include/winpr/custom-crypto.h
awakecoding: mbedtls deprecated all DES and blowfish ciphers, as for Camellia, they're not deprecated yet but they were never used, so I decided to just reduce the set of cipher IDs we map to a more reasonable subset

winpr/libwinpr/crypto/cipher.c
awakecoding: Same reason for removing the deprecated cipher IDs we map with mbedtls in the headers, this is just another mapping function

winpr/libwinpr/crypto/cipher.c
awakecoding: Same as above, deprecated and unused cipher mapping

CMakeLists.txt
awakecoding: Good catch, I'll fix this

awakecoding:
@akallabeth changes have been made according to comments, as for the removal of mbedtls id mapping of DES, Blowfish ciphers, they've been removed in mbedtls 3.x, and it's not worth using ifdefs for compatibility. As for camelia ciphers, we never needed them, and they're bound to be removed eventually, so might as well rip the band aid now. I do use the AES ciphers in other projects that use WinPR without FreeRDP though, so we're keeping those.

winpr/include/winpr/custom-crypto.h
akallabeth: we´ve released 3.0 with these, so I´m for playing it safe here and just change what is really required. (yes, should not be used anywhere, but you never know)

akallabeth:
@awakecoding your changes break the openssl build on the ci.

awakecoding:
I'll do another pass of changes on Monday to get this pull request clean and ready to go. I'll put the camellia ciphers back, but blowfish and DES will still be removed because they're no longer defined, is that okay? In other words I'll limit the ID removal to what's been removed in mbedtls 3.x

akallabeth:
@awakecoding I´d just keep all the defines in the public header, but no need to implement that (no longer existing) ones for `mbedtls`. (just return a failure where appropriate)
just want to keep the headers and defined symbols consistent for 3.0

awakecoding:
@akallabeth it should all be good now!

akallabeth (approved):
LGTM

akallabeth:
@awakecoding great, and merged!

PR #9665, opened 2023-12-16
[channels,gfx] reply to cache import offer

akallabeth:
we do not use the cache, so reply with 0 imported messages by default.

### DIFF
--- a/libfreerdp/core/peer.c
+++ b/libfreerdp/core/peer.c
@@ -93,14 +93,11 @@ static HANDLE freerdp_peer_virtual_channel_open(freerdp_peer* client, const char
 		return (HANDLE)peerChannel;
 	}
 
-	peerChannel = (rdpPeerChannel*)calloc(1, sizeof(rdpPeerChannel));
+	peerChannel = server_channel_common_new(client, index, mcsChannel->ChannelId, 128, NULL, name);
 
 	if (peerChannel)
 	{
-		peerChannel->index = index;
-		peerChannel->client = client;
 		peerChannel->channelFlags = flags;
-		peerChannel->channelId = mcsChannel->ChannelId;
 		peerChannel->mcsChannel = mcsChannel;
 		mcsChannel->handle = (void*)peerChannel;
 	}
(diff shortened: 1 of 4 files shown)


libfreerdp/core/peer.c
```diff
@@ -103,6 +103,7 @@ static HANDLE freerdp_peer_virtual_channel_open(freerdp_peer* client, const char
 		peerChannel->channelId = mcsChannel->ChannelId;
 		peerChannel->mcsChannel = mcsChannel;
 		mcsChannel->handle = (void*)peerChannel;
+		InitializeCriticalSection(&peerChannel->writeLock);
```
pnowack: In this commit (`[core,server] lock FreeRDP_WTSVirtualChannelWrite`), this code here is still there from the old version (even though it is removed in a later commit). This should be removed from this commit here though, as it would break bisecting otherwise (fails to compile).

libfreerdp/core/peer.c
```diff
@@ -122,6 +123,7 @@ static BOOL freerdp_peer_virtual_channel_close(freerdp_peer* client, HANDLE hCha
 	mcsChannel = peerChannel->mcsChannel;
 	WINPR_ASSERT(mcsChannel);
 	mcsChannel->handle = NULL;
+	DeleteCriticalSection(&peerChannel->writeLock);
```
pnowack: Same here: In this commit (`[core,server] lock FreeRDP_WTSVirtualChannelWrite`), this code here is still there from the old version. This should be removed, as it would break bisecting otherwise (fails to compile).

pnowack (commented):
The commit message body of the commit `[channels,gfx] reply to cache import offer` does not represent the actual code change, because it presents the view of one server implementation (here: the shadow server), but not every single FreeRDP-based server implementation. Other server implementations might actually intent to use the cache import offer. So, that message should be changed accordingly.

The following commit message body fits for that commit better (I chose a different reasoning, but that code change still reflect that):
```
Most FreeRDP-based servers unlikely make use of the cache import offer,
as Microsoft did not document, how the cache keys are generated.
So, add a default handler for that PDU to always notify the client,
that no entries have been imported.
```

The following title could be used for that commit too: `Add a default cache import offer handler`

Code-wise, things look alright to me.

libfreerdp/core/peer.c
akallabeth: ok, merged the last 2 commits

hardening (approved):
Seems ok

PR #9668, opened 2023-12-18
[primitives] skip benchmarking if only one available

akallabeth:


hardening (approved):
Oh really a good idea

PR #9669, opened 2023-12-18
[proxy,modules] allow compiling with old c++ compiler

joantolo:
For dyn-channel-dump.

On C++17 there's the filesystem standard library.
Newer c++ compilers use it with <filesystem>.
However, older c++ compilers use it with <experimental/filesystem>.


akallabeth:
@freerdp-bot test please

PR #9670, opened 2023-12-18
Libressl compat

akallabeth:


awakecoding:
I just built it locally, and it works! Woohoo! LibreSSL for the win ;)

PR #9674, opened 2023-12-19
Build fixes

hardening:


client/X11/xf_input.c
```diff
@@ -875,17 +877,11 @@ static int xf_input_handle_event_remote(xfContext* xfc, const XEvent* event)
 				/* fallthrough */
 				WINPR_FALLTHROUGH
 			case XI_TouchUpdate:
-				/* fallthrough */
-				WINPR_FALLTHROUGH
```
akallabeth, line 879: what not working here?

akallabeth (changes requested):
if a macro `WINPR_FALLTHROUGH` does not work for you then fix the error there and do not remove it in code.
should already define to empty if the compiler does not support the `c11` features

client/X11/xf_input.c
hardening: Well it seems like for gcc a fallthrough is when you have
```C
case item:
    ... code...
   <no break>
case item 2:
```
The fallthrough is only at the point where it's missing a break, other cases are just regular `switch/case` matches (ie when there's no code between `case`)

mfleisz: Looks like this is actually the standard way of using fallthrough: https://en.cppreference.com/w/c/language/attributes/fallthrough

PR #9675, opened 2023-12-19
[client,sdl] correctly handle AuthenticateEx

akallabeth:
handle callback with already provided credentials that did not fail (aka `RDP` or `TLS` security)


akallabeth:
* copied handling from `client_cli_authenticate_ex`

hardening (approved):

PR #9676, opened 2023-12-19
[ci,rpm] disable build-id, update dependencies

akallabeth:


akallabeth:
merging as current nightly is broken and this can not make it worse.

PR #9677, opened 2023-12-19
Rel mouse fixes

akallabeth:


akallabeth:
@hardening no, there the relative coordinates are always available and we just check `freerdp_client_use_relative_mouse_events` and if `TRUE` send the relative coordinates, else the absolute ones.

akallabeth:
@hardening this bug was in `client/common` though, so it ignored what the client was telling the library

hardening (approved):
LGTM

PR #9681, opened 2023-12-19
Bigendian fixes

akallabeth:
provides all the fixes from #9632 without the ci (that needs some clarifications and additional fixes first before ready)


hardening:
Seems ok for me. Except that I don't understand the endianess problem with these static strings in SSPI (like package names) and other places.

akallabeth:
> Seems ok for me. Except that I don't understand the endianess problem with these static strings in SSPI (like package names) and other places.

@hardening `static const WCHAR foo[] = { 'a', 'b', 'c', '\0' };` evaluates to `0x6100, 0x6200, 0x6300, 0x0000` on little endian and `0x0061, 0x0062, 0x0063, 0x0000` on big endian as `WCHAR == uint16_t`

Issue #9678, opened 2023-12-19
evdev keyboard mapping broken


akallabeth:
`freerdp_keyboard_init` tries to detect the keyboard layout and create a reverse map to get a scancode for a keypress.

when `WITH_XKBFILE` is not detected or fails, the detection falls back to `freerdp_keyboard_init_x11_evdev`.

this mapping is broken.

easily tested with a build that does not have `libxkbfile-dev` installed at compile time.

cumt2010:
Yes, we meet the same problem.
freerdp_keyboard_init_x11_evdev -> GetVirtualKeyCodeFromKeycode, parameter should use WINPR_KEYCODE_TYPE_XKB instead of WINPR_KEYCODE_TYPE_EVDEV , can get correct result.

MarekKasik:
I've got report for this issue in gnome-connections' issue tracker here: https://gitlab.gnome.org/GNOME/gnome-connections/-/issues/175 .
The issue shows in flatpak which does not have X display but Wayland only. freerdp_keyboard_init_xkbfile() fails in such case and freerdp_keyboard_init_x11_evdev() is called. It behaves as in https://github.com/FreeRDP/FreeRDP/issues/10572 then. The change mentioned in https://github.com/FreeRDP/FreeRDP/issues/9678#issuecomment-1864371063 fixes the issue though.

akallabeth:
@MarekKasik `freerdp_keyboard_init*` family of functions are deprecated and should no longer be used.

MarekKasik:
Thank you for your reply. I'll have a look at reworking of that part then.

PR #9686, opened 2023-12-20
fix:[locale,keyboard] fix keyboard input confused for linux x86_64

cumt2010:
linux x86_64 connect Windows server 2016(share desktop or vapp), keyboard typing is confusing.
freerdp_keyboard_init_x11_evdev -> GetVirtualKeyCodeFromKeycode, parameter should use WINPR_KEYCODE_TYPE_XKB instead of WINPR_KEYCODE_TYPE_EVDEV , can get correct result.

### DIFF
--- a/libfreerdp/locale/keyboard.c
+++ b/libfreerdp/locale/keyboard.c
@@ -279,11 +279,11 @@ static int freerdp_keyboard_init_x11_evdev(DWORD* keyboardLayoutId,
 	WINPR_ASSERT(x11_keycode_to_rdp_scancode);
 	for (size_t keycode = 0; keycode < count; keycode++)
 	{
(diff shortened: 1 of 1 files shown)


cumt2010:
https://github.com/FreeRDP/FreeRDP/issues/9678
no <sup>libxkbfile-dev</sup> installed.

akallabeth (changes requested):
you have a build issue, this code should not be executed. (you need `xkbfile` development headers)
your fix is therefore incorrect.

cumt2010:
> you have a build issue, this code should not be executed. (you need `xkbfile` development headers) your fix is therefore incorrect.
> 
> this should only be executed when the events are from `EVDEV` and not if you lack some development headers and get events from other sources.

got it, thanks.

PR #9690, opened 2023-12-20
Clipboard docs

akallabeth:
* adds better formatting of help text from argument struct (support indented linebreaks)
* adds better formatting in manpage from argument struct (support linebreaks)
* improve helper binary for generating docbook xml from argument struct
* improve help text for `clipboard`


akallabeth:
build and tested the generated manpages.

* the binary name (in the manpage) now matches the installed one
* the manpage name now matches the binary name
* all installed binaries have now a manpage (exceptions: sample client & server)

PR #9697, opened 2023-12-21
Release 3.1.0

akallabeth:


client/Android/Studio/freeRDPCore/src/main/cpp/android_freerdp.c
```diff
@@ -56,8 +56,7 @@
 #define TAG CLIENT_TAG("android")
 
 /* Defines the JNI version supported by this library. */
-#define FREERDP_JNI_VERSION "3.0.1-dev0"
-
+#define FREERDP_JNI_VERSION "3.1.1-dev0"
```
pnowack, line 59:
```suggestion
#define FREERDP_JNI_VERSION "3.1.0-dev0"
```

ChangeLog
```diff
@@ -1,3 +1,46 @@
...
+* mbedTLS support enhancements (#9662)
+* Improve building on mac OS (#9641)
+* New and improved manpages (#9690, #9650)
+* Unify CMake common options, add (optional) binary versioning and allow
+  building rdtk and uwac as static dependencies  (#9695)
+* SDL client improvements (#9693, #9657, #9659, #9683, #9680, #9657, #9664,
+  #9656)
+
+For a complete and detailed change log since the last release run:
+git log 3.0.1...3.0.0
```
pnowack, line 42:
```suggestion
git log 3.1.0...3.0.0
```

ChangeLog
```diff
@@ -1,3 +1,46 @@
+# 2023-12-22 Version 3.1.0
+
+A new 3.1.0 release only 10 days after 3.0.0.
```
pnowack, line 3: I think we should drop the `only 10 days after` wording here. It might express, that we would work sloppy, which is not the case here (but that wording could implicate that here, since this is the first "minor" stable update after the first major release, in a short time, while that short time is emphasized here).

akallabeth: good point. updated.
also added the `(optional)` for the new build dependencies more prominent

PR #9698, opened 2023-12-21
[client,sdl] make dialog image support optional

akallabeth:
allow disabling image support in connection dialog to avoid having to link `SDL_image`


pnowack (approved):
Change works in testing and I couldn't spot an issue in the code here

PR #9719, opened 2024-01-02
Add GetPublicKey transport IO method to access TLS public key.

llyzs:
The original NLA module access the TLS public key directly through the internal rdpTls structure, which is not available if a custom client implements its own transport IO layer (rdpTransportIo). In order for a custom transport layer implementation works with core NLA after TLS connection is established (TLSConnect), the NLA module needs to retrieve the public key through a transport IO function instead of using internal structure.

This patch adds such method in rdpTransportIo while also moves the original codes into a default implementation.


akallabeth (approved):
LGTM

mfleisz (approved):

PR #9720, opened 2024-01-02
[cmake] generate relative pkgconfig path

### LINKED ISSUE #9718: PKG_CONFIG_RELOCATABLE is broken on half the linux distros
In https://github.com/FreeRDP/FreeRDP/commit/89429885a0160156360b54001f707cb143d39d67 a default assumption was added that the pcfiledir will always be 2 depths below the prefix.

```
option(PKG_CONFIG_RELOCATABLE "Generate relocatable pkg-config files" ON)
if (PKG_CONFIG_RELOCATABLE)
    set(PKG_CONFIG_INSTALL_PREFIX "\${pcfiledir}/../..")
else()
    set(PKG_CONFIG_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX})
endif()
```

However in many systems like all the debian derivatives and Flatpak runtimes, this is not true. Instead the pkgconfig file ends up under `prefix/lib/arch-triplet/pkgconfig` or a real world `/usr/lib/x86_64-linux-gnu/pkgconfig` and thus the double `../../` relative path is broken.

No idea what to do about this, or if it should be kept on given it's a feature that only works sometimes and can be broken at others depending on the host system. It could be possible to resolve the prefix instead of hardcoding it, but something like that should be done in cmake itself imo if it wants to have generic relocatable pkg config files.

akallabeth:
generate the relative pkgconfig path from the actual install locations. fixes #9718


akallabeth:
@freerdp-bot test

akallabeth:
confirmed working by @pnowack

PR #9725, opened 2024-01-03
[libuwac] fix keyboard "sticky" keys when entering window

manio:
The commit is fixing the following problem:
When the freerdp window gets keyboard focus and is notified, it wrongly process keys which are *held*, not a freshly pressed.

The comment in the code explains it more.

### DIFF
--- a/uwac/libuwac/uwac-input.c
+++ b/uwac/libuwac/uwac-input.c
@@ -273,6 +273,8 @@ static void keyboard_handle_leave(void* data, struct wl_keyboard* keyboard, uint
                                   struct wl_surface* surface)
 {
 	struct itimerspec its = { 0 };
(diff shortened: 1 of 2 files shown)


manio:
Oh this is wrong. Have to work more about it...

manio:
The change seems to be working fine... but I have sometimes the situation, that when I enter a freerdp window and start typing I have some modifier key active and I have to press all modifier keys to have it return to normal state...
I am setting this to draft and will test it for some longer period...

manio:
Hi again,
I tested it for some time and think it's ready for merge.
Switching from Draft to ready...

akallabeth:
@freerdp-bot test

manio:
Thank you

PR #9727, opened 2024-01-04
[common,assist] fix file parser (Fixes #9726)

### LINKED ISSUE #9726: Remote Assistance File Parsing
There appears to be an issue with remote assistance file parsing in the [current master](https://github.com/FreeRDP/FreeRDP/commit/33447dc16e8f9e52b2b46cccdcc017bad1908008). I tried generating .msrcIncident from a variety of different Windows versions (Windows 11, Windows 10, Windows 7) and all of the ones I tested currently don't work due to the same issue.

When [`freerdp_assistance_parse_all_elements_of_l`](https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/common/assistance.c#L669) is done parsing, `file` contains 4 entries in `MachinePorts` but 8 entries in `MachineAddresses`.

This causes [`freerdp_settings_set_pointer_array`](https://github.com/FreeRDP/FreeRDP/blob/master/libfreerdp/common/assistance.c#L1326) to fail with an invalid offset later on when setting `FreeRDP_TargetNetAddresses`.

Here is an example .msrcIncident which can be used for reproduction. The password is `4X638PTVZTKZ`: https://gist.github.com/lemonmojo/6bef0edf563ad9631ceae8d3d29a2aef

akallabeth:
* only append N or U entry if it was found
* append port for N and U entries

### DIFF
--- a/libfreerdp/common/assistance.c
+++ b/libfreerdp/common/assistance.c
@@ -660,12 +660,20 @@ static BOOL freerdp_assistance_parse_all_elements_of_l(rdpAssistanceFile* file, 
 	if (!freerdp_assistance_parse_attr(&u, &ulen, "U", data))
 		return FALSE;
 
-	if (!ArrayList_Append(file->MachineAddresses, strndup(n, nlen)))
-		return FALSE;
-	if (!ArrayList_Append(file->MachineAddresses, strndup(u, ulen)))
-		return FALSE;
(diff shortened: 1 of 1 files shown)


hardening (commented):
Looks good

hardening (approved):

Issue #9722, opened 2024-01-03
 Failed to connect to "WindowsServer2008R2 Datacenter" using freerdp_proxy of freerdp3.1 !


Mr-LZG:
freerdp-proxy log:
```
root@debian:~/code/FreeRDP3.1/trunk# /usr/local/bin/freerdp-proxy ./config.ini
[16:51:47:479] [120030:0001d4de] [INFO][com.freerdp.proxy.server] - [main]: freerdp-proxy version info:
[16:51:47:479] [120030:0001d4de] [INFO][com.freerdp.proxy.server] - [main]:     FreeRDP version: 3.1.0
[16:51:47:480] [120030:0001d4de] [INFO][com.freerdp.proxy.server] - [main]:     Git commit: 3.1.0
[16:51:47:480] [120030:0001d4de] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.AudioInput' not found, value defaults to true.
[16:51:47:480] [120030:0001d4de] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.DeviceRedirection' not found, value defaults to true.
[16:51:47:480] [120030:0001d4de] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.VideoRedirection' not found, value defaults to true.
[16:51:47:480] [120030:0001d4de] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.CameraRedirection' not found, value defaults to true.
[16:51:47:480] [120030:0001d4de] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.PassthroughIsBlacklist' not found, value defaults to false.
[16:51:47:480] [120030:0001d4de] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Input.Multitouch' not found, value defaults to true.
[16:51:47:480] [120030:0001d4de] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Security.ServerNlaSecurity' not found, value defaults to false.
[16:51:47:483] [120030:0001d4de] [INFO][com.freerdp.proxy.config] - [pf_server_config_print]: Proxy configuration:
[16:51:47:483] [120030:0001d4de] [INFO][com.freerdp.proxy.config] - [pf_server_config_print]:   Server:
[...]

Issue #9714, opened 2023-12-30
[16bpp] color depth not supported by proxy


Mr-LZG:
freerdp_proxy log

```
root@debian:~/code/FreeRDP3.1/trunk# /usr/local/bin/freerdp-proxy ./config.ini
[13:31:07:293] [72715:00011c0b] [INFO][com.freerdp.proxy.server] - [main]: freerdp-proxy version info:
[13:31:07:294] [72715:00011c0b] [INFO][com.freerdp.proxy.server] - [main]:      FreeRDP version: 3.1.0
[13:31:07:294] [72715:00011c0b] [INFO][com.freerdp.proxy.server] - [main]:      Git commit: 3.1.0
[13:31:07:295] [72715:00011c0b] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.AudioInput' not found, value defaults to true.
[13:31:07:295] [72715:00011c0b] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.DeviceRedirection' not found, value defaults to true.
[13:31:07:296] [72715:00011c0b] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.VideoRedirection' not found, value defaults to true.
[13:31:07:296] [72715:00011c0b] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.CameraRedirection' not found, value defaults to true.
[13:31:07:296] [72715:00011c0b] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Channels.PassthroughIsBlacklist' not found, value defaults to false.
[13:31:07:296] [72715:00011c0b] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Input.Multitouch' not found, value defaults to true.
[13:31:07:297] [72715:00011c0b] [WARN][com.freerdp.proxy.config] - [pf_config_get_bool]: key 'Security.ServerNlaSecurity' not found, value defaults to false.
[13:31:07:297] [72715:00011c0b] [INFO][com.freerdp.proxy.config] - [pf_server_config_print]: Proxy configuration:
[13:31:07:297] [72715:00011c0b] [INFO][com.freerdp.proxy.config] - [pf_server_config_print]:    Server:
[...]

PR #9741, opened 2024-01-08
[server,proxy] add TlsSecLevel option to config

akallabeth:
To support legacy targets add the TlsSecLevel configuration option that is equivalent to the /tls:seclevel option of the client implementations. This allows automatic configuration of OpenSSL legacy providers if they are available.

### DIFF
--- a/include/freerdp/server/proxy/proxy_config.h
+++ b/include/freerdp/server/proxy/proxy_config.h
@@ -109,6 +109,9 @@ extern "C"
 		size_t PrivateKeyPEMLength;
 
 		wIniFile* ini;
+
+		/* target continued */
+		UINT32 TargetTlsSecLevel;
 	};
 
 	/**
--- a/server/proxy/pf_config.c
+++ b/server/proxy/pf_config.c
@@ -74,6 +74,7 @@ static const char* key_target_fixed = "FixedTarget";
 static const char* key_target_user = "User";
 static const char* key_target_pwd = "Password";
 static const char* key_target_domain = "Domain";
+static const char* key_target_tls_seclevel = "TlsSecLevel";
 
 static const char* section_clipboard = "Clipboard";
 static const char* key_clip_text_only = "TextOnly";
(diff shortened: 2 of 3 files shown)


include/freerdp/server/proxy/proxy_config.h
```diff
@@ -109,6 +109,9 @@ extern "C"
 		size_t PrivateKeyPEMLength;
 
 		wIniFile* ini;
+
+		/* target continued */
+		INT32 TargetTlsSecLevel;
```
mfleisz, line 114: The `TlsSecLevel` is an `UINT32` (using this type also eliminates introduction of these new _int32 macros/functions)

server/proxy/pf_config.c
```diff
@@ -632,6 +669,8 @@ BOOL pf_server_config_dump(const char* file)
 		goto fail;
 	if (IniFile_SetKeyValueString(ini, section_target, key_target_fixed, bool_str_true) < 0)
 		goto fail;
+	if (IniFile_SetKeyValueString(ini, section_target, key_target_tls_seclevel, "1") < 0)
```
mfleisz, line 672: Shouldn't we use `IniFile_SetKeyValueInt` here to indicate that this setting is an integer?

include/freerdp/server/proxy/proxy_config.h
akallabeth: we need to support 'not available' for this option too (aka we don´t touch the default)

server/proxy/pf_config.c
akallabeth: yes, we can use that too.

include/freerdp/server/proxy/proxy_config.h
mfleisz: So why don't we just set it to 1 by default (like we also do in `pf_server_config_dump`) and avoid checking for set/not set?

akallabeth: ok, that we can do.

mfleisz (approved):

Mr-LZG:
#9722  and #9723  do not seem to be TLS related problems. During debugging, it was found that the server failed to receive the license message. SEC_LICENSE_PKT is not set in the secureFlags returned by the server.

Mr-LZG:
Use freerdp3.2 to test and find that the problems (#9722 #9723 ) still exists.

akallabeth:
@Mr-LZG stop spamming the issue tracker.
I´ve already told you more than once that `RDP` security has issues and that it is tracked with https://github.com/FreeRDP/FreeRDP/issues/6713

if you are so eager to get this working, we´re very happy to merge pull requests fixing stuff.

Mr-LZG:
@akallabeth 
Okay, okay, sorry!
I have tested that it is true that the target device uses the RDP security layer, and the problem （ #9714 ） occurs only because the RDP security layer is enabled, and it is not a problem with openssl.
We'd love to support more types of target devices since I can't modify them.

PR #9751, opened 2024-01-09
Proxy fix

akallabeth:
fixes an issue with `xfreerdp 2.0` (and maybe others) if a deactivate/reactivate sequence occurs against a `FreeRDP3` based server


mfleisz (approved):

Mr-LZG:
Use freerdp3.2 to test and find that the problems (#9722 #9723 ) still exists.

PR #9756, opened 2024-01-10
[shadow] command line help should not be part of the log info

efferre79:
xfreerdp client uses the console for usage information, that solution is much more readable and user friendly


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #9764, opened 2024-01-11
[channels,drdynvc] Add settings to enable synchronous dynamic channels.

llyzs:
The drdynvc channel already supports synchronous mode (without threads) but is not exposed through a public setting.

This patch adds such setting so thin clients can run synchronous dynamic channels as needed.

Other legacy static channels should also support synchronous mode and should come in future PR.


akallabeth (approved):
LGTM

Issue #9766, opened 2024-01-12
Segmentation fault during auth


jdarmetzki:
**Describe the bug**
When i try to connect to my work-pc with xfreerdp >= 3.0, the client dies with a segmentation fault. 
Connecting to a VM for example works fine.

My work-pc was migrated from On-Premise AD via VPN to Azure

**To Reproduce**
Unclear - try to connect to a machine that is configured like mine :( 

**Expected behavior**
A successful connection

**Screenshots**
n/a

**Application details**
* FreeRDP version (`xfreerdp /version`)
3.1.0, latest from git as of 2024-01-11 19:00
* Command line used
xfreerdp Arbeit-duo.rdp /u:$username /monitors:0,1 -grab-keyboard   /p:$PW 
* Output of `xfreerdp /buildconfig`

<details><summary>buildconfig</summary>

```
[07:17:14:329] [20992:00005200] [WARN][com.freerdp.core.rdp] - [log_build_warn][0x5641aa5c09c0]: *************************************************
[07:17:14:329] [20992:00005200] [WARN][com.freerdp.core.rdp] - [log_build_warn][0x5641aa5c09c0]: This build is using [debug] build options:
[07:17:14:329] [20992:00005200] [WARN][com.freerdp.core.rdp] - [log_build_warn][0x5641aa5c09c0]: * 'WITH_DEBUG_ALL=ON'
[07:17:14:329] [20992:00005200] [WARN][com.freerdp.core.rdp] - [log_build_warn][0x5641aa5c09c0]: * 'WITH_DEBUG_CAPABILITIES=ON'
[07:17:14:329] [20992:00005200] [WARN][com.freerdp.core.rdp] - [log_build_warn][0x5641aa5c09c0]: * 'WITH_DEBUG_CERTIFICATE=ON'
[07:17:14:329] [20992:00005200] [WARN][com.freerdp.core.rdp] - [log_build_warn][0x5641aa5c09c0]: * 'WITH_DEBUG_CHANNELS=ON'
[07:17:14:329] [20992:00005200] [WARN][com.freerdp.core.rdp] - [log_build_warn][0x5641aa5c09c0]: * 'WITH_DEBUG_CLIPRDR=ON'
[07:17:14:329] [20992:00005200] [WARN][com.freerdp.core.rdp] - [log_build_warn][0x5641aa5c09c0]: * 'WITH_DEBUG_CODECS=ON'
[...]

PR #9771, opened 2024-01-13
[winpr,sspi] properly log all kerberos calls

akallabeth:
to ease debugging log each failure of a kerberos related function along with location and function name
#9766 did show logging was not really helpful, so add some useful information to the errors logged.


akallabeth:
@freerdp-bot test please

hardening (approved):
Looks ok

PR #9774, opened 2024-01-17
Fix freerdp_parse_username_ptr to not break up usernames with @ in them

ElliottLester:
Issue:
My company has switched to hybrid AAD/AD In order to auth correctly the username needs to contain the domain.

Fix:
looking at freerdp_parse_username_ptr it seems like the intended behavior was to not break the username for user@corp.net accounts.
In my case the server expects the format AzureAD\user@corp.net this seems to resolve my issue using /sec:nla:off.

Notes:
Not sure if this is the right fix but it seems to work for me. 
tested against 3.1.0 and master.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -1171,6 +1171,7 @@ static BOOL freerdp_parse_username_ptr(const char* username, const char** user, 
                                        const char** domain, size_t* domainlen)
 {
 	const char* p = strchr(username, '\\');
+	const char* a = strchr(username, '@');
 
 	*user = NULL;
 	*userlen = 0;
@@ -1178,7 +1179,7 @@ static BOOL freerdp_parse_username_ptr(const char* username, const char** user, 
 	*domain = NULL;
 	*domainlen = 0;
 
-	if (p)
+	if (p && !a)
 	{
 		const size_t length = (size_t)(p - username);
 		*user = &p[1];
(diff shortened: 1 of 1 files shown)


hardening (approved):
Seems ok

hardening:

client/common/cmdline.c
```diff
@@ -1190,11 +1191,14 @@ static BOOL freerdp_parse_username_ptr(const char* username, const char** user,
...
 		/* Do not break up the name for '@'; both credSSP and the
-		 * ClientInfo PDU expect 'user@corp.net' to be transmitted
-		 * as username 'user@corp.net', domain empty (not NULL!).
+		 * ClientInfo PDU expect 'AzureAD/user@corp.net' to be transmitted
+		 * as username 'AzureAD/user@corp.net',
+		 * domain empty (not NULL!).
 		 */
 		*user = username;
 		*userlen = strlen(username);
+		*domain = username;
```
akallabeth, line 1200: why do you set `domain` to `username`? that contradicts the description.

ElliottLester: if domain = NULL 
later in the auth process we will ask for it interactively 
perhaps it should be
domain = ""
domainlen = 0

ElliottLester: the original comment seems to intend the same thing 
"domain empty (not NULL!)."

but I didn't want to return a pointer to a char* on the stack.

akallabeth: @ElliottLester then why not set it to the terminating `\0` of the input string?

akallabeth (approved):
Looking good now.

just a silly question before merging, did you try `\AzureAD\user@corp.net` without these changes?
that should do the same (set domain to `""`) and username to `AzuerAD\user@corp.net`

ElliottLester:
> Looking good now.
> 
> just a silly question before merging, did you try `\AzureAD\user@corp.net` without these changes? that should do the same (set domain to `""`) and username to `AzuerAD\user@corp.net`

I did not try that. I understand now having now looked through the code that that would probably work,
but before I did that I wouldn't have know why that format works only for this client.

akallabeth:
ok, after thinking a bit about this, I´ll close this pr.
it does work for your use case, but I´m fairly sure there are (some? lots? don´t know) versions out there that do not work.

I´ll create a pull request adding a better explanation of `AzureAD` usage for the next version of FreeRDP.

ElliottLester:
yeah I guess that resolves my issue thanks!

PR #9780, opened 2024-01-18
[core,gcc] split update of RDP encryption from gcc

akallabeth:
* Update EncryptionLevel and EncryptionMethods before writing GCC data
* Use const rdpSettings* in all GCC write functions


hardening (approved):
Seems ok for me

PR #9793, opened 2024-01-22
[core,transport] make blocking mode available to transport IO interface.

llyzs:
The expected blocking mode is needed for custom transport layer and should be exposed to the transport IO interface.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test please

mfleisz (approved):

PR #9795, opened 2024-01-22
[build] fix issues detected by clang-tidy

akallabeth:


hardening:
I hate the code style where a single "if line" have mandatory brackets. Mixed with back to line rule, that changes
```C
if (cond)
    return TRUE;
```
to
```C
if (cond)
{
    return TRUE;
}
```

I don't see any gain, given that making clang-tidy happy is not a goal.

PR #9805, opened 2024-01-24
remove superfluous manpage build dependencies

### LINKED ISSUE #9804: Missing path for `${manpage}.xml` dependency
**Describe the bug**
Building FreeRDP on older versions of CMake fails with the message:
```
ninja: error: '../client/SDL/man/sdl-freerdp3.1.xml', needed by 'client/SDL/man/CMakeFiles/sdl-freerdp3.1.manpage', missing and no known rule to make it
```
This is because the build target `${manpage}.manpage` lists `${manpage}.xml` as a build dependency, here: https://github.com/FreeRDP/FreeRDP/blob/465745131e799ee40fd78030bdecebb37e2c5a0d/cmake/InstallFreeRDPMan.cmake#L40-L47 but defines no build rule for `${manpage}.xml`.  Instead, `${manpage}.xml` is generated and placed in the binary directory by a call to `configure_file` at configure time, before ninja is invoked to build the project.  Since the dependency declaration does not specify the path, some versions of CMake expect to find `${manpage}.xml` in the source directory, where it does not exist, and abort.  The latest version of CMake appears to recognize that it already generated `${manpage}.xml` and implicitly uses the path to the binary directory when constructing the the build rules.

**To Reproduce**
Create a new directory, then in that directory:
1. Place a copy of the attached [Dockerfile](https://github.com/FreeRDP/FreeRDP/files/14042081/Dockerfile.txt) (rename it to remove the .txt extension)
1. Clone FreeRDP as "FreeRDP"
1. Run the following commands:

```bash
mkdir deps pkgs

# Build Docker image
docker build --build-arg BASE_IMAGE=ubuntu:20.04 --tag freerdp-builder:ubuntu-20.04 --file Dockerfile .

# FreeRDP fails to provide FindcJSON.cmake, so find_package cannot find libcjson.
# Compile against newer package that includes cJSONConfig.cmake.
cd deps
CJSON_REPO=http://ftp.debian.org/debian/pool/main/c/cjson
curl -O $CJSON_REPO/libcjson-dev_1.7.14-1_amd64.deb -O $CJSON_REPO/libcjson1_1.7.14-1_amd64.deb
cd ..

# Build FreeRDP
docker run -it -v ./deps:/deps:ro -v ./FreeRDP:/src:ro -v ./pkgs:/out --rm freerdp-builder:ubuntu-20.04
```

**Expected behavior**
The build should complete successfully and place the new package files into the pkgs directory.

karmix:
The `${manpage}.manpage` custom target in `generate_and_install_freerdp_man_from_xml` should only declare as dependencies the build targets we wish to add to `ALL`, and should not include the sub-dependencies of those build targets.

Resolves #9804

### DIFF
--- a/cmake/InstallFreeRDPMan.cmake
+++ b/cmake/InstallFreeRDPMan.cmake
@@ -41,9 +41,6 @@ function(generate_and_install_freerdp_man_from_xml template manpage dependencies
 		${manpage}.manpage ALL
                 DEPENDS
 			${manpage}
-                        ${manpage}.xml
-                        ${template}.xml.in
-                        generate_argument_docbook
 		)
 	install_freerdp_man(${CMAKE_CURRENT_BINARY_DIR}/${manpage} 1)
     endif()


akallabeth (changes requested):
while it can be reasoned that the dependencies should move to `add_custom_command` it is no solution just to delete this and hope for the best.

karmix:
I apologize for the confusion, I was not suggesting we hope for the best.  I explained why each of these dependencies was unnecessary in the discussion of the issue, here: https://github.com/FreeRDP/FreeRDP/issues/9804#issuecomment-1908976647

akallabeth:
@karmix you did rationalize your changes, but you missed that `generate_argument_docbook` is not in the dependency list of `add_custom_command(OUTPUT ${manpage}` in your pr

karmix:
`generate_argument_docbook` is listed as a `COMMAND`.  When commands reference executables that CMake builds (as is this case), it implicitly adds them as order-only dependencies of the custom command target.  As an order-only dependency, Ninja will ensure it gets built before the custom command runs.  (If that weren't the case, there would be nothing to ensure that order in the current rules :grin: .)

I didn't wan't to modify behavior in the initial PR.  Adding `generate_argument_docbook` as a standard dependency would deviate from current behavior by causing Ninja to rerun the custom command and refresh the manpage if `generate_argument_docbook` was updated/recompiled (where the order-only dependency does not trigger a refresh).  It isn't a bad idea to add this if you are going to be doing development work on the tool.  If you'd like, I can update the PR.

akallabeth:
> 
> I didn't wan't to modify behavior in the initial PR. Adding `generate_argument_docbook` as a standard dependency would deviate from current behavior by causing Ninja to rerun the custom command and refresh the manpage if `generate_argument_docbook` was updated/recompiled (where the order-only dependency does not trigger a refresh). It isn't a bad idea to add this if you are going to be doing development work on the tool. If you'd like, I can update the PR.

that was the whole point in having it in the dependency list ;)
the tool generates the data from compiling with current sources and extract data from a struct.

akallabeth:
replaced by #9813

PR #9809, opened 2024-01-25
Win addin load fix

akallabeth:


nfedera (changes requested):
GetModuleHandle must not incrementing the reference count, which will happen when using LoadLibrary

mfleisz (approved):

PR #9810, opened 2024-01-26
[gcc,clang] add support for attribute(malloc)

akallabeth:
Mark all malloc like functions and add their appropriate free function to let the compiler complain on mismatches


akallabeth:

PR #9818, opened 2024-01-29
Shadow gfx fix

akallabeth:
delay gfx messages until the capability exchange is complete.

### DIFF
--- a/channels/rdpgfx/client/rdpgfx_main.c
+++ b/channels/rdpgfx/client/rdpgfx_main.c
@@ -314,6 +314,10 @@ static UINT rdpgfx_send_supported_caps(GENERIC_CHANNEL_CALLBACK* callback)
 			capsSet->flags = caps10Flags;
 		}
 
+		/* The following capabilities expect support for image scaling.
+		 * Disable these for builds that do not have support for that.
+		 */
+#if defined(WITH_CAIRO) || defined(WITH_SWSCALE)
 		if (!rdpgfx_is_capability_filtered(gfx, RDPGFX_CAPVERSION_105))
 		{
 			capsSet = &capsSets[pdu.capsSetCount++];
@@ -337,6 +341,7 @@ static UINT rdpgfx_send_supported_caps(GENERIC_CHANNEL_CALLBACK* callback)
 			capsSet->length = 0x4;
 			capsSet->flags = caps10Flags;
 		}
+#endif
 
 		if (!rdpgfx_is_capability_filtered(gfx, RDPGFX_CAPVERSION_107))
 		{
--- a/channels/rdpgfx/server/rdpgfx_main.c
+++ b/channels/rdpgfx/server/rdpgfx_main.c
@@ -41,6 +41,24 @@
 #define TAG CHANNELS_TAG("rdpgfx.server")
 #define RDPGFX_RESET_GRAPHICS_PDU_SIZE 340
 
+#define checkCapsAreExchanged(context) \
+	checkCapsAreExchangedInt(context, __FILE__, __func__, __LINE__)
+static BOOL checkCapsAreExchangedInt(RdpgfxServerContext* context, const char* file,
+                                     const char* fkt, size_t line)
+{
+	WINPR_ASSERT(context);
+	WINPR_ASSERT(context->priv);
+
+	const DWORD level = WLOG_TRACE;
+	if (WLog_IsLevelActive(context->priv->log, level))
+	{
+		WLog_PrintMessage(context->priv->log, WLOG_MESSAGE_TEXT, level, line, file, fkt,
+		                  "activeCapSet{Version=0x%08" PRIx32 ", flags=0x%08" PRIx32 "}",
+		                  context->priv->activeCapSet.version, context->priv->activeCapSet.flags);
+	}
+	return context->priv->activeCapSet.version > 0;
+}
+
 /**
  * Function description
  * Calculate packet size from data length.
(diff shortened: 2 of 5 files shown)


channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -233,6 +250,7 @@ static UINT rdpgfx_send_caps_confirm_pdu(RdpgfxServerContext* context,
 	else
 		Stream_Zero(s, capsSet->length);
 
+	context->priv->areCapsExchanged = TRUE;
```
pnowack, line 253: I'm not particularly happy about this commit (`[channels,rdpgfx] add server side channel checks`). The thing is, this is a race condition here. RDPGFX is in one part special (compared to other DVCs), and that is that the client side can reset the protocol (RDPGFX) at any time, as long the first accepted capability set was version 103 or later (See also `3.3.5.19 Processing an RDPGFX_CAPS_CONFIRM_PDU message`).

So, e.g. the client can send a `CapsAdvertise` PDU again, while the server side is in the process of sending the `CapsConfirm` PDU. So, the `areCapsExchanged` variable would not be able to always show the correct value. It completely depends on the actual server implementation here (how it handles that).

Overall, this is a race condition, which I don't like. I think this commit should therefore be dropped, I don't see a way to reliably handle this (since we don't know, how the API user wants to handle sudden protocol resets).

---

Some further thoughts: One alternative would be to name this variable `areInitialCapsExchanged`. Then the changes would here reflect that change just fine.

pnowack (commented):
The `[server,shadow] wait for rdpegfx channel ready ` commit looks fine. Only did a very brief look upon the wlog changes, but I think these are okay. Not a fan of the last commit though

channels/rdpgfx/server/rdpgfx_main.c
akallabeth: don´t see the race here.

1. `areCapsExchanged = true`
2. receive `CapsAdvertise -> respond and areCapsExchanged true -> true`

pnowack: > receive CapsAdvertise -> respond and areCapsExchanged true -> true

This assumes, that the server side directly handles the caps in the `CapsAdvertise` callback, but this is not necessarily true.
g-r-d for example stores the capabilities, notifies the renderer to finish current rendering operations and to prevent new ones, and then (in the thread where the graphics handling is done) answers the client with the `CapsConfirm` PDU.

akallabeth: then in your example there are still the previous caps in place and only when the server processed these the new ones.
still a `true -> true` transition from a channel state perspective.

pnowack: The thing is, the server side needs extra handling for cap version >=103 anyway, because upon confirming the caps again, it needs to recreate the whole "stage", i.e. reset graphics, recreate all surfaces, map them, submit an initial update for the mapped surfaces.
The server implementation very likely handles graphics in a different thread (g-r-d, I think krdp also uses a different thread), and if it would directly handle the caps in the `CapsAdvertise` callback, it would run into a protocol violation, when the new caps are already confirmed by submitting the graphics update before sending the RESET_GRAPHICS_PDU, the CREATE_SURFACE PDUs, etc.

akallabeth: hmm, the problem you try to point out is that this check can only confirm existence of exchanged capablilities and what you´re talking about is identifying which, lets call it `session`, the call belongs to, right?

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -314,6 +314,10 @@ static UINT rdpgfx_send_supported_caps(GENERIC_CHANNEL_CALLBACK* callback)
 			capsSet->flags = caps10Flags;
 		}
 
+		/* The followind capabilities expect support for image scaling.
```
pnowack, line 317: `followind` -> `following`

channels/rdpgfx/client/rdpgfx_main.c
```diff
@@ -314,6 +314,10 @@ static UINT rdpgfx_send_supported_caps(GENERIC_CHANNEL_CALLBACK* callback)
...
+		 * disable these for builds that do not have support for that.
```
pnowack, line 318: `disable` -> `Disable`

pnowack (commented):
Just two nits

channels/rdpgfx/client/rdpgfx_main.c
akallabeth: fixed

channels/rdpgfx/client/rdpgfx_main.c
akallabeth: fixed

channels/rdpgfx/server/rdpgfx_main.c
```diff
@@ -1213,6 +1272,8 @@ static UINT rdpgfx_recv_cache_import_offer_pdu(RdpgfxServerContext* context, wSt
  */
 static UINT rdpgfx_recv_caps_advertise_pdu(RdpgfxServerContext* context, wStream* s)
 {
+	if (checkCapsAreExchanged(context))
+		return ERROR_ALREADY_INITIALIZED;
```
pnowack, line 1276: This check here (in `rdpgfx_recv_caps_advertise_pdu`) needs to be removed, because this is actually allowed here.

pnowack (approved):
LGTM

PR #9819, opened 2024-01-29
[ci] add clang-tidy

akallabeth:


.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,28 @@
...
+on:
+  push:
+    branches: [ master, stable*  ]
+  pull_request:
+    branches: [ master, stable* ]
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
```
bwrsandman, line 11: Normally not needed

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,28 @@
+name: clang-tidy-review
+on:
+  push:
+    branches: [ master, stable*  ]
```
bwrsandman, line 4: clang-tidy review is meant to run on PRs not pushes.

bwrsandman (changes requested):
The fact that it's not positng might be that in your PR's context, it doesn't have access to the token from FreeRDP since it's running in akallabeth's context.
Not completely sure though.

When [I run it](https://github.com/openblack/openblack/blob/master/.github/workflows/ci-vcpkg.yml#L172C1-L178C29), it's in the [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) trigger which is a bit different from `pull_request` in that the tests come from the target and not the branch of the PR.

.github/workflows/clang-tidy.yml
akallabeth: ok, removed.

.github/workflows/clang-tidy.yml
akallabeth: ok, removed.

akallabeth:
@freerdp-bot test please

PR #9825, opened 2024-02-01
[channels] Add synchronous static channel setting.

llyzs:
Only rdpsnd and rdpdr use this setting in this PR for now. Note that rdpsnd could use SynchronousDynamicChannel setting depending on which channel it operates upon.

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.h
+++ b/channels/rdpdr/client/rdpdr_main.h
@@ -112,6 +112,7 @@ typedef struct
 	rdpContext* rdpcontext;
 	wStreamPool* pool;
 	wLog* log;
+	BOOL async;
 } rdpdrPlugin;
 
 BOOL rdpdr_state_advance(rdpdrPlugin* rdpdr, enum RDPDR_CHANNEL_STATE next);
--- a/channels/rdpsnd/client/rdpsnd_main.c
+++ b/channels/rdpsnd/client/rdpsnd_main.c
@@ -114,6 +114,7 @@ struct rdpsnd_plugin
 
 	size_t references;
 	BOOL OnOpenCalled;
+	BOOL async;
 };
 
 static const char* rdpsnd_is_dyn_str(BOOL dynamic)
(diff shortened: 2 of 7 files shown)


channels/rdpdr/client/rdpdr_main.h
```diff
@@ -70,6 +70,8 @@ typedef struct
 	DEVMAN* devman;
 	BOOL ignoreInvalidDevices;
 
+	BOOL async;
```
akallabeth, line 73: public header, add at end of struct!

channels/rdpsnd/client/rdpsnd_main.c
```diff
@@ -81,6 +81,7 @@ struct rdpsnd_plugin
 	BOOL attached;
 	BOOL connected;
 	BOOL dynamic;
+	BOOL async;
```
akallabeth, line 84: public header, add at end of struct!

akallabeth (commented):
overall LGTM, two things though:

1. new settings at end of structs
2. might be a nice thing to unify these settings (queue, thread, ...) in a common set of functions that are shared by the channels

llyzs:
1. Fixed.
2. Sounds like quite a bit changes, probably another change set if needed.

akallabeth (approved):
LGTM

channels/rdpsnd/client/rdpsnd_main.c
```diff
@@ -1636,13 +1652,17 @@ static UINT rdpsnd_on_data_received(IWTSVirtualChannelCallback* pChannelCallback
...
+		{
+			Stream_Release(copy);
+			return ERROR_INTERNAL_ERROR;
+		}
+		return CHANNEL_RC_OK;
 	}
-
-	return CHANNEL_RC_OK;
+	else
+		return rdpsnd_recv_pdu(plugin, copy);
```
mfleisz, line 1665: Nit-picking but imo code like this:
```
if (expr)
    return something special
return default;
```
is less error prone (i.e. to future changes that might introduce a path in the `if` or `else` that do not return) than:
```
if
    return;
else
    return
```

llyzs: Here I think it makes more sense if I just leave the last return as the last line and check error in the else path.

mfleisz (approved):
LGTM

PR #9828, opened 2024-02-02
include null byte in Authorization Request PDU

fifthdegree:
The RDS AAD Auth PDUs have no packet headers to indicate length. Instead, these packets are zero-terminated strings. Somehow, Windows accepts Authentication Request PDUs without a terminating null byte during regular connections, but not through WVD websocket gateways.

### DIFF
--- a/libfreerdp/core/aad.c
+++ b/libfreerdp/core/aad.c
@@ -505,6 +505,10 @@ static int aad_send_auth_request(rdpAad* aad, const char* ts_nonce)
 	if (stream_sprintf(s, "{\"rdp_assertion\":\"%s.%s.%s\"}", jws_header, jws_payload,
 	                   jws_signature) < 0)
 		goto fail;
+
+	/* Include null terminator in PDU */
+	Stream_Write_UINT8(s, 0);
+
 	Stream_SealLength(s);
 
 	if (transport_write(aad->transport, s) < 0)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth (commented):
maybe better add that to the `stream_sprintf` function, it should add the terminator

hardening (approved):

fifthdegree:
> maybe better add that to the `stream_sprintf` function, it should add the terminator

I'm not sure. Although right now the function is only used here, I think it would be unexpected for a function named `stream_sprintf` to include the terminator in the stream. That's my opinion; I won't argue if you insist on it, though:)

akallabeth:
> > maybe better add that to the `stream_sprintf` function, it should add the terminator
> 
> I'm not sure. Although right now the function is only used here, I think it would be unexpected for a function named `stream_sprintf` to include the terminator in the stream. That's my opinion; I won't argue if you insist on it, though:)

hmm, actually `stream_sprintf` already writes the `\0` but does advance the stream position only to the `\0` and not beyond.
so the fix looks good (was a nitpick anyway), but a `clang-format` is required to please the `qa` checker.

akallabeth:
@fifthdegree ok, fixed the clang-format. Thanks for the investigation and fix :)

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test please

PR #9833, opened 2024-02-04
[build] fix Wmismatched-deallocator warnings

akallabeth:
With custom malloc function attributes the fail path in the _New functions produces warnings due to allocator mismatches. Silence them.

follow up to #9810


akallabeth:
@freerdp-bot test

Issue #9358, opened 2023-09-04
TLS Channel binding issue with RSA-PSS signature


lemonmojo:
As discussed with @akallabeth on Matrix, this [commit](https://github.com/FreeRDP/FreeRDP/commit/3a10bcd36a062437af9db40b45260def71144b01) which was backported to 2.0 causes issues with certificates with RSA-PSS signature. Specifically, `OBJ_find_sigid_algs` returns `NID_undef`.

The implementation of `x509_utils_get_signature_alg` on [master](https://github.com/FreeRDP/FreeRDP/blob/5f8f69512b1307b341ef6cca1043b12679568bc4/libfreerdp/crypto/x509_utils.c#L719) uses `EVP_PKEY_get_default_digest_nid` instead which returns `NID_sha256` as expected and allows the connection to succeed.

So I suggest to backport the implementation of `x509_utils_get_signature_alg` to `crypto_cert_get_signature_alg` in stable 2.0.

PR #9838, opened 2024-02-05
[crypto,x509] fix tls-server-end-point signature algorithm selection

davidben:
This reverts commit 00baf58a71ccd0842e483ac034a0658e8945f887. That change appears to have been incorrect. It's described as simplying retrieving the "default signature digest", but it actually changed the function's behavior entirely. The function wasn't retrieving defaults previously.

A certificate contains, among other things, a public key and a signature. The public key is the public key of the subject. However, the signature was generated by the issuer. That is, if I get a certificate from a CA, the public key will be my public key and the signature will be my CA's signature over the certificate contents.

Now, the original code returned the digest used in the certificate's signature. That is, it tells you which signature algorithm did my *CA* use to sign my certificate.

The new code extracts the certificate's public key (my public key, not the CA's). This doesn't necessarily tell you the signature algorithm, so it then asks OpenSSL what the "default" signature algorithm would it use with the key. This notion of "default" is ad-hoc and has changed over time with OpenSSL releases. It doesn't correspond to any particular protocol semantics. It's not necessarily the signature algorithm of the certificate.

Now, looking at where this function is used, it's called by freerdp_certificate_get_signature_alg, which is called by tls_get_channel_binding to compute the tls-server-end-point channel binding. That code cites RFC 5929, which discusses picking the hash algorithm based on the certificate's signatureAlgorithm:

https://www.rfc-editor.org/rfc/rfc5929#section-4.1

That is, the old version of the code was correct and the "simplification" broke it. Revert this and restore the original version.

I suspect this went unnoticed because, almost all the time, both the old and new code picked SHA-256 and it was fine. But if the certificate was, say, signed with SHA-384, the new code would compute the wrong channel binding.


akallabeth (changes requested):
https://github.com/FreeRDP/FreeRDP/issues/9358

I don´t really get what the signing CA has to do with this here, but maybe I miss something?

channel binding uses the digest from the cipher spec (e.g. `TLS_AES_256_GCM_SHA384` is `SHA384` even if the CA signed the cert with `sha256`)

davidben:
It sounds like there's a lot of confusion here. Let me try to clarify:

> https://github.com/FreeRDP/FreeRDP/issues/9358

Ah, that was the motivation! Okay, so the correct thing to do then is revert 00baf58a71ccd0842e483ac034a0658e8945f887 and then fix it to extract the signature algorithm from PSS. This is, unfortunately, a bit tedious and kinda ill-defined, but I can see about putting that together for you.

00baf58a71ccd0842e483ac034a0658e8945f887 is not the solution for RSA-PSS and also handles PSS incorrectly. Again, you're mostly getting lucky that almost everyone uses SHA-256.

> I don´t really get what the signing CA has to do with this here, but maybe I miss something?

See RFC 5929, as cited in FreeRDP's own code. The tls-server-end-point channel binding is defined as:

> The hash function is to be selected as follows:
>
>   o  if the certificate's signatureAlgorithm uses a single hash
>       function, and that hash function is either MD5 [[RFC1321](https://www.rfc-editor.org/rfc/rfc1321)] or SHA-1
>      [[RFC3174](https://www.rfc-editor.org/rfc/rfc3174)], then use SHA-256 [[FIPS-180-3](https://www.rfc-editor.org/rfc/rfc5929#ref-FIPS-180-3)];
>
>    o  if the certificate's signatureAlgorithm uses a single hash
>       function and that hash function neither MD5 nor SHA-1, then use
>       the hash function associated with the certificate's
>       signatureAlgorithm;
> 
>    o  if the certificate's signatureAlgorithm uses no hash functions or
>       uses multiple hash functions, then this channel binding type's
>       channel bindings are undefined at this time (updates to is channel
>       binding type may occur to address this issue if it ever arises).

Note how it is talking about the certificate's **signatureAlgorithm**. That corresponds to this field here:
https://www.rfc-editor.org/rfc/rfc5280.html#section-4.1.1.2

What 00baf58a71ccd0842e483ac034a0658e8945f887 does is take the certificate's public key (https://www.rfc-editor.org/rfc/rfc5280.html#section-4.1.2.7) and then extract OpenSSL's arbitrary default hash function. In fact, notice how it changed what it returned between OpenSSL versions.

https://www.openssl.org/docs/man1.1.1/man3/EVP_PKEY_get_default_digest_nid.html
https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_get_default_digest_nid.html

> channel binding uses the digest from the cipher spec (e.g. TLS_AES_256_GCM_SHA384 is SHA384 even if the CA signed the cert with sha256)

Channel binding is a general technique. For specifically the tls-server-end-point channel binding, which is where FreeRDP uses this function, that is not true. Using the cipher suite is neither what the spec says, nor what 00baf58a71ccd0842e483ac034a0658e8945f887 implements. 

That is, none of three things are the same:
1. The PRF hash used in the TLS cipher suite
2. The signature algorithm used to sign the end-entity certificate
3. The hash algorithm that your particular version of OpenSSL happens to use by default with the end-entity certificate's public key

The one you've described is (1). The one in the spec is (2). The one in 00baf58a71ccd0842e483ac034a0658e8945f887 is (3).

davidben:
I'll add, the reason it didn't work with PSS is because PSS was defined kinda badly. Rather than being a single OID that captures the hash, it uses a single OID and then encodes a gnarly space of parameters in this RSASSA-PSS-params structure. Truly a mess and I wish people would just skip RSA-PSS. But if you all want to support it, I can help you put together code to inspect that.
https://www.rfc-editor.org/rfc/rfc4055.html

akallabeth:
@davidben ok, really nice explanation, thank you. Looks like I indeed misread the RFC quite a bit :/

davidben:
Honestly, I don't know why it didn't use the PRF hash of the cipher suite. That would have been _so_ much more sensible and straightforward. :-) I guess this way the binding is independent of the cipher suite that was negotiated, but that seems kinda unimportant.

Anyway, I'll update this PR with a fix for PSS.

davidben:
Alright, I have updated the PR with unit tests, and with RSA-PSS support.

I apologize for how truly horrible the PSS extraction function is. As far as I can tell (some context: my day job is working on cryptography in Chrome, and BoringSSL, an OpenSSL derivative), OpenSSL does not expose any useful APIs for this. And due to a serious of repeated bad decisions made when defining RSA-PSS, the encoding in X.509 and also the definition of the algorithm itself, is pointlessly complicated. I could probably rant for hours on how much of a standardization failure that was. Honestly, I think PSS in X.509 was botched so badly that we should all pretend it doesn't exist and toss it. (There's no good reason for it to be defined this way. RFC 4055 just made bad decisions.)

I did my best to match the surrounding style, but I probably got some of it wrong. Let me know what needs fixing.

I also only tested this against my system OpenSSL. Depending on how far back of OpenSSLs you all aim to support, I suspect it needs some compatibility functions. I can add those once I know which ones are needed. (Will CI flag any issues there?)

Also, ASN.1 is complicated, and OpenSSL's APIs are horrific, so let me know if anything in the PR is confusing.

hardening:

akallabeth:
> Alright, I have updated the PR with unit tests, and with RSA-PSS support.
> 
> I apologize for how truly horrible the PSS extraction function is. As far as I can tell (some context: my day job is working on cryptography in Chrome, and BoringSSL, an OpenSSL derivative), OpenSSL does not expose any useful APIs for this. And due to a serious of bad standardization decisions, both PSS's encoding in X.509 and PSS itself are pointlessly complicated. I could probably rant for hours on how much of a standardization failure that was. Honestly, I think PSS in X.509 was botched so badly that we should all pretend it doesn't exist and toss it. (There's no good reason for it to be defined this way. RFC 4055 just made bad decisions.)
> 
> I did my best to match the surrounding style, but I probably got some of it wrong. Let me know what needs fixing.
> 
> I also only tested this against my system OpenSSL. Depending on how far back of OpenSSLs you all aim to support, I suspect it needs some compatibility functions. I can add those once I know which ones are needed. (Will CI flag any issues there?)
> 
> Also, ASN.1 is complicated, and OpenSSL's APIs are horrific, so let me know if anything in the PR is confusing.

actually, looking quite nice. (well, as nice as it can be ;))
and really appreciating you take the time to fix this (the whole extraction was a headache as I´m not too deep into these RFC and I thing I managed to trip every trap along the way :/)

davidben:
> https://ci.freerdp.com//job/PullRequestTester/11360/

Oh, I missed that you all have a clang-format. Fixed, I think.

akallabeth:
@freerdp-bot test please

akallabeth (approved):
Ok, LGTM

really nice work, thank you for fixing this :)

PR #9843, opened 2024-02-07
Sample improve

akallabeth:
* fix location of cursor icon, install it
* use `winpr_image_*` to load sample server cursor image in different formats


akallabeth:
@pnowack not sure this affects `g-r-d` or not, could you verify?

PR #9844, opened 2024-02-07
[client,mac] fix CAPS handling

akallabeth:
* do not change capslock if state did not change (triggered by multiple flagsChanged calls while autorepeat of a pressed key)
* add debug log for flagsChanged


mfleisz (approved):
LGTM

PR #9855, opened 2024-02-10
[ci,tidy] remove unsupported options

akallabeth:
* workflow_dispatch does not work with plugin
* remove limited permissions


akallabeth:
@freerdp-bot test please

PR #9864, opened 2024-02-12
[server,sample] add bounds checks

akallabeth:
* test_peer_draw_icon should ignore requests if the resulting image is out of bounds of desktop size.


akallabeth (commented):
tested build:
1. cursor no longer drawn over desktop size
2. cursor stays at last drawable position

PR #9865, opened 2024-02-13
[cmake] guard cmake_policy(CMP0091)

akallabeth:
only set the policy to NEW if it is defined


akallabeth:
tested on debian 10, working

PR #9866, opened 2024-02-13
[winpr,crypto] fix WITH_INTERNAL_MD5 handling

akallabeth:
If WITH_INTERNAL_MD5=ON also use it with winpr_Digest_Init_Allow_FIPS


akallabeth:
confirmed by submitter that it is working, merging.

PR #9873, opened 2024-02-14
Bitmap formats

akallabeth:


akallabeth:
confirmed fixed by reporter.

PR #9905, opened 2024-02-26
[server,shadow] improve shadow help and startup checks

akallabeth:
a SAM file is required for the shadow server to work if NLA is activated. Better document that and abort startup if no file is available.

### DIFF
--- a/server/shadow/shadow_server.c
+++ b/server/shadow/shadow_server.c
@@ -55,6 +55,14 @@ static int shadow_server_print_command_line_help(int argc, char** argv,
 
 	printf("Usage: %s [options]\n", argv[0]);
 	printf("\n");
+	printf("Notes: By default NLA security is active.\n");
+	printf("\tIn this mode a SAM database is required.\n");
+	printf("\tProvide one with /sam-file:<file with path>\n");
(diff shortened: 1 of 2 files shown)


server/shadow/shadow_server.c
```diff
@@ -499,6 +504,13 @@ int shadow_server_parse_command_line(rdpShadowServer* server, int argc, char** a
 		if (!freerdp_settings_set_bool(settings, FreeRDP_NlaSecurity, FALSE))
 			return COMMAND_LINE_ERROR;
 	}
+
+	if (freerdp_settings_get_bool(settings, FreeRDP_NlaSecurity))
+	{
+		const char* sam = freerdp_settings_get_string(settings, FreeRDP_NtlmSamFile);
+		if (!sam)
```
ondrejholy, line 511: Just that this check probably prevents usage of the default `/etc/winpr/SAM` file unless its path isn't explicitly specified over `/sam-file:` argument. Is that intentional?

akallabeth: hmm, true.
so need to modify this to also check for a global user database.

akallabeth:
@freerdp-bot test

PR #9922, opened 2024-02-28
[client,win32] Child session fixes

hardening:
It seems like WaitFor[Single|Multiple]Object calls aren't reliable on pipes, especially on the pipe opened for childSession access. The object can be marked as signaled even if no data is available, making the connection laggy and unresponsive (nearly unusable in some cases).
This patch works around that by using ReadFileEx() with overlapped instead of simple ReadFile() and use asynchronous reads.

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -1022,7 +1022,7 @@ static DWORD WINAPI wf_client_thread(LPVOID lpParam)
 	rdpSettings* settings = context->settings;
 	WINPR_ASSERT(settings);
 
-	while (1)
+	while (!freerdp_shall_disconnect_context(instance->context))
 	{
 		HANDLE handles[MAXIMUM_WAIT_OBJECTS] = { 0 };
 		DWORD nCount = 0;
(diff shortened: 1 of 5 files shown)


client/Windows/wf_client.c
```diff
@@ -1045,7 +1045,9 @@ static DWORD WINAPI wf_client_thread(LPVOID lpParam)
 			nCount += tmp;
 		}
 
-		if (MsgWaitForMultipleObjects(nCount, handles, FALSE, 1000, QS_ALLINPUT) == WAIT_FAILED)
+		DWORD status = MsgWaitForMultipleObjectsEx(nCount, handles, 5 * 1000, QS_ALLINPUT,
```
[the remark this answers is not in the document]
hardening: Never seen such a stupid suggestion !

libfreerdp/core/childsession.c
```diff
@@ -192,15 +357,29 @@ static int transport_bio_named_uninit(BIO* bio)
...
+	if (!ptr->readEvent || ptr->readEvent == INVALID_HANDLE_VALUE)
+		goto error_event;
+
+	ptr->readOverlapped.hEvent = ptr->readEvent;
+
 	BIO_set_data(bio, ptr);
 	BIO_set_flags(bio, BIO_FLAGS_SHOULD_RETRY);
 	return 1;
+
+error_event:
```
akallabeth, line 378: shouldn´t this be cleaned up by `transport_bio_named_free` or `transport_bio_named_uninit`?

akallabeth (approved):
one little comment, otherwise LGTM

libfreerdp/core/childsession.c
hardening: I don't think that's possible as these functions rely on `BIO_set_data()` being called on the BIO (what we do at the end of transport_bio_named_new

akallabeth: then it would be useful to wrap the `cleanup` in a separate function as `ptr = BIO_get_data(); cleanup(ptr); BIO_set_data(NULL);` is what is done?

PR #9926, opened 2024-03-01
[channels,location] implement client stub channel

akallabeth:
to fix issues with RDP files that request the location channel provide a stub channel.

@pnowack do you know by chance if there is a simple library we can use or will this have to be a per platform imlementation to actually get the location?

### DIFF
--- a/channels/location/ChannelOptions.cmake
+++ b/channels/location/ChannelOptions.cmake
@@ -1,6 +1,6 @@
 
 set(OPTION_DEFAULT ON)
-set(OPTION_CLIENT_DEFAULT OFF)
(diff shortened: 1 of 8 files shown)


pnowack:
> @pnowack do you know by chance if there is a simple library we can use

There is `geoclue` for this. It is actually horrible for the server side, because it is a system service and thus having per (remote-) user session location is currently impossible with that. But for the client side, it should be okay-ish for now.

akallabeth:
@pnowack will keep that for a second pull request as `geoclue` is linux only?
with this pr we should have the channel up and running, the client just needs to hook up the `LocationStart` and `LocationStop` routines and send `LocationSend` in  between.

Issue #9925, opened 2024-03-01
Wrong colors with freerdp-shadow-cli 3 on big-endian 


ondrejholy:
**Describe the bug**
Colors are wrong when `freerdp-shadow-cli` 3.3.0 is started on big-endian. 

**To Reproduce**
1. Start `freerdp-shadow-cli` 3.3.0 on big-endian system (with a properly configured SAM file)
3. Connect to it over `xfreerdp /u:user /p:password /v:server` 3.3.0 or 2.11.5 (big-endian, or little-endian doesn't matter)
4. The colors are broken

**Expected behavior**
![Screenshot from 2024-02-28 12-48-43](https://github.com/FreeRDP/FreeRDP/assets/1051384/01de15b5-e350-4756-9e47-d46ef0d4d984)

**Screenshots**
![Screenshot from 2024-02-28 12-48-03](https://github.com/FreeRDP/FreeRDP/assets/1051384/6893710c-c1b6-485e-878a-3e0d2e9aee41)

**Application details**
[xfreerdp.txt](https://github.com/FreeRDP/FreeRDP/files/14445981/xfreerdp.txt)
[freerdp-shadow-cli.txt](https://github.com/FreeRDP/FreeRDP/files/14445983/freerdp-shadow-cli.txt)

**Environment (please complete the following information):**
Fedora 41 (s390x)

**Additional context**
The colors are ok when `freerdp-shadow-cli` is started on little-endian, or with 2.11.5.

All the unit tests from the test suite passed on the big-endian machine.

ondrejholy:
It seems that the main problem is:
```
diff --git a/server/shadow/X11/x11_shadow.c b/server/shadow/X11/x11_shadow.c
index 5c1fab18d..ef9c9062d 100644
--- a/server/shadow/X11/x11_shadow.c
+++ b/server/shadow/X11/x11_shadow.c
@@ -842,6 +842,7 @@ static int x11_shadow_screen_grab(x11ShadowSubsystem* subsystem)
                if (!empty)
                {
                        BOOL success = 0;
+                       UNIT32 format = (ImageByteOrder(subsystem->display) == LSBFirst) ? PIXEL_FORMAT_BGRX32 : PIXEL_FORMAT_XRGB32;
                        EnterCriticalSection(&surface->lock);
                        extents = region16_extents(&(surface->invalidRegion));
                        x = extents->left;
[...]

PR #9927, opened 2024-03-01
Shadow big endian

akallabeth:
@ondrejholy this should address #9925 but no way to properly test.

### DIFF
--- a/include/freerdp/server/shadow.h
+++ b/include/freerdp/server/shadow.h
@@ -325,10 +325,14 @@ extern "C"
 	WINPR_ATTR_MALLOC(shadow_server_free, 1)
 	FREERDP_API rdpShadowServer* shadow_server_new(void);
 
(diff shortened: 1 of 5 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

ondrejholy:
Thanks for looking into it. I've just made a quick test and see just a black screen now, so this is obviously not correct...

akallabeth:
@ondrejholy ok, did have a bug in there, should be fixed now. (new function now  did work on same color)

ondrejholy:
It works fine now, except for cursor colors. The `shadow_subsystem_pointer_convert_alpha_pointer_data` function also needs to take the `subsystem->format` into account. I've just tested that reverse ordering of the https://github.com/FreeRDP/FreeRDP/blob/3.3.0/server/shadow/shadow_subsystem.c#L240-L243 lines fix cursor colors on big-endian...

akallabeth:
@ondrejholy ok, did push a fix for the cursor color conversion function too

Issue #9924, opened 2024-03-01
Use vcpkg to manage dependencies


KangLin:
[Vcpkg](https://github.com/microsoft/vcpkg) helps you manage C and C++ libraries on Windows, Linux and MacOS.

it is recommended that the official freerdp use vcpkg.

Some bugs found: https://github.com/microsoft/vcpkg/pull/37015

akallabeth:
happy about pull requests addressing issues found.

as for direct integration, I´d like to hear some more arguments than `recommended` (or even better, create a pull that shows what you have in mind)
(there are various `dependency managers` for `C` and `C++` out there and all have their benefits and downsides)

PR #9931, opened 2024-03-04
vcpkg:  Use vcpkg to manage dependencies #9924

KangLin:
### Documents:
- https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html
- https://learn.microsoft.com/vcpkg/get_started/overview
- https://learn.microsoft.com/vcpkg/users/buildsystems/cmake-integration

### Usage:

- cmake >= v3.27.0
- get vcpkg

  ```
  git clone https://github.com/microsoft/vcpkg.git
  ```

- Set environment variable `VCPKG_ROOT` 
- mkdir build in FreeRDP

  ```
  cd FreeRDP
  mkdir build
  ```

- CMake configure

   ```
   cd build
   cmake .. --preset default-msvc     # in windows
   cmake .. --preset default-ninja      # in ubuntu
   ```

  If not set VCPKG_ROOT. please set CMAKE_TOOLCHAIN_FILE

  ```
  cmake .. --preset default-msvc  -DCMAKE_TOOLCHAIN_FILE="d:/source/vcpkg/scripts/buildsystems/vcpkg.cmake"
  ```

  See: 
  - https://learn.microsoft.com/vcpkg/users/buildsystems/cmake-integration
  - https://learn.microsoft.com/zh-cn/vcpkg/consume/manifest-mode?tabs=cmake%2Cbuild-cmake
- build

  ```
  cmake --build .
  ```

  The dependencies is installed `vcpkg_installed`
- install
  ```
  cmake --install . --prefix install
  ```
### It is divided into the following features(in vcpkg.json):

    freerdp(include libfreerdp, libwinpr, libwinpr-tool).
    channel
    client
    server

### Preset
- Completed

  "default"                     - Freerdp library minimal features
  "default-msvc"                - Default using Visual Studio 17 2022
  "default-ninja"               - Default using ninja
  "default-android-arm64"       - Default for android arm64
  "default-android-arm64-make"  - Default for android arm64 with make
  "default-android-x86_64"      - Default for android x86_64
  "default-android-x86_64-make" - Default for android x86_64 with make
  "freerdp"                     - Freerdp library with ffmpeg etc
  "freerdp-msvc"                - Freerdp using Visual Studio 17 2022
  "freerdp-ninja"               - Freerdp using ninja
  "freerdp-android-arm64"       - Freerdp for android arm64
  "freerdp-android-arm64-make"  - Freerdp for android arm64 with make
  "freerdp-android-x86_64"      - Freerdp for android x86_64
  "freerdp-android-x86_64-make" - Freerdp for android x86_64 with make
  "client"                      - Client
  "client-msvc"                 - Client using Visual Studio
  "client-ninja"                - Client using ninja
  "server"                      - Server
  "server-msvc"                 - Server using Visual Studio 17 2022
  "server-ninja"                - Server using ninja
  "all"                         - All
  "all-msvc"                    - All using Visual Studio 17 2022
  "all-ninja"                   - All using ninja

- Unfinished
  -  macos . I don't have a device, please add it who has a device.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,301 @@
+{
+	"version": 7,
+	"cmakeMinimumRequired": {
+		"major": 3,
+		"minor": 27,
+		"patch": 0
+	},
+	"configurePresets": [
+		{
+			"name": "vcpkg",
+			"displayName": "Configure with vcpkg. please set VCPKG_ROOT environment first",
+			"toolchainFile": "$env{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
+		},
+
+		{
+			"name": "default",
+			"displayName": "Freerdp library minimal features",
+			"description": "Freerdp library minimal features with vcpkg",
+			"inherits": "vcpkg",
+			"cacheVariables": {
+				"WITH_FFMPEG": "OFF",
+				"WITH_SWSCALE": "OFF",
+				"CHANNEL_URBDRC": "OFF",
+				"WITH_PKCS11": "OFF",
+				"WITH_CUPS": "OFF",
+				"WITH_MANPAGES": "OFF",
+				"WITH_KRB5": "OFF",
+				"WITH_CLIENT_COMMON": "ON",
+				"WITH_CLIENT": "OFF",
+				"WITH_SERVER": "OFF",
+				"WITH_PROXY": "OFF",
+				"WITH_SHADOW": "OFF",
+				"WITH_PLATFORM_SERVER": "OFF",
+				"WITH_VERBOSE_WINPR_ASSERT": "OFF",
+				"WITH_OPAQUE_SETTINGS": "OFF"
+			}
+		},
+		{
+			"name": "default-msvc",
+			"displayName": "Default using Visual Studio 17 2022",
+			"description": "Freerdp library minimal features using Visual Studio 17 2022",
+			"inherits": "default",
+			"generator": "Visual Studio 17 2022",
+			"condition": {
+				"type": "equals",
+				"lhs": "${hostSystemName}",
+				"rhs": "Windows"
+			}
+		},
+		{
+			"name": "default-ninja",
+			"displayName": "Default using ninja",
+			"description": "Freerdp library minimal features using ninja",
+			"inherits": "default",
+			"generator": "Ninja"
+		},
+		{
+			"name": "default-android",
+			"displayName": "Default for android",
+			"description": "Freerdp library minimal features for android",
+			"hidden": true,
+			"inherits": "default-ninja",
+			"cacheVariables": {
+				"VCPKG_CHAINLOAD_TOOLCHAIN_FILE":
+				    "$env{ANDROID_NDK_HOME}/build/cmake/android.toolchain.cmake"
+			}
+		},
+		{
+			"name": "default-android-arm64",
+			"displayName": "Default for android arm64",
+			"description": "Freerdp library minimal features for android arm64",
+			"inherits": "default-android",
+			"cacheVariables": {
+				"ANDROID_ABI": "arm64-v8a",
+				"VCPKG_TARGET_TRIPLET": "arm64-android"
+			}
+		},
+		{
+			"name": "default-android-arm64-make",
+			"displayName": "Default for android arm64 with make",
+			"description": "Freerdp library minimal features for android arm64 with make",
+			"inherits": "default-android-arm64",
+			"generator": "Unix Makefiles",
+			"cacheVariables": {
+				"CMAKE_MAKE_PROGRAM":
+				    "$env{ANDROID_NDK_HOME}/prebuilt/${hostSystemName}-x86_64/bin/make"
+			}
+		},
+		{
+			"name": "default-android-x86_64",
+			"displayName": "Default for android x86_64",
+			"description": "Freerdp library minimal features for android x86_64",
+			"inherits": "default-android",
+			"cacheVariables": {
+				"ANDROID_ABI": "x86_64",
+				"VCPKG_TARGET_TRIPLET": "x64-android"
+			}
+		},
+		{
+			"name": "default-android-x86_64-make",
+			"displayName": "Default for android x86_64 with make",
+			"description": "Freerdp library minimal features for android x86_64 with make",
+			"inherits": "default-android-x86_64",
+			"generator": "Unix Makefiles",
+			"cacheVariables": {
+				"CMAKE_MAKE_PROGRAM":
+				    "$env{ANDROID_NDK_HOME}/prebuilt/${hostSystemName}-x86_64/bin/make"
+			}
+		},
+
+		{
+			"name": "freerdp",
+			"displayName": "Freerdp library with ffmpeg etc",
+			"description": "Freerdp library with ffmpeg etc with vcpkg",
+			"inherits": "vcpkg",
+			"cacheVariables": {
+				"WITH_CLIENT_COMMON": "ON",
+				"WITH_CLIENT": "OFF",
+				"WITH_SERVER": "OFF",
+				"WITH_PROXY": "OFF",
+				"WITH_SHADOW": "OFF",
+				"WITH_PLATFORM_SERVER": "OFF",
+				"WITH_VERBOSE_WINPR_ASSERT": "OFF",
+				"WITH_OPAQUE_SETTINGS": "OFF",
+				"VCPKG_MANIFEST_FEATURES": "freerdp;channel"
+			}
+		},
+		{
+			"name": "freerdp-msvc",
+			"displayName": "Freerdp using Visual Studio 17 2022",
+			"description":
+			    "Freerdp library with ffmpeg etc  with vcpkg using Visual Studio 17 2022",
+			"inherits": "freerdp",
+			"generator": "Visual Studio 17 2022",
+			"condition": {
+				"type": "equals",
+				"lhs": "${hostSystemName}",
+				"rhs": "Windows"
+			}
+		},
+		{
+			"name": "freerdp-ninja",
+			"displayName": "Freerdp using ninja",
+			"inherits": "freerdp",
+			"generator": "Ninja"
+		},
+		{
+			"name": "freerdp-android",
+			"displayName": "freerdp for android",
+			"hidden": true,
+			"inherits": "freerdp",
+			"generator": "Unix Makefiles",
+			"cacheVariables": {
+				"VCPKG_CHAINLOAD_TOOLCHAIN_FILE":
+				    "$env{ANDROID_NDK_HOME}/build/cmake/android.toolchain.cmake"
+			}
+		},
+		{
+			"name": "freerdp-android-arm64",
+			"displayName": "Freerdp for android arm64",
+			"inherits": "freerdp-android",
+			"cacheVariables": {
+				"ANDROID_ABI": "arm64-v8a",
+				"VCPKG_TARGET_TRIPLET": "arm64-android"
+			}
+		},
+		{
+			"name": "freerdp-android-arm64-make",
+			"displayName": "Freerdp for android arm64 with make",
+			"inherits": "freerdp-android-arm64",
+			"generator": "Unix Makefiles",
+			"cacheVariables": {
+				"CMAKE_MAKE_PROGRAM":
+				    "$env{ANDROID_NDK_HOME}/prebuilt/${hostSystemName}-x86_64/bin/make"
+			}
+		},
+		{
+			"name": "freerdp-android-x86_64",
+			"displayName": "Freerdp for android x86_64",
+			"inherits": "freerdp-android",
+			"cacheVariables": {
+				"ANDROID_ABI": "x86_64",
+				"VCPKG_TARGET_TRIPLET": "x64-android"
+			}
+		},
+		{
+			"name": "freerdp-android-x86_64-make",
+			"displayName": "Freerdp for android x86_64 with make",
+			"inherits": "freerdp-android-x86_64",
+			"generator": "Unix Makefiles",
+			"cacheVariables": {
+				"CMAKE_MAKE_PROGRAM":
+				    "$env{ANDROID_NDK_HOME}/prebuilt/${hostSystemName}-x86_64/bin/make"
+			}
+		},
+
+		{
+			"name": "client",
+			"displayName": "Client",
+			"description": "Client with vcpkg",
+			"inherits": "vcpkg",
+			"cacheVariables": {
+				"WITH_CLIENT_COMMON": "ON",
+				"WITH_CLIENT": "ON",
+				"WITH_SERVER": "OFF",
+				"WITH_PROXY": "OFF",
+				"WITH_SHADOW": "OFF",
+				"WITH_PLATFORM_SERVER": "OFF",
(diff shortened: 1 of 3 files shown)


vcpkg.json
```diff
@@ -0,0 +1,62 @@
...
+				"openh264",
+				"opus",
+				"ffmpeg",
+				"cjson",
+				"libusb",
+				"opencl",
+				"cairo",
+				"uriparser",
+				"sdl2",
+				"sdl2-ttf"
```
akallabeth, line 30: missing `sld2-image`

vcpkg.json
```diff
@@ -0,0 +1,62 @@
...
+			]
+		},
+		"freerdp-with-ffmpeg": {
+			"description": "Enable the feature of freerdp library with ffmpeg",
+            "dependencies": [
+			    "openh264",
```
akallabeth, line 36: `openh264` and `ffmpeg` are alternatives.
while it is possible to compile against both, it is not recommended.

vcpkg.json
```diff
@@ -0,0 +1,62 @@
...
+		"channel": {
+			"description": "Enable the feature of channel",
+            "dependencies": [
+				"libusb"
+			]
+		},
+		"client": {
+			"description": "Enable client feature",
+            "dependencies": [
+			    "openh264",
```
akallabeth, line 49: same as above

CMakePresets.json
```diff
@@ -0,0 +1,117 @@
+{
+    "version": 3,
+    "cmakeMinimumRequired": {
+        "major": 3,
+        "minor": 20,
```
akallabeth, line 5: does this ave a specific reason or is this number just a guess?

CMakePresets.json
```diff
@@ -0,0 +1,117 @@
...
+    },
+    "configurePresets": [
+        {
+            "name": "freerdp",
+            "displayName": "freerdp library",
+            "cacheVariables": {
+				"WITH_FFMPEG": "OFF",
+				"WITH_SWSCALE": "OFF",
+				"CHANNEL_URBDRC": "OFF",
+				"WITH_CLIENT": "OFF"
```
akallabeth, line 16: you´re building `WITH_SERVER=ON` and `WITH_PROXY=ON`
they might actually be good things to consider for options ;)

akallabeth (commented):
Looks like a good starting point for discussion/refinement

A few notes:
* Formatting is a mess, best run the `json` through a beautifier first.
* The options seem incomplete (`server` components are not handled at all
* `cairo` and `swscale` are alternatives, no need to compile against both
* `ffmpeg/x264` and `openh264` are alternatives, no need to compile against both
* some missing (optional) dependencies (`sdl2-image`, `webp`, `png`, `jpeg`, ...)

vcpkg.json
KangLin: Which one is officially recommended？

CMakePresets.json
KangLin: Service is working now?

vcpkg.json
akallabeth: depends on what licensing terms you can be comfortable with.
* `x264` has better quality but is `GPL`
* `openh264` is `BSD`

CMakePresets.json
akallabeth: > Service is working now?

what do you mean by that?

vcpkg.json
```diff
@@ -0,0 +1,93 @@
...
+                    "name": "ffmpeg",
+                    "features": [
+                        "openh264"
+                    ]
+                },
+                "cjson",
+                "libusb",
+                {
+                    "name":"uriparser",
+                    "platform": "linux"
```
akallabeth, line 30: where does this idea come from?

KangLin: https://github.com/FreeRDP/FreeRDP/wiki/Compilation#dependencies

CMakePresets.json
KangLin: I remember when I used it last year, you said it was unfinished.

vcpkg.json
KangLin: I don't care about licensing agreements. I'm using openh264 for the time being.

vcpkg.json
KangLin: webp,png,jpeg is a dependency of sdl2-image. I only include sdl2-image. vcpkg progress it dependency.

CMakePresets.json
KangLin: WITH_SERVER=ON include server library and server application. so that whether it needs to be add WITH_SERVER_APP?
The WITH_CLIENT too.

akallabeth:
@KangLin this starts looking like something quite nice.
@pnowack @hardening @bmiklautz @mfleisz what do you think about this?

bmiklautz:
> @KangLin this starts looking like something quite nice. @pnowack @hardening @bmiklautz @mfleisz what do you think about this?

Having a preset file is a very nice idea.
As we are not using vcpkg anywhere ourselves at the moment  I'm not show how it will be kept up to date (if required) but besides that it looks good for me.

@KangLin thanks for the work!

hardening:
I don't know anything about vcpkg, but as far I can tell that seems ok.

hardening:

KangLin:
> best run `git clang-format` looks like `JSON` is already part of the files we check formatting with

```
l@l:/home/FreeRDP$ git clang-format vcpkg.json
no modified files to format
l@l:/home/FreeRDP$ git clang-format CMakePresets.json
no modified files to format
```

akallabeth:
@freerdp-bot test

akallabeth:
> > best run `git clang-format` looks like `JSON` is already part of the files we check formatting with
> 
> ```
> l@l:/home/FreeRDP$ git clang-format HEAD
> no modified files to format
> ```

well, if you already committed your changes better run `git clang-format HEAD~1` or `git clang-format origin/master` ;)

pnowack:
> @pnowack @hardening @bmiklautz @mfleisz what do you think about this?

Never heard of vcpkg, so no idea about that. What wonders me though is that if vcpkg is used to manage dependencies (I assume: instead of cmake doing that), why are here no line removals (of the old behaviour)?
Is this also a change, that is applicable for FreeRDP3 or would this add another dependency here?

akallabeth:
@pnowack https://vcpkg.io/ is, similar to https://conan.io/ a package manager for `C/C++`. it integrates into `CMake` via a toolchain file and calls to `find_package` et al result in the dependency being installed.

while this is nice to use if you just want to create some binary without the need to think about wher to get/how to install dependencies it has the (usual) drawback of `do you trust the creator of the build recipie or binaries` with these package managers on a whole different level than if you use e.g. your distributions package manager.

so, just like @bmiklautz said, the `cmake presets` are definitely fine as we can use them, but the `vcpkg` related configurations need to be maintained and I don´t see who here could do that as nobody is using it. (of course, volunteers welcome)

mfleisz:
Not a vcpkg expert either but it lgtm

KangLin:
>  What wonders me though is that if vcpkg is used to manage dependencies (I assume: instead of cmake doing that), why are here no line removals (of the old behaviour)?

See: VCPKG_MANIFEST_FEATURES in https://learn.microsoft.com/vcpkg/users/buildsystems/cmake-integration

Because the toolchain file is evaluated during the project() call, all CMake-level variables that modify a vcpkg setting must be set before the first call to project(). 
So if use VCPKG_MANIFEST_FEATURES. then need reconstruct the whole of CMakeLists.txt.
Of course this one is the best. But this one needs official support.

Another way to do this is to use cmake preset. or manual operation.

KangLin:
> so, just like @bmiklautz said, the cmake presets are definitely fine as we can use them, but the vcpkg related configurations need to be maintained and I don´t see who here could do that as nobody is using it. (of course, volunteers welcome)

Because there is no the vcpkg now, no one uses it. I think there are a lot of places where you need to use the vcpkg:
- FreeRDP libfreerdp(eg: ffmpeg, openh264, x264), clients(eg: sdl, x11) and servers require vcpkg. It's difficult for new users to deal with these dependencies.
- Use the libfreerdp library as a dependent program(eg: [RabbitRemoteControl](https://github.com/KangLin/RabbitRemoteControl), remmina etc)

akallabeth:
@KangLin while the arguments are true to set up a quick dev env,
the trust issue accompanied with such dependency managers remains (same as node.js and debian repos)
while I put a little trust in the latter, I don´t trust the first, for example

and I don´t quite get what you mean by

> Whether libfreerdp-client and libfreerdp-server are stand-alone features or integrated into freerdp will require changes to the current CMakeLists.txt, which can be discussed in the next step.

you can either build them shared or static

KangLin:
when WITH_CLIENT is ON. the libfreerdp-client and the client application are generated at the same time.
WITH_SERVER is ON. the libfreerdp-server and the server application are generated at the same time.
You need to control the build library and the application separately. eg: add WITH_SERVER_APP

KangLin:
> @KangLin while the arguments are true to set up a quick dev env, the trust issue accompanied with such dependency managers remains (same as node.js and debian repos) while I put a little trust in the latter, I don´t trust the first, for example
> 
> and I don´t quite get what you mean by

Sure, there is a crisis of trust. Would we like to be the first to eat crabs :)

akallabeth:
> when WITH_CLIENT is ON. the libfreerdp-client and the client application are generated at the same time. WITH_SERVER is ON. the libfreerdp-server and the server application are generated at the same time. You need to control the build library and the application separately. eg: add WITH_SERVER_APP

check the remaining `CMake` variables ;)
yes, by default the client and server implementations are `ON` but you can deactivate each one of them.
`ccmake` or the `cmake-gui` can help there to find them more easily

hardening:
@KangLin if I where to want to test this PR with a sample project using vcpkg, how can I demonstrate that it works ?

KangLin:
> @KangLin if I where to want to test this PR with a sample project using vcpkg, how can I demonstrate that it works ?

Use the PR can test. See the top for details.

dg0yt:
> Formatting is a mess, best run the `json` through a beautifier first.

vcpkg has `vcpkg format-manifest <filepath>`.

akallabeth:
@KangLin ok, so I´ve tested the default build option `--preset default-ninja`:
1. the preset should enable all features and not have half of them disabled (`URBDRC`, `Opus`, `WINPR_UTILS_IMAGE_*`, `WITH_CUPS`, `WITH_FFMPEG` or `WITH_FAAC WITH_FAAD2 WITH_GSM WITH_LAME`, `WITH_INTERNAL_*`, `WITH_PKCS11`, `WITH_PULSE`, `WITH_SERVER`, `WITH_SOXR`, `WITH_SWSCALE`
2. highly encouraged: `WITH_OPAQUE_SETTINGS` as we are deprecating direct struct access
3. for release builds `WITH_VERBOSE_WINPR_ASSERT=OFF`
4. respect `BUILD_SHARED_LIBS`, currently `vcpkg` related dependencies are just linked statically ignoring this.

dg0yt:
> respect `BUILD_SHARED_LIBS`, currently `vcpkg` related dependencies are just linked statically ignoring this.

`BUILD_SHARED_LIBS` is mostly for controlling the current build but not the dependencies. For controlling the library linkage for dependencies from vcpkg, the vcpkg target triplet must be chosen as needed.

CMakePresets.json
```diff
@@ -0,0 +1,234 @@
...
+			"displayName": "Default using ninja",
+			"inherits": "default",
+			"generator": "Ninja"
+		},
+		{
+			"name": "default-android",
+			"displayName": "Default for android",
+			"hidden": true,
+			"inherits": "default",
+			"generator": "Unix Makefiles",
```
akallabeth, line 55: no need for that, we're building android targets with `Ninja` as well

CMakePresets.json
```diff
@@ -0,0 +1,234 @@
...
+				"WITH_FFMPEG": "OFF",
+				"WITH_SWSCALE": "OFF",
+				"CHANNEL_URBDRC": "OFF",
+				"WITH_PKCS11": "OFF",
+				"WITH_CUPS": "OFF",
+				"WITH_MANPAGES": "OFF",
+				"WITH_KRB5": "OFF",
+				"WITH_CLIENT": "OFF",
+				"WITH_SERVER": "OFF",
+				"WITH_PROXY": "OFF"
```
akallabeth, line 30: `WITH_SERVER=OFF` also does not build the `libfreerdp-server3.so` (we did a better dependency with `WITH_CLIENT=OFF` where the library is outside with `WITH_CLIENT_COMMON`)
so better keep that on and deactivate the components `WITH_SHADOW`, `WITH_PROXY` and `WITH_PLATFORM_SERVER`

CMakePresets.json
KangLin: This is for compiling Android on Windows

KangLin:
> @KangLin ok, so I´ve tested the default build option `--preset default-ninja`:
> 
>     1. the preset should enable ~all features~ (mostly ok with preset `all-ninja`) and not have half of them disabled (`URBDRC`, `Opus`, `WINPR_UTILS_IMAGE_*`, `WITH_CUPS`, `WITH_FFMPEG` or `WITH_FAAC WITH_FAAD2 WITH_GSM WITH_LAME`, `WITH_INTERNAL_*`, `WITH_PKCS11`, `WITH_PULSE`, `WITH_SERVER`, `WITH_SOXR`, `WITH_SWSCALE`

The default is the minimum dependency. The all is the maximum dependency.

>     2. highly encouraged: `WITH_OPAQUE_SETTINGS` as we are deprecating direct struct access

I think this is best set in CMakeLists.txt.

>     3. for release builds `WITH_VERBOSE_WINPR_ASSERT=OFF`

I think this is best set in CMakeLists.txt.

>     5. I really dislike that `vcpkg` ignores system installed libraries and downloads its own set regardless of availability

Yes, so I only use it on systems that don't have pre-installed(eg: windows, android). and not used on linux(eg: ubuntu, debain etc)

dg0yt:
Given https://github.com/microsoft/vcpkg/pull/37015#issuecomment-1996286423, I wonder if it might be necessary to clarify that this PR is not relevant for consuming `freerdp` in vcpkg manifest mode (using the official vcpkg registry of ports).
This project manifest is useful for facilitating quick-start of development of freerdp and for building included apps.

KangLin:
> Given [microsoft/vcpkg#37015 (comment)](https://github.com/microsoft/vcpkg/pull/37015#issuecomment-1996286423), I wonder if it might be necessary to clarify that this PR is not relevant for consuming `freerdp` in vcpkg manifest mode (using the official vcpkg registry of ports). This project manifest is useful for facilitating quick-start of development of freerdp and for building included apps.

dg0yt:
> I really dislike that `vcpkg` ignores system installed libraries and downloads its own set regardless of availability

I do understand that and the downside of this behavior, in particular when it comes to core system services.
OTOH vcpkg has its own approach on controlling what goes into a build and affects the binary artifacts, regardless of platform. This is based on tracking the ports which are the dependencies. The artifacts shall not vary depending on whether a system libs is installed or not. And there is a pattern for how to choose system libraries: 

Provide empty overlay ports for those dependencies. 

It may seem inconvenient, but it allows for proper tracking. (Well, they won't be entirely empty, and they may even need adapt some polyfill from vcpkg.)

Issue #9932, opened 2024-03-04
Unable to print 2 times on the same printer on Windows


dev-ddoe:
**Describe the bug**
When I start RDP session with a printer redirected, it is possible to print the first time. The second time, the Windows server report an error

**To Reproduce**
1. Open a RDP session from a Windows workstation with a printer redirected
2. Open notepad
3. Print on the redirected printer => the print passed
4. Print on the redirected printer a second time
5. The server report a print error

**Expected behavior**
* No error reported server side
* The job is printed

**Application details**
* FreeRDP version: 3.3.1-dev0 (aebe9742e) from https://ci.freerdp.com/job/freerdp-nightly-windows/arch=win64,label=vs2017/1484/
* Command line used wfreerdp.exe /u:Administrator /p:pass /v:server-ip  /printer:POUET
* OS version connecting to (server side) Windows server 2016

**Environment (please complete the following information):**
 - OS: Windows 10
 - Architecture: amd64

PR #9934, opened 2024-03-04
Prevent printer closing when a job is closed

dev-ddoe:
It prevents future job to print

fix https://github.com/FreeRDP/FreeRDP/issues/9932


akallabeth:
@freerdp-bot test

PR #9944, opened 2024-03-07
make local gettimeofday helper static

awakecoding:
We have a local gettimeofday() helper for Windows in libfreerdp/utils/pcap.c that wasn't defined as static. There was no reason not to make it static, and it caused a symbol conflict with LibreSSL on Windows.

### DIFF
--- a/libfreerdp/utils/pcap.c
+++ b/libfreerdp/utils/pcap.c
@@ -38,7 +38,7 @@
 #include <sys/timeb.h>
 #include <winpr/windows.h>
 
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
pr looks ok, but grepping through code this looks like it needs some cleanup:

* there are multiple definitions of `gettimeofday`
* sometimes there are fallbacks to `clock_gettime` sometimes not

so will check this first.

akallabeth:
replaced by #9949

PR #9949, opened 2024-03-08
[winpr,sysinfo] unify time function use

akallabeth:
* Add new function winpr_GetTickCount64NS for high resolution tick count with (up to) nanosecond resolution
* Add new function winpr_GetUnixTimeNS for high resolution system time as nanoseconds since 1.1.1970
* Replace use of clock_gettime and gettimeofday in whole project with these new functions

### DIFF
--- a/winpr/libwinpr/sspi/NTLM/ntlm_compute.c
+++ b/winpr/libwinpr/sspi/NTLM/ntlm_compute.c
@@ -248,15 +248,12 @@ BOOL ntlm_write_ntlm_v2_response(wStream* s, const NTLMv2_RESPONSE* response)
 
 void ntlm_current_time(BYTE* timestamp)
 {
-	FILETIME filetime = { 0 };
-	ULARGE_INTEGER time64 = { 0 };
+	FILETIME ft = { 0 };
 
 	WINPR_ASSERT(timestamp);
 
-	GetSystemTimeAsFileTime(&filetime);
-	time64.u.LowPart = filetime.dwLowDateTime;
-	time64.u.HighPart = filetime.dwHighDateTime;
-	CopyMemory(timestamp, &(time64.QuadPart), 8);
+	GetSystemTimeAsFileTime(&ft);
+	CopyMemory(timestamp, &(ft), sizeof(ft));
 }
 
 /**
--- a/winpr/libwinpr/sysinfo/sysinfo.c
+++ b/winpr/libwinpr/sysinfo/sysinfo.c
@@ -33,8 +33,42 @@
 #include <fcntl.h>
 #endif
 
+#if !defined(_WIN32)
+#if defined(_POSIX_SOURCE) && (_POSIX_C_SOURCE >= 200112L)
+#include <time.h>
+#elif !defined(__APPLE__)
+#include <sys/time.h>
+#include <sys/sysinfo.h>
+#endif
+#endif
+
 #include "../log.h"
 #define TAG WINPR_TAG("sysinfo")
+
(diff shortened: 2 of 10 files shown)


winpr/libwinpr/sspi/NTLM/ntlm_compute.c
```diff
@@ -248,15 +248,16 @@ BOOL ntlm_write_ntlm_v2_response(wStream* s, const NTLMv2_RESPONSE* response)
...
+	} now = { 0 };
 
 	WINPR_ASSERT(timestamp);
 
-	GetSystemTimeAsFileTime(&filetime);
-	time64.u.LowPart = filetime.dwLowDateTime;
-	time64.u.HighPart = filetime.dwHighDateTime;
-	CopyMemory(timestamp, &(time64.QuadPart), 8);
+	GetSystemTimeAsFileTime(&now);
+	CopyMemory(timestamp, &(now.ul.QuadPart), sizeof(now.ul.QuadPart));
```
mfleisz, line 260: Why are we using this `union` magic here?
Wouldn't a simple `CopyMemory(timestamp, &now, sizeof(now));` work just fine?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -553,35 +544,78 @@ DWORD GetTickCount(void)
...
-	ticks = ul.QuadPart;
+	union
+	{
+		FILETIME ft;
+		ULARGE_INTEGER ul;
+	} t;
+
+	GetSystemTimeAsFileTime(&t.ft);
+
+	static ULARGE_INTEGER first = { 0 };
```
mfleisz, line 571: What do we do that static `first` stuff?
This changes the behavior of `GetTickCount64` as it will always return `0` on its first invocation (which might be unexpected since this should theoretically not be possible).

The Linux implementation now might also return negative (overflowed) values if `gettimeofday` fails.

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -553,35 +544,78 @@ DWORD GetTickCount(void)
...
-	FILETIME ft;
-	ULARGE_INTEGER ul;
-	GetSystemTimeAsFileTime(&ft);
-	ul.LowPart = ft.dwLowDateTime;
-	ul.HighPart = ft.dwHighDateTime;
-	ticks = ul.QuadPart;
+	union
+	{
+		FILETIME ft;
+		ULARGE_INTEGER ul;
```
mfleisz, line 566: Microsoft actually recommends using native 64-bit integer types instead of their `LARGE_INTEGER` structs: https://learn.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-large_integer-r1

winpr/libwinpr/sysinfo/sysinfo.c
akallabeth: hmm, true, thought it was application start, but it is system start.
problem is that `GetSystemTimeAsFileTime` value is much too large for the `GetTickCount` and we need some reasonable cutoff.

akallabeth: ok, fixed under windows by using `QueryPerformanceFrequency` with `QueryPerformanceCounter`
the fallback to `gettimeofday` now remembers the `uptime` on first measurement and returns the difference to that point in time.

akallabeth:

mfleisz (approved):
LGTM

mfleisz (approved):
LGTM

PR #9955, opened 2024-03-08
[sdl] Only blit rectangles to windows that they intersect

caseif:
This pull request modifies the `sdl_draw_to_window_rect` to skip attempting to blit rects which do not intersect the window the function is currently operating on. This prevents `SdlWindow#blit` from returning false and short-circuiting the drawing loop. In a single-window session, it's reasonable to assume that something is wrong if any of the rectangles being blitted do not interesect the window, but in a multimon session each rectangle will only be applicable for one window. Because the primary window is always handled first, all rectangles which intersect a different window will fail to be blitted to the primary one and immediately short-circuit the draw loop.

This PR doesn't modify the analogous functions for smart-sizing, because currently smart-sizing appears to be completely broken on SDL with multimon enabled.

### DIFF
--- a/client/SDL/sdl_freerdp.cpp
+++ b/client/SDL/sdl_freerdp.cpp
@@ -339,8 +339,14 @@ static bool sdl_draw_to_window_rect(SdlContext* sdl, SdlWindow& window, SDL_Surf
 		return sdl_draw_to_window_rect(sdl, window, surface, offset,
 		                               { 0, 0, surface->w, surface->h });
 	}
+
+	bool multimon = freerdp_settings_get_bool(sdl->context()->settings, FreeRDP_UseMultimon) != 0;
+	SDL_Rect windowRect = window.rect();
 	for (auto& srcRect : rects)
 	{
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
nice catch,
but I´d move the check to `sdl_draw_to_window_rect` and simply ignore regions that are not intesecting (without any multimonitor check)

there might arise scenarios where this will be true for even single monitor sessions, so no need to make it more complicated than necessary.

akallabeth:
on second thought, might be simplest to ignore these in `SdlWindow::blit`

akallabeth:
@caseif thank you for bringing this to my attention.
I´ve created #9962 to ignore `SdlWindow::blit` that does not intersect instead though, as we don´t necessarily know the updates intersect with the surface (even in single window mode, once we add resizeability without resolution changes)

caseif:
Thanks for taking care of this. This was my initial thought but I was a bit nervous about potentially obscuring an error condition, but if that's not an issue then I agree it's a much nicer solution. :)

PR #9956, opened 2024-03-08
[win32] various fix for childSession

hardening:
WAIT_TIMEOUT was not correctly handled in non blocking mode. Also there was a segfault when freeing the transport (double free).


akallabeth (approved):
LGTM

PR #9977, opened 2024-03-14
Export usage requirements

dg0yt:
Exported CMake Config may carry imported targets:
~~~
$<LINK_ONLY:cjson>
$<LINK_ONLY:ZLIB::ZLIB>
ZLIB::ZLIB
~~~
These must be resolved when the config is used.
Exported pc files did't carry these usage requirements at all.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #9990, opened 2024-03-18
[winpr,sysinfo] clock_gettime is available since POSIX 1993

akallabeth:


hardening (approved):

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -34,7 +34,7 @@
 #endif
 
 #if !defined(_WIN32)
-#if defined(_POSIX_SOURCE) && (_POSIX_C_SOURCE >= 200112L)
+#if defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 199309L)
```
VVD, line 37: If you want to check posix version then use:
```
#if defined(_POSIX_VERSION) && (_POSIX_VERSION >= 199309L)
```
instead.

akallabeth: @VVD no. `_POSIX_VERSION` is the version the system supports, `_POSIX_C_SOURCE` the version the application requests. this code should only be compiled if the application requests a `POSIX` standard that supports the functions regardless of what the system would be capable of.
if you have issues with the posix implementation you´re using being awkward, fix that or use the non posix implementations of the functions.

VVD: https://stackoverflow.com/questions/11350878/how-can-i-determine-if-the-operating-system-is-posix-in-c
https://lists.freebsd.org/pipermail/freebsd-standards/2004-March/000474.html

akallabeth: no, request is to define the symbol somewhere.
a `#define` is just one way. you can also add `CFLAGS=-D_POSIX_C_SOURCE=xxx` or let your buildsystem detect it somehow and add a compiler define.
we do it per platform and you happen to be on one where you have to do that work as we do have active test runners.

VVD: Where FreeRDP define `_POSIX_C_SOURCE`?
I can't find it anywhere.

akallabeth: then we´re back at your platform has issues with using `POSIX` along with other `API` calls, so fix that.

VVD: So `FreeRDP` doesn't define `_POSIX_C_SOURCE` anywhere, but expects it to be set by someone outside of `FreeRDP`?
That's not how it works. The application itself must define the required `_POSIX_C_SOURCE` value if the sources fully comply with it (but that's not true: for example it use non-posix u_long and ETIMEDOUT).
But if you want just check support of the posix version then use `_POSIX_VERSION` instead.

You got the direction wrong:
`_POSIX_C_SOURCE` must be defined by application for `C compiler` and `libc`.
`_POSIX_VERSION` must be defined by `C compiler` and `libc` for application.
An application can only check `_POSIX_C_SOURCE` if it defines it itself (via #define or via CFLAGS).

VVD: One more explanation: https://lwn.net/Articles/590381/#CommAnchor592882

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -579,7 +579,7 @@ ULONGLONG winpr_GetTickCount64(void)
 UINT64 winpr_GetTickCount64NS(void)
 {
 	UINT64 ticks = 0;
-#if defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 200112L)
+#if defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 199309L)
```
bsdimp, line 582: same comment.

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -632,7 +632,7 @@ UINT64 winpr_GetUnixTimeNS(void)
 	} t = { 0 };
 	GetSystemTimeAsFileTime(&t.ft);
 	return (t.u64 - FILETIME_TO_UNIX_OFFSET_S * 10000000ull) * 100ull;
-#elif _POSIX_C_SOURCE >= 200112L
+#elif defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 199309L)
```
bsdimp, line 635: same comment.

winpr/libwinpr/sysinfo/sysinfo.c
akallabeth: we´re using it exactly that way.
the application sets this (with a `CMake` platform define or by inclusion of platform headers) and we´re checking here for a conformance level (same as we check for `_WIN32` and `__APPLE__ or similar)
we do not check `_POSIX_VERSION` because that is not always the desired compatibility level (`_POSIX_C_SOURCE` is defined by the applictaion, `_POSIX_VERSION` not)

`POSIX_SOURCE` was wrong, yes, that is the reason we eliminated it.

bsdimp: It appears that your default build system (at least for FreeBSD, confirmed by grep) doesn't define `_POSIX_C_SOURCE`. `_WIN32` and `__APPLE__` are defined by the compiler, while `_POSIX_C_SOURCE` is defined by the user. Is it a requirement of the library that ports to a POSIX environment add this define somehow to the build environment? I don't see any place it's defined, and would have expected to see that in your sources.

akallabeth: we do not have a freebsd test node (well, I just did set up one, so that might change), but yes.
I´m currently restructuring the platform specific cmake settings to have them all in one place.

did a little playing around, does `FreeBSD` have something like `MacOS` `_DARWIN_C_SOURCES` or `Linux/GNU` `_DEFAULT_SOURCE` or `_GNU_SOURCE` which allows mixing `POSIX` with additional `API`?

I´ve been told the `__BSD_VISIBLE` are internal which is unexpected as they can not be reactivated once you detect `_POSIX_C_SOURCE`? (all other variables can be mixed, just that one not?)

VVD: Just use `__FreeBSD__` - all extensions already included.

PR #9992, opened 2024-03-18
Incompatible pointer fix

akallabeth:


hardening (approved):

winpr/libwinpr/clipboard/synthetic_file.c
```diff
@@ -276,7 +276,13 @@ static BOOL add_directory_contents_to_list(wClipboard* clipboard, const WCHAR* l
...
 {
 	BOOL result = FALSE;
-	const WCHAR* wildcard = "/\0*\0\0\0";
+	union
+	{
+		const char* c;
+		const WCHAR* w;
+	} wildcard;
+	const char buffer[4] = "/\0*\0\0\0";
+	wildcard.c = buffer;
```
floppym, line 285: This feels a bit messy to me. Doesn't this risk unaligned access?

Also, this will populate the data as little-endian, even on big-endian platforms. Is that intentional?

floppym: Also, you are putting 7 bytes into a 4 byte buffer.

akallabeth: yes, little endian is intentional. alignment does not matter. and no, these are not 7 bytes, they are cut short due to definition. (but I just realized that it should state `buffer[6]`, damn, fix incoming.

floppym: > these are not 7 bytes, they are cut short due to definition

The string literal is 7 bytes due to the implicit `\0` at the end. However, it seems GCC will silently drop that implicit `\0` when you assign it to an array that is undersized by 1.

PR #9994, opened 2024-03-19
[win32,chilsession] fix path computation in 32 bits

hardening:
The path was not adjusted correctly (wrong WCHAR pointer computation).


akallabeth (approved):
LGTM

PR #9995, opened 2024-03-19
Posix refactor

akallabeth:
* unify `POSIX` checks
* supply platform defaults for known systems at a single location (not completed yet)
* move `#ifdef` madness to (configurable) `CMake` file as far as possible


hardening (approved):
Seems ok for me

PR #9996, opened 2024-03-19
[cmake] unify windows resource file generation

akallabeth:
* use a macro to add resource files for executables and shared libraries


hardening (approved):

server/Windows/cli/CMakeLists.txt
```diff
@@ -25,38 +25,12 @@ set(${MODULE_PREFIX}_SRCS
...
-add_executable(${MODULE_NAME} ${${MODULE_PREFIX}_SRCS})
-
-if (WITH_BINARY_VERSIONING)
-	set_target_properties(${MODULE_NAME} PROPERTIES OUTPUT_NAME "${OUTPUT_NAME}${FREERDP_API_VERSION}")
-	set_target_properties(${MODULE_NAME} PROPERTIES PDB_NAME "${OUTPUT_NAME}${FREERDP_API_VERSION}${CMAKE_EXECUTABLE_SUFFIX}")
-else()
-	set_target_properties(${MODULE_NAME} PROPERTIES OUTPUT_NAME "${OUTPUT_NAME}")
-	set_target_properties(${MODULE_NAME} PROPERTIES PDB_NAME "${OUTPUT_NAME}${CMAKE_EXECUTABLE_SUFFIX}")
-endif()
+AddTargetWithResourceFile(${MODULE_NAME} TRUE "${FREERDP_VERSION}" ${MODULE_PREFIX}_SRCS)
```
dg0yt, line 28: This unexpectedly changed the tool name from `wfreerdp-server` (aka `OUTPUT_NAME`) to `winpr-server-cli` (aka `MODULE_NAME`). In a patch release. Was this intented? 
(Noticed in vcpkg packaging. At least the PDB files should match now.)

akallabeth: no, this was unintentional.
but `wfreerdp-server` is something that has strong warning signs all over the place as incomplete and unmaintained, so better deactivate that.

dg0yt: In vcpkg, the user must explicitly ask for feature `server`. But it is enabled in CI at least when changing the freerdp port.

Where can I find the strong warnings signs?

akallabeth: @dg0yt 
1. you get a very prominent warning during `CMake` configure run
2. you get a very prominent warning during application start

just like you see it with `wlfreerdp` (deprecation) and `sdl-freerdp` (experimental)

dg0yt: Okay, so I don't see it because I'm very succesful in maintaining MSVC builds through vcpkg CI only. ;-)

I guess it is these two lines:
https://github.com/FreeRDP/FreeRDP/blob/ef83633c0181f1a1717939a06301639cc1602dd8/server/Windows/CMakeLists.txt#L21-L22

akallabeth: if you don´t check build logs then yes.

dg0yt: I only get logs when the build fails... But at least I consider unexpected name changes noteworthy.

I still wonder what is the prefered action: Restore the original name, or simply keep the change.

PR #10012, opened 2024-03-22
client/Windows/wf_cliprdr.c: handle malloc error

chipitsine:


client/Windows/wf_cliprdr.c
```diff
@@ -2133,6 +2133,8 @@ wf_cliprdr_server_format_data_request(CliprdrClientContext* context,
 			globlemem = (char*)GlobalLock(hClipdata);
 			size = (int)GlobalSize(hClipdata);
 			buff = malloc(size);
+			if (buff == NULL)
```
akallabeth, line 2136: missing error handling

akallabeth (changes requested):
* missing error handling (e.g. `CloseClipboard` 
* `globlemem`  is also not checked for `NULL`
* I believe the root issue you wanted to fix is a `size == 0` case? that is also not handled properly

chipitsine:
`I believe the root issue you wanted to fix is a size == 0 case? that is also not handled properly` - no, I caught that error from cppcheck. I'll have a look at suggestions as well

akallabeth:
@freerdp-bot test please

akallabeth (approved):

Issue #10006, opened 2024-03-21
FreeRDP does not send relative mouse movements using the advanced input channel


sasha0552:
**Describe the bug**
FreeRDP does not send relative mouse movements using the advanced input channel. It tries to send [relative mouse events](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/9af47964-8bfc-4f43-8a15-4339a89833fa) even if they are not available, but advanced input is available.

**To Reproduce**
1. Connect to the RDP server using FreeRDP with the required parameters.
2. Run a program that handles advanced input channel.

**Expected behavior**
FreeRDP sends relative mouse movements using the advanced input channel, if available.

**Screenshots**
![image](https://github.com/FreeRDP/FreeRDP/assets/45110640/e2790edf-4a7a-427f-b702-e19de9170f1f)
*(I don't want to retake the screenshot, but I also received messages where AINPUT_FLAGS_HAVE_REL (H) was 1, but AINPUT_FLAGS_REL (R) was 0.)*

**Application details**
* FreeRDP version (`xfreerdp /version`)
**This is FreeRDP version 3.4.0-3 (2-3.4.0-3)**
* Command line used
**xfreerdp3 /v:omitted /u:omitted /p:omitted /mouse:relative:on,grab:on +mouse-relative**
* Output of `xfreerdp /buildconfig`
**Below**
* OS version connecting to (server side)
**Microsoft Windows 10 IoT Enterprise LTSC, Version 21H2 (OS Build 19044.4170)**
* If available the log output from a run with `/log-level:trace 2>&1 | tee log.txt` 
**Unnecessary**
* If you built it yourself add some notes which tag/commit/branch you have used, also your cmake parameters and
  compiler can help
**[Tag 3.4.0](https://github.com/FreeRDP/FreeRDP/tree/3.4.0), [CMake parameters](https://gitlab.archlinux.org/archlinux/packaging/packages/freerdp/-/blob/main/PKGBUILD#L75), Compiler: gcc (GCC) 13.2.1 20230801**

<details>
<summary>xfreerdp /buildconfig</summary>

```
This is FreeRDP version 3.4.0-3 (2-3.4.0-3)
[...]

PR #10013, opened 2024-03-23
[client,X11] fix detection of need to use relative movements

sasha0552:
As noted in issue #10006

>`xfreerdp` doesn't send relative mouse movements at all (only absolute; am I missing some other option to capture the mouse?).

I found flaws in the X11 client implementation; it doesn't register raw input handlers because the ainput channel may not be available at the time of connection.
Also, even if it successfully registers a handler, it sends two types of events at the same time: both absolute and relative.
This PR fixes both shortcomings. Relative movements now work correctly in `xfreerdp`, not just in `sdl-freerdp`.

Capture

### DIFF
--- a/client/X11/xf_event.c
+++ b/client/X11/xf_event.c
@@ -450,7 +450,8 @@ static BOOL xf_event_MotionNotify(xfContext* xfc, const XMotionEvent* event, BOO
 {
 	WINPR_ASSERT(xfc);
 
-	if (xfc->xi_event)
+	if (xfc->xi_event ||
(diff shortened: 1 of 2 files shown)


sasha0552:
(I don't know the purpose of `xfc->xi_event`, or `event` methods in general, when `generic` is always (?) used, so I added relmouse checks to `xf_event.c`. If necessary, I can remove these checks, move the relmouse checks in `xf_input.c` to if's, and set `xfc->xi_event` to `TRUE`, like before.)

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #10041, opened 2024-04-03
[winpr,kerberos] use profile_abandon

akallabeth:
suggested by @greghudson to just release the profile and not flush changes back to config file. See #9766


hardening (approved):
Seems ok for me

akallabeth:
@freerdp-bot test properly please?

PR #10055, opened 2024-04-06
[x11,client] always use current clipboard owner

akallabeth:


akallabeth:
confirmed working by @tsz8899

PR #10056, opened 2024-04-06
fix few resource leaks

chipitsine:


hardening:

libfreerdp/utils/http.c
```diff
@@ -126,7 +126,10 @@ BOOL freerdp_http_request(const char* url, const char* body, long* status_code,
 	{
 		blen = strlen(body);
 		if (winpr_asprintf(&headers, &size, post_header_fmt, path, hostname, blen) < 0)
+		{
+			free(hostname);
```
akallabeth, line 130: same `free` missing in else branch

akallabeth (changes requested):
LGTM, just one more free required

libfreerdp/utils/http.c
chipitsine: thanks, added

akallabeth (approved):
LGTM, thank you!

PR #10065, opened 2024-04-11
location: Increase buffer size for printing channel info

mfleisz:
The current buffer size of 32 bytes was too small to hold the version info string which resulted in a non-null terminated buffer (`_snprintf` does not null-terminate the buffer if it is too small) being printed.


akallabeth (approved):
LGTM

PR #10075, opened 2024-04-15
Coverity

akallabeth:


client/common/client_cliprdr_file.c
```diff
@@ -1981,6 +1981,7 @@ BOOL cliprdr_file_context_update_server_data(CliprdrFileContext* file_context, w
 	}
 
 	HashTable_Lock(file_context->inode_table);
+	HashTable_Lock(file_context->clip_data_table);
```
pnowack, line 1984: This looks wrong. The mutex of `file_context->inode_table` is deliberately used for all actions on the `clip_data_table` too. Basically, we need one mutex for most actions, and since the `HashTable` already contains one, that one is used (in particular here: the mutex of the `inode_table`).
So, this code wouldn't add any protection here (since it is already done), but may introduce a deadlock (haven't checked that here in detail, but possible).

akallabeth: @pnowack ok, checked again.
there is no deadlock but should not be required, correct.

akallabeth: removed the commit for the time being.

PR #10092, opened 2024-04-18
[codec,progressive] revert 4e24b966c86fdf494a782f0dfcfc43a057a2ea60

akallabeth:
do not reset progressive surface numUpdatedTiles after progressive_decompress pass. The updates might accumulate until the frameId changes, only then reset.

fixes the long standing graphical glitches when moving windows.


akallabeth:
@freerdp-bot would you please test again ?

PR #10099, opened 2024-04-18
shadow: mention in "help" output that auth defaults to "required"

Hi-Angel:
`freerdp-shadow-cli` requires authorization by default, however the help output was saying:

    -auth (default:off)
        Clients must authenticate

Fix that.

-----------------------

Alternatively, since FreeRDP is very peculiar in how it handles command line options, it's possible that the phrase `-auth (default:off)` was referring to the inverse logic of `-auth` disabling authorization and `off` inverting the disabling, and then the complete `-auth (default:off)` was presumed to imply "enabled". Tell me if that's so, in which case that still needs to be changed because that's very confusing, especially for anyone coming from other cmdline utilities background.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM, thank you!

PR #10114, opened 2024-04-22
[core,redirection] fix length of redirection strings

akallabeth:
length field is in bytes, when converting from UCS-2 use size in WCHAR


akallabeth:
@freerdp-bot test again please!

PR #10144, opened 2024-04-25
Timezone improve

akallabeth:
[winpr,timezone] refactor timezone mapping
    
    * add option to use ICU as a fallback mapping (eliminate need for
      WindowsZones mapping list)
    * extract timezone related settings from localtime_r and eliminate the
      need to parse the complex TimeZones table
    * Add new TimeZoneNameMap to map IANA to windows names (Id, Standard,
      Daylight and Display names)
    * Implement GetDynamicTimeZoneInformation
    * Add fallback mapping from `struct tm` `tm_zone` to `IANA` names if the default detection fails
    * Add new C# tool to generate mappings


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #10168, opened 2024-05-06
[winpr,sspi] skip IP addresses for kerberos auth

akallabeth:
kerberos requires hostnames to authenticate, if we connect with a IP address do not try kerberos based authentication

### DIFF
--- a/winpr/libwinpr/sspi/Kerberos/kerberos.c
+++ b/winpr/libwinpr/sspi/Kerberos/kerberos.c
@@ -39,6 +39,7 @@
 #include <winpr/crypto.h>
 #include <winpr/path.h>
 #include <winpr/wtypes.h>
+#include <winpr/winsock.h>
 
 #include "kerberos.h"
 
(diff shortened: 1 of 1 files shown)


akallabeth:
@giox069 can you test this branch and tell if it works for you?
simple check if we have an `IP` address as hostname and then fail kerberos early, should be sufficient for your case.

giox069:
@akallabeth it works, now connection is successful when I connect via IP address.

But an attempt to contact kerberos server is done even if I used an ip address for /v. I can see it via krb5 KRB5_TRACE

I tried to block outgoing traffic to port 88 of kdc
```
sudo nft flush chain ip filter OUTPUT
sudo nft add rule ip filter OUTPUT udp dport 88 drop
sudo nft add rule ip filter OUTPUT tcp dport 88 drop
```
and I have a 30 seconds delay before connection.

Kerberos is not 100% disabled... in my case I can accept it, and this patch should solve all my problems.
But I think I have a case where I cannot reach a KDC when connecting to a win11 PC... i will test it later.

akallabeth:
@giox069 yes, kerberos is not disabled here (as we only know that the machine name is an IP address later on), but at least it fails gracefully with `NTLM` fallback, which was the intention here.

PR #10172, opened 2024-05-08
Fix incompatible pointer type issues

AlessandroBono:
FreeRDP 2.11.7 doesn't compile in Fedora 40 due to use of `-Werror=incompatible-pointer-types`. This PR fixes it.


akallabeth:
@freerdp-bot test

PR #10178, opened 2024-05-10
Add support for JSON-C library

ondrejholy:
The AAD authentication and SDL prefs currently relies on the cJSON library. This is not available in RHEL for example. Let's add an option to build with the JSON-C library instead.

Related: https://github.com/FreeRDP/FreeRDP/discussions/10009


ondrejholy:
The implementation is pretty straightforward. It includes a test case for SDL prefs, however, the AAD authentication hasn't been tested. @fifthdegree , it seems that you are the author of the AAD support. Is there any chance, that you can test this, please?

fifthdegree:
Sorry, I no longer have the proper setup (or spare time :) to test this.

akallabeth:
@ondrejholy thanks for the pull!
feel free to ignore `clang-tidy` suggestions (they are sometimes useful, but not always as you can see)

cmake/FindJSONC.cmake
```diff
@@ -0,0 +1,22 @@
+# - Try to find JSON-C
+# Once done this will define
+#
+#  JSONC_FOUND - JSON-C was found
+#  JSONC_INCLUDE_DIRS - JSON-C include directories
+#  JSONC_LIBRARIES - libraries needed for linking
+
+find_package(PkgConfig REQUIRED)
```
akallabeth, line 8: this will abort configuration on a system without pkg-config

cmake/FindcJSON.cmake
```diff
@@ -0,0 +1,22 @@
+# - Try to find cJSON
+# Once done this will define
+#
+#  cJSON_FOUND - cJSON was found
+#  CJSON_INCLUDE_DIRS - cJSON include directories
+#  CJSON_LIBRARIES - libraries needed for linking
+
+find_package(PkgConfig REQUIRED)
```
akallabeth, line 8: this will abort detection on a system without pkg-config

include/freerdp/json.h
```diff
@@ -0,0 +1,64 @@
...
+#if CJSON_VERSION_MINOR <= 7
+#if CJSON_VERSION_PATCH < 13
+#define USE_CJSON_COMPAT
+#endif
+#endif
+#endif
+
+#elif defined(WITH_JSONC)
+
+#include <json-c/json.h>
```
akallabeth, line 37: not sure this is an `API` we´d like to ship as our own like this in public headers.
might be better to use some thin wrapper and use our own naming scheme here to have that abstracted.

akallabeth (changes requested):

include/freerdp/json.h
akallabeth: other option would be to just move this to internals like before with `cJSON` so that we do not expose this ourselves. (you need to redo the used symbols for `SDL` client, but well...)

libfreerdp/CMakeLists.txt
```diff
@@ -248,23 +248,15 @@ if(FAAC_FOUND)
...
-    if (NOT CJSON_LIBRARIES OR NOT CJSON_INCLUDE_DIRS)
-        find_path(CJSON_INCLUDE_DIRS
-            NAMES cjson/cJSON.h
-            REQUIRED
-        )
-        find_library(CJSON_LIBRARIES
-            NAMES cjson
-            REQUIRED
-        )
+    if (WITH_JSONC)
```
akallabeth, line 251: this should still search for the library (with `REQUIRED` keyword).
the detection before is used to determine the default value of the `WITH_AAD` option, but here we enforce the requirements.

cmake/FindJSONC.cmake
ondrejholy: Some of the dependencies already require pkg-config, but it is true that they are not built by default, or only on Linux. So I will change that to not require it. Also the `pkg_check_modules` result was not used by mistake, so I will add the missing `HINTS`...

cmake/FindcJSON.cmake
ondrejholy: I will change it here as well. I will also replace `CJSON_LIBRARIES` with `cJSON_LIBRARIES` as I have just found out that there is a bug on the master branch. The code currently relies on pkg-config results that use  `CJSON` prefix and ignores results from `find_library`. The same applies for `find_path`.

include/freerdp/json.h
ondrejholy: Well, I expected that you wouldn't like this approach. I am just preparing a wrapper to avoid the code duplication...

libfreerdp/CMakeLists.txt
ondrejholy: Isn't the `message(FATAL_ERROR "AAD support requested but cJSON not found")` enough?

Issue #9791, opened 2024-01-21
Please enable WITH_VAAPI=ON in nightly builds


shmerl:
Is there any reason to keep it at `WITH_VAAPI=OFF`? It would be nice to be able to use GPU acceleration.

Though it would be nice to have a configuration option to enable or disable it at runtime in case you expect some instability with it.

akallabeth:
pull requests welcome, but I doubt you will be able to implement something that is not `works on my machine for a tiny subset of connection parameters`

shmerl:
It can be for example some environment variable if you prefer. Something like `FREERDP_USE_VAAPI=true` if  you prefer that over connection paramters.

Other than that, I just tested it (on AMD GPU) and it works fine with VAAPI. So it's sort of  waste not to have an option to use it if it's available. I see even Debian package isn't enabling it and I suspect for the same reason - there is no way to make it configurable at runtime. I can take a look at where this can be made conditional.

shmerl:
OK, I took a look. It has a whole bunch of conditionals with
```
#ifdef WITH_VAAPI
...
```

I can add a boolean flag that would also be checked in addition to ifdefs (for parts that actually do some logic), and that flag would be set depending on some env variable like above or if may be command line parameter like `/vaapi` or something like that.

What do you think?

shmerl:
On a side note - I've been using  a VAAPI enabled rebuilt Debian package without any issues so far.

akallabeth:
@shmerl yes, it works for certain hardware combinations and connection settings (that is the reason we have it in there) but not for others and there is no easy way to detect which ones work and which do not.
so, we´re happy about pull requests to make this a runtime option, but definitely not to be enabled by default.

shmerl:
Btw, do you prefer to remove VAAPI related ifdefs and simply have a runtime check, or you want to keep compile time exclusion with ifdef and add a runtime check on top for case when compile time was enabled? It feels somewhat messy to have all those ifdefs with runtime checks on top.

akallabeth:
@shmerl keep the `#ifdef` for now (we don´t want to enable stuff not tested at all in stable releases) but if found stable we can remove these (so runtime checks on top)

shmerl:
I made an initial PR, just enabling the /vaapi switch in the initial commit. It's not being used yet (just in case I got something wrong before proceeding further):

!10181.

shmerl:
[...]

PR #10181, opened 2024-05-12
Support enabling VAAPI at runtime

shmerl:
Create a runtime switch to enable VAAPI optionally (see #9791).

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -5214,6 +5214,13 @@ static int freerdp_client_settings_parse_command_line_arguments_int(
 			if (rc != 0)
 				return rc;
 		}
+#ifdef WITH_VAAPI
+		CommandLineSwitchCase(arg, "vaapi")
+		{
+			if (!freerdp_settings_set_bool(settings, FreeRDP_Vaapi, enable))
+				return COMMAND_LINE_ERROR;
+		}
+#endif
 		CommandLineSwitchDefault(arg)
 		{
 			if (handle_option)
--- a/client/common/file.c
+++ b/client/common/file.c
@@ -108,6 +108,9 @@ struct rdp_file
 	DWORD EncodeRedirectedVideoCapture;          /* encode redirected video capture */
 	DWORD RedirectedVideoCaptureEncodingQuality; /* redirected video capture encoding quality */
 	DWORD VideoPlaybackMode;                     /* videoplaybackmode */
+#ifdef WITH_VAAPI
+	DWORD Vaapi;                                 /* enable VA-API */
+#endif
 
 	DWORD ConnectionType; /* connection type */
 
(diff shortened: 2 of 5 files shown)


client/common/cmdline.c
```diff
@@ -5214,6 +5214,13 @@ static int freerdp_client_settings_parse_command_line_arguments_int(
 			if (rc != 0)
 				return rc;
 		}
+#ifdef WITH_VAAPI
```
akallabeth, line 5217: we already have `/gdi:hw` for hardware acceleration boolean, no need for this if you don´t require specific strings to be passed.

client/common/file.c
```diff
@@ -334,6 +337,9 @@ static const char key_int_screen_mode_id[] = "screen mode id";
 static const char key_int_singlemoninwindowedmode[] = "singlemoninwindowedmode";
 static const char key_int_maximizetocurrentdisplays[] = "maximizetocurrentdisplays";
 static const char key_int_use_multimon[] = "use multimon";
+#ifdef WITH_VAAPI
```
akallabeth, line 340: not something to be passed in a `RDP` file.

libfreerdp/common/settings_str.h
```diff
@@ -261,6 +261,9 @@ static const struct settings_str_entry settings_map[] = {
 	{ FreeRDP_UseMultimon, FREERDP_SETTINGS_TYPE_BOOL, "FreeRDP_UseMultimon" },
 	{ FreeRDP_UseRdpSecurityLayer, FREERDP_SETTINGS_TYPE_BOOL, "FreeRDP_UseRdpSecurityLayer" },
 	{ FreeRDP_UsingSavedCredentials, FREERDP_SETTINGS_TYPE_BOOL, "FreeRDP_UsingSavedCredentials" },
+#ifdef WITH_VAAPI
```
akallabeth, line 264: file is autogenerated, do not edit manually

libfreerdp/core/test/settings_property_lists.h
```diff
@@ -190,6 +190,9 @@ static const size_t bool_list_indices[] = {
 	FreeRDP_UseMultimon,
 	FreeRDP_UseRdpSecurityLayer,
 	FreeRDP_UsingSavedCredentials,
+#ifdef WITH_VAAPI
```
akallabeth, line 193: file is auto generated, do not edit

akallabeth:
better just use `FreeRDP_SoftwareGdi` to check for enable/disable at runtime for now.
more finegrained options can be added later once there is something worth enabling.

shmerl:
Thanks, I'll scrap this one then and use `FreeRDP_SoftwareGdi` / `/gdi:hw` in a new one.

shmerl:
OK, I made a change and tried testing it, but I don't see it kicking in even with `/gdi:hw` set. Must be I'm missing something or I made some mistake. Let me open a PR for review anyway and I'll try to debug it a bit later when I have a chance.

PR #10183, opened 2024-05-14
Jsonc

akallabeth:
Replaces #10178


akallabeth:
@ondrejholy did take your pr and moved the stuff to `WinPR` so we don´t require to know which json library was linked.

akallabeth:
@ondrejholy ok, did implement the generic `JSON` wrapper `API` (mostly only the functions we currently use)
would be nice if you could test this and readd the tests from your pull request (did not take these parts, only renamed the functions we use)

ondrejholy:
> @ondrejholy did take your pr and moved the stuff to `WinPR` so we don´t require to know which json library was linked.

Thanks for the help with the wrapper. I'm just sorry that I didn't know you were working on it, because I spent some time unnecessarily yesterday preparing my own wrapper (which I announced at https://github.com/FreeRDP/FreeRDP/pull/10178#discussion_r1599992749, by the way). Can you please announce your intentions in advance next time?

> @ondrejholy ok, did implement the generic `JSON` wrapper `API` (mostly only the functions we currently use) would be nice if you could test this and readd the tests from your pull request (did not take these parts, only renamed the functions we use)

Ok, I'm going to take a look at your changes and redo the tests...

akallabeth:
@ondrejholy sorry, did read your post too late myself :/

akallabeth:
must fix the API stuff (wrong detection) then this is ready from my side, tests would be a great addition.

ondrejholy:
> must fix the API stuff (wrong detection) then this is ready from my side, tests would be a great addition.

Ok, let me know once you are finished...

akallabeth:
@ondrejholy ok, should be finished, if the `abi-checker` succeeds then the detection is now working.

ondrejholy:
So I cherry-picked a few more commits from the original PR, readded the unit test, and made some fixes in your changes. The test seems to work fine with both the cJSON and JSON-C dep. See https://github.com/ondrejholy/FreeRDP/tree/wip/oholy/akallabeth-jsonc. Will you rebase this PR onto my branch, or should I open PR for your fork, or here?

akallabeth:
@ondrejholy did cherry-pick your commits to this pr.

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test please

akallabeth:
@ondrejholy ok, did have to remove the SDL tests:
1. `WINPR_ASSERT` might not be defined, better use error return
2. the config file can not be overridden, the default of the user running the test is used.

ondrejholy:
Hmm, right. It is not a problem to replace `WINPR_ASSERT` with the `if-return` statements.

The config path is overridden over the `XDG_CONFIG_HOME` envar, see: https://github.com/ondrejholy/FreeRDP/blob/af592294cd4fda51468331e1af4e7667890525b5/client/SDL/test/CMakeLists.txt#L25. The `ctest --test-dir build -R TestSDLPrefs` works fine. It doesn't work when you manually start the test binary without setting `XDG_CONFIG_HOME` though. Maybe the test can verify that `XDG_CONFIG_HOME` is properly set, or set that environment variable itself, what do you think?

But since we have the JSON wrapper now, it should not be hard to add some generic tests for the wrapper itself... would you like me to prototype something?

akallabeth:
> Hmm, right. It is not a problem to replace `WINPR_ASSERT` with the `if-return` statements.
> 
> The config path is overridden over the `XDG_CONFIG_HOME` envar, see: https://github.com/ondrejholy/FreeRDP/blob/af592294cd4fda51468331e1af4e7667890525b5/client/SDL/test/CMakeLists.txt#L25. The `ctest --test-dir build -R TestSDLPrefs` works fine. It doesn't work when you manually start the test binary without setting `XDG_CONFIG_HOME` though. Maybe the test can verify that `XDG_CONFIG_HOME` is properly set, or set that environment variable itself, what do you think?
> 
> But since we have the JSON wrapper now, it should not be hard to add some generic tests for the wrapper itself... would you like me to prototype something?

I think the whole `sdl_pref_*` must be rewritten.
might be a good idea to create a class and then use an instance() method that uses the default file name can be replaced by some custom stuff while testing?

akallabeth:
I´ll have a look at the rewrite of the `SdlPref`

Issue #9506, opened 2023-10-24
Several error cases of connection


LucasMou:
**Describe the bug**
When i use cmd:`./xfreerdp /v: /u: /p: /monitors:0,1 /multimon /network:auto /audio-mode:0 /sound /microphone +clipboard /cert:ignore +gfx-progressive +gfx-small-cache +gfx-thin-client /gdi:sw /floatbar:sticky:on,default:visible,show:always /drive:hotplug,* /gfx:avc444 /f
`.The following types of connection errors occur.

**Test method**
Connect to the host hold for 10 seconds, wait for 1 second, and continue to connect to the host.

1. 
[INFO][com.freerdp.core] - ERRINFO_GRAPHICS_SUBSYSTEM_FAILED (0x0000112F):The server-side graphics subsystem is in an error state and unable to continue graphics encoding.
[ERROR][com.freerdp.core] - rdp_set_error_info:freerdp_set_last_error_ex ERRINFO_GRAPHICS_SUBSYSTEM_FAILED [0x0001112F]
[ERROR][com.freerdp.core.transport] - BIO_read returned a system error 104: Connection reset by peer
[WARN][com.freerdp.client.common] - Disconnected by server hitting a bug or resource limit [The server-side graphics subsystem is in an error state and unable to continue graphics encoding.]

I noticed that there was an issue(#4717) that mentioned this problem, and I tried to use the solution(https://github.com/RangeeGmbH/FreeRDP/commit/bf9e5ceb4ac0afc97d2d44671ff08d460947b7d9), but it doesn't seem to apply to my situation.

2.
[12:35:17:642] [52758:52780] [ERROR][com.freerdp.channels.rdpdr.client] - channel [RDPDR] received 17219, expected state RDPDR_CHANNEL_STATE_CLIENT_CAPS but have state RDPDR_CHANNEL_STATE_NAME_REQUEST, aborting.
[12:35:17:642] [52758:52780] [ERROR][com.freerdp.channels.rdpdr.client] - channel [RDPDR] received 21328, expected state RDPDR_CHANNEL_STATE_NAME_REQUEST but have state RDPDR_CHANNEL_STATE_INITIAL, aborting.

When this happens, I don't know if it will affect normal use.

3.
[15:26:55:510] [77839:77842] [INFO][com.freerdp.client.x11] - Logon Error Info SESSION_ID [UNKNOWN]

[...]

PR #10200, opened 2024-05-17
Allow PAKID_CORE_CLIENTID_CONFIRM in RDPDR_CHANNEL_STATE_NAME_REQUEST

tomsci:
Occasionally servers can send this sequence which breaks the channel state machine leading to all smartcard operations failing. This patch modifies the state transitions to tolerate the observed sequence of operations.

In my testing, this fixes the issue I mentioned in https://github.com/FreeRDP/FreeRDP/issues/9506#issuecomment-2112745653 - but it's maybe not the best way to fix the problem. For example it's possible that adding RDPDR_CHANNEL_STATE_NAME_REQUEST to the allowable states in PAKID_CORE_CLIENTID_CONFIRM breaks something else I didn't see in my testing - I'm not familiar enough with this codebase or protocol to say for sure. So consider this PR a suggestion and a chance for discussion rather than a perfect solution :)

### DIFF
--- a/channels/rdpdr/client/rdpdr_capabilities.c
+++ b/channels/rdpdr/client/rdpdr_capabilities.c
@@ -179,7 +179,28 @@ UINT rdpdr_process_capability_request(rdpdrPlugin* rdpdr, wStream* s)
 		return CHANNEL_RC_NULL_DATA;
 
 	WINPR_ASSERT(rdpdr->state == RDPDR_CHANNEL_STATE_SERVER_CAPS);
-	rdpdr_state_advance(rdpdr, RDPDR_CHANNEL_STATE_CLIENT_CAPS);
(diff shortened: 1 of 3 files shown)


akallabeth (changes requested):
looks like a works for me patch.
1. will break server side (proxy) rdpdr
2. I´d like to know a few more details about your setup (you attached some to a already closed issue), specifically if the systems are stock or if there is some additional software installed.
3. what would be awesome a session recording of such a situation (that is if you can do this for a test user, the recording will contain the whole session data and desktop) with `xfreerdp /dump:record,file:somefile`

akallabeth:
should be fixed with https://github.com/FreeRDP/FreeRDP/releases/tag/3.7.0

PR #10202, opened 2024-05-20
[core,caps] Allow invalid TS_GENERAL_CAPABILITYSET::protocolVersion

akallabeth:
some FreeRDP versions did send an invalid value of 0x0000 instead of the required 0x200. Log this kind of violation but continue.


akallabeth:
@freerdp-bot test

akallabeth:

akallabeth:
@freerdp-bot test please

hardening (approved):
Looks ok for me, in `rdp_write_general_capability_set` we should also ignore the value from the settings and always send `TS_CAPS_PROTOCOLVERSION`.

akallabeth:
@freerdp-bot test
@hardening hmm, you mean removing the setting?

hardening:
> @hardening hmm, you mean removing the setting?

I guess that at end that what we will do. But in the meantime that would prevent behaviors like in proxies where you have front settings that you copy as is for the back connection. And so if an old FreeRDP connects front you will fail to connect back because of the check in the write function...

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test please

PR #10206, opened 2024-05-22
[core] fix clientInfo with both dynamicTimeZone and reconnect cookie

hardening:
The ClientInfo packet was wrongly forged when we had dynamic timezone and a reconnect cookie (reserved fields written twice).

### DIFF
--- a/libfreerdp/core/info.c
+++ b/libfreerdp/core/info.c
@@ -545,8 +545,8 @@ static BOOL rdp_write_extended_info_packet(rdpRdp* rdp, wStream* s)
 	if (!Stream_EnsureRemainingCapacity(s, 10ull))
 		goto fail;
 
-	Stream_Write_UINT32(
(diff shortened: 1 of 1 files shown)


libfreerdp/core/info.c
```diff
@@ -560,17 +560,15 @@ static BOOL rdp_write_extended_info_packet(rdpRdp* rdp, wStream* s)
 
 		if (!Stream_EnsureRemainingCapacity(s, 4ull))
```
akallabeth, line 561: sure this check is still required?

akallabeth (approved):
LGTM,
the capacity check might not be required anymore though.

libfreerdp/core/info.c
hardening: Correct I've removed it

akallabeth (approved):
LGTM

PR #10208, opened 2024-05-23
Image copy improve, bugfixes

akallabeth:
* expose (no)overlap versions of `freerdp_image_copy`
* use dedicated `freerdp_image_copy_no_overlap` where that is possible
* fix a bug with multitransport capability reading in server code
* log TLS handshake errors


akallabeth:
@freerdp-bot test

PR #10213, opened 2024-05-24
Thread prio

akallabeth:


akallabeth:
@freerdp-bot test please

winpr/libwinpr/thread/thread.c
```diff
@@ -605,6 +614,54 @@ static BOOL winpr_StartThread(WINPR_THREAD* thread)
 	return rc;
 }
 
+BOOL SetThreadPriority(HANDLE hThread, int nPriority)
+{
+	ULONG Type = 0;
+	WINPR_HANDLE* Object = NULL;
+
+	if (!winpr_Handle_GetInfo(hThread, &Type, &Object))
```
hardening, line 622: here we should also check that the object is a thread

winpr/libwinpr/thread/thread.c
```diff
@@ -28,6 +28,14 @@
 
 #include <winpr/thread.h>
 
+#ifndef MIN
```
hardening, line 31: not your fault, but we 13 definitions of `MIN` in FreeRDP and 5 for `MAX`...

akallabeth: hmm, we already got a central definition for `freerdp` but lacking one for `winpr`

akallabeth:

PR #10222, opened 2024-05-27
Color inline

akallabeth:


hardening (approved):
Looks ok for me.
Just curious, do you have some before / after numbers ?

akallabeth:
@hardening just run the commit with the unit test on this branch and then on current master.

akallabeth:
with converting copy it is ~10-15%

PR #10225, opened 2024-05-29
[core,codecs] fix API function names

akallabeth:
Update function names to conform to rest of header.

* deprecate codecs_new and codecs_free
* add new freerdp_client_codecs_new and freerdp_client_codecs_free


hardening (approved):
Seems ok.

PR #10226, opened 2024-05-30
Ignore empty KDC URL when initializing client attributes

luckyluke:
For example, on a rdp file generated on win10 the entry is present but empty, even with simple user/pass authentication.

Without this change, authentication fails as it tries to initialize the kerberos configuration with the empty URL.


hardening:

libfreerdp/core/credssp_auth.c
```diff
@@ -205,7 +205,12 @@ static BOOL credssp_auth_client_init_cred_attributes(rdpCredsspAuth* auth)
 		ULONG buffer_size = 0;
 
 		str_size = ConvertUtf8ToWChar(auth->kerberosSettings.kdcUrl, NULL, 0);
-		if (str_size <= 0)
+		if (str_size == 0)
```
akallabeth, line 208: better fix the `if (auth->kerberosSettings.kdcUrl)` above?

PR #10227, opened 2024-05-30
SDL2: sdl_kbd: parse keyboard config once instead of at each keypress

mjt0k:
Currently, sdl-freerdp.json file is parsed at every keypress, which is just wrong.  Instead, parse it at startup and remember the key configuration for the whole session.

The same should be done for SDL3 I guess.

### DIFF
--- a/client/SDL/SDL2/sdl_kbd.cpp
+++ b/client/SDL/SDL2/sdl_kbd.cpp
@@ -555,27 +555,27 @@ BOOL sdlInput::keyboard_handle_event(const SDL_KeyboardEvent* ev)
 	const UINT32 rdp_scancode = sdl_scancode_to_rdp(ev->keysym.scancode);
 	const SDL_Keymod mods = SDL_GetModState();
 
-	if ((mods & hotkey_modmask) == hotkey_modmask)
+	if ((mods & _hotkeyModmask) == _hotkeyModmask)
 	{
 		if (ev->type == SDL_KEYDOWN)
 		{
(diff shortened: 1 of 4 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test please

mjt0k:
I runtime-tested it locally.  Unfortunately I can't test the same change with SDL3 right now (don't have SDL3 installed, will have to figure that out first)

akallabeth:
@mjt0k constructor is fine here as there is no exception case (config can be read or not, if not fallback settings are used)

akallabeth:
for sdl3 look at https://github.com/FreeRDP/FreeRDP/wiki/Compilation
install `SDL`, `SDL_ttf` and `SDL_image` from git to some location (`/tmp/testinstall`) and then point the `FreeRDP` compile to it `-DCMAKE_INSTALL_PREFIX=/tmp/testinstall` and maybe `CMAKE_PREFX_PATH` as well

mjt0k:
no, I mean, the constructor might be executed way earlier than everything else is initialized - like $HOME, $XDG_CONFIG_PATH - that stuff.  It should be initialized maybe at the beginning of connection, not when a variable (or class member) of type SdlInput is declared :)

akallabeth:
@mjt0k no, that stuff is already set up when the application starts (and only after that any constructor is executed)

akallabeth:
@freerdp-bot test please

PR #10243, opened 2024-06-04
Release 3.6.0

akallabeth:


Fantu:
Hi, why MS-RDPECAM support is not in noteworthy changes in the changelog?

akallabeth:
@Fantu because I missed that one ;)
would have been nice to get notified before the merge ;)

well, still need to finish release tests, so can add a note to that in release mail/github release page

PR #10261, opened 2024-06-06
Rdstls

akallabeth:


libfreerdp/core/gateway/arm.c
```diff
@@ -594,9 +594,93 @@ static BOOL arm_pick_base64Utf16Field(const WINPR_JSON* json, const char* name,
...
+	{
+		WINPR_JSON* adressN = WINPR_JSON_GetArrayItem(ipAddress, j);
+		if (!adressN)
+			continue;
+
+		WINPR_JSON* publicIpNode = WINPR_JSON_GetObjectItem(adressN, "publicIpAddress");
+		if (publicIpNode && WINPR_JSON_IsString(publicIpNode))
+		{
+			const char* publicIp = WINPR_JSON_GetStringValue(publicIpNode);
+			if (publicIp)
```
joe-koral-imprivata, line 652: This results in publicIp=="" for 

```
"azureInstanceNetworkMetadata": "{\"interface\":[{\"ipv4\":{\"ipAddress\":[{\"privateIpAddress\":\"10.234.60.7\",\"publicIpAddress\":\"\"}],\"subnet\":[{\"address\":\"10.234.60.0\",\"prefix\":\"24\"}]},\"ipv6\":{\"ipAddress\":[]},\"macAddress\":\"6045BDD6FD58\"}]}"
```

which results in `freerdp_settings_set_pointer_array` failing for the privateIp on line 666 because `offset >= maxOffset` is true.

akallabeth: ok, pushed a fix for empty strings

akallabeth:
@joe-koral-imprivata ok, fixed a few issues detected from the log.
still, this looks like the same error code (something not ok with credentials or similar), but that would require some server side logging to find out what it is.

as for linux code stepping: check out `qt-creator`, `nemiver` or any other `gdb/lldb` frontend ;)

akallabeth:
@joe-koral-imprivata hmm, looking at it, maybe they filter the user agent?
we set it in `arm_handle_request` so maybe try changing that to something `mstsc` or `webview2/windows` are sending?

akallabeth:
@joe-koral-imprivata also, a dump of a request JSON from a working client could help (just the keys, not the values ;))

akallabeth:
merging this as it improves the situation (although not fixing for the reporter)
@joe-koral-imprivata if you find out any more details, please open an issue with what you discovered, I´ll address that then.

PR #10264, opened 2024-06-07
[winpr,pool] fix winpr_SetThreadpoolThreadMaximum

akallabeth:
unlock the arraylist, don´t lock it twice.


akallabeth:

PR #10265, opened 2024-06-07
Base64 inline restrict

akallabeth:


akallabeth:
@freerdp-bot test please

PR #10280, opened 2024-06-17
[channel,rdpecam] add option RDPECAM_CLIENT_CHANNEL_STUB

akallabeth:
* this allows building FreeRDP with only a stubbed [MS-RDPECAM] channel
* also revert the default to `OFF` (we are in a stable release, so the user of the library must opt in to enable this)


hardening (approved):

mfleisz (approved):

channels/rdpecam/ChannelOptions.cmake
```diff
@@ -1,11 +1,7 @@
 
 set(OPTION_DEFAULT ON)
-set(OPTION_SERVER_DEFAULT ON)
-
 set(OPTION_CLIENT_DEFAULT OFF)
-if (NOT IOS AND NOT ANDROID AND CMAKE_SYSTEM_NAME STREQUAL "Linux")
-    set(OPTION_CLIENT_DEFAULT ON)
-endif()
+set(OPTION_SERVER_DEFAULT ON)
```
oleg0421, line 5: @akallabeth : I agree this is still experimental feature and should be off by default. However, with this change in `ChannelOptions.cmake`, there's no way to enable it other than modifying this file by setting:

```
set(OPTION_CLIENT_DEFAULT ON)
```
I think changes in this file are not needed. Instead, `channels/rdpecam/CMakeLists.txt` should have:
```
option(RDPECAM_CLIENT_CHANNEL_STUB "Only build [MS-RDPECAM] channel stub" ON)
```
If it's ON by default, then stub will be in place and no real code.
If built with `-DRDPECAM_CLIENT_CHANNEL_STUB=OFF`, then real code will be included.

akallabeth: @oleg0421 no, you can just set `-DCHANNEL_RDPECAM_CLIENT=ON` in your configure line ;)

PR #10286, opened 2024-06-18
[tools] improve smartcard traffic interpreter

hardening:
This patch adds interpretations for more commands and adds GPC application.


akallabeth (approved):
LGTM

PR #10292, opened 2024-06-20
[channel,rdpecam] MJPEG input format support

oleg0421:
Adding support for MJPEG camera format, using FFMPEG library decoder.

Higher end web cameras, supporting 720p and higher resolutions at 30 fps, are using MJPEG format. This allows to overcome USB 2.0 throughput limitation, that would otherwise be encountered with uncompressed formats, such as YUYV. Even laptop cameras typically use USB 2.0 bus internally, so MJPEG support is required for 720p and 1080p at 30 fps.

To check, if camera supports MJPEG format:
```
v4l2-ctl --list-formats --device=/dev/video0
```

Cc: @akallabeth, @hardening

### DIFF
--- a/channels/rdpecam/client/camera_device_main.c
+++ b/channels/rdpecam/client/camera_device_main.c
@@ -24,11 +24,16 @@
 #define TAG CHANNELS_TAG("rdpecam-device.client")
 
 /* supported formats in preference order:
- * passthrough, I420 (used as input for H264 encoder), other YUV based, RGB based
+ * H264, MJPG, I420 (used as input for H264 encoder), other YUV based, RGB based
  */
 static const CAM_MEDIA_FORMAT_INFO supportedFormats[] = {
-	/* inputFormat, outputFormat */
-	{ CAM_MEDIA_FORMAT_H264, CAM_MEDIA_FORMAT_H264 }, /* passthrough: comment out to disable */
+/* inputFormat, outputFormat */
+#if defined(WITH_INPUT_FORMAT_H264)
+	{ CAM_MEDIA_FORMAT_H264, CAM_MEDIA_FORMAT_H264 }, /* passthrough */
+#endif
+#if defined(WITH_INPUT_FORMAT_MJPG)
+	{ CAM_MEDIA_FORMAT_MJPG, CAM_MEDIA_FORMAT_H264 },
+#endif
 	{ CAM_MEDIA_FORMAT_I420, CAM_MEDIA_FORMAT_H264 },
 	{ CAM_MEDIA_FORMAT_YUY2, CAM_MEDIA_FORMAT_H264 },
 	{ CAM_MEDIA_FORMAT_NV12, CAM_MEDIA_FORMAT_H264 },
--- a/channels/rdpecam/client/encoding.c
+++ b/channels/rdpecam/client/encoding.c
@@ -88,6 +88,60 @@ static enum AVPixelFormat ecamToAVPixFormat(CAM_MEDIA_FORMAT ecamFormat)
 
 /**
  * Function description
+ * initialize libswscale
+ *
+ * @return success/failure
+ */
+static BOOL ecam_init_sws_context(CameraDeviceStream* stream, enum AVPixelFormat pixFormat)
+{
+	WINPR_ASSERT(stream);
+
+	if (stream->sws)
+		return TRUE;
+
+	/* replacing deprecated JPEG formats, still produced by decoder */
+	switch (pixFormat)
+	{
+		case AV_PIX_FMT_YUVJ411P:
+			pixFormat = AV_PIX_FMT_YUV411P;
+			break;
+
+		case AV_PIX_FMT_YUVJ420P:
+			pixFormat = AV_PIX_FMT_YUV420P;
+			break;
+
+		case AV_PIX_FMT_YUVJ422P:
+			pixFormat = AV_PIX_FMT_YUV422P;
+			break;
+
+		case AV_PIX_FMT_YUVJ440P:
+			pixFormat = AV_PIX_FMT_YUV440P;
+			break;
+
+		case AV_PIX_FMT_YUVJ444P:
+			pixFormat = AV_PIX_FMT_YUV444P;
+			break;
+
+		default:
+			break;
+	}
+
+	const int width = (int)stream->currMediaType.Width;
+	const int height = (int)stream->currMediaType.Height;
+
+	stream->sws = sws_getContext(width, height, pixFormat, width, height, AV_PIX_FMT_YUV420P, 0,
+	                             NULL, NULL, NULL);
+	if (!stream->sws)
+	{
+		WLog_ERR(TAG, "sws_getContext failed");
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+/**
+ * Function description
  *
  * @return success/failure
  */
@@ -96,27 +150,70 @@ static BOOL ecam_encoder_compress_h264(CameraDeviceStream* stream, const BYTE* s
 {
 	UINT32 dstSize = 0;
 	BYTE* srcSlice[4] = { 0 };
+	int srcLineSizes[4] = { 0 };
 	BYTE* yuv420pData[3] = { 0 };
 	UINT32 yuv420pStride[3] = { 0 };
-	prim_size_t size = { 0 };
-	size.width = stream->currMediaType.Width;
-	size.height = stream->currMediaType.Height;
+	prim_size_t size = { stream->currMediaType.Width, stream->currMediaType.Height };
 	CAM_MEDIA_FORMAT inputFormat = streamInputFormat(stream);
-	enum AVPixelFormat pixFormat = ecamToAVPixFormat(inputFormat);
+	enum AVPixelFormat pixFormat = AV_PIX_FMT_NONE;
+
+#if defined(WITH_INPUT_FORMAT_MJPG)
+	if (inputFormat == CAM_MEDIA_FORMAT_MJPG)
+	{
+		stream->avInputPkt->data = (BYTE*)srcData;
+		stream->avInputPkt->size = srcSize;
+
+		if (avcodec_send_packet(stream->avContext, stream->avInputPkt) < 0)
+		{
+			WLog_ERR(TAG, "avcodec_send_packet failed");
+			return FALSE;
+		}
+
+		if (avcodec_receive_frame(stream->avContext, stream->avOutFrame) < 0)
+		{
+			WLog_ERR(TAG, "avcodec_receive_frame failed");
+			return FALSE;
+		}
+
+		for (size_t i = 0; i < 4; i++)
+		{
+			srcSlice[i] = stream->avOutFrame->data[i];
+			srcLineSizes[i] = stream->avOutFrame->linesize[i];
+		}
+
+		/* get pixFormat produced by MJPEG decoder */
+		pixFormat = stream->avContext->pix_fmt;
+	}
+	else
+#endif
+	{
+		pixFormat = ecamToAVPixFormat(inputFormat);
+
+		if (av_image_fill_linesizes(srcLineSizes, pixFormat, (int)size.width) < 0)
+		{
+			WLog_ERR(TAG, "av_image_fill_linesizes failed");
+			return FALSE;
+		}
+
+		if (av_image_fill_pointers(srcSlice, pixFormat, (int)size.height, (BYTE*)srcData,
+		                           srcLineSizes) < 0)
+		{
+			WLog_ERR(TAG, "av_image_fill_pointers failed");
+			return FALSE;
+		}
+	}
 
 	/* get buffers for YUV420P */
-	if (h264_get_yuv_buffer(stream->h264, stream->srcLineSizes[0], size.width, size.height,
-	                        yuv420pData, yuv420pStride) < 0)
+	if (h264_get_yuv_buffer(stream->h264, srcLineSizes[0], size.width, size.height, yuv420pData,
+	                        yuv420pStride) < 0)
 		return FALSE;
 
 	/* convert from source format to YUV420P */
-	BYTE* ptr = WINPR_CAST_CONST_PTR_AWAY(srcData, BYTE*);
-	if (av_image_fill_pointers(srcSlice, pixFormat, (int)size.height, ptr, stream->srcLineSizes) <
-	    0)
+	if (!ecam_init_sws_context(stream, pixFormat))
 		return FALSE;
 
 	const BYTE* cSrcSlice[4] = { srcSlice[0], srcSlice[1], srcSlice[2], srcSlice[3] };
-	if (sws_scale(stream->sws, cSrcSlice, stream->srcLineSizes, 0, (int)size.height, yuv420pData,
+	if (sws_scale(stream->sws, cSrcSlice, srcLineSizes, 0, (int)size.height, yuv420pData,
 	              (int*)yuv420pStride) <= 0)
 		return FALSE;
 
@@ -136,11 +233,27 @@ static BOOL ecam_encoder_compress_h264(CameraDeviceStream* stream, const BYTE* s
 static void ecam_encoder_context_free_h264(CameraDeviceStream* stream)
 {
 	WINPR_ASSERT(stream);
+
 	if (stream->sws)
 	{
 		sws_freeContext(stream->sws);
 		stream->sws = NULL;
 	}
+
+#if defined(WITH_INPUT_FORMAT_MJPG)
+	if (stream->avOutFrame)
+		av_frame_free(&stream->avOutFrame); /* sets to NULL */
+
+	if (stream->avInputPkt)
+	{
+		stream->avInputPkt->data = NULL;
+		stream->avInputPkt->size = 0;
+		av_packet_free(&stream->avInputPkt); /* sets to NULL */
+	}
+
+	if (stream->avContext)
+		avcodec_free_context(&stream->avContext); /* sets to NULL */
+#endif
 
 	if (stream->h264)
 	{
@@ -148,6 +261,57 @@ static void ecam_encoder_context_free_h264(CameraDeviceStream* stream)
 		stream->h264 = NULL;
 	}
 }
+
+#if defined(WITH_INPUT_FORMAT_MJPG)
+/**
+ * Function description
+ *
+ * @return success/failure
+ */
+static BOOL ecam_init_mjpeg_decoder(CameraDeviceStream* stream)
+{
+	WINPR_ASSERT(stream);
+
+	const AVCodec* avcodec = avcodec_find_decoder(AV_CODEC_ID_MJPEG);
+	if (!avcodec)
+	{
+		WLog_ERR(TAG, "avcodec_find_decoder failed to find MJPEG codec");
+		return FALSE;
+	}
+
+	stream->avContext = avcodec_alloc_context3(avcodec);
+	if (!stream->avContext)
+	{
+		WLog_ERR(TAG, "avcodec_alloc_context3 failed");
+		return FALSE;
+	}
+
+	stream->avContext->width = stream->currMediaType.Width;
+	stream->avContext->height = stream->currMediaType.Height;
+
+	if (avcodec_open2(stream->avContext, avcodec, NULL) < 0)
+	{
+		WLog_ERR(TAG, "avcodec_open2 failed");
+		return FALSE;
+	}
+
+	stream->avInputPkt = av_packet_alloc();
+	if (!stream->avInputPkt)
+	{
+		WLog_ERR(TAG, "av_packet_alloc failed");
+		return FALSE;
+	}
+
+	stream->avOutFrame = av_frame_alloc();
+	if (!stream->avOutFrame)
+	{
+		WLog_ERR(TAG, "av_frame_alloc failed");
+		return FALSE;
+	}
+
+	return TRUE;
+}
+#endif
 
 /**
  * Function description
@@ -157,6 +321,7 @@ static void ecam_encoder_context_free_h264(CameraDeviceStream* stream)
 static BOOL ecam_encoder_context_init_h264(CameraDeviceStream* stream)
 {
 	WINPR_ASSERT(stream);
+
 	if (!stream->h264)
 		stream->h264 = h264_context_new(TRUE);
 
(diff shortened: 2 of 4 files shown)


hardening:
My webcam can do Mjeg and it produces 
```
[07:12:46:312] [226221:000373ca] [INFO][com.freerdp.channels.rdpecam-v4l.client] - [cam_v4l_stream_start]: Camera format: MJPG, width: 640, height: 480, fps: 30/1
[mjpeg @ 0x794eeb741900] mjpeg_decode_dc: bad vlc: 0:0 (0x794eeb741f90)
[mjpeg @ 0x794eeb741900] error dc
[mjpeg @ 0x794eeb741900] error y=54 x=1
[libx264 @ 0x794ec4205f80] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2 AVX
[libx264 @ 0x794ec4205f80] profile High, level 3.0, 4:2:0, 8-bit
[libx264 @ 0x794ec4205f80] frame I:2     Avg QP:24,99  size: 21742
[libx264 @ 0x794ec4205f80] frame P:279   Avg QP:24,11  size:  2724
[libx264 @ 0x794ec4205f80] mb I  I16..4: 17,0% 62,5% 20,6%
[libx264 @ 0x794ec4205f80] mb P  I16..4:  0,2%  0,6%  0,0%  P16..4: 34,3%  7,7%  6,4%  0,0%  0,0%    skip:50,9%
[libx264 @ 0x794ec4205f80] final ratefactor: 22,16
[libx264 @ 0x794ec4205f80] 8x8 transform intra:68,7% inter:76,1%
[libx264 @ 0x794ec4205f80] coded y,uvDC,uvAC intra: 58,7% 64,8% 24,8% inter: 14,1% 24,0% 0,1%
[libx264 @ 0x794ec4205f80] i16 v,h,dc,p: 33% 35% 13% 19%
[libx264 @ 0x794ec4205f80] i8 v,h,dc,ddl,ddr,vr,hd,vl,hu: 20% 15% 32%  6%  4%  4%  5%  7%  7%
[libx264 @ 0x794ec4205f80] i4 v,h,dc,ddl,ddr,vr,hd,vl,hu: 29% 12% 16%  8%  6%  8%  5% 13%  5%
[libx264 @ 0x794ec4205f80] i8c dc,h,v,p: 62% 17% 16%  4%
[libx264 @ 0x794ec4205f80] Weighted P-Frames: Y:1,4% UV:1,4%
[libx264 @ 0x794ec4205f80] ref P L0: 55,2%  9,5% 22,5% 12,7%  0,1%
[libx264 @ 0x794ec4205f80] kb/s:686,28
```
when opening webcamtest, but it doesn't work (no more success with youcam)

channels/rdpecam/client/camera_device_main.c
```diff
@@ -24,11 +24,14 @@
 #define TAG CHANNELS_TAG("rdpecam-device.client")
 
 /* supported formats in preference order:
- * passthrough, I420 (used as input for H264 encoder), other YUV based, RGB based
+ * H264, MJPG, I420 (used as input for H264 encoder), other YUV based, RGB based
+ *
+ * comment out any line to disable specific format for testing
  */
 static const CAM_MEDIA_FORMAT_INFO supportedFormats[] = {
```
hardening, line 31: Just wondering the rational of that array, as with any format we end up with doing h264 at the end...

oleg0421:
@hardening:

> [mjpeg @ 0x794eeb741900] mjpeg_decode_dc: bad vlc: 0:0 (0x794eeb741f90)
> [mjpeg @ 0x794eeb741900] error dc
> [mjpeg @ 0x794eeb741900] error y=54 x=1

This means error decoding JPEG frame :( There are few hints on Internet about this error:

https://obsproject.com/forum/threads/occasional-blinking-warning-mjpeg_decode_dc-bad-vlc-0-0-in-log.157426
https://obsproject.com/forum/threads/mjpeg-decode-error-leads-to-stream-disconnecting-from-twitch.125851
https://www.reddit.com/r/obs/comments/n8ghlp/active_usb_extension_cables_for_usb_webcams
https://www.mail-archive.com/bugs@openbsd.org/msg15026.html

The summary: USB protocol errors, that result in corrupted packets, or possibly problematic camera firmware.

My 2 cameras that support MJPEG don't show this problem.

I don't insist on this change going in, in case MJPEG format is known to cause these issues. Or we can commit the code with
```
	{ CAM_MEDIA_FORMAT_MJPG, CAM_MEDIA_FORMAT_H264 },
```
line commented out or under some `#ifdef`.

channels/rdpecam/client/camera_device_main.c
oleg0421: The `outputFormat` column in this array serves two purposes:

1. If `inputFormat` equals `outputFormat`, we do the pass through (no encoding):
https://github.com/FreeRDP/FreeRDP/blob/78044ab1b44c49774749fcc7d29b0fb9ae5a6b4b/channels/rdpecam/client/camera_device_main.c#L142
2. The MS-RDPECAM spec doesn't limit the output format to H264 only. For practical purposes (bandwidth), we keep it like that, but in theory we can have:
```
	{ CAM_MEDIA_FORMAT_MJPG, CAM_MEDIA_FORMAT_MJPG },
```
in which case we will pass through MJPG. Also, MS may add other high compression codecs in future (H265, etc), so we will be ready for that :)

hardening:
Hum rather strange, I ran `guvcview` and it seems like capture works fine with MJPEG

oleg0421:
@hardening : `guvcview` is using its own MJPEG decoder. We are using FFMPEG library decoder, which may be more sensitive to USB errors. Could you please provide more details:

- your camera make/model;
- is it USB external camera, are you using USB hub, long cable or anything special;
- is the error always the same, i.e. `error y=54 x=1` or you see different `y` and `x` values every time? Can you share a  log with multiple errors please?

If you share more details, I can try to find the same camera/model and hopefully reproduce it. Thanks!

hardening:
@oleg0421 so my cam is a `Logitech, Inc. HD Pro Webcam C920` (`046d:08e5`), connected on the monitor's USB hub. It works perfectly with guvcview, without any decoding errors of MJPEG (I mean I don't have any long cable).

The error position varies (30 < y < 54).

Debug logs: https://gist.github.com/hardening/eaf129889c90a13ea32aff9503bd7982 (running webcamtest site and then youcam)

In guvcview you're right they have their own JPEG decoder, I noticed that they check that the MPEG frame is at least `0xaf`, and except if I read the code wrongly when decoding fails it's the previous frame that is shown. Anyway I've not seen any decoding failure error.

oleg0421:
@hardening : thank you for sharing additional information!

Somehow, the log doesn't show any ffmpeg library output (no mjpeg_decode_dc: bad vlc: 0:0 error), but I can see many "Skip sample: no credits left" messages. Let me try to sort this out:

- "Skip sample: no credits left" could mean you hit congestion in your network uplink. In the latest patch that went in, we optimized H264 bitrates, but you still need to have at least 1 - 2 Mbps uplink for camera redirect to work;

- Regarding `guvcview` approach of showing the last good frame: IMO, masking USB layer 1 - 2 errors is not what you ultimately want to do. If USB setup has a problem, it will impact your video, network or other high bandwidth devices. You'd rather want to troubleshoot and eliminate this problem. Even with raw video formats (where there's no decoder), you'll see distorted video, artifacts, etc.;

- I also have Logitech C920 (earlier revision 046d:082d) and I don't experience this error.

Could you please try to connect your camera directly to USB port on your computer (bypassing monitor's USB hub)?

akallabeth:
@oleg0421 is this pull ready or still stuff to fix?

oleg0421:
> @oleg0421 with direct connection of the cam I get the same result.

@hardening: what do you mean by same result? Could you please share the new log when camera is connected directly? The log with `mjpeg_decode_dc: bad vlc: 0:0` errors.

> Just for tests I've tried skipping the mjpg to h264 transcoding (by setting mjpg -> mjpg in the convertion array), so that native mjpg is sent, but that doesn't change anything (always the no credits message).

To passthrough MJPG, the required uplink bandwidth could be around 50 Mpbs. As mentioned, `no credits left` message indicates uplink saturation.

Let's please not mix these two things up. Slow uplink is one problem.  `mjpeg_decode_dc: bad vlc: 0:0`  is a different problem, which may point to a USB hub with low throughput.

oleg0421:
> @oleg0421 is this pull ready or still stuff to fix?

@akallabeth: don't have enough information to troubleshoot @hardening's problem. Not sure what's wrong with proposed change (if anything). It would be helpful, if somebody else helps with testing (camera needs to support MJPEG format).

In general, adding MJPEG format is essential to support 720p and higher resolutions. But it may also require adding hardware acceleration (VAAPI) support for h264 encoding, because with higher resolutions h264 encoding shows high CPU load.

hardening:
I gave a look at the code and I'm wondering if the pb is that when buffers are submitted to the sampleCallback, a full frame could be composed of multiple v4l buffers and so if we don't have credits, then we could possibly drop/mistreat just a piece of the encoded frame.
So when we're capturing raw traffic that gives artifacts (like before this PR), and when the capture format is more "encoded" (MJPG), that gives "unable to decode in ffmpeg". 
Could it be an explanation @oleg0421 ? Perhaps we need another callback that is triggered when all the buffers have been treated and would be called here https://github.com/FreeRDP/FreeRDP/blob/master/channels/rdpecam/client/v4l/camera_v4l.c#L467 meaning "end of frame"  (and so we would send the sample in that callback ) ?

oleg0421:
@hardening:

> I gave a look at the code and I'm wondering if the pb is that when buffers are submitted to the sampleCallback, a full frame could be composed of multiple v4l buffers

In V4L2 API description:
https://www.kernel.org/doc/html/v4.9/media/uapi/v4l/io.html
it's not mentioned that single video frame can be stored in several buffers. When buffers are allocated, the size for each buffer is big enough to fit in entire frame based on dimensions and pixel format.
The purpose to have multiple buffers is for parallel processing, i.e. one buffer is being processed by the application, while the other is being populated by the driver with next frame. I suspect it would be enough to have 2 buffers for our case (single capture thread), but there are 4 just in case.

> and so if we don't have credits, then we could possibly drop/mistreat just a piece of the encoded frame. So when we're capturing raw traffic that gives artifacts (like before this PR), and when the capture format is more "encoded" (MJPG), that gives "unable to decode in ffmpeg". Could it be an explanation @oleg0421 ?

My feeling is that in your case buffer has incomplete frame (tail cut off) due to USB bandwidth limitation / issue, but the missing part of the frame is not in another buffer, it's just lost. Please correct me if I'm wrong.

> Perhaps we need another callback that is triggered when all the buffers have been treated and would be called here https://github.com/FreeRDP/FreeRDP/blob/master/channels/rdpecam/client/v4l/camera_v4l.c#L467 meaning "end of frame" (and so we would send the sample in that callback ) ?

Definitely, in case V4L2 API would have such a provision (that frame can be stored in multiple buffers), then we would have to reconstruct the frame before giving it to decoder. Maybe I don't see it in the documentation, please let me know.

jens-maus:
@oleg0421 I just tested this PR in my thinclient project (https://github.com/jens-maus/thinRoot) and can say that RDP MJPG camera redirection works with this PR. See here:

```
16:18:35:634] [879:00000388] [INFO][com.freerdp.channels.rdpecam-v4l.client] - [cam_v4l_stream_start]: Camera format: MJPG, width: 640, height: 480, fps: 30/1
[libx264 @ 0x7f06605d6b40] using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2
[libx264 @ 0x7f06605d6b40] profile High, level 3.0, 4:2:0, 8-bit
[libx264 @ 0x7f06605d6b40] frame I:1     Avg QP:24.57  size:  6347
[libx264 @ 0x7f06605d6b40] frame P:191   Avg QP:17.02  size:  2758
[libx264 @ 0x7f06605d6b40] mb I  I16..4: 21.8% 72.1%  6.1%
[libx264 @ 0x7f06605d6b40] mb P  I16..4:  1.4% 12.3%  0.0%  P16..4: 27.6% 10.8%  6.6%  0.0%  0.0%    skip:41.2%
[libx264 @ 0x7f06605d6b40] final ratefactor: 23.26
[libx264 @ 0x7f06605d6b40] 8x8 transform intra:89.1% inter:86.2%
[libx264 @ 0x7f06605d6b40] coded y,uvDC,uvAC intra: 28.2% 26.0% 2.4% inter: 8.2% 15.1% 0.0%
[libx264 @ 0x7f06605d6b40] i16 v,h,dc,p: 25% 41% 20% 14%
[libx264 @ 0x7f06605d6b40] i8 v,h,dc,ddl,ddr,vr,hd,vl,hu: 18% 17% 60%  1%  1%  1%  1%  1%  1%
[libx264 @ 0x7f06605d6b40] i4 v,h,dc,ddl,ddr,vr,hd,vl,hu: 24% 32% 18%  4%  5%  6%  5%  3%  3%
[libx264 @ 0x7f06605d6b40] i8c dc,h,v,p: 63% 19% 17%  2%
[libx264 @ 0x7f06605d6b40] Weighted P-Frames: Y:0.0% UV:0.0%
[libx264 @ 0x7f06605d6b40] ref P L0: 48.3%  3.4% 30.8% 17.5%
[libx264 @ 0x7f06605d6b40] kb/s:666.38
``` 

Thus, I am **not** spotting the kind of `mjpeg_decode_dc` errors @hardening has spotted in his environment.

The camera that I am currently using is:

```
$ v4l2-ctl --info --list-formats
Driver Info:
	Driver name      : uvcvideo
	Card type        : C270 HD WEBCAM
	Bus info         : usb-0000:00:15.0-1
	Driver version   : 6.6.52
	Capabilities     : 0x84a00001
		Video Capture
		Metadata Capture
		Streaming
		Extended Pix Format
		Device Capabilities
	Device Caps      : 0x04200001
		Video Capture
		Streaming
		Extended Pix Format
Media Driver Info:
	Driver name      : uvcvideo
	Model            : C270 HD WEBCAM
	Serial           : 200901010001
	Bus info         : usb-0000:00:15.0-1
	Media version    : 6.6.52
	Hardware revision: 0x00000021 (33)
	Driver version   : 6.6.52
Interface Info:
	ID               : 0x03000002
	Type             : V4L Video
Entity Info:
	ID               : 0x00000001 (1)
	Name             : C270 HD WEBCAM
	Function         : V4L2 I/O
	Flags            : default
	Pad 0x01000007   : 0: Sink
	  Link 0x02000016: from remote pad 0x100000a of entity 'Extension 8' (Video Pixel Formatter): Data, Enabled, Immutable
ioctl: VIDIOC_ENUM_FMT
	Type: Video Capture

	[0]: 'YUYV' (YUYV 4:2:2)
	[1]: 'MJPG' (Motion-JPEG, compressed)
```

Looking forward to seeing this PR integrated.

oleg0421:
@jens-maus : thank you for verifying and confirming MJPG format works!

akallabeth:
@oleg0421 can you address the remaining issues and rebase this to current master?

oleg0421:
@akallabeth : I re-based it to current master. There are no issues, other than what was originally reported by @hardening, which I believe is USB hardware specific.

channels/rdpecam/client/encoding.c
```diff
@@ -96,27 +150,70 @@ static BOOL ecam_encoder_compress_h264(CameraDeviceStream* stream, const BYTE* s
...
-	size.height = stream->currMediaType.Height;
+	prim_size_t size = { stream->currMediaType.Width, stream->currMediaType.Height };
 	CAM_MEDIA_FORMAT inputFormat = streamInputFormat(stream);
-	enum AVPixelFormat pixFormat = ecamToAVPixFormat(inputFormat);
+	enum AVPixelFormat pixFormat = AV_PIX_FMT_NONE;
+
+#if defined(WITH_INPUT_FORMAT_MJPG)
+	if (inputFormat == CAM_MEDIA_FORMAT_MJPG)
+	{
+		stream->avInputPkt->data = (BYTE*)srcData;
```
akallabeth, line 163: please use `WINPR_CAST_CONST_AWAY_PTR` macro for this type of cast

channels/rdpecam/client/encoding.c
```diff
@@ -96,27 +150,70 @@ static BOOL ecam_encoder_compress_h264(CameraDeviceStream* stream, const BYTE* s
 {
 	UINT32 dstSize = 0;
 	BYTE* srcSlice[4] = { 0 };
+	int srcLineSizes[4] = { 0 };
 	BYTE* yuv420pData[3] = { 0 };
 	UINT32 yuv420pStride[3] = { 0 };
-	prim_size_t size = { 0 };
-	size.width = stream->currMediaType.Width;
-	size.height = stream->currMediaType.Height;
+	prim_size_t size = { stream->currMediaType.Width, stream->currMediaType.Height };
```
akallabeth, line 156: please `WINPR_ASSERT(stream);` before dereferencing

channels/rdpecam/client/encoding.c
```diff
@@ -86,6 +86,60 @@ static enum AVPixelFormat ecamToAVPixFormat(CAM_MEDIA_FORMAT ecamFormat)
...
+
+		case AV_PIX_FMT_YUVJ444P:
+			pixFormat = AV_PIX_FMT_YUV444P;
+			break;
+
+		default:
+			break;
+	}
+
+	const int width = (int)stream->currMediaType.Width;
```
akallabeth, line 129: please `WINPR_ASSERT` here to avoid an unchecked integer overflow (unlikely, but better play it safe)

akallabeth (approved):
LGTM,
a few little nitpicks, but nothing blocking.

channels/rdpecam/client/encoding.c
hardening: if I'm correct static function, only called from this module => not needed, we control the caller

hardening (approved):
I'm rather sure that on my setup the problem does not come from USB connection problems as my cam works perfectly with all other software, and the same happen with direct USB connection. But let's merge this PR and I'll have a closer look when I'll have time (and BTW if we have to drop frames from the CAM that should be handled too).

oleg0421:
@hardening , @akallabeth: I was able to get a hold of Logitech Webcam C930e and do some testing with it. I observed some JPEG decoding errors such as:

```
[mjpeg @ 0x7f9d102824c0] overread 8
[mjpeg @ 0x7f9d102824c0] EOI missing, emulating
```

or
```
[mjpeg @ 0x7f9d47583c40] mjpeg_decode_dc: bad vlc: 0:0 (0x7f9d47580048)
[mjpeg @ 0x7f9d47583c40] error dc
[mjpeg @ 0x7f9d47583c40] error y=134 x=117
```

In my case, all errors show up **_only while decoding first frame from camera_**.  Beginning from second frame, there are no problems with decoding.

I inspected the dumps of the first frame and I can confirm there is random partial data corruption. My guess is that something happens during the capture start between the camera, driver and kernel, which results in this data corruption. So far I don't see anything wrong with the application code itself and this problem doesn't happen with other cameras that I have.

One simple workaround is to always drop 1st frame in case of MJPEG format and it works ok (no decoding errors). Not sure, if this is acceptable as a solution, please let me know.

hardening:
I would not be shocked if we decode beginning at the second frame, or perhaps we must not make the decoding error fatal, or we can live with a certain rate of decoding errors ?

oleg0421:
Currently when the 1st corrupted JPEG frame is being decoded, FFmpeg library detects and prints the problem, but API doesn't return an error (considers it minor). We then try to re-encode to H264. Since it's a 1st frame, it becomes bad H264 I-frame, which results in distorted image for ~10 seconds or so, until next good I-frame is generated. The effect of single corrupted frame is over-amplified.

There's a special flag in FFmpeg API (`AV_EF_EXPLODE`), that tells MJPEG decoder to abort decoding on minor error detection and return error. It will allow us to skip corrupted frames. WIll prepare a small patch in the next few days.

jens-maus:
> We then try to re-encode to H264.

Btw (and sorry for talking to a slightly different direction): Now that RDPCAM support in FreeRDP works quite well, I recognized that there is almost no Webcam out that that directly provides H264 streams anymore, even thought Logitech explicitly announces H264 support for their higher (C925e, C930) models. v4l-cal always return only raw and MJPEG support for these models. I just read that these models seem to attach H264 frames to the MJPEG streams for applications aware of that to pickup that data in that way. Does anyone in here is aware of that or perhaps knows if there is a way to get out that MJPEG attached H264 data? Or do we have any chance to tune FreeRDP or it's ffmpeg dependency to extract and use that H264 data? Because that would greatly help clients to keep CPU usage low of course...

oleg0421:
@jens-maus : a great question, something I've been wondering myself after seeing some forum posts about attaching H264 frames to the MJPEG streams by Logitech cameras. It seems unlikely to me, considering it goes against Logitech's official stance on hardware H264 encoding:

https://www.logitech.com/en-us/video-collaboration/resources/think-tank/articles/article-logitech-and-h264-encoding.html

Another consideration is:  why would they do it this way, but not as separately advertised H264 format, like older C920 cameras do? Just to give some partner competitive advantage by risking potential technical and legal issues - not worth it.

Having said that, if you or someone finds good technical details how it's done (activation of this mode and how to extract from MJPEG frames), then please share and we can try to implement it.

I recognize the importance of optimization, as current MJPEG decoding followed by H264 encoding takes a lot of CPU, especially for higher resolutions and while running on thin cliens. One way to address that could be adding hardware acceleration (VAAPI) support for H264 encoding, which seems possible, considering FreeRDP already uses VAAPI for H264 decoding.

jens-maus:
@oleg0421 I could also indeed not believe that Logitech has actually gone that road in somewhat hiding the H264 data within the MJPEG frames, but it really seems that for their cameras where the mention H264 support this is actually the case. In fact, see here a 5 year old feature request for ffmpeg which actually explains it in more detail:

https://trac.ffmpeg.org/ticket/8413

And in this ticket you will find other links to e.g. gstreamer where they already implemented a kind of demuxer to somewhat extract the H264 data from the MJPEG frames. See here:

https://github.com/GStreamer/gst-plugins-bad/blob/master/sys/uvch264/gstuvch264_mjpgdemux.c

Or for a general stackoverflow talk on that matter:

https://stackoverflow.com/questions/55439184/getting-unable-to-decode-app-fields-while-playing-usb-webcam-stream-through-ff

So it would actually be great if you could have a look at all that H264 smuggling that logitech seem to have put in their recent camera generations. It is really astonishing that even the same hardware models which previously directly provided H264 streams are now – when you buy them freshly – solely announcing MJPEG when plugged in and analyzed via `v4l-ctl`, but they use that MJPEG smuggling to actually provide certain apps (e.g. Skype & co) with some benefit while standard apps do not know about that "innovative" way of hiding the H264 data. In indeed, for thinclient environments, especially where e.g. FreeRDP might be integrated it would be great if the hided H264 data could be extracted. So please let me know what you will find out and how I could assist in getting such a demuxer implemented in FreeRDP.

hardening:
We have a special place in hell for people that put H264 in MJPEG so that they can check the H264 box !

PR #10295, opened 2024-06-21
Resolves the issue of out-of-bounds reading of uninitialized data causing request processing failure.

AsianOtter:
## Problem
An internal error occasionally occurs when pasting files to a remote desktop via cliprdr.
## Analysis
The cliprdr_packet_file_contents_request_new function always allocates 36 bytes to the wStream object, even if haveClipDataId is FALSE. However, when reading the wStream object (in cliprdr_read_file_contents_request function), it determines whether there is a ClipDataId based on its length.

### DIFF
--- a/channels/cliprdr/cliprdr_common.c
+++ b/channels/cliprdr/cliprdr_common.c
@@ -360,7 +360,7 @@ UINT cliprdr_read_file_contents_request(wStream* s, CLIPRDR_FILE_CONTENTS_REQUES
 	Stream_Read_UINT32(s, request->nPositionHigh); /* nPositionHigh (4 bytes) */
 	Stream_Read_UINT32(s, request->cbRequested);   /* cbRequested (4 bytes) */
 
-	if (Stream_GetRemainingLength(s) >= 4)
+	if (Stream_GetRemainingLength(s) >= sizeof(CLIPRDR_FILE_CONTENTS_REQUEST) - sizeof(CLIPRDR_HEADER) - sizeof(BOOL))
 	{
 		Stream_Read_UINT32(s, request->clipDataId); /* clipDataId (4 bytes) */
 		request->haveClipDataId = TRUE;
--- a/channels/cliprdr/cliprdr_common.c
+++ b/channels/cliprdr/cliprdr_common.c
@@ -360,7 +360,7 @@ UINT cliprdr_read_file_contents_request(wStream* s, CLIPRDR_FILE_CONTENTS_REQUES
 	Stream_Read_UINT32(s, request->nPositionHigh); /* nPositionHigh (4 bytes) */
 	Stream_Read_UINT32(s, request->cbRequested);   /* cbRequested (4 bytes) */
 
-	if (Stream_GetRemainingLength(s) >= sizeof(CLIPRDR_FILE_CONTENTS_REQUEST) - sizeof(CLIPRDR_HEADER) - sizeof(BOOL))
+	if (request->common.dataLen >= sizeof(CLIPRDR_FILE_CONTENTS_REQUEST) - sizeof(CLIPRDR_HEADER) - sizeof(BOOL))
 	{
 		Stream_Read_UINT32(s, request->clipDataId); /* clipDataId (4 bytes) */
 		request->haveClipDataId = TRUE;


channels/cliprdr/cliprdr_common.c
```diff
@@ -360,7 +360,7 @@ UINT cliprdr_read_file_contents_request(wStream* s, CLIPRDR_FILE_CONTENTS_REQUES
 	Stream_Read_UINT32(s, request->nPositionHigh); /* nPositionHigh (4 bytes) */
 	Stream_Read_UINT32(s, request->cbRequested);   /* cbRequested (4 bytes) */
 
-	if (Stream_GetRemainingLength(s) >= 4)
+	if (Stream_GetRemainingLength(s) >= sizeof(CLIPRDR_FILE_CONTENTS_REQUEST) - sizeof(CLIPRDR_HEADER) - sizeof(BOOL))
```
akallabeth, line 363: this is wrong.
the `wStream* s` contains all data read from the wire.
this effectively deactivates `clipDataId` reading in all cases as the `sizeof(CLIPRDR_FILE_CONTENTS_REQUEST) - sizeof(CLIPRDR_HEADER)` have already been read (`Stream_Read_UINT32(s, request->cbRequested);` et al)

akallabeth (changes requested):
your analysis is flawed.

`cliprdr_packet_file_contents_request_new` is only used when sending data to the remote end.
the data in `cliprdr_read_file_contents_request` is read from the remote end and contains exactly the data that was transmitted, there is no out of bounds read here.

channels/cliprdr/cliprdr_common.c
AsianOtter: > this is wrong. the `wStream* s` contains all data read from the wire. this effectively deactivates `clipDataId` reading in all cases as the `sizeof(CLIPRDR_FILE_CONTENTS_REQUEST) - sizeof(CLIPRDR_HEADER)` have already been read (`Stream_Read_UINT32(s, request->cbRequested);` et al)

Sorry, this is my editing error. this should be:  request->common.dataLen  >= sizeof(CLIPRDR_FILE_CONTENTS_REQUEST) - sizeof(CLIPRDR_HEADER) - sizeof(BOOL)

AsianOtter:
> your analysis is flawed.
> 
> `cliprdr_packet_file_contents_request_new` is only used when sending data to the remote end. the data in `cliprdr_read_file_contents_request` is read from the remote end and contains exactly the data that was transmitted, there is no out of bounds read here.

To add: My usage scenario is to connect through FreeRDP Proxy.

channels/cliprdr/cliprdr_common.c
AsianOtter: FreeRDP Proxy has both client and server roles, so the problem is obvious. However, in any scenario, it is more accurate to use request->common.dataLen to judge.

AsianOtter:
FreeRDP Proxy has both client and server roles, so the problem is obvious. However, in any scenario, it is more accurate to use request->common.dataLen to judge.

akallabeth:
> FreeRDP Proxy has both client and server roles, so the problem is obvious. However, in any scenario, it is more accurate to use request->common.dataLen to judge.

no, the problem is simply not there.
the proxy does only forward `cliprdr` channel data and does not parse the channel internal data (and the `capacity != length` of a stream)

with your proposed change we now have a possible out of bound read on the stream (you check data provided by network and don´t validate against what we actually got)

akallabeth:
@AsianOtter where does your issue occur? maybe we´re missing something and looking at the wrong bug/fix?

AsianOtter:
> @AsianOtter where does your issue occur? maybe we´re missing something and looking at the wrong bug/fix?



> @AsianOtter where does your issue occur? maybe we´re missing something and looking at the wrong bug/fix?

2.x.x does not support transferring files from server (remote) to client (local) through proxy, but 3.x.x can. So I tried to use 2.x.x proxy and 3.x.x xfreerdp. Then I found that "internal error" was reported when transferring files from client to server through proxy (only single text file could be transferred successfully).
After debugging and analysis, it was found that: xfreerdp read ClipDataId when reading FILE_CONTENTS_REQUEST, and rejected the request because cliprdr_file_context_server_file_range_request could not find ClipDataId. In fact, I saw that the proxy did not have ClipDataId when writing FILE_CONTENTS_REQUEST, and did not modify the length of the stream until cliprdr_packet_send sent FILE_CONTENTS_REQUEST (no call to Stream_SetLength or Stream_SealLength), that is, only when initializing in Stream_New was capacity=length=size set.
Therefore, when reading FILE_CONTENTS_REQUEST from the stream, the length of the data portion of the request should be used to determine whether it contains ClipDataId, rather than the remaining unread length of the stream.

AsianOtter:
> > FreeRDP Proxy has both client and server roles, so the problem is obvious. However, in any scenario, it is more accurate to use request->common.dataLen to judge.
> 
> no, the problem is simply not there. the proxy does only forward `cliprdr` channel data and does not parse the channel internal data (and the `capacity != length` of a stream)
> 
> with your proposed change we now have a possible out of bound read on the stream (you check data provided by network and don´t validate against what we actually got)

If my proposed change is really wrong, maybe I should add a call to Stream_SealLength between cliprdr_write_file_contents_request and cliprdr_packet_send to update the length in the stream structure member to the correct value.

AsianOtter:
After execute channels/cliprdr/client/cliprdr_main.c:913
`s = cliprdr_packet_file_contents_request_new(fileContentsRequest);`
s->length always is 36
However, if request->haveClipDataId is False, the data length is only 24 bytes, plus CLIPRDR_HEADER is 32 bytes. Therefore, you cannot use s->length or Stream_GetRemainingLength(s) to determine whether it contains ClipDataId.
Did I explain it clearly?

akallabeth:
@AsianOtter yes,
1. your assumptions are for `stable-2.0` but you change files for `master/3.x` 
2. you did miss the `Stream_GetPosition` call in `cliprdr_packet_send` (also done in 2.x) which limits to bytes actually written
3. `Stream_GetRemainingLength` is called on functions that received data over the wire (and `Stream_SealLength` has been called)

AsianOtter:
> @AsianOtter yes,
> 
> 1. your assumptions are for `stable-2.0` but you change files for `master/3.x`
> 2. you did miss the `Stream_GetPosition` call in `cliprdr_packet_send` (also done in 2.x) which limits to bytes actually written
> 3. `Stream_GetRemainingLength` is called on functions that received data over the wire (and `Stream_SealLength` has been called)

Thank you. I checked the code according to your suggestion and it is indeed logical. Does that mean that stable-2.0 and master/3.x do not have the problem I mentioned? But I did encounter the problem. I will check and analyze it next week. Have a nice weekend.

AsianOtter:
The freerdp version I used had a bug in cliprdr_server_packet_send, which has been fixed in the latest version. Thanks for your reply.

PR #10296, opened 2024-06-21
[winpr,timezone] fix windows timezone mappings

akallabeth:
On systems build without `-DWITH_TIMEZONE_ICU=ON` (and not windows) the fallback mapping was broken. 
* WINDOWS_TZID_ENTRY struct members were inverted
* Entries with multiple IANA entries (separated by space) are now each checked for a match

### DIFF
--- a/winpr/libwinpr/timezone/TimeZoneNameMapUtils.c
+++ b/winpr/libwinpr/timezone/TimeZoneNameMapUtils.c
@@ -115,7 +115,28 @@ static const char* map_fallback(const char* iana, TimeZoneNameType type)
 	for (size_t x = 0; x < WindowsZonesNrElements; x++)
 	{
 		const WINDOWS_TZID_ENTRY* const entry = &WindowsZones[x];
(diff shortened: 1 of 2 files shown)


winpr/libwinpr/timezone/TimeZoneNameMapUtils.c
```diff
@@ -115,7 +115,25 @@ static const char* map_fallback(const char* iana, TimeZoneNameType type)
...
+			const char* res = NULL;
+			char* tzid = _strdup(entry->tzid);
+			char* ctzid = tzid;
+			while (tzid)
+			{
+				char* space = strchr(tzid, ' ');
+				if (space)
+					*space++ = '\0';
+				if (strcmp(tzid, iana) == 0)
+					res = entry->windows;
```
mfleisz, line 129: Just a minor remark but `break`ing after a successful search will save some cycles

akallabeth: done

mfleisz (approved):

PR #10312, opened 2024-06-26
[gateway,http] implement proper timeouts

akallabeth:
Implement request timeouts in http_response_recv, use TcpConnectTimeout as default
@progxaker this might fix your issues. The timeouts were previously on the wrong level, now the HTTP read is measured and a timeout triggered.


progxaker:
Hello @akallabeth. Thank you for the fix, it works in a scenario: just run FreeRDP)

If you don't mind, I'd like to test it in a scenario: try to connect to a disabled AVD instance, get the TLS certificate dialog, wait a couple of minutes and click "Permanent" to see if the original problem is fixed.

mfleisz (approved):

PR #10321, opened 2024-07-01
[channel,drive] fix directory traversal detection

akallabeth:
filter out .. part of the file or directory name.

### DIFF
--- a/channels/drive/client/drive_file.c
+++ b/channels/drive/client/drive_file.c
@@ -90,6 +90,33 @@ static BOOL drive_file_fix_path(WCHAR* path, size_t length)
 	return TRUE;
 }
 
+static BOOL contains_dotdot(const WCHAR* path, size_t base_length, size_t path_length)
+{
+	WCHAR dotdotbuffer[6] = { 0 };
+	const WCHAR* dotdot = InitializeConstWCharFromUtf8("..", dotdotbuffer, ARRAYSIZE(dotdotbuffer));
+	const WCHAR* tst = path;
+
+	do
+	{
+		tst = _wcsstr(tst, dotdot);
+		if (!tst)
+			return FALSE;
+
+		/* Filter .. sequences in file or directory names */
+		if ((base_length == 0) || (*(tst - 1) == L'/') || (*(tst - 1) == L'\\'))
+		{
+			if (tst + 2 < path + path_length)
+			{
+				if ((tst[2] == '/') || (tst[2] == '\\'))
+					return TRUE;
+			}
+		}
+		tst += 2;
+	} while (TRUE);
+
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test please

channels/drive/client/drive_file.c
```diff
@@ -90,6 +90,34 @@ static BOOL drive_file_fix_path(WCHAR* path, size_t length)
...
+	const size_t len = _wcslen(path);
+
+	do
+	{
+		tst = _wcsstr(tst, dotdot);
+		if (!tst)
+			return FALSE;
+
+		/* Filter .. sequences in file or directory names */
+		if ((*(tst - 1) == L'/') || (*(tst - 1) == L'\\'))
```
mfleisz, line 107: This will read invalid memory if `path` starts with `..` (not sure if this is a possible scenario though)

channels/drive/client/drive_file.c
```diff
@@ -90,6 +90,34 @@ static BOOL drive_file_fix_path(WCHAR* path, size_t length)
...
+			return TRUE;
```
mfleisz, line 108: Shouldn't we check for both - start and end delimiters? Currently i.e. `foo/..foo/bar.txt` would return `TRUE`.

channels/drive/client/drive_file.c
akallabeth: no, we always start with an offset (the local prefix)

channels/drive/client/drive_file.c
akallabeth: hmm, true, need to check that as well.

channels/drive/client/drive_file.c
```diff
@@ -90,6 +90,33 @@ static BOOL drive_file_fix_path(WCHAR* path, size_t length)
 	return TRUE;
 }
 
+static BOOL contains_dotdot(const WCHAR* path, size_t base_length, size_t path_length)
```
mfleisz, line 93: Thinking about - why don't we just let `_wcsstr` do the job i.e. `return _wcsstr(path, "/../") || _wcsstr(path, "\\..\\");`?

akallabeth: we could, but there are 2 variants missing then ;)
(windows allows mixing `\\` and `/`)

akallabeth: there is also the case when we start with `../` (ok, not in our specific case here as we don´t expect `base_length == 0`)

mfleisz: True that - lgtm 👍

mfleisz (approved):

PR #10322, opened 2024-07-01
[server,proxy] remove sample config.ini

akallabeth:
the sample can be dynamically generated by proxy binary, do not leave an (outdated) config example in the repo.


akallabeth:
@freerdp-bot test please

PR #10326, opened 2024-07-01
[server,proxy] remove unused section from config

akallabeth:
Remote GFXSettings section as it was no longer used. This is a non breaking change, existing entries will be ignored (just like up until now) but the generated config now no longer contains it.


hardening (approved):

akallabeth:
@freerdp-bot test please

PR #10327, opened 2024-07-01
core: Fix handling of logon errors during nla_client_begin

mfleisz:
Under certain circumstances `InitializeSecurityContext` returns an error if a wrong password was provided. This PR checks the returned status code and correctly sets the last error code. This allows a client application to ask the user again for credentials.

A scenario where this happens is when a user tries to connect with a wrong password and both machines are joined to the same domain.


hardening (approved):

akallabeth:
@freerdp-bot test please

PR #10348, opened 2024-07-04
[client,common] fix /ipv4 /ipv6 without :force

Frederick888:
As arg->Value is NULL when using /ipv4 or /ipv6 alone, it segfaulted doing strncmp.

---

Btw should the help text be

- ipv4: Prefer IPv4 A record over IPv6 AAAA record
- ipv6: Prefer IPv6 AAAA record over IPv4 A record

?

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -4600,7 +4600,7 @@ static int freerdp_client_settings_parse_command_line_arguments_int(
 		}
 		CommandLineSwitchCase(arg, "ipv4")
 		{
-			if (strncmp(arg->Value, "force", 6) == 0)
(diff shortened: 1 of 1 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test please

akallabeth:
> As arg->Value is NULL when using /ipv4 or /ipv6 alone, it segfaulted doing strncmp.
> 
> Btw should the help text be
> 
>     * ipv4: Prefer IPv4 A record over IPv6 AAAA record
> 
>     * ipv6: Prefer IPv6 AAAA record over IPv4 A record
>

happy about any improvement there ;)

PR #10366, opened 2024-07-09
[core] also recognize mth:// routing token

hardening:
With Azure you also have some routing token of the form mth://xxxxxxx, so accept these as token.


akallabeth (approved):
LGTM,
nitpick comments see above :grin:

PR #10371, opened 2024-07-10
[channels,rdpdr] fix drive redirection hotplug

fujames1:
Currently, after receiving `PAKID_CORE_USER_LOGGEDON`, the client will switch to `RDPDR_CHANNEL_STATE_USER_LOGGEDON`, send Device List Announce, then switch back to `RDPDR_CHANNEL_STATE_READY.`  As a result, any drives mounted later will not be announced.

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -1686,8 +1686,8 @@ static BOOL rdpdr_check_channel_state(rdpdrPlugin* rdpdr, UINT16 packetid)
 			    RDPDR_CHANNEL_STATE_CLIENTID_CONFIRM, RDPDR_CHANNEL_STATE_READY);
 		default:
 		{
-			enum RDPDR_CHANNEL_STATE state = RDPDR_CHANNEL_STATE_READY;
-			return rdpdr_state_check(rdpdr, packetid, state, 1, state);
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
`PAKID_CORE_USER_LOGGEDON` is an optional message that might never be received.
sure it is not easier to fix the check in `rdpdr_try_send_device_list_announce_request`?

fujames1:
> `PAKID_CORE_USER_LOGGEDON` is an optional message that might never be received. sure it is not easier to fix the check in `rdpdr_try_send_device_list_announce_request`?

Yes, it can be fixed that way.   But the current way of handling `PAKID_CORE_USER_LOGGEDON` is confusing to me:  we have an `RDPDR_CHANNEL_STATE_USER_LOGGEDON` state but it is temporary and seems redundant, because we also have a `userLoggedOn` member in `rdpdrPlugin`.

akallabeth:
@fujames1 it is temporary, yes. (better for debugging if you know the exact state you were in)
the `userLoggedOn` argument for device list announce is required as the protocol strictly distinguishes between before and after this message was received. (but there is a backward compatibility as well... see ` device_announce` for some details)

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/10374
thx for the hint.

PR #10374, opened 2024-07-10
[channels,rdpdr] fix drive redirection hotplug

akallabeth:
Currently, after receiving PAKID_CORE_USER_LOGGEDON, the client will switch to RDPDR_CHANNEL_STATE_USER_LOGGEDON, send Device List Announce, then switch back to RDPDR_CHANNEL_STATE_READY.  As a result, any drives mounted later will not be announced.


akallabeth:
@fujames1 does this work for you as well? (merging if you confirm ;))

fujames1:
> @fujames1 does this work for you as well? (merging if you confirm ;))

Yes, it works for me.  thank you!

Issue #10386, opened 2024-07-17
Memory leak


pkopylov:
https://github.com/FreeRDP/FreeRDP/blob/7d6bdbaef8014f7247d3b65efc365c92f1aa5253/libfreerdp/core/gateway/tsg.c#L1342

I think there should be `goto fail` instead of return in order to free `s`.

akallabeth:
[no](https://github.com/FreeRDP/FreeRDP/blob/7d6bdbaef8014f7247d3b65efc365c92f1aa5253/libfreerdp/core/gateway/rpc_client.c#L1121)

pkopylov:
> [no](https://github.com/FreeRDP/FreeRDP/blob/7d6bdbaef8014f7247d3b65efc365c92f1aa5253/libfreerdp/core/gateway/rpc_client.c#L1121)

Now I see, the `rpc_client_write_call()` frees the `s` parameter passed into the function. Thank ypu!

PR #10387, opened 2024-07-17
Fix a memory leak in TsProxySendToServer() in the tsg.c file

pkopylov:
Fix the issue: https://github.com/FreeRDP/FreeRDP/issues/10386


akallabeth:
double free, closing.

PR #10408, opened 2024-07-22
[utils,unwind] fix path to log.h

floppym:
Bug: https://bugs.gentoo.org/936406

### DIFF
--- a/winpr/libwinpr/utils/unwind/debug.c
+++ b/winpr/libwinpr/utils/unwind/debug.c
@@ -30,7 +30,7 @@
 #include "debug.h"
 
 #include <winpr/wlog.h>
-#include "../log.h"
+#include "../../log.h"
 
 #include <dlfcn.h>
 


akallabeth (changes requested):
Looks like a `CMake` issue here:
in `winpr/libwinpr/utils/CMakeLists.txt` there should be a `winpr_include_directory_add(${CMAKE_CURRENT_SOURCE_DIR})` but there is `set(CMAKE_INCLUDE_CURRENT_DIR ON)` (which might be turned on/off externally

akallabeth:
@floppym fixed the missing `CMake` include path with https://github.com/FreeRDP/FreeRDP/pull/10410 - is this still required?

floppym:
I wasn't able to reproduce the error personally. I will apply #10410 on top of 3.6.3 in Gentoo and see if the bug gets re-opened.

floppym:
Closing for now.

Issue #2946, opened 2015-10-25
RemoteApp window minimizes when on an inactive deskop


ariscop:
[video](https://youtu.be/PFNzbvk2wUM), Command line:
`./xfreerdp /v:[server] /clipboard /u:[domain user] /app:C:\\Windows\\System32\\cmd.exe`
Tested with the version in fedora 22 and current master

This happens for all applications, not just powershell
Interestingly though, while a command is writing to powershell it wont minimize

akallabeth:
still an issue?

akallabeth:
Closing, no feedback.

PR #10409, opened 2024-07-22
Fix: Prevent RemoteAPP window minimization on desktop switch in Gnome

MatthewAllDev:
## Summary
This PR addresses an issue where RemoteAPP windows were unintentionally minimized when switching desktops in Gnome (WM Mutter).

## Changes
- Modified `xf_event_PropertyNotify` in `xf_event.c` to check for the Gnome session before setting the minimized state.
- Added a condition to skip setting the minimized state if the desktop session is Gnome.

## Testing
Tested on Gnome with Wayland+Mutter and confirmed that RemoteAPP windows are no longer minimized when switching desktops.

## Issue
Fixes #[2946](https://github.com/FreeRDP/FreeRDP/issues/2946)

## Future Work
I have identified other issues with Mutter, related to window behavior during maximization and subsequent restoration. These issues did not appear with KWin + Wayland. I will work on fixing these problems for Gnome users in future pull requests.

### DIFF
--- a/client/X11/xf_event.c
+++ b/client/X11/xf_event.c
@@ -23,6 +23,8 @@
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
 
+#include <string.h>
+
 #include <winpr/assert.h>
 
 #include <freerdp/log.h>
@@ -35,6 +37,7 @@
 #include "xf_input.h"
 #include "xf_gfx.h"
 #include "xf_graphics.h"
+#include "xf_utils.h"
 
 #include "xf_event.h"
 
@@ -982,7 +985,7 @@ static BOOL xf_event_PropertyNotify(xfContext* xfc, const XPropertyEvent* event,
 			if (status)
 			{
 				/* If the window is in the iconic state */
-				if (((UINT32)*prop == 3))
+				if (((UINT32)*prop == 3) && !IsGnome())
 				{
 					minimized = TRUE;
 					if (appWindow)
--- a/client/X11/xf_utils.h
+++ b/client/X11/xf_utils.h
@@ -20,6 +20,7 @@
 #pragma once
 
 #include <winpr/wlog.h>
+#include <winpr/wtypes.h>
 
 #include <X11/Xlib.h>
 
@@ -89,3 +90,5 @@ int LogTagAndXConvertSelection_ex(const char* tag, const char* file, const char*
 int LogDynAndXConvertSelection_ex(wLog* log, const char* file, const char* fkt, size_t line,
                                   Display* display, Atom selection, Atom target, Atom property,
                                   Window requestor, Time time);
+
+BOOL IsGnome(void);
--- a/client/X11/xf_utils.c
+++ b/client/X11/xf_utils.c
@@ -18,7 +18,9 @@
  * limitations under the License.
  */
 
+#include <string.h>
 #include <winpr/assert.h>
+#include <winpr/wtypes.h>
 
 #include "xf_utils.h"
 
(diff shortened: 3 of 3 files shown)


akallabeth (commented):
Looks like it does what you describe.
This is kind of old code (preceding my time in this project) and I´ve checked https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html
looks like this should not be handled at all here?

akallabeth:
@MatthewAllDev thank you for the pull, have to inquire why the code is there in the first place, hope someone still remembers :grin: 
could you rebase your branch though (and force push) so that we don´t have merges in the pr? thx.

MatthewAllDev:
@akallabeth Sure, I'll do it now. I also don't fully understand why this code is here, but I decided to make a quick fix that minimally impacts other window managers.

client/X11/xf_event.c
```diff
@@ -982,7 +984,7 @@ static BOOL xf_event_PropertyNotify(xfContext* xfc, const XPropertyEvent* event,
 			if (status)
 			{
 				/* If the window is in the iconic state */
-				if (((UINT32)*prop == 3))
+				if (((UINT32)*prop == 3) && strcmp(getenv("DESKTOP_SESSION"), "gnome") != 0)
```
akallabeth, line 987: sorry for the late comment, did miss this.
`getenv` might return `NULL` and `strcmp` does have undefined behavior with`NULL` arguments.

akallabeth: maybe abstract this in `xf_utils.c/h` to generically query for some environment as you mentioned other gnome specific fixes to be done?

MatthewAllDev: @akallabeth Yes, sorry, I didn't consider that. I'll create an abstraction and fix the commit.

MatthewAllDev:
@akallabeth Done. By the way, I discovered an interesting side effect of this change. On the server (Windows 10), I sometimes run an application that takes screenshots, analyzes them, and performs certain actions (clicker). When using a regular RDP connection with the /f option, this application would crash after switching desktops on the client OS. After making the changes, this issue no longer occurs."

client/X11/xf_utils.h
```diff
@@ -89,3 +89,5 @@ int LogTagAndXConvertSelection_ex(const char* tag, const char* file, const char*
 int LogDynAndXConvertSelection_ex(wLog* log, const char* file, const char* fkt, size_t line,
                                   Display* display, Atom selection, Atom target, Atom property,
                                   Window requestor, Time time);
+
+bool IsGnome();
```
akallabeth, line 93: * `bool` is defined in `stdbool.h` (or use `BOOL` with `winpr/wtypes.h`)
* `fkt()` is `C` for arbitrary arguments, use `fkt(void)` instead

akallabeth:
> @akallabeth Done. By the way, I discovered an interesting side effect of this change. On the server (Windows 10), I sometimes run an application that takes screenshots, analyzes them, and performs certain actions (clicker). When using a regular RDP connection with the /f option, this application would crash after switching desktops on the client OS. After making the changes, this issue no longer occurs.

you mean normal desktop RDP session?
this sounds like a bug in that app, as `minimize` sends `suppress output` to the server and the server no longer sends screen updates to the client

client/X11/xf_utils.h
MatthewAllDev: Fixed. Once again, I apologize. I've never written in C for operating systems, only in my youth for controllers =) I used BOOL because it is used in most cases in the project.

akallabeth: no worries, everyone has to start somewhere :grin:

MatthewAllDev:
> you mean normal desktop RDP session?

Yes, that's right.

> this sounds like a bug in that app, as `minimize` sends `suppress output` to the server and the server no longer sends screen updates to the client

Probably so. Although I can admit that this might be done intentionally for optimization, as we don't need those updates while the window is minimized. As they say, not a bug, but a feature (maybe).

akallabeth:
> Probably so. Although I can admit that this might be done intentionally for optimization, as we don't need those updates while the window is minimized. As they say, not a bug, but a feature (maybe).

there is the `-suppress-output` command line flag disabling this, you might test if that helps.

MatthewAllDev:
@akallabeth Yes, thanks. It works for me.

PR #10411, opened 2024-07-23
[core,connection] drop mstshash for RdpVersion > 5+

akallabeth:
Mimic mstsc behaviour, but keep compatible with older servers. Drop the mstshash cookie (containing the username) if the server is windows 10 or newer


akallabeth:
@hedwigz could you check this is what you expected (and works for your case too)?

akallabeth:
@freerdp-bot test please

akallabeth:
@freerdp-bot test please

akallabeth:
does only work by accident in redirection case, otherwise the `RdpVersion` is always the version the RDP client supports.

PR #10441, opened 2024-08-02
[client,x11] log variable name for XGetAtomName

akallabeth:
To better debug issues with calls log the name of the target variable to allow better tracing which call did fail


akallabeth:
@freerdp-bot test again

akallabeth:
@freerdp-bot test please

PR #10472, opened 2024-08-12
Clip rdp 2 rdp fix

akallabeth:
fixes a bug introduced with 3.x preventing clipboard to work from one FreeRDP client to another.

### DIFF
--- a/channels/cliprdr/cliprdr_common.c
+++ b/channels/cliprdr/cliprdr_common.c
@@ -290,7 +290,7 @@ wStream* cliprdr_packet_format_list_new(const CLIPRDR_FORMAT_LIST* formatList,
 			{
 				wszFormatName = NULL;
 
-				if (szFormatName)
+				if (szFormatName && (strnlen(szFormatName, 2) > 0))
 				{
 					wszFormatName = ConvertUtf8ToWCharAlloc(szFormatName, &formatNameSize);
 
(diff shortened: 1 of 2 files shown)


channels/cliprdr/cliprdr_common.c
```diff
@@ -299,9 +299,9 @@ wStream* cliprdr_packet_format_list_new(const CLIPRDR_FORMAT_LIST* formatList,
 						Stream_Free(s, TRUE);
 						return NULL;
 					}
-					formatNameSize += 1; /* append terminating '\0' */
 				}
 
+				formatNameSize++;
```
pnowack, line 304: Why moving this out of the upper `if`-block? This is the ASCII format block (short format names), where the name always uses 32 bytes (unused chars are zeros). If `szFormatName == NULL`, then now only 30 zeros are written, while before 32 where written.

pnowack (commented):
Haven't experimented with the RAW path before, but the rest of the changes look OK.

channels/cliprdr/cliprdr_common.c
akallabeth: because the remaining 2 zero bytes are written down below in line 372

akallabeth:
@tsz8899 sure this is related to this pull request?

akallabeth:
@tsz8899 please
* add comments related to https://github.com/FreeRDP/FreeRDP/pull/10477 there ;)
* what did you try to tell with the last post, does it work with the pr or not?

PR #10474, opened 2024-08-13
[crypto,tls] fix AcceptedKey checks

akallabeth:
* Add freerdp_certificate_get_pem_ex to extract PEM for cert only
* Compare only certificate without certificate chain
* Store only certificate PEM without chain for later comparison
* Do the same for locally accepted certificates, only store the `PEM` without the chain.


pnowack:
This works. Thank you!

pnowack (approved):
The code changes themselves also look good to me.

mfleisz (approved):

PR #10475, opened 2024-08-13
Set `NTLMSSP_NEGOTIATE_LM_KEY` in when sending AM on ntlmv2

tomer-shefler:
Server gets zeros as `LmChallengeResponse` because when the freerdp client sends the `AuthenticateMessage`, the  `NTLMSSP_NEGOTIATE_LM_KEY` flag should be set in order for `LmChallengeResponse` to be sent.
However, the method (`ntlm_write_AuthenticateMessage`) start with flag mask `message->NegotiateFlags` as 0 in no case the bit `NTLMSSP_NEGOTIATE_LM_KEY` is set. 
I've set it under `NTLMv2` because of a precedent in a different function.

### DIFF
--- a/winpr/libwinpr/sspi/NTLM/ntlm_message.c
+++ b/winpr/libwinpr/sspi/NTLM/ntlm_message.c
@@ -1225,6 +1225,7 @@ SECURITY_STATUS ntlm_write_AuthenticateMessage(NTLM_CONTEXT* context, const PSec
 	if (context->NTLMv2)
 	{
 		message->NegotiateFlags |= NTLMSSP_NEGOTIATE_56;
+		message->NegotiateFlags |= NTLMSSP_NEGOTIATE_LM_KEY;
 
 		if (context->SendVersionInfo)
 			message->NegotiateFlags |= NTLMSSP_NEGOTIATE_VERSION;


akallabeth (changes requested):
from [MS-NLMP](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/b38c36ed-2804-4868-a9ff-8dd3182128e4) 2.2.2.5 NEGOTIATE:
```
P (1 bit): If set, requests usage of the NTLM v2 session security. NTLM v2 session security is a
misnomer because it is not NTLM v2. It is NTLM v1 using the extended session security that is also
in NTLM v2. NTLMSSP_NEGOTIATE_LM_KEY and
NTLMSSP_NEGOTIATE_EXTENDED_SESSIONSECURITY are mutually exclusive. If both
NTLMSSP_NEGOTIATE_EXTENDED_SESSIONSECURITY and NTLMSSP_NEGOTIATE_LM_KEY are
requested, NTLMSSP_NEGOTIATE_EXTENDED_SESSIONSECURITY alone MUST be returned to the
client. NTLM v2 authentication session key generation MUST be supported by both the client and
the DC in order to be used, and extended session security signing and sealing requires support
from the client and the server in order to be used.<25> An alternate name for this field is
NTLMSSP_NEGOTIATE_EXTENDED_SESSIONSECURITY.
```

akallabeth:
@tomer-shefler might help if you tell where you experience a bug with this (use case we can test with), but the change looks not correct (or maybe there is another bug hidden)

PR #10477, opened 2024-08-13
[channels,cliprdr] refactor cliprdr_packet_format_list_new

akallabeth:
* Simplify function
* Add missing arguments for ASCII names

follow up to #10472

### DIFF
--- a/channels/cliprdr/cliprdr_common.c
+++ b/channels/cliprdr/cliprdr_common.c
@@ -237,140 +237,76 @@ wStream* cliprdr_packet_file_contents_response_new(const CLIPRDR_FILE_CONTENTS_R
 }
 
 wStream* cliprdr_packet_format_list_new(const CLIPRDR_FORMAT_LIST* formatList,
-                                        BOOL useLongFormatNames)
+                                        BOOL useLongFormatNames, BOOL useAsciiNames)
 {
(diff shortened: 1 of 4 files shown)


akallabeth:
@tsz8899 ok, refactored and simplified the `cliprdr_packet_format_list_new` function, now the strings should be of correct size in all cases.

channels/cliprdr/cliprdr_common.c
```diff
@@ -241,141 +241,85 @@ wStream* cliprdr_packet_file_contents_response_new(const CLIPRDR_FILE_CONTENTS_R
...
-
-			if (asciiNames)
-			{
-				if (szFormatName)
-					formatNameLength = strnlen(szFormatName, 32);
+	if (useLongFormatNames && useAsciiNames)
+		WLog_WARN(TAG,
+		          "called with invalid arguments useLongFormatNames=%s, useAsciiNames requires "
+		          "useLongFormatNames=false, ignoring argument.",
+		          boolstr(useLongFormatNames));
```
mfleisz, line 263: Would be easier to write `useLongFormatNames=true` instead of introducing `boolstr` and using format parameters.

akallabeth:
@mfleisz ok, fixed.

mfleisz (approved):

PR #10488, opened 2024-08-20
[core,update] ensure EndPaint before DesktopResize

akallabeth:
When DesktopResize is called EndPaint might not have been. Ensure the update-lock is released and data flushed before resizing.


hardening (approved):
Seems ok

PR #10494, opened 2024-08-20
Fix ASN.1 integer decoding

fifthdegree:
Treat ASN.1 encoded integers with a leading zero byte and the MSB of the second byte set as non-negative

### DIFF
--- a/winpr/libwinpr/utils/asn1/asn1.c
+++ b/winpr/libwinpr/utils/asn1/asn1.c
@@ -1020,7 +1020,7 @@ static size_t WinPrAsn1DecReadIntegerLike(WinPrAsn1Decoder* dec, WinPrAsn1_tag e
 	size_t ret = readTagAndLen(dec, &dec->source, &tag, &len);
 	if (!ret || (tag != expectedTag))
 		return 0;
-	if (!Stream_CheckAndLogRequiredLength(TAG, &dec->source, len) || (len > 4))
+	if (len == 0 || !Stream_CheckAndLogRequiredLength(TAG, &dec->source, len) || (len > 4))
 		return 0;
 
 	WinPrAsn1_INTEGER val = 0;
--- a/winpr/libwinpr/utils/asn1/asn1.c
+++ b/winpr/libwinpr/utils/asn1/asn1.c
@@ -1024,13 +1024,20 @@ static size_t WinPrAsn1DecReadIntegerLike(WinPrAsn1Decoder* dec, WinPrAsn1_tag e
 		return 0;
 
 	WinPrAsn1_INTEGER val = 0;
-	for (size_t x = 0; x < len; x++)
+	UINT8 v = 0;
+
+	Stream_Read_UINT8(&dec->source, v);
+	if (v & 0x80)
+		val = 0xFFFFFFFF;
+	val |= v;
+
+	for (size_t x = 1; x < len; x++)
 	{
-		INT8 v = 0;
-		Stream_Read_INT8(&dec->source, v);
+		Stream_Read_UINT8(&dec->source, v);
 		val = (WinPrAsn1_INTEGER)(((UINT32)val) << 8);
 		val |= v;
 	}
+
 	*target = val;
 	ret += len;
 


hardening (approved):
Good catch.

hardening:
@freerdp-bot can you also test that fix ?

winpr/libwinpr/utils/asn1/asn1.c
```diff
@@ -1024,13 +1024,20 @@ static size_t WinPrAsn1DecReadIntegerLike(WinPrAsn1Decoder* dec, WinPrAsn1_tag e
 		return 0;
 
 	WinPrAsn1_INTEGER val = 0;
-	for (size_t x = 0; x < len; x++)
+	UINT8 v = 0;
+
+	Stream_Read_UINT8(&dec->source, v);
+	if (v & 0x80000000)
```
hardening, line 1030: That should be `v & 0x80` ?

fifthdegree: Oops, yes. Thanks!

akallabeth:
@freerdp-bot test

hardening:
Ok the fuzzer caught a legitimate bug when the integer retrieval function is used to grab a len and with the modified function we return a negative value that pass the test of length that we do but fails later.

winpr/libwinpr/utils/asn1/asn1.c
```diff
@@ -1024,13 +1024,20 @@ static size_t WinPrAsn1DecReadIntegerLike(WinPrAsn1Decoder* dec, WinPrAsn1_tag e
 		return 0;
 
 	WinPrAsn1_INTEGER val = 0;
-	for (size_t x = 0; x < len; x++)
+	UINT8 v = 0;
+
+	Stream_Read_UINT8(&dec->source, v);
```
akallabeth, line 1029: Missing length check

akallabeth (changes requested):

winpr/libwinpr/utils/asn1/asn1.c
hardening: The length check is done 2 lines above

akallabeth: @hardening and does not handle the case `len == 0` and unconditionally reads 1 byte here

hardening: @akallabeth  oh yeah I missed that one, good catch

hardening:

fifthdegree:
> Ok the fuzzer caught a legitimate bug when the integer retrieval function is used to grab a len and with the modified function we return a negative value that pass the test of length that we do but fails later.

Not sure what you're referring to so didn't do anything about it (sorry). But do feel free to make the changes yourself!

winpr/libwinpr/utils/asn1/asn1.c
```diff
@@ -1020,17 +1020,24 @@ static size_t WinPrAsn1DecReadIntegerLike(WinPrAsn1Decoder* dec, WinPrAsn1_tag e
 	size_t ret = readTagAndLen(dec, &dec->source, &tag, &len);
 	if (!ret || (tag != expectedTag))
 		return 0;
-	if (!Stream_CheckAndLogRequiredLength(TAG, &dec->source, len) || (len > 4))
+	if (len == 0 || !Stream_CheckAndLogRequiredLength(TAG, &dec->source, len) || (len > 4))
```
akallabeth, line 1023: @fifthdegree sure that a 0 length is invalid? should that not be 0 like before this change?

fifthdegree: Yes.

>8.3 Encoding of an integer value
8.3.1 The encoding of an integer value shall be primitive. The contents octets shall consist of one or more octets.

hardening:
> 
> Not sure what you're referring to so didn't do anything about it (sorry). But do feel free to make the changes yourself!

Forget my remark I was wrong.

hardening (approved):

winpr/libwinpr/utils/asn1/asn1.c
akallabeth: @fifthdegree thx, was unsure if 0 is allowed or not, that clarifies it.

PR #10509, opened 2024-08-26
[cmake] mark dependency includes SYSTEM

akallabeth:
Mark all dependency include paths SYSTEM so warnings from system headers are excluded from ci warning statistics


akallabeth:
@freerdp-bot test please

PR #10512, opened 2024-08-26
[warnings] fix cert-err33-c

akallabeth:
* Fix various warnings from ci (see commit messages for specific types)
* Disable some `clang-tidy` checks that we do not need/use
* Disable checks on build directories


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

hardening:
Hum I'm always sceptical with these patches that look like they're there just to make these (sometime inaccurate) check tools happy. Not sure to like the heavy `4ULL * xxx * yy` notation all around the place...

akallabeth:
> Hum I'm always sceptical with these patches that look like they're there just to make these (sometime inaccurate) check tools happy. Not sure to like the heavy `4ULL * xxx * yy` notation all around the place...

that is actually a bugfix (`C` does define how the integer types are expanded, it depends on the first argument)

akallabeth:
@freerdp-bot test

mfleisz:
Just started to have a look at this beast ... but I do not know what the analysis tool has against `if (ptr)` or `if (!ptr)` as this is well defined by the standard and common practice. IMO this does not improve readability and it is also regarded bad practice by the [CppCoreGuideliness](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions).

akallabeth:
@mfleisz that will be https://clang.llvm.org/extra/clang-tidy/checks/readability/implicit-bool-conversion.html - there seem to be opposing views on this matter ^^
anyway, contained in a single commit, so can be dropped and the check disabled if consensus. (did that with a few of the checks that contradict our coding styles already)

akallabeth:
replaced by pull requests for each commit:
#10518 
#10519 
#10520 
#10521

PR #10513, opened 2024-08-26
core: relax check in freerdp_peer_new

hardening:
Not all socket kinds have the TCP_NODELAY option (especially local UNIX sockets), so don't make the setsockopt fatal.


akallabeth (approved):
LGTM

PR #10518, opened 2024-08-28
Clang tidy cleanup

akallabeth:
Since #10512 has grown to an incredibly big monster of a pull request I´ve split out the `clang-tidy` and `CMake` changes in a small separate pr.

here we:
* clean up the `clang-tidy` configuration used on ci
* disable compiler warnings for generated code in build directories


akallabeth:
merging as the changes only affect configuration and bring down the warnings by a considerable amount.
should there still be warnings that we should disable we can do that in a different pr.

PR #10528, opened 2024-08-29
Winpr perf

akallabeth:


akallabeth:
@chergert I´ve refined the trigger condition of `ResetEvent` and `SetEvent` so it is not called repeatedly.
could you validate this improves the situation?

chergert:
Thanks!

Just tested and this dropped `winpr_SubmitThreadpoolWork` down to `.08%` of perf samples here, definitely a win!

PR #10542, opened 2024-08-30
fix type conversion issue with `_Unwind_GetLanguageSpecificData`

chenrui333:
Seeing incompatible type conversion build failure with 3.8.0 release build with newer clang

```
  /tmp/freerdp-20240830-26880-qdym8p/FreeRDP-3.8.0/winpr/libwinpr/utils/unwind/debug.c:107:26: error: incompatible integer to pointer conversion assigning to 'void *' from 'uintptr_t' (aka 'unsigned long') [-Wint-conversion]
                  info->langSpecificData = _Unwind_GetLanguageSpecificData(context);
                                         ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  1 error generated.
```

relates to https://github.com/Homebrew/homebrew-core/pull/182939

### DIFF
--- a/winpr/libwinpr/utils/unwind/debug.c
+++ b/winpr/libwinpr/utils/unwind/debug.c
@@ -104,7 +104,7 @@ static _Unwind_Reason_Code unwind_backtrace_callback(struct _Unwind_Context* con
 	{
 		unwind_info_t* info = &ctx->info[ctx->pos++];
 		info->pc = _Unwind_GetIP(context);
-		info->langSpecificData = _Unwind_GetLanguageSpecificData(context);
+		info->langSpecificData = (void *) _Unwind_GetLanguageSpecificData(context);
 	}
 
 	return _URC_NO_REASON;


akallabeth:
@freerdp-bot test
@chenrui333 interesting, there seem to be differing definitions of `_Unwind_GetLanguageSpecificData` ( on my debian system it is defined as `void *_Unwind_GetLanguageSpecificData(struct _Unwind_Context *);`)

akallabeth:
@chenrui333 did add the fix for this to #10541 
as far as I can tell there are at least two differing definitions of this specific function and there is no usable way to check for it (tried https://cmake.org/cmake/help/latest/module/CheckPrototypeDefinition.html but that does not work there for some reason)

anyway, thank you for the pull!

chenrui333:
Thanks @akallabeth, gonna try in https://github.com/Homebrew/homebrew-core/pull/195079

PR #10548, opened 2024-09-03
[BSD] build-serial-port

akallabeth:
* build serial port code on more platforms
* `#ifdef` linux specific code (TODO: Implement something similar for `*BSD` @VVD maybe you can help there?)


VVD:
Commit 9738fc40f30dfbcce5f1caf7b97ef960b0a7280c build fine with -DCHANNEL_SERIAL=ON on FreeBSD 14.1 amd64 without my addition patches required for 3.8.0.

akallabeth:
@VVD virtual serial ports or similar. (don´t know how that is called in BSD speek)
just connect one end to a terminal (`screen`, `putty`, `minicom`, ... pick your poison) and the other end with `FreeRDP` to another one on the remote end.
the test is type something. should show up on the remote end.

that said, it will most likely not work right now as there are a few `#ifdef __linux__` for a few pieces, did not have time to look up replacements used on `BSD` (if you know them that would be great ;))

PR #10552, opened 2024-09-04
clang tidy fix

hardening:
This PR relaxes some checks with clang-tidy to make our life easier during PR.  Each commit has the arguments to relax these checks.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -57,6 +57,7 @@ Checks: >
     -readability-magic-numbers,
     -readability-qualified-auto,
     -readability-uppercase-literal-suffix,
+    -readability-isolate-declaration,
     -performance-no-int-to-ptr
 WarningsAsErrors: ''
 HeaderFilterRegex: ''
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -27,6 +27,7 @@ Checks: >
     -cert-env33-c,
     -clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling,
     -clang-analyzer-valist.Uninitialized,
+    -cppcoreguidelines-init-variables,
     -cppcoreguidelines-pro-bounds-array-to-pointer-decay,
     -cppcoreguidelines-owning-memory,
     -cppcoreguidelines-avoid-do-while,


akallabeth (changes requested):
no.
both of them make life a lot easier during debugging

hardening:
I made a significant effort to argument in the commit messages pointing the numerous drawbacks of these rules:

* just a matter of taste, shall not be enforced for one rule
* produces tons of false positives, and we have tools that covers the case much better for the other 

I think it deserves more than just an authoritative "no. both of them make life a lot easier during debugging".

And clearly I'm opposed to the further PRs that would be there just to make the code comply with these questionable rules. And as a more general thought the enforced clang-tidy rules shall be discussed, as many of them are presented as so called "good practice" when it's in reality just personal preferences.

akallabeth:
> I made a significant effort to argument in the commit messages pointing the numerous drawbacks of these rules:
> 
>     * just a matter of taste, shall not be enforced for one rule
> 
>     * produces tons of false positives, and we have tools that covers the case much better for the other
no, not one false positive for this rule. (There are other checks I´m currently working through which are not well adjusted for our code, but that is not the case for these rules)

> 
> 
> I think it deserves more than just an authoritative "no. both of them make life a lot easier during debugging".
> 
please provide one valid argument to not initialize variables. (the second rule is more or less a direct consequence, one declaration per line)

> And clearly I'm opposed to the further PRs that would be there just to make the code comply with these questionable rules. And as a more general thought the enforced clang-tidy rules shall be discussed, as many of them are presented as so called "good practice" when it's in reality just personal preferences.
it is not just personal preference, it is a lot of bugs that have been eliminated due to not having uninitialized memory.
while this requires the developer to write a little bit more it makes it a lot easier to debug stuff if you can reproduce an issue without having to worry about random initialized memory.

akallabeth:
@hardening you can go through the list of current warnings https://ci.freerdp.com/job/code-quality-checker/8297/clang/ and check which ones are not really ok for our stuff.
the ones with the most occurrences are the ones that we should address (integer related stuff)

hardening:
@akallabeth I gave an example of a false positive in the commit message for clang-tidy  `-cppcoreguidelines-init-variables` with stack variable initialized directly after declaration by a function that take pointers to these variables. That's one case where systematic variable initialization is not needed.

My reproach on this rule is that the algorithm is stupid and don't do any context analysis, it's just "you have ints or bools, you must initialize them, I don't care of the context, period". And we have other checkers that have better analysis and really catch the problematic cases of potentially use before init. So let's drop `-cppcoreguidelines-init-variables`, this check has too many false positive, as it doesn't report "possible variable use before init" as you pretend, it reports a much bigger set that is "uninitialized int/bool variable at declaration" which is totally different. That second set of course includes the first problematic one, but also tons of cases where initialization is not needed. It seems like clang-tidy is not the appropriate tool to treat that case.

I'd add that the false positive have made you push tons of unneeded changes, suggest and comment unneeded initialization in PRs (for integers but also for more complex objects). That's really a lot of noise for an incomplete rule.

And I agree with you, the other rule goes with it: in a world where not always initializing variables when it's not needed is ok, you can have multiple declaration on the same line, when it stays readable.

(I also feel that the systematic initializations of variable tends to produce warning about "set but not used variables" but I must check that)

akallabeth:
@hardening that example is invalid, as there is no way you can prove that the function does what you claim ;)
that is what I´m talking about.

also, if you have `set but not used variables` then they are exactly that, aka you initialized them twice (aka declared too soon)

(but most of these are from `rdpdr` where we read data from stream but not use it)

akallabeth:
> My reproach on this rule is that the algorithm is stupid and don't do any context analysis, it's just "you have ints or bools, you must initialize them, I don't care of the context, period". And we have other checkers that have better analysis and really catch the problematic cases of potentially use before init. So let's drop `-cppcoreguidelines-init-variables`, this check has too many false positive, as it doesn't report "possible variable use before init" as you pretend, it reports a much bigger set that is "uninitialized int/bool variable at declaration" which is totally different. That second set of course includes the first problematic one, but also tons of cases where initialization is not needed. It seems like clang-tidy is not the appropriate tool to treat that case.
> 

no, the argument is this variable is declared where it is not required. If it is then it is initialized.
yes, in `C` historically you had to declare at beginning of scope, but these times are gone now for > 20 years.

so, initializing a variable boils down to:
1. The variable is declared at wrong scope -> fix by moving to correct scope
2. Variable is declared at correct scope -> you need to initialize as otherwise you get heisenbugs (we use `goto` within functions to jump to error condition exit, uninitialized there means often crashes)

not initializing to:
1. Some really critical piece of code that needs this last percent of speedup (hint: we don´t really have that, all that code is already in inline assembly)
1. too lazy to write the assignment or don´t like it -> not much we can do here, we´ll always disagree

so, no, the rule is well thought out the way it is.
benefits I already experienced:
1. More reliable debugging (no more random variable values to catch)
4. Faster to find issues in compiler warnings (clean code produces less warnings, so less to analyze after each build)

hardening:
First, you keep arguing that I would be against checks, that's the point of almost all your responses: defending doing checks (straw man argument). My position is that I'm against unnecessary checks, and we had exactly the same conversation with asserts. So please arg for the pertinence of these precise checks, not the pertinence of checks in general as I'm not against checks.

> @hardening that example is invalid, as there is no way you can prove that the function does what you claim ;) that is what I´m talking about.
> 

Well you're wrong, and the other checkers that aren't just tidiers, are able to do such context propagation. A tool that is context aware knows the modified arguments of a function and can deduce from that. We even have evidences of this in the reports of the other tools. We even have cases where the tool is able to check all the possible paths, and so detect cases with unneeded conditional initialization.

Again let's use the right tools, `clang-tidy` is not the correct tool for checking "use of var before init" case.

hardening:
> > My reproach on this rule is that the algorithm is stupid and don't do any context analysis, it's just "you have ints or bools, you must initialize them, I don't care of the context, period". And we have other checkers that have better analysis and really catch the problematic cases of potentially use before init. So let's drop `-cppcoreguidelines-init-variables`, this check has too many false positive, as it doesn't report "possible variable use before init" as you pretend, it reports a much bigger set that is "uninitialized int/bool variable at declaration" which is totally different. That second set of course includes the first problematic one, but also tons of cases where initialization is not needed. It seems like clang-tidy is not the appropriate tool to treat that case.
> 
> no, the argument is this variable is declared where it is not required. If it is then it is initialized. yes, in `C` historically you had to declare at beginning of scope, but these times are gone now for > 20 years.
> 
> so, initializing a variable boils down to:
> 
>     1. The variable is declared at wrong scope -> fix by moving to correct scope
> 
>     2. Variable is declared at correct scope -> you need to initialize as otherwise you get heisenbugs (we use `goto` within functions to jump to error condition exit, uninitialized there means often crashes)
> 

I don't understand your argument here, I perfectly understand the problem and agree with you on that. Just `clang-tidy` leads to unnecessary initializations.

> 
> not initializing to:
> 
>     1. Some really critical piece of code that needs this last percent of speedup (hint: we don´t really have that, all that code is already in inline assembly)
> 
>     2. too lazy to write the assignment or don´t like it -> not much we can do here, we´ll always disagree
> 

Well one could arg that the laziness is not willing to check if the initialization is needed or not

> 
> so, no, the rule is well thought out the way it is. benefits I already experienced:
> 
>     1. More reliable debugging (no more random variable values to catch)
> 
>     2. Faster to find issues in compiler warnings (clean code produces less warnings, so less to analyze after each build)

Well, for the PR I've seen, there's probably less that 5% of real fixes and so 95% of "noise", so that doesn't bring that much reliability (these 5% that were caught by smarter tools anyway). If compilers produce false positive let's just silent these warnings and rely on more effective tools that will not force us to do mechanical but unnecessary inits.

PR #10554, opened 2024-09-04
Tidy silence

akallabeth:


libfreerdp/crypto/tls.c
```diff
@@ -309,8 +309,22 @@ static long bio_rdp_tls_ctrl(BIO* bio, int cmd, long num, void* ptr)
...
+			{
+				union
+				{
+					void (*fkt)(const SSL*, int, int);
+					ULONG_PTR uptr;
+				} cnv;
+				cnv.fkt = SSL_get_info_callback(tls->ssl);
+				*((ULONG_PTR*)ptr) = cnv.uptr;
+				status = 1;
+			}
```
mfleisz, line 327: As discussed yesterday - just making a correct cast would have avoided the union hacking and the explanation for it (the initial cast to `ULONG_PTR` is just wrong):

`*((void**)ptr) = (void*)(uintptr_t)SSL_get_info_callback(tls->ssl);`

Works fine on gcc (trunk), clang (trunk) and msvc: https://godbolt.org/z/TaTsna4xc

Now we have a 16-liner instead of a 2-liner ...

akallabeth: @mfleisz that was the one we had with the warning yesterday. (the one where we case function pointers)
the initial cast is necessary as you can not dereference a `void*` (which is the type the argument has)

akallabeth: ok, looking at your example, the `ULONG_PTR` stuff can be replaced, true.
still requires a little explanation why we just cast around the function pointer

akallabeth: for reference, that code yielded a compiler warning `-Wbad-function-cast`, but I´ll test if the `*((void**)ptr)` stuff also fixes it

PR #10563, opened 2024-09-05
Add new transport io layer design and support custom socket.

llyzs:
The original TCPConnect/Attach will remain valid until all existing gateways are migrated, and then should become obsolete.

### DIFF
--- a/include/freerdp/transport_io.h
+++ b/include/freerdp/transport_io.h
@@ -32,6 +32,23 @@
 extern "C"
 {
 #endif
+	typedef int (*pTransportLayerRead)(void* userContext, void* data, int bytes);
+	typedef int (*pTransportLayerWrite)(void* userContext, const void* data, int bytes);
+	typedef BOOL (*pTransportLayerFkt)(void* userContext);
+	typedef BOOL (*pTransportLayerWait)(void* userContext, BOOL waitWrite, DWORD timeout);
+	typedef HANDLE (*pTransportLayerGetEvent)(void* userContext);
+
+	struct rdp_transport_layer
+	{
+		ALIGN64 void* userContext;
+		ALIGN64 pTransportLayerRead Read;
+		ALIGN64 pTransportLayerWrite Write;
+		ALIGN64 pTransportLayerFkt Close;
+		ALIGN64 pTransportLayerWait Wait;
+		ALIGN64 pTransportLayerGetEvent GetEvent;
+		UINT64 reserved[64 - 6]; /* Reserve some space for ABI compatibility */
+	};
+	typedef struct rdp_transport_layer rdpTransportLayer;
 
 	typedef int (*pTCPConnect)(rdpContext* context, rdpSettings* settings, const char* hostname,
 	                           int port, DWORD timeout);
@@ -42,6 +59,10 @@ extern "C"
 	typedef BOOL (*pTransportGetPublicKey)(rdpTransport* transport, const BYTE** data,
 	                                       DWORD* length);
 	typedef BOOL (*pTransportSetBlockingMode)(rdpTransport* transport, BOOL blocking);
+	typedef rdpTransportLayer* (*pTransportConnectLayer)(rdpTransport* transport,
+	                                                     const char* hostname, int port,
+	                                                     DWORD timeout);
+	typedef BOOL (*pTransportAttachLayer)(rdpTransport* transport, rdpTransportLayer* layer);
 
 	struct rdp_transport_io
 	{
@@ -52,10 +73,12 @@ extern "C"
 		pTransportFkt TransportDisconnect;
 		pTransportRWFkt ReadPdu;  /* Reads a whole PDU from the transport */
 		pTransportRWFkt WritePdu; /* Writes a whole PDU to the transport */
-		pTransportRead ReadBytes; /* Reads up to a requested amount of bytes from the transport */
+		pTransportRead ReadBytes; /* Reads up to a requested amount of bytes */
 		pTransportGetPublicKey GetPublicKey;
 		pTransportSetBlockingMode SetBlockingMode;
-		UINT64 reserved[54]; /* Reserve some space for ABI compatibility */
+		pTransportConnectLayer ConnectLayer;
+		pTransportAttachLayer AttachLayer;
+		UINT64 reserved[64 - 12]; /* Reserve some space for ABI compatibility */
 	};
 	typedef struct rdp_transport_io rdpTransportIo;
 
@@ -77,6 +100,9 @@ extern "C"
 	FREERDP_API SSIZE_T transport_parse_pdu(rdpTransport* transport, wStream* s, BOOL* incomplete);
 	FREERDP_API rdpContext* transport_get_context(rdpTransport* transport);
 	FREERDP_API rdpTransport* freerdp_get_transport(rdpContext* context);
+
+	FREERDP_API rdpTransportLayer* transport_layer_new(rdpTransport* transport, size_t contextSize);
+	FREERDP_API void transport_layer_free(rdpTransportLayer* layer);
 
 #ifdef __cplusplus
 }
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -1336,3 +1336,200 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
 
 	return sockfd;
 }
+
+struct rdp_tcp_layer
+{
+	int sockfd;
+	HANDLE hEvent;
+};
+typedef struct rdp_tcp_layer rdpTcpLayer;
+
+static int freerdp_tcp_layer_read(void* userContext, void* data, int bytes)
+{
+	if (!userContext)
+		return -1;
+	if (!data || !bytes)
+		return 0;
+
+	rdpTcpLayer* tcpLayer = (rdpTcpLayer*)userContext;
+
+	int error = 0;
+	int status = 0;
+
+	WSAResetEvent(tcpLayer->hEvent);
+	status = _recv(tcpLayer->sockfd, data, bytes, 0);
+
+	if (status > 0)
+		return status;
+
+	if (status == 0)
+		return -1; /* socket closed */
+
+	error = WSAGetLastError();
+
+	if ((error == WSAEWOULDBLOCK) || (error == WSAEINTR) || (error == WSAEINPROGRESS) ||
+	    (error == WSAEALREADY))
+	{
+		status = 0;
+	}
+	else
+	{
+		status = -1;
+	}
+
+	return status;
+}
+
+static int freerdp_tcp_layer_write(void* userContext, const void* data, int bytes)
+{
+	if (!userContext)
+		return -1;
+	if (!data || !bytes)
+		return 0;
+
+	rdpTcpLayer* tcpLayer = (rdpTcpLayer*)userContext;
+
+	int error = 0;
+	int status = 0;
+
+	status = _send(tcpLayer->sockfd, data, bytes, 0);
+
+	if (status <= 0)
+	{
+		error = WSAGetLastError();
+
+		if ((error == WSAEWOULDBLOCK) || (error == WSAEINTR) || (error == WSAEINPROGRESS) ||
+		    (error == WSAEALREADY))
+		{
+			status = 0;
+		}
+		else
+		{
+			status = -1;
+		}
+	}
+
+	return status;
+}
+
+static BOOL freerdp_tcp_layer_close(void* userContext)
+{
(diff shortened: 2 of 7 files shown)


include/freerdp/transport_io.h
```diff
@@ -42,20 +59,26 @@ extern "C"
...
-		pTransportFkt TLSAccept;
-		pTransportAttach TransportAttach;
-		pTransportFkt TransportDisconnect;
-		pTransportRWFkt ReadPdu;  /* Reads a whole PDU from the transport */
-		pTransportRWFkt WritePdu; /* Writes a whole PDU to the transport */
-		pTransportRead ReadBytes; /* Reads up to a requested amount of bytes from the transport */
-		pTransportGetPublicKey GetPublicKey;
-		pTransportSetBlockingMode SetBlockingMode;
-		UINT64 reserved[54]; /* Reserve some space for ABI compatibility */
+		ALIGN64 pTCPConnect TCPConnect;
```
akallabeth, line 69: while most likely does not change much, this alignment might break `ABI/API` for existing 3.x

akallabeth (commented):
LGTM overall, except for the `ALIGN64` 
one question though, an existing client can work with the new API without modifications? (`ABI/API` should stay that stable at least)

mfleisz (approved):
Tested under Windows 10 and Fedora 39 connecting via HTTP and legacy RPC gateway  -worked for me 👍

include/freerdp/transport_io.h
llyzs: Without the alignment it is probably already going to break on 32-bit system whenever adding new entries. I think it is better to fix it now.

llyzs:
@akallabeth I removed the alignment just in case - leave it out for next big release.

Existing clients should work unless it also has its own custom transport implementation (because the core is now calling ConnectLayer instead of TCPConnect), and this is something I don't think there is any way to keep absolutely compatible.

akallabeth:
> @akallabeth I removed the alignment just in case - leave it out for next big release.

:+1: 

> Existing clients should work unless it also has its own custom transport implementation (because the core is now calling ConnectLayer instead of TCPConnect), and this is something I don't think there is any way to keep absolutely compatible.

hmm, so the stream dump of older clients is no longer compatible?
that might pose an issue, checking that a bit more in depth

libfreerdp/core/tcp.c
```diff
@@ -1336,3 +1336,200 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
+static BOOL freerdp_tcp_layer_wait(void* userContext, BOOL waitWrite, DWORD timeout)
+{
+	if (!userContext)
+		return FALSE;
+
+	rdpTcpLayer* tcpLayer = (rdpTcpLayer*)userContext;
+
+	int status = -1;
+	int sockfd = tcpLayer->sockfd;
+#ifdef WINPR_HAVE_POLL_H
```
akallabeth, line 1439: should there not already be an abstraction in `WinPR` for waiting?

libfreerdp/core/tcp.c
```diff
@@ -1336,3 +1336,200 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
+}
+
+static BOOL freerdp_tcp_layer_close(void* userContext)
+{
+	if (!userContext)
+		return FALSE;
+
+	rdpTcpLayer* tcpLayer = (rdpTcpLayer*)userContext;
+
+	if (tcpLayer->sockfd >= 0)
```
akallabeth, line 1422: `WINPR_ASSERT` before dereferencing or return `NULL` after a failed check

libfreerdp/core/tcp.c
```diff
@@ -1336,3 +1336,200 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
+		return -1;
+	if (!data || !bytes)
+		return 0;
+
+	rdpTcpLayer* tcpLayer = (rdpTcpLayer*)userContext;
+
+	int error = 0;
+	int status = 0;
+
+	WSAResetEvent(tcpLayer->hEvent);
```
akallabeth, line 1359: `WINPR_ASSERT` before dereferencing or return `NULL` after a failed check

libfreerdp/core/tcp.c
```diff
@@ -1336,3 +1336,200 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
+		return -1;
+	if (!data || !bytes)
+		return 0;
+
+	rdpTcpLayer* tcpLayer = (rdpTcpLayer*)userContext;
+
+	int error = 0;
+	int status = 0;
+
+	status = _send(tcpLayer->sockfd, data, bytes, 0);
```
akallabeth, line 1395: `WINPR_ASSERT` before dereferencing or return `NULL` after a failed check

libfreerdp/core/tcp.c
```diff
@@ -1336,3 +1336,200 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
+{
+	if (!userContext)
+		return FALSE;
+
+	rdpTcpLayer* tcpLayer = (rdpTcpLayer*)userContext;
+
+	int status = -1;
+	int sockfd = tcpLayer->sockfd;
+#ifdef WINPR_HAVE_POLL_H
+	struct pollfd pollset;
```
akallabeth, line 1440: maybe easier to initialize with `={0};` and remove the other 0 stuff later on

libfreerdp/core/tcp.c
```diff
@@ -1336,3 +1336,200 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
+}
+
+static HANDLE freerdp_tcp_layer_get_event(void* userContext)
+{
+	if (!userContext)
+		return NULL;
+
+	rdpTcpLayer* tcpLayer = (rdpTcpLayer*)userContext;
+
+	return tcpLayer->hEvent;
```
akallabeth, line 1482: `WINPR_ASSERT` before dereferencing or return `NULL` after a failed check

libfreerdp/core/tcp.c
```diff
@@ -1336,3 +1336,200 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
+	WINPR_ASSERT(context);
+
+	const rdpSettings* settings = context->settings;
+	WINPR_ASSERT(settings);
+
+	int sockfd = -1;
+	rdpTransportLayer* layer = NULL;
+	rdpTcpLayer* tcpLayer = NULL;
+
+	sockfd = freerdp_tcp_connect(context, hostname, port, timeout);
```
akallabeth, line 1497: move declaration here?

libfreerdp/core/tcp.c
```diff
@@ -1336,3 +1336,200 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
+		goto fail;
+
+	layer->Read = freerdp_tcp_layer_read;
+	layer->Write = freerdp_tcp_layer_write;
+	layer->Close = freerdp_tcp_layer_close;
+	layer->Wait = freerdp_tcp_layer_wait;
+	layer->GetEvent = freerdp_tcp_layer_get_event;
+
+	tcpLayer = (rdpTcpLayer*)layer->userContext;
+	tcpLayer->sockfd = -1;
```
akallabeth, line 1514: `WINPR_ASSERT` before dereferencing or return `NULL` after a failed check

libfreerdp/core/transport.c
```diff
@@ -1477,6 +1480,62 @@ static BOOL transport_default_set_blocking_mode(rdpTransport* transport, BOOL bl
...
+
+	return IFCALLRESULT(FALSE, transport->io.AttachLayer, transport, layer);
+}
+
+BOOL transport_default_attach_layer(rdpTransport* transport, rdpTransportLayer* layer)
+{
+	BIO* layerBio = NULL;
+	BIO* bufferedBio = NULL;
+
+	layerBio = BIO_new(BIO_s_transport_layer());
```
akallabeth, line 1513: Declare variable directly here

libfreerdp/core/transport.c
```diff
@@ -1477,6 +1480,62 @@ static BOOL transport_default_set_blocking_mode(rdpTransport* transport, BOOL bl
...
+	if (!layerBio)
+		goto fail;
+
+	bufferedBio = BIO_new(BIO_s_buffered_socket());
```
akallabeth, line 1517: declare variable directly here

libfreerdp/core/transport.c
```diff
@@ -1809,3 +1870,191 @@ void transport_set_early_user_auth_mode(rdpTransport* transport, BOOL EUAMode)
...
+
+rdpTransportLayer* transport_layer_new(rdpTransport* transport, size_t contextSize)
+{
+	rdpTransportLayer* layer =
+	    (rdpTransportLayer*)calloc(1, sizeof(rdpTransportLayer) + contextSize);
+	if (!layer)
+		return NULL;
+
+	if (contextSize)
+		layer->userContext = layer + 1;
```
akallabeth, line 1882: maybe declare a struct at the beginning and allocate that or just allocate the usercontext separately. (makes memory checks easier if something is done out of bound)

libfreerdp/core/transport.c
```diff
@@ -1809,3 +1870,191 @@ void transport_set_early_user_auth_mode(rdpTransport* transport, BOOL EUAMode)
...
+			status = 0;
+			break;
+	}
+
+	return status;
+}
+
+static int transport_layer_bio_new(BIO* bio)
+{
+	BIO_set_flags(bio, BIO_FLAGS_SHOULD_RETRY);
```
akallabeth, line 2020: `WINPR_ASSERT` before dereferencing or return `error` after a failed check (spot unexpected arguments early on, debugging in the ssl layers is hard)

akallabeth (approved):
some nitpicking on checks, otherwise LGTM

libfreerdp/core/tcp.c
hardening: AFAICT it's not needed as the check is already done line 1385

libfreerdp/core/tcp.c
hardening: Already done line 1417

libfreerdp/core/tcp.c
hardening: We have already an abstraction for that in WinPR (https://github.com/FreeRDP/FreeRDP/blob/master/winpr/libwinpr/synch/pollset.h) but it's not exposed currently

libfreerdp/core/tcp.c
hardening: already done line 1477

libfreerdp/core/tcp.c
hardening: Check not needed the call to `transport_layer_new` guarantees that `layer->userContext` is not NULL

libfreerdp/core/transport.c
hardening: I'm perfectly ok with @llyzs 's design, just 1 alloc means easier new/free procedures.

libfreerdp/core/tcp.c
hardening: check done line 1349

hardening (approved):
LGTM.
Udp will probably make big changes in this, but that's alright for now.

libfreerdp/core/tcp.c
llyzs: The problem in WinPR is that it has a different definition (SOCKET instead of int for sockfd) and using that requires changing a lot of other existing stuff in tcp.c which I don't think is worth the trouble.

libfreerdp/core/tcp.c
llyzs: already done a null pointer check at 1417

libfreerdp/core/tcp.c
hardening: @llyzs sure and winpr pollset would be a pain to expose with a public API, as it's meant for internal use in WinPR

libfreerdp/core/transport.c
llyzs: I made the change, the memory check stuff actually makes sense.

llyzs:
@akallabeth pushed all the changes.

akallabeth (approved):
LGTM

libfreerdp/core/tcp.c
akallabeth: true, then lets leave this for now until we have a better solution ready.

PR #10567, opened 2024-09-06
[client,common] extended parser

akallabeth:
add client common command line parser with custom command line arguments

### DIFF
--- a/include/freerdp/utils/gfx.h
+++ b/include/freerdp/utils/gfx.h
@@ -30,6 +30,8 @@ extern "C"
 {
 #endif
 
+	FREERDP_API const char* rdpgfx_caps_version_str(UINT32 capsVersion);
+
 	FREERDP_API const char* rdpgfx_get_cmd_id_string(UINT16 cmdId);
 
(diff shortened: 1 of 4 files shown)


include/freerdp/utils/gfx.h
```diff
@@ -30,6 +30,8 @@ extern "C"
 {
 #endif
 
+	FREERDP_API const char* rdpgfx_caps_version_str(UINT32 capsVersion);
```
hardening, line 33: Perhaps you could add a javadoc that would contain a "since version 3.X" as it's a public API (I should do the same in my RCG PR BTW)

hardening (approved):

include/freerdp/utils/gfx.h
akallabeth: hmm, true, we should add some doxygen stuff.

akallabeth: https://github.com/FreeRDP/FreeRDP/issues/10568

PR #10569, opened 2024-09-06
Fprintf void

akallabeth:
@hardening did research the global-init warnings, this is only valid for `C++` code, so disabled it here.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -21,6 +21,7 @@ Checks: >
     -altera-id-dependent-backward-branch,
     -altera-struct-pack-align,
     -altera-unroll-loops,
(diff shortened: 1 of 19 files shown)


hardening (commented):
Nobody ever cares of the result of `fprintf`, isn't there a way to exclude that function (and possibly others) from the check, that would make our life easier and code cleaner ?

akallabeth:
> Nobody ever cares of the result of `fprintf`, isn't there a way to exclude that function (and possibly others) from the check, that would make our life easier and code cleaner ?

not quite correct, nobody cares for `stdout` but for files it is a different matter.
anyway, will have a look.

PR #10576, opened 2024-09-09
[core,transport] fix ssl info callback function signature and some wa…

llyzs:
The SSL callback signature must be declared as an exact match to the signature that is actually being called, or crash could occur in some systems.

Also fix two functions that should declared as static (compile warning)

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -97,10 +97,8 @@ struct rdp_transport
 	BOOL earlyUserAuth;
 };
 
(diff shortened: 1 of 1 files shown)


mfleisz:
@freerdp-bot test

mfleisz (approved):
lgtm

mfleisz:
@llyzs Sure the new callback type is correct? BIO_info_cb seems to be correctly defined as `int (*)(BIO*, int, int)` in the OpenSSL headers?

PR #10579, opened 2024-09-09
[core,transport] follow up to #10576

akallabeth:
document the behaviour and reference the place it is used to avoid confusion on future code review


mfleisz (approved):
lgtm

PR #10605, opened 2024-09-14
[audin,winmm]Windows audio record support stereo, /microphone:format:1 no noise

cumt2010:
/microphone:format:1, use `WAVE_FORMAT_PCM` format, 2 channels, it's almost all noise in Windows.
After comment this, it works well, no noise, very clear

### DIFF
--- a/channels/audin/client/winmm/audin_winmm.c
+++ b/channels/audin/client/winmm/audin_winmm.c
@@ -337,16 +337,16 @@ static UINT audin_winmm_set_format(IAudinDevice* device, const AUDIO_FORMAT* for
 		{
 			/* BUG: Many devices report to support stereo recording but fail here.
 			 *      Ensure we always use mono. */
-			if (ppwfx->nChannels > 1)
-			{
-				ppwfx->nChannels = 1;
-			}
+			// if (ppwfx->nChannels > 1)
+			//{
+			//	ppwfx->nChannels = 1;
+			//}
 
-			if (ppwfx->nBlockAlign != 2)
-			{
-				ppwfx->nBlockAlign = 2;
-				ppwfx->nAvgBytesPerSec = ppwfx->nSamplesPerSec * ppwfx->nBlockAlign;
-			}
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
this check was introduced exactly because of bugs with a lot of hardware (see comment)
so a little more explanation and your test cases are required to approve this.

cumt2010:
> this check was introduced exactly because of bugs with a lot of hardware (see comment) so a little more explanation and your test cases are required to approve this.

OS: Windows 10 1703
Mic: Intel SST Audio Device (WDM)

by the way, i want use WAVE_FORMAT_OPUS format,  how to do?
1. What's the correct command line?
2. -DWITH_OPUS=ON

Thanks a lot.

akallabeth:
> by the way, i want use WAVE_FORMAT_OPUS format, how to do?

connect to `gnome-remote-desktop` ;)
check https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeai/d04ffa42-5a0f-4f80-abb1-cc26f71c9452 for supported formats by windows servers. (basically boils down to some old compression formats, `PCM` and `AAC`)
 
>     1. What's the correct command line?
`/microphone` or `/sound`.
anything else is just for *specialized*  use cases, try to avoid parameters. (the server, aka remote end, does choose from what the client supports, so if you set something the server does not support and deactivate all other stuff you end up with a broken config)

cumt2010:
> approve this.

Thanks very much, got it. 
Just use this parameter as bellow, record and playback works very well

Linux:
/microphone:sys:alsa,format:1
/sound:sys:alsa

Windows:
/microphone:format:1
/sound
add comment this, it works.

Now, A new issue for me:  the sound delay is getting worse. sometimes more than 30 secs.

mfleisz:
Just tested with wfreerdp using `/microphone /sound` and I was able to record/playback in the session without any glitches.
Closing as this doesn't fix anything generally broken.

PR #10612, opened 2024-09-16
SDL3 client: improve the treatment of clipboard data requests

hardening:
This patch adds a success flag on clipboard data requests, so that we can react correctly in the SDL handler. I've also made some errors non fatal like if you requests a kind of data that doesn't exist the client should not terminate

### DIFF
--- a/client/SDL/SDL3/sdl_clip.cpp
+++ b/client/SDL/SDL3/sdl_clip.cpp
@@ -285,6 +285,24 @@ UINT sdlClip::SendDataResponse(const BYTE* data, size_t size)
 	return _ctx->ClientFormatDataResponse(_ctx, &response);
 }
 
+UINT sdlClip::SendDataRequest(uint32_t formatID, const std::string& mime)
+{
+	CLIPRDR_FORMAT_DATA_REQUEST request = { .requestedFormatId = formatID };
+
+	_request_queue.push({ formatID, mime });
+
+	WINPR_ASSERT(_ctx);
+	WINPR_ASSERT(_ctx->ClientFormatDataRequest);
+	UINT ret = _ctx->ClientFormatDataRequest(_ctx, &request);
+	if (ret == CHANNEL_RC_OK)
+	{
+		WLog_Print(_log, WLOG_ERROR, "error sending ClientFormatDataRequest, cancelling request");
+		_request_queue.pop();
+	}
+
+	return ret;
+}
+
 std::string sdlClip::getServerFormat(uint32_t id)
 {
 	for (auto& fmt : _serverFormats)
@@ -585,60 +603,73 @@ UINT sdlClip::ReceiveFormatDataResponse(CliprdrClientContext* context,
 	ClipboardLockGuard give_me_a_name(clipboard->_system);
 	std::lock_guard<CriticalSection> lock(clipboard->_lock);
 	if (clipboard->_request_queue.empty())
-		return ERROR_INTERNAL_ERROR;
-
-	auto request = clipboard->_request_queue.front();
-	if (formatDataResponse->common.msgFlags & CB_RESPONSE_FAIL)
 	{
-		WLog_WARN(TAG, "clipboard data request for format %" PRIu32 " [%s], mime %s failed",
-		          request.format(), request.formatstr().c_str(), request.mime().c_str());
+		WLog_Print(clipboard->_log, WLOG_ERROR, "no pending format request");
 		return ERROR_INTERNAL_ERROR;
 	}
 
-	deleted_unique_ptr<void> cdata;
-	UINT32 srcFormatId = 0;
-	switch (request.format())
+	do
 	{
-		case CF_TEXT:
-		case CF_OEMTEXT:
-		case CF_UNICODETEXT:
-			srcFormatId = request.format();
-			break;
+		UINT32 srcFormatId = 0;
+		auto& request = clipboard->_request_queue.front();
+		bool success = (formatDataResponse->common.msgFlags & CB_RESPONSE_OK) &&
+		               !(formatDataResponse->common.msgFlags & CB_RESPONSE_FAIL);
+		request.setSuccess(success);
 
-		case CF_DIB:
-		case CF_DIBV5:
-			srcFormatId = request.format();
-			break;
-
-		default:
+		if (!success)
 		{
+			WLog_Print(clipboard->_log, WLOG_WARN,
+			           "clipboard data request for format %" PRIu32 " [%s], mime %s failed",
+			           request.format(), request.formatstr().c_str(), request.mime().c_str());
+			break;
+		}
 
-			auto name = clipboard->getServerFormat(request.format());
-			if (!name.empty())
+		switch (request.format())
+		{
+			case CF_TEXT:
+			case CF_OEMTEXT:
+			case CF_UNICODETEXT:
+				srcFormatId = request.format();
+				break;
+
+			case CF_DIB:
+			case CF_DIBV5:
+				srcFormatId = request.format();
+				break;
+
+			default:
 			{
-				if (name == type_FileGroupDescriptorW)
+				auto name = clipboard->getServerFormat(request.format());
+				if (!name.empty())
 				{
-					srcFormatId =
-					    ClipboardGetFormatId(clipboard->_system, type_FileGroupDescriptorW);
+					if (name == type_FileGroupDescriptorW)
+					{
+						srcFormatId =
+						    ClipboardGetFormatId(clipboard->_system, type_FileGroupDescriptorW);
 
-					if (!cliprdr_file_context_update_server_data(clipboard->_file,
-					                                             clipboard->_system, data, size))
-						return ERROR_INTERNAL_ERROR;
-				}
-				else if (name == type_HtmlFormat)
-				{
-					srcFormatId = ClipboardGetFormatId(clipboard->_system, type_HtmlFormat);
+						if (!cliprdr_file_context_update_server_data(
+						        clipboard->_file, clipboard->_system, data, size))
+							return ERROR_INTERNAL_ERROR;
+					}
+					else if (name == type_HtmlFormat)
+					{
+						srcFormatId = ClipboardGetFormatId(clipboard->_system, type_HtmlFormat);
+					}
 				}
 			}
+			break;
 		}
-		break;
-	}
 
-	const BOOL sres = ClipboardSetData(clipboard->_system, srcFormatId, data, size);
-	if (!sres)
+		if (!ClipboardSetData(clipboard->_system, srcFormatId, data, size))
+		{
+			WLog_Print(clipboard->_log, WLOG_ERROR, "error when setting clipboard data");
+			return ERROR_INTERNAL_ERROR;
+		}
+	} while (false);
+
+	if (!SetEvent(clipboard->_event))
 		return ERROR_INTERNAL_ERROR;
 
-	(void)SetEvent(clipboard->_event);
 	return CHANNEL_RC_OK;
 }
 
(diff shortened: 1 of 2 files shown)


client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -585,18 +603,25 @@ UINT sdlClip::ReceiveFormatDataResponse(CliprdrClientContext* context,
...
+	bool success = (formatDataResponse->common.msgFlags & CB_RESPONSE_OK) &&
+	               !(formatDataResponse->common.msgFlags & CB_RESPONSE_FAIL);
+	request.setSuccess(success);
+
+	if (!success)
 	{
 		WLog_WARN(TAG, "clipboard data request for format %" PRIu32 " [%s], mime %s failed",
 		          request.format(), request.formatstr().c_str(), request.mime().c_str());
-		return ERROR_INTERNAL_ERROR;
+		goto out;
```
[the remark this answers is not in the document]
hardening: stupid advice, we shall drop `cppcoreguidelines-avoid-goto`

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -671,27 +701,51 @@ const void* sdlClip::ClipDataCb(void* userdata, const char* mime_type, size_t* s
 	}
 
 	{
-		HANDLE hdl[2] = { clip->_event, freerdp_abort_event(clip->_sdl->context()) };
-		auto status = WaitForMultipleObjects(ARRAYSIZE(hdl), hdl, FALSE, INFINITE);
-		if (status != WAIT_OBJECT_0)
-			return nullptr;
+		HANDLE hdl[2] = { freerdp_abort_event(clip->_sdl->context()), clip->_event };
+		DWORD status;
```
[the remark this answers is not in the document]
hardening: false positive, we shall remove `cppcoreguidelines-init-variables`

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -671,27 +701,51 @@ const void* sdlClip::ClipDataCb(void* userdata, const char* mime_type, size_t* s
 	}
 
 	{
-		HANDLE hdl[2] = { clip->_event, freerdp_abort_event(clip->_sdl->context()) };
-		auto status = WaitForMultipleObjects(ARRAYSIZE(hdl), hdl, FALSE, INFINITE);
-		if (status != WAIT_OBJECT_0)
-			return nullptr;
+		HANDLE hdl[2] = { freerdp_abort_event(clip->_sdl->context()), clip->_event };
```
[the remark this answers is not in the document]
hardening: we should drop `cppcoreguidelines-avoid-c-arrays`, there's no rational to use complicated `C++` construct that could not behave as expected instead of plan C arrays, especially when it's for calling a `C` function.

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -743,7 +797,8 @@ bool sdlClip::mime_is_html(const std::string& mime)
 	return mime.compare(mime_html) == 0;
 }
 
-ClipRequest::ClipRequest(UINT32 format, std::string mime) : _format(format), _mime(std::move(mime))
+ClipRequest::ClipRequest(UINT32 format, const std::string& mime)
+    : _format(format), _mime(mime), _success(false)
```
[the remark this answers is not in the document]
hardening: When you see the justification of `modernize-pass-by-value`, you can eventually have a benefit with the `move` operator but you must match so many conditions that at the end the compiler is not sure to be able to do the cheap optimization. While there's lots of ways to misuse that (the funny thing is that they recognize themselves in the rule description that it's error prone), so I'd be to not enforce the "modern" thing and put off `modernize-pass-by-value`. Unfortunately the case I've changed is one where the compiler can do the optimization :-(

client/SDL/SDL3/sdl_clip.cpp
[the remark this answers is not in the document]
akallabeth: @hardening you´re in `C++` here, so good advice ;)
you can do `RAII` with `std::unique_ptr` so even `malloc` allocated data is freed up properly.

client/SDL/SDL3/sdl_clip.cpp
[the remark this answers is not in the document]
akallabeth: @hardening yes, that one is not really helpful.
not fully through with the `C` checks yet, but this one is on the list to be deactivated.

client/SDL/SDL3/sdl_clip.cpp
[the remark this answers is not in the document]
akallabeth: maybe this helps: https://stackoverflow.com/questions/48842770/how-do-i-run-a-cleanup-code-on-the-function-exit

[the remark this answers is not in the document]
akallabeth: and maybe this one https://stackoverflow.com/questions/10270328/the-simplest-and-neatest-c11-scopeguard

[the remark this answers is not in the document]
akallabeth: grepping through code, I see there are a few places left that use `goto` in the `C++` components, should create a separate pr to clean that up.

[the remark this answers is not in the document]
hardening: Well I'm perfectly aware of scopeguard, but sometime a good old `goto` is way more readable than a complicated C++ construct (the links that you gave are the perfect example of that, it seems like PERL code, but it's C++). 
Anyway I'll see if scopeguard makes sense for this precise function, but as a general rule I'm for letting people use gotos when it makes sense (and so remove the bothering rule).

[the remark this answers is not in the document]
akallabeth: @hardening please. the second example is a template class (really on hello world complexity) and a template helper function to create an instance of that.
you gain proper scoping and eliminate all the bad stuff of `goto` (just jumping around without any checks at all)
in `C` there is no real alternative, but for `C++` there is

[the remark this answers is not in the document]
hardening: So I managed to deal it with a `do { .... if(error condition) break; ... } while(0);` but I couldn't find a non over-complex way to handle the thing even with scoped guard. And in fact this is an hidden `goto`, just like is the scoped guard in a way. That "never use goto rule" is not accurate, the rule must be don't use goto inappropriately.

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -585,60 +603,74 @@ UINT sdlClip::ReceiveFormatDataResponse(CliprdrClientContext* context,
...
-		case CF_TEXT:
-		case CF_OEMTEXT:
-		case CF_UNICODETEXT:
-			srcFormatId = request.format();
-			break;
-
-		case CF_DIB:
-		case CF_DIBV5:
-			srcFormatId = request.format();
+		deleted_unique_ptr<void> cdata;
```
akallabeth, line 613: leftover?

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -585,60 +603,74 @@ UINT sdlClip::ReceiveFormatDataResponse(CliprdrClientContext* context,
...
+		UINT32 srcFormatId = 0;
+		ClipRequest& request = clipboard->_request_queue.front();
```
akallabeth, line 615: `auto &`?

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -671,27 +703,51 @@ const void* sdlClip::ClipDataCb(void* userdata, const char* mime_type, size_t* s
...
 	{
-		HANDLE hdl[2] = { clip->_event, freerdp_abort_event(clip->_sdl->context()) };
-		auto status = WaitForMultipleObjects(ARRAYSIZE(hdl), hdl, FALSE, INFINITE);
-		if (status != WAIT_OBJECT_0)
-			return nullptr;
+		HANDLE hdl[2] = { freerdp_abort_event(clip->_sdl->context()), clip->_event };
+		DWORD status;
+		do
+		{
+			status = WaitForMultipleObjects(ARRAYSIZE(hdl), hdl, FALSE, 1 * 1000);
```
akallabeth, line 710: please avoid polling loops

akallabeth (approved):
Overall LGTM

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -285,6 +285,24 @@ UINT sdlClip::SendDataResponse(const BYTE* data, size_t size)
...
+	CLIPRDR_FORMAT_DATA_REQUEST request = { .requestedFormatId = formatID };
+
+	_request_queue.push({ formatID, mime });
+
+	WINPR_ASSERT(_ctx);
+	WINPR_ASSERT(_ctx->ClientFormatDataRequest);
+	UINT ret = _ctx->ClientFormatDataRequest(_ctx, &request);
+	if (ret == CHANNEL_RC_OK)
+	{
+		WLog_ERR(TAG, "error sending ClientFormatDataRequest, cancelling request");
```
akallabeth, line 299: class has member `_log`, so better use `WLog_Print(_log, WLOG_ERROR, ...`

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -585,60 +603,74 @@ UINT sdlClip::ReceiveFormatDataResponse(CliprdrClientContext* context,
...
 	std::lock_guard<CriticalSection> lock(clipboard->_lock);
 	if (clipboard->_request_queue.empty())
-		return ERROR_INTERNAL_ERROR;
-
-	auto request = clipboard->_request_queue.front();
-	if (formatDataResponse->common.msgFlags & CB_RESPONSE_FAIL)
 	{
-		WLog_WARN(TAG, "clipboard data request for format %" PRIu32 " [%s], mime %s failed",
-		          request.format(), request.formatstr().c_str(), request.mime().c_str());
+		WLog_ERR(TAG, "no pending format request");
```
akallabeth, line 607: class has member `_log`, so better use `WLog_Print(_log, WLOG_ERROR, ...`

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -585,60 +603,74 @@ UINT sdlClip::ReceiveFormatDataResponse(CliprdrClientContext* context,
...
+		deleted_unique_ptr<void> cdata;
+		UINT32 srcFormatId = 0;
+		ClipRequest& request = clipboard->_request_queue.front();
+		bool success = (formatDataResponse->common.msgFlags & CB_RESPONSE_OK) &&
+		               !(formatDataResponse->common.msgFlags & CB_RESPONSE_FAIL);
+		request.setSuccess(success);
+
+		if (!success)
+		{
+			WLog_WARN(TAG, "clipboard data request for format %" PRIu32 " [%s], mime %s failed",
```
akallabeth, line 622: class has member `_log`, so better use `WLog_Print(_log, WLOG_WARN, ...`

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -585,60 +603,74 @@ UINT sdlClip::ReceiveFormatDataResponse(CliprdrClientContext* context,
...
+			break;
 		}
-		break;
-	}
 
-	const BOOL sres = ClipboardSetData(clipboard->_system, srcFormatId, data, size);
-	if (!sres)
+		if (!ClipboardSetData(clipboard->_system, srcFormatId, data, size))
+		{
+			WLog_ERR(TAG, "error when setting clipboard data");
```
akallabeth, line 666: class has member `_log`, so better use `WLog_Print(_log, WLOG_ERROR, ...`

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -671,27 +703,51 @@ const void* sdlClip::ClipDataCb(void* userdata, const char* mime_type, size_t* s
...
-		clip->_cache_data.insert({ mime_type, { len, ptr } });
-		*size = len;
-		return ptr.get();
+		if (request.success())
+		{
+			auto formatID = ClipboardRegisterFormat(clip->_system, mime_type);
+			auto data = ClipboardGetData(clip->_system, formatID, &len);
+			if (!data)
+			{
+				WLog_ERR(TAG, "error retrieving clipboard data");
```
akallabeth, line 740: class has member `_log`, so better use `WLog_Print(_log, WLOG_ERROR, ...`

client/SDL/SDL3/sdl_clip.cpp
hardening: Any reason for that ? We could get stuck forever there with `INFINITE` time, that also mean that we can handle a timeout for the SDL response as an improvement...

client/SDL/SDL3/sdl_clip.cpp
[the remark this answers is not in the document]
akallabeth: yes, that one we can agree on.

client/SDL/SDL3/sdl_clip.cpp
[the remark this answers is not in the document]
akallabeth: but, not initialized.
also, remove the polling construct please.

client/SDL/SDL3/sdl_clip.cpp
akallabeth: @hardening yes, power consumption. makes a huge difference when I run `xfreerdp` without a lot of stuff going on. (well, not this one alone, but these polling loops keep accumulating)
if you get stuck fix the polling function ;)

hardening: Hum, I'll change it, but AFAICT most polling loops in FreeRDP are on a model where they wait `INFINITE` on a real handle and a stop event.
If you have power consumption problems perhaps that's the sign that we have some polling loops that must have timeout (because of the real need of a timeout, or because we have an external event that can't be WaitForObjected) but the wait delay is not computed correctly. Could also be a missing `ResetEvent` somewhere..

akallabeth: @hardening I don´t have problems (now), but if you use polling loops (as we did with `xfreerdp` previously) that increases `CPU` load and power consumption/wake ups
The loop here is also not exposing any of the arguments against `INFINITE`, it just waits for clipboard events or `rdpContext::abortEvent`

client/SDL/SDL3/sdl_clip.cpp
akallabeth: as in unused ^^

client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -671,27 +702,45 @@ const void* sdlClip::ClipDataCb(void* userdata, const char* mime_type, size_t* s
 	}
 
 	{
-		HANDLE hdl[2] = { clip->_event, freerdp_abort_event(clip->_sdl->context()) };
-		auto status = WaitForMultipleObjects(ARRAYSIZE(hdl), hdl, FALSE, INFINITE);
-		if (status != WAIT_OBJECT_0)
+		HANDLE hdl[2] = { freerdp_abort_event(clip->_sdl->context()), clip->_event };
+		DWORD status = WaitForMultipleObjects(ARRAYSIZE(hdl), hdl, FALSE, 10 * 1000);
```
akallabeth, line 706: not sure a hard coded timeout is a good idea, got reports of people using `RDP` over satellite :grin: 
anyway, small nitpick for later cleanup.

akallabeth (approved):
LGTM

client/SDL/SDL3/sdl_clip.cpp
hardening: Yeah but here we're in the SDL callback so on the other hand we can't stay here forever...

akallabeth: true.
lets leave it at that and wait for the complaints :grin:

PR #10615, opened 2024-09-16
[winpr,asn1] fix sign extraction for ASN1 integer

akallabeth:
ASN1 integer might not be full 32bit integers, so extract the sign from the first byte and append 0x80000000 after the rest of the integer was successfully read.

### DIFF
--- a/winpr/libwinpr/utils/asn1/asn1.c
+++ b/winpr/libwinpr/utils/asn1/asn1.c
@@ -1023,22 +1023,28 @@ static size_t WinPrAsn1DecReadIntegerLike(WinPrAsn1Decoder* dec, WinPrAsn1_tag e
 	if (len == 0 || Stream_GetRemainingLength(&dec->source) < len || (len > 4))
 		return 0;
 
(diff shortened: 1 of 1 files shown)


winpr/libwinpr/utils/asn1/asn1.c
```diff
@@ -1023,22 +1023,33 @@ static size_t WinPrAsn1DecReadIntegerLike(WinPrAsn1Decoder* dec, WinPrAsn1_tag e
 	if (len == 0 || Stream_GetRemainingLength(&dec->source) < len || (len > 4))
 		return 0;
 
-	WinPrAsn1_INTEGER val = 0;
+	union
```
hardening, line 1026: What a price to not have compiler warnings ! There's no warning, but also no one who understand with a quick look what the code is doing...

akallabeth: true, not required here.
but reading a `union` is not really hard, don´t exaggerate ;)

mfleisz (approved):
LGTM

PR #10617, opened 2024-09-16
Sdl scope guard

akallabeth:
use RAII for SDL client cleanup code

### DIFF
--- a/client/SDL/SDL2/sdl_freerdp.cpp
+++ b/client/SDL/SDL2/sdl_freerdp.cpp
@@ -1167,6 +1167,54 @@ static void sdl_post_final_disconnect(freerdp* instance)
 		return;
 }
 
+static void sdl_client_cleanup(SdlContext* sdl, int exit_code, char* error_msg)
+{
+	WINPR_ASSERT(sdl);
+
+	rdpContext* context = sdl->context();
+	WINPR_ASSERT(context);
+
+	rdpSettings* settings = context->settings;
+	WINPR_ASSERT(settings);
+
+	sdl->rdp_thread_running = false;
+	bool showError = false;
+	if (freerdp_settings_get_bool(settings, FreeRDP_AuthenticationOnly))
+		WLog_Print(sdl->log, WLOG_INFO, "Authentication only, exit status %s [%" PRId32 "]",
+		           sdl_map_to_code_tag(exit_code), exit_code);
+	else
+	{
+		switch (exit_code)
+		{
+			case SDL_EXIT_SUCCESS:
+			case SDL_EXIT_DISCONNECT:
+			case SDL_EXIT_LOGOFF:
+			case SDL_EXIT_DISCONNECT_BY_USER:
+			case SDL_EXIT_CONNECT_CANCELLED:
+				break;
+			default:
+			{
+				std::lock_guard<CriticalSection> lock(sdl->critical);
+				if (sdl->connection_dialog && error_msg)
+				{
+					sdl->connection_dialog->showError(error_msg);
+					showError = true;
+				}
+			}
+			break;
+		}
+	}
+	free(error_msg);
+	if (!showError)
+		sdl_hide_connection_dialog(sdl);
+
+	sdl->exit_code = exit_code;
+	sdl_push_user_event(SDL_USEREVENT_QUIT);
+#if SDL_VERSION_ATLEAST(2, 0, 16)
+	SDL_TLSCleanup();
+#endif
+}
+
 /* RDP main loop.
  * Connects RDP, loops while running and handles event and dispatch, cleans up
  * after the connection ends. */
@@ -1189,49 +1237,12 @@ static DWORD WINAPI sdl_client_thread_proc(SdlContext* sdl)
 	BOOL rc = freerdp_connect(instance);
 
 	rdpContext* context = sdl->context();
+	WINPR_ASSERT(context);
+
 	rdpSettings* settings = context->settings;
 	WINPR_ASSERT(settings);
 
-	ScopeGuard guard(
-	    [&]()
-	    {
-		    sdl->rdp_thread_running = false;
-		    bool showError = false;
-		    if (freerdp_settings_get_bool(settings, FreeRDP_AuthenticationOnly))
-			    WLog_Print(sdl->log, WLOG_INFO, "Authentication only, exit status %s [%" PRId32 "]",
-			               sdl_map_to_code_tag(exit_code), exit_code);
-		    else
-		    {
-			    switch (exit_code)
-			    {
-				    case SDL_EXIT_SUCCESS:
-				    case SDL_EXIT_DISCONNECT:
-				    case SDL_EXIT_LOGOFF:
-				    case SDL_EXIT_DISCONNECT_BY_USER:
-				    case SDL_EXIT_CONNECT_CANCELLED:
-					    break;
-				    default:
-				    {
-					    std::lock_guard<CriticalSection> lock(sdl->critical);
-					    if (sdl->connection_dialog && error_msg)
-					    {
-						    sdl->connection_dialog->showError(error_msg);
-						    showError = true;
-					    }
-				    }
-				    break;
-			    }
-		    }
-		    free(error_msg);
-		    if (!showError)
-			    sdl_hide_connection_dialog(sdl);
-
-		    sdl->exit_code = exit_code;
-		    sdl_push_user_event(SDL_USEREVENT_QUIT);
-#if SDL_VERSION_ATLEAST(2, 0, 16)
-		    SDL_TLSCleanup();
-#endif
-	    });
+	ScopeGuard guard([&]() { sdl_client_cleanup(sdl, exit_code, error_msg); });
 
 	if (!rc)
 	{
--- a/client/SDL/SDL2/sdl_freerdp.cpp
+++ b/client/SDL/SDL2/sdl_freerdp.cpp
@@ -1167,7 +1167,7 @@ static void sdl_post_final_disconnect(freerdp* instance)
 		return;
 }
 
-static void sdl_client_cleanup(SdlContext* sdl, int exit_code, char* error_msg)
+static void sdl_client_cleanup(SdlContext* sdl, int exit_code, const std::string& error_msg)
 {
 	WINPR_ASSERT(sdl);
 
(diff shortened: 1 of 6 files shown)


hardening:
This PR has everything about the no gotos rule: so sure the scopeGuard thing when the cleanup part is short and limited make sense and is very convenient.

 But when it's like in `sdl_client_thread_proc` you replace something like:
```
    ...
    if (!do_the_job())
       goto terminate;
    ...
    if (!do_the_second_job())
       goto terminate;
terminate:
     ... 
     massive terminate block
     ....
```
So perfectly linear thing easy to read and understand, a very common way of doing things by:
```
   ScopeGuard guard(
	    [&]()
   {
     ... 
     massive terminate block
     ....
   }
    ...
    if (!do_the_job())
       return exit_code;
    ...
    if (!do_the_second_job())
       return exit_code;
```

So at function exit everything is done in your back: not explicit when you're looking at the code at the return statement position. And the treatment is not a trivial thing like just unsetting a lock like with a `scoped_lock`.
In terms of readability, you have first the last executed statement, so a bit confusing. So `ScopeGuard` is really useful, but in the case of `sdl_client_thread_proc` I feel like it introduces more confusion than the good old goto.

And to be more general: the no goto rule is something I've always heard from university teachers that had never done real projects. And when you hear about the justification for this rule, there was that, at the time, static analysers were not able to explore all possible execution paths and propagate variable usage tainting. There used to also have a "no early exit rule", because an early exit is an hidden goto.
You can have a reasonable usage of gotos that makes the code nicer and easier to read without going in the extreme of spaghetti programming. And the other extreme IMHO is always the last C++ZZZ standard, were the C++ committee pushes constructs that are very hard to read, very hard to understand, and at the end only a few people knows what really happens in terms of generated code: the exact definition of error-prone.

akallabeth:
> This PR has everything about the no gotos rule: so sure the scopeGuard thing when the cleanup part is short and limited make sense and is very convenient.
> 
not only then.

> But when it's like in `sdl_client_thread_proc` you replace something like:
> 
> ```
>     ...
>     if (!do_the_job())
>        goto terminate;
>     ...
>     if (!do_the_second_job())
>        goto terminate;
> terminate:
>      ... 
>      massive terminate block
>      ....
> ```
> 
> So perfectly linear thing easy to read and understand, a very common way of doing things by:
> 
> ```
>    ScopeGuard guard(
> 	    [&]()
>    {
>      ... 
>      massive terminate block
>      ....
>    }
>     ...
>     if (!do_the_job())
>        return exit_code;
>     ...
>     if (!do_the_second_job())
>        return exit_code;
> ```
> 
> So at function exit everything is done in your back: not explicit when you're looking at the code at the return statement position. And the treatment is not a trivial thing like just unsetting a lock like with a `scoped_lock`. In terms of readability, you have first the last executed statement, so a bit confusing. So `ScopeGuard` is really useful, but in the case of `sdl_client_thread_proc` I feel like it introduces more confusion than the good old goto.
> 

* well, it is done consistently at any exit point without exceptions (the odd one out here: `goto`)
* it is easy to understand if you are in `C++` as you know what a `ScopeGuard` (or `LockGuard`, ...) is 

the point you have is that the cleanup code should be moved to a function as it is way too complex (and it was way too complex before)

> And to be more general: the no goto rule is something I've always heard from university teachers that had never done real projects. And when you hear about the justification for this rule, there was that, at the time, static analysers were not able to explore all possible execution paths and propagate variable usage tainting. There used to also have a "no early exit rule", because an early exit is an hidden goto. You can have a reasonable usage of gotos that makes the code nicer and easier to read without going in the extreme of spaghetti programming. And the other extreme IMHO is always the last C++ZZZ standard, were the C++ committee pushes constructs that are very hard to read, very hard to understand, and at the end only a few people knows what really happens in terms of generated code: the exact definition of error-prone.

please stop that nonsense about smarter static analyzers.
the `goto` issues are real (fixed a lot of them in `FreeRDP` - and yes, in your code too)

it is about easy to avoid pitfalls like these nasty helper constructs: `goto` you need in `C` as there is no `RAII` possible there.
the code is exactly the same as before (the whole cleanup code) except for the small header of `ScopeGuard guard([&]() { ... });`

the whole point here is that if we keep on with ignoring warnings and issues like these then they pile up and nobody can find the issues buried beneath a ton of stuff that screams `too lazy to write properly`

hardening:
> it is easy to understand if you are in C++ as you know what a ScopeGuard

Well by `you are in C++` you don't describe a precise concept, most projects have a policy and choose to use some facilities of the C++ language and not others. At Qt: no exceptions, no massive and confusing template usage, etc... Except if I missed something I've seen such discussion were FreeRDP people would have agree on which aspects and constructs shall be used, and which shall be avoided as error-prone or too complicated.

So it's really easy to answer to your `too lazy to write properly` by `too lazy to question the rule`. In which context the rule has been written ? Is it valid in all cases (which could make it not to be enforced) ? What are the drawbacks of the rule ? Is it worth it ? 


And there's a more general question that is, is it a goal to drop all warnings in FreeRDP (even when there's clearly tools false positives) ? You even have made it a moving target as you've increased the warning level when fixing some warnings. We also have evidences of conflicting rules: you've "fixed" some warnings but these fixes show new warnings, so it seems like the rule set is not consistent.

I remember some years ago when you did a massive work with a PR that would have removed a warning about pointer casts, so that was making green buttons on the CI and compiler warnings. But on the other side, that was making the code horrible to read and fortunately @mfleisz and @nfedera brought you back to reality.

So is the future of FreeRDP a project where there's a very strict rule set mostly decided by you, and contributions must match these rules (or you ensure them afterward) ? Code readability is very low on the priority list, way after having green lights with warnings ? Checks are done even when not needed because "you need to do checks even when unnecessary , that always improves security". Some tools reports some annoying false positives, but as the tool as detected some legit bugs previously, the goal is to make the tool happy (the "it as detected real bugs previously" argument so it's not questionable).

So let's push that dystopia, we'll end up with FreeRDP being a project where the code will have 0 warnings but a poor readability. There will have unneeded self-checks in a lot of places because you never know. Contributors will be spammed by checkers of these rules when submitting a PR, and will have to comply to them. Not sure that I still wanna be part of that adventure. 

So I'm calling for an intermediate position where there's:
* more relaxed rules
* more human decisions with applying these rules instead of automating them
* perhaps have more linter/checker "shut up tags" in the code when the tool is at fault
* more discussions and consensus elaboration about these subjects

(figuring that this way too long and should have been a discussion instead of a PR comment)

akallabeth:
> > it is easy to understand if you are in C++ as you know what a ScopeGuard
> 
> Well by `you are in C++` you don't describe a precise concept, most projects have a policy and choose to use some facilities of the C++ language and not others. At Qt: no exceptions, no massive and confusing template usage, etc... Except if I missed something I've seen such discussion were FreeRDP people would have agree on which aspects and constructs shall be used, and which shall be avoided as error-prone or too complicated.
> 

well, we´re talking about basic `C++` concepts like `RAII` here - not exceptions, not template programming

> So it's really easy to answer to your `too lazy to write properly` by `too lazy to question the rule`. In which context the rule has been written ? Is it valid in all cases (which could make it not to be enforced) ? What are the drawbacks of the rule ? Is it worth it ?
>

keep to the language you´re programming in. No, `C++` is not `C` and keeping paradigms of `C` around complicates stuff.
 
> And there's a more general question that is, is it a goal to drop all warnings in FreeRDP (even when there's clearly tools false positives) ? You even have made it a moving target as you've increased the warning level when fixing some warnings. We also have evidences of conflicting rules: you've "fixed" some warnings but these fixes show new warnings, so it seems like the rule set is not consistent.
> 
> I remember some years ago when you did a massive work with a PR that would have removed a warning about pointer casts, so that was making green buttons on the CI and compiler warnings. But on the other side, that was making the code horrible to read and fortunately @mfleisz and @nfedera brought you back to reality.
>

no, that was creating a readable version of casts. having a single macro for a specific kind of cast is quite helpful as you also have a single point to modify it to your needs/compiler/...
(agree on that the first revisions were horrible, so some refinement is always appreciated)
 
> So is the future of FreeRDP a project where there's a very strict rule set mostly decided by you, and contributions must match these rules (or you ensure them afterward) ? Code readability is very low on the priority list, way after having green lights with warnings ? Checks are done even when not needed because "you need to do checks even when unnecessary , that always improves security". Some tools reports some annoying false positives, but as the tool as detected some legit bugs previously, the goal is to make the tool happy (the "it as detected real bugs previously" argument so it's not questionable).
> 

it is not some tool but your own compiler. yes, there are really nasty things in `C` but as we now run on more and more architectures each and every warning you see ends up a bug on one of them.
as for *readability* that I don´t agree with you there.
(you don´t like `unions` but really like the `!!val` bool conversion - I´d call that preference not readability) 
what we definitely need is less code repetition (as I´ve removed with the `GetFunctionPointerAs` for example, you now have a macro you can adjust to your needs/compiler/whatever instead of hundreds of places in the code you need to adjust - and a point for you, that macro still needs some refinement, but working on that.)

as for readability, I´m open to making it *readable*, but that does not mean to fall back to the first thing that comes to mind (lets reuse that error prone concept because been doing so forever)
 
> So I'm calling for an intermediate position where there's:
> 
>     * more relaxed rules
> 
so you do the check after each pull request manually?

>     * more human decisions with applying these rules instead of automating them
> 
so you do the check after each pull request manually?

>     * perhaps have more linter/checker "shut up tags" in the code when the tool is at fault
> 
already there (and being improved), but we´re talking about valid complaints of the compiler/linter here

>     * more discussions and consensus elaboration about these subjects
> 
> 
> (figuring that this way too long and should have been a discussion instead of a PR comment)
agree, we need to do some talk again, better sooner than later ^^

hardening:
`!!` is used by many projects including gnome and gtk (and very well defined behaviour for the compiler and C standard).

I feel like some other OSS projects have a much larger audience / targets than FreeRDP, but some code without all these tricks against readability, I'm wondering what we're missing here...

akallabeth:
> `!!` is used by many projects including gnome and gtk (and very well defined behaviour for the compiler and C standard).
> 
yes, and so is `union` - that is my point ;)

> I feel like some other OSS projects have a much larger audience / targets than FreeRDP, but some code without all these tricks against readability, I'm wondering what we're missing here...

we need to stick to plain `ISO C` for most as we want to support `MSVC` whereas the projects you mention use a lot of compiler specific extensions making life a lot easier.

client/SDL/SDL2/sdl_freerdp.cpp
```diff
@@ -1190,9 +1237,13 @@ static DWORD WINAPI sdl_client_thread_proc(SdlContext* sdl)
 	BOOL rc = freerdp_connect(instance);
 
 	rdpContext* context = sdl->context();
+	WINPR_ASSERT(context);
+
 	rdpSettings* settings = context->settings;
 	WINPR_ASSERT(settings);
 
+	ScopeGuard guard([&]() { sdl_client_cleanup(sdl, exit_code, error_msg); });
```
akallabeth, line 1245: @hardening hmm, this here might indeed need another refactoring.
thinking about it it would be a lot better to encapsulate the stuff in a function and call them one after the other.
e.g. the whole code doing `freerdp_connect` in one, `while (!freerdp_shall_disconnect_context(context))` in another and then before returning calling `sdl_client_cleanup`

so no need for a `ScopeGuard` any more and even better readable than before.

akallabeth:
@hardening updated the pr, now it is indeed easier to read and we don´t need the `ScopeGuard` for the thread loop anymore while still being safe with regards to resource cleanup.

PR #10618, opened 2024-09-17
[core,proxy] fix nonblocking BIO reads

akallabeth:
In case of non-blocking BIO layers the proxy read functions bailed out with an error. Retry reading in that case unless the TcpConnectTimeout is exceeded


mfleisz (approved):
lgtm

PR #10619, opened 2024-09-17
[clang-tidy] deactivate some C++ checks not fitting

akallabeth:
* cppcoreguidelines-avoid-c-arrays: most of our code is in C, so the C++ components must use these to call the library functions
* modernize-pass-by-value: not compatible with the lowest supported C++ standard of the project

@hardening you probably will love this one :grin:


hardening (approved):

hardening:
And merged !

PR #10620, opened 2024-09-17
[clang-tidy] disable cppcoreguidelines-pro-bounds-pointer-arithmetic

akallabeth:
the check is not really helpful as all it points out is use of argc/argv or some handling of C API stuff


hardening (approved):

akallabeth:
forgot the `hicpp` check doing similar stuff, updated.

hardening (approved):

PR #10622, opened 2024-09-17
[warnings] fix shorten-64-to-32

akallabeth:


channels/cliprdr/cliprdr_common.c
```diff
@@ -403,14 +403,13 @@ UINT cliprdr_read_format_list(wStream* s, CLIPRDR_FORMAT_LIST* formatList, BOOL
 	}
 	else if (!useLongFormatNames)
 	{
-		const size_t cap = Stream_Capacity(sub1);
-		formatList->numFormats = (cap / 36ULL);
-
-		if ((36ULL * formatList->numFormats) != cap)
+		const size_t cap = Stream_Capacity(sub1) / 36ULL;
+		if (cap > UINT32_MAX)
```
hardening, line 407: I'd be really curious to know in which case cap could be larger than `UINT32_MAX`, that would mean that `Stream_Capacity(sub1)` would be bigger than `UINT32_MAX * 36`, that seems a bit too much for `sub1` ...

PR #10623, opened 2024-09-17
Warn fixes stream

akallabeth:


hardening:
All the `va_list ap` thing is some of these unnecessary initializations. Let's drop that clang-tidy rule that produces so many false positives and instead of spending energy treating false positives, let's find the correct tool to correctly track uninitialized variables.

akallabeth:
> All the `va_list ap` thing is some of these unnecessary initializations. Let's drop that clang-tidy rule that produces so many false positives and instead of spending energy treating false positives, let's find the correct tool to correctly track uninitialized variables.

no, these are not false positives.
the 64bit to 32bit reduction must be checked (assert or check before cast)

akallabeth:
@hardening moved the initializer stuff to #10635 to have this pr solely for changes to the stream module.

mfleisz (commented):
Why do we need all these new outer casts?
I.e. `UINT16` + `UINT16` will yield a `UINT16` - the cast of the results to `UINT16` again seems unnecessary

akallabeth:
> Why do we need all these new outer casts? I.e. `UINT16` + `UINT16` will yield a `UINT16` - the cast of the results to `UINT16` again seems unnecessary

because the result of a `<< 8` is an integer which needs to be casted back again. (yea, annoying)

akallabeth:
> Why do we need all these new outer casts? I.e. `UINT16` + `UINT16` will yield a `UINT16` - the cast of the results to `UINT16` again seems unnecessary

@mfleisz ok, found a better way.
just do the shift on the result variable and `or` in the new bytes.

mfleisz (commented):
Looks much better now - one more minor comment: since all the `read_X_be` and `read_X_le` functions basically do the same it might be a good idea to refactor this logic out into a separate function and just call it with the correct type size.

akallabeth:
> Looks much better now - one more minor comment: since all the `read_X_be` and `read_X_le` functions basically do the same it might be a good idea to refactor this logic out into a separate function and just call it with the correct type size.

hmm, on second thought, that would mean we need to always return a `UINT64` even for `UINT8` from that unified function and cast that to the result.
we also can not remove the current inline functions as they are part of a public header.

mfleisz (approved):

PR #10626, opened 2024-09-17
[build,mac] require cJSON

akallabeth:
we build with cJSON, so enforce it during build. Prevents other installed libraries to be considered.


AlexanderProd:
Successfully built on macOS.

PR #10627, opened 2024-09-18
[core,gateway] migrate rpc gateway to use new transport layer.

llyzs:


hardening (approved):
LGTM

libfreerdp/core/gateway/rpc.c
```diff
@@ -707,10 +707,9 @@ static RpcVirtualConnection* rpc_virtual_connection_new(rdpRpc* rpc)
 
 static BOOL rpc_channel_tls_connect(RpcChannel* channel, UINT32 timeout)
 {
-	int sockfd = 0;
 	rdpTls* tls = NULL;
 	int tlsStatus = 0;
-	BIO* socketBio = NULL;
+	BIO* layerBio = NULL;
```
hardening, line 712: perhaps this could be moved line 735 so that we don't have that unnecessary NULL initialisation

mfleisz (approved):
Tested and works when connecting to an RPC gateway 👍

libfreerdp/core/gateway/rpc.c
llyzs: While I was at it I also moved the other variables and add some asserts.

akallabeth (approved):
LGTM

MartinHaimberger:
+1 LGTM

MartinHaimberger (approved):

PR #10630, opened 2024-09-18
[build,c++] add CXXCompilerFlags to (un)set warnings

akallabeth:
* Just like with the C components and CompilerFlags.cmake add a configuration for C++ that disables specific warnings only found in C++ code.
* Disable useless `C++` warnings (old standard compatibility, ...)
* Disable `-Wno-disabled-macro-expansion` as the warning is only providing useless information for our project
* Fix some constants changed for `SDL3 API`


akallabeth:
merging after:
* tested locally and works on ci as well.
* changes only (un)set compiler warnings

PR #10631, opened 2024-09-18
Warn fixes cov

akallabeth:
* Fix type casts, integer expansion is tricky.
* Fix wrong function parameter types
* assert integer narrowing range check
* deactivate `reinterpret-cast` warnings


mfleisz (approved):
lgtm

PR #10634, opened 2024-09-18
[utils,smartcard] return after free

akallabeth:
In smartcard_ListReadersA_Call first free up memory and then return in case of an error


hardening (approved):
LGTM

PR #10635, opened 2024-09-18
Initialize variables

akallabeth:


libfreerdp/core/client.c
```diff
@@ -706,11 +706,13 @@ void* freerdp_channels_get_static_channel_interface(rdpChannels* channels, const
 
 HANDLE freerdp_channels_get_event_handle(freerdp* instance)
 {
-	HANDLE event = NULL;
-	rdpChannels* channels = NULL;
-	channels = instance->context->channels;
-	event = MessageQueue_Event(channels->queue);
-	return event;
+	WINPR_ASSERT(instance);
```
hardening, line 709: Perhaps we'd better return NULL in that case as it's a public API function, so perhaps `freerdp_channels_get_event_handle(NULL)` should not assert but just return NULL and let the caller handle the situation...

akallabeth: good point

akallabeth:
checked branch again.
only initializations and range checks in this pr.

PR #10637, opened 2024-09-18
[warnings] fix mismatched-dealloc

akallabeth:
Some allocator functions have an error path where the corresponding free function is called. Since the memory in the allocator function was allocated using malloc/calloc the free function does not match. Silence warnings with pragma macros


akallabeth:
merging as:
* pr makes use of `WINPR_PRAGMA_DIAG_IGNORED_MISMATCHED_DEALLOC` in allocator functions consistent
* works locally and on ci

PR #10639, opened 2024-09-18
Kbd layout fix

akallabeth:


akallabeth:
@ondrejholy this fixes a bug that was reported from a fedora package.
namely, it fails to load the keyboard layouts from the JSON file if supplied.
with this patch applied the loading succeeds. (did also add a unit test as this area was uncovered before)

mfleisz (approved):

PR #10641, opened 2024-09-19
replace strcpy/tcscpy/wcscpy strlen/tcslen/wcslen

akallabeth:


winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -699,19 +699,18 @@ static BOOL kerberos_rd_tgt_token(const sspi_gss_data* token, char** target, krb
 		if (len == 0)
 			return TRUE;
 
-		buf = malloc(len);
+		buf = calloc(len, sizeof(char));
 		if (!buf)
```
hardening, line 703: `calloc()` not needed, the job is already done correctly with the usage of `stpcpy` and `*buf = 0`

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -661,13 +661,13 @@ static BOOL kerberos_mk_tgt_token(SecBuffer* buf, int msg_type, char* sname, cha
...
-	WinPrAsn1Decoder dec;
-	WinPrAsn1Decoder dec2;
+	WinPrAsn1Decoder dec = { 0 };
+	WinPrAsn1Decoder dec2 = { 0 };
 	BOOL error = 0;
 	WinPrAsn1_tagId tag = 0;
 	WinPrAsn1_INTEGER val = 0;
 	size_t len = 0;
-	wStream s;
+	wStream s = { 0 };
```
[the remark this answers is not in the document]
akallabeth: part of https://github.com/FreeRDP/FreeRDP/pull/10642 so leave it out here.

winpr/libwinpr/sspi/Kerberos/kerberos.c
akallabeth: reverted, https://github.com/FreeRDP/FreeRDP/pull/10642 does change stuff here too.

winpr/libwinpr/comm/comm.c
```diff
@@ -1078,15 +1078,16 @@ DWORD QueryCommDevice(LPCTSTR lpDeviceName, LPTSTR lpTargetPath, DWORD ucchMax)
...
+	if (size + 2 > ucchMax)
 	{
 		SetLastError(ERROR_INSUFFICIENT_BUFFER);
 		return 0;
 	}
 
-	_tcscpy(lpTargetPath, storedTargetPath);
-	lpTargetPath[_tcslen(storedTargetPath) + 1] = '\0'; /* 2nd final '\0' */
-	return _tcslen(lpTargetPath) + 2;
+	_tcsncpy(lpTargetPath, storedTargetPath, size);
```
mfleisz, line 1088: I think this is a bug. The `lpTargetPath` will only contain the string but not a trailing `\0` here so the final string will only be single and not double `\0` terminated.

winpr/libwinpr/comm/test/TestCommDevice.c
```diff
@@ -64,7 +64,7 @@ static int test_CommDevice(LPCTSTR lpDeviceName, BOOL expectedResult)
 			return FALSE;
 		}
 
-		if (lpTargetPath[_tcslen(lpTargetPath) + 1] != 0)
+		if (lpTargetPath[_tcsnlen(lpTargetPath, ARRAYSIZE(lpTargetPath)) + 1] != 0)
```
mfleisz, line 67: This way the `_tcsnlen` call is useless as it still might cause an out-of-bounds read if the buffer is not correctly terminated.

winpr/libwinpr/environment/test/TestEnvironmentGetEnvironmentStrings.c
```diff
@@ -26,7 +24,7 @@ int TestEnvironmentGetEnvironmentStrings(int argc, char* argv[])
 			_tprintf(_T("test failed: return %d\n"), rc);
 			goto fail;
 		}
-		length = _tcslen(p);
+		const size_t length = _tcsnlen(p, (size_t)rc);
```
mfleisz, line 27: The size specifier for this `_tcsnlen` call is wrong. `p` is in no relation to `rc` (its size is only part of it). Since we do not know or can estimate its size (and `GetEnvironmentStrings` is defined to be correctly terminated) I'd stick to `_tcslen` here.

winpr/libwinpr/comm/test/TestCommDevice.c
akallabeth: good catch.

winpr/libwinpr/environment/test/TestEnvironmentGetEnvironmentStrings.c
akallabeth: @mfleisz `rc` is always `tcslen(p) + 2` (newline + '\0') so the limit does make sense.
this way you get early termination of `tcsnlen` if the terminators are missing - well, then it most likely already failed with `_sntprintf`, but none the less)

mfleisz: strnlen is not a replacement for strlen

mfleisz: This is actually wrong since in this case `rc` will be always greater than the length of `p` so we do not gain anything from using `strnlen`.

I think [this](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strnlen-strnlen-s?view=msvc-170#remarks) gives a pretty good explanation of when to use strnlen - and it begins with "**strnlen** is not a replacement for **strlen**". Applied to our cause I think `strnlen` is misused as we are not limiting the strings length but actually extending it with `rc`.

mfleisz: Additional note: `snprintf` does not count the trailing `\0`

akallabeth: @mfleisz true. did mix up that part, but the rest is still valid (`rc > tcslen(p)` is still true)

winpr/libwinpr/sspi/Negotiate/negotiate.c
```diff
@@ -1407,9 +1407,9 @@ static SECURITY_STATUS SEC_ENTRY negotiate_AcquireCredentialsHandleW(
 		const SecPkg* pkg = MechTable[i].pkg;
 		cred->mech = &MechTable[i];
 
-		if (!kerberos && _tcscmp(pkg->name, KERBEROS_SSP_NAME) == 0)
+		if (!kerberos && _tcsncmp(pkg->name, KERBEROS_SSP_NAME, ARRAYSIZE(KERBEROS_SSP_NAME)) == 0)
```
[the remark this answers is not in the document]
akallabeth: while it adds a little overhead (aka explicitly write the length as argument) it prevents reintroducing these in code paths where we do not want such functions to be used (most of `FreeRDP`)

since it is impossible to distinguish these valid instances from the other ones the only feasible way is to just eradicate them completely and bail error once they are reintroduced.
(and we´re talking compile time overhead here)

winpr/libwinpr/environment/test/TestEnvironmentGetEnvironmentStrings.c
akallabeth: @mfleisz ok, found a way.
https://devblogs.microsoft.com/oldnewthing/20100203-00/?p=15083 limits the whole block to `32768` char.
so, a sensible upper bound for a single variable (although maybe too big) is `32768` char

akallabeth: @mfleisz as for your link, that is exactly the reason I try to eliminate these.
While some instances here might also work with `strlen` et al, most of `FreeRDP` parses data from (untrusted) network sources.
So, eliminating these altogether from our codebase (and we are not talking about a lot of instances here) gives us the freedom to tighten the pr checker further (e.g. fail if such functions are introduced again)

winpr/libwinpr/sspi/Negotiate/negotiate.c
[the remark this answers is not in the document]
hardening: I don't understand your point,  `pkg->name` will always be maintained by us as it's an internal structure, and `KERBEROS_SSP_NAME` is constant. There's no possible misuse in the future. 
More generally, all the len checking string functions pointed by this PR can be useful when it's external data that we may not trust (and we have a way to have max length), but that doesn't seem like a good case when it's for some of our internal data.

Always the same remark with trust within a code units and checks: if the point is "you never know, that doesn't hurt to add that checks", then what are the rules to add the checks ? If we start taking measures like if we don't trust our own internal data, then we'll end up with checking everything possible everywhere. We'll have a meta-FreeRDP inside FreeRDP that checks that FreeRDP does what is expected !

Note: it's not only pushing on the stack the length argument, this n argument also have to be check at every loop turn in `_tcsncmp` (so check for EOS and for n)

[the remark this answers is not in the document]
akallabeth: @hardening point being there are functions from the standard library more error prone than others.
There is a reason such function are deprecated/are warned about by a lot of checkers.

anyway, the changes to kerberos have grown too big for this pr, will split it out in a separate one.

client/common/man/generate_argument_manpage.c
```diff
@@ -27,15 +27,15 @@ static char* resize(char** buffer, size_t* size, size_t increment)
 static char* append(char** buffer, size_t* size, const char* str)
 {
 	const size_t len = strnlen(*buffer, *size);
-	const size_t add = strlen(str);
+	const size_t add = strnlen(str, 128);
```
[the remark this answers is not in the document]
akallabeth: adjusted

winpr/libwinpr/environment/test/TestEnvironmentGetEnvironmentStrings.c
```diff
@@ -7,32 +7,37 @@
...
-
-	p = lpszEnvironmentBlock;
+	LPTCH lpszEnvironmentBlock = GetEnvironmentStrings();
+	if (!lpszEnvironmentBlock)
+		goto fail;
 
+	TCHAR* p = lpszEnvironmentBlock;
 	while (p[0] && p[1])
 	{
+		/* https://devblogs.microsoft.com/oldnewthing/20100203-00/?p=15083 */
```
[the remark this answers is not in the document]
akallabeth: ok, removed.

mfleisz (approved):

PR #10643, opened 2024-09-19
[core,gateway] read leftover http body in the rdg IN channel.

llyzs:
If the HTTP body after the final HTTP request sent in the RDG IN channel remains unprocessed in the network stack, it could confuse the event polling system in some cases.


akallabeth (approved):
LGTM

PR #10644, opened 2024-09-19
Revert "[core,gateway] read leftover http body in the rdg IN channel."

llyzs:
Reverts FreeRDP/FreeRDP#10643


llyzs:
This needs to be reverted because it is apparently breaking the RD2019 gateway

mfleisz (approved):

PR #10648, opened 2024-09-20
[client,sdl] update to current sdl3-api

akallabeth:


akallabeth:
compiles with current `SDL3` 88a01fbc964fbea1203071ea3ca2c6f799d4f571 so merging.
next update will be soon :tm: enough.

PR #10663, opened 2024-09-23
[core,client] add channel poll registration API

akallabeth:
Add a new API that allows channels to register/unregister an event-handle along with a callback function to be called by the RDP main thread.
This allows background processing of channel specifics without the need for a channel specific thread.

### DIFF
--- a/include/freerdp/channels/channels.h
+++ b/include/freerdp/channels/channels.h
@@ -33,6 +33,9 @@ extern "C"
 {
 #endif
 
+	/** @since version 3.9.0 */
+	typedef BOOL (*freerdp_channel_handle_fkt_t)(rdpContext* context, void* userdata);
+
 	FREERDP_API int freerdp_channels_client_load(rdpChannels* channels, rdpSettings* settings,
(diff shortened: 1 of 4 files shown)


hardening (commented):
Nice adding. Everything looks ok for me.
I was wondering about the name `freerdp_channel_poll_fkt_t` as usually with poll in the name you kinda expect some wait operation. Perhaps `freerdp_channel_handler_fkt_t` or `freerdp_channel_event_fkt_t` ?
Not a show stopper anyway.

akallabeth:
> Nice adding. Everything looks ok for me. I was wondering about the name `freerdp_channel_poll_fkt_t` as usually with poll in the name you kinda expect some wait operation. Perhaps `freerdp_channel_handler_fkt_t` or `freerdp_channel_event_fkt_t` ? Not a show stopper anyway.

hmm, yes, you could read it like that (although `poll` does not imply `wait`)
well, lets change it

akallabeth:
@hardening ok, renamed (and now also squashed the commits), feel free to merge :grin:

PR #10668, opened 2024-09-24
[cmake,ffmpeg] fix version detection

akallabeth:
If not detected by pkg-config fall back on header file parsing to extract the version of the FFmpeg component.

### DIFF
--- a/cmake/FindFFmpeg.cmake
+++ b/cmake/FindFFmpeg.cmake
@@ -35,6 +35,7 @@
 # SPDX-FileCopyrightText: 2008 Alexander Neundorf <neundorf@kde.org>
 # SPDX-FileCopyrightText: 2011 Michael Jansen <kde@michael-jansen.biz>
 # SPDX-FileCopyrightText: 2021 Stefan Brüns <stefan.bruens@rwth-aachen.de>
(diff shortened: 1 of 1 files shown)


cmake/FindFFmpeg.cmake
```diff
@@ -76,15 +77,15 @@ endmacro()
...
+    if(NOT WIN32)
     # use pkg-config to get the directories and then use these values
     # in the FIND_PATH() and FIND_LIBRARY() calls
     find_package(PkgConfig)
 
     if(PKG_CONFIG_FOUND)
       pkg_check_modules(PC_${_component} QUIET ${_pkgconfig})
     endif()
-  endif(NOT WIN32)
+  endif(WIN32)
```
hardening, line 88: `NOT WIN32` as said by the various checkers

akallabeth:
> That kind of cmake stuff always sound like magic to me, anyway LGTM.

more like `ffmpeg` magic ;)
easy if `pkg-config` is there, otherwise you really need to dig deep to extract what you´re looking for ^^

cmake/FindFFmpeg.cmake
akallabeth: yepp, should be fixed now.

PR #10672, opened 2024-09-24
Multimon warn

akallabeth:
add checks for client settings before establishing a connection.
currently checks the monitor configuration for constraints (one primary, no gaps, no overlaps, ...)


libfreerdp/core/settings.c
```diff
@@ -277,6 +277,151 @@ void freerdp_settings_print_warnings(const rdpSettings* settings)
...
+	}
+
+	if (!havePrimary)
+	{
+		WLog_ERR(TAG, "Monitor configuration does not contain a primary monitor!");
+		rc = FALSE;
+	}
+	if (!foundOrigin)
+	{
+		WLog_ERR(TAG, "Monitor configuration must start at 0/0 for first monitor!");
```
hardening, line 402: I had in mind that the primary monitor must be in `(0,0)` but not necessarily that it is supposed to be the first of the list. IIRC I've seen windows configurations with 3 monitors organized horizontally, with the primary in the middle, and so the first of the list is not at origin, it has a negative x coordinate.

hardening: That seems confirmed in MS-RDPBCGR at 2.2.1.3.6.1. Perhaps we could add a check for the 200x200 minimum size of the virtual desktop as specified in 2.2.1.3.6

akallabeth: will do that.

akallabeth: ok, fixed.

libfreerdp/core/settings.c
```diff
@@ -277,6 +277,199 @@ void freerdp_settings_print_warnings(const rdpSettings* settings)
...
+	for (UINT32 x = 0; x < count; x++)
+	{
+		const rdpMonitor* monitor =
+		    freerdp_settings_get_pointer_array(settings, FreeRDP_MonitorDefArray, x);
+		if (monitor->x < bounds.x)
+		{
+			bounds.x = monitor->x;
+			bounds.width += monitor->width;
+		}
+		if (monitor->x >= bounds.x + bounds.width)
```
mfleisz, line 395: I don't think this condition (and the y equivalent) is correct i.e.:

image

The second monitor's Y (40) won't be larger than the first monitor's Y + height (i.e. 0 + 800) so the bounding rectangle keeps the size of just the first monitor.

mfleisz (commented):
Apart from my remark I think everything else looks good

libfreerdp/core/settings.c
```diff
@@ -277,6 +277,199 @@ void freerdp_settings_print_warnings(const rdpSettings* settings)
...
+		}
+	}
+	return FALSE;
+}
+
+static BOOL monitor_has_gaps(const rdpSettings* settings, UINT32 start, UINT32 count,
+                             const rdpMonitor* compare)
+{
+	BOOL hasNeighbor = FALSE;
+	for (UINT32 x = 0; x < count; x++)
```
hardening, line 313: can we loop with `i` (or `j`) and not `x` as when quickly reading I though that was a variable for the x axis.

libfreerdp/core/settings.c
```diff
@@ -277,6 +277,199 @@ void freerdp_settings_print_warnings(const rdpSettings* settings)
...
+	if (!hasNeighbor)
+	{
+		WLog_ERR(TAG,
+		         "Monitor configuration has gaps! Monitor %" PRIu32 " does not have any neighbor",
+		         start);
+	}
+	return !hasNeighbor;
+}
+
+static BOOL freerdp_settings_client_monitors_have_gaps(const rdpSettings* settings)
```
hardening, line 339: I may be wrong but to check that you can also add the rects of monitors to a `REGION16` and look at the intersection with its bounding rect, if the result is the same region, you don't have gaps.

akallabeth: @hardening implemented now with:
1. a function `align_rects` which checks that either left, top, bottom or right of two monitors are are aligned
2. `dijkstra` to build a graph from monitor 0 and check all others are reachable.

mfleisz (commented):
Codewise looks good
IMO some comments/explanation would be nice - I.e. I don't really know what the function `dijkstra` is doing/returning without spending quite some time figuring out what the code actually does.

PR #10676, opened 2024-09-25
Cmake file to hex

akallabeth:
* replace the custom build binary with a `CMake` script (read file as hex, transform to `C` style `0x12, 0x34, ..` contents and pass that to `configure_file` into a prepared template
* use this for `SDL` client compiled resources
* use this for opencl sources instead of `STRINGIFY`


akallabeth:
* last addition (better comment in template files to find source file for generated ones)
* properly converts any file to a C hex array

PR #10679, opened 2024-09-25
Rdpei no thread

akallabeth:


channels/rdpei/client/CMakeLists.txt
```diff
@@ -30,3 +30,9 @@ set(${MODULE_PREFIX}_LIBS
 include_directories(..)
 
 add_channel_client_library(${MODULE_PREFIX} ${MODULE_NAME} ${CHANNEL_NAME} TRUE "DVCPluginEntry")
+
+option(RDPEI_USE_CHANNEL_THREAD "use a thread to process RDPEI channel events" OFF)
```
mfleisz, line 34: Since we already have `FreeRDP_SynchronousDynamicChannels` (and the thread-less version makes only sense in that scenario) I'd check for that setting instead of introducing another build option.

channels/rdpei/client/rdpei_main.c
```diff
@@ -410,18 +427,55 @@ static UINT rdpei_add_pen_frame(RdpeiClientContext* context)
...
+
+	EnterCriticalSection(&rdpei->lock);
+
+	const UINT64 now = GetTickCount64();
+
+	/* Send an event every ~20ms */
+	if ((now < rdpei->lastPollEventTime) || (now - rdpei->lastPollEventTime < 20ULL))
+	{
+		LeaveCriticalSection(&rdpei->lock);
+		Sleep(1);
```
mfleisz, line 456: Don't really get the need for this Sleep?
I guess it is for the single-threaded callback but do we really want to yield the whole rdp/channel thread here?

channels/rdpei/client/rdpei_main.c
```diff
@@ -410,18 +427,55 @@ static UINT rdpei_add_pen_frame(RdpeiClientContext* context)
...
 	return rdpei_add_pen_frame(context);
 }
 
+static BOOL rdpei_poll_run(rdpContext* context, void* userdata)
+{
+	RDPEI_PLUGIN* rdpei = userdata;
+	WINPR_ASSERT(rdpei);
+	WINPR_ASSERT(context);
+
+	EnterCriticalSection(&rdpei->lock);
```
mfleisz, line 448: I'd suggest moving the CS locking/unlocking out into the thead proc, guarding the call to `rdpei_poll_run`

akallabeth: good idea, splitting this into `rdp_poll_run` and `rdp_poll_run_unlocked`

channels/rdpei/client/rdpei_main.c
akallabeth: yes, true. need to think about a proper way to tell the main loop to yield.

channels/rdpei/client/CMakeLists.txt
akallabeth: done.

akallabeth:
@freerdp-bot test

channels/rdpei/client/rdpei_main.c
```diff
@@ -410,18 +428,58 @@ static UINT rdpei_add_pen_frame(RdpeiClientContext* context)
...
 	{
-		WLog_ERR(TAG, "rdpei_add_frame failed with error %" PRIu32 "!", error);
+		WLog_Print(log, WLOG_ERROR, "rdpei_add_frame failed with error %" PRIu32 "!", error);
 		return error;
 	}
 
 	return rdpei_add_pen_frame(context);
 }
 
+static BOOL rdpei_poll_run_unlocked(rdpContext* context, void* userdata)
```
mfleisz, line 443: Actually I'd just have `rdpei_poll_run` and protect the call in the thread with `Enter/LeaveCriticalSection`.
Making two versions with _lock, _unlock or _nothing_ just increases the chance of picking the wrong one (as is currently done in the `freerdp_client_channel_register` call).

channels/rdpei/client/rdpei_main.c
```diff
@@ -84,9 +86,13 @@ typedef struct
 
 	CRITICAL_SECTION lock;
 	rdpContext* rdpcontext;
+
+	BOOL async;
 	HANDLE thread;
+
 	HANDLE event;
 	BOOL running;
+	UINT64 lastPollEventTime;
```
mfleisz, line 95: Just a minor comment - wrt struct packing the current layout might not be optimal - this would be better:

```
...
HANDLE event;
UINT64 lastPollEventTime;
BOOL running;
BOOL async;
```

channels/rdpei/client/rdpei_main.c
akallabeth: @mfleisz why do you think locking is not required for the `freerdp_client_channel_register` call?

mfleisz: @akallabeth sorry - thought the CS is only required due to the thread created in the channel - but of course it's not! 😅

channels/rdpei/client/rdpei_main.c
akallabeth: done

mfleisz (approved):

PR #10680, opened 2024-09-25
[core,connection] handle message channel in state

akallabeth:
When in demand active it is possible that we can receive message channel messages. In case we receive one switch parsing to the appropriate handlers.

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -1264,6 +1264,21 @@ state_run_t rdp_client_connect_demand_active(rdpRdp* rdp, wStream* s)
 	if (freerdp_shall_disconnect_context(rdp->context))
 		return STATE_RUN_QUIT_SESSION;
 
+	if (rdp->mcs->messageChannelId && (channelId == rdp->mcs->messageChannelId))
+	{
+		UINT16 securityFlags = 0;
+		if (!rdp_read_security_header(rdp, s, &securityFlags, NULL))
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

libfreerdp/core/connection.c
```diff
@@ -1264,6 +1264,21 @@ state_run_t rdp_client_connect_demand_active(rdpRdp* rdp, wStream* s)
 	if (freerdp_shall_disconnect_context(rdp->context))
 		return STATE_RUN_QUIT_SESSION;
 
+	if (rdp->mcs->messageChannelId && (channelId == rdp->mcs->messageChannelId))
```
hardening, line 1267: Looks ok for me, but don't we need to mutualize that, as I guess that we might get some inflight packets in other states, and also get some different packets ?

akallabeth: @hardening good point. waiting for feedback, and if yes wrap this into a common function.

jens-maus:
@akallabeth I checked the sources of this PR and now a connection redirection is possible against our connection broker. So yes, this PR seems to solve the issues we were having with xfreerdp3 lately. Good job!

libfreerdp/core/connection.c
akallabeth: ok, done, refactored this where possible.

akallabeth:
@freerdp-bot test please

PR #10684, opened 2024-09-26
[core,mcs] unify message channel handling

akallabeth:
@jens-maus I´ve merged #10680 without the unification changes, so your issue is fixed.

could you be so kind and also give this pr a try?
I think I got all instances now and handle the message channel the same in all places (without double RDP encryption/decryption)


jens-maus:
Thanks @akallabeth, I tested this PR and it is working now.

PR #10685, opened 2024-09-26
Adding .rdp file support to Android.

nkotixwolf:
Currently you cannot pass .rdp files to FreeRDP libraries in Android. This feature adds the support for this.

How To Test
- Do not set the new variable entitled rdpFileName under Bookmark Base and it will function the same
- Set the new variable of Bookmark Base entitled rdpFileName to a proper .rdp file on your Android device that your application has access to and it will successfully initiate an RDP connection using the .rdp file.

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/domain/BookmarkBase.java
+++ b/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/domain/BookmarkBase.java
@@ -40,6 +40,7 @@ public class BookmarkBase implements Parcelable, Cloneable
 	protected int type;
 	private long id;
 	private String label;
+	private String rdpFileName;
 	private String username;
 	private String password;
 	private String domain;
@@ -53,6 +54,7 @@ public class BookmarkBase implements Parcelable, Cloneable
 		type = parcel.readInt();
 		id = parcel.readLong();
 		label = parcel.readString();
+		rdpFileName = parcel.readString();
 		username = parcel.readString();
 		password = parcel.readString();
 		domain = parcel.readString();
@@ -73,6 +75,7 @@ public class BookmarkBase implements Parcelable, Cloneable
 		type = TYPE_INVALID;
 		id = -1;
 		label = "";
+		rdpFileName = null;
 		username = "";
 		password = "";
 		domain = "";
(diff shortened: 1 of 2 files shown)


hardening:
@freerdp-bot got and check that with a test please

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/domain/BookmarkBase.java
```diff
@@ -293,6 +308,7 @@ public void writeToSharedPreferences(SharedPreferences sharedPrefs)
 	public void readFromSharedPreferences(SharedPreferences sharedPrefs)
 	{
 		label = sharedPrefs.getString("bookmark.label", "");
+		rdpFileName = sharedPrefs.getString("bookmark.rdpFileName", null);
```
akallabeth, line 311: why default to `null` and not an empty string?

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/domain/BookmarkBase.java
```diff
@@ -73,6 +75,7 @@ private void init()
 		type = TYPE_INVALID;
 		id = -1;
 		label = "";
+		rdpFileName = null;
```
akallabeth, line 78: why default to `null` and not an empty string?

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
```diff
@@ -228,6 +228,10 @@ public static boolean setConnectionInfo(Context context, long inst, BookmarkBase
 		ArrayList<String> args = new ArrayList<String>();
 
 		args.add(TAG);
+		if (bookmark.getRdpFileName() != null) {
```
akallabeth, line 231: here you also need to check if the file name is empty, valid, ...

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
```diff
@@ -228,6 +228,10 @@ public static boolean setConnectionInfo(Context context, long inst, BookmarkBase
...
+			args.add(bookmark.getRdpFileName());
+			args.add("/gt:,no-websockets");
```
akallabeth, line 233: this is wrong and has nothing to do with `rdp-file` support

akallabeth (changes requested):
This looks ok for most parts, but:
1. the `RDP` file must most likely be imported (there is no longer filesystem level access on android without going through the appropriate frameworks except for application private data)
2. some changes are unrelated, must be removed.

PR #10688, opened 2024-09-28
Settings fix

akallabeth:
* fix settings validation issues detected by ci
* add unit test for settings validation
* fix missing `strerror_r` detection
* fix a few `clang-tidy` false positives with `// NOLINT`
* fix ci compiler flags setting
* enable `RDPECAM` and `RDPEAR` channels for `clang-tidy`


akallabeth:
tested fix, now no longer leaks or aborts connections with valid monitor setups.

PR #10695, opened 2024-10-01
Cov fixes

akallabeth:
* fix some dead code warnings (one actually being a bug)
* replace `(type)Stream_Buffer(s)` with `Stream_BufferAs(s, type)`


akallabeth:
minor changes + renaming, merging as is.

PR #10706, opened 2024-10-04
Narrow fixes more more more

akallabeth:
* fix signed/unsigned mixups
* add checks/asserts before integer narrow casts

### DIFF
--- a/libfreerdp/core/transport.c
+++ b/libfreerdp/core/transport.c
@@ -1038,7 +1038,10 @@ SSIZE_T transport_parse_pdu(rdpTransport* transport, wStream* s, BOOL* incomplet
 		pduLength = parse_default_mode_pdu(transport, s);
 
 	if (pduLength == 0)
-		return pduLength;
+		return 0;
+
+	if (pduLength > SSIZE_MAX)
+		return -1;
 
 	const size_t len = Stream_Length(s);
 	if (len > pduLength)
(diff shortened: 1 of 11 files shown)


libfreerdp/core/transport.c
```diff
@@ -1079,22 +1084,25 @@ static int transport_default_read_pdu(rdpTransport* transport, wStream* s)
 	{
 		if (!Stream_EnsureCapacity(s, 4))
 			return -1;
-		const int rc = transport_read_layer_bytes(transport, s, 4);
+		const SSIZE_T rc = transport_read_layer_bytes(transport, s, 4);
+		if (rc > INT32_MAX)
```
mfleisz, line 1088: `transport_read_layer_bytes` is supposed to just return -1, 0 or 1 so anything else should be considered an error

libfreerdp/core/transport.c
```diff
@@ -1067,9 +1070,11 @@ static int transport_default_read_pdu(rdpTransport* transport, wStream* s)
 		BYTE c = '\0';
 		do
 		{
-			const int rc = transport_read_layer(transport, &c, 1);
+			const SSIZE_T rc = transport_read_layer(transport, &c, 1);
+			if (rc > INT32_MAX)
```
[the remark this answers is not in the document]
akallabeth: good point, adjusted

libfreerdp/core/transport.c
akallabeth: good point, adjusted

mfleisz (approved):

PR #10708, opened 2024-10-04
Manpage pre generated

akallabeth:
To remove a lot of building dependencies allow to build without manpage generation, use pregenerated ones instead.


hardening (approved):
Seems ok for me, also I've never look at that part previously

PR #10714, opened 2024-10-08
[primitives,sse] fix a bug in sse4.1 primitives copy

akallabeth:


akallabeth:
@jackyzy823 I was able to reproduce this with 3.8.0, but after applying this fix I could no longer do so.
can you confirm?

hardening (approved):
LGTM

mfleisz (approved):

Issue #6508, opened 2020-10-10
Protocol Security Negotiation with RD Gateway and RD Connection Broker


ZeddieXX:
My work has RDS with RD Gateway and RD Connection Broker.  It's been a while since I used freerdp and Remmina to connect, but it used to work.

The servers are built on Windows Server 2016.  When I try to connect using the downloaded RDP file from the RDWeb portal, Remmina would try to connect and after a few minutes, would give this error: Protocol Security Negotiation 

When I try to do the same with xfreerdp, it gives the same error. Since Remmina uses freerdp to connect, I believe the bug resides on freerdp.

the command I used is:
xfreerdp company.rdp /u:myusername /d:workdomain.com /p:mypassword

Output:

[09:56:37:301] [6538:6539] [INFO][com.freerdp.core] - freerdp_connect:freerdp_set_last_error_ex resetting error state
[09:56:37:301] [6538:6539] [INFO][com.freerdp.client.common.cmdline] - loading channelEx rdpdr
[09:56:37:301] [6538:6539] [INFO][com.freerdp.client.common.cmdline] - loading channelEx rdpsnd
[09:56:37:301] [6538:6539] [INFO][com.freerdp.client.common.cmdline] - loading channelEx cliprdr
[09:56:37:301] [6538:6539] [INFO][com.freerdp.client.common.cmdline] - loading channelEx drdynvc
[09:56:38:642] [6538:6539] [INFO][com.freerdp.primitives] - primitives autodetect, using optimized
[AVHWDeviceContext @ 0x7f101c335dc0] Failed to initialise VAAPI connection: -1 (unknown libva error).
[09:56:38:643] [6538:6539] [ERROR][com.freerdp.codec] - Could not initialize hardware decoder, falling back to software: Input/output error
[09:56:38:645] [6538:6539] [INFO][com.freerdp.core.nego] - Detecting if host can be reached locally. - This might take some time.
[09:56:38:645] [6538:6539] [INFO][com.freerdp.core.nego] - To disable auto detection use /gateway-usage-method:direct
[09:56:38:647] [6538:6539] [ERROR][com.freerdp.core] - freerdp_tcp_connect:freerdp_set_last_error_ex ERRCONNECT_DNS_NAME_NOT_FOUND [0x00020005]
[09:56:38:723] [6538:6539] [INFO][com.freerdp.core] - freerdp_tcp_connect:freerdp_set_last_error_ex resetting error state
[...]

PR #10716, opened 2024-10-08
[winpr,sspi] fix kerberos return on get_credentials

akallabeth:
return SEC_E_NO_CREDENTIALS


hardening (approved):
LGTM, can't we refine by looking at the return code of `krb5_get_credentials` ?

akallabeth:
@hardening any refinement welcome, this was just a direct consequence of a bad return value that was part of a bug report, see https://github.com/FreeRDP/FreeRDP/issues/6508

PR #10719, opened 2024-10-08
server: Fix Win32 compilation errors

mfleisz:


akallabeth:
@mfleisz ah, already on that with #10718 
I think we should set that for the whole project as these defines do pose an issue with `C++` code (maybe just limited to the `C++` components?)

akallabeth:
closing in favor of https://github.com/FreeRDP/FreeRDP/pull/10718

PR #10720, opened 2024-10-09
[primitives,copy] remove bgr24 to bgrx32 optimized

akallabeth:
the function was buggy, remove until resolved.

### DIFF
--- a/libfreerdp/primitives/sse/prim_copy_sse4_1.c
+++ b/libfreerdp/primitives/sse/prim_copy_sse4_1.c
@@ -54,6 +54,9 @@ static INLINE pstatus_t sse_image_copy_bgr24_bgrx32(BYTE* WINPR_RESTRICT pDstDat
 
 	const __m128i mask = _mm_set_epi32(0xFF, 0xFF, 0xFF, 0xFF);
 	const SSIZE_T rem = nWidth % 4;
+
+	const size_t align = nSrcStep % 16;
+	const BOOL fast = (align == 0) ? TRUE : (align >= 4 - MIN(4, rem) ? TRUE : FALSE);
 	const SSIZE_T width = nWidth - rem;
 	for (SSIZE_T y = 0; y < nHeight; y++)
 	{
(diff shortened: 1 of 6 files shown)


akallabeth:
@jackyzy823 did reproduce this now on my fedora machine.
Could you try the referenced pull and check if that now works?
If yes I´ll check what exactly is going wrong here.

hardening (commented):
Looks ok

akallabeth:
@hardening just removed the optimized functions, will refine and readd them once I get confirmation this is actually causing the issue.

hardening:
Perhaps we could use primitives of ffmpeg as we kinda need ffmpeg on most platforms to have a decent FreeRDP...

akallabeth:
@hardening not for copy as far as I can see. (and only as option as it is not available on a lot of platforms by default)

jackyzy823:
Yes, this PR seems working. so far so good.

akallabeth:
@jackyzy823 ok, found the bugs ^^
the `avx2` one is tricky, the shuffle function can not cross 16byte boundaries, so we needed a workaround for that.

libfreerdp/primitives/sse/prim_copy_sse4_1.c
```diff
@@ -63,15 +67,20 @@ static INLINE pstatus_t sse_image_copy_bgr24_bgrx32(BYTE* WINPR_RESTRICT pDstDat
...
-			_mm_storeu_si128(dst, d0);
+			for (; x < width; x += 4)
+			{
+				const __m128i* src = (const __m128i*)&srcLine[(x + nXSrc) * srcByte];
+				__m128i* dst = (__m128i*)&dstLine[(x + nXDst) * dstByte];
+				const __m128i s0 = _mm_loadu_si128(src);
+				const __m128i s1 = _mm_shuffle_epi8(s0, smask);
+				const __m128i s2 = _mm_loadu_si128(dst);
+
+				__m128i d0 = _mm_blendv_epi8(s1, s2, mask);
```
jackyzy823, line 81: The thing i missed :( ,  Restoring the original alpha value  :+1:

jackyzy823:
I tested and it is so far so good under both sse and avx2. 

Thanks for your work! :+1:

PR #10724, opened 2024-10-10
[utils,smartcard] ensure '\0' termination for NDR

akallabeth:
NDR read functions do not exactly know which kind of string is read and if it contains a '\0', so ensure that we add one large enought to hold a unicode '\0'


hardening (approved):
Took me some time to find the off by one change, but LGTM

PR #10743, opened 2024-10-15
[crypto,er/der] refactor module

akallabeth:
Since these are part of the official FreeRDP API better rename the functions to have a safe namespace

* Rename functions with a freerdp_ prefix
* Fix function arguments, use size_t


hardening:
I have nothing against this PR even if we shall keep in mind that all this stuff could be replaced by using the winpr asn1 module. We could even have benefits doing so as the ASN1 encoders have the capacity of avoiding copies when you build the packet (as headers length depends of the length of inner content).

akallabeth:
@hardening oh, good point.

akallabeth:
@hardening do you remember why this was made public API?

PR #10748, opened 2024-10-16
Static assert

akallabeth:
`C11` and `C23` provide `_Static_assert` and `static_assert` so lets use that.
also, `C++11` and `C++17` provide the same

### DIFF
--- a/client/X11/xf_keyboard.c
+++ b/client/X11/xf_keyboard.c
@@ -23,7 +23,9 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdint.h>
 
+#include <winpr/assert.h>
 #include <winpr/crt.h>
 #include <winpr/path.h>
 #include <winpr/assert.h>
@@ -191,6 +193,7 @@ void xf_keyboard_release_all_keypress(xfContext* xfc)
 {
 	WINPR_ASSERT(xfc);
 
+	WINPR_STATIC_ASSERT(ARRAYSIZE(xfc->KeyboardState) <= UINT32_MAX);
 	for (size_t keycode = 0; keycode < ARRAYSIZE(xfc->KeyboardState); keycode++)
 	{
 		if (xfc->KeyboardState[keycode])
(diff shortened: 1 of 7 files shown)


client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -242,7 +242,7 @@ UINT sdlClip::MonitorReady(CliprdrClientContext* context, const CLIPRDR_MONITOR_
 		return ret;
 
 	clipboard->_sync = true;
-	SDL_ClipboardEvent ev = { SDL_EVENT_CLIPBOARD_UPDATE, 0, 0, false, NULL, 0 };
+	SDL_ClipboardEvent ev = { SDL_EVENT_CLIPBOARD_UPDATE, 0, 0, false, 0, 0 };
```
hardening, line 245: AFAICT this seems wrong, this field is a pointer not an integer...

client/SDL/SDL3/sdl_monitor.cpp
```diff
@@ -207,7 +207,7 @@ static BOOL sdl_apply_display_properties(SdlContext* sdl)
 		float vdpi = dpi;
 		SDL_Rect rect = {};
 
-		if (SDL_GetDisplayBounds(*id, &rect) < 0)
+		if (!SDL_GetDisplayBounds(*id, &rect))
 			return FALSE;
```
hardening, line 211: good catch I also had it here but forgot to push it

client/SDL/SDL3/sdl_clip.cpp
akallabeth: ok, the order is wrong, correct.
```
typedef struct SDL_ClipboardEvent
 {
     SDL_EventType type; /**< SDL_EVENT_CLIPBOARD_UPDATE */
     Uint32 reserved;
     Uint64 timestamp;   /**< In nanoseconds, populated using SDL_GetTicksNS() */
     bool owner;       /**< are we owning the clipboard (internal update) */
     Sint32 n_mime_types;     /**< number of mime types */
     const char **mime_types; /**< current mime types */
 } SDL_ClipboardEvent;
 ```

mfleisz (approved):

PR #10768, opened 2024-10-22
[winpr,stream] assert Stream_Write_(U)INT[8|16|32] ranges

akallabeth:
* Split `Stream_Write_*` into a define that checks the input variable for range violations and an `Stream_Write_*_unchecked` `static inline` function that actually writes.
* The limits are asserted so invalid values can be found during runtime


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

pnowack (approved):
This looks good to me

PR #10772, opened 2024-10-22
[warnings] unify CommandLineParseCommaSeparatedValues

akallabeth:
CommandLineParseCommaSeparatedValues does some internal hack to allow the allocated char** to be cleaned up by free.
This enforces compiler warnings as the pointer types do not match. Use a union to access the allocated memory in all types required (char*, char**, const char**)


hardening (approved):
Seems ok

PR #10780, opened 2024-10-24
Window fake monitor

akallabeth:
Enforce monitor layout requirements for windowed mode and fullscreen mode before checking the layouts meet multimonitor requirements.

* This way we always send a valid monitor layout in single monitor / windowed mode (start at 0/0 and primary=TRUE)


hardening (approved):
Seems ok.
Nitpicking: in `libfreerdp/core/settings.c` there's no good reason to call setters and getters of setters for ints (drop that and the code would look a lot simpler with less useless error checks)

akallabeth:
@hardening there is a reason for that, once settings are  fully `opaque` getters/setters can add range checks for each value. (so you can´t set an out of range value anymore)

PR #10783, opened 2024-10-25
[ci,nightly] fix missing files in RPM spec

akallabeth:


akallabeth:
successfully build nightly with `./packaging/scripts/create_rpm.sh` on `fedora 40`

PR #10788, opened 2024-10-26
[rdg,websockets] Fix Http Negotiate for Websocket

akarl10:
When using negotiate in non direct NTLM mode the success response contains an aditional token for the authentication layer. Add HTTP_STATUS_SWITCH_PROTOCOLS to the list of valid HTTP status codes where to extract the last auth token


akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #10791, opened 2024-10-28
[client,common] tighten /vmconnect checks

### LINKED ISSUE #5300: vmconnect with different port should not work Freerdp 2.0.0RC4
Steps:-
1:-xfreerdp /cert-ignore -toggle-fullscreen /title:test  /f    /vmconnect:89097b21-6aff-47ee-bd05-3977cb32f78b /d:testdom /u:test /v:x.x.x.x:2170 /p:test123
Result:- connection is established to vm console
Expected :- connection should not be established 
2:- same scenario with mstsc having port 2170 connection is not established.

akallabeth:
Fixes #5300:
* abort if +/-nego is used, /vmconnect overrides these.
* only set custom port if no /port or /v:host:port is in use allowing overriding default port


akallabeth:
@freerdp-bot test

PR #10794, opened 2024-10-28
Internal rc4

akallabeth:
fix fallback support to `RC4`:

* use `winpr_RC4_New` for assistance files
* Add new `winpr_Cipher_NewEx` for explicit length parameters
* add log messages for unsupported ciphers

### DIFF
--- a/libfreerdp/core/rdp.c
+++ b/libfreerdp/core/rdp.c
@@ -2967,14 +2967,27 @@ static void log_build_warn_cipher(rdpRdp* rdp, log_line_t* firstLine, WINPR_CIPH
 
 	char key[WINPR_CIPHER_MAX_KEY_LENGTH] = { 0 };
 	char iv[WINPR_CIPHER_MAX_IV_LENGTH] = { 0 };
-	WINPR_CIPHER_CTX* enc = winpr_Cipher_New(md, WINPR_ENCRYPT, key, iv);
-	WINPR_CIPHER_CTX* dec = winpr_Cipher_New(md, WINPR_DECRYPT, key, iv);
-	if (enc && dec)
+
+	/* RC4 only exists in the compatiblity functions winpr_RC4_*
+	 * winpr_Cipher_* does not support that. */
+	if (md == WINPR_CIPHER_ARC4_128)
 	{
-		haveCipher = TRUE;
+		WINPR_RC4_CTX* enc = winpr_RC4_New(key, sizeof(key));
+		haveCipher = enc != NULL;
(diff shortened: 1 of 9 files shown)


libfreerdp/core/rdp.c
```diff
@@ -2967,14 +2967,27 @@ static void log_build_warn_cipher(rdpRdp* rdp, log_line_t* firstLine, WINPR_CIPH
 
 	char key[WINPR_CIPHER_MAX_KEY_LENGTH] = { 0 };
 	char iv[WINPR_CIPHER_MAX_IV_LENGTH] = { 0 };
-	WINPR_CIPHER_CTX* enc = winpr_Cipher_New(md, WINPR_ENCRYPT, key, iv);
-	WINPR_CIPHER_CTX* dec = winpr_Cipher_New(md, WINPR_DECRYPT, key, iv);
-	if (enc && dec)
+
+	/* RC4 only exists in the compatiblity functions winpr_RC4_*
+	 * winpr_Cipyer_* does not support that. */
```
hardening, line 2972: typo should be `winpr_Cipher_*`

winpr/include/winpr/custom-crypto.h
```diff
@@ -265,9 +265,15 @@ extern "C"
...
 	WINPR_ATTR_MALLOC(winpr_Cipher_Free, 1)
-	WINPR_API WINPR_CIPHER_CTX* winpr_Cipher_New(WINPR_CIPHER_TYPE cipher,
-	                                             WINPR_CRYPTO_OPERATION op, const void* key,
-	                                             const void* iv);
+	WINPR_API WINPR_DEPRECATED_VAR("[since 3.10.0] use winpr_Cipher_NewEx",
+	                               WINPR_CIPHER_CTX* winpr_Cipher_New(WINPR_CIPHER_TYPE cipher,
+	                                                                  WINPR_CRYPTO_OPERATION op,
+	                                                                  const void* key,
+	                                                                  const void* iv));
+	WINPR_ATTR_MALLOC(winpr_Cipher_Free, 1)
```
hardening, line 273: Perhaps you should add the documentation so that we know that it's available since 3.10..

hardening (approved):
Seems ok even if it's not clear to me why we need `winpr_Cipher_NewEx`

winpr/include/winpr/custom-crypto.h
akallabeth: it was not available before too (well, only if you used old OpenSSL)

PR #10803, opened 2024-10-31
Fix more warnings

akallabeth:
* add `WINPR_ASSERT_AT` to provide custom locations (if the `assert` is called from an inline function or similar)
* add order value range checks (for the fuzzer) asserting out of range values
* add some compiler warnings


akallabeth:
@mfleisz @hardening rebased and added a unit test, please review.
will merge next week.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #10817, opened 2024-11-04
[core,primitive] fix rgb primitive copy without conversion.

llyzs:
Use existing BGRx copy function to support RGBx copy because there is no conversion and the color order does not actually matter. This change will preserve the alpha channel when using RGBx as surface pixel format.


mfleisz (approved):

akallabeth:
@mfleisz please set a milestone before merging (helps in tracking this later on)

PR #10822, opened 2024-11-05
[core,rdstls] do silent password check

akallabeth:
In rdstls_parse_pdu_data_type do not use logging stream functions. This function might be called with incomplete data so avoid spamming the log.


akallabeth:
@pnowack you´re lucky, the log you posted in https://matrix.to/#/!qYDoCYrKpfPzpmxxyX:matrix.org/$OYEa7Muq96pAQGTpU0FrowFQk-F5W3o2KrvsXI44-pM?via=matrix.org&via=tchncs.de&via=gitter.im contains the line of the file.

replaced with a silent version of the function so that the incomplete PDU does not spam the log.

hardening (approved):

PR #10823, opened 2024-11-05
[channels,rdpdr] send only used caps to server

akallabeth:
When using RDPDR only send capabilities to server we have activated, e.g. smartcard, drive, printer, port (serial or parallel) redirection

### DIFF
--- a/channels/rdpdr/client/rdpdr_capabilities.c
+++ b/channels/rdpdr/client/rdpdr_capabilities.c
@@ -30,6 +30,7 @@
 #include <winpr/crt.h>
 #include <winpr/stream.h>
 
+#include <freerdp/freerdp.h>
 #include <freerdp/utils/rdpdr_utils.h>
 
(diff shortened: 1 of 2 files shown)


akallabeth:
@gbhauck would be great if you could test this pr.
now the client no longer announces e.g. `smartcard` capability if `/smartcard` is not part of the command line (and it was not enabled as requirement of some other argument)

would be interesting to know if this is sufficient for your configuration to work out of the box or if not which error you receive.

gbhauck:
> @gbhauck would be great if you could test this pr. now the client no longer announces e.g. `smartcard` capability if `/smartcard` is not part of the command line (and it was not enabled as requirement of some other argument)
> 
> would be interesting to know if this is sufficient for your configuration to work out of the box or if not which error you receive.

I'm still on vacation :-)
Thanks for the work- I will try to test it next weekend !

PR #10827, opened 2024-11-06
[core,rdstls] use less verbose logging

akallabeth:
the rdstls_recv function is called with possibly incomplete data, do not log length check failures


pnowack (approved):
This works fine (tested with g-r-d (Remote Login)). The code changes also look good to me.

PR #10831, opened 2024-11-07
Winpr path

akallabeth:
* Add https://github.com/likle/cwalk for path operations 
* Add `Cwalk` as automatic external download to the build (if no checkout is found in `external/cwalk`)
* Add new functions to normalize paths and check if something is root of


akallabeth:
@freerdp-bot test

PR #10836, opened 2024-11-08
Drive backend2

akallabeth:
Depends on #10831


hardening:
Seems very promising, I'll review it !

akallabeth:
@hardening this one is ready, but waiting for a review on #10831 (it changes our dependencies, so do want some eyes on that)

PR #10849, opened 2024-11-14
[core]: Fix some issues with RAIL window state order logging

mfleisz:
This commit fixes following issues when logging window state orders:

- Always print Window and Owner Ids as integer numbers
- Log all window styles (including extended styles)
- Log window title (very useful to identify the window in the log)

### DIFF
--- a/libfreerdp/core/window.c
+++ b/libfreerdp/core/window.c
@@ -93,6 +93,15 @@ BOOL utf8_string_to_rail_string(const char* string, RAIL_UNICODE_STRING* unicode
 	return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


libfreerdp/core/window.c
```diff
@@ -93,6 +93,15 @@ BOOL utf8_string_to_rail_string(const char* string, RAIL_UNICODE_STRING* unicode
 	return TRUE;
 }
 
+static char* rail_string_to_utf8_string(const RAIL_UNICODE_STRING* unicode_string)
+{
+	WINPR_ASSERT(unicode_string);
+
+	size_t outLen = 0;
```
akallabeth, line 100: can be `NULL` if unused

akallabeth (approved):
LGTM

PR #10850, opened 2024-11-14
Pragma warn

akallabeth:
* only define `coverity` pragma when `COVERITY_BUILD` is defined
* fix a warning


akallabeth:
@freerdp-bot test

PR #10853, opened 2024-11-14
Cmake multiconfig2

akallabeth:
* use modern `CMake` options instead of old 2.x syntax
* move common parts to separate `cmake` files and include them (easier to maintain)


akallabeth:
@freerdp-bot test

PR #10861, opened 2024-11-14
do not always return FALSE when gdi is NULL

cbr-rc:
In some cases where gdi is not instantiated, this always returns false. The desired behavior when there is no gdi is to return earlier without failing all the time.


akallabeth:
@freerdp-bot test

PR #10862, opened 2024-11-15
[winpr,environment] implement threadsafe env

akallabeth:
Implement thread save versions of getenv, setenv, putenv, clearenv and unsetenv


hardening:
The problem with these changes is that if the env is modified externally we don't see the changes in the winpr functions. That may end-up with headaches for people using FreeRDP as a library...

akallabeth:
@hardening only if modified after application startup

Issue #8640, opened 2023-01-25
floatbar dragging window underneath it


itayhilan:
Describe the bug
If I try to drug the floatbar it shows a trail of black and the when you click on the mouse it`s also drugs any window/icon/object that is directly underneath it .

To Reproduce

connect to rdp session with full screen and floatbar enable . xfreerdp /v:[IP_Address] /u:testuser /f /floatbar:sticky:on,default:visible,show:fullscreen
After you in the rdp session just maximize any application inside your rdp desktop session ( e.g chrome )
try to drag down the floatbar down or to the sides and you will see black trail and the you click again on the mouse it dragging the application underneath the floatbar .
Screenshots

Application details

FreeRDP version
FreeRDP version 3.0.0-dev (3.0.0) 275741cc75
Build type: Release
Compiler: GNU, 5.4.0
Target architecture: x64

OS version connecting to (server side) - NotouchOS ( Stratodesk ) based on Ubuntu 18.4 desktop

Environment (please complete the following information):

Linux OS
Ubuntu 18.04
Architecture: amd64
Thanks
Itay

ganeshjkale:
verified on freerdp v3.0.0, issue still exist

ganeshjkale:
issue is not reproducible with freerdp v2.11.5.

@akallabeth Issue is easily reproducible with any xrdp version linux server

amazingfate:
After reverting the changes in commit: https://github.com/FreeRDP/FreeRDP/commit/e136444f51b12f45f07b94953d73f04ceb0ce502#diff-e1cd2907c74e9234d8afc9156f13970e5944beae3afee118736ef7680cef67dbL556-R642, issue is gone.
The deleted code is introduced by commit https://github.com/FreeRDP/FreeRDP/commit/d4be1717c23db7845d3344243cf3918cecb7b7ac, which is in the same pr with the commit deleting it.

PR #10882, opened 2024-11-21
[client,X11] ignore mouse events not originating in session window

amazingfate:
Fixes https://github.com/FreeRDP/FreeRDP/issues/8640
This is a relaunch of d4be1717c23db7845d3344243cf3918cecb7b7ac, which was deleted by e136444f51b12f45f07b94953d73f04ceb0ce502.


akallabeth (approved):
LGTM

would add some comment there as the code is not obvious (so it does not get dropped by accient again)

we´ve also switched to more modern `C` standard, so you can declare and initialize the variable `w` directly.

PR #10887, opened 2024-11-25
MS-RDPECAM: VAAPI h264 encoding, resolves #10762

### LINKED ISSUE #10762: [MS-RDPECAM] implement hardware accelerated H264 encoding for camera stream
Recently implemented support for [MS-RDPECAM] (https://github.com/FreeRDP/FreeRDP/issues/4971) uses CPU based H264 encoding, which results in high CPU load, especially for 720p and higher resolutions and while running on older systems. One way to address that could be adding hardware acceleration (VAAPI) support for H264 encoding, which seems possible, considering FreeRDP already uses VAAPI for H264 decoding.

Trying to find out if [MS-RDPECAM] camera redirect feature is actually being used and whether it makes sense to implement hardware accelerated H264 encoding.

Please indicate if this feature request makes sense.

oleg0421:
Implement VAAPI hardware accelerated H264 encoding for camera stream, based on ffmpeg built in support, resolves:
https://github.com/FreeRDP/FreeRDP/issues/10762

The performance improvement for xfreerdp on my Intel based system, encoding 1080p camera stream:

80% CPU without VAAPI;
<20% CPU with VAAPI enabled.

To enable, compile with:
```
-DWITH_VAAPI_H264_ENCODING=ON
```
Had to change the new `h264.h` APIs, that were added earlier by main MS-RDPECAM feature:
1. `h264_get_yuv_buffer` function was removed. Source data buffers are now allocated and managed by the user.
2. `h264_compress` now takes those source buffers as new parameters.
This should have no impact, as no other components are using these new functions yet.

Also note, that input raw format for hardware encoder is NV12, while for software encoder it is YUV420P. As it is currently implemented, the user of `h264_compress` needs to know whether VAAPI is used at compile time, in order to supply right input format. Runtime knowledge would require to modify existing APIs such as `h264_context_new`, `h264_context_reset`, which I wanted to avoid.

I had to switch from using `H264_RATECONTROL_VBR` to `H264_RATECONTROL_CQP`. CQP mode provides more consistent visual quality, whether you use software or hardware encoding. Typically, hardware encoder will produce higher bitrate than software encoder with the same CQP setting.

The `WITH_VAAPI_H264_ENCODING` option was tested on Intel based systems only. Unfortunately, I don't have any AMD or Nvidia based system to test with, but in theory it should work there as well.


akallabeth (commented):
Overall LGTM

just the `API` change is not really required and should be avoided.

There is also already a `WITH_VAAPI` option which is more likely what you were looking for with `CMAKE_DEPENDENT_OPTION`. 

we had to disable that for the generic case though as `VAAPI` does not support all resolutions we require for a normal `RDP` session (your use case of camera should be covered though)

include/freerdp/codec/h264.h
```diff
@@ -74,30 +74,20 @@ extern "C"
...
 	 * @param h264 The H264 context to use for compression
+	 * @param pSrcData A pointer to pointers to input buffers (in YUV420P or NV12 format)
+	 * @param pSrcStride A pointer to array of line byte lengths in the input buffers
 	 * @param ppDstData A pointer that will hold the allocated result buffer
 	 * @param pDstSize A pointer for the destination buffer size in bytes
 	 * @return \b >= 0 for success, \b <0 for an error
 	 * @since version 3.6.0
 	 */
-	FREERDP_API INT32 h264_compress(H264_CONTEXT* h264, BYTE** ppDstData, UINT32* pDstSize);
+	FREERDP_API INT32 h264_compress(H264_CONTEXT* h264, const BYTE** WINPR_RESTRICT pSrcData,
```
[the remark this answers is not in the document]
oleg0421: First version of MS-RDPECAM implementation relies on source video buffers for `h264_compress` being allocated in `libfreerdp/codec/h264.c, avc420_ensure_buffer()`. The allocation there (old code) assumes source video format is always YUV420P and has number of planes (3) and stride sizes hardcoded for YUV420P.

With this change, we also have to support NV12 as source video format (because it is being used by hardware encoders). NV12 format has 2 planes and different stride sizes, so the old code can't handle it.

The only solution is to allocate/deallocate source buffers outside of `h264.c` and supply them as input parameters for `h264_compress`. I don't see any problem with that, because:
1. New interface is cleaner, `h264_compress` is now a direct map to `h264->subsystem->Compress()` API.
2. Nothing else inside xfreerdp took advantage of `h264_get_yuv_buffer()` and `h264_compress()` after it was introduced in June, so there's no impact here.

I'm trying to avoid modifying the APIs at all cost, but in this case it wasn't possible.

channels/rdpecam/client/encoding.c
```diff
@@ -354,11 +367,17 @@ static BOOL ecam_encoder_context_init_h264(CameraDeviceStream* stream)
 	                             ecam_encoder_h264_get_max_bitrate(stream)))
 		goto fail;
 
+	/* Using CQP mode for rate control. It produces more comparable quality
+	 * between VAAPI and software encoding than VBR mode
+	 */
 	if (!h264_context_set_option(stream->h264, H264_CONTEXT_OPTION_RATECONTROL,
-	                             H264_RATECONTROL_VBR))
+	                             H264_RATECONTROL_CQP))
```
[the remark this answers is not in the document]
oleg0421: With VBR ratecontrol mode, we have a table of preset bitrates in `channels/rdpecam/client/encoding.c`, for example, for 1080p we have 2700kbps, which worked for software encoding. However, hardware encoders on older Intel platforms produce terrible video quality and require approximately double the bitrate to get to comparable quality.

In contrast, CQP mode ensures a given quality level across different encoders, regardless of hardware specs, etc, sometimes at the cost of higher bitrate. That is why I've decided to switch to CQP mode and propose 26 as a compromise value between quality and relatively low bitrate.

cmake/ConfigOptions.cmake
```diff
@@ -150,6 +150,11 @@ CMAKE_DEPENDENT_OPTION(WITH_VIDEO_FFMPEG "Use FFMPEG for video encoding/decoding
 	"WITH_FFMPEG" OFF)
 CMAKE_DEPENDENT_OPTION(WITH_VAAPI "Use FFMPEG VAAPI" OFF
 	"WITH_VIDEO_FFMPEG" OFF)
+CMAKE_DEPENDENT_OPTION(WITH_VAAPI_H264_ENCODING "Use FFMPEG VAAPI hardware H264 encoding" OFF
```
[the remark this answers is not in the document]
oleg0421: It means `WITH_VAAPI_H264_ENCODING` is dependent on `WITH_VIDEO_FFMPEG` being enabled, because VAAPI H264 encoding is only supported by ffmpeg library (not by openh264).

Also, to answer your concern, seems like `WITH_VAAPI_H264_ENCODING` in addition to existing `WITH_VAAPI` provides more flexibility, because VAAPI can be enabled independently for H264 encoding and decoding.

include/freerdp/codec/h264.h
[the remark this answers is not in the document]
hardening: Can't you have `h264_compressEx` with your new function signature and have `h264_compress` that would call it with the correct arguments, this way we don't change the API...

libfreerdp/codec/h264_ffmpeg.c
```diff
@@ -141,7 +178,11 @@ static BOOL libavcodec_create_encoder(H264_CONTEXT* WINPR_RESTRICT h264)
 		return TRUE;
 
 	libavcodec_destroy_encoder(h264);
+#ifdef WITH_VAAPI_H264_ENCODING
```
hardening, line 181: Here perhaps you should instead have a fallback mechanism, like we look for h264_vaapi if the option is set and if not found we go back the default implementation...

include/freerdp/codec/h264.h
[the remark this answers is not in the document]
akallabeth: @oleg0421 not necessarily unused. we´ve published versions containing these symbols.
you can mark them deprecated so we can ditch them once we make a major version jump again.

libfreerdp/codec/h264_ffmpeg.c
oleg0421: For encoding, providing run time fallback to software encoder (if hardware initialization fails) will require complete new API, including entry points to `h264_ffmpeg` and `h264_openh264` subsystems. Main reason for this is that API user needs to be notified back that fallback happened during initialization in order to feed YUV420P frames to the encoder (instead of NV12).

With currently proposed change, other h264 encoder API users (if any) will fail if `WITH_VAAPI_H264_ENCODING` is defined, because they are unaware they need to feed NV12 frames. My guess is that xfreerdp client today doesn't have any other users for encoder APIs. The server might be the only other user and needs to be compiled without `WITH_VAAPI_H264_ENCODING`.

Another needed API change will be to provide frame size (width, height) to `h264_context_new()`, because hardware encoder context initialization requires width and height. Today context initialization only happens when `h264->subsystem->Compress()` is called for the 1st time, but it's too late to notify the user which frames (YUV420P or NV12) to feed at this point.

Hope this adds some useful info and not confuses you further :)

include/freerdp/codec/h264.h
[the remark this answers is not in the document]
oleg0421: Done

libfreerdp/codec/h264_ffmpeg.c
```diff
@@ -114,6 +114,43 @@ static void libavcodec_destroy_encoder(H264_CONTEXT* WINPR_RESTRICT h264)
 	sys->codecEncoderContext = NULL;
 }
 
+#ifdef WITH_VAAPI_H264_ENCODING
+static int set_hw_frames_ctx(H264_CONTEXT* WINPR_RESTRICT h264)
+{
+	H264_CONTEXT_LIBAVCODEC* sys = (H264_CONTEXT_LIBAVCODEC*)h264->pSystemData;
```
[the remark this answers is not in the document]
oleg0421: Seems not needed: static function, already checked by calling functions

libfreerdp/codec/h264_ffmpeg.c
```diff
@@ -158,7 +199,7 @@ static BOOL libavcodec_create_encoder(H264_CONTEXT* WINPR_RESTRICT h264)
 			break;
 
 		case H264_RATECONTROL_CQP:
-			/* TODO: sys->codecEncoderContext-> = h264->QP; */
+			av_opt_set_int(sys->codecEncoderContext, "qp", h264->QP, AV_OPT_SEARCH_CHILDREN);
```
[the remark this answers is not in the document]
oleg0421: Done

channels/rdpecam/client/encoding.c
```diff
@@ -205,22 +216,21 @@ static BOOL ecam_encoder_compress_h264(CameraDeviceStream* stream, const BYTE* s
...
-
-	/* convert from source format to YUV420P */
+	/* convert from source format to YUV420P or NV12 */
 	if (!ecam_init_sws_context(stream, pixFormat))
 		return FALSE;
 
 	const BYTE* cSrcSlice[4] = { srcSlice[0], srcSlice[1], srcSlice[2], srcSlice[3] };
-	if (sws_scale(stream->sws, cSrcSlice, srcLineSizes, 0, (int)size.height, yuv420pData,
-	              (int*)yuv420pStride) <= 0)
+	if (sws_scale(stream->sws, cSrcSlice, srcLineSizes, 0, (int)size.height, stream->swsOutData,
```
akallabeth, line 224: @oleg0421 hmm, wouldn´t moving this part into the encoder function solve your issue to have different encoding formats?
this way you could just feed native format and then convert to whatever you need internally?

akallabeth: you could even expose a `out_data_free` function to wrap the cleanup calls and not link `ffmpeg` directly this way

cmake/ConfigOptions.cmake
```diff
@@ -150,6 +150,11 @@ CMAKE_DEPENDENT_OPTION(WITH_VIDEO_FFMPEG "Use FFMPEG for video encoding/decoding
...
+	"WITH_VIDEO_FFMPEG" OFF)
+if(WITH_VAAPI_H264_ENCODING)
+	add_definitions("-DWITH_VAAPI_H264_ENCODING")
```
akallabeth, line 156: better use https://cmake.org/cmake/help/latest/command/add_compile_definitions.html

include/freerdp/codec/h264.h
```diff
@@ -96,9 +97,25 @@ extern "C"
...
+	 *
+	 * @param h264 The H264 context to use for compression
+	 * @param pSrcData A pointer to pointers to input buffers (in YUV420P or NV12 format)
+	 * @param pSrcStride A pointer to array of line byte lengths in the input buffers
+	 * @param ppDstData A pointer that will hold the allocated result buffer
+	 * @param pDstSize A pointer for the destination buffer size in bytes
+	 * @return \b >= 0 for success, \b <0 for an error
+	 * @since version
+	 */
+	FREERDP_API INT32 h264_compress_ex(H264_CONTEXT* h264, const BYTE** WINPR_RESTRICT pSrcData,
```
akallabeth, line 115: so, like mentioned before:
I´d suggest moving format conversion from source format to required internal FFMPEG format from channel into this function.
this way we do not need to link the `rdpecam` channel to `ffmpeg` nor `swscale` and can freely implement another backend if we find something better.

only downside is that you need to duplicate the `CAM_MEDIA_FORMAT` enum here to allow these as input formats.

NOTE: would also move the `MJPEG` encoding to such a function

libfreerdp/codec/h264_ffmpeg.c
akallabeth: @oleg0421 see my comments above.
you are right that you need to change something, but it is sufficient to fix the new API here by just providing a source buffer and which format it used.

the extension for the `H264_CONTEXT_SUBSYSTEM` is required, yes, but that is backward compatible (just return an error if the function pointer is not initialized)

include/freerdp/codec/h264.h
oleg0421: @akallabeth : I liked your suggestion to add source raw format type to `h264_compress_ex`, but I'm not sure converting from MJPEG to raw format should also move inside `h264.c`, because it is a separate step not directly related  to h264 encoding. Also I'm not sure `CAM_MEDIA_FORMAT` is the right type to pass this new parameter. I would probably use `enum AVPixelFormat` for that, which is part of `swscale` library and has many more types (including types produced by ffmpeg MJPEG decoder).
Anyway, let me take some time to think about it more and try to come up with something.

akallabeth: @oleg0421 `AVPixelFormat` is not suitable as that is defined in an optional dependent library.

oleg0421: @akallabeth: supporting runtime fallback to software encoder (in case hardware encoder initialization fails) requires  changes of multiple layers, including `h264_ffmpeg.c` and `h264_openh264.c` subsystems, common `h264.c` layer and camera channel's `encoding.c` layer. Unfortunately, I don't currently have enough time to work on that. The `CHANNEL_RDPECAM_CLIENT` is not enabled by default anyway, users need to compile the code to enable it. They might as well try `WITH_VAAPI_H264_ENCODING` option and if it doesn't work on their system, disable it.

akallabeth: does not necessarily require changing all the backends, but as we´re chaning the required `API` to support the channel we should address this well enough so that we don´t need to rework this in a few months.

the point being, if you have the `API` you can add the implementation for others later on.

akallabeth: @oleg0421 just to be clear: I just want the `API` function `h264_compress_ex` to cover all these possibilities as changing the implementation is something we can easily do, but changing function signatures is not.

oleg0421: @akallabeth : I understand your point and have an idea how to support runtime fallback to software encoding:

1. Move `avcodec_find_encoder_by_name("h264_vaapi")` and `av_hwdevice_ctx_create()` to `libavcodec_init()`. If they fail, switch to using software encoder.
2. In other places, rely on `if (sys->hwctx)` check to support both encoders in runtime.
3. Add "get only" option `H264_CONTEXT_OPTION_ENCODER_INPUT_FORMAT` to `H264_CONTEXT_OPTION` enum with these values:
```
typedef enum
{
	H264_ENCODER_INPUT_FORMAT_YUV420P = 0,
	H264_ENCODER_INPUT_FORMAT_NV12,
} H264_ENCODER_INPUT_FORMAT;
```

4. Before using `h264_compress_ex()` API with proposed signature, call:
```
h264_context_get_option(stream->h264, H264_CONTEXT_OPTION_ENCODER_INPUT_FORMAT);
```
which will return the right format, depending on which encoder was initialized, and supply `pSrcData` in right format for `h264_compress_ex()`.

This way, openh264 library will also be supported, because it will always return YUV420P.

akallabeth: @oleg0421 nice idea with the fallback.
as for formats, I´d take the opposite approach, make the internals opaque for the caller and just tell which format the input uses.
This way you can switch internal formats as much as you like if the backend has better support for a different one without the need to modify the caller code.

hardening: > @akallabeth : I understand your point and have an idea how to support runtime fallback to software encoding:
> 
>     1. Move `avcodec_find_encoder_by_name("h264_vaapi")` and `av_hwdevice_ctx_create()` to `libavcodec_init()`. If they fail, switch to using software encoder.
> 
>     2. In other places, rely on `if (sys->hwctx)` check to support both encoders in runtime.
> 

That was exactly what I was meaning with fallback (BTW I already have implemented that in old private projects, it works great)

oleg0421: > [note] the whole `sws_scale` stuff you do you can just ditch and the current (implicit) input expectation is `H264_ENCODER_INPUT_FORMAT_YUV420P` - which works with openh264 as well. for `ffmpeg` you can just do format conversion internally without any modifications to the `S_H264_CONTEXT_SUBSYSTEM` struct

We can only move `sws_scale()` to `h264.c`, but not ditch it, in case we add format type parameter to `h264_compress_ex()`. Some layer still needs to do the conversion from YUYV422 (which cameras typically output) or another camera format to YUV420P or NV12.

We cannot put `sws_scale()` logic inside `h264_ffmpeg.c`, because `h264_openh264.c` needs same logic to convert from camera output  format to YUV420P.

The h264 encoders typically expect YUV420P input for software and NV12 for hardware encoding. It's not typically h264 encoder library responsibility to handle variety of input formats. Also sometimes video processing does up- or down-scaling before encoding, and `sws_scale()` efficiently does it together with format conversion, which is the reason to keep it at current place (`encoding.c`).

akallabeth: @oleg0421 please forget about openh264.
we can easily move it inside and have a `ffmpeg` implementation in the function as first step.
what you are talking about is not required for this pr and out of scope, but no reason not to clean that up.

the drawbacks of having it outside is that you need to not only  link the libraries, you need to know which backend is compiled in, ....

for scaling we have the `freerdp_image_scale` function and optimized functions are something out of scope here as well (and they again must not expose implementation detail to be switchable)

oleg0421: @akallabeth : what about MJPEG decoding part? Also having it  inside `h264_ffmpeg.c` makes little sense. You are thinking more in terms of ffmpeg library dependencies, but we also need to keep modularity in mind. `h264.c` is the unified interface to different h264 encoder/decoder libraries, not a container for everything ffmpeg related.
If you put everything in `h264_ffmpeg.c` now, then we'll have to duplicate the logic later for other h264 encoders.

`freerdp_image_scale` doesn't support any of those YUV formats, only RGB-based, so we need `sw_scale`.

akallabeth: well, you have only two paths:
1. pass  through directly
2. convert to h264

since that is just another input format it can be converted internally

> `freerdp_image_scale` doesn't support any of those YUV formats, only RGB-based, so we need sw_scale.

no, this just illustrates that we have functions abstracting the required functions and (just as your example) the things are currently not implemented due to lack of demand.
not an argument to reinvent the wheel.

oleg0421: @akallabeth : MJPEG is not a raw video format and can't be represented in terms of `BYTE* srcSlice[4]` and `srcLineSizes[4]`, so if you want to support it by `h264_compress_ex`, we need to change parameters to point to whole buffer and provide whole image width / height. Only this way we can get rid of channel's `swscale` and `ffmpeg` libraries dependency, but at the cost of polluting h264 encoder/decoder library and transferring the above dependencies there.

akallabeth: @oleg0421 just like `avc444_compress` then?
that already does image from `RGB` to 2x YUV420 conversion.

akallabeth: @oleg0421 note: encoding/decoding is split:
* `libfreerdp/codec/h264.c` for generic code
* `libfreerdp/codec/h264_ffmpeg.c` for the ffmpeg implementation details 
* ...

oleg0421: > @oleg0421 just like `avc444_compress` then? that already does image from `RGB` to 2x YUV420 conversion.

@akallabeth these functions use primitives such as `ssse3_RGBToYUV420_BGRX_UV`, which use manually written SSE3 optimized instructions. This is beyond my expertise and besides `sws_scale()` does these optimized conversions very efficiently.

oleg0421:
@akallabeth: please consider this new proposal. It relies on old API for encoding, only adds `h264_context_new2` with `BOOL* hwAccel` in / out parameter, allowing the client to: 1) explicitly request hw acceleration; 2) check whether hw acceleration is actually on, so it can use the right input format. This approach allows to keep compatibility with other encoder clients unaware of hw acceleration and supports runtime fallback to software encoding.

As to moving the MJPEG decoding and raw format conversions out of the channel and into h264.c, I tried that but struggled due to
dependencies on ffmpeg and swscale, which make the code difficult to manage. I think once we have solutions to replacing this logic with internal libraries, then it will make the refactoring easier, but for now let's keep it as is. Also note that main encoding API is not changing, only one addition to context initialization.

akallabeth:
@oleg0421 limited access for a few days, will take a little time (don´t have stuff to test around)

channels/rdpecam/client/encoding.c
```diff
@@ -329,7 +331,10 @@ static BOOL ecam_encoder_context_init_h264(CameraDeviceStream* stream)
 	WINPR_ASSERT(stream);
 
 	if (!stream->h264)
-		stream->h264 = h264_context_new(TRUE);
+	{
+		stream->h264hwAccel = TRUE; // request hardware acceleration
+		stream->h264 = h264_context_new2(TRUE, &stream->h264hwAccel);
```
akallabeth, line 336: use `h264_context_get_option` for this

include/freerdp/codec/h264.h
```diff
@@ -126,6 +126,16 @@ extern "C"
...
 
+	/**
+	 * @brief Creates new h264 context
+	 *
+	 * @param Compressor TRUE if used to compress, FALSE if used to decompress
+	 * @param hwAccel in: TRUE to request hw accel, out: TRUE if hw accel is on
+	 * @return new h264 context, NULL for an error
+	 * @since version
+	 */
+	FREERDP_API H264_CONTEXT* h264_context_new2(BOOL Compressor, BOOL* hwAccel);
```
akallabeth, line 137: not required, you have `h264_context_set_option` and `h264_context_get_option`

akallabeth (changes requested):
sorry it took this long to review (holidays and busy with other bugs)

* the `h264_context_new2` is not really a good `API` as you can just query with existing `h264_context_get_option`
* forcing the user of the `API` to convert to different formats is not really a good workflow. As discussed before, the proper way would be to add a new `API` that explicitly tells which format the source data has and letting the encoder function convert internally to avoid pushing the dependencies out of the core library to the channel

include/freerdp/codec/h264.h
oleg0421: @akallabeth: `h264_context_new2` is required to explicitly request hardware acceleration. By default (for those legacy clients using `h264_context_new`) hardware acceleration remains off. This is the only reasonable way to provide compatibility with legacy users of encoding API (either old API or new API introduced by camera feature).

Think of it this way: if we compile hardware acceleration in and enable it for all users, then legacy users of existing API will be broken (because they supply YUV420P format and encoder expects NV12).

Using `h264_context_set_option` to request hardware acceleration is not possible, because the encoder context (hardware or software) is created at the very beginning by `h264_context_new`.

Handling hardware acceleration transparently by accepting YUV420P format from legacy users is not acceptable either because of performance hit due to double conversion (source format -> YUV420P -> NV12) and possible video quality loss due to that.

akallabeth: @oleg0421 no, it is not required at all.
all you need to do is applying the setting if you set the `h264_context_set_option` (e.g. context reset)

as for legacy, well, not really a good argument, you can just update the whole thing (and in practice it will be updated together with the core library as it is shipped in the same package)

oleg0421: @akallabeth : I introduced the option `H264_CONTEXT_OPTION_HW_ACCEL`, as suggested (thanks!)

libfreerdp/codec/h264_mf.c
```diff
@@ -454,6 +454,10 @@ static void mf_uninit(H264_CONTEXT* h264)
 static BOOL mf_init(H264_CONTEXT* h264)
 {
 	HRESULT hr;
+
+	WINPR_ASSERT(h264);
+	h264->hwAccel = FALSE; /* not supported */
```
akallabeth, line 459: not required, context is `calloc` ed

libfreerdp/codec/h264_openh264.c
```diff
@@ -523,6 +523,7 @@ static BOOL openh264_init(H264_CONTEXT* h264)
 	static WelsTraceCallback traceCallback = openh264_trace_callback;
 
 	WINPR_ASSERT(h264);
+	h264->hwAccel = FALSE; /* not supported */
```
akallabeth, line 526: not required, context is `calloc` ed

libfreerdp/codec/h264_mediacodec.c
```diff
@@ -443,6 +443,7 @@ static BOOL mediacodec_init(H264_CONTEXT* h264)
 	media_status_t status;
 
 	WINPR_ASSERT(h264);
+	h264->hwAccel = FALSE; /* not supported */
```
akallabeth, line 446: not requred, context is `calloc` ed

cmake/ConfigOptions.cmake
```diff
@@ -160,6 +160,10 @@ option(WITH_FFMPEG "Enable FFMPEG for audio/video encoding/decoding" ON)
 cmake_dependent_option(WITH_DSP_FFMPEG "Use FFMPEG for audio encoding/decoding" ON "WITH_FFMPEG" OFF)
 cmake_dependent_option(WITH_VIDEO_FFMPEG "Use FFMPEG for video encoding/decoding" ON "WITH_FFMPEG" OFF)
 cmake_dependent_option(WITH_VAAPI "Use FFMPEG VAAPI" OFF "WITH_VIDEO_FFMPEG" OFF)
+cmake_dependent_option(WITH_VAAPI_H264_ENCODING "Use FFMPEG VAAPI hardware H264 encoding" ON "WITH_VIDEO_FFMPEG" OFF)
```
akallabeth, line 163: not quite sure, what is the difference to `WITH_VAAPI` option?

oleg0421: This is for additional flexibility: can compile with either of them on (or both). You've mentioned there were some scenarios when hardware decoding was not working due to non-standard sizes, so you've had to keep it off by default. For camera, the sizes are usually standard so we can have `WITH_VAAPI_H264_ENCODING` on by default.

libfreerdp/codec/h264.c
```diff
@@ -701,16 +725,19 @@ H264_CONTEXT* h264_context_new(BOOL Compressor)
...
 
 	h264->Compressor = Compressor;
+	h264->hwAccel = FALSE;
 	if (Compressor)
-
 	{
 		/* Default compressor settings, may be changed by caller */
 		h264->BitRate = 1000000;
 		h264->FrameRate = 30;
+		/* delay h264_context_init until 1st h264_context_reset */
```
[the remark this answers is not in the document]
oleg0421: `h264_context_init` is where `H264_CONTEXT_SUBSYSTEM`'s `Init` is called. For `h264_context_reset`, there's no special API function in `H264_CONTEXT_SUBSYSTEM`. We can do hardware (or software) device context creation once and we need to postpone calling `Init` until after `h264_context_set_option(stream->h264, H264_CONTEXT_OPTION_HW_ACCEL, TRUE)` is called to indicate client wants to request hw accel.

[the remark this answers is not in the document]
akallabeth: @oleg0421 `h264->subsystem->Uninit(h264);` followed by `h264->subsystem->Init(h264);`?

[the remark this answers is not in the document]
oleg0421: @akallabeth : to call `Uninit` / `Init` every time `h264_context_reset` is called could be performance hit. Existing `h264_context_reset`doesn't do much, except setting picture width / height. The other encoder user (server) may call it quite often (I guess), to encode different screen parts.
On the other hand, we are guaranteed `h264_context_reset` is called at least once, because it sets width / height.

libfreerdp/codec/h264_mf.c
oleg0421: Removed from here

libfreerdp/codec/h264_openh264.c
oleg0421: It's needed, because client may have set it to TRUE (asking for hw accel), but there's no support for it. Moved it to a better place though.

libfreerdp/codec/h264_mediacodec.c
oleg0421: Ok, removed from here, because mediacodec doesn't support encoder.

libfreerdp/codec/h264.c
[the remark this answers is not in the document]
akallabeth: @oleg0421 then you do something wrong.
reset should only be called on resolution change or if hardware support is enabled/disabled.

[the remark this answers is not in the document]
oleg0421: @akallabeth I'm sorry, this is confusing, but please let me try to explain:

`h264_context_reset` doesn't do anything but sets width and height. Camera channel only calls it once.

There are 2 entry points before encoding starts:
`h264_context_new`
`h264_context_reset` (which is mandatory to call for encoder to set width and height)

If we call subsystem's `Init` from `h264_context_new`, then _software_ context will be created by subsystem.
Then if we call `Uninit` / `Init` from `h264_context_reset`, the _software_ context will be deleted and _hardware_ context created. Please see 5 initialization steps at PR description above for more details.

It is possible to call `Uninit` / `Init` from `h264_context_reset`, but it will only slow down video start. This will also imply user can dynamically switch between software and hardware encoders, which is not needed.

Today there's simply no need for subsystem reset, only one time `Init`. If there's a need in the future, we can always implement it.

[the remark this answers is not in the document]
akallabeth: @oleg0421 we´re talking `5ms` here.
I´ve added a small unit test (see https://github.com/akallabeth/FreeRDP/tree/h264-test ) that measures stuff.
* most expensive in initialization is populating the initial H264 encoder related global variables (~400ms), only part of first call to `h264_context_new`
* the init/uninit takes around 5ms

so while you can move the init from `_new` to `_reset` you also need to take care of initialized contexts in `_reset` (e.g. call `_uniinit` before if initialized, everything else is unexpected. (you don´t know the usage of the function as it is part of the `API` and external projects rely on that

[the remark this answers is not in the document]
akallabeth: @oleg0421 ok, did fix a few bugs regarding h264 reset (and encoding) with #11065 
this should now eliminate the 400ms startup delay and the whole H264 uninit/init cycle now constantly takes ~5ms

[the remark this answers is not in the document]
oleg0421: @akallabeth: merged in your changes and tested - everything works, but in the trace:
```
[17:53:02:358] [11598:00002d62] [INFO][com.freerdp.codec] - [libavcodec_init]: Using VAAPI for accelerated H264 decoding
[17:53:02:367] [11598:00002d62] [INFO][com.freerdp.codec] - [libavcodec_init]: Using VAAPI for accelerated H264 decoding
[17:53:02:385] [11598:00002d62] [INFO][com.freerdp.codec] - [libavcodec_init]: Using VAAPI for accelerated H264 decoding
[17:53:02:399] [11598:00002d62] [INFO][com.freerdp.codec] - [libavcodec_init]: Using VAAPI for accelerated H264 decoding
[17:53:02:552] [11598:00002d62] [INFO][com.freerdp.channels.rdpsnd.client] - [rdpsnd_load_device_plugin]: [dynamic] Loaded fake backend for rdpsnd
[17:53:04:253] [11598:00002d62] [INFO][com.freerdp.channels.rdpsnd.client] - [rdpsnd_load_device_plugin]: [dynamic] Loaded fake backend for rdpsnd
[17:53:12:365] [11598:00002d62] [INFO][com.freerdp.codec] - [libavcodec_init]: Using VAAPI for accelerated H264 encoding
[17:53:12:525] [11598:00002d62] [INFO][com.freerdp.channels.rdpecam-v4l.client] - [cam_v4l_stream_start]: Camera format: MJPG, width: 1920, height: 1080, fps: 30/1
[17:53:12:670] [11598:00002d75] [ERROR][com.freerdp.codec] - [libavcodec_compress]: Failed to encode video frame (Resource temporarily unavailable [-11])
[17:53:12:727] [11598:00002d75] [ERROR][com.freerdp.codec] - [libavcodec_compress]: Failed to encode video frame (Resource temporarily unavailable [-11])
[17:53:17:070] [11598:00002d62] [INFO][com.freerdp.codec] - [libavcodec_init]: Using VAAPI for accelerated H264 decoding
[17:53:17:072] [11598:00002d62] [INFO][com.freerdp.codec] - [libavcodec_init]: Using VAAPI for accelerated H264 decoding
 ```

the line "Using VAAPI for accelerated H264 decoding" appears 4 times before camera is started and 2 times after. This means `h264_context_reset` is called by _decoder_ 3 times and 1 time respectively. Probably minor, you decide.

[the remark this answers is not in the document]
Comment (akallabeth) [reply libfreerdp/codec/h264.c]: > the line "Using VAAPI for accelerated H264 decoding" appears 4 times when new RDP session is started and then twice every time new decoder is created. Probably minor, you decide.

this looks like a bug? (if what you said is true) because that would mean you create the encoder 4 times? (not really a big thing though)

File: libfreerdp/codec/h264_mf.c [hunk 17]
```diff
@@ -482,6 +482,7 @@ static BOOL mf_init(H264_CONTEXT* h264)
 
 	if (h264->Compressor)
 	{
+		goto error; /* not supported */
```
akallabeth, line 485: remove, this change has nothing to do with your pull

libfreerdp/codec/h264_openh264.c
```diff
@@ -648,6 +648,7 @@ static BOOL openh264_init(H264_CONTEXT* h264)
 		}
 	}
 
+	h264->hwAccel = FALSE; /* not supported */
```
akallabeth, line 651: context is `calloc` initialized, so no need for this

akallabeth (changes requested):
function LGTM now.

so, for getting this merged, please rebase on master, sqush your commits  (the history is messy) and force push.

libfreerdp/codec/h264_mf.c
oleg0421: Done

libfreerdp/codec/h264.c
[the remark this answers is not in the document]
oleg0421: This is about decoder, not encoder. Just wanted to point out as a side effect of `h264_context_reset` change.

akallabeth (approved):
LGTM

akallabeth:
@oleg0421 ok, looks like the formatting is off, can you run `git clang-format origin/master`, commit & squash the changes and force push again?

oleg0421:
@freerdp-bot test

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@oleg0421 ok, `./scripts/cmake-format.sh --check` still fails, best run that one too (cmake file formatting)
e.g. run `./scripts/cmake-format.sh -f` and do the commit & squash procedure once more...

akallabeth:
@freerdp-bot test

akallabeth (approved):

oleg0421:
@akallabeth: don't have cmake-format. Tried to format cmake/ConfigOptions.cmake manually, but it fails again :( Could you suggest what could be wrong?

akallabeth:
@oleg0421 
```
diff --git a/cmake/ConfigOptions.cmake b/cmake/ConfigOptions.cmake
index a4395981c7..cf94704e1d 100644
--- a/cmake/ConfigOptions.cmake
+++ b/cmake/ConfigOptions.cmake
@@ -163,7 +163,7 @@ cmake_dependent_option(WITH_VIDEO_FFMPEG "Use FFMPEG for video encoding/decoding
 cmake_dependent_option(WITH_VAAPI "Use FFMPEG VAAPI" OFF "WITH_VIDEO_FFMPEG" OFF)
 cmake_dependent_option(WITH_VAAPI_H264_ENCODING "Use FFMPEG VAAPI hardware H264 encoding" ON "WITH_VIDEO_FFMPEG" OFF)
 if(WITH_VAAPI_H264_ENCODING)
-       add_definitions("-DWITH_VAAPI_H264_ENCODING")
+  add_definitions("-DWITH_VAAPI_H264_ENCODING")
 endif()
 
 option(USE_VERSION_FROM_GIT_TAG "Extract FreeRDP version from git tag." ON)
```

include/freerdp/codec/h264.h
```diff
@@ -61,6 +61,8 @@ extern "C"
 		H264_CONTEXT_OPTION_FRAMERATE,
 		H264_CONTEXT_OPTION_QP,
 		H264_CONTEXT_OPTION_USAGETYPE, /** @since version 3.6.0 */
+		H264_CONTEXT_OPTION_HW_ACCEL,  /** set to request hw accel, get to check if hw accel is on,
+		                                  @since version */
```
hardening, line 65: Don't forget to set a version here

hardening (approved):
Looks ok for me

include/freerdp/codec/h264.h
oleg0421: Ok, but what version?

akallabeth: 3.11.0 will be next (as 3.10.3 is current and we use semantic versioning)

hardening: Well I'd say 3.11 as it should be integrated for next version...

oleg0421: Done

oleg0421:
@akallabeth thank you,

```
./scripts/cmake-format.sh --check
```
passes now.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #10889, opened 2024-11-25
[client,x11] use xf_lock_x11 in xf_cliprdr_server_format_list

ondrejholy:
Currently, the xfreerdp client segfaults when the `xf_cliprdr_process_selection_request` and `xf_cliprdr_server_format_list` functions run concurrently. They are both reading and writing the `serverFormats`, `numServerFormats`, `response` and possibly some other properties. Let's use the `xf_local_x11` in the `xf_cliprdr_server_format_list` function to prevent this.

See: https://issues.redhat.com/browse/RHEL-68126


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM
thank you for the fix!

PR #10892, opened 2024-11-26
[warnings] fix lots of bugprone warnings

akallabeth:


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

libfreerdp/emu/scard/smartcard_emulate.c
```diff
@@ -1253,7 +1253,7 @@ HANDLE WINAPI Emulate_SCardAccessStartedEvent(SmartcardEmulationContext* smartca
 	WLog_Print(smartcard->log, smartcard->log_default_level, "SCardAccessStartedEvent {");
 
 	/* Not required, return random */
-	winpr_RAND(&hEvent, sizeof(hEvent));
+	winpr_RAND((void*)&hEvent, sizeof(hEvent));
```
hardening, line 1257: Strange I would have bet that cast to `void *` was implicit in C.

hardening (commented):
I don't see the added value to the adding of `(void*)` casts, except that it tends to bloat the code. And once this part of the patch will have passed we will have to use that everywhere at every new PR...

libfreerdp/emu/scard/smartcard_emulate.c
akallabeth: No, the only place these casts added are `void**` to `void*` or similar casts.

While implicit like integer narrow/widen it is (as the title suggests) often a source for buggy code (pass wrong type to function) and for that reason not necessarily without warnings

cmake/CleaningConfigureFile.cmake
```diff
@@ -6,6 +6,6 @@
 
 function(cleaning_configure_file SRC DST)
   configure_file(${SRC} ${DST} ${ARGN})
-  set_property(DIRECTORY APPEND PROPERTY ADDITIONAL_CLEAN_FILES ${DST})
+  # set_property(DIRECTORY APPEND PROPERTY ADDITIONAL_CLEAN_FILES ${DST})
```
mfleisz, line 9: Intended or debugging leftover?

mfleisz, line 9: Intended or debugging leftover?

mfleisz (commented):
Overall LGTM - just (maybe?) one leftover in the cmake config file

cmake/CleaningConfigureFile.cmake
akallabeth: true, that should be removed.

akallabeth: and fixed.

akallabeth:
@hardening the `(void*)` casts are there for the specific parts of code where we do some 2d array magic. (allocate with one `malloc` and `free`, ...)
while this works fine it is in no way the intended usage of the `API` and therefore prone to emit warnings.

PR #10894, opened 2024-11-27
[CMake,simd] unify WITH_SSE2, WITH_AVX2, WITH_NEON

akallabeth:
* Add new CMake option WITH_SIMD to enable best available instruction type
* Unify simd related defines in single header


mfleisz (commented):
It's a start but not quite happy with the changes yet:

- We still carry the bogus `WITH_SSE2` and `WITH_NEON` settings around. I see two problems with that: First, users just want to compile either with or without SIMD support (and not care about whether it be SSE, AVX or NEON). We should provide one switch `WITH_SIMD` to enable that and deprecate the old `WITH_XXX` (show a message and  just enable `WITH_SIMD`). Second problem is that they are utterly misleading since `WITH_SSE2` will define `SSE2_ENABLED` in the code (why not call it the same?) but enables not just SSE2 code paths but also every other SSE optimizations.
- We still check for architecture/compiler support in multiple areas (`DetectIntrinsicSupport.cmake` and `simd.h`)
- The MSVC compiler arguments seem not quite right (i.e. if I specify `/arch:SSE2` it is not guaranteed to support SSE3 and/or SSE4.x - there is even a designated SSE4.2 option available). However, all x64 MSVC compilers by now have removed the /arch:SSExx switches completely and just support everything SSE (they even print a diagnostic as you can see in the builder logs). MSVC x86 is the only left that supports any of the /arch:SSE switches but has SSE2 enabled by default either.
- Just a small side note: there are many occurrences of `CMAKE_C_COMPILER_ID` that have some strange character appended

As we do not really need any fain grained detection (we haven't had one until yet) I think a flexible solution could look like this:
- Just use `WITH_SIMD` to enable SIMD (`WITH_SSE2` and `WITH_NEON` will just set `WITH_SIMD=ON` and print a message). 
- Have a check if this is compiling for X86/X64 (`WITH_SIMD_SSE`) or ARM ((`WITH_SIMD_NEON`) plus a simple check for AVX2 support (as easy as checking with `check_c_compiler_flag` and setting `WITH_SIMD_AVX2`). Additionally we set SIMD compiler flag variables during detection (i.e. `SIMD_C_FLAGS_SSE2`, `SIMD_C_FLAGS_SSE3`, ...).
- Generate the `simd.h` header simply exposing `WITH_SIMD` and `WITH_SIMD_SSE/NEON/AVX2`
- `set_simd_source_file_properties` can be stripped down to just add `SIMD_C_FLAGS${INTRINSIC_TYPE}` to the compiler flags for the given files

akallabeth:
@mfleisz 
* regarding the removal of `WITH_SSE2` and `WITH_NEON` that can most likely be done, `WITH_AVX2` we need to keep (so we can test the `SSE` path and `AVX2` path of some optimizations)
* `SSE2_ENABLED` marks actual support for `SSE` while `WITH_SSE` just enables compiler detection for it. (so a different thing)
* the `simd.h` header checks need to stay as with builds for apple universal binaries are generated, where only during the compile run the actual architecture is known.
* we still need to only set the optimizations for the files with the `SIMD instructions`, adding the other flags to that code allows the compiler to generate these optimizations as well and we could end up with e.g. a non `AVX2` path containing `AVX2` instructions (at least for `gcc` and `clang`)

mfleisz (approved):

PR #10895, opened 2024-11-27
Cmake format

akallabeth:
* Add script running `cmake-format` on all `CMake` files
* apply script 

[EDIT] I´ve mostly applied the default style of `cmake-format` (except to not reformat comments and line length of 120)
so, if you want any changes speak up, I don´t care about which style is chosen, but it should be one and only one.


hardening (approved):
LGTM

PR #10900, opened 2024-11-29
[channels,drive] allow operation on main thread

akallabeth:


mfleisz (commented):
Just a question: To me it seems that the use of message queue and `freerdp_client_channel_register`/`freerdp_client_channel_unregister` is not really necessary in synchronous mode. If I understand the code correctly we just need to call `drive_process_irp` directly in `device.IRPRequest` instead of using a queue and that's it?

akallabeth:
@mfleisz yes, but this way we can handle different scenarios better - if we do register the handles we can change the thread model directly in the core without anything to be changed and eliminate the manual thread creation.

channels/drive/client/drive_main.c
```diff
@@ -56,6 +56,8 @@ typedef struct
 	wListDictionary* files;
 
 	HANDLE thread;
+	BOOL async;
+	BOOL running;
```
mfleisz, line 60: I think this is not needed anymore

PR #10904, opened 2024-12-02
[core,rdstls] fix rdstls parsing

joantolo:
When parsing the rdstls PDU, don't set the stream position at the end.

To know the pduLength is enough getting the last field length and the stream position.

This fixes transport_read_layer_bytes failing because it was getting on toRead = 0.


akallabeth:
@joantolo please explain with a little more detail.
you do reduce this to parttial parsing here so if that fixes something it smells like another bug is hidden

akallabeth:
@freerdp-bot test
@joantolo ok, added a check in `transport_default_read_pdu` to handle complete `PDU` with no additional data to read correctly.

akallabeth (approved):
LGTM

PR #10909, opened 2024-12-03
[channels,rdpgfx] fix rdpgfx_server_packet_complete_header

akallabeth:
a condition check was inverted preventing server side gfx packets to be sent.


akallabeth:
tested with shadow server and working.

PR #10915, opened 2024-12-04
[core,freerdp] fix automatic reconnect after auth

akallabeth:
In freerdp_connect do an automatic reconnect if we failed with a FREERDP_ERROR_CONNECT_TRANSPORT_FAILED
This way we can prevent connection disconnects if the remote was disconnected due to timeout (e.g. Auth-Dialog was open for a long time)


hardening (approved):
Seems ok for me

PR #10919, opened 2024-12-05
[Windows, clipboard] fix server don't support format to application exit.

KangLin:
[Windows, clipboard] fix server don't support format to application exit.
    
See: https://github.com/KangLin/RabbitRemoteControl/issues/31


akallabeth (approved):
LGTM

PR #10931, opened 2024-12-08
[client,common] fix manpage generation

akallabeth:
strip all compile/link flags for the manpage helper tool. This avoids linking/runtime issues with e.g. ASAN and other such tools used by nightly builds and others.


akallabeth:
@freerdp-bot test

PR #10937, opened 2024-12-09
bcrypt fix

akallabeth:
the `BCRYPT_*`  defines were not (consistently) `'\0` terminated. (e.g. `\0\0` in byte view)
use `C11` syntax to define strings in an easier to read manner and add a unit test to check for termination (and even byte count)


hardening (approved):
if the `u'string'` notation works, I'm delighted that it works

akallabeth:
@hardening that is part of `C 11` standard and allows declaring 16bit char type constants.

PR #10943, opened 2024-12-10
[client,x11] reset buffer length when sending

akallabeth:
When sending clipboard data reset the intermediate buffer position. This way the buffer is always 0 at beginning of reads, even if the same clipboard contents is requested in a different format.

Follow up to #10942


hardening (approved):
Seems ok

PR #10944, opened 2024-12-10
[cmake,configure_file] fix change checks

akallabeth:
* Create temporary configuration files in a different build directory
* Copy the currently required configuration file only if it is different from the currently set one
* Use copy_if_different for build target configuration update


hardening (approved):
Seems ok for me, but the CI on mac disagree...

akallabeth:
@hardening ci just uncovered a few minor bugs as this now works properly.

* the manpage genration helper tool needs to use proper paths
* `json-c` has inconsistencies in include paths (`CFLAGS` have both, `<prefix>` and `<prefix>/json-c` , the `CMake` file generated only lists `<prefix>/json-c`

PR #10964, opened 2024-12-13
Rework settings initialization to use bulk functions

hardening:
Introduce functions to do multiple initializations and use these functions to refactor `freerdp_settings_new`.


libfreerdp/core/settings.c
```diff
@@ -1294,6 +1227,76 @@ rdpSettings* freerdp_settings_new(DWORD flags)
 	return NULL;
 }
 
+BOOL freerdp_settings_set_bools(rdpSettings* settings, const FreeRDP_Settings_Bool_Value* values,
+                                size_t n)
+{
+	WINPR_ASSERT(settings);
```
akallabeth, line 1233: `values` is the thing dereferenced that should be asserted

libfreerdp/core/settings.c
```diff
@@ -1294,6 +1227,76 @@ rdpSettings* freerdp_settings_new(DWORD flags)
...
+			return FALSE;
+	}
+
+	return TRUE;
+}
+
+BOOL freerdp_settings_set_uint16s(rdpSettings* settings,
+                                  const FreeRDP_Settings_UInt16_Value* values, size_t n)
+{
+	WINPR_ASSERT(settings);
```
akallabeth, line 1247: `values` is the thing dereferenced that should be asserted

libfreerdp/core/settings.c
```diff
@@ -1294,6 +1227,76 @@ rdpSettings* freerdp_settings_new(DWORD flags)
...
+			return FALSE;
+	}
+
+	return TRUE;
+}
+
+BOOL freerdp_settings_set_int32s(rdpSettings* settings, const FreeRDP_Settings_Int32_Value* values,
+                                 size_t n)
+{
+	WINPR_ASSERT(settings);
```
akallabeth, line 1261: `values` is the thing dereferenced that should be asserted

libfreerdp/core/settings.c
```diff
@@ -1294,6 +1227,76 @@ rdpSettings* freerdp_settings_new(DWORD flags)
...
+			return FALSE;
+	}
+
+	return TRUE;
+}
+
+BOOL freerdp_settings_set_uint32s(rdpSettings* settings,
+                                  const FreeRDP_Settings_UInt32_Value* values, size_t n)
+{
+	WINPR_ASSERT(settings);
```
akallabeth, line 1275: `values` is the thing dereferenced that should be asserted

libfreerdp/core/settings.c
```diff
@@ -1294,6 +1227,76 @@ rdpSettings* freerdp_settings_new(DWORD flags)
...
+			return FALSE;
+	}
+
+	return TRUE;
+}
+
+BOOL freerdp_settings_set_strings(rdpSettings* settings,
+                                  const FreeRDP_Settings_String_Value* values, size_t n)
+{
+	WINPR_ASSERT(settings);
```
akallabeth, line 1289: `values` is the thing dereferenced that should be asserted

include/freerdp/settings.h
```diff
@@ -74,6 +74,51 @@ extern "C"
...
+	} FreeRDP_Settings_UInt32_Value;
+
+	/** \brief a string key id and its value
+	 * \since version 3.11.0
+	 */
+	typedef struct
+	{
+		FreeRDP_Settings_Keys_String keyId;
+		const char* value;
+	} FreeRDP_Settings_String_Value;
```
akallabeth, line 120: you can scope all these in a single 
```
struct foo {
int type;
union {
type1 a;
type2 b;
} u;
}
}
```

then you can unify the bulk setters to a single function.

akallabeth (changes requested):
Nice idea!

would do it with less exposed functions though (simply add a type filed in the struct so the bulk setter knows what to set)

include/freerdp/settings.h
```diff
@@ -403,6 +487,19 @@ extern "C"
...
+	/** \brief Sets some UINT32 settings values.
+	 *
+	 *  \param settings A pointer to the settings to alter, must not be NULL.
+	 *  \param values The key/values to set
+	 *  \param n The number of key/value to set.
+	 *  \since version 3.11.0
+	 *
+	 *  \return \b TRUE for success, \b FALSE for failure
+	 */
+	FREERDP_API BOOL freerdp_settings_set_uint32s(rdpSettings* settings,
```
akallabeth, line 499: maybe a good idea to make the name easier to distinguish from the single setters, name it `freerdp_settings_set_bulk(settings, values, count)`?

akallabeth:
@hardening maybe expose `FREERDP_SETTINGS_TYPE` for that?

include/freerdp/settings.h
hardening: Well that can be done as an extension, but I don't think would be that convenient to use: we'll have to specify the type and the field name in the initializer array. The arrays are already massive in `freerdp_settings_new` that would probably double them and have each initialized value spread on multiple lines: probably less readable...

akallabeth: @hardening well, not a good argument for the `API` functions, these should be well defined, minimal and complete. (just seen you did not implement all setters here)
by having a single exposed `bulk` setter that could be easily done, including future extensions.

for the initializer stuff in `settings_new`  you can do an internal wrapper though, so you don´t have to add the type.

akallabeth: @hardening and just remembered something: check the function `freerdp_settings_set_value_for_name`

hardening: So we'll have nice code internally thanks to macros and all external users of this function will have long array initializers (or will have to reimplement these macros), that sounds like the bad public `API` recipe.
Another point I've just figured is that with the union we completely loose the static type checking of settings, so why not adding a bulk function for the adventurous, but also keep the functions by types that do the checks at build time...

akallabeth: @hardening not if you use named initializers, but yes, without them there are no type checks.
as for external users, where do you need to set these in bulk?

main use is in setting defaults for settings, command line parser and rdp file
and only the default initializers do require bulk initialization.

so, before we add a lot of functions to `API` that we need to maintain I'd want to hear a good reason for that.

hardening: Well ok, if that kind of thing works, then I'm for it as it does the static check:
```C
typedef struct {
   rdp_settings_type stype;
   union {
       struct {
          FreeRDP_Settings_Keys_UInt16 keyId;
          UINT16 value;
       } t_uint16;
       struct {
          FreeRDP_Settings_Keys_UInt32 keyId;
          UINT32 value;
       } t_uint32;
    ...
   } u;
} FreeRDP_Settings_Value;

#define FREERDP_INIT_UINT32(K, V) { .stype: RDP_SETTINGS_TYPE_UINT32, t_uint32: { FREERDP_##K, V } }

FreeRDP_Settings_Value values[] = {
   FREERDP_INIT_UINT32(FakeMouseMotionInterval, 5),
};


```

akallabeth: I´d name the members less similar to the datatypes (e.g. `uint32_t -> u32` and `int32_t -> i32`) but otherwise good.
would also be possible to make the macro a `static inline` function like `static inline FreeRDP_Settings_Value FREERDP_INIT_UINT32(FreeRDP_Settings_Keys_UInt32 key, uint32_t val)`

hardening: Hum just did a few tests but unfortunately I can see exactly what I was afraid of: the compiler manage to put a `FreeRDP_Settings_Keys_Bool` in a `FreeRDP_Settings_Keys_String` without doing the type check. So for instance 
```
FreeRDP_Settings_Value values[] = {
		FREERDP_SETTING_BOOL(FreeRDP_TcpKeepAlive, TRUE),
		FREERDP_SETTING_STRING(FreeRDP_TcpKeepAlive, FALSE),
	};
```
builds, and without even a warning and so you will figure the error at runtime.

akallabeth: Then use a static inline instead of a macro?

akallabeth: @hardening also, just tested:
*  `gcc` is a bit less strict than `clang` with warnings, but a integer assignment to a string is always warned about
* assigning a string to an integer value is only warned by `clang -Weverything`

hardening: Well of course I'm bothered by an int assigned to a string, but I was even more bothered by different enums types assigned to each other without a single notice. That also happens with inline functions. That was my main point against opaque settings: the loss of static checks on the settings, and unfortunately we're at that point :-(

akallabeth:
@hardening did have a look at this: see https://github.com/akallabeth/FreeRDP/tree/settings_massive
when using `static inline` functions you get proper type checks (avoiding named initializers in the header though as `C++` only supports them quite recently, whereas `C` supports them since `C99`)

PR #10966, opened 2024-12-13
Settings tests

akallabeth:
* add some `API` documentation
* Add some `freerdp_settings_*` functions not covered previously to unit tests

### DIFF
--- a/libfreerdp/core/test/TestSettings.c
+++ b/libfreerdp/core/test/TestSettings.c
@@ -1382,6 +1382,106 @@ fail:
 	return log_result(rc, __func__);
 }
 
+static BOOL test_string_null(rdpSettings* settings, FreeRDP_Settings_Keys_String id)
+{
+	if (!freerdp_settings_set_string(settings, id, NULL))
+		return FALSE;
+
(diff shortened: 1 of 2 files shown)


libfreerdp/core/test/TestSettings.c
```diff
@@ -1382,6 +1382,106 @@ static BOOL test_validity_check(void)
...
+static BOOL test_string_null(rdpSettings* settings, FreeRDP_Settings_Keys_String id)
+{
+	if (!freerdp_settings_set_string(settings, id, NULL))
+		return FALSE;
+
+	const char* chk = freerdp_settings_get_string(settings, id);
+	return (chk == NULL);
+}
+
+static BOOL test_string_check(rdpSettings* settings, FreeRDP_Settings_Keys_String id,
```
pnowack, line 1394: Maybe something like `test_string_equal()` would be more precise as a name here?

pnowack (approved):
Just one thing, otherwise lgtm.

libfreerdp/core/test/TestSettings.c
akallabeth: true, but not really relevant.

PR #10968, opened 2024-12-13
[cmake] case insensitive simd check

akallabeth:


VVD:
Build fine with patches! Thanks!

PR #10971, opened 2024-12-15
[ci,build] add workflow to build on FreeBSD

akallabeth:


akallabeth:
merging to allow running a test build.
issues with builds fixed in follow up pull requests

PR #10989, opened 2024-12-16
[locale,keyboard] deprecate keyboard mapping

akallabeth:
* deprecate keyboard mapping functions:
  * not context sensitive but app global
  * only work reliable with XKB files
* expose functions to parse and remap RDP scancodes


ThinkChaos (approved):
This does fix the issue for me!

Thanks for you responsiveness :)

akallabeth:
@freerdp-bot test

PR #11004, opened 2024-12-19
[aad] Respect the /server-name command line option

chewi:
AAD authentication fails when you give a bare IP to `/v`.


akallabeth:
@freerdp-bot test

PR #11018, opened 2024-12-25
Fix gfxredir_server.. undeclared here

gpotter2:
Trying to include `gfxredir.h` fails with:
```
In file included from /usr/local/lib/pkgconfig/../../include/winpr3/winpr/winpr.h:22:
/usr/local/lib/pkgconfig/../../include/freerdp3/freerdp/server/gfxredir.h:95:27: error: ‘gfxredir_server_context_free’ undeclared here (not in a function); did you mean ‘audin_server_context_free’?
   95 |         WINPR_ATTR_MALLOC(gfxredir_server_context_free, 1)
```

This was apparently broken in https://github.com/FreeRDP/FreeRDP/commit/bb42d425edd9467977f4673589c05028ec4701dc.

Merry christmas!

### DIFF
--- a/include/freerdp/server/gfxredir.h
+++ b/include/freerdp/server/gfxredir.h
@@ -92,9 +92,10 @@ extern "C"
 		UINT32 confirmedCapsVersion;
 	};
 
+	FREERDP_API void gfxredir_server_context_free(GfxRedirServerContext* context);
+
 	WINPR_ATTR_MALLOC(gfxredir_server_context_free, 1)
 	FREERDP_API GfxRedirServerContext* gfxredir_server_context_new(HANDLE vcm);
-	FREERDP_API void gfxredir_server_context_free(GfxRedirServerContext* context);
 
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
that looks wrong.
you can´t just move the attribute away from the function and expect it still works.

gpotter2:
I'm not sure I understand what you mean. I changed the fix to make it clearer, but the only thing I'm doing is changing the order of 2 lines: defining the free function before using it in the macro.

I must say I haven't looked into what this macro does. All i know is that everywhere else it is used, it is always after the definition and not before it like in this file.

akallabeth (approved):
LGTM.

yes, sorry could have been clearer.
the function it uses must be declared before use of the macro

the maco itself must be part of the function declaration (basically tells that this function is an allocator and the function within the braces the free function for it)

akallabeth:
@freerdp-bot test

PR #11030, opened 2025-01-01
Sign fixes

akallabeth:
* fixes a lot of sign warnings
* problematic part: change some `API` function arguments from `unsigned` to `signed`

### DIFF
--- a/libfreerdp/codec/test/TestFreeRDPCodecProgressive.c
+++ b/libfreerdp/codec/test/TestFreeRDPCodecProgressive.c
@@ -1185,8 +1185,17 @@ static WINPR_NORETURN(void usage(const char* name))
 
 static void print_codec_stats(const char* name, UINT64 timeNS)
 {
-	const double dectimems = timeNS / 1000000.0;
-	(void)fprintf(stderr, "[%s] took %lf ms to decode\n", name, dectimems);
+	const double dectimeMS = (double)timeNS / 1000.0 / 1000.0;
+	(void)fprintf(stderr, "[%s] took %lf ms to decode\n", name, dectimeMS);
+}
+
+static UINT64 measure_diff_and_print(const char* cname, UINT32 frameId, UINT64 start)
+{
+	const UINT64 end = winpr_GetTickCount64NS();
+	const UINT64 diff = end - start;
+	const double ddiff = (double)diff / 1000000.0;
+	(void)fprintf(stderr, "frame [%s] %" PRIu32 " took %lf ms\n", cname, frameId, ddiff);
+	return diff;
 }
 
 static int test_dump(int argc, char* argv[])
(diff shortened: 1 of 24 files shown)


libfreerdp/utils/smartcard_call.c
```diff
@@ -1548,7 +1548,7 @@ static LONG smartcard_LocateCardsByATRA_Call(scard_call_context* smartcard, wStr
 }
 
 LONG smartcard_irp_device_control_call(scard_call_context* smartcard, wStream* out,
-                                       UINT32* pIoStatus, SMARTCARD_OPERATION* operation)
+                                       NTSTATUS* pIoStatus, SMARTCARD_OPERATION* operation)
```
pnowack, line 1551: Isn't NTSTATUS a signed type (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/c8b512d5-70b1-4028-95f1-ec92d35cb51e)?

hm.. on a second look, the docs mention here an unsigned type (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/1c412a84-0776-4984-b35c-3f0445fcae65). This looks contrary. Some changes here in this MR change `INT32` to `NTSTATUS`, some change `UINT32` to `NTSTATUS`. Do we use now a signed or an unsigned type?

libfreerdp/codec/test/TestFreeRDPCodecProgressive.c
```diff
@@ -1185,10 +1185,19 @@ static WINPR_NORETURN(void usage(const char* name))
 
 static void print_codec_stats(const char* name, UINT64 timeNS)
 {
-	const double dectimems = timeNS / 1000000.0;
+	const double dectimems = (double)timeNS / 1000000.0;
```
pnowack, line 1188: On one side, we have here `dectimems`, on the other one `timeNS`. Either way, whether we use snake case or camel case, this should be unique here, e.g. either `dec_time_ms` + `time_ns` or `decTimeMS` + `timeNS`

Also: I counting zeros takes some time to verify the number. Instead of `(double)timeNS / 1000000.0;`, I think `(double)timeNS / 1000.0 / 1000.0;` is much clearer (and is as a number faster verified)

libfreerdp/core/tcp.c
```diff
@@ -1163,10 +1162,16 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
 #if defined(HAVE_AF_VSOCK_H)
 		hostname = vsock;
 		sockfd = socket(AF_VSOCK, SOCK_STREAM, 0);
+		if (sockfd < 0)
```
pnowack, line 1165: I think a warning message with the `errno` would be good here (when hitting that issue).

pnowack (commented):
Don't know much about the SDL parts. Also not sure about whether NTSTATUS is now signed or unsigned, but aside from that, except for the nits here, I couldn't spot something here.

libfreerdp/utils/smartcard_call.c
akallabeth: @pnowack basically a signed type `NTSTATUS` (`int32_t` typedef)
the spec is contradictory but all status codes in use are referencing `NTSTATUS` values, so the idea is to use this type instead

libfreerdp/core/tcp.c
akallabeth: ok, done.

libfreerdp/codec/test/TestFreeRDPCodecProgressive.c
akallabeth: done

PR #11035, opened 2025-01-03
winpr/input: Add missing korean keyboard type

pnowack:
Quoting the commit here:
```
The keyboard types in 2.2.1.3.2 Client Core Data (TS_UD_CS_CORE)
([MS-RDPBCGR]) contain a korean keyboard type, which is not yet present
in WinPR. So, add it.
```


akallabeth (approved):
LGTM

PR #11037, opened 2025-01-03
[winpr,input] add missing WINPR_KBD_TYPE

akallabeth:
add mapping for uncommon keyboard types as well.

* added japanese layout corrections
* added korean layout
* map IBM compatible to IBM enhanced


hardening (approved):
Seems ok right now even if new keyboard types aren't filled

pnowack:
In my test setup here, I applied this patch on the server side, where g-r-d runs (which uses the conversion method from this MR here (virtual scancode -> virtual keycode, virtual keycode -> keycode).

Did some tests with IBM enhanced kbd type on the client side (tested with 2 different clients) (Client 1: xfreerdp3 (client side does not use this patch here)) (Client 2: Microsofts RDP client for iOS (IPhone)):
The `Windows` key seems to have no effect, same for the `Alt Gr` key, the `Del` key, `End` key, and arrow keys. (haven't checked all, but these already don't work anymore).

winpr/libwinpr/input/scancode.c
```diff
@@ -104,7 +129,426 @@ static DWORD KBD7X[128] = {
...
+	                              VK_NONE,
+	                              VK_NONE,
+	                              VK_NONE,
+	                              VK_NONE,
+	                              VK_OEM_PA1,
+	                              VK_TAB,
+	                              VK_ABNT_C2,
+	                              VK_OEM_PA2 };
+
+static const DWORD KBD8X[256] = { VK_NONE,
```
pnowack, line 296: This (`KBD8X`) and `KBD8T` should better use the same formatting, as it was done in the other already existing arrays (`KBD4X`, `KBD4T`, `KBD7X`, `KBD7T`)

winpr/libwinpr/input/scancode.c
```diff
@@ -114,81 +558,86 @@ DWORD GetVirtualKeyCodeFromVirtualScanCode(DWORD scancode, DWORD dwKeyboardType)
...
+	{
+			/* IBM and compatible */
+		case WINPR_KBD_TYPE_IBM_PC_XT:
+		case WINPR_KBD_TYPE_OLIVETTI_ICO:
+		case WINPR_KBD_TYPE_IBM_PC_AT:
+		case WINPR_KBD_TYPE_IBM_ENHANCED:
+			if (scancode & KBDMULTIVK)
+				return KBD4X_1[codeIndex];
+			if (scancode & KBDMULTIVK)
+				return KBD4X[codeIndex];
```
pnowack, line 571:
```
			if (scancode & KBDMULTIVK)
				return KBD4X_1[codeIndex];
			if (scancode & KBDMULTIVK)
				return KBD4X[codeIndex];
```
Ah, the flag check does not look correct to me. Probably, that's why some keys don't work in my testing here.

winpr/libwinpr/input/scancode.c
```diff
@@ -114,81 +558,86 @@ DWORD GetVirtualKeyCodeFromVirtualScanCode(DWORD scancode, DWORD dwKeyboardType)
...
 
-	if (dwKeyboardType == WINPR_KBD_TYPE_IBM_ENHANCED)
-		return (scancode & KBDEXT) ? KBD4X[codeIndex] : KBD4T[codeIndex];
-	else if (dwKeyboardType == WINPR_KBD_TYPE_JAPANESE)
-		return (scancode & KBDEXT) ? KBD7X[codeIndex] : KBD7T[codeIndex];
+		case WINPR_KBD_TYPE_JAPANESE:
+			if (scancode & KBDMULTIVK)
+				return KBD7X_1[codeIndex];
+			if (scancode & KBDMULTIVK)
+				return KBD7X[codeIndex];
```
pnowack, line 578: Same like above (wrong flag check)

winpr/libwinpr/input/scancode.c
```diff
@@ -114,81 +558,86 @@ DWORD GetVirtualKeyCodeFromVirtualScanCode(DWORD scancode, DWORD dwKeyboardType)
...
+			return KBD7T[codeIndex];
+		case WINPR_KBD_TYPE_KOREAN:
+			if (scancode & KBDMULTIVK)
+				return KBD8X_1[codeIndex];
+			if (scancode & KBDMULTIVK)
+				return KBD8X[codeIndex];
```
pnowack, line 584: Same like above (wrong flag check)

pnowack (changes requested):

winpr/libwinpr/input/scancode.c
pnowack: Upon checking the other conversion direction, I guess the latter check should use the `KBDEXT` flag.

winpr/libwinpr/input/scancode.c
```diff
@@ -114,81 +558,86 @@ DWORD GetVirtualKeyCodeFromVirtualScanCode(DWORD scancode, DWORD dwKeyboardType)
 	if (codeIndex > 127)
 		return VK_NONE;
 
-	if ((dwKeyboardType != WINPR_KBD_TYPE_IBM_ENHANCED) &&
-	    (dwKeyboardType != WINPR_KBD_TYPE_JAPANESE))
-		dwKeyboardType = WINPR_KBD_TYPE_IBM_ENHANCED;
+	switch (dwKeyboardType)
+	{
+			/* IBM and compatible */
+		case WINPR_KBD_TYPE_IBM_PC_XT:
```
pnowack, line 564: Nit: indentation of the comment above this `case` seems to be off

winpr/libwinpr/input/scancode.c
akallabeth: oh, looks like a `clang-format` issue, need to check.

akallabeth: indeed.
did disable the automatic formatting for these two arrays as the formatting depends on the length of the define names (tires to neatly arrange these in rows, but fails doing that)

winpr/libwinpr/input/scancode.c
akallabeth: indeeed. fixed.

pnowack (approved):
With the last push, extended key events (at least on IBM enhanced) are now fixed.

PR #11043, opened 2025-01-04
Allow to disable encryption as a server

gpotter2:
This PR makes it possible to disable encryption when acting as a server, by setting `FreeRDP_EncryptionLevel` to `ENCRYPTION_LEVEL_NONE`. Previously, this was the default but was changed dynamically to `ENCRYPTION_LEVEL_CLIENT_COMPATIBLE`, so instead it set the default to `ENCRYPTION_LEVEL_CLIENT_COMPATIBLE` and allow disabling it by setting it to None.

This is incredibly useful for debugging.

### DIFF
--- a/libfreerdp/core/nego.c
+++ b/libfreerdp/core/nego.c
@@ -1505,18 +1505,6 @@ BOOL nego_send_negotiation_response(rdpNego* nego)
 				if (!freerdp_settings_set_bool(settings, FreeRDP_UseRdpSecurityLayer, TRUE))
 					return FALSE;
 
-				if (freerdp_settings_get_uint32(settings, FreeRDP_EncryptionLevel) ==
-				    ENCRYPTION_LEVEL_NONE)
-				{
-					/**
-					 * If the server implementation did not explicitly set a
-					 * encryption level we default to client compatible
-					 */
-					if (!freerdp_settings_set_uint32(settings, FreeRDP_EncryptionLevel,
-					                                 ENCRYPTION_LEVEL_CLIENT_COMPATIBLE))
-						return FALSE;
-				}
-
 				if (freerdp_settings_get_bool(settings, FreeRDP_LocalConnection))
 				{
 					/**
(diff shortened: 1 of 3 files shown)


hardening:
Hum this would also be very useful when you wanna do RDP on top of a socket pair and clearly you don't want any encryption

akallabeth (changes requested):
Overall LGTM,

I'd appreciate a log message if that is done just like with the `FreeRDP_LocalConnection` branch

hardening:
Side note: some time ago I did a try with `LocalConnection` set on both client and server and that was not working, the nego process was unfortunately failing. Perhaps we must dig that because instead of the NULL RDP encryption, what we want is no encryption at all...

@gpotter2 I suspect that what you call easier debugging is that you can see in clear the RDP traffic. You can achieve that by using `/tls:secrets-file:your file path` and configure your wireshark TLS plugin with this path. With that freerdp will dump SSL secrets, and you'll see everything in clear even connections with TLS/NLA. If you use mimikatz to export the DC's kerberos secrets, you'll really have everything in clear including all the NLA related parts...

akallabeth:
ok, did have a look on this, works as advertised.

not sure we want this in without a compile time guard though, what is your opinion on that @hardening @mfleisz @pnowack ?
while it is nice for debugging it allows configuring an even less secure server (no encryption instead of weak one)

We can merge it and add a define guard later on before we do another release as well.

gpotter2:
I think there are cases where you could want something like this.
One of my use cases is binding FreeRDP over a a VSock. Encryption isn't really useful for those.

akallabeth:
> Ok let's do that I'll see what it gives with my test case where I run a FreeRDP based server that connects locally to qfreerdp. That would allow me to drop the automatic cert generation.

you mean merge?

akallabeth:
@freerdp-bot test

akallabeth:
@gpotter2 ok, the `ci` is failing as your code does not meet formatting style.
please run `git clang-format` on your changes. (if you are at it, best squash all changes to a single commit and force push)

akallabeth:
ok, just realized this whole functionality is already integrated. Check the `FreeRDP_LocalConnection` setting, that is what you´re looking for.

PR #11044, opened 2025-01-05
Regression in update_check_flush

gpotter2:
If I'm not mistaken, and I absolutely could be, I believe that the return value of `update_check_flush` is wrong and has been for a long time. This never was an issue before https://github.com/FreeRDP/FreeRDP/commit/5f862846638cfa92742a69a76ec160a0ff8299ab as it wasn't checked, but now that it is I believe that it's causing issues.

My understanding is that this function should:
- create `s->us` by calling `begin` if missing
- flush if the buffer is full
- return false if an error occurs

Currently doing `BeginPaint`, `CreateWindow` then `EndPaint` fails on CreateWindow because of that. I do not think that this is wanted.

If I got it wrong feel free to explain :P

### DIFF
--- a/libfreerdp/core/update.c
+++ b/libfreerdp/core/update.c
@@ -1079,19 +1079,20 @@ static BOOL update_check_flush(rdpContext* context, size_t size)
 
 	wStream* s = update->us;
 
-	if (!update->us)
+	if (!s)
 	{
-		update_begin_paint(&update->common);
-		return FALSE;
+		if (!update_begin_paint(&update->common))
+			return FALSE;
+		s = update->us;
 	}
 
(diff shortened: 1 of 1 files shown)


gpotter2:
BTW @hardening, Hi ! Thanks for having a look at my PRs.
I really likes what you did in weston regarding RDP, that was impressive work. To give a little bit of context the 4 PRs I just made are for https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/1667 which could interest you. I'm rebasing the work Microsoft did for WSLg and trying to have it merged in weston so that it stays supported in the long run.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #11052, opened 2025-01-09
log: Modify received Heartbeat PDU from debug to trace

KangLin:
log: Modify received Heartbeat PDU from debug to trace


hardening (approved):
Oh yes, I'm doing that on all my fresh checkout

hardening:
@freerdp-bot can you do a test build please

PR #11053, opened 2025-01-09
winpr: fix dependency in cmake generated file

hardening:
When a static build was done uriparser was not taken as dependency.


akallabeth (approved):
LGTM

PR #11060, opened 2025-01-10
Fix calls of Stream_EnsureCapacity

hardening:
There were various places where `Stream_EnsureCapacity` was called instead of the expected `Stream_EnsureRemainingCapacity`. When the stream position is set to 0 the 2 calls are equivalent, but we had various places where it was not always the case.


akallabeth (changes requested):

hardening:
Ok fixed the wrong fixes

akallabeth (approved):
LGTM

PR #11065, opened 2025-01-12
H264 test

akallabeth:
* add a unit test for h264 (very basic, but a start)
* fix a unnecessary startup delay if only one optimized primitive is available
* reset H264 encoder/decoder on reset
* fix a few error path leaks

### DIFF
--- a/libfreerdp/codec/yuv.c
+++ b/libfreerdp/codec/yuv.c
@@ -171,12 +171,12 @@ BOOL yuv_context_reset(YUV_CONTEXT* WINPR_RESTRICT context, UINT32 width, UINT32
 
 	if (context->useThreads)
 	{
-		const UINT32 pw = (width + TILE_SIZE - width % TILE_SIZE) / TILE_SIZE;
-		const UINT32 ph = (height + TILE_SIZE - height % TILE_SIZE) / TILE_SIZE;
+		const size_t pw = (width + TILE_SIZE - width % TILE_SIZE) / TILE_SIZE;
+		const size_t ph = height + TILE_SIZE - height % TILE_SIZE / TILE_SIZE;
 
-		/* We´ve calculated the amount of workers for 64x64 tiles, but the decoder
-		 * might get 16x16 tiles mixed in. */
-		const UINT32 count = pw * ph * 16;
+		/* calculate absolute maximum of work tiles to be used.
+		 */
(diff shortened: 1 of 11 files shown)


libfreerdp/codec/yuv.c
```diff
@@ -171,12 +171,12 @@ BOOL yuv_context_reset(YUV_CONTEXT* WINPR_RESTRICT context, UINT32 width, UINT32
 
 	if (context->useThreads)
 	{
-		const UINT32 pw = (width + TILE_SIZE - width % TILE_SIZE) / TILE_SIZE;
-		const UINT32 ph = (height + TILE_SIZE - height % TILE_SIZE) / TILE_SIZE;
+		const size_t pw = (width + TILE_SIZE - width % TILE_SIZE) / TILE_SIZE;
+		const size_t ph = height + TILE_SIZE - height % TILE_SIZE / TILE_SIZE;
```
pnowack, line 175: Why were the braces `()` around `height + TILE_SIZE - height % TILE_SIZE` for `ph` removed, but not for `pw`? Was this accidental? It looks to me, that either they should not be removed at all, or removed for both `ph` and `pw`.

akallabeth: @pnowack thx, was really an accident.

akallabeth: https://github.com/FreeRDP/FreeRDP/pull/11066

libfreerdp/primitives/primitives.c
```diff
@@ -220,7 +220,7 @@ static BOOL primitives_autodetect_best(primitives_t* prims)
 	};
 	const struct prim_benchmark* best = NULL;
 
-#if !defined(HAVE_CPU_OPTIMIZED_PRIMITIVES) && !defined(WITH_OPENCL)
+#if !defined(HAVE_CPU_OPTIMIZED_PRIMITIVES) || !defined(WITH_OPENCL)
 	{
 		struct prim_benchmark* cur = &testcases[0];
```
oleg0421, line 225: @akallabeth : this change introduced a regression! In my setup, `HAVE_CPU_OPTIMIZED_PRIMITIVES` is defined, but `testcases[0]` picks up "generic" primitives instead of "optimized". As a result, there is a slight, but noticeable CPU utilization jump in video related scenarios.

PR #11074, opened 2025-01-13
[winpr,input] fix japanese mapping

akallabeth:
use japanese 109 layout instead of 106 layout

@pnowack I may have used the wrong layout as basis for the japanese mapping, the one in the issue you mentioned in matrix https://gitlab.gnome.org/GNOME/gnome-remote-desktop/-/issues/234#note_2315460 looks like the 109 key layout https://kbdlayout.info/kbdjpn/virtualkeys?arrangement=OADG109A

so, adjusted the mapping once again and hope this now covers the use case.


akallabeth:
got feedback from https://gitlab.gnome.org/GNOME/gnome-remote-desktop/-/issues/234#note_2329947 and the mapping looks sane now as we have confirmation the scancodes match the ones listed in https://kbdlayout.info/kbdjpn/virtualkeys?arrangement=ISO105 and the keyboard type is `7` for japanese.
Maybe we should also check what exactly the `keyboard subtype 2` is (`iso/ansi/abnt/oadg`?)

PR #11077, opened 2025-01-14
Tzsimplify and primitives benchmark fix

akallabeth:


libfreerdp/primitives/primitives.c
```diff
@@ -165,6 +165,7 @@ static primitives_YUV_benchmark* primitives_YUV_benchmark_init(primitives_YUV_be
 	return ret;
 }
 
+#if defined(HAVE_CPU_OPTIMIZED_PRIMITIVES) && defined(WITH_OPENCL)
```
hardening, line 168: Shouldn't it be a `or` instead of a `and` ?

akallabeth: no, `and` because the benchmark is only run if we have more than one optimized primitive implementation.

PR #11085, opened 2025-01-14
[clang,tidy] fix warnings

akallabeth:


libfreerdp/core/capabilities.c
```diff
@@ -4540,8 +4540,7 @@ BOOL rdp_recv_demand_active(rdpRdp* rdp, wStream* s, UINT16 pduSource, UINT16 le
 	if (!Stream_CheckAndLogRequiredLength(TAG, s, 4))
 		return FALSE;
 
-	UINT32 SessionId = 0;
-	Stream_Read_UINT32(s, SessionId); /* SessionId */
+	const UINT32 SessionId = Stream_Get_UINT32(s); /* SessionId */
```
hardening, line 4543: Perhaps we could just `Stream_Seek` as it's not used

winpr/libwinpr/utils/asn1/asn1.c
```diff
@@ -686,9 +686,9 @@ size_t WinPrAsn1EncOID(WinPrAsn1Encoder* enc, const WinPrAsn1_OID* oid)
 	return WinPrAsn1EncMemoryChunk(enc, ER_TAG_OBJECT_IDENTIFIER, oid);
 }
 
-size_t WinPrAsn1EncOctetString(WinPrAsn1Encoder* enc, const WinPrAsn1_OctetString* octets)
+size_t WinPrAsn1EncOctetString(WinPrAsn1Encoder* enc, const WinPrAsn1_OctetString* oid)
```
hardening, line 689: This renaming is unfortunate, octets strings != oid

winpr/libwinpr/utils/asn1/asn1.c
```diff
@@ -742,9 +742,9 @@ size_t WinPrAsn1EncContextualOID(WinPrAsn1Encoder* enc, WinPrAsn1_tagId tagId,
 }
 
 size_t WinPrAsn1EncContextualOctetString(WinPrAsn1Encoder* enc, WinPrAsn1_tagId tagId,
-                                         const WinPrAsn1_OctetString* octets)
+                                         const WinPrAsn1_OctetString* oid)
 {
```
hardening, line 746: This renaming is unfortunate, octets strings != oid

libfreerdp/core/capabilities.c
akallabeth: keeping that as reminder, as the field has a function but we ignore it. (we do seek when the field is unused)

winpr/libwinpr/utils/asn1/asn1.c
akallabeth: ok, then we should rename the other way around.

winpr/libwinpr/utils/asn1/asn1.c
akallabeth: ok, renamed.

PR #11093, opened 2025-01-18
Make credential dialog work with SDL3

vmpn:
When using `master` branch with SDL3 the credentials dialog was not working for me.

Looking at code and API documentation seems like code was not fully changed from SDL2->SDL3 API conventions.

Tested by successfully working with real RDP host using SDL3 with wayland video driver and following flags

```
/gateway:g:<censored>,u:<censored>
/u:<censored>
/v:<censored>
/audio-mode:0
+compression
/multimon
+toggle-fullscreen
```

PR #11092, opened 2025-01-17
Make code compatible with SDL3 3.1.8+

vmpn:
Make FreeRDP work post https://github.com/libsdl-org/SDL/commit/b902b0527ba18f23f06fbf09026f27f714d8f6cd commit that renames field in SDL_ClipboardEvent struct

PR #11094, opened 2025-01-18
FreeRDP forces primary monitor (x,y) to 0, 0 which breaks multi-monitor geometry checks 

vmpn:
*  I am using SDL3 client over wayland (KWIN)
* I have 3 monitor setup with middle monitor being the primary:

Image

The multi monitor geometry checks fail due to code in `freerdp_settings_enforce_monitor_exists` introduced in 392a0857280943002b1fffc4a5fd25deedec3ed2 forcing x and y of primary monitor to 0,0

Applying proposed changes fixes the geometry checks while ensuring monitor info is shifted so that top left of main monitor is still reported at (0,0) per `2.2.1.3.6.1 Monitor Definition (TS_MONITOR_DEF)` spec, and allows me to connect to Windows 11 RDP host.

There are other patches I had to apply to make SDL3 work and test this change but they felt separate from this issue:

* https://github.com/FreeRDP/FreeRDP/pull/11092
* https://github.com/FreeRDP/FreeRDP/pull/11093

### DIFF
--- a/libfreerdp/core/gcc.c
+++ b/libfreerdp/core/gcc.c
@@ -2198,26 +2198,25 @@ BOOL gcc_write_client_monitor_data(wStream* s, const rdpMcs* mcs)
 			{
 				baseX = current->x;
 				baseY = current->y;
-				break;
 			}
 		}
 
 		for (UINT32 i = 0; i < settings->MonitorCount; i++)
 		{
 			const rdpMonitor* current = &settings->MonitorDefArray[i];
-			const UINT32 left = WINPR_ASSERTING_INT_CAST(uint32_t, current->x - baseX);
-			const UINT32 top = WINPR_ASSERTING_INT_CAST(uint32_t, current->y - baseY);
-			const UINT32 right = left + WINPR_ASSERTING_INT_CAST(uint32_t, current->width - 1);
-			const UINT32 bottom = top + WINPR_ASSERTING_INT_CAST(uint32_t, current->height - 1);
-			const UINT32 flags = current->is_primary ? MONITOR_PRIMARY : 0;
+			const INT32 left = WINPR_ASSERTING_INT_CAST(int32_t, current->x - baseX);
+			const INT32 top = WINPR_ASSERTING_INT_CAST(int32_t, current->y - baseY);
+			const INT32 right = left + WINPR_ASSERTING_INT_CAST(int32_t, current->width - 1);
+			const INT32 bottom = top + WINPR_ASSERTING_INT_CAST(int32_t, current->height - 1);
+			const INT32 flags = current->is_primary ? MONITOR_PRIMARY : 0;
 			WLog_DBG(TAG,
 			         "Monitor[%" PRIu32 "]: top=%" PRIu32 ", left=%" PRIu32 ", bottom=%" PRIu32
 			         ", right=%" PRIu32 ", flags=%" PRIu32,
 			         i, top, left, bottom, right, flags);
-			Stream_Write_UINT32(s, left);   /* left */
-			Stream_Write_UINT32(s, top);    /* top */
-			Stream_Write_UINT32(s, right);  /* right */
-			Stream_Write_UINT32(s, bottom); /* bottom */
+			Stream_Write_INT32(s, left);   /* left */
+			Stream_Write_INT32(s, top);    /* top */
+			Stream_Write_INT32(s, right);  /* right */
+			Stream_Write_INT32(s, bottom); /* bottom */
 			Stream_Write_UINT32(s, flags);  /* flags */
 		}
 	}
--- a/libfreerdp/core/settings.c
+++ b/libfreerdp/core/settings.c
@@ -1763,8 +1763,6 @@ BOOL freerdp_settings_enforce_monitor_exists(rdpSettings* settings)
 		    freerdp_settings_get_pointer_array_writable(settings, FreeRDP_MonitorDefArray, 0);
 		if (!monitor)
 			return FALSE;
-		monitor->x = 0;
-		monitor->y = 0;
 		monitor->is_primary = TRUE;
 	}
 


libfreerdp/core/gcc.c
```diff
@@ -2198,26 +2198,25 @@ BOOL gcc_write_client_monitor_data(wStream* s, const rdpMcs* mcs)
...
+			const INT32 flags = current->is_primary ? MONITOR_PRIMARY : 0;
 			WLog_DBG(TAG,
 			         "Monitor[%" PRIu32 "]: top=%" PRIu32 ", left=%" PRIu32 ", bottom=%" PRIu32
 			         ", right=%" PRIu32 ", flags=%" PRIu32,
 			         i, top, left, bottom, right, flags);
-			Stream_Write_UINT32(s, left);   /* left */
-			Stream_Write_UINT32(s, top);    /* top */
-			Stream_Write_UINT32(s, right);  /* right */
-			Stream_Write_UINT32(s, bottom); /* bottom */
+			Stream_Write_INT32(s, left);   /* left */
```
akallabeth, line 2216: nice catch, they are signed, correct.

libfreerdp/core/gcc.c
```diff
@@ -2198,26 +2198,25 @@ BOOL gcc_write_client_monitor_data(wStream* s, const rdpMcs* mcs)
...
-			const UINT32 bottom = top + WINPR_ASSERTING_INT_CAST(uint32_t, current->height - 1);
-			const UINT32 flags = current->is_primary ? MONITOR_PRIMARY : 0;
+			const INT32 left = WINPR_ASSERTING_INT_CAST(int32_t, current->x - baseX);
+			const INT32 top = WINPR_ASSERTING_INT_CAST(int32_t, current->y - baseY);
+			const INT32 right = left + WINPR_ASSERTING_INT_CAST(int32_t, current->width - 1);
+			const INT32 bottom = top + WINPR_ASSERTING_INT_CAST(int32_t, current->height - 1);
+			const INT32 flags = current->is_primary ? MONITOR_PRIMARY : 0;
 			WLog_DBG(TAG,
 			         "Monitor[%" PRIu32 "]: top=%" PRIu32 ", left=%" PRIu32 ", bottom=%" PRIu32
 			         ", right=%" PRIu32 ", flags=%" PRIu32,
```
akallabeth, line 2214: format string is now wrong (`PRIu32` vs `PRId32`)

libfreerdp/core/gcc.c
```diff
@@ -2198,26 +2198,25 @@ BOOL gcc_write_client_monitor_data(wStream* s, const rdpMcs* mcs)
...
 
 		for (UINT32 i = 0; i < settings->MonitorCount; i++)
 		{
 			const rdpMonitor* current = &settings->MonitorDefArray[i];
-			const UINT32 left = WINPR_ASSERTING_INT_CAST(uint32_t, current->x - baseX);
-			const UINT32 top = WINPR_ASSERTING_INT_CAST(uint32_t, current->y - baseY);
-			const UINT32 right = left + WINPR_ASSERTING_INT_CAST(uint32_t, current->width - 1);
-			const UINT32 bottom = top + WINPR_ASSERTING_INT_CAST(uint32_t, current->height - 1);
-			const UINT32 flags = current->is_primary ? MONITOR_PRIMARY : 0;
+			const INT32 left = WINPR_ASSERTING_INT_CAST(int32_t, current->x - baseX);
```
akallabeth, line 2207: `left = WINPR_ASSERTING_INT_CAST(` now useless, `current->x` already is of same type

libfreerdp/core/settings.c
```diff
@@ -1763,8 +1763,6 @@ BOOL freerdp_settings_enforce_monitor_exists(rdpSettings* settings)
 		    freerdp_settings_get_pointer_array_writable(settings, FreeRDP_MonitorDefArray, 0);
 		if (!monitor)
 			return FALSE;
-		monitor->x = 0;
```
akallabeth, line 1766: wrong fix, the setting here is to enforce `RDP` requirements.
correct would be to map the `0/0` to native.

akallabeth (changes requested):

libfreerdp/core/gcc.c
```diff
@@ -2198,26 +2198,25 @@ BOOL gcc_write_client_monitor_data(wStream* s, const rdpMcs* mcs)
 			{
 				baseX = current->x;
 				baseY = current->y;
-				break;
```
akallabeth, line 2201: why remove this?

akallabeth:
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/c3964b39-3d54-4ae1-a84a-ceaed311e0f6

akallabeth:
we have 
```
if (!freerdp_settings_set_uint32(settings, FreeRDP_MonitorLocalShiftX,
				                                 WINPR_ASSERTING_INT_CAST(uint32_t, monitor->x)))
					goto fail;
				if (!freerdp_settings_set_uint32(settings, FreeRDP_MonitorLocalShiftY,
				                                 WINPR_ASSERTING_INT_CAST(uint32_t, monitor->y)))
					goto fail;
```
to set the offset shift of the monitors (needs to be checked if it works in all cases)

akallabeth:
@vmpn thx for the try, but there is more to be done:
1. The local monitor layout must set a correct primary
2. the primary offsets must be set (FreeRDP_MonitorLocalShiftX, FreeRDP_MonitorLocalShiftY) on
  * initial connect
  * before display channel updates the remote configuration
3. the `MONITOR_DEF` array must be constructed from the original local layout (primary first, then the other monitors in the array)
4. the FreeRDP_MonitorLocalShiftX, FreeRDP_MonitorLocalShiftY values must be respected for mouse and touch coordinates as well as when drawing. (gfx channel GraphicsReset, rails, ...)

libfreerdp/core/gcc.c
vmpn: :facepalm: will fix, thanks.

libfreerdp/core/gcc.c
vmpn: wrong change left over, after several iterations. Will fix

akallabeth:
@vmpn better check the referenced pr I created today.
should address the issues by properly sorting the monitor array and moving the primary to 0/0, no need for shortcuts.
(still not fully tested, so will take some time before merge)

vmpn:
Thanks for taking time to review and comment, I had a feeling there would be much more to this.

My MVP was to go from recent changes made it impossible to connect to at least somewhat working, while starting to learn the code base.

In regards to you master comment points:

>   The local monitor layout must set a correct primary
Do you mean use `SDL_GetPrimaryDisplay` instead of using first display returned by `SDL_GetDisplays` as primary?



>    the primary offsets must be set (FreeRDP_MonitorLocalShiftX, FreeRDP_MonitorLocalShiftY) on
>
>  * initial connect
>  * before display channel updates the remote configuration

Will need to study code  more. I saw the shift logic local to SdlWindow class, did not realize there were more places.


>    the MONITOR_DEF array must be constructed from the original local layout (primary first, then the other monitors in the array)

Is there particular reason why primary monitor has to be first since there flag to tell which monitor is primary?

>    the FreeRDP_MonitorLocalShiftX, FreeRDP_MonitorLocalShiftY values must be respected for mouse and touch coordinates as well as when drawing. (gfx channel GraphicsReset, rails, ...)
Do you mean replace SdlWindow own _offset_x and _offset_y with common code FreeRDP_MonitorLocalShiftX, FreeRDP_MonitorLocalShiftY?


>for your case it might suffice to set the correct FreeRDP_MonitorLocalShiftX and FreeRDP_MonitorLocalShiftY values instead of removing the monitor->x = 0 part.

Is is generally feasible director to set FreeRDP_MonitorLocalShiftX and FreeRDP_MonitorLocalShiftY and leave actual monitor coordinates as-is, or there is some other complexity I am overlooking?

I am guessing there are some systems that simply provide false coordinates, but would that not be better solved by letting user specify custom override coordinates than hard forcing (0,0) for primary? I feel like I am missing context here

vmpn:
Will take a loo at your reference PR as well, maybe it will answer some of my questions

akallabeth:
@vmpn it is specified that way, see https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/c3964b39-3d54-4ae1-a84a-ceaed311e0f6

vmpn:
replaced by #11095

PR #11095, opened 2025-01-18
Multi monitor fix

akallabeth:
replaces https://github.com/FreeRDP/FreeRDP/pull/11094

now all valid monitor configurations are automatically modified and ordered according to `RDP` rules.


akallabeth:
@vmpn did have a little time to start fixing the multimonitor mess.
so, to address the main issue at its core:
1. I´ve introduced `freerdp_settings_set_monitor_def_array_sorted` that must be used to pass the native monitor configuration and align the resulting array in a `RDP` compatible way
2. fix the `FreeRDP_MonitorLocalShiftX` and `FreeRDP_MonitorLocalShiftY` types to always represent the current offset of monitors from native layout

* I´ve tried to apply that to the `SDL` clients already, but mouse and touch coordinate mapping must respect the `FreeRDP_MonitorLocalShiftX` and `FreeRDP_MonitorLocalShiftY` as well (that is a TODO)
* `xfreerdp` is quite a mess as it grew over a decade and the function must be completely rewritten (TODO)

client/SDL/SDL3/sdl_monitor.cpp
```diff
@@ -260,26 +257,25 @@ static BOOL sdl_apply_display_properties(SdlContext* sdl)
 		    SDL_GetCurrentDisplayOrientation(WINPR_ASSERTING_INT_CAST(uint32_t, *id));
 		const UINT32 rdp_orientation = sdl_orientaion_to_rdp(orientation);
 
-		auto monitor = static_cast<rdpMonitor*>(
-		    freerdp_settings_get_pointer_array_writable(settings, FreeRDP_MonitorDefArray, x));
-		WINPR_ASSERT(monitor);
+		rdpMonitor monitor = {};
 
 		/* windows uses 96 dpi as 'default' and the scale factors are in percent. */
 		const auto factor = dpi / 96.0f * 100.0f;
```
vmpn, line 263: FYI, one thing I noticed in debugging this portion of SDL3 code under kwin-wayland, is the  value called `dpi` will now be 1 for 100% scale and 2 for 200% scale.


https://wiki.libsdl.org/SDL3/README/highdpi


and further any fractional scale is not making it through as wayland API SDL3 calls is integer based.

`SDL_waylandvideo.c`
```
static void display_handle_scale(void *data,
                                 struct wl_output *output,
                                 int32_t factor)
{
    SDL_DisplayData *internal = (SDL_DisplayData *)data;
    internal->scale_factor = factor;
}
```
so dimensions of monitor will scale but scale factor will not.

For now I worked around it by remoting with 100% scale



I do see the fractional scala manager extension is wire up, but did not have time to learn more about it yet.

```
static void display_handle_global(void *data, struct wl_registry *registry, uint32_t id,
                                  const char *interface, uint32_t version)
{
...
    } else if (SDL_strcmp(interface, "wp_fractional_scale_manager_v1") == 0) {
        d->fractional_scale_manager = wl_registry_bind(d->registry, id, &wp_fractional_scale_manager_v1_interface, 1);
```

akallabeth: most likely a TODO on the SDL3 side of things. (still not officially fully stable yet)

akallabeth:
ok, base test with sdl2, sdl3 and xfreerdp work (only got 2 monitors here, so would appreciate some feedback)

akallabeth:
@vmpn does this work for you?
should now use this for `xfreerdp`, `sdl2-freerdp` and `sdl3-freerdp`

vmpn:
@akallabeth I am able to use your PR branch with kwin-wayland and SDL3 client, thank you

akallabeth:
so, lets merge this and fix any remaining issues that might pop up with a follow up pr.

PR #11105, opened 2025-01-22
[android,scripts] update android native build

akallabeth:
* update dependencies to current versions
* build openh264 with static runtime
* use curl for download instead of wget


akallabeth:
@licaon-kter ok, this should address the remaining issue (`openh264` runtime now static)

akallabeth:
@freerdp-bot test

akallabeth:
@licaon-kter ok, this looks like the android components also need an update :/

akallabeth:
@licaon-kter tracking the android UI related stuff here: https://github.com/FreeRDP/FreeRDP/issues/11106

PR #11122, opened 2025-01-30
[client] Add command line flag to disable Ctrl+Alt+M hotkey to minimize application

LordOfTheRats:
I added a command line flag to disable the Ctrl+Alt+M hotkey for minimizing the application. Some of our users need this hotkey inside the RDP session.

I guess include/freerdp/settings_types_private.h should be updated to include the new setting into the stable ABI.


akallabeth (changes requested):
I´d prefer an approach like with the SDL client, where we allow a per user configuration file to provide custom hotkeys.
keeps client specifics out of the `rdpSettings` (yea, already too late for that, but maybe we can improve it a little :grin: )

PR #11145, opened 2025-02-07
[core,orders] fix update_read_delta

akallabeth:
@lemonmojo I think your issue is fixed with this.


lemonmojo:
@akallabeth That did the trick! 🥳

PR #11147, opened 2025-02-09
[build,android] only enable required codecs for ffmpeg

akallabeth:
the ffmpeg build script now only enables required audio/video codecs to save some space
@licaon-kter this should reduce the apk size a lot

### DIFF
--- a/scripts/android-build-ffmpeg.sh
+++ b/scripts/android-build-ffmpeg.sh
@@ -122,11 +122,47 @@ function build {
         ${ARCH_OPTIONS} \
         --enable-cross-compile \
         --enable-pic \
-        --enable-jni --enable-mediacodec \
+        --enable-jni \
+	--enable-mediacodec \
         --enable-shared \
         --disable-vulkan \
         --disable-stripping \
-        --disable-programs --disable-doc --disable-avdevice --disable-avfilter --disable-avformat
+        --disable-programs \
+	--disable-doc \
+	--disable-avdevice \
+	--disable-avfilter \
+	--disable-avformat \
+	--disable-everything \
+	--enable-encoder=aac \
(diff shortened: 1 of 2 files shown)


licaon-kter:
Indeed now the APK is only 55Mb :tada:

licaon-kter:
@akallabeth should I amend the build line and remove `--openh264` as that's the default now?

akallabeth:
@licaon-kter yes, the handling for `openh264` no longer requires extra `NDK` and can be build with the default one now.

licaon-kter:
@akallabeth I just tested that and by removing `--openh264` the APK is missing `libopenh264.so`, is that as it should be?

In the log I see
```
WARNING: Disabled libopenh264_encoder because not all dependencies are satisfied: libopenh264
...
-- Skipping optional feature OpenH264 for codec (use OpenH264 library)
--     Enable feature OpenH264 using "-DWITH_OPENH264=ON"
...
```

akallabeth:
> @akallabeth I just tested that and by removing `--openh264` the APK is missing `libopenh264.so`, is that as it should be?
> 
> In the log I see
> 
> ```
> WARNING: Disabled libopenh264_encoder because not all dependencies are satisfied: libopenh264
> ...
> -- Skipping optional feature OpenH264 for codec (use OpenH264 library)
> --     Enable feature OpenH264 using "-DWITH_OPENH264=ON"
> ...
> ```

do you build with the proper config file `scripts/android-build-release.conf`?

licaon-kter:
> do you build with the proper config file scripts/android-build-release.conf?

I run `scripts/android-build-freerdp.sh --ndk $$NDK$$ --sdk $$SDK$$ --release` 

not sure if/what that uses

PR #11163, opened 2025-02-11
[primitives,sse] split headers

akallabeth:
Split out all SSL inline functions and defines from prim_internal.h. The header is included in files build without -msse and on some platforms compilation will fail due to that.


mfleisz:
I think using `LOAD_SI128` all over the place will turn out to be quite ineffective. Checking for alignment on each load might as well totally trash any speed advantage we gain from using an aligned vs an unaligned load.

Also, after reading https://stackoverflow.com/a/38383624 the suggestion is to go with `_mm_lddqu_si128` instead of `_mm_load_si128` anyways.

akallabeth:
> I think using `LOAD_SI128` all over the place will turn out to be quite ineffective. Checking for alignment on each load might as well totally trash any speed advantage we gain from using an aligned vs an unaligned load.
> 
> Also, after reading https://stackoverflow.com/a/38383624 the suggestion is to go with `_mm_lddqu_si128` instead of `_mm_load_si128` anyways.

true, replaced with unaligned access all the way (the aligned case does not really trigger in real word use cases and modern processors do not really have a huge difference on load/store times for unaligned)

bumped the requirement to `SSE3` to be able to use the newer load functions though.

mfleisz (commented):
Got two questions:
- Why do we actually still have these LOAD_ and STORE_ macros? The initial reason for LOAD was checking for unaligned/aligned loads which we do not need anymore.
- What is the reason for all these new `*_int` functions?

sur5r:
> * What is the reason for all these new `*_int` functions?

This is due to a problem I hit yesterday. When the check whether to enable certian intrinsics/primitives (SSE4.1 in my case) happens INSIDE that compilation unit, the compiler might (and did) emit instructions (`pinsrq`), a non-SSE4.1 CPU does not understand, leading to SIGILL, rendering the check moot:

```
$ objdump --dwarf=follow-links --disassemble=primitives_init_YUV_sse41.constprop.0 /usr/lib/x86_64-linux-gnu/libfreerdp3.so.3 
...
000000000015ff00 <primitives_init_YUV_sse41.constprop.0>:
  15ff00:	55                   	push   %rbp
  15ff01:	48 89 e5             	mov    %rsp,%rbp
  15ff04:	e8 c7 cc ed ff       	call   3cbd0 <primitives_get_generic@plt>
  15ff09:	bf 04 00 00 00       	mov    $0x4,%edi
  15ff0e:	f3 0f 7e 05 62 e7 09 	movq   0x9e762(%rip),%xmm0        # 1fe678 <g_Subsystem_libavcodec+0x98>
  15ff15:	00 
  15ff16:	48 89 05 2b 7f 0e 00 	mov    %rax,0xe7f2b(%rip)        # 247e48 <generic.lto_priv.9>
  15ff1d:	48 8d 05 dc 52 00 00 	lea    0x52dc(%rip),%rax        # 165200 <general_RGBToYUV420_8u_P3AC4R>
  15ff24:	66 48 0f 3a 22 c0 01 	pinsrq $0x1,%rax,%xmm0
  15ff2b:	48 8d 05 3e 86 ff ff 	lea    -0x79c2(%rip),%rax        # 158570 <general_YUV420CombineToYUV444>
  15ff32:	0f 11 05 9f 81 0e 00 	movups %xmm0,0xe819f(%rip)        # 2480d8 <pPrimitivesCpu+0xb8>
  15ff39:	f3 0f 7e 05 3f e7 09 	movq   0x9e73f(%rip),%xmm0        # 1fe680 <g_Subsystem_libavcodec+0xa0>
  15ff40:	00 
  15ff41:	66 48 0f 3a 22 c0 01 	pinsrq $0x1,%rax,%xmm0
  15ff48:	48 8d 05 21 5e ff ff 	lea    -0xa1df(%rip),%rax        # 155d70 <general_YUV444ToRGB_8u_P3AC4R>
  15ff4f:	0f 11 05 92 81 0e 00 	movups %xmm0,0xe8192(%rip)        # 2480e8 <pPrimitivesCpu+0xc8>
  15ff56:	f3 0f 7e 05 2a e7 09 	movq   0x9e72a(%rip),%xmm0        # 1fe688 <g_Subsystem_libavcodec+0xa8>
  15ff5d:	00 
  15ff5e:	66 48 0f 3a 22 c0 01 	pinsrq $0x1,%rax,%xmm0
  15ff65:	48 8d 05 a4 75 00 00 	lea    0x75a4(%rip),%rax        # 167510 <general_RGBToAVC444YUVv2>
  15ff6c:	0f 11 05 85 81 0e 00 	movups %xmm0,0xe8185(%rip)        # 2480f8 <pPrimitivesCpu+0xd8>
  15ff73:	f3 0f 7e 05 15 e7 09 	movq   0x9e715(%rip),%xmm0        # 1fe690 <g_Subsystem_libavcodec+0xb0>
  15ff7a:	00 
  15ff7b:	66 48 0f 3a 22 c0 01 	pinsrq $0x1,%rax,%xmm0
  15ff82:	0f 11 05 7f 81 0e 00 	movups %xmm0,0xe817f(%rip)        # 248108 <pPrimitivesCpu+0xe8>
  15ff89:	e8 52 b9 ed ff       	call   3b8e0 <IsProcessorFeaturePresentEx@plt>
  15ff8e:	85 c0                	test   %eax,%eax
```

In the above assembly excerpt, `pinsrq` is used before `IsProcessorFeaturePresentEx` is even called.

That's why @akallabeth split those to make the runtime check actually work. Thanks again btw!

mfleisz:
@sur5r now that makes sense - thanks for the hint!

libfreerdp/primitives/sse/prim_templates.h
```diff
@@ -86,21 +83,21 @@
 		{                                                                             \
```
mfleisz, line 83: I think we can ditch this check (and half of the code below) after the lddq/load changes

mfleisz (commented):
prim_templates still needs some more love:
- There is still `SSE3_SSD_ROUTINE` and `SSE3_SCD_PRE_ROUTINE` checking for aligned/unaligned load
- Each macro takes a `_slowWay_` argument that should be eliminated altogether (as well as all the accompanying `dptr` alignment checks)

akallabeth:
@mfleisz done, removed all alignment checks and the slow prefix iteration.

mfleisz (approved):

PR #11166, opened 2025-02-12
[warnings] fix sign conversion

akallabeth:


libfreerdp/core/server.c
```diff
@@ -1153,7 +1155,7 @@ BOOL WINAPI FreeRDP_WTSQuerySessionInformationA(HANDLE hServer, DWORD SessionId,
 
 		if (!pBuffer)
 		{
-			SetLastError(E_OUTOFMEMORY);
+			SetLastError(g_err_oom);
```
mfleisz, line 1158: We should use the correct error code (`ERROR_NOT_ENOUGH_MEMORY`) instead of casting this (correct) warning away

PR #11167, opened 2025-02-12
Fix all unused warnings

akallabeth:
* should only contain `WINPR_ATTR_UNUSED` additions and some `WLog_ERR("TODO", "TODO: implement");` log messages
* Rationale for silencing these is that the `API` requires these parameters set but we currently do not use them. Do not emit warnings during compile for these, but log something instead so that in case of use the user is warned.
* check https://ci.freerdp.com/job/code-quality-checker/9845/ we´re down to nearly nothing with regards to compiler warnings (most of the remaining being backward compatibility warnings for `C99` and such)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #11193, opened 2025-02-16
[client,sdl] Fix keyboard grab toggle logic

caseif:
This PR fixes an issue with the logic for toggling the keyboard grab state in the SDL 3 client. Previously the `grab_kbd` field was being passed as a parameter to `keyboard_grab`, which doesn't make sense because this flag is only ever mutated in that same function using the same parameter. This makes it impossible for the field to ever be set as `true`. The `grab_kbd_enabled` field seems to be the correct input parameter, given that it is toggled immediately before the call to `keyboard_grab`.


akallabeth (changes requested):
`_sdl->grab_kbd` keeps the current state of the grab, `_sdl->grab_kbd_enabled` if grabbing is enabled.
your fix looks wrong.

akallabeth:
@caseif thx for the pr, replaced it with https://github.com/FreeRDP/FreeRDP/pull/11195

PR #11196, opened 2025-02-17
[channel,rdpecam] Optional format passthrough (no H264 encoding)

oleg0421:
Adding new cmake option:
```
 -DRDPECAM_FORMAT_PASSTHROUGH=ON
 ```
to allow any camera format passthrough without encoding to H264. Useful for virtual machine environments where network bandwidth is not an issue. Helps to close https://github.com/FreeRDP/FreeRDP/issues/10935

### DIFF
--- a/channels/rdpecam/client/camera.h
+++ b/channels/rdpecam/client/camera.h
@@ -89,15 +89,8 @@ typedef struct
 
 typedef struct
 {
-	CAM_MEDIA_FORMAT inputFormat;  /* camera side */
-	CAM_MEDIA_FORMAT outputFormat; /* network side */
-
-} CAM_MEDIA_FORMAT_INFO;
-
-typedef struct
-{
 	BOOL streaming;
-	CAM_MEDIA_FORMAT_INFO formats;
+	CAM_MEDIA_FORMAT inputFormat; /* camera side */
 	CAM_MEDIA_TYPE_DESCRIPTION currMediaType;
 
 	GENERIC_CHANNEL_CALLBACK* hSampleReqChannel;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: channels/rdpecam/client/camera.h [hunk 1]
```diff
@@ -124,11 +117,16 @@ typedef struct
...
-	return stream->formats.inputFormat;
+	return stream->inputFormat;
 }
 static INLINE CAM_MEDIA_FORMAT streamOutputFormat(CameraDeviceStream* stream)
 {
-	return stream->formats.outputFormat;
+#if !defined(WITH_FORMAT_PASSTHROUGH)
+	return CAM_MEDIA_FORMAT_H264;
+#else
+	return (stream->inputFormat == CAM_MEDIA_FORMAT_MJPG_H264) ? CAM_MEDIA_FORMAT_H264
```
akallabeth, line 127: did miss this in the last pr, `CAM_MEDIA_FORMAT_MJPG_H264` is kind of a misuse of the `enum`.
no other (clean) way to use this? (e.g. split input & output format?)

akallabeth (changes requested):
this looks more like a runtime option to use (e.g. `/dev:rdpecam:format:mjpg` or similar) instead of doing this at compile time

oleg0421:
@akallabeth: command line option can be implemented. We need to keep some **limitations** in mind:

On the input (camera) side, the formats are not equal. Camera dependent choices, in priority order, are:
 - H264 advertised;
 - H264 muxed (Logitech specific);
 - MJPG;
 - one of the raw formats, such as YUYV.
 
H264 provides highest compression and it is **the only format** used on the output (network) side. We cannot offer MJPG (medium compression) or raw formats due to huge bandwidth requirements. The only exception is when RDP is used in virtual machine environment, where bandwidth is not an issue.

Also note, many cameras don't support H264 and some don't even support MJPG, and we can't do encoding from raw format to MJPG, for example. In my opinion, the only useful CLI options are the following:

1. Disable some input formats, i.e. disable H264 or disable MJPG, needed primarily for development and workarounds if there are any issues with a particular format. This is already done with cmake options.
2. Disable H264 encoding, do passthrough of source format instead. As mentioned, practical for VM environment only.

I wanted to do option 2 in this PR also as cmake option, mostly because it has rather limited use case and also because I'm not familiar with FreeRDP command line handling and requirements.

Even though command line parameters are generally useful, the complexity gets increased, more testing is required considering all possible combinations.

We can leave this PR as a reference for someone who understands the above mentioned details and is willing to implement CLI options.

Defenso-QTH:
This is a highly desired feature for my use case and would solve issue #10935.

I would prefer a runtime switch but can accept a buildtime option as a working alternative, temporary or permanent.

Defenso-QTH:
This was cancelled?

oleg0421:
@Defenso-QTH: not cancelled. Sorry, busy with day job and other things. It's an open source. Anyone can pick it up, see comment above.

PR #11226, opened 2025-02-21
[core,credssp_auth] Fix faulty string length check in `credssp_auth_client_init_cred_attributes`

thenextman:
[78acedb40ee275b6aadcb6a5fbfc8922ea30835a](https://github.com/FreeRDP/FreeRDP/commit/78acedb40ee275b6aadcb6a5fbfc8922ea30835a) introduced a faulty string length check which prevents using an explicit KDC URL.

The bounds check uses the `<=` operator instead of `>=`, allowing only strings _longer_ than `UINT16_MAX / 2`.


akallabeth (approved):
LGTM nice catch!
thank you for the fix!

akallabeth:
@freerdp-bot test

PR #11241, opened 2025-02-25
CMake: generate a .gitignore file for the build directory

yegorich:
As soon as a configuration step is successful, a `.gitignore` file is created inside an out-of-source build directory. Hence, there is no need to put its name into the root `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 **/CMakeCache.txt
 **/CMakeFiles
-build
 checker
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM
nice patch, thank you!

yegorich:
Btw, meson creates this file by default...

PR #11243, opened 2025-02-25
[client,cmdline] fix port parsing for gateway

THS-on:
9d8954f missed also adding the port to the settings.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM
thank you for the fix!

PR #11247, opened 2025-02-25
CMake: automatically create compile_commands.json file

yegorich:
This file is required to enable such features as code completion, diagnostics, go-to-definition, etc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,9 @@ if(NOT PROJECT_SOURCE_DIR STREQUAL PROJECT_BINARY_DIR)
   file(GENERATE OUTPUT .gitignore CONTENT "*")
 endif()
 
+# Create compile_commands.json file
+set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
+
 if(NOT DEFINED VENDOR)
   set(VENDOR "FreeRDP" CACHE STRING "FreeRDP package vendor")
 endif()


akallabeth (changes requested):
if you need it you can always run `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ...` as configure step.

so what is the reason to force it on for all builds?

yegorich:
I made some measurements with and without  `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`:

```
real    0m7,815s
user    0m3,449s
sys     0m10,536s

```
```
real    0m7,894s
user    0m3,445s
sys     0m10,589s

```

As one can see, there is no real difference. Hence, the overhead is minimal and the developers have everything ready for their IDE.

P.S. meson build system creates `compile_commands.json` by default. Thus, I'm always missing it when dealing with CMake :-)

hardening (approved):

hardening:

CMakeLists.txt
```diff
@@ -32,6 +32,9 @@ if(NOT PROJECT_SOURCE_DIR STREQUAL PROJECT_BINARY_DIR)
   file(GENERATE OUTPUT .gitignore CONTENT "*")
 endif()
 
+# Create compile_commands.json file
+set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
```
akallabeth, line 36: ok, justification accepted.

please change to `option(CMAKE_EXPORT_COMPILE_COMMANDS "somedescription" ON)` to allow disabling it for the ones that want to.

akallabeth (changes requested):

CMakeLists.txt
yegorich: @akallabeth this doesn't work as it is already an internal CMake option :-(

akallabeth: replaced by https://github.com/FreeRDP/FreeRDP/pull/11252

PR #11248, opened 2025-02-25
Add EditorConfig settings file

yegorich:
EditorConfig helps keeping coding styles consistent across different editors and IDEs.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[{CMakeLists.txt,*.cmake}]
+indent_style = space
+indent_size = 2
+tab_width = 2
+charset = utf-8
+max_line_length = 100
+end_of_line = lf
+trim_trailing_whitespace = true
+insert_final_newline = true


akallabeth (changes requested):
* not sure where the benefit over `.clang-format` is here
* the coding style is not consistent with the one we have now
* applying the style to all file types might be a bit too much (some are resources)

yegorich:
Do I understand it correctly that `.clang-format` checks only C code? This config handles only CMake related files so far.

EditorConfig also supports editors that don't mecessary support `.clang-format`.

I moved all settings inside CMake file filter.

yegorich:
What values were suitable for CMake files? Should the length be extended to 120?

hardening:
> What values were suitable for CMake files? Should the length be extended to 120?

Well check .clang-format we enforce various things there

yegorich:
@akallabeth I set `max_line_length` to 100 as configured in .`clang-format`

akallabeth:
@yegorich you still did not answer what the whole thing would be good for.
since the functionality already exists with our current `.clang-format`, `cmake-format` and `shfmt` usage.
from what I read the whole thing does not really have a usable command line client (ant/maven/gradle are a huge overhead)

akallabeth:
@yegorich so basically a config file that will run out of sync with the ones actually used by the ci?

yegorich:
@akallabeth yes, it will run out-of-sync. But if you compare `.editorconfig` and `.clang-format`, there are not so many options and they are not likely to change.

yegorich:
Ok. Then I'm closing this PR. This is just a nice to have feature after all.

bmiklautz:
I agree with @akallabeth duplicated changes are usually extra work on the long run or something that get's forgotten. 
However if `.editorconfig` is gaining attraction and can be used **without** plugin it would be nice to have. 

clang-format will stay our main reference for formatting but what about adding some comment in `.editorconfig` to note that the file might be out of sync and  comments in `.clang-format`  behind the corresponding setting with a note that `.editorconfig` needs update when the setting is  change. 


@yegorich thank you for the contribution.

yegorich:
The most editors like Visual Studio, neovim, xcode, GitHub, GirLab etc. (see the list on the project's page) support .editorconfig internally i.e. without the need of a plugin.

PR #11256, opened 2025-02-26
[shadow,kerberos] Fix Shadow crashing when calling krb5_get_init_cred…

XiaopengZHOU:
…s_keytab if no suitable entry is found in keytab.

In the **kerberos_AcceptSecurityContext()** function of **winpr/libwinpr/sspi/Kerberos/kerberos.c** file, there is a _do ... while_ loop iterating through the _keytab_ entries. If there are indeed entries in the _keytab_ but none are suitable, the read entries are freed with the **krb5glue_free_keytab_entry_contents()** function. But the **entry** data structure is not erased. This means that when the loop exits with the **rv == KRB5_KT_END** condition, the structure still contains invalid data (dangling pointer) and will cause **krb5_get_init_creds_keytab()** call to crash.


akallabeth (changes requested):
nice catch!

but I think the reset should happen after `krb5glue_free_keytab_entry_contents` regardless of success or not (or you'll end in a double free)

akallabeth:
ok, did cherry-pick your commit and changed the `memset` to be always done after free. See #11257

PR #11268, opened 2025-02-28
[server,shadow] readd multi rect BitmapUpdate support

akallabeth:
* Partially revert 1f83198bb1e5d7e8dfecae0e6511f88a7888c8c4
* Introduce new /bitmap-compat command line option for shadow-server
* Default to send multi rectangle BitmapUpdate


akallabeth:
@ggirol-rc does this now work for you without this patch when february updates are installed?

PR #11287, opened 2025-03-05
client: Move buffer pointer after writing RDP settings

mfleisz:


client/common/file.c
```diff
@@ -1685,12 +1685,16 @@ size_t freerdp_client_write_rdp_file_buffer(const rdpFile* file, char* buffer, s
 	if (intsize < 0)
 		return 0;
 	totalSize += (size_t)intsize;
+	if (buffer)
```
akallabeth, line 1688: * `NULL` check not required (done in line 1680)
* might be easier to always use `&buffer[totalSize]` on write functions

mfleisz: The NULL check is required in case the function is called with `buffer=NULL, size=0` (in order to calculate the size of the required buffer). If we do not check for `NULL` here the calls to`write_string_parameters`/`write_custom_parameters` will fail if we pass them some random buffer value.

akallabeth (approved):

PR #11296, opened 2025-03-06
Invert 3x deprecated

akallabeth:


CMakeLists.txt
```diff
@@ -163,9 +163,14 @@ if(WITH_FREERDP_DEPRECATED)
   add_compile_definitions(WITH_FREERDP_DEPRECATED)
 endif()
 
-option(WITH_FREERDP_3x_DEPRECATED "Build FreeRDP 3x deprecated symbols" ON)
+option(DROP_FREERDP_3x_DEPRECATED "Build FreeRDP 3x deprecated symbols" OFF)
```
mfleisz, line 166: Not sure it's a good idea to drop the `WITH_` scheme here (makes it harder to find the option in tools like ccmake) - maybe we could use something like `WITH_FREERDP_3x_DEPRECATED_REMOVED` or something along that line.

akallabeth: well, finding is easy, knowing what to look for not so much.
but since the logic is inverted a `WITH_*` is definitely misleading.

mfleisz: Just thought it should be in line with the existing `WITH_FREERDP_DEPRECATED` - but looking at that setting it seems to be broken the same way as `WITH_FREERDP_3x_DEPRECATED` is. 🤔

akallabeth: @mfleisz well, for `WITH_FREERDP_DEPRECATED` it is intended (e.g. build without by default)

akallabeth:
@ondrejholy maybe of interest for you: we did have a quirk in the last 3.13.0 release deactivating all deprecated functions unless `WITH_FREERDP_3x_DEPRECATED` is defined (by every application that uses `FreeRDP`)
this pr fixes that.

PR #11301, opened 2025-03-06
missing ConnectFlags variable in license_read_platform_challenge_packet breaks in WITH_DEBUG_LICENSE builds

eduar-hte:
`ConnectFlags` was removed in commit 9f5a00e, but it's still being logged when building `WITH_DEBUG_LICENSE` (see [here](https://github.com/FreeRDP/FreeRDP/blob/52fec2cd037987071afca501c724c655426a1c42/libfreerdp/core/license.c#L1982))


akallabeth:
@eduar-hte nice catch!

PR #11311, opened 2025-03-09
[winpr,nt] Fix incorrect name in FILE_INFORMATION_CLASS

Biswa96:
This renames FileUnknownInformation1 with FileCompletionInformation which is documented in https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ne-wdm-_file_information_class


akallabeth (approved):
LGTM

PR #11316, opened 2025-03-10
[cmake] Fix DLL install directory

Biswa96:
This installs DLL in /bin directory in mingw environment. Previously, DLL was installed in /lib. Those need to be mentioned explicitly, see https://cmake.org/cmake/help/latest/command/install.html


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #11384, opened 2025-03-20
Rel mouse change

akallabeth:
@pnowack ok, adjusted the behavior to only send relative if it is enabled in the client and there is no cursor visible.

### DIFF
--- a/client/X11/xf_input.c
+++ b/client/X11/xf_input.c
@@ -50,7 +50,8 @@
 
 #define MIN_FINGER_DIST 5
 
-static int xf_input_event(xfContext* xfc, const XEvent* xevent, XIDeviceEvent* event, int evtype);
+static int xf_input_event(xfContext* xfc, WINPR_ATTR_UNUSED const XEvent* xevent,
+                          XIDeviceEvent* event, int evtype);
 
 #ifdef DEBUG_XINPUT
 static const char* xf_input_get_class_string(int class)
@@ -744,26 +745,25 @@ static int xf_input_pens_unhover(xfContext* xfc)
 	return 0;
 }
 
-int xf_input_event(xfContext* xfc, const XEvent* xevent, XIDeviceEvent* event, int evtype)
+static bool xf_use_rel_mouse(xfContext* xfc)
+{
+	if (!freerdp_client_use_relative_mouse_events(&xfc->common))
+		return false;
+	if (!xfc->isCursorHidden)
+		return false;
+	return true;
+}
+
+int xf_input_event(xfContext* xfc, WINPR_ATTR_UNUSED const XEvent* xevent, XIDeviceEvent* event,
+                   int evtype)
 {
 	WINPR_ASSERT(xfc);
 	WINPR_ASSERT(xevent);
 	WINPR_ASSERT(event);
 
-	/* When not running RAILS we only care about events for this window.
-	 * filter out anything else, like floatbar window events
-	 */
-	const Window w = xevent->xany.window;
-
 	xfWindow* window = xfc->window;
 	if (window)
 	{
-		if (w != window->handle)
-		{
-			if (!xfc->remote_app)
-				return 0;
-		}
-
 		if (xf_floatbar_is_locked(window->floatbar))
 			return 0;
 	}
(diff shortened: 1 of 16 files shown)


client/X11/xf_input.c
```diff
@@ -785,8 +793,7 @@ int xf_input_event(xfContext* xfc, const XEvent* xevent, XIDeviceEvent* event, i
 			break;
 
 		case XI_Motion:
-			xfc->xi_event = !xfc->common.mouse_grabbed ||
-			                !freerdp_client_use_relative_mouse_events(&xfc->common);
+			xfc->xi_event = !xfc->common.mouse_grabbed || !xf_use_rel_mouse(xfc);
```
pnowack, line 796: I think absolute mouse ptr motion can always be sent. Reason is if the e.g. not a game is played inside the remote session, but video is watched in the browser or player. When the cursor is inside the window, it will become invisible while watching the video. When moving it again, it becomes visible.
Due to network latency it can happen, that the cursor shape is updated after the move events are sent. The actual mouse ptr might here not match the position where it is actually at. In that case, an extra absolute mouse ptr event is needed.
I think both client and server side need to track the cursor-shape-visible state here. The server side could always save the last absolute event, and when this transition from invisible-cursor to visible-cursor happens, it can reuse the last absolute mouse ptr event.

client/X11/xf_input.c
```diff
@@ -774,8 +783,7 @@ int xf_input_event(xfContext* xfc, const XEvent* xevent, XIDeviceEvent* event, i
 	{
 		case XI_ButtonPress:
 		case XI_ButtonRelease:
-			xfc->xi_event = !xfc->common.mouse_grabbed ||
-			                !freerdp_client_use_relative_mouse_events(&xfc->common);
+			xfc->xi_event = !xfc->common.mouse_grabbed || !xf_use_rel_mouse(xfc);
```
pnowack, line 786: (same comment here, like the one for the `XI_Motion` case)

pnowack:
I think this needs changes in `[client/common/client.c](https://github.com/FreeRDP/FreeRDP/pull/9677/files#diff-2d94332c2a112728e85ee57b6e35cee103fcdf0d3634af6e1b10f915d095ccdf)`. In #9677, the FreeRDP client code was adjusted to either always send relative events or absolute events, based on whether the server side advertised relative mouse input support.

I think we should always send absolute events, and when we have the invisible cursor situation, then send relative mouse motion events too, if the server initially advertised support for it.

client/X11/xf_input.c
akallabeth: we should test this, how the windows end does react to such a thing.

akallabeth:
@pnowack ok, update the pr:
1. add relative mouse input support to shadow server
2. fix some bugs on the x11 side

as for sending both, relative and absolute, I don't think that is feasible.
the server has no way of knowing which data to discard.

if the server did have such a thing then the whole hidden mouse pointer thing could be avoided anyway.

pnowack:
> as for sending both, relative and absolute, I don't think that is feasible.
the server has no way of knowing which data to discard.

Actually it is feasible. The server side also knows, which cursor type is currently used. So, it can select which event type is needed

> if the server did have such a thing then the whole hidden mouse pointer thing could be avoided anyway.

True, though less data to transmit, when the cursor is visible.

Won't have time today and tomorrow for another test unfortunately. Will try again on the weekend

akallabeth:
@pnowack do you happen to know a client and a server that send relative mouse input we can use for reference comparisons?

akallabeth:
@pnowack ok, tested relative mouse input now with the shadow server.
Did patch this branch with:
```
git diff
diff --git a/server/shadow/shadow_client.c b/server/shadow/shadow_client.c
index 06f6c4da2..640891128 100644
--- a/server/shadow/shadow_client.c
+++ b/server/shadow/shadow_client.c
@@ -37,6 +37,8 @@
 
 #define TAG CLIENT_TAG("shadow")
 
+#define WITH_NULL_POINTER
+
 typedef struct
 {
        BOOL gfxOpened;
@@ -2179,7 +2181,13 @@ static int shadow_client_subsystem_process_message(rdpShadowClient* client, wMes
                        if (client->activated)
                        {
                                IFCALL(update->pointer->PointerNew, context, &pointerNew);
+#if defined(WITH_NULL_POINTER)
+                               POINTER_SYSTEM_UPDATE pointer_system = { 0 };
+                               pointer_system.type = SYSPTR_NULL;
+                               IFCALL(update->pointer->PointerSystem, context, &pointer_system);
+#else
                                IFCALL(update->pointer->PointerCached, context, &pointerCached);
+#endif
                        }
 
                        break;
```

to force the pointer `NULL`  condition to be met.

works fine with `sdl3-freerdp` and `xfreerdp` so far.

akallabeth:
@pnowack any luck finding any microsoft client that supports that?
tried mstsc from a windows 11 24h2 and that one does not announce relative mouse support

akallabeth:
@pnowack ok, updated the changes to the shadow server to allow hiding the mouse cursor via command line.
so, now we have a fully functioning client/server sample to test this.

I'd vote for merging this as is and once we have a reference client we can check if we could send absolute mouse updates alongside relative ones.

PR #11433, opened 2025-03-27
[channels,rdpdr] relax state checks for PAKID_CORE_CLIENTID_CONFIRM

akallabeth:
This missage might be sent before PAKID_CORE_SERVER_CAPABILITY by some servers. Do accept these.


akallabeth:
@pacomateos10 does this work for you? I fail to reproduce the inverted message order, but reevaluating [[MS-RDPEFS]](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpefs/34d9de58-b2b5-40b6-b970-f82d4603bdb5) revealed that this message order is not strictly forbidden (just not easy to spot in the documentation)

akallabeth:
@pacomateos10 did update the pr, please recheck.

PR #11435, opened 2025-03-27
nla: send user and domain hints with smartcard logon

hardening:
When logging with a smartcard logon, the TS Credentials packet can also contains hints of the user and the domain that were used. This patch adds these informations.


akallabeth (approved):
LGTM

PR #11443, opened 2025-03-31
Various fixes related to smartcard logon server-side

hardening:
2 commits to:
*  remove a verbose log in ncrypt that happens because we're calling a function to know the size of the target string
* fix server-side user2user kerberos (also removed some unneeded `krb_log_exec_bool` logs)

### DIFF
--- a/winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
+++ b/winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
@@ -541,7 +541,8 @@ static BOOL convertKeyType(CK_KEY_TYPE k, LPWSTR dest, DWORD len, DWORD* outlen)
 			dest[0] = 0;
 		return FALSE;
 	}
-	else
+
+	if (dest)
 	{
 		if (retLen + 1 > len)
 		{
@@ -549,11 +550,8 @@ static BOOL convertKeyType(CK_KEY_TYPE k, LPWSTR dest, DWORD len, DWORD* outlen)
 			return FALSE;
 		}
 
-		if (dest)
-		{
-			memcpy(dest, r, sizeof(WCHAR) * retLen);
-			dest[retLen] = 0;
-		}
+		memcpy(dest, r, sizeof(WCHAR) * retLen);
+		dest[retLen] = 0;
 	}
 
 	return TRUE;
--- a/winpr/libwinpr/sspi/Kerberos/kerberos.c
+++ b/winpr/libwinpr/sspi/Kerberos/kerberos.c
@@ -127,9 +127,6 @@ static const WinPrAsn1_OID kerberos_OID = { 9, (void*)"\x2a\x86\x48\x86\xf7\x12\
 static const WinPrAsn1_OID kerberos_u2u_OID = { 10,
 	                                            (void*)"\x2a\x86\x48\x86\xf7\x12\x01\x02\x02\x03" };
 
-#define krb_log_exec_bool(fkt, ctx, ...)                                                     \
-	kerberos_log_msg(ctx, fkt(ctx, ##__VA_ARGS__) ? KRB5KRB_ERR_GENERIC : 0, #fkt, __FILE__, \
-	                 __func__, __LINE__)
 #define krb_log_exec(fkt, ctx, ...) \
 	kerberos_log_msg(ctx, fkt(ctx, ##__VA_ARGS__), #fkt, __FILE__, __func__, __LINE__)
 #define krb_log_exec_ptr(fkt, ctx, ...) \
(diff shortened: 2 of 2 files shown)


winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
```diff
@@ -543,7 +543,7 @@ static BOOL convertKeyType(CK_KEY_TYPE k, LPWSTR dest, DWORD len, DWORD* outlen)
 	}
 	else
 	{
-		if (retLen + 1 > len)
+		if (len && retLen + 1 > len)
```
akallabeth, line 546: Looks buggy:
1. you skip this if `len == 0`
2. a few lines below a `memcpy()` with a (unchecked) `retLen >= 0`?

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -1356,10 +1388,9 @@ static SECURITY_STATUS SEC_ENTRY kerberos_AcceptSecurityContext(
 			if (rv != 0)
 				goto cleanup;
 
-			if ((!sname || krb_log_exec_bool(krb5_principal_compare_any_realm, credentials->ctx,
-			                                 principal, entry.principal)) &&
-			    (!realm || krb_log_exec_bool(krb5_realm_compare, credentials->ctx, principal,
-			                                 entry.principal)))
+			if ((!sname ||
+			     krb5_principal_compare_any_realm(credentials->ctx, principal, entry.principal)) &&
```
akallabeth, line 1392: maybe better modify this to only log with `DEBUG/TRACE` instead of `ERROR`?

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -1325,13 +1325,45 @@ static SECURITY_STATUS SEC_ENTRY kerberos_AcceptSecurityContext(
...
+		 *  |          |
+		 *  v          v
+		 *  <host>$\x00<REALM>
+		 *
+		 */
 		if (target)
 		{
-			if (*target != 0 && *target != '@')
-				sname = target;
+			char* sname = strchr(target, '/');
```
akallabeth, line 1341: maybe easier to read as
```
if (target)
		{
			if (strncmp(target, "TERMSRV/", 8) != 0)
				goto cleanup;
			const char* sname = &target[8];
			const char* realm = strchr(sname, '@');
			if (!realm)
				goto cleanup;

			const intptr_t diff = realm - sname;
			realm++;

			char* str = NULL;
			size_t strlen = 0;
			(void)winpr_asprintf(&str, &strlen, "%.*s$@%s", diff, sname, realm);
			if (!str)
				goto cleanup;
			free(target);
			target = str;
		}
```

hardening: Hum I'm gonna take inspiration of that, but the realm is not always present, so I must adapt that.

akallabeth: @hardening just a suggestion though, your code is working as well.

winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
akallabeth: @hardening I think the check `if (len > 0)` should be part of the else above.

hardening: reworked that.

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -1325,13 +1325,45 @@ static SECURITY_STATUS SEC_ENTRY kerberos_AcceptSecurityContext(
...
+
 			realm = strchr(target, '@');
 			if (realm)
+			{
+				*realm = '$';
 				realm++;
+
+				size_t len = strlen(realm);
+				memmove(realm + 1, realm, len + 1);
+				*realm = 0;
```
akallabeth, line 1357: should have been `'@'` right?

akallabeth (changes requested):

akallabeth:
@hardening also 2 new warnings: https://ci.freerdp.com/job/code-quality-checker/10290/clang/fileName.-1652330529/ (the shadow variable might not be a bug, but it is easy to confuse if the outher and inner variables have the same names)

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -1325,13 +1325,45 @@ static SECURITY_STATUS SEC_ENTRY kerberos_AcceptSecurityContext(
...
+
+				realm++;
+			}
+			else
+			{
+				size_t len = strlen(sname);
```
akallabeth, line 1363: wrong length here, you invalidated this with the `memmove ` above

akallabeth (approved):
LGTM

PR #11446, opened 2025-03-31
Standard rdp security network issues

TolchiIsland:
2 threads tried sending network data simultaniously - 1 trying to send fastpath (e.g.  mouse events), and one sending in response to surface commands (on graphic updates).
Each thread does the following:

1. Check rdp->do_crypt and set rdp->sec_flags if needed
2. Atomically encrypt the data (guarded by lock)
3. Reset rdp->sec_flags
4. Atomically send the data (guarded by lock)

One race is caused by accessing rdp->sec_flags from multiple flags, causing them to overwrite each other. Fixed this by removing the global variable and using stack variables for better separation.
Another race is the encryption and send not synchronized, resulting in some cases packets sent not in the order they're encrypted, and the server cannot decrypt these packets. Fixed this by properly locking the whole encryption and sending altogether.

### DIFF
--- a/libfreerdp/core/activation.c
+++ b/libfreerdp/core/activation.c
@@ -75,7 +75,8 @@ BOOL rdp_recv_server_synchronize_pdu(rdpRdp* rdp, wStream* s)
 
 BOOL rdp_send_server_synchronize_pdu(rdpRdp* rdp)
 {
-	wStream* s = rdp_data_pdu_init(rdp);
+	UINT32 sec_flags = 0;
+	wStream* s = rdp_data_pdu_init(rdp, &sec_flags);
 	if (!s)
 		return FALSE;
 
@@ -87,7 +88,7 @@ BOOL rdp_send_server_synchronize_pdu(rdpRdp* rdp)
 	}
 
 	WINPR_ASSERT(rdp->mcs);
-	return rdp_send_data_pdu(rdp, s, DATA_PDU_TYPE_SYNCHRONIZE, rdp->mcs->userId);
+	return rdp_send_data_pdu(rdp, s, DATA_PDU_TYPE_SYNCHRONIZE, rdp->mcs->userId, sec_flags);
 }
 
 BOOL rdp_recv_client_synchronize_pdu(rdpRdp* rdp, wStream* s)
@@ -99,7 +100,8 @@ BOOL rdp_recv_client_synchronize_pdu(rdpRdp* rdp, wStream* s)
 
 BOOL rdp_send_client_synchronize_pdu(rdpRdp* rdp)
 {
-	wStream* s = rdp_data_pdu_init(rdp);
+	UINT32 sec_flags = 0;
+	wStream* s = rdp_data_pdu_init(rdp, &sec_flags);
 	if (!s)
 		return FALSE;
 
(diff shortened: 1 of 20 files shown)


akallabeth:
@TolchiIsland nice fix!
will review later this week, but on a first glance it looks good.

akallabeth:
@freerdp-bot test

libfreerdp/core/rdp.c
```diff
@@ -1333,10 +1365,12 @@ state_run_t rdp_recv_data_pdu(rdpRdp* rdp, wStream* s)
 	return STATE_RUN_FAILED;
 }
 
-state_run_t rdp_recv_message_channel_pdu(rdpRdp* rdp, wStream* s, UINT16 securityFlags)
+state_run_t rdp_recv_message_channel_pdu(rdpRdp* rdp, wStream* s, UINT32 securityFlags)
 {
 	WINPR_ASSERT(rdp);
 	WINPR_ASSERT(s);
+	WINPR_ASSERT(securityFlags &
```
akallabeth, line 1372: this triggers the code fuzzer, so the assertion is wrong or something is not handled correctly

TolchiIsland: Indeed my assertion was wrong. Should pass now

akallabeth:
@freerdp-bot test

libfreerdp/core/activation.c
```diff
@@ -75,7 +75,8 @@ BOOL rdp_recv_server_synchronize_pdu(rdpRdp* rdp, wStream* s)
 
 BOOL rdp_send_server_synchronize_pdu(rdpRdp* rdp)
 {
-	wStream* s = rdp_data_pdu_init(rdp);
+	UINT32 sec_flags = 0;
```
akallabeth, line 78: should everywhere be `UINT16 sec_flags`.

akallabeth:
@TolchiIsland nearly there.
you did convert the wrong way though, the security flags are `UINT16` not `UINT32`. (there is still a warning left due to this)

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

nice fix, thank you!

libfreerdp/core/activation.c
TolchiIsland: I relied on this https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/e13405c5-668b-4716-94b2-1c2654ca1ad4
Flags are up to `UINT32`, although so far only `UINT16` has been defined and the high part is reserved.
As I was working on this I decided it's better to future support the full `UINT32` (and add some asserts for now where we don't fully support it) instead of having a enlarging our tech debt along the way.
Also, the `rdp->sec_flags` I just deleted was `UINT32`, and some functions were as well (it was confused between `UINT16` and `UINT32`), so I wanted to continue its spirit.

akallabeth: @TolchiIsland understandable, but the assertions are still there (`SEC_FLAGSHI_VALID`), so no need to worry.
all fields are internal and can be refactored at any time (only thing that would give us dept is if we exposed something like this in a public `API`)

PR #11454, opened 2025-04-01
[core,tls] enable SNI when building with libreSSL

thenextman:
Building with libreSSL doesn't enable SNI.

`OPENSSL_NO_TLSEXT` will not be defined for libreSSL, `SSL_set_tlsext_host_name` is always available.


akallabeth (approved):
LGTM

PR #11455, opened 2025-04-02
[channels,client] log server format list

akallabeth:


taowap:
[log.txt](https://github.com/user-attachments/files/19566049/log.txt)

copy no english char form  window to mac, formatId = CF_TEXT, After copying, it is garbled.

PR #11484, opened 2025-04-09
[rdpei/server] fix build and channel init

mnauw:
In some build configurations, compilation of server-side RDPEI channel might fail (see commit for details).
Also, at runtime, the channel fails to initialize since `WTSVirtualChannelOpenEx(WTS_CURRENT_SESSION, ...)` (unconditionally) fails.  So it has been adjusted to what is used in other channels.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #11486, opened 2025-04-10
[client,sdl] fix clipboard updates

akallabeth:
* Ignore updates for our own data (use a special clipboard format to identify these)

we now use a trick to identify our own updates:
1. We insert the special mime type `x-special/freerdp-clipboard-update`
2. We update the client clipboard with `SDL_SetClipboardData` once (we use `_current_mimetypes` for that)
3. we clear `_current_mimetypes` after `SDL_SetClipboardData`, so further clipboard events will not retrigger an update
4. as long as the special mime type is available in the clipboard do not send a format list update to the server
5. if the clipboard is empty or changed from another application (our special mime type was erased) we send a format list update to the server


akallabeth:
@hardening read https://github.com/libsdl-org/SDL/issues/12557
basically it is set for *some* of the events, but not all.
to date there is no other reliable way to check for this

PR #11508, opened 2025-04-16
[common,settings] new settings (de)serialization API

akallabeth:
* Add rdpKey to PEM conversion
* Modify WINPR_JSON_Version to return -1 if not supported
* Add rdpSettings serialize and deserialize functions
* make pointer types more robust, also update the buffer sizes if the size is changed directly with the `freerdp_settings_set_uint32` functions


akallabeth:
@freerdp-bot test

PR #11510, opened 2025-04-19
rdpei/server: Fix incorrect PDU length read

pnowack:
The PDU length is a 32-bit unsigned integer and not a 16-bit one.


akallabeth (approved):
LGTM

PR #11520, opened 2025-04-22
[winpr,sysinfo] use a single clock to provide System and Local time

mnauw:
... rather than a different one for second and sub-second part

The sub-second is currently taken from "tick count", which (apparently) tracks (process) up-time rather than any real system time.  This then leads to "funny" timestamps in logs, which may go backwards and do not correspond to timestamps as produced by other logging.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

hardening (approved):
Seems ok for me

PR #11530, opened 2025-04-24
[rdpsnd/client] add parameters to pulse snd device plugin

akallabeth:
... to specify pulseaudio client name and stream name


akallabeth:
@mnauw thank you for this patch!

PR #11540, opened 2025-04-28
[winpr,file] Fix assert fail always when removing flags

lazy5f:
It alway fails at least on macOS clang. This fix is based on the code below which does the same operation:

https://github.com/FreeRDP/FreeRDP/blob/9bd2330c4e10d7ac2d09e671705478af38b06823/winpr/libwinpr/file/generic.c#L635

Not sure `WINPR_ASSERTING_INT_CAST` is quite necessary in this case, though.


akallabeth (approved):
LGTM

Issue #11541, opened 2025-04-29
FFMPEG API Changes: FF_PROFILE depreciation


EndlessEden:
FFmpeg Depreciated FF_PROFILE_* back in Sept.2023, Replacing it with AV_PROFILE_* ( See Commit: [8238bc0
](https://github.com/FFmpeg/FFmpeg/commit/8238bc0b5e3dba271217b1223a901b3f9713dc6e) )

As a result, FreeRDP wont build against any release after that point.

EndlessEden:
Submitted Pull Request #11542 with a fix, please review and advise. Thank you.

PR #11542, opened 2025-04-29
FF_PROFILE Depreciation

EndlessEden:
Fixes depreciated FF_PROFILE_AAC definition, replacing with AV_PROFILE_AAC, see issue #11541


akallabeth (changes requested):
Please add a fallback path for older versions (still found in many distributions)

akallabeth:
@freerdp-bot test

PR #11547, opened 2025-05-01
Support 'Restrict Credential Delegation' mode

gpotter2:
This PR fixes two issues related to RestrictedAdmin on the server side:

1. There is an inconsistency in how RestrictedMode is handled on the server side. As the following code points out, while RemoteGuard needs to be enabled explicitely, the server always considers that it supports RestrictedAdmin. (hence also why there is no RestrictedAdminSupported boolean).

https://github.com/FreeRDP/FreeRDP/blob/32956261e42ffb893c204d84c2869deeaacb7e4e/libfreerdp/core/nego.c#L1256-L1273

However, when sending the negotiation response, the server has this weird check:
https://github.com/FreeRDP/FreeRDP/blob/32956261e42ffb893c204d84c2869deeaacb7e4e/libfreerdp/core/nego.c#L1474

If my understanding is correct, and we consider that the server always supports RestrictedAdmin, it should be removed. If we don't consider that we always support RestrictedAdmin, we should fail the negotiation in the first snippet. I PR'd the first option.

2. FreeRDP's server fails to negotiate if you enable "Restrict Credential Delegation" but do not enable RemoteGuard on Freerdp.

image

Enabling this feature makes the negotiation flags be 3 (RESTRICTED_ADMIN_MODE_REQUIRED + REDIRECTED_AUTHENTICATION_MODE_REQUIRED), but it actually means "OR" as shown above. This looks somewhat off-spec, but we shouldn't fail as we support RestrictedAdmin but not RemoteGuard.

### DIFF
--- a/libfreerdp/core/nego.c
+++ b/libfreerdp/core/nego.c
@@ -59,7 +59,8 @@ struct rdp_nego
 	UINT32 RequestedProtocols;
 	BOOL NegotiateSecurityLayer;
 	BOOL EnabledProtocols[32];
-	BOOL RestrictedAdminModeRequired;
+	BOOL RestrictedAdminModeRequired;  /* Client-side */
+	BOOL RestrictedAdminModeSupported; /* Server-side */
 	BOOL RemoteCredsGuardRequired;
 	BOOL RemoteCredsGuardActive;
 	BOOL RemoteCredsGuardSupported;
@@ -1254,7 +1255,18 @@ BOOL nego_process_negotiation_request(rdpNego* nego, wStream* s)
 		return FALSE;
 	}
 	if (flags & RESTRICTED_ADMIN_MODE_REQUIRED)
-		WLog_Print(nego->log, WLOG_INFO, "RDP_NEG_REQ::flags RESTRICTED_ADMIN_MODE_REQUIRED");
+	{
+		if (nego->RestrictedAdminModeSupported)
+		{
+			WLog_Print(nego->log, WLOG_INFO, "RDP_NEG_REQ::flags RESTRICTED_ADMIN_MODE_REQUIRED");
+		}
+		else
+		{
+			WLog_Print(nego->log, WLOG_ERROR,
+			           "RDP_NEG_REQ::flags RESTRICTED_ADMIN_MODE_REQUIRED but disabled");
+			return FALSE;
+		}
+	}
 
 	if (flags & REDIRECTED_AUTHENTICATION_MODE_REQUIRED)
 	{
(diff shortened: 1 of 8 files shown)


libfreerdp/core/nego.c
```diff
@@ -1471,8 +1481,8 @@ BOOL nego_send_negotiation_response(rdpNego* nego)
 		if (freerdp_settings_get_bool(settings, FreeRDP_SupportGraphicsPipeline))
 			flags |= DYNVC_GFX_PROTOCOL_SUPPORTED;
 
-		if (freerdp_settings_get_bool(settings, FreeRDP_RestrictedAdminModeRequired))
-			flags |= RESTRICTED_ADMIN_MODE_SUPPORTED;
+		/* Server always supports RestrictedAdmin */
```
akallabeth, line 1484: not sure this is a good idea.
we use the `rdpSettings` to have a way to enable/disable things and set the default values to what is expected.
(allows easier testing of *impossible* situation handling)

akallabeth (commented):
Overall LGTM,
the removal of the server side settings check might be better left out if that is possible.

libfreerdp/core/nego.c
gpotter2: That's okay, but in this case we need to add a new `FreeRDP_RestrictedAdminSupported` flag and some extra logic

gpotter2: I've changed the PR to add a new settings to handle this behavior.

libfreerdp/core/settings.c
```diff
@@ -884,6 +884,7 @@ rdpSettings* freerdp_settings_new(DWORD flags)
 	    !freerdp_settings_set_bool(settings, FreeRDP_RdstlsSecurity, FALSE) ||
 	    !freerdp_settings_set_bool(settings, FreeRDP_NegotiateSecurityLayer, TRUE) ||
 	    !freerdp_settings_set_bool(settings, FreeRDP_RestrictedAdminModeRequired, FALSE) ||
+	    !freerdp_settings_set_bool(settings, FreeRDP_RestrictedAdminModeSupported, TRUE) ||
```
gpotter2, line 887: This is set to TRUE by default to mimic the previous behavior.

akallabeth:
@freerdp-bot test

akallabeth:
@gpotter2 you should also run the `./tools/update-settings-tests ` script to generate all required settings setter/getter. (generated files are also not properly formatted, so don't forget that)

gpotter2:
@akallabeth Done.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

akallabeth:
@gpotter2 thx for the nice improvement!
please rebase your other pr to current `master` , then I'll have a look at it.

PR #11548, opened 2025-05-01
[cmake] Fix finding ffmpeg under nonstandard prefixes

thestr4ng3r:
The prefix given to pkg_check_modules is PC\_${\_component}, not PC\_LIB${\_component}, so those variables were always read as empty, thus no HINTS were passed to the find\_* functions and ffmpeg could not be found if it was not in a default path.

For reference:
https://invent.kde.org/multimedia/ffmpegthumbs/-/commit/afcc7af91967fb66145bef0c74f811b3c58da80c
https://bugs.kde.org/show_bug.cgi?id=338280


akallabeth (approved):
LGTM

PR #11549, opened 2025-05-01
Support NLA in shadow server when running behind a Hyper-V proxy

gpotter2:
**This PR requires https://github.com/FreeRDP/FreeRDP/pull/11547**

This PR adds support for other security protocols than plain RDP when using FreeRDP shadow server inside a Hyper-V VM. In this specific case, security is handled by Hyper-V, so the FreeRDP server needs to be able to advertise for NLA support (like Windows does by default), without actually handling it itself.

Untitled Diagram
(source: me)

This just adds a `/vmconnect` flag that turns this behavior on (same name than on the client side). This was tested on Win11 24H2 using vmconnect.exe

Usage: `sudo freerdp-shadow-cli /bind-address:"vsock://1" /vmconnect`

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -1525,7 +1525,11 @@ BOOL rdp_server_accept_nego(rdpRdp* rdp, wStream* s)
 	SelectedProtocol = nego_get_selected_protocol(nego);
 	status = FALSE;
 
-	if (SelectedProtocol & PROTOCOL_RDSTLS)
+	if (freerdp_settings_get_bool(rdp->settings, FreeRDP_VmConnectMode) &&
+	    SelectedProtocol != PROTOCOL_RDP)
+		/* When behind a Hyper-V proxy, security != RDP is handled by the host. */
+		status = TRUE;
+	else if (SelectedProtocol & PROTOCOL_RDSTLS)
 		status = transport_accept_rdstls(rdp->transport);
 	else if (SelectedProtocol & PROTOCOL_HYBRID)
 		status = transport_accept_nla(rdp->transport);
--- a/server/shadow/cli/shadow.c
+++ b/server/shadow/cli/shadow.c
@@ -64,6 +64,8 @@ int main(int argc, char** argv)
 		  "Remote credential guard" },
 		{ "restricted-admin", COMMAND_LINE_VALUE_BOOL, NULL, BoolValueTrue, NULL, -1, NULL,
 		  "Restricted Admin" },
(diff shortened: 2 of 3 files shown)


akallabeth:
I assume this required #11547 ?

gpotter2:
Yes it does. I'll rebase when merged to give some clarity ^^

server/shadow/cli/shadow.c
```diff
@@ -62,6 +62,10 @@ int main(int argc, char** argv)
 		  "Clients must authenticate" },
 		{ "remote-guard", COMMAND_LINE_VALUE_BOOL, NULL, BoolValueFalse, NULL, -1, NULL,
 		  "Remote credential guard" },
+		{ "restricted-admin", COMMAND_LINE_VALUE_BOOL, NULL, BoolValueTrue, NULL, -1, NULL,
```
akallabeth, line 65: might be a good idea to split this commit as `vmconnect` and `restricted-admin` are separate things?

libfreerdp/core/connection.c
```diff
@@ -1525,7 +1525,11 @@ BOOL rdp_server_accept_nego(rdpRdp* rdp, wStream* s)
 	SelectedProtocol = nego_get_selected_protocol(nego);
 	status = FALSE;
 
-	if (SelectedProtocol & PROTOCOL_RDSTLS)
+	if (freerdp_settings_get_bool(rdp->settings, FreeRDP_VmConnectMode) &&
+		SelectedProtocol != PROTOCOL_RDP)
```
akallabeth, line 1529: so even `TLS` security? that is more or less the same as `RDP` except it is using standard `anonymous TLS` instead of `MS` grown custom `RSA` stuff

akallabeth (changes requested):

libfreerdp/core/connection.c
gpotter2: Yes, I just doubled check. For anything but RDP, vmms.exe intercepts it.
Since vmms.exe is basically doing TLS interception, it somewhat makes sense for it not to re-create a TLS connection just to talk to a local VM.

akallabeth (approved):
LGTM

server/shadow/cli/shadow.c
akallabeth: no, fine this way (just easier to track later on with separate commits.

akallabeth:
@freerdp-bot test

gpotter2:
Sorry, just git clang-formatted.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #11558, opened 2025-05-06
[core,test] refactor TestSettings

akallabeth:
* log each test case run
* flush log messages
* split large test case functions


akallabeth:
@freerdp-bot test

PR #11576, opened 2025-05-13
[rdpsnd/server] skip sending of formats if no formats specified yet

mnauw:
... which prevents sending a problematic PDU, while allowing for more flexibility when using the server API

### DIFF
--- a/channels/rdpsnd/server/rdpsnd_main.c
+++ b/channels/rdpsnd/server/rdpsnd_main.c
@@ -59,6 +59,9 @@ static UINT rdpsnd_server_send_formats(RdpsndServerContext* context)
 	BOOL status = FALSE;
 	ULONG written = 0;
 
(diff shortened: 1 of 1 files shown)


channels/rdpsnd/server/rdpsnd_main.c
```diff
@@ -59,6 +59,9 @@ static UINT rdpsnd_server_send_formats(RdpsndServerContext* context)
 	BOOL status = FALSE;
 	ULONG written = 0;
 
+	if (context->num_server_formats == 0)
```
akallabeth, line 62: looks like a bug.
1. `PCM` should always be supported 
2. the channel is *required* for `RDP` protocol to properly work (don't ask why, did not design it)
3. if not the whole channel should not be started instead of aborting setup sequence here

PR #11578, opened 2025-05-14
Timer implementation

akallabeth:
Adds a timer implementation bound to a `rdpContext`

This allows adding arbitrary timers for tasks that require it for a RDP session.
The timer API allows nanosecond, but current implementation limits it to milliseconds (to be fixed)

### DIFF
--- a/libfreerdp/core/timer.c
+++ b/libfreerdp/core/timer.c
@@ -1,0 +1,321 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ * Timer implementation
+ *
+ * Copyright 2025 Armin Novak <anovak@thincast.com>
+ * Copyright 2025 Thincast Technologies GmbH
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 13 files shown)


akallabeth:
@hardening done, now you can choose if you want to run from background thread or schedule it to be processed on main thread.

libfreerdp/core/timer.c
```diff
@@ -0,0 +1,321 @@
...
+		const uint64_t next = expire_and_reschedule(timer);
+		const uint64_t now = winpr_GetTickCount64NS();
+		if (now >= next)
+		{
+			timeout = INFINITE;
+			continue;
+		}
+
+		const uint64_t diff = next - now;
+		const uint64_t diffMS = diff / 1000;
```
volkov0aa, line 201: Shouldn't it be `diff / 1000000`?
1ns = 1s / 10^9
1ms = 1s / 10^3

Then RESIZE_MIN_DELAY_NS should also be changed from 200000UL to 200000000UL

PR #11588, opened 2025-05-15
winpr: move ndr routines from rdpear to winpr

hardening:


akallabeth (changes requested):
When moving to public API we should prefix the functions with `winpr` or similar to make them unique.

winpr/include/winpr/ndr.h
```diff
@@ -0,0 +1,222 @@
...
+	WINPR_API BOOL ndr_read_##LOWER(NdrContext* context, wStream* s, UPPER* v);             \
+	WINPR_API BOOL ndr_read_##LOWER##_(NdrContext* context, wStream* s, const void* hints,  \
+	                                   void* v);                                            \
+	WINPR_API BOOL ndr_write_##LOWER(NdrContext* context, wStream* s, UPPER v);             \
+	WINPR_API BOOL ndr_write_##LOWER##_(NdrContext* context, wStream* s, const void* hints, \
+	                                    const void* v);                                     \
+	extern WINPR_API const NdrMessageDescr ndr_##LOWER##_descr_s;                           \
+	WINPR_API NdrMessageType ndr_##LOWER##_descr(void)
+
+#define NDR_ARRAY_OF_TYPE_DECL(TYPE, UPPERTYPE)                                                  \
```
akallabeth, line 37: maybe also remove the macros from public headers.

hardening: The macros are used in this header. And usually when you want to declare a new type you'll want to use them.

hardening:
I've made the renamings but I've seen things I'd like to fix for a public API.

PR #11590, opened 2025-05-15
[winpr,sspi] add kerberos string len checks

akallabeth:


hardening (approved):
Probably doesn't hurt

PR #11600, opened 2025-05-19
feat: add login through MS identity broker via sso-mib interface

rupran:
This change enables an alternative way of acquiring the necessary access tokens through a local identity broker. In the current implementation, we need to visit URLs twice and paste back the URLs we are redirected to in order to extract authorization codes and ultimately fetch the correct access tokens for RDP (described here: <0>).

As an alternative, MS also provides the Microsoft Authentication Library (MSAL) through which authentication can be handled more or less in the background when we're using a trusted device. In particular, we can request access tokens with the same parameters as we're currently doing through the URL-based scheme.

As the MSAL bindings are not available for C, we implemented a small wrapper library called sso-mib which is available at https://github.com/siemens/sso-mib. This library translates the high-level requests (such as acquire_token_interactive) to respective messages on the D-Bus messaging bus which is used to communicate with the identity broker service on Linux. The library can be built as a .deb package and subsequently be found through PkgConfig mechanisms in CMake.

When sso-mib is not available through pkg-config, it can also be placed in `external/`, with the directory structure looking like the following. `include/` is copied from the root of the sso-mib directory and `lib/` populated with the built shared library files and symlinks.

    external/
    ├── README
    └── sso-mib
        ├── include
        │   └── sso-mib
        │       ├── mib-account.h
        │       ├── mib-client-app.h
        │       ├── mib-exports.h
        │       ├── mib-pop-params.h
        │       ├── mib-prt.h
        │       ├── mib-prt-sso-cookie.h
        │       └── sso-mib.h
        └── lib
            ├── libsso-mib.so -> libsso-mib.so.0
            ├── libsso-mib.so.0 -> libsso-mib.so.0.4.0
            └── libsso-mib.so.0.4.0

This feature is currently hidden behind a configuration switch and must be enabled via `-DWITH_SSO_MIB=ON`. If the connection to the broker fails (for example, if no identity broker is installed or running on the system), we automatically fall back to the current scheme of copy-pasting URLs.

<0>: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/e967ebeb-9e9f-443e-857a-5208802943c2

If the communication with the identity broker is successful, you should not need to copy and paste URLs anymore, but instead the identity broker should open a window on the first connection attempt, asking for your password, and then ask for confirmation to allow the remote connection:

Screenshot 2025-05-19 090115
Screenshot 2025-05-19 090135

Once the connection to the same machine has been accepted, subsequent logins should require no interaction at all as the identity broker caches the token and provides it on request.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -410,7 +410,6 @@ else()
 endif()
 
 if(WITH_SSO_MIB)
-  set(SSO_MIB_EXTERNAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external/sso-mib")
   find_package(SSO_MIB REQUIRED)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@
 # Copyright 2011 Otavio Salvador <otavio@ossystems.com.br>
 # Copyright 2011 Marc-Andre Moreau <marcandre.moreau@gmail.com>
 # Copyright 2012 HP Development Company, LLC
+# Copyright 2025 Siemens
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -400,6 +401,17 @@ endif()
 
 if(NOT WITHOUT_FREERDP_3x_DEPRECATED)
   find_feature(Wayland ${WAYLAND_FEATURE_TYPE} ${WAYLAND_FEATURE_PURPOSE} ${WAYLAND_FEATURE_DESCRIPTION})
+endif()
+
+if(UNIX)
+  option(WITH_SSO_MIB "Build with sso-mib support" OFF)
+else()
+  set(WITH_SSO_MIB OFF CACHE INTERNAL "unsupported platform")
+endif()
+
+if(WITH_SSO_MIB)
+  set(SSO_MIB_EXTERNAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/external/sso-mib")
+  find_package(SSO_MIB REQUIRED)
 endif()
 
 option(WITH_LIBRESSL "build with LibreSSL" OFF)
--- a/client/common/CMakeLists.txt
+++ b/client/common/CMakeLists.txt
@@ -55,15 +55,17 @@ endif()
 
 include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
 
+addtargetwithresourcefile(${MODULE_NAME} FALSE "${FREERDP_VERSION}" SRCS)
+
 if(WITH_SSO_MIB)
-  set(SSO_MIB_EXTERNAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../external/sso-mib")
+  if(SSO_MIB_EXTERNAL_DIR)
+    add_dependencies(${MODULE_NAME} sso-mib-external)
+  endif()
 
   include_directories(${SSO_MIB_INCLUDE_DIRS})
   add_compile_definitions(WITH_SSO_MIB)
   list(APPEND LIBS ${SSO_MIB_LIBRARIES})
 endif()
-
-addtargetwithresourcefile(${MODULE_NAME} FALSE "${FREERDP_VERSION}" SRCS)
 
 list(APPEND LIBS freerdp winpr)
 
--- a/client/common/CMakeLists.txt
+++ b/client/common/CMakeLists.txt
@@ -2,6 +2,7 @@
 # FreeRDP Client Common
 #
 # Copyright 2012 Marc-Andre Moreau <marcandre.moreau@gmail.com>
+# Copyright 2025 Siemens
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -53,6 +54,16 @@ if(WITH_FUSE)
 endif()
 
 include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
+
+option(WITH_SSO_MIB "Build with sso-mib support" OFF)
+if(UNIX AND WITH_SSO_MIB)
+  set(SSO_MIB_EXTERNAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../external/sso-mib")
+  find_package(SSO_MIB REQUIRED)
+
+  include_directories(${SSO_MIB_INCLUDE_DIRS})
+  add_compile_definitions(WITH_SSO_MIB)
+  list(APPEND LIBS ${SSO_MIB_LIBRARIES})
+endif()
 
 addtargetwithresourcefile(${MODULE_NAME} FALSE "${FREERDP_VERSION}" SRCS)
 
--- a/client/common/CMakeLists.txt
+++ b/client/common/CMakeLists.txt
@@ -55,10 +55,8 @@ endif()
 
 include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
 
-option(WITH_SSO_MIB "Build with sso-mib support" OFF)
-if(UNIX AND WITH_SSO_MIB)
+if(WITH_SSO_MIB)
   set(SSO_MIB_EXTERNAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../external/sso-mib")
-  find_package(SSO_MIB REQUIRED)
 
   include_directories(${SSO_MIB_INCLUDE_DIRS})
   add_compile_definitions(WITH_SSO_MIB)
--- a/client/common/CMakeLists.txt
+++ b/client/common/CMakeLists.txt
@@ -55,6 +55,10 @@ endif()
 
 include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
 
+if(WITH_SSO_MIB)
+  list(APPEND SRCS sso_mib_tokens.c)
+endif()
+
 addtargetwithresourcefile(${MODULE_NAME} FALSE "${FREERDP_VERSION}" SRCS)
 
 if(WITH_SSO_MIB)
--- a/client/common/client.c
+++ b/client/common/client.c
@@ -70,23 +70,7 @@
 #endif
 
 #ifdef WITH_SSO_MIB
-#include <sso-mib/sso-mib.h>
-#include <freerdp/crypto/crypto.h>
-#include <winpr/json.h>
-
-enum sso_mib_state
-{
-	SSO_MIB_STATE_INIT = 0,
-	SSO_MIB_STATE_FAILED = 1,
-	SSO_MIB_STATE_SUCCESS = 2,
-};
-
-struct MIBClientWrapper
-{
-	MIBClientApp* app;
-	enum sso_mib_state state;
-};
-
+#include "sso_mib_tokens.h"
 #endif
 
 #include <freerdp/log.h>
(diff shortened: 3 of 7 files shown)


rupran:
Documentation of the sso-mib library is now also available via GitHub Pages: https://siemens.github.io/sso-mib/

akallabeth (commented):
Thank you,
looks like a nice addition.

since I did recently update the webview support with https://github.com/FreeRDP/FreeRDP/pull/11518
this still uses the default implementation as fallback, right?

fmoessbauer:
Maintainer of the sso-mib library here.

> since I did recently update the webview support with #11518

The webview is used to perform the entra login in a webview window instead of the browser, right? Once we have the `sso-mib` integrated, we can use the `mib_client_app_acquire_prt_sso_cookie` to inject the `PrtSsoCookie` into the request. By that, the user can be SSOed in the first login (gateway) and also can pass the conditional access policies.

This is just half of the login flow, but also improves the user experience. In this MR, the whole login flow (including the generation of the PoP tokens for RDP) is done via the broker.

client/common/client.c
```diff
@@ -68,6 +69,23 @@
 #include <freerdp/utils/aad.h>
 #endif
 
+#ifdef WITH_SSO_MIB
+#include <sso-mib/sso-mib.h>
+#include <freerdp/crypto/crypto.h>
+#include <winpr/json.h>
+
+static MIBClientApp* sso_mib_client_app = NULL;
```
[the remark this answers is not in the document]
rupran: No particular reason, I'll move it in the next iteration.

akallabeth:
@fmoessbauer yes, the webview is for doing the login window (currently only for sdl-freerdp) whereas you did add the code to client-common (currently only cli with copy & paste to browser)

I'm asking as it might be better to split this to stand on its own so the best backend for the login flow can be used depending on client. (the cli version is overridden by sdl-freerdp currently by default to have a webview window pop up)

client/common/CMakeLists.txt
```diff
@@ -54,6 +55,16 @@ endif()
 
 include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
 
+option(WITH_SSO_MIB "Build with sso-mib support" OFF)
+if(UNIX AND WITH_SSO_MIB)
+  set(SSO_MIB_EXTERNAL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../external/sso-mib")
```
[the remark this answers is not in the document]
akallabeth: @rupran hmm, looks more like a https://cmake.org/cmake/help/latest/module/ExternalProject.html usecase (FetchContent is only for `CMake` based ones or just data to be downloaded)
sorry for the wrong link.

akallabeth:
@rupran @fmoessbauer is there a short howto for testing this somewhere?
might be a good idea to have that at hand to remember how to set up a test env quickly later on if there should be some issues.

client/common/CMakeLists.txt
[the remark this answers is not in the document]
rupran: This was a bit of a pain as ExternalProject_Add needs explicit dependencies at every place where the external project is used, and in combination with moving the pointer to the MIBClientApp into the global freerdp header, this required touching quite a few CMakeLists.txt files across the tree... I hope I got them all but this might require a deeper look.

Otherwise, if you know a better way of managing the dependency, I'm all ears.

rupran:
> @fmoessbauer yes, the webview is for doing the login window (currently only for sdl-freerdp) whereas you did add the code to client-common (currently only cli with copy & paste to browser)
> 
> I'm asking as it might be better to split this to stand on its own so the best backend for the login flow can be used depending on client. (the cli version is overridden by sdl-freerdp currently by default to have a webview window pop up)

I moved the implementation into a separate compilation unit as a first step, but I have not thought about how to integrate this properly in the future... as the concrete implementation is hidden through the `instance->GetAccessToken` pointer, there is not really a possibility for fallbacks at the moment other than implementing them in the actual flows.

channels/CMakeLists.txt
```diff
@@ -176,6 +176,13 @@ macro(add_channel_client_library _module_prefix _module_name _channel_name _dyna
   add_library(${_module_name} OBJECT ${${_module_prefix}_SRCS})
   set_property(TARGET ${_module_name} PROPERTY FOLDER "Channels/${CHANNEL_NAME}/Client")
 
+  if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 179: this should only be required at one place, `client/common` as the lib is not used anywhere else

channels/CMakeLists.txt
```diff
@@ -203,6 +210,13 @@ macro(
   add_library(${_module_name} OBJECT ${${_module_prefix}_SRCS})
   set_property(TARGET ${_module_name} PROPERTY FOLDER "Channels/${_channel_name}/Client/Subsystem/${CHANNEL_SUBSYSTEM}")
 
+  if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 213: same as above

channels/CMakeLists.txt
```diff
@@ -223,6 +237,12 @@ macro(add_channel_server_library _module_prefix _module_name _channel_name _dyna
   add_library(${_module_name} OBJECT ${${_module_prefix}_SRCS})
   set_property(TARGET ${_module_name} PROPERTY FOLDER "Channels/${CHANNEL_NAME}/Server")
 
+  if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 240: same as above

client/SDL/SDL2/CMakeLists.txt
```diff
@@ -66,6 +66,13 @@ endif()
 
 addtargetwithresourcefile(${MODULE_NAME} "${WIN32_GUI_FLAG}" "${PROJECT_VERSION}" SRCS)
 
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 69: same as above

client/SDL/SDL2/dialogs/CMakeLists.txt
```diff
@@ -65,6 +65,13 @@ add_subdirectory(res)
 
 add_library(sdl2-dialogs STATIC ${SRCS})
 
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 68: same as above

client/SDL/SDL3/CMakeLists.txt
```diff
@@ -69,6 +69,13 @@ set_target_properties(SDL3::Headers PROPERTIES SYSTEM TRUE)
 
 addtargetwithresourcefile(${MODULE_NAME} "${WIN32_GUI_FLAG}" "${PROJECT_VERSION}" SRCS)
 
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 72: same as above

client/SDL/SDL3/dialogs/CMakeLists.txt
```diff
@@ -78,6 +78,13 @@ add_subdirectory(res)
 
 add_library(sdl3-dialogs STATIC ${SRCS} sdl_connection_dialog_hider.hpp sdl_connection_dialog_hider.cpp)
 
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 81: same as above

client/Wayland/CMakeLists.txt
```diff
@@ -51,6 +51,14 @@ list(APPEND ${MODULE_PREFIX}_LIBS freerdp-client freerdp uwac)
 
 addtargetwithresourcefile(${MODULE_NAME} TRUE ${FREERDP_VERSION} ${MODULE_PREFIX}_SRCS)
 
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 54: same as above

libfreerdp/CMakeLists.txt
```diff
@@ -206,6 +206,13 @@ endif()
 
 addtargetwithresourcefile(${MODULE_NAME} FALSE "${FREERDP_VERSION}" LIBFREERDP_SRCS)
 
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 209: same as above

libfreerdp/codec/CMakeLists.txt
```diff
@@ -125,6 +125,11 @@ if(ANDROID AND WITH_MEDIACODEC)
 endif()
 
 add_library(freerdp-codecs OBJECT ${CODEC_SRCS})
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 128: same as above

server/proxy/CMakeLists.txt
```diff
@@ -53,6 +53,13 @@ set(PRIVATE_LIBS freerdp-client freerdp-server)
 
 set(PUBLIC_LIBS winpr freerdp)
 
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 56: same as above

server/proxy/channels/CMakeLists.txt
```diff
@@ -6,3 +6,10 @@ if(WITH_PROXY_EMULATE_SMARTCARD)
 endif()
 
 add_library(${MODULE_NAME} OBJECT ${SOURCES})
+
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 10: same as above

client/common/sso_mib_tokens.h
```diff
@@ -0,0 +1,16 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Copyright 2025 Siemens
+ */
+
+#ifndef FREERDP_CLIENT_COMMON_SSO_MIB_TOKENS_H
+#define FREERDP_CLIENT_COMMON_SSO_MIB_TOKENS_H
+
+#include <freerdp/freerdp.h>
+
+BOOL sso_mib_get_avd_access_token(freerdp* instance, char** token);
```
akallabeth, line 11: this can be integrated in 2 ways to be available by default:
1. Move the whole thing to `libfreerdp/core` and check if a token can be retrieved before calling `instance->GetAccessToken(instance, ACCESS_TOKEN_TYPE_XXX`
2. Chain the `GetAccessToken` calls and expose a new callback via `rdpClientContext` and do not touch the existing callback in client implementations.

CMakeLists.txt
```diff
@@ -402,6 +402,14 @@ if(NOT WITHOUT_FREERDP_3x_DEPRECATED)
   find_feature(Wayland ${WAYLAND_FEATURE_TYPE} ${WAYLAND_FEATURE_PURPOSE} ${WAYLAND_FEATURE_DESCRIPTION})
 endif()
 
+option(WITH_SSO_MIB "Build with sso-mib support" OFF)
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 406: would remove the `UNIX` check all over the place and just do a 
```
if (UNIX)
option(WITH_SSO_MIB "Build with sso-mib support" OFF)
else()
set(WITH_SSO_MIB OFF CACHE INTERNAL "unsupported platform")
endif()

File: include/freerdp/freerdp.h [hunk 17]
```diff
@@ -339,8 +343,13 @@ extern "C"
 
 		ALIGN64 rdpStreamDumpContext* dump; /* 64 */
 		ALIGN64 wLog* log;                  /* 65 */
+#if defined(WITH_SSO_MIB)
+		ALIGN64 MIBClientApp* mibClientApp; /* 66 */
```
akallabeth, line 347: might be better off in `rdpClientContext`

channels/CMakeLists.txt
akallabeth: you can just do a forward declaration of the type and hide details

include/freerdp/freerdp.h
rupran: Is the `rdpContext` member of `freerdp` always a `rdpClientContext` inside `client/common/*`?

channels/CMakeLists.txt
rupran: Cleaned up by moving into client.h but still unsure about the right handling of the type. Right now, I do the following, similar to two other flags that are already there:
```c
#if defined(WITH_SSO_MIB)
		ALIGN64 MIBClientApp* mibClientApp; /**< (offset 10) */
#else
		UINT64 reserved4;
#endif
```

include/freerdp/freerdp.h
rupran: Moved to `rdpClientContext` but the previous question still remains (I mean it works, but that's not a perfect indicator 😉 )

client/common/client.c
```diff
@@ -68,6 +69,19 @@
...
+#include "sso_mib_tokens.h"
+
+enum sso_mib_state
+{
+	SSO_MIB_STATE_INIT = 0,
+	SSO_MIB_STATE_FAILED = 1,
+	SSO_MIB_STATE_SUCCESS = 2,
+};
+
+static enum sso_mib_state sso_mib_state = SSO_MIB_STATE_INIT;
```
akallabeth, line 82: should this not be part of `instance/context` too?

include/freerdp/client.h
```diff
@@ -41,6 +41,10 @@
 #include <freerdp/client/encomsp.h>
 #endif
 
+#if defined(WITH_SSO_MIB)
+#include <sso-mib/sso-mib.h>
```
akallabeth, line 45: just remove this include and only include it in `client.c`
wrap the thing in a `typedef struct foo MIBClientAppWrapper ;` (or whatever you deem a good name) and define the struct itself only in `client.c`

this way you do not expose any types from the dependent lib and you only need the include paths/link options for the single file.

client/common/CMakeLists.txt
```diff
@@ -54,8 +55,22 @@ endif()
 
 include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
 
+if(UNIX AND WITH_SSO_MIB)
```
akallabeth, line 58: the `UNIX` should be removed here too

include/freerdp/freerdp.h
```diff
@@ -360,7 +360,14 @@ extern "C"
 		Disconnect_Ultimatum_channel_purged = 4
 	};
 
+#ifdef __cplusplus
```
akallabeth, line 363: useless change?

include/freerdp/freerdp.h
```diff
@@ -360,7 +360,14 @@ extern "C"
...
+}
+#endif
 #include <freerdp/client.h>
+#ifdef __cplusplus
```
akallabeth, line 367: useless change?

client/common/sso_mib_tokens.c
```diff
@@ -0,0 +1,119 @@
...
+	{
+		*token = strdup(mib_prt_get_access_token(prt));
+		rc = TRUE;
+		g_object_unref(prt);
+	}
+
+cleanup:
+	if (params)
+		g_object_unref(params);
+	if (json)
```
akallabeth, line 113: `free` and our wrappers do acceppt `NULL` as a valid argument, no need to check this here.

include/freerdp/freerdp.h
akallabeth: see my comment below regarding wrapping the type in a private struct.

client/common/client.c
rupran: Yes, I can move it over (possibly into the Wrapper struct)

include/freerdp/freerdp.h
rupran: The include of some glib headers errored out here complaining about templates under C linkage deep inside the include chain... But that might go away with the Wrapper struct as well.

include/freerdp/freerdp.h
rupran: See above

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@rupran nice, now the whole thing looks a lot cleaner.

rupran:
> @rupran nice, now the whole thing looks a lot cleaner.

Yep, I think so too, thank you for the detailed comments. I'll see what I can do about the general fallback mechanism mentioned above (https://github.com/FreeRDP/FreeRDP/pull/11600#discussion_r2097940471). Would you prefer to have this also as part of this PR or as a later improvement on top?

akallabeth:
ok, you need to reformat the code:
* `./scripts/cmake-format.sh -f` for cmake files
* `git clang-format -f origin/master` for source code

one thing (can be done later) is to add a `GetAccessToken` function pointer to `rdpClientContext` so that the common implementation can fall back to a client specific one if the `sso-mib` does not produce a token.

akallabeth:
@rupran ok, did add the delegated callback with https://github.com/FreeRDP/FreeRDP/pull/11606

akallabeth:
@rupran with that pr, you can use `freerdp_get_common_access_token` to store the default callback implementation to be called when the `sso-mib` fails and with `freerdp_set_common_access_token` you can set your own implementation that will first try the `sso-mib` and then call the previously stored callback.

akallabeth:
@freerdp-bot test

client/common/client.c
```diff
@@ -164,6 +169,13 @@ void freerdp_client_context_free(rdpContext* context)
 	if (!context)
 		return;
 
+#ifdef WITH_SSO_MIB
```
akallabeth, line 172: did miss this, if you allocate in `freerdp_client_start` then you need to deallocate in `freerdp_client_stop` and not in `freerdp_client_free`

akallabeth (changes requested):

client/common/client.c
akallabeth: ok, did run this and implemented the use of the common callback here https://github.com/akallabeth/FreeRDP/tree/sso-mib  (feel free to cherry pick the commit to your branch)

what I did find is this though:
* `mib_public_client_app_new` uses stuff from `rdpSettings` and therefore should only be called after `PreConnect` callback and not here.

So, I'd suggest you do allocate the `mibClientWrapper` in `ContextNew` and free it in `ContextFree` but only initialize it on first use of the callback.

rupran: Just so I get it right:

If I move the allocation/freeing of `client_context->mibClientWrapper` into `freerdp_client_context_new/free`, I don't want to call `mib_public_client_app_new` there as the settings will not have been initialized. Instead, you want this initialized in which callback?

Inside `sso_mib_get_access_token` itself so a client (like `xfreerdp`) would:
- set `sso_mib_get_access_token` as their primary `GetAccessToken` callback and
- call `freerdp_set_common_access_token` with `client_cli_get_access_token`?

Or do I get the meanings of the new callback mixed up here?

akallabeth: @rupran so, in short yes (sorry, did misread it first :grin: )

akallabeth:
@freerdp-bot test
@rupran ok, I do have a working version of this now, I'll merge this one and create a new pr with the changes to unify this.

akallabeth (approved):
LGTM,
will do a follow up with the remaining changes.

PR #11607, opened 2025-05-21
[core,rdp] Use a single GetCommonAccessToken variable

akallabeth:
Only store the callback in rdpRdp and access it only via getter.


rupran:
> @rupran this should fix the access code retrieval.

Yep, looks good, thanks for the quick fix!

PR #11608, opened 2025-05-21
Sso mib deferred callback

akallabeth:


client/common/sso_mib_tokens.c
```diff
@@ -0,0 +1,199 @@
...
+	if (!client_context->mibClientWrapper->app)
+	{
+		const char* client_id =
+		    freerdp_settings_get_string(context->settings, FreeRDP_GatewayAvdClientID);
+		client_context->mibClientWrapper->app =
+		    mib_public_client_app_new(client_id, MIB_AUTHORITY_COMMON, NULL, NULL);
+	}
+
+	if (!client_context->mibClientWrapper->app)
+		;
```
rupran, line 137: Stray `;`

client/common/sso_mib_tokens.c
```diff
@@ -0,0 +1,199 @@
...
+					}
+				}
+			}
+			break;
+			default:
+				break;
+		}
+	}
+	if (!rc && client_context->mibClientWrapper->GetCommonAccessToken)
+		rc = client_context->mibClientWrapper->GetCommonAccessToken(context, tokenType, token,
```
rupran, line 194: Does this needs to call context->instance->GetAccessToken(context->instance, [...])`?

Additionally, we need to move error checking around here as otherwise `scope` and `req_cnf` would be NULL in case the first request fails (and sets the `state` to `SSO_MIB_STATE_FAILED`).

akallabeth: it does call this.
* The `GetCommonAccessToken` saved in the context is the one set before, so `freerdp_common_context`
* `scope` and `req_cnf` are not `NULL`, they are extracted from the `va_list` if `` is `ACCESS_TOKEN_TYPE_AAD`

client/common/client.c
```diff
@@ -231,7 +226,16 @@ int freerdp_client_stop(rdpContext* context)
...
-	return IFCALLRESULT(CHANNEL_RC_OK, pEntryPoints->ClientStop, context);
+	const UINT rc = IFCALLRESULT(CHANNEL_RC_OK, pEntryPoints->ClientStop, context);
+
+#ifdef WITH_SSO_MIB
+	rdpClientContext* client_context = (rdpClientContext*)context;
+	if (client_context->mibClientWrapper->app)
+		g_object_unref(client_context->mibClientWrapper->app);
+	free(client_context->mibClientWrapper);
+	client_context->mibClientWrapper = NULL;
+#endif // WITH_SSO_MIB
```
rupran, line 237: Needs deduplication with the `#ifdef SSO_MIB` above

client/common/sso_mib_tokens.c
akallabeth: ah, ok, now I see, did miss that case :grin:

akallabeth:
@rupran ok, did finish this, now should work.
can you confirm?

rupran:
> @rupran ok, did finish this, now should work. can you confirm?

Will do a deeper test later but looked good on first happy path test 👍

akallabeth:
@rupran great. I did push another commit to make the whole thing fully opaque.

rupran:
Yep, this works as expected (in xfreerdp):

- If everything is working, the broker either asks for an interactive login for the RDSAAD token or silently retrieves it and opens the RDP connection
- If the first broker call fails, it falls back to the URL method for both requests as the `state` is kept as `FAILED` - I simulated this by editing the scope to some random string
- If the first broker request succeeds, but the second request fails, we need to get the RDSAAD token from the browser (with the `ms-appx-web://` redirect) - I simulated this by passing `""` into the call to `mib_pop_params_set_kid`  instead of the actual `kid` value from `req_cnf`.

Compiling with `-DWITH_SSO_MIB=OFF` is the same as before :)

PR #11610, opened 2025-05-21
Flatpak update

akallabeth:


akallabeth:
@rupran @fmoessbauer did I get the `DBus` permissions right?

packaging/flatpak/com.freerdp.FreeRDP.json
```diff
@@ -37,6 +37,7 @@
 		"--socket=cups",
 		"--socket=wayland",
 		"--filesystem=xdg-download",
+		"--talk-name=com.microsoft.identity.Broker1",
```
fmoessbauer, line 40: Hi, sorry for the late response. There's a little typo which currently breaks it: This must be: `com.microsoft.identity.broker1`. I just changed that locally and now it works!

Anyways, many thanks for making this possible!

akallabeth: ok, copied from here: https://github.com/siemens/sso-mib/blob/main/dbus/spec/com.microsoft.identity.broker1.xml 
is this case difference expected?

fmoessbauer: DBus and their names... What is in the spec is the interface name. What we need to allow is the name on the bus, which is different in this case.
image

PR #11631, opened 2025-05-26
[stream] reset pool array size after clearing

asterwyx:
StreamPool becomes inconsistent after clearing. This might be a problem in some critical situation. Reset pool array size after clearing.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test
@asterwyx nice catch and fix, thank you!

PR #11635, opened 2025-05-26
[client,sdl] Wayland fixes for the SDL backend

motor-dev:
- properly compute the offsets for each monitor, when monitors have negative offsets. ‣ SDL_VIDEODRIVER=wayland sdl-freerdp3  /list:monitor * [3] [ASUSTek COMPUTER INC VG27AQML1A S2LMQS007046 (DP-1)] 2560x1440	+0+0 [4] [ASUSTek COMPUTER INC VG27AQML1A S2LMQS007036 (HDMI-A-2)] 1440x2560	+2560+-550 If the negative offset is not taken into account, the remote desktop will be drawn with black borders at the top or left hand side of the windows.

- also ensures these offsets are not reset at each redraw.

- add monitor ID in the window title to enable wayland compositors (e.g. hyprland) to find the various windows. hyprland has one workspace per monitor and refuses to let the application choose the monitor (or sdl-freerdp does not ask nicely). sdl-freedrp /multimon opens both windows on the same screen. adding a monitor identifier in the title allows hyprland rules to find the window and move it to the proper workspace.

These changes allow sdl-freedrp to find the correct geometry on Wayland compositors (tested on plasma/kwin and hyprland) when using SDL_VIDEODRIVER=wayland

### DIFF
--- a/client/SDL/SDL3/sdl_freerdp.cpp
+++ b/client/SDL/SDL3/sdl_freerdp.cpp
@@ -355,17 +355,6 @@ static BOOL sdl_draw_to_window(SdlContext* sdl, SdlWindow& window,
 
 	if (!freerdp_settings_get_bool(context->settings, FreeRDP_SmartSizing))
 	{
-		window.setOffsetX(0);
-		window.setOffsetY(0);
-		if (gdi->width < size.w)
-		{
-			window.setOffsetX((size.w - gdi->width) / 2);
-		}
-		if (gdi->height < size.h)
-		{
-			window.setOffsetY((size.h - gdi->height) / 2);
-		}
-
 		auto surface = sdl->primary.get();
 		if (!sdl_draw_to_window_rect(sdl, window, surface, { window.offsetX(), window.offsetY() },
 		                             rects))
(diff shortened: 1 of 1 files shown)


client/SDL/SDL3/sdl_freerdp.cpp
```diff
@@ -692,9 +698,8 @@ static BOOL sdl_create_windows(SdlContext* sdl)
 
 		if (freerdp_settings_get_bool(settings, FreeRDP_UseMultimon))
 		{
-			auto r = window.rect();
-			window.setOffsetX(startX - r.x);
-			window.setOffsetY(startY - r.y);
+			window.setOffsetX(originX - monitor->x);
```
motor-dev, line 701: This is due to some magic happening in settings.c
the monitor information is rebased so that the main monitor is always at offset 0, 0
the window rect has no idea about that and so cannot be used to compute the actual offset; we compute instead the offset based on the relative monitor position compared to the origin

this fixes an issue when the main monitor itself is not at origin 0, 0. This does not cause problem when the monitor is at position 0, 0.

Tested on Plasma X11 and Plasma Wayland where the main monitor was at position (0, 550).
Tested on Hyprland where the main monitor was at position (0, 0) and the second monitor at position (2560, -550)

all configurations lead to proper placement of the picture inside the window with no black border.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

vmpn:
@motor-dev @akallabeth thank you very much for

> add monitor ID in the window title to enable wayland compositors (e.g. hyprland) to find the various windows

this will be very handy to stop having to manually position windows on multiple monitors under wayland

PR #11639, opened 2025-05-28
[core,transport] The user context memory should be managed by the caller

KangLin:


akallabeth (changes requested):
No.

Abi/api breakage and contrary our use case.

KangLin:
@akallabeth This is the only use case. Contrary to the usual method of use.
The context must be managed by the user.
As the next major version of the API?

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/11642

PR #11641, opened 2025-05-28
[core,transport] Fix core dump, when the user context memory is managed by the caller.

KangLin:
Fix core dump, when the user context memory is managed by the caller.
ag:
```
    rdpTransportLayer* layer = nullptr;
    layer = transport_layer_new(freerdp_get_transport(context), 0);
    if (!layer)
        return nullptr;
    layer->userContext = this;
```

The application is core. when close.

It is recommended to modify this interface in the next major version. see: #11639


akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/11642

PR #11642, opened 2025-05-28
[core,transport] only free userContext if userContextSize > 0

akallabeth:


akallabeth:
@KangLin ok, with this our and your use case should work.
you can just create with `userContextSize = 0` and the memory will not be touched.

mfleisz:
I am missing the issue that is fixed here? If `contextSize` is 0 the pointer will be `NULL` not causing any issue with `free()`. If it is non-zero it will be allocated and freed.

akallabeth:
> I am missing the issue that is fixed here? If `contextSize` is 0 the pointer will be `NULL` not causing any issue with `free()`. If it is non-zero it will be allocated and freed.

yes, the pointer is in the public API and can be changed after the `_new` function.

PR #11646, opened 2025-06-02
[client,common] (re)initialize fuse root in cliprdr_file_context_init

akallabeth:
The clipboard might be (re)initialized multiple times, to do initialization and deinitialization in cliprdr_file_context_init and cliprdr_file_context_uninit.


akallabeth:
@freerdp-bot test

PR #11651, opened 2025-06-03
[WaitForXXObject] use infinite timeout where possible

akallabeth:
eliminate polling loops wherever possible.


libfreerdp/core/childsession.c
```diff
@@ -151,7 +151,7 @@ static int transport_bio_named_read(BIO* bio, char* buf, int size)
 
 			if (ptr->opInProgress)
 			{
-				DWORD status = WaitForSingleObjectEx(ptr->readEvent, 500, TRUE);
+				DWORD status = WaitForSingleObjectEx(ptr->readEvent, INFINITE, TRUE);
 				switch (status)
```
hardening: I feel like we could get stuck forever when it closes. We should probably do something in `doReadOp` to set `readEvent`

libfreerdp/crypto/tls.c
```diff
@@ -1003,25 +1003,28 @@ static int pollAndHandshake(rdpTls* tls)
 
 	do
 	{
-		HANDLE event = NULL;
+		HANDLE events[] = { NULL, freerdp_abort_event(tls->context) };
```
hardening: I think we should have the abort event first, as it's the one that will get notified the less. (so that `status` can be used more accurately to know if it's an abort or real traffic)

libfreerdp/core/childsession.c
akallabeth: hmm, looking at `treatReadResult` you might be right, there should be a `SetEvent` in all success cases.

libfreerdp/crypto/tls.c
akallabeth: ?
the first event is the one more likely to fire, so it should be faster this way?

libfreerdp/core/childsession.c
akallabeth: ok, will revert this, looks like it needs more work (the notifications are not happening in all cases)

PR #11659, opened 2025-06-04
Drive unicode

akallabeth:


akallabeth:
@freerdp-bot test

winpr/libwinpr/file/file.c
```diff
@@ -1585,3 +1585,32 @@ FILE* winpr_fopen(const char* path, const char* mode)
 	return result;
 #endif
 }
+
+#if !defined(_UWP) && !defined(_WIN32)
```
lazy5f, line 1589: Isn't this `#if defined(_UWP) || !defined(_WIN32)` as in file.h?

akallabeth: no.

channels/drive/client/drive_main.c
```diff
@@ -1060,6 +1060,76 @@ static UINT drive_register_drive_path(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints,
...
+	WINPR_ASSERT(pEntryPoints);
+
+	/* Enumerate all devices: */
+	const DWORD dlen = GetLogicalDriveStringsW(0, NULL);
+
+	WCHAR* devlist = calloc(dlen + 1, sizeof(WCHAR));
+	if (!devlist)
+		return ERROR_OUTOFMEMORY;
+
+	const DWORD rc = GetLogicalDriveStringsW(sizeof(devlist) - 1, devlist);
```
mfleisz: `sizeof(devlist)` is wrong here should be `dlen`

channels/drive/client/drive_main.c
```diff
@@ -1060,6 +1060,76 @@ static UINT drive_register_drive_path(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints,
...
+{
+	UINT error = ERROR_INTERNAL_ERROR;
+
+	WINPR_ASSERT(drive);
+	WINPR_ASSERT(pEntryPoints);
+
+	/* Enumerate all devices: */
+	const DWORD dlen = GetLogicalDriveStringsW(0, NULL);
+
+	WCHAR* devlist = calloc(dlen + 1, sizeof(WCHAR));
```
mfleisz: Over-allocating by 1

winpr/libwinpr/file/file.c
```diff
@@ -1585,3 +1585,32 @@ FILE* winpr_fopen(const char* path, const char* mode)
...
+
+DWORD GetLogicalDriveStringsA(DWORD nBufferLength, LPSTR lpBuffer)
+{
+	const char path[] = "/\0";
+	const size_t len = sizeof(path);
+	if (nBufferLength < len)
+		return WINPR_ASSERTING_INT_CAST(DWORD, len);
+
+	memcpy(lpBuffer, path, len);
+	return WINPR_ASSERTING_INT_CAST(DWORD, strnlen(path, len));
```
mfleisz: returning `len - 1` would allow the compiler to evaluate this at compile time rather than have an unnecessary `strnlen` runtime call

winpr/libwinpr/path/shell.c
```diff
@@ -58,6 +60,24 @@ static char* GetPath_XDG_RUNTIME_DIR(void);
  * http://msdn.microsoft.com/en-us/library/windows/desktop/bb762188/
  */
 
+#if defined(WIN32) && !defined(_UWP)
+
+static char* win_get_known_folder(REFKNOWNFOLDERID id, BOOL system)
```
mfleisz: The `system` parameter is confusing as -1 indicates the "Default User": https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath

Since this is only used once for `FOLDERID_ProgramData` (and in that case for no obvious reason) we should remove it and just use `NULL` for the handle parameter.

akallabeth: well, the `obvious` reason is that we do want the system install program data folder, not the user specific one ;)

winpr/libwinpr/file/file.c
akallabeth: would be wrong, `len == strlen(path) + 2`

winpr/libwinpr/path/shell.c
mfleisz: The "Default User" has nothing to do with system ... it is just a template User used when creating a new user account.

The program data folder is a system folder (just like system32) and is the same for each user ;-)

winpr/libwinpr/file/file.c
mfleisz: Actually `strlen` is wrong here since the function also counts the `\0` characters of the drive strings (except the trailing `\0`):

https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlogicaldrivestringsw#return-value
> If the function succeeds, the return value is the length, in characters, of the strings copied to the buffer, not including the terminating null character.

mfleisz (approved):

Issue #11658, opened 2025-06-04
Resize appWindow pixmap


ligangcc:
Today,  I found sometimes the size of appWindow pixmap not changed when window size changed.
I run command to open an RemoteApp:

```
./xfreerdp ....  /app:hidef:off,program:"C:\\Program Files\\Autodesk\\AutoCAD 2018\\acad.exe" +clipboard +async-update +multitransport /dynamic-resolution /cert:ignore /log-level:trace
```

the key log info:

```
[15:47:51:492] [323849:0004f10c] [DEBUG][com.freerdp.core.update] - [dump_window_state_order]: WindowUpdate windowId=1770570 clientOffset=(872,2306) windowOffset=(436,1153) windowClientDelta=(436,1153) windowWidth=1105 windowHeight=39 windowRects=((0,0,1105,39)) visibleOffset=(436,1153) visibilityRects=((0,0,1105,39))
[15:47:51:492] [323849:0004f10c] [DEBUG][com.freerdp.client.x11] - [xf_rail_window_common]: fieldFlags=100df00d, windowsId=1b044ad, appWindows=0x7f019c107a40
[15:47:51:492] [323849:0004f10c] [TRACE][com.freerdp.client.x11] - [xf_MoveWindow]: XMoveResizeWindow(0x7f019c000bc0, 79691811, 436, 1153, 1105, 39)
[15:47:51:494] [323849:0004f10c] [DEBUG][com.freerdp.client.x11] - [xf_UpdateWindowArea]: xf_UpdateWindowArea...appWindow=0x7f019c107a40
[15:47:51:494] [323849:0004f10c] [TRACE][com.freerdp.client.x11] - [xf_UpdateWindowArea]: XPutImage(0x7f019c000bc0, d: {79691817}, gc: {0x7f019c107b60}, image: [0x7f0184013bb0]{32}, src_x: {436}, src_y: {1153}, dest_x: {0}, dest_y: {0}, width: {1105}, height: {39})
[15:47:51:494] [323849:0004f10c] [INFO][com.freerdp.client.x11] - [LogDynAndXPutImage_ex]: drawable geometry (0,0,1103,27)
```

I add code to get the pixmap size before XPutImage

```
LogDynAndXPutImage_ex
....
if ((width == 0) || (height == 0))
    {
        const DWORD lvl = WLOG_WARN;
        if (WLog_IsLevelActive(log, lvl))
[...]

PR #11660, opened 2025-06-05
[client,x11]#11658,when size changed of an existing appWindows, except resize the main window, resize it's pixmap also

ligangcc:
[xf_rail_window_common]: fieldFlags=100df00d, windowsId=1b044ad, appWindows=0x7f019c107a40
here, because appWindows is existed, no create occurred.
then in xf_MoveWindow, if size changed, just call LogDynAndXMoveResizeWindow to resize it's main window,
the pixmap not changed.


akallabeth:
@freerdp-bot test

Issue #11686, opened 2025-06-16
NTLM fails after server redirection


petertalaber:
Hello,

**Describe the bug**
In freerdp3.15.0 (latest) if the client gets server redirection it fails to connect with NTLM authentication afterwards.

After some debugging it turned out that the problem seems caused by that according to [MS-RDPBCGR section 2.2.13.1] in case of cleartext password is received (inUnicode format) it includes a null-terminator. Password is then completely (including null-terminator) saved to settings->RedirectionPassword (and PasswordLength) by rdp_redirection_apply_settings(). After redirection when NLA/NTLM is negotiated, password is recalled by identity_set_from_settings_with_pwd() and hash is calculated including this null-terminated password in ntlm_compute_ntlm_v2_hash(). This hash then will be rejected by the target server. Note, that in normal connection, when password is prompted or it is passed as an argument, it won't contain the null-terminator, hence hash will be accepted by the server.
All versions from 3.0.0 is affected.

Note: with earlier freerdp2 versions like 2.11.5 it was working as expected. In this versions although password is stored by rdp_redirection_apply_settings() to settings->FreeRDP_RedirectionPassword (including null-terminator) but in nla_client_init() the redirection password is passed but with RedirectionPasswordLength-1 when calling sspi_SetAuthIdentityWithUnicodePassword() hence null-terminator is dropped before hash is calculated.

**To Reproduce**
1. Having an RDS load balanced environment and freerdp client with at lest version 3.0.0
2. Make sure that connected client will be load balanced and client will receive Server Redirection PDU.
3. Make sure that client will negotiate NTLM authentication (after redirection)
4. See error as server will drop the connection.

**Expected behavior**
In a load balanced environment xclient is able to connect to the server using NTLM.

**Application details**
* FreeRDP version (`xfreerdp /version`)
This is FreeRDP version 3.15.1-dev0 (c66139112)

* Command line used:
install/bin/xfreerdp /v:localhost /u:myuser /p:mysecret

* Output of `xfreerdp /buildconfig`
freerdp.latest/freerdp# install/bin/xfreerdp /buildconfig
This is FreeRDP version [install/bin/xfreerdp] 3.15.1-dev0 (c66139112)
[...]

PR #11687, opened 2025-06-16
[core,nla] use wcslen for password length

akallabeth:


akallabeth:
@petertalaber please confirm if this fixes your issue.

petertalaber:
Hi [akallabeth](https://github.com/akallabeth)!

Thanks for the quick reply. Yes, this fix solves #11686 issue as well.

PR #11688, opened 2025-06-17
Clear channel error prior to call channel init event proc

Pollux42:
Each call to `pChannelInitEventProc[Ex]` is checked by `CHANNEL_RC_OK != getChannelError(instance->context)`.

This PR resets error code so it does not leak between different channel event.

For example, when a channel fails at CHANNEL_EVENT_CONNECTED, the channel error state remains when we send the CHANNEL_EVENT_DISCONNECTED. So tt prevents the call of `PubSub_OnChannelDisconnected(instance->context->pubSub, instance->context, &e);` even if everything goes well during that step.

### DIFF
--- a/libfreerdp/core/client.c
+++ b/libfreerdp/core/client.c
@@ -881,9 +881,6 @@ UINT freerdp_channels_disconnect(rdpChannels* channels, freerdp* instance)
 			                                            CHANNEL_EVENT_DISCONNECTED, 0, 0);
 		}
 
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
this definitely breaks stuff.

ignoring the channel error (as discussed) would be less of an issue.

akallabeth:
@freerdp-bot test

akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #11690, opened 2025-06-18
[client,common] fix -mouse-motion

akallabeth:
* Move code to client/common to have it in place for all clients without modification
* Remember if a button was pressed and only suppress move events if no button is pressed.


akallabeth:
@freerdp-bot test

PR #11692, opened 2025-06-19
[core,proxy] fix IPv4 and IPv6 length

akallabeth:


cowpiebrownrice:
I confirm that the issue https://github.com/FreeRDP/FreeRDP/issues/11725 has been resolved with this merge. After applying the changes, the build and runtime on openSUSE Tumbleweed work correctly without the previously encountered library and dependency problems. Thanks for merging!

Issue #11645, opened 2025-05-31
screen scaling ratio affects the mouse positioning of the SDL client


tsz8899:
Symptom:
When using sdl3-freerdp.exe on Windows, if the local screen scaling ratio is not 100%, the mouse positioning will be misaligned.

if by manually setting the sdl3-freerdp.exe program compatibility: Replace the high DPI behavior -system 
local screen scaling ratio is not 100%,mouse positioning accurate
but using this method will cause the display to be blurry

need screen scaling ratio and high DPI alternative Settings are required without affecting mouse positioning
Please help

tanodzh:
I have similar issue on Armbian using SDL3 client. I can see it with or without scaling, both full screen or windowed mode.

PR #11710, opened 2025-06-26
[client,sdl3] Fix: SDL3 client mouse offset in DPI scaled environments

tsz8899:
Mouse Offset Issue and Solution for FreeRDP SDL3 Client

Merge pull request https://github.com/FreeRDP/FreeRDP/issues/11645

Problem Description

When using the FreeRDP SDL3 client on Windows environments with display scaling (DPI scaling) set to anything other than 100%, the mouse cursor within the application window exhibits an offset or misalignment. This issue occurs when the program itself is not scaled by the operating system (i.e., it's DPI-aware or explicitly set not to be scaled by the system).

The core of the problem lies in the discrepancy between the physical pixel coordinates reported by SDL for mouse events and the logical coordinates expected by the FreeRDP client's internal scaling mechanism.

Specifically:
*   SDL's mouse motion and button events (`SDL_MouseMotionEvent.x`, `SDL_MouseMotionEvent.y`, `SDL_MouseButtonEvent.x`, `SDL_MouseButtonEvent.y`) report coordinates in **physical pixels**.
*   The `sdl_scale_coordinates` function, responsible for scaling coordinates between the local client window and the remote RDP session, was calculating its scaling factors (`sx`, `sy`) based on the **logical dimensions** of the SDL window.
*   When the operating system's display scaling is not 100%, the physical pixel dimensions of the window are larger than its logical dimensions. Passing physical pixel coordinates directly to a function expecting logical coordinates (which are then scaled based on logical window size) leads to an incorrect transformation, resulting in the mouse offset.

Solution

The solution involves converting the mouse event coordinates from physical pixels to logical pixels *before* they are passed to the `sdl_scale_coordinates` function. This ensures that the scaling function receives coordinates in the expected logical space, allowing it to perform the correct transformation to the RDP session's coordinate system.

The `SDL_GetWindowDisplayScale()` function is used to obtain the display scale factor of the window. This factor is then used to divide the physical pixel coordinates, effectively converting them to logical coordinates.

**Changes Made:**

The following changes were applied to `client/SDL/SDL3/sdl_touch.cpp`:

1.  **In `sdl_scale_coordinates` function:**
    *   The calculation of `sx` and `sy` (scaling factors) was adjusted to use the logical dimensions of the SDL window (`SDL_GetWindowSize`) instead of the pixel dimensions (`window.rect()`). This ensures the scaling factors are correctly derived from the logical size of the window relative to the RDP session's resolution.

2.  **In `sdl_handle_mouse_motion` and `sdl_handle_mouse_button` functions:**
    *   Before calling `sdl_scale_coordinates`, the `ev->x` and `ev->y` (physical pixel coordinates from SDL mouse events) are divided by the `display_scale` obtained from `SDL_GetWindowDisplayScale(window)`. This converts the physical pixel coordinates to logical coordinates, aligning them with the expectations of `sdl_scale_coordinates`.

These modifications ensure that the mouse coordinates are consistently handled in the logical pixel space throughout the scaling process, resolving the offset issue in non-100% display scaling environments.

In Windows screen scaling environment test, when high DPI compatibility settings are not configured, or when set to Application, System, or System (Enhanced), the mouse position remains accurate.

### DIFF
--- a/client/SDL/SDL3/sdl_touch.cpp
+++ b/client/SDL/SDL3/sdl_touch.cpp
@@ -61,10 +61,11 @@ BOOL sdl_scale_coordinates(SdlContext* sdl, Uint32 windowId, INT32* px, INT32* p
 			continue;
 		}
 
-		auto size = window.rect();
+		int logical_w, logical_h;
+		SDL_GetWindowSize(window.window(), &logical_w, &logical_h);
 
-		sx = size.w / static_cast<double>(gdi->width);
-		sy = size.h / static_cast<double>(gdi->height);
+		sx = static_cast<double>(logical_w) / static_cast<double>(gdi->width);
+		sy = static_cast<double>(logical_h) / static_cast<double>(gdi->height);
 		offset_x = window.offsetX();
 		offset_y = window.offsetY();
 		break;
@@ -202,8 +203,18 @@ BOOL sdl_handle_mouse_motion(SdlContext* sdl, const SDL_MouseMotionEvent* ev)
 	sdl->input.mouse_focus(ev->windowID);
 	const BOOL relative =
 	    freerdp_client_use_relative_mouse_events(sdl->common()) && !sdl->hasCursor();
-	auto x = static_cast<INT32>(relative ? ev->xrel : ev->x);
-	auto y = static_cast<INT32>(relative ? ev->yrel : ev->y);
+
+	// Get the window and its display scale
+	SDL_Window* window = SDL_GetWindowFromID(ev->windowID);
+	if (!window)
+		return FALSE;
+
+	float display_scale = SDL_GetWindowDisplayScale(window);
+
+	// Convert physical pixel coordinates to logical coordinates
+	auto x = static_cast<INT32>(relative ? ev->xrel : ev->x / display_scale);
+	auto y = static_cast<INT32>(relative ? ev->yrel : ev->y / display_scale);
+
 	sdl_scale_coordinates(sdl, ev->windowID, &x, &y, TRUE, TRUE);
 	return freerdp_client_send_button_event(sdl->common(), relative, PTR_FLAGS_MOVE, x, y);
 }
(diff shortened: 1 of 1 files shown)


akallabeth:
@tsz8899 nice one, thank you!
could you please run `clang-format` on your changes though?

tsz8899:
@akallabeth I've formatted the code with the clang-format tool from Visual Studio 2022. Could you please verify if this is correct now? Thanks.

akallabeth:
@freerdp-bot test

akallabeth (changes requested):
Ok, tested on debian 13/gnome with 175% display scaling:
* 8 without the patch the mouse position works correct
* with the patch it is completely off

so, looks like your patch does not always work, can't merge it in this state.

akallabeth:
https://ci.freerdp.com/job/code-quality-checker/10768/clang/fileName.-304069795/

some issues with the changes

vp1981:
Is there any issue with this PR?

I ran into the same problem on the host with two monitors and scaling, but on Hyprland (in fact, all was working fine until some time ago, maybe two-three weeks ago). Interesting point, on the host with one monitor (FullHD) without scaling sdl3-freerdp (from master) works just fine (no mouse pointer offset).

akallabeth:
@vp1981 read my comment above.
it has the same bugs as the current version, it just works for a different flavor of high resolution compositor breaking the ones that currently (kind of) work.

vp1981:
> @vp1981 read my comment above.

Sorry, I read it, but it kind of slipped my mind.

> it has the same bugs as the current version, it just works for a different flavor of high resolution compositor breaking the ones that currently (kind of) work.

Unfortunately, I don't have access to a host with GNOME (wayland preferably), but have hosts with KDE. I'll try to test the sdl3-freerdp with this PR and scaling.

Regarding SDL3, could a different behavior be due to different version of SDL3? As I mentioned, it used to work fine, but after some update (I'm using Archlinux), either Hyprland or SDL3 (more likely the latter) the sdl3-freerdp started working weirdly (no mouse offset only on the system without scaling).

Currently, the freerdp (freerd-git) package is build with SDL3 3.2.26 and SDL3_TTF (-git) 3.2.0

vp1981:
From https://github.com/libsdl-org/SDL/releases, 3.2.26:

> -Fixed destination coordinates when using scale with the software renderer

Could it be that one? Then my case is different from @tsz8899...

vp1981:
Ok, it doesn't as simple, as @akallabeth already mentioned.
1. On Hyprland with one FullHD monitor w/o scaling sdl3-freerdp, both with this PR and w/o it, everything is fine.
2. On Hyprland with two monitors (I have two hosts: 4k+4k and 4k+FullHD, scaling+scaling and scaling+n/scaling resp.) this PR gives the expected behavior (but I only checked in 4k+4k case).
3. But in KDE with scaling (125% and 75%) sdl-freerdp with this PR **gives** mouse offset.

I don't know **how** scaling is done/implemented in KDE and Hyprland, but maybe they are different things with the same term.

akallabeth:
should be fixed with [#sdl clean](https://github.com/FreeRDP/FreeRDP/pull/12173)

PR #11715, opened 2025-06-30
[common,settings] force reallocation on caps copy

akallabeth:


hardening:
Tested here and it fixes the problem

hardening (approved):

PR #11723, opened 2025-07-02
kerberos: do various tries for TGT retrieval in u2u

hardening:
In a server-side user2user scenario, you can have either a configuration with a service mapped to a user or something host based. So your keytab will contain either a `TERMSRV/<host>@<realm>` entry that is able to retrieve a TGT, or the machine account associated with `<host>$@<realm>`.

This patch makes the kerberos SSPI first try to retrieved a TGT as if it was a service mapped to a user. And if it doesn't work (or if the entry is missing), it goes for a host based attempt.

### DIFF
--- a/winpr/libwinpr/sspi/Kerberos/kerberos.c
+++ b/winpr/libwinpr/sspi/Kerberos/kerberos.c
@@ -1281,6 +1281,108 @@ static SECURITY_STATUS SEC_ENTRY kerberos_InitializeSecurityContextW(
 	return status;
 }
 
+#ifdef WITH_KRB5
+static BOOL retrieveTgtForPrincipal(KRB_CREDENTIALS* credentials, krb5_principal principal,
+                                    krb5_creds* creds)
+{
+	BOOL ret = FALSE;
+	krb5_kt_cursor cur = { 0 };
+	krb5_keytab_entry entry = { 0 };
+	if (krb_log_exec(krb5_kt_start_seq_get, credentials->ctx, credentials->keytab, &cur))
+		goto cleanup;
+
+	do
+	{
+		krb5_error_code rv =
+		    krb_log_exec(krb5_kt_next_entry, credentials->ctx, credentials->keytab, &entry, &cur);
+		if (rv == KRB5_KT_END)
+			break;
+		if (rv != 0)
+			goto cleanup;
+
+		if (krb5_principal_compare(credentials->ctx, principal, entry.principal))
+			break;
+		rv = krb_log_exec(krb5glue_free_keytab_entry_contents, credentials->ctx, &entry);
+		memset(&entry, 0, sizeof(entry));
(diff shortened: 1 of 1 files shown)


winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -1281,6 +1281,110 @@ static SECURITY_STATUS SEC_ENTRY kerberos_InitializeSecurityContextW(
...
+		if (rv == KRB5_KT_END)
+			break;
+		if (rv != 0)
+			goto cleanup;
+
+		if (krb5_principal_compare(credentials->ctx, principal, entry.principal))
+			break;
+		const krb5_error_code res =
+		    krb_log_exec(krb5glue_free_keytab_entry_contents, credentials->ctx, &entry);
+		memset(&entry, 0, sizeof(entry));
```
akallabeth, line 1306: why not just move the definition to the loop?

hardening: Oh no special reason, I think it's part of the old code I've moved out from `AcceptSecurityContext`

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -85,7 +85,7 @@ const SecPkgInfoW KERBEROS_SecPkgInfoW = {
 	KERBEROS_SecPkgInfoW_NameBuffer,   /* Name */
 	KERBEROS_SecPkgInfoW_CommentBuffer /* Comment */
 };
-
+#define WITH_KRB5
```
akallabeth: looks like slipped through debug code?

akallabeth (changes requested):

winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -1281,6 +1281,109 @@ static SECURITY_STATUS SEC_ENTRY kerberos_InitializeSecurityContextW(
...
+	ret = FALSE;
+
+	/*
+	 * if it's not working let's try with <host>$@<REALM> (note the dollar)
+	 */
+	char hostDollar[300] = { 0 };
+	if (target_princ->length < 2)
+		goto out;
+
+	snprintf(hostDollar, sizeof(hostDollar) - 1, "%s$@%s", target_princ->data[1].data,
```
akallabeth: return check or cast `(void)` to ignore.

winpr/libwinpr/sspi/Kerberos/kerberos.c
hardening: arf correct

akallabeth:
@hardening ok, `git clang-format` missing.

akallabeth (approved):
LGTM

PR #11734, opened 2025-07-08
winpr: re-introduce the credentials module

hardening:
The credential module from FreeRDP2 was removed, but sometime you still need to have support for `Cred[Marshal|Unmarshal]Credential` especially when it's a smartcard logon and that the cert sha1 is encoded using these functions.

### DIFF
--- a/winpr/include/winpr/cred.h
+++ b/winpr/include/winpr/cred.h
@@ -22,14 +22,17 @@
 #include <winpr/winpr.h>
 
 #ifdef _WIN32
+#include <windows.h>
 #include <wincred.h>
 #else
+
+#include <winpr/wtypes.h>
 
 #define CERT_HASH_LENGTH 20
 
 typedef enum
 {
-	CertCredential,
+	CertCredential = 1,
 	UsernameTargetCredential,
 	BinaryBlobCredential,
 	UsernameForPackedCredentials,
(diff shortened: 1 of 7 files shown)


winpr/libwinpr/credentials/credentials.c
```diff
@@ -0,0 +1,402 @@
...
+	{
+		WLog_ERR(TAG, "invalid string length");
+		return FALSE;
+	}
+	return TRUE;
+}
+
+static UINT cred_encode(const BYTE* bin, size_t len, WCHAR* cred)
+{
+	static const char enc[] = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789#-";
```
akallabeth: `base64`? we already have that in the lib.

hardening:
@freerdp-bot can you re test please ?

.clang-tidy
```diff
@@ -76,6 +76,7 @@ Checks: >
     -readability-magic-numbers,
     -readability-math-missing-parentheses,
     -readability-misleading-indentation,
+    -readability-non-const-parameter,
```
akallabeth: I don't see the point in disabling warnings just because some stubs were added.

which functions do you need?
Just leave out the rest, does not make a lot of sense to add stuff that is not implemented.

hardening: Well I may need the other functions in the future

winpr/libwinpr/credentials/credentials.c
hardening: unfortunately it's not base64

.clang-tidy
akallabeth: then add them *in the future*?

winpr/libwinpr/credentials/credentials.c
akallabeth: too bad :/

.clang-tidy
akallabeth: I don't like all the stubs that are still in the project (even worse the ones that do not warn you they are not implemented).
so, we can always add new functions when we need them, there is no backward compatibility anyway if the function is just a stub.

hardening:
Finally made it pass all the tests !

hardening:
Anyone to merge that ?

winpr/include/winpr/cred.h
```diff
@@ -43,7 +46,91 @@ typedef struct
...
+} CREDENTIAL_TARGET_INFORMATIONW, *PCREDENTIAL_TARGET_INFORMATIONW;
+
+typedef enum
+{
+	CredUnprotected,
+	CredUserProtection,
+	CredTrustedProtection,
+	CredForSystemProtection
+} CRED_PROTECTION_TYPE,
+    *PCRED_PROTECTION_TYPE;
```
mfleisz, line 132: Formatting?

hardening: Well if it passes the formatting test it should be ok, isn't it ? I've just ran `clang-format` on it, no custom changes from me...

PR #11739, opened 2025-07-10
[core,proxy] fix BIO read methods

akallabeth:
Introduced with ae05778644023d2b44e69c4fff59843ee367689f the return of BIO_read changed.
Revert this to original behaviour and fix a bug in the proxy code


akallabeth:
@hardening did have to change the whole thing a little. (`BIO_write` had the same bug)

PR #11741, opened 2025-07-14
[client,common] fix sso_mib_get_access_token return value in error case

rupran:
Use `FALSE` instead of `ERROR_INTERNAL_ERROR` which could be interpreted as "truthy" at the respective call site.


akallabeth (approved):
LGTM
nice catch

akallabeth:
@freerdp-bot test

Issue #11744, opened 2025-07-16
rpath is not set correctly anymore


daym:
We are trying to package this for gnu guix.  There, we have an rpath validator... and it finds problems with freerdp.  It had worked with freerdp 2.

Now (with freerdp 3.16.0) we get the following error:

```
cmake "../source" "-DCMAKE_BUILD_TYPE=Release" "-DCMAKE_INSTALL_PREFIX=/gnu/store/888yjb757h3xaqxfjlvf01d53c1zah4i-freerdp-3.16.0" "-DCMAKE_INSTALL_LIBDIR=lib" "-DCMAKE_I
NSTALL_RPATH_USE_LINK_PATH=TRUE" "-DCMAKE_INSTALL_RPATH=/gnu/store/888yjb757h3xaqxfjlvf01d53c1zah4i-freerdp-3.16.0/lib" "-DCMAKE_VERBOSE_MAKEFILE=ON" "-DCMAKE_C_FLAGS= -Wno-error=incompatible-poin
ter-types -Wno-error=int-conversion" "-DWITH_JPEG=ON" "-DWITH_SSE2=ON" "-DWITH_PULSE=ON" "-DWITH_CUPS=ON" "-DBUILD_TESTING=ON" "-DWITH_SERVER=ON" "-DWITH_SHADOW=ON" "-DWITH_PROXY=ON" "-DWITH_ADD_P
LUGIN_TO_RPATH=ON"
[...]
validating RUNPATH of 13 binaries in "/gnu/store/cs88609kzfjc74ify8n8lw7i8y3v3nad-freerdp-3.16.0/lib"...
[...]

PR #11745, opened 2025-07-16
build: Fix RUNPATH setting.

daym:
Fixes <https://github.com/FreeRDP/FreeRDP/issues/11744>.

### DIFF
--- a/server/proxy/modules/bitmap-filter/CMakeLists.txt
+++ b/server/proxy/modules/bitmap-filter/CMakeLists.txt
@@ -42,4 +42,11 @@ addtargetwithresourcefile(${PROJECT_NAME} FALSE "${PROJECT_VERSION}" SRCS FALSE)
 
 target_link_libraries(${PROJECT_NAME} winpr freerdp)
 
+# Fix RPATH for proxy plugins in subdirectory
+if(NOT WIN32 AND NOT APPLE AND NOT FREEBSD)
+  set_target_properties(${PROJECT_NAME} PROPERTIES
+    INSTALL_RPATH "\$ORIGIN/../../${CMAKE_INSTALL_LIBDIR}:\$ORIGIN/../.."
+  )
+endif()
+
 install(TARGETS ${PROJECT_NAME} DESTINATION ${FREERDP_PROXY_PLUGINDIR})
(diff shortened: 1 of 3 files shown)


akallabeth (changes requested):
this patch is wrong.
the setting is done in `cmake/ConfigureRPATH.cmake`

daym:
I had seen cmake/ConfigureRPATH.cmake already--but it would add an extraordinary amount of `..` to the search path for libraries if I  did it globally there, maybe accessing files way up as (possibly attacker-provided) dependencies. That said, if this risk is acceptable, we also can do it globally. That would be just this:

```
diff -ru orig/FreeRDP/cmake/ConfigureRPATH.cmake FreeRDP/cmake/ConfigureRPATH.cmake
--- orig/FreeRDP/cmake/ConfigureRPATH.cmake	2025-07-17 10:42:17.557667810 +0200
+++ FreeRDP/cmake/ConfigureRPATH.cmake	2025-07-17 11:32:13.692935801 +0200
@@ -15,10 +15,10 @@
     option(WITH_ADD_PLUGIN_TO_RPATH "Add extension and plugin path to RPATH" OFF)
     if(WITH_ADD_PLUGIN_TO_RPATH)
       set(CFG_INSTALL_RPATH
-          "\$ORIGIN/../${FREERDP_EXTENSION_REL_PATH}:\$ORIGIN/../${FREERDP_PLUGIN_PATH}:\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}:\$ORIGIN/.."
+          "\$ORIGIN/../${FREERDP_EXTENSION_REL_PATH}:\$ORIGIN/../${FREERDP_PLUGIN_PATH}:\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}:\$ORIGIN/..:\$ORIGIN/../.."
       )
     else()
-      set(CFG_INSTALL_RPATH "\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}:\$ORIGIN/..")
+      set(CFG_INSTALL_RPATH "\$ORIGIN/../${CMAKE_INSTALL_LIBDIR}:\$ORIGIN/..:\$ORIGIN/../..")
     endif()
   endif()
 endif(APPLE)
```

The reason the rpath doesn't work is that the plugin directories are now deeper than in freerdp 2. If you want to verify it, check the rpath in the plugin `.so` files.

akallabeth:
@daym ok, you're only talking about proxy modules.
that might explain why we can't reproduce :grin: 

I see your problem, doing a custom installation where `RPATH` is required to find the core libraries the modules will not find them.

need to think about a good way to do this, most likely a patch to `AddTargetWithResourceFile` to set it relative to `CMAKE_INSTALL_PATH`

akallabeth:
ok, replaced by https://github.com/FreeRDP/FreeRDP/pull/11747

PR #11754, opened 2025-07-30
[winpr,timezone] ensure thread-safe initialization

mnauw:
The singleton/local/static structures might crash into difficulties if it is performed in parallel on several threads.  This has also really been known to happen in a (proxy) server setting with 1 thread per connecting client.

This particular approach may or may not be convoluted (depending on taste).  Alternatives are of course possible, including hooking such initialization into some sort of `freerdp_init()` function.

### DIFF
--- a/winpr/libwinpr/timezone/TimeZoneIanaAbbrevMap.c
+++ b/winpr/libwinpr/timezone/TimeZoneIanaAbbrevMap.c
@@ -27,6 +27,7 @@
 #include <unistd.h>
 
 #include <winpr/string.h>
+#include <winpr/synch.h>
 #include "timezone.h"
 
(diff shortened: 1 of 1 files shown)


mfleisz:
Hm - this might actually be a use case for `InitOnceExecuteOnce`

hardening:
We already have that with `InitOnceExecuteOnce`

mnauw:
Adjusted to use the specified function.

hardening:
@freerdp-bot can you give a test please ?

mnauw:
Ping? Failing check looks like a CI issue, but have no info/view on that.

mfleisz:
@freerdp-bot test

PR #11761, opened 2025-08-06
Clean up bugs exposed on systems with high core counts

### LINKED ISSUE #11721: TestSynchCritical failing
I have set up a CI running builds daily on multiple architectures (aarch64, ppc64le, s390x, x86_64) and now I am getting `TestSynchCritical` failures in ~50% runs.

The test log contains (this is on aarch64, TEST1_RUNTIME_MS has been extended to 100, thus the 800 ms value)
````
5/153 Testing: TestSynchCritical
5/153 Test: TestSynchCritical
Command: "/home/jenkins/workspace/freerdp/label/aarch64/build/Testing/TestSynch" "TestSynchCritical"
Directory: /home/jenkins/workspace/freerdp/label/aarch64/build/winpr/libwinpr/synch/test
"TestSynchCritical" start time: Jul 02 07:23 UTC
Output:
Deadlock will be assumed after 800 ms.
CriticalSection failure: RecursionCount is invalid
CriticalSection failure: RecursionCount is invalid
CriticalSection failure: Possible dead lock detected
<end of output>
Test time =   0.82 sec
Test Failed.
"TestSynchCritical" end time: Jul 02 07:23 UTC
"TestSynchCritical" time elapsed: 00:00:00
````
Running the test manually gives for example (on ppc64le)
````
[jenkins@kernelci-18 Testing]$ ./TestSynch TestSynchCritical
Deadlock will be assumed after 400 ms.
[jenkins@kernelci-18 Testing]$ ./TestSynch TestSynchCritical
Deadlock will be assumed after 400 ms.
CriticalSection failure: RecursionCount is invalid
CriticalSection failure: RecursionCount is invalid
CriticalSection failure: Possible dead lock detected
[jenkins@kernelci-18 Testing]$ ./TestSynch TestSynchCritical
Deadlock will be assumed after 400 ms.
[jenkins@kernelci-18 Testing]$ ./TestSynch TestSynchCritical
Deadlock will be assumed after 400 ms.
CriticalSection failure: RecursionCount is invalid
CriticalSection failure: RecursionCount is invalid
CriticalSection failure: Possible dead lock detected
[jenkins@kernelci-18 Testing]$ ./TestSynch TestSynchCritical
Deadlock will be assumed after 400 ms.
CriticalSection failure: RecursionCount is invalid
CriticalSection failure: RecursionCount is invalid
CriticalSection failure: Possible dead lock detected
````
The difference between these systems and my s390x and x86_64 ones (where the test seems to run OK) is that the "failing" systems have many CPUs (80 and 128). Isn't there a know issue with the test? Any workarounds possible?

### PR DESCRIPTION
**Pull request by pvachon**
This PR contains two fixes for issues exposed while building and testing FreeRDP on systems with large numbers of CPU cores ( >=80 experimentally).

The first fixes a case where if the number of CPU cores is greater than the number of lines in a chunk of data to be processed with the `yuv` CODEC, the line skip size would be zero, resulting in a divide by zero error later in the processing.

The second is a fix for how winpr generates a thread ID. In the current upstream code, the thread ID is the 32 LSBs of the pointer to the current thread's `pthread_t` as returned by `pthread_self()`. This would occasionally result in a collision where two (or more) threads would have the same ID, readily reproducible with large numbers of threads when running `TestSynchCritical`. This would result in a thread incorrectly determining it holds a `CRITICAL_SECTION`, and takes the fast path assuming that it is a recursive entry into the `CRITICAL_SECTION`.

This patch replaces the (ab)use of the `pthread_self()` pointer value with an OS-provided (or runtime-provided) thread ID, for Linux, FreeBSD and macOS, For all other cases, the fallback is to XOR the 32 MSBs of the pointer returned from `pthread_self()` with the 32 LSBs, providing a bit more entropy. There is still a (distant) risk of a collision in this case, this probability is vastly diminished.

From my testing, and one other user testing, it looks like this fixes #11721.

Tested on various Linux distributions, FreeBSD 14.3 and macOS 15.

### DIFF
--- a/winpr/libwinpr/thread/thread.c
+++ b/winpr/libwinpr/thread/thread.c
@@ -29,11 +29,10 @@
 
 #include <winpr/thread.h>
 
-#if defined(__MACOSX__) || defined(__FreeBSD__)
+#if defined(__FreeBSD__)
 #include <pthread_np.h>
 #elif defined(__linux__)
 #include <sys/syscall.h>
-
 #endif
 
 #ifndef MIN
@@ -918,14 +917,13 @@ HANDLE _GetCurrentThread(VOID)
 
 DWORD GetCurrentThreadId(VOID)
 {
-#if defined(__FreeBSD__) || defined(__MACOSX__)
+#if defined(__FreeBSD__)
 	int tid = pthread_getthreadid_np();
 	return tid;
 #elif defined(__linux__)
 	pid_t tid = syscall(SYS_gettid);
 	return tid;
 #else
-#warning Using possibly broken GetCurrentThreadId
 	pthread_t tid = pthread_self();
 	/* Since pthread_t can be 64-bits on some systems, take just the    */
 	/* lower 32-bits of it for the thread ID returned by this function. */
--- a/winpr/libwinpr/thread/thread.c
+++ b/winpr/libwinpr/thread/thread.c
@@ -918,11 +918,9 @@ HANDLE _GetCurrentThread(VOID)
 DWORD GetCurrentThreadId(VOID)
 {
 #if defined(__FreeBSD__)
-	int tid = WINPR_CXX_COMPAT_CAST(DWORD, pthread_getthreadid_np());
-	return tid;
+	return WINPR_CXX_COMPAT_CAST(DWORD, pthread_getthreadid_np());
 #elif defined(__linux__)
-	pid_t tid = WINPR_CXX_COMPAT_CAST(DWORD, syscall(SYS_gettid));
-	return tid;
+	return WINPR_CXX_COMPAT_CAST(DWORD, syscall(SYS_gettid));
 #else
 	pthread_t tid = pthread_self();
 	/* Since pthread_t can be 64-bits on some systems, take just the    */
--- a/winpr/libwinpr/thread/thread.c
+++ b/winpr/libwinpr/thread/thread.c
@@ -918,10 +918,10 @@ HANDLE _GetCurrentThread(VOID)
 DWORD GetCurrentThreadId(VOID)
 {
 #if defined(__FreeBSD__)
-	int tid = pthread_getthreadid_np();
+	int tid = WINPR_CXX_COMPAT_CAST(DWORD, pthread_getthreadid_np());
 	return tid;
 #elif defined(__linux__)
-	pid_t tid = syscall(SYS_gettid);
+	pid_t tid = WINPR_CXX_COMPAT_CAST(DWORD, syscall(SYS_gettid));
 	return tid;
 #else
 	pthread_t tid = pthread_self();
--- a/winpr/libwinpr/thread/thread.c
+++ b/winpr/libwinpr/thread/thread.c
@@ -29,6 +29,13 @@
 
 #include <winpr/thread.h>
 
+#if defined(__MACOSX__) || defined(__FreeBSD__)
+#include <pthread_np.h>
+#elif defined(__linux__)
+#include <sys/syscall.h>
+
+#endif
+
 #ifndef MIN
 #define MIN(x, y) (((x) < (y)) ? (x) : (y))
 #endif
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by sharkcz** [comment]
I believe the root cause for the TestSynchCritical failures has been fixed.
Running
````
good=0; bad=0; for i in `seq 1 100`; do ./build/Testing/TestSynch TestSynchCritical && good=$(($good+1)) || bad=$(($bad+1)); done; echo "good=$good bad=$bad"
````
returns 100/100 of "passed" on both 80-core aarch64 system and 128-core (in fact 2x16 cores with 4 threads each) ppc64le system. It wasn't possible before the fix (~35% failure rate on aarch64). Both systems are running Fedora 42.

But when running the test as part of the "ctest" suite, the default 400ms timeout (`TEST1_RUNTIME_MS`) is too short for the 128-core ppc64le system. No real errors are reported (no "RecursionCount is invalid", etc), just "Possible dead lock detected".
````
good=0; bad=0; for i in `seq 1 100`; do ctest --test-dir build -R TestSynchCritical && good=$(($good+1)) || bad=$(($bad+1)); done; echo "good=$good bad=$bad"
````
passes when I update `TEST1_RUNTIME_MS` to `100`, with the reported runtimes around 650ms (greater than the 400ms limit). Perhaps the `dwDeadLockDetectionTimeMs` at line [331](https://github.com/FreeRDP/FreeRDP/blob/master/winpr/libwinpr/synch/test/TestSynchCritical.c#L331) variable should be updated based on the core count or similar.

**Comment by mfleisz** [comment]
@pvachon thanks for the contribution! Very much appreciated!

@sharkcz thanks a lot for testing and pointing out the issue with the timeout!

We are currently having some issues with our infrastructure but I'll have a look and merge this as soon as things are resolved. Thanks again guys!

**Comment by sharkcz** [comment]
you are welcome and great job from @pvachon for sure

**Comment by sharkcz** [comment]
also tested on a 160-core (dual 20-core x 4 threads) ppc64le system and no problems there with `TEST1_RUNTIME_MS` increased, the times per test are ~700ms

**Comment by mfleisz** [comment]
@freerdp-bot test

**Comment by mfleisz** [comment]
@pvachon looks like there is an issue with the `#include` on MacOS: https://ci.freerdp.com/job/osx/15157/label=macosx/console

**Comment by pvachon** [comment]
I really don't want to admit how this happened, but I did not originally include the fix for macOS in the original PR. I've updated with what I also tested on macOS, and retested on FreeBSD and Linux.

**Review by akallabeth — APPROVED** [review]
LGTM

**Comment by akallabeth** [comment]
@freerdp-bot test

File: winpr/libwinpr/thread/thread.c [hunk 1]
```diff
@@ -911,11 +917,19 @@ HANDLE _GetCurrentThread(VOID)
 
 DWORD GetCurrentThreadId(VOID)
 {
+#if defined(__FreeBSD__)
+	int tid = pthread_getthreadid_np();
+	return tid;
```
akallabeth: please use `WINPR_CXX_COMPAT_CAST` to cast to `DWORD` (aka unsigned int) to avoid sign warnings

winpr/libwinpr/thread/thread.c
```diff
@@ -911,11 +917,19 @@ HANDLE _GetCurrentThread(VOID)
...
+#elif defined(__linux__)
+	pid_t tid = syscall(SYS_gettid);
```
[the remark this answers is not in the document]
akallabeth: please use `WINPR_CXX_COMPAT_CAST` to cast to `DWORD` (aka unsigned int) to avoid sign warnings

winpr/libwinpr/thread/thread.c
pvachon: Done -- did the same for the Linux equivalent.

winpr/libwinpr/thread/thread.c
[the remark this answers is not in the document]
pvachon: Done -- did the same for the FreeBSD equivalent.

pvachon:
@akallabeth Thanks for the review -- I've added those casts explicitly and retested. Let me know if there are any other changes I can make!

winpr/libwinpr/thread/thread.c
```diff
@@ -918,10 +918,10 @@ HANDLE _GetCurrentThread(VOID)
 DWORD GetCurrentThreadId(VOID)
 {
 #if defined(__FreeBSD__)
-	int tid = pthread_getthreadid_np();
+	int tid = WINPR_CXX_COMPAT_CAST(DWORD, pthread_getthreadid_np());
```
akallabeth: ok, the destination type is still wrong.
just replace it with a return.

winpr/libwinpr/thread/thread.c
```diff
@@ -918,10 +918,10 @@ HANDLE _GetCurrentThread(VOID)
...
 	return tid;
 #elif defined(__linux__)
-	pid_t tid = syscall(SYS_gettid);
+	pid_t tid = WINPR_CXX_COMPAT_CAST(DWORD, syscall(SYS_gettid));
```
akallabeth: ok, the destination type is still wrong.
just replace it with a return.

winpr/libwinpr/thread/thread.c
pvachon: :man_facepalming: Fixed.

winpr/libwinpr/thread/thread.c
pvachon: :man_facepalming: Fixed.

pvachon:
:man_facepalming: Didn't think that one through all the way -- but I've fixed to directly return the result from `syscall(SYS_gettid)` or `pthread_getthreadid_np`.

akallabeth:
@freerdp-bot test

PR #11771, opened 2025-08-12
[client,x11] ignore floatbar events

akallabeth:
* need to check if button release events are properly propagated if the button press was on the window
* need to check if button release events are properly propagated if the button press was on the floatbar


akallabeth:
@mfweir can you test this pr? should address your issue.

Issue #11628, opened 2025-05-25
Fullchain pem certificate is not being converted well to a rdpCertificate object


OfirMatasas:
**Is your feature request related to a problem? Please describe.**
My proxy is using the FreeRDP library with a pem file as the **fullchain** server certificate, which being converted to a `rdpCertificate` on runtime using the `freerdp_certificate_new_from_pem` function.
For some reason, whenever the rdpCertificate object is being presented to the client, the content of the certificate is invalid, for not containing the intermediate chain of the certificate properly.
For instance, if the content of the original pem certificate is:
> -----BEGIN CERTIFICATE-----
> ABC
> -----END CERTIFICATE-----
> -----BEGIN CERTIFICATE-----
> 123
> -----END CERTIFICATE-----

The certificate the server is showing to the client is:
> -----BEGIN CERTIFICATE-----
> ABC
> -----END CERTIFICATE-----
> -----BEGIN CERTIFICATE-----
> ABC
> -----END CERTIFICATE-----

So as you can see, the intermediate chain of the certificate is missing, and it's being replaced with the first part of the certificate, which is kind of a weird behavior.
I followed the code, and saw that the `freerdp_certificate_new_from_pem` functions calls the `freerdp_certificate_new_from` function, which calls the `freerdp_certificate_new_from_x509` function, which passes `NULL` as the chain of the certificate I'm trying to convert from pem to rdpCertificate object (which I suspect to be the problem here).

**Describe the solution you'd like**
Whenever the `freerdp_certificate_new_from_pem` is being called with a fullchain pem certificate, there should be some kind of logic for checking if this certificate is fully-chained, and act the same (not passing NULL as the chain of the certificate).
Furthermore, if this logic I just suggested is missing, the certificate shouldn't be duplicated like that, and just keep the first part and ditch the second part of the certificate.

**Describe alternatives you've considered**
I've tried to find a way to bypass this issue with other functions in the `certificate.c` file, but couldn't find any.. Maybe I missed something?

**Additional context**
None

akallabeth:
yes, only the server certificate is extracted, no chain.
(we did not require that so far)

but we'll be happy about pull requests adding this

OfirMatasas:
Could anyone assist in finding a bypass for this issue?

PR #11775, opened 2025-08-13
Add extract_chain_from_pem function for getting cert chain

omatasas:
Based on this issue:
- https://github.com/FreeRDP/FreeRDP/issues/11628

Can anyone of the admins check this PR and say is it's sufficient for fixing this chain being missing from the certificate?

### DIFF
--- a/libfreerdp/crypto/certificate.c
+++ b/libfreerdp/crypto/certificate.c
@@ -1335,19 +1335,23 @@ fail:
 	return NULL;
 }
 
-static STACK_OF(X509)* extract_chain_from_pem(const char* pem)
+static STACK_OF(X509)* extract_chain_from_pem(const char* pem, BOOL isFile)
 {
 	if (!pem)
 	{
 		return NULL;
 	}
 
-	BIO* bio = BIO_new_mem_buf(pem, strlen(pem));
+	BIO* bio = NULL;
+	if (isFile)
+		bio = BIO_new_file(pem, "rb");
+	else
+		bio = BIO_new_mem_buf(pem, strlen(pem));
+
 	if (!bio)
 	{
 		return NULL;
 	}
-
 
 	X509* leaf = PEM_read_bio_X509(bio, NULL, NULL, NULL);
 	if (!leaf)
@@ -1355,7 +1359,6 @@ static STACK_OF(X509)* extract_chain_from_pem(const char* pem)
 		BIO_free(bio);
 		return NULL;
 	}
-
 
 	STACK_OF(X509)* chain = sk_X509_new_null();
 	if (!chain)
@@ -1381,7 +1384,7 @@ static rdpCertificate* freerdp_certificate_new_from(const char* file, BOOL isFil
 	X509* x509 = x509_utils_from_pem(file, strlen(file), isFile);
 	if (!x509)
 		return NULL;
-	rdpCertificate* cert = freerdp_certificate_new_from_x509(x509, extract_chain_from_pem(file));
+	rdpCertificate* cert = freerdp_certificate_new_from_x509(x509, extract_chain_from_pem(file, isFile));
 	X509_free(x509);
 	return cert;
 }


libfreerdp/crypto/certificate.c
```diff
@@ -1335,12 +1335,51 @@ rdpCertificate* freerdp_certificate_new_from_x509(const X509* xcert, const STACK
 	return NULL;
 }
 
+STACK_OF(X509)* extract_chain_from_pem(const char* pem)
```
akallabeth: should be `static`

libfreerdp/crypto/certificate.c
```diff
@@ -1335,12 +1335,51 @@ rdpCertificate* freerdp_certificate_new_from_x509(const X509* xcert, const STACK
...
+	return chain;
+}
+
 static rdpCertificate* freerdp_certificate_new_from(const char* file, BOOL isFile)
 {
 	X509* x509 = x509_utils_from_pem(file, strlen(file), isFile);
 	if (!x509)
 		return NULL;
-	rdpCertificate* cert = freerdp_certificate_new_from_x509(x509, NULL);
+	rdpCertificate* cert = freerdp_certificate_new_from_x509(x509, extract_chain_from_pem(file));
```
akallabeth: unhandled case  `isFile == TRUE`

libfreerdp/crypto/certificate.c
```diff
@@ -1335,12 +1335,51 @@ rdpCertificate* freerdp_certificate_new_from_x509(const X509* xcert, const STACK
...
+		return NULL;
+	}
+
+	X509* cert = NULL;
+	while ((cert = PEM_read_bio_X509(bio, NULL, NULL, NULL)) != NULL)
+	{
+		sk_X509_push(chain, cert);
+	}
+
+	return chain;
```
akallabeth, line 1379: leak `bio`?
leak `leaf`?

libfreerdp/crypto/certificate.c
omatasas: Fixed.

libfreerdp/crypto/certificate.c
omatasas: Handled.

libfreerdp/crypto/certificate.c
omatasas: Fixed.

akallabeth:
@freerdp-bot test

omatasas:
> Overall LGTM, doing commit squash && force push would help to have a single commit for this.
> 
> need to do some tests (will take till next week, sorry) to confirm this does not break stuff before merging.
> 
> thank you for the contribution!

Thank you :)
But in general, it should fix the issue and support full-chain server certificates right?
So the clients should get the certificate with the intermediate CAs if available in the original certificate we passed to FreeRDP.

I'm asking because I tried to make it work locally, and it didn't go as expected, and maybe there're other places in the code we should take care of for fixing it.

Could you please test this new functionality with a full-chain certificate, just to make sure we didn't miss anything?
@akallabeth

akallabeth:
@omatasas not sure full chain certs work as expected with `RDP` (there are some MS specifics in the protocol with the older security modes) but for newer `TLS` based tunnels it should work.
Need to test this myself though and currently busy with another task.
If I do not finish this by end of next week ping me, I might have forgotten :grin:

libfreerdp/crypto/certificate.c
```diff
@@ -1335,12 +1335,56 @@ rdpCertificate* freerdp_certificate_new_from_x509(const X509* xcert, const STACK
...
+}
+
 static rdpCertificate* freerdp_certificate_new_from(const char* file, BOOL isFile)
 {
 	X509* x509 = x509_utils_from_pem(file, strlen(file), isFile);
 	if (!x509)
 		return NULL;
-	rdpCertificate* cert = freerdp_certificate_new_from_x509(x509, NULL);
+	rdpCertificate* cert = freerdp_certificate_new_from_x509(x509, extract_chain_from_pem(file, isFile));
```
akallabeth, line 1387: possible leak of `extract_chain_from_pem(file, isFile)` here.
in case `freerdp_certificate_new_from_x509` fails the chain is leaking.

akallabeth:
replaced by #11787

PR #11780, opened 2025-08-14
[client,common] add freerdp_client_populate_settings_from_rdp_file_un…

akallabeth:
…checked

Sometimes it is required to have raw values from a RDP file converted to rdpSettings without consistency modifications. Expose this with a new function.


akallabeth:
@freerdp-bot test

PR #11787, opened 2025-08-18
Fullchain support

akallabeth:


akallabeth:
@omatasas taken your commits and squashed them.
Added missing details for fullchain support for shadow/proxy and client.

PR #11790, opened 2025-08-20
[winpr,credentials] prefer utf-8 over utf-16-LE

akallabeth:
When implementing the wide character functions convert the string to utf-8 before manipulation to avoid endianess issues.


hardening (approved):
I've not followed too much the issue with endian-ness but seems ok for me

Issue #11800, opened 2025-08-21
extend the timeout for TestSynchCritical?


sharkcz:
This is a follow-up of #11721 and the fix in #11761. With the threading code fixed I am experiencing timeouts for the `TestSynchCritical` check on some of my big machines, primarily when run from `ctest`. Could we either increase the value of `TEST1_RUNTIME_MS` (https://github.com/FreeRDP/FreeRDP/blob/master/winpr/libwinpr/synch/test/TestSynchCritical.c#L11) or set the test's timeout dynamically, eg. computed from the number of CPUs? Doubling the current value of `50` to `100` would resolve my issue, but perhaps a better solution exists.

akallabeth:
not sure how the best way to go forward on this is.
we did set a low timeout (along with some optimizations on other tests) to cut down on pull request build times.

[edit] ok, running with a higher timeout does not increase it a lot, but not sure why this is taking that long, might hide some fundamental flaw there.

sharkcz:
As far as I can tell, only my big ppc64le systems are affected. The systems have spinning rust as storage (no SSD or NVMe) and 4 threads per core. So perhaps a hw or platform thing. And as stated in https://github.com/FreeRDP/FreeRDP/pull/11761#issuecomment-3158559537 running the test standalone is OK. Somehow `ctest` makes them to run "slower".

Perhaps the timeout value could be exposed via cmake and made configurable ...

akallabeth:
since I do not really have anything to test here, can you do a pr?

sharkcz:
> since I do not really have anything to test here, can you do a pr?

for the simple increase/doubling or for something more sophisticated? The former will be simple, but latter not so much :-)

akallabeth:
@sharkcz if that does not increase the overall test time a lot then a simple doubling would do.

PR #11808, opened 2025-08-22
[winpr,synch] increase timeout for TestSynchCritical

sharkcz:
Increase the deadlock detection timeout in TestSynchCritical to accommodate longer runtime on systems with large number of CPUs/threads. The usual test run time when the threads are finishing correctly won't change.

Fixes: https://github.com/FreeRDP/FreeRDP/issues/11800

A note for the future - perhaps the timeout should be scaled based on the number of CPUs, because the number of created threads in the test is based on the number of the CPUs.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM, thank you.

PR #11809, opened 2025-08-22
Enable RDPECAM client in flatpak release

cvpcs:
Adds necessary configuration to enable RDPECAM client webcam redirection support under flatpak.

### DIFF
--- a/packaging/flatpak/com.freerdp.FreeRDP.json
+++ b/packaging/flatpak/com.freerdp.FreeRDP.json
@@ -29,7 +29,7 @@
 	],
 	"command": "sdl-freerdp",
 	"finish-args": [
-		"--device=dri",
+		"--device=all",
 		"--share=ipc",
 		"--socket=x11",
 		"--share=network",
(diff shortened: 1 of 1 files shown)


packaging/flatpak/com.freerdp.FreeRDP.json
```diff
@@ -29,7 +29,7 @@
 	],
 	"command": "sdl-freerdp",
 	"finish-args": [
-		"--device=dri",
+		"--device=all",
```
cvpcs: I'm curious what everyone's opinion of this is. I added it because it's the [only mechanism for granting webcam access](https://docs.flatpak.org/en/latest/sandbox-permissions.html#device-access), but then I noticed we also don't explicitly grant the `pcsc` socket for smartcards either despite freerdp supporting it.

Should the policy of the flatpak be to only grant minimal permissions by default and users need to explicitly grant these extra permissions to support their scenarios? Or should the flatpak request all permissions need to support the full breadth of its feature set?

akallabeth: good practice is only request minimum necessary, with user extending where required.

akallabeth: anyhow, if you want the flatpak from flatpak updated look at https://github.com/flathub/com.freerdp.FreeRDP this here is a manifest that resembles something like the other nightly builds.

cvpcs: > good practice is only request minimum necessary, with user extending where required.

Makes sense. I've reverted the permissions change.

> anyhow, if you want the flatpak from flatpak updated look at https://github.com/flathub/com.freerdp.FreeRDP this here is a manifest that resembles something like the other nightly builds.

I've also opened this PR against the flathub repo: https://github.com/flathub/com.freerdp.FreeRDP/pull/87

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #11812, opened 2025-08-25
[proxy,channels] refactor dynamic channel

akallabeth:
Split DynvcTrackerPeekFn into smaller fragments each handling one kind of packet


akallabeth:
@freerdp-bot test

PR #11814, opened 2025-08-26
[core,settings] fix ReceivedCapabilities reset

akallabeth:


winpr/libwinpr/sspi/Kerberos/kerberos.c
```diff
@@ -1356,6 +1367,7 @@ static BOOL retrieveSomeTgt(KRB_CREDENTIALS* credentials, const char* target, kr
 
 	ret = FALSE;
 
+#if defined(WITH_KRB5_MIT)
```
hardening, line 1370: That fixes the build with heimdal but does a different behaviour

akallabeth: @hardening yes, but I did not find a proper way to implement it there.
had to add https://github.com/FreeRDP/FreeRDP/pull/11815/commits/7a3030b556fa2c38ff4057fa949ee55439d09475 as `rdpear` also does not compile against `Heimdal` at all.

PR #11822, opened 2025-08-28
audin_oss: do not reset mic volume on capture start

### LINKED ISSUE #11821: Microphone volume is set to 100 on every capture start
### Current state

FreeRDP sets microphone volume to 100% on every capture start.

### Offending code

https://github.com/FreeRDP/FreeRDP/blob/5a41b7da0b931308bc829557f31eec05b8dc494b/channels/audin/client/oss/audin_oss.c#L175-L193

### Problem description 

On my setup, setting microphone volume to 100% causes noise cancelling to fail in the conferencing software (Teams) running on the remote. I have to manually set the volume back to 30% every time I male a call to make it work.

### Expected behaviour

FreeRDP should not interfere with client OS configuration. Let the user manage their audio settings by themselves.

Defenso-QTH:
Closes #11821


akallabeth:
@freerdp-bot test

PR #11828, opened 2025-08-30
Geometry fix

akallabeth:


akallabeth:
> LGTM. Given the PR name I was desperately looking for what was fixed, but in fact it's just code reshaping.

length checks ;)
there was a bug in there that only the provided length variable was checked but not the stream length.

PR #11830, opened 2025-09-01
core/redirection: Ensure stream has enough space for all parameters

pnowack:
```
While commit a08e3fc5315af8a10c9b1d1333ec3c3d1066a172 fixed the case,
where the stream did not have enough space for the target certificate,
it did miss out a few other cases where the remaining length is just
checked, but not increased when needed. Fix this by now also covering
the remaining cases.

Fixes: a08e3fc5315af8a10c9b1d1333ec3c3d1066a172
```


hardening (approved):
LGTM

PR #11837, opened 2025-09-02
fix(winpr): ncrypt_pkcs11: set correct PIV certificate labels

TheBestTvarynka:
Hi,

I am working on RDP smartcard logon support. I noticed that PIV certificate labels in the `winpr/libwinpr/ncrypt/ncrypt_pkcs11.c` do not match labels specified in the NIST SP 800-73 (section _4.3. Object Identifiers_).

I added new labels that correspond to the values from the specification. FreeRDP fails to construct a container name without this fix. I tested changes on a PIV-compatible smart card: YubiKey 5 Nano.


hardening:
@freerdp-bot can you test that PR for us please ?

PR #11843, opened 2025-09-04
[codec,yuv] wrap step calculation

akallabeth:


libfreerdp/codec/yuv.c
```diff
@@ -803,7 +811,7 @@ static BOOL pool_encode(YUV_CONTEXT* WINPR_RESTRICT context, PTP_WORK_CALLBACK c
 	{
 		const RECTANGLE_16* rect = &regionRects[x];
 		const UINT32 height = rect->bottom - rect->top;
-		const UINT32 steps = (height + context->heightStep / 2) / context->heightStep;
+		const UINT32 steps = getSteps(height, context->heightStep);
```
hardening, line 814: aren't we gonna have off by 1 errors with that patch ?

akallabeth: integer rounding, the old code did have a slight chance of rounding down instead of up

PR #11851, opened 2025-09-10
rdpecam: add some new callbacks to the HAL

hardening:
This patch adds Activate and Deactivate callbacks to the HAL, matching the messages exchanged on the channel. This is to prepare the support of a windows HAL using the microsoft media fundation framework.
It also documents the HAL's functions.


hardening:
@freerdp-bot come on do a proper test and don't care too much about `TestConnect` failing

Issue #11859, opened 2025-09-12
Opt-out of SDL hotkeys


vredez:
Please add an option to completely turn off SDL hotkeys.

A quick glance over the current implementation suggests, that there is no way to disable those hotkeys at all.
Even configuring individual hotkey mappings to empty strings in `sdl-freerdp.json`, or setting the mask-key array to empty won't do as workaround, since hard-coded fallback values will be used then instead.

It is really annoying to suddenly get disconnected from a session, just because I pressed `RSHIFT+D` (which would produce a capital `D` when touch-typing), because that's the unlucky default hotkey for disconnecting.

IMHO, remapping all hotkeys to some exotic combination is just shifting the problem. Sooner or later some software on running on the target machine expects such a combination to work.

vredez:
Update: I tried to fix this myself via #11860.

PR #11860, opened 2025-09-13
[client,sdl2,sdl3] Allow SDL hotkeys to be disabled via JSON config

vredez:
## Summary
Add a new configuration attribute `EnableHotkeys` for the SDL configuration file `sdl-freerdp.json` which defaults to `true`. Setting this to `false` disables hotkey handling for the SDL2 and SDL3 clients.

## Motivation
Hotkeys used by SDL{2,3} clients can interfere with software running on the target host. Allowing those hotkeys to be disabled completely remedies that issue (see #11859).

## How to test
- Create this example configuration file at `$XDG_CONFIG_HOME/freerdp/sdl-freerdp.json`:
   ```json
   {
      "EnableHotkeys": false
   }
   ```
- Connect to a host with the SDL{2,3} client
- No hotkeys are active (for example pressing `R_SHIFT`+`D` doesn't close the client)

### DIFF
--- a/client/SDL/SDL2/sdl_kbd.cpp
+++ b/client/SDL/SDL2/sdl_kbd.cpp
@@ -481,37 +481,40 @@ BOOL sdlInput::keyboard_handle_event(const SDL_KeyboardEvent* ev)
 {
 	WINPR_ASSERT(ev);
 	const UINT32 rdp_scancode = sdl_scancode_to_rdp(ev->keysym.scancode);
-	const SDL_Keymod mods = SDL_GetModState();
 
-	if ((mods & _hotkeyModmask) == _hotkeyModmask)
+	if (_hotkeysEnabled)
 	{
-		if (ev->type == SDL_KEYDOWN)
+		const SDL_Keymod mods = SDL_GetModState();
+		if ((mods & _hotkeyModmask) == _hotkeyModmask)
 		{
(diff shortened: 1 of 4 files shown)


akallabeth (changes requested):
Hi, thank you for this.
But we've already got the `SDL_KeyModMask` - we can simply add a special value there to disable the handling altogether. (current handling seems to not handle the case when `KMOD_NONE` is explicitly set as value, which should be fine for this)

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/11862

thank you for the effort. really appreciated (help is always *very* welcome)

did not use it as is, because I think it is easier to just set `KMOD_NONE`  in the config and handle that appropriately (and update the manpage so people find it)

vredez:
Hi, thanks for your feedback!
Setting `KMOD_NONE` is the first thing I've tried. Unfortunately, this doesn't work with the current implementation because it triggers the fallback default setting of the modifier mask.

With
```json
{
    "SDL_KeyModMask": [ "SDL_KMOD_NONE" ]
}
```
pressing `RSHIFT`+`D` still terminates the client.

akallabeth:
@vredez then try the referenced pull request I just opened ;)

vredez:
> @vredez then try the referenced pull request I just opened ;)

Sorry I did miss that, works fine 😊.

PR #11866, opened 2025-09-16
[client,x11] fix rails/desktop switch

akallabeth:
* Add TODO log entries for (non)monitored desktop messages
* Enable/disable RAILS mode according to these messages


akallabeth:
@freerdp-bot test

PR #11873, opened 2025-09-19
[core,arm] extract redirected username

akallabeth:


hardening (approved):

hardening:
Happy to see that I'm not the only one to have Azure interests ;-)

PR #11876, opened 2025-09-21
[cmake,pkg-config] properly set requires fields

akallabeth:


pnowack (commented):
Not too familiar with cmake, but what I can confirm is that this fixes building g-r-d against FreeRDP/master (no more build break with https://github.com/FreeRDP/FreeRDP/commit/d355082c32503c188ebe8dfcf9e3ebd1931072a9, when this MR is applied too).

pnowack (approved):

PR #11885, opened 2025-09-25
wst error handling

akallabeth:


hardening (approved):
Seems ok. So strange that the previous code was explicitly using signed ints...

akallabeth:
> Seems ok. So strange that the previous code was explicitly using signed ints...

most likely a leftover from the times everything was an `int` :grin:

PR #11892, opened 2025-09-26
Azure domain

akallabeth:
* Provide a default domain `AzureAD` if no domain was provided via `/d:domain` setting
* Modify `freerdp_interruptible_get_line` so that a suggested value can be provided which can be edited.

this way we can provide a sane default for (most) azure/entra setups while still allowing a user to override the defaults.

### DIFF
--- a/include/freerdp/utils/passphrase.h
+++ b/include/freerdp/utils/passphrase.h
@@ -31,9 +31,46 @@ extern "C"
 {
 #endif
 
-	FREERDP_API int freerdp_interruptible_getc(rdpContext* context, FILE* file);
+	/** @brief Read a single character from a stream.
+	 * This function will check \ref freerdp_shall_disconnect_context and abort when the session
+	 * terminates.
+	 *
+	 * The function will abort on any of <ctrl>+[cdz] or end of stream conditions as well as when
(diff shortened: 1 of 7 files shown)


akallabeth:
@RMCampos this should work for you, I'm still trying to figure out how to address the cli issue with suggested username/domain/password.
(with `sdl-freerdp` you can edit them, with the `cli` these fields are currently not shown)

akallabeth:
@freerdp-bot test

include/freerdp/utils/passphrase.h
```diff
@@ -31,9 +31,46 @@ extern "C"
 {
 #endif
 
-	FREERDP_API int freerdp_interruptible_getc(rdpContext* context, FILE* file);
+	/** @brief Read a single character from a stream.
```
hardening, line 34: As these are new functions, we should have a `@since` tag

akallabeth: no, not new, just did not have documentation.
the `suggest` stuff is new though.

hardening: oh my bad, sorry for the noise

PR #11912, opened 2025-10-02
dsp_ffmpeg: fix latency buildup during resampling

FriederHannenheim:
## Issue description
On certain devices we encountered issues where Microphone audio over Freerdp would suffer from increasing latency which could build up to multiple minutes.

After a thorough examination I have traced this issue to the resampling function. Input frames to this functions have a variable number of samples. However the output frame is set up on the first call to `swr_convert_frame` and will always keep that sample count. 

Excerpt from the documentation of `swr_convert_frame`: https://ffmpeg.org/doxygen/5.1/group__lswr.html#gac482028c01d95580106183aa84b0930c
- The output [AVFrame](https://ffmpeg.org/doxygen/5.1/structAVFrame.html) can be NULL or have fewer allocated samples than required. In this case, any remaining samples not written to the output will be added to an internal FIFO buffer, to be returned at the next call to this function or to [swr_convert()](https://ffmpeg.org/doxygen/5.1/group__lswr.html#gaa5bb6cab830146efa8c760fa66ee582a).

Because the frames after the first one usually contain more samples this FIFO keeps building up and introduces the latency.

## Proposed solution
Resetting the frame after the data has been used fixes this problem

### DIFF
--- a/libfreerdp/codec/dsp_ffmpeg.c
+++ b/libfreerdp/codec/dsp_ffmpeg.c
@@ -219,6 +219,34 @@ static void ffmpeg_close_context(FREERDP_DSP_CONTEXT* WINPR_RESTRICT context)
 	}
 }
 
+static void ffmpeg_setup_resample_frame(FREERDP_DSP_CONTEXT* WINPR_RESTRICT context,
+                                        AUDIO_FORMAT* WINPR_RESTRICT format)
+{
+	if (context->resampled->buf[0] != NULL)
+		av_frame_unref(context->resampled);
+
+	if (context->common.encoder)
+	{
+		context->resampled->format = context->context->sample_fmt;
+		context->resampled->sample_rate = context->context->sample_rate;
+	}
+	else
+	{
+		context->resampled->format = AV_SAMPLE_FMT_S16;
+
+		WINPR_ASSERT(format->nSamplesPerSec <= INT_MAX);
+		context->resampled->sample_rate = (int)format->nSamplesPerSec;
+	}
+
+#if LIBAVUTIL_VERSION_INT >= AV_VERSION_INT(57, 28, 100)
+	av_channel_layout_default(&context->resampled->ch_layout, format->nChannels);
+#else
+	const int64_t layout = av_get_default_channel_layout(format->nChannels);
+	context->resampled->channel_layout = layout;
+	context->resampled->channels = format->nChannels;
+#endif
+}
+
 static BOOL ffmpeg_open_context(FREERDP_DSP_CONTEXT* WINPR_RESTRICT context)
 {
 	int ret = 0;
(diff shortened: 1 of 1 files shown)


libfreerdp/codec/dsp_ffmpeg.c
```diff
@@ -820,7 +820,10 @@ BOOL freerdp_dsp_ffmpeg_encode(FREERDP_DSP_CONTEXT* WINPR_RESTRICT context,
 
 	if (context->context->frame_size <= 0)
 	{
-		return ffmpeg_encode_frame(context->context, context->resampled, context->packet, out);
+		if (!ffmpeg_encode_frame(context->context, context->resampled, context->packet, out))
```
akallabeth: revert, this must return here (or a default return must be added)

libfreerdp/codec/dsp_ffmpeg.c
```diff
@@ -862,6 +865,16 @@ BOOL freerdp_dsp_ffmpeg_encode(FREERDP_DSP_CONTEXT* WINPR_RESTRICT context,
 			}
 		} while (rest > 0);
 
+		av_frame_unref(context->resampled);
+		context->resampled->format = context->context->sample_fmt;
```
akallabeth: best extract that from `ffmpeg_open_context`, wrap it in some function `foo_reset` and use it here and in `ffmpeg_open_context`

akallabeth (changes requested):
Overall nice work, thank you for the pr!

libfreerdp/codec/dsp_ffmpeg.c
FriederHannenheim: I agree. Maybe the function call should be on line 816 above the call to `ffmpeg_resample_frame` so that the code isn't duplicated in the if-branches?

akallabeth: you only apply the code for the decoder, this is in the `else` branch

FriederHannenheim: I got a little confused with the braces there. It needs to be applied in both cases. The first branch of the if is for cases where the output frame size is not fixed, not for decoding.

akallabeth: yes, so you nave 2 to 3 use cases for this, best wrap it in a static function (except for the unref, that should only be used here I guess)

akallabeth:
@freerdp-bot test

FriederHannenheim:
Okay I have now refactored it into a static function with the check if av_frame_unref is required on top. What do you think of this patch?

libfreerdp/codec/dsp_ffmpeg.c
```diff
@@ -219,6 +219,33 @@ static void ffmpeg_close_context(FREERDP_DSP_CONTEXT* WINPR_RESTRICT context)
...
+		context->resampled->format = AV_SAMPLE_FMT_S16;
+
+		WINPR_ASSERT(format->nSamplesPerSec <= INT_MAX);
+		context->resampled->sample_rate = (int)format->nSamplesPerSec;
+	}
+
+#if LIBAVUTIL_VERSION_INT >= AV_VERSION_INT(57, 28, 100)
+	av_channel_layout_default(&context->resampled->ch_layout, format->nChannels);
+#else
+	context->resampled->channel_layout = layout;
```
akallabeth, line 245: you need to define `layout` here ;)

FriederHannenheim: oops

akallabeth:
@freerdp-bot test

PR #11921, opened 2025-10-10
[core,timer] fix reschedule interval

akallabeth:
time was calculated in µs instead of ms.
@vokov0aa thank you for the hint

### DIFF
--- a/channels/video/client/video_main.c
+++ b/channels/video/client/video_main.c
@@ -1160,7 +1160,7 @@ static UINT video_plugin_initialize(IWTSPlugin* plugin, IWTSVirtualChannelManage
 
 	if (status == CHANNEL_RC_OK)
 		video->context->priv->timerID =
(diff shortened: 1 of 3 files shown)


ncPUMA:
@akallabeth ,
You should increase old timer intervals too.
Here:
https://github.com/FreeRDP/FreeRDP/blob/a6ae83202c3cd0f9505e0ca7ce0ef24835ef7d2f/channels/video/client/video_main.c#L1163
and here:
https://github.com/FreeRDP/FreeRDP/blob/a6ae83202c3cd0f9505e0ca7ce0ef24835ef7d2f/client/Wayland/wlf_disp.c#L30
and here:
https://github.com/FreeRDP/FreeRDP/blob/a6ae83202c3cd0f9505e0ca7ce0ef24835ef7d2f/client/X11/xf_disp.c#L43

PR #11930, opened 2025-10-13
Align width and height for AVC444 decoding to 32

eduar-hte:
Windows seems to be aligning to 32 in current versions of RDP when sending AVC444 bitstreams. This can be noticed in non-standard resolutions (that are not multiple of 32 after aligning to 16), where the combination of the `AVC444_CHROMAv2` 
bitstream fails because the `nTotalWidth` argument to `general_ChromaV2ToYUV444` is incorrect.

To trigger the issue run the FreeRDP client with the cmd-line argument `/gfx:AVC444` and set a resolution where the width is not a multiple of 32 (even after the width value is aligned to 16). For example, I can reproduce the issue with: `/size:802x600`, `/size:1026x768` or `/size:1282x1024`.

image

I've reproduced the issue in the following versions of Windows;

- Win11 25H2 (Version 10.0.26200.6584)
- Win10 22H2 (Version 10.0.19045.6332)

The following table shows the value for the first element of `iStride` (the other two elements are that value divided by 2 for a `YUV420` frame) for each resolution. This value appears to be that of the aligned value plus 64. The table shows that the `alignedWidth` that is computed in `yuv444_combine_work_callback` doesn't match that of `iStride` (after adding 64) when the alignment is 16, but it's when it's aligned to 32.

| Width | Height | iStride (0) | alignedWidth (16) | alignedWidth (32) | alignedWidth (16) + 64 | alignedWidth (32) + 64 |
|--------|---------|------------|---------------------|---------------------|---------------------------|---------------------------|
| 800 | 600 | 864 | 800 | 800 | 864 | 864 |
| 1024 | 768 | 1088 | 1024 | 1024 | 1088 | 1088 |
| 1280 | 1024 | 1344 | 1280 | 1280 | 1344 | 1344 |
| 802 | 600 | 896 | 816 | 832 | 880 (x) | 896 |
| 1026 | 768 | 1120 | 1040 | 1056 | 1104 (x) | 1120 |
| 1282 | 1024 | 1376 | 1296 | 1312 | 1360 (x) | 1376 |

# NOTES

- 16-byte alignment was initially introduced in commit 8adc2ba
- BTW, I think the 16-byte aligned values that are currently calculated in `yuv444_combine_work_callback` always match the values in `yuv->width` & `yuv->height` used as input, as they're initialized with the width & height of the surface which are 16-byte aligned in `gdi_CreateSurface` since commit 3d1cec8.
  - That is, I think the code I updated to try and address the issue was not currently necessary, but it now is (to do 32-byte alignment).
  - I also tested a fix updating the alignment used in `gdi_CreateSurface` to set it to 32, which works, but I decided to limit the potential splash damage of this PR by only adjusting the alignment in `yuv444_combine_work_callback`.
- I could not find a reference in the RDP GFX spec that indicates the alignment requirement/change.
  - I wonder if the alignment was always 32 or if this has been a change, as if it was the latter, an alternative approach may be needed to support older versions of Windows as well (`iStride[0] - 64`?)

### DIFF
--- a/libfreerdp/codec/yuv.c
+++ b/libfreerdp/codec/yuv.c
@@ -479,7 +479,7 @@ static void CALLBACK yuv444_combine_work_callback(PTP_CALLBACK_INSTANCE instance
 	const RECTANGLE_16* rect = &param->rect;
 	WINPR_ASSERT(rect);
 
-	const UINT32 alignedWidth = yuv->width + ((yuv->width % 16 != 0) ? 16 - yuv->width % 16 : 0);
+	const UINT32 alignedWidth = yuv->width + ((yuv->width % 32 != 0) ? 32 - yuv->width % 32 : 0);
 	const UINT32 alignedHeight =
 	    yuv->height + ((yuv->height % 16 != 0) ? 16 - yuv->height % 16 : 0);
 


akallabeth:
@freerdp-bot test

akallabeth:
@eduar-hte which h264 decoder did you test with?
could be a bug in that one as well. (as you said, the spec does not mention a change and there is still the 16 pixel alignment mentioned)

eduar-hte:
> @eduar-hte which h264 decoder did you test with? could be a bug in that one as well. (as you said, the spec does not mention a change and there is still the 16 pixel alignment mentioned)

I forgot to mention that. I've reproduced this with either FFmpeg (n8.0 & n4.4.6) and OpenH264 (v2.4.1). I'm building the client on Windows but I expect the issue to happen in other platforms as well.

I initially thought it could be a decoder issue too (as I run into an issue with newer versions of OpenH264 as well), but after a while I found it was controlled by the resolution's width.

As I mentioned in the report, I think the issue happens only with AVC444 with ChromaV2, as the combination function requires the width to perform the combination. I'm referencing this in the generic implementation `general_ChromaV2ToYUV444` as it's simpler to analyze the code, but this applies to the optimized versions as well (in fact, the screen capture was made on a build using the SSE 4.1 YUV primitives).

https://github.com/FreeRDP/FreeRDP/blob/7d86a9259b2c9e3cbd7f0c91e13a182d7b41f5eb/libfreerdp/primitives/prim_YUV.c#L190
https://github.com/FreeRDP/FreeRDP/blob/7d86a9259b2c9e3cbd7f0c91e13a182d7b41f5eb/libfreerdp/primitives/prim_YUV.c#L206
https://github.com/FreeRDP/FreeRDP/blob/7d86a9259b2c9e3cbd7f0c91e13a182d7b41f5eb/libfreerdp/primitives/prim_YUV.c#L208

I wonder if this may be an implementation change in RDP to leverage AVX instructions, which use 256-bit registers and require 32-byte alignment.

pnowack:
> Windows seems to be aligning to 32 in current versions of RDP when sending AVC444 bitstreams. This can be noticed in non-standard resolutions (that are not multiple of 32 after aligning to 16)

I've done something similar for my VA-API handling in g-r-d, due to a bug that I discovered with @mfleisz some months ago. `msrdc` fails (with a protocol error message), when using software decoding instead of hardware decoding, when the width is a multiple of 16 but not also a multiple of 32. According to @mfleisz this does not happen with hardware decoding and he could reproduce the error with Ogon, what I reproduced there with g-r-d.

The height here does not need to be aligned to a multiple of 32 instead of 16. However, simply aligning the width to a multiple of 16 **should be wrong** from my testing that I did back then. The problem here is that the position of the data from the auxiliary view is not applied correctly.
FreeRDP seems seems to do the position mapping based on the alignedWidth, where the alignedWidth is a multiple of 16.

But instead of doing the position mapping of the auxiliary data based on the alignedWidth, it should be I think based on the width of the AVC stream here (which itself already must be a multiple of 16).
When aligning the width in g-r-d to 32 instead of 16, `msrdc` always works fine. FreeRDP-based clients however have the same shift in colour data (like in the picture of this MR description), when the aligned value of the width for the alignment of 16 and 32 differ.

The shift in colour data, likely happens, due to "cut" between the `U` and `V` values in the auxiliary view: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegfx/781406c3-5e24-4f2b-b6ff-42b76bf64f6d

libfreerdp/codec/yuv.c
```diff
@@ -479,9 +479,8 @@ static void CALLBACK yuv444_combine_work_callback(PTP_CALLBACK_INSTANCE instance
 	const RECTANGLE_16* rect = &param->rect;
 	WINPR_ASSERT(rect);
 
-	const UINT32 alignedWidth = yuv->width + ((yuv->width % 16 != 0) ? 16 - yuv->width % 16 : 0);
-	const UINT32 alignedHeight =
-	    yuv->height + ((yuv->height % 16 != 0) ? 16 - yuv->height % 16 : 0);
+	const UINT32 alignedWidth = (yuv->width + 31) & ~31;
+	const UINT32 alignedHeight = (yuv->height + 31) & ~31;
```
pnowack: The alignment for the height should not be needed.

pnowack:
I think the alignment of (just) the width to 32 is alright. However, I think there should be another change, that the "cut" between the `U` data and the `V` data in the auxiliary view (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegfx/781406c3-5e24-4f2b-b6ff-42b76bf64f6d) is not hardcoded based on the aligned width value. Instead the "cut" should be based on the width of the AVC stream.

libfreerdp/codec/yuv.c
eduar-hte: Should it be at least aligned to 16 (as before), as the combination is done on 16x16 macroblocks, right?

I don't think this code should depend on the initialization value being already aligned (which I mentioned is the case because the surface is already 16-byte aligned)

pnowack: > Should it be at least aligned to 16 (as before), as the combination is done on 16x16 macroblocks, right?

Yes, the old code line (just) for the height was fine, where the alignment of 16 was used.

eduar-hte: ok, I'll update the pull request to reflect this.

NOTE: I thought about updating the line to simplify the alignment code with the approach I used initially, but I'd rather minimize the changes in the pull request (so I'll go back to the previous code).

eduar-hte:
> I think the alignment of (just) the width to 32 is alright. However, I think there should be another change, that the "cut" between the `U` data and the `V` data in the auxiliary view (https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegfx/781406c3-5e24-4f2b-b6ff-42b76bf64f6d) is not hardcoded based on the aligned width value. Instead the "cut" should be based on the width of the AVC stream.

Are you referring to the `nTotalWidth` argument to `general_ChromaV2ToYUV444` as well? By the width of the AVC stream do you mean the stride value? If that's the case, I proposed a similar approach at the end of my initial report, where I suggested using `iStride[0]`. However, I'm not sure what accounts for the `64` delta in the received data, and which if not detracted would generate issues as well.

libfreerdp/codec/yuv.c
eduar-hte: BTW The aligned height is actually unused afterwards, as it's only passed to the `general_ChromaV2ToYUV444` which doesn't reference it (the height for the combination is based on the updated `RECTANGLE_16` instances).

pnowack:
> Are you referring to the nTotalWidth argument to general_ChromaV2ToYUV444 as well? By the width of the AVC stream do you mean the stride value?

No, the GFX surface size is the size of the content, that is shown on the screen. But the video size of the AVC stream can be larger than that (already due to the 16x16 alignment (and now with the 32x16 alignment)). So it has its own width and height. And the split of the `U` and `V` data in https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpegfx/781406c3-5e24-4f2b-b6ff-42b76bf64f6d is based on that video streams width (not the GFX surface one, where the eventual content later lands).

I had to find some time to be able to look where we can retrieve these values from the decoder. The decoder itself gets these values from the SPS NAL unit (16 * the `pic_width_in_mbs_minus1` value for the width, and 16 * the `pic_height_in_map_units_minus1` value for the height). But it looks like that we never actually try to retrieve these values, which is unfortunate.

Implementing that handling for all four H264 backends (MediaCodec, MediaFoundation, OpenH264, FFMPEG) would be the correct way. However, for this quick bugfix, this seems like a huge task. So I retract that call for changing that in this PR. I guess this will take some time until Microsoft adjusts the alignment for AVX512 (64-Byte alignment).

I will do a test later today with g-r-d. Aside from that, this now looks fine to me.

pnowack:
@freerdp-bot Can you do a test too?

pnowack:
> I will do a test later today with g-r-d. Aside from that, this now looks fine to me.

Did a quick test with g-r-d, where I observed the issue too when using xfreerdp. This MR fixes it there too.

akallabeth:
@freerdp-bot test

PR #11941, opened 2025-10-17
Fix build and run with optional channels

blino:
This MRs addresses several build issues when specific channels are disabled, and also an issue at runtime to avoid loading channels disabled at build time.

cc @hwti

### DIFF
--- a/channels/server/channels.c
+++ b/channels/server/channels.c
@@ -41,20 +41,36 @@
 #if defined(CHANNEL_AUDIN_SERVER)
 #include <freerdp/server/audin.h>
 #endif
+#if defined(CHANNEL_RDPSND_SERVER)
 #include <freerdp/server/rdpsnd.h>
+#endif
+#if defined(CHANNEL_CLIPRDR_SERVER)
 #include <freerdp/server/cliprdr.h>
+#endif
+#if defined(CHANNEL_ECHO_SERVER)
 #include <freerdp/server/echo.h>
+#endif
+#if defined(CHANNEL_RDPDR_SERVER)
 #include <freerdp/server/rdpdr.h>
+#endif
 #if defined(CHANNEL_RDPEI_SERVER)
 #include <freerdp/server/rdpei.h>
 #endif
+#if defined(CHANNEL_DRDYNVC_SERVER)
 #include <freerdp/server/drdynvc.h>
+#endif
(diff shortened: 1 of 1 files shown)


client/common/client.c
```diff
@@ -2248,13 +2249,19 @@ BOOL freerdp_client_handle_pen(rdpClientContext* cctx, UINT32 flags, INT32 devic
 	}
 
 	WLog_WARN(TAG, "Invalid pen %" PRId32 " flags 0x%08" PRIx32, deviceid, flags);
+#else
+	WLog_WARN(TAG, "Pen event detected but RDPEI support not compiled in. Recompile with "
+	               "-DWITH_CHANNELS=ON");
+#endif
```
hardening, line 2255: Perhaps we should be more specific (`CHANNEL_RDPEI_CLIENT`) as it seems like `-DWITH_CHANNELS=ON` is not enough

client/common/client.c
```diff
@@ -2271,6 +2278,11 @@ BOOL freerdp_client_pen_cancel_all(rdpClientContext* cctx)
 		}
 	}
 	return TRUE;
+#else
+	WLog_WARN(TAG, "Pen event detected but RDPEI support not compiled in. Recompile with "
+	               "-DWITH_CHANNELS=ON");
```
hardening: same here

hardening:
@freerdp-bot can you do a test build to see if it breaks something ?

client/common/client.c
blino: This has been copied to mimic the message already used in the following functions:
- `freerdp_handle_touch_up()`
- `freerdp_handle_touch_down()`
- `freerdp_handle_touch_motion()`

Should I fix all of them?

akallabeth:
@blino could you rebase to current `master` so the ci passes?
ok from  me as well, thank you!

client/common/client.c
akallabeth: @blino good catch, yes please.

blino: Ok, done in the same commit, let me know if you prefer to split.
Thanks for the feedback!

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #11948, opened 2025-10-20
[core,timer] ensure all scheduled timers are handled

akallabeth:
Timers might be delayed, so loop over all expired timers until none are left.


pnowack:
This indeed fixes dynamically resizing the xfreerdp3 window here. With these changes, xfreerdp3 now sends new monitor sizes again when resizing the xfreerdp3 window (and not just for the first resize).

pnowack (approved):
LGTM

PR #11960, opened 2025-10-27
[crypto,x509] set comparison function for sk_ASN1_OBJECT_find in EKU check

thenextman:
The `x509_utils_check_eku` function was using sk_ASN1_OBJECT_find() without setting a comparison function, which relies on pointer equality rather than content comparison.

This worked with OpenSSL by coincidence because OBJ_nid2obj() returns pointers to a global singleton table, so comparing the object from X509_get_ext_d2i() against OBJ_nid2obj() happened to match. However, with LibreSSL, objects parsed from certificates may be separate allocations, causing the pointer comparison to fail even when the OID values are identical.

The fix adds an asn1_object_cmp() wrapper function and calls sk_ASN1_OBJECT_set_cmp_func() to enable proper content-based comparison using OBJ_cmp().

### DIFF
--- a/libfreerdp/crypto/x509_utils.c
+++ b/libfreerdp/crypto/x509_utils.c
@@ -568,6 +568,17 @@ char* x509_utils_get_issuer(const X509* xcert)
 	return issuer;
 }
 
+static int asn1_object_cmp(const ASN1_OBJECT* const* a, const ASN1_OBJECT* const* b)
+{
+	if (!a || !b)
+		return (a == b) ? 0 : (a ? 1 : -1);
+	
+	if (!*a || !*b)
+		return (*a == *b) ? 0 : (*a ? 1 : -1);
+
+	return OBJ_cmp(*a, *b);
+}
+
 BOOL x509_utils_check_eku(const X509* xcert, int nid)
 {
 	BOOL ret = FALSE;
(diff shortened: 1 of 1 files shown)


libfreerdp/crypto/x509_utils.c
```diff
@@ -568,6 +568,11 @@ char* x509_utils_get_issuer(const X509* xcert)
 	return issuer;
 }
 
+static int asn1_object_cmp(const ASN1_OBJECT* const* a, const ASN1_OBJECT* const* b)
```
akallabeth, line 571: not sure, can you expect all objects to be `!= NULL`?

thenextman: @akallabeth Practically no because `X509_get_ext_d2i` won't create NULL objects from valid certs. However in interest of the fact that the input is probably untrusted, I added NULL checks.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@thenextman could you run `git clang-format` on your changes?

akallabeth:
ok, taken your commit, reformatted and added to #11961 
thank you for the patch!

PR #11962, opened 2025-10-28
[winpr,synch] Fix starvation in pollset_poll caused by emscripten_sleep

mfleisz:
Recent changes in `pollset_poll` added yielding the process using `emscripten_sleep` so event handlers can have time to process data and set events.

However calling `emscripten_sleep` on every iteration when checking the pollset caused process starvation which slowed down the whole session massively.

This commit fixes this issue by checking for signaled events 10 times before actually causing a yield. This avoids possible deadlocks while not slowing down the whole session pipeline.


MartinHaimberger (approved):
LGTM

PR #12005, opened 2025-11-18
fix line clipping issue

kov-serg:
Fix incorrect line drawing clipping

before fix:
freerdp-before

after fix:
freerdp-after

This fix remove nonsense clipping from gdi line drawing code.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #12011, opened 2025-11-18
Update command line usage for gateway option

alexiri:
Simple change to update the help text to show the correct, non-dreprecated syntax for `/gateway`.


akallabeth:
@alexiri nice catch!
@freerdp-bot test

PR #12031, opened 2025-12-01
Fix remote credential guard channel loading

hardening:
This patch fixes the loading of the RCG channel. It also adds a guard so the `/remoteGuard` cmdline doesn't work if the channel is not built in.

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -4860,6 +4860,7 @@ static int parse_command_line(rdpSettings* settings, const COMMAND_LINE_ARGUMENT
 			if (!freerdp_settings_set_bool(settings, FreeRDP_RestrictedAdminModeRequired, enable))
 				return fail_at(arg, COMMAND_LINE_ERROR);
 		}
+#ifndef CHANNEL_RDPEAR_CLIENT
 		CommandLineSwitchCase(arg, "remoteGuard")
 		{
 			if (!freerdp_settings_set_bool(settings, FreeRDP_RemoteCredentialGuard, TRUE))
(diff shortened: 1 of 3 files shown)


client/common/cmdline.c
```diff
@@ -4862,6 +4862,10 @@ static int parse_command_line(rdpSettings* settings, const COMMAND_LINE_ARGUMENT
 		}
 		CommandLineSwitchCase(arg, "remoteGuard")
 		{
+#ifndef CHANNEL_RDPEAR_CLIENT
+			WLog_ERR(TAG, "CHANNEL_RDPEAR_CLIENT not compiled in, remoteGuard can not work");
+			return fail_at(arg, COMMAND_LINE_ERROR);
+#endif
```
akallabeth: should be else (otherwise the following is dead code)

akallabeth: maybe do like with the h264 options, do not compile the command line option at all?

hardening: hum yes makes sense

client/common/cmdline.c
```diff
@@ -4860,13 +4860,15 @@ static int parse_command_line(rdpSettings* settings, const COMMAND_LINE_ARGUMENT
 			if (!freerdp_settings_set_bool(settings, FreeRDP_RestrictedAdminModeRequired, enable))
 				return fail_at(arg, COMMAND_LINE_ERROR);
 		}
+#ifndef CHANNEL_RDPEAR_CLIENT
```
akallabeth, line 4863: you forgot the option in the header `cmdline.h`.
if it stays there it will be shown in the help

PR #12032, opened 2025-12-02
Fix inverted ifdef

hardening:


akallabeth:
oops, did miss that one too :/

PR #12039, opened 2025-12-04
rdpear: handle basic NTLM commands and fix server-side

hardening:
This patch adds the handling of basic NTLM commands. Because there's some mysterious 4 zero bytes after pickle header in Kerberos packets, not present in NTLM commands, the patch also had to rework a bit the packet parsing / forging.

The patch also address a server-side bug when parsing supplemental creds, if the client was sending an empty list, we were considering this as an error.

And finally we also implement the parsing of MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL. This breaks the public API, anyway this was basically unused (as not parsed before) and the previous API was wrong as what we receive is MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL not MSV1_0_SUPPLEMENTAL_CREDENTIAL, so I guess the API breakage is ok.

### DIFF
--- a/channels/rdpear/common/rdpear-common/rdpear_common.h
+++ b/channels/rdpear/common/rdpear-common/rdpear_common.h
@@ -79,8 +79,9 @@ typedef enum
 	// End NTLM remote calls
 } RemoteGuardCallId;
 
-FREERDP_LOCAL RdpEarPackageType rdpear_packageType_from_name(WinPrAsn1_OctetString* package);
-FREERDP_LOCAL wStream* rdpear_encodePayload(RdpEarPackageType packageType, wStream* payload);
+FREERDP_LOCAL RdpEarPackageType rdpear_packageType_from_name(const WinPrAsn1_OctetString* package);
+WINPR_ATTR_MALLOC(Stream_Free, 1)
+FREERDP_LOCAL wStream* rdpear_encodePayload(BOOL isKerb, wStream* payload);
 
 #define RDPEAR_COMMON_MESSAGE_DECL(V)                                                            \
 	FREERDP_LOCAL BOOL ndr_read_##V(NdrContext* context, wStream* s, const void* hints, V* obj); \
--- a/include/freerdp/peer.h
+++ b/include/freerdp/peer.h
@@ -56,8 +56,18 @@ extern "C"
 	typedef BOOL (*psPeerHasMoreToRead)(freerdp_peer* peer);
 	typedef BOOL (*psPeerClose)(freerdp_peer* peer);
 	typedef void (*psPeerDisconnect)(freerdp_peer* peer);
+
+	/** callback called when we receive remote credential guard credentials during NLA
+	 * @param peer the associated freerdp_peer
+	 * @param logonCreds the KERB_TICKET_LOGON containing the TGT and the host service ticket
+	 * @param suppCreds some MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL containing NTLM hashes
+	 * @return if the treatment was successful
+	 * @bug before 3.19.0 suppCreds were a pointer to MSV1_0_SUPPLEMENTAL_CREDENTIAL, not
+	 * 		MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL as now
+	 */
 	typedef BOOL (*psPeerRemoteCredentials)(freerdp_peer* peer, KERB_TICKET_LOGON* logonCreds,
-	                                        MSV1_0_SUPPLEMENTAL_CREDENTIAL* suppCreds);
+	                                        MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL* suppCreds);
+
 	typedef BOOL (*psPeerCapabilities)(freerdp_peer* peer);
 	typedef BOOL (*psPeerPostConnect)(freerdp_peer* peer);
 	typedef BOOL (*psPeerActivate)(freerdp_peer* peer);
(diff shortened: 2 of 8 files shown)


libfreerdp/core/nla.c
```diff
@@ -1196,6 +1196,39 @@ static BOOL nla_read_KERB_TICKET_LOGON(WINPR_ATTR_UNUSED rdpNla* nla, wStream* s
 	return TRUE;
 }
 
+static MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL* nla_read_NtlmCreds(WINPR_ATTR_UNUSED rdpNla* nla,
```
akallabeth, line 1200: mark with `WINPR_ATTR_MALLOC`

libfreerdp/core/nla.c
```diff
@@ -1196,6 +1196,39 @@ static BOOL nla_read_KERB_TICKET_LOGON(WINPR_ATTR_UNUSED rdpNla* nla, wStream* s
...
+                                                                 wStream* s)
+{
+	WINPR_ASSERT(nla);
+	WINPR_ASSERT(s);
+
+	if (!Stream_CheckAndLogRequiredLength(TAG, s, 32 + 4))
+		return FALSE;
```
akallabeth: NULL?

libfreerdp/core/nla.c
```diff
@@ -1196,6 +1196,39 @@ static BOOL nla_read_KERB_TICKET_LOGON(WINPR_ATTR_UNUSED rdpNla* nla, wStream* s
...
+
+	size_t pos = Stream_GetPosition(s);
+	Stream_Seek(s, 32);
+
+	ULONG EncryptedCredsSize = Stream_Get_UINT32(s);
+	if (!Stream_CheckAndLogRequiredLength(TAG, s, EncryptedCredsSize))
+		return FALSE;
```
akallabeth: NULL?

libfreerdp/core/nla.c
```diff
@@ -1486,6 +1526,41 @@ static BOOL nla_write_TSRemoteGuardKerbCred(rdpNla* nla, WinPrAsn1Encoder* enc)
 	return ret;
 }
 
+static BOOL nla_write_TSRemoteGuardNtlmCred(rdpNla* nla, WinPrAsn1Encoder* enc,
+                                            const MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL* pntlm)
+{
+	BOOL ret = FALSE;
+	char ntlm[] = { 'N', '\0', 'T', '\0', 'L', '\0', 'M', '\0' };
```
akallabeth: why not `const BYTE`, avoids cast below?

winpr/include/winpr/secapi.h
```diff
@@ -70,6 +70,37 @@ typedef struct
...
+#define MSV1_0_CRED_REMOVED 0x4
+#define MSV1_0_CRED_CREDKEY_PRESENT 0x8
+#define MSV1_0_CRED_SHA_PRESENT 0x10
+
+typedef struct _MSV1_0_CREDENTIAL_KEY
+{
+	UCHAR Data[MSV1_0_CREDENTIAL_KEY_LENGTH];
+} MSV1_0_CREDENTIAL_KEY, *PMSV1_0_CREDENTIAL_KEY;
+
+typedef struct _MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL
```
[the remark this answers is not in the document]
hardening: Pff, this syntax is used in all windows headers, this warning is a pain in the *** !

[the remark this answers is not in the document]
akallabeth: well, windows headers are a pain, true :grin:

include/freerdp/peer.h
```diff
@@ -57,7 +57,7 @@ extern "C"
 	typedef BOOL (*psPeerClose)(freerdp_peer* peer);
 	typedef void (*psPeerDisconnect)(freerdp_peer* peer);
 	typedef BOOL (*psPeerRemoteCredentials)(freerdp_peer* peer, KERB_TICKET_LOGON* logonCreds,
-	                                        MSV1_0_SUPPLEMENTAL_CREDENTIAL* suppCreds);
+	                                        MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL* suppCreds);
```
akallabeth, line 69: Is this change avoidable?
maybe add a `@bug` doxygen to explain this change as this is public `API`

hardening: No it has be here, I made a mistake when writing the API (well it's easy to confuse `MSV1_0_SUPPLEMENTAL_CREDENTIAL` and `MSV1_0_REMOTE_SUPPLEMENTAL_CREDENTIAL`)
I'll add a doxygen comment to explain and give the version where it happens.

channels/rdpear/common/rdpear-common/rdpear_common.h
```diff
@@ -79,8 +79,8 @@ typedef enum
 	// End NTLM remote calls
 } RemoteGuardCallId;
 
-FREERDP_LOCAL RdpEarPackageType rdpear_packageType_from_name(WinPrAsn1_OctetString* package);
-FREERDP_LOCAL wStream* rdpear_encodePayload(RdpEarPackageType packageType, wStream* payload);
+FREERDP_LOCAL RdpEarPackageType rdpear_packageType_from_name(const WinPrAsn1_OctetString* package);
+FREERDP_LOCAL wStream* rdpear_encodePayload(BOOL isKerb, wStream* payload);
```
akallabeth, line 84: `WINPR_ATTR_MALLOC`

akallabeth (approved):
LGTM

Issue #11990, opened 2025-11-11
xfreerdp /dvc:rdpecam sends unrequested samples to the server side


pnowack:
Version: 3.17.2
Distribution: Archlinux

While working on RDPECAM support in g-r-d (server side), I noticed a bug with the camera implementation on the FreeRDP client side, where `xfreerdp3 /dvc:rdpecam` sends `Sample Response`s for samples, that were not requested.

Regarding reproduction, the handling for RDPECAM support in g-r-d is currently not yet part of a g-r-d release, however I can test patches for the client side.

In any case, the way I reproduced it is the following:

1. Connect with `xfreerdp3 /dvc:rdpecam` to g-r-d (compiled from my RDPECAM branch)
2. Open GNOMEs `Snapshot` application

Observation: A camera stream is now shown

3. Minimize `Snapshot`

After minimizing `Snapshot`, `Snapshot` won't request any samples anymore (because the window is not shown) and stops the PipeWire stream. Thus g-r-d too won't request any samples and queues a `Streams stop request`. g-r-d waits for the pending requested samples and then send the `Stop Streams request`.
But despite g-r-d not requesting any further samples, it still gets samples from `xfreerdp3` until the `Sucess response` is received for the `Streams Stop request`.
Currently, g-r-d handles this fatally by closing the DVC for the camera due to protocol violation (maybe a bit harsh, but that's how I notice such errors).

**What should happen instead?**

Instead of just pushing samples to the server side, only encode and send samples that were requested by the server side. The client side can already fetch a frame from the local camera, but it should only encode and send it, when the server asks for it, like the protocol documents in `2.2.3.14 Sample Response`, quoting:

`The Sample Response message is sent by the client over a device channel when a new video sample is available and there are outstanding Sample Requests`.

In this case here the condition `outstanding Sample Requests` is not fulfilled.

PR #12041, opened 2025-12-04
rdpecam: fix camera sample grabbing

hardening:
Before this patch we had a behavior where there was a credit of 8 samples that could be sent to the server with no corresponding sample request. So in the right conditions, we were having situations where the server was receiving samples that it has not requested, and so it was dropping them. The visible effect was small artifacts in the camera stream when i-frames where dropped, and more serious ones when the dropped content was containing key frames.

This issue has also been reported when xfreerdp connects on g-r-d as #11990.

This patch reworks the frame grabbing workflow: when the frame grabbing thread calls the sample callback we check if a sample is already pending, waiting to be sent to the server. If that's the case and the camera's input format supports frame dropping we just refresh the pending frame with the new one. If the input format can't drop frames (like with h264 and mjpg) we wait until the current pending frame is sent. So now frames can be sent either when we receive a sample request from the server, or when the sample callback is invoked.

### DIFF
--- a/channels/rdpecam/client/camera.h
+++ b/channels/rdpecam/client/camera.h
@@ -101,7 +101,10 @@ typedef struct
 	CAM_MEDIA_TYPE_DESCRIPTION currMediaType;
 
 	GENERIC_CHANNEL_CALLBACK* hSampleReqChannel;
-	INT nSampleCredits;
+	CRITICAL_SECTION lock;
+	volatile LONG samplesRequested;
+	wStream* pendingSample;
+	volatile BOOL haveSample;
 	wStream* sampleRespBuffer;
 
 	H264_CONTEXT* h264;
--- a/channels/rdpecam/client/camera_device_main.c
+++ b/channels/rdpecam/client/camera_device_main.c
@@ -19,6 +19,7 @@
 
 #include <winpr/assert.h>
 #include <winpr/cast.h>
+#include <winpr/interlocked.h>
 
 #include "camera.h"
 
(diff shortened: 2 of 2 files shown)


hardening:
@freerdp-bot seems like you have badly done your test, retry please

channels/rdpecam/client/camera.h
```diff
@@ -101,7 +101,10 @@ typedef struct
 	CAM_MEDIA_TYPE_DESCRIPTION currMediaType;
 
 	GENERIC_CHANNEL_CALLBACK* hSampleReqChannel;
-	INT nSampleCredits;
+	CRITICAL_SECTION lock;
+	volatile BOOL sampleRequested;
```
pnowack: This should be a queue instead of a boolean. If for whatever reason the server sends two SampleRequests in a short(er) interval (could also be due to some sudden network delay, so does not necessarily have to be a server issue), then only one of them is answered instead of both.
In such case, the second request will then later be handled when we get the frame from the camera device, but that should be fine and still protocol-conform.

hardening: Well in this case, no need for a queue, I will just set an integer for `sampleRequested`, increment it on sample requests and decrement it on sample sent

hardening: I've done that

channels/rdpecam/client/camera_device_main.c
```diff
@@ -352,14 +443,19 @@ static UINT ecam_dev_process_sample_request(CameraDevice* dev, GENERIC_CHANNEL_C
...
+
 	/* need to save channel because responses are asynchronous and coming from capture thread */
 	if (stream->hSampleReqChannel != hchannel)
 		stream->hSampleReqChannel = hchannel;
 
-	/* allow to send that many unsolicited samples */
-	stream->nSampleCredits = ECAM_MAX_SAMPLE_CREDITS;
+	UINT ret = CHANNEL_RC_OK;
+	stream->samplesRequested++;
+	if (stream->pendingSample)
```
earias-cybele: I wonder if this should check `stream->haveSample` instead. I have noticed that if there's a sample request before the camera has provided a sample, the `stream->pendingSample` has the `length` set during initialization (which is equal to `capacity`, 8294400).

This triggers an assert in `ecam_dev_send_sample_response` when executing:

`	Stream_Write(stream->sampleRespBuffer, sample, size);
`

PR #12043, opened 2025-12-05
Cleanups release

akallabeth:
* split `nla_encode_ts_credentials`
* add log messages in `rdpecam` for dropped samples.

### DIFF
--- a/channels/rdpecam/client/camera_device_main.c
+++ b/channels/rdpecam/client/camera_device_main.c
@@ -351,7 +351,7 @@ static UINT ecam_dev_process_start_streams_request(CameraDevice* dev,
 		return ERROR_INVALID_DATA;
 	}
 
-	stream->pendingSample = Stream_New(NULL, mediaType.Width * mediaType.Height * 4ull);
+	stream->pendingSample = Stream_New(NULL, 4ull * mediaType.Width * mediaType.Height);
 	if (!stream->pendingSample)
 	{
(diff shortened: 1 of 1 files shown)


channels/rdpecam/client/camera_device_main.c
```diff
@@ -123,13 +123,20 @@ static BOOL mediaSupportDrops(CAM_MEDIA_FORMAT format)
 
 static UINT ecam_dev_send_pending(CameraDevice* dev, int streamIndex, CameraDeviceStream* stream)
 {
-	BYTE* encodedSample = NULL;
-	size_t encodedSize = 0;
+	WINPR_ASSERT(dev);
+	WINPR_ASSERT(stream);
+
+	if (stream->samplesRequested <= 0)
```
hardening, line 129: This change is unfortunate because with the one in `ecam_dev_sample_captured_callback` it will flood the output with `Frame drop: No sample requested` while no frame is actually dropped, and we're in perfectly legitimate case. The frame grabbing thread grabs samples as they come from the camera, and we serve the last one when the server ask for some content.

hardening: Seems like this PR was merged too fast, and without external review :-(

akallabeth: @hardening ouch, my bad.
but I did run it, and while it prints quite a number of messages it is not that verbose (but maybe `WLOG_TRACE` would have been the better fit, true)

Issue #12048, opened 2025-12-06
echo pass | xfreerdp /from-stdin crashes


ileGITimo:
**Describe the bug**
As in the title running "echo pass | xfreerdp /from-stdin"  crashes with latest Fedora package 2:3.18.0-1.fc43 and a version just built from github master.  Running without the  pipe and entering the password when asked still works. Downgrading to Fedora 2:3.16.0-4.fc43 also works

**To Reproduce**
1. run like described above in title
2. xfreerdp crashes immediately with  "free(): double free detected in tcache 2"

**Expected behavior**
a xfreerdp connection : - )

**Application details**
* This is FreeRDP version 3.19.1-dev0 (b244d24)
* gdb /usr/local/bin/xfreerdp
* * r /from-stdin /t:TITLE /u:user /d:domain /network:lan /bpp:24 +clipboard /w:1920 /h:1080 /v:host:3389 /cert:ignore /sec:tls /sound /microphone < /tmp/password
*  Output from a gdb run with backtrace [gdb-run.txt](https://github.com/user-attachments/files/23991026/gdb-run.txt)
* [buildconfig.txt](https://github.com/user-attachments/files/23991009/buildconfig.txt)
* OS version connecting to (server side): never connects
* Built against github:master, no special instructions, with CXXOPTS=-g in command line

**Environment (please complete the following information):**
 - OS:  Linux
 - Version/Distribution: Fedora 43
 - Architecture:  x85_64

ileGITimo:
User and domain are given with /u:user and /d:domain, pasword will come later from stdin.

When client_cli_read_string is called to read username, it prints out the prompt "Username:        ", prints out the given user name, then waits for a carriage return from stdin, but stdin is the file with the pasword. Code proceeds to read the whole password and returns suggestion concatenated with the password.  This messes up something.

When then cli_read_sting is called to read domain it crashes with a bad free.
Valgrind report here [valgrind.txt](https://github.com/user-attachments/files/23996566/valgrind.txt)
Will try to further debug this later.

ileGITimo:
If instead of using "echo pass | xfreerdp" one uses "xfreerdp ... < passwd_file" and the passwd_file has two blank lines and the password in the third line, it works ...

PR #12049, opened 2025-12-06
Don't call freerdp_interruptible_get_line for user, domain or passwor…

ileGITimo:
May fix issue #12048, though not all cases were tested


akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/12050 but thx for the effort!

ileGITimo:
I did say "may fix" and  "not all cases were tested: : - )

PR #12059, opened 2025-12-10
[channels,rdpecam] allow MJPEG frame skip and direct passthrough

akallabeth:


hardening:
Can you also remove the erronous log in `ecam_dev_send_pending` ?

channels/rdpecam/client/camera_device_main.c
```diff
@@ -36,6 +36,7 @@ static const CAM_MEDIA_FORMAT_INFO supportedFormats[] = {
 #endif
 #if defined(WITH_INPUT_FORMAT_MJPG)
 	{ CAM_MEDIA_FORMAT_MJPG, CAM_MEDIA_FORMAT_H264 },
+	{ CAM_MEDIA_FORMAT_MJPG, CAM_MEDIA_FORMAT_MJPG },
```
hardening, line 39: I've tested mjpg passthrough, it fixes the problem with the first invalid frame (seems like server side they are robust to that).
Anyway if we enable that passthrough, we can entirely remove the code that does mjpg to h264 in `encoding.c` as it will never be invoked. BTW if the goal is mjpg passthrough, you should remove the line above (MJPG to H264), as otherwise it will be selected all the time.

Some more general thoughts; in the rdpecam there's basically 2 bottlenecks: acquiring data from the webcam and encode/push them to the server. 
If we have a low end device, we will possibly have low USB bandwidth and low CPU encoding possibility, that makes sense to enable `WITH_INPUT_FORMAT_MJPG`, as it will reduce the traffic on the USB port, and as it will be passthrough, no CPU spent doing the encoding. But the downside will be the the network bandwidth (mjpg is not that great).
So if we have a more powerful host, then we'd better desactivate mjpg passthrough, acquire data in YUY2 format as we may have a good USB bandwidth. Encode them in H264 as we have CPU power, and so reduce network bandwidth.
My remark is, shall we add a cmake option for mjpeg passthrough in the hybrid case where we want to acquire data from the webcam as mjpg (to save USB bandwidth) and transcode it to h264 (to save network bandwidth at the cost of CPU cycles) ?

akallabeth: @hardening well, we could add such an option.
but for the `MJPEG` to `H264` to work the other change (allow dropping of frames) is enough to make it work.

akallabeth: note: with my test the `mjpeg` is always reencoded to `h264` (remote is a windows 10)

hardening: That's what I was saying above, if you keep the line `{ CAM_MEDIA_FORMAT_MJPG, CAM_MEDIA_FORMAT_H264 },` the input format is matched and you endup with transcoding to h264, so not passthrough...

akallabeth: well, depends a lot on the server side. (that is selecting the desired mode from the available formats)
given, current algorithm there is quite primitive, but that can change with every update.

akallabeth:
> `ecam_dev_send_pending`

did change to `WLog_VRB` so this should no longer spam normal debug logs.

hardening:
> > `ecam_dev_send_pending`
> 
> did change to `WLog_VRB` so this should no longer spam normal debug logs.

The message must be removed, it is wrong: no frame is dropped, it's the standard case where the grabbing thread gets a sample ahead of the sample request so that we have the best latency (when we're requested for a sample, if we have it we can serve it instantly).

akallabeth:
> > > `ecam_dev_send_pending`
> > 
> > 
> > did change to `WLog_VRB` so this should no longer spam normal debug logs.
> 
> The message must be removed, it is wrong: no frame is dropped, it's the standard case where the grabbing thread gets a sample ahead of the sample request so that we have the best latency (when we're requested for a sample, if we have it we can serve it instantly).

ok, rephrasing it.

akallabeth:
@hardening ok, refactored the whole thing.
now the logging is more complete and frames that fail to encode are dropped.

channels/rdpecam/client/camera_device_main.c
```diff
@@ -139,12 +139,16 @@ static UINT ecam_dev_send_pending(CameraDevice* dev, int streamIndex, CameraDevi
...
 		{
-			WLog_DBG(TAG, "Frame drop or error in ecam_encoder_compress");
+			WLog_DBG(TAG, "Frame dropped: error in ecam_encoder_compress");
+			stream->haveSample = FALSE;
 			return CHANNEL_RC_OK;
 		}
 
 		if (!stream->streaming)
+		{
+			WLog_DBG(TAG, "Frame delayed: stream not running");
```
hardening: Hum it's more likely that it's because the stream is being stopped ?

akallabeth: well, then it is not running, so the message is correct?

hardening: What I meant is that the frame is not gonna be delayed, it's gonna be dropped as the stream is stopping. Nitpicking...

akallabeth:
@hardening anything remaining to fix?

channels/rdpecam/client/camera_device_main.c
akallabeth: ok, so dropped later on :grin:

PR #12103, opened 2026-01-07
[client,sdl] lock primary while used

akallabeth:


mdeslaur:
Shouldn't this issue also apply to client/SDL/SDL2/sdl_freerdp.cpp?

PR #12108, opened 2026-01-08
[server,proxy] make peer_list access thread-safe and fix leaks

LeeEirc:


akallabeth (changes requested):
remove the locks, just set `server->peer_list = ArrayList_New(TRUE);` instead.
:+1:  for the leak fix.

PR #12111, opened 2026-01-08
[enum,cast] fix implicit enum casts

akallabeth:
* check input value range if read from network
* cast properly after value range is checked
* fix wrong return values

### DIFF
--- a/libfreerdp/codec/dsp.c
+++ b/libfreerdp/codec/dsp.c
@@ -393,20 +393,20 @@ static BOOL freerdp_dsp_decode_ima_adpcm(FREERDP_DSP_CONTEXT* WINPR_RESTRICT con
 			{
 				BYTE* dst = Stream_Pointer(out);
 
-				const int channel = (i < 4 ? 0 : 1);
+				const unsigned channel = (i < 4 ? 0 : 1);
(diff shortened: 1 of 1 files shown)


hardening:
Seems like there's a lot of unneeded extra blocks

channels/audin/server/audin.c
```diff
@@ -81,7 +81,25 @@ static UINT audin_server_recv_version(audin_server_context* context, wStream* s,
 	if (!Stream_CheckAndLogRequiredLengthWLog(audin->log, s, 4))
 		return ERROR_NO_DATA;
 
-	Stream_Read_UINT32(s, pdu.Version);
+	{
```
hardening, line 84: extra block not needed (here and in many other places)

akallabeth: it is helpful though (scoping of variables)

hardening: I can't see any benefit with the scoping here and in most of the other places, but perhaps I'm missing something...

akallabeth: `scoping`? 
prevents accidental (re)use of a variable if out of scope

PR #10491, opened 2024-08-20
[core,tcp] retry all DNS entries until success

akallabeth:
When resolving some DNS name with getaddrinfo try all returned entries matching the configuration (IPv4|IPv6|any) until the socket can be connected. Fixes #5335

Issue #5335, opened 2019-04-10
xfreerdp fails to connect when one host out of several is down


bduncan:
Hi,

We rebooted our three xrdp servers today. They are load balancing behind haproxy. DNS returns three IP addresses in response to an A query against (details anonymised) `xrdp.domain`:

```
:) bduncan@fry:~$ host xrdp
xrdp.domain has address 192.168.215.211
xrdp.domain has address 192.168.215.208
xrdp.domain has address 192.168.215.209
:) bduncan@fry:~$
```

Today, one of the servers got bricked and we had to rebuild it. My DNS cached the above result for a long time.

When I run `xfreerdp /v:xrdp.domain` I expect to get a connection to one of the remaining working hosts, even though one of them gives ECONNREFUSED on 3389.

Instead, I get this error:

```
[16:16:41:077] [12220:12221] [INFO][com.freerdp.client.common.cmdline] - loading channelEx cliprdr
[16:16:41:078] [12220:12221] [INFO][com.freerdp.client.x11] - No user name set. - Using login name: bduncan
[16:16:41:084] [12220:12221] [ERROR][com.freerdp.core.transport] - BIO_should_retry returned a system error 32: Broken pipe
[16:16:41:084] [12220:12221] [ERROR][com.freerdp.core.nego] - Protocol Security Negotiation Failure
[16:16:41:084] [12220:12221] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED [0x0002000C]
[16:16:41:084] [12220:12221] [ERROR][com.freerdp.core.connection] - Error: protocol security negotiation or connection failure
```

(lots of messages about security negotiation despite the server returning a clear Connection Refused!)

I can connect to any of the remaining servers directly, but not using this DNS name.

It looks to me like xfreerdp is trying to connect to each IP address simultaneously and acts on the first response it receives. Since the failure is much easier, it arrives first, so the whole connection fails:

```
:) bduncan@fry:~$ strace -ffs 9999 -e connect xfreerdp /v:xrdp.domain +glyph-cache
connect(10, {sa_family=AF_UNIX, sun_path=@"/tmp/.X11-unix/X0"}, 20) = 0
strace: Process 12259 attached
[16:17:05:462] [12258:12259] [INFO][com.freerdp.client.common.cmdline] - loading channelEx cliprdr
[...]

PR #12112, opened 2026-01-08
Really try all DNS entries until success

ondrejholy:
The https://github.com/FreeRDP/FreeRDP/issues/5335 issue was closed over https://github.com/FreeRDP/FreeRDP/pull/10491, but the problem still persists. The commit 4286a4c tries to create a socket for all DNS entries until the socket is created, but it doesn't verify that it is actually possible to connect to that socket. This is my attempt to fix it. There are also fixes for some other issues. See commit descriptions for more details.

### DIFF
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -1077,6 +1077,53 @@ int freerdp_tcp_connect(rdpContext* context, const char* hostname, int port, DWO
 	return transport_tcp_connect(context->rdp->transport, hostname, port, timeout);
 }
 
+static struct addrinfo* reorder_addrinfo_by_preference(rdpContext* context, struct addrinfo* addr)
+{
+	WINPR_ASSERT(context);
+	WINPR_ASSERT(addr);
+
+	const BOOL preferIPv6 =
+	    freerdp_settings_get_bool(context->settings, FreeRDP_PreferIPv6OverIPv4);
+	if (!preferIPv6)
+		return addr;
+
+	struct addrinfo* ipv6Head = NULL;
+	struct addrinfo* ipv6Tail = NULL;
+	struct addrinfo* otherHead = NULL;
+	struct addrinfo* otherTail = NULL;
+
+	/* Partition the list into IPv6 and other addresses */
+	while (addr)
+	{
+		struct addrinfo* next = addr->ai_next;
+		addr->ai_next = NULL;
+
+		if (addr->ai_family == AF_INET6)
+		{
+			if (!ipv6Head)
+				ipv6Head = addr;
+			else
+				ipv6Tail->ai_next = addr;
+			ipv6Tail = addr;
+		}
+		else
+		{
+			if (!otherHead)
+				otherHead = addr;
+			else
+				otherTail->ai_next = addr;
+			otherTail = addr;
+		}
+		addr = next;
+	}
+
+	/* Concatenate the lists */
+	if (ipv6Tail)
+		ipv6Tail->ai_next = otherHead;
+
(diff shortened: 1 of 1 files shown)


libfreerdp/core/tcp.c
```diff
@@ -853,11 +853,17 @@ static BOOL freerdp_tcp_connect_timeout(rdpContext* context, int sockfd, struct
 	}
 
 	{
-		const SSIZE_T res = recv(sockfd, NULL, 0, 0);
-		if (res == SOCKET_ERROR)
+		UINT32 optval = 0;
+		socklen_t optlen = sizeof(optval);
+		if (getsockopt(sockfd, SOL_SOCKET, SO_ERROR, (void*)&optval, &optlen) < 0)
```
akallabeth, line 858: no cast required

akallabeth:
@freerdp-bot test

libfreerdp/core/tcp.c
```diff
@@ -1244,34 +1291,36 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
+					}
+				}
+
 				if (sockfd < 0)
 				{
 					const int lrc = get_next_addrinfo(context, addr->ai_next, &addr,
 					                                  FREERDP_ERROR_CONNECT_FAILED);
 					if (lrc < 0)
+					{
+						freerdp_set_last_error_if_not(context, FREERDP_ERROR_CONNECT_FAILED);
```
akallabeth, line 1317: leaking `result`

libfreerdp/core/tcp.c
```diff
@@ -1233,9 +1278,11 @@ int freerdp_tcp_default_connect(rdpContext* context, rdpSettings* settings, cons
...
-			 * If PreferIPv6OverIPv4 = TRUE we force to IPv6 if there
-			 * is such an address available, but fall back to first if not found
+			 * If PreferIPv6OverIPv4 = TRUE we reorder addresses by preference:
+			 * IPv6 addresses come first, then other addresses.
 			 */
+			result = reorder_addrinfo_by_preference(context, result);
+
 			const int rc =
 			    get_next_addrinfo(context, result, &addr, FREERDP_ERROR_DNS_NAME_NOT_FOUND);
 			if (rc < 0)
```
akallabeth, line 1288: not your code, but it looks like in this case `result` is leaking?

libfreerdp/core/tcp.c
ondrejholy: I've keep the original logic, where `freeaddrinfo(input);` is in the `get_next_addrinfo` function, though I don't think it is nice solution.

ondrejholy:
Yes it can be simply tested by adding stuff in `/etc/hosts`. Just be careful that `getaddrinfo` reorders the results in some cases, it e.g. prefers addresses to local interfaces. It is worth to check the order over `getent ahosts [host]` when testing.

libfreerdp/core/tcp.c
akallabeth: ok.
there is still a leak though, see https://ci.freerdp.com/job/linux/15645/label=debian32/console

libfreerdp/core/tcp.c
ondrejholy: I have use it as it is used on other places in this file, but can remove it.

libfreerdp/core/tcp.c
ondrejholy: Ok, I will take a look tomorrow.

akallabeth:
ok, created a new pr based on this one, will fix the leak there.
thank you for the fixes!

PR #12115, opened 2026-01-09
[client,common] fix /remoteGuard

akallabeth:
The #ifdef condition was inverted, fix it.


hardening:
I'm sorry it's my fault I had fixed that locally but forgot to push it.

hardening (approved):

PR #12133, opened 2026-01-13
[printer] Compute printer configuration file using sha256+base64

dev-ddoe:
When printer name are long, using only base64 will create a long file which can cause the issue ERROR_FILENAME_EXCED_RANGE on Windows.

### DIFF
--- a/channels/printer/client/printer_main.c
+++ b/channels/printer/client/printer_main.c
@@ -83,7 +83,10 @@ static char* get_printer_hash(const WCHAR* name, size_t length)
 	BYTE hash[WINPR_SHA256_DIGEST_LENGTH] = { 0 };
 
 	if (!winpr_Digest(WINPR_MD_SHA256, (void*)name, length, hash, sizeof(hash)))
-		return crypto_base64_encode((const BYTE*)name, length);
+	{
+		WLog_ERR(TAG, "get_printer_hash failed to create hash of printer name");
+		return NULL;
+	}
 
 	return crypto_base64_encode(hash, sizeof(hash));
 }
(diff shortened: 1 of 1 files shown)


dev-ddoe:
for example, a printer with the name `BLABLABLAou på BLABLABLA.BLABLABLA.BLABLABLA.BLA (BLABLABLABLA 2 )` does not work on Windows
I get the error: [2026-1-13 05:11:55:977] ERROR: devman_load_device_service failed!
GetLastError returned ERROR_FILENAME_EXCED_RANGE

channels/printer/client/printer_main.c
```diff
@@ -77,11 +77,22 @@ typedef enum
 static const char* filemap[] = { "PortDosName", "PnPName", "DriverName",
 	                             "CachedPrinterConfigData" };
 
+
+static char* get_printer_hash(const WCHAR* name, size_t length)
+{
+	BYTE hash[WINPR_SHA256_DIGEST_LENGTH] = { 0 };
+
+	if (!winpr_Digest(WINPR_MD_SHA256, (void*)name, length, hash, sizeof(hash)))
+		return crypto_base64_encode((const BYTE*)name, length);
```
akallabeth: better use a proper abort.
if this fails there is something seriously wrong on the system.

akallabeth: note: there is already `winpr_BinToHexString`

akallabeth: also: please squash the commits and force push, should be a single one.

dev-ddoe: For now, I used previous code as a fallback which was not checked.
I put a fixup as a proposal.

akallabeth:
ok, taking your commits and adding a few additional checks.
See #12135 for the final commit, thank you!

PR #12149, opened 2026-01-16
[codec,swscale] allow runtime loading of swscale

thenextman:
`[RDPECAM]` has hard dependencies on `swscale` and `avutil`. It can optionally use `avcodec` for the MJPEG codec. It can optionally use an H264 backend for H264 passthrough.

For Devolutions, this is problematic: we _only_ ship a static library for `freerdp3` and its dependencies.  A static link to any part of `ffmpeg` is simply not possible. However, we would like to ship `[RDPECAM]` on our Linux client.

The proposed solution adds an option alongside `WITH_SWSCALE`, `WITH_SWSCALE_LOADING`. If enabled, we compile the `codec.swscale` module and try to do a runtime load of both `swscale` and `avutil`. The module contains the necessary forward declarations to build. This allows the user to use camera redirection assuming they have ffmpeg installed and an H264 backend (in our case, that is OpenH264). MJPEG is not expected to work in this configuration.

I'm not sure if this is a path the maintainers want to take, but I've been testing on my side and found it to work well. It would make things significantly easier for us if this (_or something like it_) could be merged in the upstream project.


CMakeLists.txt
```diff
@@ -414,6 +414,10 @@ if(WITH_OPENH264 AND NOT WITH_OPENH264_LOADING)
   option(WITH_OPENH264_LOADING "Use LoadLibrary to load openh264 at runtime" OFF)
 endif(WITH_OPENH264 AND NOT WITH_OPENH264_LOADING)
 
+if(WITH_SWSCALE AND NOT WITH_SWSCALE_LOADING)
+  option(WITH_SWSCALE_LOADING "Use LoadLibrary to load swscale at runtime" OFF)
```
akallabeth, line 422: use https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

thenextman:
First, thanks for taking a look!

> Looks like this is not only about `SWSCALE` but also `FFMPEG/H264`?

Kind of. `swscale` and `avutil` are both part of ffmpeg, and they're required for using RDPECAM. But it's not all of ffmpeg, just those specific libraries.

RDPECAM can use H264 as the underlying codec and that requires a full H264 backend. Otherwise it can use mjpeg which requires the full ffmpeg.

> To get this upstream I'd suggest:
> 
> 1. Wrap all used functions in `winpr` or in `libfreerdp/utils` (not sure where this is better off) and replace the usage in the whole project with the wrapper functions (like was done with `JSON`)
> 2. do the appropriate linking/loading only there, avoid all the `#ifdef` stuff except where required (e.g. `CMake/Pkgconfig` related stuff)
> 3. link (or load) this only with winpr

Would you consider an interim approach? i.e. keep this stuff in libfreerdp but factor the code better so the `encoding` and `color` modules don't have the #ifdef stuff? Practically I don't really want to refactor the build system to link ffmpeg (or parts of it) in WinPR but pragmatically I'm also not convinced it's the proper place for that. Especially if we consider what's done for H264 currently; this is all codec related. But I do think the implementation could be cleaner and follow the model of the H264 codec more closely.

Let me know what you think.

akallabeth:
@thenextman well, now looking at it a bit more, it would most likely be best to add the encoding functions to `libfreerdp` just like we already have in `include/freerdp/codec/h264.h` or similar and wrap it up in there just like is done with all the other h264 related functions?

* `swscale` is already wrapped in `freerdp_image_scale` but I think in `rdpecam` it is used for YUV or similar color data. so maybe an additional wrapper function to support these formats? 
* then there is a `ffmpeg` based `mjpeg` encoder loop, might also just be added to codecs as generic function that is called
* there would need to be wrappers for the `new/free` functions for the required contexts as well

thenextman:
> @thenextman well, now looking at it a bit more, it would most likely be best to add the encoding functions to `libfreerdp` just like we already have in `include/freerdp/codec/h264.h` or similar and wrap it up in there just like is done with all the other h264 related functions?
> 
> * `swscale` is already wrapped in `freerdp_image_scale` but I think in `rdpecam` it is used for YUV or similar color data. so maybe an additional wrapper function to support these formats?
> * then there is a `ffmpeg` based `mjpeg` encoder loop, might also just be added to codecs as generic function that is called
> * there would need to be wrappers for the `new/free` functions for the required contexts as well

Sounds good, I will rework this next week in line with your suggestions. Do you prefer to rework this PR or close it and open a new one?

thenextman:
@akallabeth I've reworked these changes and ready for review again.

I'll re-summarize the problem and describe the changes.

`rdpecam` client has some dependencies:

|Dependency|Build Flag|Required|Notes|
| --- | --- | --- | --- |
|Video4Linux |N/A|Yes|Camera enumeration and capture|
|swscale, avutil|`WITH_SCSCALE`|Yes|Color space conversion YUV <-> RGB and scaling|
|avcodec|`WITH_FFMPEG`|Optional|MJPEG codec|
|H264|`WITH_FFMPEG`, `WITH_OPENH264`, etc|Optional|H264 codec|

So: the channel _requires_ some parts of ffmpeg (swscale, avutil) to do color space conversions and scaling. There are two codecs that might be supported, both are optional but obviously for a functional camera we need at least one. `ffmpeg` can provide both the MJPEG and H264 codecs. If we choose to ignore MJPEG, we can use OpenH264 instead and have just H264 codec.

At Devolutions we build a fully statically linked FreeRDP based client and linking to ffmpeg is not feasible due to the GPL. However we would like to, at least partially, support RDPECAM.

This PR allows the builder to specify an additional flag `WITH_SWSCALE_LOADING` alongside `WITH_SWSCALE`. This will load `swscale` and `avutil` at runtime if available. A new `FREERDP_VIDEO_CONTEXT` API wraps the functionality and supports both direct linking and runtime loading.

Existing workflows remain unaffected, but on our side we can use `WITH_SWSCALE=ON WITH_SWSCALE_LOADING=ON WITH_OPENH264=ON WITH_OPENH264_LOADING=ON` and support runtime loading of ffmpeg and OpenH264, giving `RDPECAM` support for the H264 codec assuming the user has the packages installed.

> Why is MJPEG decoder not runtime loaded?

 It would be nice to also support MJPEG here, but it's more challenging and the codec is less widely used with modern cameras. This could be a future improvement. The `swscale`/`avutil` functions we wrap are just stateless image processing. `avcodec` maintains internal codec state, has complex initialization and registration and possibly has issues with ABI stability between versions.

.claude/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "permissions": {
```
akallabeth: please remove this file

channels/rdpecam/client/camera.h
```diff
@@ -25,12 +25,8 @@
 #include <stdlib.h>
 #include <string.h>
 
-#if defined(WITH_INPUT_FORMAT_MJPG)
-#include <libavcodec/avcodec.h>
-#endif
-
-#include <libswscale/swscale.h>
-#include <libavutil/imgutils.h>
+/* Use FreeRDP video codec wrappers for MJPEG and color space conversion */
```
akallabeth: useless comment.

channels/rdpecam/client/CMakeLists.txt
```diff
@@ -17,13 +17,11 @@
 
 define_channel_client("rdpecam")
 
-if(NOT WITH_SWSCALE OR NOT WITH_FFMPEG)
-  message(FATAL_ERROR "WITH_FFMPEG and WITH_SWSCALE required for CHANNEL_RDPECAM_CLIENT")
+if(NOT WITH_SWSCALE)
```
akallabeth: useless check here, the core library might not use the same build flags.

channels/rdpecam/client/CMakeLists.txt
```diff
@@ -32,22 +30,40 @@ else()
...
 
+include(CMakeDependentOption)
+
 option(RDPECAM_INPUT_FORMAT_H264 "[MS-RDPECAM] Enable H264 camera format (passthrough)" ON)
 if(RDPECAM_INPUT_FORMAT_H264)
   add_compile_definitions("WITH_INPUT_FORMAT_H264")
 endif()
 
-option(RDPECAM_INPUT_FORMAT_MJPG "[MS-RDPECAM] Enable MJPG camera format" ON)
+if(NOT WITH_SWSCALE_LOADING)
```
akallabeth: should not be compile time, the core library might not be of the same build but use different flags.

include/freerdp/codec/video.h
```diff
@@ -0,0 +1,167 @@
...
+		VIDEO_FORMAT_ABGR = 27,
+		VIDEO_FORMAT_BGRA = 28,
+		VIDEO_FORMAT_YUV411P = 7,
+		VIDEO_FORMAT_YUVJ420P = 12,
+		VIDEO_FORMAT_YUVJ422P = 13,
+		VIDEO_FORMAT_YUVJ444P = 14,
+		VIDEO_FORMAT_YUVJ440P = 32,
+		VIDEO_FORMAT_YUVJ411P = 33,
+		VIDEO_FORMAT_YUV440P = 31,
+		VIDEO_FORMAT_RGB32 = 100, /* Will map to platform-specific BGRA/ARGB */
```
akallabeth: remove `,`

include/freerdp/codec/video.h
```diff
@@ -0,0 +1,167 @@
...
+	} FREERDP_VIDEO_FORMAT;
+
+	/**
+	 * @brief Video context options
+	 */
+	typedef enum
+	{
+		VIDEO_CONTEXT_OPTION_DECODER_TYPE, /** MJPEG decoder type */
+		VIDEO_CONTEXT_OPTION_SCALING_QUALITY, /** Scaling algorithm quality */
```
akallabeth: remove `,`

libfreerdp/codec/color.c
```diff
@@ -1239,16 +1243,35 @@ BOOL freerdp_image_scale(BYTE* WINPR_RESTRICT pDstData, DWORD DstFormat, UINT32
 		if ((srcFormat == AV_PIX_FMT_NONE) || (dstFormat == AV_PIX_FMT_NONE))
 			return FALSE;
 
+#if defined(WITH_SWSCALE_LOADING)
```
akallabeth: just always use the `freerdp_sws*` wrappers? should not make a difference (e.g. if `WITH_SWSCALE_LOADING` is not defined it should just pass through.

libfreerdp/codec/image_ffmpeg.c
```diff
@@ -0,0 +1,279 @@
...
+	WINPR_ASSERT(g_swscale.freeContext);
+	g_swscale.freeContext(ctx);
+}
+
+// =============================================================================
+// libavutil Runtime Loading
+// =============================================================================
+
+static const char* avutil_library_names[] = {
+#if defined(_WIN32)
```
akallabeth, line 218: this is asking for trouble, might be a good idea to make these names generated from the version (header) compiled against ?

libfreerdp/codec/video.c
```diff
@@ -0,0 +1,560 @@
...
+ */
+
+#include <freerdp/config.h>
+
+#include <winpr/assert.h>
+#include <winpr/wlog.h>
+
+#include <freerdp/codec/video.h>
+
+#define TAG "com.freerdp.codec.video"
```
akallabeth: use the `FREERDP_TAG` et al macros

akallabeth:
@thenextman regarding the `MJPEG` stuff, only doing a runtime check if `libfreerdp` supports it would be enough, just to remove the compile time checks.

akallabeth:
@thenextman any update on this?
needs a rebase and some cleanup.

Will prepare a new release next week, so if you want it in please address the remainining issues.

thenextman:
@akallabeth Sorry the radio silence, busy week

I appreciate the detailed review, I'll get these items fixed up and rebased as soon as possible. We really want this on our side but it's not necessary to get it in the release so no stress. I'll follow up as soon as possible. Thanks again!

include/freerdp/codec/video.h
```diff
@@ -0,0 +1,167 @@
...
+
+	typedef struct s_FREERDP_VIDEO_CONTEXT FREERDP_VIDEO_CONTEXT;
+
+	/**
+	 * @brief Video pixel format enumeration
+	 * Maps to AVPixelFormat but provides independence from FFmpeg headers
+	 */
+	typedef enum
+	{
+		VIDEO_FORMAT_NONE = -1,
```
akallabeth: might be a good idea to prefix the constants with `FREERDP_` or similar as well

include/freerdp/codec/video.h
```diff
@@ -0,0 +1,167 @@
...
+	} FREERDP_VIDEO_FORMAT;
+
+	/**
+	 * @brief Video context options
+	 */
+	typedef enum
+	{
+		VIDEO_CONTEXT_OPTION_DECODER_TYPE, /** MJPEG decoder type */
+		VIDEO_CONTEXT_OPTION_SCALING_QUALITY /** Scaling algorithm quality */
+	} VIDEO_CONTEXT_OPTION;
```
akallabeth: Might be a good idea to prefix with `FREERDP_` (public `API`!)

libfreerdp/codec/color.c
```diff
@@ -1245,22 +1245,33 @@ BOOL freerdp_image_scale(BYTE* WINPR_RESTRICT pDstData, DWORD DstFormat, UINT32
...
+		/* sws_scale expects 4-element arrays even for packed RGB formats with 1 plane */
+		const int srcStep[4] = { (int)nSrcStep, 0, 0, 0 };
+		const int dstStep[4] = { (int)nDstStep, 0, 0, 0 };
 
 		if ((srcFormat == AV_PIX_FMT_NONE) || (dstFormat == AV_PIX_FMT_NONE))
 			return FALSE;
 
-		resize = sws_getContext((int)nSrcWidth, (int)nSrcHeight, srcFormat, (int)nDstWidth,
-		                        (int)nDstHeight, dstFormat, SWS_BILINEAR, NULL, NULL, NULL);
+		if (!freerdp_swscale_available())
```
akallabeth, line 1271: can be moved to `freerdp_sws_getContext`, no need for an external extra check.

channels/rdpecam/client/camera.h
```diff
@@ -108,23 +102,13 @@ typedef struct
...
 	H264_CONTEXT* h264;
-
-#if defined(WITH_INPUT_FORMAT_MJPG)
-	AVCodecContext* avContext;
-	AVPacket* avInputPkt;
-	AVFrame* avOutFrame;
-#endif
+	FREERDP_VIDEO_CONTEXT* video;
 
 #if defined(WITH_INPUT_FORMAT_H264)
```
akallabeth: `WITH_INPUT_FORMAT_H264` can be eliminated, this should be queried from the core library

channels/rdpecam/client/camera_device_main.c
akallabeth: maybe convert this to a function that returns the formats depending on what the library tells is available

channels/rdpecam/client/encoding.c
```diff
@@ -198,113 +198,52 @@ UINT32 h264_get_max_bitrate(UINT32 height)
...
-
-	if (ecam_sws_valid(stream))
+	if (stream->video)
 		return TRUE;
 
-	ecam_sws_free(stream);
-
-	/* replacing deprecated JPEG formats, still produced by decoder */
-	switch (pixFormat)
+	if (!freerdp_video_available())
```
akallabeth: can be moved to `freerdp_video_context_new`

include/freerdp/codec/video.h
```diff
@@ -0,0 +1,167 @@
...
+	FREERDP_API BOOL freerdp_video_fill_plane_info(BYTE* data[4], int lineSize[4],
+	                                               FREERDP_VIDEO_FORMAT format, UINT32 width,
+	                                               UINT32 height, const BYTE* buffer);
+
+	/**
+	 * @brief Check if video processing is available
+	 *
+	 * @return TRUE if video codecs are available (FFmpeg loaded)
+	 */
+	FREERDP_API BOOL freerdp_video_available(void);
```
akallabeth, line 99: maybe add some additional function `FREERDP_API BOOL freerdp_video_format_available(FREERDP_VIDEO_FORMAT flags);` argument to allow checking for specific format support as well (mjpg, h264, ...)

akallabeth (approved):
Overall LGTM, some stuff in public API should be renamed.

include/freerdp/codec/video.h
```diff
@@ -0,0 +1,193 @@
...
+	 * @return TRUE if all requested features are available
+	 */
+	FREERDP_API BOOL freerdp_video_feature_available(UINT32 features);
+
+	/**
+	 * @brief Check if MJPEG decoding is available
+	 *
+	 * @return TRUE if MJPEG decoder is available
+	 */
+	FREERDP_API BOOL freerdp_video_mjpeg_available(void);
```
akallabeth: not sure this is required?

include/freerdp/codec/video.h
```diff
@@ -0,0 +1,193 @@
...
+		FREERDP_VIDEO_CONTEXT_OPTION_SCALING_QUALITY /** Scaling algorithm quality */
+	} FREERDP_VIDEO_CONTEXT_OPTION;
+
+	/**
+	 * @brief Video feature flags for capability checking
+	 */
+	typedef enum
+	{
+		FREERDP_VIDEO_FEATURE_NONE = 0,
+		FREERDP_VIDEO_FEATURE_MJPEG_DECODE = (1 << 0), /** MJPEG decoding support */
```
akallabeth: mjpeg decoding?
more likely `encoding`?

thenextman: > mjpeg decoding? more likely `encoding`?

No, I don't think so. We're decoding MJPEG from the camera and then re-encoding to H264 for transmission. Unless I'm wrong, there's no codepath where we _encode_ MJPEG.

thenextman:
@akallabeth Thanks for the patient and careful review as I refined this.

From my side I think it's good to go but please post back with further comments or issues.

include/freerdp/codec/video.h
akallabeth: true, there is the `mjpg` input variant, but there is also the variant where `mjpg` can be the target. so most likely we need both.

akallabeth:
@thenextman I've now tested this and it works well.
What I still do not like is the public `API`:

1. `FREERDP_VIDEO_FORMAT` should be more or less a copy of `CAM_MEDIA_FORMAT` and not the `FFMpeg` header constants.
2. all the functions for encoding data should be combined and hidden from the public `API`. Think of a `freerdp_video_encode(sourceformat, destinationformat)` style function
3. `freerdp_video_feature_available` should be extended to query `encoding` as well as `decoding` of a format is supported (this way `getSupportedFormats` can be better tailored to provide specific format mappings depending on formats/testcases/config files/...)
4. with the upper points in place `FREERDP_VIDEO_FEATURE` is obsolete

sorry to bring this up so late, but only today had enough time to think about all this (and adding `public API` functions that are not well designed is always a pain in the long run)

include/freerdp/codec/video.h
thenextman: Ok, maybe I am being thick here but I'm not sure I follow

If we look at existing [formats](https://github.com/FreeRDP/FreeRDP/blob/51e66d7bf0500fa0f6fbeab75d282e4f5ecfc2d4/channels/rdpecam/client/camera_device_main.c#L32), the only time MJPEG is an output format is for passthrough (no encode or decode).

So I guess you mean either

1. Add support for MJPEG as an output format (probably not, out of scope for this PR)
2. Add the MJPEG_ENCODE flag but just have it return FALSE (not available). That seems more likely.
3. Or something else, forgive me if I misunderstood completely?

thenextman:
> @thenextman I've now tested this and it works well. What I still do not like is the public `API`:
> 
> 1. `FREERDP_VIDEO_FORMAT` should be more or less a copy of `CAM_MEDIA_FORMAT` and not the `FFMpeg` header constants.
> 2. all the functions for encoding data should be combined and hidden from the public `API`. Think of a `freerdp_video_encode(sourceformat, destinationformat)` style function
> 3. `freerdp_video_feature_available` should be extended to query `encoding` as well as `decoding` of a format is supported (this way `getSupportedFormats` can be better tailored to provide specific format mappings depending on formats/testcases/config files/...)
> 4. with the upper points in place `FREERDP_VIDEO_FEATURE` is obsolete
> 
> sorry to bring this up so late, but only today had enough time to think about all this (and adding `public API` functions that are not well designed is always a pain in the long run)

That's fine, I understand the pain of getting stuck with a badly thought-out public API.

Take a look at the new commits when you get time and LMK what you think

winpr/libwinpr/utils/collections/MessageQueue.c
```diff
@@ -103,52 +94,41 @@ BOOL MessageQueue_Wait(wMessageQueue* queue)
...
+	if ((required < queue->size) || (required < count))
 		return FALSE;
 
 	if (required > queue->capacity)
 	{
 		const size_t old_capacity = queue->capacity;
-		const size_t new_capacity = required + increment;
 
-		wMessage* new_arr = (wMessage*)realloc(queue->array, sizeof(wMessage) * new_capacity);
+		/* When queue is wrapped (tail <= head), we need extra space to unwrap it.
```
akallabeth: remove this, already fixed upstream.

include/freerdp/codec/video.h
```diff
@@ -0,0 +1,274 @@
...
+	} FREERDP_VIDEO_FORMAT;
+
+	/**
+	 * @brief Video context options
+	 */
+	typedef enum
+	{
+		FREERDP_VIDEO_CONTEXT_OPTION_DECODER_TYPE, /** MJPEG decoder type */
+		FREERDP_VIDEO_CONTEXT_OPTION_SCALING_QUALITY /** Scaling algorithm quality */
+	} FREERDP_VIDEO_CONTEXT_OPTION;
```
akallabeth: no longer used?

thenextman:
> Looks a lot better now, I'll play with it some more, hope to get this done next week.

Will wait to address the most recent comments until you've taken a closer look

akallabeth:
@thenextman ok, one thing for sure:
got dozens of warnings such as these:

`freerdp/channels/rdpecam/client/encoding.c:291: warning: Format specifies type 'int' but the argument has underlying type 'unsigned int' (fix available)`

looks like the format strings are wrong on almost all places.

akallabeth:
ok, the function `freerdp_video_convert` (and helpers) is still overly complicated.

The only intended use is to convert an input stream to an output stream.
I think you hook in at the wrong place, see https://github.com/FreeRDP/FreeRDP/blob/37ace4ddc7b7e5ed5c58d3a8842792862cd1b5eb/channels/rdpecam/client/camera_device_main.c#L232 as a starting point.

you have:
* source format
* destination format
* input buffer (serialized) with a length `const BYTE* sample, size_t size`
* output buffer `stream->pendingSample`

so, most opaque (although tailored to this single use case) would be a function `BOOL freerdp_video_sample_convert(FREERDP_VIDEO_CONTEXT* context, FREERDP_VIDEO_FORMAT srcFormat, const void* srcSampleData, size_t srcSampleLength, FREERDP_VIDEO_FORMAT dstFormat wStream* output);`

* this way most of the encoding handling in `rdpecam` will no longer be required and `freerdp_video_sample_convert` can fully utilize `ffmpeg` to convert to/from the formats.
* Not even the checks for `srcFormat != dstFormat` are then required as the function can simply copy the source data to the destination if they equal.

akallabeth:
ok, moving this to another release.

thenextman:
> ok, moving this to another release.

@akallabeth Thanks, will try to go over this again later this week

thenextman:
@akallabeth I did update the approach here, will need to resolve conflicts but please let me know if you have further feedback

akallabeth:
@thenextman the only remaining thing is refactoring the encoder function and hide the ffmpeg specifics.
there is no need to expose all that stuff, a simple function as mentioned in https://github.com/FreeRDP/FreeRDP/pull/12149#issuecomment-3934031703 should be the only thing exposed, without the need for any helper functions preparing the input data. (we got a plain buffer and write to a `wStream` and we know which formats we have as input and output, that is sufficient)

akallabeth:
@thenextman ok, did have an in depth analysis of this pr (once again) and pushed some fixes to https://github.com/akallabeth/FreeRDP/tree/swscale-runtime-loader

please integrate that into your pr (this should stabilize the `API` function and reduce complexity in the `rdpecam` channel)

The remaining issue that remains is that I get really lots of warnings from improper casts, unused functions and similar.
please clean that up, then we can proceed.

akallabeth:
@thenextman wtf?
could you please take a little more care when rebasing?
the new commit is a complete mess.

thenextman:
> @thenextman wtf? could you please take a little more care when rebasing? the new commit is a complete mess.

Apologies, pushed the wrong branch (apparently) and then got called away for other things. Will fix.

akallabeth:
@freerdp-bot test

akallabeth:
@thenextman ok, this is still kind of a mess.
if you want it in the next release, please do a rebase and check each line if that is actually something you wanted to change or some `leftover` from whichever tool you're using...

PR #12162, opened 2026-01-20
Allow querying auth identity with kerberos when running as a server

jadahl:
See commits. This allows me to do

```
  status = freerdp_nla_QueryContextAttributes (rdp_context,
                                               SECPKG_ATTR_AUTH_IDENTITY,
                                               &auth_identity);
```

to get the user and realm (domain) part of a `user@realm` principal that was used for authentication.

PR #12163, opened 2026-01-20
Sspi krb heimdal

akallabeth:
follow up to #12162

### DIFF
--- a/libfreerdp/core/freerdp.c
+++ b/libfreerdp/core/freerdp.c
@@ -1337,14 +1337,11 @@ BOOL freerdp_nla_decrypt(rdpContext* context, const SecBuffer* inBuffer, SecBuff
 
 SECURITY_STATUS freerdp_nla_QueryContextAttributes(rdpContext* context, DWORD ulAttr, PVOID pBuffer)
 {
-	rdpNla* nla;
-
 	WINPR_ASSERT(context);
 	WINPR_ASSERT(context->rdp);
 
-	if (context->rdp->nla)
-		nla = context->rdp->nla;
-	else
+	rdpNla* nla = context->rdp->nla;
+	if (!nla)
 		nla = transport_get_nla(context->rdp->transport);
 
 	WINPR_ASSERT(nla);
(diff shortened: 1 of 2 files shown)


akallabeth:
@jadahl can you have a look at this and check if it works for you  (`MIT` and `Heimdal`?)

jadahl:
> @jadahl can you have a look at this and check if it works for you (MIT and Heimdal?)

Tested with MIT and it appears to still work; installing heimdal-devel on Fedora doesn't seem enough for FreeRDP to find it, even after setting the flavour to Heimdal. I suspect it to be due to it placing the pc file in /usr/lib64/heimdal/lib/pkgconfig/.

akallabeth:
@jadahl `export PKG_CONFIG_PATH=/usr/lib64/heimdal/lib/pkgconfig/` might help there ;)

jadahl:
> @jadahl export PKG_CONFIG_PATH=/usr/lib64/heimdal/lib/pkgconfig/ might help there ;)

Good point :) that made it build. Unfortunately Kerberos authentication doesn't work at all with Heimdal, it seems. 

```
[20:59:43:157] [42195:0000a4e2] [WARN][com.winpr.sspi] - [winpr_AcceptSecurityContext]: AcceptSecurityContext status SEC_E_INVALID_HANDLE [0x80090301]
[20:59:43:157] [42195:0000a4e2] [ERROR][com.freerdp.core.auth] - [credssp_auth_authenticate]: AcceptSecurityContext status SEC_E_INVALID_HANDLE [0x80090301]
[20:59:43:157] [42195:0000a4e2] [ERROR][com.freerdp.core.transport] - [transport_accept_nla]: client authentication failure
[20:59:43:157] [42195:0000a4e2] [ERROR][com.freerdp.api] - [peer_recv_callback_internal]: CONNECTION_STATE_NEGO - rdp_server_accept_nego() fail
[20:59:43:157] [42195:0000a4e2] [ERROR][com.freerdp.core.transport] - [transport_check_fds]: transport_check_fds: transport->ReceiveCallback() - STATE_RUN_FAILED [-1]
```

A related side note: in the server using FreeRDP I also link against libkrb5.so, but only the MIT variant is supported so far. That means building against FreeRDP built with Heimdal doesn't work, unless I also support both, and some how, at build time, detect what Kerberos flavor FreeRDP is using.

I think ideally you'll want to make this discoverable at compile time (without guessing based on paths and `-l` flags), either to fail early on, or to just use the same flavor automatically. E.g. a pkg-config variable or something perhaps.

akallabeth:
@jadahl on the todo list, but `Heimdal` is currently not a first class supported option anyway (just lacking time to check it all)

akallabeth:
so, merging as is, we can always fix the rest later on.

PR #12170, opened 2026-01-21
[crypto] deprecate er and der modules

akallabeth:
These are unused and not well tested, mark them deprecated to announce removal once FreeRDP 4 is released.

### DIFF
--- a/winpr/include/winpr/platform.h
+++ b/winpr/include/winpr/platform.h
@@ -516,15 +516,15 @@ WINPR_PRAGMA_DIAG_POP
 
 #if defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 202311L)
 #define WINPR_DEPRECATED(obj) [[deprecated]] obj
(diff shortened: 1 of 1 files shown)


akallabeth:
@hardening do you use these functions somewhere? you once mentioned we can replace them with the `ASN1` stuff?

hardening:
> @hardening do you use these functions somewhere? you once mentioned we can replace them with the `ASN1` stuff?

Oh yes AFAICT der and er where mostly used for things that are ASN1 at the end (except if I missed some case)

akallabeth:
> > @hardening do you use these functions somewhere? you once mentioned we can replace them with the `ASN1` stuff?
> 
> Oh yes AFAICT der and er where mostly used for things that are ASN1 at the end (except if I missed some case)

yes, in separate commits.
Fixes warnings displayed for `deprecation` symbols and a compiler warning in another. (marked as such)

akallabeth:
@freerdp-bot test

PR #12171, opened 2026-01-22
Add support for querying SECPKG_ATTR_PACKAGE_INFO to NTLM and Kerberos

jadahl:
This also needed some way to free the returned buffer, thus `freerdp_nla_FreeContextBuffer()` was also introduced.

This allows me to support both NTLM and Kerberos authentication, and applying policy in the `Logon()` callback depending on what mechanism was used to authenticate the session.

### DIFF
--- a/winpr/libwinpr/sspi/NTLM/ntlm.c
+++ b/winpr/libwinpr/sspi/NTLM/ntlm.c
@@ -897,6 +897,31 @@ static SECURITY_STATUS SEC_ENTRY ntlm_QueryContextAttributesW(PCtxtHandle phCont
 	{
 		return ntlm_computeMicValue(context, (SecBuffer*)pBuffer);
 	}
+	else if (ulAttribute == SECPKG_ATTR_PACKAGE_INFO)
+	{
+		SecPkgContext_PackageInfo* PackageInfo = (SecPkgContext_PackageInfo*)pBuffer;
+		size_t size = sizeof(SecPkgInfoA);
+		SecPkgInfoA* pPackageInfo =
+		    (SecPkgInfoA*)sspi_ContextBufferAlloc(QuerySecurityPackageInfoIndex, size);
+
+		if (!pPackageInfo)
+			return SEC_E_INSUFFICIENT_MEMORY;
+
+		pPackageInfo->fCapabilities = NTLM_SecPkgInfoA.fCapabilities;
+		pPackageInfo->wVersion = NTLM_SecPkgInfoA.wVersion;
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

include/freerdp/freerdp.h
```diff
@@ -767,6 +767,15 @@ owned by rdpRdp */
 	FREERDP_API SECURITY_STATUS freerdp_nla_QueryContextAttributes(rdpContext* context,
 	                                                               DWORD ulAttr, PVOID pBuffer);
 
+	/** Calls FreeContextbuffer on the SSPI context associated with the NLA part of the RDP context
+	 *
+	 *	\param context the RDP context
+	 *	\param pBuffer an opaque pointer to free
+	 *	\returns a SECURITY_STATUS indicating if the operation completed successfully
+	 *	\since version 3.22.0
```
jadahl, line 775: What should this be?

akallabeth: yes, 3.22.0 it will be (when this is merged)

akallabeth: @jadahl hmm, maybe add a list of supported/known good attributes to the docs of `freerdp_nla_QueryContextAttributes`?

hardening:
What about having the return of `QueryContextAttributes` being something `flat` that can be freed by a simple `free` so that `freerdp_nla_FreeContextBuffer` is just a free call. So the return value is an allocated block containing the SecPkgInfoA with strings appended, and we make the pointers of SecPkgInfoA point to the correct places...

mfleisz:
@hardening I'd rather use a dedicated free-function. On Windows the library and the application might use different heaps (i.e. different msvc runtime) so allocating in the library and free'ing in the application might cause a crash.

hardening:
@mfleisz that's perfectly compatible, I'm just advocating that `FreeContextBuffer` in winPR is just `free`.

mfleisz:
@hardening right - sorry for the confusion! 😅

jadahl:
> @mfleisz that's perfectly compatible, I'm just advocating that FreeContextBuffer in winPR is just free.

`FreeContextBuffer()` in winPR isn't new, nor is how `SecPkgInfo*`'s memory is managed, so feels a bit orthogonal to this PR. Arguably if the way `SecPkgInfo`'s memory is handled is to change, it should be the same for `winpr_QuerySecurityPackageInfoA()`.

include/freerdp/freerdp.h
jadahl: Added a single element list of supported attributes.

akallabeth:
@freerdp-bot test

PR #12173, opened 2026-01-23
Sdl3 high dpiv2

akallabeth:
so, what does this do?

* Refactor the SDL3 client, better encapsulate the classes (use getter/setter instead of direct member access)
* mark functions `[[nodiscard]]` to get warnings for unused return
* split classes to own files (`SdlContext`)
* Fix `high DPI` support in a number of cases. Now with `/dynamic-resolution` the remote display scaling is dynamically updated depending on which monitor the application is on
* simplify coordinate conversion between RDP and local window. (use SDL functions and only apply local scaling for `smart-sizing`)

TODO: check and fix remaining multimon issues:
* ~~Placement of initial window (maybe a `SDL` bug on `wayland` )~~  placed a warning for `wayland` as this is not possible currently.
* ~~add/remove monitors during operation~~

### DIFF
--- a/client/SDL/SDL3/sdl_context.cpp
+++ b/client/SDL/SDL3/sdl_context.cpp
@@ -217,7 +217,24 @@ BOOL SdlContext::postConnect(freerdp* instance)
 	if (freerdp_settings_get_bool(context->settings, FreeRDP_UseMultimon))
 	{
 		const auto driver = SDL_GetCurrentVideoDriver();
-		if (driver && strcmp(driver, "wayland") == 0)
+		bool buggy = false;
+		if (driver)
+		{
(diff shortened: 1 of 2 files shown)


mfleisz:
Two general remarks:
- Do not use `(void)` casts: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts
- There are several places that ignore the return value - this indicates either a missing check or that the function should not be marked as `[[nodiscard]]`

akallabeth:
@mfleisz ah, did not know about `std::ignore =` :+1: 
as for the use itself, still a draft, so yes, some functions might not require `[[nodiscard]]`

PR #12177, opened 2026-01-23
Winpr attr nodiscard

akallabeth:
This adds a compiler attribute to warn about unused results to the functions we expose (and most internal ones)

This might help spot issues where return values have not been checked.

@hardening @mfleisz @bmiklautz what do you think about this?


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #12193, opened 2026-01-27
peer: Disconnect if Logon() returned FALSE

jadahl:
Returning FALSE implies the client was not authorized, but it does not have any real impact on the connection, other than setting the authorized field.

Change this by disconnecting the client if a Logon() callback returned that a client was not authorized.

### DIFF
--- a/libfreerdp/core/peer.c
+++ b/libfreerdp/core/peer.c
@@ -842,8 +842,15 @@ static state_run_t peer_recv_callback_internal(WINPR_ATTR_UNUSED rdpTransport* t
 					IFCALLRET(client->Logon, client->authenticated, client, &client->identity,
 					          FALSE);
 				}
-				if (rdp_server_transition_to_state(rdp, CONNECTION_STATE_MCS_CREATE_REQUEST))
-					ret = STATE_RUN_SUCCESS;
+				if (client->Logon && !client->authenticated)
+				{
+					ret = STATE_RUN_FAILED;
+				}
+				else
+				{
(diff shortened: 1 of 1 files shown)


jadahl:
Cc: @pnowack

libfreerdp/core/peer.c
```diff
@@ -836,11 +836,15 @@ static state_run_t peer_recv_callback_internal(WINPR_ATTR_UNUSED rdpTransport* t
 					IFCALLRET(client->Logon, client->authenticated, client, &client->identity,
 					          TRUE);
 					nego_free_nla(rdp->nego);
+					if (client->Logon && !client->authenticated)
```
mfleisz: Would move this code outside the if/else block to avoid code duplication.

jadahl: Makes sense; speaking of that, should
```
rdp_server_transition_to_state()
```
be called with something else in this case?

pnowack: > Makes sense; speaking of that, should
> rdp_server_transition_to_state()
> be called with something else in this case?

could actually just do `return STATE_RUN_FAILED` or similar (`ret` is already initialized with `STATE_RUN_FAILED`). Relevant is just that the condition under the code here `if (rdp_server_transition_to_state(rdp, CONNECTION_STATE_MCS_CREATE_REQUEST))` does not trigger when `Logon()` failed

jadahl: Yes, doing that now, it avoids the explicit `client->Disconnect()` call.

akallabeth:
@jadahl taken your commit and refactored a little in https://github.com/FreeRDP/FreeRDP/pull/12196
thank you for the fix!

PR #12196, opened 2026-01-27
peer: Disconnect if Logon() returned FALSE

akallabeth:
Returning FALSE implies the client was not authorized, but it does not have any real impact on the connection, other than setting the authorized field.

Change this by disconnecting the client if a Logon() callback returned that a client was not authorized.

### DIFF
--- a/include/freerdp/peer.h
+++ b/include/freerdp/peer.h
@@ -71,8 +71,21 @@ extern "C"
 	typedef BOOL (*psPeerCapabilities)(freerdp_peer* peer);
 	typedef BOOL (*psPeerPostConnect)(freerdp_peer* peer);
 	typedef BOOL (*psPeerActivate)(freerdp_peer* peer);
+
+	/** @brief Callback after the initial RDP authentication (NLA) succeeded or anonymous tunnel was
+	 * established (RDP, TLS, ...)
+	 *
+	 *  @param peer A pointer to a peer context to work on
+	 *  @param identity A pointer to the identity of the peer
+	 *  @param automatic \b TRUE in case the connection is already authenticated, \b FALSE in case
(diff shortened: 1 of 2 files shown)


include/freerdp/peer.h
```diff
@@ -71,8 +71,21 @@ extern "C"
...
+
+	/** @brief Callback after the initial RDP authentication (NLA) succeeded or anonymous tunnel was
+	 * established (RDP, TLS, ...)
+	 *
+	 *  @param peer A pointer to a peer context to work on
+	 *  @param identity A pointer to the identity of the peer
+	 *  @param automatic \b TRUE in case the connection is already authenticated, \b FALSE in case
+	 * of \b RDP, \b TLS or similar anonymous tunnels
+	 *
+	 *  @return \b TRUE if the connection is allowed, \b FALSE if denied. Defaults to \b TRUE if the
```
jadahl, line 83: Doesn't this imply that all current users (that are unlikely to set this) will break?

jadahl: Or at least the one I'm working on doesn't currently set this callback to anything, so upgrading FreeRDP will break it.

akallabeth: no, default return is `TRUE`, aka pass the stage?

akallabeth: see https://github.com/FreeRDP/FreeRDP/blob/c8e4d8e5b0df13072e722ab24a93228a37b76ffb/include/freerdp/api.h#L94

jadahl: Ugh, I can't read I think...

akallabeth: well, I know that feeling from somewhere :grin:

PR #12197, opened 2026-01-28
[channels,rdpecam] fix PROPERTY_DESCRIPTION parsing

akallabeth:
* The Capabilities field is a mask, so allow combination of all valid values.
* Adjust header to hold a mask
* Use enum to define constants for parsed values


include/freerdp/channels/rdpecam.h
```diff
@@ -269,35 +269,41 @@ typedef struct
...
+	CAM_PROPERTY_CAPABILITY_Manual = 0x01u,
+	CAM_PROPERTY_CAPABILITY_Auto = 0x02u
 } CAM_PROPERTY_CAPABILITIES;
 
 typedef struct
 {
 	CAM_PROPERTY_SET PropertySet;
 	BYTE PropertyId;
-	CAM_PROPERTY_CAPABILITIES Capabilities;
+	UINT32 Capabilities;
```
pnowack, line 306: It actually is just one byte
```suggestion
	BYTE Capabilities;
```

channels/rdpecam/server/rdpecam-utils.h
```diff
@@ -307,19 +307,16 @@ static inline bool rdpecam_valid_CamPropertySet_(UINT8 val, wLog* log, const cha
  */
 #define rdpecam_valid_CamPropertyCapabilities(val) \
 	rdpecam_valid_CamPropertyCapabilities_((val), WLog_Get(TAG), __FILE__, __func__, __LINE__)
-static inline bool rdpecam_valid_CamPropertyCapabilities_(UINT8 val, wLog* log, const char* file,
+static inline bool rdpecam_valid_CamPropertyCapabilities_(UINT32 val, wLog* log, const char* file,
```
pnowack, line 310: same here, the capability bitmask is 8 byte long
```suggestion
static inline bool rdpecam_valid_CamPropertyCapabilities_(BYTE val, wLog* log, const char* file,
```

pnowack (commented):
This works in a test with Windows App on iOS and g-r-d

include/freerdp/channels/rdpecam.h
akallabeth: but the `enum` before did compile to 4 bytes ;)

PR #12207, opened 2026-01-29
Vendor by app

akallabeth:


libfreerdp/core/settings.c
```diff
@@ -187,119 +188,165 @@ static BOOL settings_reg_query_bool(rdpSettings* settings, FreeRDP_Settings_Keys
...
+					(void)_sntprintf(numentries, ARRAYSIZE(numentries), _T("Cache%uNumEntries"), x);
+					(void)_sntprintf(maxsize, ARRAYSIZE(maxsize), _T("Cache%uMaxCellSize"), x);
+
+					settings_reg_query_word_val(hKey, numentries, &cache.cacheEntries);
+					settings_reg_query_word_val(hKey, maxsize, &cache.cacheMaximumCellSize);
+					if (!freerdp_settings_set_pointer_array(settings, FreeRDP_GlyphCache, x,
+					                                        &cache))
+						WLog_WARN(TAG, "Failed to store GlyphCache %u", x);
+				}
+				{
```
mfleisz, line 298: This looks weird - is there a `for` missing or should this be part of the previous `for`?

akallabeth:
> The code in `utils/helpers.c` and `utils/winpr.c` are pretty much identical. Wouldn't it be better to just use the winpr functions or at least just forward the freerdp_ calls to the winpr_ ones to avoid code duplication?

no, the builds of `winpr` and `freerdp` could use different compile time flags, so better keeping them separate.

libfreerdp/core/settings.c
akallabeth: scope for local variable `cache`

Issue #12176, opened 2026-01-23
Multimonitor not working with wayland


punx120:
Hi,

I'm having issue with sdl freerdp on gnome / wayland with 4 screens. I launched it like this:

```
/usr/bin/sdl-freerdp3  /v:host /network:auto /rfx /gfx:avc444 /sound /mic:sys:pulse,format:1 /multimon /f
```

I do see 4 different windows, looks like 1 window per monitor, but they all appeared on the same monitor and i can't move them. (xfreerdp works fine).

**Application details**
```
sdl-freerdp3 /version
This is FreeRDP version 3.21.0 (3.21.0)
````
* Command line used
* Output of `xfreerdp /buildconfig`
```
This is FreeRDP version [sdl-freerdp3] 3.21.0 (3.21.0)
[...]

PR #12248, opened 2026-02-05
[client,sdl] fix multimon/fullscreen on wayland

motor-dev:
Help SDL identify the display to use for fullscreen even when compositors interfere with window positions. SDL respects the intent of a window if it is moved to the appropriate display just before it is made fullscreen.

When using the /f or /multimon flag, the initial fullscreen request will force the window to go fullscreen on the display that was passed to window creation. Subsequent fullscreen requests do not do that - if the user manually moves a window to a display and requests fullscreen, the window will go fullscreen on the current display.

fix for https://github.com/FreeRDP/FreeRDP/issues/12176

clang-format correctly  reformatted the switch statement in sdl_freedrp.cpp. Turn off whitespace differences to see the actual diff.


akallabeth:
@freerdp-bot test

akallabeth:
@motor-dev nice one, thank you!

PR #12250, opened 2026-02-06
[client,windows] Fix mouse wheel delta encoding for RDP protocol compliance

tsz8899:
This patch addresses an issue in the Windows client's mouse wheel event handling where the delta encoding did not fully comply with the RDP protocol specification.

Previously, the  value was directly used which could lead to incorrect scroll direction or magnitude interpretation on some RDP servers.

The fix implements:
- Direction is now explicitly set via .
- Magnitude (steps) is encoded as an absolute value in the lower 8 bits, using 9-bit two's complement for negative values.
- Ensures consistency with SDL/X11 encoding practices where applicable.

This improves compatibility and reliability of mouse wheel input with various RDP server implementations.

### DIFF
--- a/client/Windows/wf_event.c
+++ b/client/Windows/wf_event.c
@@ -256,21 +256,46 @@ static BOOL wf_event_process_WM_MOUSEWHEEL(wfContext* wfc, HWND hWnd, UINT Msg, 
 	WINPR_ASSERT(input);
 
 	DefWindowProc(hWnd, Msg, wParam, lParam);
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
* lots of (unrelated) changes
* the value is not in twos complement format! (yes, the old comment was wrong)

only relevant change I can see here is doing a cutoff if the value is larger than the maximum which was lacking before?

tsz8899:
Closing in favor of a cleaner, more focused PR addressing the magnitude overflow and previous feedback.

PR #12252, opened 2026-02-06
[client,rail] Fix icon dimension and size calculation integer overflows

tsz8899:
This patch addresses potential integer overflow vulnerabilities in the Windows client's RAIL icon processing. Without proper bounds checking, excessively large icon dimensions or total pixel area could lead to integer overflows during the calculation of  (bitmap size). This could result in memory corruption, crashes, or other security exploits.

The fix introduces:
- Validation for icon  and  to prevent individual dimension overflow (e.g., if one dimension exceeds ).
- Validation for the total pixel  () to ensure it does not exceed a reasonable maximum (e.g., ).
- A safer calculation for  after these dimension checks have been passed.

These checks ensure that icon data is processed safely and within expected limits, improving the robustness and security of the client.


tsz8899:
Closing due to incorrect title prefix. This PR will be re-submitted with the [client,windows] prefix as per contributor's requested convention.

PR #12253, opened 2026-02-06
[client,keyboard] Implement complete keyboard indicator synchronization

tsz8899:
This patch implements a comprehensive synchronization mechanism for keyboard indicators (Num Lock, Caps Lock, Scroll Lock, Kana Lock) between the FreeRDP Windows client and the RDP server.

Key changes include:
- Calling  in the  handler to ensure keyboard state is synchronized when the client window gains focus.
- Implementing the  callback, which allows the RDP server to update the local client's keyboard LED states.
- Registering  in  and performing an initial keyboard state sync immediately after connection establishment.

This resolves issues where local and remote keyboard indicator states could become desynchronized, improving the user experience by ensuring consistent behavior of keyboard LEDs.


tsz8899:
Closing due to incorrect title prefix. This PR will be re-submitted with the [client,windows] prefix as per contributor's requested convention.

PR #12254, opened 2026-02-06
[client,windows] Enhance event processing safety with robust NULL and state checks

tsz8899:
This patch significantly enhances the robustness and crash-prevention of the FreeRDP Windows client's event processing. It addresses potential instability issues that can arise from attempting to access invalid, stale, or already-disconnected resources during keyboard and focus events.

Key safety enhancements include:
- ****: Added  checks for  and , along with  and connection instance validation, preventing event processing on invalid windows or during disconnection.
- ****: Implemented comprehensive  and validity checks for , its window handle, the FreeRDP instance, and input callbacks before proceeding with focus-in event logic. This ensures that input events are only dispatched when the client is in a valid and connected state.

These changes aim to prevent crashes related to race conditions, window lifecycle issues, or unexpected disconnections, leading to a more stable client experience.

### DIFF
--- a/client/Windows/wf_event.c
+++ b/client/Windows/wf_event.c
@@ -88,7 +88,29 @@ LRESULT CALLBACK wf_ll_kbd_proc(int nCode, WPARAM wParam, LPARAM lParam)
 
 	if (g_parent_hWnd && g_main_hWnd)
 	{
+		/* 修复：检查窗口句柄是否仍然有效 */
+		if (!IsWindow(g_main_hWnd))
+		{
+			g_main_hWnd = NULL;
+			g_focus_hWnd = NULL;
+			return CallNextHookEx(NULL, nCode, wParam, lParam);
+		}
+
 		wfc = (wfContext*)GetWindowLongPtr(g_main_hWnd, GWLP_USERDATA);
+
+		/* 修复：检查wfc有效性，避免访问无效指针 */
+		if (!wfc)
+		{
+			return CallNextHookEx(NULL, nCode, wParam, lParam);
+		}
+
+		/* 修复：检查连接状态，断开连接时不处理键盘事件 */
+		if (!wfc->common.context.instance ||
+		    freerdp_shall_disconnect_context(&wfc->common.context))
+		{
+			return CallNextHookEx(NULL, nCode, wParam, lParam);
+		}
+
 		GUITHREADINFO gui_thread_info;
 		gui_thread_info.cbSize = sizeof(GUITHREADINFO);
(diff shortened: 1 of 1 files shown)


client/Windows/wf_event.c
```diff
@@ -88,7 +88,29 @@ LRESULT CALLBACK wf_ll_kbd_proc(int nCode, WPARAM wParam, LPARAM lParam)
 
 	if (g_parent_hWnd && g_main_hWnd)
 	{
+		/* 修复：检查窗口句柄是否仍然有效 */
```
akallabeth, line 91: english please

client/Windows/wf_event.c
```diff
@@ -218,7 +254,51 @@ void wf_event_focus_in(wfContext* wfc)
 	rdpInput* input;
 	POINT pt;
 	RECT rc;
+
+#define TAG CLIENT_TAG("windows")
```
akallabeth, line 258: move to top of file please

client/Windows/wf_event.c
```diff
@@ -118,20 +140,34 @@ LRESULT CALLBACK wf_ll_kbd_proc(int nCode, WPARAM wParam, LPARAM lParam)
...
+						/* 修复：按下时按虚拟键码置位，避免扫描码/虚拟键码不一致问题 */
+						if (p->vkCode < 256)
+							g_keystates[p->vkCode] = TRUE;
 						break;
 					case WM_KEYUP:
 					case WM_SYSKEYUP:
 					default:
-						g_keystates[p->scanCode & 0xFF] = FALSE;
+						/* 修复：弹起时按虚拟键码清零 */
+						if (p->vkCode < 256)
```
akallabeth, line 169: here a log would help to identify if there are keys we can not map.

client/Windows/wf_event.c
```diff
@@ -88,7 +88,29 @@ LRESULT CALLBACK wf_ll_kbd_proc(int nCode, WPARAM wParam, LPARAM lParam)
...
 		wfc = (wfContext*)GetWindowLongPtr(g_main_hWnd, GWLP_USERDATA);
+
+		/* 修复：检查wfc有效性，避免访问无效指针 */
+		if (!wfc)
+		{
+			return CallNextHookEx(NULL, nCode, wParam, lParam);
+		}
+
+		/* 修复：检查连接状态，断开连接时不处理键盘事件 */
+		if (!wfc->common.context.instance ||
```
akallabeth, line 108: why is this check required here? should suffice if the  `RDP` thread checks that?

akallabeth (changes requested):

PR #12258, opened 2026-02-06
[client,windows] Fix system menu text memory leak

tsz8899:
This patch addresses a memory leak in the FreeRDP Windows client related to the system menu text. Previously, the  function was used to copy the window title to the system menu, but the memory allocated for the  was not properly freed after use.

The fix ensures that the memory allocated for the system menu text is freed when  is called, preventing a gradual memory leak that could occur as window titles are updated or sessions are closed. This improves the long-term stability and resource management of the client.

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -328,6 +328,8 @@ static void wf_append_item_to_system_menu(HMENU hMenu, UINT fMask, UINT wID, con
 	if (wfc)
 		item_info.dwItemData = (ULONG_PTR)wfc;
 	InsertMenuItem(hMenu, wfc->systemMenuInsertPosition++, TRUE, &item_info);
+
+	free(item_info.dwTypeData);
 }
 
 static void wf_add_system_menu(wfContext* wfc)


client/Windows/wf_client.c
```diff
@@ -323,7 +323,7 @@ static void wf_append_item_to_system_menu(HMENU hMenu, UINT fMask, UINT wID, con
 	item_info.cbSize = sizeof(MENUITEMINFO);
 	item_info.wID = wID;
 	item_info.fType = MFT_STRING;
-	item_info.dwTypeData = _wcsdup(text);
+	item_info.dwTypeData = (LPWSTR)text;
```
akallabeth: wrong cast, remove!

client/Windows/wf_client.c
```diff
@@ -319,15 +319,24 @@ static void wf_append_item_to_system_menu(HMENU hMenu, UINT fMask, UINT wID, con
...
+	text_buffer = _wcsdup(text);
+	item_info.dwTypeData = text_buffer;
 	item_info.cch = (UINT)_wcslen(text);
+
 	if (wfc)
 		item_info.dwItemData = (ULONG_PTR)wfc;
+
 	InsertMenuItem(hMenu, wfc->systemMenuInsertPosition++, TRUE, &item_info);
+
+	free(text_buffer);
```
MartinHaimberger: Why not use free(item_info.dwTypeData) if this is a leak and forget this over complicated way (is this AI generated???)

tsz8899: Thanks for the feedback. 
I’ve updated the change to keep a single `_wcsdup`, removed the cast, and now free `item_info.dwTypeData` immediately after `InsertMenuItem`. This matches the suggested cleanup and avoids the extra buffer. Please let me know if you see any remaining concerns.

Tools get the job done,decision-making secure wins the long game.

MartinHaimberger: @tsz8899 yea tools CAN get jobs done, but if such a AI bullshit is commited, which doest not even compile or was testet before commiting, not to speak of efficient code, then the question is if it is better NOT to use that tools and think over the changes yourself.

PR #12260, opened 2026-02-06
[client,windows] Fix premature session exit when closing a RAIL window

tsz8899:
### Summary
Historically, `PostQuitMessage(0)` was called in `wf_RailWndProc` upon receiving `WM_DESTROY`, following standard single-window Win32 practices. However, in RAIL (RemoteApp) mode where multiple application windows coexist, this causes the entire client session to terminate whenever any single RemoteApp window is closed.

This patch removes the unnecessary call to ensure session continuity across multiple RAIL windows.

### Changes
- Removed `PostQuitMessage(0)` from the `WM_DESTROY` handler in `wf_RailWndProc` (client/Windows/wf_rail.c).
- This ensures that closing a single RemoteApp window only destroys that specific window instead of sending a `WM_QUIT` signal to the entire application message loop.

### DIFF
--- a/client/Windows/wf_rail.c
+++ b/client/Windows/wf_rail.c
@@ -400,7 +400,6 @@ LRESULT CALLBACK wf_RailWndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lPara
 			break;
 
 		case WM_DESTROY:
-			PostQuitMessage(0);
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
description does not match code changes?

PR #12261, opened 2026-02-06
[client,windows] Add window handle validation in RDP_EVENT_TYPE_WINDOW_NEW

tsz8899:
This patch enhances the robustness of the FreeRDP Windows client by adding explicit validation for window handles during the processing of  events.

Specifically, in the  function within , a check () is introduced before attempting to use the newly created window handle. This ensures that subsequent operations on the window (such as updating its properties or registering it with the system) are only performed if a valid window handle was successfully created.

This prevents potential crashes or undefined behavior that could occur if  fails to return a valid window handle, improving the stability of the client when handling new RAIL windows.


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #12270, opened 2026-02-06
[client,windows] Enhance disconnect safety and resource cleanup (v2)

tsz8899:
This patch, an updated version (v2), aims to significantly improve the safety and reliability of the FreeRDP Windows client during the disconnection process. It addresses various scenarios where premature cleanup or invalid state access could lead to crashes or resource leaks.

Key enhancements include:
- ****: Ensures  is explicitly set to  at the beginning of the connection phase, preventing potential use of stale process IDs.
- ****: Implements a crucial check () before attempting to send a  message to the message thread. This prevents sending messages to an invalid process ID if the thread has already exited or  is , avoiding potential crashes. It also sets  after cleanup.
- ****: Adds a  check to ensure the  function is only called if a valid process ID exists. This prevents errors when attempting to terminate an already-exited or invalid process.
- **Robustness**: These changes collectively make the disconnection path more robust against race conditions and unexpected thread/process states.

These improvements enhance the overall stability of the FreeRDP Windows client by ensuring resources are released safely and interactions with the Windows API are guarded against invalid states during disconnection.

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -1032,6 +1032,14 @@ static BOOL wf_present_gateway_message(freerdp* instance, UINT32 type, BOOL isDi
 	return TRUE;
 }
 
+/* Fix: Use an asynchronous disconnect helper thread to prevent freerdp_disconnect from blocking */
+static DWORD WINAPI wf_disconnect_thread(LPVOID lpParam)
+{
+	freerdp* instance = (freerdp*)lpParam;
+	freerdp_disconnect(instance);
+	return 0;
+}
+
 static DWORD WINAPI wf_client_thread(LPVOID lpParam)
 {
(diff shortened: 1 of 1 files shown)


client/Windows/wf_client.c
```diff
@@ -1067,7 +1075,7 @@ static DWORD WINAPI wf_client_thread(LPVOID lpParam)
 		if (freerdp_focus_required(instance))
 		{
 			wf_event_focus_in(wfc);
-			wf_event_focus_in(wfc);
+			/* Fix: Remove redundant call to avoid unnecessary overhead */
```
akallabeth, line 1078: do not remove, this works around some older windows server bugs.

tsz8899: ok

PR #12273, opened 2026-02-08
[winpr,smartcard] simplify PCSC_ReadDeviceSystemName

akallabeth:
* If we read SCARD_ATTR_DEVICE_SYSTEM_NAME_A just pass the actual destination buffer, no further processing required.
* In case of SCARD_ATTR_DEVICE_SYSTEM_NAME_W allocate a WCHAR version of the result, copy back to destination buffer.


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #12276, opened 2026-02-08
[client,windows] Fix Ctrl+Alt+Enter shortcut by syncing keystates with Virtual Key codes

tsz8899:
This patch addresses the issue where the Ctrl+Alt+Enter fullscreen toggle was non-functional in the Windows client.

Technical Details:
- The keyboard hook procedure was previously tracking key states in  using scan codes.
- However, the shortcut detection logic (, ) depends on Virtual Key codes (VK_CONTROL, VK_MENU, etc.).
- This mismatch caused the shortcut sequence to be ignored.

Fix:
Updated the keyboard hook to update  using Virtual Key codes, ensuring consistency across all shortcut detection points. All comments have been translated to English and the file has been formatted using clang-format.

### DIFF
--- a/client/Windows/wf_event.c
+++ b/client/Windows/wf_event.c
@@ -120,20 +120,28 @@ LRESULT CALLBACK wf_ll_kbd_proc(int nCode, WPARAM wParam, LPARAM lParam)
 
 				input = wfc->common.context.input;
 				rdp_scancode = MAKE_RDP_SCANCODE((BYTE)p->scanCode, p->flags & LLKHF_EXTENDED);
-				keystate = g_keystates[p->scanCode & 0xFF];
+				/* Fix: keystate tracking was using scan codes, but shortcut detection
+				   uses Virtual Key codes, preventing Ctrl+Alt+Enter from triggering.
+				   Syncing to Virtual Key codes ensures consistent shortcut detection. */
+				keystate = g_keystates[p->vkCode & 0xFF];
 
 				switch (wParam)
 				{
 					case WM_KEYDOWN:
 					case WM_SYSKEYDOWN:
-						g_keystates[p->scanCode & 0xFF] = TRUE;
+						/* Fix: set state by Virtual Key code to avoid scan code inconsistencies */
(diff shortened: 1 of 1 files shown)


client/Windows/wf_event.c
```diff
@@ -120,20 +120,28 @@ LRESULT CALLBACK wf_ll_kbd_proc(int nCode, WPARAM wParam, LPARAM lParam)
...
+				keystate = g_keystates[p->vkCode & 0xFF];
 
 				switch (wParam)
 				{
 					case WM_KEYDOWN:
 					case WM_SYSKEYDOWN:
-						g_keystates[p->scanCode & 0xFF] = TRUE;
+						/* Fix: set state by Virtual Key code to avoid scan code inconsistencies */
+						if (p->vkCode < 256)
+							g_keystates[p->vkCode] = TRUE;
```
MartinHaimberger, line 134: Doesn't this change the behavior compared to the old code?

tsz8899: After searching the entire project, it was found that the g_keystates array is only defined and used within the `wf_event.c` file (it is a static variable). This implies:
* This change will not affect any cross-module RDP communication protocol.
* This change will not affect the keyboard status seen by the RDP server (because what is sent to the server is still p->flags and rdp_scancode).
* The only changed behavior: it is merely a small cache used internally by the FreeRDP local client to determine "whether the current Ctrl/Alt is pressed".

 1. g_keystates is a static local array used exclusively for internal shortcut detection (ctrl_down, alt_ctrl_down). It is not exported or used for RDP protocol packets.
 2. The previous logic used Scan Codes to update the array but queried it using Virtual Key codes. This is logically inconsistent because Scan Codes and VK codes are not interchangeable indices.
 3. By syncing to p->vkCode, we ensure that the tracking array matches the query constants (like VK_CONTROL), restoring the functionality of shortcuts like Ctrl+Alt+Enter.
 4. Standard RDP packet transmission (input->KeyboardEvent) remains untouched as it still uses the raw rdp_scancode and flags."

PR #12277, opened 2026-02-08
[client,windows] Enhance input event safety and window transition robustness

tsz8899:
This patch addresses two critical issues in the Windows client input system:

1. Program Hang during Exit:
- Problem: The low-level keyboard hook could attempt to access an invalid window context or RDP instance during session teardown, leading to deadlocks or crashes.
- Fix: Added IsWindow() checks and robust NULL validation at the hook entry points to safely bail out if the session is no longer active.

2. Keyboard State Inconsistency during Focus Transitions:
- Problem: Switching focus between FreeRDP and other local applications sometimes resulted in stuck keys or desynchronized lock-key states.
- Fix: Reinforced wf_event_focus_in with comprehensive validation of the input interface and its callbacks, ensuring clean state synchronization when the window regains focus.

Key Improvements per Maintainer Feedback:
- Moved #define TAG to the top of the file.
- Replaced Chinese comments with clear English descriptions.
- Added debug logging for Virtual Key codes that fall outside the 0-255 tracking range.
- Removed redundant deep-structure checks in favor of interface-level validation.
- Formatted the code using clang-format.

### DIFF
--- a/client/Windows/wf_event.c
+++ b/client/Windows/wf_event.c
@@ -36,6 +36,8 @@
 
 #include <windowsx.h>
 
+#define TAG CLIENT_TAG("windows")
+
 static HWND g_focus_hWnd = NULL;
 static HWND g_main_hWnd = NULL;
 static HWND g_parent_hWnd = NULL;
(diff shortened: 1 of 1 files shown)


client/Windows/wf_event.c
```diff
@@ -118,7 +127,16 @@ LRESULT CALLBACK wf_ll_kbd_proc(int nCode, WPARAM wParam, LPARAM lParam)
...
 					return 1;
 
+				if (p->vkCode >= 256)
+				{
+					WLog_Print(TAG, WLOG_DEBUG, "Processing VK code out of tracking range: 0x%08lX",
+					           p->vkCode);
+				}
+
 				input = wfc->common.context.input;
+				if (!input || !input->KeyboardEvent)
```
MartinHaimberger, line 137: input->KeyboardEvent is not even used here ... ?!

PR #12279, opened 2026-02-08
[client,windows] Fix return value checks for mouse event functions

tsz8899:


akallabeth:
@freerdp-bot test

hardening:
Comments should have been dropped. A comment saying "when the function returns TRUE that means success" is pretty useless...

akallabeth:
@hardening true, was just too lazy :grin: 
doubt that there will be a proper fix, the pr really are automated (and most of them broken)

PR #12297, opened 2026-02-13
[client,SDL3] Fix high-DPI handling and window sizing on scaled displays

morgan9e:
Fixes DPI scaling and window sizing on scaled displays.

Removes the highDpi guard so fullscreen modes are always enumerated, since SDL_GetDisplayBounds() may return logical coordinates.

Fixes wrong client window size when used on scaled desktops by using correct logical size.

---

Note: I think there's ongoing https://github.com/FreeRDP/FreeRDP/pull/12217 that solves this fundamentally, but I thought it better be fixed on current code until refactor is done.

### DIFF
--- a/client/SDL/SDL3/sdl_monitor.cpp
+++ b/client/SDL/SDL3/sdl_monitor.cpp
@@ -193,13 +193,8 @@ int sdl_list_monitors([[maybe_unused]] SdlContext* sdl)
 	WINPR_ASSERT(rect.w > 0);
 	WINPR_ASSERT(rect.h > 0);
 
-	bool highDpi = dpi > 100;
-
-	if (highDpi)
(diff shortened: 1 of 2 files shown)


akallabeth:
@morgan9e maybe you can take a look at my unfinished pr https://github.com/FreeRDP/FreeRDP/pull/12217

I'm still working on creating a proper display map which should contain the `logical rect` and `pixel rect` for each display.

morgan9e:
> @morgan9e maybe you can take a look at my unfinished pr #12217
> 
> I'm still working on creating a proper display map which should contain the `logical rect` and `pixel rect` for each display.


Yes, sorry. maybe it's better to consider platform and coordinate differences 

I'll try out and test #12217, thanks!


(closing since it'll need rework)

PR #12298, opened 2026-02-13
[client,SDL3] Fix properly handle smart-sizing with fullscreen

morgan9e:
### Fix handling of smart-sizing + fullscreen case from SDL3 client

> If /f is specified in combination with /smart-sizing:widthxheight then we run the session in the /smart-sizing dimensions scaled to full screen ([client/X11/xf_client.c#L1246](https://github.com/FreeRDP/FreeRDP/blob/341b25d508c78fedced1a7152bff14638fddc1bf/client/X11/xf_client.c#L1246))


akallabeth:
@freerdp-bot test
@morgan9e thank you for the fix! taken your commit and added proper handling of `/multimon` and just `/smart-sizing` as well.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #12309, opened 2026-02-15
[android] Fix invert scrolling default value mismatch

### LINKED ISSUE #12308: [Android] "Invert Scrolling" default value (true) not applied on fresh install until toggled
## Description

The "Invert Scrolling" setting under **Settings > User Interface** has a default value of `true` (checked) in `settings_app_ui.xml`. However, on a fresh install of the aFreeRDP Android app, the invert scrolling behavior does not actually take effect even though the checkbox appears checked.

The scrolling only works in the inverted direction **after manually unchecking and re-checking** the "Invert Scrolling" option.

## Steps to Reproduce

1. Install aFreeRDP on an Android device (fresh install / cleared app data)
2. Open the app and go to **Settings > User Interface**
3. Observe that "Invert Scrolling" is checked by default
4. Connect to a remote desktop session
5. Scroll using touch gestures - notice scrolling is **not** inverted despite the setting being checked
6. Go back to **Settings > User Interface**
7. Uncheck "Invert Scrolling", then re-check it
8. Reconnect to the remote session
9. Scrolling is now correctly inverted

## Expected Behavior

The "Invert Scrolling" setting should be applied immediately on first launch without requiring a manual toggle. Since the default value is `true`, scrolling should be inverted from the very first session.

## Actual Behavior

The default value of `true` in the XML is not being persisted into `SharedPreferences` until the user explicitly interacts with the checkbox. The `getInvertScrolling()` method in `ApplicationSettingsActivity.java` reads from SharedPreferences with a default of `false`, which contradicts the XML default of `true`.

## Root Cause Analysis

Looking at the code:

- In `settings_app_ui.xml` (line 15-18), the CheckBoxPreference has `android:defaultValue="true"`
- In `ApplicationSettingsActivity.java` (line 281-286), `getInvertScrolling()` calls `preferences.getBoolean(...)` with a fallback default of `false`

If `PreferenceManager.setDefaultValues()` has not been called before the first read (or is called with `readAgain=false` and defaults were already partially set), the SharedPreferences key may not exist yet, causing `getBoolean()` to return `false` instead of the intended `true`.

## Suggested Fix

Ensure the fallback default in `getInvertScrolling()` matches the XML default:

```java
// In ApplicationSettingsActivity.java, line 284:
return preferences.getBoolean(
    context.getString(R.string.preference_key_ui_invert_scrolling), true);  // changed false -> true
```

Alternatively, ensure `PreferenceManager.setDefaultValues(context, R.xml.settings_app_ui, false)` is called early enough in the app lifecycle.

## Environment

- **App:** aFreeRDP (Android client)
- **FreeRDP version:** latest master
- **Android version:** 13+
- **Device:** Multiple Android devices tested

Wladefant:
The `getInvertScrolling()` fallback default was `false`, while the XML preference in `settings_app_ui.xml` defines `android:defaultValue="true"`.

On a fresh install, if `PreferenceManager.setDefaultValues()` has not yet persisted the XML defaults into SharedPreferences, the fallback value `false` is returned instead of the intended `true`. This causes the "Invert Scrolling" option to appear checked in the UI but not actually take effect until the user toggles it manually.

Change the fallback from `false` to `true` to match the XML default.

Fixes #12308

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/presentation/ApplicationSettingsActivity.java
+++ b/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/presentation/ApplicationSettingsActivity.java
@@ -282,7 +282,7 @@ public class ApplicationSettingsActivity extends AppCompatPreferenceActivity
 	{
 		SharedPreferences preferences = get(context);
 		return preferences.getBoolean(
-		    context.getString(R.string.preference_key_ui_invert_scrolling), false);
+		    context.getString(R.string.preference_key_ui_invert_scrolling), true);
 	}
 
 	public static boolean getAskOnExit(Context context)


akallabeth:
@freerdp-bot test

PR #12322, opened 2026-02-17
[winpr,utils] ensure message queue capacity

akallabeth:
the capacity should always be larger than what is actually used.

### DIFF
--- a/winpr/libwinpr/utils/test/TestQueue.c
+++ b/winpr/libwinpr/utils/test/TestQueue.c
@@ -3,56 +3,229 @@
 #include <winpr/tchar.h>
 #include <winpr/collections.h>
 
-int TestQueue(int argc, char* argv[])
+static bool wrap_test(bool (*fkt)(wQueue* queue))
 {
-	size_t item = 0;
-	size_t count = 0;
-	wQueue* queue = NULL;
-
-	WINPR_UNUSED(argc);
-	WINPR_UNUSED(argv);
-
-	queue = Queue_New(TRUE, -1, -1);
+	wQueue* queue = Queue_New(TRUE, -1, -1);
 	if (!queue)
-		return -1;
+		return false;
+
+	WINPR_ASSERT(fkt);
+	const bool rc = fkt(queue);
+	Queue_Free(queue);
+	return rc;
+}
+
+static bool check(const void* ptr, size_t pos)
+{
+	if (!ptr)
+		return false;
+	if (ptr != (void*)(pos + 23))
+		return false;
+	return true;
+}
+
+static bool append(wQueue* queue, size_t pos)
+{
+	void* ptr = (void*)(pos + 23);
+	return Queue_Enqueue(queue, ptr);
+}
+
+static bool fill_capcity(wQueue* queue, size_t* pos)
+{
+	WINPR_ASSERT(pos);
+
+	size_t cpos = *pos;
+	const size_t capacity = Queue_Capacity(queue);
+	while (Queue_Count(queue) < capacity)
+	{
+		if (!append(queue, cpos++))
+			return false;
+	}
+	*pos = cpos;
+	return true;
+}
+
+static bool drain(wQueue* queue, size_t expect)
+{
+	void* ptr = Queue_Dequeue(queue);
+	return check(ptr, expect);
+}
+
+static bool drain_capcity(wQueue* queue, size_t remain, size_t* pos)
+{
+	WINPR_ASSERT(pos);
+
+	size_t cpos = *pos;
+	while (Queue_Count(queue) > remain)
+	{
+		if (!drain(queue, cpos++))
+			return false;
+	}
+	*pos = cpos;
+
+	return true;
+}
+
+static bool test_growth_move(wQueue* queue, bool big)
+{
+	WINPR_ASSERT(queue);
+
+	const size_t cap = Queue_Capacity(queue);
+	if (cap < 4)
+		return false;
+
+	size_t wpos = 0;
+	size_t rpos = 0;
+	if (!fill_capcity(queue, &wpos))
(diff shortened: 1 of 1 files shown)


winpr/libwinpr/utils/collections/MessageQueue.c
```diff
@@ -94,29 +103,49 @@ BOOL MessageQueue_Wait(wMessageQueue* queue)
...
 
 	const size_t required = queue->size + count;
 	// check for overflow
-	if ((required < queue->size) || (required < count))
+	if ((required < queue->size) || (required < count) ||
+	    (required > (SIZE_MAX - increment) / sizeof(wMessage)))
 		return FALSE;
 
-	if (required > queue->capacity)
+	if (required >= queue->capacity)
```
mfleisz: Should be only `>` - no need to increase the size if capacity meets our requirement

winpr/libwinpr/utils/collections/MessageQueue.c
```diff
@@ -94,29 +103,49 @@ BOOL MessageQueue_Wait(wMessageQueue* queue)
...
+		 * allocated space.
+		 */
 		if (queue->tail <= queue->head)
 		{
-			CopyMemory(&(queue->array[old_capacity]), queue->array, queue->tail * sizeof(wMessage));
-			queue->tail += old_capacity;
+			size_t tocopy = queue->tail;
+			size_t slots = new_capacity - old_capacity;
+			const size_t batch = (tocopy < slots) ? tocopy : slots;
+			CopyMemory(&(queue->array[old_capacity]), queue->array, batch * sizeof(wMessage));
```
mfleisz, line 137: The queue will be filled with garbage data after this path is executed (after `CopyMemory` and `memmove_s`). It seems the queue API zeroes memory everywhere - even though I wonder why as there is no sensible data in the allocated memory block and this is just a waste of cycles imho.

winpr/libwinpr/utils/collections/Queue.c
```diff
@@ -45,6 +45,30 @@ struct s_wQueue
 	BYTE padding2[4];
 };
 
+static inline void* uptr2void(uintptr_t ptr)
```
mfleisz, line 48: Use a simple (void*) cast

winpr/libwinpr/utils/collections/Queue.c
```diff
@@ -165,27 +206,47 @@ static BOOL Queue_EnsureCapacity(wQueue* queue, size_t count)
...
 		if (queue->tail <= queue->head)
 		{
-			CopyMemory((void*)&(queue->array[old_capacity]), (void*)queue->array,
-			           queue->tail * sizeof(void*));
-			queue->tail += old_capacity;
+			const size_t tocopy = queue->tail;
+			const size_t slots = new_capacity - old_capacity;
+			const size_t batch = (tocopy < slots) ? tocopy : slots;
+
+			CopyMemory(&(queue->array[old_capacity]), queue->array, batch * sizeof(uintptr_t));
```
mfleisz, line 224: Like in MessageQueue this branch will leave garbage in the queue array

winpr/libwinpr/utils/collections/Queue.c
```diff
@@ -325,9 +376,10 @@ wQueue* Queue_New(BOOL synchronized, SSIZE_T capacity, SSIZE_T growthFactor)
 	if (!queue->event)
 		goto fail;
 
-	obj = Queue_Object(queue);
-	obj->fnObjectEquals = default_queue_equals;
-
+	{
```
mfleisz, line 368: unneeded scope

winpr/libwinpr/utils/collections/Queue.c
```diff
@@ -45,6 +45,30 @@ struct s_wQueue
...
+		const void* cv;
+		void* v;
+		uintptr_t u;
+	} cnv;
+	cnv.u = ptr;
+	return cnv.v;
+}
+
+static inline uintptr_t vptr2uptr(const void* ptr)
+{
```
mfleisz, line 54: Use a simple (uintptr_t) cast

winpr/libwinpr/utils/collections/Queue.c
akallabeth: can change to a macro, but this way it is easier to read (opinionated)

winpr/libwinpr/utils/collections/Queue.c
akallabeth: can change to a macro, but this way it is easier to read (opinionated)

winpr/libwinpr/utils/collections/Queue.c
akallabeth: scopes `wObject* obj`.

mfleisz: What for?

winpr/libwinpr/utils/collections/Queue.c
mfleisz: Not a friend of macros to be honest.
`void* ptr = (void*)queue->array[index];` simply shows a cast whereas `void* ptr = uptr2void(queue->array[index]);` could also indicate some conversion/additional manipulation (not that obvious) - but that's just my opinion.

akallabeth: ok, did use the `inline ` function to make it:
1. obvious why this is being done
2. easier to search for all locations where this is used

so, can eliminate the `union` in the inline function, but would like to keep these.

mfleisz: Ok - just one more suggestion - the naming should be consistent (`vptr2uptr` and `uptr2vptr` or `void2uptr` and `uptr2void`)

PR #12339, opened 2026-02-20
[client,sdl] ensure primary monitor is set

### LINKED ISSUE #12292: sdl-freerdp  multimonitor does not work when primary monitor is not included in the monitors list
Hello,

I am unable to start sld-freerdp in multi monitor config.

My /list:monitor output:

listing 3 monitors:
     * [3] [Dell Inc. 24"] 2560x1440	+0+187
       [4] [Huawei Technologies Co., Inc. 28"] 3840x2560	+2560+0
       [5] [LG Electronics 27"] 3840x2160	+6400+0

it works when when monitor 3(primary) is included in the monitors list
sdl-freerdp   /multimon /monitors:3,4,5  ... 

but when monitor 3 is not included:
sdl-freerdp   /multimon /monitors:4,5  ... 

is fails to start with the error

 [28526:00006f7c] [ERROR][com.freerdp.common] - [freerdp_settings_set_monitor_def_array_sorted]: Could not find primary monitor, aborting

xfreerdp works correctly in that scenario.
I am using version 3.22

Thanks

motor-dev:
Fall back on the first monitor of the list if the user does not add the OS primary monitor in the list of monitors.

fixes https://github.com/FreeRDP/FreeRDP/issues/12292

### DIFF
--- a/client/SDL/SDL3/sdl_monitor.cpp
+++ b/client/SDL/SDL3/sdl_monitor.cpp
@@ -198,11 +198,19 @@ int sdl_list_monitors([[maybe_unused]] SdlContext* sdl)
 		}
 		return TRUE;
 	}
+
+	bool foundPrimary = false;
 	for (const auto& id : sdl->monitorIds())
 	{
 		const auto monitor = sdl->getDisplay(id);
+		foundPrimary |= monitor.is_primary != 0;
 		monitors.emplace_back(monitor);
 	}
+
+	if (!foundPrimary && !monitors.empty())
+		/* The user did not add the OS primary monitor, so set the first monitor as primary */
+		monitors.front().is_primary = TRUE;
+
 	return freerdp_settings_set_monitor_def_array_sorted(settings, monitors.data(),
 	                                                     monitors.size());
 }


akallabeth (changes requested):
sorry, this is a hack and no fix.

* the issue you reference is rooted in the fact that you need to send a full monitor layout to the remote windows (which needs to conform to windows standards, no gaps, all monitors connected via at least 1 pixel)
* you need to map mouse and touch events if you do not send your full local layout
* you need to map back the screen updates to the ones you have

I doubt this will be solved with less than 10 lines of code ;)

motor-dev:
I think we are talking about different problems then. There's a very simple bug that if the user does not include the primary monitor in their monitor list, freerdp will just shut down. 
listing 2 monitors:
```
     * [3] [ASUSTek COMPUTER INC - VG27AQML1A - DP-1] 2560x1440	+0+560
       [4] [ASUSTek COMPUTER INC - VG27AQML1A - HDMI-A-2] 1440x2560	+2560+0
```
       
launching `sdl-freedrp +multimon /monitors:4` it just shuts down because the primary display (3) is not passed on the command line.

What both X11 and SDL2 are doing is to find a monitor that can be set as the primary one. Both x11 and SDL2 keep track if they have seen the primary monitor when pushing into the monitor list, and if not will add the first monitor that the user picked as the primary monitor.
I just made sure the SDL3 backend also does that so there's always a monitor considered the primary monitor in the list.

This only goes over the first hurdle (is there a primary monitor) I am not sure what happens if I pass it monitors that have gaps.

I have been using multimon on the last head revision and have not noticed any problem with mouse mapping or screen updates.

akallabeth:
@motor-dev no, this is exactly the issue I was describing.
you just picked a use case that *might* work with your fix (although this is just `/f` - fullscreen then it works as well) (therefore *hack*), but it is actually the same problem as above.

motor-dev:
alright I am afraid I can't take the full fix as I don't really understand enough of the code.

PR #12348, opened 2026-02-22
sdl-parent feature add

treytesoro:
Example usage from winform:

```
using System;
using System.Collections.Generic;
using System.ComponentModel;
using System.Data;
using System.Diagnostics;
using System.Drawing;
using System.Linq;
using System.Runtime.InteropServices;
using System.Text;
using System.Threading;
using System.Threading.Tasks;
using System.Windows.Forms;

namespace ParentWinTest
{
    public partial class Form1 : Form
    {
        // Add your custom message handling logic here
        const int WM_USER = 0x400;
        const int SDL_PARENT_UPDATED = WM_USER + 300;
        const int SDL_PARENT_RESIZED = WM_USER + 200;
        const int SDL_PARENT_FOCUSED = WM_USER + 201;
        const int SDL_PARENT_CLOSING = WM_USER + 202;

        const int WM_SYSCOMMAND = 0x0112;
        const int SC_MINIMIZE = 0xF020;
        const int WM_PARENTNOTIFY = 0x0210;

        private IntPtr childPtr = IntPtr.Zero;
        public Form1()
        {
            InitializeComponent();
        }

        private void Form1_Load(object sender, EventArgs e)
        {
            var hdl = this.Handle;
            Debug.WriteLine($"Handle: {hdl}");
        }

        protected override void WndProc(ref Message m)
        {
            if (m.Msg == WM_SYSCOMMAND)
            {
                if (m.WParam.ToInt32() == SC_MINIMIZE)
                {
                    Console.WriteLine("Form is minimizing!");
                }
            }

            if (m.Msg == WM_PARENTNOTIFY)
            {
                Debug.WriteLine("WM_PARENTNOTIFY");
            }

            if (m.Msg == SDL_PARENT_UPDATED)
            {
                Debug.WriteLine($"SDL_PARENT_UPDATED. Child handle is {m.WParam}");
                childPtr = m.WParam;
                Task.Run(() => {
                    Thread.Sleep(2000);
                    this.Invoke(new Action (() => { 
                        SynchSize();
                    }));
                });
            }

            base.WndProc(ref m);
        }

        private void Form1_ResizeEnd(object sender, EventArgs e)
        {
            if (childPtr != IntPtr.Zero)
            {
                SynchSize();
            }
        }

        private void SynchSize()
        {
            int w = this.ClientSize.Width;
            int h = this.ClientSize.Height;

            int wh = w << 16 | h;
            Win32.PostMessage(childPtr, SDL_PARENT_RESIZED, (IntPtr)wh, IntPtr.Zero);
        }

        private void Form1_SizeChanged(object sender, EventArgs e)
        {
            if (childPtr != IntPtr.Zero)
            {
                SynchSize();
            }
        }

        private void Form1_Activated(object sender, EventArgs e)
        {
            Win32.PostMessage(childPtr, SDL_PARENT_FOCUSED, IntPtr.Zero, IntPtr.Zero);
        }

        private void Form1_FormClosing(object sender, FormClosingEventArgs e)
        {
            Win32.PostMessage(childPtr, SDL_PARENT_CLOSING, IntPtr.Zero, IntPtr.Zero);
        }
    }


    class Win32
    {
        [DllImport("user32.dll")]
        public static extern Boolean PostMessage(IntPtr hwnd, Int32 msg, IntPtr wParam, IntPtr lParam);
    }
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,10 @@
 build
 checker
 abi-checker
+
+# ignore treytesoro scripts and artifacts
+.vscode/
+**/*.log
+winscripts/
+/freerdpinstall
+/libs
--- a/client/SDL/SDL3/sdl_context.cpp
+++ b/client/SDL/SDL3/sdl_context.cpp
@@ -452,16 +452,37 @@ bool SdlContext::createWindows()
 		auto did = WINPR_ASSERTING_INT_CAST(SDL_DisplayID, id);
 		auto window = SdlWindow::create(did, title, flags, w, h);
 
+		const auto parent = freerdp_settings_get_uint64(settings, FreeRDP_ParentWindowId);
+		bool isParentSet = FALSE;
+		if (parent != 0)
+		{
+			isParentSet = window.setParent(parent);
+			if (!isParentSet)
+				return false;
+		}
+
 		if (freerdp_settings_get_bool(settings, FreeRDP_UseMultimon))
 		{
 			window.setOffsetX(originX - monitor->x);
 			window.setOffsetY(originY - monitor->y);
 		}
 
+		SDL_WindowID winId = window.id(); // get value now before move ctor nulls window
+
 		_windows.insert({ window.id(), std::move(window) });
+
+		// in case of failure, do not apply hook.
+		if(isParentSet) {
+			_mainWindowID = winId;
+			SDL_SetWindowsMessageHook(SdlWindow::SDLMessageHook, reinterpret_cast<void*>(this));
+		}
 	}
 
 	return true;
+}
+
+SDL_WindowID SdlContext::getMainWindowId() {
+	return _mainWindowID;
 }
 
 bool SdlContext::updateWindowList()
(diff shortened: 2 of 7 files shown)


akallabeth:
@treytesoro wouldn't it be sufficient to create the window without borders in case a parent is to be set?

treytesoro:
> @treytesoro wouldn't it be sufficient to create the window without borders in case a parent is to be set?

:) That is honestly how I originally had it before you did the sdl-parent work.  That can be reapplied, but is there ever a use-case where a developer would want to re-parent a window later?  Say for docking, etc....  For my own rdc manager type app I'm working on, I want to be able to dock/undock from it.  With how resizing seems to reapply the border, hence the reason for reapplying the child style, I'm worried the reverse would be true as well and we'd end up with a borderless top level window.

akallabeth:
> > @treytesoro wouldn't it be sufficient to create the window without borders in case a parent is to be set?
> 
> :) That is honestly how I originally had it before you did the sdl-parent work. That can be reapplied, but is there ever a use-case where a developer would want to re-parent a window later? Say for docking, etc.... For my own rdc manager type app I'm working on, I want to be able to dock/undock from it. With how resizing seems to reapply the border, hence the reason for reapplying the child style, I'm worried the reverse would be true as well and we'd end up with a borderless top level window.

well, if there is then the dev should step up and make some noise :grin: 
I'd like to keep platform specific code to a minimum unless explicitly required. (tends to be hard to maintain)

PR #12374, opened 2026-02-25
[c23] replace NULL

akallabeth:
* Add a fallback macro for nullptr to NULL
* Replace NULL with C23 nullptr


akallabeth:
split up to smaller ones.

PR #12375, opened 2026-02-25
[core,gateway] ignore incomplete rpc header

akallabeth:


akallabeth:
@MikhailN75 this was a bug introduced with 3.22.0, the header parsing must not abort on incomplete data received.

MikhailN75:
I just tested latest master and it works well. Thanks again

PR #12377, opened 2026-02-26
[warnings] make function declaration names consistent

akallabeth:
Use the same parameter names for declaration and implementation.


akallabeth:
@freerdp-bot test

PR #12380, opened 2026-02-26
[libfreerdp] Add new define for logon error info

mfleisz:


akallabeth (changes requested):
incomplete.

* see `freerdp_get_logon_error_info_type`
* if you update this, please add all missing options

Issue #12393, opened 2026-02-27
[wlroots] sdl3-freerdp multimon broken on since last update


obi1:
Commit 0ff34a11248927627d505f2e45b8d4eb7b5ada1f broke multimon for me under Sway.

After that commit I'm getting:
```
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [freerdp_settings_client_monitors_check_primary_and_origin]: Monitor configuration virtual desktop width must be 200 <= 128 <= 32766
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [freerdp_settings_client_monitors_check_primary_and_origin]: Monitor configuration virtual desktop height must be 200 <= 64 <= 32766
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [freerdp_settings_check_client_after_preconnect]: Invalid or unsupported monitor configuration detected
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [freerdp_settings_check_client_after_preconnect]: Check if the configuration is valid.
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [freerdp_settings_check_client_after_preconnect]: If you suspect a bug create a new issue at https://github.com/FreeRDP/FreeRDP/issues/new
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [freerdp_settings_check_client_after_preconnect]: Provide at least the following log lines detailing your monitor configuration:
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [log_monitor_configuration]: [BEGIN] MonitorDefArray[2]
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [log_monitor]: [0] [primary] {0x0-64x64} [3] {64x64, orientation: ORIENTATION_LANDSCAPE [0], desktopScale: 100, deviceScale: 100}
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [log_monitor]: [1] [       ] {-64x0-64x64} [4] {64x64, orientation: ORIENTATION_LANDSCAPE [0], desktopScale: 100, deviceScale: 100}
[08:50:11:206] [192448:0002efde] [ERROR][com.freerdp.settings] - [log_monitor_configuration]: [END] MonitorDefArray[2]
[...]

PR #12398, opened 2026-02-27
[client,sdl] workaround for wlroots compositors

obi1:
workaround for #12393

### DIFF
--- a/client/SDL/SDL3/sdl_window.cpp
+++ b/client/SDL/SDL3/sdl_window.cpp
@@ -24,6 +24,7 @@
 #include "sdl_window.hpp"
 #include "sdl_utils.hpp"
 
+#include <freerdp/channels/disp.h>
 #include <freerdp/utils/string.h>
 
 SdlWindow::SdlWindow(SDL_DisplayID id, const std::string& title, const SDL_Rect& rect,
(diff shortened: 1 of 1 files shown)


client/SDL/SDL3/sdl_window.cpp
```diff
@@ -346,6 +347,32 @@ SDL_Rect SdlWindow::rect(SDL_Window* window, bool forceAsPrimary)
...
+	// don't get their actual display dimensions. The dummy window returns its creation size (64x64)
+	// instead of the display size. This causes validation errors since we require >= 200px.
+	// Workaround: If we got dimensions that are too small, query the display directly.
+	if (rect.w < DISPLAY_CONTROL_MIN_MONITOR_WIDTH || rect.h < DISPLAY_CONTROL_MIN_MONITOR_HEIGHT)
+	{
+		const auto displayID = SDL_GetDisplayForWindow(window);
+		SDL_Rect displayBounds = {};
+		if (SDL_GetDisplayBounds(displayID, &displayBounds))
+		{
+			if (forceAsPrimary)
```
akallabeth, line 360: remove, this is dead code. `rect` is initialized.

client/SDL/SDL3/sdl_window.cpp
```diff
@@ -346,6 +347,32 @@ SDL_Rect SdlWindow::rect(SDL_Window* window, bool forceAsPrimary)
...
+			{
+				rect.x = 0;
+				rect.y = 0;
+			}
+			rect.w = displayBounds.w;
+			rect.h = displayBounds.h;
+
+			const float contentScale = SDL_GetDisplayContentScale(displayID);
```
akallabeth, line 368: the display scale factor might not be the same as the actual scaling setting.
can at least this be reliably queried from the `window` instead of the `display`?

client/SDL/SDL3/sdl_window.cpp
```diff
@@ -346,6 +347,32 @@ SDL_Rect SdlWindow::rect(SDL_Window* window, bool forceAsPrimary)
 	if (!SDL_GetWindowSizeInPixels(window, &rect.w, &rect.h))
 		return {};
 
+	// On wlroots compositors (Sway, river, etc.), windows that are hidden/unmapped
+	// don't get their actual display dimensions. The dummy window returns its creation size (64x64)
+	// instead of the display size. This causes validation errors since we require >= 200px.
+	// Workaround: If we got dimensions that are too small, query the display directly.
+	if (rect.w < DISPLAY_CONTROL_MIN_MONITOR_WIDTH || rect.h < DISPLAY_CONTROL_MIN_MONITOR_HEIGHT)
```
akallabeth, line 354: would guard the whole thing with [SDL_GetPlatform](https://wiki.libsdl.org/SDL3/SDL_GetPlatform) and [SDL_GetCurrentVideoDriver](https://wiki.libsdl.org/SDL3/SDL_GetCurrentVideoDriver)

client/SDL/SDL3/sdl_window.cpp
obi1: I tried different scaling and it seemed to work. I will try getting it from windows. Probably won't be able to check before monday.

client/SDL/SDL3/sdl_window.cpp
obi1: I can do that if you think a general fallback for similar cases elsewhere isn't necessary.

Someone complained of same issue on Hyprland and they moved away from wlroots a while ago.

client/SDL/SDL3/sdl_window.cpp
akallabeth: https://wiki.libsdl.org/SDL3/README-highdpi

akallabeth:
~~ok, can not reproduce the issue here on `debian 13/sway` 
when was this introduced? not sure we should add workarounds for uncommon bugs that are only in specific versions.~~

ok, `SDL_VIDEODRIVER=wayland` is required, defaults to `x11` here even with `sway`

PR #12425, opened 2026-03-04
[client,sdl] workaround for wlroots compositors

akallabeth:
Replaces #12398


akallabeth:
@obi1 did take your commit and tailor it to sway. Feel free to add additional variables to check.
this workaround should only be applied if absolutely necessary.

PR #12433, opened 2026-03-05
[client] Fix writing of gatewayusagemethod to .rdp files

mfleisz:
When loading the gatewayusagemethod setting from a .rdp file we use `freerdp_settings_set_uint32` to set FreeRDP's gateway parameters (`GatewayEnable` and `GatewayBypassLocal`).

However when writing settings back to a .rdp file we ignore these settings. This PR introduces `freerdp_get_gateway_usage_method` that will take internal gateway paramters into account and return the corresponding gatewayusagemethod.

### DIFF
--- a/include/freerdp/settings.h
+++ b/include/freerdp/settings.h
@@ -285,6 +285,8 @@ extern "C"
 
 	FREERDP_API BOOL freerdp_set_gateway_usage_method(rdpSettings* settings,
 	                                                  UINT32 GatewayUsageMethod);
+	WINPR_ATTR_NODISCARD
+	FREERDP_API UINT32 freerdp_get_gateway_usage_method(rdpSettings* settings);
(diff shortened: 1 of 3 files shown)


include/freerdp/settings.h
```diff
@@ -285,6 +285,7 @@ extern "C"
 
 	FREERDP_API BOOL freerdp_set_gateway_usage_method(rdpSettings* settings,
 	                                                  UINT32 GatewayUsageMethod);
+	FREERDP_API UINT32 freerdp_get_gateway_usage_method(rdpSettings* settings);
```
akallabeth, line 289: add `WINPR_ATTR_NODISCARD` (public `API` should warn if a getter is called uselessly)

akallabeth: but looking at the header, can do that in another pr.

PR #12466, opened 2026-03-11
Smartcard fix smartcard-login, pass rdpContext for abort

akallabeth:


akallabeth:
@ktostam could you give this a try? should fix your problem.

ktostam:
At first glance, it looks OK.  No more locks in pcscd logs... everything seems to be working. Thank you very much for quickly diagnosing and fixing the problem.

PR #12497, opened 2026-03-16
[winpr,wlog] add WLog_SetGlobalPrefix

akallabeth:
Allow setting an application wide logger prefix to easily distinguish different application log entries.


akallabeth:
@freerdp-bot test

PR #12500, opened 2026-03-17
[channels,audin] set error when audio_format_read fails

ondrejholy:
Currently, `goto fail` is called when `audio_format_read` fails, but `error` is not changed, so `CHANNEL_RC_OK` is returned. Before the 1c5c7422 commit, `ERROR_INVALID_DATA` was returned. Let's again set that error code and also restore the debug print.


akallabeth:
@freerdp-bot test
@ondrejholy hi, thank you for the pr!

PR #12504, opened 2026-03-17
[core,update] make the PlaySound callback non-mandatory

thenextman:
If the server sends a "Play Sound" PDU but the client does not subscribe `update->PlaySound` it forces a disconnection.

I have only seen this in the wild with xrdp running the VNC backend.

All of freerdp's own clients (with the exception of `xfreerdp`) subscribe this callback but make it a no-op.

This PR makes the callback "opt in" (a client can safely leave it unsubscribed and they will not be disconnected if it's triggered).


akallabeth:
@freerdp-bot test

akallabeth:
nice one, thank you!

PR #12510, opened 2026-03-18
[codec,openh264] reject encoder ABI mismatch on runtime-loaded library

thenextman:
OpenH264 apparently doesn't follow semver, because there are ABI changes in the _encoder_ between 2.4.1 and 2.6.0 (versions tested).

This change adds a sanity check if the client is using `WITH_OPENH264_LOADING`; for encoding, the runtime version of the library must match the version we compiled against.

Decoding is unaffected. It's implicitly assumed that we're building against at least 1.6.0, and if not the preprocessor checks for `OPENH264_MAJOR` and `OPENH264_MINOR` activate different code paths anyway.

### DIFF
--- a/libfreerdp/codec/h264_openh264.c
+++ b/libfreerdp/codec/h264_openh264.c
@@ -565,6 +565,18 @@ static BOOL openh264_init(H264_CONTEXT* h264)
 
 		if (h264->Compressor)
 		{
+#if defined(WITH_OPENH264_LOADING)
+			if (sysContexts->version.uMajor != OPENH264_MAJOR ||
+			    sysContexts->version.uMinor != OPENH264_MINOR)
+			{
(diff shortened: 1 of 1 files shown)


akallabeth:
@thenextman well, from the diff  it looks like it is `sem version` compatible? (e.g. new values at end of struct)

akallabeth:
ah, ok, `SLayerBSInfo` is the problematic one

thenextman:
@akallabeth Yes, `SFrameBSInfo` is the problem, and I don't see a pragmatic solution other than forcing the versions to match. I suppose not many people have hit this (I imagine that dynamic OpenH264 loading with a fully static `freerdp` build and RDPECAM might not be a very common setup).

akallabeth:
@freerdp-bot test
@thenextman well, I suppose it is good enough then :)

PR #12511, opened 2026-03-18
[channels,video] fix wrong cast

akallabeth:
IWTSVirtualChannelCallback should be cast to GENERIC_CHANNEL_CALLBACK pointer, not GENERIC_LISTENER_CALLBACK


akallabeth:
@ondrejholy @mjt0k looks like I did mess up https://github.com/FreeRDP/FreeRDP/pull/12502

PR #12513, opened 2026-03-19
Fix Queue.c to make stable-2.0 build

joshua2a:
Stable 2.0 does not build. Revert to void pointers.


akallabeth:
I assume you're trying newer `gcc` versions?

akallabeth:
btw, you should migrate to 3.x soon, this branch is out of support.

PR #12520, opened 2026-03-20
[utils, signal] lazily initialize Windows CRITICAL_SECTION to match POSIX static mutex behavior

thenextman:
#12335 refactored signal handlers per-platform. On Windows, a critical section is used for locking and is initialized in `freerdp_handle_signals`.

_If_ the client does not call `freerdp_handle_signals`, this causes a segfault as soon as we try to lock the critical section, e.g.

```
fsig_lock()
freerdp_add_signal_cleanup_handler(void * context, void(*)(int, const char *, void *) handler
freerdp_connect_begin(rdp_freerdp * instance) 
freerdp_connect(rdp_freerdp * instance)
myclient_connect()
```

Note that when we're consuming freerdp as a library, we might not want it to take ownership of process-wide signal handling.

On the POSIX side, `PTHREAD_MUTEX_INITIALIZER` is a compile time constant. This PR uses `INIT_ONCE_STATIC_INIT` to achieve the same result on Windows.

### DIFF
--- a/libfreerdp/utils/signal_win32.c
+++ b/libfreerdp/utils/signal_win32.c
@@ -15,9 +15,17 @@
 #define TAG FREERDP_TAG("utils.signal.posix")
 
 static CRITICAL_SECTION signal_lock;
+static INIT_ONCE signal_lock_init = INIT_ONCE_STATIC_INIT;
+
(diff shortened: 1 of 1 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12527, opened 2026-03-23
RFC: Research/av1 codec extension

akallabeth:
This pr is proposing an extension to the RDP protocol to allow the use of `AV1` codec.

@pnowack @hardening 

What does it provide:
* `AV1` encoding/decoding with `AOM`
* Add missing `YUV` routines
* Client decoder and command line 
* Shadow encoder and command line

Missing:
* add encoder/decoder with hardware acceleration support

### DIFF
--- a/client/X11/xf_gfx.c
+++ b/client/X11/xf_gfx.c
@@ -443,6 +443,9 @@ static UINT xf_DeleteSurface(RdpgfxClientContext* context,
 #ifdef WITH_GFX_H264
 		h264_context_free(surface->gdi.h264);
 #endif
+#if defined(WITH_GFX_AV1)
+		freerdp_av1_context_free(surface->gdi.av1);
+#endif
 		surface->image->data = nullptr;
 		XDestroyImage(surface->image);
 		winpr_aligned_free(surface->gdi.data);


include/freerdp/codec/av1.h
```diff
@@ -0,0 +1,89 @@
...
+
+	typedef struct S_FREERDP_AV1_CONTEXT FREERDP_AV1_CONTEXT;
+
+	typedef enum
+	{
+		FREERDP_AV1_CONTEXT_OPTION_PROFILE,
+		FREERDP_AV1_CONTEXT_OPTION_RATECONTROL,
+		FREERDP_AV1_CONTEXT_OPTION_BITRATE,
+		FREERDP_AV1_CONTEXT_OPTION_FRAMERATE,
+		FREERDP_AV1_CONTEXT_OPTION_QP,
```
pnowack: `FREERDP_AV1_CONTEXT_OPTION_RATECONTROL`, `FREERDP_AV1_CONTEXT_OPTION_BITRATE`, `FREERDP_AV1_CONTEXT_OPTION_FRAMERATE`, and `FREERDP_AV1_CONTEXT_OPTION_QP` shouldn't be needed, as these are encoder settings only. The decoder side does not care about it.

However, there are some settings, were both encoder and decoder side have constraints, like:
- supported format, e.g. YUV420, YUV420_10BPP (currently only supported ones on Intel and AMD GPUs; I suppose Intel and NVIDIA will support 444 native variants in the future)
- max picture width and height

(not sure whether there are further things that we will only know by reading the AV1 spec, but these two above are definitely relevant for Vulkan Video and VA-API)

include/freerdp/channels/rdpgfx.h
```diff
@@ -157,6 +157,7 @@ typedef struct
  */
 
 #define RDPGFX_CODECID_UNCOMPRESSED 0x0000
+#define RDPGFX_CODECID_AV1 0x0001
```
pnowack: Should probably have a short comment, that this our custom codec (and currently is only valid for `RDPGFX_CAPVERSION_FRDP_1`)

include/freerdp/codec/av1.h
akallabeth: @pnowack we only care about `YUV420` and `YUV444`, the 10 bit variants are currently out of scope (maybe a future extension?)

akallabeth: what I'm currently not completely sure is how the stream parameters are passed.
with the `H264` modes the first frame contains the stream parameters that are required to configure the decoder.
with `AV1` it works, but not sure this is just because we use the same encoder/decoder implementation with the same defaults or because the stream has these encoded. (research/todo)

pnowack: > what I'm currently not completely sure is how the stream parameters are passed.

They are part of the Open Bitstream Units (https://aomediacodec.github.io/av1-spec/#obu-syntax). It's kinda the equivalent of NAL units for HEVC and AVC. So, they are directly part of the stream. Which is also why the supported max size of the decoder needs to be communicated to the server side, as when the stream arrives, it's already too late.

> with AV1 it works, but not sure this is just because we use the same encoder/decoder implementation 

Usually, software encoder/decoder implementations already take care of the stream parameter writing (here: Open Bitstream Units), but when things go in direction to hardware acceleration, they need to be manully parsed (though FFMPEG should have that part implemented).

include/freerdp/channels/rdpgfx.h
```diff
@@ -131,6 +131,13 @@ typedef struct
 #define RDPGFX_CAPS_FLAG_AVC_THINCLIENT 0x00000040U    /* 10.3+ */
 #define RDPGFX_CAPS_FLAG_SCALEDMAP_DISABLE 0x00000080U /* 10.7+ */
 
+#define RDPGFX_CAPS_FLAG_AV1_I444_SUPPORTED                                                     \
+RDPGFX_CAPS_FLAG_AVC420_ENABLED /** Custom Extension: Onyl valid if RDPGFX_CAPVERSION_FRDP_1 is \
```
pnowack: The flag value must differ, otherwise this causes problems with distributions like Ubuntu or Fedora, as they don't have AVC support, but they can have AV1 support (because: royalty-free). It also makes it easier for the server side to dynamically choose what codec would fit better for a situation.

include/freerdp/channels/rdpgfx.h
```diff
@@ -131,6 +131,13 @@ typedef struct
...
+	                               used */
+#define RDPGFX_CAPS_FLAG_AV1_I444_DISABLED                                                         \
+RDPGFX_CAPS_FLAG_AVC_DISABLED /** Custom Extension: Onyl valid if RDPGFX_CAPVERSION_FRDP_1 is used \
```
pnowack: Same here: The flag value must differ, otherwise this causes problems with distributions like Ubuntu or Fedora, as they don't have AVC support, but they can have AV1 support

include/freerdp/channels/rdpgfx.h
akallabeth: does not matter, just a number. (did this so it is obvious the numbers are the same)

include/freerdp/channels/rdpgfx.h
akallabeth: see above

include/freerdp/codec/av1.h
akallabeth: well, been bitten by `fdk-aac` so tend to be careful there ;)

include/freerdp/channels/rdpgfx.h
pnowack: It does. If e.g. a client on Fedora or Ubuntu connects, advertises `RDPGFX_CAPS_FLAG_AV1_I444_SUPPORTED`, because it supports AV1, does not mean that AV1 can used. So if then e.g. the server side sees that no GPU supports AV1, it can fallback to AVC, so it checks for `RDPGFX_CAPS_FLAG_AVC420_ENABLED`. Since it is enabled (too), it will try to use AVC, although the client side does not support it. The server side cannot here differentiate between these two cases, because the flags share the same value and that is the problem.

akallabeth: @pnowack no. the flags are only valid with regard to the capability version.

akallabeth: so, while the `10_x` capabilities share the same flags, the `8_x` ones have different ones.
and now our custom extension is no different from that.

pnowack: > no. the flags are only valid with regard to the capability version.

What will then happen with the AVC capability with `RDPGFX_CAPVERSION_FRDP_1`? Just because `RDPGFX_CAPVERSION_FRDP_1` was negotiated does not mean that AV1 can always be used (e.g. due to encoder contraints), so other codecs need to be allowed too. E.g. for multi monitor setups: For one monitor AV1 can be used mid-session, but for the other one e.g. AV1 cannot be used. Then the server has to fall back to something else. Can it then assume AVC is possible too or not?

akallabeth: you do not select that caps unless you support `AV1`?
the idea is you just announce it client side, if you can decode it. (with the `RDPGFX_CAPS_FLAG_AV1_I444_SUPPORTED` flag to signal you can decode `I444`)
and the server then checks if it supports `AV1` and `I444` and selects the capability with the appropriate flag.

akallabeth: so, if you end up having selected the `RDPGFX_CAPVERSION_FRDP_1` you already know client and server support `AV1`  (with `I420` always being available and `I444` depending on flags negotiated.

pnowack: > you do not select that caps unless you support AV1?

Well, the server side would only select the caps set, when supported. But that does not mean that creating an AV1 encode session would not fail later. For example, NVENC only allows creating (globally on the whole computer) maximum of 8 NVENC sessions (older drivers even less). So, the GPU and driver support AV1, but you never know whether that limit is hit and then a fallback to another codec is needed. Both GPU and driver support it, but whether creating that encode session is successful is another story.
Another situation would be if encoding AV1 is supported by the server side, but the client side requested a resolution for which AV1 is not supported, but maybe via AVC. The server side does not know what resolution the client chooses in the future and it also can change mid-session via the DISPLAY channel.
So, AV1 can be supported, but it is not always possible to use it within a session. Sometimes only for some monitors (e.g. one monitor AV1, the second one AVC). AVC is the next best option for these fallback situations. But AVC can only be used if the client side actually supports (which is not the case on e.g. Fedora and Ubuntu, but AV1 is).

akallabeth: ok, the current implementation is more or less `AV1` or `RFX/progressive` as fallback.

* regarding `GPU` reaches the `AV1` encoder limit, then there is always software encoder fallback (yes, slow, but that will be always slow)
* resolution restrictions of `H264` and `AV1` are usually the same, but you're right, never say never :grin:

akallabeth: in that case we can just use some unused flag constants instead.

PR #12529, opened 2026-03-23
fix locking in shadow_lobby.c

ggirol-rc:
when WITH_RDTK is not defined, shadow_client_init_lobby unlocks a lock which has never been locked.

This regression was introduced in f4c991f902882ae46b4ab883964c02a5649bad1d the commit making RDTK optional.

### DIFF
--- a/server/shadow/shadow_lobby.c
+++ b/server/shadow/shadow_lobby.c
@@ -40,12 +40,12 @@ BOOL shadow_client_init_lobby(rdpShadowServer* server)
 	if (!lobby)
 		return FALSE;
 
+	EnterCriticalSection(&lobby->lock);
 #if defined(WITH_RDTK)
 	rdtkEngine* engine = rdtk_engine_new();
 	if (!engine)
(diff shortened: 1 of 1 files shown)


server/shadow/shadow_lobby.c
```diff
@@ -40,12 +40,12 @@ BOOL shadow_client_init_lobby(rdpShadowServer* server)
 	if (!lobby)
 		return FALSE;
 
+	EnterCriticalSection(&lobby->lock);
 #if defined(WITH_RDTK)
 	rdtkEngine* engine = rdtk_engine_new();
 	if (!engine)
 		return FALSE;
```
akallabeth, line 47: you should unlock it here as well or you just moved from one locking issue to another.

akallabeth:
ok, cherry-picked your commit, fixed the locking and added to https://github.com/FreeRDP/FreeRDP/pull/12528 

ggirol-rc:
Thank you !

PR #12530, opened 2026-03-23
winpr: improve libunwind backtraces

ggirol-rc:
Even in the ideal case where dladdr succeeds, the message used to display the base address of the library (to undo ASLR) and the starting address of the function, but not the address which triggered the assertion. As a result, it's impossible to go back to a line number even with debug symbols and addr2line.

If we display this address, then it becomes possible.


akallabeth:
@freerdp-bot test

PR #12534, opened 2026-03-23
Fix process exit code on non-pidfd platforms (macOS, BSD)

### LINKED ISSUE #12533: WinPR CreateProcess: wrong exit code and wait loop issues on non-pidfd platforms (macOS, BSD)
## Summary

On platforms without `pidfd` support (macOS, BSD), `WaitForSingleObject` for process handles uses a `waitpid` polling loop that has two bugs:

### Bug 1: Raw waitpid status used as exit code

`ProcessCleanupHandle()` in `process.c` and the non-pidfd wait path in `wait.c` store the raw `waitpid` status directly as the exit code:

```c
process->dwExitCode = (DWORD)process->status;
```

On Unix, the raw `waitpid` status encodes the exit code in bits 8-15. A process exiting with code 1 produces raw status `0x0100` (256), which gets stored as `dwExitCode = 256` instead of `1`. The fix is to use `WEXITSTATUS()`:

```c
if (WIFEXITED(process->status))
    process->dwExitCode = (DWORD)WEXITSTATUS(process->status);
else if (WIFSIGNALED(process->status))
    process->dwExitCode = (DWORD)(128 + WTERMSIG(process->status));
```

**Affected files:**
- `winpr/libwinpr/synch/wait.c` line 162
- `winpr/libwinpr/thread/process.c` line 554

### Bug 2: Wait loop exits without final waitpid check

The polling loop in `WaitForSingleObjectEx` (non-pidfd path) decrements `dwMilliseconds` by 50ms slices:

```c
} while (dwMilliseconds > 50);
return WAIT_TIMEOUT;
```

When `dwMilliseconds` reaches 50, the loop exits without a final `waitpid` check after the last sleep. A process that finishes during the last 50ms slice is reported as `WAIT_TIMEOUT` even though it has exited.

**Affected file:** `winpr/libwinpr/synch/wait.c` lines 155-185

### Observed in

`TestThreadCreateProcess` fails on macOS with:
```
Process exited with code: 0x00000100   (should be 0x00000000)
```

After fixing the WEXITSTATUS extraction, exit code becomes `0x00000001` (properly extracted). The exit code 1 itself appears to be a separate issue with how `EnvironmentBlockToEnvpA` or pipe fd handling works on macOS — tracked separately.

### Environment

- FreeRDP 3.24.1
- macOS 15.x (arm64)
- No pidfd support → non-pidfd wait path used

Kotivskyi:
## Summary

Fix two bugs in the non-pidfd `WaitForSingleObject` path for process handles, affecting macOS and BSD.

## Changes

**`winpr/libwinpr/synch/wait.c`** — Non-pidfd wait loop:
- Use `WIFEXITED`/`WEXITSTATUS`/`WIFSIGNALED`/`WTERMSIG` to properly decode waitpid status instead of storing raw status as exit code
- Restructure polling loop: always check `waitpid` before sleeping, check timeout only after failed `waitpid` — prevents missing a process that exits during the last sleep slice

**`winpr/libwinpr/thread/process.c`** — `ProcessCleanupHandle` (pidfd path):
- Same `WEXITSTATUS` fix for the pidfd-based cleanup handler

## Before/After

```
# Before (raw waitpid status leaked as exit code):
Process exited with code: 0x00000100   (should be 0x00000000 for exit(0))

# After (properly decoded):  
Process exited with code: 0x00000001   (correct extraction via WEXITSTATUS)
```

## Test results

Full test suite: **153/154 pass** (same as unpatched baseline).

`TestThreadCreateProcess` still fails — the remaining issue is that `printenv` exits with code 1 on macOS, likely due to a separate bug in `EnvironmentBlockToEnvpA` or pipe fd handling. This PR fixes the exit code *reporting*; the test's underlying `printenv` failure is a pre-existing issue.

All other tests pass, including all thread/synch/process tests.

Fixes #12533

### DIFF
--- a/winpr/libwinpr/synch/wait.c
+++ b/winpr/libwinpr/synch/wait.c
@@ -152,14 +152,17 @@ DWORD WaitForSingleObjectEx(HANDLE hHandle, DWORD dwMilliseconds, BOOL bAlertabl
 		 * 		If not (on other platforms) we do a waitpid */
 		WINPR_PROCESS* process = (WINPR_PROCESS*)Object;
 
-		do
+		while (TRUE)
 		{
-			DWORD status = 0;
-			DWORD waitDelay = 0;
 			int ret = waitpid(process->pid, &(process->status), WNOHANG);
 			if (ret == process->pid)
 			{
-				process->dwExitCode = (DWORD)process->status;
+				if (WIFEXITED(process->status))
+					process->dwExitCode = (DWORD)WEXITSTATUS(process->status);
+				else if (WIFSIGNALED(process->status))
+					process->dwExitCode = (DWORD)(128 + WTERMSIG(process->status));
+				else
+					process->dwExitCode = (DWORD)process->status;
 				return WAIT_OBJECT_0;
 			}
 			else if (ret < 0)
@@ -171,18 +174,22 @@ DWORD WaitForSingleObjectEx(HANDLE hHandle, DWORD dwMilliseconds, BOOL bAlertabl
 				return WAIT_FAILED;
 			}
 
-			/* sleep by slices of 50ms */
-			waitDelay = (dwMilliseconds < 50) ? dwMilliseconds : 50;
+			if (dwMilliseconds == 0)
+				return WAIT_TIMEOUT;
 
-			status = SleepEx(waitDelay, bAlertable);
+			/* sleep by slices of 50ms */
+			DWORD waitDelay = 50;
+			if (dwMilliseconds != INFINITE)
+			{
+				if (dwMilliseconds < 50)
+					waitDelay = dwMilliseconds;
+				dwMilliseconds -= waitDelay;
+			}
+
+			DWORD status = SleepEx(waitDelay, bAlertable);
 			if (status != 0)
 				return status;
-
-			dwMilliseconds -= waitDelay;
-
-		} while (dwMilliseconds > 50);
-
-		return WAIT_TIMEOUT;
+		}
 	}
 
 	if (Type == HANDLE_TYPE_MUTEX)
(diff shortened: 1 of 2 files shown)


Kotivskyi:
Note: I'm not sure how correct this fix is. Fix is generated by Claude Code. Please review carefully

akallabeth:
@Kotivskyi will take some time, this part is (mostly) unused, only `rdp2tcp` channel uses `CreateProcess`.
so need to dig into that first myself.

winpr/libwinpr/synch/wait.c
```diff
@@ -152,14 +152,17 @@ DWORD WaitForSingleObjectEx(HANDLE hHandle, DWORD dwMilliseconds, BOOL bAlertabl
...
-		do
+		while (TRUE)
 		{
-			DWORD status = 0;
-			DWORD waitDelay = 0;
 			int ret = waitpid(process->pid, &(process->status), WNOHANG);
 			if (ret == process->pid)
 			{
-				process->dwExitCode = (DWORD)process->status;
+				if (WIFEXITED(process->status))
```
akallabeth, line 160: Might be a good idea to unify that in a simple (internal) helper function as this is used more than once.

winpr/libwinpr/thread/process.c
```diff
@@ -551,7 +551,14 @@ static DWORD ProcessCleanupHandle(HANDLE handle)
 	if (process->fd > 0)
 	{
 		if (waitpid(process->pid, &process->status, WNOHANG) == process->pid)
-			process->dwExitCode = (DWORD)process->status;
+		{
+			if (WIFEXITED(process->status))
```
akallabeth, line 555: See above

akallabeth (changes requested):
Can you please check `TestThreadCreateProcess` and why it currently passes on mac os although this is not working as you suggest?
(see https://ci.freerdp.com/job/osx/label=macosx/16451/console for an example of a pass on mac os)

[edit] or if that was a misunderstanding, could you add a check for the correct return code if not there?

akallabeth:
@freerdp-bot test

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/12586

Issue #7707, opened 2022-03-07
TODO: smartcard related items


hardening:
NLA smartcard  logon PR has been merged, anyway I'm seeing items that still need work:

* [x] make smartcard emulation a runtime option and not a build option, so that we can choose at runtime if we have real smartcard or the emulation (that a matter of having a struct with all smartcard functions that we would fill with the correct functions);
* [x] implement container name computation under linux and Mac so that NLA smartcard logon works from scratch without having to specify by hand the correct containerName (retrieved by connecting the card under windows). The algorithm involves retrieving the guid of the card for PIV cards (and pick the container id of the cert), and for GIDS cards it's another algorithm (EFS.dir I think)
* [x] make TLS smartcard logon more convenient, like if there's only one smartcard with one certificate, then automatically pick the user and domain from the smartcard

akallabeth:
@hardening Just a note: last option (TLS) already works, that is how we do logon here.
Do I miss something here or is it maybe dependent on device in use?

hardening:
Here if I just run `xfreerdp /v:<target> /sec:tls /smartcard-logon:pin:<pin>` with just one smartcard plugged in a user and a domain are asked, while we could guess them.

akallabeth:
@hardening we use `xfreerdp /v:<target> /sec:tls /smartcard-logon /p:<pin> +password-is-pin` as without the last setting the `info` packet does not send the correct flags.

akallabeth:
@hardening I´d add another item for:
* `CredMarshalCredential` implementation in `WinPR` so we don´t need the `#ifdef _WIN32` switches in `nla.c`
* Switch all functions in `nla.c` to the `*W` (unicode) versions as we don´t want to limit usernames/passwords to `ANSI`

hardening:
There's a fundamental limitation with using `CredMarshalCredential`: with an existing kerberos cache I'm not sure we can ensure that this cache has been established with pkinit by a given certificate (given by SHA1). `CredMarshalCredential` is so much a system oriented things with lots of implicits under windows. 
I was thinking that when it's FreeRDP that is initiating the cache, we could also create a meta data file (cache: /tmp/krb5cc_1000 meta: /tmp/krb5cc_1000.meta) and put the pkinit parameters in it (like slotId and certId) so that we can retrieve them with an emulated `CredMarshalCredential` call in winPR. That would do the trick when it's a cache initiated by FreeRDP but not when you establish your kerberos cache by hand.
[...]

PR #12559, opened 2026-03-30
fix: smartcard logon with ECC keys and minidriver-assigned container names

dko-strd:
## Summary

This fixes two issues that prevent smartcard auto-logon over RDP when using
ECC-based PIV smartcards (e.g. YubiKey with ECCP384 certificates):

1. **Wrong credential provider in TSSmartCardCreds**: The default CSP
   ("Microsoft Base Smart Card Crypto Provider") does not support ECC keys.
   Changed to the KSP ("Microsoft Smart Card Key Storage Provider") and set
   KeySpec to 0 (XCN_AT_NONE) for CNG compatibility.

2. **Container name mismatch**: FreeRDP computes container names from the
   card's CHUID GUID + PIV certificate tag. However, Windows minidrivers
   (e.g. ykmd.dll) may assign different container names during certificate
   enrollment, which are stored in the MSCMAP object on the card (PIV tag
   0x5FFF10). This mismatch causes the server to fail with
   STATUS_SMARTCARD_NO_KEY_CONTAINER (0xC0000384).

   The fix reads the MSCMAP via GET DATA APDU before falling back to the
   CHUID-based method. Each 107-byte record contains a UTF-16 container name
   and a PIV slot byte, which is matched against the certificate's slot.
   Cards without an MSCMAP (no Windows minidriver enrollment) fall back
   gracefully to the existing CHUID-based algorithm.

This is related to the TODO list in #7707, which marked "implement container
name computation under linux" as done. That implementation only covered the
CHUID-based algorithm, which works for the Microsoft inbox minidriver
(msclmd.dll) but not for third-party minidrivers like ykmd.dll that use
enrollment-time container names stored in the MSCMAP.

Also updates `check_for_piv_container_name()` to use MAX_CONTAINER_NAME_LEN
(39) instead of PIV_CONTAINER_NAME_LEN (36), since MSCMAP container names
can be up to 39 characters.

## Verification

Before this patch, `xfreerdp /list:smartcard` shows the CHUID-derived name:
```
0: CN = "REDACTED"
    * containerName: ad6818ee-78e6-b44c-acd5-f0ce3e5fc105
```

The Windows minidriver (ykmd.dll) uses a different name (`certutil -scinfo`):
```
Key Container = te-Smartcard-Anmeldung-ECC-593b68-61208
```

After this patch, `xfreerdp /list:smartcard` shows the MSCMAP-derived name,
matching the minidriver:
```
0: CN = "REDACTED"
    * containerName: te-Smartcard-Anmeldung-ECC-593b68-61208
```

## How to test

1. YubiKey (or other PIV card) with ECC certificate enrolled via Windows
   CA using a third-party minidriver (e.g. ykmd.dll)
2. Before this patch: `xfreerdp /smartcard-logon /smartcard /sec:nla` fails
   with STATUS_SMARTCARD_NO_KEY_CONTAINER (visible in Windows Security Event
   Log as Event 4625 with substatus 0xC0000384)
3. After this patch: auto-logon works without specifying a container name
4. Verify with `xfreerdp /list:smartcard` that the containerName now shows
   the MSCMAP-derived name instead of the CHUID-derived one
5. Cards without MSCMAP should continue to work as before (CHUID fallback)

The container name can still be overridden via `/smartcard-logon:container:<name>`
if needed.

### DIFF
--- a/winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
+++ b/winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
@@ -899,6 +899,163 @@ static SECURITY_STATUS NCryptP11EnumKeys(NCRYPT_PROV_HANDLE hProvider, LPCWSTR p
 	return NTE_NO_MORE_ITEMS;
 }
 
+static BOOL piv_check_sw(DWORD buf_len, const BYTE* buf, BYTE expected_sw1)
+{
+	return (buf_len >= 2) && (buf[buf_len - 2] == expected_sw1);
+}
+
+static BOOL piv_check_sw_success(DWORD buf_len, const BYTE* buf)
+{
+	return (buf_len >= 2) && (buf[buf_len - 2] == 0x90) && (buf[buf_len - 1] == 0x00);
+}
+
+static SECURITY_STATUS get_piv_container_name_from_mscmap(SCARDHANDLE card,
+                                                          const SCARD_IO_REQUEST* pci,
+                                                          const BYTE* piv_tag, BYTE* output,
+                                                          size_t output_len)
+{
+	BYTE buf[258] = WINPR_C_ARRAY_INIT;
+	BYTE mscmap_buf[2148] = WINPR_C_ARRAY_INIT;
+	DWORD buf_len = sizeof(buf);
+	DWORD mscmap_total = 0;
+
+	if (SCardTransmit(card, pci, APDU_PIV_GET_MSCMAP, sizeof(APDU_PIV_GET_MSCMAP), nullptr, buf,
+	                  &buf_len) != SCARD_S_SUCCESS)
+		return NTE_NOT_FOUND;
+
+	if (piv_check_sw_success(buf_len, buf))
+	{
+		mscmap_total = buf_len - 2;
+		if (mscmap_total > sizeof(mscmap_buf))
+			return NTE_NOT_FOUND;
+		memcpy(mscmap_buf, buf, mscmap_total);
+	}
+	else if (piv_check_sw(buf_len, buf, 0x61))
+	{
+		mscmap_total = buf_len - 2;
+		if (mscmap_total <= sizeof(mscmap_buf))
+			memcpy(mscmap_buf, buf, mscmap_total);
+
+		while (piv_check_sw(buf_len, buf, 0x61) && mscmap_total < sizeof(mscmap_buf))
+		{
+			BYTE get_resp[5] = { 0x00, 0xC0, 0x00, 0x00, 0x00 };
+			get_resp[4] = buf[buf_len - 1];
+			buf_len = sizeof(buf);
+			if (SCardTransmit(card, pci, get_resp, sizeof(get_resp), nullptr, buf, &buf_len) !=
+			    SCARD_S_SUCCESS)
+				break;
+			DWORD chunk = piv_check_sw_success(buf_len, buf) ? (buf_len - 2)
+			              : piv_check_sw(buf_len, buf, 0x61) ? (buf_len - 2)
+			                                                 : 0;
+			if (chunk == 0 || mscmap_total + chunk > sizeof(mscmap_buf))
+				break;
+			memcpy(mscmap_buf + mscmap_total, buf, chunk);
+			mscmap_total += chunk;
+		}
+		if (!piv_check_sw_success(buf_len, buf))
+			return NTE_NOT_FOUND;
+	}
+	else
+		return NTE_NOT_FOUND;
+
+	/* Strip TLV wrappers: outer tag 0x53, inner tag 0x81 */
+	const BYTE* mscmap_data = mscmap_buf;
+	DWORD mscmap_data_len = mscmap_total;
+
+	for (int tlv_pass = 0; tlv_pass < 2; tlv_pass++)
+	{
+		if (mscmap_data_len < 2)
+			break;
+		BYTE tlv_tag = mscmap_data[0];
+		if (tlv_tag != 0x53 && tlv_tag != 0x81)
+			break;
+		size_t hdr = 2;
+		if (mscmap_data[1] == 0x82 && mscmap_data_len > 4)
+			hdr = 4;
+		else if (mscmap_data[1] == 0x81 && mscmap_data_len > 3)
+			hdr = 3;
+		mscmap_data += hdr;
+		mscmap_data_len -= (DWORD)hdr;
+	}
+
+	/* Map PIV tag to slot byte */
+	BYTE target_slot = 0;
+	for (size_t i = 0; i < ARRAYSIZE(piv_tag_to_slot); i++)
+	{
+		if (memcmp(piv_tag, piv_tag_to_slot[i].tag, 3) == 0)
+		{
+			target_slot = piv_tag_to_slot[i].slot;
+			break;
+		}
+	}
+	if (target_slot == 0)
+		return NTE_NOT_FOUND;
+
+	/* Search MSCMAP records (107 bytes each) for matching slot */
+	size_t num_records = mscmap_data_len / MSCMAP_RECORD_SIZE;
+	for (size_t i = 0; i < num_records; i++)
+	{
+		const BYTE* record = mscmap_data + (i * MSCMAP_RECORD_SIZE);
+		if (record[MSCMAP_SLOT_OFFSET] == target_slot)
+		{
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@dko-strd wow, nice additon, thank you!
looks good at a first glance, but need some time to check it properly.

winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
```diff
@@ -914,14 +955,127 @@ static SECURITY_STATUS get_piv_container_name(NCryptP11KeyHandle* key, const BYT
...
+		BYTE mscmap_buf[2148] = WINPR_C_ARRAY_INIT;
+		DWORD mscmap_len = 0;
+		DWORD mscmap_total = 0;
+		BOOL mscmap_found = FALSE;
+
+		buf_len = sizeof(buf);
+		if (SCardTransmit(card, pci, APDU_PIV_GET_MSCMAP, sizeof(APDU_PIV_GET_MSCMAP), nullptr,
+		                  buf, &buf_len) == SCARD_S_SUCCESS)
+		{
+			if (buf_len >= 2 && buf[buf_len - 2] == 0x90 && buf[buf_len - 1] == 0x00)
```
hardening: I'd suggest adding ` && (buf_len >= 2)` in the if above, this way you can remove this test later in the block.

winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
```diff
@@ -914,14 +955,127 @@ static SECURITY_STATUS get_piv_container_name(NCryptP11KeyHandle* key, const BYT
 	if ((buf[buf_len - 2] != 0x90 || buf[buf_len - 1] != 0) && buf[buf_len - 2] != 0x61)
 		goto out;
 
+	/* Try reading the MSCMAP (container map) from the card first.
+	 * Windows minidrivers store the actual container names there, which may differ
+	 * from the CHUID-derived names (e.g. after CHUID rotation or template enrollment). */
+	{
```
hardening: With your addings, I'd suggest to split the function to test for `MCSMAP` and for `PIV` in their own functions.

dko-strd:
This has been bugging me for a while. Last year I pushed for a (partially) YubiKey-based authentication setup in our organization. After the final decision to go with ECC certificates, things stopped working smoothly on Linux clients — smartcard redirection into the remote session was fine, but the actual RDP auto-logon with `/smartcard-logon` failed silently. Since I'm actively resisting Windows 11 usage outside of a VM, fixing this was long overdue.

I've tested the changes with both a YubiKey 5 and a Nitrokey 3, and both work correctly after the patch. That said, I'm aware this touches a sensitive code path and I definitely don't want to be the one who temporarily breaks smartcard support for everyone 😅 — so I'm grateful for any findings from the community.

Also addressed the review feedback in the latest commit — thanks @hardening for the quick and helpful review!

dko-strd:
One more note: I was only able to test this against certificates issued by a current AD CS running on Windows Server 2022 Datacenter. Unfortunately I don't have the means to quickly verify whether the MSCMAP format differs on older or newer Windows Server releases.

hardening:
I can test that the PIV path still works here

winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
```diff
@@ -858,6 +899,163 @@ static SECURITY_STATUS NCryptP11EnumKeys(NCRYPT_PROV_HANDLE hProvider, LPCWSTR p
...
+
+		while (piv_check_sw(buf_len, buf, 0x61) && mscmap_total < sizeof(mscmap_buf))
+		{
+			BYTE get_resp[5] = { 0x00, 0xC0, 0x00, 0x00, 0x00 };
+			get_resp[4] = buf[buf_len - 1];
+			buf_len = sizeof(buf);
+			if (SCardTransmit(card, pci, get_resp, sizeof(get_resp), nullptr, buf, &buf_len) !=
+			    SCARD_S_SUCCESS)
+				break;
+			DWORD chunk = piv_check_sw_success(buf_len, buf) ? (buf_len - 2)
```
hardening, line 947: We don't like very much these nested ternary-operator, if you change that to a regular `if-then-else` things will be more clear.

winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
```diff
@@ -858,6 +899,163 @@ static SECURITY_STATUS NCryptP11EnumKeys(NCRYPT_PROV_HANDLE hProvider, LPCWSTR p
...
+			return NTE_NOT_FOUND;
+		memcpy(mscmap_buf, buf, mscmap_total);
+	}
+	else if (piv_check_sw(buf_len, buf, 0x61))
+	{
+		mscmap_total = buf_len - 2;
+		if (mscmap_total <= sizeof(mscmap_buf))
+			memcpy(mscmap_buf, buf, mscmap_total);
+
+		while (piv_check_sw(buf_len, buf, 0x61) && mscmap_total < sizeof(mscmap_buf))
```
hardening, line 939: If I'm not mistaken at the first loop turn `piv_check_sw` is called "twice" as we've just checked that condition 4 lines above (in the if). I'd suggest to treat that in the loop itself (in the status checking that could possibly be a switch)

winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
```diff
@@ -858,6 +899,163 @@ static SECURITY_STATUS NCryptP11EnumKeys(NCRYPT_PROV_HANDLE hProvider, LPCWSTR p
...
+		{
+			BYTE get_resp[5] = { 0x00, 0xC0, 0x00, 0x00, 0x00 };
+			get_resp[4] = buf[buf_len - 1];
+			buf_len = sizeof(buf);
+			if (SCardTransmit(card, pci, get_resp, sizeof(get_resp), nullptr, buf, &buf_len) !=
+			    SCARD_S_SUCCESS)
+				break;
```
hardening, line 946: If I'm correct here we could just return `NTE_NOT_FOUND`

hardening:
May I suggest that you add a function that sends a request and reads the response taking in account either the `9000` or the `61XX` case, that will simplify a lot the interactions with the smartcard (even not including sw1-sw2 in the response buffer). That means that the callers of this function can concentrate on the response content, not on if it was transmitted with chunks or not.
With that it's gonna bring a massive simplification of the code.

akallabeth:
I can confirm it works with smartcard emulation (the old `PIV` path)

akallabeth:
@freerdp-bot test

akallabeth:
@dko-strd do you have a small howto for generating compatible key/cert pairs?
I'd like to test this, but the docs for setting up a full domain are a bit overwhelming.

got an older VM setup for testing (server 2019) so can I simply add these new certs to the existing RSA keys? (sorry for the questions, but I don't usually use this)

akallabeth:
@dko-strd ok, there are still some problems left:
`MS_SCARD_PROV` is used also when enumerating the certs in `NCryptEnumStorageProviders` and `winpr_NCryptOpenStorageProviderEx` and should most likely be replaced there too?

libfreerdp/core/nla.c
```diff
@@ -234,14 +234,10 @@ static BOOL nla_adjust_settings_from_smartcard(rdpNla* nla)
...
-		{
-			WLog_ERR(TAG, "unable to set CSP name");
-			goto out;
-		}
-		if (!settings->CspName &&
-		    !freerdp_settings_set_string(settings, FreeRDP_CspName, MS_SCARD_PROV_A))
+		/* Use KSP instead of legacy CSP — the CSP does not support ECC keys,
+		 * which are common on modern PIV smartcards. The KSP supports both RSA and ECC. */
+		if (!freerdp_settings_set_string(settings, FreeRDP_CspName,
+		                                 "Microsoft Smart Card Key Storage Provider"))
```
akallabeth, line 240: use `MS_SMART_CARD_KEY_STORAGE_PROVIDER` macro for this.

akallabeth: also, keep the old path as fallback in case the name can not be set.

akallabeth:
@dko-strd could you address the remaining issues?
did test and it is working, so would merge it then.

akallabeth:
Replaced by #12585

PR #12560, opened 2026-03-30
Fix remote credential with windows11h2

hardening:
The PDU fields is now checked and needs to be set to 7.


hardening:
Well this patch is the result of a byte to byte traffic comparison between mstsc and FreeRDP, trying to understand why it was not working with FreeRDP. mstsc sets that to 7, and IIRC when developing RCG 0 was not ok with previous versions of windows (but I may have a memory failure).

akallabeth:
@hardening if I read the spec correctly this was changed and backported to (some) windows versions, so maybe behavior differs depending on patch level?
anyway, merging this, just wanted to raise the question to be clarified as I'm not too familiar with that part of the protocol :grin:

PR #12579, opened 2026-04-08
[channels,rdpewa] fix crash with invalid free()

bluca:
Commit 8592765db27d7a0608e6133bcd3f2258a2eeacd9
refactored to do a manual memset and free. But it
invalidates the pointer that is then passed to free(), which crashes the client every time the user enters a PIN and presses accept.

Just use memset() to avoid invalidating the pointer.


recolic:
Thanks for your work!

akallabeth:
oh, you were faster, did prepare that already with #12581 
thank you, but closing as the other pr has some additional changes.

PR #12589, opened 2026-04-11
ci: add workflow for building stable windows builds

sitiom:
Github Actions workflow for building stable Windows builds, and uploading them to https://github.com/FreeRDP/FreeRDP/releases when a release is created (64-bit and 32-bit, built with vs2022).

I have tested the workflow on my fork: https://github.com/sitiom/FreeRDP/actions/runs/24284549257

### DIFF
--- a/.github/workflows/windows-release.yml
+++ b/.github/workflows/windows-release.yml
@@ -1,0 +1,132 @@
+name: windows-release
+
+on:
+  workflow_dispatch:
+  release:
+    types:
+      - published
+
+permissions:
+  contents: write
+
+jobs:
+  build:
+    runs-on: windows-2022
+    strategy:
+      matrix:
+        include:
+          - arch: x64
+            platform: x64
+            triplet: x64-windows-static
+            msvc_arch: x64
+          - arch: x86
+            platform: Win32
+            triplet: x86-windows-static
+            msvc_arch: amd64_x86
+
+    steps:
+      - name: Check out source
+        uses: actions/checkout@v6
+
+      - name: Set up Developer Command Prompt
+        uses: ilammy/msvc-dev-cmd@v1
+        with:
+          arch: ${{ matrix.msvc_arch }}
+
+      - name: Set up vcpkg
+        id: vcpkg
+        uses: johnwason/vcpkg-action@v7
+        with:
+          pkgs: >-
+            cjson
+            openssl
+            opus
+            sdl3
+            sdl3-image
+            sdl3-ttf
+            uriparser
+            zlib
+          triplet: ${{ matrix.triplet }}
+          cache-key: windows-release-${{ matrix.arch }}
+          token: ${{ github.token }}
+
+      - name: Configure CMake
+        run: |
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
before we can accept such a pr we'd need someone to maintain these stable builds.

.github/workflows/windows-release.yml
```diff
@@ -0,0 +1,137 @@
...
+            -DVCPKG_MANIFEST_MODE=OFF `
+            "-DCMAKE_INSTALL_PREFIX=$installDir" `
+            -DWITH_SAMPLE=ON `
+            -DWITH_CLIENT_SDL2=ON `
+            -DWITH_CLIENT_SDL3=ON `
+            -DWITH_CLIENT_SDL_VERSIONED=ON `
+            -DWITH_WINPR_TOOLS=ON `
+            -DWITH_WINPR_TOOLS_CLI=ON `
+            -DWITH_JSON_DISABLED=OFF `
+            -DWITH_CJSON_REQUIRED=OFF `
```
hardening: Shouldn't it be ON if you install the corresponding vcpkg package ?

.github/workflows/windows-release.yml
```diff
@@ -0,0 +1,137 @@
...
+
+          cmake --fresh -S $root -B $buildDir `
+            -G "Visual Studio 17 2022" `
+            -A "${{ matrix.platform }}" `
+            -C (Join-Path $root 'packaging\windows\preload.cmake') `
+            "-DCMAKE_TOOLCHAIN_FILE=$toolchain" `
+            "-DVCPKG_TARGET_TRIPLET=${{ matrix.triplet }}" `
+            -DVCPKG_MANIFEST_MODE=OFF `
+            "-DCMAKE_INSTALL_PREFIX=$installDir" `
+            -DWITH_SAMPLE=ON `
```
hardening, line 68: Not sure you need that for a release package...

hardening:
Thank you for proposing that, I don't know if we'll take it at the end, but I know how much effort it takes to make these workflow work correctly.

.github/workflows/windows-release.yml
sitiom: I was trying to match the buildconfig from the [nightly builds](https://ci.freerdp.com/job/freerdp-nightly-windows/), but I think ON should be the correct value here.

.github/workflows/windows-release.yml
sitiom: Maybe for completeness, some packages do include the sample client/server 🤔

.github/workflows/windows-release.yml
```diff
@@ -0,0 +1,137 @@
...
+
+      - name: Set up vcpkg
+        id: vcpkg
+        uses: johnwason/vcpkg-action@v7
+        with:
+          pkgs: >-
+            cjson
+            openssl
+            opus
+            sdl2
```
akallabeth: sdl2 client is deprecated and should better not be used anymore.

PR #12592, opened 2026-04-13
[codec,progressive] fix underflow guard in progressive_rfx_quant_sub

ondrejholy:
Currently, `progressive_rfx_quant_sub` checks `q1->HH1 < q2->HL1` before computing `q1->HL1 - q2->HL1`. This is a copy-paste error — the guard compares the wrong field (`HH1` instead of `HL1`), leaving the `HL1` subtraction unprotected against underflow. Let's fix the guard to compare `q1->HL1 < q2->HL1`.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM, nice catch!

PR #12594, opened 2026-04-13
fix: wfreerdp floatbar visibility

### LINKED ISSUE #11048: Floatbar not visible - windows wfreerdp.exe
Hello,

we are using wfreerdp.exe on our windows terminal servers (the bug is on a normal win11 pc too). We have no floatbar in the newer versions of free rdp (3.8, 3.9 and 3.10.02)
In the older version 2.2 we can get a floatbar, but in this version is a bug too, so we cant use it permanently. It would be nice, if we could use the newest version too inclusiv the floatbar :)

We are using this commanline: G:\XXX\wfreerdp-2.2 /cert:ignore /clipboard /f /floatbar:sticky:on,default:visible,show:fullscreen /smart-sizing /client-hostname:XX /u:%USERNAME%@XX.local /v:XX.XX.XX.XX

Best regards
Christoph

sitiom:
- Fixes #11048 

Fix the Windows `wfreerdp` floatbar so it is repositioned correctly once the main window becomes visible.

### DIFF
--- a/client/Windows/wf_floatbar.c
+++ b/client/Windows/wf_floatbar.c
@@ -731,3 +731,29 @@ BOOL wf_floatbar_toggle_fullscreen(wfFloatBar* floatbar, BOOL fullscreen)
 
 	return TRUE;
 }
+
+BOOL wf_floatbar_reset_position(wfFloatBar* floatbar)
+{
+	RECT rect = WINPR_C_ARRAY_INIT;
+	const int y = -(int)floatbar->offset;
+
+	if (!floatbar || !floatbar->parent || !floatbar->hwnd)
+		return FALSE;
+
+	if (!GetWindowRect(floatbar->parent, &rect))
+		return FALSE;
+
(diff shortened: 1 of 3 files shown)


client/Windows/wf_floatbar.c
```diff
@@ -731,3 +731,29 @@ BOOL wf_floatbar_toggle_fullscreen(wfFloatBar* floatbar, BOOL fullscreen)
 
 	return TRUE;
 }
+
+BOOL wf_floatbar_reset_position(wfFloatBar* floatbar)
+{
+	RECT rect = WINPR_C_ARRAY_INIT;
+	const int y = -(int)floatbar->offset;
```
akallabeth: * null dereference, move below null checks
* use `WINPR_ASSERTING_INT_CAST` to cast to int

akallabeth (changes requested):

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12609, opened 2026-04-16
[winpr,ntlm] fix message cleanup across the SSPI lifecycle

### LINKED ISSUE #12603: [Fuzzing] memory leak in NTLM AUTHENTICATE parser: ntlm_read_AuthenticateMessage fail path leaks up to 5 security-buffer allocations
## Severity
**High** — pre-authentication remote memory leak, unbounded per connection. Enables a small unauthenticated botnet to exhaust server RAM in minutes (DoS). Not directly exploitable for code execution.

## Affected function / file
- `ntlm_read_AuthenticateMessage` — `winpr/libwinpr/sspi/NTLM/ntlm_message.c` (bare `return status;` at the `fail:` label near **line 1227**; cf. the success path at lines **1220–1224** which calls `ntlm_free_message_fields_buffer` on all six fields).
- Per-field allocation happens in `ntlm_read_message_fields_buffer` at `ntlm_message.c:361` (`malloc(fields->Len)`).
- Reachable **pre-auth** via the NTLMSSP handshake.

## Root cause
`ntlm_read_AuthenticateMessage` allocates six security-buffer fields one at a time (`DomainName`, `UserName`, `Workstation`, `LmChallengeResponse`, `NtChallengeResponse`, `EncryptedRandomSessionKey`), each via `ntlm_read_message_fields_buffer` → `malloc(fields->Len)`.

The success path calls `ntlm_free_message_fields_buffer` on every field. The failure path (`fail:` label) is a bare `return status;` with **no** cleanup:

```c
fail:
    return status;   /* <-- leaks every already-allocated field buffer */
```

If earlier fields are allocated successfully and a later field's `Len` / `MaxLen` / `Offset` fails the sanity check (e.g. buffer offset beyond received data), every already-allocated earlier buffer leaks.

`Len` is an attacker-controlled `UINT16` (0–65535), so up to **~320 KB per malformed request** (5 fields × ~64 KB) can be leaked pre-auth. At typical connection rates a small botnet can exhaust server RAM in minutes.

The analogous `ntlm_read_ChallengeMessage` fail path only frees `TargetName`, but the full field set is not allocated there before the fail-point, so that parser is likely fine — should be re-audited as part of this fix.

## Reproduction
Discovered via a new libFuzzer harness (`TestFuzzNTLMMessage`) that dispatches between `ntlm_read_{Negotiate,Challenge,Authenticate}Message` by a 1-byte prefix, run under `ASAN_OPTIONS=detect_leaks=1`. 2,977 dedup'd leak reproducers were saved.

Canonical reproducer (65 bytes, hex):

```
6b 4e 54 4c 4d 53 53 50 00 03 00 00 00 50 00 00
00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
00 00 00 00 c0 08 00 00 00 00 00 00 00 60 00 00
00 41 00 00 00 00 00 00 00 00 00 00 00 00 00 00
04
```

(The first byte `0x6b` is the harness dispatch selector; bytes `4e 54 4c 4d 53 53 50 00` are the `NTLMSSP\0` signature; `03 00 00 00` selects the AUTHENTICATE message type.)

To reproduce from scratch:

```bash
# Save the 65-byte input
python3 -c "import sys; sys.stdout.buffer.write(bytes.fromhex('6b4e544c4d5353500003000000500000000000000000000000000000000000000000000000c00800000000000000600000004100000000000000000000000000000004'))" > poc_ntlm.bin

# Run under LSan (libFuzzer enables LSan by default on ASan builds)
ASAN_OPTIONS=detect_leaks=1 ./TestFuzzNTLMMessage poc_ntlm.bin
```

Abridged LSan output:

```
[ERROR] NTLM_MESSAGE_FIELDS::Buffer offset 161 beyond received data 64
Direct leak of 8 byte(s) in 1 object(s) allocated from:
  #0 malloc
  #1 ntlm_read_message_fields_buffer  winpr/libwinpr/sspi/NTLM/ntlm_message.c:361
  #2 ntlm_read_AuthenticateMessage    winpr/libwinpr/sspi/NTLM/ntlm_message.c:1019
SUMMARY: AddressSanitizer: 8 byte(s) leaked in 1 allocation(s).
```

## Suggested fix
Convert the `fail:` label in `ntlm_read_AuthenticateMessage` to free all six message-field buffers, mirroring the success-path cleanup at lines 1219–1224:

```c
fail:
    ntlm_free_message_fields_buffer(&(message->DomainName));
    ntlm_free_message_fields_buffer(&(message->UserName));
    ntlm_free_message_fields_buffer(&(message->Workstation));
    ntlm_free_message_fields_buffer(&(message->LmChallengeResponse));
    ntlm_free_message_fields_buffer(&(message->NtChallengeResponse));
    ntlm_free_message_fields_buffer(&(message->EncryptedRandomSessionKey));
    return status;
```

`ntlm_free_message_fields_buffer` already tolerates NULL buffers (it checks `fields->Buffer != NULL` before freeing), so calling it on not-yet-allocated fields is safe.

Then explicitly re-audit `ntlm_read_ChallengeMessage`'s fail path for the same pattern once the AUTHENTICATE fix is in — the set of fields allocated before its fail-point is narrower, but worth confirming.

## Disclosure
Found via offline fuzzing campaign on commit `55149bc0a39d5936c9985427c4cdf6908216eae6` (master tip at run time). Reporting here since the bug is in open-source code, pre-auth reachable, and the fix is mechanical.

parasol-aser:
Fixes #12603.

## Why this change is needed

@akallabeth correctly noted that the production success path does not
leak — `ntlm_ContextFree` already releases everything owned by the
context under normal SSPI usage. The leak this patch fixes is on the
**parser failure path** only, so it never triggers for well-formed
NTLM traffic.

However it is reachable pre-auth with a malformed AUTHENTICATE token.
`ntlm_read_AuthenticateMessage`'s `fail:` label was previously a bare
`return status;`, while the success path freed six message-field
buffers. When `ntlm_read_message_fields_buffer` succeeded for earlier
fields and a later sanity check failed, every already-allocated
earlier buffer leaked. `Len` is attacker-controlled `UINT16` so up to
~320 KB per request can be leaked by an unauthenticated attacker —
unbounded remote memory pressure over many connections.

Reproducer from the linked issue (65 bytes, pre-auth); see #12603 for
full hex. Under ASan+LSan this produces:
```
[ERROR] NTLM_MESSAGE_FIELDS::Buffer offset 161 beyond received data 64
Direct leak of 8 byte(s) in 1 object(s) allocated from:
  #0 malloc
  #1 ntlm_read_message_fields_buffer  ntlm_message.c:361
  #2 ntlm_read_AuthenticateMessage    ntlm_message.c:1019
```

## Scope of the change

One function touched: `ntlm_read_AuthenticateMessage`
(`winpr/libwinpr/sspi/NTLM/ntlm_message.c`). No changes to
`ntlm_ContextFree`, no new helpers, no touching the Negotiate or
Challenge parsers, no test or harness changes bundled in.

- Move the success-path cleanup (the six `ntlm_free_message_fields_buffer` calls) into the shared `fail:` label.
- Route the remaining post-allocation `return …` statements through `goto fail` with `status` set so they also release the buffers.
- `NtChallengeResponse.Buffer` is aliased into `context->NtChallengeResponse.pvBuffer` between its mid-function assignment (~L1048) and the reallocation inside `ntlm_compute_ntlm_v2_response()`. Skip freeing the message field while that alias is live — `ntlm_ContextFree()` then frees exactly once via `sspi_SecBufferFree(&context->NtChallengeResponse)`. An explanatory comment documents the invariant.

Diff: **+30 / -17** in one function.

## Test plan

- `ninja -C build-test TestSspi`
- `ctest --output-on-failure -R 'TestNTLM|TestInitializeSecurityContext|TestCredentialZeroing'` — all pass
- `ASAN_OPTIONS=detect_leaks=1 ./fuzz/TestFuzzNTLMMessage fuzz/crashes/TestFuzzNTLMMessage/leak-001c6d5eb53b6c82bc4ad8516b4b214692df5d35` — clean (was leaking on master)

### DIFF
--- a/winpr/libwinpr/sspi/NTLM/ntlm_message.c
+++ b/winpr/libwinpr/sspi/NTLM/ntlm_message.c
@@ -1125,24 +1125,30 @@ SECURITY_STATUS ntlm_read_AuthenticateMessage(NTLM_CONTEXT* context, PSecBuffer 
 	{
 		const SECURITY_STATUS rc = ntlm_compute_lm_v2_response(context); /* LmChallengeResponse */
 		if (rc != SEC_E_OK)
-			return rc;
+		{
+			status = rc;
+			goto fail;
+		}
 	}
 
 	{
 		const SECURITY_STATUS rc = ntlm_compute_ntlm_v2_response(context); /* NtChallengeResponse */
 		if (rc != SEC_E_OK)
-			return rc;
+		{
+			status = rc;
+			goto fail;
+		}
 	}
 
 	/* KeyExchangeKey */
 	if (!ntlm_generate_key_exchange_key(context))
-		return SEC_E_INTERNAL_ERROR;
+		goto fail;
 	/* EncryptedRandomSessionKey */
 	if (!ntlm_decrypt_random_session_key(context))
-		return SEC_E_INTERNAL_ERROR;
+		goto fail;
 	/* ExportedSessionKey */
 	if (!ntlm_generate_exported_session_key(context))
-		return SEC_E_INTERNAL_ERROR;
+		goto fail;
 
 	if (flags & MSV_AV_FLAGS_MESSAGE_INTEGRITY_CHECK)
 	{
@@ -1150,7 +1156,7 @@ SECURITY_STATUS ntlm_read_AuthenticateMessage(NTLM_CONTEXT* context, PSecBuffer 
 
 		if (!ntlm_compute_message_integrity_check(context, messageIntegrityCheck,
 		                                          sizeof(messageIntegrityCheck)))
-			return SEC_E_INTERNAL_ERROR;
+			goto fail;
 		CopyMemory(
 		    &((PBYTE)context->AuthenticateMessage.pvBuffer)[context->MessageIntegrityCheckOffset],
 		    message->MessageIntegrityCheck, sizeof(message->MessageIntegrityCheck));
@@ -1166,7 +1172,8 @@ SECURITY_STATUS ntlm_read_AuthenticateMessage(NTLM_CONTEXT* context, PSecBuffer 
 			winpr_HexDump(TAG, WLOG_ERROR, message->MessageIntegrityCheck,
 			              sizeof(message->MessageIntegrityCheck));
 #endif
-			return SEC_E_MESSAGE_ALTERED;
+			status = SEC_E_MESSAGE_ALTERED;
+			goto fail;
 		}
 	}
 	else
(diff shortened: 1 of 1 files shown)


akallabeth:
@parasol-aser is this change required?
I can't produce any leaks with the current implementation (and do prefer not to change critical paths without justification)

parasol-aser:
Thanks for pushing back on this, that was the right call — the first version of this PR touched `ntlm_ContextFree` and added alias-aware helpers, none of which is needed for the actual bug. I've force-pushed a much narrower patch.

**You are correct that the production success path does not leak**, so this is not observable under normal NTLM usage. The leak is strictly on the parser failure path: `ntlm_read_AuthenticateMessage`'s `fail:` label was a bare `return status;` while the success path freed six message-field buffers. On malformed input where `ntlm_read_message_fields_buffer` succeeds for earlier fields and a later sanity check fails, the already-allocated earlier buffers leak — up to ~320 KB per request (5 × attacker-controlled `UINT16` Len), reachable pre-auth, so a small number of malformed NTLMSSP tokens per connection adds up to remote memory pressure. The canonical 65-byte reproducer and LSan trace are in #12603.

**Revised scope (vs the first push):**
- Only `ntlm_read_AuthenticateMessage` is touched. `ntlm_ContextFree`, the NEGOTIATE/CHALLENGE parsers, and the test files are untouched.
- Success-path cleanup (the six `ntlm_free_message_fields_buffer` calls) moves into the shared `fail:` label, and the remaining post-allocation `return …` statements become `goto fail` with `status` set, so both exits release the six buffers.
- One invariant to be aware of: `NtChallengeResponse.Buffer` is aliased into `context->NtChallengeResponse.pvBuffer` between the mid-function assignment and the reallocation inside `ntlm_compute_ntlm_v2_response()`. The `fail:` label skips freeing `message->NtChallengeResponse` while that alias is live, so `sspi_SecBufferFree(&context->NtChallengeResponse)` in `ntlm_ContextFree()` frees it exactly once. A comment documents this.

Diff is now **+30 / -17** in a single function. `TestNTLM`, `TestInitializeSecurityContext`, and `TestCredentialZeroing` continue to pass; the `leak-001c6d5e…` reproducer from #12603 is clean under ASan+LSan after the change.

parasol-aser:
@akallabeth too bad the above comment is too verbose (made by my coding agent), but the leak is on the parser failure path. The new fuzzers are in https://github.com/FreeRDP/FreeRDP/pull/12610

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12619, opened 2026-04-16
[client,Windows] optimize floatbar positioning and visibility behavior

tsz8899:
## Summary
- optimize floatbar positioning and visibility behavior for the Windows client on current upstream
- improve floatbar repositioning, show/hide behavior, and related state synchronization
- fix incorrect maximize/restore button visibility switching in the floatbar window

## Main changes
- route floatbar position updates through the `wf_floatbar_reset_position()` call path
- improve floatbar repositioning after desktop resize, window move, window resize, and exit-size-move
- replace the previous hide/show behavior with two stable states:
  - full bar
  - 2px strip
- adjust floatbar width dynamically based on the window title while keeping a minimum width
- fix maximize/restore button visibility switching for different window states
- improve floatbar resource cleanup on disconnect

- reorganize floatbar positioning and visibility logic around the current upstream call flow
- consolidate repositioning, visibility state switching, and lifecycle handling to reduce inconsistent states

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -1216,7 +1216,7 @@ static DWORD WINAPI wf_client_thread(LPVOID lpParam)
 		DWORD wait_result;
 
 		/* Start disconnection in a separate thread to prevent hang on laggy networks */
-		disconnect_thread = CreateThread(NULL, 0, wf_disconnect_thread, instance, 0, NULL);
+		disconnect_thread = CreateThread(nullptr, 0, wf_disconnect_thread, instance, 0, nullptr);
 		if (disconnect_thread)
 		{
 			/* Wait for 2 seconds, if timeout, give up and exit */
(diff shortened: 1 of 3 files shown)


client/Windows/wf_client.c
```diff
@@ -1537,10 +1594,16 @@ static int wfreerdp_client_stop(rdpContext* context)
...
+		wait_result = WaitForSingleObject(wfc->keyboardThread, 1000);
+		if (wait_result == WAIT_TIMEOUT)
+		{
+			WLog_ERR(TAG, "Keyboard thread cleanup timeout, terminating");
+			TerminateThread(wfc->keyboardThread, 1);
+		}
 		(void)CloseHandle(wfc->keyboardThread);
-		wfc->keyboardThread = nullptr;
+		wfc->keyboardThread = NULL;
 		wfc->keyboardThreadId = 0;
```
hardening, line 1607: This change seems wrong as if I'm correct we're going on `nullptr` instead of `NULL`

tsz8899: @hardening thanks
use nullptr in floatbar redesign changes

What further adjustments are needed

akallabeth (changes requested):
sorry, this looks like a huge mess.

changes all over the place, not really related to each other and some hacks (why do you spawn a thread for disconnect? if you need that you did something seriously wrong)

PR #12629, opened 2026-04-18
fix: prevent PostQuitMessage in RemoteApp WM_DESTROY handler

### LINKED ISSUE #4074: wfreerdp remoteapp unusable
As it stands now multiple windows are spawned, this holds true for all versions. On some versions the remoteapp program will not start or  wfreerdp immediately closes. Resizing or moving the window makes the program crash and all sorts of graphical glitches, unfortunately.

Tested on 2.0 and previous released versions.

On line 632 rail_main.c
if (message.id == WMQ_QUIT)
			break;

This is with certainty a bug it has to change to avoid unexpected shutdown of wfreerdp; it will actually make wfreedp return when (any) window is closed. 

On line 419 wf_rail.c
case WM_CLOSE:
	DestroyWindow(hWnd);
        break;
case WM_DESTROY:
	PostQuitMessage(0);
	break;

This is redundant and causes bugs wfreerdp will e.g. hang after this call. 
Function:
static BOOL wf_rail_window_delete(rdpContext* context,
                                  WINDOW_ORDER_INFO* orderInfo)
should be used to handle this; it is implemented as a callback.

wfreerdp seems to miss wm_move messages causing the mouse pointer and expected position to be misaligned. The problem here is that messages are not captured outside the rectangle of a window (having focus) . I have fixed this in my version.

wfreerdp main window is still shown despite the app option this issue should be addressed the app option should be handled in "wf_post_connect(freerdp* instance)" function located in file "wf_client.c" at line 304.

The syntax for remoteapp:
wfreerdp.exe /v:[server IP/Hostname] /app:"||ALIAS"

If someone needs a remoteapp environment I will try and make it available. We should actually have certain environments readily available for development and testing purposes.

ZackaryShen:
## Summary

Fix for issue #4074: wfreerdp remoteapp unusable

### Problem

When using wfreerdp in RemoteApp mode (`/app:"||ALIAS"`), closing a RemoteApp window triggers `PostQuitMessage(0)` in the `wf_RailWndProc` WM_DESTROY handler. This unexpectedly quits the entire wfreerdp session instead of just closing the RemoteApp application window.

### Fix

Comment out `PostQuitMessage(0)` from the `wf_RailWndProc` WM_DESTROY handler. The `DestroyWindow(hWnd)` call in the WM_CLOSE handler is appropriate for RemoteApp windows -- the window should be destroyed when closed. However, `PostQuitMessage(0)` should only be called when the main FreeRDP session window is closed, not individual RemoteApp child windows.

### Changes

- `client/Windows/wf_rail.c`: Commented out `PostQuitMessage(0)` from WM_DESTROY case in `wf_RailWndProc`

Fixes #4074

### DIFF
--- a/client/Windows/wf_rail.c
+++ b/client/Windows/wf_rail.c
@@ -400,7 +400,7 @@ LRESULT CALLBACK wf_RailWndProc(HWND hWnd, UINT msg, WPARAM wParam, LPARAM lPara
 			break;
 
 		case WM_DESTROY:
-			PostQuitMessage(0);
+			/* PostQuitMessage(0) - REMOVED: closing RemoteApp window should not quit the session */
 			break;
 
 		default:


akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #12631, opened 2026-04-19
(sdl3): probe display with 64x64 window instead of fullscreen

morgan9e:
fullscreen probe causes screen to flash e.g. on macOS.

Use 64x64 minimal window for probing display information for non-wlroots,
and keep fullscreen probe for wlroots edgecase.

### DIFF
--- a/client/SDL/SDL3/sdl_window.cpp
+++ b/client/SDL/SDL3/sdl_window.cpp
@@ -345,15 +345,8 @@ bool SdlWindow::fill(SDL_Window* window, Uint8 r, Uint8 g, Uint8 b, Uint8 a)
 	return SDL_FillSurfaceRect(surface, &rect, color);
 }
 
-rdpMonitor SdlWindow::query(SDL_Window* window, SDL_DisplayID id, bool forceAsPrimary)
+static rdpMonitor buildMonitor(SDL_DisplayID id, bool forceAsPrimary, const SDL_Rect& r, float dpi)
 {
-	if (!window)
-		return {};
-
-	const auto& r = rect(window, forceAsPrimary);
-	const float factor = SDL_GetWindowDisplayScale(window);
-	const float dpi = std::roundf(factor * 100.0f);
-
 	WINPR_ASSERT(r.w > 0);
 	WINPR_ASSERT(r.h > 0);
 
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
* 1x1 windows are not allowed on all platforms
* you got unrelated changes en masse, why?

morgan9e:
> * 1x1 windows are not allowed on all platforms
> 
>     * you got unrelated changes en masse, why?

fixed to use 64x64 window.

although what's unrelated changes may I ask?

akallabeth:
> > * 1x1 windows are not allowed on all platforms
> >   
> >   * you got unrelated changes en masse, why?
> 
> fixed to use 64x64 window.
> 
> although what's unrelated changes may I ask?

you only need to change the fullscreen flag?
or what are all the other changes for?

morgan9e:
> > > * 1x1 windows are not allowed on all platforms
> 
> > >   
> 
> > >   * you got unrelated changes en masse, why?
> 
> > 
> 
> > fixed to use 64x64 window.
> 
> > 
> 
> > although what's unrelated changes may I ask?
> 
> 
> 
> you only need to change the fullscreen flag?
> 
> or what are all the other changes for?

I thought fullscreen probing was needed to properly get screen size and scale no matter the driver we are using (x11, wayland, macOS)

but it makes desktop flash so I was thinking about handling each driver seperately...


but still is seems like wlroots requires fullscreen probing no matter what so fallback for those too

akallabeth:
ok, I don't think this will work.
scaling factors are only available from the window, the ones from display do not (always) tell the truth for the specific monitor, sorry

MrVampy:
Hey @morgan9e — came across this PR after hitting an adjacent problem in the same area. Sharing in case it's useful context.

The specific bug I ran into on wlroots (Sway, Hyprland) and Smithay (Niri): hidden/unmapped windows are placed on the *currently-focused* display regardless of `SDL_WINDOWPOS_CENTERED_DISPLAY(id)` passed to `createDummy()`. So on a mixed-scale setup (scale-2 laptop + three scale-1 externals), every per-display dummy-window query lands on the laptop — the externals get reported with the laptop's 2x pixel density, inflating their reported resolution.

The 64x64 change in this PR doesn't address that placement issue — a smaller probe window on wlroots still gets pinned to the focused output.

FWIW, I sketched a hybrid: keep the window-based probe as the authoritative source (per @akallabeth's point that window-scoped APIs are SDL's intended query surface), but fall back to `SDL_GetDisplayBounds(id)` + `SDL_GetDesktopDisplayMode(id)->pixel_density` specifically when `XDG_SESSION_DESKTOP` / `XDG_CURRENT_DESKTOP` matches a known-affected compositor (sway / Hyprland / river / wlroots / niri). Overridable via env var. That narrows the non-standard path to the specific compositors where the window placement bug manifests:

https://github.com/MrVampy/FreeRDP/commit/b5397f4

Possibly complementary to your flash-reduction fix — they address different symptoms. Not suggesting re-opening anything; just putting it in the thread for visibility since both PRs touch the same probe path.

morgan9e:
Thanks, i'll also have a look and try to fix and think of optimal solution here and there..

PR #12635, opened 2026-04-19
(sdl3): Fix oversized window on HiDPI Wayland

morgan9e:
Fix oversized window on HiDPI Wayland.

On HiDPI Wayland sessions, the SDL3 FreeRDP window is created at the requested pixel count interpreted as logical pixels, which the compositor then scales up to physical pixels.

server surface appears smaller than wayland client window on windowed mode.

| Before |  After |
| ------ | ------- |
| Before | After |

### DIFF
--- a/client/SDL/SDL3/sdl_context.cpp
+++ b/client/SDL/SDL3/sdl_context.cpp
@@ -1117,6 +1117,8 @@ bool SdlContext::handleEvent(const SDL_WindowEvent& ev)
 		case SDL_EVENT_WINDOW_MOUSE_ENTER:
 			return restoreCursor();
 		case SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED:
+			if (!resizeToScale(window))
+				return false;
 			if (isConnected())
 			{
 				if (!window->fill())
@@ -1128,6 +1130,8 @@ bool SdlContext::handleEvent(const SDL_WindowEvent& ev)
 			}
 			break;
 		case SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED:
+			if (!resizeToScale(window))
+				return false;
 			if (!window->fill())
 				return false;
 			if (!drawToWindow(*window))
@@ -1452,6 +1456,17 @@ int SdlContext::argumentHandler(const COMMAND_LINE_ARGUMENT_A* arg, void* custom
 		}
 	}
 	return 0;
+}
+
+bool SdlContext::resizeToScale(SdlWindow* window)
+{
+	if (freerdp_settings_get_bool(context()->settings, FreeRDP_SmartSizing))
+		return true;
+	if (!useLocalScale())
+		return true;
+	if (!window)
+		return false;
+	return window->resizeToScale();
 }
 
 bool SdlContext::useLocalScale() const
--- a/client/SDL/SDL3/sdl_window.cpp
+++ b/client/SDL/SDL3/sdl_window.cpp
@@ -28,7 +28,7 @@
 
 SdlWindow::SdlWindow(SDL_DisplayID id, const std::string& title, const SDL_Rect& rect,
                      [[maybe_unused]] Uint32 flags)
-    : _displayID(id)
+    : _initialW(rect.w), _initialH(rect.h), _displayID(id)
 {
 	auto props = SDL_CreateProperties();
 	SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_TITLE_STRING, title.c_str());
@@ -36,6 +36,7 @@ SdlWindow::SdlWindow(SDL_DisplayID id, const std::string& title, const SDL_Rect&
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_Y_NUMBER, rect.y);
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_WIDTH_NUMBER, rect.w);
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, rect.h);
+	SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_RESIZABLE_BOOLEAN, true);
 
 	if (flags & SDL_WINDOW_HIGH_PIXEL_DENSITY)
 		SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_HIGH_PIXEL_DENSITY_BOOLEAN, true);
@@ -48,12 +49,12 @@ SdlWindow::SdlWindow(SDL_DisplayID id, const std::string& title, const SDL_Rect&
 
 	_window = SDL_CreateWindowWithProperties(props);
 	SDL_DestroyProperties(props);
-
-	std::ignore = resizeToPixelSize({ rect.w, rect.h });
 	SDL_SetHint(SDL_HINT_APP_NAME, "");
 	std::ignore = SDL_SyncWindow(_window);
 
 	_renderer = SDL_CreateRenderer(_window, nullptr);
+
+	std::ignore = resizeToScale();
 
 	_monitor = query(_window, id, true);
 }
@@ -61,8 +62,9 @@ SdlWindow::SdlWindow(SDL_DisplayID id, const std::string& title, const SDL_Rect&
 SdlWindow::SdlWindow(SdlWindow&& other) noexcept
     : _window(other._window), _renderer(other._renderer), _renderTarget(other._renderTarget),
       _gdiTexture(other._gdiTexture), _gdiTextureW(other._gdiTextureW),
-      _gdiTextureH(other._gdiTextureH), _displayID(other._displayID), _offset_x(other._offset_x),
-      _offset_y(other._offset_y), _monitor(other._monitor)
+      _gdiTextureH(other._gdiTextureH), _initialW(other._initialW), _initialH(other._initialH),
+      _displayID(other._displayID), _offset_x(other._offset_x), _offset_y(other._offset_y),
+      _monitor(other._monitor)
 {
 	other._window = nullptr;
 	other._renderer = nullptr;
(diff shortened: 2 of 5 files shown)


client/SDL/SDL3/sdl_window.cpp
```diff
@@ -228,13 +234,21 @@ void SdlWindow::minimize()
...
+		return;
+	if (SDL_GetWindowFlags(_window) & SDL_WINDOW_FULLSCREEN)
+		return;
+
 	auto sc = scale();
-	const int iscale = static_cast<int>(sc * 100.0f);
-	auto w = 100 * size.x / iscale;
-	auto h = 100 * size.y / iscale;
-	return resize({ w, h });
+	if (sc < 1.0f)
```
akallabeth: remove this, the check does not make sense.

akallabeth:
@morgan9e which environment did you test wit h?
can not reproduce your issue here on debian 13/gnome

morgan9e:
> @morgan9e which environment did you test wit h? can not reproduce your issue here on debian 13/gnome

Hi, I'm on Fedora 44, GNOME 50, Wayland on 160%.

client/SDL/SDL3/sdl_window.cpp
```diff
@@ -28,14 +28,15 @@
...
-    : _displayID(id)
+    : _displayID(id), _bufferW(rect.w), _bufferH(rect.h)
 {
 	auto props = SDL_CreateProperties();
 	SDL_SetStringProperty(props, SDL_PROP_WINDOW_CREATE_TITLE_STRING, title.c_str());
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_X_NUMBER, rect.x);
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_Y_NUMBER, rect.y);
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_WIDTH_NUMBER, rect.w);
 	SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_HEIGHT_NUMBER, rect.h);
+	SDL_SetBooleanProperty(props, SDL_PROP_WINDOW_CREATE_RESIZABLE_BOOLEAN, true);
```
akallabeth, line 39: not a good idea, this makes every window resizeable, but that should only be true if the display channel is enabled and supported or `/smart-sizing` is enabled

client/SDL/SDL3/sdl_window.cpp
```diff
@@ -228,13 +230,28 @@ void SdlWindow::minimize()
...
+	float pd = SDL_GetWindowPixelDensity(_window);
+	if (pd <= 0.0f)
+		pd = 1.0f;
+
+	const int targetW = static_cast<int>(std::ceil(static_cast<float>(_bufferW) / pd));
+	const int targetH = static_cast<int>(std::ceil(static_cast<float>(_bufferH) / pd));
+
+	int curW = 0;
+	int curH = 0;
+	SDL_GetWindowSize(_window, &curW, &curH);
```
akallabeth: return check missing or use `std::ignore`

client/SDL/SDL3/sdl_window.hpp
```diff
@@ -67,7 +67,7 @@ class SdlWindow
 	void fullscreen(bool enter, bool forceOriginalDisplay);
 	void minimize();
 
-	[[nodiscard]] bool resizeToPixelSize(const SDL_Point& size);
+	void resizeToScale();
```
akallabeth: why drop the return value?

akallabeth (changes requested):

akallabeth:
you might want to add `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` to the events you're interested in

akallabeth:
@morgan9e ok, indeed on `fedora 43 ` I can reproduce this as well.
overall the approach looks sensible, but you'd need to add this check to all scale related events and do some error checking.

akallabeth:
just tested:
* `/dynamic-resolution` no longer works
* `/smart-sizing` no longer works

client/SDL/SDL3/sdl_window.cpp
```diff
@@ -228,13 +232,28 @@ void SdlWindow::minimize()
...
-bool SdlWindow::resizeToPixelSize(const SDL_Point& size)
+bool SdlWindow::resizeToScale()
 {
-	auto sc = scale();
-	const int iscale = static_cast<int>(sc * 100.0f);
-	auto w = 100 * size.x / iscale;
-	auto h = 100 * size.y / iscale;
-	return resize({ w, h });
+	if (!_window || _bufferW <= 0 || _bufferH <= 0)
+		return true;
```
akallabeth: should this not be a failure condition?

morgan9e:
> just tested:
> 
>     * `/dynamic-resolution` no longer works
> 
>     * `/smart-sizing` no longer works

Does it work now?

Also for fullscreen, resizing does not works if we don't set that.. and since we lock resizing on postConnect

akallabeth:
@morgan9e with a3efd39e598fafc4289de266e0625a491883064c `/dynamic-resolution` and `/smart-sizing` work here now.

client/SDL/SDL3/sdl_window.cpp
```diff
@@ -28,14 +28,15 @@
 
 SdlWindow::SdlWindow(SDL_DisplayID id, const std::string& title, const SDL_Rect& rect,
                      [[maybe_unused]] Uint32 flags)
-    : _displayID(id)
+    : _displayID(id), _bufferW(rect.w), _bufferH(rect.h)
```
akallabeth: initialization order is wrong. with `C++` you need to initialize in order of declaration in header.
also, you might want to rename `_bufferW` and `_bufferH` as I'm not sure from the names what the value might mean. `_currentTargetW` or something simmilar might make it more clear.

akallabeth: or just keep the whole `SDL_Rect` used?

morgan9e: Thanks, fixed.

akallabeth:
@freerdp-bot test

client/SDL/SDL3/sdl_context.cpp
```diff
@@ -1117,6 +1109,8 @@ bool SdlContext::handleEvent(const SDL_WindowEvent& ev)
 		case SDL_EVENT_WINDOW_MOUSE_ENTER:
 			return restoreCursor();
 		case SDL_EVENT_WINDOW_DISPLAY_SCALE_CHANGED:
+			if (!window->resizeToScale())
```
akallabeth: this should not be done unconditionally.
the `resizeToScale` is only required if there is no dynamic resizing or smart scale in in use.
to rely on an external `SDL` event to behave as you expect on all platforms is not really something I'd consider stable.
so, best readd the `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` event and do not apply in case this is not required.

morgan9e: but even if there's dynamic-resolution or /smart-sizing, the initial window size is still wrong. (although they do fill the window later if window is resized once)

akallabeth: no.
you do set the size in the constructor, if `SDL` later changes the details of the window then the initially set values are already wrong and need to be updated.

so, to be clear, just skip the `resizeToScale` if you enter from an `SDL_Event`, not in the constructor.

morgan9e: but don't we not know the scale factor yet in constructor?

akallabeth: and?
if the scale factor changes then you need to tell that to the remote as well (in case of `/dynamic-resolution`)?
in case of `/smart-sizing` the update will stretch anyway on next draw.

client/SDL/SDL3/sdl_window.cpp
```diff
@@ -248,6 +248,10 @@ bool SdlWindow::resizeToScale()
 	int curH = 0;
 	if (!SDL_GetWindowSize(_window, &curW, &curH))
 		return false;
+	
+	if (curW != _requestedW || curH != _requestedH)
```
akallabeth: not sure what this change is for?
[edit] ok, this breaks the whole thing? how should both, this and the next check pass to allow resizing?

client/SDL/SDL3/sdl_context.cpp
akallabeth: the patch attached does work here for these modes:

[0001-update-scale-only-in-static-mode.patch](https://github.com/user-attachments/files/26929359/0001-update-scale-only-in-static-mode.patch)

morgan9e: yeah that works, thnaks

client/SDL/SDL3/sdl_context.cpp
```diff
@@ -1454,6 +1458,17 @@ int SdlContext::argumentHandler(const COMMAND_LINE_ARGUMENT_A* arg, void* custom
 	return 0;
 }
 
+bool SdlContext::resizeToScale(SdlWindow* window)
+{
+	if (useLocalScale())
```
akallabeth: ok, this is my bad, but this condition was inverted :/

akallabeth (changes requested):
Ok, if the latest comment is fixed this should work with `wayland` and `macos`.
see https://wiki.libsdl.org/SDL3/README-highdpi

`windows` and `x11` are most likely not working (can't test with `x11` as `XWayland` does not pass the scale for my setup)

client/SDL/SDL3/sdl_window.cpp
morgan9e: since we are resizing against constructor size, I thought it was better to check if's actually that size before we resize it

morgan9e:
> Ok, if the latest comment is fixed this should work with `wayland` and `macos`. see https://wiki.libsdl.org/SDL3/README-highdpi
> 
> `windows` and `x11` are most likely not working (can't test with `x11` as `XWayland` does not pass the scale for my setup)

we can guard it against Wayland I guess, but i think it's fine as is?

client/SDL/SDL3/sdl_window.cpp
```diff
@@ -228,13 +230,32 @@ void SdlWindow::minimize()
...
+
+	const int targetW = static_cast<int>(std::ceil(static_cast<float>(_initialW) / pd));
+	const int targetH = static_cast<int>(std::ceil(static_cast<float>(_initialH) / pd));
+
+	int curW = 0;
+	int curH = 0;
+	if (!SDL_GetWindowSize(_window, &curW, &curH))
+		return false;
+	
+	if (curW != _initialW || curH != _initialH)
```
akallabeth: * why rename to `_initialW` and `_initialH`?
* the check is still wrong. (you might have more than 1 monitor with different scaling settings)

akallabeth (changes requested):

akallabeth:
> > Ok, if the latest comment is fixed this should work with `wayland` and `macos`. see https://wiki.libsdl.org/SDL3/README-highdpi
> > `windows` and `x11` are most likely not working (can't test with `x11` as `XWayland` does not pass the scale for my setup)
> 
> we can guard it against Wayland I guess, but i think it's fine as is?

guess you're right with this, the `SDL_GetWindowPixelDensity` will return `1.0` on `windows / x11` cases, so `SDL_GetWindowSize` will return the correct sizes.

client/SDL/SDL3/sdl_window.cpp
akallabeth: you must update the `values after resizing the window though.

morgan9e: I thought it was better naming, i'll revert it if you don't like it

But if they passes through that, aren't we resizing against wrong size?? We're resizing to _initial/pd here

akallabeth: no?
it just means that the size actually changed from the initially reported value and you should do the fit?
currently it is either this or the next check return true and `resize` is never triggered?

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

morgan9e:
oops sorry left trailing spaces

akallabeth:
@freerdp-bot test

akallabeth:
@morgan9e please squash the commits

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

akallabeth:
thank you for the fix!

loki1368:
sdl-freerdp /v:192.168.0.5 /size:3840x2160 /smart-sizing /scale-desktop:175 +clipboard -grab-keyboard +force-console-callbacks

With version 3.25 my connections(windows 10, linux xrdp) is broken. In fullscreen mode I have picture like Before. Hyprland. Previous versions working well.

PR #12648, opened 2026-04-21
[client,X11] fix residual race in xf_clipboard_formats_free

ondrejholy:
Commit 58409406 locked xf_cliprdr_free_formats() but reset lastSentFormats/lastSentNumFormats after releasing the lock. This left a narrow window where xf_clipboard_changed() could observe a freed-but-non-NULL pointer if the X11 event thread acquired the lock between the unlock and the reset.

Move the pointer/count reset inside the critical section so that any reader holding the lock always sees a consistent nullptr/0 or a valid allocation.

Made-with: Cursor

### DIFF
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -869,13 +869,14 @@ static void xf_clipboard_formats_free(xfClipboard* clipboard)
 {
 	WINPR_ASSERT(clipboard);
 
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM,
thank you!

PR #12651, opened 2026-04-21
[cache,bitmap] fix off-by-one in bitmap_cache_put bounds check

ondrejholy:
Commit ffad58fd fixed a heap out-of-bounds access by overallocating the cells array by 1, citing "older RDP servers do send a off by 1 cache index". However, the advisory (GHSA-h666-rfw3-jhvj) identifies this as a malicious server attack with no legitimate protocol basis for a server to send cacheId == maxCells.

The overallocation also introduced a secondary issue: the initialization loop in bitmap_cache_new runs for i < maxCells only, leaving cells[maxCells].entries == NULL. As a result, a request with cacheId == maxCells still crashes the client via NULL pointer dereference in bitmap_cache_put instead of the original OOB write.

The correct fix is to change the boundary check in bitmap_cache_put from strict greater-than (id > maxCells) to id >= maxCells, making it consistent with bitmap_cache_get which already uses the correct check. With the bounds check fixed, the +1 overallocation from ffad58fd is no longer needed and is reverted.

Made-with: Cursor

### DIFF
--- a/libfreerdp/cache/bitmap.c
+++ b/libfreerdp/cache/bitmap.c
@@ -396,6 +396,19 @@ rdpBitmapCache* bitmap_cache_new(rdpContext* context)
 		cell->number = nr;
 	}
 
+	/* initialize the overallocated extra slot for old RDP servers that send
+	 * cacheId == maxCells; use a minimal allocation since no protocol-negotiated
+	 * capacity exists for this slot */
+	{
+		BITMAP_V2_CELL* extra = &bitmapCache->cells[bitmapCache->maxCells];
+		/* allocate an extra entry for BITMAP_CACHE_WAITING_LIST_INDEX */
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
sorry, this fix is bad.
the overallocation was correct to satisfy (really old, ok) `RDP` servers.

the correct fix would be to ensure that the `0` set also applies to that overallocated portion.

ondrejholy:
So what about now?

akallabeth:
@ondrejholy hmm, best check with an old `XP` server and check if that is still required. (I think it was a `XP` thing)

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12653, opened 2026-04-21
Add web-based remote desktop client with GitHub Pages support

GreatJack12:
This PR adds a web-based remote desktop client that enables streaming Windows desktops to web browsers with 60fps 1080p quality.

## Changes
- Added web-client/ directory with HTML/JavaScript interface
- Added WebSocket proxy server for RDP streaming  
- Added GitHub Actions workflow for automatic Pages deployment
- Updated main README with web client information
- Added comprehensive setup documentation

## Features
- 60fps streaming at 1080p resolution
- Low-latency WebSocket communication
- Cross-platform browser support
- Easy setup with FreeRDP shadow server

Once merged, GitHub Pages can be enabled at https://freerdp.github.io/FreeRDP/


akallabeth:
@GreatJack12 wow, looks like a huge pr.

* could you add a description for devs to have an idea what we're reviewing?
* could you please remove binary data (the whole `nodejs` stuff) from the pr? we're not checking in such data.

akallabeth:
@GreatJack12 Sorry but this is not an RDP client implementation.

The frontend simply opens a WebSocket connection and displays jpeg frames received from a backend service.
The backend appears to be a Node.js wrapper around ffmpeg screen capture outputting mjpeg over WebSocket.

There is actual zero RDP protocol handling here.

The README and PR description currently present this as a "web-based remote desktop client", which is misleading.
It is effectively a non-interactive remote screen JPEG streamer.

The reference to freerdp-shadow also does not match the actual implementation.
It is mentioned in the documentation but not used by the code - and it wouldn't make any sense either.

Additionally a large number of generated/vendor artifacts were committed, binaries were included in the PR and entire dependency/module directories were committed, which we do not accept.

If you want to propose browser support in the future, it would need to be based on real RDP protocol integration.

PR #12666, opened 2026-04-23
cmake: Findyuv: Use correct pkgconfig name

Conan-Kudo:
The pkgconfig name for libyuv in Linux distributions is `libyuv` rather than `yuv`. Use the correct name so it can be picked up properly.


akallabeth (approved):
LGTM

PR #12490, opened 2026-03-15
[ncrypt,smartcardlogon] support ECC keys in PKCS#11 smartcard enumeration

dko-strd:
## Summary

- Remove hardcoded `CKK_RSA` filter from `public_key_filter` in `collect_keys()` so that ECC (ECDSA) keys on smartcards (e.g. YubiKey PIV) are enumerated alongside RSA keys during `/smartcard-logon`
- Pass PKCS#11 `certid=` in the PKINIT identity string so MIT krb5's pkinit module can select the correct certificate when multiple certificates are present on the same token

Fixes #12489

## Details

The PKCS#11 key enumeration in `ncrypt_pkcs11.c:collect_keys()` used a hardcoded `CKA_KEY_TYPE = CKK_RSA` filter, causing `/smartcard-logon` to report "no certificates found" for any non-RSA key. Removing this filter allows all key types that support verification (`CKA_VERIFY = TRUE`) to be discovered. The per-object key type is still retrieved via `CKA_KEY_TYPE` and correctly mapped downstream — `convertKeyType()` already handles `CKK_EC` → `BCRYPT_ECDSA_ALGORITHM`.

Without the `certid=` parameter in the PKINIT identity string, MIT krb5's pkinit module fails with *"There are N certs, but there must be exactly one"* when multiple certificates exist on the token. The cert ID is already available in `scCert->keyName` and is now extracted and appended to `pkinitArgs`.

## Test plan

- [x] Verified with YubiKey 5 (ECCP384 in PIV slots 9a, 9c, 82, 83) — keys are now enumerated (`slot has 4 objects` vs `0 objects` before)
- [x] Certificate-to-key matching via `CKA_ID` works correctly for ECC keys
- [x] PKINIT identity string correctly includes `certid=` — MIT krb5 selects the right certificate
- [x] RSA keys continue to work (no filter regression — `CKA_VERIFY` still applies)
- [ ] End-to-end smartcard logon with ECC (blocked by unrelated KDC-side PKINIT/ECDSA interop issue with MIT krb5, not a FreeRDP problem)

PR #12684, opened 2026-04-26
fix(winpr,ncrypt): support PIV retired key slots for smartcard logon

dko-strd:
## Summary

Smartcard logon from a PIV **retired key management slot** (0x82–0x95)
fails on the Windows server with `NTE_SILENT_CONTEXT` because FreeRDP
sends an empty container name. Two interacting bugs in `ncrypt_pkcs11.c`:

1. `piv_cert_tags[]` only listed the four standard PIV slots; retired-slot
   labels (`"Retired Certificate for Key Management N"`) matched no entry.
2. The `NCRYPT_PROPERTY_NAME` `NTE_NOT_FOUND` fallback path (GIDS / raw
   `CKA_LABEL` branch) never set `*pcbResult` on the size-query call, so
   the caller allocated a 2-byte buffer and the second call wrote nothing.

This PR adds the 20 retired-slot label-to-PIV-tag mappings (the matching
`piv_tag_to_slot[]` entries already existed) and sets `*pcbResult` to the
WCHAR-converted label size in the fallback branch.

Although nominally for retired key-management keys, these slots are in
practice the only place to store **multiple smartcard-logon-capable
authentication certificates on a single PIV token**. This is exactly
what's needed to run a tiered-administration model (separate T0/T1/T2
admin identities) from one YubiKey instead of carrying one card per
tier.

Builds on the smartcard-logon work in #12490.

Verified end-to-end against Windows Server 2022 with a YubiKey-PIV cert
enrolled into a retired key-management slot — the auto-derived
CHUID-based container name now matches `certutil -scinfo` output and
logon succeeds.

### DIFF
--- a/winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
+++ b/winpr/libwinpr/ncrypt/ncrypt_pkcs11.c
@@ -91,6 +91,27 @@ static const piv_cert_tags_t piv_cert_tags[] = {
 	{ "Certificate for Digital Signature", { 0x5F, 0xC1, 0x0A } },
 	{ "Certificate for Key Management", { 0x5F, 0xC1, 0x0B } },
 	{ "Certificate for Card Authentication", { 0x5F, 0xC1, 0x01 } },
+
+	{ "Retired Certificate for Key Management 1", { 0x5F, 0xC1, 0x0D } },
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #12688, opened 2026-04-26
[client, android] Encrypt bookmark database

### LINKED ISSUE #2042: Android - aFreeRDP - Client stores passwords as cleartext
In the android client, stored passwords can be found as unencrypted clear text in the sqlite database.

The database file is /data/data/com.freerdp.afreerdp/databases/bookmarks.db
The table [tbl_manual_bookmarks] contains the following columns
[password]
[gateway_password]

Both of these fields should have the option of being encrypted with a user supplied master password.
 As it stands, the current state of the client is wholly open for exploitation.

svncibrahim:
**Name**
[client, android] Encrypt bookmark database with SQLCipher

**Description**

- Fiexes #2042

As discussed, this PR has been split. This part only handles internal database encryption. Export/import will be provided in a separate PR.

Key Changes:

*   **Database Encryption:** The internal SQLite database is now fully encrypted using SQLCipher. We don't have to manually check or encrypt specific fields.
*   **Hardware-Bound Key:** Uses `AndroidKeyStore` to generate a key that is permanently bound to the device hardware. No user password is required.
*   **Migration:** Safely migrates existing plaintext databases to the new encrypted SQLCipher format using [sqlcipher_export()](https://www.zetetic.net/sqlcipher/sqlcipher-api/index.html#sqlcipher_export), keeping existing user bookmarks safe. 

**Instructions on how to test**
Build the Android client (aFreeRDP).

1. Upgrade from an older version of the app and verify existing bookmarks are correctly migrated and still work.
2. Add a bookmark with a password, restart the app, and verify it still connects.
3. Check the app data directory (`/data/data/com.freerdp.afreerdp/databases/`) and verify the `bookmarks.db` file is fully encrypted and no `.wal` or `.shm` files are left readable.

**Environments Tested:**
Emulators: Android 7.0 (API 24), Android 14 (API 34)
Physical Device: Android 16 (API 36)

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/data/AppDatabase.java
+++ b/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/data/AppDatabase.java
@@ -5,11 +5,18 @@ package com.freerdp.freerdpcore.data;
 import android.content.Context;
 
 import androidx.annotation.NonNull;
+
+import java.io.File;
 import androidx.room.Database;
 import androidx.room.Room;
 import androidx.room.RoomDatabase;
 import androidx.room.migration.Migration;
 import androidx.sqlite.db.SupportSQLiteDatabase;
+
+import com.freerdp.freerdpcore.security.KeystoreHelper;
+
+import net.zetetic.database.sqlcipher.SQLiteDatabase;
+import net.zetetic.database.sqlcipher.SupportOpenHelperFactory;
 
 @Database(entities = { BookmarkEntity.class }, version = AppDatabase.DB_VERSION,
           exportSchema = false)
@@ -17,6 +24,11 @@ public abstract class AppDatabase extends RoomDatabase
 {
 	static final int DB_VERSION = 11;
 	private static final String DB_NAME = "bookmarks.db";
+
+	static
+	{
+		System.loadLibrary("sqlcipher");
+	}
 
 	private static volatile AppDatabase instance;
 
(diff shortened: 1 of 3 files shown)


akallabeth:
@svncibrahim 
* might be a better idea to just use https://github.com/sqlcipher/sqlcipher as the database for `room`, then everything is encrypted.
* might also be a good idea to use a hardware bound token as a key by default (so you don't need to enter a password but the database is only usable on the same hardware)

svncibrahim:
@akallabeth Thanks for the review!

**Hardware-bound token:** The KeystoreHelper in this PR already does exactly this. It natively uses [AndroidKeyStore](https://developer.android.com/privacy-and-security/keystore) (TEE) to lock the encryption key to the device's hardware. No password is required for normal usage.

**SQLCipher:**  I avoided it to keep a purely native Android solution without introducing external .so dependencies to the project.

**Export/Import:** Even with a fully encrypted SQLCipher database, a hardware-bound .db file cannot be transferred to a new device. We still strictly need this PR's JSON + Master Password flow to securely move bookmarks between devices.

If the project prefers full-database encryption and adding the SQLCipher dependency is acceptable, I’m happy to update the PR. Let me know!

akallabeth:
@svncibrahim
* I agree with the import/export, there a separate (not hardware bound) secret is required.
* for internal database a fully encrypted solution might still be preferred though. (no need to check if a field is delicate or not)
* note for import: the standard way to exchange configuration for `RDP` sessions are `rdp` files, which use a `key:type:value` format, the core library supports parsing this, but not sure this can be reused for the java components

svncibrahim:
@akallabeth 
- Thanks for the feedback! I agree that encrypting the entire database with SQLCipher is a much cleaner approach, so I will refactor the PR to implement it.  
- I also think using the standard .rdp format for import/export is a great idea, and I will investigate how we can properly parse and handle it on the Java side.

akallabeth:
>     * I also think using the standard .rdp format for import/export is a great idea, and I will investigate how we can properly parse and handle it on the Java side.

if custom keys are used all elements could be mapped. (best use some common prefix so it is obvious that it is a `aFreeRDP` extension)

encrypted credentials could be implemented by adding a key that tells that encryption is active (or just always expect it on?)

svncibrahim:
agreed on custom keys, using an `afreerdp` prefix makes sense.

for credentials, standard windows rdp already expects them to be encrypted by default anyway. we can just assume encryption is always on too. but to avoid mixing our encrypted blob with Windows DPAPI, maybe we should use a custom key like `afreerdp_password`. What do you think?

akallabeth:
@svncibrahim yes, the standard keys for credentials do not work properly (the https://learn.microsoft.com/en-us/windows/win32/api/dpapi/nf-dpapi-cryptprotectdata function only works with the same user account as well)

akallabeth:
@svncibrahim I think this pr should be split, one for the encrypted db and one for the import/export function

svncibrahim:
@akallabeth Agreed. I wanted to finish the CMake cleanups first. Now that they're done, I'll split the PR and work on both parts this week.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

akallabeth:
@svncibrahim nice implementation!

But I did try to debug in android studio and this was the result (looks like something on the gradle side is missing):

```
FATAL EXCEPTION: pool-4-thread-1
Process: com.freerdp.afreerdp, PID: 14669
java.lang.UnsatisfiedLinkError: No implementation found for long net.zetetic.database.sqlcipher.SQLiteConnection.nativeOpen(java.lang.String, int, java.lang.String, boolean, boolean) (tried Java_net_zetetic_database_sqlcipher_SQLiteConnection_nativeOpen and Java_net_zetetic_database_sqlcipher_SQLiteConnection_nativeOpen__Ljava_lang_String_2ILjava_lang_String_2ZZ) - is the library loaded, e.g. System.loadLibrary?
	at net.zetetic.database.sqlcipher.SQLiteConnection.nativeOpen(Native Method)
	at net.zetetic.database.sqlcipher.SQLiteConnection.open(SQLiteConnection.java:226)
	at net.zetetic.database.sqlcipher.SQLiteConnection.open(SQLiteConnection.java:202)
	at net.zetetic.database.sqlcipher.SQLiteConnectionPool.openConnectionLocked(SQLiteConnectionPool.java:474)
	at net.zetetic.database.sqlcipher.SQLiteConnectionPool.open(SQLiteConnectionPool.java:188)
	at net.zetetic.database.sqlcipher.SQLiteConnectionPool.open(SQLiteConnectionPool.java:180)
	at net.zetetic.database.sqlcipher.SQLiteDatabase.openInner(SQLiteDatabase.java:1032)
	at net.zetetic.database.sqlcipher.SQLiteDatabase.open(SQLiteDatabase.java:1017)
	at net.zetetic.database.sqlcipher.SQLiteDatabase.openDatabase(SQLiteDatabase.java:844)
	at net.zetetic.database.sqlcipher.SQLiteOpenHelper.getDatabaseLocked(SQLiteOpenHelper.java:359)
	at net.zetetic.database.sqlcipher.SQLiteOpenHelper.getWritableDatabase(SQLiteOpenHelper.java:278)
	at net.zetetic.database.sqlcipher.SupportHelper.getWritableDatabase(SupportHelper.java:60)
	at androidx.sqlite.driver.SupportSQLiteDriver.open(SupportSQLiteDriver.android.kt:57)
	at androidx.sqlite.driver.SupportSQLiteDriver.open(SupportSQLiteDriver.android.kt:33)
	at androidx.room.coroutines.PassthroughConnectionPool.connection$lambda$0(PassthroughConnectionPool.kt:47)
	at androidx.room.coroutines.PassthroughConnectionPool.$r8$lambda$5PfN97KPsXXXOapSC9IzcedL70w(Unknown Source:0)
	at androidx.room.coroutines.PassthroughConnectionPool$$ExternalSyntheticLambda0.invoke(D8$$SyntheticClass:0)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:83)
	at androidx.room.coroutines.PassthroughConnectionPool.useConnection(PassthroughConnectionPool.kt:58)
	at androidx.room.RoomConnectionManager.useConnection(RoomConnectionManager.android.kt:138)
	at androidx.room.RoomDatabase.useConnection(RoomDatabase.android.kt:619)
	at androidx.room.util.DBUtil__DBUtil_androidKt$performBlocking$1$1.invokeSuspend(DBUtil.android.kt:260)
	at androidx.room.util.DBUtil__DBUtil_androidKt$performBlocking$1$1.invoke(Unknown Source:8)
	at androidx.room.util.DBUtil__DBUtil_androidKt$performBlocking$1$1.invoke(Unknown Source:4)
	at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:42)
	at kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(Builders.common.kt:156)
	at kotlinx.coroutines.BuildersKt.withContext(Unknown Source:1)
	at androidx.room.util.DBUtil__DBUtil_androidKt$performBlocking$1.invokeSuspend(DBUtil.android.kt:72)
	at androidx.room.util.DBUtil__DBUtil_androidKt$performBlocking$1.invoke(Unknown Source:8)
	at androidx.room.util.DBUtil__DBUtil_androidKt$performBlocking$1.invoke(Unknown Source:4)
	at androidx.room.coroutines.RunBlockingUninterruptible_androidKt$runBlockingUninterruptible$1$1.invokeSuspend(RunBlockingUninterruptible.android.kt:52)
	at androidx.room.coroutines.RunBlockingUninterruptible_androidKt$runBlockingUninterruptible$1$1.invoke(Unknown Source:8)
	at androidx.room.coroutines.RunBlockingUninterruptible_androidKt$runBlockingUninterruptible$1$1.invoke(Unknown Source:4)
	at kotlinx.coroutines.intrinsics.UndispatchedKt.startCoroutineUndispatched(Undispatched.kt:20)
	at kotlinx.coroutines.CoroutineStart.invoke(CoroutineStart.kt:360)
	at kotlinx.coroutines.AbstractCoroutine.start(AbstractCoroutine.kt:124)
	at kotlinx.coroutines.BuildersKt__Builders_commonKt.launch(Builders.common.kt:52)
	at kotlinx.coroutines.BuildersKt.launch(Unknown Source:1)
```

svncibrahim:
Thanks for testing! It was working on my side, so I likely missed something in the setup. I’ll investigate

svncibrahim:
Okay, I found the issue. The library was being loaded from the wrong place.
I fixed it, can you test again?

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/data/AppDatabase.java
```diff
@@ -30,8 +37,13 @@ public static AppDatabase getInstance(Context context)
 			{
 				if (instance == null)
 				{
+					System.loadLibrary("sqlcipher");
```
akallabeth: would it not be better to load it from a `static {}` block?

akallabeth: like here: https://github.com/FreeRDP/FreeRDP/blob/531c4b1e346542d9d8e945c2c89e9c891ad15604/client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java#L75

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/data/AppDatabase.java
svncibrahim: Done, thanks!

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12690, opened 2026-04-26
rdpecam-v4l: stop the capture thread when streaming is cleared

### LINKED ISSUE #12689: RDPECAM: Camera light remains on after streaming ends
FreeRDP 3.24.0 connecting to a Windows server from a FreeBSD client. But the issue was confirmed by a Linux user as well.

It appears the `/dev/video0` is never closed.

The cause is this only check in the while loop: if the camera sends frames fast enough, the streaming never ends, and the camera remains open.
https://github.com/FreeRDP/FreeRDP/blob/89a8ed852bfdee071f04d955079d937e57165452/channels/rdpecam/client/v4l/camera_v4l.c#L477

(This code hasn't changed since 3.24.0.)

Defenso-QTH:
If a camera fills buffers as fast as we dequeue them, the DQBUF inner loop in cam_v4l_stream_capture_thread() never returns -1 and the thread can't reach the outer `while (stream->streaming)` check. cam_v4l_stream_stop()'s WaitForSingleObject() on the capture thread therefore blocks forever, VIDIOC_STREAMOFF and close(fd) are never reached, and /dev/videoN stays open until the client process exits. As a consequence the webcam activity LED stays on after the user closes the remote app.

Re-check stream->streaming in the inner loop so the thread can unwind between two DQBUFs; the existing stop path then closes the fd within one capture-thread iteration.

Sponsored-by: Defenso

Fixes #12689

### DIFF
--- a/channels/rdpecam/client/v4l/camera_v4l.c
+++ b/channels/rdpecam/client/v4l/camera_v4l.c
@@ -473,8 +473,8 @@ static DWORD WINAPI cam_v4l_stream_capture_thread(LPVOID param)
 			buf.type = V4L2_BUF_TYPE_VIDEO_CAPTURE;
 			buf.memory = V4L2_MEMORY_MMAP;
 
-			/* dequeue buffers until empty */
-			while (ioctl(fd, VIDIOC_DQBUF, &buf) != -1)
+			/* dequeue buffers until empty, or until we are asked to stop */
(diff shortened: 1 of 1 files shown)


akallabeth (approved):
LGTM,
nice catch!

akallabeth:
@freerdp-bot test
@Defenso-QTH ok, that is a nasty one. Could not reproduce this on my end although the machine is more than fast enought to actually hit this.

PR #12706, opened 2026-04-28
[cmake,deps] use alias target for sso-mib

akallabeth:
* Alias target for sso-mib
* Remove externalproject build of sso-mib


hardening:
I guess this must fix what was preventing to build sso-mib+freerdp with accendino

Issue #12627, opened 2026-04-18
xfreerdp hang/freeze KDE desktop for 1s if kliper and fcitx5 clipboard both enabled.


jackyzy823:
# **Describe the bug**

xfreerdp hang/freeze KDE desktop for 1s if kliper and fcitx5 clipboard both enabled.

# **To Reproduce**

1. Install Fedora KDE 43 and update all to the latest
2. Install fcitx5 and set KDE settings Virtual Keyboard to fcitx5 and restart
3. Make sure klipper is enabled (kde system tray settings ->general ->clipboard -> show when relevant) and fcitx5 clipboard addon is enabled (fcitx5 config -> addons -> advanced -> clipboard checkbox is checked). (And that's why i [said](https://bugs.kde.org/show_bug.cgi?id=518205#c9) i can't reproduce this in a new KDE installation VM)
3. Use `xfreerdp` to connect any remtoe rdp service (gnome / windows)
4. Open a kwrite/kate locally, copy (Ctrl+C) some text from remote,  try using mouse to click on kwrite/kate icon on local taskbar
5. You could observe a freeze/hang (about 1s) on kde desktop
6. If this behavior is not obvious for you , you could install a buggy version (6.24.0) [kf6-kguiaddons](https://kojipkgs.fedoraproject.org//packages/kf6-kguiaddons/6.24.0/1.fc43/x86_64/kf6-kguiaddons-6.24.0-1.fc43.x86_64.rpm) . In this version poll timeout is mistakenly extended from 1s to [10s](https://invent.kde.org/frameworks/kguiaddons/-/commit/c43abe7587c2018facef7ca2f16cd1c5480586b7#line_cc63e0754_A280).
7. If you disable either klipper (kde system tray settings ->general ->clipboard ->  never show disabled) or fcitx5 clipboard addon , the behavior disappear.

# **Expected behavior**

Copy operation won't hang KDE desktop.

# **Application details**

* FreeRDP version (`xfreerdp /version`) 3.24.2
* OS version connecting to (server side) Windows 10

# **Environment (please complete the following information):**

 - OS: Linux
 - Version/Distribution: Fedora 43 , KDE 6.6.4, KDE framework 6.25.0 (kguiaddons could be 6.24.0)
 - Architecture: x86_64

# **TL;DR**

[...]

PR #12707, opened 2026-04-28
[client,x11] handling multiple selection requests and queued to send server format data request.

jackyzy823:
When format data response received, dispatching selection notify to matched selection requests.

For #12627

-----

After reading icccm docs carefully, i recognized that responding with property = none in selection notify is actually correct, this means conversion failed from owner side.

So here what i do is to make ~~duplicate~~ multiple selection requests at the same time won't be treated as a failure.

When the first multiple selection request received in process_selection_request after xfreerdp announce own the clipboard,  it is put in pending_responds, and since there's only one in pending_responds , a format data request was sent to remote.

In the window of after format data request sent and before format data response received, if another selection request received, if it has the same source (remote) format id , it will be put in to pending_responds, otherwise, it will be put in queued_responds.

After format data response received, we iterate the pending_responds, if target format was found in cache, then return cached dst data, else convert data to desired target format. and then provide data and send selection notify.

After  all requests in pending_responds is processed, we change queued_responds, if not null, put the first one (and all other having same source format id) to pending_responds, and send format data request to remote.

---
Well code is quite ugly .....

### DIFF
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -78,6 +78,13 @@ typedef struct
 	char* formatName;
 } RequestedFormat;
 
+typedef struct
+{
+	XSelectionEvent* expectedResponse;
+	RequestedFormat* requestedFormat;
+	BOOL data_raw_format;
+} SelectionResponse;
+
 struct xf_clipboard
 {
 	xfContext* xfc;
@@ -110,11 +117,8 @@ struct xf_clipboard
 	wHashTable* cachedData;
 	wHashTable* cachedRawData;
 
-	BOOL data_raw_format;
-
-	RequestedFormat* requestedFormat;
-
-	XSelectionEvent* respond;
+	wArrayList* pending_responses;
+	wQueue* queued_responses;
 
 	Window owner;
 	BOOL sync;
@@ -195,6 +199,17 @@ static BOOL requested_format_replace(RequestedFormat** ppRequestedFormat, UINT32
 
 	*ppRequestedFormat = requested;
 	return TRUE;
+}
+
+static void selection_response_free(void* ptr)
+{
+	SelectionResponse* selection_response = (SelectionResponse*)ptr;
+	if (!selection_response)
+		return;
+
+	free(selection_response->expectedResponse);
+	requested_format_free(&selection_response->requestedFormat);
+	free(selection_response);
 }
 
 static void xf_cached_data_free(void* ptr)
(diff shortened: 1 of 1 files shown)


client/X11/xf_cliprdr.c
```diff
@@ -110,11 +117,8 @@ struct xf_clipboard
 	wHashTable* cachedData;
 	wHashTable* cachedRawData;
 
-	BOOL data_raw_format;
-
-	RequestedFormat* requestedFormat;
-
-	XSelectionEvent* respond;
+	wArrayList* pending_responds;
```
akallabeth: typo, should be `responses`

client/X11/xf_cliprdr.c
```diff
@@ -110,11 +117,8 @@ struct xf_clipboard
...
+	wQueue* queued_responds;
```
akallabeth: typo, should be `responses`

client/X11/xf_cliprdr.c
```diff
@@ -78,6 +78,13 @@ typedef struct
 	char* formatName;
 } RequestedFormat;
 
+typedef struct
+{
+	XSelectionEvent* respond;
```
akallabeth: typo, should be `response` or better `expectedResponse`

client/X11/xf_cliprdr.c
```diff
@@ -78,6 +78,13 @@ typedef struct
...
+	RequestedFormat* requestedFormat;
+	BOOL data_raw_format;
+} SelectionRespond;
```
akallabeth: typo, should be `SelectionResponse`

client/X11/xf_cliprdr.c
```diff
@@ -197,6 +201,16 @@ static BOOL requested_format_replace(RequestedFormat** ppRequestedFormat, UINT32
 	return TRUE;
 }
 
+static void selection_respond_free(void* ptr)
+{
+	SelectionRespond* selection_respond = (SelectionRespond*)ptr;
```
akallabeth: `free` functions usually allow `nullptr` as argument

client/X11/xf_cliprdr.c
```diff
@@ -2193,198 +2279,299 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
...
-			free(pDstData);
-			return CHANNEL_RC_OK;
-		}
-		if (!HashTable_Insert(clipboard->cachedData, format_to_cache_slot(dstFormatId),
-		                      cached_data))
+		UINT32 nextFormatId = next->requestedFormat->formatToRequest;
+		const xfCliprdrFormat* cformat =
+		    xf_cliprdr_get_client_format_by_atom(clipboard, next->respond->target);
+
+		wQueue* backlog = Queue_New(FALSE, -1, -1);
```
akallabeth, line 2539: missing `nullptr` check (oom et al)

akallabeth (changes requested):
Overall LGTM.

some nitpicks:
* variable naming
* `xf_cliprdr_server_format_data_response` has grown quite huge, might benefit from splitting into smaller functions
* some `nullptr` checks are missing

in any case, thank you, nice work!

akallabeth:
@freerdp-bot test

client/X11/xf_cliprdr.c
akallabeth: also, there can only be 1 pending response (there are no identifiers in the messages, so parallel requests do not work)

jackyzy823: > also, there can only be 1 pending response (there are no identifiers in the messages, so parallel requests do not work)

Hi , i think if these requests having same remote format id  between the time window of after the format data request of the first one, and the format data response. They could be appended in the pending list.

Also after processing current format data response, it would be nature to append all request which having next remote format id into pending list.

akallabeth: :+1:

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
there are a few warnings left to address, see
https://ci.freerdp.com/job/code-quality-checker/12346/clang/

(return checks, unused variables and one `// NOLINTNEXTLINE(clang-analyzer-unix.Malloc): sometext` for the allocated memories

client/X11/xf_cliprdr.c
```diff
@@ -2193,198 +2281,303 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
...
+	SelectionResponse* next = Queue_Peek(clipboard->queued_responses);
+	if (next)
 	{
-		cached_data = xf_cached_data_new(pDstData, DstSize);
-		if (!cached_data)
+		UINT32 nextFormatId = next->requestedFormat->formatToRequest;
+		const xfCliprdrFormat* cformat =
+		    xf_cliprdr_get_client_format_by_atom(clipboard, next->expectedResponse->target);
+
+		wQueue* backlog = Queue_New(FALSE, -1, -1);
```
akallabeth, line 2539: looking at this again, why use this intermediate queue at all?

akallabeth:
@jackyzy823 ok, taken your commit and did the cleanups. Added to https://github.com/FreeRDP/FreeRDP/pull/12712

PR #12709, opened 2026-04-29
[core,settings] add auto reconnect triggered flag

akallabeth:
* Add `FreeRDP_SessionHasBeenReconnected` setting indicating that a session is (or was) reconnected
* Set flag on `rdp_client_reconnect`

### DIFF
--- a/client/common/test/default-settings.json
+++ b/client/common/test/default-settings.json
@@ -42,6 +42,7 @@
 		"FreeRDP_HiDefRemoteApp": true,
 		"FreeRDP_IPv6Enabled": false,
 		"FreeRDP_AutoReconnectionEnabled": false,
+		"FreeRDP_SessionHasBeenReconnected": false,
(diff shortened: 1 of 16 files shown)


mfleisz (commented):
I think the name `AutoReconnectionInProgress` is a bit misleading:
- I wouldn't call it `AutoReconnect` as it is more of a client setting and has actually nothing to do with calling `freerdp_reconnect`
-  It suggests that we are currently in the process of an auto reconnect but the value remains true, even after we successfully reconnected (reconnect no longer in progress)
- Possible alternatives: `IsReconnectAttempt` or `IsReconnect` or maybe an enum would be even more expressive i.e.:

```
enum ConnectionAttempt
{
    Initial,
    Reconnect
};
```

akallabeth:
@mfleisz hmm, `SessionHasBeenReconnected` might be a better name then?

mfleisz (approved):

PR #12694, opened 2026-04-27
Add VideoToolbox H.264 support for ffmpeg

fstanis:
This is a simpler implementation than #12337 - it uses ffmpeg (setting `-hwaccel videotoolbox`) and largely mirrors `WITH_VAAPI`.

Tested by building with `-DWITH_VIDEOTOOLBOX=ON` and then connecting to a windows client on macos.

PR #12711, opened 2026-04-29
Force YUV420P when videotoolbox is used

fstanis:
Quick follow-up to #12694 to address a VideoToolbox quirk: if it outputs NV12 (which is the default), ffmpeg rejects it because it's a semi-planar format and the output expects a planar one.

This PR forces it to use YUV420P by setting `sw_format`. There may be a better way to do this, but this seemed like the easiest fix.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #12741, opened 2026-05-06
Fix for GHSA-j9q5-7g8m-jc9v, disable NDR pointer aliasing

hardening:
The usage of aliased pointers can be used to trigger a double free, or a type confusion (using the same memory location twice and pretend it's 2 different types of objects). As pointer aliasing is never seen in practice with RDPEAR, let's just return an error when we see an aliased pointer.


akallabeth (approved):
LGTM

PR #12753, opened 2026-05-07
[core,codec] Fix invalid overlap check

mfleisz:
The overlap check returned true if src and dst were placed directly next to each other in memory.

`dststart` can be located directly at the address `srcend` and `dstend` can be at the exact same address as `srcstart` so do not check for equality at these two places.


akallabeth (approved):
LGTM

PR #12754, opened 2026-05-07
fix(crypto): check if should retry on reading certificate

SLAVONchick:
This PR fixes the issue when certificate can be read, but the BIO_read function returns status < 0. 
According to [OpenSSL docs](https://docs.openssl.org/master/man3/BIO_read/#notes) this might not be an indication of an error.
In my particular case, when I was trying to connect to corporate remote desktop, the certificate was read successfully from RD Gateway, but the status was still `-1`.

### DIFF
--- a/libfreerdp/crypto/certificate.c
+++ b/libfreerdp/crypto/certificate.c
@@ -1556,7 +1556,8 @@ static BOOL bio_read_pem(BIO* bio, char** ppem, size_t* plength)
 		ERR_clear_error();
 
 		const int status = BIO_read(bio, &pem[offset], (int)(length - offset));
-		if (status < 0)
+		const int should_retry = BIO_should_retry(bio);
+		if (status < 0 && should_retry <= 0)
 		{
 			WLog_ERR(TAG, "failed to read certificate");
 			goto fail;
--- a/libfreerdp/crypto/certificate.c
+++ b/libfreerdp/crypto/certificate.c
@@ -28,6 +28,7 @@
 #include <stdio.h>
 #include <string.h>
 
+#include <sys/select.h>
 #include <winpr/assert.h>
 #include <winpr/wtypes.h>
 #include <winpr/crt.h>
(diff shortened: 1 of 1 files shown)


libfreerdp/crypto/certificate.c
```diff
@@ -1556,7 +1556,8 @@ static BOOL bio_read_pem(BIO* bio, char** ppem, size_t* plength)
 		ERR_clear_error();
 
 		const int status = BIO_read(bio, &pem[offset], (int)(length - offset));
-		if (status < 0)
+		const int should_retry = BIO_should_retry(bio);
+		if (status < 0 && should_retry <= 0)
```
akallabeth: this fix is buggy:
* `status = -1`
* `should_retry = 1`
* down below `offset += (size_t)-1` <-- kaboom

akallabeth (changes requested):

libfreerdp/crypto/certificate.c
SLAVONchick: that's right, thank you! I've implemented waiting on the socket in this case. Though, I'm not quite sure what the correct timeout should be.

akallabeth (changes requested):
wtf?
please don't do AI stuff here, this is just awful.

akallabeth:
ok, replaced by https://github.com/FreeRDP/FreeRDP/pull/12755

SLAVONchick:
Excuse me, I didn't do use AI at all. And #12755 will just hang in my case.

akallabeth:
* well, then what did you try to do with the `select` in that part of the code? 
* can you give more details on which problem you try to fix? ( openssl version, cerfiticate, ...) as this smells fishy. (We do only use these with a `BIO_s_mem()` which already contain all the data, so where does your timeout come from?

akallabeth:
refactored the whole thing with https://github.com/FreeRDP/FreeRDP/pull/12758

SLAVONchick:
> * well, then what did you try to do with the `select` in that part of the code?
> 
>     * can you give more details on which problem you try to fix? ( openssl version, cerfiticate, ...) as this smells fishy. (We do only use these with a `BIO_s_mem()` which already contain all the data, so where does your timeout come from?

As I pointed out in the comments I've tried to wait until the socket is ready to be read from.

Yes, you're right, the certificate data is read correctly, but then it tries to read more using `BIO_read()` and in normal situation it would return 0 and correctly end the loop. In my situation the `BIO_read()` returns -1 and `BIO_should_retry()` returns 8, so the loop continues executing.

I'm using OpenSSL version 3.5.5.

akallabeth:
@SLAVONchick ok, sounds strange you have to wait on this routine which should already have completed writing.
Maybe a `BIO_flush` was missing?

anyway, did rewrite the whole thing, does that work for you?

SLAVONchick:
@akallabeth Discarded of all the changes I've made and rebuilt from scratch and yeah, now it connects without any issues. I really appreciate the help.

PR #12758, opened 2026-05-08
Unify PEM read routines

akallabeth:


akallabeth:
@SLAVONchick did have a look on that routine you have trouble with and found we had the same stuff implemented multiple times.
can you check if this pr works for you as well?

SLAVONchick:
@akallabeth Hi! Sorry for the delay, tested against master and it seems to work without any issues.

PR #12759, opened 2026-05-08
Fix sdl3 clipboard for files

jackyzy823:
Most files related code for clipboard have been implemented, so i just mimic the xfreerdp's one to fill the gap and make it work.

### DIFF
--- a/client/SDL/SDL3/sdl_clip.cpp
+++ b/client/SDL/SDL3/sdl_clip.cpp
@@ -138,6 +138,8 @@ sdlClip::sdlClip(SdlContext* sdl)
 	std::stringstream ss;
 	ss << s_mime_freerdp_update << "-" << _uuid;
 	_mime_uuid = ss.str();
+
+	std::ignore = cliprdr_file_context_set_locally_available(_file, TRUE);
 }
 
 sdlClip::~sdlClip()
@@ -231,6 +233,7 @@ bool sdlClip::handleEvent(const SDL_ClipboardEvent& ev)
 
 	bool textPushed = false;
 	bool imgPushed = false;
+	bool filePushed = false;
 
 	for (size_t i = 0; i < nformats; i++)
 	{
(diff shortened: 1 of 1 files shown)


client/SDL/SDL3/sdl_clip.cpp
```diff
@@ -526,6 +540,15 @@ UINT sdlClip::ReceiveServerFormatList(CliprdrClientContext* context,
...
 
 	clipboard->_current_mimetypes.clear();
+
+	{
+		ClipboardLockGuard systemlock(clipboard->_system);
+		std::scoped_lock lock(clipboard->_lock);
+		auto res = cliprdr_file_context_notify_new_server_format_list(filecontext);
+		if (res != CHANNEL_RC_OK)
+			return res;
+	}
```
jackyzy823, line 550: Actually, not sure if this should be placed after `clipboard->_current_mimetypes.clear();` or before it. But i think `_current_mimetypes` should be empty if `cliprdr_file_context_notify_new_server_format_list` failed.

akallabeth:
@freerdp-bot test

akallabeth:
@jackyzy823 the sdl client also runs on windows and mac os, so do you properly hide the formats if the file clipboard is not available?

jackyzy823:
Struggling to cross compile sdl-freerdp for win with mingw64, however it compiles. 
And also download sdl-freerdp (aka original one) for win from jenkins ci .

<del>
It looks like original one doesn't support well on windows too.

Behavior
1. Remote to local text copy: OK
2. Local to remote text copy: not work
3. Remote to local file copy: OK
4. Local to remote file copy: not work and a liitle buggy, mouse show busy cursor.
And mine one behaves the same ....
</del>


----------

~~ok after testing , i found that windows sdl-freerdp won't go through sdlclip, in in contrast, sdlclip will bother clipboard functionality.~~

Wrong, it's virtualbox guest additions :(

akallabeth:
@jackyzy823 ok, working on linux, lets just merge and check the next nightly.
we can always do a follow up if there should be issues there.

akallabeth (approved):
LGTM

PR #12763, opened 2026-05-10
add sso-mib dependency to client cmake and pkg-config

sfalken:
When built WITH_SSO_MIB=ON the dependencies on sso-mib aren't being exported to dependent builds.  (e.g. when building krdp against freerdp)

This change adds that dependency to the .cmake and .pc files for the FreeRDP-Client so "downstream" builds will pick it up, when FreeRDP is built with sso-mib


Conan-Kudo (approved):

akallabeth:
@sfalken why should this dependency be exposed?
all use is entirely internal, if a downstream project wants to use it it should link it in themselves?

Conan-Kudo:
Without it, krdc fails to compile properly because of a missing link dependency. So it doesn't appear to be entirely internal.

sfalken:
> @sfalken why should this dependency be exposed? all use is entirely internal, if a downstream project wants to use it it should link it in themselves?

Sorry.  I misspoke, because I'm apparently suffering from dumb this morning.    This dependency is passed down to krdc, not krdp, and only affects rdp *clients* that build against freerdp.

That being said, hold off.   because of my dumb, I was building krdp against freerdp.  Which wasn't the problem I was trying to solve in the first place.  krdc is the problem. 

I'm not going to abandon this PR as of yet, but this PR doesn't actually fix the problem with krdc right now.

akallabeth:
@sfalken the dependency is internal, there is no doubt about that. (e.g. `xfreerdp` or `sdl-freerdp` use this without being linked to it)
so, if something downstream fails to link it is missing a dependency itself.

akallabeth:
ok, just compiled krdc 6c8b9b10902c13c3c64d632485fe97a33e8e6fc0 against a freerdp 1d93a2daf888c2cd7debc98c6d8bf3553853af08 with `-DWITH_SSO_MIB=ON` and there are no linking issues?

sfalken:
This patch is completely un-neccessary, and I was chasing a red herring.   Sorry for the noise.

PR #12768, opened 2026-05-11
fix 400 (actually too long request) when upgrading to websockets on connecting to Azure Virtual Desktops

tothi:
Received HTTP 400 on gateway response while trying to upgrade to websockets. Debugged the 400 message details and it was a too long request. Found that the Bearer token was sent duplicate times: once in the header and also as a GET parameter.

The fix attempt is not sending the Bearer token in headers if it is sent as (the ClmTk) GET parameter also.

Moreover, also removed the user agent from the GET parameters because it is also sent via Headers in order to make the request shorter.

Seems to be working in my situtation (where the Bearer token was really long).

Reference: https://github.com/FreeRDP/FreeRDP/issues/11951


akallabeth:
@tothi thank you, will check (hopefully) this or next week. (sorry, little busy)

akallabeth:
@freerdp-bot test

akallabeth:
@tothi hmm, do you have some details/links/... regarding your new refresh token?
is there a corresponding `rdpw` file key? (all others have such an equivalent)

akallabeth:
@tothi ok, taken your first commit and added a second bugfix with https://github.com/FreeRDP/FreeRDP/pull/12770
the refresh token changes need some discussion before we can merge these.

akallabeth:
@tothi ping? can you give some insight in why you added the refresh token commit?

tothi:
> @tothi ping? can you give some insight in why you added the refresh token commit?

oh no. accidentally pushed that refresh token commit to the same branch before the pull request was merged. sorry. it is not relevant, it is a feature update I use now (using a refresh token on host authentication after the gateway instead of proceeding with the std oauth flow).

going to fix it...

akallabeth:
ok, closing this then, the other commits have already been merged.

PR #12777, opened 2026-05-13
Android db non null

akallabeth:


svncibrahim:
Hi @akallabeth 

Tested on a device with an existing install , still crashes with the same `IllegalStateException`:

```
Migration didn't properly handle: bookmarks(com.freerdp.freerdpcore.data.BookmarkEntity).
Expected:  Column { name = 'async_channel', defaultValue = 'false' }
Found:     Column { name = 'async_channel', defaultValue = 'undefined' }
```

Adding `defaultValue` to `@ColumnInfo` tells Room what to *expect* in the schema, but doesn't rewrite the existing database. The table was originally created in `MIGRATION_10_11` without `DEFAULT` clauses, so upgraded databases will always read back `'undefined'` for those columns.

A new migration that recreates the table with proper `DEFAULT` clauses is needed to fix this for existing users. Full logcat attached.

[logcat.txt](https://github.com/user-attachments/files/27700452/logcat.txt)

akallabeth:
@svncibrahim yes, not ready, doing some refinements.

akallabeth:
@svncibrahim ok, this should work now, added a new scheme version and a migration path to add the missing default values in the existing database.

svncibrahim:
Hi @akallabeth 

Two bugs found after testing on an existing install:

- **Data corruption in MIGRATION_13_14**

All text and integer fields (label, hostname, username, etc.) are overwritten with their own column name after migration. Screenshot attached.


- **Crash on any write operation**  HistoryEntity schema hash mismatch

```
IllegalStateException: Room cannot verify the data integrity.
Expected identity hash: d4c907792fd5ffa7e999ded418cd6be5
Found:                  8df863c636f12f85de62e684d3451745
```
Logcat attached. [logcat.txt](https://github.com/user-attachments/files/27710389/logcat.txt)



Screenshot_2026-05-13-15-52-20-18_bfe5c884e366e53e8d644a47ee8284f1

svncibrahim:
To reproduce:

1. Check out an older DB version (e.g. abfce2c8), build and install the app, add a bookmark with a recognizable label.
2. Check out upstream/master, build and install over the existing app (do not uninstall).
3. Launch the app, all bookmark fields now show their column name instead of the saved value.

svncibrahim:
This is fixed by #12783

PR #12782, opened 2026-05-14
[winpr,sspi] replace password-length heuristic with explicit hash

ramnes:
WinPR currently detects pass-the-hash credentials by adding 512 to `PasswordLength` and checking `PasswordLength > 512` downstream. So any plaintext password longer than 512 chars (a JWT used as the RDP password, for example) is then mistakenly going through `ntlm_convert_password_hash`, and then NLA fails with `PasswordHash has invalid length`.

This PR adds a new dedicated flag instead: `SEC_WINPR_AUTH_IDENTITY_PASSWORD_HASH`. It's set on the PTH path and read by `ntlm_compute_ntlm_v2_hash`. I chose the `0x00800000` value to be high enough to sit clearly in the extension space above Microsoft's contiguous block, and low enough to stay clear of the `SSPIPFC_USE_MASK` (`0xFF000000`) reserved range.

`/pth:<hex>` behavior is unchanged on the wire.

In the meantime, I dropped the existing `SSPI_CREDENTIALS_PASSWORD_HASH` macro because it seems that it never got wired up, and because it's colliding with `SEC_WINNT_AUTH_IDENTITY_ANSI` anyway. Happy to bring it back if there's any reason to.

To reproduce the original bug:

```
xfreerdp /v:<host> /u:<user> /p:<a-password-of-more-than-512-chars>
```

Before: NLA fails with `PasswordHash has invalid length`. After: the long plaintext password is sent through and NLA proceeds normally.

`TestSspiPasswordHash` covers both the long-plaintext regression and the explicit-flag PTH path. PTH itself (`/pth:<32-hex>`) keeps working unchanged.

FWIW, the Devolutions/sspi-rs project fixed the same heuristic recently: https://github.com/Devolutions/sspi-rs/pull/619.


akallabeth:
@freerdp-bot test

hardening (approved):
That seems ok for me, the only question I have is that it changes the public API for that feature. Anyway we can probably live with it as: PTH is not that much used and the previous way to invoke it was so fishy...

hardening:
Note: we should probably heavily document these change in the release notes...

ramnes:
Thanks for the quick review and merge! Let me know if there’s anything else I can help with on this front. :)

PR #12783, opened 2026-05-15
[client,android] fix data corruption in MIGRATION_13_14

svncibrahim:
The UPDATE statements in MIGRATION_13_14 wrapped source column names in single quotes, causing SQLite to treat them as string literals. Every text field (label, hostname, username, etc.) was overwritten with its
own column name instead of the saved value after migration.

Remove the single quotes from column identifiers in all UPDATE statements.

Screenshot showing data corruption after migration, all bookmark fields display their column name (label, hostname, username, etc.) instead of the saved values.

Screenshot_2026-05-13-15-52-20-18_bfe5c884e366e53e8d644a47ee8284f1


akallabeth:
@freerdp-bot test

akallabeth:
@svncibrahim oh, indeed. nice fix, thx.

akallabeth (approved):
LGTM

PR #12793, opened 2026-05-18
feat(dev): Enable more options for nightly builds

a5ehren:
The current nightly builds do not exercise a lot of the optional build flags that are available, which can lead to problems later on. 

With that in mind, this PR:
- Attempts to detect a maximal config for a given distro+version
- Enables those options
- Builds the package
- Adds a local build script that launches containerized builds for all supported distros
- Adds libaom to the flatpak and enables AV1

These setups exposed 4 build errors:
- TSMF channel did not build against modern GStreamer due to type changes in Rectangle
- X11 disconnect helpers were not exported as public APIs, which blocked their use when the client is a shlib
- `TestNCryptSmartcard` cannot run under ASAN
- Ancient distros (gcc-7 on OS Leap 15) still had the C++ filesystem stdlib as "experimental" and a different so-name


cmake/Findaom.cmake
```diff
@@ -7,6 +7,8 @@
 #  AOM_INCLUDE_DIRS - the AOM include directory
 #  AOM_LIBRARIES - libaom library
 
+set(AOM_MIN_VERSION "2.0")
```
a5ehren, line 10: note here - libaom 2.0 is actually the first public/stable version released by AOM/Google.

Some very old distros that you techincally support, like Ubuntu 20 or Debian Bullseye, have beta versions of libaom that your av1 code builds against but provides such an awful experience that it may as well not build.

akallabeth (changes requested):
Not quite sure what to think about this.

there are some interesting and promising parts in this, but as of right now it looks like stuff jumbled together.

maybe split this up into separate commits by feature?

a5ehren:
> Not quite sure what to think about this.
> 
> there are some interesting and promising parts in this, but as of right now it looks like stuff jumbled together.
> 
> maybe split this up into separate commits by feature?

Sure, there some places that I can split it.

client/X11/xf_client.c
```diff
@@ -1717,7 +1717,7 @@ static DWORD WINAPI xf_client_thread(LPVOID param)
 	return exit_code;
 }
 
-int xf_exit_code_from_disconnect_reason(DWORD reason)
+FREERDP_API int xf_exit_code_from_disconnect_reason(DWORD reason)
```
akallabeth, line 1720: this is header only

cmake/CompilerSanitizerOptions.cmake
```diff
@@ -32,6 +32,37 @@ if(CMAKE_COMPILER_IS_CLANG OR CMAKE_COMPILER_IS_GNUCC)
     add_compile_options(-fsanitize=address)
     add_compile_options(-fsanitize-address-use-after-scope)
     add_link_options(-fsanitize=address)
+    if(CMAKE_C_COMPILER_ID MATCHES ".*Clang.*")
```
akallabeth, line 35: not sure why this code is required?

winpr/libwinpr/ncrypt/test/CMakeLists.txt
```diff
@@ -18,6 +18,10 @@ foreach(test ${${MODULE_PREFIX}_TESTS})
   add_test(${TestName} ${TESTING_OUTPUT_DIRECTORY}/${MODULE_NAME} ${TestName})
 endforeach()
 
+if(WITH_SANITIZE_ADDRESS)
```
akallabeth, line 21: our ci runs this test with `-fsanitize=address` all the time?

cmake/CompilerSanitizerOptions.cmake
a5ehren: This is to fix a build problem on Tumbleweed and Rawhide where the linker can't find Clang's ASAN library for some reason, so you have to force it into the rpath. There is probably a more elegant way to do it, but I haven't looked in detail.

akallabeth: ok.
well, we don't (generally) add such things to the main `CMake` files (already enough workarounds for various problems there, so no need to add another layer of these)

maybe you're just lacking a search path (`CMAKE_PREFIX_PATH` or `CMAKE_MODULE_PATH`) that you can add in the corresponding toolchain or preload file?

winpr/libwinpr/ncrypt/test/CMakeLists.txt
a5ehren: maybe caused by the cmake thing? I'll look at it again

a5ehren: This is the message, only shows up on Ubuntu 26.04 (which your CI runner is not using yet):
```
==6575==You are trying to dlopen a libpcsclite.so.1 shared library with RTLD_DEEPBIND flag which is incompatible with sanitizer runtime (see https://github.com/google/sanitizers/issues/611 for details). If you want to run libpcsclite.so.1 library under sanitizers please remove RTLD_DEEPBIND from dlopen flags.
```
Given that FreeRDP does not use that flag, I'm not entirely sure what's going on.

akallabeth: @a5ehren there was a bug in debian/sid a few weeks back (fixed in the meantime) - sounds like ubuntu did not (yet) apply the fix?

a5ehren: ah, cool. let me find that documentation to verify

a5ehren:
Superceded by 12814

PR #12811, opened 2026-05-26
Improve FIPS mode support

ondrejholy:
This series improves FIPS compliance across the board:
- Disable NTLM in FIPS mode — NTLM uses MD4/RC4 which are not FIPS-approved. When FIPS is active, AuthenticationPackageList is set to "!ntlm" so only Kerberos is used for NLA.
- Fix RC4 FIPS override for OpenSSL 3 — The winpr_RC4_New_Allow_FIPS function used EVP_CIPH_FLAG_NON_FIPS_ALLOW which is no longer honored in OpenSSL 3. Use EVP_CIPHER_fetch with "fips=no" property instead, matching the existing MD5 fix.
- Disable RDP Security in FIPS mode on OpenSSL 3.4+ — The RDP FIPS encryption method uses 3DES which is no longer approved in the OpenSSL 3.4 FIPS provider. On these versions, RDP Security is disabled so only NLA/TLS (using AES) is negotiated.
- Suppress expected FIPS cipher warnings — The crypto build-time checks produced misleading warnings for algorithms that are intentionally blocked under FIPS. Test _Allow_FIPS variants where available (RC4, MD5) and silently skip algorithms expected to be unavailable (MD4, HMAC-MD5). A single INFO message now summarizes what is unavailable.

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -325,14 +325,30 @@ BOOL rdp_client_connect(rdpRdp* rdp)
 	/* FIPS Mode forces the following and overrides the following(by happening later
 	 * in the command line processing):
 	 * 1. Forces the only supported RDP encryption method to be FIPS.
-	 * 2. Disables NTLM authentication (not FIPS compliant due to MD4/RC4). */
+	 * 2. Disables NTLM authentication (not FIPS compliant due to MD4/RC4).
+	 * 3. Disables RDP Security if 3DES is not available in the crypto provider. */
 	if (settings->FIPSMode || winpr_FIPSMode())
 	{
+
 		settings->EncryptionMethods = ENCRYPTION_METHOD_FIPS;
+
 		if (!settings->AuthenticationPackageList)
 		{
 			if (!freerdp_settings_set_string(settings, FreeRDP_AuthenticationPackageList, "!ntlm"))
 				return FALSE;
+		}
+
+		const BYTE key[WINPR_CIPHER_MAX_KEY_LENGTH] = WINPR_C_ARRAY_INIT;
+		const BYTE iv[WINPR_CIPHER_MAX_IV_LENGTH] = WINPR_C_ARRAY_INIT;
+		WINPR_CIPHER_CTX* ctx = winpr_Cipher_NewEx(WINPR_CIPHER_DES_EDE3_CBC, WINPR_ENCRYPT, key,
+		                                           sizeof(key), iv, sizeof(iv));
+		if (ctx)
+		{
+			winpr_Cipher_Free(ctx);
+		}
+		else
+		{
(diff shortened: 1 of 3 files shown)


libfreerdp/core/rdp.c
```diff
@@ -28,6 +28,11 @@
 #include <winpr/assert.h>
 #include <winpr/cast.h>
 #include <winpr/json.h>
+#include <winpr/ssl.h>
+
+#ifdef WITH_OPENSSL
```
akallabeth: this should not leak here, we try to keep the implementation includes in winpr or `crypto` subdirectory
the checks are not openssl specific but to `FIPS` compliance level I assume, so all implementations providing this will be affected.

libfreerdp/core/connection.c
```diff
@@ -324,10 +328,24 @@ BOOL rdp_client_connect(rdpRdp* rdp)
...
+
 		settings->EncryptionMethods = ENCRYPTION_METHOD_FIPS;
+
+		if (!settings->AuthenticationPackageList)
+		{
+			if (!freerdp_settings_set_string(settings, FreeRDP_AuthenticationPackageList, "!ntlm"))
+				return FALSE;
+		}
+
+#if defined(WITH_OPENSSL) && (OPENSSL_VERSION_NUMBER >= 0x30400000L)
```
akallabeth: this check should adhere to the comment, e.g. check for fips compliance and not implementation (e.g. openssl)

akallabeth (changes requested):
Overall LGTM.

TODO:
* Add the `FIPS` checks lower down so no implementation headers are required up at connection settings
* Is licensing in this mode working? That one is also using `RC4` and will prevent connections to systems that have that enabled (although it never uses any of these algorithms for encryption of the connection)

akallabeth:
@ondrejholy also, wondering why `TLS` is allowed but `NLA` with `NTLM` not (the former leaks plain text passwords to the server, so in any case worse than `NLA`, but that might be due to compliance with `FIPS` I assume?)

ondrejholy:
>     * Is licensing in this mode working? That one is also using `RC4` and will prevent connections to systems that have that enabled (although it never uses any of these algorithms for encryption of the connection)

Licensing is using `winpr_RC4_New_Allow_FIPS`, which this PR fixes.

> @ondrejholy also, wondering why `TLS` is allowed but `NLA` with `NTLM` not (the former leaks plain text passwords to the server, so in any case worse than `NLA`, but that might be due to compliance with `FIPS` I assume?)

Yes, this is about FIPS compliance. NTLM uses MD4/RC4 that are not approved by FIPS, but TLS uses AES/SHA that are approved.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12814, opened 2026-05-27
test(packaging): add container distro builds

a5ehren:
Add a container-based package build test for deb, rpm, and Arch packaging across more distributions than current CI covers.

Notably, OpenSuSE 15/16/Tumbleweed + Arch + various Debians. The list is controlled in the new script.

Adjust the nightly RPM spec and Debian prep script mainly to make those package builds pass under the expanded distro matrix. Include C++ filesystem link detection and sanitizer linker-flag handling needed by older or distro-specific toolchains.


packaging/rpm/freerdp-nightly.spec
```diff
@@ -99,18 +115,22 @@ BuildRequires: dbus-glib-devel
...
+BuildRequires: (webkitgtk6.0-devel or webkit2gtk4.1-devel or webkit2gtk4.0-devel)
+%endif
+%endif
 
-%if 0%{?fedora} >= 36 || 0%{?rhel} >= 8
-BuildRequires: (ffmpeg-free-devel or ffmpeg-devel)
+%if 0%{?fedora} || 0%{?rhel} > 9 || 0%{?suse_version} >= 1600
+BuildRequires: pcsc-lite-devel
+BuildRequires: libcbor-devel >= 0.10.2
+BuildRequires: libfido2-devel >= 1.14.0
```
a5ehren, line 133: These two version bars were being enforced by cmake but not the specfile. this just moves the failure to an earlier stage.

packaging/rpm/freerdp-nightly.spec
```diff
@@ -168,16 +191,24 @@ cp %{_sourcedir}/source_version freerdp-nightly-%{version}/.source_version
...
     -DWITH_FFMPEG=ON \
     -DWITH_DSP_FFMPEG=ON \
+%else
+    -DWITH_FFMPEG=OFF \
 %endif
 %if 0%{?rhel} <= 8
     -DALLOW_IN_SOURCE_BUILD=ON \
 %endif
+%if 0%{?fedora} || 0%{?rhel} > 9 || 0%{?suse_version}
+    -DWITH_WEBVIEW=ON \
```
a5ehren, line 208: I know the webview is experimental, but enabling it for front-running repos makes sense with how often webkitGTK changes stuff

cmake/CompilerSanitizerOptions.cmake
```diff
@@ -28,22 +28,35 @@ endif()
...
     add_link_options(-fsanitize=memory)
   elseif(WITH_SANITIZE_THREAD)
+    set(FREERDP_SANITIZER_ENABLED ON)
     add_compile_options(-fsanitize=thread)
     add_link_options(-fsanitize=thread)
   endif()
 
+  # OpenSuSE 16+ clang adds this flag by default to any %cmake section in RPM specfiles,
+  # but it causes some ASAN builds to fail as they intentionally leave some symbols undefined
+  if(CMAKE_COMPILER_IS_CLANG AND FREERDP_SANITIZER_ENABLED)
```
akallabeth: not sure this is the right place to meddle with linker flags.
maybe better off in a toolchain file? (I assume the options are system defaults?)

cmake/CompilerSanitizerOptions.cmake
```diff
@@ -28,22 +28,35 @@ endif()
...
+    foreach(linker_flags CMAKE_SHARED_LINKER_FLAGS CMAKE_MODULE_LINKER_FLAGS)
+      string(REPLACE "-Wl,--no-undefined" "" ${linker_flags} "${${linker_flags}}")
+    endforeach()
+  endif()
+
+  # Also make sure this flag is left off if we are doing sanitizer builds
   option(WITH_NO_UNDEFINED "Add -Wl,--no-undefined" OFF)
-  if(WITH_NO_UNDEFINED)
+  if(WITH_NO_UNDEFINED AND NOT (CMAKE_COMPILER_IS_CLANG AND FREERDP_SANITIZER_ENABLED))
```
akallabeth: revert this, this is a functional change. If I enable the option I want the flag, no magic dropping it silently

packaging/scripts/prepare_deb_freerdp-nightly.sh
```diff
@@ -1,8 +1,21 @@
...
-ln -s packaging/deb/freerdp-nightly debian
-git rev-parse --short HEAD >.source_version
+if [ -L debian ]; then
+  rm debian
+elif [ -e debian ]; then
+  echo "debian already exists" >&2
+  exit 1
+fi
+
+cp -a packaging/deb/freerdp-nightly debian
```
akallabeth, line 15: why do you create a copy here?

packaging/scripts/test_distro_package_builds.sh
```diff
@@ -0,0 +1,1167 @@
+#!/usr/bin/env bash
+
+set -uo pipefail
+
+SCRIPT_PATH=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
```
akallabeth: this script is way too complicated.
maybe better split it up and use some podman/dockerfiles for the images?

cmake/CompilerSanitizerOptions.cmake
a5ehren: SuSE 16 puts this in their /usr/lib/rpm/macros.d/macros.cmake when you install rpm-build:
```
-DCMAKE_EXE_LINKER_FLAGS="%{?build_ldflags} -Wl,--as-needed -Wl,--no-undefined -Wl,-z,now" \\\
-DCMAKE_MODULE_LINKER_FLAGS="%{?build_ldflags} -Wl,--as-needed" \\\
-DCMAKE_SHARED_LINKER_FLAGS="%{?build_ldflags} -Wl,--as-needed -Wl,--no-undefined -Wl,-z,now" \\\
```
I can only guess they think that no one wants to do ASAN in a packaged build

packaging/scripts/prepare_deb_freerdp-nightly.sh
a5ehren: It's a guard against dpkg builders writing state back into the root folder. More of a concern for the local script than upstream or CI.

packaging/scripts/test_distro_package_builds.sh
a5ehren: Agreed. I wanted to see if you were interested before I put work into splitting it up. Started out simple when I was just doing the RPM but then it got out of control.

cmake/CompilerSanitizerOptions.cmake
akallabeth: @a5ehren well, you won't put it in a relase build, but we're building nightlies here right?

a5ehren:
accidentally nuked my branch, going to start over

PR #12815, opened 2026-05-27
[client,android] Add printer channel support for Android

### LINKED ISSUE #2882: Printer redirection on Android ?
Hi,

Would like to know if FreeRDP has support for printer redirection on Android. I went through various posts and here are my observations.

1) We need to build with -DWITH_CUPS=ON while generating make files using cmake.
2) We need to populate the printer name and printer driver to rdpSettings Structure. Went thorugh the command line usage.

Want to know if remote printing works on Android devices ? From the documentation looks like on linux/unix systems CUPS is used as the backend system for printing. 

Let me know if somebody has tried or any info on this.

Thanks,
Mahesh

svncibrahim:
## [client,android] Add printer channel support for Android
- Fixes  #2882 
### Description
End-to-end RDP printer redirection for aFreeRDP:

- **Backend:** New `printer_android` subsystem saves incoming print data as PDF to
  `/sdcard/Download/`. Uses `Microsoft Print to PDF:android` driver to request
  PDF output server-side.

- **Notification:** On job completion, a system notification appears with **Open**
  and **Print** actions. **Print** forwards the PDF to the Android print subsystem
  via a transparent proxy activity.

- **Setting:** Per-bookmark "Redirect Printer" toggle in Advanced Settings,
  backed by a Room DB migration (v15 -> v16).

- **WinPR ARM64 workaround:** `InterlockedPushEntrySList` silently fails on
  Android ARM64 when pointer tagging sets bit 63, causing the async thread to
  exit immediately. Async mode is disabled for the printer channel on Android as
  a temporary fix.  The proposed root fix (`#if !defined(__LP64__)` guard in
  `interlocked.c`) is tracked in #12816 .

### Testing Instructions
1. Enable **Redirect Printer** in bookmark Advanced Settings.
2. Connect and print a document from the remote session.
3. Verify notification appears with Open/Print actions.
4. Tap **Print** -> Android print dialog opens with the PDF.

### Environments Tested
- Physical Device: Android 16 (API 36)

### DIFF
--- a/channels/printer/client/android/printer_android.c
+++ b/channels/printer/client/android/printer_android.c
@@ -1,0 +1,282 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ * Print Virtual Channel - Android backend
+ *
+ * Copyright 2026 Ibrahim Sevinc <ibrahim.sevinc.mail@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+
+#include <winpr/crt.h>
+#include <winpr/assert.h>
+#include <freerdp/client/printer.h>
+#include <freerdp/channels/log.h>
+
+#define TAG CHANNELS_TAG("printer.client.android")
+
+#define PRINT_OUTPUT_DIR "/sdcard/Download/"
+#define PRINT_OUTPUT_PREFIX "rdp_print_"
+#define ANDROID_PRINTER_NAME "aFreeRDP Print"
+#define ANDROID_PRINTER_DRIVER "Microsoft Print to PDF"
+
+typedef struct
+{
+	rdpPrinterDriver driver;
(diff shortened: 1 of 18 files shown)


channels/printer/client/android/printer_android.c
```diff
@@ -0,0 +1,281 @@
...
+}
+
+static void printer_android_close_printjob(rdpPrintJob* printjob)
+{
+	rdpAndroidPrintJob* aj = (rdpAndroidPrintJob*)printjob;
+	rdpAndroidPrinter* ap = (rdpAndroidPrinter*)printjob->printer;
+
+	if (aj->file)
+	{
+		fclose(aj->file);
```
akallabeth, line 75: missing `aj->file = nullptr` set

channels/printer/client/printer_main.c
```diff
@@ -1012,9 +1013,15 @@ static UINT printer_register(PDEVICE_SERVICE_ENTRY_POINTS pEntryPoints, rdpPrint
 	printer_dev->rdpcontext = pEntryPoints->rdpcontext;
 	printer_dev->printer = printer;
 
+#if defined(ANDROID)
```
akallabeth: remove, will be fixed in the referenced pr anyway.

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
```diff
@@ -404,6 +404,9 @@ public static boolean setConnectionInfo(Context context, long inst, BookmarkBase
 		args.add("/clipboard");
 		args.add("/disp");
 
+		if (advanced.getRedirectPrinter())
+			args.add("/printer:aFreeRDP Print,Microsoft Print to PDF:android,default");
```
akallabeth: the `:android` part should not be required, it is the only backend available on android

akallabeth (changes requested):
Overall LGTM,
some small nitpicks left

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12821, opened 2026-05-28
[client,windows] honor /from-stdin in wfreerdp

scottgeigel2:
## Summary
  `wfreerdp.exe /from-stdin` has silently ignored piped credentials since 1e63d8c49 (Mar 2023). `echo pw | wfreerdp.exe /v:host /u:user /from-stdin` drops the password and pops the GUI
  login dialog instead. This restores the behavior originally added in 07ea60e96.

  Three stacked issues:

  1. `freerdp_passphrase_read` on Windows unconditionally invoked `CredUICmdLinePromptForCredentialsW`, which does not consume piped stdin. Now reads stdin via `fgets` when `from_stdin`
  is set, suppressing console echo via `SetConsoleMode` on an interactive tty to match the Unix `tcsetattr`-based behavior.

  2. `wf_authenticate_ex` had a `/from-stdin` branch but it was gated on `wfc->isConsole`, which `wf_has_console()` returns FALSE for piped stdin. Replaced with a delegation: when
  `/from-stdin` is set, `wf_pre_connect` overrides `instance->AuthenticateEx` with the common `client_cli_authenticate_ex`, eliminating duplicated stdin-reading logic.

  3. The `FreeConsole()` call and GUI cert/gateway callback installation lived in `wfreerdp_client_new`, which runs **before** `freerdp_client_settings_parse_command_line`. That meant
  stdin was already severed and GUI dialogs already wired up before `/from-stdin` was even visible. Moved to `wf_pre_connect`, which runs after command-line parsing.

  ## Test plan
  - [ ] Piped: `echo "mypassword" | wfreerdp.exe /v:host /u:user /from-stdin /cert:ignore` → connects without GUI prompt
  - [ ] Interactive console: `wfreerdp.exe /v:host /u:user /from-stdin /cert:ignore` from a real cmd/PowerShell → prompts on stdout, password input is not echoed, connects
  - [ ] No regression: `wfreerdp.exe /v:host /u:user` (GUI launch, no `/from-stdin`) → CredUI dialog appears as before
  - [ ] Gateway: `echo -e "user\ngwpw\npw" | wfreerdp.exe /v:host /u:user /from-stdin /g:gw /gu:gwuser` → all three credentials consumed in order

### DIFF
--- a/client/Windows/wf_client.c
+++ b/client/Windows/wf_client.c
@@ -52,6 +52,7 @@
 #include <freerdp/locale/locale.h>
 #include <freerdp/locale/keyboard.h>
 #include <freerdp/codec/region.h>
+#include <freerdp/client.h>
 #include <freerdp/client/cmdline.h>
 #include <freerdp/client/channels.h>
 #include <freerdp/channels/channels.h>
@@ -542,6 +543,12 @@ static BOOL wf_authenticate_ex(freerdp* instance, char** username, char** passwo
 	WINPR_ASSERT(username);
 	WINPR_ASSERT(domain);
 	WINPR_ASSERT(password);
+
+	/* /from-stdin: delegate to the common CLI handler instead of prompting via GUI.
+	 * Settings are fully parsed by the time auth callbacks fire, so this branch
+	 * reliably observes the user's choice. */
+	if (freerdp_settings_get_bool(instance->context->settings, FreeRDP_CredentialsFromStdin))
+		return client_cli_authenticate_ex(instance, username, password, domain, reason);
 
 	const WCHAR auth[] = L"Target credentials requested";
 	const WCHAR authPin[] = L"PIN requested";
--- a/libfreerdp/utils/passphrase.c
+++ b/libfreerdp/utils/passphrase.c
@@ -29,6 +29,7 @@
 #ifdef _WIN32
 
 #include <stdio.h>
+#include <string.h>
 #include <io.h>
 #include <conio.h>
 #include <wincred.h>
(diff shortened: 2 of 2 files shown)


akallabeth (changes requested):
please drop the unrelated changes.

scottgeigel2:
thanks for taking a look, I force-pushed over my old commit.

client/Windows/wf_client.c
```diff
@@ -237,6 +250,34 @@ static BOOL wf_pre_connect(freerdp* instance)
...
+	 * is set, route auth through the common CLI and keep the console attached
+	 * (the latter only matters if stdin wasn't a tty to begin with). */
+	if (freerdp_settings_get_bool(settings, FreeRDP_CredentialsFromStdin))
+	{
+		WLog_INFO(TAG, "/from-stdin set: routing auth through CLI");
+		wfc->isConsole = TRUE;
+		instance->AuthenticateEx = client_cli_authenticate_ex;
+	}
+
+#ifdef WITH_WIN_CONSOLE
```
akallabeth: why move this block here?

scottgeigel2: If I understand right, freerdp_client_context_new gets called before the CLI options get parsed. So I was trying to avoid using the Windows functions if from-stdin was being used.
I don't know if this would be better, but maybe I could try adding the is from stdin check in those windows specific handlers and just have them call the cli counterpart instead?

scottgeigel2: I pushed a 2nd commit to show you what I was talking about. Maybe this is better?

client/Windows/wf_client.c
```diff
@@ -982,6 +993,11 @@ static DWORD wf_verify_changed_certificate_ex(freerdp* instance, const char* hos
 	WCHAR* caption;
 	int what = IDCANCEL;
 
+	if (freerdp_settings_get_bool(instance->context->settings, FreeRDP_CredentialsFromStdin))
```
akallabeth: certificate options are not related to `/from-stdin`

client/Windows/wf_client.c
```diff
@@ -1030,6 +1046,10 @@ static DWORD wf_verify_changed_certificate_ex(freerdp* instance, const char* hos
 static BOOL wf_present_gateway_message(freerdp* instance, UINT32 type, BOOL isDisplayMandatory,
                                        BOOL isConsentMandatory, size_t length, const WCHAR* message)
 {
+	if (freerdp_settings_get_bool(instance->context->settings, FreeRDP_CredentialsFromStdin))
```
akallabeth: gateway message options are not related to `/from-stdin`

akallabeth (changes requested):
looking good now except for the certificate/gateway hooks.
what you're (re)implementing is `+force-console-callbacks`

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #12826, opened 2026-05-30
cmake: add server-minimal preset for lightweight static server builds

grioghar:
## Problem

Building a minimal static FreeRDP server library currently requires passing 25+ `-DWITH_*=OFF` flags on the cmake command line. There is no documentation of the correct combination, and discovering it requires trial and error — particularly because several flags interact non-obviously:

- `-DWITH_FFMPEG=OFF` alone is insufficient; `-DWITH_SWSCALE=OFF` and `-DWITH_DSP_FFMPEG=OFF` are independent probes
- `-DWITH_CLIENT=OFF` does not disable the proxy; `-DWITH_PROXY=OFF` is separate
- `-DWITH_CAIRO=OFF` is needed to suppress rdtk's cairo dependency even when `-DWITH_RDTK=OFF` is set

## Solution

Add a `CMakePresets.json` with two presets:

- **`server-minimal`** — `MinSizeRel` static build with only `rdpgfx`, `cliprdr`, and `rdpsnd` server channels; no client, proxy, shadow, media codecs, or platform display libraries
- **`server-minimal-debug`** — same but `Debug` with ASAN/UBSAN

## Usage

```bash
cmake --preset server-minimal \
  -DOPENSSL_ROOT_DIR=$(brew --prefix openssl)
cmake --build --preset server-minimal
```

## Testing

Verified on macOS 14 (arm64) and macOS 13 (x86_64) as part of the [macos-rdp-server](https://github.com/grioghar/macos-rdp-server) project CI, which builds FreeRDP 3.10.3 statically using exactly these flags.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,77 @@
+{
+  "version": 3,
+  "cmakeMinimumRequired": { "major": 3, "minor": 22, "patch": 0 },
+  "configurePresets": [
+    {
+      "name": "base",
+      "hidden": true,
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build/${presetName}",
+      "cacheVariables": {
+        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON"
+      }
+    },
+    {
+      "name": "server-minimal",
+      "displayName": "Minimal static server library",
+      "description": "Builds only the server-side libraries (libfreerdp, libwinpr, rdpgfx/cliprdr/rdpsnd channels) as static archives with no client, proxy, shadow, or media-codec dependencies. Suitable for embedding in a lightweight RDP server without requiring FFmpeg, Cairo, Opus, X11, or Wayland.",
+      "inherits": "base",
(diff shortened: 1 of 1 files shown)


akallabeth:
@grioghar oh, nice idea!
would just use a better name, `minimal-server` sounds more like a full blown server binary.

`minimal-deps-static-lib-server` or something similar so it is obvious what is actually build?

akallabeth:
@grioghar ok, taken your commit and modifed the presets with https://github.com/FreeRDP/FreeRDP/pull/12989
so far there is a `minimal-library`, `minimal-server`, `minimal-client` and `minimal-binaries` preset available.

I've removed default config options (`BUILD_SHARED_LIBS` et al) so far as that are default `CMake` options and the presets should provide a base configuration that is reusable.

Thinking about also adding a `full/all on` preset with the same 4 options to allow having a fullfeatured `lib/client/server/tool`

akallabeth:
@grioghar so thank you for your contribution!

PR #12827, opened 2026-05-30
perf(rdpgfx-server): pre-allocate send buffer to eliminate per-frame malloc

grioghar:
## Problem

`rdpgfx_server_packet_send` allocates a new `wStream` for the ZGFX compressed output on every call. For an H.264 server at 60fps this produces 60+ malloc+free pairs per second per connected client, each allocating several hundred kilobytes for the compressed frame data.

```c
/* current code — allocates on every packet */
fs = Stream_New(NULL, SrcSize + 7 + (SrcSize / ZGFX_SEGMENTED_MAXSIZE + 1) * 4);
```

## Change

Add `send_stream` (`wStream*`) to `RdpgfxServerPrivate`, initialised with 1 MiB capacity at context creation. `rdpgfx_server_packet_send` resets and reuses this buffer via `Stream_SetPosition` + `Stream_EnsureCapacity` instead of allocating.

- `Stream_EnsureCapacity` grows the buffer transparently if a frame exceeds 1 MiB
- `send_stream` is freed in `rdpgfx_server_context_free`
- No API changes; no behaviour changes

## Thread safety

Send functions are called from the application thread. The channel's own thread only reads incoming messages via `input_stream`. The new `send_stream` is accessed from a single thread and requires no additional synchronisation — same threading model as the existing `input_stream`.

## Expected impact

At 60fps, each frame triggers one ZGFX output buffer allocation sized to `SrcSize + overhead`. For a 1080p H.264 stream this is typically 100–500 KB per frame. Pre-allocating eliminates this steady-state allocation at the cost of a single 1 MiB allocation at session start.

Note: this change has been code-reviewed but not yet integrated into a running production server. I am submitting it for review and will update this PR once runtime validation is complete.


akallabeth (changes requested):
please remove the preset commit

PR #12834, opened 2026-06-02
Fix copying multiple items of the same type between xfreerdp sessions

FriederHannenheim:
# What is broken?
With two xfreerdp sessions running, copying text from one to the other always pastes the first copied text, ignoring any text that is copied after it. It also happens on other media types.

# Reproducing
- Open 2 xfreerdp sessions
- open text editors in both sessions
- in the first session, write 2 (non-identical) lines of text
- copy the first line of text and paste it into the second session
- copy the second line of text from the first session and paste it into the second session
- You will see that the first line of text is pasted again instead of the second line

## Why it doesn't work
In xf_cliprdr.c `xf_clipboard_changed` only compares the advertised selection formats. Since text is selected in both cases the selection formats are the same and no new formats are advertised to the server. 

https://github.com/FreeRDP/FreeRDP/blob/5ea7306dd4875cf1f52cef65fbd30936c626973a/client/X11/xf_cliprdr.c#L937

Sending `CLIPRDR_FORMAT_LIST` is the way the server is notified of clipboard changes on the client. If no new formats are advertised the server believes that the clipboard stayed the same on the client and won't send another `CLIPRDR_FORMAT_DATA_REQUEST` to receive the new clipboard contents.

### Why it does work with other clients
This doesn't happen when copying text in x11 clients other than xfreerdp since `xf_cliprdr_get_client_formats` returns an empty format list if the clipboard owner is not a xfreerdp window. 
https://github.com/FreeRDP/FreeRDP/blob/5ea7306dd4875cf1f52cef65fbd30936c626973a/client/X11/xf_cliprdr.c#L789
This empty list gets passed into `xf_cliprdr_send_format_list` which then calls `xf_clipboard_copy_formats` which copies it into `lastSentFormats`.
Then a `XConvertSelection` request is issued. The client owning the clipboard answers with `SelectionNotify` which contains the list of formats. This ends up in `xf_cliprdr_send_format_list` again where it is compared to the empty list. As they are not equal these formats get advertised to the server.

If the clipboard owner is a xfreerdp window the first call to `xf_cliprdr_send_format_list` is made with an already filled out list of formats so the `lastSentFormats` are not cleared and `xf_clipboard_changed` returns false on the next call, so no new formats are advertised.

# The fix
To fix this I reverted commit a51f4ccaaa8d4a2232725dc555827705ea8f0429. I believe it doesn't work in practice the way it is implemented right now. Since `xf_clipboard_changed` only compares the clipboard formats it does not notice a change when the new clipboard contents are of the same format as the old. 

Theoretically we could try fetching the clipboard contents in `xf_clipboard_changed` to see if they have changed. This would be one way of fixing the bug. 

However, I believe removing the duplicate ClientFormatList announcement filter entirely is the better way. The only case I see (please correct me if I'm wrong) in which it is useful is if the user copies the same content multiple times. In that case this code saves some packets that advertise the same formats again to the server.

---
This pull request was made as part of my work at openthinclient GmbH

### DIFF
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -157,7 +157,7 @@ static const char type_FileGroupDescriptorW[] = "FileGroupDescriptorW";
 static const char type_HtmlFormat[] = "HTML Format";
 
 static void xf_cliprdr_clear_cached_data(xfClipboard* clipboard);
-static UINT xf_cliprdr_send_client_format_list(xfClipboard* clipboard, BOOL force);
+static UINT xf_cliprdr_send_client_format_list(xfClipboard* clipboard);
 static void xf_cliprdr_set_selection_owner(xfContext* xfc, xfClipboard* clipboard, Time timestamp);
 
 static void requested_format_free(RequestedFormat** ppRequestedFormat)
@@ -288,7 +288,7 @@ static BOOL xf_cliprdr_update_owner(xfClipboard* clipboard)
 static void xf_cliprdr_check_owner(xfClipboard* clipboard)
 {
 	if (xf_cliprdr_update_owner(clipboard))
-		xf_cliprdr_send_client_format_list(clipboard, FALSE);
+		xf_cliprdr_send_client_format_list(clipboard);
 }
 
 static BOOL xf_cliprdr_is_self_owned(xfClipboard* clipboard)
@@ -836,7 +836,7 @@ static void xf_cliprdr_provide_server_format_list(xfClipboard* clipboard)
 }
 
 static UINT xf_cliprdr_send_format_list(xfClipboard* clipboard, const CLIPRDR_FORMAT* formats,
-                                        UINT32 numFormats, BOOL force)
+                                        UINT32 numFormats)
 {
 	union
 	{
@@ -879,7 +879,7 @@ static void xf_cliprdr_get_requested_targets(xfClipboard* clipboard)
 {
 	UINT32 numFormats = 0;
 	CLIPRDR_FORMAT* formats = xf_cliprdr_get_client_formats(clipboard, &numFormats);
-	xf_cliprdr_send_format_list(clipboard, formats, numFormats, FALSE);
+	xf_cliprdr_send_format_list(clipboard, formats, numFormats);
 	xf_cliprdr_free_formats(formats, numFormats);
 }
 
@@ -1255,7 +1255,7 @@ static BOOL xf_cliprdr_process_selection_notify(xfClipboard* clipboard,
 	{
 		if (xevent->property == None)
 		{
-			xf_cliprdr_send_client_format_list(clipboard, FALSE);
+			xf_cliprdr_send_client_format_list(clipboard);
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
sorry, breaks lots of use cases.
most prominently that huge datasets need to be copied multiple times if detection is removed.

please refine the check, not remove it

FriederHannenheim:
How would you go about refining the check? 

I guess one could cache the CLIPRDR_FORMAT_DATA_REQUEST data and compare that in xf_clipboard_changed. But with this implementation you'd cache huge datasets in ram and comparing them isn't cheap either.

FriederHannenheim:
I've now thought a bit more about the problem. I don't think copying huge datasets multiple times is a problem that happens a lot. For a dataset to be copied multiple times the following would need to happen:
- The user copies the dataset on the client
- The user pastes the dataset on the server (causing the dataset to be copied to the server)
- The user copies the same dataset on the client again, causing CLIPRDR_FORMAT_LIST to be sent again
- The user pastes the dataset on the server (causing another copy)

If the dataset is not copied again on the client the server caches the data and it won't be transferred again. I think it's fair to transfer the dataset again if it was freshly copied on the client.

Edit: The way the the check is implemented right now this also only happens if the user copies the dataset in another xfreerdp instance since the check does not work when the clipboard owner is not an instance of xfreerdp.

To improve this check by also comparing the contents of the clipboard could actually introduce more transfers of the dataset:
- xfreerdp sessions foo and bar
- user copies dataset in foo
- bar receives `XFixesSelectionNotifyEvent` and fetches the dataset (from foo which needs to transfer it from the server) to compare it to the last sent dataset
- user copies dataset again in foo 
- bar receives `XFixesSelectionNotifyEvent` again and fetches the dataset again to compare it to the last sent dataset

This all happens without even pasting anything into bar. With the check removed this just results in the `ClientFormatList` being announced multiple times which is not that big.

akallabeth:
@FriederHannenheim hmm, the original commit introducing this was a51f4ccaaa8 
and the reason was that due to network delays the duplicated announcement would provoke an update loop between remote and local (there is no good way to check for where the data originated)

FriederHannenheim:
> @FriederHannenheim hmm, the original commit introducing this was [a51f4cc](https://github.com/FreeRDP/FreeRDP/commit/a51f4ccaaa8d4a2232725dc555827705ea8f0429) and the reason was that due to network delays the duplicated announcement would provoke an update loop between remote and local (there is no good way to check for where the data originated)

Could you elaborate on the original issue? I don't think I understand it.

akallabeth:
@FriederHannenheim we had issues with reliably identifying clipboard changes (for example https://github.com/Keruspe/GPaste/issues/262 - that is not the only application, just an example)

FriederHannenheim:
But that seems like it was a bug in the clipboard manager no? I've tested this PR with Gpaste 45 and I did not see the history looping image described.

I don't think this is a thing that can be fixed in freerdp. The clipboard protocol is already lazy and does not copy the data unless it needs to. If there are issues with this they probably come from misbehaving applications.

akallabeth:
@freerdp-bot test

akallabeth:
@FriederHannenheim well, we can merge and see if there are regressions reported.

akallabeth (approved):

FriederHannenheim:
If there are any, feel free to ping me.

PR #12845, opened 2026-06-04
winpr/input: Fix numpad mapping for japanese keyboards

pnowack:
```
Currently, when pressing any numpad key on japanese keyboards with
gnome-remote-desktop, no number is entered regardless of the numlock
setting. The reason seems to be that those keys are hard coded to
VK_HOME, VK_UP, VK_LEFT, etc..
To fix this issue, use the dedicated virtual keycode values for the
numpad, like VK_NUMPAD7, VK_NUMPAD8 and such.
```


akallabeth (approved):
LGTM

PR #12848, opened 2026-06-06
Fix(wfreerdp): Refresh Windows frame after fullscreen restore

zorjen122:
### Problem
  Run wfreerdp with /f and /floatbar, /smart-sizing, and leave the full screen with the floatbar restore button, it will not refresh the window frame, so you can only switch through the taskbar to brush the frame.

```cmdline: .\wfreerdp.exe /v:[host]:[port] /p:[password] /u:[username] /f /floatbar /smart-sizing```


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #12849, opened 2026-06-06
Restore fullscreen when maximizing a toggled fullscreen window

zorjen122:
### Full screen experience
  When a Windows client session starts in fullscreen mode and is toggled back to
windowed mode, pressing the window maximize button should restore fullscreen
instead of only maximizing the decorated window (like mstsc).


akallabeth (approved):

akallabeth:
@freerdp-bot test

PR #12855, opened 2026-06-08
[channels,cliprdr] reset stream after use

### LINKED ISSUE #12853: cliprdr server: infinite PDU reprocessing loop / 100% CPU after 07e717d23
# **Describe the bug**

The cliprdr server channel enters an infinite loop after handling a single clipboard PDU. cliprdr_server_read() (channels/cliprdr/server/cliprdr_main.c) processes one PDU and sets cliprdr->totalExpectedBytes = 0, but never resets the receive stream cliprdr->s.

This is a regression from commit 07e717d23 ([channels,cliprdr] refactor server channel), which removed the Stream_ResetPosition(s) that previously ran after cliprdr_server_receive_pdu() and added a Stream_SealLength(s) before it, without restoring the buffer afterwards.

# **To Reproduce**

1. Run a FreeRDP-based RDP server that enables the cliprdr server channel (CliprdrServerContext).
2. Connect any RDP client and let the clipboard channel complete its capability/format exchange.
3. Observe a server thread climbing to ~100% CPU and steadily growing memory.

# **Expected behavior**

After each PDU the receive buffer is reset so the next read starts from a clean state, and the read thread exits when the channel is closed.

# **Application details**

- FreeRDP version:  (master; version 3.26.1)
- Affected component: channels/cliprdr/server/cliprdr_main.c, cliprdr_server_read().

OS: Linux

# **Additional context**

Suggested fix in cliprdr_server_read(): after cliprdr_server_receive_pdu(), restore the buffer with Stream_SetPosition(s, 0) + Stream_SetLength(s, Stream_Capacity(s)), and treat a 0-byte read on a signaled channel as a close (return an error so the loop exits).

akallabeth:


maordadush:
@akallabeth Thanks for the quick fix — this does stop the reprocessing loop / CPU spin. 
One thing to flag, though: Stream_ResetPosition(s) alone looks insufficient. The same refactor (07e717d23) also added Stream_SealLength(s) just before cliprdr_server_receive_pdu(), which sets the stream length to the size of the PDU just processed. Stream_ResetPosition() only rewinds the position (pointer = buffer) — it doesn't restore the length. So on the next read, Stream_SafeSeek() checks against the shrunk length (Stream_GetRemainingLength) and fails as soon as a following PDU needs more bytes than the previous one.

Applying the same one-line change on top of the current master, the server read aborts shortly after with:

> [cliprdr_server_read] invalid length, got 16, require at least 24 [element size=1]
> ...
> CheckEventHandle failed with error 1359   (ERROR_INTERNAL_ERROR)

which makes cliprdr_server_thread exit and the clipboard channel stop working.

Restoring the length together with the position fixes it:

```
const UINT error = cliprdr_server_receive_pdu(context, s, &cliprdr->header);
Stream_SetPosition(s, 0);
Stream_SetLength(s, Stream_Capacity(s));
cliprdr->totalExpectedBytes = 0;
```

parasol-aser:
**`channels/cliprdr/server/cliprdr_main.c:cliprdr_server_read` leaves the reusable stream at length 0**

After the first CLIPRDR PDU is dispatched, the new `Stream_SetLength(s, 0)` leaves the persistent input stream with readable length 0. The next non-empty client PDU is still written into the buffer, but `Stream_SafeSeek(s, BytesReturned)` checks readable length rather than capacity, so a normal multi-PDU clipboard exchange can fail the channel before the second PDU is parsed.

Fix: restore the stream length after `WTSVirtualChannelRead` and before `Stream_SafeSeek`, for example to the previous stream position plus `BytesReturned`, and cover it with a two-PDU server read test.

maordadush:
@akallabeth with `Stream_SetLength(s, 0)` the next non-empty PDU fails immediately at the header seek (invalid length, got 0, require at least 8 → CheckEventHandle failed 1359). Restoring the length fixes it, either` length = position + BytesReturned` before the seek, or resetting to `Stream_Capacity(s)` after dispatch

akallabeth:
@maordadush oh, wrong branch pushed, now it should be ok.

PR #12856, opened 2026-06-09
[core,gateway] validate auth blob length in rdg_process_extauth_sspi

metsw24-max:
**Missing length check in rdg_process_extauth_sspi**

A malicious or intermediary RDG gateway can send a PKT_TYPE_EXTENDED_AUTH_MSG whose declared `authBlobLen` is larger than the data actually received, and the handler then `malloc`s that size and `Stream_Read`s it straight from the packet buffer with no validation, reading past the end of the received data. The sibling response handlers all guard their reads with `Stream_CheckAndLogRequiredLength` first, so I have added the same checks here before the header fields and before the blob copy.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM, good catch!

PR #12858, opened 2026-06-09
[core,event] add StateChanged event

akallabeth:
* Define new StateChanged event type
* Trigger new event whenever the connection state changes


mfleisz (commented):
LGTM except the unnecessary scope in `rdp_set_state`

PR #12862, opened 2026-06-09
[core,nego] bound cookie tag check to remaining length

metsw24-max:
**Out-of-bounds read in nego_read_request_token_or_cookie**

When a server parses the optional routing token or cookie of an X.224 Connection Request, the only guard guarantees 15 remaining bytes, but the first `memcmp` against `"Cookie: mstshash="` reads 17, so a request whose token region is 15 or 16 bytes long reads one or two bytes past the buffer. Gated that compare on the remaining length; behaviour is unchanged since the cookie path already requires at least 19 bytes.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12868, opened 2026-06-10
[client,x11] release normal keys before modifiers

### LINKED ISSUE #12789: xfreerdp switch local window keybaord focus issue
# **To Reproduce**

1.  under local Gnome, run `xfreerdp` connecting to remote Windows
2.  in remote , open a `notepad` in the front
3.  click on `xfreerdp`'s title bar ( local side ) , otherwise step 4's alt +tab will be triggered remotely.
4. use alt + tab to trigger local side window switch
5. then  go back `xfreerdp` (either use  alt + tab again or click on Gnome top left corner )
6. you could observe that `notepad`'s menubar -> File is triggered ( highlighted as blue)

# **Expected behavior**
 `notepad`'s menubar -> File is not triggered 

# **Application details**

* FreeRDP version (`xfreerdp /version`) 3.26.0
* Command line used : `xfreerdp /v:xxx /u:xxxx /p:xxxx`
* Output of `xfreerdp /buildconfig`
* OS version connecting to (server side) Windows 10

 - OS: [e.g. Linux/Windows/Android/..] Linux
 - Version/Distribution: [e.g. Debian 10, Windows 2008, Android 10] Fedora Workstation 44 (Gnome)
 - Architecture: [amd64, arm]: amd64

# **Additional context**
I think the problem is that `alt` key down event (and maybe release event) is sent to remote , but `tab` key down event is not. so `notepad` only get `alt` pressed and use it to trigger menu.

I'm not familiar with  `focus` , so when step 3 , clicking on xfreerdp title bar (local side) , should  xfreerdp be considered as focused or not ? Under this situation, Should local window manager send KeyPress event of `alt` to xfreerdp and should xfreerdp pass it to remote ?

-----

Also I've found some interesting code here which handle win-key issue.

https://github.com/FreeRDP/FreeRDP/blob/7445d013dcadf7a6ef2b96d7f006efaeb798adff/client/X11/xf_keyboard.c#L949-L953,   i tested adding RDP_SCANCODE_LMENU and RDP_SCANCODE_RMENU here could solve the problem, i guess we should add all modifier key here?

and furthermore, you don't need send  `RDP_SCANCODE_TAB` release event  here , you could send any key except current modifier key.

and furthermore again, you don't need to send  https://github.com/FreeRDP/FreeRDP/blob/7445d013dcadf7a6ef2b96d7f006efaeb798adff/client/X11/xf_keyboard.c#L955 this release event if current key is modifier key.

I guess , this is a implementation detail (or quirks) of Window's RDP server side , like if you send a key release event which keycode is not previous down , then the previous down key is canceled? 

I'm not sure what other rdp server (or desktop env) will behave if we apply this workaround on all modifier key.

akallabeth:


jackyzy823:
Hi, thanks for the fix. 

Just a question: Should `RDP_SCANCODE_LSHIFT` and `RDP_SCANCODE_RSHIFT` be treated as modifier keys too?

akallabeth:
@jackyzy823 don't think so, shift just changes the keycode sent, while modifiers are different keys (that can do something in combination with others)

PR #12877, opened 2026-06-12
Ntlm hostname

### LINKED ISSUE #12840: Allow per-session NTLM workstation name override via rdpSettings
# Is your feature request related to a problem? Please describe.

When using FreeRDP in a proxy, gateway, or containerized environment, the NTLM workstation name sent in the `NTLMSSP_AUTH` message is always derived from `gethostname()` via `GetComputerNameExA(ComputerNameNetBIOS)` in `winpr/libwinpr/sspi/NTLM/ntlm.c`. There is no way to override this per-session through `rdpSettings` or any API.

The existing `/client-hostname:` option only affects the RDP Client Info PDU (`ClientHostname` in GCC Conference Create), but has no effect on the NTLM workstation field. This was previously reported in #5007 and confirmed by maintainers but never addressed.

This is a problem for:
- **RDP proxies and gateways** that broker connections on behalf of other machines -- the target sees the proxy's hostname instead of the actual originating machine's identity.
- **Containerized deployments** (Docker, Kubernetes, etc.) where `gethostname()` returns an ephemeral container ID that is meaningless to the target and may not match any known host.
- **Endpoint security products** on target machines that validate the NTLM workstation name against known hosts and block connections from unrecognized hostnames.

# Describe the solution you'd like

A per-session setting (e.g., `FreeRDP_NtlmWorkstation` or reusing `FreeRDP_ComputerName`) that, when set, is passed through the NLA/CredSSP layer into the WinPR NTLM SSPI provider and used by `ntlm_SetContextWorkstation()` instead of falling back to `gethostname()`. The existing behavior should be preserved when no override is set.

# Describe alternatives you've considered

- **WinPR file-based registry (`HKLM.reg`)**: The `WorkstationName` key under `Software\<vendor>\WinPR\NTLM` can override the workstation globally. However, this is process-wide (the `Reg` instance is a singleton), so it cannot support per-session overrides in multi-session applications like proxies or FreeRDP's built-in server/proxy.
- **Setting the OS hostname** (`sethostname()`) before connecting: This is process-global, requires elevated privileges, and breaks concurrent sessions.
- **Setting `ClientHostname` / `ClientAddress`**: These only affect the RDP Client Info PDU, not the NTLM authentication layer.

# Additional context

Relevant code in `winpr/libwinpr/sspi/NTLM/ntlm.c`:
- `ntlm_SetContextWorkstation()` (~line 131): Falls back to `GetComputerNameExA(ComputerNameNetBIOS)` when no workstation is provided.
- `ntlm_ContextNew()` (~line 244): Reads `WorkstationName` from the WinPR registry, but only as a process-global default.
- `ntlm_InitializeSecurityContextW()` (~line 671): If workstation is still unset, calls `ntlm_SetContextWorkstation(context, nullptr)` which triggers `gethostname()`.

The NTLM SSPI layer currently has no access to `rdpSettings`, so the override needs to be threaded through the CredSSP/NLA bridge (e.g., via `SEC_WINNT_AUTH_IDENTITY` extension or a new SSPI attribute).

Related issue: #5007

akallabeth:


omatasas:
@akallabeth  Thank you for adding this! 🙏

PR #12879, opened 2026-06-12
Expose the correlationId in settings

hardening:
Before this patch a random correlationId was generated stored locally for RD gateway connections, but sometime you may want to reuse it for some other component. So these changes make it stored in settings.


akallabeth (commented):
~~Might be better to store the string representation in the settings, for a new datatype all getter/setter/defaults/... must be adjusted (that is why the unit tests fail)~~

can not reproduce with local builds, checking...

[edit] ok, if you use a custom type you need to manually add the whole shebang for settings copy, cleanup, ...
so the initial suggestion with using a guid string might still be the best approach.

PR #12886, opened 2026-06-14
[build,android] fix incremental build for ExternalProject dependencies

svncibrahim:
## Description

Before this fix, every Java code change triggered a full rebuild of all native
dependencies and **took more than 10 minutes**, even though nothing in the C/C++
sources had changed. This made iterating on the Android UI very slow.

The Android client builds its third-party deps (OpenSSL, FFmpeg, OpenH264,
libpng, libjpeg-turbo, WebP, Opus, cJSON, uriparser) via `ExternalProject_Add`,
which re-ran its update/download step on every incremental build.

Adding `UPDATE_COMMAND ""` disables that step so each dependency builds once and
is skipped afterwards. Also:

- **FreeRDP**: `BUILD_ALWAYS ON` so local source changes are always picked up.
- **OpenH264**: dropped the explicit `BINARY_DIR` override for consistency.

Build-system only, no runtime behavior changes.

## Testing Instructions

1. Clean build, all deps build and the app links.
2. Change a Java or FreeRDP source, incremental build is fast and deps are not
   re-downloaded or rebuilt.

## Environments Tested

- Android NDK build (Linux host)


akallabeth:
@svncibrahim hmm, does not look clean.
running the externalproject build manually reveals that the only places that take time are `openh264` (running clean on configure) and `ffmpeg`

svncibrahim:
> @svncibrahim hmm, does not look clean.
> running the externalproject build manually reveals that the only places that take time are `openh264` (running clean on configure) and `ffmpeg`

Let me check again, I tested only on Android Studio.

akallabeth:
@svncibrahim you might want to change `DOWNLOAD_EXTRACT_TIMESTAMP ON` as that prevents timestamp issues on more modern `CMake` versions (e.g. redownload of existing sources due to timestamp mismatches)

svncibrahim:
@akallabeth Found the root cause and pushed a cleaner version.

Only the GIT-based projects actually needed the fix. 

Their update step re-runs on every build. For openh264 that cascades: the update re-runs **PATCH_COMMAND**, which re-runs **CONFIGURE_COMMAND** (**make clean**), which means a full rebuild. And since ffmpeg has "**DEPENDS openh264 opus**", it rebuilds too. 

The URL-based projects have no real update step, so UPDATE_COMMAND "" did nothing there and I dropped it.

DOWNLOAD_EXTRACT_TIMESTAMP only applies to URL archive extraction, so it is a no-op on the GIT projects. I removed it there and kept it ON for the URL ones, as you suggested.

As a follow-up it might be nice to move all the deps to a single type, either all git or all url based, so the whole setup stays consistent.

svncibrahim:
This second commit moves the remaining URL deps (cjson, ffmpeg, openssl, uriparser) to Git, eliminating the need to maintain SHA256 hashes. It also standardizes version variables and argument ordering. 

I kept it separate so we can easily revert if you prefer tarballs.

client/Android/cmake/ExternalFreeRDP.cmake
```diff
@@ -40,7 +40,7 @@ set(FREERDP_EXTRA_CMAKE_ARGS
 
 ExternalProject_Add(
   freerdp
-  DOWNLOAD_EXTRACT_TIMESTAMP OFF
+  BUILD_ALWAYS ON
```
akallabeth, line 43: this is wrong, it forces a rebuild on every build.

client/Android/cmake/ExternalFFmpeg.cmake
```diff
@@ -53,11 +53,12 @@ endif()
...
   INSTALL_DIR ${DEPS_INSTALL_DIR}
-  DOWNLOAD_EXTRACT_TIMESTAMP OFF
   SOURCE_DIR ${CMAKE_SOURCE_DIR}/external/ffmpeg
   BINARY_DIR ${CMAKE_BINARY_DIR}/external/ffmpeg
-  URL https://github.com/FFmpeg/FFmpeg/archive/refs/tags/${FFMPEG_VERSION}.tar.gz
-  URL_HASH SHA256=dd308201bb1239a1b73185f80c6b4121f4efdfa424a009ce544fd00bf736bb2e
+  GIT_REPOSITORY https://github.com/FFmpeg/FFmpeg.git
+  GIT_TAG ${FFMPEG_VERSION}
+  GIT_SHALLOW TRUE
+  UPDATE_COMMAND ""
```
akallabeth: why override the update command? see problem below.

client/Android/cmake/ExternalOpenH264.cmake
```diff
@@ -17,12 +17,11 @@ endif()
 
 ExternalProject_Add(
   openh264
-  DOWNLOAD_EXTRACT_TIMESTAMP OFF
   SOURCE_DIR ${CMAKE_SOURCE_DIR}/external/openh264
-  BINARY_DIR ${CMAKE_BINARY_DIR}/external/openh264
   GIT_REPOSITORY https://github.com/cisco/openh264.git
   GIT_TAG ${OPENH264_VERSION}
   GIT_SHALLOW TRUE
+  UPDATE_COMMAND ""
```
akallabeth: so basically here we have the only change required?
might be a better idea to just check if the patch was already applied and skip in the `PATCH_COMMAND`?

akallabeth: https://stackoverflow.com/questions/48167884/check-if-git-apply-was-already-applied

akallabeth: hmm, a simple `git am --3way patch.txt` should suffice actually, can you give that a try?

svncibrahim: Sure, I'll give it a try and let you know the result.

akallabeth:
@svncibrahim ok, #12890 should fix it.

1. Uses `git am --3way` to avoid changing file timestamps in source
2. copies sources (only if changed) to binary directory and builds there, avoiding problems with the multi arch builds.

svncibrahim:
@akallabeth  Thanks for the review. 

I measured the actual rebuild behavior on `arm64-v8a` to back the trade-offs with numbers.

- **openh264 patch:** now `git am --3way <patch>` as you suggested.
- **`UPDATE_COMMAND ""`:** replaced with the idiomatic `UPDATE_DISCONNECTED ON`.

`UPDATE_DISCONNECTED` on the 3 custom-build deps (ffmpeg, openssl, openh264) **is mandatory**: without it they fully rebuild every build (~10 min). 

On top of that I tested three setups:

| # | `UPDATE_DISCONNECTED` | `BUILD_ALWAYS` | no-change rebuild | edit to a FreeRDP source |
|---|---|---|---|---|
| 1 | 3 custom deps only | off | ~35 s | caught (dep cascade) |
| 2 | all 9 deps | off | 0.0 s | silently missed |
| 3 | all 9 deps | on | 0.1 s | caught (~1 s) |

Option 1's 35 s is mostly an `ffmpeg ./configure` re-run (18.8 s), triggered every build by the still-connected `opus`. 

Disconnecting all 9 removes that, but ExternalProject does not track `SOURCE_DIR` edits, so without `BUILD_ALWAYS` a real source edit is silently missed (verified: 0.0 s no-op, object not recompiled, stale lib installed).

Detection actually already works today, but only as an accidental side-effect of that cascade, paid for with the 18.8 s reconfigure. 

`BUILD_ALWAYS ON` has no effect on clean/CI builds (everything is built once regardless). It only affects incremental dev builds: it fixes that silent stale-lib bug at ~0.1 s cost.

Option 3 seems like the best balance to me, but let me know what you think.

akallabeth:
@svncibrahim well:
* you disabled source change tracking with `UPDATE_DISCONNECTED`
* The `FFMPEG` configure step will always run, but not the build
* with `OpenH264` if `UPDATE_COMMAND ""` (or `UPDATE_DISCONNECTED`) no changes will be triggered

akallabeth:
@svncibrahim easiest way would be to just disable `FFMPEG` for debug builds, that way the configure step will not run if you are developing.

svncibrahim:
@akallabeth I will try disabling FFMPEG for debug. Thanks! 

The goal behind the change: pinned deps are never edited, so they should build once and stay inert, while only FreeRDP itself is tracked. Disconnecting all 9 deps freezes them, so a no-change rebuild does nothing and the FFMPEG configure step never runs. BUILD_ALWAYS then tracks FreeRDP's own sources, which ExternalProject does not watch otherwise. 

I'll test the FFMPEG-off route too and report back.

akallabeth:
@svncibrahim I don't like this meddling with the build system.
core issue is that we got a problem with integrating the build systems of `openh264` and `ffmpeg` properly (well mostly `ffmpeg` as it triggers a full rebuild when `openh264` reinstalls the files even without a change)

closing this, will merge #12890 which disables most native deps for debug builds

svncibrahim:
Sounds good, thanks. #12890 works for me, I'll rebase on top of it.

PR #12887, opened 2026-06-14
[client,android] Add RAIL/RemoteApp window support

### LINKED ISSUE #11655: is there any plan to support remoteapp in android
rail/remoteapp mode is missing in android client

open remoteapp in android as xfreerdp does.

N/A

**Additional context**
N/A

Hi, is there any plan to support remoteapp in android? since there is no implement for rail mode in android client, i notice rail mode is only supported in xfreerdp client and it seems not stable yet.

so do you have any plan to support rail channel in android client?

svncibrahim:
- Fixes #11655

## Description

RAIL (RemoteApp) support for aFreeRDP: remote application windows are shown as floating overlays instead of the full remote desktop.

- Window create/move/hide/destroy via the RAIL channel; contents from the GFX `UpdateWindowFromSurface` hook.
- Server-authoritative z-order and active window (single `OnRailMonitoredDesktop` callback), so the topmost window is the one clicks   hit.
- Remote desktop surface hidden in RemoteApp mode (no logon/wallpaper behind the windows).
- Overlays track `SessionView` zoom; input maps back via its inverse scale  matrix.
- Per-window bitmaps reused (realloc only on size change); overlay logic  isolated in `RailWindowManager`.

Non-RemoteApp sessions are unaffected.

### Supported

- Multiple windows: every top-level window the server sends is rendered as its
  own overlay, including windows opened from the launched app (e.g. starting cmd
  and launching another program from it). Z-order and the active window follow
  the server.
- Move and resize: windows can be moved and resized, and the overlays update
  accordingly.
- Menus, tooltips and popups render with their transparency (rounded corners /
  drop shadows).
- Pinch-zoom: overlays scale with the session and clicks still map to the
  correct window.
- The remote desktop background is hidden in RemoteApp mode (only the app
  windows are shown).

## Screenshot

Screenshot_2026-06-14-18-11-28-90_bfe5c884e366e53e8d644a47ee8284f1

Screenshot_2026-06-14-18-45-06-96_bfe5c884e366e53e8d644a47ee8284f1

## Testing Instructions

- In a bookmark's advanced settings, set the RemoteApp program name, then connect > the app window shows with no remote desktop background.
- Open overlapping windows (e.g. Notepad + Open dialog) > z-order correct, clicking brings the right window forward.
- Pinch-zoom > windows scale and stay aligned, clicks still land correctly.
- Normal (no RemoteApp name set) session > no regression.

## Environments Tested

- Physical Device: Android 16 (API 36)

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/cpp/android_freerdp.c
+++ b/client/Android/Studio/freeRDPCore/src/main/cpp/android_freerdp.c
@@ -49,6 +49,7 @@
 #include "android_jni_utils.h"
 #include "android_cliprdr.h"
 #include "android_disp.h"
+#include "android_rail.h"
 #include "android_freerdp_jni.h"
 
 #if defined(WITH_GPROF)
@@ -62,6 +63,10 @@
 
 static jclass gJavaActivityClass;
 static jmethodID gOnPointerSetMethod;
+static jmethodID gOnRailWindowUpdateMethod;
+
+static UINT android_UpdateWindowFromSurface(RdpgfxClientContext* context, gdiGfxSurface* surface);
+
 static void android_OnChannelConnectedEventHandler(void* context,
                                                    const ChannelConnectedEventArgs* e)
 {
@@ -84,6 +89,17 @@ static void android_OnChannelConnectedEventHandler(void* context,
 	else if (strcmp(e->name, DISP_DVC_CHANNEL_NAME) == 0)
 	{
 		android_disp_init(afc, (DispClientContext*)e->pInterface);
+	}
+	else if (strcmp(e->name, RAIL_SVC_CHANNEL_NAME) == 0)
+	{
+		android_rail_init(afc, (RailClientContext*)e->pInterface);
+	}
+	else if (strcmp(e->name, RDPGFX_DVC_CHANNEL_NAME) == 0)
+	{
+		freerdp_client_OnChannelConnectedEventHandler(context, e);
+		RdpgfxClientContext* gfx = (RdpgfxClientContext*)e->pInterface;
+		if (gfx)
+			gfx->UpdateWindowFromSurface = android_UpdateWindowFromSurface;
 	}
 	else
 		freerdp_client_OnChannelConnectedEventHandler(context, e);
@@ -111,6 +127,10 @@ static void android_OnChannelDisconnectedEventHandler(void* context,
 	else if (strcmp(e->name, DISP_DVC_CHANNEL_NAME) == 0)
 	{
 		android_disp_uninit(afc, (DispClientContext*)e->pInterface);
+	}
+	else if (strcmp(e->name, RAIL_SVC_CHANNEL_NAME) == 0)
+	{
+		android_rail_uninit(afc, (RailClientContext*)e->pInterface);
 	}
 	else
 		freerdp_client_OnChannelDisconnectedEventHandler(context, e);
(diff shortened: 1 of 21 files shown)


client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
```diff
@@ -382,10 +382,11 @@ public static boolean setConnectionInfo(Context context, long inst, BookmarkBase
 
 		if (!advanced.getRemoteProgram().isEmpty())
 		{
-			args.add("/shell:" + advanced.getRemoteProgram());
+			args.add("/app:program:" + advanced.getRemoteProgram());
```
akallabeth, line 388: hmm, alternate shell is a different beast from rails, why remove?

akallabeth (commented):
Wow, nice work!

but do I see that correct that this is (currently) still quite limited?
e.g. no window resize, no handling for multiple remote apps (you can start new ones from the one you launch, e.g. if you start `cmd`, ... ?

might be a good idea to hide this behind a `CMake` option off by default in that case.

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/services/LibFreeRDP.java
svncibrahim: You're right, I missed that, it was just reusing the existing "remote program" text field directly.

svncibrahim:
> Wow, nice work!
> 
> but do I see that correct that this is (currently) still quite limited? e.g. no window resize, no handling for multiple remote apps (you can start new ones from the one you launch, e.g. if you start `cmd`, ... ?
> 
> might be a good idea to hide this behind a `CMake` option off by default in that case.

Thanks! Both actually work fine

I tested it with multiple windows (e.g. launching cmd and opening another app from it) and resizing has no problem either.

 I forgot to mention this in the description, adding it now.

Screenshot_2026-06-14-18-45-06-96_bfe5c884e366e53e8d644a47ee8284f1

akallabeth:
@svncibrahim ok, nice! did just guess, did not yet have time to test.
just surprised that it works with such a small `android_rail_window_state` implementation, that is usually the place where the details go wrong (with `xfreerdp` that is)

svncibrahim:
Yeah, I was expecting more issues there as well. I've only done a very quick test so far, so I'm sure there are still bugs hiding in there. Feel free to test it if you get a chance.

akallabeth:
@svncibrahim I'd still hide it behind a debug flag or similar for the time being, until we know it really works. (a lot of issues / discussions for `xfreerdp` resolve around bugs we have there, so I'd want to make sure everyone knows this is experimental for the time being.

svncibrahim:
@akallabeth  sorry I miss that part.

Agreed it should be opt-in and experimental. I'd rather do it as a runtime flag than compile-time: an "Experimental" section in settings, off by default and labeled as such.

It's safe at runtime since RAIL only kicks in when /app: is set. Toggle off means we never send it, so nothing changes and the RAIL code never runs.

I've also got an experimental camera feature coming, so they could share the same section. Runtime toggle ok with you, or would you prefer a compile-time CMake option?

svncibrahim:
A runtime flag also means people can actually test it without a custom build, so we'd get more feedback.

akallabeth:
@svncibrahim just be careful, if `RDP` file support is there then it can be activated that way too, so best cover that path as well.

svncibrahim:
> @svncibrahim just be careful, if `RDP` file support is there then it can be activated that way too, so best cover that path as well.

Done. RemoteApp is now behind an experimental toggle (off by default) in a new Experimental section.

The gate is in setConnectionInfo, so all paths are covered: bookmark UI, .rdp import, and the freerdp:// URI passthrough.

akallabeth (commented):
I'd add the check in `PostConnect` only.
There we know the connection settings parsed from command line (wherever from) and can abort the session and display a dialog informing the user that that is experimental/needs to be enabled/...

client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/data/AppDatabase.java
```diff
@@ -171,6 +172,13 @@ private static String toHex(byte[] bytes)
 		return sb.toString();
 	}
 
+	private static final Migration MIGRATION_16_17 = new Migration(16, 17) {
+		@Override public void migrate(@NonNull SupportSQLiteDatabase db)
+		{
+			db.execSQL("ALTER TABLE 'bookmarks' ADD 'alternate_shell' TEXT NOT NULL DEFAULT '';");
```
akallabeth, line 178: should that not be filled with the values from `remoteProgram` before this change?

svncibrahim: Right, I'll backfill it.

svncibrahim:
> I'd add the check in `PostConnect` only. There we know the connection settings parsed from command line (wherever from) and can abort the session and display a dialog informing the user that that is experimental/needs to be enabled/...

I'll move it to PostConnect and show a dialog instead of silently falling back. Will update.

client/Android/Studio/freeRDPCore/src/main/cpp/android_freerdp.c
```diff
@@ -332,6 +352,59 @@ static BOOL android_Pointer_SetDefault(rdpContext* context)
...
+	jint* dst = (*env)->GetIntArrayElements(env, pixels, NULL);
+	if (dst)
+	{
+		/* Copy each row, forcing coloured pixels opaque (ARGB_8888 would otherwise blend the
+		 * active window's frame away) while keeping transparent black so menu corners/shadows
+		 * stay see-through. */
+		for (UINT32 row = 0; row < height; row++)
+		{
+			const UINT32* src = (const UINT32*)(surface->data + (size_t)row * surface->scanline);
+			jint* line = dst + (size_t)row * width;
```
akallabeth: `freerdp_image_copy`?

svncibrahim: Done, switched the manual row copy to freerdp_image_copy.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM,
lets merge and see if there are issues down the road.

svncibrahim:
Thanks!
I'll watch for any issues and follow up.

PR #12890, opened 2026-06-15
[client,android] use git am --3way

akallabeth:
The patch command previously always changed the timestamp triggering an expensive rebuild of native components. git am --3way patch.txt should avoid that.


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #12900, opened 2026-06-16
[tools] make the generator create more clang-format friendly code

hardening:
A few updates on the python script that generates files from the settings so that the raw output generates less diffs with clang-format.

### DIFF
--- a/tools/update-settings-tests
+++ b/tools/update-settings-tests
@@ -2,6 +2,29 @@
 import os
 import sys
 
+MAX_PER_LINE = 100
+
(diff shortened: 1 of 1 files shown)


akallabeth (commented):
* you mean `clang-format`, right?
* there are some additional newlines added that should not be there

otherwise LGTM

hardening:
Yeah that makes the changes smaller but it remains some (I've not managed to removed them yet)

akallabeth:
@hardening ok. if you could rename the commit then we're good to merge.

akallabeth (approved):
LGTM, nice change!

PR #12901, opened 2026-06-16
[client, x11] restore previous quirks to modifier keys release

jackyzy823:
still for #12789
Not sure if RDP_SCANCODE_UNKNOWN is proper. (and if not proper, i would change it back to RDP_SCANCODE_TAB)


akallabeth (changes requested):
We already release all non modifier keys pressed before the modifiers.
this looks like something else is going on at your end.

akallabeth:
@jackyzy823 you might want to try to remove `xf_sync_kbd_state` from `xf_keyboard_release_all_keypress` instead.

akallabeth:
#12927

PR #12910, opened 2026-06-17
[winpr,image] bound unaligned bitmap read to declared image size

metsw24-max:
The BMP reader `winpr_image_bitmap_read_buffer` checks that `biSizeImage` bytes are available and then copies the pixels one row at a time, reading `scanline` bytes and seeking `image->scanline - scanline` of presumed row padding. When the rows are not a multiple of four bytes wide the decoder takes its unaligned workaround and sets `scanline` to the packed `width * bytesPerPixel`, but it still seeks the alignment padding that a packed buffer never has, so the source pointer advances `image->scanline * height` bytes overall while only `biSizeImage` were validated. If the input is sized exactly to its header plus `biSizeImage`, the trailing rows read past the end of the buffer. The bytes are attacker controlled in practice because the clipboard image synthesisers feed peer supplied DIB data straight into `winpr_image_read_buffer` (`clipboard_synthesize_image_bmp_to_format`). I noticed it because a crafted 24bpp bitmap trips the `Stream_Read` capacity assertion in a debug build. The fix advances the destination by `image->scanline` and drops the source seek so each row only consumes its real `scanline` bytes; the aligned path is untouched since that seek was already zero, and the unaligned rows now also land at the correct stride. A regression test for an unaligned 24bpp bitmap is included in `TestImage` (run via `TestWinPRUtils TestImage`).

### DIFF
--- a/winpr/libwinpr/utils/image.c
+++ b/winpr/libwinpr/utils/image.c
@@ -519,8 +519,7 @@ static int winpr_image_bitmap_read_buffer(wImage* image, const BYTE* buffer, siz
 					for (size_t index = 0; index < image->height; index++)
 					{
 						Stream_Read(s, pDstData, scanline);
-						Stream_Seek(s, image->scanline - scanline);
-						pDstData += scanline;
+						pDstData += image->scanline;
 					}
 				}
 				else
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

Issue #5909, opened 2020-02-26
RemoteApp Window Size unable to change


KeepHell:
if you start a RemoteApp for example:

xfreerdp /v:<SERVER> /d:DOMAIN /u:USER /p:'PW'  -floatbar /cert-ignore /rfx /gdi:hw +clipboard +aero +fonts /app:"||APP"

RemoteApp start as expected, but its not possible to handle the window, you could resize it to half of the monitor, so if you click anywhere other than the RemoteApp Window, the RemoteApp window  gets maximized...
 
Issue only exist in Plasma, Gnome is working.
thanks.

akallabeth:
Well, which version did you try, which distro?
Did you give our nightly builds a try? https://github.com/FreeRDP/FreeRDP/wiki/PreBuilds

PR #12915, opened 2026-06-18
[client,rail] add /rail-local-move-size option

chibisuke:
Adds `/rail-local-move-size:[on|off]` (setting `FreeRDP_RemoteAppLocalMoveSize`,
**default `on` — no behavior change**), controlling whether the client advertises
`TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE`.

The client currently always advertises local move/size, so the server hands RemoteApp
window move/resize to the local WM via `_NET_WM_MOVERESIZE`. On some window managers
(notably KWin / KDE Plasma on X11) this handoff misbehaves and RemoteApp windows get stuck
during interactive resize (see #5909). With the option `off`, the client does not advertise it and the server
performs server-side move/resize instead.

**Testing** (KDE Plasma/KWin X11, Windows Server RemoteApp, notepad):
- default → resize hang present, `TS_RAIL_ORDER_LOCALMOVESIZE` PDUs received;
- `-rail-local-move-size` → resize works (server-side), no such PDUs.

### DIFF
--- a/channels/rail/client/client_rails.c
+++ b/channels/rail/client/client_rails.c
@@ -22,7 +22,10 @@ UINT client_rail_server_start_cmd(RailClientContext* context)
 	const rdpSettings* settings = rail->rdpcontext->settings;
 	WINPR_ASSERT(settings);
 
-	clientStatus.flags = TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE;
+	clientStatus.flags = 0;
+
+	if (freerdp_settings_get_bool(settings, FreeRDP_RemoteAppLocalMoveSize))
+		clientStatus.flags |= TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE;
 
 	if (freerdp_settings_get_bool(settings, FreeRDP_AutoReconnectionEnabled))
 		clientStatus.flags |= TS_RAIL_CLIENTSTATUS_AUTORECONNECT;
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -5534,6 +5534,11 @@ static int parse_command_line(rdpSettings* settings, const COMMAND_LINE_ARGUMENT
 			if (rc != 0)
 				return fail_at(arg, rc);
 		}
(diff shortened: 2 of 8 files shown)


client/common/cmdline.c
```diff
@@ -5534,6 +5534,11 @@ static int parse_command_line(rdpSettings* settings, const COMMAND_LINE_ARGUMENT
 			if (rc != 0)
 				return fail_at(arg, rc);
 		}
+		CommandLineSwitchCase(arg, "rail-local-move-size")
```
akallabeth, line 5537: `/tune: <setting:value>,<setting:value>` is for such specific settings, no need to add this to generic command line

channels/rail/client/client_rails.c
```diff
@@ -22,7 +22,10 @@ UINT client_rail_server_start_cmd(RailClientContext* context)
 	const rdpSettings* settings = rail->rdpcontext->settings;
 	WINPR_ASSERT(settings);
 
-	clientStatus.flags = TS_RAIL_CLIENTSTATUS_ALLOWLOCALMOVESIZE;
+	clientStatus.flags = 0;
+
+	if (freerdp_settings_get_bool(settings, FreeRDP_RemoteAppLocalMoveSize))
```
akallabeth, line 27: Maybe use a setting like `FreeRDP_RemoteAppFeatureFlags` insted to control the default flags set?

akallabeth (changes requested):

channels/rail/client/client_rails.c
chibisuke: Sure, I can do that — but I think it'd be pretty inconvenient, because `/tune` always wants integer values. The user would have to figure out that `0x2D5` is the default and that `-0x1` clears `ALLOWLOCALMOVESIZE`.

How would you like me to proceed?

akallabeth: as this is a quite internal setting (and here only required to work around some quirks) I think that is reasonable.

akallabeth:
ok, made the flags a setting with https://github.com/FreeRDP/FreeRDP/pull/12927

PR #12916, opened 2026-06-18
Adjustments for building on Windows with MSVC

sldr:
## Change library type from SHARED to UNKNOWN in detect_package
Fix winpr-3.dll link bug. JsonDetect.cmake declares the target SHARED IMPORTED but sets the .lib as IMPORTED_LOCATION, and never sets IMPORTED_IMPLIB. On Windows, linking a shared imported target requires the import library via IMPORTED_IMPLIB_<CONFIG>. Since that's missing, CMake substitutes its sentinel cjson-NOTFOUND for the import library, and that flows through the file API into winpr-3.dll's link line.
When linking winpr-3.dll, the link can fail if pkg-config finds cjson. For an UNKNOWN IMPORTED library, CMake links IMPORTED_LOCATION directly as a link item on every platform, no separate import library needed. The .lib full path gets linked, cjson-NOTFOUND disappears, and it still should work on Linux/macOS (not able to test).

## Eliminate warning C5103 in WTSAPI32_LOAD_PROC macro
When building with MSVC compiler you will get many warnings:
```
C:\FreeRDP\winpr\libwinpr\wtsapi\wtsapi.c(68): warning C5103: pasting ',' and 'WTS_STOP_REMOTE_CONTROL_SESSION_FN' does not result in a valid preprocessing token
C:\FreeRDP\winpr\libwinpr\wtsapi\wtsapi.c(57): note: in expansion of macro 'WTSAPI32_LOAD_PROC'
```

## Cast keys[j] to uint32_t to eliminate warning C4244: possible loss of data
Investigated if safe to typecast away warning below:
`C:\FreeRDP\channels\rdpdr\client\rdpdr_main.c(384): warning C4244: 'initializing': conversion from 'ULONG_PTR' to 'const uint32_t', possible loss of data`

### DIFF
--- a/channels/rdpdr/client/rdpdr_main.c
+++ b/channels/rdpdr/client/rdpdr_main.c
@@ -381,7 +381,7 @@ static LRESULT CALLBACK hotplug_proc(HWND hWnd, UINT Msg, WPARAM wParam, LPARAM 
 									{
 										if (device_ext->automount)
 										{
-											const uint32_t ids[] = { keys[j] };
+											const uint32_t ids[] = { (uint32_t)keys[j] };
 											WINPR_ASSERT(rdpdr->context.RdpdrUnregisterDevice);
 											error = rdpdr->context.RdpdrUnregisterDevice(
 											    &rdpdr->context, ARRAYSIZE(ids), ids);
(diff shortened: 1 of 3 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

sldr:
THANKS!!!

PR #12927, opened 2026-06-19
Rails feature setting, xfreerdp kbd sync

### LINKED ISSUE #12932: [urbdrc] Redirected USB device name truncated by one character (off-by-one in string-descriptor NUL termination)
# Describe the bug

When a USB device is redirected via the dynamic USB channel (`urbdrc`, libusb
backend), the device's product name shown on the server (Windows) side is
**missing its last character** for many devices.

Root cause is an off-by-one in the NUL-termination of the product string
descriptor in `libusb_udev_control_query_device_text()`
(`channels/urbdrc/client/libusb/libusb_udevice.c`). The terminator is written
*over* the last character of the copied string instead of *after* it, and the
reported `BufferSize` does not account for an appended terminator.

This is device-dependent, which explains why it only happens "in some cases":
devices whose string descriptor declares exactly the text length (the
spec-compliant case, `bLength == 2 + 2*N`, no trailing NUL) lose their last
character; devices that pad/NUL-terminate their descriptor happen to escape it.

Affected code (current `master`, lines 891-901):

```c
const size_t rchar = _wcsnlen((WCHAR*)&data[2], sizeof(data) / sizeof(WCHAR));
len = MIN((BYTE)ret - 2, slen);
len = MIN(len, inSize);
len = MIN(len, rchar * sizeof(WCHAR) + sizeof(WCHAR));
memcpy(Buffer, &data[2], len);

/* Just as above, the returned WCHAR string should be '\0'
 * terminated, but never trust hardware to conform to specs... */
Buffer[len - 2] = '\0';   /* <-- overwrites the last WCHAR of the name */
Buffer[len - 1] = '\0';
*BufferSize = (BYTE)len;
```

A USB string descriptor carries no NUL terminator; the text is exactly
`(bLength - 2) / 2` WCHARs. After the `MIN` chain, `len` is the full byte length
of the copied text, so `Buffer[len - 2]`/`Buffer[len - 1]` land on the **last
character**, not after it.

# To Reproduce

1. Redirect a USB device whose `iProduct` string descriptor is spec-compliant
   (declared length equals the text, no embedded/trailing NUL WCHAR), e.g. via
   `/usb:id,dev:<vid>:<pid>`.
2. On the server side, view the redirected device's friendly name (Device
   Manager / Devices and Printers).
3. The name is missing its final character (e.g. `My USB Devic` instead of
   `My USB Device`).

Walk-through for `"MyDevice"` (8 chars), `inSize = 0xFF`:

<details><summary>byte-level trace</summary>

```
ret = slen = 18  (2-byte header + 16 bytes of text), rchar = 8
len = MIN(16, 18)            = 16
len = MIN(16, 255)           = 16
len = MIN(16, 8*2 + 2 = 18)  = 16      # the +sizeof(WCHAR) slot never materializes
memcpy copies 16 bytes       -> "MyDevice"
Buffer[14] = Buffer[15] = 0  -> overwrites the 8th WCHAR ('e')
*BufferSize = 16
result sent to server: "MyDevic\0"   (last character dropped)
```

Devices that instead declare a trailing NUL (`bLength` one WCHAR longer than the
text) make the `rchar * sizeof(WCHAR) + sizeof(WCHAR)` cap win, so the overwrite
lands on the device's own NUL and no character is lost — hence the
device-to-device variability.

</details>

# Expected behavior

The full product name is preserved, NUL-terminated *after* the text, with
`BufferSize` including the terminator.

# Proposed fix

Append the terminator after the string, reserve room for it, and report the
correct size:

```diff
 const size_t rchar = _wcsnlen((WCHAR*)&data[2], sizeof(data) / sizeof(WCHAR));
+/* Reserve room for the WCHAR null terminator appended below. */
+const size_t maxLen =
+    (inSize >= sizeof(WCHAR)) ? (size_t)(inSize - sizeof(WCHAR)) : 0;
 len = MIN((BYTE)ret - 2, slen);
-len = MIN(len, inSize);
-len = MIN(len, rchar * sizeof(WCHAR) + sizeof(WCHAR));
+len = MIN(len, rchar * sizeof(WCHAR));
+len = MIN(len, maxLen);
 memcpy(Buffer, &data[2], len);

 /* Just as above, the returned WCHAR string should be '\0'
  * terminated, but never trust hardware to conform to specs... */
-Buffer[len - 2] = '\0';
-Buffer[len - 1] = '\0';
-*BufferSize = (BYTE)len;
+Buffer[len] = '\0';
+Buffer[len + 1] = '\0';
+*BufferSize = (BYTE)(len + sizeof(WCHAR));
```

Note the `maxLen` cap (`inSize - sizeof(WCHAR)`) is required: simply moving the
terminator to `Buffer[len]`/`Buffer[len + 1]` without it would write up to 2
bytes past the caller's `inSize`-byte buffer when `len == inSize`. The result
then matches the sibling `DeviceTextLocationInformation` branch, which already
appends the terminator after the text and sets `BufferSize` to include it.

# Application details

* FreeRDP version: confirmed on `master` @ `02991e7b3cc0b9800b09030c0e9c80bab877d668`
  and in release `3.27.1`; the code is long-standing.
* Command line used: any USB redirection, e.g.
  `xfreerdp /v:<host> /u:<user> /usb:id,dev:<vid>:<pid>` (not command-line specific).
* Output of `xfreerdp /buildconfig`: N/A — defect identified by source
  inspection; it is not build- or option-dependent.
* OS version connecting to (server side): any Windows host (the truncated name
  is what the server displays).
* Log output: N/A.

# Environment

 - OS: Linux (also affects any platform using the libusb `urbdrc` backend, e.g. macOS)
 - Version/Distribution: N/A (source-level defect, backend-independent)
 - Architecture: N/A (platform-independent)

# Additional context

Found via code review while investigating field reports of redirected USB device
names missing their last character. The fallback path
(`"Generic Usb String"`) and the `DeviceTextLocationInformation` path are not
affected. A minor, pre-existing concern in the same block (unrelated to this
bug): `_wcsnlen((WCHAR*)&data[2], sizeof(data) / sizeof(WCHAR))` passes a max of
128 WCHARs but only 127 fit after the 2-byte header, so a 255-byte descriptor
with no NUL and a non-zero final byte could over-read 2 bytes past `data`.

### LINKED ISSUE #12934: TestSmartcardPack fails on s390x after recent changes
Our CI says a change from the last ~24 hours caused that the `TestSmartcardPack` test started to fail on our s390x system (Fedora 42). Likely some endianness issue got introduced as little endian platforms are OK.

from `Testing/Temporary/LastTest.log`
````
139/161 Testing: TestSmartcardPack
139/161 Test: TestSmartcardPack
Command: "/home/sharkcz/projects/FreeRDP/build/Testing/TestFreeRDPUtils" "TestSmartcardPack"
Directory: /home/sharkcz/projects/FreeRDP/build/libfreerdp/utils/test
"TestSmartcardPack" start time: Jun 19 13:06 UTC
Output:
[13:06:48:348] [2802085:002ac1a5] [ERROR][com.freerdp.scard.pack] - [smartcard_ndr_read_atrmask]: LocateCards_ATRMask[0]::cbAtr 603979776 exceeds 36
cbAtr=36: expected accept, unpack returned 0xC000003E
<end of output>
Test time =   0.03 sec
Test Failed.
"TestSmartcardPack" end time: Jun 19 13:06 UTC
"TestSmartcardPack" time elapsed: 00:00:00
````

### PR DESCRIPTION
**Pull request by akallabeth**

### CODE REVIEW

File: include/freerdp/rail.h [hunk 1]
```diff
@@ -523,33 +545,35 @@ typedef struct
...
-#define RDP_RAIL_ORDER_GET_APPID_RESP 0x000F
-#define RDP_RAIL_ORDER_LANGUAGEIMEINFO 0x0011
-#define RDP_RAIL_ORDER_COMPARTMENTINFO 0x0012
-#define RDP_RAIL_ORDER_HANDSHAKE_EX 0x0013
-#define RDP_RAIL_ORDER_ZORDER_SYNC 0x0014
-#define RDP_RAIL_ORDER_CLOAK 0x0015
-#define RDP_RAIL_ORDER_POWER_DISPLAY_REQUEST 0x0016
-#define RDP_RAIL_ORDER_SNAP_ARRANGE 0x0017
-#define RDP_RAIL_ORDER_GET_APPID_RESP_EX 0x0018
+#if !defined(WITHOUT_FREERD_3x_DEPRECATED)
```
chibisuke: I think this is a typo (missing P) here.

chibisuke:
Thx a lot, I was still digging through the code and trying to figure out how to best approach this. Much appreciated.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #12933, opened 2026-06-19
[emu,scard] bound select-by-AID compare to GIDS AID length

metsw24-max:
**Out-of-bounds read in vgids_ins_select**
A SELECT by AID compares the APDU bytes against the 11-byte `g_MsGidsAID` using the server-supplied `Lc` as the length, and `Lc` is only capped at `ISO_AID_MAX_SIZE` (16). An `Lc` of 12 to 16 whose first 11 bytes match the AID makes the `memcmp` read past the constant. A selector longer than our AID cannot match anyway, so it now returns "file not found" before the compare.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12936, opened 2026-06-20
[client,sdl] throw a real exception instead of invalid bare throw

TBX3D:
The dialog loops bail to their catch with a bare throw; but with no exception in flight that calls std::terminate(). Throw std::exception() as sdl_disp.cpp does.

### DIFF
--- a/client/SDL/SDL2/dialogs/sdl_input_widgets.cpp
+++ b/client/SDL/SDL2/dialogs/sdl_input_widgets.cpp
@@ -146,13 +146,13 @@ int SdlInputWidgetList::run(std::vector<std::string>& result)
 		while (running)
 		{
 			if (!clear_window(_renderer))
-				throw;
+				throw std::exception();
 
 			if (!update(_renderer))
-				throw;
+				throw std::exception();
 
 			if (!_buttons.update(_renderer))
-				throw;
+				throw std::exception();
 
 			SDL_Event event = {};
 			SDL_WaitEvent(&event);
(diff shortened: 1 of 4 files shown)


akallabeth:
could you please explain the issue you have here?
a bare `throw` will be caught by a `catch(...)` which is done here.
how does `std::terminate` mix in here, that is not the point of these?

TBX3D:
the distinction here is try-body vs catch handler.

these `throw;` are in the try *body*, and `run()` is called from `sdl_dialogs.cpp` (`ilist.run(result)` / `slist.run()`) in normal control flow, with no exception in flight. a bare `throw;` rethrows whatever exception is currently being handled; when there is none, [except.throw] makes it call `std::terminate()` rather than throw something the local `catch (...)` could catch. so on e.g. `clear_window()` failing it aborts instead of falling through to `res = -2`.

same shape, minimal:

```cpp
int run() {
    int res = -1;
    try {
        throw;        // no active exception
    } catch (...) {
        res = -2;     // never reached
    }
    return res;
}
```
```
terminate called without an active exception
Aborted (core dumped)   // exit 134
```

a bare rethrow does what you describe when it runs *inside* a catch, like `client/SDL/common/scoped_guard.hpp:35`, which is correct and the PR leaves alone. throwing a concrete `std::exception()` in these loops keeps the intended path: break, hit `catch (...)`, return `-2`.

akallabeth:
@TBX3D oh, indeed :grin: 
not using exceptions often, so this was something new.

then the only problem remaining: why do you get there, that is an error path? (pr is :+1: )

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #12940, opened 2026-06-20
bound copy-offset source read in mppc_decompress

metsw24-max:
**Out-of-bounds read in the mppc_decompress match copy**

A copy token whose `CopyOffset` is larger than the amount of data decompressed since the last reset (for example a copy token right after `PACKET_AT_FRONT`/`PACKET_FLUSHED`, when `HistoryPtr` still sits at offset 0) wraps the source index to the tail of the 64k history buffer. The destination write is bounded against `HistoryBufferEnd`, but `SrcPtr` is then read forward `LengthOfMatch` bytes with no source bound, so the forward copy runs past the end of `HistoryBuffer` and a malicious server can over-read up to ~64k of heap.

The cause is that the masking only keeps the start index in range; it silently turns an invalid back-reference into a tail wrap instead of rejecting it. Rejecting a `CopyOffset` that points before the start of the valid history keeps both the read and the write in bounds, and matches the sibling decoders: ncrush splits the copy at the wrap boundary and xcrush bounds both ends, so this was the only bulk decoder missing the check.

To test: `TestFreeRDPCodecMppc` gains a regression case that feeds such a token after `PACKET_AT_FRONT` and expects `mppc_decompress` to refuse it. Built with `-fsanitize=address` and `BUILD_TESTING_INTERNAL=ON`; before the change the case reports a heap-buffer-overflow read at the copy loop, after it the input is rejected and the existing decode tests still pass.

### DIFF
--- a/libfreerdp/codec/mppc.c
+++ b/libfreerdp/codec/mppc.c
@@ -417,6 +417,16 @@ int mppc_decompress(MPPC_CONTEXT* mppc, const BYTE* pSrcData, UINT32 SrcSize,
 			return -1005;
 		}
 
+		/* CopyOffset points back into the data already decompressed since the last
+		 * reset. A value larger than that references bytes before the start of the
+		 * history; the masking below would wrap it to the tail of the buffer and the
+		 * forward copy would then read past HistoryBufferEnd. Reject such offsets. */
+		if (CopyOffset > (UINT32)(HistoryPtr - HistoryBuffer))
+		{
+			WLog_ERR(TAG, "invalid CopyOffset, references data before history start");
+			return -1006;
+		}
+
 		SrcPtr = &HistoryBuffer[(HistoryPtr - HistoryBuffer - CopyOffset) &
 		                        (CompressionLevel ? 0xFFFF : 0x1FFF)];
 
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM, nice catch!

and the unit test is really appreciated 😁

akallabeth (changes requested):
Ok, the check is buggy.

just tested a connection to a windows 10 and the decompression routine aborts

akallabeth:
@metsw24-max I think the only valid check you can do here is  `SrcPtr >= `HistoryBuffer` and `SrcPtr + LengthOfMatch <= HistoryBufferEnd`

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/12942

metsw24-max:
You're right, the offset check was too strict. It rejected legitimate wrap-around references, which is what aborted the decode against Windows 10. Bounding the resolved SrcPtr the way you did in #12942 keeps the read in range without assuming the back-reference can't wrap, so that's the correct fix. Thanks for catching it on real traffic and carrying it over.

PR #12943, opened 2026-06-22
[channels,tsmf] bound visible rect read in update_geometry_info

metsw24-max:
tsmf_ifman_update_geometry_info seeks the input stream to pos + numGeometryInfo, where numGeometryInfo is a server-supplied uint32 and stream_setposition only validates the target offset, not that any bytes remain there. it then reads cbVisibleRect and walks num_rects = cbVisibleRect / 16 entries of 16 bytes each with no length check, so a media server that lands the position at or near the end of the buffer reads past it, an out-of-bounds heap read of the channel pdu. the only guard was the initial GUID_SIZE+32 check, which the seek consumes. add the missing length checks before the cbVisibleRect read and before the rect loop, and guard the calloc, so the parse stays within the received data.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

Issue #5208, opened 2019-01-21
Improve error message


hardening:
With latest master, connecting on a server with a non running RDP server, I get this:

```
$ ./output-ogon/client/X11/xfreerdp /v:172.16.100.134 /u:hardening /p:XXXX /rfx /bpp:32
[14:22:45:185] [18813:18814] [INFO][com.freerdp.client.common.cmdline] - loading channelEx cliprdr
[14:22:45:202] [18813:18814] [ERROR][com.freerdp.core.transport] - BIO_should_retry returned a system error 32: Relais brisé (pipe)
[14:22:45:202] [18813:18814] [ERROR][com.freerdp.core.nego] - Protocol Security Negotiation Failure
[14:22:45:202] [18813:18814] [ERROR][com.freerdp.core] - freerdp_set_last_error ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED [0x0002000C]
[14:22:45:202] [18813:18814] [ERROR][com.freerdp.core.connection] - Error: protocol security negotiation or connection failure
```

Except by looking at the code, I guess nobody can guess that the error is a `connection refused` ! There must be a way to improve error message, at least for such common cases.

akallabeth:
Extension of #3006
Agree on that. Should do that after we've merged pending pulls, a final refinement for error handling in v2 would be great.

informatimago:
There are two things here:

1- In normal use. Then I think we should display at INFO level important state transitions.  This can inform the user whether it keeps working normally or when an error is eventually detected.

2- In case of error. Here it is important to show a specific and instructive error message, and I would say also to exit early (since we don't have code to handle errors and retry, but instead we just close down and exit, we could just exit from the function that detected and reported the error).

The only problem in this second case, and demonstrated in the example log above, is that some error is detected in very deep internal functions (BIO_should_retry) that don't know the reason why such an error can occur.  The actual error is reported by the caller (Protocol Security Negotiation Failure); we could (should) have exited then.  The last two error messages reported by the callers are not more informative, or even introduce confusion (is it security negotiation or connection failure now?).

In the code, I think we should distinguish "library" side where errors are detected and reported as result codes, but not logged (or logged at DEBUG level), from "application" side which makes the interface between the core functionality implemented in the "library" side and the user. This "application" side can detect the errors, report them to the user with an ERROR level log, and exit immediately.  Being able to exit immediately upon error would also simplify a lot the error handling code spread everywhere.

[...]

PR #12947, opened 2026-06-22
[core,tcp] log why a tcp connect failed

TBX3D:
When a TCP connection fails the client only reports a generic `ERRCONNECT_CONNECT_FAILED` with no reason. The errno is already retrieved in `freerdp_tcp_connect_timeout` but logged at `WLOG_DEBUG`, so at the default log level the user cannot tell whether the port was closed, the host was unreachable, and so on.

Thread that errno out of `freerdp_tcp_connect_timeout` and log it once at `WLOG_ERROR` in `freerdp_host_connect`, only after every resolved address has been tried. The per-attempt detail stays at debug level, so a multi-address host does not produce a line per address.

Connecting to a closed port, before:

```
[ERROR][com.freerdp.core] - [get_next_addrinfo]: ERRCONNECT_CONNECT_FAILED [0x00020006]
[ERROR][com.freerdp.core] - [freerdp_tcp_default_connect]: Couldn't get socket ip address
```

after (one line added):

```
[ERROR][com.freerdp.core] - [get_next_addrinfo]: ERRCONNECT_CONNECT_FAILED [0x00020006]
[ERROR][com.freerdp.core] - [freerdp_host_connect]: unable to connect to 127.0.0.1: Connection refused [111]
[ERROR][com.freerdp.core] - [freerdp_tcp_default_connect]: Couldn't get socket ip address
```

Only the single-host path is affected; the `TargetNetAddresses` multi-connect path is unchanged.

Tested by pointing the client at a closed port (refused, reason now shown), a reachable TCP listener (gets past TCP, no false reason logged), and an unreachable host (times out, no errno logged).

refs #5208

### DIFF
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -808,7 +808,7 @@ static BOOL freerdp_tcp_is_hostname_resolvable(rdpContext* context, const char* 
 }
 
 static BOOL freerdp_tcp_connect_timeout(rdpContext* context, int sockfd, struct sockaddr* addr,
-                                        size_t addrlen, UINT32 timeout)
+                                        size_t addrlen, UINT32 timeout, INT32* connectError)
 {
 	BOOL rc = FALSE;
 	HANDLE handles[2] = WINPR_C_ARRAY_INIT;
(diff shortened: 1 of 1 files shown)


TBX3D:
To be clear this doesn't close the issue, just partially handles it.

libfreerdp/core/tcp.c
```diff
@@ -868,6 +868,8 @@ static BOOL freerdp_tcp_connect_timeout(rdpContext* context, int sockfd, struct
 			char ebuffer[256] = WINPR_C_ARRAY_INIT;
 			WLog_DBG(TAG, "connect failed with error: %s [%" PRId32 "]",
 			         winpr_strerror(optval, ebuffer, sizeof(ebuffer)), optval);
+			if (connectError)
```
akallabeth, line 871: why not just change the log level of the entry above?

akallabeth:
@TBX3D nice idea, thank you. but I don't like the duplicate log messages (`WLog_DBG` was already there)
added this to the logs I've already had prepared after the last test run and created https://github.com/FreeRDP/FreeRDP/pull/12950

PR #12955, opened 2026-06-23
[channels,remdesk] validate ctl pdu DataLength against received stream

metsw24-max:
The RC_CTL control PDUs on the remdesk server channel carry a DataLength field that remdesk_read_channel_header takes straight off the wire without checking it against the bytes actually received. The three control handlers that parse the trailing UTF-16 strings (remote_control_desktop, authenticate, verify_password) then use DataLength-4 as the scan and conversion bound. Since the stream is only ever sealed to the amount the client actually sent, a client that advertises a larger DataLength than it supplies makes the null-terminator scan walk off the end of the channel buffer, and a DataLength below 4 underflows the bound outright. I noticed it while comparing these handlers against the length checks the other remdesk PDUs already carry. The fix clamps the working length to whatever Stream_GetRemainingLength reports and rewrites the scans as counted loops so the terminator test can no longer read one element past the buffer. I have kept the change inside the three handlers and left the surrounding flow untouched, which should keep it straightforward to review.

### DIFF
--- a/channels/remdesk/server/remdesk_main.c
+++ b/channels/remdesk/server/remdesk_main.c
@@ -171,17 +171,23 @@ static UINT remdesk_recv_ctl_remote_control_desktop_pdu(RemdeskServerContext* co
 	size_t cchStringW = 0;
 	REMDESK_CTL_REMOTE_CONTROL_DESKTOP_PDU pdu = WINPR_C_ARRAY_INIT;
 	UINT error = 0;
-	UINT32 msgLength = header->DataLength - 4;
+
+	if (header->DataLength < 4)
+		return ERROR_INVALID_DATA;
+
+	size_t cchMax = header->DataLength - 4;
+	const size_t remaining = Stream_GetRemainingLength(s);
+	if (cchMax > remaining)
+		cchMax = remaining;
+	cchMax /= sizeof(WCHAR);
+
 	const WCHAR* pStringW = Stream_ConstPointer(s);
 	const WCHAR* raConnectionStringW = pStringW;
 
-	while ((msgLength > 0) && pStringW[cchStringW])
-	{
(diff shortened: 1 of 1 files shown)


channels/remdesk/server/remdesk_main.c
```diff
@@ -171,17 +171,23 @@ static UINT remdesk_recv_ctl_remote_control_desktop_pdu(RemdeskServerContext* co
 	size_t cchStringW = 0;
 	REMDESK_CTL_REMOTE_CONTROL_DESKTOP_PDU pdu = WINPR_C_ARRAY_INIT;
 	UINT error = 0;
-	UINT32 msgLength = header->DataLength - 4;
+
+	if (header->DataLength < 4)
```
akallabeth: check in wrong place.
this should be done one level higher before reading these 4 bytes

akallabeth (changes requested):
Are you trying to implement a server that actually uses this?
would be interesting to have a look.

changes look ok, some modifications required.

channels/remdesk/server/remdesk_main.c
metsw24-max: Moved it up into remdesk_recv_ctl_pdu, right before the msgType read, since that's where those 4 bytes are consumed. The three per-handler guards are redundant now so I've dropped them; the clamp against Stream_GetRemainingLength stays in the handlers.

metsw24-max:
No, I'm not implementing a remdesk server, I came across this while comparing the CTL string handlers against the length checks the other remdesk PDUs already carry. Pushed the fix moving the DataLength check up a level as you suggested.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12956, opened 2026-06-23
[codec,planar] range-check before control byte read in plane rle

metsw24-max:
The two RLE plane decoders in `planar.c` are near-identical, but commit 52f1c87 tightened the input range check in `planar_decompress_plane_rle_only` and left the sibling `planar_decompress_plane_rle` untouched. The sibling still reads the control byte through `*srcp` first and only afterwards compares the advanced position against `SrcSize`, so the bounds check happens one byte too late. On a truncated or crafted plane the dereference can touch the byte just past the end of the plane buffer before the function bails out. I noticed it while reading the diff for the range-check fix and checking whether the same shape existed elsewhere. The change here is the same reorder applied in that commit: compute the offset, reject when `cur + 1` would exceed `SrcSize`, then read. It keeps the two decoders consistent and the read strictly inside the validated region; behaviour for well-formed planes is unchanged.

### DIFF
--- a/libfreerdp/codec/planar.c
+++ b/libfreerdp/codec/planar.c
@@ -398,14 +398,15 @@ static inline INT32 planar_decompress_plane_rle(const BYTE* WINPR_RESTRICT pSrcD
 
 		for (INT32 x = 0; x < (INT32)nWidth;)
 		{
-			const BYTE controlByte = *srcp;
-			srcp++;
-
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM, nice catch

PR #12962, opened 2026-06-24
[channels,rdpsnd] bound client format pdu length to UINT16

metsw24-max:
**16-bit overflow building the client formats PDU**

The SNDC_FORMATS response size is summed into a `UINT16` from `18 + cbSize` per format, but `cbSize` comes straight from the server's announced format list, so a list totalling more than 65535 bytes wraps the accumulator and `Stream_New` hands back a buffer too small for the fixed 24-byte header written into it. The sibling `rdpsnd_server_send_formats` already bounds the mirror case at `UINT16_MAX`, so I widened the accumulator to `size_t` and applied the same check before allocating; normal format lists are unaffected.

Repro: a malicious server sends an `SNDC_FORMATS` PDU carrying PCM formats whose `cbSize` extension blocks sum past 64 KiB.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #12971, opened 2026-06-25
[channels,urbdrc] bound msusb descriptor reads to received length

metsw24-max:
The USB redirection channel parses the MS-RDPEUSB configuration, interface and pipe descriptors out of a SELECT_CONFIGURATION or SELECT_INTERFACE request sent by the server. The three readers in msusb.c sized their reads with Stream_CheckAndLogRequiredCapacity, which reports the room left in the underlying allocation rather than the amount of data actually received; when the channel buffer comes from the stream pool its capacity is usually larger than the sealed length, so a server that declares more interfaces or pipes than it sent walks the reader off the end of the message and into stale pool bytes that are then echoed back in the result PDU. msusb_msconfig_read is worse, because after the interface list it reads the six byte configuration trailer with no length check at all; the only guard is the initial (3 + NumInterfaces) * 2 test, well below the 12 * NumInterfaces + 6 bytes it goes on to consume, so a single interface followed by a truncated trailer already reads past the buffer (caught by the stream assertions in debug builds, a heap over-read once they are compiled out). I noticed it while checking why the URB request handlers validate with required-length but the descriptor helpers they call do not. The fix moves the three helpers to Stream_CheckAndLogRequiredLength and adds the missing trailer check, which leaves valid descriptors working and rejects the truncated ones, and should keep the parsing easier to reason about going forward.

Testing: a small TestMsUsb case under channels/urbdrc/common/test covers a complete descriptor plus the three malformed shapes; run it with `ctest -R TestMsUsb`.

### DIFF
--- a/channels/urbdrc/common/CMakeLists.txt
+++ b/channels/urbdrc/common/CMakeLists.txt
@@ -23,3 +23,7 @@ set_property(TARGET urbdrc-common PROPERTY FOLDER "Channels/${CHANNEL_NAME}/Comm
 
 freerdp_client_pc_add_library_private(urbdrc-common)
 channel_install(urbdrc-common ${FREERDP_ADDIN_PATH} "FreeRDPTargets")
+
+if(BUILD_TESTING_INTERNAL OR BUILD_TESTING)
+  add_subdirectory(test)
+endif()
--- a/channels/urbdrc/common/msusb.c
+++ b/channels/urbdrc/common/msusb.c
@@ -60,7 +60,7 @@ static MSUSB_PIPE_DESCRIPTOR** msusb_mspipes_read(wStream* s, UINT32 NumberOfPip
 {
 	MSUSB_PIPE_DESCRIPTOR** MsPipes = nullptr;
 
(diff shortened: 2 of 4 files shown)


akallabeth:
@freerdp-bot test

metsw24-max:
The sanitiser run caught a real leak on the new error path, so thanks for kicking off CI. When the trailer length check bailed, msusb_msconfig_read freed the interface array but not the descriptors inside it: NumInterfaces wasn't set on the config yet, so the cleanup loop ran zero times. Fixed by recording NumInterfaces right after the interface list is read, so the failure path frees everything. TestMsUsb passes clean now, confirmed with a leak check locally as well.

akallabeth:
@metsw24-max did you test to redirect a `USB` device?
this path is quite fickle and small changes break the usb redirection.

from code changes alone it looks good now (lets see if ci passes)

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

did a quick test with redirecting my webcam and it still works, nice work!

PR #12981, opened 2026-06-29
[client,ios] remove some deprecation warnings

akallabeth:


bho3538:
@akallabeth I checked the ios project `client/iOS/cmake/CMakeLists.txt` change. When generating the iOS Xcode project from the terminal (with the command from `docs/README.ios`), `-DCMAKE_C_STANDARD:STRING="23"` fails with C_STANDARD is set to invalid value '"23"'. Dropping the quotes (`-DCMAKE_C_STANDARD:STRING=23`) resolves it. Could you confirm? (Or should I open a new PR?)

generate_ios_simulator_project_error

akallabeth:
@bho3538 ok, please do create a new pr. (looks like the bug is not triggered by the ci, so I missed that)
did have a look and the client build and it should be modified to be like the android build. (did miss that before as well)

e.g. 
* `client/ios` not added to the `FreeRDP` build
* The `client/ios` `CMake` project should be a self contained project that just builds the dependencies with `ExternalProject`, including the `FreeRDP` libraries

I'll do an update on the `ci` to properly build the project then

bho3538:
@akallabeth I create new PR (https://github.com/FreeRDP/FreeRDP/pull/12990). Could you check it?

PR #12949, opened 2026-06-23
feat(client): Modernize iOS client. (iFreeRDP)

bho3538:
Update iOS client (iFreeRDP) to support modern iOS devices.

[Changed]
- Modern device support (require 64bit device, iOS 15 or higher version)
- Support iPad keyboard & mouse
- Improved on-screen cursor movement on touch (like official MS ios Windows App)

https://github.com/user-attachments/assets/fa5e73f0-f79b-4d59-bd89-3bfcec8e8d6b

The build method is updated in the 'docs/README.ios' document.

There are still many features left to be added (clipboard, dark mode support, etc.), but for now i modified it to be available on the latest iOS devices.

PR #12990, opened 2026-07-01
[client,ios] build the iOS client like the Android client and fix some error.

bho3538:
- An error causing syntax errors when creating an Xcode project has been fixed. (`-DCMAKE_C_STANDARD:STRING="23"` --> `-DCMAKE_C_STANDARD:STRING=23`)
- The iOS client is no longer built as part of the main FreeRDP tree like Android build.  I am not familiar with CMake configuration, so I received assistance from AI for this section)


client/iOS/cmake/CMakeLists.txt
```diff
@@ -41,45 +41,39 @@ if(WITH_FFMPEG)
...
-    -DWITH_MANPAGES:BOOL=OFF
-    -DFREERDP_IOS_EXTERNAL_SSL_PATH:PATH=${DEPS_INSTALL_DIR}
-    -DWITH_INTERNAL_MD4:BOOL=ON
-    -DWITH_INTERNAL_RC4:BOOL=ON
-    -DWITH_VIDEOTOOLBOX:BOOL=ON
-    -DWITH_FFMPEG:BOOL=${WITH_FFMPEG}
-    -DWITH_VIDEO_FFMPEG:BOOL=${WITH_FFMPEG}
-    -DWITH_DSP_FFMPEG:BOOL=OFF
-    -DWITH_SWSCALE:BOOL=OFF
+# arguments shared by the FreeRDP library build and the iFreeRDP app build
```
akallabeth: remove this file, should be `client/iOS/CMakeLists.txt` that just includes the `ExternalProjects`.
Remove the `add_subdirectory` for `iOS` in `client/CMakeLists.txt` like with the android client

client/iOS/CMakeLists.txt
```diff
@@ -25,12 +25,16 @@ project(iFreeRDP VERSION ${FREERDP_DEFAULT_PROJECT_VERSION})
 
 message("project ${PROJECT_NAME} is using version ${PROJECT_VERSION}")
 
-list(APPEND CMAKE_MODULE_PATH ${PROJECT_CURRENT_SOURCE_DIR}/../../cmake/)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../../cmake/)
```
akallabeth: add the `cmake` subdirectory to find `ExternalProjects` and add them here.

akallabeth:
@bho3538 best just take the `CMake` build files from the android client and modify the arguments passed so that the proper iOS compiler/options are used.

I've further simplified the android build with https://github.com/FreeRDP/FreeRDP/pull/12992 so you now only need the following modifications in `client/iOS/CMakeLists.txt`:
1. Remove the `client/iOS` from the main build in `client/CMakeLists.txt`
2. copy `client/Android/cmake` to `client/iOS/cmake`
   * modify the `External*cmake` files to build on `iOS`, e.g. compiler paths, defines, ...
4. in `client/iOS/CMakeLists.txt` add
   * Add `client/iOS/cmake` to the module search path
   * `include(ExternalDeps)` 
   * `add_dependencies(${MODULE_NAME} freerdp)`

bho3538:
I've checked, I will fix and commit again.

akallabeth:
@bho3538 btw, you should have a look at https://cmake.org/cmake/help/latest/variable/CMAKE_XCODE_ATTRIBUTE_an-attribute.html#variable:CMAKE_XCODE_ATTRIBUTE_%3Can-attribute%3E and remove the manual code signing variables.

akallabeth:
@bho3538 did some cleanups in https://github.com/akallabeth/FreeRDP/pull/new/ios-build but did not have time to adjust the external builds from android to iOS.

code siging should also be done in the external build, but for `CMake` based projects it most likely is sufficient to pass the `CMAKE_XCODE_*` properties down

client/iOS/cmake/ExternalDeps.cmake
```diff
@@ -0,0 +1,43 @@
+include(GNUInstallDirs)
+
+set(DEPS_INSTALL_DIR "${CMAKE_BINARY_DIR}/deps")
+file(MAKE_DIRECTORY ${DEPS_INSTALL_DIR})
+
+option(WITH_OPENSSL "Build and enable OpenSSL" ON)
+option(WITH_FFMPEG "Build and enable FFmpeg codec support" ON)
```
akallabeth, line 9: why not add options for the other libraries as well here?

client/iOS/CMakeLists.txt
```diff
@@ -25,22 +29,26 @@ project(iFreeRDP VERSION ${FREERDP_DEFAULT_PROJECT_VERSION})
...
-set(FREERDP_IOS_BUNDLE_IDENTIFIER "com.freerdp.ifreerdp" CACHE STRING "iFreeRDP iOS bundle identifier")
-set(FREERDP_IOS_DEVELOPMENT_TEAM "" CACHE STRING "iFreeRDP iOS development team identifier")
-set(FREERDP_IOS_CODE_SIGN_STYLE "" CACHE STRING "iFreeRDP iOS code signing style")
+list(APPEND CMAKE_MODULE_PATH ${FREERDP_SOURCE_DIR}/cmake/)
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/)
+include(CommonConfigOptions)
+include(ProjectCStandard)
+include(ProjectCXXStandard)
+
+set(CMAKE_C_STANDARD 23 CACHE STRING "project default" FORCE)
```
akallabeth, line 40: remove this, already handled by `include(ProjectCStandard)`

client/iOS/CMakeLists.txt
```diff
@@ -25,22 +29,26 @@ project(iFreeRDP VERSION ${FREERDP_DEFAULT_PROJECT_VERSION})
...
+set(CMAKE_C_STANDARD_REQUIRED ON CACHE BOOL "project default" FORCE)
```
akallabeth: remove this, already handled by include(ProjectCXXStandard)

client/iOS/CMakeLists.txt
```diff
@@ -25,22 +29,26 @@ project(iFreeRDP VERSION ${FREERDP_DEFAULT_PROJECT_VERSION})
...
+set(CMAKE_XCODE_ATTRIBUTE_IPHONEOS_DEPLOYMENT_TARGET "15.0" CACHE STRING "project default")
```
akallabeth: https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html

client/iOS/cmake/ExternalDeps.cmake
```diff
@@ -0,0 +1,43 @@
...
+    -DCMAKE_TOOLCHAIN_FILE:FILEPATH=${_toolchain_abs} -DPLATFORM:STRING=${PLATFORM}
+    -DDEPLOYMENT_TARGET:STRING=${DEPLOYMENT_TARGET} -DIOS_TARGET_SDK:STRING=${DEPLOYMENT_TARGET}
+    -DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE} -DBUILD_SHARED_LIBS:BOOL=ON
+)
+
+set(IOS_NATIVE_DEPS "")
+
+if(WITH_OPENSSL)
+  include(ExternalOpenSSL)
+  list(APPEND IOS_NATIVE_DEPS openssl)
```
akallabeth, line 61: Should not be required, if you install to a intermediate target just add that to the search path

client/iOS/cmake/ExternalDeps.cmake
```diff
@@ -0,0 +1,43 @@
...
+  set(OPENSSL_SSL_LIBRARY ${DEPS_INSTALL_DIR}/${CMAKE_INSTALL_LIBDIR}/libssl.a)
+  set(OPENSSL_CRYPTO_LIBRARY ${DEPS_INSTALL_DIR}/${CMAKE_INSTALL_LIBDIR}/libcrypto.a)
+  set(OPENSSL_LIBRARIES ${OPENSSL_SSL_LIBRARY} ${OPENSSL_CRYPTO_LIBRARY})
+  set(OPENSSL_FOUND ON)
+endif()
+
+set(IOS_FFMPEG_ARGS "")
+if(WITH_FFMPEG)
+  include(ExternalFFmpeg)
+  list(APPEND IOS_NATIVE_DEPS ffmpeg)
```
akallabeth, line 96: Should not be required, if you install to a intermediate target just add that to the search path

client/iOS/cmake/ExternalFreeRDP.cmake
```diff
@@ -1,27 +1,85 @@
...
     -DWITH_VIDEOTOOLBOX:BOOL=ON
     -DWITH_FFMPEG:BOOL=${WITH_FFMPEG}
     -DWITH_VIDEO_FFMPEG:BOOL=${WITH_FFMPEG}
     -DWITH_DSP_FFMPEG:BOOL=OFF
     -DWITH_SWSCALE:BOOL=OFF
-    ${_ios_ffmpeg_args}
+    ${IOS_FFMPEG_ARGS}
 )
 
+if(DEFINED CMAKE_XCODE_ATTRIBUTE_DEVELOPMENT_TEAM)
```
akallabeth: just add these once in the `ExternalDeps.cmake` to `IOS_CMAKE_ARGS`?

client/iOS/cmake/ExternalDeps.cmake
akallabeth: most likely you're looking for https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

akallabeth:
@bho3538 a few build fixes here: https://github.com/akallabeth/FreeRDP/tree/ios-cmake (mainly the `CMake` based builds)

client/iOS/CMakeLists.txt
bho3538: I have a question.
If remove this line, iOS projects using C11 (global freerdp settings in `cmake/ProjectCStandard.cmake`)
C11 does not have `nullptr` keyword and it makes build error.

In that case, as I mentioned earlier, you need to modify it as shown in the link below. Do you still need to remove this CMake configuration?
https://github.com/FreeRDP/FreeRDP/pull/12949#discussion_r3492849730

image
I pulled the branch from this link and attempted to build it: https://github.com/akallabeth/FreeRDP/tree/ios-cmake

client/iOS/cmake/ExternalFreeRDP.cmake
```diff
@@ -29,10 +28,9 @@ set(FREERDP_EXTRA_CMAKE_ARGS
     -DWITH_VIDEOTOOLBOX:BOOL=ON
     -DWITH_FFMPEG:BOOL=${WITH_FFMPEG}
     -DWITH_VIDEO_FFMPEG:BOOL=${WITH_FFMPEG}
-    -DWITH_DSP_FFMPEG:BOOL=${WITH_FFMPEG}
-    -DWITH_SWSCALE:BOOL=${WITH_FFMPEG}
-    -DWITH_SWSCALE:BOOL=${WITH_URIPARSER}
-    ${IOS_FFMPEG_ARGS}
+    # iOS FFmpeg builds only avcodec/avutil (no swresample/swscale)
```
akallabeth, line 31: why?

akallabeth:
@bho3538 ok, taken your commits and created #12997 
this is building on my test machine now with all dependencies (except `uriparser` due to a bug on their end)

client/iOS/cmake/ExternalFreeRDP.cmake
bho3538: Because ios app does not support audio yet. For this reason, only the option for H.264 video decoding is enabled in FFmpeg.(so enabling that option causes a build error). It seems we can simply enable that option along with the relevant FFmpeg compilation settings when the app adds audio support in the future.

PR #12995, opened 2026-07-02
[winpr,string] fix winpr_strnstr needle length (gateway SIGSEGV)

kapott:
### Summary

`winpr_strnstr` (added in 3521712b6) uses `strnlen(needle, hlen)` to measure the needle. That clamps the needle length to the haystack window, so:

- the `if (hlen < needle_len) return nullptr;` guard is unreachable (`needle_len` can never exceed `hlen`), and
- a partial needle matches on its prefix; at `hlen == 0` the function returns `haystack` instead of `NULL`.

### Impact

`http_response_recv()` in `libfreerdp/core/gateway/http.c` calls `winpr_strnstr` in a loop with a shrinking `hlen`:

```c
winpr_strnstr(line, "\r\n", payloadOffset - (line - buffer) - 2)
```

When `line` reaches `buffer + payloadOffset - 2`, `hlen == 0`. Correct behavior returns `NULL` and ends the loop. The bug returns a false match, `line += 2` runs past the buffer, and the next iteration's `hlen` underflows to `~SIZE_MAX`, so `winpr_strnstr` scans far out of bounds -> SIGSEGV.

Reproduced connecting through an RD Gateway on a glibc system (fallback path, `WINPR_HAVE_STRNSTR` unset): the crash fires right after 2FA. Every gateway HTTP response drives the loop to `hlen == 0`, so it is deterministic (100%), not a race. The HTTP response is server-controlled, so this is remotely reachable (crash / OOB read, DoS-class).

Resolved backtrace:

```
winpr_strnstr        winpr/libwinpr/crt/string.c
http_response_recv   libfreerdp/core/gateway/http.c:1327
rdp_client_connect
freerdp_connect
```

Introduced in 3521712b6 (2026-06-30); master only, no release affected.

### Reproduction

Unit-level (deterministic, no network):

```c
winpr_strnstr("ab", "\r\n", 0);  /* returned "ab", must be NULL */
winpr_strnstr("x!", "xy", 1);    /* returned "x!", must be NULL */
```

### Fix

Measure the needle with `strlen(needle)`. This restores the `hlen < needle_len` guard, so the loop terminates at `hlen == 0` with no underflow. Only the fallback path (`!WINPR_HAVE_STRNSTR`, e.g. glibc) was affected; the native `strnstr` path is unchanged.

Adds `test_winpr_strnstr()` to `TestString.c` covering the regression cases and basic matches.

### DIFF
--- a/winpr/libwinpr/crt/string.c
+++ b/winpr/libwinpr/crt/string.c
@@ -848,10 +848,20 @@ char* winpr_strnstr(char* haystack, const char* needle, size_t hlen)
 	WINPR_ASSERT(haystack || (hlen == 0));
 	WINPR_ASSERT(needle);
 
+	/* Keep the contract identical on the native and fallback paths: nothing to
+	 * find in an empty window, and SIZE_MAX would overflow the hlen + 1 bound
+	 * below. */
+	if (hlen == 0)
+		return nullptr;
+	if (hlen == SIZE_MAX)
+		return nullptr;
+
 #if defined(WINPR_HAVE_STRNSTR)
 	return strnstr(haystack, needle, hlen);
 #else
-	const size_t needle_len = strnlen(needle, hlen);
+	/* Bound the needle scan to the haystack: needle_len becomes hlen + 1 when
+	 * the needle is longer than hlen, which the check below rejects. */
+	const size_t needle_len = strnlen(needle, hlen + 1);
 
 	if (0 == needle_len)
 		return haystack;
--- a/winpr/libwinpr/crt/test/TestString.c
+++ b/winpr/libwinpr/crt/test/TestString.c
@@ -104,6 +104,55 @@ fail:
 	return rc;
 }
 
+static BOOL test_winpr_strnstr(void)
+{
+	struct strnstr_test
+	{
+		const char* haystack;
+		const char* needle;
+		size_t hlen;
+		int expected; /* offset of the match, or -1 when NULL is expected */
+	};
+	const struct strnstr_test tests[] = {
+		/* regression: a needle longer than hlen must never match on a prefix.
+		 * winpr_strnstr used to clamp the needle length to hlen, so these
+		 * returned the haystack instead of NULL (see gateway http_response_recv
+		 * SIGSEGV). */
+		{ "ab", "\r\n", 0, -1 },
+		{ "x!", "xy", 1, -1 },
+		{ "abc", "abc", 2, -1 },
+		/* an empty needle returns the haystack */
+		{ "abc", "", 5, 0 },
+		/* basic matches */
+		{ "a\r\nb", "\r\n", 4, 1 },
+		{ "xxab", "ab", 4, 2 },
+		{ "xxxx", "ab", 4, -1 },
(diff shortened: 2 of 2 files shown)


winpr/libwinpr/crt/string.c
```diff
@@ -851,7 +851,7 @@ char* winpr_strnstr(char* haystack, const char* needle, size_t hlen)
 #if defined(WINPR_HAVE_STRNSTR)
 	return strnstr(haystack, needle, hlen);
 #else
-	const size_t needle_len = strnlen(needle, hlen);
+	const size_t needle_len = strlen(needle);
```
akallabeth: the `needle` must be smaller or equal than `haystack`, so the correct thing would be `const size_t needle_len = strnlen(needle, hlen + 1);`

akallabeth:
```
iff --git a/winpr/libwinpr/crt/string.c b/winpr/libwinpr/crt/string.c
index 48260f9f54..14600075ab 100644
--- a/winpr/libwinpr/crt/string.c
+++ b/winpr/libwinpr/crt/string.c
@@ -851,8 +851,12 @@ char* winpr_strnstr(char* haystack, const char* needle, size_t hlen)
 #if defined(WINPR_HAVE_STRNSTR)
        return strnstr(haystack, needle, hlen);
 #else
-       const size_t needle_len = strlen(needle);
+       if (hlen == 0)
+               return nullptr;
+       if (hlen == SIZE_MAX)
+               return nullptr;
 
+       const size_t needle_len = strnlen(needle, hlen + 1);
        if (0 == needle_len)
                return haystack;
```

akallabeth (changes requested):

akallabeth:
@kapott nice catch, and thank you for the unit test.
would prefer to keep the `strnlen` though, to have the check bounded (even if only in relation to `haystack`)

kapott:
Thanks for the review. Adopted your approach: kept `strnlen` but bounded with `strnlen(needle, hlen + 1)` so a needle longer than `hlen` yields `needle_len == hlen + 1` and is rejected by the length check, and added the `hlen == 0` / `hlen == SIZE_MAX` guards to keep `hlen + 1` from overflowing. Extended the unit test with those two guard cases. Pushed.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM, thank you!

kapott:
Thanks for the rapid response and handling!

akallabeth:
@kapott looks like your unit test does not (yet) pass?
looks like it does not work with `strnstr` cases where the string is actually shorter than the length you supply?

akallabeth:
@freerdp-bot test

akallabeth (approved):

kapott:
Hm, I might have jumped the gun there.. The real gap is that the fallback scans past an embedded NUL while native strnstr stops at it, so they disagree whenever hlen is longer than the actual string. I think the clean fix is to make the fallback behave like native strnstr (stop at NUL or hlen, whichever comes first), which also lets the SIZE_MAX/hlen+1 guards go away.

PR #13001, opened 2026-07-03
[codec,dsp] ensure out capacity in opus encode

metsw24-max:
**Wrong stream in the opus encode capacity check**

`freerdp_dsp_encode_opus` reserves `max_size` bytes on `context->common.buffer`, but `opus_encode` is handed `Stream_Pointer(out)` together with `max_size` as its output limit, so the destination stream `out` is never grown. When `out` has less remaining capacity than the packet the encoder emits, the write runs past its buffer.

Cause is the reservation targeting the scratch buffer instead of `out`. This is the encode side of the decode fix in 0ed1f95; the other encoders (adpcm, gsm610, mp3, faac) already grow `out`, so opus was the odd one out. Reachable once a peer negotiates `WAVE_FORMAT_OPUS`, where the rdpsnd server and audin client encode into a reused channel stream.

Fix reserves the space on `out`, matching the decode path a few lines above.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13002, opened 2026-07-03
[winpr,string] match winpr_strnstr fallback to native strnstr

kapott:
### Summary

Follow-up to #12995 (merged as 444bcbba). That PR fixed the RD Gateway
crash, but left a behavioral gap between the two code paths in
winpr_strnstr: when hlen is larger than the actual string, the fallback
kept scanning up to hlen bytes past an embedded NUL, while the native
strnstr path stops at the NUL. The fallback also needed hlen == 0 /
hlen == SIZE_MAX guards to keep the strnlen(needle, hlen + 1) bound from
overflowing.

Demonstration of the divergence on the merged code:

    winpr_strnstr("ab\0cd", "cd", 5)  // fallback: found at offset 3
                                      // native:   NULL (stops at NUL)

### Change

Rewrite the fallback to mirror native strnstr: measure the needle with
strlen and walk the haystack until the NUL terminator or hlen is
reached, whichever comes first. Both paths now behave identically for
every input, the special-case guards go away, and there is no hlen + 1
overflow to guard against.

This is a correctness/consistency fix, not a crash fix - the gateway
http_response_recv() loop is bounded by hlen and still terminates, so
the original SIGSEGV cannot recur.

### Tests

TestString.c: aligned the SIZE_MAX case with native semantics (an
over-long hlen just searches the whole string) and added an embedded-NUL
case that fails on the old fallback and passes on both paths now.

### DIFF
--- a/winpr/libwinpr/crt/string.c
+++ b/winpr/libwinpr/crt/string.c
@@ -848,33 +848,20 @@ char* winpr_strnstr(char* haystack, const char* needle, size_t hlen)
 	WINPR_ASSERT(haystack || (hlen == 0));
 	WINPR_ASSERT(needle);
 
-	/* Keep the contract identical on the native and fallback paths: nothing to
-	 * find in an empty window, and SIZE_MAX would overflow the hlen + 1 bound
-	 * below. */
-	if (hlen == 0)
-		return nullptr;
-	if (hlen == SIZE_MAX)
-		return nullptr;
-
 #if defined(WINPR_HAVE_STRNSTR)
 	return strnstr(haystack, needle, hlen);
 #else
-	/* Bound the needle scan to the haystack: needle_len becomes hlen + 1 when
-	 * the needle is longer than hlen, which the check below rejects. */
-	const size_t needle_len = strnlen(needle, hlen + 1);
+	/* Match native strnstr semantics: search at most hlen characters and stop at
+	 * the haystack NUL terminator, whichever comes first. */
+	const size_t needle_len = strlen(needle);
 
 	if (0 == needle_len)
 		return haystack;
 
-	if (hlen < needle_len)
-		return nullptr;
-
-	for (size_t i = 0; i <= hlen - needle_len; i++)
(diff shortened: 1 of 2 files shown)


winpr/libwinpr/crt/string.c
```diff
@@ -848,33 +848,20 @@ char* winpr_strnstr(char* haystack, const char* needle, size_t hlen)
...
-
 #if defined(WINPR_HAVE_STRNSTR)
 	return strnstr(haystack, needle, hlen);
 #else
-	/* Bound the needle scan to the haystack: needle_len becomes hlen + 1 when
-	 * the needle is longer than hlen, which the check below rejects. */
-	const size_t needle_len = strnlen(needle, hlen + 1);
+	/* Match native strnstr semantics: search at most hlen characters and stop at
+	 * the haystack NUL terminator, whichever comes first. */
+	const size_t needle_len = strlen(needle);
```
akallabeth, line 856: why not `const size_t needle_len = strnlen(needle, MIN(SIZE_MAX, hlen + 1));` ?
keeps the search bound and eliminates the `hlen == SIZE_MAX` check as well.

akallabeth (approved):

winpr/libwinpr/crt/string.c
kapott: Thanks! I went with strlen(needle) on purpose - native strnstr uses the full needle length too, so this keeps the two paths identical. MIN(SIZE_MAX, hlen + 1) doesn't actually drop the guard though: hlen + 1 wraps to 0 before MIN sees it, so at hlen == SIZE_MAX you end up with strnlen(needle, 0) == 0 and a false match (returns haystack for a needle that isn't there). Bounding the needle read by hlen is also what caused the original crash - needle is a NUL-terminated C-string, so there's nothing unbounded to guard against here. Happy to add an explicit bound if you'd prefer, but strlen matches native and sidesteps the overflow entirely.

PR #13005, opened 2026-07-03
[channels,rdpecam] fix DeviceAdded notification for non-ASCII device names

Fantu:
I had problems with the webcam Microsoft LifeCam HD 5000 not working on rdpecam and wasted a lot of time in many tests and debugging (because there was issue also in other cases out of freerdp which made me assume otherwise and do useless tests), finally found the problem in the webcam name containing non-ascii characters, made this patch with claude and tested it on the rpi4 arm64 [adding it to the current trixie-backports version](https://github.com/M2Rbiz/freerdp3-pkg/commits/m2r-pkg/) and now also the ms webcam works.
I think fix the problem also for other microsoft webcam with same issue and other if name containing non-ascii characters.

Thanks a lot to the people who implement rdpecam and helped to fix, test and improve it, it's very helpful and was really needed.

### DIFF
--- a/channels/rdpecam/client/camera_device_enum_main.c
+++ b/channels/rdpecam/client/camera_device_enum_main.c
@@ -19,6 +19,7 @@
 
 #include <winpr/assert.h>
 #include <winpr/cast.h>
+#include <winpr/string.h>
 
 #include "camera.h"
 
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
This change looks suspicious.
`Stream_Write_UTF16_String_From_UTF8` does the same as your proposed *fix* 

so, could you please explain where there is a problem, as in where it fails for your device?

akallabeth (approved):
Ok, figured out your problem with some tests:
the `UTF-8` character count is different than the `WCHAR` character count, so the resulting name is truncated/\`\0` extended.

akallabeth:
@freerdp-bot test

akallabeth:
ok, taken your commit and added to https://github.com/FreeRDP/FreeRDP/pull/13012 (with formatting fixed)
thank you for pointing this one out.

PR #13008, opened 2026-07-04
[core,rdstls] add endpoint FedAuth token authentication

thomasDOTwtf:
Implement rdstls_write_authentication_request_with_fedauth_token so FreeRDP can authenticate against Hyper-V VM console-connect endpoints that require an endpoint FedAuth token instead of user credentials or an auto-reconnect cookie.

FedAuth token handover implemented for:
- FreeRDP_EndpointFedAuthToken settings entry
- endpointfedauth .rdp file field
- /endpointfedauth command line switch

When the token is set the RDSTLS security protocol is enabled automatically and the authentication request is sent using RDSTLS version 2 with data type 0x03. This is needed for usage of the FedAuth token method agains a Hyper-V host.

## How to test?

For testing the FedAuth against the Host you will need the following:
* System Center Virtual Machine Manager for issuing the tokens
* Hyper-V Host running at least one virtual machine
* Remote Desktop Gateway with Console Connect Federated Authentication Plugin

If a lab environment is needed for merging it please contact me.

The tokens can be generated with powershell on SCVMM:
`$tok = New-SCVMConnectFedAuth -VMMServer $vmm -VM $vm
`

Following command line parameter should be used to hand over the tokens:
```
/v:$($tok.Host):$($tok.VMRCPort)
/gateway:g:<rdp gw fqdn>,type:http,access-token:$($tok.HostToken)
/pcb:$($tok.VMID)
/endpointfedauth:$(q $tok.VMToken)
```
Beware of the quoting of the gateway access token.

## Not covered by this PR

  - Server side handling of RDSTLS_DATA_FEDAUTH_TOKEN in rdstls_process_authentication_request_* (this change only implements the client sender)
  - Session resume with an expired token, because it never works without a new token provided by the user

## Compatibility

  - New setting slot 1157 is placed in a previously padded region, guarded by SETTINGS_DEPRECATED
  - Public function signatures unchanged
  - Existing RDSTLS password and auto-reconnect-cookie paths are unmodified
  - Providing the FedAuth endpoint token is handled in a new branch
  - /sec:rdstls remains valid. Existing .rdp files that don't contain endpointfedauth work unchanged

### DIFF
--- a/libfreerdp/core/connection.c
+++ b/libfreerdp/core/connection.c
@@ -435,7 +435,15 @@ BOOL rdp_client_connect(rdpRdp* rdp)
 	nego_enable_tls(rdp->nego, settings->TlsSecurity);
 	nego_enable_nla(rdp->nego, settings->NlaSecurity);
 	nego_enable_ext(rdp->nego, settings->ExtSecurity);
-	nego_enable_rdstls(rdp->nego, settings->RdstlsSecurity);
+	/* An endpoint FedAuth token is delivered via RDSTLS. Enable the protocol
+	 * implicitly so the token can be used without an additional /sec flag. */
+	{
+		const char* fedAuthToken =
+		    freerdp_settings_get_string(settings, FreeRDP_EndpointFedAuthToken);
+		const BOOL enableRdstls =
+		    settings->RdstlsSecurity || (fedAuthToken && (*fedAuthToken != '\0'));
+		nego_enable_rdstls(rdp->nego, enableRdstls);
+	}
 	nego_enable_aad(rdp->nego, settings->AadSecurity);
 
 	if (settings->MstscCookieMode)
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -5298,6 +5298,11 @@ static int parse_command_line(rdpSettings* settings, const COMMAND_LINE_ARGUMENT
 			if (!freerdp_settings_set_bool(settings, FreeRDP_NegotiateSecurityLayer, enable))
 				return fail_at(arg, COMMAND_LINE_ERROR);
 		}
+		CommandLineSwitchCase(arg, "endpointfedauth")
+		{
+			if (!freerdp_settings_set_string(settings, FreeRDP_EndpointFedAuthToken, arg->Value))
+				return fail_at(arg, COMMAND_LINE_ERROR_MEMORY);
+		}
 		CommandLineSwitchCase(arg, "pcb")
 		{
 			if (!freerdp_settings_set_bool(settings, FreeRDP_SendPreconnectionPdu, TRUE))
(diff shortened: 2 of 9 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@thomasDOTwtf thank you for this addition.

* as for the `rdp file` option, do you have some docs/links at hand, a quick search returned empty on my end
* as for the `RDSTLS` related changes, https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/35910151-abed-463b-aaf0-dd95010c2dc6 still only lists version 1, is there some other documentation for that or was this just trial/error?

libfreerdp/core/connection.c
```diff
@@ -435,7 +435,15 @@ BOOL rdp_client_connect(rdpRdp* rdp)
 	nego_enable_tls(rdp->nego, settings->TlsSecurity);
 	nego_enable_nla(rdp->nego, settings->NlaSecurity);
 	nego_enable_ext(rdp->nego, settings->ExtSecurity);
-	nego_enable_rdstls(rdp->nego, settings->RdstlsSecurity);
+	/* An endpoint FedAuth token is delivered via RDSTLS. Enable the protocol
+	 * implicitly so the token can be used without an additional /sec flag. */
+	{
```
akallabeth, line 440: not sure this automatism is correct, should be set in command line parser if required. doing it here makes assumptions for all clients using this feature

thomasDOTwtf:
> @thomasDOTwtf thank you for this addition.
> 
> * as for the `rdp file` option, do you have some docs/links at hand, a quick search returned empty on my end
> * as for the `RDSTLS` related changes, https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/35910151-abed-463b-aaf0-dd95010c2dc6 still only lists version 1, is there some other documentation for that or was this just trial/error?

Unfortunately the only active implementation I know of is inside the Windows Azure Pack which generated these RDP files based on Service Provider Foundation for SCVMM. I observed the files generated to have the correct contents. For the RDSTLS it was trial/error to get the connection working. My main motivation was to get the possibility of a console connection without need for user/pass from within guacamole.

libfreerdp/core/connection.c
thomasDOTwtf: That's true. I'm going to move that part to be more clear about the implicit selection.

akallabeth:
@thomasDOTwtf ah, ok. Azure only related stuff is not published by MS :/

akallabeth:
@thomasDOTwtf btw, please `git clang-format` your changes.

akallabeth:
@thomasDOTwtf ok, did have a few minutes and did the cleanup myself.
See https://github.com/FreeRDP/FreeRDP/pull/13026 - taken your commit and squashed the cleanups in that one.

Thank you for your contribution, highly appreciated!

PR #13011, opened 2026-07-06
[client,common] /smartcard-logon pass PEM directly

akallabeth:
Allow base64 encoded PEM data to be passed as an alternative to file paths

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -965,17 +965,24 @@ fail:
 	return FALSE;
 }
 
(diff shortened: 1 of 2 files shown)


hardening:
Hum there's a world where a legitimate file path can be interpreted as correct base64...

akallabeth:
> Hum there's a world where a legitimate file path can be interpreted as correct base64...

good point, should have used `crypto_base64url_decode` to be sure.
fixed.

mfleisz (commented):
LGTM

mfleisz (approved):

PR #13034, opened 2026-07-10
[channels,rail] seal server receive stream to received PDU length

metsw24-max:
**Unsealed receive buffer in rail_server_handle_messages**

The server reads each PDU into a reused stream but never trims its length to the bytes actually received, so every order handler's length check measures the 4096-byte allocation rather than the PDU. A short `TS_RAIL_ORDER_EXEC` whose `ExeOrFileLength` exceeds the delivered body then makes `rail_read_exec_order` read past it into stale buffer contents before handing the result to `ClientExec`; the sibling `rdpei` and `ainput` servers already seal their receive streams, so this does the same after checking the body was fully delivered.

### DIFF
--- a/channels/rail/server/rail_main.c
+++ b/channels/rail/server/rail_main.c
@@ -1556,6 +1556,11 @@ UINT rail_server_handle_messages(RailServerContext* context)
 	RailServerPrivate* priv = context->priv;
 	wStream* s = priv->input_stream;
 
+	/* The input stream is reused between PDUs and its length still holds the
+	 * previous PDU's size, so restore it to the full buffer before reading. */
+	if (!Stream_SetLength(s, Stream_Capacity(s)))
+		return ERROR_INTERNAL_ERROR;
+
 	/* Read header */
 	if (!Stream_EnsureRemainingCapacity(s, RAIL_PDU_HEADER_LENGTH))
 	{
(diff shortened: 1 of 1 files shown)


channels/rail/server/rail_main.c
```diff
@@ -1556,6 +1556,11 @@ UINT rail_server_handle_messages(RailServerContext* context)
 	RailServerPrivate* priv = context->priv;
 	wStream* s = priv->input_stream;
 
+	/* The input stream is reused between PDUs and its length still holds the
+	 * previous PDU's size, so restore it to the full buffer before reading. */
+	if (!Stream_SetLength(s, Stream_Capacity(s)))
```
akallabeth, line 1561: from the description this should be `Stream_ResetPosition(s)

channels/rail/server/rail_main.c
```diff
@@ -1598,6 +1603,18 @@ UINT rail_server_handle_messages(RailServerContext* context)
 		return ERROR_INTERNAL_ERROR;
 	}
 
+	if (bytesReturned != (DWORD)(orderLength - RAIL_PDU_HEADER_LENGTH))
```
akallabeth, line 1606: this should return ok but not handle the message as the read was incomplete?

akallabeth (changes requested):
looks like you found an issue here:
1. The function should try to read enough data to extract the header
2. the function should try to read the expected data
3. if a short read occurs retry later
4. after handling a complete `PDU` reset the stream

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/13037

PR #13037, opened 2026-07-10
[channels,rail] rail_server_handle_messages

### LINKED ISSUE #13041: Android client does not invoke common auto-reconnect after event-handle failure
The Android client can enable FreeRDP automatic reconnection and negotiate the standard MS-RDPBCGR reconnect cookie, but its native event loop exits when `freerdp_check_event_handles()` fails. The current source still contains a TODO referring to the X11-specific `xf_auto_reconnect()` helper, so the common reconnect implementation is never called by the Android frontend.

This makes short network interruptions and Wi-Fi handovers terminate the Android session even when FreeRDP has enough protocol state to reclaim it in place. It also means application wrappers have to replace the entire native client context and lose the benefit of the server reconnect cookie.

The closest precedents appear to be:

- #4880, which unified reconnect helpers for other frontends but did not integrate Android
- #7168, which documents the need to keep frontend events and cancellation responsive during retry
- #10389 and #10390, which fixed SDL by invoking auto-reconnect at the `freerdp_check_event_handles()` failure boundary

A proposed Android implementation is available in draft PR #13040. It:

1. calls `client_auto_reconnect_ex()` at that same failure boundary;
2. keeps explicit disconnect responsive and discards input queued for the failed transport rather than replaying it against a potentially changed desktop state after recovery;
3. makes channel PubSub subscriptions idempotent because a successful reconnect runs `PreConnect` again; and
4. unsubscribes those handlers when the Android client context is freed.

The reconnect policy, retry count, delay, error filtering, and protocol-cookie handling remain in FreeRDP's common implementation.

akallabeth:
* Split function into subcomponents
* Ensure the expected data was read

### DIFF
--- a/channels/rail/server/rail_main.c
+++ b/channels/rail/server/rail_main.c
@@ -23,6 +23,7 @@
 
 #include <freerdp/freerdp.h>
 #include <freerdp/channels/log.h>
+#include <freerdp/utils/channel_pdu_tracker.h>
 
 #include <winpr/crt.h>
 #include <winpr/synch.h>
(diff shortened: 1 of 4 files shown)


hardening:
IMHO the real fix is to use a `ChannelPduTracker`

akallabeth:
> IMHO the real fix is to use a `ChannelPduTracker`

good point @hardening

akallabeth:
@hardening done.

Issue #10389, opened 2024-07-18
SDL client does not reconnect properly


eyalt:
**Describe the bug**
SDL client does not reconnect when encountering a network problem.

**To Reproduce**
1. Run SDL client with auto reconnect (`+auto-reconnect`)
2. Connect to a server
3. Run command `sudo ss --kill dport = 3389` (or other OS specific alternative)
4. Client exits and doesn't try to reconnect

**Expected behavior**
Client will attempt to reconnect (up to max retries), and will continue in case of a successful connection

**Application details**
* FreeRDP version (`xfreerdp /version`): `This is FreeRDP version 3.6.4-dev0 (fa0232048)`
* Command line used: `+auto-reconnect` (and connection credentials)
* OS version connecting to (server side): Windows 11
* If you built it yourself add some notes which tag/commit/branch you have used, also your cmake parameters and
  compiler can help - Latest commit `fa0232048c256d73e40bc4cc711fd70b1b238561`

**Environment (please complete the following information):**
 - OS: Linux
 - Version/Distribution: Ubuntu 23.04
 - Architecture: amd64

**Additional context**
Function `sdl_client_thread_proc` has a call to `client_auto_reconnect` but seems to be out of place.
Right now it only tries to reconnect when `WaitForMultipleObjects` fails, but the correct place should be in the case `freerdp_check_event_handles` fails (this is the flow in `xf_client_thread`, the X11 client), as this is the function that fails when a connection is closed (and other circumstances).

akallabeth:
@eyalt thx for the nice bug report.
hope the referenced pull request did fix your issue.

TolchiIsland:
Thank you for your quick respone.
I also noted a race in matrix chat (that I actually encountered myself), I don't believe this pull requests addresses it, am I correct? Is there an intention to fix that as well?

akallabeth:
@TolchiIsland ok, missed that (thought you just copied the same text ^^)
fixed that now as well.

Issue #7168, opened 2021-07-12
Session frozen when you lose network for a few seconds (Unrecoverable)


FutureKidBuu:
**freerdp version** 2.3.2

I have used parameters such as **+auto-reconnect** and **/timeout** without result, the session is frozen.

**Use case:** Unplug the network wire for 15-20 seconds and plug it in again.

The log when the session has frozen but you already have a network connection:

![image](https://user-images.githubusercontent.com/25820234/125269174-be7c6900-e308-11eb-922b-3d0daafc04b7.png)

akallabeth:
@FutureKidBuu 
1. does it recover after a few seconds?
2. The screenshot of the log is not very helpful ;)

FutureKidBuu:
@akallabeth No, the current session is no longer recoverable, I have to close the window and relaunch freerdp.

FutureKidBuu:
@akallabeth log, It can be hell ;) (due to xf_lock_x11_ methods
xf_unlock_x11_)
[trace_freerdp.log](https://github.com/FreeRDP/FreeRDP/files/6800509/trace_freerdp.log)

akallabeth:
@FutureKidBuu ok, the log is missing quite a lot of stuff (and your compile did ignore lots of compile warnings about excessive/dangerous logging)

akallabeth:
I´d try a `xfreerdp <yourargs> 2>&1 |tee log.txt` to get a consistent log with errors and warnings

2fly2:
client_auto_reconnect_ex 2nd prameter is NULL,  add windows message loop in window_events can fix it.

akallabeth:
@2fly2 which version and client do you use?

2fly2:
@akallabeth  freerdp version 3.0.0-dev,  freerdp git version c1084c7, branch master.  on windows.   but is the same as linux.
for(x=0; x < 50; x++)
{
    if(!ICALLRESULT(TRUE, window_event,  instance))
        return FALSE;
    Sleep(100);
}

window_event will not run because it is NULL.

akallabeth:
@2fly2 `xfreerdp` and `wlfreerdp` will call `handle_window_events` (wayland is just a stub though)
the windows client uses the version without callback though, there it might happen that the window is not updated until the session is reconnected

2fly2:
[...]

PR #13040, opened 2026-07-11
[client,android] add automatic reconnect

mhoev:
## What changed

- invoke `client_auto_reconnect_ex()` when the Android event loop detects an RDP handle failure
- keep explicit Android disconnect requests responsive during the delay between reconnect attempts

## Why

The Android client can enable FreeRDP automatic reconnection and negotiate the standard reconnect cookie, but its native event loop currently exits when `freerdp_check_event_handles()` fails. The common auto-reconnect implementation is therefore never called by the Android frontend.

This patch connects Android to the existing common mechanism at the same failure boundary used by other frontends. Retry policy, retry count, delay, error filtering, and protocol-cookie handling remain in FreeRDP's common implementation.

The callback passed to `client_auto_reconnect_ex()` only checks the context abort flag. FreeRDP invokes it during the delay between failed attempts, allowing the existing Android disconnect entry point to cancel promptly without adding Android event-queue policy to the reconnect path.

Related context: #4880, #7168, #10389, and #10390.

## Validation

- rebased on current FreeRDP `master`
- Android `:freeRDPCore:externalNativeBuildDebug` passed for `arm64-v8a` and `armeabi-v7a`
- downstream JNI build passed for `arm64-v8a` and `armeabi-v7a`
- formatting and `git diff --check` passed

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/cpp/android_event.c
+++ b/client/Android/Studio/freeRDPCore/src/main/cpp/android_event.c
@@ -195,6 +195,36 @@ BOOL android_check_handle(freerdp* inst)
 	return TRUE;
 }
 
+BOOL android_check_reconnect_handle(freerdp* inst)
+{
+	androidContext* aCtx;
+
+	if (!inst || !inst->context)
+		return FALSE;
+
+	aCtx = (androidContext*)inst->context;
+	if (!aCtx->event_queue || !aCtx->event_queue->isSet)
+		return FALSE;
+
+	if (WaitForSingleObject(aCtx->event_queue->isSet, 0) != WAIT_OBJECT_0)
+		return TRUE;
+	if (!ResetEvent(aCtx->event_queue->isSet))
+		return FALSE;
+
+	/* Input queued before reconnect belongs to the failed transport. Do not replay it after the
+	 * session is reclaimed, but keep an explicit disconnect responsive during retry delays. */
+	while (android_peek_event(aCtx->event_queue))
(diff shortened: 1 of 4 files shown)


akallabeth (changes requested):
this looks wrong.
* Channel handlers are already registered, why do it again?
* what is the handling during reconnect supposed to do?

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/13037

mhoev:
Thanks, you are right about the channel handlers. I had incorrectly assumed that reconnect reruns `PreConnect`; `freerdp_reconnect()` calls `rdp_client_reconnect()` directly, so the subscription bookkeeping and cleanup were unnecessary and have been removed.

The event callback was intended to keep an explicit Android disconnect responsive during the delay between failed reconnect attempts. I also had downstream input-queue policy there, which does not belong in this upstream integration. The callback now only checks `freerdp_shall_disconnect_context()`, matching the existing Android disconnect path's abort flag.

The draft is now reduced to one file: call `client_auto_reconnect_ex()` at the existing event-handle failure boundary and provide the cancellation-only callback. I rebuilt the current `master` Android project successfully for `arm64-v8a` and `armeabi-v7a`.

mhoev:
Thanks, I saw #13037 after posting the update. It supersedes this draft, so there is no need to reopen or review #13040 further. We are aligning the downstream integration with its `PostFinalDisconnect` subscription cleanup.

PR #13044, opened 2026-07-12
[codec,av1] bound decode output to decoded frame size

metsw24-max:
**Out-of-bounds read decoding AV1 frames smaller than the surface**

The colour conversion in `freerdp_av1_decompress` uses the destination surface size as its region, but the source pixels come from the decoded image planes, which aom sizes to the decoded frame. A server that sends a frame smaller than the surface makes the conversion read past those planes, so the converted region is now bounded to the decoded frame. Same shape as the h264 handling in f6380da.

Repro: a 32x32 frame decoded into a tall surface reads far past the Y plane and faults (SIGBUS); clean after the change.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13067, opened 2026-07-15
[channels,drdynvc] fix channel unref on create request send failure

metsw24-max:
**Bad channel unref on create-request send failure**

On a failed create-response send, `drdynvc_process_create_request` unrefs the channel returned by `dvcman_create_channel`, but it only owns that reference when creation actually succeeded. For a duplicate, unknown or rejected channel id the returned pointer is NULL, a live channel this call never referenced, or (on the `OnNewChannelConnection` failure branch, which unlike its sibling never cleared the pointer) an already-freed one, so the unref becomes a null dereference, a premature free of a channel still in use, or a double free, reachable from a server that sends a create request for such a channel id when the response write fails. Limited the rollback unref to the success path and cleared the freed pointer in that failure branch so the helper no longer hands back a dangling channel.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13078, opened 2026-07-16
[codec,h264] support more hardware decoders

### LINKED ISSUE #13069: [nvidia] VAAPI H.264 decode fails due to missing `NV12 -> YUV420` format conversion
# Environment
- FreeRDP 3.29.0 (also reproduces on current master, commit 743f437)
- Client: xfreerdp (X11) and xfreerdp3 (SDL3), Linux (CachyOS/Arch)
- GPU: NVIDIA RTX 4070, proprietary driver, VAAPI via `libva-nvidia-driver` (NVDEC backend)
- `vainfo` confirms healthy VAAPI stack: `VAProfileH264Main/High/ConstrainedBaseline : VAEntrypointVLD` all present
- Build flags: `WITH_VAAPI=ON WITH_FFMPEG=ON WITH_VIDEO_FFMPEG=ON WITH_OPENH264=OFF`

# Summary
Client-side hardware-accelerated H.264 decode via VAAPI fails unconditionally on every single frame once the server actually sends AVC420/AVC444 GFX data, producing either a black screen (region never gets painted) or `avc444_decompress failure: -1, ignoring update` spam, depending on codec mode. RFX Progressive/ClearCodec work fine; the bug is specific to the AVC420/444 + VAAPI hardware decode path.

# Reproduction
1. Connect with `/gfx:AVC444:on` (or `AVC420:on`) to a Windows RDP host that actually negotiates AVC (needs `AVC444ModePreferred` policy or equivalent server-side conditions — with default policy the server mostly stays on RFX Progressive and the bug never triggers, which is likely why this has gone unnoticed).
2. Play video or generate enough on-screen motion for the server to switch the affected region to AVC444/420.
3. Client log (`/log-level:INFO /log-filters:com.freerdp.codec:TRACE,com.freerdp.gdi:TRACE`) shows, for every frame:

```
[ERROR][com.freerdp.codec] - [libavcodec_decompress]: Failed to transfer video frame (status=-38) (Function not implemented)
[WARN][com.freerdp.codec] - [log_decompress]: H264 decompress failed with -1
[WARN][com.freerdp.gdi] - [gdi_SurfaceCommand_AVC444]: avc444_decompress failure: -1, ignoring update.
```

`-38` is `-ENOSYS`. Screen region stays black since `gdi_SurfaceCommand_AVC444` just drops the failed update.

# Root cause
In `libfreerdp/codec/h264_ffmpeg.c`:

- `libavcodec_decompress()` unconditionally sets `sys->videoFrame->format = AV_PIX_FMT_YUV420P;` (line ~333) before decoding, regardless of whether hardware (VAAPI) decode is active.
- When `sys->hwctx` is set, `avcodec_receive_frame()` fills `sys->hwVideoFrame` (an `AV_PIX_FMT_VAAPI` hw surface), then the code calls `av_hwframe_transfer_data(sys->videoFrame, sys->hwVideoFrame, 0)` to copy it to system memory — with `sys->videoFrame->format` still forced to `YUV420P` from the line above.
- But the hwframes context's `sw_format` was negotiated in `libavcodec_get_format()` as `AV_PIX_FMT_NV12` (or `AV_PIX_FMT_P010` for 10bpc) — never `YUV420P`. This isn't a driver quirk, it's how VAAPI decode hardware works: 4:2:0 decode surfaces are NV12 (2 planes, interleaved UV), full stop.
- `av_hwframe_transfer_data()` requires the destination format to be one of the surface's supported software formats. Since `YUV420P` isn't in that list, it fails with `AVERROR(ENOSYS)` on every call — this isn't intermittent or a race, it's 100% of frames, always.

Notably, the codebase already has NV12-aware plumbing for exactly this situation — `H264_CONTEXT::hwAccel` toggles 2-plane vs 3-plane buffer allocation and stride math in `h264.c` (`yuv_ensure_buffer`, lines ~58/74). But that flag is only ever set on the **VAAPI encoder** (server) path (`h264_ffmpeg.c`, guarded by `WITH_VAAPI_H264_ENCODING`), never on the client **decode** path. So even a correct NV12 transfer wouldn't help downstream today: `avc420_decompress`/`avc444_process_rects` → `yuv420_context_decode`/`yuv444_context_decode` in `yuv.c` unconditionally treat `pYUVData[0..2]` as three separate planar Y/U/V buffers and have no NV12 branch at all.

(For reference: this is a different bug from #11040, which was a chroma reverse-filter math error in the *software* AVC444v2 path, fixed for 3.11.0.)

# Fix
Rather than threading `hwAccel`/NV12 support all the way through `h264.c`/`yuv.c` (larger, riskier change touching the shared decode pipeline used by both hw and sw paths), the attached patch keeps the existing "downstream always gets YUV420P" contract and fixes it at the source: transfer the hw surface into an intermediate frame with format left unset (so FFmpeg negotiates the correct NV12/P010 sw_format itself), then `sws_scale()` that down to YUV420P/YUV420P10LE before handing frames to the rest of the pipeline. `libswscale` is already an FFmpeg component FreeRDP detects and links (`WITH_SWSCALE`), no new build dependency.

Scoped to the VAAPI path only (checked via `sys->hw_pix_fmt == AV_PIX_FMT_VAAPI` at runtime) — VideoToolbox's hwframe context already negotiates `sw_format = AV_PIX_FMT_YUV420P` directly (see the `#ifdef WITH_VIDEOTOOLBOX` branch in `libavcodec_get_format()`), so that path was never broken and is left untouched.

Patch attached (`vaapi_nv12_fix.patch`, applies to `libfreerdp/codec/h264_ffmpeg.c` against current master / 3.29.0).

# Verification
Built locally (`cmake -DWITH_VAAPI=ON ...`, gcc, Unix Makefiles) and tested against a  Windows 11 RDP host:

- **Before**: 0/41 AVC444 frames decoded successfully — 100% failure rate, black screen.
- **After**: 398/398 AVC444 frames decoded successfully, zero transfer errors, video renders correctly, confirmed visually.

# Patch
See attached `[vaapi_nv12_fix.patch](https://github.com/user-attachments/files/30070925/vaapi_nv12_fix.patch)`.

akallabeth:
For decoders that do not natively support AV_PIX_FMT_YUV420P add a conversion step so that the post processing pipeline can handle the data correctly.

### DIFF
--- a/libfreerdp/codec/h264_ffmpeg.c
+++ b/libfreerdp/codec/h264_ffmpeg.c
@@ -25,6 +25,7 @@
 #include <freerdp/codec/h264.h>
 #include <libavcodec/avcodec.h>
 #include <libavutil/opt.h>
+#include <libswscale/swscale.h>
 
 #include "h264.h"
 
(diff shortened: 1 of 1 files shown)


akallabeth:
@kurtdc6794 tested this on my hardware so far, but since I lack `nvidia` cards can only emulate the missing format support.

you should have a log line like
```
[14:37:33:380] [314985:0004ce83] [DEBUG][com.freerdp.codec] - [libavcodec_decompress]: formats: { src:AV_PIX_FMT_VAAPI, dst:AV_PIX_FMT_YUV420P } using swscale format conversion` in your `/log-level:debug
```
output, that means the format conversion path was taken.

this is what I get on my machines:
```
[14:40:03:249] [316279:0004d391] [DEBUG][com.freerdp.codec] - [libavcodec_decompress]: formats: { src:AV_PIX_FMT_VAAPI, dst:AV_PIX_FMT_YUV420P } using hardware supported copy
```

KurtDC6794:
@akallabeth  built #13078 and tested against the same live Windows RDP host (AVC444 forced via AVC444ModePreferred), NVIDIA/NVDEC via libva-nvidia-driver.

my log showed exactly what you expected:
```
 [DEBUG][com.freerdp.codec] - [libavcodec_decompress]: formats: { src:AV_PIX_FMT_VAAPI, dst:AV_PIX_FMT_YUV420P } using swscale format conversion
 ```

 Connection reached CONNECTION_STATE_ACTIVE and stayed there for the full session, zero decode errors/failures in the log (no decompress failure/ignoring update), video rendered correctly. Looks good from here.

PR #13092, opened 2026-07-20
[core,message] clone notify icon state order for async update

metsw24-max:
**Use-after-free in the async notify icon proxy**

`update_message_NotifyIconCreate`/`NotifyIconUpdate` shallow-copy the notify icon state, so the queued order keeps borrowed pointers to the tooltip, info tip and icon buffers that `update_recv_notification_icon_info_order` frees the moment the callback returns, and the update thread then reads freed memory. The window state and window icon proxies already deep-copy, so this clones the sub-buffers the same way and frees them in the queue teardown.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13104, opened 2026-07-24
[client,android] send clipboard format list response

### LINKED ISSUE #13097: Android client breaks host clipboard when connected
# **Describe the bug**

When aFreeRDP is connected, copy and paste stops working reliably on the host machine itself.

Pasting does nothing for several seconds. Repeated attempts queue up, then all paste at once. This lasts as long as the Android session is connected, and stops as soon as it disconnects.

It affects pasting between two ordinary applications on the host. You do not have to touch the tablet, or copy anything to or from it.

# **To Reproduce**

1. Connect aFreeRDP to a GNOME Remote Desktop host with clipboard redirection active.
2. On the host, copy some text in any application.
3. On the host, paste it into any other application.
4. Nothing happens. Try again a few times, then wait.
5. All the queued pastes land at once.

Disconnecting the Android session makes the host clipboard behave normally again immediately.

# **Expected behavior**

Having an Android client connected should not interfere with the host's clipboard.

Copy and paste on the host should stay responsive whether or not a session is connected.

# **Application details**

* FreeRDP version: aFreeRDP 3.26.0 (current F-Droid release, versionCode 3026000)
* Server side: GNOME Remote Desktop on Ubuntu 26.04, GNOME Shell 50.1, Wayland session, screen sharing in "extend" mode
* I use the tablet as an additional monitor, so it is a display surface only and runs no applications of its own.
* Connection configured through the aFreeRDP app UI.

# **Environment**

 - OS: Android (client) / Linux (server)
 - Version/Distribution: Android 16 on Samsung SM-X700 (Galaxy Tab S8); server Ubuntu 26.04
 - Architecture: arm64 (client), amd64 (server)

# **Additional context**

On the host, the RDP server logs the following once for each failed paste attempt:

```
[RDP.CLIPRDR] Possible protocol violation: Client did not send format list response (Timeout reached)
```

alongside, at matching timestamps:

```
gnome-shell: Could not fetch selection data: Remote selection transfer was cancelled
```

I am not a FreeRDP developer, so please treat the following as pointers rather than a diagnosis as I may well be misreading this.

Going by the server's complaint that the client "did not send format list response", I had a look at the Android clipboard code and it *appears* that the client may not send that acknowledgement. In `client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c`, `android_cliprdr_server_format_list()` seems to cache the server's formats and request format data, but I could not find a `ClientFormatListResponse` call on any path through it. Searching the whole of `client/Android/` for that symbol returned no matches at all.

For comparison, the other clients do appear to send it: 
`xf_cliprdr_send_client_format_list_response()` in `client/X11/xf_cliprdr.c`
`sdlClip::SendFormatListResponse()` in `client/SDL/SDL3/sdl_clip.cpp`.

I also could not see anywhere that the shared channel code would send it on the client's behalf: `channels/cliprdr/client/cliprdr_format.c` appears to call `context->ServerFormatList()`, log an error if it fails, then free and return.

I checked this against master on Jul 21 2026.

One other server-side error turned up in an earlier session on the same setup. I have no idea whether it is connected to any of the above, and I could not reproduce it on demand:

```
cliprdr_read_format_list: invalid length, got 2, require at least 4   (error 1359)
```

I am filing a separate report against gnome-remote-desktop as well, since it seems reasonable that an unresponsive client should not be able to affect the host's own clipboard, whatever the client is doing wrong.

svncibrahim:
The Android client cached the server's clipboard formats and requested data, but never
sent the Format List Response PDU that MS-RDPECLIP requires after every Format List PDU.

Servers that wait for it (e.g. GNOME Remote Desktop) block their own clipboard until the
request times out, so host-side copy/paste stalls for several seconds while an aFreeRDP
client is connected.

Send `ClientFormatListResponse` on both the empty-list and normal paths, mirroring the
X11 (`xf_cliprdr_send_client_format_list_response`) and SDL (`sdlClip::SendFormatListResponse`)
clients.

Fixes #13097

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c
+++ b/client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c
@@ -33,6 +33,17 @@
 #include "android_jni_utils.h"
 #include "android_jni_callback.h"
 
+static void android_cliprdr_free_formats(CLIPRDR_FORMAT* formats, UINT32 numFormats)
+{
+	if (!formats)
+		return;
+
+	for (UINT32 index = 0; index < numFormats; index++)
+		free(formats[index].formatName);
+
+	free(formats);
+}
+
 UINT android_cliprdr_send_client_format_list(CliprdrClientContext* cliprdr)
 {
 	UINT rc = ERROR_INTERNAL_ERROR;
@@ -85,7 +96,7 @@ UINT android_cliprdr_send_client_format_list(CliprdrClientContext* cliprdr)
 	rc = afc->cliprdr->ClientFormatList(afc->cliprdr, &formatList);
 fail:
 	free(pFormatIds);
-	free(formats);
+	android_cliprdr_free_formats(formats, numFormats);
 	return rc;
 }
 
@@ -196,6 +207,21 @@ static UINT android_cliprdr_server_capabilities(CliprdrClientContext* cliprdr,
 	return CHANNEL_RC_OK;
 }
 
+static UINT android_cliprdr_send_client_format_list_response(CliprdrClientContext* cliprdr,
+                                                             BOOL status)
+{
+	CLIPRDR_FORMAT_LIST_RESPONSE formatListResponse = WINPR_C_ARRAY_INIT;
+
+	formatListResponse.common.msgType = CB_FORMAT_LIST_RESPONSE;
+	formatListResponse.common.msgFlags = status ? CB_RESPONSE_OK : CB_RESPONSE_FAIL;
+	formatListResponse.common.dataLen = 0;
+
+	if (!cliprdr || !cliprdr->ClientFormatListResponse)
+		return ERROR_INVALID_PARAMETER;
+
+	return cliprdr->ClientFormatListResponse(cliprdr, &formatListResponse);
+}
+
 /**
  * Function description
  *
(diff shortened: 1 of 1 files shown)


client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c
```diff
@@ -85,7 +85,12 @@ UINT android_cliprdr_send_client_format_list(CliprdrClientContext* cliprdr)
 	rc = afc->cliprdr->ClientFormatList(afc->cliprdr, &formatList);
 fail:
 	free(pFormatIds);
-	free(formats);
+	if (formats)
+	{
+		for (UINT32 index = 0; index < numFormats; index++)
```
akallabeth: there should be a helper function for that.

akallabeth (approved):

akallabeth:
@svncibrahim nice fix.
FYI: limited internet for the following weeks, so I'll respond delayed and can not do all stuff I usually do here.

client/Android/Studio/freeRDPCore/src/main/cpp/android_cliprdr.c
svncibrahim: Added a local android_cliprdr_free_formats helper mirroring xf_cliprdr_free_formats, used in both spots.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #13119, opened 2026-07-29
[winpr,ntlm] require 16 byte signature buffer before access

metsw24-max:
**Unchecked 16 byte signature access in the NTLM package**

The four signature entry points only test the `SECBUFFER_TOKEN` for NULL, yet its size is peer controlled: SPNEGO's `mechListMIC` length reaches `mic.cbBuffer` verbatim and `negotiate_mic_exchange` forwards anything non-empty, so a one byte MIC reads 15 bytes past an exactly sized token allocation, and the outgoing MIC slot carved out of the leftover output buffer can be under 16 bytes as well, which makes the sign side a write. Kerberos already rejects an undersized buffer in the same four calls, so this mirrors those checks and their status codes; `TestSspiNTLM` now covers both directions.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #13121, opened 2026-07-30
[client,sdl] stop drawing when update queue is empty, fixes high CPU

cizra:
SdlWindow treats an empty rectangle list as a request to redraw the full surface. The SDL_EVENT_USER_UPDATE handler passed the empty result of SdlContext::pop to drawToWindows before terminating its loop.

Check for an empty result first to avoid unnecessary full-frame texture uploads, including those caused by stale update events after the queue has already been drained.

Before this patch, my CPU usage was 60% of a single core on a 3840x2400 resolution client, when idle. After the patch, it's 8%.

Assisted-by: gpt-5.6-sol

### DIFF
--- a/client/SDL/SDL3/sdl_context.cpp
+++ b/client/SDL/SDL3/sdl_context.cpp
@@ -1508,6 +1508,9 @@ bool SdlContext::useLocalScale() const
 
 bool SdlContext::drawToWindows(const std::vector<SDL_Rect>& rects)
 {
+	if (rects.empty())
+		return true;
+
 	for (auto& window : _windows)
 	{
 		if (!drawToWindow(window.second, rects))


akallabeth (changes requested):
change looks wrong.
* why filter at that level and let the empty rect handling down the line stay?
* ~~where do you get the empty rectangle updates from?~~

akallabeth:
@cizra reasoning sounds solid, just the location for the check might be better off in `SdlContext::drawToWindows`

cizra:
Empty rectangles come from https://github.com/FreeRDP/FreeRDP/blob/0b4f031f3575a42ddec441ffdb99cd8a4ef77047/client/SDL/SDL3/sdl_context.cpp#L1698 - they're the queue-is-empty sentinel.

However, https://github.com/FreeRDP/FreeRDP/blob/0b4f031f3575a42ddec441ffdb99cd8a4ef77047/client/SDL/SDL3/sdl_window.cpp#L326 interprets the empty rectangle in a different way - there, it's "redraw the whole mess".

I'll update the patch.

cizra:
Tested manually, works fine. CPU is low.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13123, opened 2026-07-30
Avoid reallocating MonitorDefArray in enforce_monitor_exists()

### LINKED ISSUE #13122: freerdp_settings_enforce_monitor_exists() silently reallocates MonitorDefArray, invalidating cached pointers
## Summary

`freerdp_settings_enforce_monitor_exists()` (called during `PreConnect` at `libfreerdp/core/freerdp.c:124`) may reallocate `MonitorDefArray` via `freerdp_settings_set_monitor_def_array_sorted()` → `freerdp_settings_set_pointer_len()` → `free(old)` + `calloc(new)`. Any pointer previously obtained via `freerdp_settings_get_pointer(settings, FreeRDP_MonitorDefArray)` becomes dangling.

Callers that cache this pointer and write to it after `PreConnect` will corrupt the heap. This caused a crash in Remmina 1.4.39 (heap buffer overflow writing to a reallocated 1-element array as if it had 2 elements).

## Affected versions

FreeRDP 3.15.0 (and likely all 3.x). Still present in master.

## Root cause

`freerdp_settings_enforce_monitor_exists()` at `libfreerdp/core/settings.c:1773`:

```c
if (!useMonitors || (count == 0))
{
    // ...
    if (!freerdp_settings_set_monitor_def_array_sorted(settings, &monitor, 1))
        return FALSE;
}
```

`freerdp_settings_set_monitor_def_array_sorted()` at `libfreerdp/common/settings.c:2440`:

```c
if (!freerdp_settings_set_pointer_len(settings, FreeRDP_MonitorDefArray, NULL, count))
    return FALSE;
```

Which calls `freerdp_settings_set_pointer_len_()`:

```c
void* old = freerdp_settings_get_pointer_writable(settings, id);
free(old);              // <-- frees the old array
// ...
copy = calloc(len, size);  // <-- allocates a new, possibly smaller array
```

The initial allocation in `freerdp_settings_new()` is 32 elements. After `enforce_monitor_exists` with `multimon=0`, the array is shrunk to 1 element. A caller holding the old 32-element pointer now writes to freed memory.

## Impact

Any FreeRDP API consumer that:
1. Gets `MonitorDefArray` pointer before `PreConnect`
2. Writes to it after `PreConnect`

...will corrupt the heap. Remmina 1.4.39 does exactly this, causing SEGV/ABRT.

## Evidence

Kernel log showing the fault address matches monitor physical width:
```
segfault at 25400000001   (0x254 = 596 = physicalWidth of 2nd monitor, 0x01 = orig_screen)
segfault at 25400000019   (same base, different field offset)
```

Backtrace (ABRT from heap corruption detected by `free()`):
```
#7  __libc_free (libc.so.6)
#8  libfreerdp3.so.3 + 0x5ee40
#9  libfreerdp3.so.3 + 0x9e23f
#10 freerdp_settings_free
#11 libfreerdp3.so.3 + 0xd5810
#12 freerdp_context_free
#13 remmina-plugin-rdp.so + 0xd601
```

## Does NOT reproduce with xfreerdp3

```
xfreerdp3 /v:<server> /monitors:0,1   # works fine
```

FreeRDP's own `/monitors:` path correctly handles the array. The bug was exclusively in Remmina's direct-write pattern (now fixed in Remmina master).

## Suggested improvements

1. **Document** that `freerdp_settings_enforce_monitor_exists()` may invalidate `MonitorDefArray` pointers. Add a comment in the header and implementation.

2. **Add a debug warning** when the array is reallocated to a smaller size:
```c
if (count < freerdp_settings_get_uint32(settings, FreeRDP_MonitorDefArraySize))
    WLog_WARN(TAG, "MonitorDefArray shrunk from %u to %zu elements, cached pointers are invalid",
              freerdp_settings_get_uint32(settings, FreeRDP_MonitorDefArraySize), count);
```

3. **Consider** not shrinking the array (keep the original 32-element allocation, just update `MonitorCount`). This would avoid invalidating pointers entirely. The trade-off is slightly higher memory usage (32 × sizeof(rdpMonitor) ≈ 1KB), which is negligible.

Option 3 is the safest fix — it eliminates the entire class of "cached pointer invalidated by reallocation" bugs:

```c
// In freerdp_settings_enforce_monitor_exists(), instead of:
if (!freerdp_settings_set_monitor_def_array_sorted(settings, &monitor, 1))
    return FALSE;

// Use direct array write (array is already large enough from init):
rdpMonitor* arr = freerdp_settings_get_pointer_writable(settings, FreeRDP_MonitorDefArray);
arr[0] = monitor;
freerdp_settings_set_uint32(settings, FreeRDP_MonitorCount, 1);
```

rotanid:
## What

`freerdp_settings_enforce_monitor_exists()` called `set_monitor_def_array_sorted()` which reallocates `MonitorDefArray` via `free()`+`calloc()`. This silently invalidated any pointer previously obtained via `freerdp_settings_get_pointer()`, causing heap corruption in callers that cached the pointer.

## Root cause

See #13122 for full analysis.

## Fix

When the array already has room (always true after initialization, which allocates 32 elements), write directly to `arr[0]` instead, replicating the exact same field assignments (`x=0`, `y=0`, `is_primary=TRUE`, `MonitorLocalShiftX/Y`). Falls back to `set_monitor_def_array_sorted()` only if the array is empty (defensive edge case).

## Safety

- The direct-write path produces **identical results** to `set_monitor_def_array_sorted()` for a single primary monitor at (0,0): same struct fields, same shift values, same `MonitorCount`
- The fallback path preserves the old behavior for the impossible-but-defensive case where `MonitorDefArraySize == 0`
- No API change, no struct change, no behavioral change for callers that don't cache pointers

## Testing

Tested on Debian 13 trixie with FreeRDP 3.15.0. Remmina 1.4.39 (which caches the pointer and writes to it) no longer crashes.

Closes #13122

### DIFF
--- a/libfreerdp/core/settings.c
+++ b/libfreerdp/core/settings.c
@@ -1812,8 +1812,37 @@ BOOL freerdp_settings_enforce_monitor_exists(rdpSettings* settings)
 		monitor.attributes.orientation = orientation;
 		monitor.attributes.desktopScaleFactor = desktopScaleFactor;
 		monitor.attributes.deviceScaleFactor = deviceScaleFactor;
-		if (!freerdp_settings_set_monitor_def_array_sorted(settings, &monitor, 1))
-			return FALSE;
+
+		/*
+		 * If the MonitorDefArray already has room for at least 1 entry,
+		 * write directly instead of calling set_monitor_def_array_sorted().
+		 * The latter reallocates the array (free + calloc), which invalidates
+		 * any pointer previously obtained via freerdp_settings_get_pointer().
+		 * See https://github.com/FreeRDP/FreeRDP/issues/13122
+		 */
+		const UINT32 arraySize =
+		    freerdp_settings_get_uint32(settings, FreeRDP_MonitorDefArraySize);
+		if (arraySize >= 1)
+		{
+			rdpMonitor* arr =
+			    freerdp_settings_get_pointer_writable(settings, FreeRDP_MonitorDefArray);
+			WINPR_ASSERT(arr);
+			arr[0] = monitor;
+			arr[0].x = 0;
+			arr[0].y = 0;
+			arr[0].is_primary = TRUE;
+			if (!freerdp_settings_set_int32(settings, FreeRDP_MonitorLocalShiftX,
+			                                 monitor.x))
+				return FALSE;
+			if (!freerdp_settings_set_int32(settings, FreeRDP_MonitorLocalShiftY,
+			                                 monitor.y))
+				return FALSE;
+		}
+		else
+		{
+			if (!freerdp_settings_set_monitor_def_array_sorted(settings, &monitor, 1))
+				return FALSE;
+		}
 	}
 
 	return TRUE;


akallabeth (changes requested):
working around a usage issue.

PR #13127, opened 2026-07-31
Document that enforce_monitor_exists() may reallocate MonitorDefArray

rotanid:
## What

Adds documentation warnings to `freerdp_settings_enforce_monitor_exists()` in both the header declaration (`settings.h`) and the implementation (`settings.c`).

## Why

As discussed in #13122, this function may reallocate `MonitorDefArray` via `freerdp_settings_set_monitor_def_array_sorted()`. Pointers obtained via `freerdp_settings_get_pointer(settings, FreeRDP_MonitorDefArray)` before this call are invalidated. This caused a heap-corruption crash in Remmina 1.4.39 (which cached the pointer and wrote past the reallocated boundary).

The maintainer's position (#13122) is that callers should only rely on `rdpContext` being stable — not on settings-internal pointers. This PR documents that contract so future consumers don't repeat the mistake.

## Changes

- `libfreerdp/core/settings.h`: Added Doxygen comment on `freerdp_settings_enforce_monitor_exists()` declaring that `MonitorDefArray` pointers may be invalidated
- `libfreerdp/core/settings.c`: Matching comment on the implementation

No code changes, no behavioral changes.

Refs #13122

### DIFF
--- a/include/freerdp/settings.h
+++ b/include/freerdp/settings.h
@@ -69,6 +69,16 @@ extern "C"
 	 * is especially important for the proxy where you have a RDP client and RDP server in the same
 	 * application context)
 	 *
+	 * \warning Pointers obtained from settings getters (e.g.
+	 * freerdp_settings_get_pointer(), freerdp_settings_get_string(), ...) are not
+	 * guaranteed to remain valid after any operation that modifies the settings.
+	 * Internal functions such as freerdp_settings_set_pointer_len() and
+	 * freerdp_settings_enforce_monitor_exists() may reallocate the underlying
+	 * buffers, and rdp_reset_runtime_settings() may even replace the entire
+	 * rdpSettings object. Callers must re-fetch pointers after such operations
+	 * rather than caching them. Only the rdpContext pointer is guaranteed stable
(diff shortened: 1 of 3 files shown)


akallabeth (changes requested):
this function is internal, so not public API.

the reallocation is true for the whole `rdpSettings` struct, see https://github.com/FreeRDP/FreeRDP/blob/9415f2d11e4cbc4e25d3d9fd0c4271e2e05d5c58/libfreerdp/core/rdp.c#L2813 and https://github.com/FreeRDP/FreeRDP/blob/9415f2d11e4cbc4e25d3d9fd0c4271e2e05d5c58/libfreerdp/core/rdp.c#L2823

rotanid (commented):
Good point — moved the docs to the public header instead. The rdpSettings group doc now has a general warning about pointer instability, and I added shorter notes on get_pointer() and get_pointer_writable(). Dropped the internal function comments.

include/freerdp/settings.h
akallabeth, line 60: I think this is the documentation you're looking for ;)

PR #13129, opened 2026-07-31
[urbdrc] Fix TRANSFER_OUT completion framing

### LINKED ISSUE #13112: urbdrc: bulk TransferOut completion is selected from OutputBufferSize instead of transfer direction
I noticed that FreeRDP selects `URB_COMPLETION` versus `URB_COMPLETION_NO_DATA` only from `OutputBufferSize`.

https://github.com/FreeRDP/FreeRDP/blob/0b4f031f3575a42ddec441ffdb99cd8a4ef77047/channels/urbdrc/client/data_transfer.c#L860

But that should only apply to TransferIn. Per [MS-RDPEUSB 3.3.5.3.7](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/d1f22b46-0d7a-4d74-86e4-6bd8a06b702f):

> The TRANSFER_OUT_REQUEST message contains the OutputBufferSize field. This indicates the amount of data the server is sending to the device. When the client sends URB_COMPLETION_NO_DATA message to the server to report the final result of the TRANSFER_OUT_REQUEST, the OutputBufferSize value MUST NOT be greater than the OutputBufferSize value in TRANSFER_OUT_REQUEST message.

My understanding is that TransferOut should instead return:

- `URB_COMPLETION_NO_DATA`
- `OutputBufferSize = actual_length`
- no trailing output buffer

Only a TransferIn with non-zero returned data should use `URB_COMPLETION`.

Interestingly, FreeRDP used this direction-aware condition before the asynchronous transfer rewrite:

https://github.com/FreeRDP/FreeRDP/blob/15ad496bbad8037300501ccb6dfc14aa582b5083/channels/urbdrc/client/data_transfer.c#L786-L789

The current behavior causes my RDPEUSB server to receive an unexpected payload after a bulk OUT transfer, followed by an RDP transport disconnect.

Is the current behavior intentional for compatibility with a Windows URBDRC implementation, or was the transfer direction accidentally lost during the asynchronous rewrite?

uchouT:
## Description

FreeRDP selected between `URB_COMPLETION` and `URB_COMPLETION_NO_DATA` based only on `OutputBufferSize`.

For a `TRANSFER_OUT_REQUEST`, this caused the input buffer sent to the USB device to be serialized again as trailing completion data.

According to MS-RDPEUSB, a TransferOut result should use `URB_COMPLETION_NO_DATA`. Its `OutputBufferSize` reports the transferred byte count, but no trailing `OutputBuffer` is present.

This change preserves the outer transfer direction through asynchronous callbacks and selects the completion framing as follows:

- TransferIn with returned data: `URB_COMPLETION`
- TransferIn without returned data: `URB_COMPLETION_NO_DATA`
- TransferOut: `URB_COMPLETION_NO_DATA`, while retaining the transferred
  byte count in `OutputBufferSize`

See:
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpeusb/d1f22b46-0d7a-4d74-86e4-6bd8a06b702f

## Manual testing

I connected FreeRDP to a Windows Server 2022 and redirected a USB mass-storage device (a USB flash drive).

For the same 31-byte bulk TransferOut request:

```text
[READ ] TRANSFER_OUT_REQUEST ... length=67
urb_bulk_or_interrupt_transfer: ep:0x1 transfer_type 2 ... OutputBufferSize:0x1f
```

Before this change:

```text
[WRITE] URB_COMPLETION ... FunctionId=00000101, length=67
```

After this change:

```text
[WRITE] URB_COMPLETION_NO_DATA ... FunctionId=00000102, length=36
```

`0x1f` is 31 bytes, and the previous 67-byte completion consisted of the 36-byte result plus those 31 bytes as trailing data. The new completion contains only the 36-byte result.

A TransferIn returning one byte continued to use the expected framing in both versions:

```text
[WRITE] URB_COMPLETION ... FunctionId=00000101, length=37
```

No problems were observed with the redirected USB flash drive during this test. No other types of USB devices have been tested.

Full sanitized logs:

[freerdp-before.log](https://github.com/user-attachments/files/30602376/freerdp-before.log)
[freerdp-after.log](https://github.com/user-attachments/files/30602380/freerdp-after.log)

Timestamps log noise were removed.

Fixes #13112

### DIFF
--- a/channels/urbdrc/client/data_transfer.c
+++ b/channels/urbdrc/client/data_transfer.c
@@ -97,15 +97,25 @@ fail:
 	return status;
 }
 
+/* [MS-RDPEUSB] 2.2.7.2 and 2.2.7.3:
+ * Only a TRANSFER_IN_REQUEST that returns data carries an OutputBuffer.
+ * TRANSFER_OUT_REQUEST reports the transferred byte count in OutputBufferSize,
+ * but always uses URB_COMPLETION_NO_DATA. */
+static UINT32 urb_completion_payload_size(int transferDir, UINT32 outputBufferSize)
+{
+	return (transferDir == USBD_TRANSFER_DIRECTION_IN) ? outputBufferSize : 0;
+}
+
 static UINT urb_write_completion(WINPR_ATTR_UNUSED IUDEVICE* pdev,
                                  GENERIC_CHANNEL_CALLBACK* callback, BOOL noAck, wStream* out,
                                  UINT32 InterfaceId, UINT32 MessageId, UINT32 RequestId,
-                                 UINT32 usbd_status, UINT32 OutputBufferSize)
+                                 UINT32 usbd_status, UINT32 OutputBufferSize, int transferDir)
 {
 	if (!out)
 		return ERROR_INVALID_PARAMETER;
 
-	if (Stream_Capacity(out) < OutputBufferSize + 36)
+	const UINT32 payloadSize = urb_completion_payload_size(transferDir, OutputBufferSize);
+	if (Stream_Capacity(out) < payloadSize + 36ULL)
 	{
 		Stream_Free(out, TRUE);
 		return ERROR_INVALID_PARAMETER;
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

uchouT:
formatted @akallabeth please re-launch test

akallabeth:
@freerdp-bot test

PR #13132, opened 2026-08-02
[client,sdl] do not alpha blend the remote framebuffer

### LINKED ISSUE #13131: [SDL3] Missing SDL_SetTextureBlendMode causes stripes at /bpp:24 and a black screen at /bpp:32; X11 client unaffected
# **Describe the bug**

The SDL3 client creates its textures in a format **with an alpha channel** (`SDL_PIXELFORMAT_BGRA32`) but never calls `SDL_SetTextureBlendMode()`. SDL3 defaults textures that have an alpha channel to `SDL_BLENDMODE_BLEND`, so the contents of the GDI buffer's fourth byte are interpreted as **opacity**.

That fourth byte is not opacity. It is whatever the colour conversion left behind, and what it contains depends on the negotiated colour depth:

| `/bpp` | alpha byte after conversion | result on screen |
| --- | --- | --- |
| 16 | always filled to opaque | correct |
| 24 | partially left at 0 | **horizontal stripes / torn image** |
| 32 | top byte ends up zero | **fully black window** |

Two blend stages are affected, and both are wrong:

1. `_gdiTexture` is blended onto `_renderTarget` (the persistent dirty-rect accumulator),
2. `_renderTarget` is then blended onto the backbuffer.

**The X11 client is not affected**, which is what makes this look like a server or protocol problem at first. `xf_get_local_color_format()` in `client/X11/xf_graphics.c` picks `PIXEL_FORMAT_BGRX32` for depth 24 — same memory layout, but **X instead of A**, so no alpha channel and no blending. The SDL3 client hard-codes `PIXEL_FORMAT_BGRA32` in `sdl_context.cpp` regardless of depth.

There is a call to `SDL_SetSurfaceBlendMode(_primary.get(), SDL_BLENDMODE_NONE)` in `sdl_context.cpp`, but that applies to the *surface*, not to the textures created from it, so it has no effect here.

# **To Reproduce**

1. Connect to an xrdp server with the SDL3 client:
   `sdl-freerdp /v:<host> /u:<user> /bpp:24 /cert:ignore`
2. The image shows horizontal stripes / tearing.
3. Repeat with `/bpp:32` → the window is entirely black (input still works, so the session itself is fine).
4. Repeat with `/bpp:16` → correct.
5. Repeat all of the above with `xfreerdp` against the same server → **all correct**, including 24 and 32.

# **Expected behavior**

The remote framebuffer should be copied to the screen unmodified. The GDI buffer's fourth byte is padding, not opacity, and must not be used for blending.

# **Application details**

* FreeRDP version: **3.30.0** (Homebrew, unmodified). Also reproduced with **3.26.0**.
* Affected file is **byte-identical in 3.26.0, 3.30.0 and current `master`**, down to the same line numbers.
* Command line: `sdl-freerdp /v:<host> /u:<user> /p:<pw> /bpp:24 /w:1600 /h:1000 /cert:ignore`
* Server: **xrdp 0.9.24** on Linux Mint 22.3, Xorg backend, `max_bpp` set to the depth under test.
* SDL 3.4.12.

<details><summary>/buildconfig</summary>

```
This is FreeRDP version [/opt/homebrew/bin/sdl-freerdp] 3.30.0 (n/a)
Build type:          Release
CFLAGS:               -fvisibility=hidden -fno-omit-frame-pointer -Wredundant-decls -fsigned-char -Wimplicit-function-declaration -Wno-jump-misses-init -Wno-c++-keyword -fvisibility=hidden -O3 -DNDEBUG
Compiler:            AppleClang, 21.0.0.21000099
Target architecture: ARM
WITH_CLIENT_SDL=ON WITH_CLIENT_SDL3=ON WITH_CLIENT_SDL2=OFF WITH_SDL_LINK_SHARED=ON
WITH_X11=ON WITH_SWSCALE=ON WITH_FFMPEG=ON WITH_GFX_H264=ON WITH_OPENSSL=ON
```

</details>

# **Environment**

- OS: macOS 26.6 (25G72)
- Architecture: arm64 (Apple Silicon)
- Both `sdl-freerdp` and `xfreerdp` from the same Homebrew `freerdp 3.30.0` build, so **frontend is the only variable**.

# **Additional context**

## Evidence

Because both frontends come from the same package build, version and build options can be ruled out. Same server, same session, same command line apart from the binary:

| Frontend / variant (all 3.30.0, `/bpp:24`) | Result |
| --- | --- |
| `xfreerdp` (X11 via XQuartz) | correct |
| `sdl-freerdp`, default | stripes |
| `sdl-freerdp /gdi:sw` | black |
| `sdl-freerdp /bpp:16` | correct |
| `sdl-freerdp` with blend mode forced off at runtime | **correct** |

The last row is the actual proof rather than an inference. I interposed `SDL_CreateTexture` via `DYLD_INSERT_LIBRARIES` and called `SDL_SetTextureBlendMode(t, SDL_BLENDMODE_NONE)` on every texture the client creates, changing nothing else. The stripes disappeared immediately at `/bpp:24`.

## Affected code (`master`)

`client/SDL/SDL3/sdl_window.cpp`, `SdlWindow::ensureRenderTarget()`, line 286:

```cpp
_renderTarget =
    SDL_CreateTexture(_renderer, SDL_PIXELFORMAT_BGRA32, SDL_TEXTUREACCESS_TARGET, w, h);
if (!_renderTarget)
    SDL_LogError(SDL_LOG_CATEGORY_RENDER, "SDL_CreateTexture (render target): %s",
                 SDL_GetError());
```

`client/SDL/SDL3/sdl_window.cpp`, `SdlWindow::blit()`, line 502:

```cpp
_gdiTexture = SDL_CreateTexture(_renderer, surface->format, SDL_TEXTUREACCESS_STREAMING,
                                surface->w, surface->h);
```

`surface->format` is `BGRA32` here, because `sdl_context.cpp:285` does `gdi_init(instance, PIXEL_FORMAT_BGRA32)` unconditionally.

## Suggested fix

Two lines. Verified against a real xrdp server at 24 bpp:

```diff
--- a/client/SDL/SDL3/sdl_window.cpp
+++ b/client/SDL/SDL3/sdl_window.cpp
@@ -286,6 +286,11 @@
 	_renderTarget =
 	    SDL_CreateTexture(_renderer, SDL_PIXELFORMAT_BGRA32, SDL_TEXTUREACCESS_TARGET, w, h);
 	if (!_renderTarget)
 		SDL_LogError(SDL_LOG_CATEGORY_RENDER, "SDL_CreateTexture (render target): %s",
 		             SDL_GetError());
+	else
+		/* The GDI buffer's fourth byte is padding, not opacity. SDL3 defaults
+		 * textures with an alpha channel to SDL_BLENDMODE_BLEND, which would
+		 * blend that padding against the backbuffer. */
+		SDL_SetTextureBlendMode(_renderTarget, SDL_BLENDMODE_NONE);
 }
@@ -502,6 +507,10 @@
 		_gdiTexture = SDL_CreateTexture(_renderer, surface->format, SDL_TEXTUREACCESS_STREAMING,
 		                                surface->w, surface->h);
 		if (!_gdiTexture)
 		{
 			SDL_LogError(SDL_LOG_CATEGORY_RENDER, "SDL_CreateTexture: %s", SDL_GetError());
 			return false;
 		}
+		/* Same reasoning as in ensureRenderTarget(): copy, do not blend. */
+		SDL_SetTextureBlendMode(_gdiTexture, SDL_BLENDMODE_NONE);
 		_gdiTextureW = surface->w;
 		_gdiTextureH = surface->h;
 	}
```

An alternative — arguably better — fix would be to mirror the X11 client and choose a format without an alpha channel (`PIXEL_FORMAT_BGRX32`) in `sdl_context.cpp`, but that touches the whole GDI path, whereas setting the blend mode is local and matches what the surface-level code already tries to do.

I am happy to open a pull request if that is welcome.

nojan01:
Fixes #13131.

## What it does

The SDL3 client creates both of its textures in a format **with an alpha channel** — `sdl_context.cpp` calls `gdi_init(instance, PIXEL_FORMAT_BGRA32)` unconditionally — but never calls `SDL_SetTextureBlendMode()`. SDL3 defaults textures that have an alpha channel to `SDL_BLENDMODE_BLEND`, so the fourth byte of the GDI buffer is used as opacity.

That byte is padding, not opacity:

| `/bpp` | alpha byte after conversion | result |
| --- | --- | --- |
| 16 | always opaque | correct |
| 24 | partly zero | horizontal stripes |
| 32 | zero | black window |

This patch sets `SDL_BLENDMODE_NONE` on both textures, so the remote framebuffer is copied rather than blended.

## Why the X11 client is not affected

`xf_get_local_color_format()` picks `PIXEL_FORMAT_BGRX32` for depth 24 — same memory layout, **X instead of A**, hence no alpha channel and no blending. The SDL3 path hard-codes `BGRA32` regardless of depth.

`sdl_context.cpp` already calls `SDL_SetSurfaceBlendMode(_primary.get(), SDL_BLENDMODE_NONE)`, which shows the intent, but that applies to the *surface* and not to the textures created from it.

## How to test

Against an xrdp server (I used xrdp 0.9.24 on Linux Mint 22.3, Xorg backend):

```
sdl-freerdp /v:<host> /u:<user> /bpp:24 /cert:ignore
```

Before: horizontal stripes. With `/bpp:32`: black window, input still working.
After: correct at 16, 24 and 32.

## Evidence

Before writing the patch I verified the cause rather than inferring it. Homebrew ships `xfreerdp` and `sdl-freerdp` from the same 3.30.0 build, so the frontend is the only variable — same server, same session, same command line:

| Variant (3.30.0, `/bpp:24`) | Result |
| --- | --- |
| `xfreerdp` (X11) | correct |
| `sdl-freerdp`, default | stripes |
| `sdl-freerdp /gdi:sw` | black |
| `sdl-freerdp /bpp:16` | correct |
| `sdl-freerdp`, blend mode forced off at runtime | **correct** |

For the last row I interposed `SDL_CreateTexture` via `DYLD_INSERT_LIBRARIES` and called `SDL_SetTextureBlendMode(t, SDL_BLENDMODE_NONE)` on every texture, changing nothing else. The stripes disappeared immediately.

The affected code is byte-identical in 3.26.0, 3.30.0 and current `master`. I have been running this patch on 3.26.0 in production for a day at 24 bpp without regressions.

## Alternative considered

Mirroring the X11 client by choosing `PIXEL_FORMAT_BGRX32` in `sdl_context.cpp` would arguably be the more principled fix, but it touches the whole GDI path. Setting the blend mode is local and matches what the surface-level code already tries to do. Happy to switch if you prefer that direction.

## Checklist

- `clang-format --style=file` reports no changes on the modified file
- No API or behaviour change beyond the blend mode

### DIFF
--- a/client/SDL/SDL3/sdl_window.cpp
+++ b/client/SDL/SDL3/sdl_window.cpp
@@ -288,6 +288,13 @@ void SdlWindow::ensureRenderTarget()
 	if (!_renderTarget)
 		SDL_LogError(SDL_LOG_CATEGORY_RENDER, "SDL_CreateTexture (render target): %s",
 		             SDL_GetError());
+	else
+	{
+		/* The fourth byte of the GDI buffer is padding, not opacity. SDL3 defaults
+		 * textures that have an alpha channel to SDL_BLENDMODE_BLEND, which would
+		 * blend that padding against the backbuffer. */
+		SDL_SetTextureBlendMode(_renderTarget, SDL_BLENDMODE_NONE);
+	}
 }
 
 bool SdlWindow::drawRect(SDL_Surface* surface, SDL_Point offset, const SDL_Rect& srcRect)
@@ -506,6 +513,11 @@ bool SdlWindow::blit(SDL_Surface* surface, const SDL_Rect& srcRect, SDL_Rect& ds
 			SDL_LogError(SDL_LOG_CATEGORY_RENDER, "SDL_CreateTexture: %s", SDL_GetError());
 			return false;
 		}
+
+		/* Same reasoning as in ensureRenderTarget(): copy the remote framebuffer,
+		 * do not blend it. */
+		SDL_SetTextureBlendMode(_gdiTexture, SDL_BLENDMODE_NONE);
+
 		_gdiTextureW = surface->w;
 		_gdiTextureH = surface->h;
 	}


akallabeth (changes requested):
breaks RAILS

this sounds like some bug happening long before (the alpha channel should be correct in any case)

PR #13133, opened 2026-08-03
[client,sdl] RemoteApp (RAIL) support for the SDL3 client

### LINKED ISSUE #9078: [SDL] rails support
support remote application mode

svncibrahim:
## Summary

Adds RemoteApp (RAIL) support to the SDL3 client on X11 and Wayland. Each RemoteApp window becomes its own borderless SDL window.

This is experimental. A one-time warning is logged when RAIL mode starts.

The Wayland backend is the more experimental of the two. On a Wayland session, running with `SDL_VIDEODRIVER=x11` (XWayland) is the recommended way to use RemoteApp today. 

## Common to both backends

Each window is a borderless SDL window with the server-drawn frame as its content.

Window lifecycle, z-order, icons, min/max limits, work-area reporting and activation are wired up as usual. Specific to this client:

- Popups, menus and tooltips are SDL popup windows parented to their owner. No taskbar slot, no focus steal, positionable on Wayland.
- Local move and resize follow the X11 client's model. Input is suppressed while the WM drags, server geometry is ignored during the drag, and on release the client sends a synthetic button-up plus a `ClientWindowMove` with the final rect.
- Damage-driven paint. GFX pixels are deep-copied each frame, since the channel thread can free the source during a resize. Popups and layered windows keep real alpha, other surfaces are forced opaque.
- The desktop mirror is not output-suppressed while RAIL is on. Otherwise the server stops sending updates and input freezes.
- Menu drop shadows are drawn only when anchored to a visible popup. A window's own frame-edge shadow has no popup to anchor to, and would lag the window, so it is left out.

The resize border differs per backend.

## X11 backend

X11 needs WM protocols SDL does not expose, so the client uses libX11 directly. Without libX11 these become no-op stubs, and only move/resize is unavailable.

- Move and resize use `_NET_WM_MOVERESIZE`.
- Resize margin: the window is grown over the margin and its real bounds are published with `_GTK_FRAME_EXTENTS`. The margin takes input, and snap, tile and maximize still match the visible part.
- Opaque resizes stay in step with server frames via `_NET_WM_SYNC_REQUEST`.
- A compositor check gates transparent windows, so they are used only where they actually blend.

## Wayland backend

Wayland does not let a client place or freely resize its own toplevel, and several RAIL behaviours have no protocol at all. This backend works, but it is a reduced version of the X11 one.

- Render and input are server-relative, so the local position does not matter. Reported coordinates are clamped to the desktop.
- Interactive resize is server-driven. Grabbing the window's own resize border is forwarded to the server, which replies with a local move/size, and the client hands the resize to the compositor via `xdg_toplevel_resize`. The grab area is the visible border, so it is thinner than the wide invisible margin X11 offers.
- Move and resize use `xdg_toplevel.move` and `xdg_toplevel.resize`.
- Popups are xdg_popups and need a parent. A popup that arrives before its owner is deferred and retried.
- A full-display `WS_POPUP` (a slideshow) cannot be an xdg_popup, so it is realized as an SDL fullscreen window.

Not available on Wayland, for lack of protocol:

- Always on top. Topmost windows are ordinary windows.
- Z-order restacking. There is no focus-neutral restack path.
- Client-set window positions. The local layout follows the compositor.

A wider, X11-style grab margin needs SDL to expose window border insets, which it does not yet. I have opened an issue and a PR upstream for opt-in border-inset properties and will track it there. Once that lands, and FreeRDP's minimum SDL includes it, the Wayland backend can use the insets with `SDL_HitTest` and the two backends converge.

## Known issues and limitations

- Wayland lacks always on top, restacking and client positioning, as above. Prefer `SDL_VIDEODRIVER=x11` on a Wayland session.
- Multi-monitor and HiDPI scaling are untested.
- Windows and macOS are untested. The native move/resize helpers are stubbed there anyway.
- No tray icon support yet.
- ServerSystemParam is not wired.

## Testing

Servers: Windows 10 and Windows 11 RemoteApp hosts.

Clients: GNOME 50 (Mutter) has no native X11 session, so there I tested XWayland and native Wayland. For native X11 I used KDE and XFCE. Most testing was on GNOME, with a few Mutter-specific quirks.

My coverage is not wide, so there are almost certainly bugs I have not caught. This touches a lot of window manager and compositor behaviour, so testing on other WMs, compositors, servers or multi-monitor setups would help a lot.

All help, testing and feedback is appreciated.


svncibrahim:
A few areas worth exercising if anyone gets a chance to test this. It is not exhaustive and some of it overlaps, but I am listing it in case it helps. A Windows RemoteApp host and the usual apps (Explorer, Notepad, an Office app) are enough to hit most of it.

The backends behave differently, so ideally try more than one:
- native X11 (KDE, XFCE, ...)
- native Wayland (GNOME/Mutter, KDE, ...)
- XWayland (SDL x11 driver on a Wayland session)

### Windows and geometry
- [ ] Open several app windows at once: each is its own window with correct content, title and icon.
- [ ] Move a window by dragging: it follows smoothly, no drift or jumps.
- [ ] Resize from every edge and corner: content updates without lag or flicker (on X11 it stays in step with the server frames).
- [ ] Grab the very edge, i.e. the invisible resize border just outside the visible frame: the resize starts there, and the right resize cursor shows (on Wayland this goes through the subsurface band).
- [ ] Try to resize past the app's minimum and maximum size: it clamps.
- [ ] Minimize, then restore from the taskbar.
- [ ] Maximize: it fills the work area and does not cover the panel/taskbar; restore returns to the previous size and place.
- [ ] Drag a maximized window: it restores and follows the pointer.
- [ ] Snap/tile by dragging to a screen edge, and with the WM tile shortcuts (Super+Left/Right, etc.): the window and the server geometry agree.
- [ ] Close a window; close all windows: they go away cleanly, nothing left behind.

### Rendering
- [ ] Windows with rounded corners or per-pixel alpha (Win10/11 shadows, glass, etc.) render without black corners or frames.
- [ ] Content stays correct during and after a resize (no tearing, no stale pixels, no black bands).

### Z-order, focus and dialogs
- [ ] Click between overlapping windows: the clicked one comes to the front and the stacking stays correct.
- [ ] Switch focus by clicking another window or with Alt-Tab: typing goes to the active window.
- [ ] Open a modal dialog (File > Open / Save As): it appears above its parent, blocks it, and is not hidden behind other windows.
- [ ] A dialog owned by a window stays above that window.
- [ ] Floating tool windows / task panes / palettes (e.g. Office panes) sit relative to their owner and stay attached.
- [ ] Resize a window that is on top of others: the ones underneath are not disturbed.

### Popups, menus and shadows
- [ ] Right-click context menus appear at the pointer with no offset or drift.
- [ ] Dropdowns and submenus (e.g. the Office ribbon or File menu) are positioned correctly, no shift.
- [ ] Cascading submenus (menu > submenu > sub-submenu) line up correctly.
- [ ] The menu drop shadow looks right and does not spill past the menu.
- [ ] Clicking away dismisses the menu; opening and closing menus quickly leaves no ghosts or stuck popups.
- [ ] Tooltips show and clear.

### Fullscreen
- [ ] PowerPoint slideshow (F5): it goes fullscreen, covers the panel, and takes keyboard focus immediately (arrows/space advance slides without having to click first). Esc exits.

### Input
- [ ] Keyboard reaches the focused app, including special and Unicode characters.
- [ ] Mouse clicks, drag and scroll work; focus follows clicks.

### Session
- [ ] Disconnect and reconnect: the session windows come back and nothing is stuck.

### Known gaps (no need to file these)
- Multi-monitor and HiDPI/scaling are untested.
- Windows and macOS clients are not built or tested; the move/resize helpers are stubs there.

A single flow that touches a lot at once: open an Office app, open a couple of documents so there are several windows, tile two side by side, open File > Open in one (check it is on top and blocks its own window), right-click on the content (check the menu position and its shadow), start the PowerPoint slideshow (check fullscreen and keyboard), then move and resize the windows around.

geldot:
Here are my testing results with this branch on  a Wayland / KWin / hdpi / dual monitor Linux client connecting to Windows 11 Professional.

Config:

```
remoteapplicationmode:i:1
remoteapplicationname:s:Word
remoteapplicationprogram:s:"C:\Program Files\Microsoft Office\root\Office16\WINWORD.EXE"
disableremoteappcapscheck:i:1
desktopwidth:i:3840
desktopheight:i:2160
desktopscalefactor:i:150
screen mode id:i:1
```

`X11/xfreerdp`: connects and the program runs and works.

`SDL3/sdl-freerdp`: a black window with centered "Please wait" text and Windows login spinner appears and doesn't go away, and the program itself (Word) never appears.

I've tried removing the extra options (width/height/scale factor etc.) and it makes no difference - sdl-freerdp always gets stuck at the login screen.

geldot:
Update. The below hack seems to help `sdl-freerdp` get past the stale Windows login screen and with this Word is working in program mode:

```patch
diff --git i/client/SDL/SDL3/sdl_context.cpp w/client/SDL/SDL3/sdl_context.cpp
index d742ff67c..4b47a5713 100644
--- i/client/SDL/SDL3/sdl_context.cpp
+++ w/client/SDL/SDL3/sdl_context.cpp
@@ -1627,7 +1627,7 @@ bool SdlContext::useLocalScale() const

 bool SdlContext::drawToWindows(const std::vector<SDL_Rect>& rects)
 {
-       if (rects.empty())
+       if (rects.empty() && !_rail.enabled())
                return true;

        for (auto& window : _windows)
```

More hidpi/multi-display observations coming up shortly. First impressions are not bad at all.

geldot:
For multi-monitor, the `sendWorkArea()` hack won't work. It makes more sense to let the existing `/span` implementation set up a workarea that covers all Wayland monitors (or all displays in SDL terminology). Then, enforce the max size in `SdlRail::reportAndAdopt()` and `SdlRailWindow::reconcile()` instead. This allows app windows to be maximized and restored correctly while being bounded to the screen they're currently on.

svncibrahim:
Hi @geldot 

Thanks, this is really helpful, and good catch on the drawToWindows change. The login/blank window one is clearly a big issue, I'll take a look at it right away.

Thanks for testing HiDPI and multi-monitor too. Like I mentioned in the issue, I haven't had a chance to test those setups myself and figured something would break there. My setup doesn't really allow it right now, but I'm planning to do some multi-monitor testing over the next few weeks. Until then, any testing you do is very welcome.

svncibrahim:
@geldot I just pushed a fix for the login/blank window issue based on your findings. Could you please pull the latest commits and test if the problem is resolved on your end?

akallabeth:
@freerdp-bot test

akallabeth:
@svncibrahim are all the platform specific parts really required?
kind of defeats the purpose of having SDL in the first place (as in the least amount of platform specific code to ease maintenance)

svncibrahim:
> @svncibrahim are all the platform specific parts really required? kind of defeats the purpose of having SDL in the first place (as in the least amount of platform specific code to ease maintenance)

Hi @akallabeth,

I want the same thing. Most of the platform-specific code came from the Wayland side rather than by choice: getting and setting absolute window positions isn't possible on Wayland at all, so the code branched to work around that.

The resize margin is a different case, one that SDL just didn't expose, so I upstreamed it as libsdl-org/SDL#16108 (looking positive so far). Once it lands, sdl_wayland_band goes away, which is probably the biggest reduction.

This took a lot of trial and error, so the code is rougher than I'd like and it's spread across the commits. There are probably mistakes in there, so feedback is very welcome.

akallabeth:
@svncibrahim ok, then I'd skip the compatibility with current sdl and just require a version that contains the fix?
no need to make it compatible with something that will be quite a burden (to test, ...) on our end for a new feature.

as for the wayland positioning, is there not a experimental protocol addressing that already available? (did forget the name though, or I might have mixed that one up with something similar)

svncibrahim:
@akallabeth 

On compatibility, agreed, I'll drop the fallback and just require the fixed version. The problem is the fix is only on the 3.6 milestone with no release date yet, so I can't require it right now. I'll ask upstream if it can go into 3.4 too. Also, the resize margin makes windows a lot easier to resize, so I think it's worth having.

On positioning, do you mean xx-zones-v1 (https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/264, https://wayland.app/protocols/xx-zones-v1)? That's the one I found, but it's very new and no compositor supports it yet, so I don't think it helps us for now.

svncibrahim:
Small update, I dropped the raw libwayland dependency. Wayland resize is now server-driven using the window's own border, so it works perfectly on the current SDL release. 

The tradeoff is a thinner grab area (visible border only) compared to X11. I'll restore the wider margin once SDL exposes border insets, but for now, this keeps the backend minimal.

geldot:
I've taken the latest version for a spin. The login screen issue is resolved, and I'm glad to see the resize border hack gone as it was glitchy (https://github.com/libsdl-org/SDL/issues/16058).

One visual issue: In Word, some windows have correct transparency for the rounded corners, while some have a black background.

image

My sense is that hidpi will be a little more complex (more than just adding `SDL_WINDOW_ALLOW_HIGHDPI`), while multi monitor is straightforward and using the approach I suggested above works well (I'll try to prepare a patch).

The main question beyond that is I think code cleanup and architectural. The functionality is very compelling.

svncibrahim:
Hi @geldot,

Thanks for testing, glad login and the resize border are sorted.

I tested against a Windows 10 host, which has no rounded corners, so I probably missed the black background. Looks like a transparency classification bug on my side, I'll look into it.

Also, where are you testing multi monitor (and X11 or Wayland)? On my side Wayland multi monitor works well so far.

On multi monitor and anything else, patches are very welcome, thank you.

akallabeth:
@freerdp-bot test

svncibrahim:
Hi @geldot,

I set up a Windows 11 host and confirmed the black background issue. While testing, I actually found several more GFX and transparency bugs.

The root of the problem seems to be that Windows 10 didn't use transparency nearly as much as Windows 11 does, so my initial assumptions in the code were off.

I'll need a bit more time to refactor some parts to handle this properly. I'll be testing on both Win10 and Win11 in parallel from now on. I will push an update once the refactoring is done. Thanks!

svncibrahim:
Squashed the history into 6 commits and folded the fixes and cleanups into their origin commits.

Main changes since the last push:

- Transparency: popup and menu black backgrounds are fixed by taking the surface format from RDPEGFX instead of guessing, plus refined layered drop shadow rendering
Screenshot From 2026-08-24 04-37-43

- Window properties: WS_EX_TOPMOST via SDL_SetWindowAlwaysOnTop

- Pointer: bounds and allocation guards for custom cursors, **fixes crashes** on UAC prompts and invalid cursor updates

akallabeth:
@svncibrahim where do we stand here?
is this ready for merge or not usable at all?

svncibrahim:
Hi @akallabeth,

It is almost ready. I just need to fix one or two minor things and do some last-minute touches. I was quite busy last week so I couldn't get to it, but the PR should be fully ready by this weekend.

akallabeth:
@svncibrahim great!
then I'll add it to the next release milestone.

svncibrahim:
@akallabeth ready from my side. Rebased on current master, description is up to date.

Wayland works well enough for daily use, but it is still incomplete because of its known limitations. For the full RemoteApp experience I would recommend `SDL_VIDEODRIVER=x11` for now.

One thing you might not have seen: libsdl-org/SDL#14292 (draft) adds the experimental `xx-zones-v1` protocol. If that lands it covers most of what RemoteApp needs on Wayland. Compositor support is the missing piece today.

Tested against Windows 10 and 11 hosts, on KDE, XFCE, XWayland and native Wayland. Wider testing very welcome.

akallabeth:
@freerdp-bot test

akallabeth:
@svncibrahim ok, looks like there is a version guard missing for `SDL_HINT_VIDEO_X11_ENABLE_XSYNC_EXT`

svncibrahim:
> @svncibrahim ok, looks like there is a version guard missing for `SDL_HINT_VIDEO_X11_ENABLE_XSYNC_EXT`

Fixed, thanks. `SDL_HINT_VIDEO_X11_ENABLE_XSYNC_EXT` is SDL 3.4.10+, now behind a `SDL_VERSION_ATLEAST(3, 4, 10)` guard.

akallabeth:
@freerdp-bot test

akallabeth (commented):
Please review the remaining compiler warnings.

you can ignore the ones from the generated wayland source/header, these will be handled later.

svncibrahim:
Fixed in the push above: the sdl_* warnings are gone.

Only NOLINTs left are on the Xlib `XClientMessageEvent` union and its `long[5]` member, scoped to those lines.

akallabeth:
@freerdp-bot test

svncibrahim:
~~Sorry, missed two locally. Fixing now.~~

**Both fixed.**

akallabeth:
@freerdp-bot test

akallabeth:
@svncibrahim ok, ci looking good now. 
please have a little pationce, I'll do a final review but have currently not that much time, so might take a day to two.

akallabeth:
https://github.com/FreeRDP/FreeRDP/pull/13425 follow up

PR #13134, opened 2026-08-03
[core,server] fix length underflow in wts_read_drdynvc_pdu

metsw24-max:
**Length underflow in wts_read_drdynvc_pdu**

`length` tracks what is left of the received DYNVC PDU, but the variable-length ChannelId is subtracted from it without checking that the field actually fits, so a single byte PDU wraps it to `SIZE_MAX` and the truncated `0xFFFFFFFF` reaches `wts_queue_receive_data`, which then copies that many bytes out of the ~1600 byte channel buffer. The sibling `wts_read_drdynvc_data_first` already guards the same subtraction, so this mostly brings the two into line; rejecting the short PDU also stops the ChannelId being taken from stale buffer content, which the check inside `wts_read_variable_uint` does not catch because `channel->receiveData` reports its allocation as its length.

Added a regression test under `libfreerdp/core/test` that drives the server receive path with a truncated DYNVC PDU. It fails on master and passes with the change; the rest of the suite is unaffected.

### DIFF
--- a/libfreerdp/core/server.c
+++ b/libfreerdp/core/server.c
@@ -297,7 +297,7 @@ static BOOL wts_read_drdynvc_pdu(rdpPeerChannel* channel)
 	/* receiveData is allocated with a fixed chunk size and never sealed, so its stream length
 	 * is the allocation and not the number of bytes received. Parse through a substream
 	 * limited to the received data so the stream bounds checks apply to the PDU. */
-	wStream sbuffer = { 0 };
+	wStream sbuffer = WINPR_C_ARRAY_INIT;
 	wStream* s = Stream_StaticInit(&sbuffer, Stream_Buffer(channel->receiveData), length);
 
 	const UINT8 value = Stream_Get_UINT8(s);
--- a/libfreerdp/core/server.c
+++ b/libfreerdp/core/server.c
@@ -150,17 +150,18 @@ static unsigned wts_read_variable_uint(wStream* s, int cbLen, UINT32* val)
 	}
 }
 
-static BOOL wts_read_drdynvc_capabilities_response(rdpPeerChannel* channel, UINT32 length)
+static BOOL wts_read_drdynvc_capabilities_response(rdpPeerChannel* channel, wStream* s)
 {
 	UINT16 Version = 0;
 
 	WINPR_ASSERT(channel);
 	WINPR_ASSERT(channel->vcm);
-	if (length < 3)
+	WINPR_ASSERT(s);
+	if (!Stream_CheckAndLogRequiredLength(TAG, s, 3))
 		return FALSE;
 
-	Stream_Seek_UINT8(channel->receiveData); /* Pad (1 byte) */
-	Stream_Read_UINT16(channel->receiveData, Version);
+	Stream_Seek_UINT8(s); /* Pad (1 byte) */
+	Stream_Read_UINT16(s, Version);
 	DEBUG_DVC("Version: %" PRIu16 "", Version);
 
 	if (Version < 1)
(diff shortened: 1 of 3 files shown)


akallabeth (changes requested):
useless.

length is always >= val, check `wts_read_variable_uint` return values.

`length == Stream_Length()` and the number of bytes extracted from the stream are checked.

[edit] did you have a real issue here or was that just a generic review based fix? The handling could use some refactoring utilizing a substream (so the whole length extraction is not required at all)

metsw24-max:
**Why the stream checks do not cover this**

`channel->receiveData` is created with `Stream_New(nullptr, chunkSize)` (server.c:2187) and `Stream_New` sets `length = capacity`. Nothing seals it afterwards, the receive path only writes and resets the position, so `Stream_Length()` stays at the chunk size for the life of the channel. The received byte count is the stream position, which is what the local `length` was initialised from at server.c:293.

So `length` and `Stream_Length()` are not the same thing here, and the `Stream_CheckAndLogRequiredLength` inside `wts_read_variable_uint` was measuring against the ~1600 byte allocation rather than against the PDU. For a PDU that ends before its ChannelId the read still succeeds, out of whatever the previous PDU left in the buffer, and `length -= val` then wraps.

**Was it real**

Yes. Master with `WITH_DEBUG_DVC`, feeding a valid 6 byte DYNVC_CREATE_RSP and then a 1 byte one:

```
[wts_read_drdynvc_pdu]: Cmd CREATE_REQUEST_PDU ChannelId 1 length 4
[wts_read_drdynvc_pdu]: Cmd CREATE_REQUEST_PDU ChannelId 1 length 18446744073709551615
```

The client only sent `0x10`, so the ChannelId 1 is the stale byte from the previous PDU. `0xFFFFFFFF` then passes the `length < 4` check. Sending `0x30` instead puts the same value into `wts_read_drdynvc_data`, and the test binary dies with SIGBUS in the `CopyMemory` in `wts_queue_receive_data`.

**Change**

Agreed the one line check was the wrong shape, so I have reworked it the way you suggested. `wts_read_drdynvc_pdu` now parses through a substream limited to the received bytes, which makes the existing bounds checks meaningful, and the manual length extraction is gone from all four helpers. Test suite is 191/191 and the truncated PDU is rejected on both paths.

libfreerdp/core/server.c
```diff
@@ -290,20 +289,24 @@ static BOOL wts_read_drdynvc_pdu(rdpPeerChannel* channel)
...
 	if ((length < 1) || (length > UINT32_MAX))
 		return FALSE;
 
-	Stream_ResetPosition(channel->receiveData);
-	const UINT8 value = Stream_Get_UINT8(channel->receiveData);
-	length--;
+	/* receiveData is allocated with a fixed chunk size and never sealed, so its stream length
+	 * is the allocation and not the number of bytes received. Parse through a substream
+	 * limited to the received data so the stream bounds checks apply to the PDU. */
+	wStream sbuffer = { 0 };
```
akallabeth: prefer `WINPR_C_ARRAY_INIT`

akallabeth (approved):
LGTM,
nice catch

akallabeth:
@freerdp-bot test

libfreerdp/core/server.c
metsw24-max: Done, switched it to WINPR_C_ARRAY_INIT.

akallabeth (approved):

akallabeth:
@freerdp-bot test

akallabeth:
@metsw24-max sorry did merge in wrong order (had a fix already as part of #13160 ) but will merge your unit test

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/13162

PR #13136, opened 2026-08-03
[winpr,sspi] dispatch by package index instead of a mixed-width name

hamed7ir:
### The problem

`sspi_winpr` stored the security package's **name** in the upper pointer of the
`SecHandle` and read it back to select the dispatch table. The value was written
and read at different character widths — stored as TCHAR/WCHAR under `-DUNICODE`,
but compared as a narrow string — so the lookup could not match the name it had
itself stored.

With `WITH_NATIVE_SSPI=OFF`, the context and credential wrappers therefore failed
to resolve their own package, and NLA aborted with `SEC_E_SECPKG_NOT_FOUND`.

`schannel.c` writes explicit `SCHANNEL_PACKAGE_NAME_A` and `_W` values into that
same field, so both widths are genuinely in use on the same handle type — the
field carries no indication of which one it currently holds.

### The change

Store a 1-based package identifier (`SSPI_PACKAGE_ID`, `sspi.h`) in the handle
instead of a name, and add by-index table lookups
(`sspi_GetSecurityFunctionTable{A,W}ByIndex`) for the dispatch wrappers. The enum
order matches `SecPkgInfo{A,W}_LIST` and `SecurityFunctionTable{A,W}_NAME_LIST`,
so dispatch no longer depends on TCHAR width in either direction.

I went with an index rather than normalising everything to one width because the
field is untyped and both the A and W paths are legitimate. A single-width
convention would have to be honoured by every writer and every reader with
nothing enforcing it at compile time, so the same class of bug would be one new
call site away.

One thing worth your opinion: the index dispatch is correct as long as the enum
stays in step with those two lists, and nothing currently enforces that. I'm
happy to add a static assertion or an init-time check if you'd rather have it
guarded.

### How to test

The existing WinPR test suite already catches this, so no new test is needed.

Configure with the internal SSPI and internal tests enabled, on Windows with
`UNICODE` defined:

```
cmake -S . -B build -G Ninja -DBUILD_TESTING_INTERNAL=ON -DWITH_NATIVE_SSPI=OFF -DWITH_INTERNAL_RC4=ON -DWITH_INTERNAL_MD4=ON -DWITH_INTERNAL_MD5=ON
cmake --build build --target TestSspi
ctest --test-dir build -R "Sspi|Credential|SecurityContext|SecurityPackage"
```

On current master (`1cf992dc2), two of those tests fail:

| Test | master | this branch |
|---|---|---|
| TestQuerySecurityPackageInfo | pass | pass |
| TestEnumerateSecurityPackages | pass | pass |
| **TestInitializeSecurityContext** | **fail** | pass |
| **TestAcquireCredentialsHandle** | **fail** | pass |
| TestSspiNTLM | pass | pass |
| TestSspiPasswordHash | pass | pass |
| TestCredentialZeroing | pass | pass |
| | 5/7, ctest exit 8 | 7/7, ctest exit 0 |

The failure on master prints the error this description names:

```
DeleteSecurityContext status: 0x80090305      (SEC_E_SECPKG_NOT_FOUND)
```

`0x80090305` appears once in the master run and not at all with this change.

The configuration is presumably why this went unnoticed: the default Windows
build uses the OS SSPI (`WITH_NATIVE_SSPI=ON`), so the internal dispatch path
these tests exercise is never reached. A CI configuration with
`WITH_NATIVE_SSPI=OFF` would have caught it.

If you would rather see it end to end, connect with NLA using
`wfreerdp /v:<host> /u:<user> /p:<password>` against a build configured the same
way: authentication fails before the change and succeeds after.

Verified on Windows x64 (MSVC, vcpkg static OpenSSL/zlib); also in use on
Windows ARM32. The change is not platform-specific — it affects any `UNICODE`
build using the internal SSPI.

### Notes

Rebased on current master, single commit, no merges. Checked locally against the
required checks: `codespell` clean on the changed files, no CMake files touched,
and no new compiler warnings at `/W4` or clang-tidy findings on the changed lines
when compared with master's own copies of the same files.

### DIFF
--- a/winpr/libwinpr/sspi/sspi.h
+++ b/winpr/libwinpr/sspi/sspi.h
@@ -55,6 +55,20 @@ void* sspi_SecureHandleGetUpperPointer(SecHandle* handle);
 void sspi_SecureHandleSetUpperPointer(SecHandle* handle, void* pointer);
 void sspi_SecureHandleFree(SecHandle* handle);
 
+/* Package identity stored (1-based; 0 == unset) in a SecHandle upper-pointer.
+ * Replaces the former mixed-width _T() name-string scheme so dispatch never
+ * depends on TCHAR/char width. Order MUST match SecPkgInfo{A,W}_LIST /
+ * SecurityFunctionTable{A,W}_NAME_LIST in sspi_winpr.c. */
+typedef enum
+{
+	SSPI_PACKAGE_NONE = 0,
+	SSPI_PACKAGE_NTLM = 1,
+	SSPI_PACKAGE_KERBEROS = 2,
+	SSPI_PACKAGE_NEGOTIATE = 3,
+	SSPI_PACKAGE_CREDSSP = 4,
+	SSPI_PACKAGE_SCHANNEL = 5
+} SSPI_PACKAGE_ID;
+
 enum SecurityFunctionTableIndex
 {
 	EnumerateSecurityPackagesIndex = 1,
--- a/winpr/libwinpr/sspi/sspi.h
+++ b/winpr/libwinpr/sspi/sspi.h
@@ -55,6 +55,26 @@ void* sspi_SecureHandleGetUpperPointer(SecHandle* handle);
 void sspi_SecureHandleSetUpperPointer(SecHandle* handle, void* pointer);
 void sspi_SecureHandleFree(SecHandle* handle);
 
+/* Package identity of a SecHandle (1-based; 0 == unset). The order MUST match
+ * SecPkgInfo{A,W}_LIST and SecurityFunctionTable{A,W}_NAME_LIST in sspi_winpr.c;
+ * that is asserted at compile time there. */
+typedef enum WINPR_C23_ENUM_TYPE(uint32_t)
+{
+	SSPI_PACKAGE_NONE = 0,
+	SSPI_PACKAGE_NTLM = 1,
+	SSPI_PACKAGE_KERBEROS = 2,
+	SSPI_PACKAGE_NEGOTIATE = 3,
+	SSPI_PACKAGE_CREDSSP = 4,
+	SSPI_PACKAGE_SCHANNEL = 5,
+	SSPI_PACKAGE_COUNT /* number of identifiers, SSPI_PACKAGE_NONE included; keep last */
+} SSPI_PACKAGE_ID;
+
+/* Typed access to the package identity. Shares the upper-pointer slot and its encoding
+ * with sspi_SecureHandleGet/SetUpperPointer, but never forms a pointer from the
+ * identifier, so no integer-to-pointer conversion appears at any call site. */
+WINPR_ATTR_NODISCARD SSPI_PACKAGE_ID sspi_SecureHandleGetPackageId(SecHandle* handle);
+void sspi_SecureHandleSetPackageId(SecHandle* handle, SSPI_PACKAGE_ID id);
+
 enum SecurityFunctionTableIndex
 {
 	EnumerateSecurityPackagesIndex = 1,
--- a/winpr/libwinpr/sspi/sspi_winpr.c
+++ b/winpr/libwinpr/sspi/sspi_winpr.c
@@ -304,6 +304,22 @@ void sspi_SecureHandleSetUpperPointer(SecHandle* handle, void* pointer)
 	handle->dwUpper = (ULONG_PTR)(~((size_t)pointer));
 }
 
+SSPI_PACKAGE_ID sspi_SecureHandleGetPackageId(SecHandle* handle)
+{
+	if (!handle || !SecIsValidHandle(handle) || !handle->dwUpper)
+		return SSPI_PACKAGE_NONE;
+
+	return (SSPI_PACKAGE_ID)(~((size_t)handle->dwUpper));
+}
+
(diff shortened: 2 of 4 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@hamed7ir pr looks good but the casts produce quite a number of warnings.
might be a good idea to use the proper cast macros to avoid that.

winpr/libwinpr/sspi/sspi.h
```diff
@@ -55,6 +55,20 @@ void* sspi_SecureHandleGetUpperPointer(SecHandle* handle);
 void sspi_SecureHandleSetUpperPointer(SecHandle* handle, void* pointer);
 void sspi_SecureHandleFree(SecHandle* handle);
 
+/* Package identity stored (1-based; 0 == unset) in a SecHandle upper-pointer.
+ * Replaces the former mixed-width _T() name-string scheme so dispatch never
+ * depends on TCHAR/char width. Order MUST match SecPkgInfo{A,W}_LIST /
+ * SecurityFunctionTable{A,W}_NAME_LIST in sspi_winpr.c. */
+typedef enum
```
akallabeth: maybe use the `WINPR_C23_ENUM_TYPE` macro

winpr/libwinpr/sspi/sspi_winpr.c
```diff
@@ -995,6 +995,8 @@ void sspi_GlobalFinish(void)
 	sspi_ContextBufferAllocTableFree();
 }
 
+/* By-NAME lookups: used by callers that pass a real package name string
```
akallabeth: please remove the comments, they are referring to outdated information

winpr/libwinpr/sspi/sspi_winpr.c
```diff
@@ -1040,6 +1042,26 @@ static const SecurityFunctionTableW* sspi_GetSecurityFunctionTableWByNameA(const
...
+ * (WCHAR name read as char*). */
+static const SecurityFunctionTableA* sspi_GetSecurityFunctionTableAByIndex(UINT_PTR id)
+{
+	if ((id < SSPI_PACKAGE_NTLM) || (id > ARRAYSIZE(SecurityFunctionTableA_NAME_LIST)))
+		return nullptr;
+
+	return SecurityFunctionTableA_NAME_LIST[id - 1].SecurityFunctionTable;
+}
+
+static const SecurityFunctionTableW* sspi_GetSecurityFunctionTableWByIndex(UINT_PTR id)
```
akallabeth: * `WINPR_ATTR_NODISCARD` might be a good idea
* as argument use `const void*` to avoid casts all over the code
* * even better, pass `phCredential` directly and extract it in this helper

winpr/libwinpr/sspi/sspi_winpr.c
```diff
@@ -1040,6 +1042,26 @@ static const SecurityFunctionTableW* sspi_GetSecurityFunctionTableWByNameA(const
 	return table;
 }
 
+/* By-INDEX lookups: used ONLY by the context/credential dispatch wrappers, whose
+ * handle upper-pointer stores a 1-based SSPI_PACKAGE_ID (see sspi.h). Width-agnostic
+ * replacement for the old name-string dispatch, which mis-fired under -DUNICODE
+ * (WCHAR name read as char*). */
+static const SecurityFunctionTableA* sspi_GetSecurityFunctionTableAByIndex(UINT_PTR id)
```
akallabeth: * `WINPR_ATTR_NODISCARD` might be a good idea
* as argument use `const void*` to avoid casts all over the code
* even better, pass `phCredential` directly and extract it in this helper

akallabeth (approved):
LGTM

hamed7ir:
> @hamed7ir pr looks good but the casts produce quite a number of warnings. might be a good idea to use the proper cast macros to avoid that.

Thanks, all done , though one thing : the call-site casts are
pointer→int, and those aren't what's warning. The noise is from the int→pointer
casts on the write side (performance-no-int-to-ptr), so just passing the handle
into the helpers would have left them all in place. Fixed both: the helpers take
the handle now, and typed get/set accessors handle the encoding without ever
forming a pointer. No raw casts left in the diff.
I can't reproduce your exact set though , .clang-tidy disables that check, so I
had to force it on, and I'm on MSVC rather than gcc. If anything's still warning
after this, let me know which build.
Also: WINPR_C23_ENUM_TYPE on the enum, WINPR_ATTR_NODISCARD on both lookups,
dropped the outdated comments (there was a third in sspi.h), and renamed
ByIndex → ByHandle since it's not an index anymore , sorry for moving your
comment anchors.
I went ahead and added the static assert I'd asked about, since it was cheap:
it ties the lists to SSPI_PACKAGE_COUNT so adding a package without updating
them won't compile. Easy to drop if you'd rather not.
Rebased on master, 7/7 with WITH_NATIVE_SSPI=OFF. The negotiate.c change in
cbdf4d9c7 doesn't overlap , it reads the lower pointer, this only touches the
upper slot. Updated the stale master hash in the description too.

akallabeth:
@freerdp-bot test

akallabeth:
looks like there are some functions now in code that is also compiled on windows (and fails there)

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

PR #13144, opened 2026-08-07
[server,utils] fix clipboard image transfer channel buffer

wubw0508:
ChannelPduTracker allocated and validated against CHANNEL_PDU_LENGTH, which does not cover the largest virtual-channel chunk plus its PDU header.

Use CHANNEL_CHUNK_MAX_LENGTH with the header size for the receive buffer and its bounds checks. This lets remote desktop clipboard image transfers receive full channel PDUs without truncation or rejecting valid chunks.

### DIFF
--- a/libfreerdp/utils/channel_pdu_tracker.c
+++ b/libfreerdp/utils/channel_pdu_tracker.c
@@ -28,7 +28,7 @@ struct ChannelPduTracker
 {
 	HANDLE vc;
 	wStream* currentPacket;
-	char buffer[CHANNEL_PDU_LENGTH];
+	char buffer[CHANNEL_CHUNK_MAX_LENGTH + sizeof(CHANNEL_PDU_HEADER)];
 	size_t offset;
 	wLog* log;
 };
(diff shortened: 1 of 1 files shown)


wubw0508:

Yes, I can prove it's effective

As shown in the figure below, the left side is the client machine, and the right side is the page where Remmina connects to the server. There are clipboard on both machines. After either machine takes a screenshot, the image will enter the clipboard on both the client and server sides. The function of transferring images to the clipboard can only take effect if this submission is added.

ex

hardening:
Are you sure you're running the last git version because the bug this PR is supposed to fix should have been already fixed by 40d9202cd95551600c07437ce6bd5ecd7d31e57b fixing #13102 which looks like exactly what you're describing. So except if I missed something this fix is wrong.

wubw0508:
> Are you sure you're running the last git version because the bug this PR is supposed to fix should have been already fixed by [40d9202](https://github.com/FreeRDP/FreeRDP/commit/40d9202cd95551600c07437ce6bd5ecd7d31e57b) fixing #13102 which looks like exactly what you're describing. So except if I missed something this fix is wrong.

https://github.com/FreeRDP/FreeRDP/commit/40d9202cd95551600c07437ce6bd5ecd7d31e57b fixing
The commit can slove the question

PR #13145, opened 2026-08-07
[client,android] replace deprecated KeyboardView

svncibrahim:
## Overview

`android.inputmethodservice.KeyboardView` has been deprecated since API 29. aFreeRDP used two of them (an extended key panel and a separate modifier row) alongside the system IME, behind two independent toggle buttons. This replaces all of it with a single `ExtendedKeyboardView`: a persistent modifier bar with a collapsible, paged key panel.

## Key Changes

- **One keyboard, one button.** The "system keyboard" and "extended keyboard" toggles become a single toggle, and the duplicate buttons in the floating toolbar and the touch pointer are removed.
- **New `ExtendedKeyboardView`.** A persistent bar (Esc, Tab, Shift, Ctrl, Win, Alt, Del, panel toggle) over a collapsible panel with two pages: function/navigation keys and numpad. Replaces 4 XML keyboard definitions and 20 bitmap key assets.
- **Long press to lock a modifier.** A tap arms the modifier for the next key only, a long press latches it. Locking used to be a double tap within 800 ms, so the second tap meant "lock" or "turn off" depending only on timing. A small lamp on each modifier key shows off / armed / locked.
- **Ctrl+Alt+Del without opening the panel.** `Del` and `Tab` moved onto the persistent bar next to the modifiers. They used to live on the extended key page, so sending Ctrl+Alt+Del meant expanding the keyboard first.

- **Themed and responsive.** Light and dark palettes, orientation specific key sizes, key auto repeat while held, and a layered back press. The navigation bar also leaves immersive mode while the IME is up, so the back gesture works on the first swipe.
- **No ActionBar on the session screen.** `SessionActivity` carried an ActionBar it never showed. Below API 35, AppCompat folds that ActionBar's height into the insets handed to the content, which the session read as the status bar inset and turned into a black band above the remote desktop.

### Old keyboard:

Screenshot_20260807_113920
Screenshot_20260807_113942

---
### New keyboard (collapsed bar, expanded panel, light and dark):
Screenshot_20260807_125929
 Screenshot_20260807_130017 Screenshot_20260807_130024 Screenshot_20260807_130624 

## How to Test
1. Build, run, and connect to a session.
2. Toggle the keyboard from the floating toolbar or the touch pointer. The key bar and the system IME come up together.
3. Tap `▼` to expand the panel, `⇆` to switch between the function page and the numpad.
4. Tap `Ctrl`, press a letter, and confirm Ctrl releases itself. Long press `Ctrl` and confirm it stays down until pressed again. The lamp should follow.
5. Hold `⌫` for auto repeat, press back to walk down the layers, and rotate with the panel open.
6. With "hide status bar" off, confirm there is no black band above the remote desktop.

## Environments Tested

- Android 14 (API 34) phone emulator, 720p
- Android 16 (API 36) tablet


client/Android/Studio/freeRDPCore/src/main/java/com/freerdp/freerdpcore/presentation/ExtendedKeyboardView.java
```diff
@@ -0,0 +1,449 @@
+/*
+   Android Extended On-screen Keyboard
+
+   Copyright 2026 Ibrahim Sevinc <ibrahim.sevinc.mail@gmail.com>
+
+   This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0.
```
hardening, line 6: Hum the best expert of these subjects but aren't we Apache licence in FreeRDP. I don't know if Mozilla licence mixes so well with Apache...

svncibrahim: I used Mozilla licence because the other files in the Android client already have that
header, so I just followed them. I did the same in my earlier Android PRs.

Happy to change it to Apache if you prefer.

akallabeth: @hardening android client is MPL ;)

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM, need to do a runtime check though.

PR #13149, opened 2026-08-10
Vmonitors

hardening:
A draft PR just to see if this functionality would be useful, for now it basically works, but I've seen some remaining bugs during vmonitor resize with `/dynamic-resolution`.


akallabeth:
@hardening interesting idea, might be helpful for testing as well.

PR #13152, opened 2026-08-11
[libfreerdp,crypto] add compilation opt-in TLS AEAD data limit with RFC 8446 threshold

jclairembault:
OpenSSL does not enforce the AEAD cryptographic data limits itself, see https://github.com/openssl/openssl/issues/23566. To respect https://datatracker.ietf.org/doc/html/rfc8446#section-5.5 an explicit rekey is required.

Add a WITH_TLS_DATA_LIMIT compile-time opt-in (OFF by default, to avoid interoperability issues not yet detected) that tracks per-direction plaintext bytes and, once the threshold is crossed, refreshes the TLS 1.3 traffic keys in place via SSL_key_update. On non-TLS-1.3 connections a rekey is not possible, so accounting is disabled and the connection is kept; if a TLS 1.3 rekey fails the connection is cut.

The threshold is RFC 8446 5.5's bound for AES-GCM: 2^24.5 records of 2^14 bytes (~362 GiB).

It may be tested by reducing the threshold to something lower in the code and check the the rekey is effective via wireshark.


akallabeth:
@freerdp-bot test

jclairembault:
I've just fixed the formatting issue in CMakeList.txt, sorry about that I had just check the C files with clang...

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13169, opened 2026-08-13
[winpr,sspi] fix server-side SPNEGO mechanism fallback

### LINKED ISSUE #13168: Server NLA: SPNEGO fallback after unsupported optimistic mech (NEGOEX) fails with SEC_E_INVALID_HANDLE
> **Heads up:** the failure itself was hit and verified by hand against a real deployment (Windows 11 client → gnome-remote-desktop), but the root-cause analysis and the proposed patch were identified and authored by Claude (AI assistant), with the fix hand-tested end-to-end. Please review with that in mind.

## Environment

- **Server:** gnome-remote-desktop 50.0 (Ubuntu 26.04) using distro freerdp3 / libwinpr3 **3.30.0** (`3.30.0+dfsg-0ubuntu0.26.04.2`). The affected code is unchanged in current `master`.
- **Client:** Windows 11, Microsoft Entra (Azure AD)-joined, built-in mstsc.
- Applies to any WinPR-based RDP server doing NLA with username/password (credentials/SAM), not just gnome-remote-desktop.

## Summary

When a Windows 11 client (observed with Entra-joined machines) connects with NLA, its CredSSP `negTokenInit` carries `mechTypes = [NEGOEX, NTLM]` with an **optimistic NEGOEX mech token**. WinPR does not implement NEGOEX, so `negotiate_AcceptSecurityContext` correctly falls back to NTLM from the mech list — but the fallback path is broken: authentication always fails with `SEC_E_INVALID_HANDLE` and the client gets a generic credential error, regardless of the password entered.

## Server log (WLOG_LEVEL=DEBUG)

```
[DEBUG][com.freerdp.core.nego] - [nego_process_negotiation_request]: RDP_NEG_REQ: RequestedProtocol: [SSL|HYBRID|HYBRID_EX][0x0000000b]
[DEBUG][com.freerdp.core.connection] - [rdp_server_accept_nego]: Negotiated Security: RDSTLS:0 NLA:1 TLS:0 RDP:0
[DEBUG][com.freerdp.core.auth] - [credssp_auth_init]: Using package: Negotiate (cbMaxToken: 12256 bytes)
[DEBUG][com.freerdp.core.nla] - [nla_server_authenticate]: Receiving Authentication Token
[DEBUG][com.winpr.negotiate] - [negotiate_AcceptSecurityContext]: Mechanism: SPNEGO (1.3.6.1.5.5.2)
[DEBUG][com.winpr.negotiate] - [negotiate_read_neg_token]: Reading negTokenInit...
[DEBUG][com.winpr.negotiate] - [negotiate_read_neg_token]:      mechTypes [0] (26 bytes)        <-- [NEGOEX, NTLM]
[DEBUG][com.winpr.negotiate] - [negotiate_read_neg_token]:      mechToken [2] (323 bytes)       <-- optimistic NEGOEX token
[DEBUG][com.winpr.negotiate] - [negotiate_AcceptSecurityContext]: Requested mechanism: NTLM (1.3.6.1.4.1.311.2.2.10)
[DEBUG][com.winpr.negotiate] - [negotiate_AcceptSecurityContext]: Accepted mechanism: NTLM (1.3.6.1.4.1.311.2.2.10)
[DEBUG][com.winpr.negotiate] - [negotiate_write_neg_token]: Writing negTokenResp...
[DEBUG][com.winpr.negotiate] - [negotiate_write_neg_token]:      negState [0] (3)               <-- request-mic, no mech token
[DEBUG][com.winpr.negotiate] - [negotiate_write_neg_token]:      supportedMech [1] (NTLM (1.3.6.1.4.1.311.2.2.10))
<client answers with negTokenResp carrying an NTLM NEGOTIATE message>
[WARN][com.winpr.sspi] - [winpr_AcceptSecurityContext]: AcceptSecurityContext status SEC_E_INVALID_HANDLE [0x80090301]
[ERROR][com.freerdp.core.auth] - [credssp_auth_authenticate]: AcceptSecurityContext status SEC_E_INVALID_HANDLE [0x80090301]
[ERROR][com.freerdp.core.transport] - [transport_accept_nla]: client authentication failure
[ERROR][com.freerdp.core.peer] - [peer_recv_callback_internal]: CONNECTION_STATE_NEGO - rdp_server_accept_nego() fail
```

## Root cause

`winpr/libwinpr/sspi/Negotiate/negotiate.c`, `negotiate_AcceptSecurityContext`:

1. **Round 1:** the optimistic mech OID (NEGOEX, `1.3.6.1.4.1.311.2.2.30`) is not in `MechTable`, so `init_context.mech` stays `NULL` and the sub-mech `AcceptSecurityContext` is never invoked. The fallback loop then picks NTLM from `mechTypes` and a `NEGOTIATE_CONTEXT` is created — but `context->sub_context` remains the zero-initialized handle from `NEGOTIATE_CONTEXT_init()` (`{ .dwLower = 0, .dwUpper = 0 }`). The server replies `negState = request-mic, supportedMech = NTLM` with no mech token (correct per SPNEGO).
2. **Round 2:** the client restarts with an NTLM NEGOTIATE message. The server calls:

   ```c
   status = context->mech->pkg->table->AcceptSecurityContext(
       sub_cred, &context->sub_context, &mech_input, ...);
   ```

   passing the still-empty `&context->sub_context` as `phContext`. `ntlm_AcceptSecurityContext` begins with:

   ```c
   /* behave like windows SSPIs that don't want empty context */
   if (phContext && !phContext->dwLower && !phContext->dwUpper)
       return SEC_E_INVALID_HANDLE;
   ```

   so the empty-but-non-NULL handle is rejected and authentication dies. (The Kerberos SSP has the same guard, so a Kerberos fallback would fail identically.)

The path triggers whenever round 1 ends without a created sub-context, i.e. for **any** of:

- the optimistic mech is not in `MechTable` (NEGOEX — the Windows 11 / Entra-joined case);
- the optimistic mech is supported but its `AcceptSecurityContext` fails (e.g. Kerberos u2u with no keytab — note `kerberos_AcceptSecurityContext` invalidates `phNewContext` on first-call failure, returning the handle to the empty state);
- the client sends a `negTokenInit` with **no optimistic token at all**, which is perfectly legal SPNEGO.

Clients whose optimistic token is NTLM itself (e.g. xfreerdp without Kerberos credentials) work fine against the same server, which is why this went unnoticed.

**Regression scope:** the empty-context guards were introduced by 58a3919435 ("winpr: add some checks in InitializeSecurityContext and AcceptSecurityContext", 2022-10-10), so all 3.x releases are affected; 2.x is not (an empty handle resolved to a NULL lower pointer and was treated as "create a fresh context").

Notably, the **client side already handles the symmetric situation correctly**: when the acceptor picks a different mech than the optimistic one, `negotiate_InitializeSecurityContextW` deletes the optimistic sub-context and passes `NULL` into the newly selected mech (the `/* Clean up the optimistic mech */` block). The server-side fallback path is missing the same treatment.

There is also a latent stale-context hazard on the server path: `ntlm_AcceptSecurityContext` publishes its context into `phNewContext` *before* parsing the NEGOTIATE message and does not clean up on parse failure, so a failed supported-optimistic-mech round can leave a sub-context belonging to mech A in the handle that round 2 then feeds to fallback mech B's `AcceptSecurityContext` (type confusion), in addition to leaking it. The fix below addresses this too by cleaning up the partial sub-context at fallback time, mirroring the client side.

## Proposed fix (tested)

Two hunks in `negotiate_AcceptSecurityContext`, mirroring the existing client-side mech-change handling:

1. On optimistic-mech failure, delete any partially created sub-context before falling back (fixes the leak/type-confusion above; both `ntlm_DeleteSecurityContext` and `kerberos_DeleteSecurityContext` invalidate the handle back to the empty state and tolerate empty handles).
2. On the following round, pass `NULL` for `phContext` when the sub-context is absent, so the selected mech allocates a fresh context. The emptiness test uses `sspi_SecureHandleGetLowerPointer`, which returns NULL exactly for the zero-initialized/invalidated handle states.

```diff
--- a/winpr/libwinpr/sspi/Negotiate/negotiate.c
+++ b/winpr/libwinpr/sspi/Negotiate/negotiate.c
@@ negotiate_AcceptSecurityContext (round 1, optimistic mech failed)
 		if (IsSecurityStatusError(status))
 		{
 			if (!init_context.spnego)
 				return status;
 
+			/* Clean up any partially created sub context before falling back to
+			 * another mechanism (mirrors the mech change handling in
+			 * negotiate_InitializeSecurityContextW) */
+			if (init_context.mech)
+				init_context.mech->pkg->table->DeleteSecurityContext(&init_context.sub_context);
+
 			init_context.mic = TRUE;
 			first_mech = init_context.mech;
 			init_context.mech = nullptr;
 			output_token.mechToken.cbBuffer = 0;
 		}
@@ negotiate_AcceptSecurityContext (later round, SPNEGO branch)
 			/* Use the output buffer to store the optimistic token */
 			if (output_buffer)
 				CopyMemory(&output_token.mechToken, output_buffer, sizeof(SecBuffer));
 
-			status = context->mech->pkg->table->AcceptSecurityContext(
-			    sub_cred, &context->sub_context, &mech_input, fContextReq | context->mech->flags,
-			    TargetDataRep, &context->sub_context, &mech_output, pfContextAttr, ptsTimeStamp);
+			/* If the client's optimistic mechanism was not accepted the sub
+			 * context was never created (or was cleaned up); pass NULL so the
+			 * selected mechanism creates a fresh context instead of rejecting
+			 * the empty handle with SEC_E_INVALID_HANDLE */
+			{
+				PCtxtHandle sub_context = sspi_SecureHandleGetLowerPointer(&context->sub_context)
+				                              ? &context->sub_context
+				                              : nullptr;
+				status = context->mech->pkg->table->AcceptSecurityContext(
+				    sub_cred, sub_context, &mech_input, fContextReq | context->mech->flags,
+				    TargetDataRep, &context->sub_context, &mech_output, pfContextAttr,
+				    ptsTimeStamp);
+			}
```

Verified with a build against 3.30.0 loaded into gnome-remote-desktop: the Windows 11 client completes the full CredSSP exchange — NEGOEX optimistic token declined, NTLM fallback with `request-mic`, challenge/authenticate, `mechListMIC` verified — and NLA authentication succeeds. A non-fallback NLA login (xfreerdp → same server, NTLM optimistic token) still works, so the normal path is unaffected.

Security considerations reviewed:

- The SPNEGO downgrade protection is untouched: the fallback path still sets `context->mic = TRUE`, and a final token without a MIC is still rejected (`SEC_E_INVALID_TOKEN`).
- The NULL-context branch is not attacker-steerable after the sub-context exists: handle emptiness is only produced by "never created" or `DeleteSecurityContext`, never by client input, and the `NEGOTIATE_STATE_NEGORESP` gate on processing mech tokens is unchanged.
- Net memory effect is positive (the partial sub-context previously leaked on fallback).

## Steps to reproduce

1. Any WinPR-based NLA server with username/password auth (gnome-remote-desktop is the easiest: enable Desktop Sharing with credentials).
2. Connect with mstsc from an Entra-joined Windows 11 machine (these send `mechTypes = [NEGOEX, NTLM]` with a NEGOEX optimistic token in CredSSP).
3. Correct credentials fail; server logs `AcceptSecurityContext status SEC_E_INVALID_HANDLE [0x80090301]`.

mjpowersjr:
> **Heads up:** the failure was hit and verified by hand against a real deployment (Windows 11 client → gnome-remote-desktop), but the root-cause analysis and this patch were identified and authored by Claude (AI assistant), with the fix hand-tested end-to-end. Please review with that in mind.

Fixes #13168

## What

Server-side `negotiate_AcceptSecurityContext` fails NLA with `SEC_E_INVALID_HANDLE` whenever round 1 ends without a created sub-context — i.e. when the client's optimistic SPNEGO mech is unsupported (NEGOEX, sent by Windows 11 Entra-joined clients), when a supported optimistic mech fails, or when the client legally sends no optimistic token at all. The next round passed the empty `sub_context` handle to the selected mechanism's `AcceptSecurityContext`, which rejects empty handles since 58a3919435 (so all 3.x affected, 2.x not).

## How

Mirrors the existing mech-change handling on the client side (`negotiate_InitializeSecurityContextW`, the "Clean up the optimistic mech" block):

1. On optimistic-mech failure, delete any partially created sub-context before falling back. Besides restoring the handle to its empty state (both `ntlm_DeleteSecurityContext` and `kerberos_DeleteSecurityContext` invalidate it), this also fixes a latent leak/type-confusion: `ntlm_AcceptSecurityContext` publishes its context into `phNewContext` before parsing and does not clean up on failure, so a failed mech-A round could feed mech A's context to fallback mech B in the next round.
2. Pass `NULL` instead of the empty handle so the selected mechanism creates a fresh context. Emptiness is tested with `sspi_SecureHandleGetLowerPointer`, which returns NULL exactly for the zeroed/invalidated handle states.

Security notes: the SPNEGO downgrade protection is unchanged (fallback still sets `mic = TRUE`; a final token without a MIC is still rejected), and the NULL-context branch is not attacker-steerable once a sub-context exists (handle emptiness only arises from "never created" or `DeleteSecurityContext`, never from client input).

## Testing

- Built against 3.30.0 and loaded into gnome-remote-desktop 50.0 (Ubuntu 26.04): a previously-failing Windows 11 Entra-joined mstsc client now completes the full CredSSP exchange (NEGOEX optimistic token declined → NTLM fallback with `request-mic` → challenge/authenticate → `mechListMIC` verified) and NLA authentication succeeds.
- Regression: non-fallback NLA login (xfreerdp → same server, NTLM optimistic token) still authenticates.
- Compile-checked against current master (both patch sites are identical between 3.30.0 and master).

Full protocol traces and analysis in #13168.

### DIFF
--- a/winpr/libwinpr/sspi/Negotiate/negotiate.c
+++ b/winpr/libwinpr/sspi/Negotiate/negotiate.c
@@ -1126,6 +1126,12 @@ static SECURITY_STATUS SEC_ENTRY negotiate_AcceptSecurityContext(
 			if (!init_context.spnego)
 				return status;
 
+			/* Clean up any partially created sub context before falling back to
+			 * another mechanism (mirrors the mech change handling in
+			 * negotiate_InitializeSecurityContextW) */
+			if (init_context.mech)
+				init_context.mech->pkg->table->DeleteSecurityContext(&init_context.sub_context);
+
 			init_context.mic = TRUE;
 			first_mech = init_context.mech;
 			init_context.mech = nullptr;
@@ -1213,8 +1219,15 @@ static SECURITY_STATUS SEC_ENTRY negotiate_AcceptSecurityContext(
 			if (output_buffer)
 				CopyMemory(&output_token.mechToken, output_buffer, sizeof(SecBuffer));
 
+			/* If the client's optimistic mechanism was not accepted the sub
+			 * context was never created (or was cleaned up); pass NULL so the
+			 * selected mechanism creates a fresh context instead of rejecting
+			 * the empty handle with SEC_E_INVALID_HANDLE */
+			PCtxtHandle sub_context = sspi_SecureHandleGetLowerPointer(&context->sub_context)
+			                              ? &context->sub_context
+			                              : nullptr;
 			status = context->mech->pkg->table->AcceptSecurityContext(
-			    sub_cred, &context->sub_context, &mech_input, fContextReq | context->mech->flags,
+			    sub_cred, sub_context, &mech_input, fContextReq | context->mech->flags,
 			    TargetDataRep, &context->sub_context, &mech_output, pfContextAttr, ptsTimeStamp);
 
 			if (IsSecurityStatusError(status))


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #13176, opened 2026-08-14
[codec,h264] fix offset region change detection

zx1991:
## Description

`detect_changes` currently starts its tile loop at `regionRect->left` and
`regionRect->top`, then adds those origins a second time when constructing each
tile rectangle. For dirty regions away from `(0, 0)`, the leading part of the
region is therefore omitted and the generated metadata starts at an incorrect
offset.

Iterate tile-relative offsets from zero before translating them into surface
coordinates.

## Testing

- Added a regression case that encodes a full frame followed by a changed
  `(32, 16)-(160, 112)` region.
- Confirmed that the new test fails with the original loop and passes with this
  change.
- `ctest --test-dir build-pr-make --output-on-failure` (161/161 passed)
- `ctest --test-dir build-pr-make --output-on-failure -R '^TestFreeRDPCodecH264$'`
- `clang-format --style=file --dry-run --Werror libfreerdp/codec/h264.c libfreerdp/codec/test/TestFreeRDPCodecH264.c`

### DIFF
--- a/libfreerdp/codec/h264.c
+++ b/libfreerdp/codec/h264.c
@@ -301,9 +301,9 @@ static BOOL detect_changes(BOOL firstFrameDone, const UINT32 QP, const RECTANGLE
 	}
 	else
 	{
-		for (size_t y = regionRect->top; y < regionRect->bottom; y += 64)
+		for (size_t y = 0; y < regionRect->bottom - regionRect->top; y += 64)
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #13178, opened 2026-08-14
[winpr,sysinfo] use CLOCK_MONOTONIC as a fallback

brad0:
CLOCK_MONOTONIC_RAW is a Linux-specific clock ID exposed
via glibc and setting it to a value of 4 has a different
meaning for different libcs or no meaning at all.

### DIFF
--- a/winpr/libwinpr/sysinfo/sysinfo.c
+++ b/winpr/libwinpr/sysinfo/sysinfo.c
@@ -327,8 +327,10 @@ BOOL GetSystemTimeAdjustment(WINPR_ATTR_UNUSED PDWORD lpTimeAdjustment,
 	return FALSE;
 }
 
-#ifndef CLOCK_MONOTONIC_RAW
-#define CLOCK_MONOTONIC_RAW 4
+#ifdef CLOCK_MONOTONIC_RAW
+#define CLOCK_ID CLOCK_MONOTONIC_RAW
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -586,7 +582,7 @@ UINT64 winpr_GetTickCount64NS(void)
 #if defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE >= 199309L)
 	struct timespec ts = WINPR_C_ARRAY_INIT;
 
-	if (clock_gettime(CLOCK_MONOTONIC_RAW, &ts) == 0)
```
akallabeth, line 589: better do a fallback to `CLOCK_MONOTONIC` if `CLOCK_MONOTONIC_RAW` is not defined.

akallabeth (changes requested):

winpr/libwinpr/sysinfo/sysinfo.c
brad0: Done.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13185, opened 2026-08-14
[core,nego] report why security negotiation failed

thenextman:
When we connect to a server with NLA required and the client has NLA disabled, the server refuses with a `RDP_NEG_FAILURE` PDU holding `HYBRID_REQUIRED_BY_SERVER`. The failure code is logged at `WLOG_WARN` and then thrown away, and the connection falls back to the blanket `ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED`.

I understand that from the perspective of freerdp's own clients, it's basically a non-issue (defaults should always enable NLA, the reason is visible in the terminal, etc). For a third-party client it's trickier, yes this scenario is becoming vanishingly small but it does still occur that users misconfigure their session and although the failure is _known_ by the library, we have to invest a support cycle to understand the problem.

Notably mstsc.exe does report this distinctly to the user

> The remote computer requires Network Level Authentication, which your computer does not
> support. For assistance, contact your system administrator or technical support.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #13186, opened 2026-08-14
[core,gateway] bound tsg authorize tunnel response to received length

metsw24-max:
**Out-of-bounds read in the TSG authorise-tunnel response parser**

`tsg_ndr_read_packet_response_data` guards its two NDR field reads with `Stream_CheckAndLogRequiredCapacityOfSizeWLog`, and `TsProxyAuthorizeTunnelReadResponse` reads `responseDataLen` with no length check at all. The RPC stub arrives in the reused `pdu->s` buffer, which is sealed with `Stream_SealLength` to the received size while its capacity stays at the larger pool allocation, so on a short response from the gateway the capacity checks still pass and the reads step past the sealed message into stale pool bytes. Switched the two checks to the length form used everywhere else in the file and added the missing check before `responseDataLen`; the malicious gateway is the trust boundary here and well-formed responses behave the same as before.

### DIFF
--- a/libfreerdp/core/gateway/tsg.c
+++ b/libfreerdp/core/gateway/tsg.c
@@ -1879,7 +1879,7 @@ static BOOL tsg_ndr_read_packet_response_data(rdpTsg* tsg, wStream* s,
 {
 	WINPR_ASSERT(tsg);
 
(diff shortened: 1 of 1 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13191, opened 2026-08-15
uwac: only use O_TMPFILE if it exists

brad0:
use O_TMPFILE if it exists, which is a Glibc-ism, otherwise
fallback to the fallback path for other libcs.

Also add missing check for mkostemp() and setting
UWAC_HAVE_MKOSTEMP.


brad0:
Also looking over the rest of the related code that is there I saw a variable by the name of ```UWAC_HAVE_MKOSTEMP``` which is never set anywhere.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13192, opened 2026-08-15
[cmake] use pkg-config to detect epoll-shim

brad0:


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

cmake/PlatformDefaults.cmake
```diff
@@ -109,7 +109,14 @@ if(USE_PLATFORM_DEFAULT)
   add_compile_definitions("WINPR_TIMEZONE_FILE=\"${WINPR_TIMEZONE_FILE}\"")
 
   if(EPOLLSHIM)
-    find_path(EPOLLSHIM_INCLUDE_DIR NAMES sys/epoll.h sys/timerfd.h HINTS /usr/local/include/libepoll-shim)
-    find_library(EPOLLSHIM_LIBS NAMES epoll-shim libepoll-shim HINTS /usr/local/lib)
+    find_package(PkgConfig QUIET)
+    pkg_check_modules(PC_EPOLLSHIM epoll-shim)
+
+    find_path(EPOLLSHIM_INCLUDE_DIR
```
akallabeth: not sure, but should this be `REQUIRED`?
(as in if we enter here this must be found or else abort?

cmake/PlatformDefaults.cmake
```diff
@@ -109,7 +109,14 @@ if(USE_PLATFORM_DEFAULT)
...
+              NAMES sys/epoll.h sys/timerfd.h
+              HINTS /usr/local/include/libepoll-shim ${PC_EPOLLSHIM_INCLUDE_DIRS})
+    find_library(EPOLLSHIM_LIBS
```
akallabeth: not sure, but should this be `REQUIRED`?
(as in if we enter here this must be found or else abort?

cmake/PlatformDefaults.cmake
```diff
@@ -109,7 +109,12 @@ if(USE_PLATFORM_DEFAULT)
   add_compile_definitions("WINPR_TIMEZONE_FILE=\"${WINPR_TIMEZONE_FILE}\"")
 
   if(EPOLLSHIM)
-    find_path(EPOLLSHIM_INCLUDE_DIR NAMES sys/epoll.h sys/timerfd.h HINTS /usr/local/include/libepoll-shim)
-    find_library(EPOLLSHIM_LIBS NAMES epoll-shim libepoll-shim HINTS /usr/local/lib)
+    find_package(PkgConfig QUIET REQUIRED)
```
akallabeth, line 112: sure you want `pkg-config` to be required and not the `find_path` and `find_library` below?

brad0: As in require pkg-config to detect epoll-shim. If pkg-config doesn't exist then the reset would fail. Am I misunderstanding how this works? Still a CMake noob.

akallabeth: well, `pkg-config` here is just a detection helper here?
the `find_path` and `find_library` are the ones detecting the include and library related things used later on?

brad0: Well in the sense that epoll-shim is required for FreeBSD/OpenBSD and pkg-config is required to detect it. A required helper.

akallabeth: then the `find_path` and `find_library` below are useless?

akallabeth: see https://cmake.org/cmake/help/latest/module/FindPkgConfig.html on which variables are defined by `pkg_check_modules`

akallabeth:
ok, taken your commit and fixed the `pkg-config` usage. added to https://github.com/FreeRDP/FreeRDP/pull/13198

PR #13210, opened 2026-08-18
[utils] fix ChannelPduTracker partial read handling

ElCruncharino:
ChannelPduTracker_poll() treats a short read of the current chunk as
corrupted data instead of waiting for the rest, killing whichever
channel hits it (rdpdr, rail, cliprdr).

Reproduced against a local krdp build (https://invent.kde.org/plasma/krdp,
with unmerged changes) by copying a file over a few KB into the
clipboard: the cliprdr thread dies with "CheckEventHandle failed".
This is a FreeRDP bug, not krdp-specific.

Fix: only error when too much data arrives, not too little.

Tested up to 25MB, byte for byte.

### DIFF
--- a/libfreerdp/utils/channel_pdu_tracker.c
+++ b/libfreerdp/utils/channel_pdu_tracker.c
@@ -65,6 +65,13 @@ wStream* ChannelPduTracker_poll(ChannelPduTracker* tracker, BOOL* ok)
 	}
 
 	const size_t actual = recvSz - sizeof(CHANNEL_PDU_HEADER);
+	if (actual < header->length)
+	{
+		/* Rest of this chunk hasn't arrived over the wire yet, keep polling. */
+		*ok = TRUE;
+		return nullptr;
+	}
+
 	if (actual != header->length)
 	{
 		WLog_Print(tracker->log, WLOG_ERROR,


akallabeth:
@freerdp-bot test

akallabeth (commented):
LGTM,
@hardening your opinion?

libfreerdp/utils/channel_pdu_tracker.c
```diff
@@ -65,6 +65,13 @@ wStream* ChannelPduTracker_poll(ChannelPduTracker* tracker, BOOL* ok)
 	}
 
 	const size_t actual = recvSz - sizeof(CHANNEL_PDU_HEADER);
+	if (actual < header->length)
```
hardening, line 68: This should never happen, so if we get that, the bug is somewhere else

hardening: I mean even the `if (recvSz < sizeof(CHANNEL_PDU_HEADER))` line 54 is unexpected and should never happen (should change that to an assert of return `FALSE`). Can we have a larger view of the usage of `ChannelPduTracker` where this patch is needed ?

hardening (changes requested):

libfreerdp/utils/channel_pdu_tracker.c
ElCruncharino: That's fair.

On the usage question: `ChannelPduTracker` has three callers (cliprdr, rail, rdpdr server threads), and each calls `poll()` once per event-loop iteration with no retry loop of its own. It's the only place in the stack that reassembles a chunk arriving in pieces.

`WTSVirtualChannelRead` doesn't guarantee framed delivery, so turning that into an assert would encode an assumption the API itself doesn't make.

I went back and tried to reproduce the failure I originally saw, a multi-chunk cliprdr transfer killing the channel thread. Here's the log from when I first hit it:
```
[14:38:22:160] [DEBUG][com.freerdp.channels.cliprdr.server] - [cliprdr_server_receive_format_list]: ClientFormatList: numFormats: 2
[14:38:22:160] [DEBUG][com.freerdp.channels.cliprdr.server] - [cliprdr_server_format_list_response]: ServerFormatListResponse
[14:38:22:160] [DEBUG][com.freerdp.channels.cliprdr.server] - [cliprdr_server_format_data_request]: ClientFormatDataRequest
[14:38:22:165] [DEBUG][com.freerdp.channels.cliprdr.server] - [cliprdr_server_file_contents_request]: ServerFileContentsRequest: streamId: 0x00000001
[14:38:22:165] [DEBUG][com.freerdp.channels.cliprdr.server] - [cliprdr_server_receive_format_data_response]: CliprdrClientFormatDataResponse
[14:38:22:199] [ERROR][com.freerdp.channels.cliprdr.server] - [cliprdr_server_thread]: CheckEventHandle failed with error 13!
```
I couldn't get it to happen again on a fresh attempt, outside some krdc-specific flakiness unrelated to this code path.

Fair to call it transient rather than something `ChannelPduTracker` reliably causes. But my issue DID resolve itself in testing with the fix in this PR that I thought was probably the issue, so it wasn't exactly aimless FWIW.

While chasing it I found three other bugs in the clipboard format-negotiation path instead, and it's possible these played a role in that flakiness. Will open a separate PR.

PR #13211, opened 2026-08-18
[codec,av1] use dav1d for AV1 decoding

### LINKED ISSUE #12680: Use dav1d for AV1 decoding
# **Is your feature request related to a problem? Please describe.**

# **Describe the solution you'd like**

# **Describe alternatives you've considered**

# **Additional context**

ElCruncharino:
Closes #12680

dav1d becomes the preferred AV1 decoder with -DWITH_DAV1D=ON. libaom
stays the only encoder since dav1d doesn't encode. Falls back to libaom
decode if dav1d isn't built. Falls back further to the existing stub if
neither is built.

Also fixes a bug found while testing. WITH_LIBYUV's RGB/YUV conversion
ignored the caller's pixel format. It always assumed packed BGRA8888 and
overran the row stride for anything else. This hit encode and the old
libaom decode path too, not just dav1d.

Added AV1's first test coverage. It encodes with libaom and decodes with
whichever backend is active. That's what caught the bug above.

Verified live against freerdp-shadow encoding with libaom and xfreerdp
decoding with dav1d. About 2000 frames. No errors. Pixel-identical to
source.


akallabeth:
@freerdp-bot test

cmake/Finddav1d.cmake
```diff
@@ -0,0 +1,41 @@
...
+find_path(DAV1D_INCLUDE_DIR NAMES dav1d/dav1d.h PATH_SUFFIXES include HINTS ${DAV1D_ROOT} ${PC_DAV1D_INCLUDE_DIRS})
+find_library(DAV1D_LIBRARY NAMES dav1d PATH_SUFFIXES lib HINTS ${DAV1D_ROOT} ${PC_DAV1D_LIBRARY_DIRS})
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(DAV1D DEFAULT_MSG DAV1D_LIBRARY DAV1D_INCLUDE_DIR)
+
+if(DAV1D_INCLUDE_DIR AND DAV1D_LIBRARY)
+  set(DAV1D_FOUND ON CACHE BOOL "DAV1D")
+  set(DAV1D_LIBRARIES ${DAV1D_LIBRARY} CACHE STRING "DAV1D")
+  set(DAV1D_INCLUDE_DIRS ${DAV1D_INCLUDE_DIR} CACHE STRING "DAV1D")
```
akallabeth, line 28: preferred way to do this is creating a `IMPORTED` target that contains all the details like `pkg-config` does with`PkgConfig::<name of library>`

akallabeth (approved):
LGTM

PR #13208, opened 2026-08-18
[scard,pack] fix NDR pointer validation for empty card handles

joantolo:
When unpacking REDIR_SCARDHANDLE, the NDR pointer was passed as nullptr to smartcard_ndr_pointer_read_(), preventing it from accepting zero-length handles. This is inconsistent with how smartcard_unpack_redir_scard_context_() handles the same scenario. Capture the NDR pointer value and validate consistency between cbHandle and pbHandleNdrPtr, matching the pattern used for context unpacking.

Issue found when testing smartcard redirection from gnome-remote-desktop with mstsc client redirecting ([ref](https://gitlab.gnome.org/GNOME/gnome-remote-desktop/-/merge_requests/406#note_2845101)).

PR #13215, opened 2026-08-19
[scard,pack] remove incorrect stream length checks in unpack functions

joantolo:
Remove premature Stream_CheckAndLogRequiredLengthWLogEx checks from both smartcard_unpack_redir_scard_context_() and smartcard_unpack_redir_scard_handle_(). These checks validate cbContext/cbHandle bytes are available, but the actual data is read later in the deferred ref phase. The proper validation happens in the _ref functions right before reading.

This adds the missing fix I tried to add after comment: https://github.com/FreeRDP/FreeRDP/pull/13208#discussion_r3806608821


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13219, opened 2026-08-19
fix use-after-free when winpr_aligned_recalloc fails

metsw24-max:
winpr_aligned_offset_recalloc frees the caller's original block on the allocation-failure path (the goto fail branch runs winpr_aligned_free before returning null), unlike winpr_aligned_offset_realloc just above it which returns null and leaves the block intact. every caller follows the tmp = winpr_aligned_recalloc(p, ...); if (!tmp) return idiom, so a failed grow leaves them holding a freed pointer and a stale length; nsc_context_initialize (bmp.width*bmp.height for an nscodec surface come straight off the wire) and the clearcodec glyph/temp/vbar buffers then reuse that pointer, giving a use-after-free write of attacker-supplied pixels plus a double free when the codec context is torn down. the fix returns null without touching memblock on failure so the behaviour matches realloc and the windows _aligned_recalloc contract. TestAlignment now forces the failing path and checks the original block still holds its contents.

### DIFF
--- a/winpr/libwinpr/crt/alignment.c
+++ b/winpr/libwinpr/crt/alignment.c
@@ -220,7 +220,7 @@ void* winpr_aligned_offset_recalloc(void* memblock, size_t num, size_t size, siz
 	newMemblock = winpr_aligned_offset_malloc(size * num, alignment, offset);
 
 	if (!newMemblock)
-		goto fail;
+		return nullptr;
 
 	pNewMem = WINPR_ALIGNED_MEM_STRUCT_FROM_PTR(newMemblock);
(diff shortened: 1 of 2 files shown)


winpr/libwinpr/crt/alignment.c
akallabeth: no longer used?

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/13224

PR #13220, opened 2026-08-19
[client,x11] clear errno before keyboard pipe read

### LINKED ISSUE #13218: [client,x11] /kbd:pipe exits on stale errno after FIFO EOF
`/kbd:pipe` can exit immediately on macOS and remove its FIFO.

The FIFO `read()` returns 0, but `errno` can still contain `ETIMEDOUT`
from an earlier operation. The current code interprets that stale value
as a read error.

Expected behavior: the FIFO remains available until the RDP session
ends.

gornkv:
`read()` can return 0 without clearing `errno`. On macOS, a stale
`ETIMEDOUT` caused the keyboard pipe reader to exit and the FIFO to be
removed.

Clear `errno` before reading so the existing EOF handling retries as
intended.

Closes #13218.

Tested by building `xfreerdp-client` with `WITH_X11=ON` on macOS.

### DIFF
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -1314,6 +1314,7 @@ static BOOL xf_process_pipe(rdpContext* context, const char* pipe)
 	while (!freerdp_shall_disconnect_context(context))
 	{
 		char buffer[64] = WINPR_C_ARRAY_INIT;
+		errno = 0;
 		ssize_t rd = read(fd, buffer, sizeof(buffer) - 1);
 		if (rd == 0)
 		{


akallabeth (changes requested):
wrong fix.
just set `errno = 0;` before `read`

akallabeth:
and please, leave out the AI generated filler stuff, if I want to read opengroup specs I'll do so, having these attached to every issue/pull request is annoying.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13222, opened 2026-08-19
[winpr,sspi] add custom extensions

akallabeth:


MartinHaimberger (approved):
Looks good to me. +1

winpr/libwinpr/sspi/NTLM/ntlm.c
```diff
@@ -519,6 +519,22 @@ static SECURITY_STATUS SEC_ENTRY ntlm_AcquireCredentialsHandleW(
 
 	if (settings)
 	{
+		const size_t size = sizeof(SEC_WINPR_NTLM_SETTINGS);
+		if (settings->size != size)
+		{
+			switch (settings->size)
+			{
+				case 72:
```
hardening: Perhaps that should that be set as a constant somewhere ?

akallabeth: replaced by a check for size >= sizeof(struct)

PR #13223, opened 2026-08-19
[codec,yuv] fix pool_decode_rect loop variable reuse

ondrejholy:
Currently, `pool_decode_rect` uses `waitCount` both as the `for` loop
variable and as the worker slot index. When `numRegionRects` exceeds
`work_object_count`, resetting `waitCount` to 0 restarts the loop from
the beginning instead of continuing with the remaining rectangles,
causing it to never terminate. This was introduced by commit https://github.com/FreeRDP/FreeRDP/commit/1cc64eb58b2f5ce9e4ad8aa8610e085f696e578c.
Let's use a separate loop variable `x` to iterate over `regionRects`
so that `waitCount` only tracks the worker slot position.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/libfreerdp/codec/yuv.c
+++ b/libfreerdp/codec/yuv.c
@@ -558,7 +558,7 @@ static BOOL pool_decode_rect(YUV_CONTEXT* WINPR_RESTRICT context, BYTE type,
 	}
 
 	/* case where we use threads */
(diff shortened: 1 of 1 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):

PR #13225, opened 2026-08-19
[channels,cliprdr] fix server-side format-list negotiation bugs

ElCruncharino:
Two related bugs in server-side clipboard format-list negotiation, found while investigating #13210 (unrelated to it).

**1. `cliprdr_packet_format_list_new()` write length mismatch**

Clamps the destination to 16 WCHARs in short-format-name mode but still passes the unclamped UTF-8 source length into `Stream_Write_UTF16_String_From_UTF8`. Any format name over 15 characters broke the packet build against a short-name peer. `FileGroupDescriptorW` is 20 characters, so it hit this every time.

Checked directly: built the function standalone with a 20-character format name in short-name mode. Fails with "insufficient buffer supplied" before the fix, builds a correct 44-byte packet with the name safely truncated after.

**2. `cliprdr_server_format_data_request()` buffer sizing**

Sized its wire buffer from caller-supplied `msgFlags`/`dataLen`, inconsistent with the client-side sender, which hardcodes `0`/`4`. The payload is always exactly one UINT32 (`requestedFormatId`), so a caller that left `dataLen` at its zero-initialized default got a stream 4 bytes too small for the `Stream_Write_UINT32` that follows.

Checked the same way: reproduced the exact stream-construction sequence. Zero bytes free where 4 are needed before the fix, 4 free after.

**End to end:** built `freerdp-shadow-cli` from this branch, connected with a stock `xfreerdp` client, and transferred a real file over the clipboard on a live RDP session. Came through byte for byte, and format names actually negotiated long against a real client, no workarounds needed.

### DIFF
--- a/channels/cliprdr/server/cliprdr_main.c
+++ b/channels/cliprdr/server/cliprdr_main.c
@@ -1476,9 +1476,6 @@ CliprdrServerContext* cliprdr_server_context_new(HANDLE vcm)
 	cliprdr->log = WLog_Get(TAG);
 	cliprdr->vcm = vcm;
 
-	/* negotiation below only narrows this, never turns it on */
-	context->useLongFormatNames = TRUE;
-
 	return context;
 
 fail:
--- a/channels/cliprdr/server/cliprdr_main.c
+++ b/channels/cliprdr/server/cliprdr_main.c
@@ -1476,6 +1476,9 @@ CliprdrServerContext* cliprdr_server_context_new(HANDLE vcm)
 	cliprdr->log = WLog_Get(TAG);
 	cliprdr->vcm = vcm;
 
+	/* negotiation below only narrows this, never turns it on */
+	context->useLongFormatNames = TRUE;
+
 	return context;
 
 fail:
--- a/channels/cliprdr/server/cliprdr_main.c
+++ b/channels/cliprdr/server/cliprdr_main.c
@@ -338,8 +338,8 @@ static UINT cliprdr_server_format_data_request(CliprdrServerContext* context,
 		WLog_Print(cliprdr->log, WLOG_WARN, "called with invalid type %08" PRIx32,
 		           formatDataRequest->common.msgType);
 
-	wStream* s = cliprdr_packet_new(CB_FORMAT_DATA_REQUEST, formatDataRequest->common.msgFlags,
-	                                formatDataRequest->common.dataLen);
+	/* payload is always one UINT32, hardcode size like the client-side sender */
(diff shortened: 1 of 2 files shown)


channels/cliprdr/server/cliprdr_main.c
```diff
@@ -1476,6 +1476,9 @@ CliprdrServerContext* cliprdr_server_context_new(HANDLE vcm)
 	cliprdr->log = WLog_Get(TAG);
 	cliprdr->vcm = vcm;
 
+	/* negotiation below only narrows this, never turns it on */
+	context->useLongFormatNames = TRUE;
```
pnowack: No, this is server implementation specific, and not FreeRDP specific. If you want to advertise `CB_USE_LONG_FORMAT_NAMES` (or any capability), you set that in your server implementation after creating the context and not in FreeRDP for any other FreeRDP based server too.

This applies to other capabilities (like e.g. `CB_CAN_LOCK_CLIPDATA`, `CB_STREAM_FILECLIP_ENABLED` , etc.) too.
Depending on how you do your CLIPRDR implementation you still need to check what the client here reports. Clients are not obligated to support long format names, and clients are not obligated to even send their capabilities PDU in CLIPRDR (the capabilities PDU was only added in later versions by Microsoft).

ElCruncharino: Fair, and makes sense once you point it out. I'd noticed none of the server implementations I looked at set this before Start(), so it always negotiated short names in practice. I read that as a gap in the shared context rather than something every implementation is expected to opt into itself. Reverted and updated the PR description to match.

I'll set it explicitly in my own server implementation instead.

pnowack: > I'd noticed none of the server implementations I looked at set this before Start()

That would be wrong, because `Start()` already sends the current set server capabilities, then sends the MonitorReady PDU, then sets up the channel thread, where all received CLIPRDR PDUs from the client are handled. Thus, the capabilities, that the server wants to advertise, have to be set before calling that function.

Also see `3.2.5.1.3 Sending a Client Clipboard Capabilities PDU`:
```
If the Server Capabilities store indicates that the server does not support a particular feature, then the
client SHOULD NOT indicate support for that feature.
```
Clients that send their clipboard capabilities are "allowed" to set capabilities that the server side did not announce, but they should not to. So, always set on your server side what you want to use in the first place.

akallabeth (approved):
does not hurt.

akallabeth:
@freerdp-bot test

PR #13235, opened 2026-08-20
Windows DOS file attributes support for Linux

jlabrous:
Add Linux filesystem support for preserving Windows DOS attributes when FreeRDP creates or updates files. The implementation handles filesystems that store attributes in extended attributes and FAT filesystems that expose the attributes through FAT_IOCTL_SET_ATTRIBUTES.
Supported filesystems are ntfs, cifs, fat/vfat and fall-back with user.DOSATTRIB (used by WineHQ and smbserver)


akallabeth:
@freerdp-bot test

winpr/CMakeLists.txt
```diff
@@ -251,6 +251,8 @@ if(NOT IOS)
   check_include_files(sys/sockio.h WINPR_HAVE_SYS_SOCKIO_H)
   check_include_files(syslog.h WINPR_HAVE_SYSLOG_H)
   check_include_files(sys/select.h WINPR_HAVE_SYS_SELECT_H)
+  check_include_files(sys/xattr.h WINPR_HAVE_SYS_XATTR_H)
```
akallabeth, line 254: better use [CheckSourceCompiles](https://cmake.org/cmake/help/latest/module/CheckSourceCompiles.html#module:CheckSourceCompiles) here. There is an incompatible function definition on mac os (and maybe others) that will fail the compilation.

akallabeth (changes requested):

akallabeth:
thank you for your pull request.

I've taken your commit, fixed the detection of the feature (breaks on osx) and cleaned up commit history (merges) with https://github.com/FreeRDP/FreeRDP/pull/13239

winpr/CMakeLists.txt
jlabrous: I saw that ! I will take care in source code ...

PR #13236, opened 2026-08-20
[client,x11] rail: re-send work area when it changes mid-session

oraturk75:
## What this fixes

`xf_rail_send_workarea()` is called once, at RAIL activation. The comment on the current code states plainly that a mid-session `_NET_WORKAREA` change is not re-sent.

The result is that after any change to the host's work area — panel resized or moved, resolution change, monitor hotplug, desktop switch — the server keeps maximizing RemoteApp windows to a work area that no longer exists, for the remainder of the session. Maximized windows either overlap the panel or stop short of the screen edge.

This is easy to hit on a long-lived session and there is currently no way to recover from it short of reconnecting.

## What it does

Subscribes to `PropertyChangeMask` on the root window — preserving any mask already selected there, since the clipboard code also selects on the root window — and watches `_NET_WORKAREA` and `_NET_CURRENT_DESKTOP`.

`_NET_CURRENT_DESKTOP` is included because `_NET_WORKAREA` is indexed by the current desktop, so switching desktops changes the effective work area while the property itself is unchanged.

Sends are coalesced rather than emitted per event. Two reasons:

1. Window managers publish several distinct, individually valid work areas while a panel is being reconfigured. On Cinnamon, restoring a 44px panel publishes the panel-less geometry first, then the settled one.
2. The server's window-geometry reply is asynchronous. A reply to an earlier send can arrive after a later send has already gone out, and the client then applies the superseded geometry. Observed directly: two sends 83 ms apart, the reply to the first arriving 102 ms after the second, leaving the window at the stale size.

So a one-shot timer is re-armed on every relevant `PropertyNotify` and only the value that survives the settle interval is sent. The timer runs in mainloop context via `freerdp_timer_add(..., mainloop=true)`, so the Xlib reads and the channel write stay on the main thread.

Because the timer disarms on fire and needs a fresh event plus a full interval to fire again, the settle interval is also the guaranteed minimum spacing between sends. 300 ms leaves roughly 60% headroom over the ~185 ms reply latency measured on this setup. A value equal to the last one successfully sent is suppressed outright.

**Known bound:** this holds while the server's geometry-reply latency stays below the settle interval. A server slow enough to exceed 300 ms could still invert replies. Fixing that in general would need request/reply correlation that `TS_RAIL_ORDER_SYSPARAM` does not provide.

Failure to subscribe to the root-window events warns and continues — work-area tracking degrades, the session does not.

## How to test

1. Start a RemoteApp session: `xfreerdp3 /v:<host> /app:program:notepad.exe`
2. Maximize the app window. It should stop at the panel edge.
3. Change your panel's height (or move it to another screen edge).
4. Maximize again — the window should respect the new work area. On master it uses the old one.
5. Repeat with a resolution change, and with a desktop switch on a multi-desktop WM.

With `/log-level:DEBUG` the client logs each decision:

```
Deferring work area 0,0 1664x936: still settling
Deferring work area 0,0 1664x892: still settling
Sending work area 0,0 1664x892
```

The transient value is visible as deferred and never sent. Repeating an unchanged value logs `Suppressing work area ...: unchanged`.

Tested on Ubuntu 26.04 / Cinnamon / X11 against Windows 11 Pro. Clipboard round-trip and a non-RAIL full-desktop connection were checked for regressions, since the change touches the root-window event mask.

### DIFF
--- a/client/X11/xf_event.c
+++ b/client/X11/xf_event.c
@@ -974,6 +974,15 @@ static BOOL xf_event_PropertyNotify(xfContext* xfc, const XPropertyEvent* event,
 	WINPR_ASSERT(xfc);
 	WINPR_ASSERT(event);
 
+	const Window root = DefaultRootWindow(xfc->display);
+	if ((event->window == root) &&
+	    ((event->atom == xfc->NET_WORKAREA) || (event->atom == xfc->NET_CURRENT_DESKTOP)))
+	{
+		if (!xfc->remote_app || !xfc->rail)
+			return TRUE;
+		return xf_rail_schedule_workarea(xfc);
+	}
+
 	/*
 	 * This section handles sending the appropriate commands to the rail server
 	 * when the window has been minimized, maximized, restored locally
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -29,6 +29,7 @@
 #include <winpr/print.h>
 
 #include <freerdp/client/rail.h>
+#include <freerdp/timer.h>
 
 #include "xf_window.h"
 #include "xf_rail.h"
@@ -36,6 +37,14 @@
 
 #include <freerdp/log.h>
 #define TAG CLIENT_TAG("x11")
+/*
+ * The settle interval is also the minimum spacing between sends, because the timer disarms on fire
+ * and needs a fresh event plus a full interval to fire again. That spacing is what prevents a stale
+ * server geometry reply from overtaking the next send. 300 ms leaves roughly 60% headroom over the
+ * ~185 ms reply latency measured here, so a loaded host or server has room before replies can
+ * invert; one slower than the interval still can.
+ */
+#define WORKAREA_SETTLE_INTERVAL_NS 300000000UL
 
 static const char* error_code2str(UINT32 code)
 {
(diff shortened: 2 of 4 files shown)


akallabeth (commented):
* can you explain why you added a settle timer? (the explanation regarding superseded updates is not really convincing)
* why don't  you just listen for the corresponding [ConfigureNotify](https://tronche.com/gui/x/xlib/events/window-state-change/configure.html) event and send it there?

akallabeth:
@oraturk75 ok, thank you for your pull request.

I've taken your commit and refactored some things to better fit the rest of the code with https://github.com/FreeRDP/FreeRDP/pull/13238

I've removed the timer as it complicates things (and, as you already mentioned there are high latency links).
did work so far here (debian13/gnome), hope on your end as well?

oraturk75:
Thanks, and thanks for taking it further. The refactor is clearly better placed than what I had.

**On the settle timer.** I added it because on Cinnamon, restoring a panel publishes two distinct work areas in one burst, and the second send raced the server's reply to the first:

```
29.864  -> send work area 936   (transient, panel-less)
29.947  -> send work area 892   (settled)
30.049  <- server WindowUpdate windowHeight=936
30.084  ConfigureNotify h=936   -- stale value applied
```

The window then stayed at 936 for the rest of the session. What I missed is *why* it stayed: it had been promoted to `_NET_WM_STATE_FULLSCREEN`, because a work area equal to the full screen makes maximized and fullscreen coincide, and that state is sticky.

I re-ran the same case on your version just now (3f40fa109, Cinnamon/X11, Win11 Pro). The inversion still occurs, but it recovers:

```
10:434  -> send 936
10:509  -> send 888
10:588  <- server WindowUpdate 936
10:613  ConfigureNotify h=936   -- stale, as before
10:842  <- server WindowUpdate 890   -- server corrects
10:843  ConfigureNotify h=888   -- settles correctly
```

So roughly 230 ms of wrong geometry rather than a stuck window. The timer was treating a symptom of the sticky fullscreen promotion, not the race itself. Your version is fine here. No stuck window, no crash, and I'd rather have the lower latency.

**On ConfigureNotify.** It does not fire for the case this is mostly about. A panel resize doesn't change the root window's geometry, only `_NET_WORKAREA`. Measured with `xev -root` on Cinnamon, resizing the panel 48 -> 64 -> 48:

```
15 PropertyNotify   (3 of them _NET_WORKAREA)
 0 ConfigureNotify
```

`ConfigureNotify` would catch resolution changes and monitor hotplug, but would miss panel changes entirely.

Also confirmed working on my end: panel 48 -> 64 -> 48 with a maximized RemoteApp, correct final geometry, clipboard fine, no assertion.

PR #13247, opened 2026-08-24
[client] rail: add an opt-in launch FIFO for additional RemoteApps

### LINKED ISSUE #498: RAIL: Reuse an existing RemoteApp connection when opening another app on the same server
Currently every time a new RDP connection is started even though applications from a single server are requested. It should be possible to have only one connection for multiple Remote Apps.

oraturk75:
Part of #498.

## What this changes

On a server configured for one interactive session, starting a second `xfreerdp
/app:` takes the session from the first. Existing applications disappear and
return on the new connection, making multiple RemoteApp launches disruptive.

This adds a control path to the existing connection, so later launches send
another Client Execute instead of opening a new connection. MS-RDPERP permits
further Client Execute PDUs after `ExecuteApp`.

**Off by default.** The X11 client enables it with `+rail-multi-exec`, only in
RemoteApp mode.

The initial version used a Unix socket, follower mode and request/response
protocol. Following @akallabeth's review, this current version uses a reusable
common FIFO component with fire-and-forget records. It is roughly 1800 lines
smaller, with no follower, response framing, correlation or launch queue.

## FIFO protocol

`client/common/rail_ipc.c` owns a named FIFO and calls `RailClientContext::ClientExecute`
when a record arrives. Only X11 is wired in this PR, but another client can reuse
the common component with its own adapter.

The FIFO is `$XDG_RUNTIME_DIR/freerdp-rail-v1-<sha256>.fifo`, mode 0600. The key
hashes a versioned, length-prefixed serialization of the settings that identify
a session. It excludes credentials, presentation settings and values learned
after connecting.

A record is labelled UTF-8 lines terminated by one empty line:

```
program=||notepad
working-directory=
arguments=
flags=0

```

Only `program` is required. Keys may appear in any order. Duplicate or unknown
keys reject the record, and `flags` accepts decimal or `0x` hexadecimal. After
malformed input, the parser discards through the next empty line and continues.

Submit the complete record in one `write()` no larger than the FIFO's
`_PC_PIPE_BUF`, so concurrent writers cannot interleave:

```sh
printf '%s\n\n' 'program=||notepad' > "$fifo"
```

## Safety and lifetime

The runtime directory is opened once and checked for ownership and permissions,
and later operations are relative to that descriptor. The FIFO is validated
before and after opening, then removed at teardown only if its identity still
matches.

A `flock()` on a sidecar `.lock` file elects one primary `xfreerdp` process for
each session key. A second opted-in process exits before `freerdp_connect()`,
avoiding a competing connection. The lock file remains because replacing its
locked inode creates a race; the kernel releases the lock when the primary exits.

Records remain in the kernel FIFO until the RAIL channel is ready. Reads are
bounded so FIFO traffic cannot starve the RDP connection, and they pause during
desktop resynchronization or channel loss.

## Launch behavior

A write tells the writer that the kernel accepted the bytes, and nothing more.
Server and local failures appear in the FreeRDP log.

Responses are intentionally out of scope. Returning an Execute result to the
originating writer would require a separate return path, correlation, and
additional lifetime and failure handling, reintroducing much of the complexity
removed from the earlier prototype. It can be evaluated separately if a concrete
caller needs it; this PR makes no commitment to that extension.

Normal RemoteApp mode keeps its existing behavior: a rejected launch ends the
session. With `+rail-multi-exec`, a rejection is logged and the session remains
available for later records, even if no application has launched yet. This
policy follows the selected mode rather than whether an earlier launch succeeded.

The accepted cost is that a bad first path can leave the client connected with
no window until a later valid record arrives or the server's own session policy
ends it.

With `+rail-multi-exec`, the initial `/app:program:` is optional: `xfreerdp`
completes the RAIL preamble and waits for the first FIFO record. The shared
helper skips only an empty Client Execute; all non-empty Windows and Android
calls remain unchanged, while an empty Windows call still receives the preamble.

## Known limits

- Receiving side only. A later `xfreerdp` process using `+rail-multi-exec`
  currently exits before connecting instead of forwarding its `/app` request.
  Automatic submit-and-exit was deferred to keep this the primary-only
  implementation @akallabeth asked for.
- An `xfreerdp` process started without `+rail-multi-exec` does not participate
  and can still take the Windows session from the opted-in process. Launchers
  relying on connection reuse must enable the option consistently.
- Values cannot contain LF or NUL.
- Records are capped by `_PC_PIPE_BUF` (4096 bytes on Linux), below RAIL's
  16000-byte argument limit.
- POSIX only; construction reports unsupported elsewhere.
- The opt-in is an X11 command-line option rather than a FreeRDP setting.

## Testing

The common component test now builds without linker wrapping or a visibility
override. Real conditions cover descriptor exhaustion, read failure, wait
failure and the per-pass byte budget.

Synthetic ownership changes, forced failure of an allocation-only WinPR call,
the validation-to-open FIFO substitution, and a fictional `_PC_PIPE_BUF` were
removed rather than preserved behind production test seams. The production
defenses remain unchanged. Deterministic coverage of the individual ownership
clauses and the validation-to-open recheck is no longer claimed.

The X11 wrappers remain because they substitute for unavailable X and RDP
servers. Removing each one was verified to break its intended test.

The seven `client/common` and `client/X11` tests pass normally and under
AddressSanitizer with no leaks. The changed sources build without new warnings;
`clang-format`, `cmake-format` and `git diff --check` are clean.

Live against Windows 11 with one interactive session:

- an empty initial program completed the RAIL handshake without sending a
  Client Execute;
- a nonexistent first program was rejected without ending the connection;
- three applications then launched through the FIFO over one TCP connection
  and one guest session, without replacing the earlier windows or their owning
  process;
- another bad launch was contained and the following launch still worked;
- teardown removed the FIFO and retained the sidecar lock.

### DIFF
--- a/client/X11/test/CMakeLists.txt
+++ b/client/X11/test/CMakeLists.txt
@@ -1,0 +1,45 @@
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  set(MODULE_NAME "TestXfRailIpc")
+
+  add_executable(${MODULE_NAME} TestXfRailIpc.c)
+  target_include_directories(${MODULE_NAME} PRIVATE .. ../../../channels/rail/client)
+  target_link_libraries(${MODULE_NAME} PRIVATE freerdp winpr freerdp-client xfreerdp-client)
+  set_target_properties(${MODULE_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${TESTING_OUTPUT_DIRECTORY}")
+
+  add_test(NAME ${MODULE_NAME} COMMAND ${TESTING_OUTPUT_DIRECTORY}/${MODULE_NAME})
+  set_property(TARGET ${MODULE_NAME} PROPERTY FOLDER "FreeRDP/Client/X11/Test")
+
+  target_compile_options(${MODULE_NAME} PRIVATE "-fvisibility=default")
+  target_link_options(
+    ${MODULE_NAME} PRIVATE "-Wl,--wrap=XGetWindowAttributes" "-Wl,--wrap=XSelectInput" "-Wl,--wrap=XGetWindowProperty"
+  )
+
+  set(MODULE_NAME "TestXfClientRailIpc")
+
+  add_executable(${MODULE_NAME} TestXfClientRailIpc.c)
+  target_include_directories(${MODULE_NAME} PRIVATE ..)
+  target_link_libraries(${MODULE_NAME} PRIVATE freerdp winpr freerdp-client xfreerdp-client)
+  set_target_properties(${MODULE_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${TESTING_OUTPUT_DIRECTORY}")
+  target_link_options(
+    ${MODULE_NAME}
+    PRIVATE
+    "-Wl,--wrap=freerdp_connect"
+    "-Wl,--wrap=freerdp_disconnect"
+    "-Wl,--wrap=freerdp_get_event_handles"
+    "-Wl,--wrap=freerdp_check_event_handles"
+    "-Wl,--wrap=XPending"
+  )
+
+  add_test(NAME ${MODULE_NAME} COMMAND ${TESTING_OUTPUT_DIRECTORY}/${MODULE_NAME})
+  set_property(TARGET ${MODULE_NAME} PROPERTY FOLDER "FreeRDP/Client/X11/Test")
+endif()
+
+set(MODULE_NAME "TestXfCmdLine")
+
+add_executable(${MODULE_NAME} TestXfCmdLine.c)
+target_include_directories(${MODULE_NAME} PRIVATE ..)
+target_link_libraries(${MODULE_NAME} PRIVATE freerdp winpr freerdp-client xfreerdp-client)
+set_target_properties(${MODULE_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${TESTING_OUTPUT_DIRECTORY}")
+
+add_test(NAME ${MODULE_NAME} COMMAND ${TESTING_OUTPUT_DIRECTORY}/${MODULE_NAME})
+set_property(TARGET ${MODULE_NAME} PROPERTY FOLDER "FreeRDP/Client/X11/Test")
--- a/client/X11/xf_client.c
+++ b/client/X11/xf_client.c
@@ -1590,7 +1590,6 @@ static DWORD WINAPI xf_client_thread(LPVOID param)
 	freerdp* instance = (freerdp*)param;
 	WINPR_ASSERT(instance);
 
-	const BOOL status = freerdp_connect(instance);
 	rdpContext* context = instance->context;
 	WINPR_ASSERT(context);
 	xfContext* xfc = (xfContext*)instance->context;
@@ -1598,6 +1597,22 @@ static DWORD WINAPI xf_client_thread(LPVOID param)
 
 	rdpSettings* settings = context->settings;
 	WINPR_ASSERT(settings);
+	BOOL status = FALSE;
+
+	if (xfc->railMultiExec && freerdp_settings_get_bool(settings, FreeRDP_RemoteApplicationMode) &&
+	    !freerdp_settings_get_bool(settings, FreeRDP_AuthenticationOnly))
+	{
+		/* Execute failures are nonfatal in this mode, so a connected session must always
+		 * retain the FIFO as a working submission path. */
+		xfc->railIpc = freerdp_client_rail_ipc_new(settings, xfc->log);
+		if (!xfc->railIpc)
+		{
+			exit_code = XF_EXIT_PRE_CONNECT_FAILED;
+			goto end;
+		}
+	}
+
+	status = freerdp_connect(instance);
 
 	if (!status)
 	{
@@ -1637,6 +1652,10 @@ static DWORD WINAPI xf_client_thread(LPVOID param)
 		HANDLE handles[MAXIMUM_WAIT_OBJECTS] = WINPR_C_ARRAY_INIT;
 		DWORD nCount = 0;
 		handles[nCount++] = inputEvent;
+
+		const HANDLE railIpcEvent = freerdp_client_rail_ipc_get_event(xfc->railIpc);
+		if (railIpcEvent)
+			handles[nCount++] = railIpcEvent;
 
 		/*
 		 * win8 and server 2k12 seem to have some timing issue/race condition
@@ -1694,6 +1713,9 @@ static DWORD WINAPI xf_client_thread(LPVOID param)
 			}
 		}
 
+		if (!freerdp_client_rail_ipc_check_event(xfc->railIpc))
+			WLog_Print(xfc->log, WLOG_ERROR, "RAIL IPC failed to process launch input");
+
 		if (!handle_window_events(instance))
 			break;
 	}
@@ -1716,6 +1738,8 @@ disconnect:
 
 	freerdp_disconnect(instance);
 end:
+	freerdp_client_rail_ipc_free(xfc->railIpc);
+	xfc->railIpc = nullptr;
 	ExitThread(exit_code);
 	return exit_code;
 }
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -1049,6 +1049,10 @@ xf_rail_monitored_desktop(WINPR_ATTR_UNUSED rdpContext* context,
 		WLog_Print(xfc->log, WLOG_WARN, "WINDOW_ORDER_TYPE_DESKTOP flag missing!");
 		return FALSE;
 	}
+	if ((orderInfo->fieldFlags &
+	     (WINDOW_ORDER_FIELD_DESKTOP_ARC_BEGAN | WINDOW_ORDER_FIELD_DESKTOP_HOOKED)) &&
+	    !freerdp_client_rail_ipc_set_ready(xfc->railIpc, FALSE))
+		WLog_Print(xfc->log, WLOG_ERROR, "RAIL IPC failed to revoke launch readiness");
 
 	if ((orderInfo->fieldFlags & WINDOW_ORDER_FIELD_DESKTOP_ARC_BEGAN) &&
 	    (orderInfo->fieldFlags & WINDOW_ORDER_FIELD_DESKTOP_HOOKED))
@@ -1072,15 +1076,13 @@ xf_rail_monitored_desktop(WINPR_ATTR_UNUSED rdpContext* context,
 		if (!xf_rail_select_workarea_events(xfc))
 			return FALSE;
 
-		const char* app =
-		    freerdp_settings_get_string(context->settings, FreeRDP_RemoteApplicationProgram);
-		if ((app != nullptr) && (strnlen(app, 1) > 0))
-		{
-			if (client_rail_server_start_cmd(xfc->rail) != CHANNEL_RC_OK)
-				return FALSE;
-			if (xf_rail_send_workarea(xfc) != CHANNEL_RC_OK)
-				return FALSE;
-		}
+		if (client_rail_server_start_cmd(xfc->rail) != CHANNEL_RC_OK)
+			return FALSE;
+		if (xf_rail_send_workarea(xfc) != CHANNEL_RC_OK)
+			return FALSE;
+
+		if (!freerdp_client_rail_ipc_set_ready(xfc->railIpc, TRUE))
+			WLog_Print(xfc->log, WLOG_ERROR, "RAIL IPC failed to grant launch readiness");
 	}
 	if (orderInfo->fieldFlags & WINDOW_ORDER_FIELD_DESKTOP_ACTIVE_WND)
 	{
@@ -1112,6 +1114,8 @@ static BOOL xf_rail_non_monitored_desktop(rdpContext* context,
 		WLog_Print(xfc->log, WLOG_WARN, "TODO: implement WINDOW_ORDER_TYPE_DESKTOP");
 		return FALSE;
 	}
+	if (!freerdp_client_rail_ipc_set_ready(xfc->railIpc, FALSE))
+		WLog_Print(xfc->log, WLOG_ERROR, "RAIL IPC failed to revoke launch readiness");
 	if (orderInfo->fieldFlags & WINDOW_ORDER_FIELD_DESKTOP_NONE)
 	{
 		WLog_Print(xfc->log, WLOG_WARN, "TODO: implement WINDOW_ORDER_FIELD_DESKTOP_NONE");
@@ -1401,6 +1405,13 @@ int xf_rail_init(xfContext* xfc, RailClientContext* rail)
 	xfc->rail = rail;
 	xf_rail_register_update_callbacks(context->update);
 	rail->custom = (void*)xfc;
+	if (!freerdp_client_rail_ipc_attach(xfc->railIpc, rail))
+	{
+		WLog_Print(xfc->log, WLOG_ERROR, "RAIL IPC failed to attach the RAIL channel");
+		rail->custom = nullptr;
+		xfc->rail = nullptr;
+		return 0;
+	}
 	rail->ServerExecuteResult = xf_rail_server_execute_result;
 	rail->ServerSystemParam = xf_rail_server_system_param;
 	rail->ServerLocalMoveSize = xf_rail_server_local_move_size;
@@ -1436,10 +1447,10 @@ fail:
 
 int xf_rail_uninit(xfContext* xfc, RailClientContext* rail)
 {
-	WINPR_UNUSED(rail);
-
 	if (xfc->rail)
 	{
+		if (!freerdp_client_rail_ipc_detach(xfc->railIpc, rail))
+			WLog_Print(xfc->log, WLOG_ERROR, "RAIL IPC failed to detach the RAIL channel");
 		xfc->rail->custom = nullptr;
 		xfc->rail = nullptr;
 	}
--- a/client/X11/xf_rail.c
+++ b/client/X11/xf_rail.c
@@ -1159,13 +1159,14 @@ static UINT xf_rail_server_execute_result(RailClientContext* context,
 
 	xfContext* xfc = (xfContext*)context->custom;
 	WINPR_ASSERT(xfc);
-
 	if (execResult->execResult != RAIL_EXEC_S_OK)
 	{
 		WLog_Print(
 		    xfc->log, WLOG_ERROR, "RAIL exec error: execResult=%s [0x%08" PRIx32 "] NtError=0x%X\n",
 		    error_code2str(execResult->execResult), execResult->execResult, execResult->rawResult);
-		freerdp_abort_connect_context(&xfc->common.context);
+
+		if (!xfc->railMultiExec)
+			freerdp_abort_connect_context(&xfc->common.context);
 	}
 
 	return CHANNEL_RC_OK;
--- a/client/common/rail_ipc.c
+++ b/client/common/rail_ipc.c
@@ -1,0 +1,1190 @@
+/**
+ * FreeRDP: A Remote Desktop Protocol Implementation
+ * Common RAIL launch IPC
+ *
+ * Copyright 2026 Tony Dursun <oraturk75@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <freerdp/config.h>
+#include <freerdp/client/rail_ipc.h>
+
+#include <errno.h>
+#include <inttypes.h>
+#include <limits.h>
+#include <stdlib.h>
+#include <string.h>
+
+#if !defined(_WIN32)
+#include <fcntl.h>
+#include <sys/file.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#endif
+
+#include <winpr/assert.h>
+#include <winpr/cast.h>
+#include <winpr/crt.h>
+#include <winpr/custom-crypto.h>
+#include <winpr/environment.h>
+#include <winpr/print.h>
+#include <winpr/string.h>
+#include <winpr/thread.h>
+
+#define RAIL_IPC_HASH_VERSION 2U
+#define RAIL_IPC_NAME_PREFIX "freerdp-rail-v1-"
+#define RAIL_IPC_NAME_CAPACITY 96U
+#define RAIL_IPC_MAX_RECORDS_PER_PASS 16U
+#define RAIL_IPC_READ_BUFFER_SIZE RAIL_IPC_MAX_RECORDS_PER_PASS
+#define RAIL_IPC_MAX_BYTES_PER_PASS (64U * 1024U)
+#define RAIL_IPC_KNOWN_EXEC_FLAGS                                                    \
+	(TS_RAIL_EXEC_FLAG_EXPAND_WORKINGDIRECTORY | TS_RAIL_EXEC_FLAG_TRANSLATE_FILES | \
+	 TS_RAIL_EXEC_FLAG_FILE | TS_RAIL_EXEC_FLAG_EXPAND_ARGUMENTS |                   \
+	 TS_RAIL_EXEC_FLAG_APP_USER_MODEL_ID)
+
+struct s_rail_client_ipc
+{
+	wLog* log;
+	DWORD ownerThreadId;
+	char* path;
+	RailClientContext* rail;
+	BOOL ready;
+	BOOL enabled;
+#if !defined(_WIN32)
+	int runtimeDirectoryFd;
+	int lockFd;
+	int readFd;
+	int heldWriteFd;
+	HANDLE readEvent;
+	char fifoName[RAIL_IPC_NAME_CAPACITY];
+	char lockName[RAIL_IPC_NAME_CAPACITY];
+	dev_t fifoDevice;
+	ino_t fifoInode;
+	uid_t fifoOwner;
+	BOOL fifoIdentityValid;
+	size_t pipeBuffer;
+	char* record;
+	size_t recordBytes;
+	BOOL atLineStart;
+	BOOL recordInvalid;
+	const char* recordError;
+#endif
+};
+
+WINPR_ATTR_NODISCARD
+static BOOL rail_ipc_is_owner_thread(const RailClientIpcContext* ipc)
+{
+	WINPR_ASSERT(ipc);
+	return ipc->ownerThreadId == GetCurrentThreadId();
+}
+
+/* The component has no internal locking; every lifecycle entry point runs on its owner thread. */
+static void rail_ipc_assert_owner_thread(const RailClientIpcContext* ipc)
+{
+	WINPR_ASSERT(rail_ipc_is_owner_thread(ipc));
+}
+
+static void rail_ipc_write_uint32_be(BYTE data[4], UINT32 value)
+{
+	data[0] = WINPR_ASSERTING_INT_CAST(BYTE, (value >> 24U) & UINT32_C(0xFF));
+	data[1] = WINPR_ASSERTING_INT_CAST(BYTE, (value >> 16U) & UINT32_C(0xFF));
+	data[2] = WINPR_ASSERTING_INT_CAST(BYTE, (value >> 8U) & UINT32_C(0xFF));
+	data[3] = WINPR_ASSERTING_INT_CAST(BYTE, value & UINT32_C(0xFF));
+}
+
+WINPR_ATTR_NODISCARD
+static BOOL rail_ipc_digest_field(WINPR_DIGEST_CTX* digest, const void* data, size_t length)
+{
+	BYTE encodedLength[4] = WINPR_C_ARRAY_INIT;
+
+	if (length > UINT32_MAX)
+		return FALSE;
+
+	rail_ipc_write_uint32_be(encodedLength, WINPR_ASSERTING_INT_CAST(UINT32, length));
+	if (!winpr_Digest_Update(digest, encodedLength, sizeof(encodedLength)))
+		return FALSE;
+
+	return (length == 0) || winpr_Digest_Update(digest, data, length);
+}
+
+WINPR_ATTR_NODISCARD
+static BOOL rail_ipc_digest_string(WINPR_DIGEST_CTX* digest, const char* value)
+{
+	if (!value)
+		return rail_ipc_digest_field(digest, nullptr, 0);
+	return rail_ipc_digest_field(digest, value, strlen(value));
+}
+
+WINPR_ATTR_NODISCARD
+static BOOL rail_ipc_digest_ascii_ci_n(WINPR_DIGEST_CTX* digest, const char* value, size_t length,
+                                       BOOL stripTrailingDot)
+{
+	char* canonical = nullptr;
+	BOOL rc = FALSE;
+
+	if (!value)
+		return rail_ipc_digest_field(digest, nullptr, 0);
+	while (stripTrailingDot && (length > 0) && (value[length - 1U] == '.'))
+		length--;
+	canonical = strndup(value, length);
+	if (!canonical)
+		return FALSE;
+	for (size_t x = 0; x < length; x++)
+	{
+		if ((canonical[x] >= 'A') && (canonical[x] <= 'Z'))
+			canonical[x] = (char)(canonical[x] + ('a' - 'A'));
+	}
+	rc = rail_ipc_digest_field(digest, canonical, length);
+	free(canonical);
+	return rc;
+}
+
+WINPR_ATTR_NODISCARD
+static BOOL rail_ipc_digest_ascii_ci(WINPR_DIGEST_CTX* digest, const char* value,
+                                     BOOL stripTrailingDot)
+{
+	return rail_ipc_digest_ascii_ci_n(digest, value, value ? strlen(value) : 0, stripTrailingDot);
+}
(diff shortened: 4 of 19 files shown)


akallabeth (changes requested):
Ok, this look promising.

What I do not quite understand are the many changes not contained in the `IPC` server component.

What you basically try to do is:

1. For the second instance:
   1. Build a hash of your command line to identify
   2. Check if there is an open session for it
   3. tell that session which application to open
   4. quit second instance
2. For the server instance
   1. Open a `IPC` instance to listen for new applications to start
   2. open if requested
   3. tear down the `IPC` on application terminate

* so while you set up the `IPC` server as I'd imagine (did not check implementation yet) you did not implement the application launch mechanism but do other stuff like configuration there as well?
* there are unrelated changes, I'd stick to purely functional ones here at first, if you have cleanup related stuff move that to a separate pr.

oraturk75:
Thanks for looking at it.

The launch mechanism is implemented in the coordinator. A follower's request is
parsed and queued in the IPC component, `xf_rail_ipc_dispatch()` pins the channel
and calls `ClientExecute` on the live connection, and
`xf_rail_ipc_server_execute_result()` matches the result back to the single
in-flight request and answers the follower, which then exits. The interposition
point is `xf_rail_client_execute()`, installed when the channel is initialised —
routing the leader's own launch through the same coordinator is what keeps one
owner for each result.

On the extra changes, you are right that not all of them belong here.
During hardening I folded in a few adjacent fixes that should have stayed
separate. I'll remove them from this PR.

**The shared builder and the public API go.** `client_rail_build_exec_order()`
and `client_rail_validate_exec_order()` in
`channels/rail/client/client_rails.c` and `include/freerdp/client/rail.h` are the
only places this patch touches code outside the X11 client. I will build the
Execute order privately in the client instead, which also restores the existing
`FreeRDP_ShellWorkingDirectory` behaviour.

**The `void` conversion of the RemoteApp-mode helpers goes too.** That one is
cleanup and does not need to be here.

**I will regroup the series into four functional commits** — transport,
coordinator, session identity and election, and tests — which should make the
shape much easier to follow than it is now.

The hooks in `xf_rail.c` and `xf_client.c` will remain: they provide the
channel-state and event-loop integration the coordinator needs to know when the
connection is usable and to have its listener in the client's wait set.

The Execute abort policy — aborting the connection only when nothing has ever
launched on it, rather than on any failed Execute — is required here, because
one bad launch must not close applications already running on a shared
connection. My preference is to keep it in this PR. If you would rather have it
as a prerequisite PR, I can split it out.

akallabeth:
@oraturk75 the code looks overly complicated.

1. best start with a server / primary only implementation
2. define a text based protocol as `IPC` (best newline separated, so there are no argument handling issues)
3. you can test the whole thing then with `echo -e arg1\narg2\narg3 >/path/to/pipe` 

you already identified correctly how to launch an application, no need to make the `IPC` complicated binary and request/response

akallabeth:
@oraturk75 also, considering ~~#9169~~ https://github.com/FreeRDP/FreeRDP/pull/13133 is quite far it would be better to move that to `client/common` so this can be reused.

oraturk75:
Understood, and agreed on both counts. I'll rework it as a reusable primary-side component in `client/common`: a 0600 FIFO in `$XDG_RUNTIME_DIR`, newline-separated records, fire-and-forget with no reply — so the second-instance side comes out of this PR entirely and the whole thing is drivable with `echo`. That drops the binary framing, the follower, the request/response correlation and the outcome plumbing.

Reading your example as three lines per launch — program, working directory, arguments — so `echo -e` writes one record. Say if you meant something else.

The `+rail-multi-exec` opt-in stays as it is in the X11 client for this PR; the common component is just callable, so SDL can wire it in however suits when it gets there. I'll update this PR when it's ready.

akallabeth:
@oraturk75 oh, and as for number of arguments, please check if you can set all options of the `RAIL_EXEC_ORDER`, I think there is also a 4th argument necessary (maybe optional) that carries the flags (as integer or maybe string parsed ?)

[edit] it would also be possible to do optional success/failure responses by just writing back some string when executing.
normal terminal uses `<ctrl>+d` or similar (e.g. `\n\n`, ...) to trigger execution/parsing of input so you can always resync if you don't know if the state machine is ready

oraturk75:
Flags as a fourth field makes sense. I will parse them as an integer.

One ambiguity remains with positional lines: the working directory and arguments may be empty, so a blank line can represent either field data or the record separator. After malformed input, the parser cannot reliably resynchronise.

I propose labelled lines terminated by an empty line:

```
program=||notepad
working-directory=
arguments=
flags=0

```

Only program is required. Omitted keys default to empty or zero. Unknown keys reject a v1 record, and an incompatible grammar would use a new FIFO version. The common case remains:

```sh
printf 'program=||notepad\n\n' > "$fifo"
```

I will leave responses out of v1 and keep it fire-and-forget. A shared FIFO cannot route a result to the writer that submitted a record. That would require a separate reply channel with correlation, or a per-writer reply pipe. I can revisit that after the simple path works.

If you prefer positional fields, please say so and I will follow that format.

akallabeth:
@oraturk75 sounds good.

oraturk75:
Force-pushed a redesigned implementation based on your feedback. The control path is now a reusable client/common FIFO component, and launch submission is fire-and-forget. This removes the follower, reply, and correlation machinery and makes the implementation roughly 1800 lines smaller. The FIFO uses labelled records with decimal or 0x hexadecimal flags, as agreed above.

include/freerdp/client/rail_ipc.h
```diff
@@ -0,0 +1,86 @@
...
+	 */
+	WINPR_ATTR_NODISCARD
+	FREERDP_API HANDLE freerdp_client_rail_ipc_get_event(const RailClientIpcContext* ipc);
+
+	/**
+	 * Process ready FIFO input on the client-loop thread.
+	 *
+	 * Detached, not-ready, and permanently disabled states are no-ops.
+	 */
+	FREERDP_API void freerdp_client_rail_ipc_check_event(RailClientIpcContext* ipc);
```
akallabeth: add failure return

include/freerdp/client/rail_ipc.h
```diff
@@ -0,0 +1,86 @@
...
+
+	/** Attach a RAIL channel on the client-loop thread. */
+	FREERDP_API void freerdp_client_rail_ipc_attach(RailClientIpcContext* ipc,
```
akallabeth: add failure return

include/freerdp/client/rail_ipc.h
```diff
@@ -0,0 +1,86 @@
...
+	                                                RailClientContext* rail);
+
+	/** Detach the current RAIL channel on the client-loop thread. */
+	FREERDP_API void freerdp_client_rail_ipc_detach(RailClientIpcContext* ipc,
+	                                                RailClientContext* rail);
+
+	/** Set whether the attached RAIL channel can accept Client Execute orders. */
+	FREERDP_API void freerdp_client_rail_ipc_set_ready(RailClientIpcContext* ipc, BOOL ready);
```
akallabeth: add failure return

client/common/rail_ipc.c
```diff
@@ -0,0 +1,1153 @@
...
+
+static void rail_ipc_write_uint32_be(BYTE data[4], UINT32 value)
+{
+	data[0] = WINPR_ASSERTING_INT_CAST(BYTE, (value >> 24U) & UINT32_C(0xFF));
+	data[1] = WINPR_ASSERTING_INT_CAST(BYTE, (value >> 16U) & UINT32_C(0xFF));
+	data[2] = WINPR_ASSERTING_INT_CAST(BYTE, (value >> 8U) & UINT32_C(0xFF));
+	data[3] = WINPR_ASSERTING_INT_CAST(BYTE, value & UINT32_C(0xFF));
+}
+
+static BOOL rail_ipc_digest_field(WINPR_DIGEST_CTX* digest, const void* data, size_t length)
```
akallabeth, line 107: suggest to use `WINPR_ATTR_NODISCARD` for all functions returning a value (except some rare cases where the return check is optional)

client/common/rail_ipc.c
```diff
@@ -0,0 +1,1153 @@
...
+	rail_ipc_assert_owner_thread(ipc);
+	if (ipc->rail != rail)
+		return;
+	ipc->ready = FALSE;
+	ipc->rail = nullptr;
+}
+
+void freerdp_client_rail_ipc_set_ready(RailClientIpcContext* ipc, BOOL ready)
+{
+	if (!ipc)
```
akallabeth, line 1171: assert or add a failure return

client/common/rail_ipc.c
```diff
@@ -0,0 +1,1153 @@
...
+		return;
+	rail_ipc_assert_owner_thread(ipc);
+	WINPR_ASSERT(!ipc->rail || (ipc->rail == rail));
+	ipc->rail = rail;
+	ipc->ready = FALSE;
+}
+
+void freerdp_client_rail_ipc_detach(RailClientIpcContext* ipc, RailClientContext* rail)
+{
+	if (!ipc || !rail)
```
akallabeth: assert or add a failure return

client/common/rail_ipc.c
```diff
@@ -0,0 +1,1153 @@
...
+		rail_ipc_log_errno(ipc, WLOG_ERROR, "reading the launch FIFO", errno);
+		rail_ipc_disable(ipc);
+		break;
+	}
+#endif
+}
+
+void freerdp_client_rail_ipc_attach(RailClientIpcContext* ipc, RailClientContext* rail)
+{
+	if (!ipc || !rail)
```
akallabeth: assert or add a failure return

client/common/rail_ipc.c
```diff
@@ -0,0 +1,1153 @@
...
+#if defined(_WIN32)
+	return nullptr;
+#else
+	return (ipc->enabled && ipc->rail && ipc->ready) ? ipc->readEvent : nullptr;
+#endif
+}
+
+void freerdp_client_rail_ipc_check_event(RailClientIpcContext* ipc)
+{
+	if (!ipc)
```
akallabeth, line 1070: assert or add a failure return

client/common/test/CMakeLists.txt
```diff
@@ -24,3 +24,23 @@ foreach(test ${${MODULE_PREFIX}_TESTS})
...
 
 set_property(TARGET ${MODULE_NAME} PROPERTY FOLDER "FreeRDP/Client/Test")
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  add_executable(TestClientRailIpc TestClientRailIpc.c ../rail_ipc.c)
+  target_compile_options(TestClientRailIpc PRIVATE "-fvisibility=default")
+  target_link_libraries(TestClientRailIpc PRIVATE freerdp winpr)
+  target_link_options(
+    TestClientRailIpc
+    PRIVATE
```
akallabeth: what are these for?

client/common/test/CMakeLists.txt
```diff
@@ -24,3 +24,23 @@ foreach(test ${${MODULE_PREFIX}_TESTS})
 endforeach()
 
 set_property(TARGET ${MODULE_NAME} PROPERTY FOLDER "FreeRDP/Client/Test")
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  add_executable(TestClientRailIpc TestClientRailIpc.c ../rail_ipc.c)
+  target_compile_options(TestClientRailIpc PRIVATE "-fvisibility=default")
```
akallabeth: what is this for?

akallabeth (commented):
looks a lot better now.

the unit test code looks too complicated, did not read through it yet, just added notes to the cmake related stuff.

client/X11/xf_client.c
```diff
@@ -1590,14 +1590,36 @@ static DWORD WINAPI xf_client_thread(LPVOID param)
...
+	BOOL status = FALSE;
+
+	if (xfc->railMultiExec && freerdp_settings_get_bool(settings, FreeRDP_RemoteApplicationMode) &&
+	    !freerdp_settings_get_bool(settings, FreeRDP_AuthenticationOnly))
+	{
+		const char* program =
+		    freerdp_settings_get_string(settings, FreeRDP_RemoteApplicationProgram);
+		if (!program || (program[0] == '\0'))
+		{
+			WLog_ERR(TAG, "+rail-multi-exec requires a non-empty RemoteApplicationProgram");
```
akallabeth: isn't this the other way around?

with `/rail-multi-exec` no program/commandline/... needs to be provided as that can be done later using the pipe?

client/X11/xf_client.c
```diff
@@ -1668,7 +1694,11 @@ static DWORD WINAPI xf_client_thread(LPVOID param)
 		waitStatus = WaitForMultipleObjects(nCount, handles, FALSE, INFINITE);
 
 		if (waitStatus == WAIT_FAILED)
+		{
+			if (xfc->railIpc)
```
akallabeth: this looks wrong.
please add proper return checks below where the problem occurs.

client/X11/xf_client.c
```diff
@@ -1694,6 +1724,8 @@ static DWORD WINAPI xf_client_thread(LPVOID param)
 			}
 		}
 
+		freerdp_client_rail_ipc_check_event(xfc->railIpc);
```
akallabeth: return check!

client/X11/xf_rail.c
```diff
@@ -1049,6 +1049,9 @@ xf_rail_monitored_desktop(WINPR_ATTR_UNUSED rdpContext* context,
 		WLog_Print(xfc->log, WLOG_WARN, "WINDOW_ORDER_TYPE_DESKTOP flag missing!");
 		return FALSE;
 	}
+	if (xfc->railIpc && (orderInfo->fieldFlags & (WINDOW_ORDER_FIELD_DESKTOP_ARC_BEGAN |
+	                                              WINDOW_ORDER_FIELD_DESKTOP_HOOKED)))
+		freerdp_client_rail_ipc_set_ready(xfc->railIpc, FALSE);
```
akallabeth: why guard with `xfc->railIpc != nullptr`? just handle that in `freerdp_client_rail_ipc_set_ready`

client/X11/xf_rail.c
```diff
@@ -1112,6 +1117,8 @@ static BOOL xf_rail_non_monitored_desktop(rdpContext* context,
 		WLog_Print(xfc->log, WLOG_WARN, "TODO: implement WINDOW_ORDER_TYPE_DESKTOP");
 		return FALSE;
 	}
+	if (xfc->railIpc)
```
akallabeth: move the check to the function

client/X11/xf_rail.c
```diff
@@ -1159,13 +1166,16 @@ static UINT xf_rail_server_execute_result(RailClientContext* context,
...
+	if (execResult->execResult == RAIL_EXEC_S_OK)
+		xfc->railExecSucceeded = TRUE;
+	else
 	{
 		WLog_Print(
 		    xfc->log, WLOG_ERROR, "RAIL exec error: execResult=%s [0x%08" PRIx32 "] NtError=0x%X\n",
 		    error_code2str(execResult->execResult), execResult->execResult, execResult->rawResult);
-		freerdp_abort_connect_context(&xfc->common.context);
+
+		if (!xfc->railExecSucceeded)
```
akallabeth: not sure this check works here as intended.
while this is fatal for single program rails sessions it should just be a warning for multi exec sessions, the `railExecSucceeded` check is not required.

client/X11/xf_rail.c
```diff
@@ -1400,6 +1410,7 @@ int xf_rail_init(xfContext* xfc, RailClientContext* rail)
 	xfc->rail = rail;
 	xf_rail_register_update_callbacks(context->update);
 	rail->custom = (void*)xfc;
+	freerdp_client_rail_ipc_attach(xfc->railIpc, rail);
```
akallabeth: return check!

client/X11/xf_client.c
oraturk75: Yes, agreed: with /rail-multi-exec the initial program should be optional. To preserve the normal RAIL startup ordering without adding another API, I propose changing client_rail_server_start_cmd() so it always sends Client Information, optional language-bar information, and SystemParam, but skips only ClientExecute when RemoteApplicationProgram is empty. Callers with a non-empty program keep their current behavior. Does that match what you had in mind?

client/X11/xf_client.c
oraturk75: Agreed, the railIpc-specific WAIT_FAILED handling is wrong. I read your suggestion as: restore this branch to upstream behavior, make the new IPC functions return status, and check those returns at their call sites below. Is that what you intended, rather than changing WAIT_FAILED to be fatal for every X11 session?

oraturk75:
@akallabeth I am waiting for your confirmation on two review points before I revise the overlapping X11 code:

1. [Empty initial program](https://github.com/FreeRDP/FreeRDP/pull/13247#discussion_r3880415589): I proposed always sending the normal RAIL startup information, but skipping `ClientExecute` when the initial program is empty.

2. [`WAIT_FAILED` handling](https://github.com/FreeRDP/FreeRDP/pull/13247#discussion_r3880415836): I understood your comment as restoring the existing upstream behavior there and checking the new IPC return values at their actual call sites.

Your other inline comments are clear. You also mentioned that you had not reviewed the unit test code yet, so I am not treating the review as complete. These two points affect the same revision as the other X11 changes, so I would rather confirm them now than revise that code twice.

akallabeth:
@oraturk75 oh, sorry.

1. yes, sounds good.
2. yes, exactly. better check where there is a direct connection with what happened, makes debugging later easier.

akallabeth:
@oraturk75 as for the unit test, the very compiler specific `-Wl,--wrap=function` stuff looks overcomplicated for simple unit tests.
since the `IPC` stuff should be self contained anyway, there should be no need to wrap anything.

oraturk75:
@akallabeth Revision pushed. I applied the current review comments: direct
return checks, an optional initial program, mode-based Execute failure handling,
and no linker wrapping in the common IPC test. The simplified test code is also
ready for review.

I kept this as the primary-only implementation you asked for. A later `xfreerdp`
process currently detects the existing primary and exits without forwarding its
`/app` request, so launchers such as WinApps cannot use this transparently yet.

Should the submit-and-exit side be added here or in a follow-up? It does not need
the removed machinery: a later `xfreerdp` process would write one record to the
FIFO and exit on the write result, with no follower, reply, correlation or queue.
I changed `Fixes #498` to `Part of #498` until that half exists.

include/freerdp/client/rail_ipc.h
oraturk75: Done. `freerdp_client_rail_ipc_check_event()` now returns `BOOL`. Operational
FIFO failures return `FALSE` after disabling input, while inactive and optional
no-op states return `TRUE`.

include/freerdp/client/rail_ipc.h
oraturk75: Done. `freerdp_client_rail_ipc_attach()` now returns `BOOL`. Invalid or
conflicting attachments return `FALSE` without changing the current channel.

include/freerdp/client/rail_ipc.h
oraturk75: Done. `freerdp_client_rail_ipc_set_ready()` now returns `BOOL`. Enabling input
without an attached channel returns `FALSE`.

client/common/rail_ipc.c
oraturk75: Done. All value-returning internal helpers and public declarations in the new
component now use `WINPR_ATTR_NODISCARD`.

client/common/rail_ipc.c
oraturk75: Changed to a `BOOL` result. A null IPC context remains a successful no-op
because the component is optional; enabling input without an attached channel
returns `FALSE`.

client/common/rail_ipc.c
oraturk75: Changed to a `BOOL` result. A null IPC context remains a successful no-op; a
null or mismatched channel returns `FALSE` without changing the attachment.

client/common/rail_ipc.c
oraturk75: Changed to a `BOOL` result. A null IPC context remains a successful no-op; a
null or conflicting channel returns `FALSE` without changing the attachment.

client/common/rail_ipc.c
oraturk75: Changed to a `BOOL` result. A null IPC context remains a successful no-op.
Missing `ClientExecute`, wait failure, unexpected EOF and read failure return
`FALSE` after disabling FIFO input.

client/common/test/CMakeLists.txt
oraturk75: These were failure-injection wrappers. I removed all seven. The common test now
uses real conditions for descriptor exhaustion, wait failure, read failure and
the per-pass byte budget. Descriptor exhaustion and the byte budget replaced
injected conditions; the wait and read failures already used real conditions,
leaving their wrappers unused. Deterministic coverage of the individual
ownership clauses, the validation-to-open race, an allocation-only event
failure and a fictional pipe size was deliberately dropped instead of retaining
compiler-specific interception. The production checks remain unchanged.

client/common/test/CMakeLists.txt
oraturk75: Removed. The common test now builds with the project's default visibility and
no linker wrapping.

client/X11/xf_client.c
oraturk75: Implemented as agreed. The X11 non-empty-program gate is removed, and the
shared helper sends the normal RAIL preamble but skips `ClientExecute` when the
program is empty. Non-empty calls are unchanged; Android never passes an empty
program, and a Windows caller with an empty one now receives the preamble
without an Execute. The live test confirmed an empty startup reached RAILS mode
and launched its first application from the FIFO.

client/X11/xf_client.c
oraturk75: Restored this branch to the upstream behavior and removed its IPC-specific
wrapper and test. The new IPC results are now checked at the operations that
produce them, where the failure can be logged directly.

client/X11/xf_client.c
oraturk75: Done. The result is checked and logged here. An operational failure disables
the optional FIFO input but does not end the RDP session.

client/X11/xf_rail.c
oraturk75: Done. Null-context handling is inside `freerdp_client_rail_ipc_set_ready()`, so
the call is unconditional and its result is checked and logged at the readiness
transition.

client/X11/xf_rail.c
oraturk75: Done. The null-context check moved into
`freerdp_client_rail_ipc_set_ready()`, and the result is checked and logged
here.

client/X11/xf_rail.c
oraturk75: You were right, and `railExecSucceeded` is removed. Execute failures remain
fatal in normal RemoteApp mode, but are logged and nonfatal when
`+rail-multi-exec` is selected, including before the first successful launch.
The accepted cost is that a bad first path can leave `xfreerdp` connected with
no window until a later valid record arrives or the server's session policy
ends it. The cold-rejection case was also verified against the live Windows
server.

client/X11/test/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
akallabeth, line 1: why do you add this check?

client/X11/xf_rail.c
oraturk75: Done. The attach result is checked here. Failure is logged, `xfc->rail` and
`rail->custom` are rolled back, and RAIL initialization returns failure.

include/freerdp/client/rail_ipc.h
```diff
@@ -0,0 +1,93 @@
...
+	FREERDP_API void freerdp_client_rail_ipc_free(RailClientIpcContext* ipc);
+
+	/**
+	 * Create the primary-side FIFO for the session described by settings.
+	 *
+	 * Returns nullptr when the transport is unavailable, setup is unsafe, or another primary
+	 * already owns the session key. Must be called before connecting the RDP client.
+	 */
+	WINPR_ATTR_MALLOC(freerdp_client_rail_ipc_free, 1)
+	WINPR_ATTR_NODISCARD
```
akallabeth, line 51: not required, already contained in `WINPR_ATTR_MALLOC`

client/X11/test/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+  target_compile_options(${MODULE_NAME} PRIVATE "-fvisibility=default")
+  target_link_options(
+    ${MODULE_NAME} PRIVATE "-Wl,--wrap=XGetWindowAttributes" "-Wl,--wrap=XSelectInput" "-Wl,--wrap=XGetWindowProperty"
+  )
+
+  set(MODULE_NAME "TestXfClientRailIpc")
+
+  add_executable(${MODULE_NAME} TestXfClientRailIpc.c)
+  target_include_directories(${MODULE_NAME} PRIVATE ..)
+  target_link_libraries(${MODULE_NAME} PRIVATE freerdp winpr freerdp-client xfreerdp-client)
```
akallabeth, line 21: only add `xfreerdp-client`, the rest is already in the public link interface

client/X11/test/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+  add_test(NAME ${MODULE_NAME} COMMAND ${TESTING_OUTPUT_DIRECTORY}/${MODULE_NAME})
+  set_property(TARGET ${MODULE_NAME} PROPERTY FOLDER "FreeRDP/Client/X11/Test")
+
+  target_compile_options(${MODULE_NAME} PRIVATE "-fvisibility=default")
+  target_link_options(
+    ${MODULE_NAME} PRIVATE "-Wl,--wrap=XGetWindowAttributes" "-Wl,--wrap=XSelectInput" "-Wl,--wrap=XGetWindowProperty"
+  )
+
+  set(MODULE_NAME "TestXfClientRailIpc")
```
akallabeth, line 17: not sure this test is a good idea.
* it is highly compiler dependent
* it tests stuff that should already be tested with the `TestClientRailpc`

akallabeth:
@oraturk75 lets keep the secondary handling separate from this, lets get this into shape first.
Overall looking good now, the unit tests are still very compiler/platform specific, so this needs some adjustments.

akallabeth:
@freerdp-bot test

oraturk75:
@akallabeth You are right about the compiler-dependent wrapping. I inspected both X11 test targets, and I can remove all eight wrappers and both `-fvisibility=default` overrides without adding production test hooks or changing runtime behaviour.

`TestXfClientRailIpc` can use the existing public `PreConnect` callback for construction, empty-start and active-primary checks. I will use an unroutable address so unexpected network access fails loudly. `TestXfRailIpc` can set readiness directly for its Execute-policy, readiness-revocation, teardown, attach and detach checks.

Both test targets will remain, but four assertion groups have no portable replacement without a real RDP or X server:

- `freerdp_check_event_handles()` is processed before ready FIFO input in one client-loop pass;
- a component failure is contained and the X11 client loop continues;
- successful `ARC_COMPLETED` sends the preamble, work area and readiness grant in order;
- the same path with an empty initial program sends no Execute.

My preference is to drop these four assertion groups and keep both targets with zero wrappers. The component's own behaviour remains covered by its common test. The cost is that client-loop ordering or X11 startup-sequence regressions would only be caught by live testing. Does that tradeoff work for you?

I will also make the other three changes:

- Remove the redundant `WINPR_ATTR_NODISCARD` from `_new`, since `WINPR_ATTR_MALLOC` already includes it.
- Link only `xfreerdp-client`; the remaining libraries arrive through its public interface.
- Replace the Linux-only guards in both X11 and common test CMake files with `if(NOT WIN32)`. These tests require POSIX functionality and cannot run on Windows, but they do not need to be restricted only to Linux.

akallabeth:
@freerdp-bot test

akallabeth:
@oraturk75 looks like there are plenty of warnings created with this pr.
see https://ci.freerdp.com/job/code-quality-checker/13352/clang/

best compile with `-Weverything` (`clang`) and use `clang-tidy` to identify problems & fix these warnings.

oraturk75:
@akallabeth I'm sorry; I should have run both QA configurations before pushing. They are in my checklist now.

I reproduced all 69 findings and fixed the 21 outside the two wrapper-dependent X11 tests. The remaining 48 are in those two files, so they depend on your decision about the four assertion groups discussed above. Once you confirm, I will finish the cleanup, rerun both QA configurations and push everything all together.

hardening:
Using a NamedPipe from WinPR would make it run on all platforms, isn't it ?

oraturk75:
@hardening WinPR NamedPipe is something I failed to consider earlier. When @akallabeth asked for a newline protocol that could be tested with `echo -e arg1\narg2\narg3 > /path/to/pipe`, I focused only on a POSIX FIFO solution and did not evaluate WinPR NamedPipe.

That being said, I have now investigated if the current WinPR NamedPipe implementation could replace it. Here are my findings:

- On Linux, WinPR NamedPipe uses an `AF_UNIX` `SOCK_STREAM` socket. Normal shell redirection fails, while a WinPR-linked client works, so it would require a special socket-aware writer.
- Primary election breaks. The current Unix implementation accepts `FILE_FLAG_FIRST_PIPE_INSTANCE` but does not honour it, and deletes the existing endpoint before binding. In my test, a second server replaced the active primary and both reported success.
- The Unix security is not sufficient for this feature. WinPR creates `$TMPDIR/.pipe` and its endpoint with `07777` permissions and ignores the supplied security attributes, while the current FIFO is `0600` inside a private, validated `XDG_RUNTIME_DIR`.
- NamedPipe would require accepting, reading, disconnecting and reconnecting individual writers, along with multiple-client, partial-record and wait-handle management. This brings back much of the connection-management code that was previously removed. A lot of code and complexity from the earlier design that @akallabeth asked me to remove.
- The portability benefit is not available today because only the X11 client uses this component. The public API already hides the transport details, so the transport can be replaced later without changing its current callers.

Bottom line: **No, the current WinPR NamedPipe implementation is not a suitable replacement for this PR. My recommendation is to keep the POSIX FIFO.**

hardening:
> @hardening WinPR NamedPipe is something I failed to consider earlier. When @akallabeth asked for a newline protocol that could be tested with `echo -e arg1\narg2\narg3 > /path/to/pipe`, I focused only on a POSIX FIFO solution and did not evaluate WinPR NamedPipe.
> 
> That being said, I have now investigated if the current WinPR NamedPipe implementation could replace it. Here are my findings:
> 
[...]

Well clearly it's not _your_ findings it's the message returned by your AI. Perhaps next time rephrase so that we can know if it's the raw output of the AI, or if you have entirely/partially checked by yourself these assertions (for me "I asked the AI and here's the result, what do you think ?", is perfectly legitimate, but we need to know if some human eyes went the AI's assertions for a check).

Anyway correct the WinPR implementation should be improved. I kinda disagree with @akallabeth with the "echo thing", I would find it perfectly legitimate that the spoken protocol on the unix socket is not entirely trivial clear text (if you want my opinion a json per line would be a good compromise to encode the thing we need). Also any trivial python script you write is able to write in a local unix socket so the `do it with echo+ pipe` requirement doesn't apply IMHO.

I feel like the remote rail IPC must be something not fixed only for the linux, and if we go with your architecture that's clearly what's gonna happen, except if I missed something.

akallabeth:
@hardening `JSON` would also be ok as serialization format, but having a simple file to just write to makes a whole lot of things easier to work with.

oraturk75:
@hardening You are right that saying "my findings" might gave you the wrong impression. Two AI assistants independently ran the probes, and the raw output from both is saved and reproducible. However, I did not simply copy/paste raw AI output: I worked through each finding, questioned the conclusions and edited the response myself. I did not personally run the probes, and I will be more clear about who did what from now on.

From where I stand: a portable transport is the right long-term direction, and a socket may be the eventual solution we should get to. The first version here originally used a Unix socket, though a POSIX one rather than a portable transport, and it was dropped at review. The current design follows the `echo -e ... > /path/to/pipe` requirement from that review, and a FIFO is what satisfies that.

This is already the second design in this PR. I'd rather not rebuild it a third time here, again.

Let me see what small, bounded changes I can make to improve and strengthen this current PR while laying real groundwork to give the future portable transport work a head start. I will report back with a proposal.

@akallabeth I will fold this into the same revision as your review comments. The four assertion groups I asked about are still open, so let me know if you want me to drop them or if you would rather see what the proposal does with them first.

oraturk75:
@hardening @akallabeth I finished that inspection. I found one small-ish change that is worth adding.

Only one new change has clear value: extract the record decoder from the FIFO transport. The FIFO reads the incoming bytes and tells the decoder its maximum safe record size. The decoder understands and validates the text, and then produces a standard RAIL launch request. The FIFO code keeps everything specific to FIFO operation and security.

This is the actual separation that helps both the current implementation and a future transport.

- The decoder can be tested without creating or reading a FIFO.
- A future socket can send its bytes through the same decoder.
- Each transport can provide its own maximum message size.
- If the text format later becomes JSON, the parsing change has one defined home.
- Our prototype (AI assisted heavily) behaved identically to the current FIFO parser on Linux.
- It is verified on Linux; other platforms still need CI.

I also considered JSON records, but I would leave that change out of this PR. Format and transport are two separate concerns. And now with the decoder extracted, making a later change to JSON is a lot easier. Doing it now buys nothing this PR needs but only re-work, re-test (and such), so I will keep the current format. Any further consideration of JSON belongs in future portable transport work, whenever it is taken up.

I will also finish the wrapper-free X11 rewrite and the QA cleanup. Getting both targets to zero wrappers still costs the four assertion groups I mentioned before: client-loop ordering, failure containment, and the two `ARC_COMPLETED` startup paths. The component itself stays covered by its own test, but those X11 regressions would only show up in live testing. The decoder does not change any of that. I would rather drop the four groups, unless @akallabeth wants some of the coverage kept.

I also looked at a Darwin runtime-directory fallback. I left it out. No macOS client uses this component yet, and I have no way to test it there. If you want it and there is macOS testing available, I can add it.

I looked for other useful changes, but the existing design already keeps most transport-specific details contained. The decoder is all I found. I did not want to pad the list with speculative work. If you both agree, I will do the decoder and finish the wrapper and QA cleanup in one revision. The FIFO protocol, shell interface, election and security model stay as they are.

PR #13249, opened 2026-08-24
[gdi] document cache requirement for gdi_init

maordadush:
Per review feedback on the original patch: `gdi_init`/`gdi_init_ex` require the `rdpCache`, which since `c41d16a38` ("[core,cache] move cache from gdi to core") is created by the core during the connect sequence rather than by `gdi_init_ex` itself. Calling them outside that sequence dereferences a NULL cache through `brush_cache_register_callbacks()` and `bitmap_cache_register_callbacks()`.

Neither the header declarations nor the existing doxygen blocks carried any lifecycle note, so this documents the precondition on both functions.

No functional change.

### DIFF
--- a/libfreerdp/gdi/gdi.c
+++ b/libfreerdp/gdi/gdi.c
@@ -1413,6 +1413,10 @@ BOOL gdi_resize_ex(rdpGdi* gdi, UINT32 width, UINT32 height, UINT32 stride, UINT
  * @param instance A pointer to the instance to use
  * @param format The color format for the local framebuffer
  * @return \b TRUE for success, \b FALSE for failure
+ *
+ * @note Must be called after the connect sequence has completed, e.g. from the
+ * \b PostConnect callback: GDI requires the \b rdpCache created by the
+ * core during connection, after capability negotiation.
  */
 BOOL gdi_init(freerdp* instance, UINT32 format)
 {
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
`gdi_init` must be called from `PostConnect` callback anyway so this must exist here?
it is public `API` but that still does mean lifecycle requirements must be met.

maordadush:
For a connecting client, yes -- the cache exists before `PostConnect` fires, so the guard is always true there and both registrations run exactly as before. It's a no-op for every in-tree caller. Agreed the cache belongs after capability negotiation; `bitmap_cache_new()`/`glyph_cache_new()` size from negotiated settings, so `connection.c` is the right owner.

The case I hit is a consumer that never runs the connect sequence at all -- `PostConnect` never fires, so no cache is ever created. It drives only the GFX pipeline and never receives legacy bitmap/brush orders, so the caches are meaningless to it, but `gdi_graphics_pipeline_init()` takes an `rdpGdi*` and `gdi_init()` is the only public way to get one. `cache_new()` isn't exported, so that lifecycle requirement can't be met from outside libfreerdp except by connecting to a real server.

Two ways I'd be happy to land this: keep the guard (no-op for connecting clients), or drop it and document the precondition on `gdi_init`/`gdi_init_ex` -- the header carries no lifecycle note today. Which do you prefer?

akallabeth:
@maordadush well, since the cache must exist for `gdi` to work properly it is better to just document the requirement

maordadush:
Makes sense - done. Dropped the guard and replaced it with a `@note` on both `gdi_init` and `gdi_init_ex` stating that they must be called after the connect sequence completes (e.g. from `PostConnect`) because GDI requires the `rdpCache` the core creates after capability negotiation. Comment-only, no functional change.

akallabeth (approved):
LGTM

PR #13253, opened 2026-08-24
Fixes for TestFuzzServerSeedGen

ligurio:
Needed for https://github.com/google/oss-fuzz/pull/16042#issuecomment-5388469492

### DIFF
--- a/libfreerdp/core/tcp.c
+++ b/libfreerdp/core/tcp.c
@@ -1013,6 +1013,13 @@ BOOL freerdp_tcp_set_keep_alive_mode(const rdpSettings* settings, int sockfd)
 	const BOOL keepalive = (freerdp_settings_get_bool(settings, FreeRDP_TcpKeepAlive));
 	UINT32 optval = 0;
 	socklen_t optlen = 0;
+	BOOL isTcp = FALSE;
+	struct sockaddr_storage addr = WINPR_C_ARRAY_INIT;
+	socklen_t addrlen = sizeof(addr);
+
+	if (getsockname(sockfd, (struct sockaddr*)&addr, &addrlen) == 0)
+		isTcp = (addr.ss_family == AF_INET) || (addr.ss_family == AF_INET6);
+
 	optval = keepalive ? 1 : 0;
 	optlen = sizeof(optval);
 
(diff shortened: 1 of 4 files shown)


libfreerdp/core/transport.c
```diff
@@ -865,6 +865,12 @@ static SSIZE_T transport_read_layer(rdpTransport* transport, BYTE* data, size_t
 
 		if (status <= 0)
 		{
+			if (status == 0)
```
akallabeth, line 868: this adds a bug, read the code below...

libfreerdp/core/tcp.c
```diff
@@ -1013,6 +1013,13 @@ BOOL freerdp_tcp_set_keep_alive_mode(const rdpSettings* settings, int sockfd)
 	const BOOL keepalive = (freerdp_settings_get_bool(settings, FreeRDP_TcpKeepAlive));
 	UINT32 optval = 0;
 	socklen_t optlen = 0;
+	BOOL isTcp = FALSE;
+	struct sockaddr_storage addr = WINPR_C_ARRAY_INIT;
+	socklen_t addrlen = sizeof(addr);
+
+	if (getsockname(sockfd, (struct sockaddr*)&addr, &addrlen) == 0)
+		isTcp = (addr.ss_family == AF_INET) || (addr.ss_family == AF_INET6);
```
akallabeth, line 1021: this does not check for `TCP` but `IP`

akallabeth (changes requested):

libfreerdp/core/tcp.c
akallabeth:
```
int xoptval = 0;
	socklen_t xoptlen = sizeof(xoptval);
	const int rc = getsockopt(sockfd, SOL_SOCKET, SO_TYPE, &xoptval, &xoptlen);
if (xoptval == SOL_STREAM) {... }
```

akallabeth:
@ligurio ping? 
I think most of this stuff is ok, but drop the `EOF` patch, that breaks error handling.
as for the `TCP` check, I've already added the correct socket option to check for that (maybe we get `UDP` support soon :tm: - as in next decade)

akallabeth:
@ligurio ok, taken most of the commits and added them to https://github.com/FreeRDP/FreeRDP/pull/13257

PR #13262, opened 2026-08-25
Fix utf-8 encoding of LS and PS in winpr_str_has_newlines

jasonmli8:
Old utf-8 byte codes were incorrect, assuming the unicode ID translated directly to bytes. According to https://www.compart.com/en/unicode/U+2028, the utf-8 encoding should be 0xE2 0x80 0xA8 and 0xE2 0x80 0xA9 for LS and PS respectively. This change properly checks that.

### DIFF
--- a/winpr/libwinpr/crt/string.c
+++ b/winpr/libwinpr/crt/string.c
@@ -963,16 +963,10 @@ BOOL winpr_str_has_newlines(const char* str)
 			case 0x0c:       /* FF form feed */
 			case (char)0x85: /* NEL next line */
 				return TRUE;
-			case 0x20:
+			case (char)0xE2:
 			{
-				switch (*str)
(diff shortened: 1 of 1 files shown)


akallabeth (approved):
LGTM. 
nice catch!

akallabeth:
@freerdp-bot test

jasonmli8:
Okay my bad for all the force pushes, but I think now it should be good.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@jasonmli8 please just do `git clang-format -f` until there are no changes left (`clang-format` might not finish in one run)

akallabeth:
@freerdp-bot test

PR #13263, opened 2026-08-25
Relax check in window_state_order_clone

jasonmli8:
Accept cloning of orders where windowRects or visibilityRects is 0. rectangles_clone would return nullptr when numWindowRects or numVisibilityRects == 0. These get treated the same as a memory allocation error and frees the clone when the empty list is expected.


akallabeth:
@jasonmli8 may I ask how you hit this condition?

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #13265, opened 2026-08-25
[core,test] ignore SIGPIPE in TestFuzzServerSeedGen

ligurio:
The relay threads and the FreeRDP transport write to sockets whose peer end may already be closed during teardown. A plain write()/send() to such a socket raises SIGPIPE, which terminates the whole process with a signal (observed as exit code 141) instead of failing the call.

Ignore SIGPIPE so the write returns EPIPE and the existing error paths (relay loop breaks on a short write, transport treats it as a closed connection) handle the situation cleanly.

Needed for google/oss-fuzz#16042


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13266, opened 2026-08-25
[client,x11] fix cliprdr infinite request loop

ElCruncharino:
`queued_responses[0]` gets read to build the next format-data request, but it never actually gets removed from the queue. Only duplicates at index 1+ get folded into `pending_responses` and resolved. So the head entry never resolves: the response comes back, `pending_responses` is empty, and we fall through to the same still-queued entry and fire the same request again. Forever.

It doesn't appear to be an edge case. `queued_responses` ends up with more than one entry any time a Format List PDU announces more than one format at once, which MS-RDPECLIP requires ("the sender MUST enumerate all of the formats currently available from the local system clipboard"). Just copying a file while there's also text on the clipboard is enough.

Tested against a server that offers text plus a file list together, connected with `xfreerdp`. Before: hundreds of thousands of duplicate requests a second, still climbing minutes later. After: 3 requests, then the logs were quiet.

Fix is one line: start the fold-in scan at index 0 instead of 1, so the head entry gets dequeued and resolved like everything else.

### DIFF
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -2456,7 +2456,7 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
 		const xfCliprdrFormat* cformat =
 		    xf_cliprdr_get_client_format_by_atom(clipboard, next->expectedResponse->target);
 
-		size_t index = 1;
+		size_t index = 0;
 		while ((next = ArrayList_GetItem(clipboard->queued_responses, index)) != nullptr)
 		{
 			if (next->requestedFormat->formatToRequest == nextFormatId)


client/X11/xf_cliprdr.c
```diff
@@ -2456,7 +2456,7 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
 		const xfCliprdrFormat* cformat =
 		    xf_cliprdr_get_client_format_by_atom(clipboard, next->expectedResponse->target);
 
-		size_t index = 1;
+		size_t index = 0;
```
akallabeth, line 2459: now you kill the element being requested?

akallabeth:
@ElCruncharino which server did you test against?
can not reproduce against any windows VM I have available.

client/X11/xf_cliprdr.c
ElCruncharino: No, `SetItem(..., nullptr)` just overwrites the slot. next already holds the real object before that overwrite, so `RemoveAt` right after only no-ops on an empty slot. Looks like the object survives and gets appended to `pending_responses`.

ElCruncharino:
> @ElCruncharino which server did you test against? can not reproduce against any windows VM I have available.

Reproduced this against FreeRDP's own shadow server, but I discovered it while building the file functionality of [KRDP](https://invent.kde.org/plasma/krdp/-/merge_requests/214).

Wired a minimal CliprdrServerContext into freerdp-shadow-cli (just enough to send a two-format list: `CF_UNICODETEXT` + `FileGroupDescriptorW`) and ran it against xfreerdp in an isolated X11 session. Same result: hundreds of thousands of duplicate requests per second with the bug present, clean exchange after the fix.

Why Windows doesn't do it, your guess is as good as mine.

akallabeth:
@ElCruncharino do you have that code (the shadow modifications) pushed somewhere to have a look at it?

akallabeth:
@ElCruncharino ok, windows does not announce files & text that might explain why it is not reproducible there.

ElCruncharino:
> @ElCruncharino do you have that code (the shadow modifications) pushed somewhere to have a look at it?

Pushed it here: https://github.com/ElCruncharino/FreeRDP/tree/cliprdr-storm-repro

It's based on the commit right before this fix, with a minimal `CliprdrServerContext` wired into `freerdp-shadow-cli`'s connect/free hooks (`test_cliprdr_storm.c`/`.h`) that just announces `CF_UNICODETEXT` + `FileGroupDescriptorW` together in one Format List PDU. Should be just enough to reproduce against a stock server.

To reproduce:
```
git checkout cliprdr-storm-repro
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug -DWITH_SHADOW_X11=ON -DWITH_FUSE=ON -DWITH_X11=ON -DWITH_SERVER=ON
ninja -C build freerdp-shadow-cli xfreerdp

Xvfb :99 -screen 0 1280x1024x24 &
DISPLAY=:99 ./build/server/shadow/cli/freerdp-shadow-cli /port:3391 -auth -sec-nla /sec:tls &
./build/client/X11/xfreerdp /v:localhost:3391 /u:test /p:test /cert:ignore +clipboard /sec:tls /log-level:DEBUG
```
Should see the client log flooded with repeating `ClientFormatDataRequest`/`ServerFormatDataResponse` pairs at the same millisecond timestamp, climbing unbounded. With this PR, should settle down.

And that lines up with what you found: if Windows never announces files and text together in one list, `queued_responses` never gets past 1 entry there, so the bug's precondition just doesn't occur on your VMs.

Lmk if you spot any gotchas with my testing method.

akallabeth:
@ElCruncharino ok, did run that, don't even hit the code you touched :/
clipboard is getting text just fine and in files I can paste the file (well, I can initiate but obviously no data is copied)

debian13/gnome/wayland here, what do you run?

akallabeth:
@ElCruncharino did patch in

```
diff --git a/client/X11/xf_cliprdr.c b/client/X11/xf_cliprdr.c
index 58e2724b41..8c16faedb0 100644
--- a/client/X11/xf_cliprdr.c
+++ b/client/X11/xf_cliprdr.c
@@ -2450,6 +2450,7 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
        WINPR_ASSERT(ArrayList_Count(clipboard->pending_responses) == 0);
 
        SelectionResponse* next = ArrayList_GetItem(clipboard->queued_responses, 0);
+       WLog_INFO("xxxx", "[0] next=%p [%" PRIuz "]", next, ArrayList_Count(clipboard->queued_responses));
        if (next)
        {
                UINT32 nextFormatId = next->requestedFormat->formatToRequest;
@@ -2459,6 +2460,7 @@ xf_cliprdr_server_format_data_response(CliprdrClientContext* context,
                size_t index = 1;
                while ((next = ArrayList_GetItem(clipboard->queued_responses, index)) != nullptr)
                {
+                       WLog_INFO("xxxx", "[%" PRIuz "] next=%p [0x%08" PRIx32":0x%08" PRIx32 "]", index, next, next->requestedFormat->formatToRequest, nextFormatId);
                        if (next->requestedFormat->formatToRequest == nextFormatId)
                        {
                                /* First set item to nullptr, then remove/free the element. Avoids double free */
```

to get log messages whenever the stuff is executed, but no such log entry there.
clipboard manager shows the `[Files] hi` text

akallabeth:
@ElCruncharino ok, got it.
do you happen to use a clipboard manager that fetches all available data whenever it becomes available?

the path you're in happens only if the formats are requested very quickly in succession before the responses arrive.

akallabeth (approved):
LGTM,
nice catch!

ElCruncharino:
I don't run a dedicated clipboard manager, but I do run KDE Plasma (KWin) on Wayland, and functionally that's the same thing here: KWin's own Xwayland clipboard bridge (`src/xwayland/clipboard.cpp`, `Xwl::Clipboard::selectionClaimed()`) fires `requestTargets()` automatically every time CLIPBOARD ownership changes, which happens the instant `xfreerdp` claims it to offer the server's data. So it's baked into the compositor, which is why I found nothing when I checked for one earlier.

Caught it directly too: setting a multi-format clipboard entry collapses to a single plain-text one within about a second with no action from me, KWin's bridge claiming and re-publishing it. That's gotta be the requester creating the race on my end, and probably why GNOME's equivalent bridge doesn't hit it the same way.

akallabeth:
@freerdp-bot test

akallabeth:
@ElCruncharino ok, but that means that `RDP` clients are more or less unusable on `KDE`?
I mean if you have a workload copying huge data (only on the server side) then the compositor tries to copy all that data to the client regardless of a paste event or not?

(correct me if I got that wrong)

ElCruncharino:
> @ElCruncharino ok, but that means that `RDP` clients are more or less unusable on `KDE`? I mean if you have a workload copying huge data (only on the server side) then the compositor tries to copy all that data to the client regardless of a paste event or not?
> 
> (correct me if I got that wrong)

Two separate things: KWin's bridge only asks for the format list, not data, so that alone isn't a bandwidth concern.

But Plasma's clipboard-history applet does keep a persistent history, which means it likely does eagerly pull actual content to cache it, not just on paste. Not RDP-specific though, it does that for any clipboard source, local or remote. So less "RDP is unusable on KDE," more "Plasma always eager-fetches for history," pre-existing behavior.

On the loop itself: more likely the history applet alone requests more than one format back to back for a single entry, which by itself is enough. I haven't caught that directly in a trace though, so treat it as inference, not confirmed.

Candidly, not a lot of Linux clients exist with clipboard file support, and I've only just been building that side of things in Remmina and in KRDC, the former of which isn't merged just yet. the xfreerdp3 was the only one I had to work with as a reference, so I can't imagine this has been seen a bunch in the wild.

PR #13272, opened 2026-08-26
Release cleanup

akallabeth:


winpr/include/winpr/print.h
```diff
@@ -41,9 +41,9 @@ extern "C"
 	                                size_t width);
 
 	WINPR_ATTR_MALLOC(free, 1)
-	WINPR_API char* winpr_BinToHexString(const BYTE* data, size_t length, BOOL space);
+	WINPR_API char* winpr_BinToHexString(const void* data, size_t length, BOOL space);
```
hardening, line 44: API change ?

akallabeth: yes, but a compatible one. (no size change)
prepared an update of the suppression file should the `ABI` checker complain.

akallabeth: should reduce the required casts considerably.

PR #13273, opened 2026-08-26
[winpr] fix TestThreadCreateProcess

hardening:
In WinPR we have respected the windows API that states that values passed to UpdateProcThreadAttribute must survive until DeleteProcThreadAttributeList is called. It happens that even in C (depending on compiler settings), local variable are altered when exiting a block, so in the unitary test let's move the `handles` declaration a bit higher in the function so that it lives for the whole function.


akallabeth (approved):
LGTM

PR #13278, opened 2026-08-26
[client,sdl3] fix flickering on screens with DPI != 1.0

hardening:
This patch reduces the flickering on screens with an exotic DPI (also doesn't remove it completely).


akallabeth (commented):
Not sure this works.

`SDL_GetDisplayContentScale` is not reliable, only data gathered from real windows actually tells the scale in use.

hardening:
Hum but this is a bit chicken and egg as the flickering happens at window creation, we don't have a window to know the DPI. So even if the value is not 100% accurate that may help (at least here it reduces the flickering). And IIRC the size is adjusted very quickly after that on the first update...

akallabeth:
@hardening true.

hardening:
@freerdp-bot come on git is there so do your test work please

akallabeth (approved):
LGTM

PR #13280, opened 2026-08-26
create alpha capable GFX surfaces for XRGB_8888

### LINKED ISSUE #13244: RemoteApp: Windows 11 menu drop shadow renders as an opaque black bar
# **Describe the bug**

In a Windows 11 RemoteApp (RAIL) session, some Explorer context menus render with solid black bars along their edges.

The bar is the menu's DWM drop shadow, which the server sends as RGB black plus a low per-pixel alpha ramp. The alpha is lost between the `RDPGFX_CODECID_ALPHA` surface command and the frame reaching the client.

It is deterministic per menu, not a timing race: the same menu is always broken and another is always clean, across reconnects and restarts.

# **To Reproduce**

1. Start a RemoteApp session against a Windows 11 host launching `explorer.exe`
2. Stay in the default **Home** folder, right click empty space in the file list
3. The context menu shows solid black bars along its right and bottom edges

In the same session, navigating to **This PC** and right clicking empty space gives a correct menu with no bar. Home always shows it, This PC never does.

Reproduced with `xfreerdp`. Also seen with a local SDL3 client build carrying RAIL support, which is not upstream, so `xfreerdp` is the one to reproduce with.

# **Expected behavior**

The menu's drop shadow should render as a soft translucent edge, the way the This PC menu does in the same session.

# **Screenshots**

Home, black bar along the right edge and bottom edge:

Image
Image

This PC, same session, correct:

Image

# **Application details**

* FreeRDP version: `3.30.1-dev0`, built from source
* Command line: `xfreerdp /v:<host> /u:<user> /cert:ignore /app:program:explorer.exe,hidef:on`
* Server side: Windows 11
* Also present on current master `962656442`
* Build: Release, GCC 16.2.1, x64, `WITH_SIMD=ON WITH_AVX2=ON WITH_GFX_H264=ON WITH_CLIENT_SDL3=ON`

 - OS: Linux
 - Version/Distribution: Arch Linux
 - Architecture: amd64

# **Additional context**

Numbers below come from a temporary local probe that dumps `surface->data` and reads back a single pixel after each command that writes to the surface. Not submitted, only used to locate this.

### Surface contents

Mapped GFX surface for both menus in one session, full 32 bit pixels, rightmost 10 columns of the middle row.

This PC (clean), 288x279:

```
0D000000 0B000000 09000000 07000000 06000000 04000000 03000000 02000000 01000000 01000000
```

Home (broken), 200x165:

```
0D000000 0B000000 FF000000 FF000000 FF000000 FF000000 FF000000 FF000000 FF000000 FF000000
```

The ramp starts, then goes fully opaque for the last 8 pixels. The left and top edges of the same surface sample as expected.

### Where the alpha is lost

Pixel (196, 82) of the broken menu's surface, after each command:

```
gdi_CacheToSurface             (196,82) = 00000000
RDPGFX_CODECID_CLEARCODEC      (196,82) = 00000000
RDPGFX_CODECID_ALPHA           (196,82) = 03000000
RDPGFX_CODECID_CAPROGRESSIVE   (196,82) = FF000000
```

The clean menu never receives a CAPROGRESSIVE command for its surface.

### Findings

1. The surface is created from `GFX_PIXEL_FORMAT_XRGB_8888`, so `surface->format` is `PIXEL_FORMAT_BGRX32` and `FreeRDPColorHasAlpha()` is false
2. `gdi_SurfaceCommand_Alpha` still stores per-pixel alpha in the 4th byte of that surface, measured as 0x03 above
3. `progressive_decompress` copies its tiles with `FREERDP_KEEP_DST_ALPHA`
4. The guard drops the flag for BGRX32, so the copy falls through to a full 4 byte path:

```c
if (((flags & FREERDP_KEEP_DST_ALPHA) != 0) && FreeRDPColorHasAlpha(DstFormat))
```

5. The progressive tile buffer is allocated with `memset(tile->data, 0xFF, dataLen)`, so the 4th byte arrives as 0xFF

`sse_image_copy_no_overlap_dst_alpha` already dispatches `PIXEL_FORMAT_BGRX32` to `sse_image_copy_bgrx32_bgrx32`, which writes only bytes 0..2.

The guard appears in four places, and the SSE and AVX2 variants override `prims->copy_no_overlap`:

- `libfreerdp/primitives/prim_copy.c`
- `libfreerdp/primitives/sse/prim_copy_sse4_1.c`
- `libfreerdp/primitives/sse/prim_copy_avx2.c`
- `libfreerdp/codec/color.c` (in `freerdp_image_copy_overlap`)

### What I tried

Gating on pixel width instead:

```c
if (((flags & FREERDP_KEEP_DST_ALPHA) != 0) && (FreeRDPGetBytesPerPixel(DstFormat) == 4))
```

Fixes the artifact, but fails `TestFreeRDPCodecInterleaved`, which decodes into a `PIXEL_FORMAT_RGBX32` buffer through `interleaved_decompress` (also passing `FREERDP_KEEP_DST_ALPHA`) and whose reference data encodes the current behaviour of the padding byte. `TestPrimitives` passes.

Happy to test any patch or hypothesis, I still have the instrumentation and a reliable reproducer.

svncibrahim:
- Fixes #13244.

Windows 11 creates RemoteApp menu surfaces as `GFX_PIXEL_FORMAT_XRGB_8888` and then sends `RDPGFX_CODECID_ALPHA` for them. The alpha lands in the 4th byte of a `PIXEL_FORMAT_BGRX32` surface, where `FreeRDPColorHasAlpha()` is false, so the next `RDPGFX_CODECID_CAPROGRESSIVE` command drops `FREERDP_KEEP_DST_ALPHA` and overwrites it with the tile buffer's `0xFF` padding. The menu's DWM drop shadow then renders as an opaque black bar.

Both surface creation paths now map `XRGB_8888` to `PIXEL_FORMAT_BGRA32`, so the existing `FREERDP_KEEP_DST_ALPHA` guard works and the alpha survives. Both already prefill the surface with `0xFF`, so a surface that never receives alpha data stays opaque.

## Measurements

In one Explorer session `xfreerdp` created 14 XRGB surfaces and all 14 received an ALPHA command. It also created 2 ARGB ones, neither received any.

Sampling one pixel of the broken menu's surface (200x165, the one from the report) after each command, same session and same menu, only the mapping changed:

```
                              before     after
RDPGFX_CODECID_CLEARCODEC     00000000   00000000
RDPGFX_CODECID_ALPHA          03000000   03000000
RDPGFX_CODECID_CAPROGRESSIVE  FF000000   03000000
```

The sequence repeated four times per session in both runs. The black bar is visible on screen in the before run and gone in the after run.

### DIFF
--- a/client/X11/xf_gfx.c
+++ b/client/X11/xf_gfx.c
@@ -356,7 +356,7 @@ static UINT xf_CreateSurface(RdpgfxClientContext* context,
 			break;
 
 		case GFX_PIXEL_FORMAT_XRGB_8888:
-			surface->gdi.format = PIXEL_FORMAT_BGRX32;
+			surface->gdi.format = PIXEL_FORMAT_BGRA32;
 			break;
 
 		default:
--- a/libfreerdp/gdi/gfx.c
+++ b/libfreerdp/gdi/gfx.c
@@ -1295,7 +1295,7 @@ static UINT gdi_CreateSurface(RdpgfxClientContext* context,
 			break;
 
 		case GFX_PIXEL_FORMAT_XRGB_8888:
-			surface->format = PIXEL_FORMAT_BGRX32;
+			surface->format = PIXEL_FORMAT_BGRA32;
 			break;
 
 		default:


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM.
nice catch!

PR #13284, opened 2026-08-26
[winpr,handles] allow to Close a HANDLE listed in PROC_THREAD_ATTRIBUTE_HANDLE_LIST

hardening:
The goal of this patch is to correctly track the usage of HANDLEs especially if they are referenced in a PROC_THREAD_ATTRIBUTE_HANDLE_LIST. Before the changes if you UpdateProcThreadAttribute with a list of handles to pass to a child process and close some of these handles, when CreateProcess will be executed a crash will most probably happen because it will try to access freed objects. So this patch introduces refcounting on HANDLEs, meaning that if you Close a HANDLE it does not free its associated object if the HANDLE is referenced in a PROC_THREAD_ATTRIBUTE_HANDLE_LIST.
For now we also disable the DuplicateHandle function as: it's not used anywhere in WinPR/ FreeRDP, and it doesn't reproduce the behavior of the equivalent Win32 function (for that we would need to have our HANDLEs that would "point" to internal WinPR objects with ref counters in them, so that when you Close the duplicated HANDLE it doesn't affect the original one, just unref the internal object).

### DIFF
--- a/winpr/libwinpr/comm/comm.c
+++ b/winpr/libwinpr/comm/comm.c
@@ -1502,7 +1502,6 @@ BOOL CommCloseHandle(HANDLE handle)
 	if (pComm->fd_read_event > 0)
 		close(pComm->fd_read_event);
 
-	free(pComm);
 	return TRUE;
 }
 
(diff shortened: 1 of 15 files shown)


akallabeth (approved):
LGTM,
nice work!

PR #13285, opened 2026-08-27
[winpr,cmake] fix libs passed back to the pkg-config file on OpenBSD

brad0:
OpenBSD does not have libdl or librt.


akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/13288

PR #13298, opened 2026-08-28
[winpr,json] read JSON files in binary mode

SebastienGllmt:
This PR fixes reading of `sdl-freerdp.json` on Windows

Cause: text mode translates CRLF → LF on Windows (causing `fread` returns fewer bytes than the file size previously measured with `_ftelli64`)

This means that the [exact byte-count check](https://github.com/FreeRDP/FreeRDP/blob/b2a12143658cfb8667406f224488a1b5dcbd221d/winpr/libwinpr/utils/json/json.c#L58-L60) rejected valid JSON files (since the CRLF → LF translation modifies the bytes).

Fix: open files in binary mode (and add a CRLF regression test)

### DIFF
--- a/winpr/libwinpr/utils/json/json.c
+++ b/winpr/libwinpr/utils/json/json.c
@@ -26,7 +26,7 @@
 
 WINPR_JSON* WINPR_JSON_ParseFromFile(const char* filename)
 {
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

PR #13300, opened 2026-08-29
[NTLM] Fix UTF-8 context attribute conversion

fabienr:
Running make test from openbsd port (https://marc.info/?l=openbsd-ports&m=178779328763275&w=2),

x11/freerdp $ make retest 'TEST_ENV=MALLOC_OPTIONS=C' 'TEST_FLAGS=-R TestSspiNTLM --output-on-failure -V'
===>  Regression tests for freerdp-3.31.0
...
TestSspi(30163) in free(): canary corrupted 0x11833aa1dc0[7]@7/16

(gdb) bt
...
#7  0x0000011764f59b23 in _libc_free (ptr=0x11833aa1dc0) at /usr/src/lib/libc/stdlib/malloc.c:1729
#8  0x000001155f7e1b31 in test_attributes_read (table=0x1180f7a0320 <winpr_SecurityFunctionTableA.llvm.15257515958526039962>, context=0x1176efc2f00, expected=0x1155f7dd25c,
    expectedLen=<optimized out>) at .../freerdp-3.31.0/winpr/libwinpr/sspi/test/TestSspiNTLM.c:198
#9  test_attributes (table=0x1180f7a0320 <winpr_SecurityFunctionTableA.llvm.15257515958526039962>, context=0x1176efc2f00)
    at .../freerdp-3.31.0/winpr/libwinpr/sspi/test/TestSspiNTLM.c:255
#10 0x000001155f7e0bf8 in test_default (arg=0x72263a9c88d0, fkt=0x0) at .../freerdp-3.31.0/winpr/libwinpr/sspi/test/TestSspiNTLM.c:780
#11 0x000001155f7e04b9 in TestSspiNTLM (argc=<optimized out>, argv=<optimized out>) at .../freerdp-3.31.0/winpr/libwinpr/sspi/test/TestSspiNTLM.c:1070        
#12 0x000001155f7dfb4c in main (ac=<optimized out>, av=0x72263a9c8a70) at .../build-amd64/winpr/libwinpr/sspi/test/TestSspi.c:226

I didn't make the hard job seeking the issue but an llm does.
I carefully check the reasoning step by step and the fix looks correct.

### DIFF
--- a/winpr/libwinpr/sspi/NTLM/ntlm.c
+++ b/winpr/libwinpr/sspi/NTLM/ntlm.c
@@ -1209,11 +1209,33 @@ static SECURITY_STATUS utf8len(const UNICODE_STRING* str, void* pBuffer)
 	WINPR_ASSERT(str);
 	WINPR_ASSERT(pBuffer);
 	ULONG* val = (ULONG*)pBuffer;
-	const SSIZE_T rc = ConvertWCharNToUtf8(str->Buffer, str->Length, nullptr, 0);
+	const size_t wlen = str->Length / sizeof(WCHAR);
+	const SSIZE_T rc = ConvertWCharNToUtf8(str->Buffer, wlen, nullptr, 0);
 	if (rc < 0)
 		return SEC_E_INVALID_PARAMETER;
 	*val = WINPR_ASSERTING_INT_CAST(ULONG, rc);
 	return SEC_E_OK;
+}
+
+WINPR_ATTR_NODISCARD
+static SECURITY_STATUS utf8str(const UNICODE_STRING* str, void* pBuffer)
+{
+	WINPR_ASSERT(str);
+	WINPR_ASSERT(pBuffer);
+	ULONG len = 0;
+
+	const SECURITY_STATUS status = utf8len(str, &len);
+	if (status != SEC_E_OK)
+		return status;
+	if (len == 0)
+	{
+		((char*)pBuffer)[0] = '\0';
+		return SEC_E_OK;
+	}
+
+	const size_t wlen = str->Length / sizeof(WCHAR);
+	const SSIZE_T rc = ConvertWCharNToUtf8(str->Buffer, wlen, pBuffer, (size_t)len);
+	return rc < 0 ? SEC_E_INVALID_PARAMETER : SEC_E_OK;
 }
 
 WINPR_ATTR_NODISCARD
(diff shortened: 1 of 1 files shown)


fabienr:
Sorry I got distracted and sent it without further explanation.

UNICODE_STRING.Length is in bytes, but ConvertWCharNToUtf8() expects a WCHAR count and a UTF-8 destination capacity. The getters passed the byte count for both, so ICU converted the stored NUL and appended another NUL.
That second NUL exceeded the buffer allocated from the reported UTF-8 length by one byte.

Using MALLOC_OPTIONS=C, it detects the canary corruption due to this one-byte overflow.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

winpr/libwinpr/sspi/NTLM/ntlm.c
```diff
@@ -1209,13 +1209,35 @@ static SECURITY_STATUS utf8len(const UNICODE_STRING* str, void* pBuffer)
...
+	if (status != SEC_E_OK)
+		return status;
+	if (len == 0)
+	{
+		((char*)pBuffer)[0] = '\0';
+		return SEC_E_OK;
+	}
+
+	const size_t wlen = str->Length / sizeof(WCHAR);
+	const SSIZE_T rc = ConvertWCharNToUtf8(str->Buffer, wlen, pBuffer, (size_t)len + 1);
```
akallabeth: where is the `+1` from?

fabienr: Actually it would be better to pass the buffer size as an argument to ntlm_QueryContextAttributesA(), which is way more invasive, as this affects all the ->QueryContextAttributes() calls and requires a version major bump (I guess).

I didn't study all the ConvertWCharNToUtf8(), this function returns the length **without** the NUL termination byte. The problem is we have to guess the destination buffer size, so there is an assumption about the buffer being properly initialized based on a previous call to *_LEN. Also, this is a string conversion; thus, the buffer must end with a NUL.

Without the +1, pBuffer may have no NUL termination, which may be dangerous (unbound read).

I didn't dig deep to check how the different int_WideCharToMultiByte() handles short buffers and if it returns an error when there is no room for NUL. On OpenBSD we have strlcpy() for this issue; the function will write a NUL and return a truncated string if there is not enough space.

I tried without +1 and test passed, but it only does calloc/memcmp without NUL check, which makes me think (at least the ICU backend on openbsd) the code does not enforce a NUL termination byte. If I change the test to malloc() instead of calloc() (because calloc init to NULL) and then use _tcscmp() instead of memcmp(), the test fails.

TestSspiNTLM.c:170 : TCHAR* buffer = calloc(expectedLen + 1, sizeof(TCHAR));
credssp_auth.c:554 : void* WorkstationName = calloc(len + 1, sizeof(WCHAR));

akallabeth: @fabienr the `API` is a windows compatible one (given, the attributes are our extensions), so change is impossible.
The length is queried first (the length attribute) and `\0` termination is up to the caller.

the `+1` could be an out of bound write.

akallabeth: both, `utf16len` and `utf8len` return a `strlen` like number (without terminating `\0`)

fabienr: sure, then I removed the +1 and let the callers handle NUL termination correctly.

winpr/libwinpr/sspi/NTLM/ntlm.c
```diff
@@ -1209,13 +1209,35 @@ static SECURITY_STATUS utf8len(const UNICODE_STRING* str, void* pBuffer)
...
+	WINPR_ASSERT(str);
+	WINPR_ASSERT(pBuffer);
+	ULONG len = 0;
+
+	const SECURITY_STATUS status = utf8len(str, &len);
+	if (status != SEC_E_OK)
+		return status;
+	if (len == 0)
+	{
+		((char*)pBuffer)[0] = '\0';
```
akallabeth, line 1232: this might also be an out of bound write.

akallabeth:
taken your commit, added to https://github.com/FreeRDP/FreeRDP/pull/13308 and fixed the remaining out of bound write.

PR #13304, opened 2026-08-29
[winpr,clipboard] honor DIB header size in BMP offsets

SebastienGllmt:
This PR fixes copying images from Firefox through an xrdp session.

There are three issues that are bundled together:
1. Fix `clipboard_prepend_bmp_header` discarding `readBitmapInfoHeader` info (was constructing bfOffBits using a fixed size)
2. Fix support for V4/V5 headers, which already contain the color masks in biSize
3. Fix support for older BMP assets like 8-bpp images with 256-color palettes (the default indexed-color palette calculation divided the number of entries by eight, which is incorrect for these). Strictly speaking, this one isn't required to get Firefox "copy image" working in Firefox, but is fixed together  so this PR is a clean DIB pixel-offset fix

Adds regression coverage for:

- `BITMAPV4HEADER`
- Legacy `BITMAPINFOHEADER` with `BI_BITFIELDS`
- Indexed DIB palettes

I also tested on Windows manually by copying an image from Firefox through xrdp.

### DIFF
--- a/winpr/libwinpr/clipboard/synthetic.c
+++ b/winpr/libwinpr/clipboard/synthetic.c
@@ -392,7 +392,8 @@ static void* clipboard_synthesize_cf_dibv5(wClipboard* clipboard, UINT32 formatI
 #endif
 
 static void* clipboard_prepend_bmp_header(const WINPR_BITMAP_INFO_HEADER* pInfoHeader,
-                                          const void* data, size_t size, UINT32* pSize)
+                                          size_t offset, const void* data, size_t size,
+                                          UINT32* pSize)
 {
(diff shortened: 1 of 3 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@SebastienGllmt so this fixes a problem with `xrdp` as server, nothing related to windows RDP (or FreeRDP based) correct?

akallabeth:
@SebastienGllmt works against windows and g-r-d so consider it LGTM.

akallabeth (approved):
LGTM

PR #13305, opened 2026-08-30
[codec,h264] separate YUV allocation and logical strides

SebastienGllmt:
This fixes the regression in [bea10804a](https://github.com/FreeRDP/FreeRDP/commit/bea10804af9c9356979bd1af173993ab795377a6) (which broke H.264 on my Windows machine)

image

**Disclaimer**: while for my previous PRs I am able to understand what is going on and review the code, for this one I just don't have enough background knowledge on H.264 (or the codebase given the size of the diff) to confidently know whether or not this is the right fix. It does work on my machine at least. Everything below is an AI summary which I can only understand at a high level

## Motivation

Commit [bea10804a](https://github.com/FreeRDP/FreeRDP/commit/bea10804af9c9356979bd1af173993ab795377a6) intentionally changed the H.264 YUV buffers to allocate an additional alignment block.
Some decoders write aligned dimensions even when the visible dimensions are already aligned, so retaining that additional capacity is necessary for memory safety.

However, the allocation stride was also stored in `H264_CONTEXT::iStride`.
That made the amount of allocated storage and the logical layout of the decoded frame the same value, even though they represent different properties.

This is incorrect for Media Foundation output.

Microsoft defines `MF_MT_DEFAULT_STRIDE` as the number of bytes needed to advance from one image row to the next.
The value is stored as `UINT32`, but Microsoft requires consumers to interpret it as a signed 32-bit value.
For a contiguous `IMFMediaBuffer`, this attribute describes the buffer’s row layout.
See [MF_MT_DEFAULT_STRIDE](https://learn.microsoft.com/en-us/windows/win32/medfound/mf-mt-default-stride-attribute).

Microsoft also explicitly notes that buffers with identical image dimensions can have different strides.
Its recommended approach for transforms is to keep the source and destination strides separate and process the image one row at a time.
See [Image Stride](https://learn.microsoft.com/en-us/windows/win32/medfound/image-stride)

The Media Foundation backend negotiates `MFVideoFormat_IYUV`.
Microsoft defines IYUV as planar 8-bit 4:2:0 with the same memory layout as I420.
See [Video Subtype GUIDs](https://learn.microsoft.com/en-us/windows/win32/medfound/video-subtype-guids)

Consequently, allocation padding must not change the stride used to locate rows or Y, U, and V planes in the Media Foundation output buffer.

## Failure mode

The problem is reproducible with a 1804×1128 frame for which Media Foundation reports a stride of 1808 bytes.

Because 1808 is already 16-byte aligned, the unconditional alignment calculation introduced by `bea10804a` changes it to 1824.
FreeRDP then uses 1824 to interpret an output buffer whose actual row stride remains 1808.

For this frame:

- The correct Y-plane size is `1808 × 1128 = 2,039,424` bytes.
- Each chroma plane is `904 × 564 = 509,856` bytes.
- The complete valid IYUV frame is therefore `3,059,136` bytes.
- Interpreting the frame with a stride of 1824 instead attempts to consume `3,086,208` bytes.

This shifts the U and V plane offsets and attempts to read 27,072 bytes beyond the valid frame layout.
The resulting row and plane displacement appears as progressive diagonal corruption and incorrect chroma.

## Implementation

The change separates two invariants:

- `logicalStride` describes the decoded image layout exposed through `iStride`.
- `allocationStride` and `allocationHeight` describe the amount of backing storage reserved for decoder padding.

For an already-aligned decoder stride, the logical stride is preserved unchanged.
For an unaligned decoder stride, the existing round-up-to-16 behavior is preserved.
For both cases, the allocation still extends to the next alignment block exactly as required by `bea10804a`.

The Media Foundation backend now retains its reported source stride separately.
Decoded IYUV data is copied one row at a time using independent source and destination strides.
Plane offsets are calculated from the source layout rather than from the destination allocation layout.

The output length returned by `IMFMediaBuffer::Lock` is also validated before any plane is accessed.
Microsoft defines the current length as the amount of valid data in the buffer and states that readers must not read beyond it.
See [Working with Media Buffers](https://learn.microsoft.com/en-us/windows/win32/medfound/working-with-media-buffers)

The size calculations use checked arithmetic and ceiling chroma dimensions.
This also makes the helper well-defined for odd frame dimensions and rejects truncated buffers before copying.

## Regression safety

The allocation formula introduced by `bea10804a` is preserved for every valid width and height.
This change therefore does not reduce the storage available to decoders that write an additional aligned block.

The logical-stride behavior is covered separately from allocation capacity:

- An aligned source stride of 1808 remains logically 1808 while allocating an 1824-byte row.
- An unaligned source stride of 1804 continues to round up logically and physically to 1808.
- An already-aligned height continues to receive an additional 16-row allocation block.
- Independent source and destination strides are exercised by row-copy tests.
- Odd dimensions exercise ceiling chroma width and height calculations.
- A source buffer one byte shorter than the required IYUV layout is rejected.

These tests are backend-independent, so the buffer and copy invariants do not depend on Media Foundation being available on the test host.

## Testing

- Built successfully on Windows with MSVC and `WITH_MEDIA_FOUNDATION=ON`.
- `TestFreeRDPCodecH264Buffer` passes.
- `TestClipboardFormats` and `TestImage` pass.
- Manually tested with an xrdp H.264 session at 1804×1128 and a Media Foundation stride of 1808.
- The diagonal displacement and chroma corruption are no longer present.

### DIFF
--- a/libfreerdp/codec/h264.c
+++ b/libfreerdp/codec/h264.c
@@ -41,16 +41,28 @@ static BOOL avc444_ensure_buffer(H264_CONTEXT* h264, DWORD nDstHeight);
 static BOOL yuv_ensure_buffer(H264_CONTEXT* h264, UINT32 stride, UINT32 width, UINT32 height)
 {
 	BOOL isNull = FALSE;
-	UINT32 pheight = height;
+	UINT32 logicalStride = 0;
+	UINT32 allocationStride = 0;
+	UINT32 allocationHeight = 0;
+	UINT32 allocationStrideUV = 0;
 
 	if (!h264)
 		return FALSE;
 
 	if (stride == 0)
 		stride = width;
+	if ((stride == 0) || (height == 0) || (stride < width))
+		return FALSE;
 
-	stride += 16 - stride % 16;
-	pheight += 16 - pheight % 16;
+	const UINT32 stridePadding = 16 - stride % 16;
+	const UINT32 heightPadding = 16 - height % 16;
+	if ((stride > UINT32_MAX - stridePadding) || (height > UINT32_MAX - heightPadding))
+		return FALSE;
+
+	allocationStride = stride + stridePadding;
+	allocationHeight = height + heightPadding;
+	logicalStride = (stride % 16 == 0) ? stride : allocationStride;
+	allocationStrideUV = (allocationStride + 1) / 2;
 
 	const size_t nPlanes = h264->hwAccel ? 2 : 3;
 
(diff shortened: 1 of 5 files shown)


akallabeth (changes requested):
looks overly complicated.

this affects the media foundation backend only right?

SebastienGllmt:
> this affects the media foundation backend only right?

The core regression, yes

Everything around that I would say is probably just AI guesswork of where you envisioned this part of the codebase to go, but it's probably easier for you to close this PR and tackle this in whatever way you want (if/when you ever have the will to fix the issue) given you modified this code path recently instead of trying to convey to me how you wanted this to go

akallabeth:
@SebastienGllmt well, was hard to guess from the `AI` generated inflated error description.
(the core problem tends to get lost in that sea of words)

so yes, the changes should be contained within the `Media Foundation` file only.
The other backends can just take the buffer stride and work with that.

akallabeth:
@SebastienGllmt ok, the media foundation backend did not use the correct strides during copy, fixed with https://github.com/FreeRDP/FreeRDP/pull/13308

PR #13306, opened 2026-08-30
feat: [client, mac, x11] support jp YEN/RO/EISUU/KANA keys

cawcaw2023:
This patch enables the four Japanese JIS keys (YEN, RO, EISUU, and KANA) under Mac + X11.

**Limitation**

XQuartz does not deliver the EISUU and KANA key events, so only the YEN and RO keys become usable out of the box after this change.  
(I have not tested alternative X11 implementations for macOS.)

**Workaround for EISUU / KANA**

With the combination of xfreerdp and XQuartz, the EISUU and KANA keys are dropped.  
You can work around this by temporarily remapping them to F19/F20 with Karabiner-Elements (only while XQuartz is the frontmost application) and restoring the original scancodes via xfreerdp’s remap feature:
(Confirmed on macOS 26.6.2 / XQuartz 2.8.6)

**Karabiner-Elements > Complex Modifications**
```
{
    "description": "map EISUU/KANA to F19/20 only when XQuartz is the frontmost application.",
    "manipulators": [
        {
            "conditions": [
                {
                    "bundle_identifiers": [
                        "^org\\.xquartz\\.X11$"
                    ],
                    "type": "frontmost_application_if"
                }
            ],
            "from": { "key_code": "japanese_eisuu" },
            "to": [{ "key_code": "f19" }],
            "type": "basic"
        },
        {
            "conditions": [
                {
                    "bundle_identifiers": [
                        "^org\\.xquartz\\.X11$"
                    ],
                    "type": "frontmost_application_if"
                }
            ],
            "from": { "key_code": "japanese_kana" },
            "to": [{ "key_code": "f20" }],
            "type": "basic"
        }
    ]
}
```

**xfreerdp launch options**
`xfreerdp /v:host /u:user /p:pass /kbd:layout:Japanese,remap:0x6a=0x71,remap:0x6b=0x72`

If this PR is accepted, I would like to add the workaround section above as a new section at the end of the existing Keyboard Wiki page, under a heading such as:
**Mac + XQuartz: Enabling Japanese EISUU / KANA keys**

**Tested with**
- Virtualbox Linux host 7.2.16
- Windows 11 VM guest / (native RDP | VRDP)
- debian 13 VM guest / VRDP
- xfreerdp (apple sillicon)


client/X11/xf_keyboard.c
```diff
@@ -687,8 +687,11 @@ static int load_map_from_xkbfile(xfContext* xfc)
 #if defined(__APPLE__)
 				const DWORD vkcode =
 				    GetVirtualKeyCodeFromKeycode((UINT32)i - 8u, WINPR_KEYCODE_TYPE_APPLE);
-				xfc->X11_KEYCODE_TO_VIRTUAL_SCANCODE[i] =
```
akallabeth, line 690: breaks keyboard mapping, rejected.

akallabeth (changes requested):

client/X11/xf_keyboard.c
cawcaw2023: Hi, Thanks for the review. I have a question about the fix approach.
Instead of:
"If no match is found in the WINPR_KBD_TYPE_IBM_ENHANCED table, also search the　WINPR_KBD_TYPE_JAPANESE table."
should it be changed to:
"When the startup options include /kbd:layout:Japanese, search WINPR_KBD_TYPE_JAPANESE instead of WINPR_KBD_TYPE_IBM_ENHANCED."
Would this be a safe change?

akallabeth: @cawcaw2023 no.
proper fix would be to add an option to select the keyboard type manually. (I think autodetection is not possible?)
you can use the japanese layout on a IBM type keyboard too, so these are not necessarily connected.

akallabeth (commented):
@cawcaw2023 the definition of the input mapping looks fine (although I need to check the change for `KBD7_T71` ) while the apple keyboard mapping changes look like they need some thought.
The original code just works around some differences in apple and IBM keyboard mappings.
Are these meant for Japanese Apple keyboards or just generic japanese keyboards?

client/X11/xf_keyboard.c
cawcaw2023: Understood. I realize I was conflating keyboard type and layout.

cawcaw2023:
> Are these meant for Japanese Apple keyboards or just generic japanese keyboards?

It's intended for Japanese Apple keyboards. Compared to the generic layout, most keys are shared except for those related to IME ON/OFF, so I ended up writing the patch without at all considering whether coexisting with Type7 would be fine.

client/X11/xf_keyboard.c
cawcaw2023: @akallabeth Should I modify the patch to look up the WINPR_KBD_TYPE_JAPANESE table instead of WINPR_KBD_TYPE_IBM_ENHANCED, but only when /kbd:type:japanese or /kbd:type:7 is specified?

akallabeth: that sounds reasonable.

PR #13311, opened 2026-08-31
Various WinPR addings

hardening:
Start implementing GetSystemPowerStatus
Implement utility functions to pass handle between processes

### DIFF
--- a/winpr/include/winpr/handle.h
+++ b/winpr/include/winpr/handle.h
@@ -38,6 +38,43 @@ extern "C"
 #define WINPR_FD_WRITE_BIT 1u
 #define WINPR_FD_WRITE (1u << WINPR_FD_WRITE_BIT)
 
+	/**
+	 * Exports an handle to a string platform independently (so exports the HANDLE
+	 * pointer as hexa under WIN32, and under UNIXes a single character identifying the
+	 * handle's type followed by the file descriptor number in hexa, e.g. "P4" for a pipe)
+	 * formatting it with the pseudo format string with {} for the handle (for instance
+	 * `--fdIn={}`). You can use winpr_importHandleFromString to quickly get back a
+	 * HANDLE from the output of this call.
+	 *
+	 * Only pipe HANDLEs (as returned by CreatePipe) can be exported for now under UNIXes.
+	 *
+	 * 	@param h the HANDLE to export
+	 * 	@param format the format string with {} for the handle value
+	 * 	@param outStr the output buffer
+	 * 	@param outSz outStr size
+	 * 	@return if the operation completed successfully
+	 *
+	 * 	@since version 3.32.0
+	 */
+	WINPR_ATTR_NODISCARD
+	WINPR_API BOOL winpr_exportHandleToString(HANDLE h, const char* format, char* outStr,
+	                                          size_t outSz);
+
+	/**
+	 * Imports a HANDLE that was previously exported as a string by winpr_exportHandleToString.
+	 * Under WIN32 it just converts the string to a HANDLE, and under UNIXes it creates a new
+	 * HANDLE.
+	 *
+	 * @param str the input string
+	 * @param format the format string that was used by winpr_exportHandleToString with {} for the
+	 * handle
+	 * @returns the created HANDLE or INVALID_HANDLE if something wrong happened
+	 *
+	 * @since version 3.32.0
+	 */
+	WINPR_ATTR_NODISCARD
+	WINPR_API HANDLE winpr_importHandleFromString(const char* str, const char* format);
+
 #ifndef _WIN32
 
 #define DUPLICATE_CLOSE_SOURCE 0x00000001
--- a/winpr/include/winpr/sysinfo.h
+++ b/winpr/include/winpr/sysinfo.h
@@ -344,6 +344,49 @@ extern "C"
 #define GetComputerNameEx GetComputerNameExA
 #endif
 
+/** @brief mirrors the real Win32 ACLineStatus values */
+#define AC_LINE_OFFLINE 0x00
+#define AC_LINE_ONLINE 0x01
+#define AC_LINE_BACKUP_POWER 0x02
+#define AC_LINE_UNKNOWN 0xFF
+
+/** @brief mirrors the real Win32 BatteryFlag bit values (combinable, except UNKNOWN/NO_BATTERY) */
+#define BATTERY_FLAG_HIGH 0x01
+#define BATTERY_FLAG_LOW 0x02
+#define BATTERY_FLAG_CRITICAL 0x04
+#define BATTERY_FLAG_CHARGING 0x08
+#define BATTERY_FLAG_NO_BATTERY 0x80
+#define BATTERY_FLAG_UNKNOWN 0xFF
+
+#define BATTERY_PERCENTAGE_UNKNOWN 0xFF
+#define BATTERY_LIFE_UNKNOWN 0xFFFFFFFF
+
+	/** @brief mirrors the real Win32 SYSTEM_POWER_STATUS struct.
+	 *  @since version 3.32.0
+	 */
+	typedef struct
+	{
+		BYTE ACLineStatus;
+		BYTE BatteryFlag;
+		BYTE BatteryLifePercent;
+		BYTE SystemStatusFlag;
+		DWORD BatteryLifeTime;
+		DWORD BatteryFullLifeTime;
+	} SYSTEM_POWER_STATUS, *LPSYSTEM_POWER_STATUS;
+
+	/** @brief queries the local machine's AC/battery power state. Supported on Linux, Android,
+	 *  macOS and FreeBSD; on other non-Windows platforms this always reports "no battery,
+	 *  unknown state" rather than failing, matching the real API's behavior on desktop systems
+	 *  with no battery present.
+	 *
+	 *  @param lpSystemPowerStatus receives the current power status
+	 *  @return TRUE on success, FALSE (with GetLastError() set) if the status could not be
+	 *  determined at all
+	 *  @since version 3.32.0
+	 */
+	WINPR_ATTR_NODISCARD
+	WINPR_API BOOL GetSystemPowerStatus(LPSYSTEM_POWER_STATUS lpSystemPowerStatus);
+
 #endif
 
 #if (!defined(_WIN32)) || (defined(_WIN32) && (_WIN32_WINNT < 0x0600))
--- a/winpr/libwinpr/handle/handle.c
+++ b/winpr/libwinpr/handle/handle.c
@@ -21,6 +21,10 @@
 #include <winpr/config.h>
 
 #include <winpr/handle.h>
+#include <winpr/assert.h>
+
+#include "../log.h"
+#define TAG WINPR_TAG("handle")
 
 #ifndef _WIN32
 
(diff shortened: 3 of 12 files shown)


winpr/libwinpr/pipe/pipe.h
```diff
@@ -70,6 +70,8 @@ BOOL NamedPipeRead(PVOID Object, LPVOID lpBuffer, DWORD nNumberOfBytesToRead,
 BOOL NamedPipeWrite(PVOID Object, LPCVOID lpBuffer, DWORD nNumberOfBytesToWrite,
                     LPDWORD lpNumberOfBytesWritten, LPOVERLAPPED lpOverlapped);
 
+HANDLE winpr_Pipe_FromFd(int fd);
```
akallabeth, line 74: maybe add `WINPR_ATTR_NODISCARD`?
(and, when at it, also for the rest of the functions here?)

winpr/libwinpr/pipe/pipe.c
```diff
@@ -529,6 +529,28 @@ BOOL CreatePipe(PHANDLE hReadPipe, PHANDLE hWritePipe, LPSECURITY_ATTRIBUTES lpP
...
+	if (!pPipe)
+	{
+		WLog_ERR(TAG, "error allocating pipe");
+		return INVALID_HANDLE_VALUE;
+	}
+
+	pPipe->fd = fd;
+	WINPR_HANDLE_SET_TYPE_AND_MODE(pPipe, HANDLE_TYPE_ANONYMOUS_PIPE, WINPR_FD_READ);
+	pPipe->common.ops = &ops;
+	return (HANDLE)pPipe;
```
akallabeth: better use `return &pPipe->common;`

winpr/include/winpr/sysinfo.h
```diff
@@ -344,6 +344,49 @@ extern "C"
 #define GetComputerNameEx GetComputerNameExA
 #endif
 
+/** @brief mirrors the real Win32 ACLineStatus values */
+#define AC_LINE_OFFLINE 0x00
```
akallabeth, line 348: maybe use `enum WINPR_C23_ENUM_TYPE(uint32_t)` or similar to have (some) type and range checks?

winpr/include/winpr/sysinfo.h
```diff
@@ -344,6 +344,49 @@ extern "C"
...
+#define AC_LINE_ONLINE 0x01
+#define AC_LINE_BACKUP_POWER 0x02
+#define AC_LINE_UNKNOWN 0xFF
+
+/** @brief mirrors the real Win32 BatteryFlag bit values (combinable, except UNKNOWN/NO_BATTERY) */
+#define BATTERY_FLAG_HIGH 0x01
```
akallabeth, line 354: maybe use `enum WINPR_C23_ENUM_TYPE(uint32_t)` or similar to have (some) type and range checks?

winpr/libwinpr/handle/handle.c
```diff
@@ -39,11 +43,180 @@
...
-
 #include "../handle/handle.h"
-#include "../log.h"
-#define TAG WINPR_TAG("handle")
+
+#endif
+
+/* finds the "{}" placeholder in `format`, splitting it into the literal text before and after it.
+ * Returns FALSE (and leaves *prefixLen/*suffix untouched) if `format` has none. */
+static BOOL winpr_handle_format_split(const char* format, size_t* prefixLen, const char** suffix)
```
akallabeth, line 53: Maybe add `WINPR_ATTR_NODISCARD` to help the compiler? (now not really relevant, but when changing any code you get a warning for unused results)

winpr/libwinpr/handle/handle.c
```diff
@@ -39,11 +43,180 @@
...
+	return TRUE;
+}
+
+#ifndef _WIN32
+/* single-character tag identifying a handle's type in the exported value (see
+ * winpr_exportHandleToString()/winpr_importHandleFromString()), so a POSIX-side import knows how
+ * to rebuild the right kind of HANDLE around the fd. Only pipes (as returned by CreatePipe()) are
+ * supported for now; add a case here (and a matching one in winpr_importHandleFromString()) when
+ * another handle type needs to cross a fork()+exec(). */
+static char winpr_handle_export_type_tag(ULONG type)
```
akallabeth, line 74: Maybe add `WINPR_ATTR_NODISCARD` to help the compiler? (now not really relevant, but when changing any code you get a warning for unused results)

winpr/libwinpr/handle/handle.c
```diff
@@ -39,11 +43,180 @@
...
+		return FALSE;
+	}
+	const int valueLenInt = snprintf(value, sizeof(value), "%c%x", typeTag, fd);
+#endif
+	if (valueLenInt <= 0)
+		return FALSE;
+	const size_t valueLen = (size_t)valueLenInt;
+
+	const size_t suffixLen = strlen(suffix);
+	if (prefixLen + valueLen + suffixLen + 1 > outSz)
```
akallabeth: possible overflow, `suffixLen` is unbound, so is `prefixLen`. (yes, usage error, but still should be handled)

winpr/include/winpr/handle.h
```diff
@@ -38,6 +38,43 @@ extern "C"
...
+	 *
+	 * @param str the input string
+	 * @param format the format string that was used by winpr_exportHandleToString with {} for the
+	 * handle
+	 * @returns the created HANDLE or INVALID_HANDLE if something wrong happened
+	 *
+	 * @since version 3.32.0
+	 */
+	WINPR_ATTR_NODISCARD
+	WINPR_API HANDLE winpr_importHandleFromString(const char* str, const char* format);
```
akallabeth, line 76: Maybe add a length for the input string (or a second function that takes an additional length)?
It is most likely from a untrusted source, so having proper bounds might help.

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+#if defined(__linux__)
+
+static BOOL sysinfo_read_sysfs_string(const char* path, char* buffer, size_t size)
+{
+	FILE* fp = fopen(path, "r");
+	if (!fp)
+		return FALSE;
+
+	BOOL rc = FALSE;
+	if (fgets(buffer, (int)size, fp))
```
akallabeth: `WINPR_ASSERTING_INT_CAST` (and better also check the input size does fit the range)

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+	{
+		size_t len = strlen(buffer);
```
akallabeth, line 593: `strnlen`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+		while ((len > 0) && ((buffer[len - 1] == '\n') || (buffer[len - 1] == '\r')))
+			buffer[--len] = '\0';
+		rc = TRUE;
+	}
+	(void)fclose(fp);
```
akallabeth, line 602: maybe log an error here, so problems are visible to the user?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
 	return rc;
 }
 
+#if defined(__linux__)
+
+static BOOL sysinfo_read_sysfs_string(const char* path, char* buffer, size_t size)
+{
+	FILE* fp = fopen(path, "r");
+	if (!fp)
```
akallabeth, line 587: maybe log an error here that problems are visible to the user?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+	*value = v;
+	return TRUE;
+}
+
+/** @brief reads /sys/class/power_supply/, the same kernel interface on both desktop Linux and
+ *  Android (no ANDROID-specific carve-out needed: Android exposes the identical sysfs nodes). */
+static BOOL sysinfo_get_power_status_platform(LPSYSTEM_POWER_STATUS status)
+{
+	DIR* dir = opendir("/sys/class/power_supply");
+	if (!dir)
```
akallabeth, line 626: maybe log an error/debug here so problems are visible to the user?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+	while ((entry = readdir(dir)) != nullptr)
+	{
+		char path[PATH_MAX] = { 0 };
+		char value[64] = { 0 };
+		long long v = 0;
+
+		if (entry->d_name[0] == '.')
+			continue;
+
+		(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/type", entry->d_name);
```
akallabeth: better `sizeof(path)-1`

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+		if (!sysinfo_read_sysfs_string(path, value, sizeof(value)))
+			continue;
+
+		if (strcmp(value, "Battery") == 0)
+		{
+			(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/present", entry->d_name);
```
akallabeth: better `sizeof(path)-1`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+			if (sysinfo_read_sysfs_int(path, &v) && (v == 0))
+				continue; /* battery slot present in sysfs, no battery physically installed */
+
+			haveBattery = TRUE;
+
+			(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/status", entry->d_name);
```
akallabeth: better `sizeof(path)-1`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+			if (sysinfo_read_sysfs_string(path, value, sizeof(value)) &&
+			    (strcmp(value, "Charging") == 0))
+				charging = TRUE;
+
+			(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/capacity",
```
akallabeth: better `sizeof(path)-1`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+			               entry->d_name);
+			if (sysinfo_read_sysfs_int(path, &v))
+				percent = v;
+
+			(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/time_to_empty_now",
```
akallabeth: better `sizeof(path)-1`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+			               entry->d_name);
+			if (sysinfo_read_sysfs_int(path, &v) && (v > 0))
+				timeToEmpty = v;
+
+			/* not every driver exposes "capacity" directly; fall back to computing it from
+			 * energy_now/energy_full when present */
+			if (percent < 0)
+			{
+				(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/energy_now",
```
akallabeth: better `sizeof(path)-1`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+				               entry->d_name);
+				(void)sysinfo_read_sysfs_int(path, &energyNow);
+				(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/energy_full",
```
akallabeth: better `sizeof(path)-1`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+				               entry->d_name);
+				(void)sysinfo_read_sysfs_int(path, &energyFull);
+
+				if ((energyNow < 0) || (energyFull <= 0))
+				{
+					(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/charge_now",
```
akallabeth: better `sizeof(path)-1`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+					               entry->d_name);
+					(void)sysinfo_read_sysfs_int(path, &energyNow);
+					(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/charge_full",
```
akallabeth: better `sizeof(path)-1`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+				}
+
+				if ((energyNow >= 0) && (energyFull > 0))
+					percent = (energyNow * 100) / energyFull;
+			}
+		}
+		else
+		{
+			/* Mains / USB / Wireless: any of these reporting online means we're on AC power */
+			(void)snprintf(path, sizeof(path), "/sys/class/power_supply/%s/online", entry->d_name);
```
akallabeth: better `sizeof(path)-1`?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+			if (sysinfo_read_sysfs_int(path, &v) && (v != 0))
+				onAC = TRUE;
+		}
+	}
+	(void)closedir(dir);
```
akallabeth, line 718: maybe log an error/debug so problems are user visible?

winpr/libwinpr/sysinfo/sysinfo.c
```diff
@@ -566,6 +579,377 @@ BOOL GetComputerNameExW(COMPUTER_NAME_FORMAT NameType, LPWSTR lpBuffer, LPDWORD
...
+}
+
+#elif defined(__MACOSX__)
+
+static BOOL sysinfo_get_power_status_platform(LPSYSTEM_POWER_STATUS status)
+{
+	BOOL rc = TRUE;
+	CFTypeRef blob = IOPSCopyPowerSourcesInfo();
+	if (!blob)
+	{
```
akallabeth, line 765: Maybe log an error here so problems are user visible?

akallabeth (commented):
Overall LGTM.

(lots of) nitpicks

winpr/include/winpr/sysinfo.h
hardening: It's there to mimick the windows API, so better use defines IMHO...

winpr/include/winpr/handle.h
hardening: I don't see why, since when NULL terminated string aren't ok ?

akallabeth: @hardening they are (mostly) ok, but we've had multiple occasions where we get some string from the wire and need to ensure that the data is `NULL` terminated outside.
having a function take in the length as well eliminates that requirement.

(so, for such cases the version with length is easier to handle as the available data is usually a buffer and the length of it)

winpr/libwinpr/sysinfo/sysinfo.c
hardening: AFAICT not needed `fgets` puts a `\0` and is guaranteed to not read more than `size-1`

akallabeth: true. but also, `size` is known.

winpr/libwinpr/sysinfo/sysinfo.c
hardening: Here it's not an error, just an host without any battery

winpr/libwinpr/sysinfo/sysinfo.c
hardening: Not a good idea here as we test many files that can be legitimately missing

akallabeth: well, a `trace` level log would still be ok (naming it error was my mistake)

winpr/include/winpr/handle.h
hardening: Well these functions are there mostly to help mechanizing handle passing around `CreateProcess` so not the case you're citing, I'm keeping them with only `const char *`.

winpr/libwinpr/sysinfo/sysinfo.c
hardening: ok but no need to do the extra check, we're already in controlled conditions

winpr/libwinpr/sysinfo/sysinfo.c
hardening: This function may be called regularly when we implement the batterystatus channel, and IMHO the logs are way too verbose

winpr/include/winpr/sysinfo.h
hardening: Same here this is defines in win32 headers, let's do the same to avoid possible problems

winpr/libwinpr/sysinfo/sysinfo.c
akallabeth: @hardening why should there be a log if this call does not fail?

akallabeth:
@hardening there are a few minor warnings left at https://ci.freerdp.com/job/code-quality-checker/13242/clang/#

winpr/libwinpr/sysinfo/sysinfo.c
hardening: My remark was more that I don't want extra logs when it works, and the only "real" error cases will report `unable ro read string in XXXX` (as no battery is a legitimate case)

akallabeth: @hardening ok.
this here is more of a `closedir` fails (which would be quite surprising if opening worked)

hardening: Well if `closedir` fails it doesn't change anything, so I guess we can safely ignore the return value

akallabeth (approved):

PR #13316, opened 2026-09-01
[codec,h264] configure ffmpeg via env variables

akallabeth:
* Add FREERDP_HWCTX_DEVICE to override default acceleration device
* Add FREERDP_HWCTX_TYPE to set the hardware acceleration to use
* Update manpage


akallabeth:
@freerdp-bot test

PR #13324, opened 2026-09-03
[codec,fdk] ignore decoder errors

akallabeth:
ignore decoder errors, allows recovering stream after invalid data


akallabeth:
@freerdp-bot test

PR #13325, opened 2026-09-03
[warnings,s(n)printf] fix missing return value checks

akallabeth:
* explicitly cast to (void) if unused intentionally
* add return checks where possible


akallabeth:
@freerdp-bot test

PR #13327, opened 2026-09-03
[core,gateway] bound ARM gateway responses by their own timeout, not TcpConnectTimeout

sjtrotter:

Branch: `arm/response-timeout` (3 commits on `upstream/master`). Independent of the companion PRs #13328 (OAuth hardening), #13329 (sovereign clouds) and #13330 (rdp-file mapping); all four merge cleanly together.

## Problem

`http_response_recv()` waits for every byte of a gateway HTTP response under `FreeRDP_TcpConnectTimeout` (15 s by default). That bound describes how long a TCP connect may take; it has nothing to do with how long a server-side operation behind an established connection may run. The AVD ARM gateway holds `POST /api/arm/v2/connections` open while it orchestrates a session host and regularly needs more than 15 s before it sends the status line. The read times out, `arm_handle_request()` fails, and the connection is aborted with "error retrieving ARM configuration" right after a successful sign-in.

Raising `FreeRDP_TcpConnectTimeout` to 60 s is the known workaround (we shipped it in a Remmina plugin), but it lengthens every TCP connect in the session. Retrying is not a safe fix either: the request is not idempotent, and ARM's existing `retry` path only covers a *completed* HTTP 400 "session host deallocated" answer, never a missing one.

## Change

1. `[common,settings] add FreeRDP_GatewayArmResponseTimeout` — UINT32, default 60000 ms, at the first free gateway slot (2027); `sizeof(rdpSettings)` and existing indices unchanged; getters, name table, defaults, property list and JSON fixtures updated.
2. `[core,gateway] do not bound ARM responses by TcpConnectTimeout` — new `FREERDP_LOCAL http_response_recv_ex()` taking the bound; `http_response_recv()` wraps it with the old value so RDG and WST are byte-for-byte unchanged. Only `arm.c` uses the new setting, and logs both bounds once per resolve at DEBUG. The bound applies to the status/header phase and again to the body phase.
3. `[client,cmdline] add /gateway:arm-timeout` — `arm-timeout:<1-600000 ms>` sub-option, with TestClientCmdLine cases.

## Behaviour change

A genuinely dead ARM gateway now takes up to 60 s per phase to fail instead of 15 s; `/gateway:arm-timeout` lowers it. RDG and WST paths are unaffected.

## How to test

Unit: `ctest` (TestClientCmdLine, TestSettings, TestClientRdpFile fixtures). Manual: connect to an AVD workspace whose session host is deallocated or slow to start; previously the ARM step failed after ~15 s, now it waits up to the configured bound.

## Review history

Three adversarial review rounds (Claude, Codex, security) over the combined series before splitting. All reviewers rated this part as the clearest core-owned change; the only items raised were the per-phase timeout semantics (now documented) and a redundant cast (removed).

---

### DIFF
--- a/client/common/cmdline.c
+++ b/client/common/cmdline.c
@@ -4368,6 +4368,19 @@ static BOOL parse_gateway_options(rdpSettings* settings, const COMMAND_LINE_ARGU
 				allowHttpOpts = FALSE;
 			}
 
+			const char* to = option_starts_with("timeout:", argval);
+			if (to)
+			{
+				LONGLONG val = 0;
+				if (!value_to_int(to, &val, 1, 600000))
+					goto fail;
+				if (!freerdp_settings_set_uint32(settings, FreeRDP_GatewayResponseTimeout,
+				                                 (UINT32)val))
+					goto fail;
+				validOption = TRUE;
+				allowHttpOpts = FALSE;
+			}
+
 			if (allowHttpOpts)
 			{
 				if (option_equals(argval, "no-websockets"))
--- a/include/freerdp/settings_types_private.h
+++ b/include/freerdp/settings_types_private.h
@@ -536,7 +536,10 @@ struct rdp_settings
 	SETTINGS_DEPRECATED(ALIGN64 char* GatewayHttpMsUserAgent);      /** 2026
 		                                                             * @since version 3.18.0
 		                                                             */
-	UINT64 padding2112[2112 - 2027];                                /* 2027 */
+	SETTINGS_DEPRECATED(ALIGN64 UINT32 GatewayResponseTimeout);     /** 2027
+		                                                             * @since version 3.32.0
+		                                                             */
+	UINT64 padding2112[2112 - 2028];                                /* 2028 */
 
 	/**
 	 * RemoteApp
--- a/libfreerdp/core/gateway/http.c
+++ b/libfreerdp/core/gateway/http.c
@@ -1139,7 +1139,7 @@ static SSIZE_T http_response_recv_line(rdpTls* tls, HttpResponse* response)
 
 	SSIZE_T payloadOffset = -1;
 	const UINT32 timeoutMS =
-	    freerdp_settings_get_uint32(tls->context->settings, FreeRDP_TcpConnectTimeout);
+	    freerdp_settings_get_uint32(tls->context->settings, FreeRDP_GatewayResponseTimeout);
 	const UINT64 startMS = GetTickCount64();
 	while (payloadOffset <= 0)
 	{
@@ -1200,7 +1200,7 @@ static BOOL http_response_recv_body(rdpTls* tls, HttpResponse* response, BOOL re
 
 	const UINT64 startMS = GetTickCount64();
 	const UINT32 timeoutMS =
-	    freerdp_settings_get_uint32(tls->context->settings, FreeRDP_TcpConnectTimeout);
+	    freerdp_settings_get_uint32(tls->context->settings, FreeRDP_GatewayResponseTimeout);
 
 	if ((response->TransferEncoding == TransferEncodingChunked) && readContentLength)
 	{
--- a/client/common/cmdline.h
+++ b/client/common/cmdline.h
@@ -194,7 +194,7 @@ static const COMMAND_LINE_ARGUMENT_A global_cmd_args[] = {
 	  "g:<gateway>[:<port>],u:<user>,d:<domain>,p:<password>,usage-method:["
 	  "direct|detect],access-token:<"
 	  "token>,type:[rpc|http[,no-websockets][,extauth-sspi-ntlm]|auto[,no-websockets][,extauth-"
-	  "sspi-ntlm]]|arm,url:<wss://url>,bearer:<oauth2-bearer-token>",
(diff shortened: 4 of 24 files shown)


client/common/cmdline.c
```diff
@@ -4368,6 +4368,19 @@ static BOOL parse_gateway_options(rdpSettings* settings, const COMMAND_LINE_ARGU
 				allowHttpOpts = FALSE;
 			}
 
+			const char* armto = option_starts_with("arm-timeout:", argval);
```
akallabeth: renate to `timeout` maybe? it is already in a list of `/gateway:type:http,timeout:1234,...`

include/freerdp/settings_types_private.h
```diff
@@ -536,7 +536,10 @@ struct rdp_settings
 	SETTINGS_DEPRECATED(ALIGN64 char* GatewayHttpMsUserAgent);      /** 2026
 		                                                             * @since version 3.18.0
 		                                                             */
-	UINT64 padding2112[2112 - 2027];                                /* 2027 */
+	SETTINGS_DEPRECATED(ALIGN64 UINT32 GatewayArmResponseTimeout);  /** 2027
```
akallabeth: Better name it `GatewayResponseTimeout`, can be reused for any gateway type.

libfreerdp/core/gateway/http.c
```diff
@@ -1132,14 +1132,12 @@ static BOOL sleep_or_timeout_(rdpTls* tls, UINT64 startMS, UINT32 timeoutMS, con
...
 }
 
-static SSIZE_T http_response_recv_line(rdpTls* tls, HttpResponse* response)
+static SSIZE_T http_response_recv_line(rdpTls* tls, HttpResponse* response, UINT32 timeoutMS)
 {
 	WINPR_ASSERT(tls);
 	WINPR_ASSERT(response);
 
 	SSIZE_T payloadOffset = -1;
-	const UINT32 timeoutMS =
```
akallabeth, line 1141: why move this from here? could simply be replaced by the new timeout setting, no need for all the other changes.

libfreerdp/core/gateway/http.c
```diff
@@ -1191,16 +1189,14 @@ static SSIZE_T http_response_recv_line(rdpTls* tls, HttpResponse* response)
...
+                                    size_t payloadOffset, size_t bodyLength, UINT32 timeoutMS)
 {
 	BOOL rc = FALSE;
 
 	WINPR_ASSERT(tls);
 	WINPR_ASSERT(response);
 
 	const UINT64 startMS = GetTickCount64();
-	const UINT32 timeoutMS =
-	    freerdp_settings_get_uint32(tls->context->settings, FreeRDP_TcpConnectTimeout);
```
akallabeth, line 1203: why move that from here?

akallabeth (changes requested):
Might indeed be a good idea to split these two.
but:
* most of the changes here are not required
* it is already possible to change the `TcpConnectTimeout` from command line

client/common/cmdline.c
sjtrotter: done: the sub-option is now `/gateway:timeout:<ms>`

include/freerdp/settings_types_private.h
sjtrotter: done: renamed to `GatewayResponseTimeout`

libfreerdp/core/gateway/http.c
sjtrotter: the indirection was to keep the new bound ARM-only. dropped: http_response_recv_ex is gone, arm.c untouched, and the two reads in `http_response_recv` now just use `GatewayResponseTimeout`

libfreerdp/core/gateway/http.c
sjtrotter: same as above -- reverted to the original structure, just the setting name changed at this read.

sjtrotter:
thanks for the review! reworked to the more minimal form.

the idea was to leave the other gateway types unaffected. I hit the issue with the AVD ARM gateway; it had held the connection open while a host is brought up. when a host is running, response time is 1-2 seconds so i reverted the default back to 15s as well.

commits are amended in place and re based on current master.

akallabeth (approved):
LGTM,
nice addition, thank you!

akallabeth:
@freerdp-bot test

akallabeth:
@sjtrotter do you consider this ready?

sjtrotter:
> @sjtrotter do you consider this ready?

yep! marked active, ready for final review.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth (changes requested):
looks like you missed a few unit test files which now do not contain the new settings value.

sjtrotter:
Okay i'll check it out!

sjtrotter:
added the setting to the three testcase fixtures and the unchecked twins.

akallabeth:
@freerdp-bot test

sjtrotter:
noticed that jenkins failed (again? or maybe i never pushed the changes). not sure what happened, but rebased with ten fixtures this time and local ci-like checks pass.

akallabeth:
@freerdp-bot test

akallabeth (approved):

akallabeth:
thank you for the pr!

PR #13328, opened 2026-09-03
[client,common] AAD/AVD OAuth hardening: state, PKCE, validated callback parsing, safe logging

sjtrotter:

Branch: `aad/oauth-hardening` (9 commits on `upstream/master`). Independent of #13327 (ARM response timeout) and #13329 (sovereign clouds); all merge cleanly together.

## Why

Since 3.16, libfreerdp-client builds the AAD/AVD authorization and token URLs (`freerdp_client_get_aad_url()`) and exchanges the code for a token (`client_common_get_access_token()`). That makes it an OAuth public client in the RFC 6749/7636/9700 sense, but it did the first half only: no `state`, no PKCE, and the callback (the URL the browser lands on) was parsed separately and loosely by every frontend. The SDL webview matched the redirect as a string prefix; the stdin flow took the first `code=` it found anywhere; Remmina accepted the first redirect it saw; KRDC has a commented-out `GetAccessToken` where someone started and stopped. We hit all of this building a Linux client for AVD with Entra sign-in, and an independent security review of the Remmina flow concluded these are properties of the library-built exchange, not of any one frontend.

Two further issues surfaced on the way: the redirect-URI *format* settings are printf format strings used unvalidated in `winpr_asprintf` (`xfreerdp /azure:avd-token:%n` is a write primitive on master), and OAuth response bodies and callback URIs were logged at ERROR, including authorization codes and `error_description` text naming the account.

The browser itself, threads, cancellation and smart-card UI stay in the frontends. The contract becomes: host a browser, hand the library every URL it navigates to, get back CODE / ERROR / UNRELATED / INVALID.

## Commits

1. `[core,aad] validate tenant IDs for discovery` — the tenant id is checked (ASCII alnum, `-`, `.`, ≤128, not dots-only) before core interpolates it into the `.well-known` URL; a missing base/tenant is an error instead of an assert. New `TestAadTenant`.
2. `[core,aad] keep AAD response bodies out of the error log` — nonce response body and JSON parser detail move from ERROR to DEBUG.
3. `[core,aad] add a test seam for the wellknown configuration` — `FREERDP_LOCAL freerdp_utils_aad_set_wellknown()` installs a parsed discovery document so the OAuth tests run offline. Not public API; the tests that use it are gated by `BUILD_TESTING_INTERNAL`.
4. `[client,common] validate AVD redirect URI format strings` — `GatewayAvdAccessAadFormat`/ GatewayAvdAccessTokenFormat` accept only literal text, `%%` and at most the expected number of `%s`; tenant id validated; `/azure` help text documents it.
5. `[client,common] add state and PKCE to AAD authorization requests` — 256-bit `state`, RFC 7636 S256, `code_verifier` on the token request. The transaction lives in one pointer taken from `rdpClientContext`'s reserved padding (`rdpClientAadOAuth* aad_oauth`; size unchanged, compile-time asserted), freed on context teardown, one per context, released once the matching token request has been built (a failed build keeps it so the request can be retried). The `code` passed to the token request types is now percent-encoded into the body, so callers pass the percent-decoded code (behaviour change, see below).
6. `[client,common] add freerdp_client_aad_parse_callback` — exact scheme/host/port/path match after strict percent-decoding (malformed escapes and `%00` rejected), no userinfo/fragment, constant-time `state` compare, exactly one of `code`/`error`, duplicates rejected, the first terminal answer consumes the transaction (replays → INVALID). Replaces the stdin-paste parser; the pasted line is trimmed of whitespace and wrapping `<>`/quotes and scrubbed after use. Neither the URI nor any parsed value is logged.
7. `[client,common] keep the token endpoint response out of the error log` — body moves from ERROR to DEBUG.
8. `[utils,http] keep OAuth credentials out of HTTP debug dumps` — found during the hardware test: `freerdp_http_request()` dumps every request and response body at DEBUG, which for the token endpoint means the authorization code, the PKCE verifier and the access, refresh and id tokens land in any DEBUG log a user attaches to a bug report. A `FREERDP_LOCAL` redaction pass now masks the values of `code`, `code_verifier`, `client_secret`, `assertion`, `refresh_token` (form bodies) and `access_token`, `refresh_token`, `id_token` (JSON bodies) as `<redacted, N bytes>`; the discovery document, the nonce exchange and the non-secret token fields  (`token_type`, `scope`, `expires_in`) stay visible. Dumps are also skipped entirely when DEBUG is not active. 14 offline test cases (`TestHttpRedact`, built with `BUILD_TESTING_INTERNAL`).
9. `[client,sdl] use validated AAD callback parsing, stop logging URIs` — the webview navigation listener uses the parser (UNRELATED and INVALID keep navigating, CODE and ERROR finish), stops logging URIs/codes/errors, scrubs the token body, and no longer passes a null pointer to `std::string` when a request could not be built.

## Why credentials in logs matter (rationale for commits 2, 7, 8 and 9)

None of this is a replay of the RDP handshake; the nonce and proof-of-possession key protect that. The risk is plain credential theft from a file that people routinely attach to bug reports, paste into chat, back up, and leave in home directories. Before this series, a DEBUG log of a successful AVD sign-in contained, verbatim:

- the **authorization code** and **PKCE verifier**, adjacent: together they let a reader who gets there first complete the sign-in;
- the **AVD access token** (`token_type: Bearer`): possession is proof, so for about an hour a reader can call the AVD ARM and gateway APIs as the user, including opening a session to their desktop, with no card, PIN or conditional-access prompt;
- the **refresh token**: long-lived (typically 90 days, rolling), not PKCE-bound, not possession-bound; a reader can mint fresh access tokens for months. This is the artifact that turns a leaked log into standing access;
- the **id token**: not a credential, but signed PII (UPN, name, tenant, object id);
- on failure paths at ERROR level, `error_description` text that names the account.

(The session-host token is `token_type: pop` and bound to an in-memory key the log does not contain, so it is not usable by a reader; the redaction covers it anyway.) We observed exactly this during hardware testing of the series: a DEBUG log of one Government sign-in held a live refresh token. Commit 8 masks the field values in the HTTP dumps, commits 2 and 7 keep response bodies out of ERROR, and commit 9 stops the SDL webview logging callback URIs, which carry the authorization code.

## Compatibility

- ABI: `rdpClientContext` gains one pointer inside reserved padding; size unchanged.
- New public API (`@since version 3.32.0`): `rdpClientAadOAuth` (opaque), `freerdp_client_aad_callback_result`, `freerdp_client_aad_parse_callback()`, `freerdp_client_aad_reset()`. The variadic arguments of each `freerdp_client_aad_type` are now documented in `include/freerdp/client.h`.
- **Behaviour change** for `freerdp_client_get_aad_url()` callers: the `code` argument of the two `*_TOKEN_REQUEST` types must be the percent-*decoded* authorization code (what `freerdp_client_aad_parse_callback()` returns). A frontend that keeps passing the raw query value would double-encode a code containing an escape. A frontend that builds its own authorization URL (Remmina today) gets no PKCE and no transaction; a token request without a preceding authorization request still works and logs a warning.
- Builds with `WITH_AAD=OFF` (the OAuth tests skip).

## How to test

Unit: `ctest`; with `-DBUILD_TESTING_INTERNAL=ON` `TestHttpRedact` and the full `TestClientAad` (format matrix, tenant matrix, PKCE binding, transaction lifetime, 30+ callback classifications incl. NUL/malformed-escape/replay cases) runs offline against a seeded discovery document.
Manual (done 2026-09-03, US Government AVD, PIV): `sdl-freerdp workspace.rdpw` built with `WITH_WEBVIEW=OFF`; both `Browse to:` URLs carried `state`, `code_challenge` and `code_challenge_method=S256`; both pasted redirects (which Entra extended with an extra `session_state` parameter) were accepted; the token exchange, ARM orchestration and RDS-AAD negotiation succeeded and the desktop appeared.

Memory: the combined series' full suite ran under AddressSanitizer + LeakSanitizer with no reports; the URI decoder/parser was fuzzed with 18 M random URIs under ASan+UBSan and the format validator against a 49-entry corpus during review.

## Review history

Three adversarial review rounds (Claude, Codex, security). Notable outcomes: strict percent-decoding and duplicate/valueless-key rejection were added after a `%00` origin-check bypass was demonstrated; the transaction became single-use with replay protection; the webview stopped aborting on the first malformed hit; the well-known setter was demoted from public API to a test seam after a placement review. The hardware test also exposed the pre-existing DEBUG dump of token bodies (fixed here as commit 8). One pre-existing, test-only bug was found along the way (`TestClientRdpFile` double-frees when `TMPDIR` is unusable); a separate one-commit fix is proposed in #13332.

---

### DIFF
--- a/client/SDL/common/aad/CMakeLists.txt
+++ b/client/SDL/common/aad/CMakeLists.txt
@@ -19,7 +19,7 @@
 option(WITH_WEBVIEW "Build with WebView support for AAD login popup browser" OFF)
 if(WITH_WEBVIEW)
   set(SRCS sdl_webview.hpp webview_impl.hpp sdl_webview.cpp)
-  set(LIBS winpr)
+  set(LIBS winpr freerdp-client)
 
   include(FetchContent)
 
--- a/client/SDL/common/aad/sdl_webview.cpp
+++ b/client/SDL/common/aad/sdl_webview.cpp
@@ -20,8 +20,10 @@
 #include <string>
 #include <sstream>
 #include <cstdlib>
+#include <cstring>
 #include <memory>
 
+#include <winpr/crt.h>
 #include <winpr/string.h>
 #include <freerdp/log.h>
 #include <freerdp/utils/aad.h>
@@ -30,6 +32,20 @@
 #include "webview_impl.hpp"
 
 #define TAG CLIENT_TAG("SDL.webview")
+
+/** @brief Release a string that held credential material
+ *
+ *  A token request body carries the authorization code and the PKCE code verifier, so it is
+ *  scrubbed before it is released, like the bodies client/common builds for the paste flow.
+ *
+ *  @param str The string to scrub and release, may be \b nullptr
+ */
+static void free_secret(char* str)
+{
+	if (str)
+		SecureZeroMemory(str, strlen(str));
+	free(str);
+}
 
 static std::string from_settings(const rdpSettings* settings, FreeRDP_Settings_Keys_String id)
 {
@@ -71,20 +87,28 @@ static BOOL sdl_webview_get_rdsaad_access_token(freerdp* instance, const char* s
 	auto settings = context->settings;
 	WINPR_ASSERT(settings);
 
-	std::shared_ptr<char> request(freerdp_client_get_aad_url((rdpClientContext*)instance->context,
-	                                                         FREERDP_CLIENT_AAD_AUTH_REQUEST,
-	                                                         scope),
-	                              free);
+	auto cctx = reinterpret_cast<rdpClientContext*>(instance->context);
+	std::shared_ptr<char> request(
+	    freerdp_client_get_aad_url(cctx, FREERDP_CLIENT_AAD_AUTH_REQUEST, scope), free);
+	if (!request)
+	{
+		WLog_ERR(TAG, "failed to build the AAD authorization request");
+		return FALSE;
+	}
+
 	const std::string title = "FreeRDP WebView - AAD access token";
 	std::string code;
-	auto rc = webview_impl_run(title, request.get(), code);
+	auto rc = webview_impl_run(title, request.get(), cctx, code);
 	if (!rc || code.empty())
 		return FALSE;
 
-	std::shared_ptr<char> token_request(
-	    freerdp_client_get_aad_url((rdpClientContext*)instance->context,
-	                               FREERDP_CLIENT_AAD_TOKEN_REQUEST, scope, code.c_str(), req_cnf),
-	    free);
+	std::shared_ptr<char> token_request(freerdp_client_get_aad_url(cctx,
+	                                                               FREERDP_CLIENT_AAD_TOKEN_REQUEST,
+	                                                               scope, code.c_str(), req_cnf),
+	                                    free_secret);
+	if (!token_request)
+		return FALSE;
+
 	return client_common_get_access_token(instance, token_request.get(), token);
 }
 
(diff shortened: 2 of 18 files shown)


akallabeth:
some interesting changes (the log filter for example) but the OAuth parsing is most likely better off in an external library that actually specializes in that.
we most likely should drop the custom generation in favor of something like that.

sjtrotter:
Sorry for the early close before comment. I agree that the oauth maybe belongs somewhere else; do you have somewhere in mind already?

akallabeth:
@sjtrotter not yet. need to have a look at the available options (and the availability of these on different distributions/systems)

sjtrotter:
I have been doing a lot of thinking about it and was wondering if a freedesktop portal might be the right answer?

For my requirements i need the smartcard authentication, which i think translates cleanly into a portal (by itself) but i'm less sure about the oauth bits. Something like a WebAuthenticator portal-- something more generalized so that it could potentially support more protocols too (OAuth, SAML, etc). I'm sketching some stuff out on my github to see if it makes sense.

sjtrotter:
I've drafted out a couple of desktop portals for submission into freedesktop:

https://github.com/sjtrotter/xdg-desktop-portal-webauth

https://github.com/sjtrotter/xdg-desktop-portal-certificate < for the smartcard support

(again created with ai assistance)

hoping you might be able to review at your leisure to see the viability of supporting it; potentially #13340 could be affected/related? I had drafted a plugin/helper for it that nearly does the same thing.

akallabeth:
@sjtrotter ok, the auth flow changes have been merged, we can now harden the url passing.

akallabeth:
@sjtrotter ok, had a look at this and implemented the `OAuth2` `state` and `PKCE` with https://github.com/FreeRDP/FreeRDP/pull/13433 (along with unifying `zfree / znfree`)

Issue #11032, opened 2025-01-02
AzureAD required changes


akallabeth:
@akallabeth Sorry I don't have the time to actually do any patching, but since so many people seem to be having trouble with this I figure I can at least try  to provide some (hopefully) helpful hints.

- You can use `login.microsoftonline.com` for getting the openid-configuration no matter the tenant type.
- The redirect_uri *must* be `https://login.microsoftonline.com/common/oauth2/nativeclient` (or the ms-appx-web one). These values are hard-configured in azure and you can only use ones that are on the list.
- You changed the redirect uri in one place in common/client.c to `ms-appx-web://`. That won't work correctly for the command line (e.g. firefox will refuse to open an unknown url scheme and won't put the url in the address bar).
- For gov tenants the scope parameter needs to be `www.wvd.azure.us` instead of `www.wvd.microsoft.com`.
- If we have a username for AVD we  can use that to lookup the tenant. Bonus: we can also use it to prefill the username field in the browser login forms.
- If we don't have a specific tenantid we can skip retrieving the openid config because we won't learn anything new from it anyway.

Happy New Year to everyone!

_Originally posted by @fifthdegree in https://github.com/FreeRDP/FreeRDP/issues/10796#issuecomment-2567980975_

jbrewkeggin:
@fifthdegree @akallabeth @trizzle-git

I got this working for the Government AVD, using main branch, using the following changes:



Additionally, I used options: /azure:use-tenantid:off,ad:"login.microsoftonline.us"

and changed my "wvd endpoint pool" and "remoteapplicationprogram" values in my RDPW file.

acidrop:
> [@fifthdegree](https://github.com/fifthdegree) [@akallabeth](https://github.com/akallabeth) [@trizzle-git](https://github.com/trizzle-git)
> 
> I got this working for the Government AVD, on main branch, using the following changes:
> 
> client.c patch
> Additionally, I used options: /azure:use-tenantid:off,ad:"login.microsoftonline.us"
> and changed my "wvd endpoint pool" and "remoteapplicationprogram" values in my RDPW file.

Just wanted to add that the patch worked for me, too. I gave it a shot on Azure Public Cloud with both Azure Virtual Desktops and Windows 365 Cloud PCs, and it passed.

clst:
I can also confirm the patch working great applied on 3.13.0

Thanks a lot! 👍

adampankow:
[...]

PR #13329, opened 2026-09-03
[utils,aad] AVD sovereign clouds: cloud table, automatic selection from the ARM gateway host, /gateway:cloud:

sjtrotter:

Branch: `aad/sovereign-cloud` (3 commits on `upstream/master`). Independent of #13327 (ARM response timeout) and #13328 (OAuth hardening, whose PKCE is what keeps the transiting code unusable); all merge cleanly together.

## Why

FreeRDP only knows the commercial cloud: `libfreerdp/core/settings.c` hard-codes `login.microsoftonline.com`, the
`wvd.microsoft.com` scope, the redirect format and the AVD public client id as defaults. A `.rdpw` whose gateway lives in `*.wvd.azure.us` fails against `login.microsoftonline.us` unless every frontend carries its own copy of the Government constants (Remmina's copy is an 82-line header). The constants are public service metadata that belongs next to the commercial ones already in libfreerdp; *choosing* a cloud automatically is client policy and stays in libfreerdp-client.

## Commits

1. `[utils,aad] add sovereign cloud table for AVD authentication` (libfreerdp) — opaque `FreeRDP_AadCloud`, entries `commercial` and `usgov` only (`.wvd.azure.us` → authority `login.microsoftonline.us`, scope `https://www.wvd.azure.us/.default …`, and the `/common/oauth2/nativeclient` redirect registered for the AVD public client), `freerdp_utils_aad_cloud_by_name()`, `freerdp_utils_aad_cloud_for_gateway()` (case-insensitive suffix match on a label boundary), getters, and the explicit `freerdp_utils_aad_apply_cloud()` which writes the three cloud-owned strings, preallocating all replacements before transferring ownership so it cannot leave a partial update. `TestAadCloud` covers lookups, getters and apply. No other cloud is included; nothing unverified was added.
2. `[client,common] select the AVD sovereign cloud of an ARM gateway` (libfreerdp-client) — `freerdp_client_settings_apply_avd_cloud()` (`@since 3.32.0`): when `GatewayArmTransport` is set and the gateway host matches a cloud, apply each value only while that setting still holds the commercial default (an explicit `/azure:ad:`, `avd-scope:` or `avd-access:` is never overwritten); derive `GatewayAvdUseTenantid` from a non-default tenant id unless `/azure:use-tenantid` was given explicitly. It runs exactly once per parse: at the end of the public `.rdp` entry points for API users, and once after all command-line options (the command line loads an `.rdp` without applying, so a later option can still change the gateway or authority). New `/gateway:cloud:<commercial|usgov>`  applies a table entry at the point it is parsed (last option wins against `/azure:` on the same line), suppresses the automatic selection, and gets the same tenant derivation; an unknown name is a command-line error. API frontends call the function once before connecting. TestClientCmdLine cases cover automatic selection, mixed-case hosts, tenant derivation, `use-tenantid:off` honoured, customized scope surviving, non-ARM unchanged, explicit `cloud:` with and without a tenant id, `cloud:` versus a later or earlier `/azure:avd-scope:`, an `.rdp` followed by a commercial gateway or a custom authority on the command line, and the unknown-name error.
3. `[client,common] test AVD rdp file coverage` — synthetic `avd-usgov.rdp` / `avd-commercial.rdp` fixtures in a new `client/common/test/rdp-avd/` (outside the snapshot-scanned `rdp-testcases/`), asserting the AVD keys map as documented and that population selects the cloud for the Government file and leaves the commercial and non-ARM files alone.

## Behaviour change

ARM connections whose gateway host ends in `.wvd.azure.us` now authenticate against `login.microsoftonline.us` with the Government scope unless the operator set the authority/scope/format explicitly. Commercial behaviour is unchanged. Tenant-specific discovery is enabled for such connections only when a non-default tenant id is present and `use-tenantid` was not explicitly set.

## Design notes for reviewers

- The automatic selection was originally placed in core (`rdp_client_connect()`); two independent placement reviews argued that core can only guess intent from defaults and then needs undo state, whereas the parsers know what was given explicitly. It now lives in libfreerdp-client with an explicit override. Core keeps only data and an explicit setter.
- The Government redirect deliberately points at `login.microsoftonline.com/common/oauth2/nativeclient`: that is what the Remmina implementation validated on hardware, and FreeRDP issue #11032 documents that the AVD client's redirect is registered in Azure with only the scope changing per cloud. Re-verified on hardware on 2026-09-03 in both directions: the Government tenant issued codes to this redirect and the sign-in completed, and the sovereign form `https://login.microsoftonline.us/<tenant>/oauth2/nativeclient` was rejected by Entra with AADSTS50011 ("does not match the redirect URIs configured for the application a85cf173-…"). The commercial redirect is the only one registered for the AVD client; PKCE (companion PR) is what keeps the transiting code unusable.
- An earlier revision made any `.rdp` `aadtenantid` force `GatewayAvdUseTenantid`; dropped, because upstream made tenant use opt-in deliberately (f25be351b).

## How to test

Unit: `ctest` (TestAadCloud, TestClientCmdLine, TestClientRdpFile). Manual: `sdl-freerdp workspace.rdp` for a Government workspace; expect a DEBUG line from `freerdp_client_settings_apply_avd_cloud` and the authorize URL on `login.microsoftonline.us`. To force a cloud: `/gateway:g:<host>,type:arm,cloud:usgov`.

## Review history

Three adversarial review rounds over the combined series, a dedicated placement review (which moved the automatic selection out of core and made the cloud struct opaque), and a verification round on the split branch, which found that the explicit `cloud:` option skipped tenant derivation and that an `.rdp` followed by command-line options applied the selection twice; both fixed.

---

### DIFF
--- a/client/common/test/TestClientCmdLine.c
+++ b/client/common/test/TestClientCmdLine.c
@@ -107,6 +107,70 @@ static BOOL check_settings_smartcard_no_redirection(rdpSettings* settings)
 	return result;
 }
 
+static const char avd_commercial_scope[] =
+    "https%3A%2F%2Fwww.wvd.microsoft.com%2F.default%20openid%20profile%20offline_access";
+static const char avd_commercial_redirect[] = "https%%3A%%2F%%2F%s%%2F%s%%2Foauth2%%2Fnativeclient";
+static const char avd_usgov_scope[] =
+    "https%3A%2F%2Fwww.wvd.azure.us%2F.default%20openid%20profile%20offline_access";
+static const char avd_usgov_redirect[] =
+    "https%%3A%%2F%%2Flogin.microsoftonline.com%%2Fcommon%%2Foauth2%%2Fnativeclient";
+
+static BOOL check_avd_cloud(rdpSettings* settings, const char* authority, const char* scope,
+                            const char* redirect, BOOL useTenantid)
+{
+	const struct
+	{
+		FreeRDP_Settings_Keys_String key;
+		const char* expected;
+	} strings[] = { { FreeRDP_GatewayAzureActiveDirectory, authority },
+		            { FreeRDP_GatewayAvdScope, scope },
+		            { FreeRDP_GatewayAvdAccessAadFormat, redirect } };
+
+	for (size_t x = 0; x < ARRAYSIZE(strings); x++)
+	{
+		const char* val = freerdp_settings_get_string(settings, strings[x].key);
+		if (!val || (strcmp(val, strings[x].expected) != 0))
+		{
+			TEST_FAILURE("Expected %s = %s, but got %s!\n",
+			             freerdp_settings_get_name_for_key(strings[x].key), strings[x].expected,
+			             val ? val : "(null)");
+			return FALSE;
+		}
+	}
+
+	if (freerdp_settings_get_bool(settings, FreeRDP_GatewayAvdUseTenantid) != useTenantid)
+	{
+		TEST_FAILURE("Expected GatewayAvdUseTenantid = %d!\n", useTenantid);
+		return FALSE;
+	}
+
+	return TRUE;
+}
+
+static BOOL check_settings_avd_commercial(rdpSettings* settings)
+{
+	return check_avd_cloud(settings, "login.microsoftonline.com", avd_commercial_scope,
+	                       avd_commercial_redirect, FALSE);
+}
+
+static BOOL check_settings_avd_usgov(rdpSettings* settings)
+{
+	return check_avd_cloud(settings, "login.microsoftonline.us", avd_usgov_scope,
+	                       avd_usgov_redirect, FALSE);
+}
+
+static BOOL check_settings_avd_usgov_tenantid(rdpSettings* settings)
+{
+	return check_avd_cloud(settings, "login.microsoftonline.us", avd_usgov_scope,
+	                       avd_usgov_redirect, TRUE);
+}
+
+static BOOL check_settings_avd_usgov_custom_scope(rdpSettings* settings)
+{
+	return check_avd_cloud(settings, "login.microsoftonline.us", "custom-scope", avd_usgov_redirect,
+	                       FALSE);
+}
+
 typedef struct
 {
 	int expected_status;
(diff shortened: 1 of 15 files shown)


libfreerdp/core/aad.c
```diff
@@ -33,9 +33,109 @@
...
+{
+	const char* name;
+	const char* gateway_suffix;
+	const char* authority;
+	const char* avd_scope;
+	const char* avd_redirect_format;
+};
+
+static const FreeRDP_AadCloud aad_clouds[] = {
+	{ "commercial", nullptr, "login.microsoftonline.com",
```
akallabeth, line 58: is there some link with official documentation on these or do we need to just need to poll these periodically?

I've found https://learn.microsoft.com/en-us/cli/azure/cloud?view=azure-cli-latest#az_cloud_register so it would suggest we should try to load some `JSON` config for parsing custom data (from configuration/installed locations/user provided?)

client/common/test/TestClientCmdLine.c
```diff
@@ -107,6 +107,76 @@ static BOOL check_settings_smartcard_no_redirection(rdpSettings* settings)
 	return result;
 }
 
+static const char avd_commercial_scope[] =
```
akallabeth, line 110: * might be a good idea to move this to some external configuration files (with the defaults compiled in) so there is an easy way to update without recompile.
* if there is a way to query this somewhere it would be even better

libfreerdp/core/aad.c
sjtrotter: I'll do some research and see if I can find out!

client/common/test/TestClientCmdLine.c
sjtrotter: copy, reading as related to above(?) - doing research on querying and looking at implementing a conf file.

sjtrotter:
had the ai dig into it. there's an ARM metadata endpoint at:

https://management.azure.com/metadata/endpoints?api-version=2022-09-01

and the gov/china hosts that give the Entra authority and the ARM endpoint per cloud, but it doesn't carry the specific AVD fields we need auto-selection for -- the wvd.* scope resource and the gateway/rdweb suffix. They exist in the "Required FQDNs for Azure Virtual Desktop" docs:

https://learn.microsoft.com/en-us/azure/virtual-desktop/required-fqdn-endpoint 

https://docs.azure.cn/en-us/virtual-desktop/required-fqdn-endpoint 

https://learn.microsoft.com/en-us/entra/identity-platform/authentication-national-cloud

i haven't found an anonymous endpoint that lists all sovereign clouds or their AVD parameters. even az coud lists' defaults are a compiled table in azure-cli's cloud.py:

 https://github.com/Azure/azure-cli/blob/dev/src/azure-cli-core/azure/cli/core/cloud.py

definitely agree that some sort of cloud config file would make it easier to maintain/distribute.

sjtrotter:
dev'ed out a schema and config hierarchy based on the json config loader for review.

akallabeth:
@sjtrotter interesting.
I think we should bring #13340 into shape first and then add this.

I'll propose a `JSON` exchange format for the helpers there so we can provide the base data for `OAuth2` there and no longer create the `url` in `FreeRDP` but in the helper binary (which will have more options available)

akallabeth:
@sjtrotter ok, we've released a new version doing a big overhaul of the Azure/Entra authentication components.
could you rebase your work on that?

PR #13331, opened 2026-09-03
[client,cmdline] honor explicitly disabled multimon and smart-sizing

sjtrotter:
> **Disclosure:** This pull request, including the code and this description, was prepared with AI assistance (Anthropic Claude agents, reviewed adversarially by OpenAI Codex, orchestrated by the author).
> **Status:** The author has **not yet exercised these changes interactively**; they build and pass the unit tests (FreeRDP master, Fedora 44, gcc 16) and the new TestClientCmdLine cases cover the behaviour described below.

Branch: `cmdline/honor-disabled-display-options` (one commit on top of `upstream/master`).

## Problem

An `.rdp` file that enables multimon or smart sizing cannot be overridden by a later option on the same command line, because `/multimon`, `/smart-sizing` and `/dynamic-resolution` turned their setting on whenever the switch was present, regardless of the value or sigil:

- `/multimon:off` still enabled multimon; `-multimon` also enabled it (the option was `COMMAND_LINE_VALUE_OPTIONAL` only, so the sigil was never inspected).
- `/smart-sizing:off` errored because the option only accepted `<width>x<height>`; `-smart-sizing` enabled it.
- `parse_smart_sizing_options()` tripped "Smart sizing and dynamic resolution are mutually exclusive" even when the intent was to disable smart sizing.

We hit this shipping a Linux client for Azure Virtual Desktop profiles authored on a different display layout, where a windowed single-monitor fallback needs `file.rdp /multimon:off -smart-sizing`.

## Change

- `multimon` and `smart-sizing` become `COMMAND_LINE_VALUE_BOOL | COMMAND_LINE_VALUE_OPTIONAL` (precedent: `clipboard`), so `+`/`-` sigils are honoured and an explicit `off` parses, while `force` and `<width>x<height>` keep working.
- A small `option_bool_value()` helper distinguishes the `BoolValueTrue`/`BoolValueFalse` sentinel from a literal value.
- Disabling multimon clears `UseMultimon`, `ForceMultimon` and `MonitorIds`; disabling smart sizing clears `SmartSizing`. The mutual-exclusion check now only rejects an attempt to *enable* one option while the other is set.
- A literal that is neither `off` nor the option's documented value is rejected with `COMMAND_LINE_ERROR_UNEXPECTED_VALUE` instead of silently meaning "enable".
- `dynamic-resolution` deliberately stays `COMMAND_LINE_VALUE_BOOL` only. Adding `OPTIONAL` would make WinPR's space-separated parser swallow the next argument (`xfreerdp --dynamic-resolution host` would eat `host`), so only the `+`/`-` sigil form toggles it, exactly as before.
- Help format strings for multimon and smart-sizing now advertise `off`.

## Behaviour change (please review)

`-multimon` and `-smart-sizing` previously *enabled* the feature (identical to the bare or `+` form) because the minus sigil was never consulted. They now *disable* it, consistent with every other BOOL option. A command line that relied on the minus form to enable one of these gets the opposite result.

## Tests

`client/common/test/TestClientCmdLine.c` gains cases driven by a new fixture `client/common/test/rdp-cmdline/disabled-display-options.rdp` (turns multimon with selected monitors, smart sizing and dynamic resolution on) followed by `/multimon:force` + `/multimon:off`, `/multimon:force` + `-multimon`, `-smart-sizing`, `/smart-sizing:off`, `-dynamic-resolution`, asserting the targeted setting and its dependents are cleared and the other two are untouched. Regression cases: `--dynamic-resolution server.example` is not swallowed; `--multimon force` / `--smart-sizing 1024x768` keep their pre-existing space-separated behaviour; `/multimon:banana`, `/smart-sizing:garbage`, `/dynamic-resolution:off|on|banana` are rejected. The fixture lives outside `rdp-testcases/` on purpose: that directory is bulk-scanned by TestClientRdpFile and requires JSON snapshots.

Builds: FreeRDP master, `-DWITH_AAD=ON` and `OFF`; ctest 167/167.

## Review history (for transparency)

Two independent review rounds (Claude and Codex) before this draft. Round 1 found that an earlier version had added `OPTIONAL` to `dynamic-resolution`, regressing POSIX parsing; that was reverted. Round 2 found only a commit-message inaccuracy, fixed.

---


akallabeth (changes requested):
I think you did mix up something there...

akallabeth:
@freerdp-bot test

sjtrotter:
> I think you did mix up something there...

could you say which part? happy to adjust... did I miss something?

akallabeth:
@sjtrotter first draft looked like it was doing something else (or maybe github had a browser cache issue again, quite annoying lately)

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@sjtrotter do you consider this ready now?

(note: should change the unit test like the rdp parser test and just load the test cases and expected results from `cmdline/rdp/json` files instead of hard coding them in a separate pr)

akallabeth (commented):
please remove the AI comments in code, they are misleading.

sjtrotter:
Okay will remove the comments and apply your recommendations for the tests, then mark ready for review.

sjtrotter:
comments removed.
just making sure, you wanted the tests in a separate PR? ready to open/push if so.

akallabeth:
@freerdp-bot test

akallabeth (changes requested):
I'd propose a helper function like `parse_on_off_option` to return the parser result.

you can use `PARSE_ON_OFF_RESULT` and assume the optional result on `PARSE_NONE` and continue with handling, should make it easier to read.

sjtrotter:
> I'd propose a helper function like `parse_on_off_option` to return the parser result.
> 
> you can use `PARSE_ON_OFF_RESULT` and assume the optional result on `PARSE_NONE` and continue with handling, should make it easier to read.

gotcha -- looking into this now.

sjtrotter:
reworked to use the helper function.

client/common/cmdline.c
```diff
@@ -2162,6 +2162,21 @@ static PARSE_ON_OFF_RESULT parse_on_off_option(const char* value)
 	return PARSE_FAIL;
 }
 
+static PARSE_ON_OFF_RESULT parse_on_off_argument(const COMMAND_LINE_ARGUMENT_A* arg)
+{
+	WINPR_ASSERT(arg);
+
+	if (arg->Value == BoolValueTrue)
+		return PARSE_NONE;
```
akallabeth: should be `PARSE_ON`?

client/common/cmdline.c
```diff
@@ -2162,6 +2162,21 @@ static PARSE_ON_OFF_RESULT parse_on_off_option(const char* value)
...
+	if (arg->Value == BoolValueFalse)
+		return PARSE_OFF;
+	if (option_equals("on", arg->Value))
+		return PARSE_ON;
+	if (option_equals("off", arg->Value))
+		return PARSE_OFF;
+	return PARSE_FAIL;
```
akallabeth: should be `PARSE_NONE` as `PARSE_FAIL` would indicate an error?

client/common/cmdline.c
```diff
@@ -2908,7 +2923,19 @@ static int parse_dynamic_resolution_options(rdpSettings* settings,
 	WINPR_ASSERT(settings);
 	WINPR_ASSERT(arg);
 
-	const BOOL val = arg->Value != nullptr;
+	BOOL val = TRUE;
+
+	switch (parse_on_off_argument(arg))
+	{
+		case PARSE_NONE:
```
akallabeth: should not be allowed here?

client/common/cmdline.c
```diff
@@ -2929,21 +2956,37 @@ static int parse_smart_sizing_options(rdpSettings* settings, const COMMAND_LINE_
...
+		case PARSE_ON:
+			break;
+		case PARSE_OFF:
+			val = FALSE;
+			break;
+		default:
+			size = arg->Value;
```
akallabeth, line 2973: should fail the parsing

client/common/cmdline.c
```diff
@@ -2929,21 +2956,37 @@ static int parse_smart_sizing_options(rdpSettings* settings, const COMMAND_LINE_
 	WINPR_ASSERT(settings);
 	WINPR_ASSERT(arg);
 
-	if (freerdp_settings_get_bool(settings, FreeRDP_DynamicResolutionUpdate))
+	BOOL val = TRUE;
+	const char* size = nullptr;
+
+	switch (parse_on_off_argument(arg))
+	{
+		case PARSE_NONE:
```
akallabeth: should be the optional argument case

client/common/cmdline.c
```diff
@@ -4925,16 +4968,29 @@ static int parse_command_line(rdpSettings* settings, const COMMAND_LINE_ARGUMENT
...
 		{
-			if (!freerdp_settings_set_bool(settings, FreeRDP_UseMultimon, TRUE))
-				return fail_at(arg, COMMAND_LINE_ERROR);
-
-			if (arg->Flags & COMMAND_LINE_VALUE_PRESENT)
+			switch (parse_on_off_argument(arg))
 			{
-				if (option_equals(arg->Value, str_force))
-				{
+				case PARSE_NONE:
```
akallabeth: should be the optional case

client/common/cmdline.c
```diff
@@ -4925,16 +4968,29 @@ static int parse_command_line(rdpSettings* settings, const COMMAND_LINE_ARGUMENT
...
+				case PARSE_OFF:
+					if (!freerdp_settings_set_bool(settings, FreeRDP_UseMultimon, FALSE))
+						return fail_at(arg, COMMAND_LINE_ERROR);
+					if (!freerdp_settings_set_bool(settings, FreeRDP_ForceMultimon, FALSE))
+						return fail_at(arg, COMMAND_LINE_ERROR);
+					if (!freerdp_settings_set_pointer_len(settings, FreeRDP_MonitorIds, nullptr, 0))
+						return fail_at(arg, COMMAND_LINE_ERROR);
+					break;
+				default:
+					if (!option_equals(arg->Value, str_force))
```
akallabeth, line 4991: should be the failure case

akallabeth (changes requested):

sjtrotter:
ah shoot alright, i'll get the changes fixed.

sjtrotter:
okay reworked as suggested, lmk if i missed anything!

client/common/cmdline.c
```diff
@@ -2162,6 +2162,23 @@ static PARSE_ON_OFF_RESULT parse_on_off_option(const char* value)
...
 
+static PARSE_ON_OFF_RESULT parse_on_off_argument(const COMMAND_LINE_ARGUMENT_A* arg)
+{
+	WINPR_ASSERT(arg);
+
+	if (arg->Value == BoolValueTrue)
+		return PARSE_ON;
+	if (arg->Value == BoolValueFalse)
+		return PARSE_OFF;
+	if (strnlen(arg->Value, 1) == 0)
```
akallabeth, line 2173: why is this a failure?

client/common/cmdline.c
```diff
@@ -2162,6 +2162,23 @@ static PARSE_ON_OFF_RESULT parse_on_off_option(const char* value)
 	return PARSE_FAIL;
 }
 
+static PARSE_ON_OFF_RESULT parse_on_off_argument(const COMMAND_LINE_ARGUMENT_A* arg)
```
akallabeth, line 2165: `WINPR_ATTR_NODISCARD` does not hurt here

akallabeth (changes requested):

client/common/cmdline.c
sjtrotter: Hm. I'm not sure, I'll get it reviewed.

client/common/cmdline.c
sjtrotter: Ok! I'll take a look.

client/common/cmdline.c
sjtrotter: Ah, do you mean it should fail none?

akallabeth: why do you even bother to check this case?

sjtrotter: ah, gotcha. should be removed in the latest push.

sjtrotter:
new push. should be simplified and use the annotation. tests trimmed to match as well.

client/common/cmdline.c
```diff
@@ -4925,16 +4959,31 @@ static int parse_command_line(rdpSettings* settings, const COMMAND_LINE_ARGUMENT
...
+				case PARSE_ON:
+					if (!freerdp_settings_set_bool(settings, FreeRDP_UseMultimon, TRUE))
+						return fail_at(arg, COMMAND_LINE_ERROR);
+					break;
+				case PARSE_OFF:
+					if (!freerdp_settings_set_bool(settings, FreeRDP_UseMultimon, FALSE))
+						return fail_at(arg, COMMAND_LINE_ERROR);
+					if (!freerdp_settings_set_bool(settings, FreeRDP_ForceMultimon, FALSE))
+						return fail_at(arg, COMMAND_LINE_ERROR);
+					if (!freerdp_settings_set_pointer_len(settings, FreeRDP_MonitorIds, nullptr, 0))
```
akallabeth: ok, what is this now?

akallabeth (changes requested):

client/common/cmdline.c
sjtrotter: Ah it was clearing the monitor id's if multimon is off. realizing that the monitors can still be passed/be useful even if multimon is forced off. reverting this bit. (and updating the tests).

sjtrotter:
edits should be in, also rebased to master.

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13332, opened 2026-09-03
[client,common,test] fix TestClientRdpFile double free on bad TMPDIR

sjtrotter:
> **Disclosure:** This pull request, including the code and this description, was prepared with AI assistance (Anthropic Claude agents, orchestrated by the author).
> **Status:** Verified only by unit tests and AddressSanitizer on Fedora 44 / gcc 16 against FreeRDP master; no interactive testing needed for a test-only change.

Branch: `test/rdpfile-tmpdir` (one commit on top of `upstream/master`).

## Problem

`TestClientRdpFile` aborts with `free(): invalid pointer` when the temporary directory is unusable, for example:

```
TMPDIR=/nonexistent build/Testing/TestClient TestClientRdpFile
```

How it was found: while reviewing an unrelated series (#13328), one of our AI review agents ran the suite in a read-only sandbox whose `TMPDIR` does not exist and reported the abort. It did not reproduce under ctest, direct runs, glibc malloc checking, or AddressSanitizer in a normal environment, until the environment was narrowed to `TMPDIR` pointing at a missing directory. It reproduces on unmodified `upstream/master` (the same binary aborts before any of our changes), and AddressSanitizer then pinpointed the double free. This is a **test-only** bug; no
library code is affected and no shipped client can hit it.

## Cause

In `TestClientRdpFile()` the `rdpFile* file` used for the ASCII parsing checks is freed, and the same variable is reused later for a new `rdpFile*` only after `CreateDirectoryA()` under `GetKnownPath(KNOWN_PATH_TEMP)` succeeds. When that call fails the code jumps to `fail:` before the reassignment, and `fail:` frees the already-freed pointer again. AddressSanitizer confirms a heap-use-after-free in `freerdp_client_rdp_file_free()`. `GetKnownPath()` and `CreateDirectoryA()` behave correctly; this is purely a test bug.

## Change

- Move the "write the settings out to a temporary `.rdp` file" logic into its own helper with its own local `file`, so the earlier   pointer can never be revisited on the error path.
- Log why `GetKnownPath()`/`CreateDirectoryA()` failed, so an unusable temp directory is a clean non-zero test failure instead of an abort.
- Add a POSIX-only regression check that forks a child with `TMPDIR` pointing at a nonexistent directory and asserts the child exits cleanly rather than by signal.

## Verification

- `TMPDIR=/nonexistent`: exits with a clean failure and a log line naming the directory; no abort.
- `TMPDIR=/tmp`: exit 0 as before. Full `ctest`: 167/167.
- ASan build: the use-after-free is gone.

---

### DIFF
--- a/client/common/test/TestClientRdpFile.c
+++ b/client/common/test/TestClientRdpFile.c
@@ -7,9 +7,15 @@
 #include <winpr/windows.h>
 #include <winpr/path.h>
 #include <winpr/crypto.h>
+#include <winpr/environment.h>
 
 #include <freerdp/client/file.h>
 #include <freerdp/channels/rdpecam.h>
+
+#if !defined(_WIN32)
+#include <unistd.h>
+#include <sys/wait.h>
+#endif
 
 static const BYTE testRdpFileUTF16[] = {
 	0xff, 0xfe, 0x73, 0x00, 0x63, 0x00, 0x72, 0x00, 0x65, 0x00, 0x65, 0x00, 0x6e, 0x00, 0x20, 0x00,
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
Ok, the bug is real, this fix is a mess though.

akallabeth:
https://github.com/FreeRDP/FreeRDP/pull/13333

sjtrotter:
makes sense, saw your approach on #13333 - appreciate you taking a look!

PR #13333, opened 2026-09-03
[client,common] fix TestClientRdpFile

akallabeth:
* Ensure resources are not reused or double freed
* Move tests to subfunctions


akallabeth:
@freerdp-bot test

PR #13334, opened 2026-09-03
[channels,pulse] survive hot-removal of the active audio device

shakeelosmani:
## What

Hardens the PulseAudio backends (rdpsnd + audin) against the active audio device changing or disappearing mid-session.

## Trigger / real-world incident

SDL3 client (master f91da730), Linux Mint 22 / PipeWire, connected to an Azure Virtual Desktop host with `/sound:sys:pulse /microphone`. During a Teams call running on the built-in audio device, the user plugged in a USB-C headset; wireplumber switched the default device and migrated the live playback + capture streams onto it, and **the session froze at that moment** (stale image, unresponsive, had to be killed). The capture stream additionally delivered holes that were fed to the AAC encoder as samples:

```
[aac @ ...] Input contains (near) NaN/+-Inf
```

(same long-standing symptom as #6072). Follow-up to the robustness work merged in #13234.

## Bugs fixed

1. **rdpsnd/pulse — the freeze**: `pa_stream_begin_write()` can succeed while returning an empty buffer (e.g. the stream is suspended or migrating because the default device changed). `rdpsnd_pulse_play()` treated that as progress: with `length == 0` the write loop never advances, spinning forever **with the threaded mainloop lock held** — which starves the pulse mainloop thread (the stream can never become writable again) and hangs the channel thread that called `Play`. Bail out and drop the remainder of the sample instead.
2. **audin/pulse**: `audin_pulse_stream_request_callback()` ignored the `pa_stream_peek()` contract — the return value was unchecked, `length == 0` still called `pa_stream_drop()`, and a hole (`data == NULL`, `length > 0`, what a vanishing/suspended capture device produces) was forwarded to the receive callback as real sample data (a NULL/garbage buffer for the DSP encoder). Now: check the return, skip empty buffers, drop holes without forwarding.
3. **rdpsnd/pulse**: the stream state callback cleared `pulse->stream` on `PA_STREAM_FAILED`/`TERMINATED` without releasing the `pa_stream_new()` reference — leaked stream object on every device death.
4. **rdpsnd/pulse**: `rdpsnd_pulse_close()` tested `pulse->stream` once, then waited for `pa_stream_drain()`. If the stream failed while draining, the state callback cleared `pulse->stream` and close continued into `pa_stream_disconnect(NULL)` — an assertion abort inside libpulse.
5. **rdpsnd/pulse**: the context state callback releases `pulse->context` on `PA_CONTEXT_FAILED`, but `rdpsnd_pulse_connect()` dereferenced the context again after `pa_threaded_mainloop_wait()` without re-checking, and `rdpsnd_pulse_get_volume()` / `rdpsnd_pulse_set_volume()` checked context/stream *before* taking the mainloop lock. All pointer checks now happen under the lock / after each wait.

## Testing

- Built on master (f91da730) with SDL3 client, `WITH_PULSE=ON`.
- Simulated device swap/unplug without hardware: load `module-null-sink`, make it the default sink and its monitor the default source (streams migrate onto it), then `pactl unload-module` while the session runs — repeated cycles against a Windows 11 25H2 host with `/sound:sys:pulse /microphone`; client stays alive and responsive, audio recovers onto the previous device.
- Regular sessions (connect, logon, sound/mic channel load, clean shutdown) unaffected.

### DIFF
--- a/channels/audin/client/pulse/audin_pulse.c
+++ b/channels/audin/client/pulse/audin_pulse.c
@@ -327,9 +327,23 @@ static void audin_pulse_stream_request_callback(pa_stream* stream, size_t length
 	const void* data = nullptr;
 	AudinPulseDevice* pulse = (AudinPulseDevice*)userdata;
 	UINT error = CHANNEL_RC_OK;
-	pa_stream_peek(stream, &data, &length);
-	error =
-	    IFCALLRESULT(CHANNEL_RC_OK, pulse->receive, &pulse->format, data, length, pulse->user_data);
+
+	if (pa_stream_peek(stream, &data, &length) < 0)
+	{
+		WLog_Print(pulse->log, WLOG_WARN, "pa_stream_peek failed (%d)",
+		           pa_context_errno(pulse->context));
+		return;
+	}
+
+	/* length == 0: buffer empty, no fragment to drop.
+	 * data == nullptr with length > 0: a hole in the record stream (e.g. the
+	 * capture device vanished); it carries no samples, only drop it. */
+	if (length == 0)
+		return;
+
+	if (data)
+		error = IFCALLRESULT(CHANNEL_RC_OK, pulse->receive, &pulse->format, data, length,
+		                     pulse->user_data);
 	pa_stream_drop(stream);
 
 	if (error && pulse->rdpcontext)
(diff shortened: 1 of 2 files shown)


akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth (approved):
Overall LGTM,,

PR #13335, opened 2026-09-04
[winpr,library] implement GetModuleFileNameA on OpenBSD with getexecpath()

brad0:


winpr/libwinpr/library/library.c
```diff
@@ -378,6 +378,32 @@ DWORD GetModuleFileNameA(HMODULE hModule, LPSTR lpFilename, DWORD nSize)
...
+	char path[PATH_MAX] = WINPR_C_ARRAY_INIT;
+	const int rc = getexecpath(path, sizeof(path));
+
+	if (rc != 0)
+	{
+		SetLastError(ERROR_INTERNAL_ERROR);
+		return 0;
+	}
+
+	const size_t length = strlen(path);
```
akallabeth: prefer `strnlen`

winpr/libwinpr/library/library.c
```diff
@@ -378,6 +378,32 @@ DWORD GetModuleFileNameA(HMODULE hModule, LPSTR lpFilename, DWORD nSize)
 
 	CopyMemory(lpFilename, buffer, nSize - 1);
 	lpFilename[nSize - 1] = '\0';
+	SetLastError(ERROR_INSUFFICIENT_BUFFER);
+	return nSize;
+#elif defined(__OpenBSD__)
+	char path[PATH_MAX] = WINPR_C_ARRAY_INIT;
+	const int rc = getexecpath(path, sizeof(path));
```
akallabeth: why use the intermediate buffer `path` here?

akallabeth (changes requested):
not sure why you use a intermediate buffer instead of directly passing in the destination buffer/size?

winpr/libwinpr/library/library.c
brad0: I was mimicking the macOS path.

brad0:
Is this something more along the lines of what you were looking for?

akallabeth (approved):
LGTM

akallabeth:
@brad0 yes, looks good now.
thank you!

akallabeth:
@freerdp-bot test

brad0:
It is draft as I don't want it committed yet. But the review is useful.

akallabeth:
@brad0 trying to get rid of the static file path buffers, they tend to not work reliably anywhere (the path limits have shifted in recent decades and the 260 char limit from `MAX_PATH` is somewhat limiting now)

brad0:
As soon as I feel its ready I'll let you know. Just lining it up.

brad0:
Hrmm. With this changed now TestLibraryGetModuleFileName is failing.

```
13: TestLibraryGetModuleFileName: GetModuleFileNameA unexpectedly returned 0 instead of 2
  1/162 Test  #13: TestLibraryGetModuleFileName .............***Failed    0.03 sec
```

akallabeth:
@brad0 ah, the `getexecpath` does always add `\0` right?
then you'll need the intermediate buffer :/
but better `calloc` it with `nSize + 1`

brad0:
> @brad0 ah, the `getexecpath` does always add `\0` right? then you'll need the intermediate buffer :/ but better `calloc` it with `nSize + 1`

Yes, always NUL terminated.

```
int
getexecpath(char *buf, size_t size)
{
	size_t r;

	if (_execpath == NULL) {
		errno = ENOENT;
		return (-1);
	}
	if (strlen(_execpath) + 1 > size) {
		errno = ERANGE;
		return (-1);
	}
	(void) strlcpy(buf, _execpath, size);
	return (0);
}
```

brad0:
I tried this. It is failing with the same test just as above.

winpr/libwinpr/library/library.c
```diff
@@ -380,6 +380,42 @@ DWORD GetModuleFileNameA(HMODULE hModule, LPSTR lpFilename, DWORD nSize)
...
+	if (rc != 0)
+	{
+		free(path);
+		SetLastError(ERROR_INTERNAL_ERROR);
+		return 0;
+	}
+
+	const size_t length = strnlen(path, nSize + 1);
+
+	if (length < nSize)
```
akallabeth: this branch handling is not required.
would do a
```
memset(lpFilename, nSize);
memcpy(lpFileName, path, length);
```

this always works and ensures `\0` termination where `length < nSize`

akallabeth (changes requested):

akallabeth:
@brad0 should it fail with the suggestion as well I'd recommend stepping through the test with a debugger and check the output of the function. (both, size and string returned)

akallabeth:
@freerdp-bot test

winpr/libwinpr/library/library.c
akallabeth: oh, you're right, that is also broken.
fix in https://github.com/FreeRDP/FreeRDP/pull/13337 (hope I did not miss anything)

akallabeth:
@freerdp-bot test

akallabeth:
@brad0 our jenkins is back up and the pr looks fine now.
does it work for you as expected as well?

winpr/libwinpr/library/library.c
```diff
@@ -380,6 +380,38 @@ DWORD GetModuleFileNameA(HMODULE hModule, LPSTR lpFilename, DWORD nSize)
 	lpFilename[nSize - 1] = '\0';
 	SetLastError(ERROR_INSUFFICIENT_BUFFER);
 	return nSize;
+#elif defined(__OpenBSD__)
+	char* path = calloc(1, nSize + 1);
+
+	if (path == NULL)
```
akallabeth: we're using `nullptr` from `C23` (a fallback is provided when compiled with older standards by `winpr/wtypes.h`

winpr/libwinpr/library/library.c
```diff
@@ -380,6 +380,38 @@ DWORD GetModuleFileNameA(HMODULE hModule, LPSTR lpFilename, DWORD nSize)
 	lpFilename[nSize - 1] = '\0';
 	SetLastError(ERROR_INSUFFICIENT_BUFFER);
 	return nSize;
+#elif defined(__OpenBSD__)
+	char* path = calloc(1, nSize + 1);
```
akallabeth: possible `DWORD` overflow `nSize + 1`, replace with `nSize + 1ull`

akallabeth: also, do a check `nSize < SIZE_MAX - 1` (if 32bit `size_t` is still a thing)

winpr/libwinpr/library/library.c
```diff
@@ -380,6 +380,38 @@ DWORD GetModuleFileNameA(HMODULE hModule, LPSTR lpFilename, DWORD nSize)
...
+
+	free(path);
+
+	if (length >= nSize)
+	{
+		SetLastError(ERROR_INSUFFICIENT_BUFFER);
+		return nSize;
+	}
+
+	return (DWORD)length;
```
akallabeth: better use `WINPR_ASSERTING_INT_CAST(DWORD, length)` (just to be safe if some check before gets changed and this fails)

akallabeth (changes requested):

winpr/libwinpr/library/library.c
brad0: 32-bit size_t is not a thing. *BSD's have used 64-bit size_t since day 1.

brad0: I added the first suggestion.

winpr/libwinpr/library/library.c
```diff
@@ -364,6 +364,44 @@ static DWORD mac_get_module_file_name(char* lpFilename, uint32_t nSize)
 }
 #endif
 
+#if defined(__OpenBSD__)
+WINPR_ATTR_NODISCARD
+static DWORD openbsd_get_module_file_name(char* lpFilename, uint32_t nSize)
+{
+	char* path = calloc(1, PATH_MAX);
```
akallabeth: wtf?
this revision has
1. fixed size 260 maximum path length
2. out of bound write should `nSize` be `< PATH_MAX`

akallabeth (changes requested):

winpr/libwinpr/library/library.c
```diff
@@ -364,6 +368,59 @@ static DWORD mac_get_module_file_name(char* lpFilename, uint32_t nSize)
 }
 #endif
 
+#if defined(__OpenBSD__)
+WINPR_ATTR_NODISCARD
+static DWORD openbsd_get_module_file_name(char* lpFilename, uint32_t nSize)
+{
+	size_t size = nSize + 1;
```
akallabeth: integer overflow (again)

winpr/libwinpr/library/library.c
```diff
@@ -364,6 +368,59 @@ static DWORD mac_get_module_file_name(char* lpFilename, uint32_t nSize)
...
+	while (getexecpath(path, size) != 0)
+	{
+		if (errno != ERANGE)
+		{
+			free(path);
+			SetLastError(ERROR_INTERNAL_ERROR);
+			return 0;
+		}
+
+		size *= 2;
```
akallabeth: exponential growth, you're trying to kill the application with OOM?
use a constant growth value, like `size += PATH_MAX`

winpr/libwinpr/library/library.c
```diff
@@ -364,6 +368,59 @@ static DWORD mac_get_module_file_name(char* lpFilename, uint32_t nSize)
...
+			return 0;
+		}
+
+		path = tmp;
+	}
+
+	const size_t length = strnlen(path, size);
+
+	memset(lpFilename, 0, nSize);
+	memcpy(lpFilename, path, MIN(length, (size_t)nSize));
```
akallabeth: useless cast

akallabeth (changes requested):

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

akallabeth:
@brad0 so, is this ready?

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13345, opened 2026-09-05
[channels,urbdrc] report the negotiated device speed

uchouT:
ADD_DEVICE derived DeviceIsHighSpeed from bcdUSB, which states the USB version a device complies with rather than the speed it enumerated at.

Ask the bus instead. libusb_get_device_speed reports the negotiated speed, so add IUDEVICE::query_device_speed and map its result. bcdUSB remains the fallback for a speed libusb cannot determine or this build cannot name, since it does bound the speed from above.

Note that [MS-RDPEUSB] 2.2.11 distinguishes only full from high speed, so a SuperSpeed device is still announced as high speed. That is a limit of the protocol.

### DIFF
--- a/channels/urbdrc/common/urbdrc_types.h
+++ b/channels/urbdrc/common/urbdrc_types.h
@@ -125,6 +125,19 @@ enum device_descriptor_table
 	B_NUM_CONFIGURATIONS = 17
 };
 
+/**
+ * Speed a redirected device enumerated at.
+ */
+enum device_speed
+{
+	DEVICE_SPEED_UNKNOWN = 0,
+	DEVICE_SPEED_LOW,
+	DEVICE_SPEED_FULL,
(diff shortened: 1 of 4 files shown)


channels/urbdrc/common/urbdrc_types.h
```diff
@@ -125,6 +125,19 @@ enum device_descriptor_table
 	B_NUM_CONFIGURATIONS = 17
 };
 
+/**
+ * Speed a redirected device enumerated at.
+ */
+enum device_speed
```
akallabeth, line 131: `WINPR_C23_ENUM(uint32_t)` might be a good idea

akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13352, opened 2026-09-07
[channels,urbdrc] give the device speed a type

uchouT:
Follow-up to #13345, addressing the `WINPR_C23_ENUM_TYPE` suggestion there.

query_device_speed returned int, so enum device_speed was only a set of named constants: nothing related the accessor to it, and its underlying type never reached the ABI.

Return the enum instead, and fix its underlying type with WINPR_C23_ENUM_TYPE. urbdrc_main.h now names the enum, so it includes urbdrc_types.h rather than relying on every consumer to have included it first.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13365, opened 2026-09-09
l10n: add Swedish Android localization

yeager:
Adds complete Swedish Android resources for aFreeRDP and a small, optional gettext foundation for user-facing xfreerdp help.

- adds `values-sv/strings.xml` with all 182 resource identifiers aligned with the English catalog;
- keeps non-translatable values unchanged and validates all Android key, array and placeholder parity;
- adds an opt-out `WITH_XFREERDP_I18N` CMake option that compiles/installs `xfreerdp.mo` when gettext and libintl are available;
- localizes xfreerdp help under the standard locale, while leaving protocol/library error codes and diagnostic logging stable.

Validation: `msgfmt --check --check-format`, `msgcmp` against the xgettext-generated POT, Android resource key/array/placeholder parity, and `git diff --check`. Android Gradle lint could not run in this environment because no Android SDK is installed.

### DIFF
--- a/client/Android/Studio/freeRDPCore/src/main/res/values-sv/strings.xml
+++ b/client/Android/Studio/freeRDPCore/src/main/res/values-sv/strings.xml
@@ -1,0 +1,197 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="yes">Ja</string>
+    <string name="no">Nej</string>
+    <string name="cancel">Avbryt</string>
+    <string name="save">Spara</string>
+    <string name="cont">Fortsätt</string>
+    <string name="login">Logga in</string>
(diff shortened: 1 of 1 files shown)


akallabeth (changes requested):
Ok for the android part, but the `xfreerdp` related thing should move to another pr.
(there are lots of strings to translate so merging it in this state is a no go)

akallabeth (changes requested):
LGTM.

please squash the commits and force push so we have a clean history.

yeager:
@akallabeth Done — the PR now contains only the Android localization and has been squashed to commit 62b46b7. Thank you for the review.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

Issue #11712, opened 2025-06-27
Right Shift Key Ineffective in SDL3 Client on Windows


tsz8899:
In the FreeRDP SDL3 client running on the Windows platform, the right Shift key does not function as expected within the remote session.

Symptoms
When the right Shift key is pressed:
*   Combining it with letter keys (e.g., 'a') does not produce uppercase letters in the remote session (it remains 'a').
*   Combining it with number or symbol keys (e.g., '1' or ';') does not produce the corresponding shifted characters in the remote session (it remains '1' or ';').
*   In contrast, the left Shift key on the keyboard functions correctly, enabling proper uppercase input and symbol switching in the remote session.
This issue specifically affects the right Shift key input in the FreeRDP SDL3 client on Windows.

akallabeth:
Right shift is the default meta key.
Explained in help of sdl-freerdp.

PR #13369, opened 2026-09-09
[client,sdl] handle Windows E0 36 as right shift

SebastienGllmt:
## Problem

On Windows with a CJK IME, Right Shift can be delivered with the extended-key bit set, producing the platform scancode `E0 36`.

SDL [preserves](https://github.com/libsdl-org/SDL/blob/main/src/video/windows/SDL_windowsevents.c) this value (doesn't convert it to a regular right-shift), but its Windows scancode table maps the extended-key bit set version (`E0 36`) to `SDL_SCANCODE_UNKNOWN` [here](https://github.com/libsdl-org/SDL/blob/main/src/events/scancodes_windows.h)

Arguably maybe this should be changed in SDL itself, but I mention it here for two reasons:
1. Other projects have handled it on their side: https://github.com/glfw/glfw/issues/2050
2. Other people have mentioned it on this repo: #11712 (note: unclear if this user was on a CJK IME and freed up right-shift using `sdl-freerdp.json` like I did, or if they just didn't realize right-shift was reserved)

Given this, I leave it up to you if you want to merge this, or just have this PR help anybody else who hits this issue find the answer

## Why `E0 36` should become `36`

Right-shift is not differentiated from shift in general in either:
- Windows (docs explicitly states that Right Shift is not an extended key [here](https://learn.microsoft.com/en-us/windows/win32/inputdev/about-keyboard-input#extended-key-flag))
- RDP spec (defines the extended-key flag for keys such as Right Alt, but not Right Shift. See [here](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rdpbcgr/7acaec9f-c8a6-4ee9-87d6-d9b89cf56489))

## Implementation

I simply normalize the event to `RDP_SCANCODE_RSHIFT` when all of the following hold:

- the client is running on Windows;
- SDL mapped the event to `SDL_SCANCODE_UNKNOWN`
- `SDL_KeyboardEvent.raw` is `E0 36`;

## Testing

- `git diff --check` passes.
- The SDL3 client sources, including `sdl_input.cpp`, compile
  successfully with MSVC.

### DIFF
--- a/client/SDL/SDL3/sdl_input.cpp
+++ b/client/SDL/SDL3/sdl_input.cpp
@@ -550,7 +550,14 @@ bool sdlInput::extract(const std::string& token, uint32_t& key, uint32_t& value)
 
 bool sdlInput::handleEvent(const SDL_KeyboardEvent& ev)
 {
-	const UINT32 rdp_scancode = scancode_to_rdp(ev.scancode);
+	UINT32 rdp_scancode = scancode_to_rdp(ev.scancode);
+#if defined(_WIN32)
+	/* CJK IMEs on Windows may set the extended bit for Right Shift, producing
+	 * the invalid scan code E0 36. SDL maps that event to SDL_SCANCODE_UNKNOWN,
+	 * so normalize it in the same way as the native Windows client. */
+	if ((rdp_scancode == RDP_SCANCODE_UNKNOWN) && (ev.raw == 0xE036))
+		rdp_scancode = RDP_SCANCODE_RSHIFT;
+#endif
 	const SDL_Keymod mods = SDL_GetModState();
 
 	if (_hotkeysEnabled && (mods & _hotkeyModmask) == _hotkeyModmask)


akallabeth:
@SebastienGllmt is there an upstream SDL issue for this filed?

akallabeth:
@SebastienGllmt and could this affect other keys as well? (the `E0` bit set on other keys and then mapped to unknown)

SebastienGllmt:
> could this affect other keys as well?

Maybe in theory, but at least I didn't have any other issue and glfw also only handles this case specifically (see the specific lines [here](https://github.com/glfw/glfw/blob/92dcf4ce74f2e2554a98fea09be7c705c17daa5a/src/win32_window.c#L730-L732))

`E0` can't be stripped in general (there exists cases where you don't want to strip it, like in`E0 2A` for print screen), so unless somebody finds another case where this is required, it's a one-case only patch

> is there an upstream SDL issue for this filed?

Not that I could find (either in the SDL codebase, or in references in other projects like glfw). I don't know enough about SDL to know if this is truly something they should change or if this is intended behavior for them.

akallabeth:
@SebastienGllmt ok.
best way to find out is filing an issue I guess, could you do that?

will merge this for now, but would be great to get some clarification.

akallabeth:
@freerdp-bot test

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

Issue #12355, opened 2026-02-23
Clipboard file transfer is aborted in xfreerdp when clipboard contents changes during copy


alexbers:
**Describe the bug**
When copying files from an RDP server to the local machine via clipboard (e.g. paste in a local file manager), if clipboard contents changes during transfer (user copies something else locally or on the server), the ongoing file transfer is immediately aborted. 

There are three things that cause this behavior in client/common/client_cliprdr_file.c file:
1. cliprdr_file_context_notify_new_client_format_list function destroys all FUSE entries when the local clipboard changes, even though the server's file is still valid and accessable
2. cliprdr_file_context_notify_new_server_format_list function destroys all clip data entries including those with active LOCK_CLIPBOARD_DATA held by the server), even though the locked data is available on the server
3. CB_CAN_LOCK_CLIPDATA is commented out in the client's capability flags. Without the client advertising it the server may ignore LOCK_CLIPBOARD_DATA PDUs from the client, rendering the locking mechanism ineffective

**To Reproduce**

1. Connect to a Windows RDP server with xfreerdp (clipboard redirection enabled, built with the FUSE support)
2. On the remote desktop, right click on a directory with many files (or a large file) -> copy
3. In a local file manager, paste the copied content
4. While the transfer is in progress, copy something else to the clipboard (e.g. select text and ctrl-c or copy the remote file several times)
5. The ongoing transfer is aborted with an I/O error

**Expected behavior**
The file transfer should continue to completion regardless of clipboard changes. Changing the clipboard should not interrupt ongoing transfers

**Screenshots**
![Image](https://github.com/user-attachments/assets/6908be38-bb54-4bd1-9663-aa42934d31ae)

**Application details**
* FreeRDP version 3.22.1-dev0 (ce7258485). Previous versions have the same behavior
* Command line used: xfreerdp /proxy:socks5://127.0.0.1:8891 /u:DOMAIN\\user /v:rdp.somesite.com:3389 /smart-sizing:3600x2000 /size:1800x1000 /scale:180
[...]

PR #13374, opened 2026-09-09
[client,cliprdr] retain locked transfers across clipboard changes

yeager:
Changing the clipboard currently deletes FUSE inodes and cancels pending reads, including transfers backed by a server-side clipboard lock.

Retain retired clipboard selections while FUSE has lookup references, open files/directories or pending requests. Release the server lock once those references are gone. Selections that were never used are released at the next clipboard change. Repeated requests for the same locked file list reuse its existing inodes.

Related to #12355. This addresses the receiving-side lifetime. Local lock capability advertisement still needs work on sending-side snapshots; Windows/RDS interoperability is untested.

The new Linux test mounts the real FUSE filesystem and uses a test peer for file-content replies. It covers a pending read across a local clipboard change, opening another file in an old directory, remote clipboard changes, two concurrent selections, repeated file-list requests, 100 further clipboard changes, and releasing locks on disconnect. It skips if FUSE mounting is unavailable.

The original test failed before the fix. All four client tests pass with the final test, both in a Debug build and with AddressSanitizer (including leak detection), on Linux aarch64.

`ctest --test-dir build -R '^TestClient' --output-on-failure`


yeager:
Rebased onto current master without conflicts. The focused regression test remains part of the patch; local validation was limited by the reduced dependency configuration, which excludes client-common tests when no frontend is enabled.

yeager:
FUSE test investigation on trv2: the test reaches the mount setup, but the guest denies the FUSE mount with EPERM. The strace shows fusermount3 failing the mount before the regression scenario starts, so this is a VM capability restriction rather than a failure in the patch or its test logic.

yeager:
While running the new FUSE regression test, I found and fixed an infinite loop in file-list validation: a filename containing a normal dot, such as `test.bin`, retried the same character indefinitely. The scanner now advances past a non-`..` dot. I also start the test peer before exposing its FUSE files, so an early metadata request can receive a response.

Validation: `TestClientClipboardTransfer` passes on a real FUSE mount on Linux.

yeager:
Rebased onto current master (5d70e0c). Validation on trv2: built `TestClient` and ran `sudo ctest -R "^TestClientClipboardTransfer$"` on a real FUSE mount (1/1 passed).

Issue #7699, opened 2022-03-06
Vaapi improve DRI device selection on multi gpu setups


astralaster:
I did look up how the h264 code (freerdp/src/FreeRDP-2.5.0/libfreerdp/codec/h264_ffmpeg.c) uses Vaapi currently and it will always select the DRI device 128:
```
#ifdef WITH_VAAPI
#define VAAPI_DEVICE "/dev/dri/renderD128"
#endif
```
This will work correctly if device 128 is the GPU which supports vaapi. But on multi gpu setups the device 128 is often a nvidia gpu and the integrated intel gpu which should be used by vaapi is 129.

I looked up the vaapi code from intel themselves and they did implement the same hardcoded selection. But they chaged it on something more robust, which enumerates all devices and checks them if they are intel devices before using them for vaapi.

If I find the time I will try to improve this myself, but currently im not sure if one should copy the intel enumeration code or if it is possible to access it via the vaapi lib.

akallabeth:
@astralaster we´re happy to accept improvements regarding `VAAPI` support, but currently it is not an option that is suitable for generic deployment but mostly limited to dedicated setups (there are too many rough edges with hardware limitations)

PR #13375, opened 2026-09-09
[codec] try available VAAPI render devices on Linux

yeager:
On Linux, the FFmpeg backend always defaults to `/dev/dri/renderD128`. If that GPU cannot initialize VAAPI, it falls back to software even when another render device supports VAAPI.

Try the available render devices until VAAPI initialization succeeds, and retain the selected path. Explicit JSON and environment device overrides still select only that device. Other hardware backends keep their existing selection behavior.

Related to #7699. This selects a device that can initialize VAAPI; it does not rank GPUs or probe individual H.264 profiles.

Validation: built with FFmpeg 8 and hardware acceleration enabled. The device-selection test covers an unusable first device, gaps in device numbering, no usable devices, enumeration failure, explicit device success/failure, and a non-VAAPI backend. Physical multi-GPU validation is still needed.


yeager:
Rebased onto current master in 57b630a5f. Validation on trv2: configured with FFmpeg video and hardware acceleration enabled, built TestFreeRDPCodecHardwareDevice successfully, and the test passed. Runtime coverage with a physical VAAPI render node remains pending GPU passthrough.

yeager:
Validation on Linux with FFmpeg hardware acceleration enabled: `TestFreeRDPCodecHardwareDevice` passes. The test covers render-node scanning, explicit device selection, no-match/error paths, and a non-VAAPI device type.

yeager:
Rebased onto current master (5d70e0c). Updated validation configuration for the current build option: `-DWITH_FFMPEG_HWACCEL=ON`. Built and ran `TestFreeRDPCodecHardwareDevice` on trv2; 1/1 passed. The VM has no physical GPU passthrough, so this covers discovery and fallback paths rather than hardware decoding.

yeager:
Rebased onto current master (1e8b630). The VAAPI selection code and its focused test compile with WITH_FFMPEG_HWACCEL against FFmpeg 8 and libva headers. git diff --check and Gitleaks pass. The available local FFmpeg SDK does not provide a CMake-detectable version for the full hardware-acceleration target, and the GPU test host is currently unreachable, so live VAAPI decoding remains to be run there.

Issue #5226, opened 2019-01-29
Freerdp Scale every single screen


matzemagkekse:
I use Freerdp on many Ubuntu PCs. Now I have a Problem with the scaling in the remote session. I have 2 monitors at each workstation. Now I have the problem with different screen resolutions. I would like to scale each monitor individually with one command. Is that possible? 

Example, I have one monitor with 1920 x 1080 (24 inch) with scale 150 and a Monitor with 1600 x 900 (19,5 inch) with scale 100. Many of my users who work are 70 years old and they can not read such a small font anymore. Is this considered?

Thanks for answers, Matze

matzemagkekse:
> I use Freerdp on many Ubuntu PCs. Now I have a Problem with the scaling in the remote session. I have 2 monitors at each workstation. Now I have the problem with different screen resolutions. I would like to scale each monitor individually with one command. Is that possible?
> 
> Example, I have one monitor with 1920 x 1080 (24 inch) with scale 150 and a Monitor with 1600 x 900 (19,5 inch) with scale 100. Many of my users who work are 70 years old and they can not read such a small font anymore. Is this considered?
> 
> Thanks for answers, Matze

alandoyle95:
would love to have this, using this in a homemade thin client OS using arch for a company and they have 60+ users that don't have matching screens, and some of them have vision issues, so scaling would be great.

PR #13377, opened 2026-09-09
[client,x11] allow scale factors per monitor

yeager:
Add `/monitor-scale:id:desktop:device[,id:desktop:device...]` to the X11 client. For example:

```
xfreerdp /v:server /multimon /monitor-scale:0:150:100,1:100:100
```

IDs are those shown by `/list:monitor`. Unspecified monitors use the global scale settings. Each selected monitor carries its scale values in the initial monitor attributes and subsequent Display Control layouts. Desktop scale must be 100–500; device scale must be 100, 140 or 180. Invalid values and duplicate IDs are rejected.

Related to #5226. These are server-side DPI scale requests, not client-side bitmap scaling. Windows/application behavior with mixed DPI still needs live validation.

Built the X11 client and passed an Xvfb test with two XRandR monitors. The test checks command-line parsing, detected monitor attributes, the layout sent after an XRandR change, global fallback, and malformed input without partially changing the configuration.


yeager:
Rebased onto current master and resolved the X11 test CMake conflict. TestX11MonitorScale builds and passes under Xvfb.

client/X11/test/CMakeLists.txt
```diff
@@ -5,4 +5,11 @@ if(XVFB_RUN_EXECUTABLE AND WITH_XRANDR AND X11_Xrandr_FOUND)
   add_test(NAME TestX11Monitor COMMAND ${XVFB_RUN_EXECUTABLE} -a -s "-screen 0 3120x1920x24 -nolisten tcp"
                                        $<TARGET_FILE:TestX11Monitor>
   )
+
+  add_executable(TestX11MonitorScale TestX11MonitorScale.c)
+  target_include_directories(TestX11MonitorScale PRIVATE ..)
+  target_link_libraries(TestX11MonitorScale PRIVATE xfreerdp-client freerdp-client freerdp winpr ${X11_Xrandr_LIB})
```
akallabeth, line 11: better use the imported targets, see https://cmake.org/cmake/help/latest/module/FindX11.html

client/X11/xf_monitor.h
```diff
@@ -25,9 +25,14 @@
 #include <freerdp/types.h>
 
 #include "xf_types.h"
+#include <winpr/cmdline.h>
 
 FREERDP_API int xf_list_monitors(xfContext* xfc);
 FREERDP_API BOOL xf_detect_monitors(xfContext* xfc, UINT32* pMaxWidth, UINT32* pMaxHeight);
 FREERDP_API void xf_monitors_free(xfContext* xfc);
 
+COMMAND_LINE_ARGUMENT_A* xf_monitor_arguments(size_t* count);
```
akallabeth, line 34: `WINPR_ATTR_NODISCARD`

client/X11/xf_monitor.h
```diff
@@ -25,9 +25,14 @@
...
+int xf_monitor_handle_option(const COMMAND_LINE_ARGUMENT_A* arg, void* userData);
```
akallabeth, line 35: `WINPR_ATTR_NODISCARD`

client/X11/xf_monitor.c
```diff
@@ -52,6 +53,74 @@
...
+		if (*value++ != ',')
+			return COMMAND_LINE_ERROR_UNEXPECTED_VALUE;
+	}
+	memcpy(xfc->monitorScales, scales, sizeof(scales));
+	return 0;
+}
+
+void xf_monitor_apply_scale(xfContext* xfc, rdpMonitor* monitor)
+{
+	const rdpSettings* settings = xfc->common.context.settings;
```
akallabeth, line 108: `WINPR_ASSERT(xfc); WINPR_ASSERT(monitor);`

akallabeth:
any specific reason you only added that to `xfreerdp` and not `client/common`?

yeager:
Rebased onto current master (5d70e0c) and addressed the review comments: the new test links `X11::Xrandr`, result-returning declarations use `WINPR_ATTR_NODISCARD`, and scale helpers validate their inputs. Validation on trv2: built and ran `TestX11MonitorScale` under Xvfb; 1/1 passed.

yeager:
The option is X11-specific in this revision because its IDs are the XRandR monitor IDs reported by /list:monitor, and the scale values can only be applied after the X11 frontend has enumerated those monitors. Adding the option to client/common without a frontend-neutral monitor identifier would expose the same spelling with different semantics in other clients.

I can move the parsed scale map into rdpSettings and provide a common helper that applies it to frontend monitor definitions if that is the intended cross-client API. Please let me know whether that is preferred; it requires defining a stable frontend-neutral monitor-ID contract.

yeager:
Rebased onto current master (1e8b630). Reconfigured the X11 build from scratch, built TestX11MonitorScale, and ran it under Xvfb successfully (1/1). git diff --check and Gitleaks also pass.

PR #13387, opened 2026-09-10
[channels,rdpewa] prefer built-in user verification

yeager:
Honour the CTAP `uv` option for credential creation and assertions.

Authenticators with built-in user verification now attempt it before requesting a PIN. If verification is blocked or invalid, the client retries with a PIN when one is available.

The authenticator list now reports built-in verification status instead of PIN status.

Validation: `git diff --check`; formatted with `git clang-format`. A local build cannot reach the rdpewa target because this environment lacks FFmpeg development libraries.

### DIFF
--- a/channels/rdpewa/client/rdpewa_fido.c
+++ b/channels/rdpewa/client/rdpewa_fido.c
@@ -66,6 +66,34 @@ static void zfree(char* str)
 	free(orig);
 }
 
+static bool rdpewa_fido_get_pin(rdpContext* context, char** pin)
+{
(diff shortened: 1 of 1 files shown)


akallabeth (commented):
Look ok, but need to test this. (will take a few days until I get my hands on all required keys)

akallabeth:
@freerdp-bot test

akallabeth:
already merged as part of https://github.com/FreeRDP/FreeRDP/pull/13399

PR #13388, opened 2026-09-10
[client,sdl] add fullscreen floatbar

yeager:
Adds a small top-level control bar while the SDL3 client is fullscreen.

The three controls minimize the client, leave fullscreen, or disconnect. The bar is an SDL popup and its input is consumed locally. It is created and destroyed on the SDL UI thread.

Validation: `git diff --check`; formatted with `git clang-format`. Runtime testing still needs an SDL fullscreen session on GNOME/Wayland.


yeager:
Validation update: built `sdl3-freerdp` successfully with a reduced local configuration, in addition to `git diff --check` and `git clang-format`.

client/SDL/SDL3/sdl_context.hpp
```diff
@@ -218,6 +220,8 @@ class SdlContext
 	bool _resizeable = false;
 	bool _grabMouse = false;
 	bool _grabKeyboard = false;
+	SDL_Window* _floatbar = nullptr;
```
akallabeth, line 223: I'd suggest to move the `Floatbar` to a separate class wrapping implementation details

akallabeth (changes requested):
when running with `/f +floatbar` I get:
```
[13:36:13:189] [388955:0005ef5b] [ERROR][com.freerdp.client.SDL] - [sdl_run]: [exception] sdl->setFloatbar {SDL_EVENT_USER_FLOATBAR}{SDL:Popup windows must specify either the 'SDL_WINDOW_TOOLTIP' or the 'SDL_WINDOW_POPUP_MENU' flag}{sdl_run [/home/user/src/freerdp/client/SDL/SDL3/sdl_freerdp.cpp:276-14]}
[13:36:13:197] [388955:0005ef88] [ERROR][com.freerdp.core] - [freerdp_check_event_handles]: freerdp_check_fds() failed - 0

```

yeager:
Fixed in 91c502846. The floatbar now uses SDL_WINDOW_POPUP_MENU, which SDL3 requires for popup windows that receive pointer input. Rebuilt sdl3-freerdp successfully.

yeager:
The SDL3 popup-menu correction is pushed and the affected target builds successfully. @akallabeth Could you please re-review when convenient?

yeager:
Follow-up in the latest commit: the SDL floatbar now honors the established FreeRDP_Floatbar setting. It is not created unless +floatbar is enabled, and its visibility follows show:fullscreen, show:window, or show:always. Rebuilt sdl3-freerdp successfully.

yeager:
Rebased onto current master and resolved the conflicts while preserving both the new SDL RAIL event handling and the floatbar event path. The branch is updated. A local SDL build now reaches final linking but fails on an existing unresolved client_rail_server_start_cmd reference from the updated master baseline; the floatbar sources compile before that link step.

yeager:
Rebased onto current master. The popup now uses SDL_WINDOW_POPUP_MENU, which addresses the reported SDL requirement. GitHub reports the branch as mergeable.

akallabeth (changes requested):
does not work.

yeager:
The reported SDL popup flag error is fixed in the current branch. Could you share the exact command line and the observed failure for the current head? I want to reproduce the remaining problem before changing the implementation further.

akallabeth:
@yeager please run your code before you declare it fixed.
there simply is no floatbar there, it is not crashing/disconnecting but that is about all this does.

yeager:
The floatbar is now a dedicated `SdlFloatbar` component. It owns popup creation, rendering, cleanup, window identification, and click-to-action mapping; `SdlContext` only dispatches the resulting action. The popup retains `SDL_WINDOW_POPUP_MENU`, which fixes the reported SDL error with `/f +floatbar`.

Validation: built `sdl3-freerdp` from this branch with the minimal client preset and SDL enabled.

yeager:
Found and fixed the remaining floatbar placement issue. A popup at `(0, 0)` was partly occluded by its parent on GNOME/X11; it is now centered along the parent window’s top edge.

Validation: rebuilt `sdl3-freerdp`, then created the production `SdlFloatbar` in an isolated GNOME/X11 session. The popup rendered visibly with all three controls.

yeager:
Rebased onto current master (5d70e0c) and rebuilt the SDL3 client on GNOME/X11 test host trv2. The build completed successfully. The floatbar remains a separate component and uses SDL popup-menu flags required by SDL3.

yeager:
Rebased onto current master (1e8b630). Reconfigured and built sdl3-freerdp from scratch. git diff --check and Gitleaks pass.

PR #13389, opened 2026-09-10
[codec,h264] fix broken MediaFoundation decoder build

tsz8899:
## Summary

Commit 8fbe04b ("[codec,h264] fix missing stride adjustments for Media Foundation") left `libfreerdp/codec/h264_mf.c` in a state that does not compile (reproduced with MSVC 2022, ninja `FAILED: [code=2]` on `h264_mf.c`):

* a stray closing brace `}` after `getSrcOffset()`
* type-less declarations in the copy loops (`const srcOffset = ...`, `const dstOffset = ...`, `const dstUOffset/dstVOffset = ...`) — implicit int is not valid C
* `getSrcOffset()` returned `line` instead of the computed `offset`, so the stride / bottom-up adjustments never took effect

## Changes

* remove the stray closing brace
* add the missing `UINT32` type specifiers
* return the computed `offset` from `getSrcOffset()`, restoring the intended behavior of 8fbe04b

### DIFF
--- a/libfreerdp/codec/h264_mf.c
+++ b/libfreerdp/codec/h264_mf.c
@@ -191,8 +191,7 @@ static UINT32 getSrcOffset(H264_CONTEXT_MF* sys, UINT32 line, BOOL isUV)
 	offset *= sys->stride;
 	if (isUV)
 		offset /= 2u;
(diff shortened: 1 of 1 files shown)


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

PR #13391, opened 2026-09-10
[client,sdl] add virtual monitor switching

yeager:
Adds `/sdl-virtual-monitors:<count>` to the SDL3 client.

The option reports equally sized monitors based on the selected physical display while creating one local window. `Ctrl+Alt+PageUp` and `Ctrl+Alt+PageDown` switch the visible virtual monitor and translate local pointer coordinates to that monitor.

Validation: `git diff --check`, `git clang-format`, and a reduced SDL3 build of `sdl3-freerdp`.

### DIFF
--- a/client/SDL/SDL3/sdl_context.cpp
+++ b/client/SDL/SDL3/sdl_context.cpp
@@ -19,6 +19,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <cstdlib>
 
 #include "sdl_context.hpp"
 #include "sdl_config.hpp"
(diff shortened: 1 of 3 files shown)


hardening:
Is it me or your PR is a subset of what is done in https://github.com/FreeRDP/FreeRDP/pull/13149 ?

yeager:
Thanks for pointing this out. #13149 covers the same use case with a more complete virtual-monitor layout and dynamic-resolution handling, so I am closing this PR in favor of that work.

PR #13401, opened 2026-09-12
[client,android] horizontal scroll and middle click

svncibrahim:
Follow-up on the touch pointer review: horizontal scroll and middle click.

Scroll used to send one notch per fixed amount of drag, so long pages meant dragging over and over. Now you press and hold and how far you move sets the speed. Horizontal works the same way, middle click is a new button, and I redrew the three mouse icons to match.

Wheel flag encoding moved into one helper in `Mouse`, same values as before.

image

**Tested on OPD2514, Android 16.**


akallabeth (approved):
LGTM,
nice refinement!

akallabeth:
@freerdp-bot test

PR #13402, opened 2026-09-12
[client,android] add touchpad two-finger scroll

svncibrahim:
Two-finger scroll on an attached touchpad did nothing in a session.

Since Android 14 the gesture no longer arrives as `ACTION_SCROLL`. It comes in on the touch path as `ACTION_MOVE` with `CLASSIFICATION_TWO_FINGER_SWIPE` and the travel in `AXIS_GESTURE_SCROLL_X/Y_DISTANCE`, so the mouse wheel handler never saw it. This path is only needed on API 34 and later, mice and older devices keep the existing one.

`SessionView` now reads those axes and turns the distance into wheel notches, batched historical samples included. `ScrollView2D` no longer mistakes the gesture for a drag, which otherwise pans the viewport instead of scrolling the session.

**Tested on OPD2514, Android 16 (API 36), with the attached keyboard touchpad.**


akallabeth:
@freerdp-bot test

akallabeth (commented):
LGTM

akallabeth (approved):

PR #13406, opened 2026-09-14
[core,rdstls] Fix version validation macro

ondrejholy:
Currently, `rdstls_is_version_supported` ignores its second argument because
of the `versions` versus `version` typo. This can select a different local
variable or fail to compile. However, all calls pass local `version`, so this
works currently and does not cause a bug. Let us pass the supplied argument
instead.

Co-Authored-By: GPT-5 <noreply@openai.com>


akallabeth (approved):
LGTM, nice catch!

akallabeth:
@freerdp-bot test

PR #13415, opened 2026-09-15
[core,security] add ExtSecurity to all clients and server

akallabeth:
* Add missing configuration options for ExtSecurity
* Promote ExtSecurity to first class mode (same as NLA)


akallabeth:
@freerdp-bot test

PR #13416, opened 2026-09-15
[winpr,utils] fix UTF-16 stream write length and padding

yerbolgmailcom:
## Problem

After 49838ba4abfa67bcf508d3bca869f2cfcafbc9e6, `Stream_Write_UTF16_String_From_UTF8` leaves its return count at zero, advances by a character count expressed as bytes, and then pads the entire field again when `fill` is enabled.

This breaks RDPECAM device announcements: for `UVC Camera (046d:0825)`, the following channel name starts at byte 93 instead of byte 48 (including the two-byte header). A receiver reads an empty channel name at the expected offset. Windows 11 Camera fails before opening the per-device DVC.

## Change

Keep the aligned conversion buffer, return the actual UTF-16 character count, write only the converted bytes, and pad only the remaining field. Reject oversized strings instead of silently truncating them. Check capacity even for an empty padded field.

## Validation

- Added a regression test covering aligned and unaligned destinations, padded and unpadded writes, return count, stream position, following channel-name bytes, and rejection of truncation. It fails before the fix and passes afterward.
- Full Linux build passed.
- TestStream, TestBitStream, TestStreamPool, TestFreeRDPCodecH264, TestClientRdpFile, TestClientChannels, and TestClientCmdLine passed (7/7).
- A standalone announcement probe against the installed library now reports byte 48 and the correct channel name, matching 3.31.1.
- Windows live camera retest is pending. This PR contains only the string serialization fix and its test.


winpr/libwinpr/utils/stream.c
```diff
@@ -468,22 +468,23 @@ SSIZE_T Stream_Write_UTF16_String_From_UTF8(wStream* s, size_t wcharLength, cons
                                             size_t length, BOOL fill)
 {
 	SSIZE_T rc = 0;
+	if (!Stream_CheckAndLogRequiredCapacityOfSize(STREAM_TAG, s, wcharLength, sizeof(WCHAR)))
```
akallabeth, line 471: this is a nice catch.

winpr/libwinpr/utils/stream.c
```diff
@@ -468,22 +468,23 @@ SSIZE_T Stream_Write_UTF16_String_From_UTF8(wStream* s, size_t wcharLength, cons
...
 		WCHAR* str = ConvertUtf8NToWCharAlloc(src, length, &wlen);
 		if (!str)
 			return -1;
 
-		size_t len = wcharLength;
-		if (len > wlen)
-			len = wlen;
-		Stream_Write(s, str, len * sizeof(WCHAR));
-		Stream_Seek(s, wcharLength - len);
+		if (wlen > wcharLength)
```
akallabeth, line 481: why this change?
if the `WCHAR` string is longer it should truncated like it was done before?

winpr/libwinpr/utils/stream.c
```diff
@@ -468,22 +468,23 @@ SSIZE_T Stream_Write_UTF16_String_From_UTF8(wStream* s, size_t wcharLength, cons
...
+		{
+			free(str);
+			return -1;
+		}
+		rc = (SSIZE_T)wlen;
```
akallabeth, line 486: `WINPR_ASSERTING_INT_CAST` please

akallabeth (changes requested):

akallabeth:
replaced by https://github.com/FreeRDP/FreeRDP/pull/13421
thank you for the fix, adjusted it a little and added to the referenced pr.

PR #13418, opened 2026-09-16
[core,gateway] do not require a final auth token

thenextman:
Connecting through an MS RD Gateway fails during HTTP authentication whenever the negotiated mechanism needs the acceptor's final token to complete. Practically, Kerberos fails and NTLM works.

Tested on Linux with MIT Kerberos against an MS RD Gateway with Websocket transport enabled, using credentials that yield a Kerberos ticket for HTTP/<gateway fqdn> and the command line `/v:<target> /gateway:g:<gateway fqdn>,u:<user>,d:<domain> /d:<domain> /u:<user> '/auth-pkg-list:kerberos,!ntlm' /log-level:DEBUG`

The gateway does not send a final SPNEGO token when it completes the authentication. It answers the request carrying the client's last token with 101 Switching Protocols and no WWW-Authenticate header at all. With current master and MIT Kerberos, I get:

> [credssp_auth_authenticate]: Authentication in progress... (output token size: 1640)
> [rdg_establish_data_connection]: Unexpected HTTP status: HTTP_STATUS_SWITCH_PROTOCOLS
> [rdg_establish_data_connection]: [0] HTTP/1.1 101 Switching Protocols
> [rdg_establish_data_connection]: [1] Upgrade: websocket
> [rdg_establish_data_connection]: [2] Server: Microsoft-HTTPAPI/2.0
> [rdg_establish_data_connection]: [3] Sec-WebSocket-Accept: C+LNKQrHCrcAvLDHQKKgE1NKF7o=
> [rdg_establish_data_connection]: [4] Connection: Upgrade
> [rdg_establish_data_connection]: [5] Date: Fri, 11 Sep 2026 16:18:20 GMT
> [nego_connect]: Failed to connect
> [rdp_client_connect]: ERRCONNECT_SECURITY_NEGO_CONNECT_FAILED [0x0002000C]

I also tested on FreeRDP on Windows with Windows SSP and the same result.

Every header is listed and none is WWW-Authenticate. The Kerberos context is waiting for an `AP_REP` that can never arrive, so `rdg_recv_auth_token()` returns `FALSE` for lack of a token and `rdg_establish_data_connection()` gives up on a connection the gateway has already authenticated and upgraded.

Two further observations about the 101 response rather than Kerberos or any specific SSP:

- With `no-websockets`, the same gateway answers 200 OK carrying accept-completed and the exchange completes normally. Only the upgrade response drops the header.
- NTLM escapes the bug for an unrelated reason: when NTLM is the only available mechanism, both WinPR (negotiate.c, "Using direct NTLM") and native SSPI send bare `NTLMSSP` rather than wrapping it in `SPNEGO`, so they are already finished at the `AUTHENTICATE` message and never wait for a fourth round.

### Changes

`rdg_recv_auth_token()` and `wst_recv_auth_token()` now report "no token present" to the caller instead of treating it as an error, and the loops end the exchange when that happens, letting the existing HTTP status handling decide the outcome.

### Why it is safe

- The result check is unchanged. A response with no token still goes through the same status handling: `HTTP_STATUS_DENIED` sets `FREERDP_ERROR_CONNECT_ACCESS_DENIED` and fails. Only responses the gateway sends on success continue, and reaching them already required the gateway to accept the credentials
- Nothing uses the context afterwards. Both loops call `credssp_auth_free()` immediately after, and the channel is protected by TLS; the context never signs or encrypts anything. Requiring `SEC_E_OK` bought nothing, which is why the change is confined to these two gateway loops
- An undeliverable token cannot be verified by anyone. Where the confirmation never arrives, no client can check it. Where it does arrive, behaviour is unchanged: the token is consumed and verified exactly as before.
- It also fixes a latent bug in both loops: when the package produced no further output token while not yet complete, the same response was parsed again on the next iteration. They now leave the loop.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

hardening (approved):

PR #13424, opened 2026-09-16
[uwac,client] fix invisible Wayland cursor and stuck shortcut inhibitor

berkeleyinc:
Two independent Wayland defects found while getting `wlfreerdp` to work against a Windows 11 host from a Hyprland session. Each fix is a separate commit.

## 1. Mouse cursor never visible

`UwacRegisterCursor()` was only called from the `wl_seat` and
`wl_data_device_manager` global handlers. Compositors may announce `wl_seat`
before `wl_compositor`, and Hyprland does exactly that, so at both call sites
`display->compositor` was still `NULL`, the function bailed out early and
`seat->pointer_surface` stayed `NULL` for the whole session.

`set_cursor_image()` then called `wl_pointer.set_cursor()` with a `NULL`
surface, which per protocol hides the cursor. The result was a remote desktop
with no mouse pointer.

The fix creates the cursor surface when `wl_compositor` is announced, covering
seats that were already registered, and makes `UwacRegisterCursor()`
idempotent so the existing call sites remain safe.

This is announcement-order dependent: on compositors that announce
`wl_compositor` before `wl_seat`, the old code already worked.

## 2. Local shortcuts remain inhibited after keyboard leave

`keyboard_handle_leave()` reports `UWAC_EVENT_POINTER_LEAVE`, but the event
switch in `wlfreerdp.c` had no case for it, so the event was silently dropped.

The `zwp_keyboard_shortcuts_inhibitor_v1` object created for
`UWAC_EVENT_KEYBOARD_ENTER` was therefore only released by pressing RightCtrl.
After clicking into the remote desktop, local compositor shortcuts remained
inhibited even after the window lost keyboard focus.

The fix handles `UWAC_EVENT_POINTER_LEAVE` and releases the inhibitor,
mirroring the X11 client's `xf_event_LeaveNotify()` behavior.

## How to test

Use a compositor that announces `wl_seat` before `wl_compositor` (Hyprland in
this reproduction) and enable Wayland protocol logging:

```
WAYLAND_DEBUG=1 wlfreerdp3 /v:<host> /u:<user> /p:<password> /cert:ignore
```

Cursor:

* Before: only the window surface is created and
  `wl_pointer.set_cursor(..., nil, ...)` hides the cursor.
* After: a cursor surface is also created and `set_cursor` receives that valid
  surface and the pointer-enter serial. The pointer is visible over the remote
  desktop.

Shortcut inhibitor:

* Before: `wl_keyboard.leave` has no following inhibitor destruction, and local
  compositor shortcuts remain unavailable.
* After: the leave event is followed by
  `zwp_keyboard_shortcuts_inhibitor_v1.destroy`, and local shortcuts work as
  soon as the remote window loses keyboard focus.

## Verification

* Branch is linear on the current PR base; no merge commits.
* `git clang-format --diff origin/master` reports no changes.
* Full `minimal-client` build with Wayland/UWAC enabled completes successfully
  (788/788 targets).
* Both behaviors were reproduced before the changes and verified in a live
  Hyprland session after the changes using `WAYLAND_DEBUG=1`.


client/common/cmdline.h
```diff
@@ -244,10 +244,10 @@ static const COMMAND_LINE_ARGUMENT_A global_cmd_args[] = {
 	{ "gp", COMMAND_LINE_VALUE_REQUIRED, "<password>", nullptr, nullptr, -1, nullptr,
 	  "[DEPRECATED, use /gateway:p:<password>] Gateway password" },
 #endif
-	{ "grab-keyboard", COMMAND_LINE_VALUE_BOOL, nullptr, BoolValueTrue, nullptr, -1, nullptr,
-	  "Grab keyboard focus, forward all keys to remote" },
-	{ "grab-mouse", COMMAND_LINE_VALUE_BOOL, nullptr, BoolValueTrue, nullptr, -1, nullptr,
-	  "Grab mouse focus, forward all events to remote" },
+	{ "grab-keyboard", COMMAND_LINE_VALUE_BOOL | COMMAND_LINE_VALUE_OPTIONAL, "on|off",
```
akallabeth: invalid change, revert

client/common/cmdline.h
```diff
@@ -244,10 +244,10 @@ static const COMMAND_LINE_ARGUMENT_A global_cmd_args[] = {
...
+	  BoolValueTrue, nullptr, -1, nullptr, "Grab keyboard focus, forward all keys to remote" },
+	{ "grab-mouse", COMMAND_LINE_VALUE_BOOL | COMMAND_LINE_VALUE_OPTIONAL, "on|off", BoolValueTrue,
```
akallabeth: invalid change revert!

akallabeth (changes requested):
please check the command line syntax before changing options

client/common/cmdline.h
berkeleyinc: Reverted by dropping the full command-line commit. `/grab-mouse:off` was not valid FreeRDP boolean syntax; the supported disable form is `-grab-mouse`.

client/common/cmdline.h
berkeleyinc: Agreed. `COMMAND_LINE_VALUE_BOOL` already uses the boolean sigil syntax; `-grab-keyboard` disables this default-on option. I dropped the entire command-line commit and its tests from the branch, rather than retaining custom `:on|off` parsing.

akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

akallabeth:
nice patch, thank you for your pull request!

PR #13429, opened 2026-09-16
[client,sdl] link the RAIL client entry point

yeager:
SDL RemoteApp support calls `client_rail_server_start_cmd`, but the implementation lives in the RAIL client object library and was not linked into `sdl3-freerdp`. This makes an SDL build fail at final linking with an unresolved symbol.

The SDL target now includes the RAIL client object library when it is available.

Validation: configured from current master and built `sdl3-freerdp` successfully.


akallabeth (changes requested):
Care to explain what this should be good for in plain english?
(aka, which bug do you have, how can that be triggeret, ...)

yeager:
SDL RAIL now calls `client_rail_server_start_cmd`, but the SDL target did not link the `rail-client` object library that defines it. A standard SDL build therefore failed at the final link step with an undefined reference to that symbol whenever RAIL was enabled. This adds the existing RAIL client object code to the SDL target, matching the call site; no RAIL behavior is changed.

yeager:
Verified against current master (5d70e0c): the SDL3 client builds successfully, and `client_rail_server_start_cmd` is now provided by `libfreerdp-client`, which incorporates the RAIL client objects. Adding the same objects directly to the SDL executable would duplicate the implementation after rebasing. Closing this as upstream has resolved the link path.

Issue #11231, opened 2025-02-22
Active rdp sessions shouldn't prevent locking (screensaver) of the local machine


TheShanMan:
**Is your feature request related to a problem? Please describe.**
I often work in a remote session and occasionally I forget to minimize the freerdp window when I walk away. My assumption is that whenever I walk away from my computer, it will lock and turn off the monitors shortly thereafter. But in these forgetful situations, I might come back an hour or 2 later and see my monitors on and my local machine unlocked. This is especially problematic because I'm connecting to a work computer and my employer expects me to protect work computers from unauthorized access.

**Describe the solution you'd like**
I would like freerdp to not prevent locking (screensaver) of the local machine just because the window is active (not minimized). Lack of mouse movement and keyboard activity should lead to activation of the local machine's screensaver. If this goes too far, I would at least like to be able to opt in to this behavior (command line option).

**Describe alternatives you've considered**
The only alternative currently is to minimize the window, but the problem is I occasionally forget to do that.

**Additional context**
Discussion of my issue started here: https://github.com/FreeRDP/FreeRDP/discussions/11221

akallabeth:
@TheShanMan did have a look at it, and it boils down to have mouse/keybaord grab or not: 
adding `-grab-mouse -grab-keyboard` allows sleeping, but you will not be able to grab all keyboard shortcuts.

akallabeth:
@TheShanMan well, it is the `XGrabKeyboard` that causes this, and there is no real way around that to get all keyboard input.

so, I´d say for our use case this is out of scope, but if someone is willing to do some research and provide a pr we´ll happily merge it.

TheShanMan:
That's a shame. Maybe I'll have to resort to a bash script monitoring xprintidle and running gnome-screensaver-command -l once a threshold is reached.

zorjen122:
Under what conditions did this problem arise and what is the current progress? My tests under Ubuntu26.04 LTS system, GNOME Flashback and lightdm window management show that it will lock the screen, regardless of whether the rdp is currently playing video or whether the mouse focus is in the window.

TheShanMan:
Might relate to me using X11 rather than Wayland? I haven't tried in 26.04 either.

PR #13430, opened 2026-09-16
[client,x11] add local idle screensaver option

yeager:
Adds `/local-idle-screensaver:<seconds>` to xfreerdp. When enabled, the X11 client forces the local screensaver after the configured period without client input while retaining keyboard grab during active use.\n\nValidation: built `xfreerdp`; the option is listed by `/help`.\n\nFixes #11231.


yeager:
Added an Xvfb regression test for the idle timeout. `TestX11LocalIdleScreensaver` passes.

yeager:
Validated on an isolated GNOME/X11 session on Linux: TestX11LocalIdleScreensaver passes both through CTest and directly against the GNOME X11 display. The shell remained running after the screensaver request.

yeager:
Also built and ran TestX11LocalIdleScreensaver on Linux with MIT Kerberos development support enabled. The test passes.

akallabeth:
what is this for?
there is already `/prevent-session-lock` to keep the session active and by default the local screensaver activates just fine.
The time until the screensaver activates is not something the application should control. (e.g. out of scope)

PR #13437, opened 2026-09-17
no Password: prompt and don't mess with tty if stdin is a pipe

ileGITimo:
If , on linux/Fedora, running  freerdp as:
  some_command_that_supplies_passwd  | xfreerdp /from-stdin
one still gets a "Password: " prompt and errors from tc(get/set)attr that TCSANOW is an illegall ioctl, as input is a pipe.

This pull removes the prompt and makes set_terminal_nonblock return TRUE if stdin is not a tty.
Tested, on linux Fedora, both as described above and without the pipe input.
Similarl changes may be needed for Windows (and others?) but I'm not setup to compile and test there.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test

ileGITimo:
@akallabeth: This commit is the reason I caught the typo on set_terminal_nonblock :-)

PR #13441, opened 2026-09-17
[client,common] implement common client AAD helper

akallabeth:
* Move AAD helper binary handling from SDL client to common
* Use winpr_zfree to free token data
* Use AAD helper also for xfreerdp


akallabeth:
@hardening did refactor the AAD helper binary handling a little, now every client can reuse it.
did not yet have time to port the code to autodetect the binary locations, the rest is working well.

akallabeth:
@hardening ok, ready now.
* added the auto detection
* added configuration file for user / system config

did remove the compile time helper filter though. The helpers can be build out of tree and we don't know which one are shipped, so better try them all.

* did not test on windows/mac os yet.
* as of now removed the `CLI` fallback for the time being, if required could add another public API function with this option available.

client/common/aad_helper.c
```diff
@@ -532,3 +644,384 @@ void aad_auth_helper_stop(AadAuthHelper* helper)
...
+		return FALSE;
+	}
+
+	char* redirectUrl = nullptr;
+	size_t redirectUrlLen = 0;
+	const AadAuthHelperNavigateStatus status = aad_helper_navigate(
+	    helper, "FreeRDP WebView - AAD access token", request, &redirectUrl, &redirectUrlLen);
+	winpr_zfree(request);
+
+	if (status == AAD_AUTH_HELPER_NAVIGATE_CANCELLED)
```
hardening, line 748: here as all the treatment are similar perhaps there would have a way to do something like
```C
switch(status)  {
case AAD_AUTH_HELPER_NAVIGATE_OK:
    break;
case AAD_AUTH_HELPER_NAVIGATE_CANCELLED:
case AAD_AUTH_HELPER_NAVIGATE_TIMEOUT:
default:
   WLog_ERR(TAG, "[aad-auth] %s", someErrToLongString(status));
    winpr_znfree(redirectUrl, redirectUrlLen);
    return FALSE;
}

hardening (approved):
minor non blocking comment, but LGTM

client/common/aad_helper.c
akallabeth: will do in a follow up, I think we can merge both to one function.

PR #13453, opened 2026-09-19
reject unterminated applicationId in rail get appid resp

metsw24-max:
rail_read_server_get_appid_resp_order reads applicationId with Stream_Read_UTF16_String, which fills all 260 code units and never NUL-terminates, and unlike the extended sibling rail_read_get_application_id_extended_response_order it never checks that a terminator is present. a malicious server can send a TS_RAIL_ORDER_GET_APPID_RESP whose applicationId is 260 non-zero code units, so the unterminated buffer, the last member of the stack RAIL_GET_APPID_RESP_ORDER, is handed to the ServerGetAppIdResponse callback and any consumer treating it as a c string reads past the struct. this adds the same _wcsnlen guard the extended handler already uses, which also matches the null-terminated applicationId [MS-RDPERP] requires; only a response missing its terminator is rejected, well-formed ones are unchanged. to test, drive the rail client with a GET_APPID_RESP order whose 260 applicationId code units are all non-zero: before the change _wcsnlen(applicationId, 260) returns 260 and the buffer is forwarded unterminated, after it the order is rejected.


akallabeth (approved):
LGTM

akallabeth:
@freerdp-bot test
@metsw24-max by convention the string should be `\0` terminated, but clients do have the `WCHAR` string and the maximum size from `RAIL_GET_APPID_RESP_ORDER` available and should use the `wcsnlen`, ... length checking function nonetheless.

PR #13459, opened 2026-09-20
[client,x11] fix RAIL window repaint volume (invalid region never reset, updates for unmapped windows)

arthgirard:
## Problem

RemoteApp sessions with xfreerdp felt noticeably choppier than a full desktop
session against the same Windows 11 host: laggy scrolling in Word, slow menu
hover highlighting. CPU sampling showed Xwayland at ~90 % of a core while
xfreerdp itself was far from saturated.

Tracing the X11 calls (`WLOG_FILTER=com.freerdp.client.x11:TRACE`) and the
server surface commands (`WLOG_FILTER=com.freerdp.gdi:TRACE`) for a
maximized Word window on a 5120x2832 screen:

| | server surface commands | client `XPutImage` |
|---|---|---|
| peak pixel volume | ~8 Mpx/s | ~500 Mpx/s (~2 GB/s) |
| wire | ~0.7 MB/s | – |

Every frame the client blitted the same 6 rectangles covering the entire
window. The full desktop session with the same content: ~20 Mpx/s.

## Cause

1. `xf_AppUpdateWindowFromSurface()` (surface-to-window mapped RAIL path)
   paints `surface->invalidRegion` but never clears it. `xf_OutputUpdate()`
   does for output mapped surfaces. So damage accumulates across frames
   until it covers the whole window; from then on every update re-uploads
   the complete window regardless of what the server actually sent.

2. Windows also exposes its hidden desktop window ("Program Manager",
   spanning the whole virtual screen) as a RAIL window with a mapped
   surface, and keeps repainting it. Together with (1) that was a ~100 MB
   `XPutImage` per frame into a *withdrawn* X window on a multi-monitor
   setup.

## Fix

* Commit 1: clear `surface->invalidRegion` after painting (all exit paths),
  mirroring `xf_OutputUpdate()`. 3-line change, fixes the bulk of the
  problem on its own.
* Commit 2: skip the blit while `appWindow->is_mapped` is FALSE, mark the
  window stale, and repaint the whole window from its surface on
  `MapNotify` (new `xf_AppWindowRepaintFromSurface()`), so nothing is
  left blank when a window is shown/restored. Also avoids uploads for
  minimized windows.

## Result

Same maximized-Word scroll test (Fedora 44, GNOME 50 / XWayland, mesa iris):

| | before | after |
|---|---|---|
| Xwayland CPU peak | 91 % | 21 % |
| xfreerdp CPU avg | 24–35 % | 7 % |
| client `XPutImage` volume | ~500 Mpx/s | matches server damage |

Tested against Windows 11 Pro (dockur/windows VM), M365 Word/Excel,
`/multimon`, `/gfx:RFX` (server negotiated ClearCodec/Alpha/Progressive),
FreeRDP 3.31.1 with these two changes backported, and this branch compiled
on master.

Note: a MIT-SHM path was considered and rejected — `perf` showed Xwayland's
cost is mesa's `glTexSubImage2D` staging copy in glamor, not the socket.


akallabeth (approved):
LGTM, nice one

akallabeth:
@freerdp-bot test

PR #13463, opened 2026-09-21
[server,proxy] add TargetCertPolicy config option

akallabeth:


akallabeth:
@freerdp-bot test

server/proxy/pf_client.c
```diff
@@ -884,19 +884,155 @@ static void pf_client_context_free(freerdp* instance, rdpContext* context)
...
+	char* got = freerdp_certificate_get_fingerprint_by_hash_ex(cert, hash, useSeparator);
+	if (got)
+		rc = _stricmp(TargetCertHash, got) == 0;
+
+	winpr_zfree(got);
+	return rc;
+}
+
+WINPR_ATTR_NODISCARD
+static int pf_client_verify_pinned(pClientContext* pc, const BYTE* data, size_t length)
```
mfleisz, line 904: Looks like this function is either returning 0 or 1 so a BOOL type might be more appropriate

server/proxy/pf_client.c
```diff
@@ -884,19 +884,155 @@ static void pf_client_context_free(freerdp* instance, rdpContext* context)
...
+		               "certificate policy PINNED: neigter PEM nor HASH set for "
+		               "comparison, deny access");
+	}
+
+	freerdp_certificate_free(cert);
+	return rc;
+}
+
+WINPR_ATTR_NODISCARD
+static int pf_client_verify_by_policy(pClientContext* pc, const BYTE* data, size_t length)
```
mfleisz, line 1001: Also seems to return either 0 or 1 so BOOL might be the better return choice

mfleisz (commented):
Two small remarks - overall LGTM

server/proxy/pf_client.c
akallabeth: kept as the parent function expects an `int` as well.

server/proxy/pf_client.c
akallabeth: kept as the parent function expects an `int` as well.

akallabeth: we might some day implement a cert store which would return a different value (or return negative as error)

PR #13464, opened 2026-09-21
Harden again

akallabeth:


channels/drive/client/drive_main.c
```diff
@@ -224,10 +225,12 @@ static UINT drive_process_irp_create(DRIVE_DEVICE* drive, IRP* irp)
 		if (allocationSize > 0)
 		{
 			const BYTE buffer[] = { '\0' };
-			if (!drive_file_seek(file, allocationSize - sizeof(buffer)))
-				return ERROR_INTERNAL_ERROR;
-			if (!drive_file_write(file, buffer, sizeof(buffer)))
+			if (!drive_file_seek(file, allocationSize - sizeof(buffer)) ||
+			    !drive_file_write(file, buffer, sizeof(buffer)))
+			{
+				ListDictionary_Remove(drive->files, key);
```
mfleisz: Think it is better to move the `ListDictonary_Add` after this block so we can avoid adding/removing an entry here.

mfleisz (commented):
- Handle check is missing for `serial_process_irp_close`
- `utils_set_umask` is defined twice - could be defined in winpr and reused in libfreerdp to avoid code duplication

channels/drive/client/drive_main.c
akallabeth: would require a manual free, so nothing gained.

akallabeth:
> * Handle check is missing for `serial_process_irp_close`
>
not required, the `CloseHandle` already returns `FALSE` in that case.
on the other hand, https://learn.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-closehandle is a bit vague there, so better check it outside.
 
>     * `utils_set_umask` is defined twice - could be defined in winpr and reused in libfreerdp to avoid code duplication
did not want to export a new function for this, when keeping internal we can change whenever we like.

akallabeth:
@freerdp-bot test

channels/drive/client/drive_main.c
mfleisz: Yes but besides better readable code it will also avoid unnecessary alloc/lock/unlock/free operations

PR #13465, opened 2026-09-21
OSS-Fuzz: Disable libpulse manually for oss-fuzz build

arthurscchan:
Since 10th September, FreeRDP's CMake has enabled its PulseAudio audio backends by default on Linux whenever libpulse is detected. This change has caused the OSS-Fuzz build to fail ever since. The reason is that OSS-Fuzz uses separate images for building and for fuzzing, and libpulse is never shipped into the runtime image. Because the library is present at build time, the backends are enabled automatically, and the fuzzing check in different docker image then fails. This PR fixes that by explicitly disabling PulseAudio for the OSS-Fuzz build, so that the fuzz targets run normally without depending on a library that is unused and not shipped.


akallabeth:
@freerdp-bot test

akallabeth (approved):
LGTM

